Issue #1375, opened 2020-04-10
Port shared data store changes and conventions


aaemnnosttv:
This is a catch-all issue for porting changes to shared datastore code from the Analytics component refactoring branch to develop

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All actions that have controls must have dedicated actions for state mutations
* Async actions should follow the following conventions:
    - Use a single action creator generator function to yield all actions for normal/success flow
    - E.g. an action to fetch a "foo" resource would have the following actions
    - `START_FETCH_FOO` - Sets `fetchingFoo: true` in the store
    - `FETCH_FOO` - has associated **`control`** to perform async fetch
    - `FINISH_FETCH_FOO` - Sets `fetchingFoo: false` and potentially other state for a successful completion
        - `receive*` actions are still separate and don't always share the same action name
    - `CATCH_FETCH_FOO` - called by consuming code in the `catch` block surrounding the `yield actions.fetchFoo()` and should always require an `error`
        - Should accept an `error` parameter, and the action should always have `payload.error`
    - The action creator for `fetchFoo` should always return the result from the primary main action, _after_ yielding the `FINISH_FETCH_FOO` action

#### All Stores (core/site, etc)
* Add `constants` sub-module
    - Move and export `STORE_NAME` from this module to avoid circular dependencies with store sub-modules
    - Store sub-modules should only import this from `constants`
* Ensure all action creators include a `payload`, even if empty.

#### Settings Store
* Add selectors:
    - ~~`getSavedSettings`~~
    - ~~`getSaved{Setting}`~~
* Add actions:
    - `rollbackSettings` - "Restores" current settings with saved settings
* Update `actions.set{Setting}` to allow falsy values (invariant should only prevent setting `undefined`)

#### E2E
* Add Redux action logging to e2e-utilities and opt-in via `DEBUG_REDUX=1` env var

#### Webpack + Storybook
* Update main webpack config to allow bits of it to be imported for use by custom config for storybook  
This is necessary for Storybook to work - all Site Kit external assets (e.g. `googlesitekit-data`) need to be converted to aliases

#### Shared Components
* Update the Material `Switch` to use a unique ID by default
* Upgrade `@material/react-select` to latest (resolves bug where disabled state does not change when prop changes)

#### Added Shims
A few module import shims are necessary to make sure consuming code always references the same version.
[...]

Issue #1174, opened 2020-02-26
Introduce notification functionality to core/site store


felixarntz:
This is a follow-up to #999 and #1110 (also related #1000 and #1173).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `core/site` store should have an additional selector `getNotifications`:
    * The selector should issue an API request to `core/site/data/notifications` to get a list of notification objects.
        * See #1110 / #1170 for typical properties to expect on notifications (probably doesn't matter much for implementation of this, but fyi).
    * The selector should however also support notifications added by JS (i.e. when receiving the server response, it shouldn't override the JS-added notifications, but only the ones previously received from server).
* The `core/site` store should have actions `addNotification( notification )` and `removeNotification( id )`:
    * Both of these should only work client-side for now, i.e. simply modify the JS-based notifications array.
    * All `notification` objects must include `id`, so that they can be removed using that.

## Implementation Brief

In a new `googlesitekit/datastore/site/notifications.js`:
* Use `createNotificationsStore` from #1205 with parameters `core`, `site`, `notifications` to generate the notifications store object needed and export it.
* Export `INITIAL_STATE` from that file as well (the initial state of just the notifications part).

In `googlesitekit/datastore/site/index.js`:
* Include the results in the existing `core/site` store definition, and merge the notifications store's `INITIAL_STATE` into the actual store's `INITIAL_STATE`.

## QA Brief

* 

## Changelog entry

* Add notifications functionality to `core/site` datastore.

tofumatt:
> Include the results in the existing `core/site` store definition, and merge the notifications store's `INITIAL_STATE` into the actual store's `INITIAL_STATE`.

I'm a bit confused by this part of the brief—what does that entail in code? `INITIAL_STATE` is the uninitialised state code used to bootstrap a store, but I don't understand why we'd merge one store's `INITIAL_STATE` into another.

From the sounds of this issue we want to add notifications to `core/site`; based on the approach in #999 I'd think that would entail a new file (similar to `assets/js/googlesitekit/datastore/site/connection.js`) that contained notification data for the store.

So my approach would be to use the code in https://github.com/google/site-kit-wp/pull/1205/files#diff-084d6f9b3b4d5a065e515e4b91e750d3 but use it as `assets/js/googlesitekit/datastore/site/notifications.js`, then include that in the main `core/site` similar to the other two files in that folder.

felixarntz:
[...]

PR #1388, opened 2020-04-14
Introduce createFetchStore utility function to reduce boilerplate (metaprogramming)

felixarntz:
## Summary

Addresses issue #1288

## Relevant technical choices

* `createFetchStore` returns a (partial) store object with all the internal infrastructure necessary to power an API request-based action _or_ selector.
* By only using this for the internal parts, we maintain the chance to properly document our actual public selectors and actions.
* The store object returned sticks to the patterns established via #1375 / #1387.
* PR includes some follow-up fixes to #1174, partially related to the changes here, but also a bug just uncovered here while adjusting that code (see https://github.com/google/site-kit-wp/pull/1388/commits/fc0faaf1a019b23c687462ce906867b2bd77af25).

See function docs for more details and https://github.com/google/site-kit-wp/issues/1288#issuecomment-613212108 for an example of how existing code could use it.

## Checklist


assets/js/googlesitekit/data/create-fetch-store.test.js
```diff
@@ -0,0 +1,328 @@
...
+			...state,
+			data: {
+				...( state.data || {} ),
+				[ aParam ]: response,
+			},
+		};
+	},
+	argsToParams: ( objParam, aParam ) => {
+		invariant( 'object' === typeof objParam, 'objParam is required.' );
+		invariant( 'undefined' !== typeof aParam, 'aParam is required.' );
```
aaemnnosttv, line 63: Let's use `isPlainObject` and `aParam !== undefined` instead.

assets/js/googlesitekit/data/create-fetch-store.test.js
```diff
@@ -0,0 +1,328 @@
...
+	beforeEach( () => {
+		registry = createRegistry();
+
+		storeDefinition = createFetchStore( STORE_PARAMS );
+		registry.registerStore( STORE_NAME, storeDefinition );
+		dispatch = registry.dispatch( STORE_NAME );
+		store = registry.stores[ STORE_NAME ].store;
+		select = registry.select( STORE_NAME );
+
+		apiFetchSpy = jest.spyOn( { apiFetch }, 'apiFetch' );
```
aaemnnosttv, line 92: This was added to many other tests, but I don't think we need it. It's mostly been removed in the `jest-fetch-mock` refactoring issue.

assets/js/googlesitekit/data/create-fetch-store.js
```diff
@@ -0,0 +1,252 @@
...
+ * @param {?Function} options.reducerCallback Optional. Callback function to modify state based on the API response.
+ *                                            Will be used inside the reducer. The  function receives the store's state
+ *                                            object as first parameter, the API response as second parameter, and the
+ *                                            params object for the request (see above) as third parameter. If not
+ *                                            provided, the default will return the unmodified state.
+ * @param {?Function} options.argsToParams    Optional. Function that should validate expected arguments for the
+ *                                            internal fetch action and parse them into an named parameters object,
+ *                                            with the argument names used as keys. If not provided, the default
+ *                                            function will return an empty object, essentially indicating that no
+ *                                            arguments are supported/required.
```
aaemnnosttv, line 76: Let's use the optional arg syntax (i.e. `[options.reducerCallback]`) rather than a nullable type for optional keys.

I would suggest we rename `options` to `args` as well as the whole object is not optional.

assets/js/googlesitekit/data/create-fetch-store.test.js
```diff
@@ -0,0 +1,328 @@
...
+		unsubscribeFromAll( registry );
+		apiFetchSpy.mockRestore();
+	} );
+
+	describe( 'actions', () => {
+		it( 'includes the expected actions', () => {
+			const fetchStoreDefinition = createFetchStore( {
+				baseName: 'SaveSomeData',
+				controlCallback: async () => true,
+				reducerCallback: ( state ) => state,
```
aaemnnosttv, line 109: `reducerCallback` is optional so I would omit it from this test, to use only the minimal required args - otherwise it might look like the presence of this is required to get a `receive` action.

assets/js/googlesitekit/data/create-fetch-store.test.js
```diff
@@ -0,0 +1,328 @@
...
+				dispatch.fetchGetSomeData();
+				expect( consoleErrorSpy ).toHaveBeenCalledWith( 'objParam is required.' );
+
+				dispatch.fetchGetSomeData( 123 );
+				expect( consoleErrorSpy ).toHaveBeenCalledWith( 'objParam is required.' );
+
+				dispatch.fetchGetSomeData( {} );
+				expect( consoleErrorSpy ).toHaveBeenCalledWith( 'aParam is required.' );
+
+				consoleErrorSpy.mockClear();
```
aaemnnosttv, line 131: Make sure to restore rather than clear to ensure it's back to it's normal state for other tests.
```suggestion
				consoleErrorSpy.mockRestore();
```
Also, I think this kind of cleanup should always happen in an `afterEach` or another lifecycle callback as an error in the test would otherwise cause the cleanup to be skipped and cause problems in subsequent tests.

assets/js/googlesitekit/data/create-fetch-store.test.js
```diff
@@ -0,0 +1,328 @@
...
+
+				const action = fetchStoreDefinition.actions.fetchSaveSomeData();
+
+				// Catch invariant to get exactly the error we expect.
+				let error;
+				try {
+					invariant( false, 'requiredParam is required.' );
+				} catch ( err ) {
+					error = err;
+				}
```
aaemnnosttv, line 155: It would be good to extract this to a utility as I would imagine this is probably used more than once in these tests.

I'm curious though, would it not pass with simply `new Error( 'requiredParam is required.' )` ?

assets/js/googlesitekit/data/create-fetch-store.test.js
```diff
@@ -0,0 +1,328 @@
...
+
+				expect( action.next().value ).toEqual( {
```
aaemnnosttv, line 157: For testing generators, I think it might be better to treat the response as a collection rather than checking a specific `yield`. In this case I think we can get all actions with `[ ...action ]` and then we could `expect.arrayContaining` the `{ response, error }`.

assets/js/googlesitekit/data/create-fetch-store.test.js
```diff
@@ -0,0 +1,328 @@
...
+					controlCallback: async () => true,
+					reducerCallback: ( state ) => state,
+				} );
+
+				const action = fetchStoreDefinition.actions.fetchSaveSomeData();
+
+				expect( action.next().value.type ).toEqual( 'START_FETCH_SAVE_SOME_DATA' );
+				expect( action.next().value.type ).toEqual( 'FETCH_SAVE_SOME_DATA' );
+				expect( action.next( 42 ).value.type ).toEqual( 'RECEIVE_SAVE_SOME_DATA' );
+				expect( action.next().value.type ).toEqual( 'FINISH_FETCH_SAVE_SOME_DATA' );
```
aaemnnosttv, line 175: For these I think we could list out the actions to ensure the order is correct while keeping the assertions separate.
E.g.
```js
const [ startAction, fetchAction, receiveAction, finishAction ] = action;
expect( startAction ) // ...;
expect( fetchAction ) // ...;
//...
```

assets/js/googlesitekit/data/create-fetch-store.test.js
```diff
@@ -0,0 +1,328 @@
...
+					response: undefined,
+					error,
+				} );
+			} );
+
+			it( 'yields the expected actions for a success request', () => {
+				const fetchStoreDefinition = createFetchStore( {
+					baseName: 'SaveSomeData',
+					controlCallback: async () => true,
+					reducerCallback: ( state ) => state,
```
aaemnnosttv, line 167: Let's not pass `reducerCallback` in tests where it isn't relevant/necessary.

assets/js/googlesitekit/data/create-fetch-store.test.js
```diff
@@ -0,0 +1,328 @@
...
+				const action = fetchStoreDefinition.actions.fetchSaveSomeData();
+
+				const error = {
+					code: 'this-went-wrong',
+					message: 'This went wrong.',
+				};
+
+				expect( action.next().value.type ).toEqual( 'START_FETCH_SAVE_SOME_DATA' );
+				expect( action.next().value.type ).toEqual( 'FETCH_SAVE_SOME_DATA' );
+				expect( action.throw( error ).value.type ).toEqual( 'CATCH_FETCH_SAVE_SOME_DATA' );
```
aaemnnosttv, line 198: Same as above.

assets/js/googlesitekit/data/create-fetch-store.test.js
```diff
@@ -0,0 +1,328 @@
...
+				fetch
+					.doMockOnceIf(
+						/^\/google-site-kit\/v1\/core\/test\/data\/some-data/
+					)
+					.mockResponseOnce(
+						JSON.stringify( expectedResponse ),
+						{ status: 200 }
+					);
+
+				const { response, error } = await dispatch.fetchGetSomeData( {}, 'value-to-key-response-by' );
```
aaemnnosttv, line 216: This threw me off for a bit. It's not immediately obvious that the second parameter becomes `aParam` which is used in the reducer. Maybe this is just because the store definition is so far away from this spot, but just thought I'd  mention it.

assets/js/googlesitekit/data/create-fetch-store.test.js
```diff
@@ -0,0 +1,328 @@
...
+				expect( store.getState().data ).toEqual( undefined );
+			} );
+
+			it( 'sets flag for request being in progress', async () => {
+				fetch
+					.doMockOnceIf(
+						/^\/google-site-kit\/v1\/core\/test\/data\/some-data/
+					)
+					.mockResponseOnce(
+						JSON.stringify( { someValue: 42 } ),
```
aaemnnosttv, line 254: If the response is arbitrary, let's annotate that with a comment so we're not looking for `someValue` or 42 in the test 😄

assets/js/googlesitekit/data/create-fetch-store.test.js
```diff
@@ -0,0 +1,328 @@
...
+				dispatch.fetchGetSomeData( ...requestArgs );
+
+				// Now it should be in progress.
+				expect( select.isFetchingGetSomeData( ...requestArgs ) ).toEqual( true );
+
+				// A request for other arguments however is not in progress.
+				expect( select.isFetchingGetSomeData( {}, 'anotherValue' ) ).toEqual( false );
+
+				await subscribeUntil( registry,
+					() => store.getState().data !== undefined,
```
aaemnnosttv, line 272: Let's wait until the selector `hasFinishedResolution` - I think we should generally await selectors this way unless the test requires otherwise for some reason (I can't think of why that would be necessary though).

assets/js/googlesitekit/data/create-notifications-store.js
```diff
@@ -72,9 +67,29 @@ export const createNotificationsStore = ( type, identifier, datapoint, {
 		// object so we can know if a client notification was added and then
 		// removed from state.
 		clientNotifications: client ? undefined : {},
-		isFetchingNotifications: false,
 	};
 
+	const fetchGetNotificationsStore = createFetchStore( {
```
tofumatt, line 72: Seeing this in action really illustrates the value of the new approach. Seeing these kind of diff is awesome.

The new `createFetchStore` code is still quite readable for how much is abstracted away, which is impressive. I'm happy we discussed the API so much previous to the review, because this is sooooo much more succinct without being opaque.

All that is to say: great job on this. 🎉

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -271,9 +267,7 @@ describe( 'core/user authentication', () => {
 				muteConsole( 'error' );
 				registry.select( STORE_NAME ).getGrantedScopes();
 				await subscribeUntil( registry,
-					// TODO: We may want a selector for this, but for now this is fine
```
tofumatt, line 274: Thanks for tidying all of these up 🙂

assets/js/googlesitekit/data/create-fetch-store.js
```diff
@@ -0,0 +1,252 @@
...
+	argsToParams,
+} ) => {
+	invariant( baseName, 'baseName is required.' );
+	invariant( 'function' === typeof controlCallback, 'controlCallback is required.' );
+
+	if ( 'function' !== typeof reducerCallback ) {
+		reducerCallback = ( state ) => {
+			return { ...state };
+		};
+	}
```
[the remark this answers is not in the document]
felixarntz: Do you mean handling this as an actual default value for the argument instead of conditionally doing that here? If so, I'm rather opposed because while the clause here doesn't seem as elegant, I think it's easier to read the code like this.

assets/js/googlesitekit/data/create-fetch-store.js
```diff
@@ -0,0 +1,252 @@
...
+
+	let requiresParams = true;
+	if ( 'function' !== typeof argsToParams ) {
+		requiresParams = false;
```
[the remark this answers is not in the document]
felixarntz: I've updated this to actually test the function, which is a bit more reliable than this original approach. Even when `argsToParams` is provided, it could have logic that is okay for `params` to end up empty.

assets/js/googlesitekit/data/create-fetch-store.test.js
felixarntz: Unfortunately not, the errors that `invariant` throws have a different format.

assets/js/googlesitekit/data/create-fetch-store.js
```diff
@@ -0,0 +1,252 @@
...
+		},
+	};
+
+	return {
+		actions,
+		controls,
+		reducer,
+		resolvers: {},
+		selectors,
+	};
```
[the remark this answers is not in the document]
felixarntz: Good point, I'll add that.

assets/js/googlesitekit/data/create-fetch-store.test.js
felixarntz: Using `mockRestore` here doesn't work, it will result in errors elsewhere:
```
    TypeError: global.console.error.mockClear is not a function

      1 | beforeEach( () => {
    > 2 | 	global.console.error.mockClear();
        | 	                     ^
      3 | 	global.console.warn.mockClear();
      4 | 	global.console.log.mockClear();
      5 | 	global.console.info.mockClear();

      at Object.mockClear (tests/js/setup-before-after.js:2:23)
```

I think we can leave this as is for now and look into it separately, there are multiple places in our existing code place where this approach is used.

assets/js/googlesitekit/data/create-fetch-store.test.js
felixarntz: I tried that first, but we'll want to test with specific values here (receiving `42` / throwing an error below), so I think individual `yield`s are more applicable here.

assets/js/googlesitekit/data/create-fetch-store.test.js
felixarntz: See above.

assets/js/googlesitekit/data/create-fetch-store.test.js
felixarntz: See above, how would we trigger an error like this by just destructuring the generator?

assets/js/googlesitekit/data/create-fetch-store.test.js
felixarntz: Yeah it's probably because the definition is so far away. What happens with these arguments is entirely up to the `argsToParams` logic.

assets/js/googlesitekit/data/create-fetch-store.test.js
felixarntz: There's no resolver involved here that we can rely on, and we shouldn't rely on the `isFetchingGetSomeData` selector since that's what's actually being tested below.

Since we're really just testing a plain fetch store here, all that we can do is inspect the store state directly to know that the reducer has completed setting the data as expected.

assets/js/googlesitekit/data/create-fetch-store.js
```diff
@@ -0,0 +1,252 @@
...
+	const FINISH_FETCH = `FINISH_FETCH_${ constantBaseName }`;
+	const CATCH_FETCH = `CATCH_FETCH_${ constantBaseName }`;
+	const RECEIVE = `RECEIVE_${ constantBaseName }`;
+
+	const fetchCreator = `fetch${ pascalCaseBaseName }`;
+	const receiveCreator = `receive${ pascalCaseBaseName }`;
+	const isFetching = `isFetching${ pascalCaseBaseName }`;
+
+	const actions = {
+		[ fetchCreator ]: function*( ...args ) { // eslint-disable-line object-shorthand
```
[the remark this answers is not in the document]
aaemnnosttv: I find it's easier to identify as a generator with the astrisk before the function like this. It also negates the need for an eslint override 👍

assets/js/googlesitekit/data/create-fetch-store.test.js
aaemnnosttv: Oh okay, thanks for clarifying. It should still be called in a lifecycle callback though rather than just at the end of the test.

assets/js/googlesitekit/data/create-fetch-store.test.js
aaemnnosttv: The error would be triggered when invoking the generator function - that part would still happen. It's the response that you can destructure into a list of all of its `yield`ed values as an array which can be listed out into variables that you can make the same assertions on, but in a more readable way. I'll take a closer look and give you a concrete example in a bit, unless there's something I'm missing here that would prevent this from being possible.

assets/js/googlesitekit/data/create-fetch-store.test.js
aaemnnosttv: Oh right - the resolver would be defined outside of this along with its selector 👍

assets/js/googlesitekit/data/create-fetch-store.test.js
aaemnnosttv: Looking back at this now, I see why you're stepping through the actions manually and that my suggestion indeed would not work for this test due to specific action yields being called or errors thrown directly. However, this test and `yields the expected actions for an error request` feel like they're testing implementation details and directly interact with the generator functions which none of our other datastore tests do.

It seems like the subsequent tests for `makes a network request based on controlCallback` and `dispatches an error if the request fails` cover this same behavior but at the same level of abstraction as all of our other tests, using mocked http requests. 

It could be that these generator tests tests were useful during the initial development but I don't think they're necessary anymore and that the http-based tests give us more "bang for our buck" in terms of coverage, confidence, and consistency.

Thoughts @tofumatt ?

assets/js/googlesitekit/data/create-fetch-store.test.js
aaemnnosttv: You're right, I misunderstood what was happening here 😄 

See my comment here https://github.com/google/site-kit-wp/pull/1388#discussion_r436688633

assets/js/googlesitekit/data/create-fetch-store.test.js
tofumatt: Yeah, I'm in agreement with @aaemnnosttv here. Generator tests and testing each yield is testing the implementation probably a bit too much, and it'd be better to test the output than these. Especially as this isn't testing an action yielded until the end.

There are tools like `redux-saga-tester` that allow us to use helpers that let us wait for specific actions to be dispatched if we want to check the flow of actions dispatched inside a generator action—we don't need that here, but if we _do_ want to be testing implementation/yielded action dispatch order of a generator action, we should use tools like that. They're a bit nicer in terms of their API, and also less prone to illegitimate failures; if the order of actions here changed then this test could fail unless the number of `next` calls changed too.

assets/js/googlesitekit/data/create-fetch-store.js
```diff
@@ -0,0 +1,270 @@
...
+		[ isFetching ]: ( state, ...args ) => {
+			if ( state[ isFetching ] === undefined ) {
+				return false;
+			}
+
+			let params;
+			try {
+				params = argsToParams( ...args );
+			} catch ( err ) {
+				// If parameters are invalid, fail silently here. It likely is
```
tofumatt, line 253: Minor documentation tweak here: arguably it's not really "failing" silently here. This is expected behaviour because the arguments provided aren't yet valid because other selectors aren't ready.

That said, I'd expect `argsToParams` to handle invalid params accordingly and return an appropriately empty object or whatever if dependent selectors pass it `undefined` or `null` values.

Because this code catches _any_ error in `argsToParams` and will silence it and ignore it, it's potentially dangerous and obscuring real errors. It's a bit more work for developers, but it might be better to have `argsToParams` have to handle empty arguments than take this approach. Maybe we should ditch this try/catch block.

assets/js/googlesitekit/data/create-fetch-store.js
```diff
@@ -0,0 +1,270 @@
...
+			return !! state[ isFetching ][ stringifyObject( params ) ];
+		},
+	};
+
+	return {
+		INITIAL_STATE,
+		actions,
+		controls,
+		reducer,
+		resolvers: {},
```
tofumatt, line 267: Technically we can omit this key rather than pass an empty object, right? Not a big deal either way.

assets/js/googlesitekit/data/create-fetch-store.test.js
```diff
@@ -0,0 +1,319 @@
...
+		return {
+			...state,
+			data: {
+				...( state.data || {} ),
+				[ aParam ]: response,
+			},
+		};
+	},
+	argsToParams: ( objParam, aParam ) => {
+		invariant( isPlainObject( objParam ), 'objParam is required.' );
```
tofumatt, line 62: I wonder if having checks like this in `argsToParams` is the right approach; see my comment above about catching errors in this function.

assets/js/googlesitekit/data/create-fetch-store.test.js
tofumatt: This is okay for now, but going forward if we want to test a lot of async actions like this we might want to hook up a until like `redux-saga-tester` or similar to allow us to write more flexible generator tests. These tests are all fine but the issue and worth keeping, but if a _new_ yielded action is added in-between two actions here the tests will fail even though they're largely still behaving as-intended.

Having to add extra `action.next()` here when we modify the actions seems onerous to me. But that's only if we start writing a lot of tests around these behaviours. If these lower-level libraries have these types of lower-level tests, maybe that's okay. 🙂

stories/module-analytics-setup.stories.js
```diff
@@ -64,8 +64,8 @@ storiesOf( 'Analytics Module/Setup', module )
 
 		const setupRegistry = ( registry ) => {
 			registry.dispatch( STORE_NAME ).setSettings( {} );
-			registry.dispatch( STORE_NAME ).receiveAccounts( [] );
-			registry.dispatch( STORE_NAME ).receiveExistingTag( null );
+			registry.dispatch( STORE_NAME ).receiveGetAccounts( [] );
+			registry.dispatch( STORE_NAME ).receiveGetExistingTag( null );
 			registry.stores[ STORE_NAME ].store.dispatch( { type: 'START_FETCH_ACCOUNTS_PROPERTIES_PROFILES' } );
```
aaemnnosttv, line 69: I believe this action has changed - Storybook no longer shows a loading state here so it must have 😄

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -28,147 +28,83 @@ import API from 'googlesitekit-api';
...
+	controlCallback: ( { accountID } ) => {
+		return API.set( 'modules', 'analytics', 'create-property', { accountID } );
+	},
+	reducerCallback: ( state, property, { accountID } ) => {
+		return {
+			...state,
+			properties: {
+				...state.properties || {},
+				[ accountID ]: [
+					...( state.properties || {} )[ accountID ] || [],
```
aaemnnosttv, line 66:
```suggestion
					...( state.properties[ accountID ] || [] ),
```

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -28,147 +28,83 @@ import API from 'googlesitekit-api';
...
+const fetchCreatePropertyStore = createFetchStore( {
+	baseName: 'createProperty',
+	controlCallback: ( { accountID } ) => {
+		return API.set( 'modules', 'analytics', 'create-property', { accountID } );
+	},
+	reducerCallback: ( state, property, { accountID } ) => {
+		return {
+			...state,
+			properties: {
+				...state.properties || {},
```
aaemnnosttv, line 64:
```suggestion
				...state.properties,
```

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -420,10 +263,27 @@ export const resolvers = {
...
-			const { response, error } = yield actions.fetchPropertiesProfiles( accountID );
-			if ( response && response.properties ) {
+			const { response, error } = yield fetchGetPropertiesProfilesStore.actions.fetchGetPropertiesProfiles( accountID );
+
+			if ( response ) {
+				const { dispatch } = registry;
+
+				dispatch( STORE_NAME ).receiveGetProperties( response.properties, { accountID } );
+
+				if ( response.profiles.length && response.profiles[ 0 ] && response.profiles[ 0 ].webPropertyId ) {
```
aaemnnosttv, line 273:
```suggestion
				if ( response.profiles?.[ 0 ]?.webPropertyId ) {
```

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -511,18 +371,34 @@ export const selectors = {
 	 * @param {string} accountID The Analytics Account ID to check for property creation.
 	 * @return {boolean} `true` if fetching a properties, `false` if not.
 	 */
-	isDoingGetProperties( state, accountID ) {
-		const { isFetchingPropertiesProfiles } = state;
+	isDoingGetProperties: createRegistrySelector( ( select ) => ( state, accountID ) => {
+		// Check if dispatch calls right after fetching are still awaiting.
+		if ( state.isAwaitingPropertiesProfilesCompletion ) {
+			return true;
+		}
```
aaemnnosttv, line 378: I think this flag isn't necessary and may lead to a race condition where two requests could overwrite the value here one way or the other since it isn't keyed by the args.

Instead, I think we could simply return `select( STORE_NAME ).isResolving( 'getProperties', [ accountID ] )`. That might mean that in this case we don't use the `isFetchingGetPropertiesProfiles` selector where we would normally, but I think that's okay. If the resolver is resolving `getProperties` for the given account ID, then we can consider `isDoingGetProperties` true.

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -378,28 +239,10 @@ export const reducer = ( state, { type, payload } ) => {
...
-
+		case RECEIVE_PROPERTIES_PROFILES_COMPLETION: {
 			return {
 				...state,
-				error,
-				isFetchingPropertiesProfiles: {
-					...state.isFetchingPropertiesProfiles,
-					[ accountID ]: false,
-				},
+				isAwaitingPropertiesProfilesCompletion: false,
```
aaemnnosttv, line 245: I'm not sure we need this additional flag - see my other comment in the isDoing selector.

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -36,95 +36,73 @@ import { STORE_NAME, ACCOUNT_CREATE, PROPERTY_CREATE, FORM_ACCOUNT_CREATE } from
...
+		return API.get( 'modules', 'analytics', 'accounts-properties-profiles', data, {
+			useCache: false,
+		} );
+	},
+	reducerCallback: ( state ) => {
+		// Actual accounts, properties, profiles are set by resolver with
+		// custom logic, hence here we just set a flag.
+		return {
+			...state,
+			isAwaitingAccountsPropertiesProfilesCompletion: true,
```
aaemnnosttv, line 54: I'm not sure this flag is necessary. See my comments in the properties/profile stores, but I think we can get the same data with built-in resolution selectors which is better since it's args specific (that doesn't matter as much for this one since there are no args, but I don't think it's needed.

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -36,95 +36,73 @@ import { STORE_NAME, ACCOUNT_CREATE, PROPERTY_CREATE, FORM_ACCOUNT_CREATE } from
...
+		};
+	},
+	argsToParams: ( data ) => {
+		invariant( 'object' === typeof data, 'data must be an object.' );
```
aaemnnosttv, line 58: Let's use `isPlainObject` here 👍

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -36,95 +36,73 @@ import { STORE_NAME, ACCOUNT_CREATE, PROPERTY_CREATE, FORM_ACCOUNT_CREATE } from
...
+	},
+	reducerCallback: ( state, accountTicket ) => {
+		const { id } = accountTicket;
+		return {
+			...state,
+			accountTicketID: id,
+		};
+	},
+	argsToParams: ( data ) => {
+		invariant( 'object' === typeof data, 'data must be an object.' );
```
aaemnnosttv, line 76: Let's use `isPlainObject` here 👍

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -321,11 +210,31 @@ export const resolvers = {
...
+			const { response } = yield fetchGetAccountsPropertiesProfilesStore.actions.fetchGetAccountsPropertiesProfiles( {
 				existingPropertyID: existingTag,
 			} );
 
 			if ( response ) {
+				const { dispatch } = registry;
+
+				dispatch( STORE_NAME ).receiveGetAccounts( response.accounts );
+
+				if ( response.properties.length && response.properties[ 0 ] && response.properties[ 0 ].accountId ) {
```
aaemnnosttv, line 222:
```suggestion
				if ( response.properties?.[ 0 ]?.accountId ) {
```

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -321,11 +210,31 @@ export const resolvers = {
...
+					const accountID = response.properties[ 0 ].accountId;
+					dispatch( STORE_NAME ).receiveGetProperties( response.properties, { accountID } );
+				}
+
+				if ( response.profiles.length && response.profiles[ 0 ] && response.profiles[ 0 ].webPropertyId ) {
```
aaemnnosttv, line 227:
```suggestion
				if ( response.profiles?.[ 0 ]?.webPropertyId ) {
```

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -396,7 +305,18 @@ export const selectors = {
...
+			return true;
+		}
+		// Since isFetchingGetAccountsPropertiesProfiles (via createFetchStore)
+		// holds information based on specific values but we only need
+		// generic information here, we need to check whether ANY such
+		// request is in progress.
+		if ( 'object' !== typeof state.isFetchingGetAccountsPropertiesProfiles ) {
+			return false;
+		}
+		return Object.values( state.isFetchingGetAccountsPropertiesProfiles ).some( ( value ) => value );
```
aaemnnosttv, line 319: Would need to become a registry selector, but now about this?
```suggestion
		return select( STORE_NAME ).isResolving( 'getAccounts' );
```

assets/js/googlesitekit/data/create-fetch-store.test.js
felixarntz: Yeah that's what I was thinking, we don't need to go that deep for most of our functionality, but since this is used in so many places, I wanted to make sure to have very solid coverage.

assets/js/googlesitekit/data/create-fetch-store.js
felixarntz: That's a fair point, but I think we can go with this for now, let's see if we run into issues because of it. `argsToParams` should be fairly basic, just check for required parameters and validate as needed. Since this is only a (usually internal) selector to check whether a request is in progress, I'd even more so say the problem vector here is rather low.

assets/js/googlesitekit/data/create-fetch-store.js
felixarntz: I think it's better to include it so that the function returns a full store. We're using "partial store objects" wherever we combine them with `Data.combineStores` (so that it wouldn't matter), but any store object that gets exported or returned from a function should be complete in itself I'd argue.

assets/js/googlesitekit/data/create-fetch-store.test.js
```diff
@@ -0,0 +1,319 @@
...
+						return {
+							requiredParam,
+						};
+					},
+				} );
+
+				expect( () => {
+					const response = {};
+					fetchStoreDefinition.actions.receiveSaveSomeData( response );
+				} ).toThrow( 'params is required.' );
```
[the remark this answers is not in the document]
felixarntz: When inspecting closer, it seems we don't need to `muteConsole` when expecting errors to be thrown - only where it's not explicitly caught like that, or where we use a `consoleErrorSpy`.

assets/js/googlesitekit/data/create-notifications-store.test.js
```diff
@@ -178,24 +178,24 @@ describe( 'createNotificationsStore store', () => {
...
 
-		describe( 'receiveNotifications', () => {
-			it( 'requires the notifications param', () => {
+		describe( 'receiveGetNotifications', () => {
+			it( 'requires the response param', () => {
 				expect( () => {
-					dispatch.receiveNotifications();
-				} ).toThrow( 'notifications is required.' );
+					dispatch.receiveGetNotifications();
+				} ).toThrow( 'response is required.' );
```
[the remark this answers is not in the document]
felixarntz: See above, this isn't needed here. I've covered some other cases I found though where we just issue API requests without mocks.

assets/js/googlesitekit/data/create-settings-store.js
```diff
@@ -186,91 +165,14 @@ export const createSettingsStore = ( type, identifier, datapoint, {
...
-			} );
-		},
-		[ FETCH_SAVE_SETTINGS ]: ( { payload } ) => {
-			const { values } = payload;
-			return API.set( type, identifier, datapoint, values );
-		},
-	};
+	const controls = {};
 
 	const reducer = ( state = INITIAL_STATE, { type, payload } ) => { // eslint-disable-line no-shadow
```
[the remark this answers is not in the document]
felixarntz: Yes it's because `type` is also a parameter of `createSettingsStore`.

assets/js/googlesitekit/data/create-settings-store.test.js
```diff
@@ -101,27 +101,28 @@ describe( 'createSettingsStore store', () => {
 			} );
 		} );
 
-		describe( 'fetchSettings', () => {
+		describe( 'fetchGetSettings', () => {
 			it( 'does not require any params', () => {
+				muteConsole( 'error' ); // Ignore the API fetch failure here.
```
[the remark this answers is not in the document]
felixarntz: Hmm I just saw that now, I went the exact opposite way and added missing `muteConsole` calls to these places 😄 

I think that's just less code and as long as it's annotated makes sense for these where we don't care about the actual request.

assets/js/googlesitekit/datastore/site/connection.test.js
```diff
@@ -60,24 +62,24 @@ describe( 'core/site connection', () => {
 	} );
 
 	describe( 'actions', () => {
-		describe( 'fetchConnection', () => {
+		describe( 'fetchGetConnection', () => {
 			it( 'does not require any params', () => {
 				expect( () => {
-					registry.dispatch( STORE_NAME ).fetchConnection();
+					registry.dispatch( STORE_NAME ).fetchGetConnection();
```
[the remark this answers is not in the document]
felixarntz: See above

assets/js/modules/analytics/datastore/accounts.js
felixarntz: I generally like this suggestion, however since this would couple the function to the resolver, at that point we might as well have components that need this information call `isResolving` themselves. I wonder whether we should rather rely on that than introducing an unnecessary selector.

assets/js/modules/analytics/datastore/properties.js
felixarntz: See my other comment in response to your suggestion, I think that's a good point (and I don't like the flag used here and in `accounts.js` overall), but we should look into it separately.

For this specifically, it makes sense these requests could easily conflict, so I've fixed it by keying by `accountID`.

stories/module-analytics-setup.stories.js
felixarntz: Not sure how this passed before actually - we were passing accounts here which should be enough to not be considered loading state. So actually we shouldn't pass it here. But I fail to understand how this previously was working... 🤔

aaemnnosttv: It's this line where we manually put it in a `isDoingGetAccounts` state.
```js
registry.stores[ STORE_NAME ].store.dispatch( { type: 'START_FETCH_ACCOUNTS_PROPERTIES_PROFILES' } );
```
I fixed it 👍

assets/js/modules/analytics/datastore/accounts.js
aaemnnosttv: Yeah, I don't think the logic can be entirely replaced by the one selector, but I think this should cover it:
```js
return select( STORE_NAME ).isResolving( 'getAccounts' )
    || return select( STORE_NAME ).isFetchingGetAccountsPropertiesProfiles()
```
In most cases the resolver is what makes the request (I believe tests are the only current exception here), so there is already some coupling to the resolver in that way. However, it is possible to dispatch the action directly so that's what the "OR" logic handles. As an aside, I don't think the `typeof isFetchingGetAccountsPropertiesProfiles` checks are necessary anymore now that `INITIAL_STATE[ isFetching ]` is set.

I would much more prefer using something built-in that we already have and omitting a `reducerCallback` for these "combined requests" than adding new state and actions as a workaround.

I'm not suggesting we make this change everywhere, that's indeed a conversation for another issue, but I think it makes sense for `isDoingGetAccounts` and `isDoingGetProperties`.

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -408,7 +329,14 @@ export const selectors = {
...
 	 */
 	isDoingCreateAccount( state ) {
-		return !! state.isFetchingCreateAccount;
+		// Since isFetchingCreateAccount (via createFetchStore)
+		// holds information based on specific values but we only need
+		// generic information here, we need to check whether ANY such
+		// request is in progress.
+		if ( 'object' !== typeof state.isFetchingCreateAccount ) {
+			return false;
+		}
```
aaemnnosttv, line 338:
```suggestion
```

aaemnnosttv:
@googlebot I consent.

assets/js/googlesitekit/data/create-settings-store.js
```diff
@@ -427,14 +256,21 @@ export const createSettingsStore = ( type, identifier, datapoint, {
...
 		 */
 		isDoingSaveSettings( state ) {
-			return state.isFetchingSaveSettings;
+			// Since isFetchingSaveSettings (via createFetchStore)
+			// holds information based on specific values but we only need
+			// generic information here, we need to check whether ANY such
+			// request is in progress.
+			if ( 'object' !== typeof state.isFetchingSaveSettings ) {
+				return false;
+			}
```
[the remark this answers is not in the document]
felixarntz: Fixed.

assets/js/modules/analytics/datastore/accounts.js
felixarntz: I'd rather revisit this separately throughout, potentially our `isDoingGet*` selectors are entirely unnecessary and we should instead rely on checking resolvers only. This would also mean though that, alongside `receive*` calls in test setups we would also need to mark the relevant resolvers as finished.

felixarntz:
@aaemnnosttv Thanks for pushing the additional changes. I've added a few more commits based on your feedback. Regarding the workarounds for the `isAwaiting*Completion` state properties, let's keep these for now, and I'll open an issue for looking into how we should deal with `isDoingGet*` in the future.

PR #1370, opened 2020-04-09
Handle the Analytics provisioning TOS callback.

adamsilverstein:
## Summary

Handle the Terms of Service redirect during Analytics provisioning.

Addresses issue #1213.  

## Relevant technical choices

* Adds a new hook on `admin_init` to watch for the `gatoscallback` parameter.
* Check the returned account_ticket_id against the one stored in a transient at the beginning of the process.
* Delete the transient.
* Redirect to the dashboard on success, or back to the analytics screen with an `error_code` on failure.

Note: currently I am not retrieving and storing the `internalWebPropertyID` - do we need this value? If not, I will create a follow up issue to remove it everywhere.

## Checklist

- [ ] My code is tested and passes existing unit tests.
- [ ] My code has an appropriate set of unit tests which all pass.
- [ ] My code is backward-compatible with WordPress 4.7 and PHP 5.6.
- [ ] My code follows the [WordPress](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) coding standards.
- [ ] My code has proper inline documentation.
- [ ] I have signed the Contributor License Agreement (see <https://cla.developers.google.com/>).

Issue #1211, opened 2020-03-04
Introduce Analytics account provisioning JS datastore actions/selectors and component


adamsilverstein:
## Feature Description

Show a helpful screen to users who don't have an Analytics account, they get a setup screen with a "Setup Account" action button. (Note: the exact UI/language for this screen is still to be determined.)

### Details
When activating Analytics, the user first goes thru the auth flow, then arrives at the Analytics setup screen. When no user account is detected, the view should show a (new) screen with all the [required GA account fields](https://developers.google.com/analytics/devguides/config/provisioning/v3/user-controlled#technical-ticket): eg. account name, website url, profile name, view name and an action button. 

Clicking the action button sends all fields to a REST request to the plugin (to a new `provision` data endpoint covered in a separate issue). 

Clicking the action button also sends an analytics event tracking event:
`trackEvent( 'analytics_setup', 'new_account_setup_clicked' );`

Display loading indicator on submission, then handle success or error response.

Error screen: Language/UI TBD.

Success Screen:
* Language (TBD) to confirm the account was successfully created “Configure Analytics” button takes the user to the dashboard.
* Clicking button also sends tracking event:
`trackEvent( 'analytics_setup', 'analytics_configured' );`

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

Users with no GA account or users who choose to create a new account should be able to do so directly from the plugin. 
Full description of the flow is already in "Details" above, high-level ACs here: 
* Display new screen with all necessary fields (as many as possible pre-filled) for the user to approve or edit
    * Account: pre-filled name (Site Title) --> user should be able to edit this 
    * Property: pre-filled domain of the site URL --> user should be able to edit this 
    * Profile: pre-filled (all website data) --> user should be able to edit this 
    * Timezone: pre-filled (from WP site's timezone) --> user should be able to pick this from dropdown 
* Redirect smoothly to GA TOS for the user to approve (see https://developers.google.com/analytics/devguides/config/provisioning/v3/user-controlled#technical-tos-flow-url), using the account ticket ID generated via #1212
* Redirect to main plugin dashboard and show success message for the new account + it will take some time for data to aggregate (top banner) 
* Handle errors during the process and display relevant messages for user to take action 
* Finally take the user to the main dashboard with the standard congrats banner at the top.

## Implementation Brief

* Create a new component to display when the user has no Analytics account: `AnalyticsSetupNoAccountNotice`.
[...]

PR #1389, opened 2020-04-14
Add new Analytics provisioning 'AccountCreate' component.

adamsilverstein:
## Summary

Addresses issue #1211

## Checklist

- [ ] My code is tested and passes existing unit tests.
- [ ] My code has an appropriate set of unit tests which all pass.
- [ ] My code is backward-compatible with WordPress 4.7 and PHP 5.6.
- [ ] My code follows the [WordPress](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) coding standards.
- [ ] My code has proper inline documentation.
- [ ] I have signed the Contributor License Agreement (see <https://cla.developers.google.com/>).

PR #1311, opened 2020-03-30
Add Analytics datastore components and related infra

aaemnnosttv:
## Summary

Addresses issue #1101 

## Checklist

- [x] My code is tested and passes existing unit tests.
- [x] My code has an appropriate set of unit tests which all pass.
- [x] My code is backward-compatible with WordPress 4.7 and PHP 5.6.
- [x] My code follows the [WordPress](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) coding standards.
- [x] My code has proper inline documentation.
- [x] I have signed the Contributor License Agreement (see <https://cla.developers.google.com/>).

PR #1390, opened 2020-04-14
Add Analytics provisioning datastore parts.

adamsilverstein:
## Summary

Datastore `createAccoutTicket` feature. 

Addresses issue #1211

From https://github.com/google/site-kit-wp/pull/1370

## Checklist


felixarntz:
@adamsilverstein Just FYI: Note that we're further simplifying the implementation of API-based datastore actions and selectors, because right now it involves a lot of boilerplate code (as you're noticing in this PR).

See https://github.com/google/site-kit-wp/pull/1387/files#diff-084d6f9b3b4d5a065e515e4b91e750d3R117 for an example of the new simplified approach, and then there's also https://github.com/google/site-kit-wp/pull/1388/files which introduces a utility function which will make it much less boilerplate code. Since these are both WIP, I think it's okay to keep this PR following our current convention, but just wanted to let you know - may be good to have a look in the future.

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -80,12 +82,95 @@ export const actions = {
 			type: RECEIVE_ACCOUNTS_PROPERTIES_PROFILES_FAILED,
 		};
 	},
+
+	/**
+	 * Creates a new Analytics account.
+	 *
+	 * Creates a new Analytics account for a user.
```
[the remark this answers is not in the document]
adamsilverstein: 👍

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -80,12 +82,95 @@ export const actions = {
...
+	 * @return {Function} Generator function action.
+	 */
+	*createAccount( { accountName, propertyName, profileName, timezone } ) {
+		invariant( accountName, 'accountName is required.' );
+		invariant( propertyName, 'propertyName is required.' );
+		invariant( profileName, 'profileName is required.' );
+		invariant( timezone, 'timezone is required.' );
+
+		try {
+			const createAccountTicket = yield actions.fetchCreateAccount( { accountName, propertyName, profileName, timezone } );
```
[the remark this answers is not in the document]
adamsilverstein: 👍

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -33,6 +33,8 @@ const FETCH_ACCOUNTS_PROPERTIES_PROFILES = 'FETCH_ACCOUNTS_PROPERTIES_PROFILES';
 const RECEIVE_ACCOUNTS = 'RECEIVE_ACCOUNTS';
 const RECEIVE_ACCOUNTS_PROPERTIES_PROFILES_COMPLETED = 'RECEIVE_ACCOUNTS_PROPERTIES_PROFILES_COMPLETED';
 const RECEIVE_ACCOUNTS_PROPERTIES_PROFILES_FAILED = 'RECEIVE_ACCOUNTS_PROPERTIES_PROFILES_FAILED';
+const FETCH_CREATE_ACCOUNT = 'FETCH_CREATE_ACCOUNT';
+const RECEIVE_CREATE_ACCOUNT_FAILED = 'RECEIVE_CREATE_ACCOUNT_FAILED';
```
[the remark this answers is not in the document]
adamsilverstein: Ok, I that makes sense - I will add that.

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -122,6 +207,12 @@ export const reducer = ( state, { type, payload } ) => {
 				isFetchingAccountsPropertiesProfiles: false,
 			};
 		}
+		case RECEIVE_CREATE_ACCOUNT_FAILED:
+			const { error } = payload;
+			return {
+				...state,
+				error,
```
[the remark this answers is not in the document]
adamsilverstein: 👍

adamsilverstein:
@felixarntz Thanks for the review, I have addressed your feedback.

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -80,12 +83,97 @@ export const actions = {
...
+	 * @param {Object} args.accountTicket Google Analytics create account ticket object.
+	 */
+	receiveCreateAccount( { accountTicket } ) {
+		invariant( accountTicket, 'accountTicket is required.' );
+
+		// Once we have an account ticket, redirect the user to accept the Terms of Service.
+		const { id } = accountTicket;
+		if ( id ) {
+			document.location = `https://analytics.google.com/analytics/web/?provisioningSignup=false#management/TermsOfService/?api.accountTicketId=${ id }`;
+		}
```
felixarntz, line 144: Strictly we'd also need an action creator here, and we'd need to set the request flag to false. However, since we do a redirect here which will destroy the context, I don't think it's a big deal here.

@tofumatt We talked about this before, and I think it makes sense to have that redirect in the datastore, since it must always happen. Let me know if you have a preference to potentially include this in the datastore in another way.

Potentially it would be cleaner to have a `redirectToTOS` action so that at least the create-account part can be tested in isolation from that.

adamsilverstein:
Added a selector in https://github.com/google/site-kit-wp/pull/1390/commits/99f2e1a4a0fc8384bccb861c86f1400b79e1f766. I'll work on adding some test coverage.

adamsilverstein:
Added some test coverage in https://github.com/google/site-kit-wp/pull/1390/commits/fac987a4d314e69ab841e7f1fcbd3000e78b27f4

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -80,12 +84,101 @@ export const actions = {
...
+	*createAccount( { accountName, propertyName, profileName, timezone } ) {
+		invariant( accountName, 'accountName is required.' );
+		invariant( propertyName, 'propertyName is required.' );
+		invariant( profileName, 'profileName is required.' );
+		invariant( timezone, 'timezone is required.' );
+
+		try {
+			const accountTicket = yield actions.fetchCreateAccount( { accountName, propertyName, profileName, timezone } );
+			yield {
+				type: RECEIVE_CREATE_ACCOUNT,
```
felixarntz, line 109: There always needs to be a `payload` object per convention, so here we should use an empty `payload` object with the `type`.

Also this should probably be something like `CREATE_ACCOUNT_FINISHED`, since right now it's using the same name that intuitively belongs to the `receiveCreateAccount` action creator, while in the implementation here it actually doesn't - so let's use a different name.

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -80,12 +84,101 @@ export const actions = {
...
+		} catch ( error ) {
+			// TODO: Implement an error handler store or some kind of centralized
+			// place for error dispatch...
+			return actions.receiveCreateAccountFailed( { accountName, error } );
+		}
+	},
+
+	*fetchCreateAccount( { accountName, propertyName, profileName, timezone } ) {
+		yield {
+			type: CREATE_ACCOUNT_STARTED,
```
felixarntz, line 122: Same here, there should be an empty object.

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -80,12 +84,101 @@ export const actions = {
...
+	 * Adds a account ticket to process.
+	 *
+	 * Adds the newly-created account ticket to the data store.
+	 *
+	 * @since n.e.x.t
+	 * @private
+	 *
+	 * @param {Object} args               Argument params.
+	 * @param {Object} args.accountTicket Google Analytics create account ticket object.
+	 */
```
felixarntz, line 141: Just FYI, we've decided to not document all these internal `fetch...`/`receive...` etc actions, controls, etc. No need to remove these docs from here, but wanted to let you know for the future. So from the new actions here, theoretically only `createAccount` needs to have docs.

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -208,6 +322,16 @@ export const selectors = {
 
 		return error || null;
 	},
+
+	/**
+	 * Indicates whether account creation is currently in progress.
+	 *
```
felixarntz, line 328: This is missing the since annotation.

assets/js/modules/analytics/datastore/__fixtures__/create-account.json
```diff
@@ -0,0 +1,6 @@
+{
+  "accountName": "Test Account",
+  "propertyName": "Test Property",
+  "profileName": "Test Profile",
+  "timezone": "Test Timezone."
+}
```
felixarntz, line 6: Would be great to use a fixture here that looks like an actual response from our API endpoint (account ticket object with nested objects for accounts etc). We should at least make that change as we start testing against the real API - it'll be easy then to record a real response and place some dummy values in there.

felixarntz, line 6: One thing that we need to change right away is add a test account ticket ID in here - we'll need to test that such an ID is taken from the API response as expected.

assets/js/modules/analytics/datastore/accounts.test.js
```diff
@@ -60,7 +60,94 @@ describe( 'modules/analytics accounts', () => {
...
+				await subscribeUntil( registry,
+					() => (
+						registry.select( STORE_NAME ).isDoingCreateAccount() !== true
+					),
+				);
+
+				// Ensure the proper body parameters were sent.
+				expect( JSON.parse( fetch.mock.calls[ 0 ][ 1 ].body ).data ).toMatchObject(
+					{ accountName, propertyName, profileName, timezone }
+				);
```
felixarntz, line 89: See above comment, this test is missing assertions for the API response being handled as expected. We should be able to look at `document.location` here to ensure it is the expected Analytics TOS URL with the same account ticket ID that we put in the `createAccount` fixture.

assets/js/modules/analytics/datastore/accounts.test.js
```diff
@@ -60,7 +60,94 @@ describe( 'modules/analytics accounts', () => {
 	} );
 
 	describe( 'actions', () => {
+		describe( 'createAccount', () => {
+			it( 'creates an account and adds it to the store ', async () => {
```
felixarntz, line 64: This description isn't correct here, should maybe be something like "creates an account ticket and redirects to the TOS".

assets/js/modules/analytics/datastore/accounts.test.js
```diff
@@ -60,7 +60,94 @@ describe( 'modules/analytics accounts', () => {
...
+						JSON.stringify( response ),
+						{ status: 500 }
+					);
+
+				muteConsole( 'error' );
+				registry.dispatch( STORE_NAME ).createAccount( { accountName, propertyName, profileName, timezone } );
+
+				await subscribeUntil( registry,
+					() => (
+						registry.select( STORE_NAME ).getError()
```
felixarntz, line 137: Better use something else like the `isDoingCreateAccount` selector here to decouple what the wait logic from the error, because that's what this actually tests below.

assets/js/modules/analytics/datastore/accounts.test.js
```diff
@@ -60,7 +60,94 @@ describe( 'modules/analytics accounts', () => {
...
+				);
+
+				expect( registry.select( STORE_NAME ).getError() ).toMatchObject( response );
+
+				// Ignore the request fired by the `getAccounts` selector.
+				muteConsole( 'error' );
+				const accounts = registry.select( STORE_NAME ).getAccounts();
+				// No accounts should have been added yet, as the property creation
+				// failed.
+				expect( accounts ).toEqual( undefined );
```
felixarntz, line 148: No need for these lines 143-148 here, that is unrelated to what the `createAccount` action is doing, since it's not actually adding any accounts.

assets/js/modules/analytics/datastore/accounts.js
adamsilverstein: renamed, added empty payload.

assets/js/modules/analytics/datastore/accounts.js
adamsilverstein: added empty payload object.

assets/js/modules/analytics/datastore/accounts.js
adamsilverstein: 👍

assets/js/modules/analytics/datastore/accounts.js
adamsilverstein: added

assets/js/modules/analytics/datastore/accounts.test.js
adamsilverstein: 👍 Will add that, I wasn't sure document.location would be set properly.

assets/js/modules/analytics/datastore/accounts.test.js
adamsilverstein: updated

assets/js/modules/analytics/datastore/accounts.test.js
adamsilverstein: Right, changed that here.

assets/js/modules/analytics/datastore/accounts.test.js
adamsilverstein: Hmm, jsdom does not support navigation directly and throws an error. This approach might work to enable testing:

https://github.com/jsdom/jsdom/issues/2112#issuecomment-361195043

felixarntz: Let's try using `location.assign` then, that sounds good to me.

assets/js/modules/analytics/datastore/accounts.js
felixarntz: @tofumatt @aaemnnosttv @adamsilverstein I'm not aware of any downside of using `window.location.assign()` instead of `document.location` - just pinging you all in case you know of any reason why we shouldn't be using it.

assets/js/modules/analytics/datastore/accounts.test.js
```diff
@@ -36,8 +36,15 @@ import * as fixtures from './__fixtures__';
...
 describe( 'modules/analytics accounts', () => {
 	let apiFetchSpy;
+	let windowLocationSpy;
 	let registry;
 	let store;
+	let redirect;
+
+	const accountName = 'Test Account';
+	const propertyName = 'Test Property';
+	const profileName = 'Test Profile';
```
felixarntz, line 46: Let's take these three values from `fixtures.accountTicket` instead, they're duplicated here, and it's not clear that they should be equal because of that. Also, let's define these variables in each test where they're needed, it makes the individual tests easier to understand just by looking at the test code itself.

assets/js/modules/analytics/datastore/accounts.test.js
```diff
@@ -48,6 +55,15 @@ describe( 'modules/analytics accounts', () => {
 		store = registry.stores[ STORE_NAME ].store;
 
 		apiFetchSpy = jest.spyOn( { apiFetch }, 'apiFetch' );
+
+		/* eslint-disable no-restricted-globals */
+		windowLocationSpy = jest.spyOn( window.location, 'assign' );
```
felixarntz, line 60: Maybe `locationAssignSpy` would be a more precise name? `windowLocationSpy` sounds like it would cover all functions of `window.location`.

assets/js/modules/analytics/datastore/accounts.test.js
```diff
@@ -57,10 +73,83 @@ describe( 'modules/analytics accounts', () => {
...
+						registry.select( STORE_NAME ).isDoingCreateAccount() !== true
+					),
+				);
+
+				// Ensure the proper body parameters were sent.
+				expect( JSON.parse( fetch.mock.calls[ 0 ][ 1 ].body ).data ).toMatchObject(
+					{ accountName, propertyName, profileName, timezone }
+				);
+
+				expect( redirect ).toEqual( 'https://analytics.google.com/analytics/web/?provisioningSignup=false#management/TermsOfService/?api.accountTicketId=abc123' );
```
felixarntz, line 105: Maybe use template literals to get the `abc123` from the `fixtures.accountTicket` object instead? Right now it's not clear where this value is looking from in the test.

assets/js/modules/analytics/datastore/__fixtures__/create-account.json
adamsilverstein: updated fixture

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -80,12 +84,105 @@ export const actions = {
...
+	*fetchCreateAccount( { accountName, propertyName, profileName, timezone } ) {
+		yield {
+			payload: {},
+			type: CREATE_ACCOUNT_STARTED,
+		};
+
+		return {
+			payload: { accountName, propertyName, profileName, timezone },
+			type: FETCH_CREATE_ACCOUNT,
+		};
```
felixarntz, line 130: Not a big deal, but it feels cleaner to have the `CREATE_ACCOUNT_FINISHED` action creator part in here too, instead of separately above. You should be able to do `response = yield ...` on the above object, then add the `yield` for `CREATE_ACCOUNT_FINISHED`, and at the very end `return response;`. See e.g. https://github.com/google/site-kit-wp/pull/1311/files#diff-0794af76964966f1028087bd15b64452R57-R60, https://github.com/google/site-kit-wp/pull/1311/files#diff-0794af76964966f1028087bd15b64452R71-R74, and https://github.com/google/site-kit-wp/pull/1311/files#diff-0794af76964966f1028087bd15b64452R87 for a similar code example.

assets/js/modules/analytics/datastore/accounts.js
adamsilverstein: I can add an inline comment to note that this is so we can test the location has been set. I think I can just use `location.assign` here (dropping the `window` part and remove the linting comments.

felixarntz: Not sure a comment is necessary. If there's no trade-off here, maybe we should be using this throughout our codebase for the future, since generally a function like this is better for testing.

assets/js/modules/analytics/datastore/accounts.test.js
adamsilverstein: Good suggestion, will change.

assets/js/modules/analytics/datastore/accounts.test.js
adamsilverstein: Good idea, I think this can be changed to just `location` as well, I'll give that a try.

assets/js/modules/analytics/datastore/accounts.test.js
adamsilverstein: 👍

assets/js/modules/analytics/datastore/accounts.js
adamsilverstein: ok, i'll try it that way.

assets/js/modules/analytics/datastore/accounts.js
aaemnnosttv: @adamsilverstein you're getting the eslint error because we don't allow use of `window` - use `global` instead.

I agree that we should use methods on `location` for changes.  I did some looking to see if there's any reason why we shouldn't use it and couldn't find anything. Assigning a URL to `location` is also valid per [the HTML standard](https://html.spec.whatwg.org/multipage/history.html#the-location-interface) but probably harder to test.

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -80,12 +84,105 @@ export const actions = {
...
+	 * @param {Object} args       Argument params.
+	 * @param {Object} args.error Error object.
+	 * @return {Object} Redux-style action.
+	 */
+	receiveCreateAccountFailed( { error } ) {
+		return {
+			payload: { error },
+			type: RECEIVE_CREATE_ACCOUNT_FAILED,
+		};
+	},
```
aaemnnosttv, line 171: Per the new conventions, we don't need a dedicated action creator for the fail (now called `CATCH_`) case as this is handled in the main action's generator.

assets/js/modules/analytics/datastore/accounts.test.js
aaemnnosttv: > Good idea, I think this can be changed to just `location` as well, I'll give that a try.

Agreed. I think `locationSpy` would be better here. One thing to keep in mind (although I don't think it makes a difference here) is that Jest proxies the calls through to the real implementation, so if you don't want that you can set your own implementation.

assets/js/modules/analytics/datastore/accounts.test.js
```diff
@@ -57,10 +73,83 @@ describe( 'modules/analytics accounts', () => {
...
+						JSON.stringify( fixtures.accountTicket ),
+						{ status: 200 }
+					);
+
+				muteConsole( 'error' );
+
+				registry.dispatch( STORE_NAME ).createAccount( { accountName, propertyName, profileName, timezone } );
+				await subscribeUntil( registry,
+					() => (
+						registry.select( STORE_NAME ).isDoingCreateAccount() !== true
```
aaemnnosttv, line 96: Nitpick, but I think it's easier to read this way:
```suggestion
						registry.select( STORE_NAME ).isDoingCreateAccount() === false
```

assets/js/modules/analytics/datastore/accounts.test.js
aaemnnosttv: Why does this URL have 2 query strings?

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -208,6 +326,18 @@ export const selectors = {
...
+	/**
+	 * Indicates whether account creation is currently in progress.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {boolean} True if an account is being created, false otherwise.
+	 */
+	isDoingCreateAccount( state ) {
+		return state.isCreatingAccount;
```
aaemnnosttv, line 339: This should be initialized in the `INITIAL_STATE` and be called `isFetchingCreateAccount`

Also, we `!!` the return value to guarantee it always returns a boolean.

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -80,12 +84,105 @@ export const actions = {
...
+		};
+
+		return {
+			payload: { accountName, propertyName, profileName, timezone },
+			type: FETCH_CREATE_ACCOUNT,
+		};
+	},
+
+	/**
+	 * Adds a account ticket to process.
```
aaemnnosttv, line 134:
```suggestion
	 * Adds an account ticket to process.
```

assets/js/modules/analytics/datastore/__fixtures__/account-ticket.json
```diff
@@ -0,0 +1,15 @@
+{
```
aaemnnosttv, line 1: So far all of the fixtures here are snapshots of the response for a given request. The name of the fixture should then match the request that returns it.

This seems to be the response for `fetchCreateAccount` so this should probably be called `create-account.json` (we don't have fixtures for the main action creator since it's `response` would be the same.

assets/js/modules/analytics/datastore/accounts.test.js
adamsilverstein: removed

assets/js/modules/analytics/datastore/accounts.test.js
adamsilverstein: That is a typo! Fixing.

assets/js/modules/analytics/datastore/__fixtures__/account-ticket.json
```diff
@@ -0,0 +1,15 @@
...
+    "name": "Test Account"
+  },
+  "webproperty": {
+    "name": "Test Property"
+  },
+  "profile": {
+    "name": "Test Profile"
+
+  },
+  "redirectUri": "https://redirect.com?gatoscllback"
```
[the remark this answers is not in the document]
adamsilverstein: adjusted

assets/js/modules/analytics/datastore/__fixtures__/account-ticket.json
```diff
@@ -0,0 +1,15 @@
+{
+  "id": "abc123",
+  "kind": "analytics#accountTicket",
+  "account": {
+    "name": "Test Account"
+  },
+  "webproperty": {
```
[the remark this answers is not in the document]
adamsilverstein: It is all lowercase, ref: https://developers.google.com/analytics/devguides/config/provisioning/v3/reference/provisioning#resource

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -33,6 +33,10 @@ const FETCH_ACCOUNTS_PROPERTIES_PROFILES = 'FETCH_ACCOUNTS_PROPERTIES_PROFILES';
 const RECEIVE_ACCOUNTS = 'RECEIVE_ACCOUNTS';
 const RECEIVE_ACCOUNTS_PROPERTIES_PROFILES_COMPLETED = 'RECEIVE_ACCOUNTS_PROPERTIES_PROFILES_COMPLETED';
 const RECEIVE_ACCOUNTS_PROPERTIES_PROFILES_FAILED = 'RECEIVE_ACCOUNTS_PROPERTIES_PROFILES_FAILED';
+const FETCH_CREATE_ACCOUNT = 'FETCH_CREATE_ACCOUNT';
+const CREATE_ACCOUNT_FINISHED = 'CREATE_ACCOUNT_FINISHED';
+const RECEIVE_CREATE_ACCOUNT_FAILED = 'RECEIVE_CREATE_ACCOUNT_FAILED';
+const CREATE_ACCOUNT_STARTED = 'CREATE_ACCOUNT_STARTED';
```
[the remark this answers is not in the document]
adamsilverstein: Got it, renamed.

assets/js/modules/analytics/datastore/accounts.js
adamsilverstein: moved the try catch in https://github.com/google/site-kit-wp/pull/1390/commits/26e7f06b0ee2bce7eab7540d487c41a0a47cc2ad

adamsilverstein:
> I agree that we should use methods on location for changes. I did some looking to see if there's any reason why we shouldn't use it and couldn't find anything. Assigning a URL to location is also valid per the HTML standard but probably harder to test.

Changing this to `location` works and is testable.

assets/js/modules/analytics/datastore/__fixtures__/account-ticket.json
adamsilverstein: Ah, I was naming it after the object it was a fixture for (the "Account Ticket". I renamed and changed it to `createAccount` in https://github.com/google/site-kit-wp/pull/1390/commits/20a4f07055283cb62d046f5534c6c2db474c230d.

assets/js/modules/analytics/datastore/accounts.js
adamsilverstein: Fixed in https://github.com/google/site-kit-wp/pull/1390/commits/4e32769c2ddb9168939a1f9101fac2826a3a1ce1.

assets/js/modules/analytics/datastore/accounts.js
adamsilverstein: Added in https://github.com/google/site-kit-wp/pull/1390/commits/15a35872b4625c2ebde23c2ca1241033b7158687.

assets/js/modules/analytics/datastore/accounts.test.js
adamsilverstein: Good suggestion. Possible because before the correct setup  (and !! casting) you suggested I wasn't certain this would `=== false`. I am now though and changed in https://github.com/google/site-kit-wp/pull/1390/commits/28571eef3c39f85248bd972864b2d1c129ab8db4.

assets/js/modules/analytics/datastore/accounts.js
adamsilverstein: @aaemnnosttv Can you review what I have now and let me know if that looks right?

adamsilverstein:
@aaemnnosttv & @felixarntz Thanks for the review and feedback, I have addressed each issue. Please let me know if I missed anything or need to further adjust the action creator (see comment above).

assets/js/modules/analytics/datastore/accounts.js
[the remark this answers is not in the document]
adamsilverstein: removed.

assets/js/modules/analytics/datastore/accounts.js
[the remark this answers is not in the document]
adamsilverstein: now called `accountTicket`

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -80,12 +85,104 @@ export const actions = {
...
+	/**
+	 * Adds an account ticket to process.
+	 *
+	 * @since n.e.x.t
+	 * @private
+	 *
+	 * @param {Object} args               Argument params.
+	 * @param {Object} args.accountTicket Google Analytics create account ticket object.
+	 */
+	receiveCreateAccount( { accountTicket } ) {
```
felixarntz, line 145: No need to wrap this in an object, this could become `receiveCreateAccount( accountTicket )` for simpler usage (the `accountTicket` is also directly the above `response` from the API).

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -80,12 +85,104 @@ export const actions = {
...
+	 * Logs an error with account creation.
+	 *
+	 * @since n.e.x.t
+	 * @private
+	 *
+	 * @param {Object} args       Argument params.
+	 * @param {Object} args.error Error object.
+	 * @return {Object} Redux-style action.
+	 */
+	receiveCreateAccountFailed( { error } ) {
```
felixarntz, line 166: This function name is now "out of sync" with the action type it dispatches `CATCH_FETCH_CREATE_ACCOUNT`. We don't have this type of function elsewhere in our code per latest conventions, so it'd be easiest to remove this function entirely and move the return statement for `CATCH_FETCH_CREATE_ACCOUNT` directly into the `catch` clause within `fetchCreateAccount()`.

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -80,12 +85,104 @@ export const actions = {
...
+	 */
+	*createAccount( { accountName, propertyName, profileName, timezone } ) {
+		invariant( accountName, 'accountName is required.' );
+		invariant( propertyName, 'propertyName is required.' );
+		invariant( profileName, 'profileName is required.' );
+		invariant( timezone, 'timezone is required.' );
+
+		const accountTicket = yield actions.fetchCreateAccount( { accountName, propertyName, profileName, timezone } );
+
+		return actions.receiveCreateAccount( { accountTicket } );
```
felixarntz, line 109: Let's move this call directly into `fetchCreateAccount()` (after retrieving the `response`), like it is handled elsewhere. This will also simplify this action creator to only call the `yield actions.fetchCreateAccount` above.

felixarntz:
@adamsilverstein I updated `feature/ga-provisioning` with the GA refactoring bits which are mostly completed now. Can you make sure to merge into this branch to ensure things work? There are a few merge conflicts that need to be resolved.

assets/js/modules/analytics/datastore/accounts.js
adamsilverstein: Moved in https://github.com/google/site-kit-wp/pull/1390/commits/acb3c3134ca03c53c4058b593fbae64dd371f123, please review.

assets/js/modules/analytics/datastore/accounts.js
adamsilverstein: ok, changing. Using this more by convention.

assets/js/modules/analytics/datastore/accounts.js
tofumatt: I don't like the inclusion of the `window.location` call here because:

1. It makes the data store more difficult to test; we need to mock `window.location` to test a data store, which doesn't make a whole lot of sense to me.
2. Separation of concerns: adjusting the location of the page isn't a datastore concern; it's a component/UI concern. This is logic that is better-placed in a `useEffect()` call of the component that dispatches whatever action triggers this behaviour. If the `id` is received then we perform this redirect, but in the UI (eg the component), not in the datastore.

Our datastore should not ever be concerned with routing or page URL in such a _direct_ way. I think it would be better to exclude this from the PR and incorporate this logic into the component that dispatches these actions.

-----

If we want a more generic way to handle redirects I'm even fine with a `setRedirect( url )` action and a `getRedirect()` selector in a high-level component that handles this at an app-wide level (some component high up in the component tree whose only job it is is to select that data and redirect with `window.location` when it happens). But I don't think this should be done here.

felixarntz: Thanks for the feedback @tofumatt, that's what I had pinged you about before.

That makes sense to me, it's a bit odd that the datastore handles this, but since it destroys the page context anyway I wasn't sure what's recommended here.

Based on this, I suggest we modify `receiveCreateAccount` to return an action creator object `{ payload: { accountTicket }, type: RECEIVE_CREATE_ACCOUNT }`, and then add a `RECEIVE_CREATE_ACCOUNT` case in the reducer that sets a state property `accountTicketTOSURL` or `accountTicketTermsOfServiceURL` (the latter sounds better, but also is _very_ long, so I'm not sure). A selector `getAccountTicketTOSURL()` or `getAccountTicketTermsOfServiceURL()` would then allow to get that URL once set, so that the component can take care of redirecting there. (Note that the selector/state property should be more specific to what this URL is - it shouldn't involve "redirect" in its name, since that has been a kind of "reserved" term for OAuth redirect URL in our plugin, so describing what this URL is for I think is more accurate.)

adamsilverstein: Ok, thanks for the feedback @tofumatt I agree with your points and was originally planning to have the redirect in the component itself based on the returned value. I will make it so.

assets/js/modules/analytics/datastore/accounts.js
adamsilverstein: removed & moved return in https://github.com/google/site-kit-wp/pull/1390/commits/33abf75cfa23406300c0c0c0f39504f7ba26e128

assets/js/modules/analytics/datastore/accounts.js
adamsilverstein: Changed in https://github.com/google/site-kit-wp/pull/1390/commits/d4f1bc132be3e220d3b73747b1d7ed8b7f6b9c99

assets/js/modules/analytics/datastore/accounts.js
adamsilverstein: @felixarntz I changed this as you suggested in https://github.com/google/site-kit-wp/pull/1390/commits/d4f1bc132be3e220d3b73747b1d7ed8b7f6b9c99 - one other option would be to set only the ticket or id, then construct the url in the client. Sticking with setting the URL here and I'll add the redirect logic to the component.

adamsilverstein:
Addressed feedback, resolving merge conflicts next...

adamsilverstein:
Ready for another review! Added the redirect handling to the component: https://github.com/google/site-kit-wp/pull/1389/commits/cb6e02749a0d63d98c4a34b20bc23b4b4d3a0e33

assets/js/modules/analytics/datastore/accounts.js
felixarntz: I think dealing with the URL in the datastore keeps that functionality encapsulated there. From component POV, the only thing it needs is the URL, not the account ticket ID, that's why I thought exposing the full URL would make more sense there.

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -35,13 +35,20 @@ const FETCH_ACCOUNTS_PROPERTIES_PROFILES = 'FETCH_ACCOUNTS_PROPERTIES_PROFILES';
...
+const CATCH_FETCH_CREATE_ACCOUNT = 'CATCH_FETCH_CREATE_ACCOUNT';
 
 const RECEIVE_ACCOUNTS = 'RECEIVE_ACCOUNTS';
 const RESET_ACCOUNTS = 'RESET_ACCOUNTS';
 
 export const INITIAL_STATE = {
 	accounts: undefined,
 	isFetchingAccountsPropertiesProfiles: false,
+	isFetchingCreateAccount: false,
+	accountTicketTermsOfServiceURL: false,
```
felixarntz, line 51: This should be `undefined` initially. Any values not loaded should use `undefined`, the only exception from that are the flags for e.g. requests being in progress, which can just be booleans (like the two above).

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -135,6 +142,60 @@ export const actions = {
...
+				payload: {},
+				type: FINISH_FETCH_CREATE_ACCOUNT,
+			};
+		} catch ( error ) {
+			return {
+				payload: { error },
+				type: CATCH_FETCH_CREATE_ACCOUNT,
+			};
+		}
+		return accountTicket;
```
felixarntz, line 188: This should preferably return `{ response, error }` instead (with `response` being `accountTicket`) for consistency with the other APIs. `error` would be the error caught above, or `undefined`. See e.g. https://github.com/google/site-kit-wp/blob/a1e01f986e03f2b5dfc1b23ad06e76579cff6192/assets/js/modules/analytics/datastore/accounts.js#L87

assets/js/modules/analytics/datastore/accounts.js
adamsilverstein: Set to undefined in https://github.com/google/site-kit-wp/pull/1390/commits/1e50efb5cb1acd850de1dae2cd14cb3d0dc60dcf.

assets/js/modules/analytics/datastore/accounts.js
adamsilverstein: updated in https://github.com/google/site-kit-wp/pull/1390/commits/5f10856a2893b2ed696a559e4e1066d42a445d90

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -161,31 +161,33 @@ export const actions = {
...
 		};
 
 		try {
-			accountTicket = yield {
+			response = yield {
 				payload: { accountName, propertyName, profileName, timezone },
 				type: FETCH_CREATE_ACCOUNT,
 			};
-
+			const { accountTicket } = response;
```
felixarntz, line 176: This doesn't work I think? The `response` _is_ the `accountTicket` I thought, but here you're destructuring it. I think the `response` can simply be passed to `receiveCreateAccount` below.

adamsilverstein: Oh, right - fixing.

adamsilverstein: Corrected in https://github.com/google/site-kit-wp/pull/1390/commits/8d2616d5a8ee0eb54c9bb104f154314dabc850bc.

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -161,31 +161,33 @@ export const actions = {
...
 			yield {
 				payload: {},
 				type: FINISH_FETCH_CREATE_ACCOUNT,
 			};
-		} catch ( error ) {
-			return {
+		} catch ( e ) {
+			error = e;
+			yield {
 				payload: { error },
```
aaemnnosttv: The payload here should be the same in for all actions, except for the `CATCH` case which has the added error.

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -161,31 +161,33 @@ export const actions = {
 		invariant( profileName, 'profileName is required.' );
 		invariant( timezone, 'timezone is required.' );
 
-		let accountTicket;
+		let response,
+			error;
 		yield {
 			payload: {},
```
aaemnnosttv, line 167: Payload should be the same for all actions except catch which gets the added `error`.

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -194,6 +265,35 @@ export const reducer = ( state, { type, payload } ) => {
...
+				...state,
+				isFetchingCreateAccount: true,
+			};
+		}
+
+		case RECEIVE_CREATE_ACCOUNT: {
+			const { accountTicket: { id } } = payload;
+			return {
+				...state,
+				accountTicketTermsOfServiceURL: `https://analytics.google.com/analytics/web/?provisioningSignup=false#management/TermsOfService/&api.accountTicketId=${ id }`,
```
aaemnnosttv, line 293: Is this URL going to have the same problem as other deep links for users who sign in with multiple accounts?

That is, if you go to this URL but the user you've authenticated Site Kit with is not your primary logged in Google account (e.g. `authuser=1`) will it still work?

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -135,6 +142,62 @@ export const actions = {
...
+	 * @param {string} args.propertyName Google Analytics property name.
+	 * @param {string} args.profileName  Google Analytics profile name.
+	 * @param {string} args.timezone     Google Analytics timezone.
+	 * @return {Function} Generator function action.
+	 */
+	*createAccount( { accountName, propertyName, profileName, timezone } ) {
+		invariant( accountName, 'accountName is required.' );
+		invariant( propertyName, 'propertyName is required.' );
+		invariant( profileName, 'profileName is required.' );
+		invariant( timezone, 'timezone is required.' );
```
aaemnnosttv, line 162: This is somewhat inconsistent in the codebase, but I've been trying to add a bit more context to the error message otherwise it's a pain to track down where it's coming from.
E.g. "accountName is required to create an account."

assets/js/modules/analytics/datastore/__fixtures__/create-account.json
```diff
@@ -0,0 +1,15 @@
...
+    "name": "Test Account"
+  },
+  "webproperty": {
+    "name": "Test Property"
+  },
+  "profile": {
+    "name": "Test Profile",
+    "timezone": "America/Los Angeles"
+  },
+  "redirectUri": "https://redirect.com/wp-admin?gatoscallback"
```
aaemnnosttv, line 14: Should this not be:
```suggestion
  "redirectUri": "https://redirect.com/wp-admin?gatoscallback=1"
```

assets/js/modules/analytics/datastore/accounts.js
adamsilverstein: So include `{ accountName, propertyName, profileName, timezone }` as payload for each yielded action? Will do.

assets/js/modules/analytics/datastore/accounts.js
adamsilverstein: Got it!

assets/js/modules/analytics/datastore/accounts.js
adamsilverstein: Great question, we'll have to test that. Have we been able to fix that elsewhere at all?

assets/js/modules/analytics/datastore/accounts.js
adamsilverstein: 👍 improved in https://github.com/google/site-kit-wp/pull/1390/commits/407a38b75c35236bf697fe5ccd2aa20c89674866

assets/js/modules/analytics/datastore/accounts.js
adamsilverstein: https://github.com/google/site-kit-wp/pull/1390/commits/4b6fa6aedd59bcd4a6f5002d7378eba06276330d

assets/js/modules/analytics/datastore/__fixtures__/create-account.json
adamsilverstein: Eagle eye! Fixed in https://github.com/google/site-kit-wp/pull/1390/commits/a08b6f47508812cda73b72d292ee95faedb76831.

adamsilverstein:
@aaemnnosttv Thanks for the additional feedback, which I addressed.

assets/js/modules/analytics/datastore/accounts.js
aaemnnosttv: Not that I am aware of. As far as I know, we don't have a way to detect which index to use for `authuser`. I would expect that to maybe be available after going through the oAuth flow though 🤔

felixarntz: I have found random threads describing that `authuser` (at least on certain services like Google Drive) also supports passing the Google account's email address. We should try if that works here.

While AdSense doesn't support the `authuser` query parameter and instead uses a `/u/{index}` affix in its URLs, I've already verified that at least there passing the email is supported - it will then internally redirect to the correct index number. So let's try it with the `authuser` parameter and email for Analytics.

adamsilverstein: I'll give that a try, thanks!

adamsilverstein: @felixarntz Brilliant! I verified this works as desired. I first tried the provisioning url with one google account while logged into another in Analytics and was shown an "Error: Access denied" screen. I then edited the url adding &authuser={my other email} - with this URL, Analytics redirected me to a login (as the authuser). when I completed that I reached the TOS screen correctly.

Do you think we should send the email address stored in the user profile along with the request from the plugin, or look the user up on the service/proxy side (and I assume use their first available email address)?

aaemnnosttv: Great find! I think we should pass the email through from the `Profile` as this is set after the user authenticates, so it should always match. This is also what we use for the login hint so I think it would make sense to use here since that's essentially what it is.

@adamsilverstein does the address need to be URL encoded or its fine as-is?

felixarntz: @adamsilverstein @aaemnnosttv Since we now need the user's email address, this is theoretically blocked by #1176, which brings that data into a datastore. But we shouldn't be waiting for this here, as it would delay the work too much. Because whatever we'll do here will be a temporary solution, I think it's okay to just implement the minimal PHP-side part from #1176 (implement a `_googlesitekitUserData` global on the PHP side that outputs this information) - then we can directly access that global from here, and once #1176 is really implemented switch to using the selector.

aaemnnosttv: I confirmed that this works with Search Console as well! 😄 

I just opened a new issue to address this in our existing deep links here: https://github.com/google/site-kit-wp/issues/1456

felixarntz: @adamsilverstein @aaemnnosttv Actually, let's do that in a follow-up PR though. Adding the email address to the URL shouldn't block this one from being completed - it wasn't a requirement in the first place and is rather an enhancement which we'll later need to apply to many areas.

assets/js/modules/analytics/datastore/accounts.test.js
```diff
@@ -39,6 +39,11 @@ describe( 'modules/analytics accounts', () => {
 	let registry;
 	let store;
 
+	const accountName = fixtures.createAccount.account.name;
+	const propertyName = fixtures.createAccount.webproperty.name;
+	const profileName = fixtures.createAccount.profile.name;
+	const timezone = fixtures.createAccount.profile.timezone;
+
```
aaemnnosttv, line 46: These should be declared more locally to the tests that use them (perhaps within `describe( 'createAccount'`) as these variables aren't relevant to the whole file.

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -135,6 +142,62 @@ export const actions = {
...
+	 * @return {Function} Generator function action.
+	 */
+	*createAccount( { accountName, propertyName, profileName, timezone } ) {
+		invariant( accountName, 'accountName is required to create an account.' );
+		invariant( propertyName, 'propertyName is required to create an account.' );
+		invariant( profileName, 'profileName is required to create an account.' );
+		invariant( timezone, 'timezone is required to create an account.' );
+
+		let response,
+			error;
```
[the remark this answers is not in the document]
adamsilverstein: Ok, thanks for the tip, will update.

assets/js/modules/analytics/datastore/accounts.test.js
```diff
@@ -60,6 +65,78 @@ describe( 'modules/analytics accounts', () => {
...
+				fetch
+					.doMockIf(
+						/^\/google-site-kit\/v1\/modules\/analytics\/data\/create-account-ticket/
+					)
+					.mockResponse(
+						JSON.stringify( fixtures.createAccount ),
+						{ status: 200 }
+					);
+
+				muteConsole( 'error' );
```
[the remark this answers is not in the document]
adamsilverstein: I will confirm, may not be required and only added while building the test.

assets/js/modules/analytics/datastore/accounts.test.js
adamsilverstein: 👍

assets/js/modules/analytics/datastore/accounts.js
adamsilverstein: 👍

adamsilverstein: > does the address need to be URL encoded or its fine as-is?

In my test I used as is without encoding.

assets/js/modules/analytics/datastore/accounts.js
tofumatt: Thanks for the fixes on this 👍

assets/js/modules/analytics/datastore/accounts.test.js
adamsilverstein: Turns out the query string part is part of the hash and the `?` is what is specified in the [documentation](https://developers.google.com/analytics/devguides/config/provisioning/v3/user-controlled#technical-tos-flow-url).

assets/js/modules/analytics/datastore/accounts.test.js
adamsilverstein: Moved in https://github.com/google/site-kit-wp/pull/1390/commits/0793a77d37d938987a6496259573efa95ff6152d

assets/js/modules/analytics/datastore/accounts.test.js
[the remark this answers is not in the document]
adamsilverstein: I added an inline doc explaining these, it is required for an error free test run.

assets/js/modules/analytics/datastore/accounts.js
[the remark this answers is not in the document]
adamsilverstein: updated

felixarntz:
I'll merge this now - it says the Travis-CI build is still running, but this must be some issue with Travis-CI/GitHub, since it actually already successfully completed a day ago: https://travis-ci.com/github/google/site-kit-wp/builds/162531183

Issue #1403, opened 2020-04-17
Require only clientID for AdSense datastore functions


felixarntz:
Since the AdSense account ID can be derived from the AdSense client ID, functions in the `modules/adsense` datastore that require both should be modified to only require the `clientID` and parse the account ID from there.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Functions in the `modules/adsense` datastore that require both should be modified to only require the `clientID` and parse the account ID from there.
* The respective response data should be keyed only by the `clientID`.
* The `accountID` should be parsed from `clientID` to make the API request.

## Implementation Brief

* The only relevant part of the AdSense store is the URL channels bit, so exclusively code in there needs to be updated.
* Merge #1405.

## QA Brief

* 

## Changelog entry

* N/A

tofumatt:
IB ✅ Sounds good to me—I had a look at the code in the PR as well and that seems solid. 👍

tofumatt:
(I'm un-assigning @aaemnnosttv fro this as I assumed we were both assigned to see who got to it first, but this is a straightforward AC/IB/issue. If I'm wrong to do it feel free to assign him back, but otherwise I'll move this to code review as it's already implemented.)

tofumatt:
Looks good; assigning to you for merge. 😄

tofumatt:
QA ✅

Code is updated everywhere and everything still works as-expected.

PR #1405, opened 2020-04-18
Simplify URL channels parts of AdSense datastore by only requiring clientID as argument

felixarntz:
## Summary

Addresses issue #1403 

## Relevant technical choices

* The `accountID` is only needed when making the request to the server, hence this PR moves the related logic as closely to this point, which is the control issuing the API request. In case the account ID cannot be parsed (i.e. the client ID is invalid), an error is thrown that matches the error which the server-side would throw if the client ID passed to it was invalid. This nicely abstracts away the fact that the server works slightly different (as in it requires the account ID too).

## Checklist

### DIFF
--- a/assets/js/modules/adsense/datastore/urlchannels.js
+++ b/assets/js/modules/adsense/datastore/urlchannels.js
@@ -20,6 +20,7 @@
  * External dependencies
  */
 import invariant from 'invariant';
+import { __ } from '@wordpress/i18n';
 
 /**
  * Internal dependencies
@@ -27,6 +28,7 @@ import invariant from 'invariant';
 import API from 'googlesitekit-api';
 import Data from 'googlesitekit-data';
 import { STORE_NAME } from './index';
+import { parseAccountID } from '../util';
 
 // Actions
 const FETCH_URLCHANNELS = 'FETCH_URLCHANNELS';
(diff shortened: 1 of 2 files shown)


assets/js/modules/adsense/datastore/urlchannels.js
```diff
@@ -94,7 +93,19 @@ export const actions = {
 };
 
 export const controls = {
-	[ FETCH_URLCHANNELS ]: ( { payload: { accountID, clientID } } ) => {
+	[ FETCH_URLCHANNELS ]: ( { payload: { clientID } } ) => {
+		const accountID = parseAccountID( clientID );
+		if ( 'undefined' === typeof accountID ) {
+			// Mirror the API response that would happen for an invalid client ID.
```
tofumatt, line 99: Why do this locally if the server will throw this sort of response? I don't think the check is incorrect, but I don't know why we'd re-implement it on the client side if the API already handles this.

assets/js/modules/adsense/datastore/urlchannels.test.js
```diff
@@ -127,22 +125,44 @@ describe( 'modules/adsense URL channels', () => {
...
 
 				expect( fetch ).toHaveBeenCalledTimes( 1 );
 
-				const urlchannels = registry.select( STORE_NAME ).getURLChannels( fakeAccountID, fakeClientID );
+				const urlchannels = registry.select( STORE_NAME ).getURLChannels( fakeClientID );
 				expect( urlchannels ).toEqual( undefined );
 			} );
+
+			it( 'dispatches an error when account ID cannot be parsed from client ID', async () => {
+				const clientResponse = {
```
tofumatt, line 145: Seems like this should be mocking a response rather than doing it client-side (see my comment above).

assets/js/modules/adsense/datastore/urlchannels.js
felixarntz: The server requires passing the `accountID` parameter, so we need to parse that on the client. If parsing fails, we know the `clientID` is invalid, and I just wanted to mirror the same response the server would provide for that case.

Also this avoids issuing the server request unnecessarily, since we'd already know here that it will fail.

tofumatt: Ah, okay, that makes sense. 👍

Issue #1296, opened 2020-03-26
Tag Manager: Layout inconsistency in settings after installing AMP


aaemnnosttv:
## Bug Description

If Tag Manager is set up before AMP is installed/enabled, only a web container will be configured in the settings. After, if AMP is enabled, the empty value for the setting creates an inconsistency in the layout

## Steps to reproduce

1. Set up Site Kit with Tag Manager (no AMP)
1. Install and activate AMP in either Transitional or Reader modes (i.e. paired)
1. Go to Settings > Tag Manager
1. See AMP Container ID header out of alignment with other headers

## Screenshots

![image](https://user-images.githubusercontent.com/1621608/77669693-d900c680-6f8d-11ea-9504-29699aaf23cc.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The layout of the Tag Manager settings panel should not look off if there is no AMP container selected while AMP is in use.

## Implementation Brief

* Create a new shared component for displaying a setting value, e.g. `DisplaySetting` (I'm open to better names for this 😄 )
* It should accept the `value` as a prop
* Internally, it should return the value if present, or `"\u00A0"` (unicode for `&nbsp;` which we can't use because React will escape it)
* Update settings views to use `DisplaySetting` for displaying each primitive value.  
I.e. those that are currently `{ value || false }`

If we decide in the future to display something else instead (e.g. "not set", etc) we only need to make the change in one place.

## QA Brief

* Follow the steps to reproduce above
* Ensure the layout is not [distorted](https://user-images.githubusercontent.com/1621608/77669693-d900c680-6f8d-11ea-9504-29699aaf23cc.png)

## Changelog entry

* Fix layout issue in Tag Manager settings panel when no container is selected.

tofumatt:
I would prefer it explicitly say "Not set" rather than displaying an empty string. I think it makes things clearer, more consistent, and doesn't have the potential to look like a bug. But we can discuss that in the PR or later; the basis of this is IB ✅

aaemnnosttv:
@tofumatt I agree, but that can be a discussion for another issue as you said 👍

cole10up:
**Tested**

Installed latest 1.7.1 zip. Activated Analytics and Tag manager without an AMP account. Activate AMP in Transitional mode. Noticed:

![image](https://user-images.githubusercontent.com/57115907/80072549-40f9ec80-84fb-11ea-9d89-bab2c5665cf1.png)

[...]

PR #1414, opened 2020-04-21
Add DisplaySetting component.

jqlee85:
## Summary

Addresses issue [#1296](https://github.com/google/site-kit-wp/issues/1296)

## Relevant technical choices

Adds a DisplaySetting component which takes `value` as a prop, returning the value if it is valid, or `"\u00A0"` if it is falsy. The component is used in `TagmanagerSetup` to display setting values.

## Checklist

### DIFF
--- a/assets/js/components/display-setting.test.js
+++ b/assets/js/components/display-setting.test.js
@@ -1,0 +1,55 @@
+/**
+ * DisplaySetting tests.
+ *
+ * Site Kit by Google, Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Internal dependencies
+ */
+import DisplaySetting from './display-setting';
+import { render } from '../../../tests/js/test-utils';
+
+describe( 'DisplaySetting', () => {
+	it( 'returns non-empty string unchanged', () => {
+		const { container } = render( <DisplaySetting value="Test Value" /> );
+		expect( container.textContent ).toBe( 'Test Value' );
+	} );
+
+	it( 'returns unicode value for &nbsp; if no value prop provided', () => {
+		const { container } = render( <DisplaySetting /> );
+		expect( container.textContent ).toBe( '\u00A0' );
+	} );
+
+	it( 'returns unicode value for &nbsp; if value prop is undefined', () => {
+		const { container } = render( <DisplaySetting value={ undefined } /> );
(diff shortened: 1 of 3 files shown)


assets/js/components/display-setting.test.js
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Internal dependencies
+ */
+import DisplaySetting from './display-setting';
+import { render } from '../../../tests/js/test-utils';
+
+describe( 'DisplaySetting', () => {
+	it( 'returns non-empty string unchanged', () => {
+		const { container } = render( <DisplaySetting value="Test Value" /> );
+		expect( container.textContent ).toBe( 'Test Value' );
```
aaemnnosttv, line 28: React Testing Library uses [`jest-dom`](https://github.com/testing-library/jest-dom) so we can actually use a more expressive matcher for this like so:
```suggestion
		expect( container ).toHaveTextContent( 'Test Value' );
```

Would you please update these tests to use this instead?

aaemnnosttv:
@felixarntz this is beyond the scope of the original issue, but do you think we should extend this to the other modules now or do it as part of the rewrite for each as we transition to the datastore? It probably would be relatively low effort to do. What do you think?

felixarntz:
@aaemnnosttv Let's keep it scoped to GTM for now, and let's open an issue to expand to other modules.

assets/js/components/display-setting.test.js
jqlee85: Yeah, that's the method I tried originally, but `toHaveTextContent()` either normalizes white space or replaces `&nbsp;`s (see reference below) which prevents us from being able to test for that `&nbsp;`. Testing with `textContent` was a workaround. We could use `toHaveTextContent()` for the first case with the non-empty string though. Let me know if I should change that or keep them consistent within this file. (Or if there's a preferred workaround).

[https://github.com/testing-library/jest-dom/blob/d6655dbd0736f9ab8391b59570a0ecd631c93916/src/to-have-text-content.js#L11-L13](https://github.com/testing-library/jest-dom/blob/d6655dbd0736f9ab8391b59570a0ecd631c93916/src/to-have-text-content.js#L11-L13)

aaemnnosttv: That makes sense @jqlee85. I see there is an option to turn that off with `{ normalizeWhitespace: false }` as the second parameter to the matcher, but that's a bit overly verbose for each I think.

On its face, I like your approach better as that's still quite clear and is less verbose but the text content-specific matcher gives us more specific feedback in the event of a failure.

Let's go with the `toHaveTextContent()` matcher for the better feedback (unless there's another reason why we can't), but also for consistency with how we make assertions on text content. We can DRY up the usage with a variable for `options` above that they can all share. Maybe this variable could even be called `preserveWhitespace` to be clear about what it's doing?

If for some reason we still can't use this matcher, then let's just document the exception with a comment that explains why 😄

jqlee85: @aaemnnosttv the `{normalizeWhitespace: false}` option doesn't resolve the issue, unfortunately. It just causes toHaveTextContent() to explicitly replace **nbsp**s with spaces:

`htmlElement.textContent.replace(/\u00a0/g, ' ') // Replace &nbsp; with normal spaces`.

It's as if toHaveTextContent() was specifically designed to break this test :laughing:.

I've updated the test to use toHaveTextContent for the first case, and added comments to the rest to document why we're using toBe instead.

Issue #1415, opened 2020-04-21
Fix AdSense store logic to handle actions correctly


felixarntz:
## Bug Description

The AdSense store was still implemented using old conventions, which itself does not warrant it to be fixed right here, however the old conventions actually were resulting in a problem because some actions had both a control and a reducer, which does not work. See #1361 and #1375 for more information. 

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The AdSense store should use latest fetch logic with individual actions for issuing the request and setting request flags. This logic can then quickly be replaced with `createFetchStore` (#1288) in the future.

## Implementation Brief

* Use latest fetch store conventions, with `START_FETCH...`, `FINISH_FETCH...`, and `CATCH_FETCH_...` actions.
* Pass `useCache: false` to API requests (except for reports which should be cached).
* Merge #1416.

## QA Brief

* 

## Changelog entry

* N/A

tofumatt:
All good; assigned to @felixarntz for merge 😄

tofumatt:
QA ✅

Caching is disabled in the right places and this all functions well. Moving to approval.

PR #1416, opened 2020-04-21
Fix bugs from existing AdSense datastore implementation and update to latest conventions

felixarntz:
## Summary

Addresses issue #1415

## Checklist

### DIFF
--- a/assets/js/modules/adsense/datastore/accounts.js
+++ b/assets/js/modules/adsense/datastore/accounts.js
@@ -30,9 +30,11 @@ import { STORE_NAME } from './index';
 
 // Actions
 const FETCH_ACCOUNTS = 'FETCH_ACCOUNTS';
+const START_FETCH_ACCOUNTS = 'START_FETCH_ACCOUNTS';
+const FINISH_FETCH_ACCOUNTS = 'FINISH_FETCH_ACCOUNTS';
+const CATCH_FETCH_ACCOUNTS = 'CATCH_FETCH_ACCOUNTS';
+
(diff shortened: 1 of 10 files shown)


assets/js/modules/adsense/datastore/accounts.js
```diff
@@ -88,13 +99,15 @@ export const actions = {
 
 export const controls = {
 	[ FETCH_ACCOUNTS ]: () => {
-		return API.get( 'modules', 'adsense', 'accounts' );
+		return API.get( 'modules', 'adsense', 'accounts', undefined, {
```
tofumatt, line 102: Will this take `null` as an argument for `queryParams` instead of `undefined`? I know it's functionally the same, but it might make the code a bit clearer to read is all.

If not, no worries for now.

assets/js/modules/adsense/datastore/tags.js
```diff
@@ -307,16 +349,8 @@ export const selectors = {
 	 * @return {?Object} Object with string `accountID` and boolean `permission` properties; `undefined` if not loaded.
 	 */
 	getTagPermission( state, clientID ) {
```
tofumatt, line 351: This is a nice simplification 👍

assets/js/modules/adsense/datastore/accounts.js
felixarntz: I chose to use `undefined` just to make sure it's exactly the same value as there would be otherwise. Probably `null` would work, but I'd say it's not a big deal anyway :)

Issue #1368, opened 2020-04-09
Analytics set up with no accounts triggers re-auth notice


aaemnnosttv:
## Bug Description

Currently when setting up the Analytics module using a Google account that has no Analytics accounts, the "Issue accessing data" notice is displayed when it should not.

The reason for this is that when a data request fails with an error that has a `data.reason: "insufficientPermissions"` in the response, it triggers this notice.

This exception is thrown by the Google API, which returns this response:
```json
{
  "error": {
    "errors": [
      {
        "domain": "global",
        "reason": "insufficientPermissions",
        "message": "User does not have any Google Analytics account."
      }
    ],
    "code": 403,
    "message": "User does not have any Google Analytics account."
  }
}
```

## Steps to reproduce

1. Set up Site Kit using a Google user that has no Analytics accounts
1. Connect Analytics to initialize the module's set up
1. See error

## Screenshots

![image](https://user-images.githubusercontent.com/1621608/78890682-c34fce80-7a6e-11ea-9dd7-c0427a8ff59d.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `GET:accounts-properties-profiles` datapoint should return a successful response with empty arrays for each if the user has no accounts
* All error handling for no-accounts cases in client code, E2E mocks, etc should be removed and be updated to use the new count-based logic.

## Implementation Brief

* Update `GET:accounts-properties-profiles` in `Analytics::create_data_request` to return a function that makes the same request in a try/catch block
    * Needs to use `$this->with_client_defer( false )` to dispatch the request immediately
* Only catch `Google_Service_Exception`s (any other will be caught by the default error handler)
* If an exception is caught in the case of no accounts, (exception message === `User does not have any Google Analytics account.` - unfortunately there is not a better identifier for this) then return a new `\Google\Site_Kit_Dependencies\Google_Service_Analytics_Accounts` with its items set to an empty array
* If any other exception is caught, re-throw it
* Update `GET:accounts-properties-profiles` in `Analytics::parse_data_request` to return early (and not request properties-profiles) if `$accounts` are empty
* Update consuming client code, E2E mocks, etc to use the new count-based logic

## QA Brief

* Set up Site Kit with a Google user that has no Analytics account
* Initialize the set up for Analytics
*  The "Issue accessing data" notice should not be displayed
E.g.
[...]

PR #1420, opened 2020-04-22
Fix reauth notice when no Analytics account

aaemnnosttv:
## Summary

Addresses issue #1368

## Relevant technical choices

* Updated datapoint to return successful response with empty arrays for each if no accounts
* No changes necessary in current Analytics `setup.js` component

image

image

image

## Checklist

### DIFF
--- a/includes/Modules/Analytics.php
+++ b/includes/Modules/Analytics.php
@@ -519,9 +519,9 @@ final class Analytics extends Module
 					return true;
 				};
 			case 'GET:accounts-properties-profiles':
-				$restore = $this->with_client_defer( false );
+				return function () use ( $data ) {
+					$restore_defer = $this->with_client_defer( false );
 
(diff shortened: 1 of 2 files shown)


includes/Modules/Analytics.php
```diff
@@ -518,7 +519,24 @@ protected function create_data_request( Data_Request $data ) {
 					return true;
 				};
 			case 'GET:accounts-properties-profiles':
-				return $this->get_service( 'analytics' )->management_accounts->listManagementAccounts();
+				$restore = $this->with_client_defer( false );
```
felixarntz, line 522: Let's move this into the anonymous function below. It should only change the client defer mode right before the request, here that _may_ not be the case depending on how the surrounding code deals with the returned function.

Also, I think it would be better naming the variable something more precise (we're using `$restore_defer` in other cases).

Issue #1122, opened 2020-02-11
Check to see if E2E environment is running when running test:e2e command


tofumatt:
## Feature Description

The E2E test suite requires a set of Docker containers (started with `npm run env:start`). This requirement makes sense and most of the time doesn't create issues, but is a bit new-contributor-hostile and is also an easy thing to mess up after a Docker restart or system restart. If you restart your machine, update Docker, or otherwise forget to start the E2E test environment, you'll run the test the first time only to see:

```
 ● Test suite failed to run
    Error: net::ERR_CONNECTION_REFUSED at http://localhost:9002/wp-admin/plugins.php
```

This is easily avoidable by having the E2E test script first try to connect to http://localhost:9002 (via a modified script that makes a `curl` request to http://localhost:9002, and if it can't connect running `npm run env:start` first.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Running `npm run test:e2e` on a machine that has not run the E2E Docker environment (started by `npm run env:start`) does not fail with a "connection refused" error.

## Implementation Brief

* Add a script (or npm `pre` script) to `npm run test:e2e` that tries to `curl` http://localhost:9002. If there is a response, run `return 0` right away; if not, run `npm run env:start` then `return 0`.

## QA Brief

* On a new checkout of the repo, run: `composer install && npm install && npm run build`
* Run `npm run env:stop` to ensure env is stopped since it is global
* Run `npm run test:e2e`

The E2E tests should run as if the environment was running; no error should be output and the script should exit normally.

## Changelog entry

*

tofumatt:
(I moved this one along into IB review as it is extremely straightforward to write and largely a developer quality-of-life issue. If that's a bad thing let me know and we can push it back. This is also not high-priority, but came to mind after biting me and I figured I should just file 😄)

aaemnnosttv:
The e2e environment has a `restart: always` policy so it should restart automatically once started even after rebooting. Are you sure this isn't the case?

I agree that the experience could be better for a first-time run, where it may not be obvious that a separate command is necessary to start the environment before the test command.

tofumatt:
Ah, I see! I was likely getting confused with my `10updocker` site re: restarting, so this might be more a first-run experience. In that case: more annoying to QA, and possibly much less of an issue—though still something to be (quickly and straightforwardly) improved.

aaemnnosttv:
[...]

PR #1423, opened 2020-04-22
Enhancment/1122 check for env

ryanwelcher:
## Summary

Addresses issue #1122 

## Relevant technical choices

I have created a `pretest:e2e` to check for and start the env. If the script detects that the env is not running, it outputs the message `e2e environment is not running, starting now.`

## Checklist


bin/local-env/env-check.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# Include useful functions
+. "$(dirname "$0")/includes.sh"
+
+# Check for the site status
+if [[ -z $(curl -s --head --fail $npm_package_config_E2E_URL) ]]; then
+	echo -e $(status_message 'e2e environment is not running, starting it now.')
+	npm run env:start
+	exit 0
```
aaemnnosttv, line 10: These `exit` statements are unnecessary, the script will `exit 0` automatically if there is no error.

We should however, add a `set -e` at the top to make sure the script exits with an error if `env:start` fails.

package.json
```diff
@@ -20,7 +23,8 @@
     "test:analyze": "npm run build:production -- --env.analyze=true --json --progress --profile > /tmp/stats.json && cp ./dist/assets/js/*.js /tmp && webpack-bundle-analyzer /tmp/stats.json",
     "test:js": "wp-scripts test-unit-js --config=tests/js/jest.config.js",
     "test:js:watch": "wp-scripts test-unit-js --config=tests/js/jest.config.js --watch",
-    "test:e2e": "WP_BASE_URL=http://localhost:9002 wp-scripts test-e2e --config=tests/e2e/jest.config.js",
+    "pretest:e2e": "./bin/local-env/env-check.sh",
+    "test:e2e": "WP_BASE_URL=$npm_package_config_E2E_URL wp-scripts test-e2e --config=tests/e2e/jest.config.js",
```
aaemnnosttv, line 27: Let's pass the URL using the command option instead of the env variable. It's more clear what the actual command is this way I think:

```suggestion
    "test:e2e": "wp-scripts test-e2e --config=tests/e2e/jest.config.js --wordpress-base-url=$npm_package_config_E2E_URL",
```

bin/local-env/env-check.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# Include useful functions
+. "$(dirname "$0")/includes.sh"
+
+# Check for the site status
+if [[ -z $(curl -s --head --fail $npm_package_config_E2E_URL) ]]; then
+	echo -e $(status_message 'e2e environment is not running, starting it now.')
```
aaemnnosttv, line 8: Let's capitalize E2E here 👍 
Also, we should maybe end with `...` for consistency with other messaging in similar commands.

aaemnnosttv:
Sorry @ryanwelcher - I just realized that the localhost and port are not as flexible as I thought and your initial approach for assuming `localhost` and grabbing the port from the container was the correct one.

It would be great to allow for this to be flexible to allow for configuration in the future but the rest of the code doesn't support it so there's no point in making it configurable yet.

It might be nice if we could extract a function for getting the env URL but I'll leave that for another issue.

Apologies for the back and fourth on this one, the changes I requested above are still valid, but we should go back to what you originally had for building the env URL as localhost + the port from docker-compose, and nix the added config for now.

Issue #1179, opened 2020-02-26
Introduce core/modules store in googlesitekit-modules


felixarntz:
## Feature Description

There should be a `core/modules` store on `googlesitekit.data` that allows to retrieve information about the available modules and their activation status. Some of the store selectors and actions should be wrapped with high-level functions on `window.googlesitekit.modules`. This issue is partly dependent on #1063 and #1178.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `googlesitekit-modules` script (introduced in #1063) should register a `core/modules` store on `window.googlesitekit.data` (via `googlesitekit/data` module).
* The store should follow patterns established in previous stores (see #999, #1175).
* The store should have the following "public" selectors:
    * `getModules()`: Should initially resolve via REST API call to `core/modules/data/list`.
    * `getModule( slug )`: Should probably just call `getModules()` selector and find the respective slug. Alternatively call `core/modules/data/info`, but that may result in unnecessary requests.
    * `isModuleActive( slug )`: Should get that respective boolean from the respective module information (can probably use `getModule()` selector).
* The store should have the following "public" actions:
    * `activateModule( slug )`: Should (if the module isn't already active) issue a `POST` request to `core/modules/data/info` with `active=true`.
    * `deactivateModule( slug )`: Should (if the module isn't already deactivated) issue a `POST` request to `core/modules/data/info` with `active=false` (or similarly without `active` parameter).
* In addition to registering the store on `googlesitekit.data`, the `googlesitekit-modules.js` script should expose the following convenience wrapper functions on `window.googlesitekit.modules` (in addition to the existing `createModuleStore( slug )`):
    * `activateModule( slug )`: Calls `googlesitekit.data.dispatch( 'core/modules' ).activateModule( slug )`
    * `deactivateModule( slug )`: Calls `googlesitekit.data.dispatch( 'core/modules' ).deactivateModule( slug )`

## Implementation Brief

* Use the `assets/js/googlesitekit-modules.js` script to register a `core/modules` store.
* Add selectors to the store:
  * `getModules()` with a resolver that uses data from the `core/modules/data/list` API.
  * `getModule( slug )`: Use `createRegistrySelector` to call `getModules()` above and extract the specific slug's data. It should return `null` if the slug can't be found.
  * `isModuleActive( slug )`: Should get the respective boolean from the respective module information (using the `getModule()` selector above). Should return `null` if the module doesn't exist.
[...]

Issue #1298, opened 2020-03-27
Implement googlesitekit-widgets asset with "core/widgets" datastore


felixarntz:
## Feature Description

There should be a `js/googlesitekit/widgets` module that registers a `core/widgets` store. It should have an entrypoint file `googlesitekit-widgets`. The file should also use the handle `googlesitekit-widgets` in PHP.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Add new `js/googlesitekit/widgets/datastore` module that exports a `core/widgets` datastore and `js/googlesitekit/widgets` that for now just imports that. The module will later do more things so it needs to be there in addition to `js/googlesitekit/widgets/datastore`, although right now that's the only surface point.
* The module should have an accompanying entrypoint file `js/googlesitekit-widgets.js` with a PHP handle of the same name.
* The store should have the following actions:
    * `registerWidgetArea( slug, settings )`
    * `assignWidgetArea( slug, contextSlugs )`
    * `registerWidget( slug, settings )`
    * `assignWidget( slug, widgetAreaSlugs )`
* The store should have the following selectors:
    * `isWidgetAreaRegistered( slug )`
    * `isWidgetRegistered( slug )`
    * `getWidgetAreas( contextSlug )`
    * `getWidgets( widgetAreaSlug )`
* The datastore should fully operate on the client side. So most likely this will only need `googlesitekit-data` as the sole dependency in PHP.
* See doc for more details on requirements for these actions and selectors.

## Implementation Brief

* Create a `js/googlesitekit/widgets/datastore/index.js` that creates a module, similar to other datastores. This will likely be a simple datastore, similar to `core/site`.
* Create a `js/googlesitekit/widgets/index.js` that exposes our Widgets API.
* Create an `assets/js/googlesitekit-widgets.js` file. Import `Widgets` from `assets/js/googlesitekit/widgets` and expose it on `googlesitekit.widgets`.

### Widget actions

* `registerWidgetArea( widgetAreaSlug, settings )`: Create a widget area with the specified slug and settings. This saves the data in the `widgetAreas` section of the store. If the `slug` supplied is already in use, it should dispatch a failed action and log a `console.error` message to alert the developer/user. Registering a widget with an existing slug should not crash the app. Essentially: whoever registers a `slug` first claims it. The `settings` object takes the following attributes:
    * `priority`: numeric priority, like WordPress hook priorities (lower means higher priority, default is 10)
[...]

PR #1425, opened 2020-04-22
Add "core/modules" store

tofumatt:
## Summary

Addresses issue #1179.

## Relevant technical choices

I opted not to skip requests if modules were already activated/deactivated in local state, for two reasons:

1. UI/components should be conditionally allowing dispatches to `activate(slug)`/`deactivate(slug)` based on the existing `isModuleActive(slug)` check.
2. There's no harm in sending the extra request _if_ we fail to check above (as suggested), but the added code complexity didn't really warrant it given it shouldn't happen and nothing should break if it does.

It's pretty easy to add if you decide you'd really like it, but I opted for simplicity and letting the components manage that dispatch 😄

This re-fetches all modules anytime one is updated, as I figured other props might change after activation/deactivation and it made sense to load things again. This was better/less complex code than checking to see if the module was already loaded and then modifying only its `active` property. It also seemed more accurate if any other props changed.

## Checklist


assets/js/googlesitekit/datastore/modules/index.js
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+/**
+ * Internal dependencies
+ */
+import Data from 'googlesitekit-data';
+import { createNotificationsStore } from '../../data/create-notifications-store';
+import modules from './modules';
+
+const notifications = createNotificationsStore( 'core', 'modules', 'notifications' );
```
felixarntz, line 26: Not sure why you added a notifications part to the store here? There are no general notifications for modules, also this wasn't part of the ACs :)

assets/js/googlesitekit/datastore/modules/modules.js
```diff
@@ -0,0 +1,434 @@
...
+
+		try {
+			response = yield {
+				payload: { active, slug },
+				type: FETCH_SET_MODULE_STATUS,
+			};
+
+			if ( response.success === true ) {
+				// Fetch (or re-fetch) all modules, with their updated status.
+				yield actions.fetchModules();
```
felixarntz, line 114: Nice! 👍 

I guess this is mainly to account for dependency modules? Theoretically otherwise we could just modify the one module activated/deactivated in the store directly, but I see making the extra request for fresh data is a more bulletproof approach.

assets/js/googlesitekit/datastore/modules/modules.js
```diff
@@ -0,0 +1,434 @@
...
+	 * @return {Object}         Object with {response, error}
+	 */
+	*setModuleStatus( slug, active ) {
+		invariant( slug, 'slug is required.' );
+		invariant( active !== undefined, 'active is required.' );
+
+		let response, error;
+
+		yield {
+			payload: { active, slug },
```
felixarntz, line 102: Not a big deal, but maybe we should turn these two into a `params` object right away, in preparation for #1288? This would not only be good to account for the meta-programming simplification, it's also generally a good idea to separate parameters passed from the response/error received (more relevant for e.g. `CATCH_FETCH_SET_MODULE_STATUS` below), like discussed the other week.

assets/js/googlesitekit/datastore/modules/modules.js
```diff
@@ -0,0 +1,434 @@
...
+	 * Receive the API respond for module (de)activation.
+	 *
+	 * @since n.e.x.t
+	 * @private
+	 *
+	 * @param  {string}  slug    Name of slug to activate/deactivate.
+	 * @param  {boolean} active  `true` to activate; `false` to deactivate.
+	 * @return {Object} Redux-style action.
+	 */
+	receiveSetModuleStatus( slug, active ) {
```
felixarntz, line 183: Currently unused since the new module data is updated via `fetchModules` instead. So let's remove it then.

assets/js/googlesitekit/datastore/modules/modules.js
```diff
@@ -0,0 +1,434 @@
...
+	 *     "analytics"
+	 *   ],
+	 *   "dependants": []
+	 * }
+	 * ```
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {Object|undefined} Modules available on the site.
```
felixarntz, line 341: Shouldn't this be `?Object`?

assets/js/googlesitekit/datastore/modules/modules.js
```diff
@@ -0,0 +1,434 @@
...
+	/**
+	 * Get the list of modules registered for use with Site Kit.
+	 *
+	 * A module is a section of Site Kit that relates to a particular service,
+	 * like Google Analytics or Google PageSpeed modules. They can provide
+	 * admin-only features (like PageSpeed Insights), frontend-only features,
+	 * or both (eg. Analytics, which can install Analytics <script> tags in the
+	 * frontend, and show dashboards in the WordPress Admin).
+	 *
+	 * Returns a hash of objects with the shape, keyed by slug, when successful:
```
felixarntz, line 321: Is "hash" the right term here? Maybe we should say "map"?

assets/js/googlesitekit/datastore/modules/modules.js
```diff
@@ -0,0 +1,434 @@
...
+
+		// Return `undefined` if modules haven't been loaded yet.
+		if ( modules === undefined ) {
+			return undefined;
+		}
+
+		// A module with this slug couldn't be found; return `null` to signify the
+		// "not found" state.
+		if ( modules[ slug ] === undefined ) {
+			return null;
```
felixarntz, line 372: 👍

assets/js/googlesitekit/datastore/modules/modules.js
```diff
@@ -0,0 +1,434 @@
...
+	 * Get a specific module by slug.
+	 *
+	 * Returns a specific module by its slug.
+	 * Returns `undefined` if state is still loading or if said module doesn't exist.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @param {string} slug  Module slug.
+	 * @return {Object|undefined} A specific module object.
```
felixarntz, line 359: See above, should probably be `?Object`. Also I think it would be good to clarify in the description "..., or null if invalid slug provided."

assets/js/googlesitekit/datastore/modules/modules.js
```diff
@@ -0,0 +1,434 @@
...
+
+		// Return `undefined` if modules haven't been loaded yet.
+		if ( module === undefined ) {
+			return undefined;
+		}
+
+		// A module with this slug couldn't be found; return `null` to signify the
+		// "not found" state.
+		if ( module === null ) {
+			return null;
```
felixarntz, line 403: I guess that makes sense here - I'm _a bit_ questioning whether we should just return `false`, to simplify the signature here. Logically, if a module doesn't exist, it's also not active, so returning `false` could still be appropriate.

assets/js/googlesitekit/datastore/modules/modules.js
```diff
@@ -0,0 +1,434 @@
...
+	 *
+	 * Returns `true` if the module exists and is active.
+	 * Returns `false` if the module exists but is not active.
+	 * Returns `undefined` if state is still loading or if no module with that slug exists.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @param {string} slug  Module slug.
+	 * @return {Object|undefined} A specific module object.
```
felixarntz, line 390: `?boolean` - also description should be fixed.

assets/js/googlesitekit/datastore/modules/modules.js
```diff
@@ -0,0 +1,434 @@
...
+	 *
+	 * Returns `true` if the module exists and is changing its `active` flag.
+	 * Returns `false` if the module exists but is not changing its `active` flag.
+	 * Returns `undefined` if state is still loading or if no module with that slug exists.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @param {string} slug  Module slug.
+	 * @return {boolean|undefined} Activation change status.
```
felixarntz, line 420: --> `?boolean`

assets/js/googlesitekit/datastore/modules/modules.js
```diff
@@ -0,0 +1,434 @@
...
+	 */
+	isChangingModuleActivation: ( state, slug ) => {
```
felixarntz, line 422: Maybe `isSettingModuleStatus`? This would be more in line with the other names used throughout the datastore. Both names are not that great I think, but since this is also a very specific selector, I don't have a strong opinion.

assets/js/googlesitekit/datastore/modules/modules.js
```diff
@@ -0,0 +1,434 @@
+/**
+ * core/site data store: connection info.
```
felixarntz, line 2: Needs to be updated.

assets/js/googlesitekit/datastore/modules/modules.test.js
```diff
@@ -0,0 +1,607 @@
+/**
+ * core/site data store: modules info tests.
```
felixarntz, line 2: Same here.

assets/js/googlesitekit/modules/index.js
```diff
@@ -21,10 +21,20 @@
 /**
  * Internal dependencies
  */
+import Data from 'googlesitekit-data';
 import { createModuleStore } from './create-module-store';
+// This import has a side-effect: it automatically registers the "core/modules"
+// store on `googlesitekit.data`.
+import { STORE_NAME } from '../datastore/modules';
```
felixarntz, line 28: In #1298 we decided to have the widgets datastore in `assets/js/googlesitekit/widgets/datastore` (instead of `assets/js/googlesitekit/datastore/widgets`) - here it is the other way around. We should decide on one approach and stick with it. From a module standpoint the first makes more sense - it clarifies that the datastore is part of the `widgets` (here `modules`) module. The second approach makes more sense from a "let's keep all datastores together" standpoint - however we'll not be doing this for module stores at least anyway, so I'm still not fully convinced. Let's discuss further as needed.

assets/js/googlesitekit/datastore/modules/index.js
tofumatt: Yeah, I saw that, but I assumed we might want notifications for them, but I guess they'd be scoped to the modules themselves? I figured… maybe registering a module failed, etc. But I'll remove it for now—we can add it later if needed.

assets/js/googlesitekit/datastore/modules/modules.js
tofumatt: Yeah, I was thinking anything that's got dependencies/interconnected-ness would benefit from this. We can already use this, but third-party modules might especially benefit from this.

assets/js/googlesitekit/datastore/modules/modules.js
tofumatt: `state.modules` has the potential to be `undefined`, so this is correct (though as mentioned in #1468, it should be wrapped in parens).

assets/js/googlesitekit/datastore/modules/modules.js
tofumatt: I suppose so, but I like clarifying the reason for the value here. Code that doesn't care if it exists or not can always check for a truthy value, but this exposes a bit more context for those components that might care about _why_ it's returning false-y without needing other selectors.

assets/js/googlesitekit/datastore/modules/modules.test.js
tofumatt: Copypasta strikes again! 😅

assets/js/googlesitekit/modules/index.js
tofumatt: Fair enough. I'll move these into `assets/js/googlesitekit/modules/datastore` 👍

tofumatt:
This took awhile to loop back to—thanks for your patience on it! I've addressed all the review comments; this should be ready for another review 🙂

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -0,0 +1,415 @@
...
+	 * Returns `false` if the module exists but is not changing its `active` flag.
+	 * Returns `undefined` if state is still loading or if no module with that slug exists.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @param {string} slug  Module slug.
+	 * @return {(boolean|undefined)} Activation change status; `undefined` if state is still loading or if no module with that slug exists.
+	 */
+	isSettingModuleActivation: ( state, slug ) => {
```
felixarntz, line 403: See previous comment where I suggested to name this `isSettingModuleStatus`: you changed the "changing" to "setting", but still use "activation". I like that better than "status" indeed, but let's change it elsewhere then too.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -0,0 +1,415 @@
...
+	 * supplied for the give `slug`.
+	 *
+	 * @since n.e.x.t
+	 * @private
+	 *
+	 * @param  {string}  slug   Slug of the module to activate/deactivate.
+	 * @param  {boolean} active `true` to activate; `false` to deactivate.
+	 * @return {Object}         Object with {response, error}
+	 */
+	*setModuleStatus( slug, active ) {
```
felixarntz, line 94: Let's make this `setModuleActivation`.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -0,0 +1,415 @@
...
+import Data from 'googlesitekit-data';
+import { STORE_NAME } from './constants';
+
+const { createRegistrySelector } = Data;
+
+// Actions
+const START_FETCH_SET_MODULE_STATUS = 'START_FETCH_SET_MODULE_STATUS';
+const FETCH_SET_MODULE_STATUS = 'FETCH_SET_MODULE_STATUS';
+const FINISH_FETCH_SET_MODULE_STATUS = 'FINISH_FETCH_SET_MODULE_STATUS';
+const CATCH_FETCH_SET_MODULE_STATUS = 'CATCH_FETCH_SET_MODULE_STATUS';
```
felixarntz, line 37: Let's rename these to `*_MODULE_ACTIVATION`.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -0,0 +1,415 @@
...
+const START_FETCH_MODULES = 'START_FETCH_MODULES';
+const FETCH_MODULES = 'FETCH_MODULES';
+const FINISH_FETCH_MODULES = 'FINISH_FETCH_MODULES';
+const CATCH_FETCH_MODULES = 'CATCH_FETCH_MODULES';
+const RECEIVE_MODULES = 'RECEIVE_MODULES';
+
+export const INITIAL_STATE = {
+	modules: undefined,
+	isFetchingSetModuleStatus: {},
```
felixarntz, line 46: --> `isFetchingSetModuleActivation`

assets/js/googlesitekit/modules/index.js
```diff
@@ -21,10 +21,20 @@
 /**
  * Internal dependencies
  */
+import Data from 'googlesitekit-data';
 import { createModuleStore } from './create-module-store';
+// This import has a side-effect: it automatically registers the "core/modules"
+// store on `googlesitekit.data`.
+import { STORE_NAME } from '../modules/datastore';
```
felixarntz, line 28: This can become simply `./datastore` after the rename.

assets/js/googlesitekit/modules/datastore/modules.js
tofumatt: Whoops, you're right! My bad; I'll update it.

Issue #1402, opened 2020-04-17
Require only propertyID for Analytics datastore profile functions


felixarntz:
Since the Analytics account ID can be derived from the Analytics property ID, functions in the `modules/analytics` datastore that require both should be modified to only require the `propertyID` get parse the account ID from there.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Public-facing functions in the `modules/analytics` datastore that require both should be modified to only require the `propertyID` get parse the account ID from there.
* The respective response data should be keyed only by the `propertyID`.
* The `accountID` should be parsed from `propertyID` to make the API request.

#### `assets/js/modules/analytics/datastore/profiles.js`
Update the following functions to no longer accept an `accountID` and to parse it from the `propertyID` instead
* `selectors.getProfiles`
    * `resolvers.getProfiles`
    * Update `RECEIVE_PROFILES` to groupBy `webPropertyId` only
* `selectors.isDoingCreateProfile`
* `selectors.isDoingGetProfiles`
* `actions.createProfile`
* `actions.fetchProfiles`

* Update tests as neccessary

## QA Brief

* 

## Changelog entry

* N/A

felixarntz:
IB ✅ 

@aaemnnosttv What were your thoughts for at which point to parse the account ID? #1405 implements pretty much the same thing for AdSense, so I was curious whether you were thinking about doing it the same way or had another idea.

aaemnnosttv:
@felixarntz in some cases (like selectors) we won't need to parse it anymore since the data won't be keyed by that anymore.

I think the general idea is to push it as close to the `control` as possible since our API is the only thing that requires it. Maybe it's best to do it inside the control itself to minimize the amount of parsing needed since requiring them separately there is somewhat of an implementation detail. There may be a few cases where it's required outside of that, but that will probably be an exception if at all.

felixarntz:
@aaemnnosttv Ah okay, that's what I was thinking (and did in #1405) too, so SGTM

ShahAaron:
Confirmed code is updated in the following file (assets/js/modules/analytics/datastore/profiles.js) correctly:

- selectors.getProfiles
- resolvers.getProfiles

- - Update RECEIVE_PROFILES to groupBy webPropertyId only

- selectors.isDoingCreateProfile
- selectors.isDoingGetProfiles
- actions.createProfile
- actions.fetchProfiles

PR #1431, opened 2020-04-23
Remove accountID from profiles datastore API

aaemnnosttv:
## Summary

Addresses issue #1402

## Checklist

### DIFF
--- a/assets/js/modules/analytics/datastore/profiles.js
+++ b/assets/js/modules/analytics/datastore/profiles.js
@@ -281,14 +281,14 @@ export const reducer = ( state, { type, payload } ) => {
 		}
 
 		case RECEIVE_CREATE_PROFILE: {
-			const { accountID, propertyID, profile } = payload;
+			const { propertyID, profile } = payload;
 
 			return {
 				...state,
 				profiles: {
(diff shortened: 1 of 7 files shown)


assets/js/modules/analytics/datastore/profiles.js
```diff
@@ -188,13 +173,17 @@ export const actions = {
 };
 
 export const controls = {
-	[ FETCH_CREATE_PROFILE ]: ( { payload: { accountID, propertyID } } ) => {
+	[ FETCH_CREATE_PROFILE ]: ( { payload: { propertyID } } ) => {
+		const { accountID } = parsePropertyID( propertyID );
```
felixarntz, line 177: This _could_ theoretically fail. You're catching a possible invalid `propertyID` before in `createProfile` and `fetchProfiles`, so it shouldn't be an issue. I did it slightly differently for #1403, returning an API-like error from the control. I think that approach looks less elegant, but still works seamlessly and I'd say it is a tiny bit more bulletproof (e.g. if `parsePropertyID` changed to not only be dependent on `isValidPropertyID`). But I'm okay with the approach you took here, so let's proceed.

Issue #1432, opened 2020-04-23
PHP Tests only run against latest WP


aaemnnosttv:
## Bug Description

Currently, PHP tests only run against the latest version of WP (and trunk/bleeding/nightly). This leaves us vulnerable to breaking something on an older version of WP, only to catch it with our E2E tests which do test with WP 4.7. This has happened before and isn't immediately obvious that the failure is WP version specific. 

We should update one job to be testing against the _oldest_ supported version - it makes sense to do this with the same job as we use for the minimum required PHP.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The Travis job for running PHP tests with the minimum supported version of PHP (5.6) should be updated to use the minimum supported version of WordPress (4.7)

## Implementation Brief

* Update this job definition to use WP 4.7  
https://github.com/google/site-kit-wp/blob/7f806f6719b3a201240dd7995ce4fd5baad13e15/.travis.yml#L104-L106

## QA Brief

* The PHP tests job that runs with version 5.6 should be configured to use `WP_VERSION=4.7` and pass

## Changelog entry

*

felixarntz:
Good catch @aaemnnosttv!

IB ✅

cole10up:
**Tested**

Kicked off an E2E test targeting 4.7

![image](https://user-images.githubusercontent.com/57115907/80678279-7a88a580-8a6f-11ea-871d-78682fadcf53.png)
![image](https://user-images.githubusercontent.com/57115907/80678305-83797700-8a6f-11ea-9afc-d23ecec9fb85.png)

Kicked off a test targeting latest WP

![image](https://user-images.githubusercontent.com/57115907/80678498-da7f4c00-8a6f-11ea-99a9-0b78d766d6c2.png)

Passed QA ✅

aaemnnosttv:
@cole10up this one is about PHP (PHPUnit) tests, not E2E tests 🙂 
See the second job here:
https://github.com/google/site-kit-wp/runs/626569804

PR #1438, opened 2020-04-24
Updated WP version to 4.7 for unit testing

ShahAaron:
## Summary

Addresses issue #1432 

## Checklist


aaemnnosttv:
Note the failure is due to the use of `http_api_debug` which wasn't invoked for blocked requests [until 5.3](https://github.com/WordPress/WordPress/commit/eeba1c1244ee17424c8953dc416527a97560f6cc), so this is why the PHP test job is currently failing (because we block all requests during tests).

The one test needs to be updated to use a different hook for spying on the request, then this should be good to go 😄

tests/phpunit/integration/Core/Authentication/Google_ProxyTest.php
```diff
@@ -54,19 +54,27 @@ function () {
 			}
 		);
 
-		$spy = new MethodSpy();
-		add_action( 'http_api_debug', array( $spy, 'callback' ), 10, 5 );
+		$pre_args = null;
+		$pre_url  = null;
+
+		// Use pre_http_request to support WP < 5.3
```
aaemnnosttv, line 60: Please elaborate a little more in this comment as to why this is necessary.
Specifically because `http_api_debug` is not fired for blocked requests before WP 5.3 😄

ShahAaron: done

Issue #1116, opened 2020-02-11
URLs displayed with duplicate path segment (for language/WPML, but also generally)


ernee:
## Bug Description
**Modified for UI issue only** 

URLs are displayed with a duplicate path when using the [WPML plugin](https://wpml.org/) _different languages in directories_ setting with the _use directory for default language_ sub-option checked: 

<img width="1100" alt="WPML different languages in directories" src="https://user-images.githubusercontent.com/8108643/78640051-eb1f1680-7863-11ea-93aa-7b159a21f863.png">

Site Kit chooses the default language URL as the Site URL:

<img width="1054" alt="Screen Shot 2020-04-10 at 11 09 40 AM" src="https://user-images.githubusercontent.com/8108643/79472863-34383e80-7fb9-11ea-9de4-b20c631ac713.png">

URLs reported in the UI append the directory additionally:

<img width="519" alt="Screen Shot 2020-04-13 at 6 36 05 PM" src="https://user-images.githubusercontent.com/8108643/79473003-6184ec80-7fb9-11ea-880a-b8532afffa7a.png">

Further context about the setting:

https://wpml.org/faq/server-setting-for-languages-in-directories/
https://wpml.org/documentation/getting-started-guide/language-setup/cannot-activate-language-directories/

Related support topic:

https://wordpress.org/support/topic/doesnt-handle-multiple-languages-correctly/

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Site Kit should be fully compatible with WPML.
* In particular: with the WPML settings: "different languages in directories" and "use directory for default language sub-option" checked, and the SiteURL set to `{domain}/{default_language}`, the URLS displayed in the "Top content over the last 28 days" table on the WordPress Dashboard, and for the "Most popular content" table on the Site Kit Dashboard should appear correctly.

## Implementation Brief

* When constructing the url to display for top a pages use only the `url` (path) data returned from the API (don't prepend the site URL):
  -  In **WPAnalyticsDashboardWidgetTopPagesTable** [`assets/js/modules/analytics/wp-dashboard/wp-dashboard-widget-top-pages-table.js`](https://github.com/google/site-kit-wp/blob/d0c0d9dc0f89e8516d17eb8b78d600a1780130ac/assets/js/modules/analytics/wp-dashboard/wp-dashboard-widget-top-pages-table.js#L52) and
[...]

PR #1449, opened 2020-04-27
Removed siteUrl for WPML compliance

ShahAaron:
## Summary

Addresses issue #1116 

## Checklist


assets/js/modules/analytics/dashboard/dashboard-widget-popular-pages-table.js
```diff
@@ -80,7 +79,7 @@ class AnalyticsDashboardWidgetPopularPagesTable extends Component {
 		const links = [];
 		const dataMapped = map( data[ 0 ].data.rows, ( row, i ) => {
 			const [ title, url ] = row.dimensions;
-			links[ i ] = siteURL + url;
+			links[ i ] = url;
```
felixarntz, line 82: Let's prefix these with a slash if they don't start with one. Looks like the `url` values returned don't include that, which looks a bit strange and is also different from how the Analytics frontend renders these paths. Starting in a slash clarifies that this is a path, and also accounts for the case of the home page, where right now this would end up an empty string, when it should rather be just the slash.

assets/js/modules/analytics/wp-dashboard/wp-dashboard-widget-top-pages-table.js
```diff
@@ -49,7 +48,7 @@ class WPAnalyticsDashboardWidgetTopPagesTable extends Component {
 		const links = [];
 		const dataMapped = map( data[ 0 ].data.rows, ( row, i ) => {
 			const [ title, url ] = row.dimensions;
-			links[ i ] = siteURL + url;
+			links[ i ] = url;
```
felixarntz, line 51: See above.

assets/js/modules/analytics/dashboard/dashboard-widget-popular-pages-table.js
ShahAaron: Done :)

assets/js/modules/analytics/wp-dashboard/wp-dashboard-widget-top-pages-table.js
ShahAaron: Completed

ShahAaron:
@felixarntz I've update getDataTableFromData so it does not trim links. Also updated the Visual Regression tests

assets/js/components/data-table.js
```diff
@@ -71,14 +70,14 @@ export const getDataTableFromData = ( data, headers, options ) => {
 								{ cell }
 							</Link>
 
-							{ showURLs && '' !== trim( link, '/' ) &&
```
felixarntz, line 74: The check here itself shouldn't be removed, only the `trim` part of it. This should probably become `showURLs && link`.

ShahAaron: @felixarntz I removed it as I felt it was covered by the ternary operator wrapper from line 62

felixarntz: @ShahAaron Oh you're right - sorry about that, all good then.

felixarntz:
Not sure what's happening with Travis-CI, but the build actually went through and succeeded.

Issue #1434, opened 2020-04-24
Dashboard search input not validated


aaemnnosttv:
## Bug Description

![image](https://user-images.githubusercontent.com/1621608/80126810-262b8480-859c-11ea-9f58-81bb385216d1.png)

Currently the search input on the Site Kit dashboard does not validate the input at all.

It allows the form to be submitted, even with nothing input which raises an error in the console (with no feedback).

## Steps to reproduce

1. Go to Site Kit dashboard
1. Click on "View Data" with nothing in the input
1. See error in console with no feedback given

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The dashboard search should only be submittable with valid input

## Implementation Brief
* In the `PostSearcher` component, modify the `onClick` to ensure that `match` is defined.
* Introduce `state.canSubmit` with a default value of `false` and set the `disabled` prop to be the opposite of it's value.
* Set `state.canSubmit` to be true when both `state.selection` is set and there is a match.
* If defined, set the `document.location` normally.
* Update to use `global.location` as we use `global` for all globals instead of `window` or other implied globals like `document`

Merge: #1452

## QA Brief

* 

## Changelog entry

* Ensure dashboard search form can only be submitted with valid content.

aaemnnosttv:
@ryanwelcher this is a good start - a few thoughts for you:

* Ideally we would also set the `disabled` state on the _View Data_ button based on the validity of the input
    - To do this, would need to have the value of the input in state (i.e. `Autocomplete` would be a "controlled" component  - which doesn't seem to be possible)
   - Let's disable the button based on the presence of `state.selection` since this is required to match with in the `onClick`
* Regarding the `onClick` - as long as there is a `selection` there should always be a `match` since all options are simply a list of posts by title - we should of course prevent the error if there is somehow not a `match` but I think adding additional messages is probably not necessary - let me know if this is not the case though.
* Please update this to use `global.location` as we use `global` for all globals instead of `window` or other implied globals like `document`

Please update your IB and I think this should be good to go 👍

ryanwelcher:
> * Ideally we would also set the `disabled` state on the _View Data_ button based on the validity of the input
>   * To do this, would need to have the value of the input in state (i.e. `Autocomplete` would be a "controlled" component  - which doesn't seem to be possible)
>   * Let's disable the button based on the presence of `state.selection` since this is required to match within the `onClick`

[...]

PR #1452, opened 2020-04-27
Validate Dashboard Search Input

ryanwelcher:
## Summary

Addresses issue #1434

## Checklist


assets/js/components/post-searcher.js
```diff
@@ -46,12 +47,14 @@ import Layout from './layout/layout';
 class PostSearcher extends Component {
 	constructor( props ) {
 		super( props );
-
+		this.disallowedSelections = [ 'No results found' ];
```
aaemnnosttv, line 50: This is a translated string, so this reference must also be. We can probably just extract the string itself to a property as well so all usage references same value rather than repeating it multiple times in the component.

assets/js/components/post-searcher.js
```diff
@@ -168,11 +185,13 @@ class PostSearcher extends Component {
 										<Button
 											onClick={ this.onClick }
 											className="googlesitekit-post-searcher__button"
+											disabled={ ! this.state.canSubmit }
 										>
 											{ __( 'View Data', 'google-site-kit' ) }
 										</Button>
 									</div>
 								</div>
+								{ 0 < this.state.message.length && <p>{ this.state.message }</p> }
```
aaemnnosttv, line 194: This shouldn't be added as the only visual change which has been approved is the disabled state of the submit button. Originally the issue mentioned adding feedback, but this was removed before execution 🙂

assets/js/components/post-searcher.js
```diff
@@ -112,28 +115,42 @@ class PostSearcher extends Component {
...
 
 	onConfirm( selection ) {
-		this.setState( {
-			selection,
-		} );
+		const { results } = this.state;
+		let match;
+		// Check to that the selection is "valid".
+		if ( ! this.disallowedSelections.includes( selection ) ) {
+			match = find(
```
aaemnnosttv, line 122: Not sure why lodash was used for this before but we should be able to simply use `results.find` instead. We might only want to add a guard at the very top to return early `if ( ! Array.isArray( results ) )`.

assets/js/components/post-searcher.js
```diff
@@ -112,28 +115,42 @@ class PostSearcher extends Component {
...
+				results,
+				( result ) => {
+					return result.post_title === selection;
+				}
```
aaemnnosttv, line 126:
```suggestion
				( result ) => result.post_title === selection
```

assets/js/components/post-searcher.js
```diff
@@ -156,7 +173,7 @@ class PostSearcher extends Component {
 						<div className="mdc-layout-grid__inner">
 							<div className="mdc-layout-grid__cell mdc-layout-grid__cell--span-12">
 								<div className="googlesitekit-post-searcher">
-									<label className="googlesitekit-post-searcher__label" htmlFor="autocomplete">{ __( 'Title or URL', 'google-site-kit' ) }</label>
+									<label className="googlesitekit-post-searcher__label" htmlFor="autocomplete">{ __( 'Title', 'google-site-kit' ) }</label>
```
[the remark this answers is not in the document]
ryanwelcher: updated.

assets/js/components/post-searcher.js
```diff
@@ -112,28 +115,42 @@ class PostSearcher extends Component {
...
-			'googlesitekit-dashboard',
-			{
-				id: match.id,
-				permaLink: match.permalink,
-				pageTitle: selection,
-			}
-		);
+		const { canSubmit, selection, match } = this.state;
+		if ( canSubmit ) {
+			document.location = getSiteKitAdminURL(
```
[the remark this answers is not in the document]
ryanwelcher: Updated.

assets/js/components/post-searcher.js
ryanwelcher: Updated.

assets/js/components/post-searcher.js
ryanwelcher: Updated

assets/js/components/post-searcher.js
ryanwelcher: Updated

assets/js/components/post-searcher.js
ryanwelcher: Updated.

assets/js/components/post-searcher.js
```diff
@@ -112,28 +115,37 @@ class PostSearcher extends Component {
 	}
 
 	onConfirm( selection ) {
-		this.setState( {
-			selection,
-		} );
+		const { results } = this.state;
+		let match;
+		// Check to that the selection is "valid".
```
aaemnnosttv, line 120:
```suggestion
		// Check that the selection is "valid".
```

assets/js/components/post-searcher.js
```diff
@@ -112,28 +115,37 @@ class PostSearcher extends Component {
...
+		if ( ! Array.isArray( results ) || ! this.disallowedSelections.includes( selection ) ) {
```
aaemnnosttv, line 121: This logic is slightly off. `results` must always be an array here because we are going to call `find` on it immediately after this check. We also want to make sure that the selection is not in the selections that are not allowed (i.e. `&&`).

I like the idea of `disallowedSelections`, but since there's only one, and we also have that as a property, I think we can do without it until we have more while still preserving the same readability as it's logical that you shouldn't be able to select the `noResultsMessage`.

assets/js/components/post-searcher.js
```diff
@@ -112,28 +115,37 @@ class PostSearcher extends Component {
...
+			match = results.find( ( result ) => result.post_title === selection );
+			if ( selection && ! isUndefined( match ) ) {
```
aaemnnosttv, line 123: This is a bit of a nitpick, but let's not use lodash for this. We've agreed to use a simple strict comparison with `undefined` so that's what we should use going forward for consistency. Sorry for not mentioning this before, but lodash does the same thing internally so it doesn't offer much benefit, other than semantics which are not bad, so this is just from a consistency standpoint.

In this case specifically, I think we can simply check `match` as a basic truthy check as it will either be a post `Object`, or `undefined` 🙂 
```suggestion
			if ( selection && match ) {
```

assets/js/components/post-searcher.js
```diff
@@ -112,28 +115,37 @@ class PostSearcher extends Component {
...
-		document.location = getSiteKitAdminURL(
-			'googlesitekit-dashboard',
-			{
-				id: match.id,
-				permaLink: match.permalink,
-				pageTitle: selection,
-			}
-		);
+		const { canSubmit, selection, match } = this.state;
+		if ( canSubmit ) {
```
aaemnnosttv, line 139: This conditional is unnecessary as this event handler won't be invoked if the button clicked when disabled.
We don't usually double-guard handlers like this when it's already covered where it's invoked.

Instead I think it would be a good idea to check that there is a valid `match` instead 👍 
```suggestion
		if ( match && match.id ) {
```

assets/js/components/post-searcher.js
```diff
@@ -112,28 +115,37 @@ class PostSearcher extends Component {
 	}
 
 	onConfirm( selection ) {
-		this.setState( {
-			selection,
-		} );
+		const { results } = this.state;
+		let match;
```
aaemnnosttv, line 119: This outer declaration is unnecessary as it's only set once. This can simply be `const match = results.find` below.

assets/js/components/post-searcher.js
```diff
@@ -21,9 +21,9 @@
  */
 import {
 	map,
-	find,
 	debounce,
 	trim,
+	isUndefined,
```
aaemnnosttv, line 26: See below
```suggestion
```

assets/js/components/post-searcher.js
```diff
@@ -112,28 +113,36 @@ class PostSearcher extends Component {
...
-		document.location = getSiteKitAdminURL(
-			'googlesitekit-dashboard',
-			{
-				id: match.id,
-				permaLink: match.permalink,
-				pageTitle: selection,
-			}
-		);
+		const { selection, match } = this.state;
+		if ( match && match.ID ) {
```
ryanwelcher, line 136: Just wanted to note here that I needed to change this from `id` to `ID`. It looks like in the released plugin the id is not being added to the URL.

assets/js/components/post-searcher.js
ryanwelcher: Updated

assets/js/components/post-searcher.js
ryanwelcher: Updated.

assets/js/components/post-searcher.js
ryanwelcher: Updated.

assets/js/components/post-searcher.js
ryanwelcher: Updated.

assets/js/components/post-searcher.js
aaemnnosttv: Good catch! You're right, it's returning full `WP_Post`s, not the normal post object you would expect from a core REST controller. It looks like the ID isn't currently used, but maybe I'm missing it. Either way 👍

Issue #1269, opened 2020-03-19
Finalize Analytics setup/settings refactoring with datastore


felixarntz:
This is a tracking task, to merge the branch [`refactor/analytics-data-store`](https://github.com/google/site-kit-wp/tree/refactor/analytics-data-store) to `develop` once all related refactoring has been completed.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* N/A

## Implementation Brief

* N/A

## QA Brief

* 

## Changelog entry

* N/A

cole10up:
Positive tests setting up Analytics using a new property, existing property, disconnecting and reconnecting all passed.

Passed QA ✅

PR #1454, opened 2020-04-27
Analytics datastore refactoring

aaemnnosttv:
## Summary

Addresses issue #1269

## Checklist


aaemnnosttv:
@felixarntz @tofumatt the only thing missing here seems to be the CLA consent from you two so I'm going to move this to review.

tofumatt:
@googlebot I consent.

stories/module-analytics-settings.stories.js
```diff
@@ -0,0 +1,210 @@
...
+					isSaving={ isSaving }
+					screenID={ module.screenID }
+					error={ error }
+				/>
+			</div>
+		</WithTestRegistry>
+	);
+}
+
+storiesOf( 'Analytics Module Settings', module )
```
felixarntz, line 98: Let's change this to `Analytics Module/Settings`. This will make `Settings` be a separate "sub-story" of `Analytics Module`, which makes sense here.

stories/module-analytics-setup.stories.js
```diff
@@ -0,0 +1,176 @@
...
+
+function Setup( props ) {
+	return (
+		<WithTestRegistry { ...props }>
+			<SetupWrapper />
+		</WithTestRegistry>
+	);
+}
+
+storiesOf( 'Analytics Module Setup', module )
```
felixarntz, line 60: Same here, let's use `Analytics Module/Setup`.

stories/module-analytics-settings.stories.js
aaemnnosttv: I didn't realize stories could be organized like this. Awesome! 👍

aaemnnosttv:
@felixarntz I made a few updates here that were missed before. Mostly storybook updates, but a few tweaks to components, etc:

* Added the same `useEffect` we use in `SettingsEdit` to `SetupMain` for enforcing the existing tag values. This didn't get updated when we changed from doing this in the datastore to in the component (we only updated `SettingsEdit`).
* I moved the `ErrorNotice` component from both "main" components and into the inner view components. The reason for this is that a few views (loading/progress, and `ExistingTagError`) we don't want to render the error (particularly in the `ExistingTagError` case as it would show two errors on the screen). This was a fairly simple change.
* One update to a tag permission selector which was still using both property and account IDs but also because we have a dedicated selector for that.
* Updated the `INITIAL_STATE.siteInfo` to be `undefined` (rather than an object of `undefined` values for consistency, but also for simplifying the check in the resolver)
* Added the early return in `resolvers.getSiteInfo` if `siteInfo` was already set. This was causing errors in one of the component tests even though the data was being received in the test setup, because the resolver was not returning early like we do in others. This is done in a single commit so we can cherry pick it for another PR, etc if we need it in develop sooner than this PR will be (although I hope that's not the case 😄 )

Storybook fixes were things like removing old actions like `applyProperty` that didn't exist anymore, updating `receiveExistingTag` to get the string propertyID and a few other minor changes, but these should all be working now.

aaemnnosttv:
Regarding that last commit, I had to make one more change to the `core/site` store related to the `siteInfo` change from before where it was always returning an object which is inconsistent with our other selectors. I updated it to return `undefined` if the data isn't loaded yet and updated the sub-selectors to use `select(...) || {}` to prevent errors while preserving the behavior there. I talked to @tofumatt about this before making this change since he did the original datastore work there. The intention was just to prevent the specific-selectors from failing. Tests are all good now! ✅

Issue #1400, opened 2020-04-17
Create a "combineStores" utility function to combine stores


tofumatt:
## Feature Description

Currently, we create "combined" stores like this: https://github.com/google/site-kit-wp/blob/507e81f4af2b52e77db6a71710777d5ea086302b/assets/js/googlesitekit/modules/create-module-store.js#L59

But that's a bit verbose and makes combining multiple stores together a bit annoying. We should create a utility that combines stores—even ones that lack certain keys.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Create a `combineStores( …args )` function that accepts a list of stores and combines their actions, reducers, etc. for multiple stores and returns a single store. It should not require that stores have all store properties (eg. `initialState`, `actions`, etc.) and should use sane defaults for any missing values.
* It should still fail if duplicate keys are found, similar to the `collectActions`, etc. functions.

## Implementation Brief

* Add a `combineStores( …args )` function that accepts full stores as arguments, each with the object shape:

```js
{
  INITIAL_STATE,
  actions,
  controls,
  reducer,
  resolvers,
  selectors,
}
```

Note that _every value should be optional_, because it's possible to have a store that doesn't have `controls`, or even `selectors`, etc. We can provide sane defaults for other things like `reducer` or `initialState` if none are provided. (Though equally, I'd be fine if `initialState` was also required.)

* This should act like an amalgamation of this type of code: https://github.com/google/site-kit-wp/blob/c302d8cb77374e394ba4711a93b3d7ca33e43802/assets/js/googlesitekit/datastore/site/index.js#L38-L85, but allow for developers to simply submit entire stores. So the above code would change to:

```js
const store = Data.combineStores(
	{
		actions: Data.commonActions,
		controls: Data.commonControls,
	},
	connection,
	info,
	reset,
	notifications,
);

Data.registerStore( STORE_NAME, store );
```

## QA Brief

* 

## Changelog entry

* Add `googlesitekit.data.combineStores` utility function to combine multiple datastore objects.

felixarntz:
IB ✅ 

Let's have all parts of the object optional including `INITIAL_STATE`.

ShahAaron:
Tested

PR code
Run e2e tests - passed
Run JS Unit tests - passed
Run Visual regression tests - passed

Passed QA ✅

felixarntz:
[...]

PR #1455, opened 2020-04-27
Add combineStores() utility function and tests.

jqlee85:
## Summary

Addresses issue [#1400](https://github.com/google/site-kit-wp/issues/1400)

## Relevant technical choices

Adds a `combineStores()` function to `assets/js/googlesitekit/data/utils.js` which takes a list of stores as arguments and returns a combined store object. Relevant unit tests added to `assets/js/googlesitekit/data/utils.test.js`.

## Checklist


assets/js/googlesitekit/data/utils.test.js
```diff
@@ -124,6 +128,231 @@ describe( 'data utils', () => {
 		} );
 	} );
 
+	describe( 'combineStores()', () => {
+		const combineStoresFakeActionOne = {
```
aaemnnosttv, line 132: I think the `combinedStores*` and `fake` prefixes are unnecessary here.

Also, generally having so many things defined outside can make it difficult to read the tests themselves because you need to switch back and fourth to see what is what. If possible, it would be great if any of these could be moved inline to the test that uses it. Having concise, clear names helps a lot here too 👍

assets/js/googlesitekit/data/utils.js
```diff
@@ -212,6 +212,55 @@ export const collectName = ( ...args ) => {
...
+/**
+ * Combines multiple stores.
+ *
+ * @since n.e.x.t
+ *
+ * @param {...Object} args A list of objects, each a store containing one or more of the following keys: INITIAL_STATE, actions, controls, reducer, resolvers, selectors
+ * @return {Object} The combined store.
+ */
+export const combineStores = ( ...args ) => {
+	const combined = {};
```
aaemnnosttv, line 224: Minor change here but I think it would be a little cleaner to return the new object here rather than build a variable
```suggestion
	const combined = {};
	return {
		INITIAL_STATE: collectState( /*  */ ),
		actions: collectActions( /*  */ ),
		controls: collectControls( /*  */ ),
		selectors: collectSelectors( /*  */ ),
		resolvers: collectResolvers( /*  */ ),
		reducer: collectReducers( /*  */ ),
	};
```
You can keep the comments for each if you'd like (I think they're a bit more relevant in this context).

assets/js/googlesitekit/data/utils.js
```diff
@@ -212,6 +212,55 @@ export const collectName = ( ...args ) => {
...
+
+	const defaultReducer = ( state, action ) => { // eslint-disable-line no-unused-vars
+		return { ...state };
+	};
```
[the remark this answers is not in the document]
tofumatt: We shouldn't be mutating the state object ever, but returning a new one using a spread operator is a handy way to nip the potential of that "working" somewhere in the bud. That said, we do `return state` in `assets/js/modules/analytics/datastore/settings.js` from the looks of it, so 🤷‍♂️

Can remove the `action` arg to ditch the ESLint warning though; no need for that to be there for a "default" reducer.

assets/js/googlesitekit/data/utils.js
tofumatt: Yeah, reading through this it'd be a lot neater to nix the `combined = {};` variable and return the newly-built-in-place object.

assets/js/googlesitekit/data/utils.test.js
```diff
@@ -124,6 +128,231 @@ describe( 'data utils', () => {
...
+			INITIAL_STATE: { three: 3 },
+		};
+
+		const storeFour = {
+			reducer: combineStoresFakeReducerFour,
+		};
+
+		const storeFive = {
+			actions: {
+				fakeActionOne() {
```
tofumatt, line 237: Looks like this action is defined above but rewritten here. Seems like we could consolidate the number of things defined here in addition to @aaemnnosttv's comment about colocating the variables closer to their tests.

assets/js/googlesitekit/data/utils.test.js
```diff
@@ -124,6 +128,231 @@ describe( 'data utils', () => {
...
+		it( 'should combine stores missing any of the keys', () => {
+			const combinedStoreMissing = {
+				INITIAL_STATE: collect( storeOne.INITIAL_STATE, storeTwo.INITIAL_STATE, storeThree.INITIAL_STATE ),
+				actions: collectActions( storeOne.actions, storeTwo.actions ),
+				controls: collectControls( storeOne.controls, storeTwo.controls ),
+				reducer: collectReducers( storeOne.reducer, storeTwo.reducer ),
+				resolvers: collectResolvers( storeOne.resolvers, storeTwo.resolvers ),
+				selectors: collectSelectors( storeOne.selectors, storeTwo.selectors ),
+			};
+			expect( JSON.stringify( combineStores( storeOne, storeTwo, storeThree ) ) ).toBe( JSON.stringify( combinedStoreMissing ) );
```
tofumatt, line 302: I don't think this test is quite right. Stringifying these stores changes them in pretty fundamental ways: the actions, controls, reducers, and selectors all contain functions, which may not stringify accurately. Additionally: the aim here is to ensure two things:

1. The combined store has all of the actions, initial state, etc. of each individual store.
2. The utility doesn't error when certain pieces of a store (eg. controls, or even initial state) are _not_ supplied.

It sounds like the name of this test intends to test the second requirement, but it provides every required key. You seem to be thinking of each store as a complete set of items, and omitting parts of the third store, but the utility has no concept of what we consider "the entire store", so this test doesn't seem to be doing anything other than testing what things `stringify` to. I think this one can be removed, unless I'm missing something 😅

assets/js/googlesitekit/data/utils.test.js
```diff
@@ -124,6 +128,231 @@ describe( 'data utils', () => {
...
+				actions: collectActions( storeOne.actions, storeTwo.actions ),
+				controls: collectControls( storeOne.controls, storeTwo.controls ),
+				reducer: collectReducers( storeOne.reducer, storeTwo.reducer, storeFour.reducer ),
+				resolvers: collectResolvers( storeOne.resolvers, storeTwo.resolvers ),
+				selectors: collectSelectors( storeOne.selectors, storeTwo.selectors ),
+			};
+			expect( JSON.stringify( combineStores( storeOne, storeTwo, storeFour ) ) ).toBe( JSON.stringify( combinedStoreMissingTwo ) );
+		} );
+
+		it( 'INITIAL_STATEs, reducers, and actions should all work together even if provided by separate incomplete stores', () => {
```
tofumatt, line 315: This test case is better but there aren't any tests that completely omit a key for every store. I think it'd be worth adding a test with stores that completely lack every key:

* `INITIAL_STATE`
* `actions`
* `controls`
* `reducer`
* `resolvers`
* `selectors`

assets/js/googlesitekit/data/utils.test.js
```diff
@@ -124,6 +128,231 @@ describe( 'data utils', () => {
...
+			expect( state ).toEqual( { one: 1, two: 2 } );
+
+			// Reducer from one incomplete store responds to an action added by a different incomplete store.
+			state = combinedStore.reducer( state, combinedStore.actions.fakeActionOne() );
+			expect( state ).toEqual( { one: true, two: 2 } );
+		} );
+
+		it( 'should error if action keys are duplicated', () => {
+			expect( () => {
+				combineStores( storeOne, storeFive );
```
tofumatt, line 333: As @aaemnnosttv mentioned: without these stores being defined inside this test it's difficult to reason about what's happening. I'm _assuming_ that `storeOne` and `storeFive` have duplicated action names, but I can't tell without tracking them down elsewhere in the file. I think it'd be good to put those definitions inline.

A good aim for tests is to optimise for readability and to ignore the desire to "not repeat yourself". Tests—especially their fixtures/variables—should _frequently_ be repeated if needed. We'll only be writing a unit test once—and with a public, frequently-used API like this it's unlikely to change frequently. We should make it as easy-as-possible to reason about a single broken test 😄

assets/js/googlesitekit/data/utils.test.js
```diff
@@ -124,6 +128,231 @@ describe( 'data utils', () => {
...
+			} ).toThrow( /collect\(\) cannot accept collections with duplicate keys. Your call to collect\(\) contains the following duplicated functions: fakeActionOne./ );
+		} );
+
+		it( 'should error if control keys are duplicated', () => {
```
tofumatt, line 337: Similar to the duplicate checks for everything, I think it'd be good to have a test for each key being entirely omitted, eg:

```js
expect( () => {
  combineStores( {
    INITIAL_STATE: {},
    actions: {},
    controls: {},
    reducer: {},
    resolvers: {},
    selectors: undefined,
  } );
} ).not.toThrow();
```

for each of:

* `INITIAL_STATE`
* `actions`
* `controls`
* `reducer`
* `resolvers`
* `selectors`

assets/js/googlesitekit/data/utils.test.js
jqlee85: Yep, that makes sense. I'll go through and make these more collocated and readable. Thanks!

assets/js/googlesitekit/data/utils.js
jqlee85: Yeah I was considering that, but didn't since I needed the `INITIAL_STATE` to be used for building the `reducer` as well. But I can just build that initial state first and for the rest I'll just return the newly-built-in-place object. The comments seem like overkill when the object's being built in place, since it's a little clearer what's happening.

assets/js/googlesitekit/data/utils.test.js
jqlee85: 👍

assets/js/googlesitekit/data/utils.test.js
jqlee85: 👍

assets/js/googlesitekit/data/utils.js
```diff
@@ -212,6 +212,55 @@ export const collectName = ( ...args ) => {
...
 
+/**
+ * Combines multiple stores.
+ *
+ * @since n.e.x.t
+ *
+ * @param {...Object} args A list of objects, each a store containing one or more of the following keys: INITIAL_STATE, actions, controls, reducer, resolvers, selectors
+ * @return {Object} The combined store.
+ */
+export const combineStores = ( ...args ) => {
```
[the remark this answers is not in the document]
jqlee85: :+1:

assets/js/googlesitekit/data/utils.js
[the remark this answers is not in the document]
jqlee85: :+1:

assets/js/googlesitekit/data/utils.test.js
jqlee85: :+1:

assets/js/googlesitekit/data/utils.test.js
jqlee85: :+1:

assets/js/googlesitekit/data/utils.test.js
jqlee85: Makes sense, I've changed tactics, instead of testing the whole object (which required the stringification), I've switched to specifically testing the combined stores for the presence of each individual action, control.

Also this test has been removed, in favor of testing for the combination of several stores which each contain only one of the keys, then testing to makes sure they work together as expected once combined.

Let me know if that all makes sense, thanks!

felixarntz:
Should we migrate existing code to use `combineStore` separately from this issue/PR?

assets/js/googlesitekit/data/utils.js
```diff
@@ -212,6 +212,44 @@ export const collectName = ( ...args ) => {
...
+ *
+ * @param {...Object} stores A list of objects, each a store containing one or more of the following keys: INITIAL_STATE, actions, controls, reducer, resolvers, selectors
+ * @return {Object} The combined store.
+ */
+export const combineStores = ( ...stores ) => {
+	const defaultReducer = ( state ) => {
+		return { ...state };
+	};
+
+	const combinedInitialState = collectState(
```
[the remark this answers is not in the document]
tofumatt: Oh, nevermind; I see why it isn't—it's used below as well. 👍

aaemnnosttv:
> Should we migrate existing code to use `combineStore` separately from this issue/PR?

I think that makes sense as updating existing code wasn't within the scope of the original issue.

jqlee85:
@googlebot I consent.

aaemnnosttv:
@jqlee85 - @tofumatt needs to add the consent comment since it's your PR 😄

felixarntz:
I'm personally fine with either approach, doing it in this PR or a separate one.

I'd say the advantage of doing it right away is that we dogfood our own API before finalizing it (I'm sure in this particular case the API is solid, but it's still a good practice to use it if we can).
The disadvantage is that this becomes more code to review, and if we need to change the API because something isn't working as expected, it may become overly complex.

Your call :)

jqlee85:
I've integrated [googlesitekit/datastore/site/index.js](https://github.com/google/site-kit-wp/blob/8d3e1486db1096f347a6e5bf4b46198ac3cc5548/assets/js/googlesitekit/datastore/site/index.js). Let me know if that looks alright.

But looking at [create-module-store.js]( https://github.com/google/site-kit-wp/blob/8d3e1486db1096f347a6e5bf4b46198ac3cc5548/assets/js/googlesitekit/modules/create-module-store.js), I see it adds a STORE_NAME to the object. Should I update the util to handle STORE_NAMEs as well? Or I could just to assign that property after the object is created?

aaemnnosttv:
@jqlee85 @tofumatt @felixarntz Let's keep this PR limited to the original scope. I'll create a new issue for the refactoring, which should still be done now, but it's important from a project-management standpoint that we account for it separately.

@jqlee85 move any commits you have made regarding the refactoring from this branch to a new one for use with the new issue (https://github.com/google/site-kit-wp/issues/1476).

---

> But looking at [create-module-store.js]( https://github.com/google/site-kit-wp/blob/8d3e1486db1096f347a6e5bf4b46198ac3cc5548/assets/js/googlesitekit/modules/create-module-store.js), I see it adds a STORE_NAME to the object. Should I update the util to handle STORE_NAMEs as well? Or I could just to assign that property after the object is created?

I don't think `combineStores` needs to be concerned with `STORE_NAME` as this is only needed when we call `Data.registerStore()` with the resulting aggregate store.

assets/js/googlesitekit/datastore/site/index.js
```diff
@@ -26,64 +26,20 @@ import reset from './reset';
...
-		info.reducer,
-		reset.reducer,
-		notifications.reducer,
-	)
+const store = Data.combineStores(
+	{
+		actions: Data.addInitializeAction( Data.commonActions ),
+		controls: Data.commonControls,
+		reducer: Data.addInitializeReducer( {}, ( state ) => ( { ...state } ) ),
+	},
```
aaemnnosttv, line 36: @tofumatt should this maybe be exposed as `Data.commonStore` since all stores will need this?

tofumatt: Yeah, I think that makes sense actually 👍

jqlee85:
@aaemnnosttv I've reverted the last commit which will be moved to [the new branch](https://github.com/google/site-kit-wp/tree/enhancement/1476-refactor-datastores-combinestores-utility) for 1476.

tofumatt:
Sorry, the CLA always confuses me. 😅

tofumatt:
@googlebot I consent.

felixarntz:
@jqlee85 I opened #1502 for the migration of existing stores. Would be great if you could open a PR for that against this branch here, so we can review separately, then merge together.

I think it makes sense for us to wait with merging this until we've verified it works as expected when used - I'm not worried it doesn't, but I think since we can already make use of it in existing code, we should.

aaemnnosttv:
@jqlee85 tests are failing here from removing the `Data.collect*` utilities in https://github.com/google/site-kit-wp/pull/1455/commits/9db9c515442ed638b194f05040c34ef717b2ec3c. This should only happen in #1476 after existing stores have been refactored to use `combineStores` 😄

jqlee85:
@aaemnnosttv yeah, the tests are passing in #1476. Should I revert the commit removing the `collect...` methods in this branch?

Issue #1175, opened 2020-02-26
Scaffold googlesitekit-user package and core/user store


felixarntz:
## Feature Description

This is very much similar to #999, introducing a separate `googlesitekit-user` JS file and respective module that registers a `core/user` store on `googlesitekit.data`. The implementation should follow the same patterns established in #999 (even the selectors and actions are very similar in behavior).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a JS file `googlesitekit-user.js` that is registered in PHP as `googlesitekit-user`.
* The script should depend on `googlesitekit-data` and `googlesitekit-api`.
* The script should register a `core/user` store on `googlesitekit.data`.
* The store should (for this initial implementation) have the following selectors:
    * `getAuthentication()`: resolves to the `core/user/data/authentication` route and "pass-through" its response<sup>1</sup>
    * `isAuthenticated()`: Should use the respective property from the above selector
    * `getRequiredScopes()`: Should use the respective property from the above selector
    * `getGrantedScopes()`: Should use the respective property from the above selector
    * `getError()` returns `state.error` set by internal `CATCH_*` actions  
    This doesn't belong to any other internal piece so it should be defined in its own file/sub-store
* The store should (for this initial implementation) have the following actions:
    * `disconnect()`: calls `core/user/data/disconnect` route

<sup>1</sup> The `core/user/data/authentication` should be modified in PHP to just rename `isAuthenticated` to `authenticated` (consistency with naming of other such parameters in e.g. `core/site/data/connection`). Any existing references in PHP or JS code should be adjusted.

## Implementation Brief

* Create a new file in `assets/js/googlesitekit/datastore/user/index.js` that creates a new `'core/user'` store using `'googlesitekit-data'`'s `registerStore` .
* Add an asset entry in `Assets.php` to enqueue the file.
* Expose a `getAuthentication()` selector that uses data from the `core/user/data/authentication` endpoint
* Expose a `getRequiredScopes()`: Should use the respective property from the above selector
* Expose a `getGrantedScopes()`: Should use the respective property from the above selector
* Add a `disconnect()` action that calls the `core/user/data/disconnect` API endpoint for the current user
* Add appropriate Jest tests.

## QA Brief

* 

## Changelog entry

* Introduce `core/user` datastore for managing user-specific data in JS.

felixarntz:
IB ✅

felixarntz:
[...]

PR #1463, opened 2020-04-28
Scaffold googlesitekit-user package and core/user store.

ryanwelcher:
## Summary

Addresses issue #1175 

## Checklist


assets/js/googlesitekit/datastore/user/authentication.js
```diff
@@ -0,0 +1,199 @@
...
+				type: CATCH_FETCH_AUTHENTICATION,
+			};
+		}
+
+		return { response, error };
+	},
+	/**
+	 * Stores connection info received from the REST API.
+	 *
+	 * @since 1.5.0
```
aaemnnosttv, line 79:
```suggestion
	 * @since n.e.x.t
```

assets/js/googlesitekit/datastore/user/authentication.js
```diff
@@ -0,0 +1,199 @@
...
+	 *
+	 * Returns an object with the shape when successful:
+	 * ```
+	 * {
+	 *   authenticated: <Boolean>,
+	 * }
+	 * ```
+	 *
+	 * @private
+	 * @since 1.8.0
```
aaemnnosttv, line 155:
```suggestion
	 * @since n.e.x.t
```

assets/js/googlesitekit/datastore/user/authentication.js
```diff
@@ -0,0 +1,199 @@
...
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {Object|undefined} User authentication info.
```
aaemnnosttv, line 158:
```suggestion
	 * @return {?Object} User authentication info.
```

assets/js/googlesitekit/datastore/user/authentication.js
```diff
@@ -0,0 +1,199 @@
...
+	 *
+	 * @since 1.8.0
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {boolean|undefined} Site connection status.
+	 */
+	isAuthenticated: createRegistrySelector( ( select ) => () => {
+		const authentication = select( STORE_NAME ).getAuthentication();
+
+		return typeof authentication !== 'undefined' ? authentication.authenticated : authentication;
```
aaemnnosttv, line 178: We have mixed use in the codebase, but let's standardize on `var === undefined` rather than using `typeof`

assets/js/googlesitekit/datastore/user/authentication.js
```diff
@@ -0,0 +1,199 @@
...
+		const { authentication } = state;
+		return authentication;
+	},
+	/**
+	 * Gets the Site Kit authentication status for this user.
+	 *
+	 * Returns `true` if the site is connected to Site Kit, `false` if
+	 * not. Returns `undefined` if the connection info is not available/loaded.
+	 *
+	 * @since 1.8.0
```
aaemnnosttv, line 170: 👀

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -0,0 +1,220 @@
...
+ */
+import API from 'googlesitekit-api';
+import {
+	createTestRegistry,
+	muteConsole,
+	subscribeUntil,
+	unsubscribeFromAll,
+} from 'tests/js/utils';
+import { STORE_NAME } from './constants';
+import './index';
```
aaemnnosttv, line 35: This shouldn't be necessary
```suggestion
```

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -0,0 +1,220 @@
...
+
+	describe( 'actions', () => {
+		describe( 'fetchAuthentication', () => {
+			it( 'does not require any params', () => {
+				expect( () => {
+					registry.dispatch( STORE_NAME ).fetchAuthentication();
+				} ).not.toThrow();
+			} );
+		} );
+		describe( 'receiveConnection', () => {
```
aaemnnosttv, line 72: `receiveConnection` ? 😉

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -0,0 +1,220 @@
...
+	createTestRegistry,
+	muteConsole,
+	subscribeUntil,
+	unsubscribeFromAll,
+} from 'tests/js/utils';
+import { STORE_NAME } from './constants';
+import './index';
+
+describe( 'core/user authentication', () => {
+	const responseAuthenticated = { authenticated: true };
```
aaemnnosttv: Please define this closer to the assertions. It's only one line, so even if it's added to every test, that's okay.

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -0,0 +1,220 @@
...
+				);
+
+				const connection = registry.select( STORE_NAME ).getAuthentication();
+
+				expect( fetch ).toHaveBeenCalledTimes( 1 );
+				expect( connection ).toEqual( undefined );
+			} );
+		} );
+		describe( 'isAuthenticated', () => {
+			it( 'uses a resolver get all authentication info', async () => {
```
aaemnnosttv, line 154: Description is a bit off-ish here.
```suggestion
			it( 'uses a resolver to load the authenticated value if not set yet', async () => {
```

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -0,0 +1,220 @@
...
+					() => store.getState().isFetchingAuthentication === false,
+				);
+
+				const isAuthenticated = registry.select( STORE_NAME ).isAuthenticated();
+
+				expect( fetch ).toHaveBeenCalledTimes( 1 );
+				expect( isAuthenticated ).toEqual( undefined );
+			} );
+
+			it( 'returns undefined if connection info is not available', async () => {
```
aaemnnosttv, line 205: 👀 `connection info`

webpack.config.js
```diff
@@ -92,6 +92,7 @@ const webpackConfig = ( mode ) => {
 				// New Modules (Post-JSR).
 				'googlesitekit-api': './assets/js/googlesitekit-api.js',
 				'googlesitekit-data': './assets/js/googlesitekit-data.js',
+				'googlesitekit-user': './assets/js/googlesitekit-user.js',
```
aaemnnosttv, line 95: This should be 
```suggestion
				'googlesitekit-user': './assets/js/googlesitekit-datastore-user.js',
```

includes/Core/Assets/Assets.php
```diff
@@ -445,6 +445,16 @@ private function get_assets() {
 					),
 				)
 			),
+			new Script(
+				'googlesitekit-user',
+				array(
+					'src'          => $base_url . 'js/googlesitekit-user.js',
```
aaemnnosttv, line 451:
```suggestion
					'src'          => $base_url . 'js/googlesitekit-datastore-user.js',
```

includes/Core/Assets/Assets.php
```diff
@@ -445,6 +445,16 @@ private function get_assets() {
 					),
 				)
 			),
+			new Script(
+				'googlesitekit-user',
```
aaemnnosttv, line 449:
```suggestion
				'googlesitekit-datastore-user',
```

assets/js/googlesitekit-user.js
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * External dependencies
+ */
+
```
aaemnnosttv, line 22: This is included in some other files, mostly as a hack to satisfy `@wordpress/eslint-plugin` for the `dependency-groups` rule because it couldn't understand our aliases. That isn't a problem anymore (because we got rid of them) so this can be removed 👍

assets/js/googlesitekit-user.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * Entrypoint for the "core/user data store.
```
aaemnnosttv, line 2: Orphan quote

assets/js/googlesitekit/datastore/user/authentication.js
```diff
@@ -0,0 +1,199 @@
...
+		return {
+			payload: { authentication },
+			type: RECEIVE_AUTHENTICATION,
+		};
+	},
+};
+
+export const controls = {
+	[ FETCH_AUTHENTICATION ]: () => {
+		return API.get( 'core', 'user', 'authentication' );
```
aaemnnosttv, line 96: It looks like this wasn't part of the AC/IB but this route should be added to our preloaded paths (`googlesitekit_apifetch_preload_paths`).

We don't have any preloaded for core routes yet, so I'm not sure exactly where this should go yet, but perhaps `REST_Routes`. @felixarntz ?

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -0,0 +1,335 @@
...
+					() => store.getState().isFetchingAuthentication === false,
+				);
+
+				const requiredScopes = registry.select( STORE_NAME ).getRequiredScopes();
+
+				expect( fetch ).toHaveBeenCalledTimes( 1 );
+				expect( requiredScopes ).toEqual( undefined );
+			} );
+
+			it( 'returns undefined if connection info is not available', async () => {
```
aaemnnosttv, line 326: 👀  `connection info`

assets/js/googlesitekit/datastore/user/authentication.js
```diff
@@ -0,0 +1,235 @@
...
+	/**
+	 * Gets the Site Kit authentication status for this user.
+	 *
+	 * Returns `true` if the user is authenticated, `false` if
+	 * not. Returns `undefined` if the authentication info is not available/loaded.
+	 *
+	 * @since 1.8.0
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {boolean|undefined} Site connection status.
```
aaemnnosttv, line 173:
```suggestion
	 * @return {?boolean} User authenticated status.
```

assets/js/googlesitekit/datastore/user/authentication.js
```diff
@@ -0,0 +1,235 @@
...
+	/**
+	 * Gets the granted scopes for the user.
+	 *
+	 * Returns an array of granted scopes or undefined
+	 * if authentication info is not available/loaded.
+	 *
+	 * @since 1.8.0
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {Array|undefined} Array of granted scopes
```
aaemnnosttv, line 190:
```suggestion
	 * @return {?Array} Array of granted scopes
```

assets/js/googlesitekit/datastore/user/authentication.js
```diff
@@ -0,0 +1,235 @@
...
+		return typeof authentication !== 'undefined' ? authentication.authenticated : authentication;
+	} ),
+
+	/**
+	 * Gets the granted scopes for the user.
+	 *
+	 * Returns an array of granted scopes or undefined
+	 * if authentication info is not available/loaded.
+	 *
+	 * @since 1.8.0
```
aaemnnosttv, line 187: 👀

assets/js/googlesitekit/datastore/user/authentication.js
```diff
@@ -0,0 +1,235 @@
...
+		return typeof authentication !== 'undefined' ? authentication.grantedScopes : authentication;
+	} ),
+
+	/**
+	 * Gets the required scopes for the user.
+	 *
+	 * Returns an array of required scopes or undefined
+	 * if authentication info is not available/loaded.
+	 *
+	 * @since 1.8.0
```
aaemnnosttv, line 204: 👀

assets/js/googlesitekit/datastore/user/authentication.js
tofumatt: Is this better? I guess if that's the convention: fine, but that reads like the return type is optional. Or is anything. We don't return `true|false|void`; we return `true|false|undefined`. 🤷

assets/js/googlesitekit/datastore/user/authentication.js
```diff
@@ -0,0 +1,235 @@
...
+	 * Returns an array of granted scopes or undefined
+	 * if authentication info is not available/loaded.
+	 *
+	 * @since 1.8.0
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {Array|undefined} Array of granted scopes
+	 */
+	getGrantedScopes: createRegistrySelector( ( select ) => () => {
+		const authentication = select( STORE_NAME ).getAuthentication();
```
tofumatt, line 193: A little trick to returning `undefined` if `select( STORE_NAME ).getAuthentication();` is `undefined` is to do:

```js
const authentication = select( STORE_NAME ).getAuthentication() || {};

return authentication.grantedScopes;
```

Dunno if folks like that better, but we do it in some places and it's a bit easier to write the return statement. Might be harder to follow though.

This could work too:

```js
const { grantedScopes } = select( STORE_NAME ).getAuthentication() || {};

return grantedScopes;
```

Which do you find the most readable? 🙂

includes/Core/Assets/Assets.php
tofumatt: I guess the AC/IB sort of differ here, but I'm inclined to agree with @aaemnnosttv's suggestion. This is part of the datastore so let's scope it.

webpack.config.js
tofumatt: I find it awkward that we're scoping it in one place but not another. Why even have `googlesitekit-user` be it's own entrypoint, come to think of it? When isn't user info required when interacting with the API?

assets/js/googlesitekit/datastore/user/authentication.js
felixarntz: We have so far used `?` over `|undefined`, so I think we should continue to do so. Unless this is actually incorrect - then we should update all instances though (which we have a ton of).

webpack.config.js
felixarntz: Agreed, this should be similar to `googlesitekit-datastore-site` in the way it is approached regarding naming, file structure etc.

assets/js/googlesitekit/datastore/user/authentication.js
aaemnnosttv: Actually, that was a mistake. The nullable syntax is just for parameters I think, not return type. Disregard that suggestion! 😄

assets/js/googlesitekit/datastore/user/authentication.test.js
ryanwelcher: @aaemnnosttv these values don't change, I'm not sure it's a good idea to repeat the variable in multiple places. Are you concerned about readability here?

assets/js/googlesitekit/datastore/user/authentication.js
ryanwelcher: Personally, I find the second one easier to read.

assets/js/googlesitekit/datastore/user/authentication.js
ryanwelcher: @aaemnnosttv I have not addressed yes but will once @felixarntz responds.

felixarntz: Good point @aaemnnosttv (sorry, I had missed this ping earlier). The `core/site` `connection` route is currently not preloaded either though, so I think this is something we should revisit in a follow-up issue.

assets/js/googlesitekit/datastore/user/authentication.js
```diff
@@ -0,0 +1,233 @@
...
+	*getAuthentication() {
+		const registry = yield Data.commonActions.getRegistry();
+
+		const authenticationInfo = registry.select( STORE_NAME ).getAuthentication();
+
+		if ( ! authenticationInfo ) {
+			yield actions.fetchAuthentication();
+		}
+	},
+};
```
felixarntz, line 224: Nitpicking, but can we put this `resolvers` block above `selectors` like elsewhere?

includes/Core/Authentication/Authentication.php
```diff
@@ -602,10 +602,10 @@ private function inline_js_setup_data( $data ) {
 
 		$data['isSiteKitConnected'] = $this->credentials->has();
 		$data['isResettable']       = $this->options->has( Credentials::OPTION );
-		$data['isAuthenticated']    = ! empty( $access_token );
+		$data['authenticated']      = ! empty( $access_token );
```
felixarntz, line 605: I think changing this wasn't part of the ACs - maybe it was slightly vague, but only the API route should be modified, we don't need to update this legacy data here. Having this change here also makes this PR a lot bigger with more side effects.

includes/Core/Authentication/Authentication.php
```diff
@@ -703,9 +703,9 @@ private function get_rest_routes() {
 							$access_token = $oauth_client->get_access_token();
 
 							$data = array(
-								'isAuthenticated' => ! empty( $access_token ),
-								'requiredScopes'  => $oauth_client->get_required_scopes(),
-								'grantedScopes'   => ! empty( $access_token ) ? $oauth_client->get_granted_scopes() : array(),
+								'authenticated'  => ! empty( $access_token ),
```
felixarntz, line 706: Only here this actually needs to be changed (plus references to here of course).

assets/js/util/refresh-authentication.js
```diff
@@ -31,7 +31,7 @@ export const refreshAuthentication = async () => {
 
 		// We should really be using state management. This is terrible.
 		global.googlesitekit.setup = global.googlesitekit.setup || {};
-		global.googlesitekit.setup.isAuthenticated = response.isAuthenticated;
+		global.googlesitekit.setup.authenticated = response.authenticated;
```
felixarntz, line 34: I think this is the only line in JS legacy code that should be updated based on the renaming. `global.googlesitekit.setup.isAuthenticated = response.authenticated;` would be sufficient to keep the old name for the legacy JS data (which will be phased out in the long term) while supporting the new name in the API route.

All updates to the JS components above wouldn't be necessary.

ryanwelcher:
@felixarntz thank you for the feedback. I was overzealous in my renaming here. I'll revert the naming now and re-request a review.

ryanwelcher:
@felixarntz I've made the changes you mentioned. All tests are passing locally for me so I am not sure why the failure in Travis.

felixarntz:
@ryanwelcher Thanks for making the changes, looks good! The Travis-CI failure is probably just due to some e2e test instability that occasionally causes this, I just restarted that job.

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -0,0 +1,334 @@
...
+					.doMockOnceIf( apiEndpoint )
+					.mockResponseOnce(
+						JSON.stringify( responseAuthenticated ),
+						{ status: 200 }
+					);
+
+				const initialIsAuthenticated = registry.select( STORE_NAME ).getGrantedScopes();
+				// The autentication info will be its initial value while the authentication
+				// info is fetched.
+				expect( initialIsAuthenticated ).toEqual( undefined );
```
aaemnnosttv, line 225: Copy pasta (references to authentication info, `responseAuthenticated`, etc for granted scopes)

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -0,0 +1,334 @@
...
+	createTestRegistry,
+	muteConsole,
+	subscribeUntil,
+	unsubscribeFromAll,
+} from 'tests/js/utils';
+import { STORE_NAME } from './constants';
+
+describe( 'core/user authentication', () => {
+	const responseAuthenticated = { authenticated: true, requiredScopes: [], grantedScopes: [] };
+	const apiEndpoint = /^\/google-site-kit\/v1\/core\/user\/data\/authentication/;
```
aaemnnosttv, line 38: I'm not against this being shared between all tests but let's rename it to make the consuming code a bit clearer.
```suggestion
	const coreUserDataEndpointRegExp = /^\/google-site-kit\/v1\/core\/user\/data\/authentication/;
```

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -0,0 +1,334 @@
...
+				await subscribeUntil( registry,
+					// TODO: We may want a selector for this, but for now this is fine
+					// because it's internal-only.
+					() => store.getState().isFetchingAuthentication === false,
+				);
+
+				const grantedScopes = registry.select( STORE_NAME ).getGrantedScopes();
+
+				expect( fetch ).toHaveBeenCalledTimes( 1 );
+				expect( grantedScopes ).toEqual( undefined );
```
aaemnnosttv, line 262: Missing expectation about error 😄 (there may be other places that need this in other older tests too)

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -0,0 +1,334 @@
...
+			} );
+
+			it( 'returns undefined if authentication info is not available', async () => {
+				// This triggers a network request, so ignore the error.
+				muteConsole( 'error' );
```
aaemnnosttv, line 267: See above

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -0,0 +1,334 @@
...
+					.doMockOnceIf( apiEndpoint )
+					.mockResponseOnce(
+						JSON.stringify( responseAuthenticated ),
+						{ status: 200 }
+					);
+
+				const initialIsAuthenticated = registry.select( STORE_NAME ).getRequiredScopes();
+				// The autentication info will be its initial value while the authentication
+				// info is fetched.
+				expect( initialIsAuthenticated ).toEqual( undefined );
```
aaemnnosttv, line 285: Copy pasta references to authentication info 😄

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -0,0 +1,334 @@
...
+				await subscribeUntil( registry,
+					// TODO: We may want a selector for this, but for now this is fine
+					// because it's internal-only.
+					() => store.getState().isFetchingAuthentication === false,
+				);
+
+				const requiredScopes = registry.select( STORE_NAME ).getRequiredScopes();
+
+				expect( fetch ).toHaveBeenCalledTimes( 1 );
+				expect( requiredScopes ).toEqual( undefined );
```
aaemnnosttv, line 322: Error assertion missing

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -0,0 +1,334 @@
...
+			} );
+
+			it( 'returns undefined if authentication info is not available', async () => {
+				// This triggers a network request, so ignore the error.
+				muteConsole( 'error' );
```
aaemnnosttv, line 327: Same as above

assets/js/googlesitekit/datastore/user/disconnect.js
```diff
@@ -0,0 +1,133 @@
...
+ */
+import invariant from 'invariant';
+
+/**
+ * Internal dependencies
+ */
+import API from 'googlesitekit-api';
+
+// Actions
+const START_DISCONNECT = 'START_DISCONNECT';
```
aaemnnosttv, line 30: This should be `START_FETCH_DISCONNECT` as `FETCH_DISCONNECT` is the main action.

assets/js/googlesitekit/datastore/user/disconnect.js
```diff
@@ -0,0 +1,133 @@
...
+			payload: {},
+			type: START_DISCONNECT,
+		};
+
+		try {
+			response = yield {
+				payload: {},
+				type: FETCH_DISCONNECT,
+			};
+			yield actions.receiveDisconnect( response );
```
aaemnnosttv, line 53: I think the receive here is unnecessary as there is no useful response to store here. The API literally just returns `true` right now but the request succeeding is enough by itself for now 😄

assets/js/googlesitekit/datastore/user/disconnect.js
```diff
@@ -0,0 +1,133 @@
...
+
+// Actions
+const START_DISCONNECT = 'START_DISCONNECT';
+const FETCH_DISCONNECT = 'FETCH_DISCONNECT';
+const FINISH_DISCONNECT = 'FINISH_DISCONNET';
+const CATCH_FETCH_DISCONNECT = 'CATCH_FETCH_DISCONNECT';
+const RECEIVE_DISCONNECT = 'RECEIVE_DISCONNECT';
+
+export const INITIAL_STATE = {
+	isDisconnecting: false,
```
aaemnnosttv, line 37: This state should be `isFetchingDisconnect`

assets/js/googlesitekit/datastore/user/disconnect.js
```diff
@@ -0,0 +1,133 @@
...
+	},
+
+	receiveDisconnect( disconnect ) {
+		invariant( disconnect, 'disconnect is required.' );
+
+		return {
+			payload: { disconnect },
+			type: RECEIVE_DISCONNECT,
+		};
+	},
```
aaemnnosttv, line 76: Not necessary 😄

assets/js/googlesitekit/datastore/user/disconnect.test.js
```diff
@@ -0,0 +1,50 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Internal dependencies
+ */
+import { createTestRegistry, unsubscribeFromAll } from 'tests/js/utils';
```
aaemnnosttv, line 22: Relative import please 😇

assets/js/googlesitekit/datastore/user/index.js
```diff
@@ -0,0 +1,77 @@
...
+import authentication from './authentication';
+import disconnect from './disconnect';
+import { STORE_NAME } from './constants';
+
+export const INITIAL_STATE = Data.collectState(
+	authentication.INITIAL_STATE,
+	disconnect.INITIAL_STATE
+);
+
+export { STORE_NAME };
```
aaemnnosttv, line 32: Let's keep this as the first export (I believe all stores currently do this). It can go right below the import 👍

webpack.config.js
```diff
@@ -92,6 +92,7 @@ const webpackConfig = ( mode ) => {
 				// New Modules (Post-JSR).
 				'googlesitekit-api': './assets/js/googlesitekit-api.js',
 				'googlesitekit-data': './assets/js/googlesitekit-data.js',
+				'googlesitekit-datastore-user': './assets/js/googlesitekit-datastore-user.js',
```
aaemnnosttv, line 95: Super nitpick, but can we keep alpha-sort here? (i.e. after datastore-site)

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -0,0 +1,334 @@
...
+				await subscribeUntil( registry,
+					// TODO: We may want a selector for this, but for now this is fine
+					// because it's internal-only.
+					() => store.getState().isFetchingAuthentication === false,
+				);
+
+				const connection = registry.select( STORE_NAME ).getAuthentication();
+
+				expect( fetch ).toHaveBeenCalledTimes( 1 );
+				expect( connection ).toEqual( undefined );
```
aaemnnosttv, line 148: Missing error assertion.

assets/js/googlesitekit/datastore/user/disconnect.js
aaemnnosttv: @tofumatt we don't need `disconnect` state, do we? I suppose it could be used for something but currently disconnect happens with a page reload so not sure if it would really be useful since we keep spinners etc going with local state 🤷

tofumatt: I'm personally of the opinion that it's useful to have a state like that—even if right now it's used to force a redirect (`if (isDisconnected) { window.location = '/sign-in'; }` or whatever).

It doesn't really hurt, but it's inline with our other modules so I think is fine.

aaemnnosttv: @ryanwelcher please disregard my comments about removing this! 😄

ryanwelcher: Just to be clear, I'll add `disconnected` to the initial state and create a selector for `isDisconnected`?

assets/js/googlesitekit/datastore/user/disconnect.js
```diff
@@ -0,0 +1,144 @@
...
+};
+
+export const controls = {
+	[ FETCH_DISCONNECT ]: () => {
+		return API.get( 'core', 'user', 'disconnect' );
```
felixarntz, line 92: This should use `API.set` since that is a `POST` endpoint.

assets/js/googlesitekit/datastore/user/disconnect.js
```diff
@@ -0,0 +1,144 @@
...
+
+// Actions
+const START_FETCH_DISCONNECT = 'START_FETCH_DISCONNECT';
+const FETCH_DISCONNECT = 'FETCH_DISCONNECT';
+const FINISH_DISCONNECT = 'FINISH_DISCONNET';
+const CATCH_FETCH_DISCONNECT = 'CATCH_FETCH_DISCONNECT';
+const RECEIVE_DISCONNECT = 'RECEIVE_DISCONNECT';
+
+export const INITIAL_STATE = {
+	isFetchingDisconnecting: false,
```
felixarntz, line 37: Nit-picking, but I think `isFetchingDisconnect` is more appropriate here.

assets/js/googlesitekit/datastore/user/disconnect.js
```diff
@@ -0,0 +1,144 @@
...
+			};
+		}
+		default: {
+			return { ...state };
+		}
+	}
+};
+
+export const resolvers = {};
+export const selectors = {};
```
felixarntz, line 135: I'm noticing this was not part of the ACs, but can we add a selector here `isDoingDisconnect()`, which returns the value of `isFetchingDisconnect`? For actions specifically, having a selector to see whether that action is currently in progress is useful (for selectors that rely on APIs we don't need that, since the return values of them already indicate the "loading" state).

assets/js/googlesitekit/datastore/user/index.js
```diff
@@ -0,0 +1,83 @@
...
+	error.selectors
+);
+
+const store = {
+	actions,
+	controls,
+	reducer,
+	resolvers,
+	selectors,
+};
```
felixarntz, line 78: Let's keep an eye out on #1400 - if that gets merged before, we should leverage it here. If this gets merged first, we should make sure this code is updated as part of #1476.

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -0,0 +1,358 @@
...
+	} );
+
+	describe( 'actions', () => {
+		describe( 'fetchAuthentication', () => {
+			it( 'does not require any params', () => {
+				expect( () => {
+					registry.dispatch( STORE_NAME ).fetchAuthentication();
+				} ).not.toThrow();
+			} );
+		} );
```
aaemnnosttv, line 70: We can probably remove this test. We don't have tests for most fetch actions I think because they're implicitly covered when testing "uses a resolver to make a network request".

I'm not against keeping it, but as it is it will log an error in the console because the request will fail as there's no fetch mock in place.

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -0,0 +1,358 @@
...
+					.doMockOnceIf( coreUserDataEndpointRegExp )
+					.mockResponseOnce(
+						JSON.stringify( coreUserDataExpectedResponse ),
+						{ status: 200 }
+					);
+
+				const initialConnection = registry.select( STORE_NAME ).getAuthentication();
+				// The authentication info will be its initial value while the authentication
+				// info is fetched.
+				expect( initialConnection ).toEqual( undefined );
```
aaemnnosttv, line 93:
```suggestion
				const initialAuthentication = registry.select( STORE_NAME ).getAuthentication();
				// The authentication info will be its initial value while the authentication
				// info is fetched.
				expect( initialAuthentication ).toEqual( undefined );
```

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -0,0 +1,358 @@
...
+						{ status: 500 }
+					);
+
+				muteConsole( 'error' );
+				registry.select( STORE_NAME ).getAuthentication();
+				await subscribeUntil( registry,
+					// TODO: We may want a selector for this, but for now this is fine
+					// because it's internal-only.
+					() => store.getState().isFetchingAuthentication === false,
+				);
```
aaemnnosttv, line 143: Nitpick: we do this in a number of places but it isn't necessary to reference this in the store directly and I don't think there's really any place where we actually need to.

We should be using `hasFinishedResolution( 'getAuthentication' )` like above or using another selector.

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -0,0 +1,358 @@
...
+
+				const connection = registry.select( STORE_NAME ).getAuthentication();
+
+				expect( fetch ).toHaveBeenCalledTimes( 1 );
+				expect( connection ).toEqual( undefined );
```
aaemnnosttv, line 148:
```suggestion
				const authentication = registry.select( STORE_NAME ).getAuthentication();

				expect( fetch ).toHaveBeenCalledTimes( 1 );
				expect( authentication ).toEqual( undefined );
```

assets/js/googlesitekit/datastore/user/authentication.test.js
aaemnnosttv: @ryanwelcher this was resolved but it doesn't seem like it was added.

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -0,0 +1,358 @@
...
+					.doMockOnceIf( coreUserDataEndpointRegExp )
+					.mockResponseOnce(
+						JSON.stringify( coreUserDataExpectedResponse ),
+						{ status: 200 }
+					);
+
+				const initialIsGrantedScopes = registry.select( STORE_NAME ).getGrantedScopes();
+				// The granted scope info will be its initial value while the granted scope
+				// info is fetched.
+				expect( initialIsGrantedScopes ).toEqual( undefined );
```
aaemnnosttv, line 233:
```suggestion
				const initialGrantedScopes = registry.select( STORE_NAME ).getGrantedScopes();
				// The granted scope info will be its initial value while the granted scope
				// info is fetched.
				expect( initialGrantedScopes ).toEqual( undefined );
```

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -0,0 +1,358 @@
...
+					.doMockOnceIf( coreUserDataEndpointRegExp )
+					.mockResponseOnce(
+						JSON.stringify( coreUserDataExpectedResponse ),
+						{ status: 200 }
+					);
+
+				const initialIRequiredScopes = registry.select( STORE_NAME ).getRequiredScopes();
+				// The required scope info will be its initial value while the required scope
+				// info is fetched.
+				expect( initialIRequiredScopes ).toEqual( undefined );
```
aaemnnosttv, line 301:
```suggestion
				const initialRequiredScopes = registry.select( STORE_NAME ).getRequiredScopes();
				// The required scope info will be its initial value while the required scope
				// info is fetched.
				expect( initialRequiredScopes ).toEqual( undefined );
```

assets/js/googlesitekit/datastore/user/disconnect.js
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * Internal dependencies
+ */
+import API from 'googlesitekit-api';
+
+// Actions
+const START_FETCH_DISCONNECT = 'START_FETCH_DISCONNECT';
+const FETCH_DISCONNECT = 'FETCH_DISCONNECT';
+const FINISH_DISCONNECT = 'FINISH_DISCONNET';
```
aaemnnosttv, line 32: Bit of a typo in the value but this is slightly off in the formatting as well.
```suggestion
const FINISH_FETCH_DISCONNECT = 'FINISH_FETCH_DISCONNECT';
```

assets/js/googlesitekit/datastore/user/disconnect.js
```diff
@@ -0,0 +1,157 @@
...
+			response = yield {
+				payload: {},
+				type: FETCH_DISCONNECT,
+			};
+
+			yield actions.receiveDisconnect( response );
+
+			yield {
+				payload: {},
+				type: FINISH_DISCONNECT,
```
aaemnnosttv, line 59:
```suggestion
				type: FINISH_FETCH_DISCONNECT,
```

assets/js/googlesitekit/datastore/user/disconnect.js
```diff
@@ -0,0 +1,157 @@
...
+	/**
+	 * Stores the disconnection info received from the REST API.
+	 *
+	 * @since n.e.x.t
+	 * @private
+	 *
+	 * @param {Object} disconnect Disconnection info from the API.
+	 * @return {Object} Redux-style action.
+	 */
+	receiveDisconnect( disconnect ) {
```
aaemnnosttv, line 80: I think this should mirror the state, which would also save renaming the var in the reducer.
```suggestion
	receiveDisconnected( disconnected ) {
```

assets/js/googlesitekit/datastore/user/disconnect.js
```diff
@@ -0,0 +1,157 @@
...
+		case RECEIVE_DISCONNECT: {
+			const { disconnect: disconnected } = payload;
+
+			return {
+				...state,
+				disconnected,
+			};
+		}
+
+		case FINISH_DISCONNECT: {
```
aaemnnosttv, line 114:
```suggestion
		case FINISH_FETCH_DISCONNECT: {
```

assets/js/googlesitekit/datastore/user/disconnect.js
```diff
@@ -0,0 +1,157 @@
...
+	switch ( type ) {
+		case START_FETCH_DISCONNECT: {
+			return {
+				...state,
+				isFetchingDisconnect: true,
+			};
+		}
+
+		case RECEIVE_DISCONNECT: {
+			const { disconnect: disconnected } = payload;
```
aaemnnosttv, line 106: See above
```suggestion
			const { disconnected } = payload;
```

assets/js/googlesitekit/datastore/user/disconnect.js
```diff
@@ -0,0 +1,157 @@
...
+	 * Returns whether a disconnect is occuring.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {boolean} Is a disconnect ocurring or not.
+	 */
+	isDoingDisconnect( state ) {
+		const { isDoingDisconnect } = state;
+		return isDoingDisconnect;
```
aaemnnosttv, line 146: This should reference `state.isFetchingDisconnect` as there is no `isDoingDisconnect` in state.

We use `isDoing` as a prefix for public selectors for async actions. They usually reference `isFetching..` internally.

assets/js/googlesitekit/datastore/user/disconnect.test.js
```diff
@@ -0,0 +1,68 @@
...
+
+	describe( 'disconnect', () => {
+		// Create a mock to avoid triggering a network request error.
+		// The return value is irrelevant to the test.
+		fetch
+			.doMockOnceIf( coreUserDataDisconnectEndpointRegExp )
+			.mockResponseOnce(
+				JSON.stringify( {} ),
+				{ status: 200 }
+			);
```
aaemnnosttv, line 45: Let's move this within the test that triggers it below. I think this probably works as-is but let's avoid adding code in the body of a `describe` like this and put it either in an individual test or in a `before(All|Each)` function.

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -0,0 +1,365 @@
...
+					.doMockOnceIf( coreUserDataEndpointRegExp )
+					.mockResponseOnce(
+						JSON.stringify( coreUserDataExpectedResponse ),
+						{ status: 200 }
+					);
+
+				const initialIGrantedScopes = registry.select( STORE_NAME ).getGrantedScopes();
+				// The granted scope info will be its initial value while the granted scope
+				// info is fetched.
+				expect( initialIGrantedScopes ).toEqual( undefined );
```
aaemnnosttv, line 240: Typo (extra 👁️ )

```suggestion
				const initialGrantedScopes = registry.select( STORE_NAME ).getGrantedScopes();
				// The granted scope info will be its initial value while the granted scope
				// info is fetched.
				expect( initialGrantedScopes ).toEqual( undefined );
```

assets/js/googlesitekit/datastore/user/disconnect.js
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * Internal dependencies
+ */
+import API from 'googlesitekit-api';
+
+// Actions
+const START_FETCH_DISCONNECT = 'START_FETCH_DISCONNECT';
+const FETCH_DISCONNECT = 'FETCH_DISCONNECT';
+const FINISH_FETCH_DISCONNECT = 'FINISH_DISCONNET';
```
aaemnnosttv, line 32: This one was missed - not a big deal because all other references use the variable, but there's also a typo.
```suggestion
const FINISH_FETCH_DISCONNECT = 'FINISH_FETCH_DISCONNECT';
```

assets/js/googlesitekit/datastore/user/disconnect.js
```diff
@@ -0,0 +1,157 @@
...
+	 * @private
+	 *
+	 * @param {Object} disconnect Disconnection info from the API.
+	 * @return {Object} Redux-style action.
+	 */
+	receiveDisconnected( disconnect ) {
+		invariant( disconnect, 'disconnect is required.' );
+
+		return {
+			payload: { disconnect },
```
aaemnnosttv, line 84: The payload in the reducer expects this to be `disconnected` (maybe missing a test here?)
```suggestion
	receiveDisconnected( disconnected ) {
		invariant( disconnected !== undefined, 'disconnected is required.' );

		return {
			payload: { disconnected },
```
(`disconnected` is a boolean so we need to adjust the logic for `invariant` accordingly as well to be able to accept `false`)

assets/js/googlesitekit/datastore/user/disconnect.js
```diff
@@ -0,0 +1,157 @@
...
+		return { response, error };
+	},
+
+	/**
+	 * Stores the disconnection info received from the REST API.
+	 *
+	 * @since n.e.x.t
+	 * @private
+	 *
+	 * @param {Object} disconnect Disconnection info from the API.
```
aaemnnosttv, line 77:
```suggestion
	 * @param {boolean} disconnected Disconnection response from the API.
```

assets/js/googlesitekit/datastore/user/index.js
aaemnnosttv: The PR for #1400 has been merged but the PR for #1476 (#1484) has not yet, so this can be done there.

Issue #1101, opened 2020-02-05
Split Analytics module setup and settings into dedicated components


aaemnnosttv:
## Feature Description

Currently the `AnalyticsSetup` component is overly complex and becoming more and more difficult to maintain. In line with the goals and trajectory of the future architecture for JSR, we should split this into multiple dedicated components with less logical complexity in each.

The main separation to make is between the setup and settings components, however there will need to be some pieces which are shared between both, such as the dropdowns shown during setup and settings-edit.

This iteration should both serve as the approach to use for other modules but also as a stepping stone to get as close as we can to the end goal of having each module responsible for its full output rather than just the inner contents. This step will eventually be taken but should happen later when all modules can be migrated at once. With this in mind, we should architect the new components in such a way that makes the next step as easy as possible.

In addition to this separation of component responsibilities, many enhancements and general cleanup can be done as well such as:

* Replacing magic values with named constants
* Improving flow control with better error handling
* Avoid mixing placeholder options/values in state with API responses
* Introduce utilities for validating account/property/profile IDs

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* `AnalyticsSetup` should be responsible for rendering the module's setup component only
* A new component should be introduced for rendering the module's settings component
    - A new component should be introduced for rendering the common interface shared between module setup and edit settings
    - When not in edit mode, settings views should always display current saved values and messaging
        - Supporting data which is not part of settings can be requested on demand but should be cached temporarily in the browser where possible  
        E.g. Property name
* New components should be covered by component tests with Jest
* Existing E2E tests should require minimal changes
* New components should follow JSR guidelines for file structure

## Implementation Brief

* Create a new `AnalyticsSettingsView` component for displaying saved settings
* Create a new `AnalyticsSettingsEdit` component for rendering the settings edit view composed of shared components with setup
    - Use separate components to only trigger API requests when `componentDidMount` in edit mode  
    (currently these happen more often than necessary due to a shared component)
* Extract dedicated components for settings inputs
    - `AnalyticsAccountPropertyProfileSelect` - dropdowns for account, property, and profile 
    - `AnalyticsUseSnippetToggle` - updated to be consistent with those for TagManager and AdSense using a `Switch` internally with alternate descriptions for on/off states  
    See #1048
    - `AnalyticsIPAnonymizationToggle`
    - `AnalyticsExclusionsListToggles` - different in that it renders a list of toggles for each optional exclusion
* General improvements
[...]

PR #1465, opened 2020-04-29
1101 QA Feedback

aaemnnosttv:
## Summary

Addresses issue #1101 (Follow up to #1454)

## Relevant technical choices

* Updated `SettingsView` to display the `ErrorNotice` - shows expected error when attempting to create a property in an account with insufficient permissions:
image
* Added base styles for `<legend>` elements which mirror the styles we have for `<p>` tags for now (now styling is consistent between 1.7.1 and current for "Exclude in Analytics" setting)
* Removed bottom margin from "Please select the account information below" text above inputs with inline style (it might be a good idea to add a few utility classes for things like this in the future)

## Checklist


aaemnnosttv:
One thing that we're still looking into (which wasn't raised in QA) but that I noticed is how `ROLLBACK_SETTINGS` is being dispatched during `submitChanges` when it shouldn't. It doesn't seem to be causing a problem, but it's a bit concerning that it doesn't appear to be behaving as expected.

assets/js/modules/analytics/setup/setup-form.js
```diff
@@ -60,14 +60,14 @@ export default function SetupForm( { finishSetup } ) {
 		>
 			<ErrorNotice />
 
+			<ExistingTagNotice />
+
 			{ ( !! accounts.length && ! hasExistingTag ) && (
-				<p>
+				<p style={ { marginBottom: 0 } }>
```
tofumatt, line 66: 😱

Any chance we can get a CSS class for this? I know it's just in one place but feels like a dangerous precedent to set.

assets/js/modules/analytics/settings/settings-view.js
```diff
@@ -45,6 +45,7 @@ export default function SettingsView() {
 	return (
 		<div className="googlesitekit-setup-module googlesitekit-setup-module--analytics">
 
+			<ErrorNotice />
```
felixarntz, line 48: You moved the `ErrorNotice` component around recently to a void duplication (see e.g. https://github.com/google/site-kit-wp/pull/1454/commits/0e19750e7b5b1c794d92a40be29c4c654d7ef37e) - just want to make sure adding this here doesn't re-introduce such duplication potentially?

assets/js/modules/analytics/setup/setup-form.js
aaemnnosttv: It isn't the first inline style in a component, but sure I'll add a utility 👍

assets/js/modules/analytics/settings/settings-view.js
aaemnnosttv: Yeah, I think it probably does. I've updated it to prevent that 👍

assets/sass/utilities/_margin.scss
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+.googlesitekit-plugin {
+
+	.googlesitekit-mb-0 {
```
tofumatt, line 21: Just a nitpick but—it might be nice to label this more explicitly. Eg:
```suggestion
	.googlesitekit-margin-bottom-0 {
```

Makes it easier to reason about when you see it in the component.

Issue #1176, opened 2020-02-26
Implement additional googlesitekit-user store selectors


felixarntz:
## Feature Description

Following #1175, the `core/user` store should include further selectors for data that is passed to JS via PHP. These changes should be added to the `googlesitekit-user` script. This will be very similar to the work executed in #1000.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `core/user` store should have the following additional selectors (for data passed from PHP):
    * `getUser()`
    * `isVerified()`
* all these selectors should be sourced with data that is available on pageload via PHP; in other words, the script should receive a `before_print` callback in PHP that adds the data to some specific JS global (`_googlesitekitUserData`), solely for consumption by that script
    * `user` object with `id`, `name`, `email`, `picture` (see https://github.com/google/site-kit-wp/blob/1.3.1/includes/Core/Assets/Assets.php#L575-L580)
    * `verified` boolean
    * this data should run through a new `googlesitekit_user_data` filter
    * the filter should then be leveraged by the `Authentication` class to modify the data like in https://github.com/google/site-kit-wp/blob/1.3.1/includes/Core/Authentication/Authentication.php#L526-L530 and https://github.com/google/site-kit-wp/blob/1.3.1/includes/Core/Authentication/Authentication.php#L573-L577
    * none of the existing code around similar filters should be modified or removed (although duplicate code), as it will be faded out in the future once we migrate everything

## Implementation Brief

* Add a `data_callback` callback in `Assets.php` for `googlesitekit-datastore-user`, similar to the one here: https://github.com/google/site-kit-wp/pull/1276/files#diff-f878941036c14d763ba1cda6ed924dcdR365, for the extra data needed by `getUser()`.
* Add a `googlesitekit_user_data` filter in PHP that returns https://github.com/google/site-kit-wp/blob/1.3.1/includes/Core/Assets/Assets.php#L575 by default. Update https://github.com/google/site-kit-wp/blob/1.3.1/includes/Core/Authentication/Authentication.php#L526-L530 and https://github.com/google/site-kit-wp/blob/1.3.1/includes/Core/Authentication/Authentication.php#L573-L577 to use `add_filter('googlesitekit_user_data', function ($data) { return $new_data } )` to add data to the filtered data rather than getting and extending the object directly.
[...]

PR #1483, opened 2020-05-01
Adding additional core/user store selectors

ryanwelcher:
## Summary

Addresses issue #1176

## Relevant technical choices

This PR is branched from the PR for #1175 which will need to be merged to `develop` first.

## Checklist


assets/js/googlesitekit/datastore/user/user-info.js
```diff
@@ -0,0 +1,212 @@
...
+	 *
+	 * @private
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {(Object|undefined)} User info.
+	 */
+	getUser( state ) {
+		const { user, verified } = state;
+		return { user, verified };
```
felixarntz, line 125: This is a bit odd I think - this selector should rather just return the `user` object.

assets/js/googlesitekit/datastore/user/user-info.js
```diff
@@ -0,0 +1,212 @@
...
+	 *
+	 * Returns the true if the user is verified, false if not verified, or `undefined` if the user info is not available/loaded.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {(boolean|undefined)} The user ID.
+	 */
+	isVerified: createRegistrySelector( ( select ) => () => {
+		const { verified } = select( STORE_NAME ).getUser() || {};
```
felixarntz, line 199: This value should come directly from the datastore state instead (no need for `createRegistrySelector` here). What this will need though (based on the other comment that `getUser` should not include `verified`) is a resolver that is essentially the same as the resolver for `getUser`, just checking for the `isVerified` selector instead.

assets/js/googlesitekit/datastore/user/user-info.js
```diff
@@ -0,0 +1,212 @@
...
+			return { ...state };
+		}
+	}
+};
+
+export const resolvers = {
+	*getUser() {
+		const { select } = yield Data.commonActions.getRegistry();
+
+		const { user, verified } = select( STORE_NAME ).getUser();
```
felixarntz, line 88: Based on the other comments, this only should receive `user`, and shouldn't need to worry about `verified`.

Basically, both `getUser` and `isVerified` resolvers would result in the same logic, but since the check for whether a value is already set, that shouldn't be a problem. Plus it's not making an API request, so no real overhead anyway.

includes/Core/Authentication/Authentication.php
```diff
@@ -586,6 +586,21 @@ private function inline_js_admin_data( $data ) {
...
+				$new_data     = array(
+					'id'      => $current_user->ID,
+					'email'   => $data['userData']['email'],
+					'name'    => $current_user->display_name,
+					'picture' => $data['userData']['picture'],
+				);
+				$user['user'] = $new_data;
+				return $user;
+			}
+		);
```
felixarntz, line 602: Let's decouple this filter hook from `inline_js_admin_data` (since that is legacy) and hook it in in the `register` method. The logic here regarding getting the (Google account) profile data can be duplicated, that's fine in this case (legacy vs non-legacy). We also don't need to use `wp_get_current_user`, let's only update the parts of the `$user['user']` data that need to be updated (`email` and `picture`, if set).

includes/Core/Authentication/Authentication.php
```diff
@@ -633,6 +648,15 @@ private function inline_js_setup_data( $data ) {
 			$data['isVerified'] = false;
 		}
 
+		add_filter(
+			'googlesitekit_user_data',
+			function( $user ) use ( $data ) {
+				$user['verified'] = $data['isVerified'];
+				return $user;
+			}
+		);
```
felixarntz, line 657: Same here, let's move this to the `register` method and duplicate the relevant logic, since `inline_js_setup_data` is legacy code that this shouldn't be tied into.

ryanwelcher:
@googlebot I fixed it.

ryanwelcher:
@felixarntz I addressed your feedback but just wanted to call out that I actually added an action ( and tests ) to handle when `isVerified()` is called. It's very much the same as `getUser` but I thought adding it in would provide the structure for future changes if needed.

On the PHP side, I moved the hook as instructed but just wanted to confirm that the user `email` and `picture` are coming from the WP user and not `$this->profile` in the `Authentication` class.

assets/js/googlesitekit/datastore/user/user-info.js
```diff
@@ -0,0 +1,256 @@
...
+		if ( select( STORE_NAME ).isVerified() !== undefined ) {
+			return;
+		}
+
+		if ( ! global._googlesitekitUserData ) {
+			global.console.error( 'Could not load core/user info.' );
+			return;
+		}
+
+		yield actions.receiveUserIsVerified( { ...global._googlesitekitUserData } );
```
felixarntz, line 143: I think this is not correct, it should only pass the `verified` bit as single parameter (no object).

assets/js/googlesitekit/datastore/user/user-info.js
```diff
@@ -0,0 +1,256 @@
...
+		if ( select( STORE_NAME ).getUser() !== undefined ) {
+			return;
+		}
+
+		if ( ! global._googlesitekitUserData ) {
+			global.console.error( 'Could not load core/user info.' );
+			return;
+		}
+
+		yield actions.receiveUserInfo( { ...global._googlesitekitUserData } );
```
felixarntz, line 128: Regarding my previous comment, this should maybe become `actions.receiveUser( global._googlesitekitUserData.user )`.

includes/Core/Authentication/Authentication.php
```diff
@@ -260,6 +260,21 @@ function () {
...
+				if ( $profile_data ) {
+					if ( $profile_data ) {
+						$user['user']['email']   = $profile_data['email'];
+						$user['user']['picture'] = $profile_data['photo'];
+					}
+				}
+				$user['verified'] = $this->verification->has();
+				return $user;
+			}
+		);
```
felixarntz, line 276: 👍

assets/js/googlesitekit/datastore/user/user-info.js
```diff
@@ -0,0 +1,256 @@
...
+	 *
+	 * @since n.e.x.t
+	 * @private
+	 *
+	 * @param {Object} userInfo User info, usually supplied via a global variable from PHP.
+	 * @return {Object} Redux-style action.
+	 */
+	receiveUserInfo( userInfo ) {
+		invariant( userInfo, 'userInfo is required.' );
+		const { user, verified } = userInfo;
```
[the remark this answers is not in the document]
felixarntz: Or alternatively we all bring it back into one (that is how the `core/site` store has it) - the key change we'd need to make from how that was before here though would be that we would add a `getUserInfo` selector`, and that would have the single necessary resolver that populates all data from the global. Then the `getUser` selector would select from `getUserInfo`, as would `isVerified`.

ryanwelcher:
@felixarntz I went with the approach of keeping them separate. Feedback addressed.

Issue #1476, opened 2020-05-01
Refactor existing datastores with combineStores


aaemnnosttv:
This issue is a follow-up to #1400 to update our existing code to use the new utility.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All existing datastore definitions should use `Data.combineStores` for aggregating stores into the single store which is registered on `Data`

## Implementation Brief

* Update following store creations to use to use `combineStores()` utility.
  * https://github.com/google/site-kit-wp/blob/develop/assets/js/googlesitekit/datastore/site/index.js
  * https://github.com/google/site-kit-wp/blob/develop/assets/js/googlesitekit/modules/create-module-store.js
  * https://github.com/google/site-kit-wp/blob/develop/assets/js/modules/analytics/datastore/index.js 

* Update any existing tests or files to that rely on previous exports from the updated files, including (but not limited to):
  * https://github.com/google/site-kit-wp/blob/develop/assets/js/googlesitekit/datastore/site/index.test.js
  * https://github.com/google/site-kit-wp/blob/develop/assets/js/googlesitekit/datastore/site/info.test.js
  * https://github.com/google/site-kit-wp/blob/develop/assets/js/googlesitekit/datastore/site/notifications.test.js

## QA Brief

* 

## Changelog entry

* N/A

aaemnnosttv:
> Update any existing tests or files to that rely on previous exports from the updated files, including

This part shouldn't be necessary as the exports from the index shouldn't change at all. This is only about refactoring how the store components are assembled _before_ exporting.

Otherwise, IB ✅

felixarntz:
I assigned all of you @jqlee85 @aaemnnosttv @tofumatt just because it would be good to get all of your thoughts on https://github.com/google/site-kit-wp/pull/1484#discussion_r420489638.

aaemnnosttv:
@felixarntz I think this should be good to go now. Assigning to you since you had previously requested changes, but also because it is not tagged with the release yet.

ShahAaron:
Develop pulled 
Tests run JS & PHP Unit, e2e & Visual Regression - passed

Code checked - Data.combineStores added - passed 
A/C's - passed

QA ✅

PR #1484, opened 2020-05-02
Enhancement/1476 refactor datastores combinestores utility

jqlee85:
## Summary

Addresses issue [#1476](https://github.com/google/site-kit-wp/issues/1476)

## Relevant technical choices

Convert following files to use Data.combineStores()
* https://github.com/google/site-kit-wp/blob/develop/assets/js/googlesitekit/datastore/site/index.js
* https://github.com/google/site-kit-wp/blob/develop/assets/js/googlesitekit/modules/create-module-store.js
* https://github.com/google/site-kit-wp/blob/develop/assets/js/modules/analytics/datastore/index.js
* Removes initialize logic, which was not functioning as expected

## Checklist


jqlee85:
@aaemnnosttv heads up, this is ready for review. I haven't moved the issue into CR yet, because the issue is waiting on IB review as well.

tofumatt:
@googlebot I consent.

assets/js/googlesitekit/datastore/site/index.test.js
```diff
@@ -25,9 +25,11 @@ import apiFetch from '@wordpress/api-fetch';
  * Internal dependencies
  */
 import { createTestRegistry, unsubscribeFromAll } from 'tests/js/utils';
-import { INITIAL_STATE } from './index';
+import { default as datastore } from './index';
```
felixarntz, line 28: I think this can be simplified to `import datastore from './index';`? Maybe let's name it `store` for consistency with the other ones.

assets/js/googlesitekit/datastore/site/info.test.js
```diff
@@ -25,9 +25,12 @@ import {
 	subscribeUntil,
 	unsubscribeFromAll,
 } from 'tests/js/utils';
-import { INITIAL_STATE } from './index';
+// import { INITIAL_STATE } from './index';
```
felixarntz, line 28: This comment should be removed.

assets/js/googlesitekit/modules/create-module-store.js
```diff
@@ -66,37 +66,21 @@ export const createModuleStore = ( slug, {
...
-		} );
+		} )
+		: {};
+
+	const combinedStore = Data.combineStores(
+		notificationsStore,
+		settingsStore,
+	);
+
+	const STORE_NAME = settingsStore.STORE_NAME ? [ notificationsStore.STORE_NAME ] : [ notificationsStore.STORE_NAME, settingsStore.STORE_NAME ];
```
felixarntz, line 81: Should this condition not be inverted?

assets/js/googlesitekit/modules/create-module-store.js
```diff
@@ -66,37 +66,21 @@ export const createModuleStore = ( slug, {
...
 
-		STORE_NAME.push( settingsStore.STORE_NAME );
-		INITIAL_STATE.push( settingsStore.INITIAL_STATE );
-		actions.push( settingsStore.actions );
-		controls.push( settingsStore.controls );
-		reducer.push( settingsStore.reducer );
-		resolvers.push( settingsStore.resolvers );
-		selectors.push( settingsStore.selectors );
-	}
+	combinedStore.STORE_NAME = Data.collectName( ...STORE_NAME );
```
felixarntz, line 83: We could potentially get rid of this `collectName` logic by setting the `storeName` default ourself on top of this function (e.g. in line 64 do `storeName = storeName || `modules/${ slug }`;`). This will ensure we pass that value to `createNotificationsStore` and `createSettingsStore` and it will be returned just like that, so we already know it's going to be the same name.

assets/js/googlesitekit/modules/datastore/index.js
```diff
@@ -21,49 +21,15 @@
...
-	)
-);
-
-export const controls = Data.collectControls(
-	Data.commonControls,
-	modules.controls,
-);
-
-export const reducer = Data.addInitializeReducer(
-	INITIAL_STATE,
```
felixarntz, line 45: Seeing this, maybe we missed something in #1455 - I didn't spot that before. We're now no longer passing this `INITIAL_STATE` to the `initializeReducer`, so how is that going to know the initial state? It now receives an empty object in https://github.com/google/site-kit-wp/pull/1455/files#diff-f22c2d5fdacaede6124fc9b262153a4bR343. Maybe we'll need to modify our approach here slightly.

Should `Data.combineStores` always add the `Data.commonStore` parts to it? If we do that, we'd need to ensure these specific items are ignored when checking for duplicates since they'd be included even on "hierarchically nested" calls to `Data.combineStores`. I think that'd be okay though since it would simply ensure the `Data.commonStore` pieces are included once in every store.

Or we remove the whole `initialize` logic entirely. I don't think we're using it anywhere, and while it's a good idea, we can revisit later. Maybe it was a bit premature.

What do you think @aaemnnosttv @tofumatt?

assets/js/googlesitekit/datastore/site/info.test.js
jqlee85: Ah yes, thanks for the catch.

assets/js/googlesitekit/datastore/site/index.test.js
jqlee85: Ah yes. I used `dataStore` because `store` was used in the inner scope. Bit since we only need it to get `INITIAL_STATE`, we should be able to override it.

assets/js/googlesitekit/modules/create-module-store.js
jqlee85: Ah, yes it should be inverted, thanks!

assets/js/googlesitekit/modules/create-module-store.js
jqlee85: Yeah that makes sense. I've updated to simplify the storeName logic.

assets/js/googlesitekit/modules/datastore/index.js
aaemnnosttv: > Should `Data.combineStores` always add the `Data.commonStore` parts to it?

I don't think so. Having a `Data.commonStore` to include in the combination if needed is quite minimal as it is, so I don't think it's worth doing this by default (at least for this utility).

> If we do that, we'd need to ensure these specific items are ignored when checking for duplicates since they'd be included even on "hierarchically nested" calls to `Data.combineStores`. I think that'd be okay though since it would simply ensure the `Data.commonStore` pieces are included once in every store.

@tofumatt and I talked about this potentially coming up and the idea of allowing for specific things to be duplicated. I'm not a huge fan of the idea though. Maybe we should just remove the whole error-on-duplicate-key thing as I'm not sure the benefits outweigh the compromises.

> Or we remove the whole `initialize` logic entirely. I don't think we're using it anywhere, and while it's a good idea, we can revisit later. Maybe it was a bit premature.

I thought we were using it with the reset functionality, but if we're not then let's get rid of it. If we've made it this far without needing it, it's probably safe to let it go.

felixarntz: @aaemnnosttv I think our checking for duplicates there is quite useful as it's easy to name selectors/actions the same across different parts of a store - maybe at some point we'd combine them differently and would have forgotten about a potential conflict, so I'd prefer we kept that logic.

> I thought we were using it with the reset functionality, but if we're not then let's get rid of it. If we've made it this far without needing it, it's probably safe to let it go.

You're right, we use it during reset. However, the usage there is neither really accurate nor needed. It's not accurate because it only resets that single store, and not necessary since we always reload the entire page anyway. I think adding that there was also a bit premature - let's drop the entire initialization logic and remove references to it. Could you take care of that @jqlee85?

jqlee85: @felixarntz I've removed the initialization logic and references to it. Let me know if this looks right. Just want to make sure it's not having implications I'm not aware of.

assets/js/googlesitekit/datastore/site/index.test.js
```diff
@@ -25,13 +25,15 @@ import apiFetch from '@wordpress/api-fetch';
...
+import store from './index';
 import { STORE_NAME } from './constants';
 
+const { INITIAL_STATE } = store;
+
 describe( 'core/site store', () => {
 	let apiFetchSpy;
 	let registry;
-	let store;
+	let store; // eslint-disable-line no-shadow
```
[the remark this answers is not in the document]
jqlee85: Yep, makes sense. I've made these updates.

assets/js/googlesitekit/datastore/site/info.test.js
```diff
@@ -25,9 +25,11 @@ import {
 	subscribeUntil,
 	unsubscribeFromAll,
 } from 'tests/js/utils';
-import { INITIAL_STATE } from './index';
+import store from './index';
 import { STORE_NAME } from './constants';
 
+const { INITIAL_STATE } = store;
+
```
aaemnnosttv, line 32: Please restore to the previous usage now that the exports are available.

assets/js/googlesitekit/datastore/site/notifications.test.js
```diff
@@ -29,6 +29,6 @@ describe( 'core/site notifications', () => {
 			'fetchNotifications',
 			'receiveNotifications',
 		];
-		expect( Object.keys( actions ) ).toEqual( expect.arrayContaining( actionsToExpect ) );
+		expect( Object.keys( store.actions ) ).toEqual( expect.arrayContaining( actionsToExpect ) );
```
aaemnnosttv, line 32: Please restore to the previous usage now that the exports are available.

assets/js/googlesitekit/modules/datastore/index.test.js
```diff
@@ -49,7 +49,7 @@ describe( 'core/modules store', () => {
 		it( 'has the appropriate initial state', () => {
 			const state = store.getState();
 
-			expect( state ).toEqual( INITIAL_STATE );
+			expect( state ).toEqual( dataStore.INITIAL_STATE );
```
aaemnnosttv, line 52: Please restore to the previous usage now that the exports are available.

assets/js/modules/analytics/datastore/index.js
```diff
@@ -42,71 +42,15 @@ const baseModuleStore = Modules.createModuleStore( 'analytics', {
...
-export const resolvers = Data.collectResolvers(
-	baseModuleStore.resolvers,
-	accounts.resolvers,
-	properties.resolvers,
-	profiles.resolvers,
-	settings.resolvers,
-	tags.resolvers,
-);
-
-export const selectors = Data.collectSelectors(
```
aaemnnosttv, line 93: Please restore these exports similar to the others.

jqlee85: :+1:

Issue #1479, opened 2020-05-01
Do not request proxy notifications if proxy is not being used


felixarntz:
## Bug Description

Even when the proxy is not in use the `core/site/data/notifications` endpoint tries to make a request to the proxy, which will always fail obviously because there is no client ID.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `core/site/data/notifications` endpoint should only issue a request to the proxy if it is actually in use.

## Implementation Brief

* Add `Authentication` instance as optional third parameter to `Notifications::__construct()`, and set it as class property. If not provided, create from `$context` (only to simplify tests, our production code must _always_ provide it).
* Pass the `Authentication` instance `$authentication` as third parameter when instantiating `Notifications` in the main `Plugin` class.
* Modify `core/site/data/notifications` route logic in `Notifications` class to check `$this->authentication->get_oauth_client()->using_proxy()`. If not, immediately return with an empty array `WP_REST_Response`.
* Nice side-effect to implement here too: Since the `Notifications` class now receives an `Authentication` instance, let's replace manual instantiation of the `Credentials` object in the `Notifications` constructor with `$this->authentication->credentials()`.

## QA Brief

* 

## Changelog entry

* Do not request remote notifications if the site is not connected to the remote.

aaemnnosttv:
IB ✅

cole10up:
**Tested**

Ran a series of tests with the production ready SK 1.8.0 zip. Installed and activated SK. Activated modules and performed a series of tests within the proxy around exiting TOS according to this ticket: https://github.com/google/site-kit-wp/issues/1213

Removed and added modules from settings page and the dashboard after initial login. Was unable to experience any issues on my end.

Passed QA ✅

PR #1487, opened 2020-05-04
Added authentication to notification class

ShahAaron:
## Summary

Addresses issue #1479 

## Checklist


includes/Core/Notifications/Notifications.php
```diff
@@ -129,6 +140,10 @@ private function get_rest_routes() {
 								return new WP_Error( 'exception', $e->getMessage() );
 							}
 
+							if ( $this->authentication->get_oauth_client()->using_proxy() ) {
```
felixarntz, line 143: This should be the opposite.

ShahAaron: Corrected

tests/e2e/specs/dashboard/notifications.test.js
```diff
@@ -18,56 +18,92 @@ const goToSiteKitDashboard = async () => {
...
-	} );
-
-	it( 'displays core site notifications on the main dashboard', async () => {
-		// Add the test notification (by default there are none).
-		await wpApiFetch( {
-			path: 'google-site-kit/v1/e2e/core/site/notifications',
-			method: 'post',
-			data: testSiteNotification,
+	// The proxy test cannot currently be done and needs to be skipped TODO tests need to to be fixed to handle proxy tests.
+	describe.skip( 'when using proxy', () => {
```
felixarntz, line 22: Can you clarify why we cannot currently test this? I'd think we can mock the API response? We should be able to simulate the proxy being used by filtering in proxy credentials. Thoughts @aaemnnosttv?

includes/Core/Notifications/Notifications.php
```diff
@@ -47,6 +48,14 @@ class Notifications {
 	 */
 	private $options;
 
+	/**
+	 * Authentication instance.
+	 *
+	 * @since 1.4.0
```
felixarntz, line 54: This should say `@since n.e.x.t`.

tests/e2e/specs/dashboard/notifications.test.js
aaemnnosttv: We don't currently have the infrastructure in place to fake proxy authentication in E2E. We could add it here, but I thought it would be a bit out-of-scope for this issue and is actually a symptom of a larger problem. I created an issue for it here: https://github.com/google/site-kit-wp/issues/1503

Issue #1470, opened 2020-04-29
Fix Analytics store bug with empty array API responses


felixarntz:
## Bug Description

The properties and profiles parts of the `modules/analytics` datastore currently rely on the response items to key them in the datastore state. This isn't really necessary because the individual array items in a single response will _always_ apply to the same account ID / property ID because of how the API works. Furthermore, relying on the response results in a bug when an empty array is returned - in this case there's no item to look at, so the datastore's state property will not be filled, leaving the state in the same way it would have been if the data had not been loaded yet at all.

This is a follow-up to #1269, hence high priority to also have in the coming release.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* `receiveProperties` and `receiveProfiles` (plus their reducer cases) in the `modules/analytics` datastore must not rely on response items to determine the key in the datastore state, but instead rely on the original parameter that the data was requested with, to ensure this also works for empty arrays where there's nothing to look at.

## Implementation Brief

* Instead of using `groupBy` based on response items, add an additional object with the respective parameter to the `receive...` signature (`receiveProperties( properties, { accountID } )` and `receiveProfiles( profiles, { propertyID } )`), and rely on that value in the reducer for the key.
* Nesting in a parameters object is strictly not necessary here, but makes sense to prepare for #1288, which will do that throughout, internally.
* A similar fix was recently done in the AdSense refactoring PR (only difference is that there it was for something called clients, which is almost similar to Analytics properties), so basically we need to do exactly the same here: https://github.com/google/site-kit-wp/pull/1410/commits/637da8414b701980cfa85beba0bf2ba9471e8f63
* All calls to these two `receive...` functions need to be updated to still work, including Jest tests and Storybook. Having to pass the parameter makes this more explicit, but a lot of other code already does that, and it's necessary here too (see AdSense Storybook example: https://github.com/google/site-kit-wp/pull/1410/commits/2ef371b4deb2fe2b4079f0389faab7f654246160)

## QA Brief

* 

## Changelog entry

* N/A

aaemnnosttv:
IB ✅

aaemnnosttv:
@felixarntz there are a few exceptions where this bug is not 100% fixable just by passing along params.

@ShahAaron brought it to my attention that there are two cases where we run into the same problem:
* `fetchAccountsPropertiesProfiles`
* `fetchPropertiesProfiles`

[...]

PR #1488, opened 2020-05-04
Bug/1470 analytics store bug

ShahAaron:
## Summary

Addresses issue #1470 

## Checklist


assets/js/modules/analytics/common/account-select.test.js
```diff
@@ -114,9 +114,11 @@ describe( 'AccountSelect', () => {
 	it( 'should pre-select the property and profile IDs when changed', () => {
 		const { accounts, properties, profiles } = fixtures.accountsPropertiesProfiles;
 		const { getByText, container, registry } = render( <AccountSelect />, { setupRegistry } );
+		const propertyID = fixtures.accountsPropertiesProfiles.properties[ 0 ].id;
+		const accountID = fixtures.accountsPropertiesProfiles.properties[ 0 ].accountId;
```
felixarntz, line 118: Nit-picking, but this can be simplified to reference `properties[ 0 ]` since that is already retrieved 2 lines above.

assets/js/modules/analytics/common/profile-select.test.js
```diff
@@ -38,28 +38,30 @@ apiFetchMock.mockImplementation( ( ...args ) => {
 
 const setupRegistry = ( { dispatch } ) => {
 	const { id, webPropertyId, accountId } = fixtures.propertiesProfiles.profiles[ 0 ];
+	const propertyID = webPropertyId;
```
felixarntz, line 41: It feels a bit strange to re-assign that here, as that line itself has no context on why it's reassigned. I'd prefer if we could pass `{ propertyID: webPropertyId }` below instead (that's how we do that elsewhere too).

This also applies to other similar re-assignments below - let's rather pass the values within the objects as applicable, since then there is more clear context on why this "rename" / "reassignment" is necessary (because of an object expecting specific properties).

assets/js/modules/analytics/common/profile-select.test.js
```diff
@@ -38,28 +38,30 @@ apiFetchMock.mockImplementation( ( ...args ) => {
...
-	dispatch( STORE_NAME ).receiveProfiles( fixtures.propertiesProfiles.profiles );
+	dispatch( STORE_NAME ).receiveProfiles( fixtures.propertiesProfiles.profiles, { propertyID } );
 };
 
 const setupRegistryWithExistingTag = ( { dispatch } ) => {
 	const existingTag = {
 		accountID: fixtures.accountsPropertiesProfiles.profiles[ 0 ].accountId,
 		propertyID: fixtures.accountsPropertiesProfiles.profiles[ 0 ].webPropertyId,
 	};
+	const propertyID = existingTag.propertyID;
```
felixarntz, line 53: See above.

assets/js/modules/analytics/common/profile-select.test.js
```diff
@@ -38,28 +38,30 @@ apiFetchMock.mockImplementation( ( ...args ) => {
...
 	dispatch( STORE_NAME ).setProfileID( id );
-	dispatch( STORE_NAME ).receiveProfiles( fixtures.accountsPropertiesProfiles.profiles );
+	dispatch( STORE_NAME ).receiveProfiles( fixtures.accountsPropertiesProfiles.profiles, { propertyID } );
 	dispatch( STORE_NAME ).receiveExistingTag( existingTag.propertyID );
 };
 
 const setupEmptyRegistry = ( { dispatch } ) => {
 	dispatch( STORE_NAME ).setSettings( {} );
-	dispatch( STORE_NAME ).receiveProfiles( [] );
+	dispatch( STORE_NAME ).receiveProfiles( [], {} );
```
felixarntz, line 64: This doesn't really work - let's pass any property ID here so that it is valid. Since we don't pass anything here, `propertyID` will be `undefined`, resulting in the empty array not being assigned correctly.

assets/js/modules/analytics/common/property-select.test.js
```diff
@@ -53,7 +54,7 @@ const setupRegistryWithExistingTag = ( { dispatch } ) => {
 
 const setupEmptyRegistry = ( { dispatch } ) => {
 	dispatch( STORE_NAME ).setSettings( {} );
-	dispatch( STORE_NAME ).receiveProperties( [] );
+	dispatch( STORE_NAME ).receiveProperties( [], {} );
```
felixarntz, line 57: See above.

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -61,8 +61,16 @@ export const actions = {
...
 			yield actions.receiveAccounts( response.accounts );
-			dispatch( STORE_NAME ).receiveProperties( response.properties );
-			dispatch( STORE_NAME ).receiveProfiles( response.profiles );
+
+			if ( response.properties.length && response.properties[ 0 ] && response.properties[ 0 ].accountId ) {
+				const accountID = response.properties[ 0 ].accountId;
+				dispatch( STORE_NAME ).receiveProperties( response.properties, { accountID } );
+			}
+
+			if ( response.properties.length && response.properties[ 0 ] && response.properties[ 0 ].id ) {
```
felixarntz, line 70: I'd suggest we use `response.profiles[ 0 ].webPropertyId` here instead, to make sure it's based on the profiles retrieved. The value _should_ be the same, but I think it's a bit more accurate like that.

assets/js/modules/analytics/datastore/profiles.js
```diff
@@ -162,11 +161,11 @@ export const actions = {
 		};
 	},
 
-	receiveProfiles( profiles ) {
+	receiveProfiles( profiles, { propertyID } ) {
 		invariant( Array.isArray( profiles ), 'profiles must be an array.' );
```
felixarntz, line 165: Can we add an `invariant` call here to ensure `propertyID` is not `undefined`? This ensures that we always have to pass it, which we should (see other comments on empty objects in tests above).

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -117,13 +116,17 @@ export const actions = {
...
 			const { dispatch } = yield Data.commonActions.getRegistry();
-			yield actions.receiveProperties( properties );
-			dispatch( STORE_NAME ).receiveProfiles( profiles );
+			yield actions.receiveProperties( properties, { accountID } );
 
 			if ( matchedProperty ) {
 				yield actions.receiveMatchedProperty( matchedProperty );
 			}
 
+			if ( properties.length && properties[ 0 ] && properties[ 0 ].id ) {
```
felixarntz, line 125: See other comment, I think this would be more accurate relying on `profiles[ 0 ].webPropertyId` instead.

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -240,13 +243,14 @@ export const actions = {
 	 * @private
 	 *
 	 * @param {Array} properties Properties to add.
+	 * @param {Object} accountID Account ID to add.
 	 * @return {Object} Redux-style action.
 	 */
-	receiveProperties( properties ) {
+	receiveProperties( properties, { accountID } ) {
 		invariant( Array.isArray( properties ), 'properties must be an array.' );
```
felixarntz, line 250: See other comment, let's also add an `invariant` call here to ensure `accountID` is passed.

assets/js/modules/analytics/settings/settings-edit.test.js
```diff
@@ -47,10 +47,12 @@ describe( 'SettingsEdit', () => {
 			const { accounts, properties, profiles } = fixtures.accountsPropertiesProfiles;
 			existingTag.accountID = profiles[ 0 ].accountId;
 			existingTag.propertyID = profiles[ 0 ].webPropertyId;
+			const accountID = existingTag.accountID;
+			const propertyID = existingTag.propertyID;
```
felixarntz, line 51: See other similar comments, I think this would be easier to understand if the assignment happened in the `receive...` calls below. Alternatively, this can be simplified to `const { accountID, propertyID } = existingTag`.

assets/js/modules/analytics/common/account-select.test.js
ShahAaron: Fixed

assets/js/modules/analytics/common/profile-select.test.js
ShahAaron: Fixed

assets/js/modules/analytics/common/profile-select.test.js
ShahAaron: Fixed

assets/js/modules/analytics/common/profile-select.test.js
ShahAaron: Fixed

assets/js/modules/analytics/common/property-select.test.js
ShahAaron: Fixed

assets/js/modules/analytics/datastore/properties.js
ShahAaron: Fixed

assets/js/modules/analytics/datastore/profiles.js
ShahAaron: Fixed

assets/js/modules/analytics/settings/settings-edit.test.js
ShahAaron: Fixed

assets/js/modules/analytics/datastore/accounts.js
ShahAaron: @felixarntz based on a conversation with @aaemnnosttv it was felt that as we are getting the `accountID` from `response.properties[ 0 ].accountId` that in order to insure we get the correct `propertyID` we should get it from the same place ( `response.properties[ 0 ].id` ). Happy to make the change to `response.profiles[ 0 ].webPropertyId` if you think that is the correct way :)

assets/js/modules/analytics/datastore/properties.js
ShahAaron: As above for `accounts.js` I'm happy to make the change if that is the correct thing to do :)

assets/js/modules/analytics/datastore/accounts.js
felixarntz: @ShahAaron Yeah I think it would be better to rely on the property ID via `profiles` since those with 100% certainty contain the correct property ID. We cannot assume that the first property returned is also the exact property that profiles were requested for, so this is actually problematic.

assets/js/modules/analytics/datastore/properties.js
felixarntz: Yep, that is more accurate, the current logic will likely result in errors as soon as you have more than one property.

assets/js/modules/analytics/datastore/accounts.js
ShahAaron: Fixed

assets/js/modules/analytics/datastore/properties.js
ShahAaron: Fixed

Issue #1459, opened 2020-04-27
Build and deploy storybook static files to GitHub pages for develop and master


felixarntz:
## Feature Description

We should make Storybook available on google.github.io/site-kit-wp.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* When pushing to `develop` or `master`, Storybook static files should be built and written to the `gh-pages` branch.
* In the `gh-pages` branch, a folder should be added per branch (e.g. so that the URLs will end up like `https://google.github.io/site-kit-wp/storybook/develop` and `https://google.github.io/site-kit-wp/storybook/master`).
* This must **not** happen for pull-requests.

## Implementation Brief

There is an easy-to-use Travis-CI integration for deploying to GitHub pages (which e.g. the Web Stories plugin is using: https://github.com/google/web-stories-wp/blob/d181a4ea75cff53db730920fb41c41e2624a0c96/.travis.yml#L130-L141). However, that appears not to support deploying only partial content - you can specify a `local_dir`, but not a `remote_dir`. Hence, using a GitHub action workflow makes more sense. A lot of the necessary code can be taken from the existing `build-pr.yml` workflow.

* Add `npm run build:storybook` command that stores storybook in a `storybook-static` folder (ignore that folder in git).
* Add `build-storybook.yml` workflow that runs `npm run build:storybook`, then uploads the files as artifacts, then downloads it and integrates them into the `gh-pages` branch (in a `storybook/{$branch_name}` folder) and commits.
* Merge #1457.

## QA Brief

* 

## Changelog entry

* N/A

aaemnnosttv:
This looks good to me @felixarntz - my only question is around the use of GitHub pages for the repo. You probably thought about this but it seems unlikely that Storybook is the only thing we might want to build like this to make available on the web in the future (e.g. API documentation).

* Idea 1: deploy the files to a sub-directory of the pages repo, so it would use `google.github.io/site-kit-wp/storybook` as a base URL instead
* Idea 2: Not use GitHub Pages at all. Netlify might be a really nice fit here and wouldn't even require any custom GitHub actions. Basically, we'd connect the repo to a new (free) site on Netlify and configure the build command and target directory to use for public files. It also has a feature called ["Deploy Previews"](https://www.netlify.com/blog/2016/07/20/introducing-deploy-previews-in-netlify/) which handle the branch/PR use-case really nicely.
[...]

PR #1509, opened 2020-05-06
Fix/1459 storybook assets

felixarntz:
## Summary

Addresses issue #1459

Follow-up to #1457 

## Relevant technical choices

* Use relative `./assets` instead of root-relative `/assets/` as `googlesitekit.admin.assetsRoot`.
* Build static images and SVGs into the `storybook-static` directory so that it works standalone.

See https://felixarntz.github.io/site-kit-wp/storybook/develop/ for demo (using these commits).

## Checklist


felixarntz:
@aaemnnosttv 🤦 Made this change, looks like that did it on https://felixarntz.github.io/site-kit-wp/storybook/develop/

aaemnnosttv:
@felixarntz I noticed that we can avoid the need for an extra `build:storybook-static` command by using `dist` as the output directory for `build-storybook`. Since this command cleans the output directory first (`rimraf` is actually unnecessary here), we just need to run the normal `build:static` afterwards and then `dist` is the complete storybook build.

It's a minor optimization and not worth blocking this over but I'll create a follow-up PR here for your consideration.

PR #1513, opened 2020-05-06
Fix/1101 finishing touches

aaemnnosttv:
## Summary

Addresses issue #1101

## Relevant technical choices

* Removes class causing smaller font size in `TrackingExclusionSwitches`  
image
* Adds missing parent IDs to `receive*` calls in storybook

## Checklist

### DIFF
--- a/stories/module-analytics-settings.stories.js
+++ b/stories/module-analytics-settings.stories.js
@@ -152,7 +152,7 @@ storiesOf( 'Analytics Module/Settings', module )
 		const setupRegistry = ( { dispatch } ) => {
 			dispatch( STORE_NAME ).receiveExistingTag( null );
 			dispatch( STORE_NAME ).receiveAccounts( accounts );
-			dispatch( STORE_NAME ).receiveProperties( properties, { accountID: profiles[ 0 ].accountId } );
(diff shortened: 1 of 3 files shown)


stories/module-analytics-settings.stories.js
```diff
@@ -152,8 +152,8 @@ storiesOf( 'Analytics Module/Settings', module )
 		const setupRegistry = ( { dispatch } ) => {
 			dispatch( STORE_NAME ).receiveExistingTag( null );
 			dispatch( STORE_NAME ).receiveAccounts( accounts );
-			dispatch( STORE_NAME ).receiveProperties( properties );
-			dispatch( STORE_NAME ).receiveProfiles( profiles );
+			dispatch( STORE_NAME ).receiveProperties( properties, { accountID: profiles[ 0 ].accountId } );
```
felixarntz, line 155: Similar to my comment on the original PR, we should always look at the correct array here - for `receiveProperties` this should always be based on `properties`, and for `receiveProfiles` it should always be based on `profiles`. It's probably not a big deal for the data here, but this is complex enough, so I'd prefer if we used "correct" examples throughout.

PR #1514, opened 2020-05-07
Build and deploy storybook from dist

aaemnnosttv:
## Summary

Addresses issue #1459

## Relevant technical choices

@felixarntz I haven't tested this yet other than locally so it would be good to try with your fork

- Follow up to https://github.com/google/site-kit-wp/pull/1509#issuecomment-625103313
- Simplifies the storybook build and deployment by using the default output directory for our built assets, omitting the need for additional build commands

* `build-storybook` automatically cleans the output directory (`dist`) first when it runs
* builds static storybook into `dist`
* Runs `build:static` to add svgs and images into `dist`
* Updates GitHub action to reference `dist` as the source/destination directory that was previously `storybook-static`

## Checklist


felixarntz:
@aaemnnosttv This should work - is there any side-effect though from including it in `dist`? For example we need to make sure this is never part of the ZIP bundle.

felixarntz:
@aaemnnosttv Unrelated to the original change here, but maybe we should fix https://github.com/google/site-kit-wp/runs/653791325?check_suite_focus=true here? I think we could wrap the whole `git add`, commit, push etc in a `if [[ `git status --porcelain` ]]; then` to prevent errors when there are no storybook changes. Maybe we should even add this to `build-pr.yml` as well (although the problem is much less likely to happen there)?

aaemnnosttv:
@felixarntz 

> This should work - is there any side-effect though from including it in dist? For example we need to make sure this is never part of the ZIP bundle.

I don't think there is any risk of that happening. All of our main build commands clear the `dist` directory as the first step, so even if storybook was in there and then you ran the `release-zip` command, it wouldn't be in the resulting zip because the directory would be deleted before performing the build. I can't think of any problems that could happen as part of this. Anything that goes in `dist` is the result of a build. Storybook is no different.

> I think we could wrap the whole git add, commit, push etc in a if [[ git status --porcelain ]]; then to prevent errors when there are no storybook changes. Maybe we should even add this to build-pr.yml as well (although the problem is much less likely to happen there)?

That's what this line is for:
https://github.com/google/site-kit-wp/blob/392fd967ba868ccf47f0002ee5e3a2fdee80df00/.github/workflows/remove-pr.yml#L19

That isn't in this branch but I prefer this over a wrapping conditional unless of course it's not working 😄

Issue #1524, opened 2020-05-07
Analytics adsenseLinked always false


aaemnnosttv:
## Bug Description

The special internal Analytics option for storing the linked state with AdSense (`adsenseLinked`) is always false, even after it is saved as `true`.

## Steps to reproduce

1. Setup Analytics
1. Setup AdSense in a connected state
1. Trigger Analytics `GET:report` with AdSense metrics
1. See Analytics `adsenseLinked` option is still `false`

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The Analytics setting for `adsenseLinked` should be `true` when the following criteria is met:
    - Analytics and AdSense are both active and set up
    - Analytics `GET:report` is requested with AdSense metrics

## Implementation Brief

* Sanitize `adsenseLinked` as a boolean in `Analytics\Settings::get_sanitize_callback` rather than always setting it to `false`
* Merge #1525  

## QA Brief

* 

## Changelog entry

* Fix bug where Analytics table would never include AdSense metrics even with a successful AdSense connection.

felixarntz:
IB ✅ 

(except PR is #1525) 😄

Alnetkhatiyan:
<script async src="https://pagead2.googlesyndication.com/pagead/js/adsbygoogle.js?client=ca-pub-1635336603446911"
     crossorigin="anonymous"></script> need to implement my code I can't do it

Alnetkhatiyan:
Please help for code implementation 
Regard

jamesozzie:
@Alnetkhatiyan Thanks for reaching out. We provide Site Kit initial support requests from the WordPress support forums. If you'd like to open a [new support topic](https://wordpress.org/support/plugin/google-site-kit/#new-post) there we'd be happy to assist. 

Note also that for assistance with setting up the plugin on your WordPress site there is [a guide on the plugin website](https://sitekit.withgoogle.com/documentation/install/). By following the guide you'll find there is no requirement to manually insert your ad code, the plugin can do this for you.

PR #1527, opened 2020-05-07
Really fix adsenseLinked bug

felixarntz:
## Summary

Addresses issue #1524

## Relevant technical choices

* Ensure adsenseLinked force filter is unhooked after it is hooked.
* Handle nested request parameters as arrays and cast them to make sure. This only applies to this one location causing the bug here. The original reason where this was introduced is https://github.com/google/site-kit-wp/pull/935/files#diff-1142b30534124ce88a43f90e0ba72e30L397 (parameters are cast to objects, but only top-level ones are, nested ones remain arrays).
* Ensure the `adsenseLinked` setting is modified for both successful and failing `ga:adsense...` requests (logic previously was in datapoint response handler which only is called on success).

## Checklist

### DIFF
--- a/includes/Modules/AdSense.php
+++ b/includes/Modules/AdSense.php
@@ -82,9 +82,18 @@ final class AdSense extends Module implements Module_With_Screen, Module_With_Sc
 			/**
 			 * Release filter forcing unlinked state.
 			 *
-			 * @see \Google\Site_Kit\Modules\Analytics\Settings::register
+			 * This is hooked into 'init' (default priority of 10), so that it
+			 * runs after the original filter is added.
+			 *
+			 * @see \Google\Site_Kit\Modules\Analytics::register()
+			 * @see \Google\Site_Kit\Modules\Analytics\Settings::register()
 			 */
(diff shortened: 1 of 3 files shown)


includes/Modules/AdSense.php
```diff
@@ -82,9 +82,18 @@ function( $data ) {
...
-			 * @see \Google\Site_Kit\Modules\Analytics\Settings::register
+			 * This is hooked into 'init' (default priority of 10), so that it
+			 * runs after the original filter is added.
+			 *
+			 * @see \Google\Site_Kit\Modules\Analytics::register()
+			 * @see \Google\Site_Kit\Modules\Analytics\Settings::register()
 			 */
-			remove_filter( 'googlesitekit_analytics_adsense_linked', '__return_false' );
+			add_action(
+				'init',
```
aaemnnosttv, line 92: We should probably hook this on `googlesitekit_init` which is still in the same action but should always run after everything else is loaded 👍 
```suggestion
				'googlesitekit_init',
```

aaemnnosttv:
@googlebot I consent.

Issue #1177, opened 2020-02-26
Introduce notification functionality to core/user store


felixarntz:
## Feature Description

This is a follow-up to #1175. It's also related to #1174 and #1176, although these could even be worked on at the same time (except maybe because of annoying merge conflicts).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `core/user` store should have an additional selector `getNotifications`:
    * Contrary to the `getNotifications()` of `core/site` (see #1174), this should not issue an API request for now (as there are no server-side user notifications yet). This should only look at JS notifications.
    * See #1110 / #1170 for typical properties to expect on notifications (probably doesn't matter much for implementation of this, but fyi).
* The `core/user` store should have actions `addNotification( notification )` and `removeNotification( id )`:
    * Both of these should only work client-side for now (just like the selector here).
    * All `notification` objects must include `id`, so that they can be removed using that.

## Implementation Brief

* Add the actions, selectors, etc. from  `createNotificationsStore( 'core', 'user', 'notifications', { server: false } )` to the existing `core/user` datastore.
* Integrate the revised client/server-only notification store from the WIP PR at #1330

## QA Brief

* 

## Changelog entry

* Add notifications functionality to `core/user` store.

tofumatt:
Had a think about this, and rather than adding similar selections/actions/state to the `core/user` store as what we expose in `createNotificationsStore`, I think the best option is to add two additional options to `createNotificationsStore`'s `options` argument: `client` and `server`. Both would be enabled by-default, but toggling either off would allow for modules to opt-out of either server or client notifications.

The code changes would be extremely minimal. We'd initialise `INITIAL_STATE` like so:

```js
const INITIAL_STATE = {
		serverNotifications: server ? undefined : null,
		// Initialize clientNotifications as undefined rather than an empty
		// object so we can know if a client notification was added and then
		// removed from state.
		clientNotifications: client ? undefined : null,
		isFetchingNotifications: false,
	};
```

and only enable the `getNotifications` resolver if `server` is `true`. That way, we can add a notifications store to `core/user` by using: `createNotificationsStore( 'core', 'user', 'notifications', { server: false } )` and not have to write any new selectors. We'll need a few new tests to cover the new options, but it'll be much more extensible and DRY. 😄

felixarntz:
[...]

PR #1528, opened 2020-05-08
Feature/1177 notification to core user

ShahAaron:
## Summary

Addresses issue #1177 

## Checklist


ShahAaron:
@googlebot I consent.

tofumatt:
@googlebiot I consent.

assets/js/googlesitekit/data/create-notifications-store.js
```diff
@@ -63,11 +67,11 @@ export const createNotificationsStore = ( type, identifier, datapoint, {
 	const STORE_NAME = storeName || `${ type }/${ identifier }`;
 
 	const INITIAL_STATE = {
-		serverNotifications: undefined,
+		serverNotifications: server ? undefined : null,
```
felixarntz, line 70: Why is this `null`? I think this can always be `undefined`, regardless of `server`. If server notifications are disabled, this is being ignored entirely I think? cc @tofumatt

assets/js/googlesitekit/data/create-notifications-store.js
```diff
@@ -63,11 +67,11 @@ export const createNotificationsStore = ( type, identifier, datapoint, {
...
 		// Initialize clientNotifications as undefined rather than an empty
 		// object so we can know if a client notification was added and then
 		// removed from state.
-		clientNotifications: undefined,
+		clientNotifications: client ? undefined : null,
```
felixarntz, line 74: I think this should not be `null`, but an empty object if `client` is `false`. Otherwise, in the `getNotifications` selector, the code will always try to destructure `null`. cc @tofumatt

assets/js/googlesitekit/datastore/user/notifications.test.js
```diff
@@ -0,0 +1,32 @@
...
+ */
+import { actions } from './index';
+
+describe( 'core/user notifications', () => {
+	it( 'has appropriate notification methods', () => {
+		const actionsToExpect = [
+			'addNotification',
+			'removeNotification',
+		];
+		expect( Object.keys( actions ) ).toEqual( expect.arrayContaining( actionsToExpect ) );
```
felixarntz, line 30: It may be good to also check for existence of the `getNotifications` selector here.

tofumatt:
@googlebot I consent.

assets/js/googlesitekit/data/create-notifications-store.js
tofumatt: That's true, but I thought explicitly marking them as `null` was more "correct". We use `undefined`, generally, to indicate that something is loading/not-yet-available. By setting this to `null` explicitly, no code that checks for it being `undefined` would make any assumptions about the state of data or where it is in a notifications-request-flow.

I get that it doesn't _super_ matter here, because we don't load the resolver when `server` is false-y. But I thought it best to be as explicit as possible.

tofumatt: (That said… an empty object would be fine too. Maybe even better.)

assets/js/googlesitekit/data/create-notifications-store.js
tofumatt: Ah, good point. I'm good with that change 👍

assets/js/googlesitekit/data/create-notifications-store.js
felixarntz: Thanks @tofumatt, that makes sense. Let's go with an empty object since `null` is not a value this should ever have. Our `undefined` is for loading state, and `null` would work if this was actually an "object" - it's a map of objects and not a single object though (although of course technically an object), so the empty object is more appropriate here I think, since it's an empty map.

ShahAaron: Fixed

assets/js/googlesitekit/data/create-notifications-store.js
ShahAaron: Fixed

assets/js/googlesitekit/datastore/user/notifications.test.js
ShahAaron: Added

assets/js/googlesitekit/data/create-notifications-store.js
```diff
@@ -259,10 +263,23 @@ export const createNotificationsStore = ( type, identifier, datapoint, {
 
 	const resolvers = {
 		*getNotifications() {
-			yield actions.fetchNotifications();
+			try {
+				const notifications = yield actions.fetchNotifications();
+				return actions.receiveNotifications( notifications );
+			} catch ( err ) {
+				return actions.receiveNotificationsFailed();
+			}
```
felixarntz, line 271: This block here is old code and shouldn't be there. It can actually result in an error because `receiveNotificationsFailed` does not exist. It should be reverted to what it was before (line 262).

ShahAaron: I have removed this block and reverted this back to yield actions.fetchNotifications();

Issue #1500, opened 2020-05-04
Scaffold `modules/pagespeed-insights` datastore


felixarntz:
## Feature Description

Similar to #1224, #1248, #1385, #1498 (but a lot simpler here), PageSpeed Insights should receive a `modules/pagespeed-insights` datastore.

Note that (as with the other refactorings) we'll initially focus on purely setup/settings. Since there is no PSI-specific data dependency in that, we technically don't need the datastore yet. So this issue is really about scaffolding an empty datastore (we'll need it as soon as we start refactoring the dashboard).

As part of this issue, we should add one minor change to the `core/site` store to prepare for refactoring PSI settings.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

### New store `modules/pagespeed-insights`
* Introduce datastore `modules/pagespeed-insights` that is registered on `googlesitekit.data` and exported from `assets/js/modules/pagespeed-insights/datastore`.
* Rename `assets/js/modules/pagespeed-insights/index.js` to `assets/js/modules/pagespeed-insights/index.legacy.js` and update relevant import references.
* Add new `assets/js/modules/pagespeed-insights/index.js` that for now simply imports from `./datastore`.
* Introduce JS entrypoint asset `googlesitekit-modules-pagespeed-insights.js` with similar handle in PHP. It should simply import from `./modules/pagespeed-insights`.
* The datastore should simply use `createModuleStore` but not expose any other selectors/actions.

### Existing store enhancements `core/site`
* Modify the existing `getAdminURL()` selector to become `getAdminURL( page, args )` (both parameters optional). If `page` is provided, the `adminURL` should receive a corresponding `page` query parameter. If `args` are provided, these should be added as query parameters too. If `adminURL` is `undefined` (not loaded), the entire selector should return `undefined`. This is now essentially our old `getSiteKitAdminURL( page, args )` function in a datastore way.

## Implementation Brief

* Create new directory `datastore` in `assets/js/modules/pagespeed-insights` and create `index.js` in it.
* Add new `assets/js/modules/pagespeed-insights/index.js` that for now simply imports from `./datastore`.
* Rename `assets/js/modules/search-pagespeed-insights/index.js` to `assets/js/modules/pagespeed-insights/index.legacy.js` and update relevant import references.
* Create `googlesitekit-modules-pagespeed-insights.js` in `assets/js` and import `./modules/pagespeed-insights` into it.
[...]

Issue #1426, opened 2020-04-23
Expand `modules/pagespeed-insights` datastore with selector to get performance data


tofumatt:
## Feature Description

Following up on #1500, the `modules/pagespeed-insights` store should have a selector to get a report (via REST API).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a new selector `getReport( url : string, strategy : string )`, that fetches its data via the `GET:pagespeed` PSI datapoint.
* See AdSense report functionality (https://github.com/google/site-kit-wp/pull/1248/files#diff-c0ad9b78d9856b70285f9e72745ca13a) for inspiration (although that partially still uses old fetch infrastructure conventions).

## Implementation Brief

* Add `INITIAL_STATE.reports = {}`
    - `state.reports` should be keyed by `${ strategy }::${ url }`  
    In the future this can be updated or converted to use `stringifyObject` if needed for more complex keys, but this should be good for now.
* Add `selectors.getReport( url, strategy )`
    - returns `state.reports[ ${ strategy }::${ url } ]`
* Add `resolvers.getReport()`
    - Calls `fetchGetReport` if selector returns `undefined`
* Add `controls[ FETCH_GET_REPORT ]` which calls `API.get( 'modules', 'pagespeed-insights', 'pagespeed', { strategy, url } )`
* Add Jest test coverage for behavior for selector, resolver, and control as we do for other datastore components

## QA Brief

* 

## Changelog entry

* Add `getReport` selector to `modules/pagespeed-insights` datastore to get UX reports.

felixarntz:
IB ✅ 

Let's leave the PHP datapoint unchanged for now so that we don't need to change anything in legacy JS code. The nice thing is that we can now in our new JS API look out for more consistent names, and then retroactively update the PHP side if it's worth it, once no old code is using that anymore.

aaemnnosttv:
@felixarntz I just revised the signature of the selector in the AC here to use required positional arguments instead of an object with defaults after discussing this with @tofumatt. See here for more info https://github.com/google/site-kit-wp/pull/1561#pullrequestreview-424348918

tofumatt:
Code review ✅  from me, just assigning to @aaemnnosttv for merge.

ShahAaron:
Tested

Confirmed 
* `INITIAL_STATE.reports = {}` - Confirmed added
* `state.reports` keyed by `${ strategy }::${ url }` - Confirmed added
[...]

PR #1539, opened 2020-05-10
Feature/1500 scaffold pagespeed insights datastore

jqlee85:
## Summary

Addresses issue #1500

## Relevant technical choices

* Added scaffold store for PageSpeed Insights
* Updated `getAdminURL` selector in `core/site` datastore to take parameters and add them to url using WordPress' `addQueryArgs()` utility.
* Added tests for new `getAdminURL` selector.

## Checklist


assets/js/googlesitekit/datastore/site/info.js
```diff
@@ -161,10 +162,23 @@ export const selectors = {
 	 * @param {Object} state Data store's state.
 	 * @return {(string|undefined)} This site's admin URL.
 	 */
-	getAdminURL: createRegistrySelector( ( select ) => () => {
+	getAdminURL: createRegistrySelector( ( select ) => ( state, page = undefined, args = {} ) => {
 		const { adminURL } = select( STORE_NAME ).getSiteInfo() || {};
 
-		return adminURL;
```
felixarntz, line 167: Rather than the slightly complex logic in lines 172ff, I think it would be simpler to return early here if `adminURL` is `undefined`.

assets/js/googlesitekit/datastore/site/info.js
```diff
@@ -161,10 +162,23 @@ export const selectors = {
...
+		// Prevent a page in args from overriding main page argument
+		const { page: extraPage, ...queryArgs } = args; // eslint-disable-line no-unused-vars
+
+		// Add query arguments to URL if supplied
+		const fullURL = ( adminURL && ( page || Object.keys( queryArgs ).length ) )
```
felixarntz, line 172: This may not have been formulated in the ACs precisely: The `args` object should only be considered when `page` is present, i.e. calling like `getAdminURL( undefined, { arg1: value1 } )` should have no effect.

assets/js/googlesitekit/datastore/site/info.js
```diff
@@ -161,10 +162,23 @@ export const selectors = {
 	 * @param {Object} state Data store's state.
```
felixarntz, line 162: The new parameters should be documented here.

assets/js/googlesitekit/datastore/site/info.js
```diff
@@ -161,10 +162,23 @@ export const selectors = {
...
 	 * @return {(string|undefined)} This site's admin URL.
 	 */
-	getAdminURL: createRegistrySelector( ( select ) => () => {
+	getAdminURL: createRegistrySelector( ( select ) => ( state, page = undefined, args = {} ) => {
```
felixarntz, line 165: I don't think we need a `page` default here, since it's `undefined` anyway.

assets/js/googlesitekit/datastore/site/info.js
jqlee85: Oh yeah, thanks!

assets/js/googlesitekit/datastore/site/info.js
jqlee85: Yeah, that makes sense. Now that I think about it, we can return early `adminURL` early if either `adminURL` or `page` are `undefined`, since we'll either be returning `undefined` or the URL without any additional args.

assets/js/googlesitekit/datastore/site/info.js
jqlee85: Ah, got it. I've updated the function and the unit test to cover that.

jqlee85:
@felixarntz thanks for the feedback! I'd actually already implemented the createTestRegistry stuff for #1426, so I brought those changes over and  updated `getAdminURL()` and the tests, let me know if those look good.

Also looks like we need someone's consent now for the CLA. Maybe that's @aaemnnosttv?

assets/js/googlesitekit/data/index.js
```diff
@@ -52,15 +43,6 @@ const Data = createRegistry();
...
 // developers can use them.
-Data.addInitializeAction = addInitializeAction;
-Data.addInitializeReducer = addInitializeReducer;
-Data.collectActions = collectActions;
-Data.collectControls = collectControls;
-Data.collectReducers = collectReducers;
-Data.collectResolvers = collectResolvers;
-Data.collectSelectors = collectSelectors;
-Data.collectState = collectState;
-Data.collectName = collectName;
```
felixarntz, line 63: Shouldn't these changes be part of #1484 instead?

assets/js/googlesitekit/data/utils.js
```diff
@@ -27,49 +27,6 @@ import invariant from 'invariant';
...
-			}
-
-			default: {
-				return { ...state };
-			}
-		}
-	};
-
-	return collectReducers( initialState, reducer, initializeReducer );
-};
```
felixarntz, line 72: See above.

assets/js/googlesitekit/datastore/site/info.js
```diff
@@ -159,12 +160,29 @@ export const selectors = {
 	 * @since 1.7.0
 	 *
 	 * @param {Object} state Data store's state.
+	 * @param {string} page Page query argument to add to admin URL.
```
felixarntz, line 163: This should use `{(string|undefined)}` for the type, and per our (and WordPress) docs conventions there should be an "Optional. " at the beginning of the description. It would also be good to add something like "If not provided, the base admin URL is returned.".

assets/js/googlesitekit/datastore/site/info.js
```diff
@@ -159,12 +160,29 @@ export const selectors = {
...
+	 * @param {Object} args Additional query arguments to add to admin URL.
```
felixarntz, line 164: Similarly, this should use `{(Object|undefined)}`, and also provide the "Optional. ".

assets/js/googlesitekit/data/index.js
jqlee85: Ah yeah, sorry about that, I inadvertently merged them in when troubleshooting a build issue. I'll remove the extra changes.

assets/js/googlesitekit/datastore/site/info.js
jqlee85: :+1: thanks.

assets/js/googlesitekit/datastore/site/info.js
jqlee85: Thanks! Updated.

aaemnnosttv:
@tofumatt would you please add your consent comment here?

assets/js/googlesitekit/datastore/site/info.js
```diff
@@ -159,12 +160,29 @@ export const selectors = {
...
 		const { adminURL } = select( STORE_NAME ).getSiteInfo() || {};
 
-		return adminURL;
+		// Return adminURL if undefined, or if no page supplied
+		if ( typeof adminURL === 'undefined' || typeof page === 'undefined' ) {
+			return adminURL;
+		}
+
+		// Prevent a page in args from overriding main page argument
+		const { page: extraPage, ...queryArgs } = args; // eslint-disable-line no-unused-vars
```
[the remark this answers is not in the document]
felixarntz: Good point about adding the comment, this will clarify it. The logic doesn't need to be changed, having `page` as the first query parameter makes sense.

assets/js/googlesitekit/datastore/site/info.test.js
```diff
@@ -74,6 +74,64 @@ describe( 'core/site site info', () => {
...
+
+			it( 'returns the adminURL with page query parameter if the page argument is supplied', async () => {
+				global[ baseInfoVar ] = baseInfo;
+				global[ entityInfoVar ] = entityInfo;
+
+				expect( global[ baseInfoVar ] ).not.toEqual( undefined );
+				expect( global[ entityInfoVar ] ).not.toEqual( undefined );
+
+				const adminURL = registry.select( STORE_NAME ).getAdminURL( 'testpage' );
+				expect( adminURL ).toEqual( 'http://something.test/wp-admin?page=testpage' );
```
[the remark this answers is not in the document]
felixarntz: @aaemnnosttv Good catch! I think we should couple this function to `admin.php`, at least by default. We could do the following:
* If you call `getAdminURL( 'googlesitekit-dashboard' )`, you get `{adminURL}/admin.php?page=googlesitekit-dashboard`.
* If you call `getAdminURL( 'admin.php?page=googlesitekit-dashboard' )`, you get the same result.

We would in ~99% of cases use the first since it's a lot simpler, but we would support more custom cases where maybe we want to link to somewhere other than Site Kit pages.

We could use a check for something like `.php?page=` in the `page` parameter to check whether it's the simple format or the "full" format. If someone provides a full path, we should also ensure that it's concatenated with the base admin URL properly, with a single trailing slash.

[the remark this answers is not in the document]
felixarntz: cc @jqlee85

assets/js/googlesitekit/datastore/site/info.js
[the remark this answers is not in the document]
jqlee85: Yeah, I figured we wanted the page to be first rather than tacked on to the end, hence this (admittedly slightly odd) approach.
@felixarntz @aaemnnosttv Just to clarify, I should leave L176 as is, and update the comment at L175 to: `// Since page should be first query arg, create queryArgs without 'page' to prevent a 'page' in args from overriding it`?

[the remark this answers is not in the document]
felixarntz: @jqlee85 Sounds good to me (just with a period at the end)!

assets/js/googlesitekit/datastore/site/info.js
```diff
@@ -159,12 +160,29 @@ export const selectors = {
...
+
+		// Add query arguments to URL if supplied
```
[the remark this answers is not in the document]
jqlee85: :+1:

felixarntz:
@jqlee85 Note that there are some (easy-to-resolve) merge conflicts since the `modules/search-console` store was just merged which touches similar parts of the codebase.

assets/js/modules/pagespeed-insights/index.legacy.js
```diff
@@ -0,0 +1,70 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * External dependencies
+ */
+
```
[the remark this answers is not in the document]
jqlee85: :+1:

assets/js/googlesitekit/datastore/site/info.js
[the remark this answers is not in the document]
jqlee85: Oh, just now seeing the convo below that may affect this. 😆

assets/js/googlesitekit/datastore/site/info.test.js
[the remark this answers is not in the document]
aaemnnosttv: @felixarntz just to clarify, it sounds like we're treating the first argument as a special value, which can contain a query string. If the first arg contains a query string, the first part would be used as the path, and the query string would be parsed and merged into the `queryArgs` to be applied (if any). If the first arg did not contain a query string, then the path would default to `admin.php` with the first arg as `queryArgs.page`, which would take precedence over a passed `page` key in the args object. Is that correct?

assets/js/googlesitekit/datastore/site/info.test.js
```diff
@@ -74,6 +74,64 @@ describe( 'core/site site info', () => {
 	} );
 
 	describe( 'selectors', () => {
+		describe( 'getAdminURL', () => {
+			it( 'returns the adminURL on its own if no page argument is supplied', async () => {
+				global[ baseInfoVar ] = baseInfo;
+				global[ entityInfoVar ] = entityInfo;
```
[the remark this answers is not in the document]
jqlee85: :+1:

felixarntz:
@jqlee85 Something went wrong here somehow, I think it may be because of how the code from #1484 at some point had accidentally gotten mixed up with this here. After merging `develop` back into this, a lot of the (correct) changes from there were somehow reverted back, I think potentially because they had been reverted here, so that took precedence.

I think I fixed it all now, so this should be good to go (also my comment from earlier around `createModuleStore` was only because of these merge issues).

assets/js/googlesitekit/datastore/site/info.test.js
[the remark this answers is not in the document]
felixarntz: Yep 👍

Issue #1490, opened 2020-05-04
Move using_proxy to Credentials class


aaemnnosttv:
## Feature Description

Currently the `using_proxy` method is part of the `OAuth_Client` class.

The `OAuth_Client` class is only exposed through the `Authentication` class.

Since logic for `using_proxy` is entirely dependent on `Credentials` it makes sense to move this method to `Credentials`.

Current references to this method and what it would become:
```diff
- Authentication->OAuth_Client->using_proxy()
+ Authentication->Credentials->using_proxy()
```

```diff
- OAuth_Client->using_proxy()
+ OAuth_Client->Credentials->using_proxy()
```

```diff
- Debug_Data->Authentication->OAuth_Client->using_proxy()
+ Debug_Data->Credentials->using_proxy()
```

This helps flatten the chain to access this common data and no longer requires the `OAuth_Client` to be initialized to get it.

It also has the added benefit of making this data more readily available to other classes that already have access to `Credentials`.

E.g. in #1479 the Notifications class needs access to `using_proxy` - this becomes much more direct via `Credentials` which it already has an instance of.
```diff
- Notifications->Authentication->OAuth_Client->using_proxy()
+ Notifications->Credentials->using_proxy()
```

> Originally it seemed that `Authentication` would be the better place for this method but after closer evaluation, it seems to be a better fit on `Credentials`. This is a horizontal shift for classes that already access this method through `Authentication`.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `OAuth_Client::using_proxy` method is deprecated and calls `_deprecated_function` when called
	- Returns `OAuth_Client->Credentials->using_proxy()`
* The current logic for `using_proxy` is added to `Credentials` including tests
* All current usage is updated to call `using_proxy` via `Credentials`

## Implementation Brief

* Add `Credentials::using_proxy` method, basically:
```php
$creds = $this->get();
return (bool) preg_match( '/\.apps\.sitekit\.withgoogle\.com$', $creds['oauth2_client_id'] );
```
* Deprecate `OAuth_Client::using_proxy`
* Update all usage to call `using_proxy` via `Credentials`

## QA Brief

* 

## Changelog entry

* N/A

felixarntz:
IB ✅

aaemnnosttv:
## QA ✅ 

* ⚠️ One instance of the deprecated use is present, although I believe this is due to #1211 being merged in after this one 😄   
[...]

PR #1543, opened 2020-05-11
Moved using_proxy to the Credentials class

ShahAaron:
## Summary

Addresses issue #1490 

## Checklist


includes/Core/Authentication/Clients/OAuth_Client.php
```diff
@@ -205,7 +205,8 @@ public function get_client() {
 	 * @return Google_Site_Kit_Client|Google_Site_Kit_Proxy_Client
 	 */
 	private function setup_client() {
-		if ( $this->using_proxy() ) {
+		$credentials = $this->get_client_credentials();
```
aaemnnosttv, line 208: `get_client_credentials` does not return the same `Credentials` instance that is available via `$this->credentials` which you're already using. This line is safe to remove as it isn't being used.

includes/Core/Authentication/Clients/OAuth_Client.php
```diff
@@ -621,7 +622,8 @@ function( $scope ) {
 
 		// TODO: In the future, once the old authentication mechanism no longer exists, this check can be removed.
 		// For now the below action should only fire for the proxy despite not clarifying that in the hook name.
-		if ( $this->using_proxy() ) {
+		$credentials = $this->get_client_credentials();
```
aaemnnosttv, line 625: Same here; please remove this line.

includes/Core/Authentication/Clients/OAuth_Client.php
```diff
@@ -687,23 +689,14 @@ private function refresh_profile_data() {
...
-		// If no credentials yet, assume true.
-		if ( ! is_object( $credentials ) || empty( $credentials->web->client_id ) ) {
-			return true;
-		}
-
-		// If proxy credentials, return true.
-		if ( false !== strpos( $credentials->web->client_id, '.apps.sitekit.withgoogle.com' ) ) {
-			return true;
-		}
+		_deprecated_function( __METHOD__, 'n.e.x.t', 'Credentials::using_proxy()' );
```
aaemnnosttv, line 697: Let's update the replacement function to use the equivalent of `__METHOD__` for that method.
```suggestion
		_deprecated_function( __METHOD__, 'n.e.x.t', Credentials::class . '::using_proxy' );
```

includes/Core/Authentication/Credentials.php
```diff
@@ -113,4 +113,21 @@ protected function get_default() {
...
+	 * filter.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return bool True if proxy authentication is used, false otherwise.
+	 */
+	public function using_proxy() {
+		$creds = $this->get();
+
+		return ! preg_match( '/\.apps\.googleusercontent\.com$/', $creds['oauth2_client_id'] );
```
aaemnnosttv, line 131: Let's update this as we discussed.

* If there is no `$creds['oauth2_client_id']`, return `true`
* Update the matching logic used here to check if the client ID is a proxy site ID (same as IB)

tests/phpunit/integration/Core/Authentication/Clients/OAuth_ClientTest.php
```diff
@@ -307,20 +307,19 @@ function ( Request $request ) {
 	}
 
 	public function test_using_proxy() {
+		$this->setExpectedDeprecated( OAuth_Client::class . '::using_proxy' );
 		$context = new Context( GOOGLESITEKIT_PLUGIN_MAIN_FILE );
+		$client  = new OAuth_Client( $context );
```
aaemnnosttv, line 312: This same test needs to be added to `CredentialsTest` (only without the deprecation expectation of course 😄 )
The test itself will need to be slightly different but keep in mind that whenever we're adding a new public method for a class that it will generally need a test.

As for the `fake_*` methods.. these should probably be pulled into a shared trait that both test cases can use.

includes/Core/Util/Debug_Data.php
```diff
@@ -88,6 +97,7 @@ public function __construct(
 		$this->options        = $options ?: new Options( $this->context );
 		$this->user_options   = $user_options ?: new User_Options( $this->context );
 		$this->authentication = $authentication ?: new Authentication( $this->context, $this->options, $this->user_options );
+		$this->credentials    = $this->authentication->credentials();
```
aaemnnosttv, line 100: This isn't necessary, and we don't assign instance properties directly from another class anywhere else in a constructor. Please remove this line (and the added property above).

includes/Core/Util/Debug_Data.php
```diff
@@ -214,7 +224,7 @@ private function get_amp_mode_field() {
 	 */
 	private function get_site_status_field() {
 		$is_connected = $this->authentication->credentials()->has();
-		$using_proxy  = $this->authentication->get_oauth_client()->using_proxy();
+		$using_proxy  = $this->credentials->using_proxy();
```
aaemnnosttv, line 227: This line should reference the same `Credentials` instance as the line above 😄

includes/Core/Authentication/Clients/OAuth_Client.php
ShahAaron: Fixed

includes/Core/Authentication/Clients/OAuth_Client.php
ShahAaron: Fixed

includes/Core/Authentication/Clients/OAuth_Client.php
ShahAaron: Fixed

includes/Core/Authentication/Credentials.php
ShahAaron: Fixed

tests/phpunit/integration/Core/Authentication/Clients/OAuth_ClientTest.php
ShahAaron: All done.

I've made an attempt at a new trait :)

includes/Core/Util/Debug_Data.php
ShahAaron: Fixed

includes/Core/Util/Debug_Data.php
ShahAaron: Fixed

includes/Core/Util/Authentication_Trait.php
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Trait Google\Site_Kit\Core\Util\Authentication_Trait
+ *
+ * @package   Google\Site_Kit\Core\Util
+ * @copyright 2020 Google LLC
+ * @license   https://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
+ * @link      https://sitekit.withgoogle.com
+ */
+
+namespace Google\Site_Kit\Core\Util;
```
aaemnnosttv, line 11: This trait should only ever be used for tests so this should be in `tests/phpunit/includes` instead.

E.g. https://github.com/google/site-kit-wp/blob/3167c6cbbf003ce9c52a8713b63f26b512f96b28/tests/phpunit/includes/OptionsTestTrait.php#L11

includes/Core/Util/Authentication_Trait.php
```diff
@@ -0,0 +1,63 @@
...
+	/**
+	 * Creates fake Google authentication credentials for testing.
+	 *
+	 * @since n.e.x.t
+	 */
+	public function fake_authentication() {
+		add_filter(
+			'googlesitekit_oauth_secret',
+			function () {
+				return wp_json_encode(
```
aaemnnosttv, line 31: This is somewhat unrelated but would be good to update here - we no longer require that the secret is JSON encoded - it can be pre-decoded (as an array) so we can simply remove the wrapping `wp_json_encode` call here and return the array as-is. Same below 👍

tests/phpunit/integration/Core/Authentication/Clients/OAuth_ClientTest.php
```diff
@@ -26,6 +27,7 @@
  * @group Authentication
  */
 class OAuth_ClientTest extends TestCase {
+	use Authentication_Trait;
 
 	const SITE_ID   = '12345678.apps.sitekit.withgoogle.com';
 	const CLIENT_ID = '12345678.apps.googleusercontent.com';
```
aaemnnosttv: These constants refer to values that are in the trait now. Let's get rid of these constants and update the methods on the trait to also return the fake credentials set in the filter. We don't care about the whole thing just the array set to the `web` key, so it returns an array with `client_id` and `client_secret`.

Then we can update the references in the test to use the client ID returned from the `fake_` methods rather than the class constant 👌

includes/Core/Util/Authentication_Trait.php
ShahAaron: Fixed

includes/Core/Util/Authentication_Trait.php
ShahAaron: Fixed

tests/phpunit/integration/Core/Authentication/Clients/OAuth_ClientTest.php
ShahAaron: Fixed

aaemnnosttv:
@ShahAaron One final suggestion I have would be to rename the new trait to `Fake_Authentication_Trait` to be _slightly_ less ambiguous as we also have an `Authentication` class.

Otherwise, we should be just about done here 😄

tests/phpunit/integration/Core/Authentication/Clients/OAuth_ClientTest.php
```diff
@@ -29,18 +29,15 @@
...
 	public function test_get_client() {
 		$client = new OAuth_Client( new Context( GOOGLESITEKIT_PLUGIN_MAIN_FILE ) );
 
 		$this->assertInstanceOf( 'Google\Site_Kit\Core\Authentication\Clients\Google_Site_Kit_Client', $client->get_client() );
 	}
 
 	public function test_refresh_token() {
-		$this->fake_authentication();
-		$user_id = $this->factory()->user->create();
+		$fake_auth = $this->fake_authentication();
```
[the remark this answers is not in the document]
ShahAaron: All unused variables have been removed

tests/phpunit/includes/Authentication_Trait.php
```diff
@@ -0,0 +1,73 @@
...
+						'client_secret' => 'test-client-secret',
+					),
+				);
+			}
+		);
+
+		return array(
+			'client_id'     => '12345678.apps.googleusercontent.com',
+			'client_secret' => 'test-client-secret',
+		);
```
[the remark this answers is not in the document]
ShahAaron: Variables passed to the filters now

tests/phpunit/includes/Authentication_Trait.php
```diff
@@ -0,0 +1,73 @@
...
+	 * @return array Fake proxy authentication details including client_id & client_secret.
+	 */
+	public function fake_proxy_authentication() {
+		add_filter(
+			'googlesitekit_oauth_secret',
+			function () {
+				return array(
+					'web' => array(
+						'client_id'     => '12345678.apps.sitekit.withgoogle.com',
+						'client_secret' => 'test-client-secret',
```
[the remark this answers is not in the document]
ShahAaron: Client secret has been renamed

includes/Core/Authentication/Clients/OAuth_Client.php
```diff
@@ -687,23 +687,14 @@ private function refresh_profile_data() {
...
-		// If no credentials yet, assume true.
-		if ( ! is_object( $credentials ) || empty( $credentials->web->client_id ) ) {
-			return true;
-		}
-
-		// If proxy credentials, return true.
-		if ( false !== strpos( $credentials->web->client_id, '.apps.sitekit.withgoogle.com' ) ) {
-			return true;
-		}
+		_deprecated_function( __METHOD__, 'n.e.x.t', 'Credentials::class::using_proxy' );
```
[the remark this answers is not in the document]
ShahAaron: The above change has been made

tests/phpunit/includes/Fake_Authentication_Trait.php
```diff
@@ -0,0 +1,71 @@
...
+namespace Google\Site_Kit\Tests;
+
+/**
+ * Trait for getting fake authentication credentials.
+ *
+ * @since n.e.x.t
+ * @access private
+ * @ignore
+ */
+trait Fake_Authentication_Trait {
```
felixarntz, line 20: I like it! Maybe in the future we can add some other utility methods here to also take care of user authentication and verification, so that we can fake the entire setup, which may be useful for some tests (e.g. permissions rely on all of that).

Issue #1545, opened 2020-05-11
Fix CTAs about linking Analytics with AdSense to show up


felixarntz:
## Bug Description

There are two AdSense widgets that are sourced by the Analytics API (currently part of the Analytics module codebase, hooked into AdSense-specific UI via filters):

* full-width widget at the bottom of AdSense screen (below "Performance" widget, with similar heading)
* half-width widget at the bottom of Site Kit dashboard screen (on the right of the AdSense "Earnings" widget)

Both of these widgets display the most popular pages sorted by AdSense earnings. In order to get them to show up though, the user needs to link their Analytics and AdSense accounts. At least one of the two widgets has a CTA to inform the user about that, however due to a bug that CTA is not being displayed. The other widget doesn't provide it, but it should have one too.

The two widgets also have some error and zero data handling problems, partially because the error detection logic looks at error response formats that we generally don't use (maybe we did at some point way back?). The same applies to even the default error handler in the `withData` higher order component.

While looking into this issue, several more places for enhancements/fixes surfaced (too many blue boxes is obviously nothing new), but let's not get caught up in other UX issues too much at this point and focus on fixing the underlying problem.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* If the Analytics API returns the "Restricted metric(s): ..." error, the existing CTA should be displayed in both the above AdSense top pages tables: _You need to link Analytics and AdSense to get report for your top earning pages. Learn more: https://support.google.com/adsense/answer/6084409_
* If there is no data, the dashboard widget should show the regular CTA, while the AdSense module page widget should just not render at all.
    * While there will be already other similar blue CTAs in both screens in such a case, the difference is that in the dashboard it would look weird if we displayed nothing since the user would only see a half widget. Since we cannot modify outside UI at this point, it's okay to have double CTA there (like we have in some other cases too anyway). On the AdSense screen it's easy to just return null while keeping things looking fine because there the widget is full width.
* Ensure error format of having the error at the root of the response is supported by error handlers. Old error handling logic should be kept just in case, because we're not sure why it was added - all of this will get refactored away at some point anyway.

## Implementation Brief

* Modify `getDataError` in `AdSenseDashboardWidgetTopPagesTableSmall` (the dashboard one) to support the current error response format (by looking at existence of all `code`, `message`, and `data.status`) to ensure the CTA to link Analytics and AdSense is displayed.
[...]

PR #1546, opened 2020-05-11
Fix bugs with error/zero-data handling around Analytics/AdSense linked widgets

felixarntz:
## Summary

Addresses issue #1545

## Relevant technical choices

* Added some comments around the existing error conditions, which I maintained, since there could still be something that this somehow applies to. No need to deep dive and remove as much as possible since this code will all be removed in the long term.

## Screenshots

Here is what the two CTAs look like:
Screenshot 2020-05-11 at 15 47 47

Screenshot 2020-05-11 at 15 48 33

## Checklist

### DIFF
--- a/assets/js/modules/analytics/common/adsense-link-cta.js
+++ b/assets/js/modules/analytics/common/adsense-link-cta.js
@@ -1,0 +1,36 @@
+/**
+ * Analytics AdSense Link CTA component.
+ *
+ * Site Kit by Google, Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import CTA from '../../../components/notifications/cta';
+
+export default function AdSenseLinkCTA() {
+	return (
+		<CTA
+			title={ __( 'Restricted metric(s)', 'google-site-kit' ) }
+			description={ __( 'You need to link Analytics and AdSense to get report for your top earning pages. Learn more: https://support.google.com/adsense/answer/6084409 ', 'google-site-kit' ) }
+		/>
+	);
+}
--- a/assets/js/modules/analytics/common/adsense-link-cta.js
+++ b/assets/js/modules/analytics/common/adsense-link-cta.js
@@ -29,8 +29,10 @@ import CTA from '../../../components/notifications/cta';
 export default function AdSenseLinkCTA() {
 	return (
 		<CTA
-			title={ __( 'Restricted metric(s)', 'google-site-kit' ) }
-			description={ __( 'You need to link Analytics and AdSense to get report for your top earning pages. Learn more: https://support.google.com/adsense/answer/6084409 ', 'google-site-kit' ) }
+			title={ __( 'Link Analytics and AdSense', 'google-site-kit' ) }
+			description={ __( 'You need to link Analytics and AdSense to get reports for your top earning pages.', 'google-site-kit' ) }
+			ctaLink="https://support.google.com/adsense/answer/6084409"
+			ctaLabel={ __( 'Learn more', 'google-site-kit' ) }
 		/>
 	);
 }
--- a/assets/js/modules/analytics/dashboard/dashboard-widget-analytics-adsense-top-pages.js
+++ b/assets/js/modules/analytics/dashboard/dashboard-widget-analytics-adsense-top-pages.js
@@ -36,6 +36,7 @@ import { TYPE_MODULES } from '../../../components/data';
 import { getDataTableFromData, TableOverflowContainer } from '../../../components/data-table';
 import Layout from '../../../components/layout/layout';
 import PreviewTable from '../../../components/preview-table';
+import ctaWrapper from '../../../components/notifications/cta-wrapper';
 import AdSenseLinkCTA from '../common/adsense-link-cta';
 import { analyticsAdsenseReportDataDefaults, isDataZeroForReporting } from '../util';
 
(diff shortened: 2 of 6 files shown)


assets/js/modules/analytics/dashboard/dashboard-widget-analytics-adsense-top-pages.js
```diff
@@ -146,24 +144,39 @@ const isDataZero = () => {
...
-	if ( ! data || ! data.error ) {
-		return false;
-	}
+	if ( data.code && data.message && data.data && data.data.status ) {
+		// Specifically looking for string "badRequest"
+		if ( data.data.reason && 'badRequest' === data.data.reason ) {
+			return AnalyticsAdSenseDashboardWidgetTopPagesTable.renderLayout(
+				<CTA
+					title={ __( 'Restricted metric(s)', 'google-site-kit' ) }
+					description={ __( 'You need to link Analytics and AdSense to get report for your top earning pages. Learn more: https://support.google.com/adsense/answer/6084409 ', 'google-site-kit' ) }
```
aaemnnosttv, line 153: We should use `ctaLabel` and `ctaLink` props for "Learn more" and the URL rather than putting these in the description as plain text.

assets/js/modules/analytics/dashboard/dashboard-widget-analytics-adsense-top-pages.js
```diff
@@ -146,24 +144,39 @@ const isDataZero = () => {
...
 const getDataError = ( data ) => {
-	if ( ! data || ! data.error ) {
-		return false;
-	}
+	if ( data.code && data.message && data.data && data.data.status ) {
+		// Specifically looking for string "badRequest"
+		if ( data.data.reason && 'badRequest' === data.data.reason ) {
+			return AnalyticsAdSenseDashboardWidgetTopPagesTable.renderLayout(
+				<CTA
+					title={ __( 'Restricted metric(s)', 'google-site-kit' ) }
```
aaemnnosttv, line 152: We should use a different title for this message as this is more of an under-the-hood detail. Using the word "restricted" may also sound like the user has done something "forbidden" when they haven't.

Maybe something like "Connect Analytics with AdSense for more data" to call the user to action?

assets/js/modules/analytics/dashboard/dashboard-widget-analytics-adsense-top-pages.js
```diff
@@ -146,24 +144,39 @@ const isDataZero = () => {
  *
  */
 const getDataError = ( data ) => {
-	if ( ! data || ! data.error ) {
-		return false;
-	}
+	if ( data.code && data.message && data.data && data.data.status ) {
+		// Specifically looking for string "badRequest"
+		if ( data.data.reason && 'badRequest' === data.data.reason ) {
```
aaemnnosttv, line 149: This could be simplified into a single conditional:
```js
const { code, message, data, reason } = data;
if ( code && message && status && data && 'badRequest' === reason ) {
```
No need to check for `reason` and `reason ===` as it will simply be `undefined` if not there since `data` is already checked for.

assets/js/modules/analytics/dashboard/dashboard-widget-top-earning-pages-small.js
```diff
@@ -107,23 +107,17 @@ class AdSenseDashboardWidgetTopPagesTableSmall extends Component {
...
  * @return {(HTMLElement|null)} Returns HTML element markup with error message if it exists.
  */
 const getDataError = ( data ) => {
-	if ( data && data.error_data ) {
-		const errors = Object.values( data.error_data );
-
+	if ( data.code && data.message && data.data && data.data.status ) {
 		// Specifically looking for string "badRequest"
-		if ( errors[ 0 ] && 'badRequest' === errors[ 0 ].reason ) {
+		if ( data.data.reason && 'badRequest' === data.data.reason ) {
```
aaemnnosttv, line 120: Same as above, this could be a bit cleaner and simplified.

assets/js/modules/analytics/util/index.js
```diff
@@ -462,19 +461,6 @@ export const getTopPagesReportDataDefaults = () => {
 		},
 	];
 
-	if ( getModulesData().analytics.settings.adsenseLinked ) {
```
aaemnnosttv, line 465: If we're not using this, should we not also remove the setting and all of the added filters, etc around it?

felixarntz: Potentially, I thought about that too. I think we can keep it for now and revisit separately. We should think about potential other use-cases for that connection and see if we still need it then.

assets/js/modules/analytics/dashboard/dashboard-widget-analytics-adsense-top-pages.js
felixarntz: I'll remove the `data.data.reason` check since that's unnecessary, however the two clauses itself should remain separate because the first one in itself also results in `data.message` to be returned (see below).

assets/js/modules/analytics/dashboard/dashboard-widget-top-earning-pages-small.js
felixarntz: See above, I'll only omit the one unnecessary check.

felixarntz:
Updated screenshot: 
Screenshot 2020-05-13 at 16 56 10

assets/js/modules/analytics/common/adsense-link-cta.js
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Internal dependencies
+ */
+import CTA from '../../../components/notifications/cta';
+
+export default function AdSenseLinkCTA() {
+	return (
+		<CTA
+			title={ __( 'Link Analytics and AdSense', 'google-site-kit' ) }
+			description={ __( 'You need to link Analytics and AdSense to get reports for your top earning pages.', 'google-site-kit' ) }
```
aaemnnosttv, line 33: This sounds a bit like an instruction rather than a choice the user has, if they want. 
Perhaps something like "Get reports for your top earning pages by linking your Analytics and AdSense accounts."?

marrrmarrr: Yeah, agree with Evan on the wording -- let's make clear what benefit they're getting from this if they decide to go through with the action. 

Other than that, why are the blue rectangles different sizes? The one in the topmost section has a white margin, while the other two span the entire widget. (Not blocking, just looks strange, would be good to have uniform UI.)

aaemnnosttv:
```suggestion
			description={ __( 'Get reports for your top earning pages by linking your Analytics and AdSense accounts.', 'google-site-kit' ) }
```

Issue #1510, opened 2020-05-06
Add a generic, "core/forms" store for forms


tofumatt:
Issue #1211 required some forms that didn't warrant creating a bunch of actions, selectors, etc. for their data, but we shouldn't be using `useState` to manage form data because it requires passing around state from components and negates a lot of the wins we get from using a Redux-backed datastore.

We talked about this on a call today, and we should have a generic, `'core/forms'` data store with a very basic API for getting/setting state across forms in the app.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a new file, `assets/js/googlesitekit/datastore/forms`, that registers a new form data store, under `'core/forms'`.
* It should provide a single selector: `getValue( formName, key )`
* It should provide a single action: `setValues( formName, { key: value } )` where the second argument is a map of keys to set and their respective values.

## Implementation Brief

* Create a new datastore in `assets/js/googlesitekit/datastore/forms` that registers a new form data store, under `'core/forms'`.
* Create a `assets/js/googlesitekit-forms.js` that registers the `core/forms` store on the default registry.
* Add a `googlesitekit-forms` asset in `Assets.php`, similar to `googlesitekit-api`, `googlesitekit-data`, etc. that would load the `core/form` datastore into the default registry.
* Add a selector: `getValue( formName, key )`. This should return the value stored in the `state[ formName ][ key]`; `undefined` if the `formName` or `key` doesn't exist.
* Add an action: `setValues( formName, { key: value } )` where the second argument is a map of keys to set and their respective values. If the specified `formName` doesn't exist; it should be created.

## QA Brief

* Since this is an engineering-only feature for now, specific QA isn't really needed.

## Changelog entry

* Introduce `core/forms` datastore to manage form state.

tofumatt:
Flagging this as a 7 estimate, but for what it's worth it's closer to 3-4 hours than 7 and doesn't require much specialised QA. 😄

felixarntz:
@tofumatt 
Maybe we should have `setValues( formName, { key: value } )` and `setValue( formName, key, value )`? The latter could technically call the former. I just think your current suggestion is a bit confusing using a singular "value" in the name but accepting multiple. I think it should use the plural, and having the other one would be more of a convenience thing if we need it.

[...]

PR #1552, opened 2020-05-13
Added new forms datastore and registered it.

ShahAaron:
## Summary

Addresses issue #1510 

## Checklist


assets/js/googlesitekit/datastore/forms/index.js
```diff
@@ -0,0 +1,91 @@
...
+ * External dependencies
+ */
+import Data from 'googlesitekit-data';
+import invariant from 'invariant';
+import { STORE_NAME } from './constants';
+
+// Actions
+const SET_FORM_VALUES = 'SET_FORM_VALUES';
+
+export { STORE_NAME };
```
aaemnnosttv, line 29: Let's move this up as the first export for consistency with other stores.

assets/js/googlesitekit/datastore/forms/index.js
```diff
@@ -0,0 +1,91 @@
...
+
+export const INITIAL_STATE = {
+};
```
aaemnnosttv, line 32:
```suggestion
export const INITIAL_STATE = {};
```

assets/js/googlesitekit/datastore/forms/index.js
```diff
@@ -0,0 +1,91 @@
...
+
+export const actions = {
+	setValues( formName, formData ) {
+		invariant( formName, 'form name is required.' );
```
aaemnnosttv, line 36: The field referenced in the message should match the name of the parameter. Also it's a good idea to make the message a bit more specific to where it's used. That way, when we see the error we know where it's happening.  Otherwise `x is required` is not very useful unless you're unit testing that exact function. Take a look at some of the other `invariant` uses in the codebase to see what I mean. What you have here is consistent with much of the invariant use we have, so this is a suggestion I have for everyone.

In this case though, something like `formName is required for setting values.` would be great.

Also, `formData` should be covered by `invariant` as well as a required valid value 😄

assets/js/googlesitekit/datastore/forms/index.js
```diff
@@ -0,0 +1,91 @@
...
+export const controls = {};
+
+export const reducer = ( state, { type, payload } ) => {
+	switch ( type ) {
+		case SET_FORM_VALUES: {
+			const { formName, formData } = payload;
+
+			return {
+				...state,
+				[ formName ]: formData,
```
aaemnnosttv, line 54: This isn't in the AC, but I believe the intention here is to merge the given `formData` with any existing `formData` for that form. That way you don't need to pass the complete state of the entire form every time, but can also use partial objects, similar to [`setState`](https://reactjs.org/docs/react-component.html#setstate).

@tofumatt - can you confirm this is the intended behavior?

One important detail here is that any time we add data into the store from an object, it's important that we always add that to the store _as a new object_. This usually is done using a spread like we do with `state` itself, but the important thing is that the data which is in the store no longer has a reference _outside_ the store where it could be mutated by reference. Spreading an object only makes shallow copies, so it isn't perfect, but it's a reasonable precaution that should always be taken in this context.

assets/js/googlesitekit/datastore/forms/index.js
```diff
@@ -0,0 +1,91 @@
...
+export const resolvers = {};
+
+export const selectors = {
+	getValue( state, formName, key ) {
+		if ( ! formName ) {
+			return undefined;
+		}
+		if ( ! key ) {
+			return undefined;
+		}
```
aaemnnosttv, line 73: This conditional isn't really necessary as `undefined` will be returned anyways for any falsey key.

I wouldn't really mind keeping it just for the sake of being explicit, but we don't make this check for other stores that also return undefined for the same reason, so let's remove it.

assets/js/googlesitekit/datastore/forms/index.js
tofumatt: Ah, yes, that is 100% the intention. You're right—I didn't explicitly set that out in the AC, and it's implied because of the intended use for the store, but I should have mentioned it.

So yes, all existing data should be kept when `setValues()` is dispatched. It should, as @aaemnnosttv mentioned, basically behave like React's `setState()`. 👍

includes/Core/Assets/Assets.php
```diff
@@ -481,6 +481,15 @@ private function get_assets() {
 					),
 				)
 			),
+			new Script(
+				'googlesitekit-datastore-forms',
+				array(
+					'src'          => $base_url . 'js/googlesitekit-datastore-forms.js',
```
tofumatt, line 487: This will need a related webpack entry point added as well, eg. https://github.com/google/site-kit-wp/blob/develop/webpack.config.js#L96

assets/js/googlesitekit/datastore/forms/index.js
```diff
@@ -0,0 +1,83 @@
...
+export { STORE_NAME };
+
+const SET_FORM_VALUES = 'SET_FORM_VALUES';
+
+export const INITIAL_STATE = {};
+
+export const actions = {
+	setValues( formName, formData ) {
+		invariant( formName, 'formName is required for setting values.' );
+		invariant( formData !== undefined, 'formData is required.' );
```
tofumatt, line 35: Technically we want `formData` to be an object, so the `!== undefined` check is overly verbose. Checking for truthiness is fine.

assets/js/googlesitekit/datastore/forms/index.js
```diff
@@ -0,0 +1,83 @@
...
+import { STORE_NAME } from './constants';
+
+export { STORE_NAME };
+
+const SET_FORM_VALUES = 'SET_FORM_VALUES';
+
+export const INITIAL_STATE = {};
+
+export const actions = {
+	setValues( formName, formData ) {
```
tofumatt, line 33: This should have a JSDoc block—the function should be marked as `@private` for now, but it'd be nice to document this function as I imagine we'll be using it a fair bit.

assets/js/googlesitekit/datastore/forms/index.js
```diff
@@ -0,0 +1,83 @@
...
+		default: {
+			return { ...state };
+		}
+	}
+};
+
+export const resolvers = {};
+
+export const selectors = {
+	getValue( state, formName, key ) {
```
tofumatt, line 66: This should have a JSDoc block as well.

assets/js/googlesitekit/datastore/forms/index.js
```diff
@@ -0,0 +1,83 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * External dependencies
+ */
+import Data from 'googlesitekit-data';
```
tofumatt, line 22: This isn't an external dependency; it's internal to the project so should be put in an

```js
/**
 * Internal dependencies
 */
```

section beneath `invariant`.

assets/js/googlesitekit/datastore/forms/index.js
```diff
@@ -0,0 +1,83 @@
...
+import invariant from 'invariant';
+import { STORE_NAME } from './constants';
```
tofumatt, line 24: This is internal too 😄

assets/js/googlesitekit/datastore/forms/index.js
```diff
@@ -0,0 +1,83 @@
...
+export const controls = {};
+
+export const reducer = ( state, { type, payload } ) => {
+	switch ( type ) {
+		case SET_FORM_VALUES: {
+			const { formName, formData } = payload;
+
+			return {
+				...state,
+				[ formName ]: formData,
```
tofumatt, line 53:
```suggestion
				[ formName ]: { ...( state[ formName ] || {} ), ...formData },
```

would do the trick here.

You'll want to add some unit tests to make sure this behaves as expected, things like:

* it should work with a `formName` not previously assigned
* it should work with a `formName` previously assigned
* it should preserve data from `state[ formName ]` when new data is assigned

ShahAaron: @tofumatt a new test has now been added

assets/js/googlesitekit/datastore/forms/index.test.js
```diff
@@ -0,0 +1,93 @@
...
+ * Internal dependencies
+ */
+import {
+	createTestRegistry,
+	unsubscribeFromAll,
+} from 'tests/js/utils';
+import { STORE_NAME } from './constants';
+
+describe( 'core/forms store', () => {
+	const formName = 'Form 1';
```
tofumatt, line 29: We should move these variables down into each test's code. Them being here avoids repetition, but it makes the test code much harder to read/reason about because the variables we use as input to the tests aren't scoped to a particular test nor near them.

It's helpful to approach test code differently than the rest of our code: here we want a hyper-focus on readability and context, even if that means repeating things across many tests 😄

assets/js/googlesitekit/datastore/forms/index.test.js
```diff
@@ -0,0 +1,93 @@
...
+		key3: 'value3',
+		key4: 'value4',
+	};
+	let registry;
+
+	beforeEach( () => {
+		registry = createTestRegistry();
+	} );
+
+	afterEach( () => {
```
tofumatt, line 45: This doesn't need to be here; we aren't using any `subscribeUntil` calls.

assets/js/googlesitekit/datastore/forms/index.test.js
```diff
@@ -0,0 +1,93 @@
...
+
+	describe( 'selectors', () => {
+		describe( 'getValue', () => {
+			it( 'works with a formName not previously assigned', async () => {
+				await registry.dispatch( STORE_NAME ).setValues( formName, { ...formData } );
+
+				const { key2 } = formData;
+				const form = registry.select( STORE_NAME ).getValue( formName, Object.keys( formData )[ 1 ] );
+				expect( form ).toEqual( key2 );
+			} );
```
tofumatt, line 72: Minor thing, but we usually leave an empty newline between each `it()` test block. 🙂

assets/js/googlesitekit/datastore/forms/index.test.js
```diff
@@ -0,0 +1,93 @@
...
+		} );
+	} );
+
+	describe( 'selectors', () => {
+		describe( 'getValue', () => {
+			it( 'works with a formName not previously assigned', async () => {
+				await registry.dispatch( STORE_NAME ).setValues( formName, { ...formData } );
+
+				const { key2 } = formData;
+				const form = registry.select( STORE_NAME ).getValue( formName, Object.keys( formData )[ 1 ] );
```
tofumatt, line 70: Using `Object.keys( formData )[ 1 ]` makes this test difficult to parse. It's not clear to me from here what's going on. Better would be much more explicit:

```suggestion
				const form = registry.select( STORE_NAME ).getValue( 'Form 1', 'key1' );
```

assets/js/googlesitekit/datastore/forms/index.test.js
```diff
@@ -0,0 +1,93 @@
...
+				expect( () => {
+					registry.dispatch( STORE_NAME ).setValues( formName );
+				} ).toThrow( 'formData must be an object.' );
+			} );
+		} );
+	} );
+
+	describe( 'selectors', () => {
+		describe( 'getValue', () => {
+			it( 'works with a formName not previously assigned', async () => {
```
tofumatt, line 66: This test spec isn't accurate; as mentioned above regarding the `TypeError`—this actually _won't work_ with an unassigned `formName`. Since you're assigning it below this test passes, but if line 67 is removed this test will fail with `TypeError: Cannot read property 'key2' of undefined`.

assets/js/googlesitekit/datastore/forms/index.test.js
```diff
@@ -0,0 +1,93 @@
...
+				await registry.dispatch( STORE_NAME ).setValues( formName, { ...formData } );
```
tofumatt, line 67: This isn't async so we don't _need_ the `await`, though it's fine 😏

assets/js/googlesitekit/datastore/forms/index.test.js
```diff
@@ -0,0 +1,93 @@
...
+
+				const { key2 } = formData;
+				const form = registry.select( STORE_NAME ).getValue( formName, Object.keys( formData )[ 1 ] );
+				expect( form ).toEqual( key2 );
+			} );
+			it( 'works with a formName previously assigned', async () => {
```
tofumatt, line 73:
```suggestion
			it( 'works with an existing formName', async () => {
```

assets/js/googlesitekit/datastore/forms/index.js
```diff
@@ -0,0 +1,104 @@
...
+	 * @since n.e.x.t
+	 * @private
+	 *
+	 * @param {string} formName Name of the form.
+	 * @param {Object} formData Form data supplied as a map of keys to set and their respective values.
+	 * @return {Object} Redux-style action.
+	 */
+	setValues( formName, formData ) {
+		invariant( formName, 'formName is required for setting values.' );
+		invariant( typeof formData === 'object', 'formData must be an object.' );
```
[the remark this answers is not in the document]
ShahAaron: @tofumatt I've used `formData instanceof Object && formData.constructor === Object` to check for an object

assets/js/googlesitekit/datastore/forms/index.test.js
ShahAaron: @tofumatt I've changed this test

assets/js/googlesitekit/datastore/forms/index.test.js
```diff
@@ -19,73 +19,111 @@
...
-				await registry.dispatch( STORE_NAME ).setValues( formName, { ...formData } );
+			it( 'works with a formName that does not exist', () => {
+				const formName = 'Form 1';
+				const formData = {
+					key1: 'value1',
+					key2: 'value2',
+				};
+				const newFormName = 'Form 2';
+
+				registry.dispatch( STORE_NAME ).setValues( formName, { ...formData } );
```
aaemnnosttv, line 76: No need to spread the `formData` in unless you're combining objects inline. This can simply be `formData`.

Same goes for other tests.

assets/js/googlesitekit-datastore-forms.js
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Internal dependencies
+ */
+import './googlesitekit/datastore/forms';
```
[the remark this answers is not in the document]
ShahAaron: @tofumatt is there anything I should do to mitigate this side-effect ?

assets/js/googlesitekit/datastore/forms/index.test.js
```diff
@@ -19,73 +19,111 @@
...
 	describe( 'actions', () => {
 		describe( 'setValues', () => {
 			it( 'requires the formName param', () => {
 				expect( () => {
 					registry.dispatch( STORE_NAME ).setValues();
 				} ).toThrow( 'formName is required for setting values.' );
 			} );
+
 			it( 'requires the formData param', () => {
+				const formName = 'Form 1';
```
aaemnnosttv, line 41: Extracting the form name to a variable is a bit necessary in most cases here. It would be more clear to use a self-describing string literal like "form-name".

Minor nitpick: let's avoid using numbered names unless the test is uses multiple forms. I find using letters for generic names for multiple things, and numbers only when order is significant to be a good rule of thumb that helps improve readability.

assets/js/googlesitekit/datastore/forms/index.test.js
```diff
@@ -19,73 +19,111 @@
...
 				} ).toThrow( 'formData must be an object.' );
 			} );
+
+			it( 'requires the formData param to be an object not an array', () => {
+				const formName = 'Form 1';
+
+				expect( () => {
+					registry.dispatch( STORE_NAME ).setValues( formName, [] );
+				} ).toThrow( 'formData must be an object.' );
+			} );
```
aaemnnosttv, line 54: 👍

assets/js/googlesitekit/datastore/forms/index.test.js
```diff
@@ -19,73 +19,111 @@
...
+				const form = registry.select( STORE_NAME ).getValue( newFormName, 'key2' );
+				expect( form ).toEqual( undefined );
+			} );
+
+			it( 'works with formData where the key does not exist', () => {
+				const formName = 'Form 1';
+				const formData = {
+					key1: 'value1',
+					key2: 'value2',
+				};
```
aaemnnosttv, line 87: These can be inlined

assets/js/googlesitekit/datastore/forms/index.js
```diff
@@ -82,10 +82,10 @@ export const selectors = {
 	 * @param {Object} state Data store's state.
 	 * @param {string} formName Name of the form.
 	 * @param {string} key Key supplied from formData.
-	 * @return {string|undefined} Value stored in state by formName and key or undefined.
+	 * @return {*} Value stored in state by formName and key or undefined.
 	 */
 	getValue( state, formName, key ) {
-		return state[ formName ][ key ];
+		return state[ formName ] ? state[ formName ][ key ] : undefined;
```
aaemnnosttv, line 88: I find this to be a bit easier to read, what do you think?
```suggestion
		const formData = state[ formName ] || {};

		return formData[ key ];
```

assets/js/googlesitekit/datastore/forms/index.test.js
```diff
@@ -19,73 +19,111 @@
...
+
+				registry.dispatch( STORE_NAME ).setValues( formName, { ...formData } );
 
 				const { key2 } = formData;
-				const form = registry.select( STORE_NAME ).getValue( formName, Object.keys( formData )[ 1 ] );
+				const form = registry.select( STORE_NAME ).getValue( formName, 'key2' );
```
aaemnnosttv, line 105: `form` is not an accurate name for the returned value since you're getting a single value by key from the form.

The destructured `key2` is also a bit unnecessary.

Can we add a similar test for `setValues`? We should set values with one set, and then set them again leaving at least one of the original keys untouched, then assert that the untouched key and updated keys return their expected values.

ShahAaron: @aaemnnosttv I've added a new test for updating keys

assets/js/googlesitekit/datastore/forms/index.js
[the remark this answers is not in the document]
tofumatt: I actually meant that we can just check for truthiness, I don't think we need to be this specific about the checks. Though I guess this works. Checking for object types can be iffy in JS, I'd rather be more permissive is all.

assets/js/googlesitekit-datastore-forms.js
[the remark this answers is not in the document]
tofumatt: There's no need to mitigate, I meant it's worth adding a comment so folks know that's what's happening here if they see the code; otherwise it looks like an import that's not doing anything.

assets/js/googlesitekit/datastore/forms/index.js
```diff
@@ -0,0 +1,106 @@
...
+	 * @since n.e.x.t
+	 * @private
+	 *
+	 * @param {string} formName Name of the form.
+	 * @param {Object} formData Form data supplied as a map of keys to set and their respective values.
+	 * @return {Object} Redux-style action.
+	 */
+	setValues( formName, formData ) {
+		invariant( formName, 'formName is required for setting values.' );
+		invariant( formData instanceof Object && formData.constructor === Object, 'formData must be an object.' );
```
tofumatt, line 45: I think checking truthiness is enough here; the docs spell out the usage but this sort of stuff seems fragile and not worth running on every call to me. 🤷‍♂️

assets/js/googlesitekit/datastore/forms/index.js
```diff
@@ -0,0 +1,106 @@
...
+export const selectors = {
+	/**
+	 * Gets the existing form by formName and key.
+	 *
+	 * @since n.e.x.t
+	 * @private
+	 *
+	 * @param {Object} state Data store's state.
+	 * @param {string} formName Name of the form.
+	 * @param {string} key Key supplied from formData.
```
tofumatt, line 84:
```suggestion
	 * @param {string} key Get data stored in this key.
```

assets/js/googlesitekit/datastore/forms/index.js
```diff
@@ -0,0 +1,106 @@
...
+	 * @return {*} Value stored in state by formName and key or undefined.
```
tofumatt, line 85:
```suggestion
	 * @return {*} Value stored in state by formName and key. Returns `undefined` if formName or key isn't found.
```

assets/js/googlesitekit/datastore/forms/index.test.js
```diff
@@ -0,0 +1,117 @@
...
+			} );
+
+			it( 'requires the formData param to be an object not a string', () => {
+				expect( () => {
+					registry.dispatch( STORE_NAME ).setValues( 'form-name', 'formData' );
+				} ).toThrow( 'formData must be an object.' );
+			} );
+
+			it( 'works with formData where the key value is updated, and the untouched key and updated key return expected values', () => {
+				const formData = { key1: 'value1', key2: 'value2' };
```
tofumatt, line 59: These can be moved right into the dispatch calls to tidy up this test a bit more. No need to define test variables if they're only used once.

assets/js/googlesitekit/datastore/forms/index.test.js
```diff
@@ -0,0 +1,117 @@
...
+				} ).toThrow( 'formData must be an object.' );
+			} );
+
+			it( 'requires the formData param to be an object not a string', () => {
+				expect( () => {
+					registry.dispatch( STORE_NAME ).setValues( 'form-name', 'formData' );
+				} ).toThrow( 'formData must be an object.' );
+			} );
+
+			it( 'works with formData where the key value is updated, and the untouched key and updated key return expected values', () => {
```
tofumatt, line 58: These could probably be a few separate tests:

- one for ensuring the keys are set on an unused `formName`
- one for ensuring the keys are set on an already-defined `fornName`
- one for ensuring the appropriate keys are overwritten and unaffected keys are ignored
- one ensuring an empty object doesn't overwrite the store (this is implied to be the expected behaviour to me, but we actually haven't defined it)

assets/js/googlesitekit/datastore/forms/index.test.js
```diff
@@ -0,0 +1,117 @@
...
+
+				const formUpdatedValue = registry.select( STORE_NAME ).getValue( 'form-name', 'key1' );
+				expect( formUpdatedValue ).toEqual( 'value3' );
+			} );
+		} );
+	} );
+
+	describe( 'selectors', () => {
+		describe( 'getValue', () => {
+			it( 'works with a formName that does not exist', () => {
```
tofumatt, line 77: This test name implies that `getValue( 'form-name' )` works even if `registry.dispatch( STORE_NAME ).setValues( 'form-name', {} )` has never been called, but before you check the result of `getValue` you _do_ set data for a different `formName`. Because they're so closely named and the setting of the form name doesn't impact the test at all, this made me do a double-take.

At first, I thought this test wasn't testing the appropriate behaviour, but then I saw the differing `formName` values and saw that it is. But why is `registry.dispatch( STORE_NAME ).setValues( 'form-name', formData );` being called at all? It doesn't impact this test but makes it confusing to read.

This test can be simplified and made less confusing to parse 🙂, eg:

```js
it( 'works with a formName that does not exist', () => {
  const formValue = registry.select( STORE_NAME ).getValue( 'new-form-name', 'key2' );
  expect( formValue ).toEqual( undefined );
} );
```

assets/js/googlesitekit/datastore/forms/index.test.js
```diff
@@ -0,0 +1,117 @@
...
+			it( 'works with an existing formName', () => {
+				const formData = { key1: 'value1', key2: 'value2' };
+
+				registry.dispatch( STORE_NAME ).setValues( 'form-name', formData );
+
+				const formValue = registry.select( STORE_NAME ).getValue( 'form-name', 'key2' );
+				expect( formValue ).toEqual( 'value2' );
+			} );
+
+			it( 'preserves data from state when new data is assigned', () => {
```
tofumatt, line 104: This is more of a test of `setValues` than `getValue`, but it's a good test (one of the ones I suggested above). I think this should just be moved into the test for the `setValues` action.

assets/js/googlesitekit/datastore/forms/index.test.js
ShahAaron: New Tests added

assets/js/googlesitekit/datastore/forms/index.js
ShahAaron: As agreed this check is to be left as is

assets/js/googlesitekit-datastore-forms.js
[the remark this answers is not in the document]
ShahAaron: Extra comment has been added

assets/js/googlesitekit-datastore-forms.js
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Internal dependencies
+ *
+ * Registers the core/forms datastore.
```
tofumatt, line 22: Nitpick, but the:

```js
/**
 * Internal dependencies
 */
```

dependencies comment blocks aren't supposed to have anything else in them. Best to put this on a single-line comment above the import, like:

```js
// This import has side effects; it registers the core/forms datastore
// on the default data store (`googlesitekit.data`).

```

assets/js/googlesitekit/datastore/forms/index.test.js
```diff
@@ -0,0 +1,127 @@
...
+				} ).toThrow( 'formData must be an object.' );
+			} );
+
+			it( 'requires the formData param to be an object not an array', () => {
+				expect( () => {
+					registry.dispatch( STORE_NAME ).setValues( 'form-name', [] );
+				} ).toThrow( 'formData must be an object.' );
+			} );
+
+			it( 'requires the formData param to be an object not a string', () => {
```
tofumatt, line 52: You could add another test here ensuring an object _doesn't_ throw, especially an empty one:

```js
it( 'does not throw if formData is an object', () => {
    expect( () => {
        registry.dispatch( STORE_NAME ).setValues( 'form-name', {} );
    } ).not.toThrow();
} );
```

assets/js/googlesitekit/datastore/forms/index.test.js
```diff
@@ -0,0 +1,127 @@
...
+			it( 'works with formData where the key value is updated, and the untouched key returns an expected value', () => {
+				registry.dispatch( STORE_NAME ).setValues( 'form-name', { key1: 'value1', key2: 'value2' } );
+
+				registry.dispatch( STORE_NAME ).setValues( 'form-name', { key1: 'value3' } );
+
+				const formValue = registry.select( STORE_NAME ).getValue( 'form-name', 'key2' );
+				expect( formValue ).toEqual( 'value2' );
+			} );
+
+			it( 'works with formData where the key value is updated, and the updated key returns an expected value', () => {
```
tofumatt, line 67:
```suggestion
			it( 'returns a newly-set value if a new value for an existing key is set', () => {
```

assets/js/googlesitekit/datastore/forms/index.test.js
```diff
@@ -0,0 +1,127 @@
...
+				} ).toThrow( 'formData must be an object.' );
+			} );
+
+			it( 'requires the formData param to be an object not a string', () => {
+				expect( () => {
+					registry.dispatch( STORE_NAME ).setValues( 'form-name', 'formData' );
+				} ).toThrow( 'formData must be an object.' );
+			} );
+
+			it( 'works with formData where the key value is updated, and the untouched key returns an expected value', () => {
```
tofumatt, line 58:
```suggestion
			it( 'does not overwrite unrelated keys', () => {
```

assets/js/googlesitekit/datastore/forms/index.test.js
```diff
@@ -0,0 +1,127 @@
...
+			it( 'works with formData where the key value is updated, and the updated key returns an expected value', () => {
+				registry.dispatch( STORE_NAME ).setValues( 'form-name', { key1: 'value1', key2: 'value2' } );
+
+				registry.dispatch( STORE_NAME ).setValues( 'form-name', { key1: 'value3' } );
+
+				const formValue = registry.select( STORE_NAME ).getValue( 'form-name', 'key1' );
+				expect( formValue ).toEqual( 'value3' );
+			} );
+
+			it( 'works with formData where the key value is updated with an empty object without overwriting the store', () => {
```
tofumatt, line 76:
```suggestion
			it( 'does not overwrite unrelated keys when an empty object is supplied to formData', () => {
```

assets/js/googlesitekit/datastore/forms/index.test.js
```diff
@@ -0,0 +1,127 @@
...
+
+				const formValue = registry.select( STORE_NAME ).getValue( 'form-name', 'key2' );
+				expect( formValue ).toEqual( 'value2' );
+			} );
+		} );
+	} );
+
+	describe( 'selectors', () => {
+		describe( 'getValue', () => {
+			it( 'works with a formName that does not exist', () => {
```
tofumatt, line 107: 👍

assets/js/googlesitekit/datastore/forms/index.test.js
```diff
@@ -0,0 +1,127 @@
...
+			} );
+
+			it( 'works with formData where the key does not exist', () => {
+				registry.dispatch( STORE_NAME ).setValues( 'form-name', { key1: 'value1', key2: 'value2' } );
+
+				const formValue = registry.select( STORE_NAME ).getValue( 'form-name', 'key3' );
+				expect( formValue ).toEqual( undefined );
+			} );
+
+			it( 'works with an existing formName', () => {
```
tofumatt, line 119:
```suggestion
			it( 'works with an existing formName and key', () => {
```

assets/js/googlesitekit/datastore/forms/index.js
```diff
@@ -0,0 +1,106 @@
...
+};
+
+const store = {
+	INITIAL_STATE,
+	actions,
+	controls,
+	reducer,
+	resolvers,
+	selectors,
+};
```
felixarntz, line 101: @ShahAaron @tofumatt Can we split this file with the actual form logic out into a `forms.js` file? I know it seems not as necessary here since likely this will remain the only piece of this store, but there's also no harm in splitting this up and it makes the whole architecture easier to overview.

The `index.js` file should only be the entrypoint that connects the datastore, includes `Data.commonStore` etc. At a minimum, `Data.commonStore` is currently missing from here and should be baked in.

ShahAaron: @felixarntz I've split the store into `index.js` and `forms.js` and added `Data.commonStore`

PR #1561, opened 2020-05-15
Enhancement/1426 pagespeed performance data selector

jqlee85:
## Summary

Addresses issue #1426

Created `getReport` selector which takes `options.strategy` and `options.url` and gets that report, first checking if it has already been loaded in the store, and fetching it if not. Followed the conventions used in analytics data store for fetching and testing.

## Checklist


assets/js/modules/pagespeed-insights/datastore/report.test.js
```diff
@@ -0,0 +1,168 @@
...
+				fetch
+					.doMockOnceIf(
+						/^\/google-site-kit\/v1\/modules\/pagespeed-insights\/data\/pagespeed/
+					)
+					.mockResponseOnce(
+						JSON.stringify( fixtures.desktopReport ),
+						{ status: 200 }
+					);
+
+				const options = { strategy: 'desktop', url: 'http://testsite.10uplabs.com/' };
```
aaemnnosttv, line 88: Let's use the same URL we use in other tests here
```suggestion
				const options = { strategy: 'desktop', url: 'http://example.com/' };
```

I won't highlight all of these so please make sure this is updated in all places.

assets/js/modules/pagespeed-insights/datastore/report.test.js
```diff
@@ -0,0 +1,168 @@
...
+				// Ensure the proper parameters were passed.
+				expect( fetch.mock.calls[ 0 ][ 0 ] ).toMatchQueryParameters(
+					options
+				);
+
+				expect( initialReport ).toEqual( undefined );
+				await subscribeUntil( registry,
+					() => (
+						registry.select( STORE_NAME ).getReport( options ) !== undefined
+					),
```
aaemnnosttv, line 101: I realize many of the older stores don't do this, but it's better to use the `hasFinishedResolution( selector, selectorArgs )` selector that's part of every store as this will fail faster if the selector is implemented improperly. It's also a bit more clear as to what you're waiting for 😄 
```suggestion
					() => registry.select( STORE_NAME ).hasFinishedResolution( 'getReport', [ options ] )
```

assets/js/modules/pagespeed-insights/datastore/report.test.js
```diff
@@ -0,0 +1,168 @@
...
+				const report = registry.select( STORE_NAME ).getReport( options );
+
+				expect( fetch ).toHaveBeenCalledTimes( 1 );
+				expect( report ).toEqual( fixtures.desktopReport );
+			} );
+
+			it( 'does not make a network request if report for given options is already present', async () => {
+				const options = {
+					strategy: 'mobile',
+					url: 'http://sitekit.10uplabs.com/',
```
aaemnnosttv, line 113:
```suggestion
					url: 'http://example.com/',
```

assets/js/modules/pagespeed-insights/datastore/__fixtures__/desktop-report.json
```diff
@@ -0,0 +1,5 @@
+{
+    "report": {
+        "data": "PageSpeed Insights Desktop Report Data"
+    }
+}
```
aaemnnosttv, line 5: Fixtures should be specific to the datapoint they request, with any relevant variation as part of it; i.e. `pagespeed--desktop.json`.

Apart from that, this isn't an actual response from the API so it doesn't offer much value as a fixture. We will need real data for these requests to use in Storybook as soon as the components are refactored so we may as well add it here.

Please rename these fixtures and fill them with the response from the API, using `https://example.com/` as the `url`, one for each strategy. You can do this with the plugin by setting your custom site URL in the dev settings and clearing your local+session storage so it makes the requests again.

assets/js/modules/pagespeed-insights/datastore/__fixtures__/index.js
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export { default as mobileReport } from './mobile-report';
+export { default as desktopReport } from './desktop-report';
```
aaemnnosttv, line 20: `pagespeedDesktop` and `pagespeedMobile`

assets/js/modules/pagespeed-insights/datastore/__fixtures__/desktop-report.json
jqlee85: Ah ok, I had done something like that originally, but removed it as we weren't actually testing anything within those responses.

assets/js/modules/pagespeed-insights/datastore/report.test.js
```diff
@@ -0,0 +1,168 @@
...
+
+	describe( 'actions', () => {
+
+	} );
+
+	describe( 'selectors', () => {
+		describe( 'getReport', () => {
+			it( 'uses a resolver to make a network request', async () => {
+				global[ baseInfoVar ] = baseInfo;
+				global[ entityInfoVar ] = {};
```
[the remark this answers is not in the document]
jqlee85: OK, yeah actually I think these aren't even necessary for these tests so I'll just remove them.

assets/js/modules/pagespeed-insights/datastore/report.js
```diff
@@ -0,0 +1,222 @@
...
+};
+
+export const resolvers = {
+	*getReport( options = {} ) {
+		if ( options.strategy === undefined ) {
+			options.strategy = 'mobile';
+		}
+		if ( options.url === undefined ) {
+			options.url = registry.select( CORE_SITE ).getReferenceSiteURL();
+		}
```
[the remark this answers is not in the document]
jqlee85: :+1:

jqlee85:
@aaemnnosttv I've implemented those changes, let me know if there's anything else, thanks!

aaemnnosttv:
@googlebot I consent.

assets/js/modules/pagespeed-insights/datastore/report.js
```diff
@@ -0,0 +1,228 @@
...
+		invariant( typeof strategy === 'string', 'a valid strategy is required to fetch a report.' );
+		let response, error;
+
+		yield {
+			payload: { strategy, url },
+			type: START_FETCH_REPORT,
+		};
+
+		try {
+			const report = yield {
```
tofumatt, line 71: Looks like this code isn't covered by a test for its response, because this constant is being created here, sent to the `receiveReport` action, but then never used again. From the structure of this action, it looks like this should be assigned to the `response` variable you define above and return below.

```suggestion
			response = yield {
```

assets/js/modules/pagespeed-insights/datastore/report.js
```diff
@@ -0,0 +1,228 @@
...
+				payload: { strategy, url },
+				type: FETCH_REPORT,
+			};
+
+			yield actions.receiveReport( report, { strategy, url } );
```
tofumatt, line 76:
```suggestion
			yield actions.receiveReport( response, { strategy, url } );
```

assets/js/modules/pagespeed-insights/datastore/report.js
```diff
@@ -0,0 +1,228 @@
...
+				payload: {
+					error,
+					strategy,
+					url,
+				},
+				type: CATCH_FETCH_REPORT,
+			};
+		}
+
+		return { response, error };
```
tofumatt, line 94: As mentioned above: `response` will always be `undefined` for this function. 😅

assets/js/modules/pagespeed-insights/datastore/report.test.js
```diff
@@ -0,0 +1,136 @@
...
+
+	afterAll( () => {
+		API.setUsingCache( true );
+	} );
+
+	afterEach( () => {
+		unsubscribeFromAll( registry );
+	} );
+
+	describe( 'actions', () => {
```
tofumatt, line 51: We shouldn't leave empty describe blocks, so normally I'd say to remove this… but in this case I'd say it's a good place to add a test for the `fetchReport()` action that wasn't returning the correct data above 😄

assets/js/modules/pagespeed-insights/datastore/report.js
jqlee85: Ah, thanks for the catch!

assets/js/modules/pagespeed-insights/datastore/report.js
jqlee85: :+1:

assets/js/modules/pagespeed-insights/datastore/report.js
jqlee85: :+1:

assets/js/modules/pagespeed-insights/datastore/report.test.js
jqlee85: Makes sense, added a test.

Issue #1380, opened 2020-04-13
Remove Analytics module as requirement for Tag Manager module


felixarntz:
## Feature Description

It should not be mandatory for the Analytics module to be set up before being able to set up the Tag Manager module.

**PR for this should be opened against `feature/ga-gtm-separation`.**

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Tag Manager should no longer have the Analytics module as a dependency.
* All messaging related to this dependency should no longer be visible.

## Implementation Brief

* In `includes/Modules/Tag_Manager::setup_info` remove the `depends_on` array entry.
* in `assets/js/components/setup-module.js` capture the name of `requiredModule` and display it via a placeholder in where the text is currently displayed as `Set up Analytics to gain access to %s`

## QA Brief

* 

## Changelog entry

* Remove Analytics module as requirement for Tag Manager module, decoupling them to be individual.

aaemnnosttv:
A little context to help with the definition here: Site Kit has the concept of dependencies between modules. Currently Tag Manager can only be activated after Analytics is activated because it declares Analytics as a dependency in `Tag_Manager::setup_info`.

This is used on the client side in a few places:
https://github.com/google/site-kit-wp/search?q=requiredModule&unscoped_q=requiredModule

Most of the behavior is not specific to Analytics, but there is some Analytics specific messaging. This messaging should become dynamic rather than assuming any module that is blocked by a parent needs Analytics 😄 

Analytics is also a dependency for Optimize, although I'm not sure why. @felixarntz is this still necessary?

felixarntz:
@aaemnnosttv Thanks for providing some extra context here.
> Analytics is also a dependency for Optimize, although I'm not sure why. @felixarntz is this still necessary?

I don't think so, but let's look at that separately. Not as big of a priority to decouple.

aaemnnosttv:
✅ IB

ShahAaron:
@felixarntz should I just replace the work `Analytics` in the text `Set up Analytics to gain access to %s` with the required module ?

aaemnnosttv:
PR is merged but not adding a release here as it is still only targeting a feature branch.

cole10up:
**Tested**

Built a feature/ga-gtm-separation zip

Installed SK and activated.

Notice dashboard shows Tag Manager separate and available to install outside of Analytics:
[...]

PR #1563, opened 2020-05-15
Enhancement/1380 remove analytics requirement

ShahAaron:
## Summary

Addresses issue #1380 

## Checklist

### DIFF
--- a/assets/js/components/setup-module.js
+++ b/assets/js/components/setup-module.js
@@ -96,6 +96,7 @@ class SetupModule extends Component {
 		} = this.props;
 
 		let blockedByParentModule = false;
+		let parentModule;
 
 		const { modules } = global.googlesitekit;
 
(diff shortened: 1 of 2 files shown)


ShahAaron:
@googlebot I consent.

ShahAaron:
@googlebot I consent.

assets/js/components/setup-module.js
```diff
@@ -106,6 +107,7 @@ class SetupModule extends Component {
 			requiredModules.forEach( ( requiredModule ) => {
 				if ( ! modules[ requiredModule ].setupComplete ) {
 					blockedByParentModule = true;
+					parentModule = requiredModule.replace( /(^\w{1})|(\s{1}\w{1})/g, ( match ) => match.toUpperCase() );
```
aaemnnosttv, line 110: We can't simply transform a module slug into a name for display like this as the module name is a translatable string. We need to use `modules[ requiredModule ].name` if it exists.

ShahAaron: @aaemnnosttv if the `modules[ requiredModule ].name` does not exist, is there a fallback ?

aaemnnosttv: The required module should always exist. It already assumes this above where it checks if `setupComplete` so we can assume it exists here as well.

ShahAaron: Fixed as requested

Issue #1499, opened 2020-05-04
Refactor Search Console settings using datastore


felixarntz:
## Feature Description

Similar to #1101, #1014, and #1386 (but a lot simpler here!), let's refactor the Search Console settings (no setup involved here) to use the datastore introduced by #1498.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All changes should be made within the `assets/js/modules/search-console` directory.
* There should be a `SettingsView` component that displays the Search Console settings. It should look the same as it does today, except:
    * "Connected URL" should say "Connected Property"
    * The value displayed should not be the site URL, but the Search Console property ID (via `getPropertyID` selector)
* There should be a `SettingsMain` component for the settings panel, that for now just always displays `SettingsView`. While it is not strictly necessary, it's good for consistency with the other module components infrastructure, and it keeps it internal to the `SettingsMain` component that Search Console does not have any way to edit settings (i.e. there is no need for a `SettingsEdit` component).
* The components should only access datastore data, never anything from globals.
* There should be a Search Console/Settings section in Storybook with the settings panel in different states, similar to how it was added for the issues linked above.
* The implementation should follow patterns established as part of the issues linked above.

## Implementation Brief

This work is dependant on the datastore introduced by #1498
All changes to be in `assets/js/modules/search-console`
* Remane `assets/js/modules/search-console/settings/search-console-settings-status.js` to `assets/js/modules/search-console/settings/settings-view.js`
    * In `assets/js/modules/search-console/settings/settings-view.js` change "Connected URL" to say "Connected Property"
    * In `assets/js/modules/search-console/settings/settings-view.js` change the value displayed should not be the site URL, but the Search Console property ID (via `getPropertyID` selector)
* Create `SettingsMain` in `assets/js/modules/search-console/settings/settings-main.js` that for now just always displays `SettingsView`. See #1101 
* Move filter for adding the modules settings component from `index.legacy.js` to `index.js`. Note to insure that the new component needs to be wrapped using `RegistryProvider` before being passed to the filter. 

`<Data.RegistryProvider value={ Data }>
        .......
</Data.RegistryProvider>`

* There should be a Search Console/Settings section in Storybook ( `stories/module-search-console-settings.stories.js` ) with the settings panel in different states, similar to how it was added for the issues linked above. See https://github.com/google/site-kit-wp/blob/refactor/1014-adsense-setup-components/stories/module-analytics-settings.stories.js

[...]

PR #1570, opened 2020-05-18
Enhancement/1499 search console using datastore

ShahAaron:
## Summary

Addresses issue #1499 

## Checklist


assets/js/modules/search-console/settings/settings-view.js
```diff
@@ -1,5 +1,5 @@
 /**
- * SearchConsoleSettingStatus component.
+ * Search Console Settings View component
  *
  * Site Kit by Google, Copyright 2019 Google LLC
```
felixarntz, line 4: Nitpick, but would be good to update the year here since this is kind of a new file although it shows up here as "renamed".

assets/js/modules/search-console/settings/settings-view.js
```diff
@@ -19,27 +19,27 @@
...
-				mdc-layout-grid__cell--span-12
-			">
+/**
+ * Internal dependencies
+ */
+import Data from 'googlesitekit-data';
+import { STORE_NAME } from '../datastore/constants';
+
+export default function SettingsView() {
+	const { useSelect } = Data;
```
felixarntz, line 31: This would be better to move right below the imports like in other datastore-using components, so that it is not unnecessarily run on every render.

stories/module-search-console-settings.stories.js
```diff
@@ -0,0 +1,118 @@
...
+		<WithTestRegistry callback={ callback }>
+			<div style={ { background: 'white' } }>
+				<SettingsModule
+					key={ module.slug + '-module' }
+					slug={ module.slug }
+					name={ module.name }
+					description={ module.description }
+					homepage={ module.homepage }
+					learnmore={ module.learnMore }
+					active={ module.active }
```
felixarntz, line 77: You need to also pass `setupComplete` here like in the other module's Storybook setups.

stories/module-search-console-settings.stories.js
```diff
@@ -0,0 +1,118 @@
...
+					hasSettings={ true }
```
felixarntz, line 78: Search Console is the one current special case so for visual parity (see hack in https://github.com/google/site-kit-wp/blob/1.8.1/assets/js/components/settings/settings-modules.js#L150), so let's pass `false` here.

assets/js/modules/search-console/settings/settings-view.js
ShahAaron: Year has been updated

assets/js/modules/search-console/settings/settings-view.js
ShahAaron: Completed

stories/module-search-console-settings.stories.js
ShahAaron: setupComplete has been added

stories/module-search-console-settings.stories.js
ShahAaron: hasSettings has been set to false

Issue #1573, opened 2020-05-18
Include update routine to check affected sites


felixarntz:
## Feature Description

Site owners whose sites were actually affected by the security vulnerability allowing unprivileged accounts to get verified and thus gain Search Console access should be informed and presented with steps to resolve the issue.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* On update, if there are WP user accounts that do not have the `googlesitekit_setup` capability but still have a verification token set, a request to Site Kit Service should be issued to flag them.
    * Site Kit Service will provide a notification to these sites with how they can resolve the issue.
* The plugin should support displaying more than one notification in the dashboard, in case there are multiple that the user has not dismissed (this should be the case already now, but needs to be tested specifically).

## Implementation Brief

* Introduce class `Migration_1_8_1` similar to existing migration classes in `includes/Core/Util`, with public `register()` and `migrate()` methods.
* Add private method to the class `check_unauthorized_verified_users()`:
    * If the site is not using the proxy (tiny edge-case), bail and do nothing.
    * Run a `WP_User_Query` to get all users (set reasonable limit of 20) that have a `googlesitekit_site_verification_meta` or `googlesitekit_site_verification_file` user option set, and then filter them by whether `! user_can( $user_id, Permissions::SETUP )`, to get all users that are verified despite not being authorized to do so. If there are no such users, all is fine, bail early.
    * Disconnect these users by wiping all their user meta that starts with `googlesitekit_` (similar logic as in `Authentication::disconnect()`).
    * Issue a `POST` request to the proxy's `/notifications/mark/` route with `notification_id` set to "verification_leak" and `notification_state` set to "required", as well as identifiers for these unauthorized users in an `identifiers` parameter.
* Initialize class instance from `Plugin` main class.

## QA Brief

* 

## Changelog entry

* Check for users that verified through Site Kit without being authorized to do so, and disconnect them from Site Kit.

adamsilverstein:
Overall IB looks good. One question:

> Site Kit Service will provide a notification to these sites with how they can resolve the issue.

Do these show up on the dashboard and persist until dismissed?

felixarntz:
@adamsilverstein 
> Do these show up on the dashboard and persist until dismissed?

Yes it would be dismissible and show until then.

felixarntz:
QA ✅ 

[...]

PR #1574, opened 2020-05-19
Add update routine to check affected sites

felixarntz:
## Summary

Addresses issue #1573

## Relevant technical choices

* Also adds a few missing capability checks that should not currently have major implications but should definitely be added to prevent potential leaks.

## Checklist


tests/phpunit/includes/TestCase.php
```diff
@@ -143,6 +146,38 @@ function () {
...
+							'client_id'     => self::CLIENT_ID,
+							'client_secret' => 'test-client-secret',
+						),
+					)
+				);
+			}
+		);
+	}
+
+	protected function fake_proxy_site_connection() {
```
aaemnnosttv, line 165: How about we cherry pick the added `Fake_Authentication_Trait` instead?
https://github.com/google/site-kit-wp/blob/71b8bb2fb9fb60bc5ee8cb8166a670c53cc1bc9a/tests/phpunit/includes/Fake_Authentication_Trait.php

includes/Core/Util/Migration_1_8_1.php
```diff
@@ -0,0 +1,262 @@
...
+		if ( ! $this->authentication->credentials()->has() ) {
+			return false;
+		}
+
+		$result = $this->clear_and_flag_unauthorized_verified_users();
+
+		// Update database version.
+		$this->options->set( 'googlesitekit_db_version', self::DB_VERSION );
+
+		return $result;
```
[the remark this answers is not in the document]
felixarntz: That's a fair point, I don't think the return value hurts, but we should be able to do the same kind of testing without it, which is more appropriate.

includes/Core/Util/Migration_1_8_1.php
```diff
@@ -0,0 +1,262 @@
...
+		// is irrelevant.
+		if ( ! $this->authentication->get_oauth_client()->using_proxy() ) {
+			return false;
+		}
+
+		// Only run routine once site credentials present, otherwise it is not
+		// possible to connect to the authentication service.
+		if ( ! $this->authentication->credentials()->has() ) {
+			return false;
+		}
```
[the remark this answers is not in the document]
felixarntz: Yeah I took that into consideration, I think it's acceptable here. I think we should ensure we wait with the routine until the site is connected to the proxy. The checks that are made here aren't expensive, so I think only updating the DB version / running the routine once these requirements are met makes sense.

We shouldn't decouple the user detection/cleanup from the proxy API call since it needs to send related information there.

tests/phpunit/includes/TestCase.php
felixarntz: Good point, will do that.

includes/Core/Util/Migration_1_8_1.php
```diff
@@ -0,0 +1,257 @@
...
+	 *
+	 * @since 1.8.1
+	 *
+	 * @return WP_User[] User objects of verified users. Maximum of 20.
+	 */
+	private function get_verified_users() {
+		return get_users(
+			array(
+				'number'     => 20,
+				'meta_query' => array( // phpcs:ignore WordPress.DB.SlowDBQuery
```
aaemnnosttv, line 242: Since we're specifically looking for users who shouldn't be verified, should we also add `role__not_in` with `array( 'administrator' )` so we don't potentially hit the limit with users we know should be capable?

tests/phpunit/integration/Core/Util/Migration_1_8_1Test.php
```diff
@@ -0,0 +1,249 @@
...
+		);
+
+		$this->fake_proxy_site_connection();
+		$this->stub_mark_notifications_response();
+
+		$this->migration->migrate();
+		$this->assertReceivedIdentifiers( $expected_identifiers );
+		foreach ( $problem_cases as $user ) {
+			$this->assertNoSiteKitUserOptions( $user->ID );
+		}
```
aaemnnosttv, line 141: 1. We should also assert that the non-problem users' data is untouched
1. Nitpick: let's stick with the PHPUnit convention for naming negated assertions using "Not" with the positive form e.g.  
    - `assertUserHasSiteKitOptions`
    - `assertUserNotHasSiteKitOptions`
  Maybe these should be most specific to `hasVerification` and `hasAccessToken` or `hasConnection`? 

This assertion is fine with me though if it means the user has/not-has any Site Kit user options.

tests/phpunit/integration/Core/Util/Migration_1_8_1Test.php
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	public function test_migrate_without_problem_cases() {
+		$this->create_user( 'administrator', self::FILE_VERIFICATION );
+		$this->create_user( 'subscriber', false );
+
+		$this->fake_proxy_site_connection();
+
+		$this->migration->migrate();
+		$this->assertEquals( Migration_1_8_1::DB_VERSION, $this->get_db_version() );
```
aaemnnosttv, line 103: Should also test that:
* admin still has Site Kit options.
* no API request was made

tests/phpunit/integration/Core/Util/Migration_1_8_1Test.php
```diff
@@ -0,0 +1,249 @@
...
+		$this->migration->migrate();
+		$this->assertNotEquals( Migration_1_8_1::DB_VERSION, $this->get_db_version() );
+	}
+
+	public function test_migrate_without_users() {
+		$this->fake_proxy_site_connection();
+
+		$this->migration->migrate();
+		$this->assertEquals( Migration_1_8_1::DB_VERSION, $this->get_db_version() );
+	}
```
aaemnnosttv, line 94: Ideally these should also assert no API request was made.

includes/Core/Util/Migration_1_8_1.php
```diff
@@ -0,0 +1,257 @@
...
+			return $response;
+		}
+
+		$response_code = wp_remote_retrieve_response_code( $response );
+		if ( 200 !== $response_code ) {
+			$body    = wp_remote_retrieve_body( $response );
+			$decoded = json_decode( $body, true );
+			return new WP_Error( $response_code, ! empty( $decoded['error'] ) ? $decoded['error'] : $body );
+		}
+		return true;
```
aaemnnosttv, line 180: `$response` and subsequent return values are no longer used and can be removed. That should also mean that we can set `blocking => false`.

includes/Core/Util/Migration_1_8_1.php
felixarntz: Using the role wouldn't be reliable, we should rely on capability checks. If a developer intentionally modified the site to also enable the permission for e.g. editors, we should cater for that. The only way to properly deal with capabilities like that is on the code level, these permission checks are made in `clear_unauthorized_verified_users()`.

includes/Core/Util/Migration_1_8_1.php
felixarntz: I think this method returning values is not problematic. I agree it's weird for an action hook callback method to return something, but arbitrary methods like this can certainly return values, which would have the benefit to make testing them easier (even if we don't have direct coverage here).

tests/phpunit/integration/Core/Util/Migration_1_8_1Test.php
felixarntz: Good catch on testing non-problem users' data. Regarding the scope, I think the current approach of checking for any Site Kit user options makes most sense since that's what the logic should do (effectively disconnect the user).

includes/Plugin.php
```diff
@@ -158,6 +158,8 @@ function() use ( $options, $transients, $assets ) {
 				( new Core\Notifications\Notifications( $this->context, $options, $authentication ) )->register();
 				( new Core\Util\Debug_Data( $this->context, $options, $user_options, $authentication, $modules ) )->register();
 				( new Core\Admin\Standalone( $this->context ) )->register();
+				( new Core\Util\Migration_1_3_0( $this->context, $options, $user_options ) )->register();
+				( new Core\Util\Migration_1_8_1( $this->context, $options, $user_options, $authentication ) )->register();
```
adamsilverstein, line 162: Noting the 1_3_0 migration class is moved inside the `init` callback here. I assume that is fine, maybe better, and I assume these need to load in order.

felixarntz: Yeah they need to load in order, that's why I moved this too. Furthermore, we were previously not passing the `User_Options` instance to `Migration_1_3_0` (because the outer method does not have awareness of it), which was technically okay, but still something we shouldn't do, to make sure we consistently use the same instances in the plugin.

adamsilverstein:
Nice work @felixarntz!

PR #1578, opened 2020-05-20
Improve getAccountTicketTermsOfServiceURL URL logic.

adamsilverstein:
Address URL construction bug discovered in QA on https://github.com/google/site-kit-wp/issues/1211.

## Summary

Ensure `getAccountTicketTermsOfServiceURL` returns undefined until email is available in the data store.

Before this change the original URL was returned without the user email and this was used as the redirect. This URL fails if the user is logged into a different Google account in their browser.

## Checklist

### DIFF
--- a/assets/js/modules/analytics/datastore/accounts.js
+++ b/assets/js/modules/analytics/datastore/accounts.js
@@ -22,18 +22,12 @@
 import invariant from 'invariant';
 
 /**
- * WordPress dependencies
- */
-import { addQueryArgs } from '@wordpress/url';
-
-/**
  * Internal dependencies
  */
 import API from 'googlesitekit-api';
 import Data from 'googlesitekit-data';
 import { isValidAccountSelection } from '../util';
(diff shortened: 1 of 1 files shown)


adamsilverstein:
@felixarntz testing this I don't think the `authuser` field works correctly here. When I am (logged into a different account in my browser) I am redirected and the `?api.accountTicketId={ID}` part of my URL is stripped during the redirect and I get an error: `Error: Account token is missing.`.

I wonder if there is a way we can detect if the suer is logged into multiple accounts?

felixarntz:
@adamsilverstein Good catch! I thought it would make sense to still return the base URL until the email is loaded since it would work for the majority of users (anyone who's only logged in with that one account), but yeah the trade-off is worse for the users that _do_ have multiple accounts. Did you verify with @cole10up that that was the reason the problem occurred? Multiple signed-in Google accounts?

adamsilverstein:
Debugging some more with Evan, I think the biggest issue is not being logged in as the user trying to provision the account. Evan may have a slight improvement for the URL construction, I'm going to close this PR for now, in our testing the email address was helpful when you were logged into more than one account, analytics switched to the proper account.

Issue #1566, opened 2020-05-15
Request write scopes only when needed


felixarntz:
## Feature Description

Site Kit currently requests scopes on a per-module basis. Whenever you activate a module, you have to grant access to all scopes that using this module _may_ need. The "may" is an important word here: For example, you will only need Analytics write permissions if you're creating a new account, property, or profile. However, currently all users need to grant these permissions even if they'll never use this functionality. Using incremental scopes is a best practice with OAuth, and while we're doing a decent job at it handling it per module, we should improve that specifically for scopes that allow write access, so that these are really only requested when they're needed.

This is especially time-critical because of the Analytics provisioning workstream which introduced a new write scope (for the first time since original plugin release). With the current logic in place, every existing user with Analytics active would see a prompt to grant that scope, which would be very much out of context, and they may not want to use this feature at all.

As part of this issue, we will mainly focus on Analytics, since the implementation relies on the write-related client logic being refactored already (using datastores). We can also add the change to AdSense, although that is trivial here because we don't expose any write logic (the API only has a single route that writes data anyway). This issue will only partially address the issue for Tag Manager though, since we'll need to complete #1386 first, but it must not be a blocker for this one.

Two modules that should not be changed are Search Console and Site Verification. Site Verification doesn't differentiate between read and write scopes anyway, and for Search Console we can continue to always request write permissions, since they are already needed during setup. We can reconsider that separately in the future.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `get_scopes` methods of the following modules should be modified to only include non-write scopes:
    * For AdSense, `https://www.googleapis.com/auth/adsense` should be replaced with `https://www.googleapis.com/auth/adsense.readonly`.
    * For Analytics, all scopes other than `https://www.googleapis.com/auth/analytics.readonly` should be removed.
* The `Authentication::needs_reauthenticate()` method in PHP needs to be updated so that lower permission required scopes we request don't result in `true` if the user already has granted access to a higher permission scope (e.g. if we request `adsense.readonly` but the user already granted `adsense`). This is so that users from older versions are not unexpectedly and unnecessarily asked to re-authenticate.
* All module datapoints should be annotated with specific scopes that they require if their functionality requires more scopes than the regular module ones (the read-only ones). This should use an alternate version of the `get_datapoint_services()` method, while not breaking backward-compatibility.
[...]

Issue #189, opened 2019-07-10
Add explanation how to re-activate modules in Settings page


bjackson27:
If a user doesn't approve all scopes, we show an error "not all scopes approved", but no information how to fix that. Add language to tell the user they need to disconnect the service and begin setup from scratch.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria
For cases where the user hasn't approved all requested scopes during the setup, display the following error message:  

* Heading: _Site Kit can't access necessary data_ 
* Text: _Site Kit can't access the relevant data from $PRODUCT because you haven't granted all API scopes requested during setup. To use Site Kit, you'll need to redo the setup for $PRODUCT – make sure to approve all API scopes at the authentication stage._ 
* Button: _Redo setup_ 

The text above should include variants for the following cases:
- Scopes from _multiple_ services/products are missing
- We're not able to determine which service/product is responsible for the missing required scope(s)

This is the same message displayed in #1114.

## Implementation Brief

In the `DashboardAuthAlert` component:
* Change the title prop to _Site Kit can't access necessary data_
* Identify product/services missing scopes by parsing the list of unsatisfied scopes from `global.googlesitekit.setup.unsatisfiedScopes`
* Parse product/service IDs from each scope in the list `https://www.googleapis.com/auth/{productOrServiceID}.potential.other.segments`
   - Note: the `productOrServiceID` above generally maps to a module slug, but there are exceptions (e.g. `webmasters` -> `search-console`) which need to be accounted for
   - `.potential.other.segments` may or may not be present, where it is a dot-separated string irrelevant to the matching of `productOrServiceID`
* Set the description prop to _Site Kit can't access the relevant data from $PRODUCT because you haven't granted all API scopes requested during setup. To use Site Kit, you'll need to redo the setup for $PRODUCT – make sure to approve all API scopes at the authentication stage._  using `sprintf` to pass the product name.
   - Include a variant that lists the names of multiple products
   - Include a generic variant for handling the rare case where no product or service is inferrable
   - _Exact verbaige TBD_
* change the `ctaLabel` prop to _Redo setup_ 

## Changelog entry

* Clarify message for when user needs to reauthenticate to grant required scopes.

marrrmarrr:
@bjackson27 presumably this would show up not just in the Settings, but also during the setup process. Do we have screenshots of how this would look during setup? That would help define the acceptance criteria.

felixarntz:
IB ✅

[...]

PR #1582, opened 2020-05-20
Add server-side infra for on-demand write scopes

aaemnnosttv:
## Summary

Addresses issue #1566

## Checklist


felixarntz:
@aaemnnosttv @tofumatt I'll defer to you whether to include the JS code in here too or do it in a separate PR against the same issue.

includes/Core/Authentication/Clients/OAuth_Client.php
```diff
@@ -404,10 +404,54 @@ public function get_required_scopes() {
...
+	 *
+	 * @return bool true if any required scopes are not satisfied, otherwise false.
+	 */
+	public function requires_additional_scopes() {
+		if ( ! $this->get_access_token() ) {
+			return false;
+		}
+
+		return ! $this->has_sufficient_scopes( $this->get_required_scopes() );
+	}
```
felixarntz, line 443: Can we rename this to `needs_reauthentication()`? This is closer to how it is in JS and how it was before. The current name is slightly confusing because "additional scopes" has a special meaning for us (scopes that are not the base required scopes but the user has granted them).

includes/Core/Authentication/Clients/OAuth_Client.php
```diff
@@ -539,7 +598,8 @@ public function get_authentication_url( $redirect_url = '' ) {
 		$this->user_options->set( self::OPTION_REDIRECT_URL, $redirect_url );
 
 		// Ensure the latest required scopes are requested.
-		$this->get_client()->setScopes( $this->get_required_scopes() );
+		$scopes = array_merge( $this->get_required_scopes(), $this->get_granted_additional_scopes(), $additional_scopes );
```
felixarntz, line 601: Per earlier suggestion and conversation with Thierry, we may want to remove `$this->get_granted_additional_scopes()` from right here again so that these scopes are not re-requested next time other scopes are requested (assuming they are likely one-time usage). Storing these additional scopes still makes sense so that the server knows they were granted.

includes/Core/Authentication/Exception/Insufficient_Scopes_Exception.php
```diff
@@ -0,0 +1,76 @@
...
+	 * @since n.e.x.t
+	 *
+	 * @return WP_Error
+	 */
+	public function to_wp_error() {
+		return new WP_Error(
+			static::WP_ERROR_CODE,
+			$this->getMessage(),
+			array(
+				'status' => 401, // Unauthorized.
```
felixarntz, line 71: May be more appropriate to use 403 here? I think 401 is more like "you're not logged in" where 403 is more like "you don't have permissions" (Google APIs also return 403 for scope errors).

includes/Core/Util/Scopes.php
```diff
@@ -0,0 +1,89 @@
...
+			'https://www.googleapis.com/auth/tagmanager.edit.containers',
+		),
+		'https://www.googleapis.com/auth/webmasters.readonly' => array(
+			'https://www.googleapis.com/auth/webmasters',
+		),
+	);
+
+	/**
+	 * Tests if the given scope is satisfied by the given list of available scopes.
+	 *
```
felixarntz, line 48: Missing since annotation.

includes/Core/Util/Scopes.php
```diff
@@ -0,0 +1,89 @@
...
+			return ! array_diff( $satisfying_scopes, $available_scopes );
+		}
+
+		// Return true if any of the scopes are present, otherwise false.
+		return (bool) array_intersect( $satisfying_scopes, $available_scopes );
+	}
+
+	/**
+	 * Tests if all the given scopes are satisfied by the list of available scopes.
+	 *
```
felixarntz, line 75: Missing since annotation.

includes/Core/Modules/Module.php
```diff
@@ -461,6 +487,65 @@ public function is_connected() {
...
+
+		if ( empty( $definitions[ $datapoint_key ]['scopes'] ) ) {
+			return;
+		}
+
+		$datapoint = $definitions[ $datapoint_key ];
+
+		// If the datapoint requires specific scopes, ensure they are satisfied.
+		if ( ! $this->authentication->get_oauth_client()->has_sufficient_scopes( $datapoint['scopes'] ) ) {
+			$exception = new Insufficient_Scopes_Exception( $datapoint['request_scopes_message'] );
```
felixarntz, line 516: We _may_ be better off separating the `error.message` from `error.data.requestScopesMessage` in JS, as originally in the IB (right now the error as implemented here will only include `error.message`). Depending on what messages we're going to use, what's logged (and potentially displayed) as general error messages may not be a good message to present to the user, and vice-versa. We are effectively "catching" the error in JS and then showing a modal with a more actionable explanation, so we could argue this should not be the same error message that should have been displayed if we didn't build the client-side mechanism.

includes/Modules/Analytics.php
```diff
@@ -589,6 +585,43 @@ protected function get_datapoint_services() {
...
 
+	/**
+	 * Gets map of datapoint to definition data for each.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array Map of datapoints to their definitions.
+	 */
+	protected function get_datapoint_definitions() {
+		$map = parent::get_datapoint_definitions();
```
felixarntz, line 596: This is okay for now, but we should do a follow-up issue where all datapoints are migrated here (also for all other modules, and `get_datapoint_services()` should be deprecated.

includes/Modules/Analytics.php
```diff
@@ -589,6 +585,43 @@ protected function get_datapoint_services() {
...
+			)
+		);
+
+		$map['POST:create-profile'] = array_merge(
+			$map['POST:create-profile'],
+			array(
+				'scopes'                 => array( 'https://www.googleapis.com/auth/analytics.edit' ),
+				'request_scopes_message' => __( 'Additional permissions are required to create a new Analytics view.', 'google-site-kit' ),
+			)
+		);
```
felixarntz, line 620: Note that `POST:settings` _may_ currently require `https://www.googleapis.com/auth/analytics.edit` too. I don't think we're using the datapoint this way anymore though (we no longer pass non-valid property ID/profile ID there to create them right?), so I think it'd be better to just remove that functionality rather than adjust for the legacy edge-case here.

includes/Modules/Tag_Manager.php
```diff
@@ -427,6 +427,30 @@ protected function get_datapoint_services() {
...
+	 * Gets map of datapoint to definition data for each.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array Map of datapoints to their definitions.
+	 */
+	protected function get_datapoint_definitions() {
+		$map = parent::get_datapoint_definitions();
+
+		// TODO: remove this once datapoint exists.
```
felixarntz, line 440: See above, eventually all datapoints should be in here anyway, no more `get_datapoint_services()`.

includes/Core/Authentication/Clients/OAuth_Client.php
aaemnnosttv: I thought about that and "reauthentication" is a bit vague.  which is that there are required scopes which are not satisfied. I hear you about the inconsistency of terms though. Maybe `has_insufficient_required_scopes` ? I don't mean to bike shed on this but just giving it a bit more thought now that it's becoming public.

I'll rename this for now in the interest of time, but let me know if you have any other ideas.

includes/Core/Authentication/Clients/OAuth_Client.php
aaemnnosttv: I like this idea. I've removed them here so that they are not automatically re-requested in subsequent auth URLs. I have a few ideas as to how we could make this better but it's not something we have bandwidth for in this release so this is a good first step.

includes/Core/Authentication/Exception/Insufficient_Scopes_Exception.php
aaemnnosttv: That makes sense. Updated 👍

includes/Core/Modules/Module.php
aaemnnosttv: I understand there is a potential benefit of separating an error message from a user-facing display message. I don't think we do that anywhere else and caught exceptions are turned into user-facing `WP_Error`s with the result of `getMessage()` as the message users will see. I'm open to ideas here, but it seems to me that the `$message` is intended to be user-facing as the class, `code`, and other data is more relevant for internal use.

includes/Modules/Analytics.php
aaemnnosttv: I didn't realize that was still there. The refactored components only use the dedicated datapoints so I'll remove the old behavior from `POST:settings`. I think it doesn't even need the settings datapoint in there at all since this is handled by `Modules` now.

One thing we were doing in the settings datapoint that we aren't in the new datapoints is when we created a new property and profile, we set the profile as the default profile for that property. 

I've removed the old settings datapoint handler for now. I think we can look at restoring this functionality in a follow-up issue, if necessary. I wonder if Analytics makes it the default view anyways since it's the first one created?

includes/Core/Authentication/Clients/OAuth_Client.php
felixarntz: I think the `needs_reauthentication` is fairly clear for this boolean value, given that it's a boolean I would argue there isn't even a need to know the "internals" on why reauthentication is needed - the other thing is we've been using this terminology throughout our codebase, I don't think it's worth changing.

includes/Core/Modules/Module.php
felixarntz: That's fair, let's go with this for now.

includes/Modules/Analytics.php
felixarntz: That's a good catch, let's test that. If it doesn't, we should fix this because it would be a regression. It would be weird if you created a new property and profile and then that new profile wouldn't be the one selected.

aaemnnosttv: It would still be selected because if a property does not have a `defaultProfileId` then we select the first, so it shouldn't be a regression in terms of UX.

If we wanted to add support for this to the `create-profile` datapoint, we would have to add a new parameter or something as we would only want to do this when the two are created together, not every time we create a new profile.

Is that something you feel we need to add right now? Technically, this is already the case with 1.8.

I'm not sure it's really necessary though, as it would only make a difference if the user created both a property and a profile, and then created another profile, but expected the first to be selected by default (which would probably happen anyways due to the current logic). It seems like an extreme edge case. Is there more to this that I'm missing?

felixarntz:
@aaemnnosttv Iterated a bit on this:
* Added `OAuth_Client::get_unsatisfied_scopes()`, to be used for #189.
* Fixed a bug that occurred because the error code for an invalid datapoint request was changed from `invalid_datapoint` to `400`. While the `WP_Error`s created from Google API exceptions use a numeric code that is interpreted as status code, we should generally stick to WordPress conventions of using a more descriptive error code, while handling the status via `data.status` for the REST API. Changing the error code here actually resulted in several requests failing, because we have some special handling around the `invalid_datapoint` error code for module notifications and module settings requests.
* Cleared old `POST:settings` Analytics datapoint completely.
* Some misc updates, like allowing `Insufficient_Scopes_Exception` instances to be populated from constructor so that we can just throw them (makes PHPCS recognize correctly whether we need more `throws` annotations).

felixarntz:
@aaemnnosttv Oh well, I just noticed you had introduced the `Invalid_Datapoint_Exception`, however as part of #1585, so technically this here would still be broken. I think it makes sense anyway to keep all standalone server-side changes in this PR here, so I'll just resolve the merge conflict in the other by combining our efforts 😂

aaemnnosttv:
Looks good to me @felixarntz !

Issue #1507, opened 2020-05-05
Ensure activating/deactivating a module refreshes user authentication state


felixarntz:
## Bug Description

Activating/deactivating a module may change the scopes that the current user needs to grant access for. While #1175 introduced the `core/user` store and #1179 introduced the `core/modules` store, these now need to integrate with each other so that changing module activation state results in the user authentication data to be refetched.

This is somewhat an enhancement, but also a bug since currently this would result in a broken experience.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Changing module activation state via the `core/modules` store actions should trigger a refetch of the `core/user/data/authentication` API datapoint (via `core/user` store action). The API call to activate/deactivate the module must happen first, so that the authentication data in the following call acts upon the new module state. This should be integrated in a way that the `core/modules` action only completes once it the `core/user` API fetch call has been completed.
* It must be ensured that this re-fetch actually performs the fetch. The `core/user/data/authentication` data is preloaded via the `apiFetch` preloading middleware, but here it must be bypassed (see #1478).

## Implementation Brief

In`assets/js/googlesitekit/modules/datastore/modules.js`:
* Create a new action constant called `REFETCH_AUTHENICATION`
* Add a `yield` statement to both `activateModule` and `deactivateModule` actions before the return that yields the following object`{ payload: {}, type: REFETCH_AUTHENTICATION}`
* import the STORE_NAME from `assets/js/googlesitekit/datastore/user` as `CORE_USER_STORE_NAME`
* Add a control for `REFETCH_AUTHENTICATION` [that uses `createRegistryControl`](https://github.com/google/site-kit-wp/blob/11e70c4c2cd51861a0db446da0fe995501f1ca12/assets/js/modules/analytics/datastore/settings.js#L66) from `Data` that that accepts  `params` object parameter and calls`registry.dispatch( CORE_USER_STORE_NAME).fetchAuthentication()`

In `assets/js/googlesitekit/datastore/user/authentication.js`
* Modify the `FETCH_AUTHENTICATION` control  to call to bypass caching with`return API.get( 'core', 'user', 'authentication', null, { useCache: false } )`

Note if #1175 has not been merged, this work should branch from the associated PR.

## QA Brief

* 

## Changelog entry

* Enhance functionality of new `core/modules` store so that module activation/deactivation results in authentication data to be refreshed.

felixarntz:
IB generally ✅ 

[...]

PR #1584, opened 2020-05-21
Bug/1507 ensure activating deactivating module refreshes auth state

jqlee85:
## Summary

Addresses issue #1507

## Relevant technical choices

Adds yield in `activateModule` and `deactivateModule` that returns `{ payload: {}, type: REFETCH_AUTHENTICATION }`.

## Checklist


felixarntz:
@jqlee85 Quick update, I think adding a `timestamp` query parameter to `refreshAuthentication` is all that's needed here to make this work. Then #1611 (IB has more context) is about fixing this throughout.

jqlee85:
@felixarntz sounds good. It looks like the `refreshAuthentication` utility already passes a unique timestamp each time:
https://github.com/google/site-kit-wp/blob/af5b679c11ba92b266f1a659f9b840823d8173b7/assets/js/util/refresh-authentication.js#L33

Is there something else that needs to be done then or does that resolve the issue?

felixarntz:
@jqlee85 The `refreshAuthentication` utility function isn't used by any of the new code though, it's only legacy code. So we'll need to implement that same timestamp addition to the `REFETCH_AUTHENTICATION` store action. Probably easiest to duplicate the logic that is in the `FETCH_AUTHENTICATION` control and add the timestamp there.

jqlee85:
@felixarntz ah yes, I see. I've made that update. Let me know if there are any further issues.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -197,6 +210,9 @@ export const controls = {
 	[ FETCH_MODULES ]: () => {
 		return API.get( 'core', 'modules', 'list', null, { useCache: false } );
 	},
+	[ REFETCH_AUTHENICATION ]: () => {
+		return API.get( 'core', 'user', 'authentication', null, { useCache: false, timestamp: Date.now() } );
```
felixarntz, line 214: I'm not sure this is right. Looking at `API.get`, the `timestamp` argument here wouldn't currently do anything, it needs to be in the `data` parameter (which is currently `null`) in order to affect the request URL.

jqlee85: @felixarntz Ah yes, I mistook the arguments there, thanks. I've moved the timestamp into the `data` arg, and added an `expect()` in the tests to make sure `timestamp` is included as a param for those GET requests to the auth endpoint. I also checked this locally, and confirmed the timestamp shows up in the GET request during the module deactivation. Let me know if there's anything else to button up here.

PR #1586, opened 2020-05-22
Refactored dashboard auth to check using required scopes and granted …

ShahAaron:
…scopes.

## Summary

Addresses issue #189 

## Relevant technical choices

Diverged from the IB as `currentAdminPage` returns the referring page and not the module which has not been setup correctly.

Changed to use `requiredScopes` and `grantedScopes` to retrieve the module

## Checklist


assets/js/components/notifications/dashboard-auth-alert.js
```diff
@@ -24,22 +24,23 @@ import { __, sprintf } from '@wordpress/i18n';
...
-	const { currentAdminPage } = global.googlesitekit.admin;
-	const product = currentAdminPage
-		.replace( /googlesitekit|module|-/g, ' ' )
-		.replace( /(^\w{1})|(\s{1}\w{1})/g, ( match ) => match.toUpperCase() )
-		.trim();
+	const { requiredScopes, grantedScopes } = global.googlesitekit.setup;
+
+	const productUrl = requiredScopes.filter( ( e ) => ! grantedScopes.includes( e ) )[ 0 ];
+	const productSlug = productUrl ? productUrl.toString().split( '/' ).pop() : '';
+	const product = getModulesData()[ productSlug ] ? getModulesData()[ productSlug ].name : '';
```
aaemnnosttv, line 36: This isn't sufficient to get the slug of the service/module as many scopes have `.readonly` or other variants tagged on the end. Let's use a regex to match `https://www.googleapis.com/auth/{THIS}.whatevermaybe`.

We should keep the page-based logic as a fallback as we should never use an empty string for `product` here.
Let's rename `product` to `productName` as well 👍

assets/js/components/notifications/dashboard-auth-alert.js
```diff
@@ -24,22 +24,23 @@ import { __, sprintf } from '@wordpress/i18n';
...
 
 	return (
 		<Notification
 			id="authentication error"
-			title={ __( 'Site Kit can’t access necessary data', 'google-site-kit' ) }
+			title={ __( "Site Kit can't access necessary data", 'google-site-kit' ) }
 			/* translators: %1$s: Product name */
-			description={ sprintf( __( 'Site Kit can’t access the relevant data from %1$s because you haven’t granted all API scopes requested during setup. To use Site Kit, you’ll need to redo the setup for %1$s – make sure to approve all API scopes at the authentication stage.', 'google-site-kit' ), product ) }
+			description={ sprintf( __( "Site Kit can't access the relevant data from %1$s because you haven't granted all API scopes requested during setup. To use Site Kit, you'll need to redo the setup for %1$s – make sure to approve all API scopes at the authentication stage. ", 'google-site-kit' ), product ) }
```
aaemnnosttv, line 43: This was probably done automatically by your editor, but these lines should be restored to the previous versions here https://github.com/google/site-kit-wp/pull/1586/commits/43ee72e9959db04816a871222f749abc4f81c342

ryanwelcher: Updated.

assets/js/components/notifications/dashboard-auth-alert.js
ryanwelcher: Updated

ryanwelcher:
@googlebot I consent.

ryanwelcher:
@googlebot I consent.

assets/js/components/notifications/dashboard-auth-alert.js
```diff
@@ -28,18 +28,26 @@ import Notification from '../notifications/notification';
...
-	const { currentAdminPage } = global.googlesitekit.admin;
-	const product = currentAdminPage
-		.replace( /googlesitekit|module|-/g, ' ' )
-		.replace( /(^\w{1})|(\s{1}\w{1})/g, ( match ) => match.toUpperCase() )
-		.trim();
-
+	const { unsatisfiedScopes } = global.googlesitekit.setup;
+	let message = '';
+	const missingScopes = [ ...new Set( unsatisfiedScopes.map( ( item ) => item.match( /https:\/\/www.googleapis.com\/auth\/([a-z]+)/ )[ 1 ] ) ) ];
+	const moduleNames = missingScopes.map( ( module ) => module.charAt( 0 ).toUpperCase() + module.slice( 1 ) );
```
aaemnnosttv, line 34: We should not be transforming the name from the scope like this as the matched string is only a "slug" at best. Instead we should be using `getModulesData()[ slug ].name` to get the localized name for the service (in a safer way of course to not assume the slug exists).

Also, this doesn't consider the few cases we have where the slug in the scope does not match exactly with one our module slugs. There are a few examples in the IB, but that's not an exhaustive list. We need to account for this mapping though.

assets/js/components/notifications/dashboard-auth-alert.js
```diff
@@ -28,18 +28,26 @@ import Notification from '../notifications/notification';
...
+	if ( 1 < moduleNames.length ) {
+		/* translators: %1$s: Product name */
+		message = sprintf( __( 'Site Kit can’t access the some relevant data because you haven’t granted all API scopes requested during setup. To use Site Kit, you’ll need to redo the setup for: %1$s – make sure to approve all API scopes at the authentication stage.', 'google-site-kit' ), moduleNames.join( ', ' ) );
```
aaemnnosttv, line 37: * In this case, let's use a single `%s` placeholder as only a single value is used once.
* The translators comment is slightly inaccurate as it is a list of multiple product names
* The list separator should be translatable as well. See https://github.com/google/site-kit-wp/blob/f4967c34d3d91eefddb9446b73614e1341e48b2f/includes/Core/Util/Debug_Data.php#L271-L272
* Let's split the function call over a few lines to keep it from running overly long and "hiding" the parameters filling the placeholders. (same applies to other `sprintf` calls here)

assets/js/components/notifications/dashboard-auth-alert.js
```diff
@@ -28,18 +28,26 @@ import Notification from '../notifications/notification';
...
+	} else if ( moduleNames.length === 1 ) {
+		/* translators: %1$s: Product name */
+		message = sprintf( __( 'Site Kit can’t access the relevant data from %1$s because you haven’t granted all API scopes requested during setup. To use Site Kit, you’ll need to redo the setup for %1$s – make sure to approve all API scopes at the authentication stage.', 'google-site-kit' ), moduleNames[ 0 ] );
+	} else {
+		// Generic error message here.
+		message = __( 'Site Kit can’t access the relevant data from because you haven’t granted all API scopes requested during setup. Please redo setup.', 'google-site-kit' );
```
aaemnnosttv, line 43: I realize we're pending final verbiage here but the "from because" here is a bit off 😄

aaemnnosttv:
@googlebot I consent.

assets/js/components/notifications/dashboard-auth-alert.js
```diff
@@ -24,26 +29,95 @@ import { __, sprintf } from '@wordpress/i18n';
...
+
+	let message;
+
+	switch ( messageID ) {
+		case MESSAGE_MULTIPLE:
+			/* translators: used between list items, there is a space after the comma. */
+			const listSeparator = __( ', ', 'google-site-kit' );
+			message = sprintf(
+				/* translators: %s: List of product names */
+				__( 'Site Kit can’t access the some relevant data because you haven’t granted all permissions requested during setup. To use Site Kit, you’ll need to redo the setup for: %s – make sure to approve all permissions at the authentication stage.', 'google-site-kit' ),
```
felixarntz, line 97:
```suggestion
				__( 'Site Kit can’t access all relevant data because you haven’t granted all permissions requested during setup. To use Site Kit, you’ll need to redo the setup for: %s – make sure to approve all permissions at the authentication stage.', 'google-site-kit' ),
```

assets/js/components/notifications/dashboard-auth-alert.js
```diff
@@ -24,26 +29,95 @@ import { __, sprintf } from '@wordpress/i18n';
...
 			id="authentication error"
 			title={ __( 'Site Kit can’t access necessary data', 'google-site-kit' ) }
-			/* translators: %1$s: Product name */
-			description={ sprintf( __( 'Site Kit can’t access the relevant data from %1$s because you haven’t granted all API scopes requested during setup. To use Site Kit, you’ll need to redo the setup for %1$s – make sure to approve all API scopes at the authentication stage.', 'google-site-kit' ), product ) }
-			handleDismiss={ () => {} }
+			description={ message }
 			format="small"
 			type="win-error"
-			isDismissable={ true }
+			isDismissable={ false }
```
felixarntz, line 120: Was this broken before? I see `handleDismiss` above is empty. Otherwise I'm curious why we make the notice non-dismissible here, since that's not part of the ACs afaict?

assets/js/components/notifications/dashboard-auth-alert.js
```diff
@@ -24,26 +29,95 @@ import { __, sprintf } from '@wordpress/i18n';
...
+			break;
+		case MESSAGE_SINGULAR:
+			message = sprintf(
+				/* translators: %1$s: Product name */
+				__( 'Site Kit can’t access the relevant data from %1$s because you haven’t granted all permissions requested during setup. To use Site Kit, you’ll need to redo the setup for %1$s – make sure to approve all permissions at the authentication stage.', 'google-site-kit' ),
+				moduleNames[ 0 ]
+			);
+			break;
+		case MESSAGE_GENERIC:
+			message = __( 'Site Kit can’t access all relevant data because you haven’t granted all permissions requested during setup. Please redo setup.', 'google-site-kit' );
```
felixarntz, line 109:
```suggestion
			message = __( 'Site Kit can’t access all relevant data because you haven’t granted all permissions requested during setup. To use Site Kit, you’ll need to redo the setup – make sure to approve all permissions at the authentication stage.', 'google-site-kit' );
```

assets/js/components/notifications/dashboard-auth-alert.js
aaemnnosttv: I believe so. I ran into this while testing where the notification should have been displayed but it wasn't because it was dismissed (very confusing). `handleDismiss` is not a prop for this component, but an instance method, so that wouldn't have done anything before.

aaemnnosttv:
> This requires an additional PHP change though I think, because this component now relies on the core/user store which is currently not a dependency here in PHP (googlesitekit-datastore-user). It may be best to add googlesitekit-datastore-site, googlesitekit-datastore-user, and googlesitekit-datastore-forms to the general $dependencies array in PHP anyway - I think we actually need that due to the changes in #1566

I think it probably makes sense for all the `core` datastores to always be available (at least the ones we have now). I'll add these to the list 👍

assets/js/components/notifications/dashboard-auth-alert.js
aaemnnosttv: Thanks, these are better 👍

PR #1591, opened 2020-05-25
Enhance/1211 back button

aaemnnosttv:
## Summary

Addresses issue #1211 (follow-up)

image

## Relevant technical choices

* Adds a "Back" button to the new Account Create screen  
This prevents a user from being stuck in this view by selecting the "Set up a new account" option.
    - Only adds this link if the user has any accounts, otherwise the Account Create view is the only one available to them anyways
* Adds a story to storybook for "No Accounts" using new component

## Checklist

### DIFF
--- a/assets/js/modules/analytics/common/account-create.js
+++ b/assets/js/modules/analytics/common/account-create.js
@@ -91,7 +91,7 @@ export default function AccountCreate() {
 	const { rollbackSettings } = useDispatch( STORE_NAME );
 	const handleBack = useCallback( () => rollbackSettings() );
 
-	if ( isDoingCreateAccount || isNavigating ) {
+	if ( isDoingCreateAccount || isNavigating || accounts === undefined ) {
 		return <ProgressBar />;
 	}
 
--- a/assets/js/modules/analytics/common/account-create.js
+++ b/assets/js/modules/analytics/common/account-create.js
@@ -26,6 +26,7 @@ import { useCallback, useState, useEffect } from '@wordpress/element';
  * Internal dependencies
  */
 import Button from '../../../components/button';
+import Link from '../../../components/link';
 import ProgressBar from '../../../components/progress-bar';
 import { trackEvent } from '../../../util';
 import TimezoneSelect from './account-create/timezone-select';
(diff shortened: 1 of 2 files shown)


assets/js/modules/analytics/common/account-create.js
```diff
@@ -123,12 +129,23 @@ export default function AccountCreate() {
...
-			</Button>
+			<div className="googlesitekit-setup-module__action">
+				<Button
+					disabled={ ! canSubmitAccountCreate }
+					onClick={ handleSubmit }
+				>
+					{ __( 'Create Account', 'google-site-kit' ) }
+				</Button>
+
+				{ ( accounts && !! accounts.length ) && (
```
felixarntz, line 140: Probably okay here, but this may result in the Back button suddenly appearing and possibly not being there right away although it should be. I wonder if it would be a cleaner approach to catch `accounts === undefined` in the if clause further above, returning the `ProgressBar` when accounts aren't loaded yet.

assets/js/modules/analytics/common/account-create.js
```diff
@@ -85,6 +87,10 @@ export default function AccountCreate() {
 		[ createAccount, setIsNavigating ]
 	);
 
+	// If the user clicks "Back", rollback settings to restore saved values, if any.
+	const { rollbackSettings } = useDispatch( STORE_NAME );
+	const handleBack = useCallback( () => rollbackSettings() );
```
felixarntz, line 92: Why do we need to rollback settings here?

aaemnnosttv: The minimum action needed is to change the `accountID` to not be `account_create`.

I thought it made sense to use the rollback for this as otherwise all previous selections would be lost as they are cleared when selecting `account_create`. This only really makes a difference on the settings view of course, but slightly better than just clearing the `accountID` I think.

assets/js/modules/analytics/common/account-create.js
aaemnnosttv: That shouldn't ever happen because that logic is already handled by `SettingsEdit`. I suppose it wouldn't hurt to add though.

felixarntz: Yeah I'd prefer for components to not rely on parent components where avoidable.

assets/js/modules/analytics/common/account-create.js
```diff
@@ -85,6 +87,10 @@ export default function AccountCreate() {
...
+
 	if ( isDoingCreateAccount || isNavigating ) {
```
felixarntz, line 94:
```suggestion
	if ( isDoingCreateAccount || isNavigating || accounts === undefined ) {
```

Issue #1593, opened 2020-05-26
Clear legacy caches when submitting module changes


felixarntz:
## Bug Description

With AdSense and Analytics setup/settings already refactored, both of these modules now invalidate the browser cache via `googlesitekit-api`, to ensure no stale data is being displayed (which could apply to e.g. different accounts). However, what is not covered currently is the cache used by the legacy API layer `assets/js/components/data`, which is at this point actually more relevant because pretty much the entire dashboard and module pages rely on that.

We should fix this for the coming release, which includes the AdSense refactoring and adds the prominent Analytics provisioning feature.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `submitChanges` actions of `modules/adsense` and `modules/analytics` should also invalidate caches for the legacy API layer `assets/js/components/data` (see `invalidateCacheGroup` function).
* These function calls should be annotated with TODOs to explain why they're there and that they should be removed in the future.

## Implementation Brief

* In `assets/js/modules/adsense/settings.js` and `assets/js/modules/analytics/settings.js`:
    * Import `legacyDataAPI` from `assets/js/components/data`.
    * Call `legacyDataAPI.invalidateCacheGroup( 'modules', 'adsense' )` / `legacyDataAPI.invalidateCacheGroup( 'modules', 'analytics' )` right after the existing calls to invalidate the new caches.
    * Add TODO comments to both legacy API calls.

## QA Brief

For both Analytics and AdSense modules:
* Set up the module normally
* Go to the main SK dashboard (this primes the caches)
* Refresh the page, you should not see network requests for the module data
* Edit the module settings and make a change (e.g. toggle snippet output)
* Save settings
* Navigate back to SK dashboard while observing network requests
    * Network requests should be made for the module data on the first request to the main dashboard

## Changelog entry

* Ensure all module caches are cleared when modifying Analytics settings.

aaemnnosttv:
IB ✅

ryanwelcher:
Tested locally:
* Loaded Dashboard to prime caches and then confirmed that page reload fires no network requests.
* Changed AdSense settings to not insert the snippet and confirmed that Dashboard made a network request for module data. 
* Refreshing Dashboard did not fire another request for module data.
* Change Analytics settings to not insert the snippet and confirmed that Dashboard made a network request for module data.
* Refreshing Dashboard did not fire another request for module data.

QA ✅

PR #1595, opened 2020-05-27
Invalidate legacy cache groups on submit changes

aaemnnosttv:
## Summary

Addresses issue #1593

## Relevant technical choices

* Kept `dataAPI` name in import for consistency

## Checklist

### DIFF
--- a/assets/js/modules/adsense/datastore/settings.js
+++ b/assets/js/modules/adsense/datastore/settings.js
@@ -220,6 +220,7 @@ export const controls = {
 		}
 
 		await API.invalidateCache( 'modules', 'adsense' );
+		// TODO: Remove once legacy dataAPI is no longer used.
 		dataAPI.invalidateCacheGroup( TYPE_MODULES, 'adsense' );
 
 		return {};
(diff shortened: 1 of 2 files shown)


assets/js/modules/adsense/datastore/settings.js
```diff
@@ -219,6 +220,9 @@ export const controls = {
 		}
 
 		await API.invalidateCache( 'modules', 'adsense' );
+		// TODO: Remove once legacy dataAPI is no longer used.
+		dataAPI.invalidateCacheGroup( TYPE_MODULES, 'adsense' );
```
felixarntz, line 224: Just curious whether importing the object significantly increases the bundle size for these module assets? If so, adding an extra export for just the function would reduce that overhead?

aaemnnosttv: Good question 😄 

Apparently it has no effect on the AdSense module bundle because it was already in there to be used by `isAdsenseConnectedAnalytics`.
https://github.com/google/site-kit-wp/blob/3a21b1a16689d3e147b7a44bd0d07df4de604b17/assets/js/modules/adsense/util/index.js#L29

For Analytics, it increased the production bundle size from `28.72KB` to `34.4KB`. I would say that is probably not worth splitting out another export for, although I would imagine the amount of code we actually need would probably leave this under 30KB, but that's just a guess.

felixarntz: Let's just keep it as it is for now.

PR #1599, opened 2020-05-28
Update scope handling with unsatisfied scopes

aaemnnosttv:
## Summary

Addresses issue #1566

## Relevant technical choices

* Updates scope granted status in Site Health using new scope-satisfaction logic
* Added `unsatisfiedScopes` to `googlesitekit.setup` data
* Added `unsatisfiedScopes` to `core/user/data/authentication` response
* Updated `refreshAuthentication` to include `unsatisfiedScopes`
* Updated `get_unsatisfied_scopes` to always return a list as `array_filter` preserves keys which causes the array to be encoded as an object otherwise

## Checklist

### DIFF
--- a/includes/Core/Authentication/Authentication.php
+++ b/includes/Core/Authentication/Authentication.php
@@ -732,9 +732,10 @@ final class Authentication {
 							$access_token = $oauth_client->get_access_token();
 
 							$data = array(
-								'authenticated'  => ! empty( $access_token ),
-								'requiredScopes' => $oauth_client->get_required_scopes(),
-								'grantedScopes'  => ! empty( $access_token ) ? $oauth_client->get_granted_scopes() : array(),
+								'authenticated'     => ! empty( $access_token ),
+								'requiredScopes'    => $oauth_client->get_required_scopes(),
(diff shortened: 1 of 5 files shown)


includes/Core/Authentication/Authentication.php
```diff
@@ -633,7 +633,8 @@ private function inline_js_setup_data( $data ) {
 		$data['isAuthenticated']    = ! empty( $access_token );
 		$data['requiredScopes']     = $auth_client->get_required_scopes();
 		$data['grantedScopes']      = ! empty( $access_token ) ? $auth_client->get_granted_scopes() : array();
-		$data['needReauthenticate'] = $data['isAuthenticated'] && $auth_client->needs_reauthentication();
+		$data['unsatisfiedScopes']  = ! empty( $access_token ) ? $auth_client->get_unsatisfied_scopes() : array();
+		$data['needReauthenticate'] = $auth_client->needs_reauthentication();
```
aaemnnosttv, line 637: `needs_reauthentication` already checks for the access token internally, so no need to double-check it here.

Issue #1559, opened 2020-05-14
Add getAdminScreenURL and getAdminReauthURL selectors to base module store


felixarntz:
## Feature Description

stores created by `createModuleStore` should include new selectors for module-specific admin URLs. These can be built on top of the enhanced version of `getAdminURL` from the `core/site` store, which was implemented as part of #1500.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `core/user/data/authentication` REST route should also return `needsReauthentication`, which uses `Authentication::need_reauthenticate()` (will be migrated to `OAuth_Client` class via #1566, just fyi - depending on timeline this may overlap).
* The `core/user` store should receive an additional selector `needsReauthentication()` that exposes the above.
* `createModuleStore` should receive a new `options` argument `adminPage`. The default should be "googlesitekit-dashboard". Modules with their own screens (Search Console, Analytics, AdSense) need to be amended so that they provide the slugs for their own screens when calling `createModuleStore`.
* `createModuleStore` should receive a new `options` argument `requiresSetup`. The default should be a boolean `true`. Modules that do not require setup (only PageSpeed Insights for now) need to be amended so that they provide `false` for this argument.
* Store objects returned by `createModuleStore` should have two additional selectors:
    * `getAdminScreenURL( queryArgs )`: Should use `getAdminURL` from `core/site` passing `adminPage` as the selector's `page` parameter, and `queryArgs` can be passed through.
    * `getAdminReauthURL()`: Should mirror the exact behavior of the legacy `getReAuthURL` utility function, based on `getAdminScreenURL`, whether the module `requiresSetup`, and whether the user needs to re-authenticate (requested scopes and granted scopes are available via `core/user` score, let's introduce a new selector to simplify this check, see below).
* Refactored module code based on the datastore that currently still relies on one of the old utility functions (e.g. `getReAuthURL`, used at least by refactored AdSense code) should be modified to use the new selectors.

## Implementation Brief

* Update the `core/user/data/authentication` REST endpoint in `includes/Core/Authentication/Authentication.php` to also return `needsReauthentication` which will get its value from the `OAuth_Client::needs_reauthentication()` function.

* Add `needsReauthentication()` selector to `assets/js/googlesitekit/datastore/user/authentication.js` which connects to the needsReauthentication value returned from the above endpoint.

* Add these arguments to `options` in `createModuleStore()` in `assets/js/googlesitekit/modules/create-module-store.js`
[...]

PR #1604, opened 2020-05-29
Added needsreauthentication to core user authentication endpoint.

ShahAaron:
## Summary

Addresses issue #1559 

## Checklist


assets/js/googlesitekit/modules/create-info-store.js
```diff
@@ -0,0 +1,102 @@
...
+		 * Returns admin screen url.
+		 *
+		 * @since n.e.x.t
+		 *
+		 * @param {Object} state Data store's state.
+		 * @param {(Object|undefined)} queryArgs Query arguments to add to admin URL.
+		 * @return {(string|undefined)} The admin screen url.
+		 */
+		getAdminScreenURL: createRegistrySelector( ( select ) => ( state, queryArgs ) => {
+			return select( CORE_SITE ).getAdminURL( adminPage, queryArgs );
```
tofumatt, line 65: Passing `queryArgs` here won't work as-expected, because the first argument for `getAdminURL` is `page`. So this is going to pass the query args to the wrong argument. Since `page` is an optional argument, it looks like you want:

```suggestion
			return select( CORE_SITE ).getAdminURL( adminPage, undefined, queryArgs );
```

assets/js/googlesitekit/modules/create-info-store.js
```diff
@@ -0,0 +1,102 @@
...
+		/**
+		 * Returns admin reauth url.
+		 *
+		 * @since n.e.x.t
+		 *
+		 * @param {Object} state Data store's state.
+		 * @param {(Object|undefined)} queryArgs Query arguments to add to admin URL.
+		 * @return {(string|undefined)} The admin reauth url.
+		 */
+		getAdminReauthURL: createRegistrySelector( ( select ) => ( state, queryArgs ) => {
```
tofumatt, line 77: There aren't any mention of `queryArgs` for this selector in the IB, so I think you can remove it, from my understanding.
```suggestion
		getAdminReauthURL: createRegistrySelector( ( select ) => ( state ) => {
```

assets/js/googlesitekit/modules/create-info-store.js
```diff
@@ -0,0 +1,102 @@
...
+			const noSetupQueryArgs = ! requiresSetup ? {
+				notification: 'authentication_success',
+				reAuth: undefined,
+			} : {};
+
+			if ( ! needsReauthentication ) {
+				return select( CORE_SITE ).getAdminURL( adminPage, noSetupQueryArgs );
+			}
+
+			return select( CORE_SITE ).getAdminURL( adminPage, queryArgs );
```
tofumatt, line 89: If removed above, you'll need to remove the args here too (fwiw, these also would have went to the `page` arg, not the extra `queryArgs` argument):

```suggestion
			return select( CORE_SITE ).getAdminURL( adminPage );
```

assets/js/googlesitekit/modules/create-info-store.js
ShahAaron: Done

assets/js/googlesitekit/modules/create-info-store.js
ShahAaron: Done

assets/js/googlesitekit/modules/create-info-store.js
ShahAaron: Fixed. I've changed this to use getAdminScreenURL

assets/js/googlesitekit/modules/create-info-store.js
ShahAaron: @tofumatt could I confirm this with you. The first argument is in fact `page` but I'm passing `adminPage` to that, so the second argument should be `queryArgs` not undefined otherwise no args get passed.

assets/js/googlesitekit/datastore/user/authentication.js
```diff
@@ -239,6 +239,22 @@ export const selectors = {
...
 		return unsatisfiedScopes;
 	} ),
+
+	/**
+	 * Checks reauthentication status for this user.
+	 *
+	 * Returns true if any required scopes are not satisfied or undefined
+	 * if reauthentication info is not available/loaded.
+	 *
+	 * @since 1.9.0
```
felixarntz, line 249: Should be `@since n.e.x.t`, exact version numbers will always be added as part of the release preparation.

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -430,5 +431,50 @@ describe( 'core/user authentication', () => {
 				expect( unsatisfiedScopes ).toEqual( undefined );
 			} );
 		} );
+
+		describe( 'needsReauthentication', () => {
```
felixarntz, line 435: This selector should also have the "uses a resolver get all authentication info" test like the other ones do.

assets/js/googlesitekit/modules/create-info-store.js
```diff
@@ -0,0 +1,99 @@
...
+const { createRegistrySelector } = Data;
+
+/**
+ * Creates a store object that has selectors for managing site info.
+ *
+ * @since 1.6.0
+ * @private
+ *
+ * @param {Object}  options               Options to consider for the store.
+ * @param {number}  options.storeName     Store name to use.
```
felixarntz, line 35: Since the store name here is required for the logic below to work, I'd argue it should have a default based on a required `slug` parameter (like `createModuleStore`). We'll also need this `slug` parameter further below in `getAdminReauthURL` anyway (see other comment).

assets/js/googlesitekit/modules/create-info-store.js
```diff
@@ -0,0 +1,99 @@
...
+ * @param {string}  options.adminPage     Store admin page. Default is 'googlesitekit-dashboard'.
+ * @param {boolean} options.requiresSetup Store flag, for requires setup. Default is 'true'
+ * @return {Object} The info store object.
+ */
+export const createInfoStore = ( {
+	storeName = undefined,
+	adminPage = undefined,
+	requiresSetup = undefined,
```
felixarntz, line 43: These two arguments are missing their defaults here.

assets/js/googlesitekit/modules/create-info-store.js
```diff
@@ -0,0 +1,99 @@
...
+	};
+	const resolvers = {};
+	const selectors = {
+		/**
+		 * Returns admin screen url.
+		 *
+		 * @since n.e.x.t
+		 *
+		 * @param {(Object|undefined)} queryArgs Query arguments to add to admin URL.
+		 * @return {(string|undefined)} The admin screen url.
```
felixarntz, line 61:
```suggestion
		 * @return {(string|undefined)} The admin screen URL.
```

assets/js/googlesitekit/modules/create-info-store.js
```diff
@@ -0,0 +1,99 @@
...
+	const INITIAL_STATE = {};
+	const actions = {};
+	const controls = {};
+	const reducer = ( state ) => {
+		return { ...state };
+	};
+	const resolvers = {};
+	const selectors = {
+		/**
+		 * Returns admin screen url.
```
felixarntz, line 56:
```suggestion
		 * Returns admin screen URL.
```

assets/js/googlesitekit/modules/create-info-store.js
```diff
@@ -0,0 +1,99 @@
...
+		 *
+		 * @param {(Object|undefined)} queryArgs Query arguments to add to admin URL.
+		 * @return {(string|undefined)} The admin screen url.
+		 */
+		getAdminScreenURL: createRegistrySelector( ( select ) => ( queryArgs ) => {
+			return select( CORE_SITE ).getAdminURL( adminPage, queryArgs );
+		} ),
+
+		/**
+		 * Returns admin reauth url.
```
felixarntz, line 68:
```suggestion
		 * Returns admin reauthentication URL.
```

assets/js/googlesitekit/modules/create-info-store.js
```diff
@@ -0,0 +1,99 @@
...
+		 *
+		 * @since n.e.x.t
+		 *
+		 * @return {(string|undefined)} The admin reauth url.
```
felixarntz, line 72:
```suggestion
		 * @return {(string|undefined)} The admin reauthentication URL.
```

assets/js/googlesitekit/modules/create-info-store.js
```diff
@@ -0,0 +1,99 @@
...
+			const noSetupQueryArgs = ! requiresSetup ? {
+				notification: 'authentication_success',
+				reAuth: undefined,
+			} : {};
+
+			if ( ! needsReauthentication ) {
+				return select( STORE_NAME ).getAdminScreenURL( noSetupQueryArgs );
+			}
+
+			return select( CORE_SITE ).getAdminScreenURL();
```
felixarntz, line 86: The logic in this method is not quite right. It needs to match _exactly_ what the current `getReAuthURL()` utility function does, except having the specifics being applied based on `adminPage` and `requiresSetup`. Some issues here are:
* Missing query parameters `slug` and `reAuth` (the `status` value from the original `getReAuthURL` function should always be `true` here, since we only use this URL for module activations as of now, never deactivations).
* Missing usage of the `connectURL`: The URL that's being built here should be like the `redirect` variable in the original `getReAuthURL` function; if `! needsReauthentication`, it should be returned as is, otherwise it should be added as `redirect` query parameter to the `connectURL` (with similar encoding etc. like in the original function), alongside `status: true`.
* This function should never return the general `select( CORE_SITE ).getAdminScreenURL()`.

assets/js/googlesitekit/modules/create-module-store.js
```diff
@@ -68,16 +73,25 @@ export const createModuleStore = ( slug, {
 		storeName,
 	} );
 
+	const infoStore = createInfoStore( {
```
felixarntz, line 76: See other comment, let's pass `slug` here as first parameter.

assets/js/modules/adsense/settings/settings-setup-incomplete.js
```diff
@@ -48,7 +47,7 @@ export default function SettingsSetupIncomplete() {
 				className="googlesitekit-settings-module__edit-button"
 				href={
 					// TODO: Replace this function with something datastore-driven.
-					getReAuthURL( 'adsense', true )
+					useSelect( ( select ) => select( STORE_NAME ).getAdminReauthURL() )
```
felixarntz, line 50: This should be better assigned to a variable further above with the other hooks instead of inline here.

assets/js/modules/pagespeed-insights/settings/settings-view.js
```diff
@@ -26,11 +26,11 @@ import { __, sprintf } from '@wordpress/i18n';
  */
 import Data from 'googlesitekit-data';
 import { sanitizeHTML } from '../../../util';
-import { STORE_NAME as CORE_SITE } from '../../../googlesitekit/datastore/site/constants';
+import { STORE_NAME } from '../datastore/constants';
 const { useSelect } = Data;
 
 export default function SettingsView() {
-	const dashboardPermalink = useSelect( ( select ) => select( CORE_SITE ).getAdminURL( 'googlesitekit-dashboard' ) );
+	const dashboardPermalink = useSelect( ( select ) => select( STORE_NAME ).getAdminReauthURL() );
```
felixarntz, line 33: This should not use `getAdminReauthURL`, but `getAdminScreenURL` - it's just for viewing the data, not for setting up the module.

assets/js/modules/search-console/datastore/index.js
```diff
@@ -29,6 +29,7 @@ const baseModuleStore = Modules.createModuleStore( 'search-console', {
 	settingSlugs: [
 		'propertyID',
 	],
+	adminPage: 'googlesitekit-module-search-console',
```
felixarntz, line 32: This should also receive `requiresSetup: false`, since Search Console does not have any setup in the plugin.

package-lock.json
```diff
@@ -35482,7 +35482,7 @@
       "dependencies": {
         "strip-ansi": {
           "version": "3.0.1",
-          "resolved": "http://registry.npmjs.org/strip-ansi/-/strip-ansi-3.0.1.tgz",
+          "resolved": "https://registry.npmjs.org/strip-ansi/-/strip-ansi-3.0.1.tgz",
```
felixarntz, line 35485: The changes in this file seem irrelevant, so that should be reverted?

tofumatt:
@googlebot I consent.

assets/js/googlesitekit/modules/create-info-store.js
ShahAaron: Slug added and defaults for `storename`

assets/js/modules/search-console/datastore/index.js
ShahAaron: Added with default to false

assets/js/googlesitekit/datastore/user/authentication.js
ShahAaron: Done

assets/js/googlesitekit/modules/create-info-store.js
ShahAaron: Defaults added

assets/js/googlesitekit/modules/create-info-store.js
ShahAaron: Done

assets/js/googlesitekit/modules/create-info-store.js
ShahAaron: Done

assets/js/googlesitekit/modules/create-info-store.js
ShahAaron: Done

assets/js/googlesitekit/modules/create-info-store.js
ShahAaron: Done

assets/js/googlesitekit/modules/create-module-store.js
ShahAaron: Slug added to parameters

assets/js/modules/adsense/settings/settings-setup-incomplete.js
ShahAaron: Variable added - adminReauthURL

assets/js/modules/pagespeed-insights/settings/settings-view.js
ShahAaron: Done

assets/js/googlesitekit/datastore/user/authentication.test.js
ShahAaron: Test added

assets/js/googlesitekit/modules/create-info-store.js
ShahAaron: @felixarntz could I clarify this change with you. 

I've added `reAuth` to the parameter list for `getAdminReauthURL`, status is obtained as it is passed into the store. These 2 arguments are then passed to `getAdminScreenURL`.

My question surrounds the use of `connectURL or `adminURL` as the original function `getReAuthURL` required these as it was building the entire URL string but the new `getAdminURL` doesn't require them as it builds the base url and and adds the args to it. So my question is do we actually need `connectURL or `adminURL`?

felixarntz: @ShahAaron We need `getAdminURL` to build the URL to the actual module page (or dashboard). But then, if `needsReauthentication` is `true`, we also need the connect URL (see `getReAuthURL` function's last few lines). There's the `getConnectURL` selector in `core/user` to retrieve that, and we'd need to pass the original URL built as `redirectURL` argument there.

felixarntz:
@ShahAaron Something minor I just thought about that we'll need to do in this PR as well is add `googlesitekit-datastore-site` and `googlesitekit-datastore-user` as dependencies to the `googlesitekit-modules` asset in PHP, in the `Assets` class. That's necessary since now `createModuleStore` which is part of `googlesitekit-modules` relies on data from `core/site` and `core/user`.

assets/js/googlesitekit/modules/create-info-store.js
ShahAaron: I have refactored this to include `connectURL` where the `needsReauthentication` is true

Issue #1491, opened 2020-05-04
Adsense connection stalled by Ad blocker


ernee:
## Bug Description

When using an Ad blocker extension in the browser, Adsense gets stuck at "locating accounts" and does not complete the process. Console error shows ERR_BLOCKED_BY_CLIENT on its request. 

Disabling Ad blocker on the page allows the process to complete.  

## Steps to reproduce

1. Connect the Adsense service in a browser that has an Ad blocker extension (tested[ with this one](https://chrome.google.com/webstore/detail/adblock-%E2%80%94-best-ad-blocker/gighmmpiobklfepjocnamgkkbiglidom?hl=en-US))
2. On the Adsense admin screen, see the message that Adsense is "locating accounts" with progress bar loading

## Screenshots

![Adsense and Adblock](https://user-images.githubusercontent.com/8108643/80989525-e8272f80-8de9-11ea-81cf-83f53ebf6600.png)

## Additional Context

 - PHP Version: 7.3.16
 - OS: MacOS
 - Browser: Chrome
 - Plugin Version: 1.7.1
 - Device: MacBook Air

related support topic: https://wordpress.org/support/topic/google-site-kit-adsense/

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* When an adblocker browser plugin is active, the AdSense setup flow should display the adblocker warning as expected.
* This should be tested with a few popular adblocker extensions (particularly, with the original point of failure extension linked above).

## Implementation Brief

* Rename our adblocker "detection" script from `ads.js` to `pagead2.ads.js` so that it is more appropriately blocked for AdSense requests (that string is blocked also by AdBlock extension), which is what matters most for Site Kit usage.

## QA Brief

* 

## Changelog entry

* Fix ad blocker detection failing for popular AdBlock browser extension.

felixarntz:
@ernee Have you been able to replicate that?

This is an expected error, but the user shouldn't even be able to get there typically, since we would show the AdBlocker warning that is baked into the plugin. So it looks like we have a bug that the warning doesn't show here.

ernee:
@felixarntz yes, the above is from my personal site. No Adblocker warning was displayed for me- just the stalled experience.

jamesozzie:
Similar issue in the following support topic: [https://wordpress.org/support/topic/site-kit-adsense-activation/#post-12887100](https://wordpress.org/support/topic/site-kit-adsense-activation/#post-12887100)

[...]

PR #1626, opened 2020-06-02
Changed the file name for ads blocker javascript file.

ShahAaron:
## Summary

Addresses issue #1491 

## Checklist


webpack.config.js
```diff
@@ -113,7 +113,7 @@ const webpackConfig = ( mode ) => {
 				'googlesitekit-admin': './assets/js/googlesitekit-admin.js',
 				'googlesitekit-module': './assets/js/googlesitekit-module.js',
 				// Needed to test if a browser extension blocks this by naming convention.
-				ads: './assets/js/ads.js',
+				ads: './assets/js/adsense.js',
```
felixarntz, line 116: Would be good to rename the entrypoint here too, just for consistency.

webpack.config.js
```diff
@@ -113,7 +113,7 @@ const webpackConfig = ( mode ) => {
 				'googlesitekit-admin': './assets/js/googlesitekit-admin.js',
 				'googlesitekit-module': './assets/js/googlesitekit-module.js',
 				// Needed to test if a browser extension blocks this by naming convention.
-				adsense: './assets/js/adsense.js',
+				'dfp.js.ads': './assets/js/dfp.js.ads.js',
```
aaemnnosttv, line 116: I think it makes more sense to use use a pattern that matches AdSense as much as possible rather than a service we're not using.

I found that giving the script a name of `pagead2.ads.js` was also detected by AdBlock as well as another popular adblocker AdGuard. `pagead2` comes from the URL we use for loading `adsbygoogle.js` (unfortunately this is not detected).

The various patterns used by AdBlock can be found here:
https://easylist-downloads.adblockplus.org/easylist.txt
There are a few other lists it uses but those seem to be for exceptions or other rules.
Some of these patterns also seem to be matching for CSS classes or other selectors, not just URL fragments.

felixarntz: SGTM, will change it to use `pagead2.ads.js`.

Issue #1630, opened 2020-06-03
Ensure Site Kit tracking events are sent


felixarntz:
## Bug Description

In order to record tracking events with `gtag.js`, we're pushing event data onto the `dataLayer` stack as documented. However, as `gtag.js` works asynchronously, our current approach does not ensure these events are actually collected by Analytics - especially since the actual code we're writing simply adds data to an observed array.

[`gtag.js` provides a mechanism to know when an event was actually sent](https://developers.google.com/analytics/devguides/collection/gtagjs/sending-data#know_when_an_event_has_been_sent), so we should use this in the plugin. It will make the `trackEvent` function asynchronous, but that is okay. In most cases, we shouldn't even need to update the function calls - it's only really necessary in the cases where right afterwards e.g. a navigation away from the current page happens (e.g. module activation/deactivation or setup completion).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Site Kit's tracking events should use the `event_callback` argument for `gtag.js` to "ensure" the event is actually sent - "ensure" in double quotes because there should still be a 1 second timeout in case there is a problem, since event tracking should not block the user or result in any user-facing error.
* If an event is not sent within the 1 second timeout, a console warning should be triggered.
* Code that calls the `trackEvent` function should do so asynchronously in every case where there is a navigation away from the page right after.

## Implementation Brief

* Return a promise in `trackEvent` that always resolves (consuming code should await, but doesn't need awareness of whether the event was sent in time or not).
* Pass `event_callback` among the arguments to `dataLayerPush`, making it resolve the promise.
* Use `setTimeout` with 1000ms to resolve the promise, triggering a console warning in this case.
* In the `event_callback` function, clear the timeout. Also add a safeguard in place to ensure the promise is only resolved once: Theoretically (probably not practically) the `event_callback` could kick in after 1 second, but in this case the promise would already been resolved by the timeout.
* Merge #1631.

## QA Brief

* 

## Changelog entry

* N/A

aaemnnosttv:
IB ✅

tofumatt:
QA ✅

PR #1631, opened 2020-06-03
Use asynchronous event tracking and add awareness for whether it was successfully submitted

felixarntz:
## Summary

Addresses issue #1630 

## Relevant technical choices

* Improves relevant test coverage for `trackEvent`.

## Checklist

### DIFF
--- a/assets/js/components/activation/activation-app.js
+++ b/assets/js/components/activation/activation-app.js
@@ -43,15 +43,19 @@ export class ActivationApp extends Component {
 			buttonLabel = __( 'Go to Dashboard', 'google-site-kit' );
 		}
 
+		const onButtonClick = async ( event ) => {
+			event.preventDefault();
+			await trackEvent( 'plugin_setup', proxySetupURL ? 'proxy_start_setup_banner' : 'goto_sitekit' );
+			global.location.assign( buttonURL );
+		};
+
 		return (
 			<Fragment>
 				<NotificationCounter />
 				<ActivationMain
 					buttonURL={ buttonURL }
 					buttonLabel={ buttonLabel }
-					onButtonClick={ () => {
-						trackEvent( 'plugin_setup', proxySetupURL ? 'proxy_start_setup_banner' : 'goto_sitekit' );
-					} }
+					onButtonClick={ onButtonClick }
 				/>
 			</Fragment>
 		);
--- a/assets/js/util/tracking/createTrackEvent.js
+++ b/assets/js/util/tracking/createTrackEvent.js
@@ -55,7 +55,7 @@ export default function createTrackEvent( config, dataLayerTarget ) {
 				}
 				resolved = true;
 				if ( ! eventSent ) {
-					console.warn( `Tracking event "${ eventName }" took too long to fire:`, eventData ); // eslint-disable-line no-console
+					global.console.warn( `Tracking event "${ eventName }" (category "${ eventCategory }") took too long to fire.` );
 				}
 				resolve();
 			};
--- a/assets/js/util/tracking/createTrackEvent.js
+++ b/assets/js/util/tracking/createTrackEvent.js
@@ -21,6 +21,7 @@ export default function createTrackEvent( config, dataLayerTarget ) {
 	 * @param {string} eventName The event category. Required.
 	 * @param {string} eventLabel The event category. Optional.
 	 * @param {string} eventValue The event category. Optional.
+	 * @return {Promise} Promise that always resolves.
 	 */
 	return function trackEvent( eventCategory, eventName, eventLabel = '', eventValue = '' ) {
 		const {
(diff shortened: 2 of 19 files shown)


assets/js/util/tracking/index.test.js
```diff
@@ -68,35 +76,76 @@ describe( 'trackEvent', () => {
...
+			[ DATA_LAYER ]: {
+				push: ( args ) => args[ 2 ].event_callback(),
+			},
+		};
+		const { trackEvent } = createTracking( { trackingEnabled: true }, dataLayer );
+
+		const consoleWarnSpy = jest.spyOn( global.console, 'warn' );
+		await fakeTimeouts( () => trackEvent( 'test-category', 'test-name', 'test-label', 'test-value' ) );
+		expect( consoleWarnSpy ).not.toHaveBeenCalled();
+		consoleWarnSpy.mockClear();
```
aaemnnosttv, line 134: This should use `mockRestore` to return the `warn` to the original unmocked function
```suggestion
		consoleWarnSpy.mockRestore();
```

assets/js/util/tracking/index.test.js
```diff
@@ -68,35 +76,76 @@ describe( 'trackEvent', () => {
...
+				push: () => {},
+			},
+		};
+		const { trackEvent } = createTracking( { trackingEnabled: true }, dataLayer );
+
+		muteConsole( 'warn' );
+		const consoleWarnSpy = jest.spyOn( global.console, 'warn' );
+		await fakeTimeouts( () => trackEvent( 'test-category', 'test-name', 'test-label', 'test-value' ) );
+		expect( consoleWarnSpy ).toHaveBeenCalledWith( 'Tracking event "test-name" (category "test-category") took too long to fire.' );
+		consoleWarnSpy.mockClear();
```
aaemnnosttv, line 149:
```suggestion
		consoleWarnSpy.mockRestore();
```

assets/js/util/tracking/index.test.js
```diff
@@ -68,35 +76,76 @@ describe( 'trackEvent', () => {
...
+		const dataLayer = {
+			[ DATA_LAYER ]: {
+				push: () => {},
+			},
+		};
+		const { trackEvent } = createTracking( { trackingEnabled: true }, dataLayer );
+
+		muteConsole( 'warn' );
+		const consoleWarnSpy = jest.spyOn( global.console, 'warn' );
+		await fakeTimeouts( () => trackEvent( 'test-category', 'test-name', 'test-label', 'test-value' ) );
```
aaemnnosttv, line 147: We should probably also assert that the promise resolved with `false` as well here.
```suggestion
		const resolvedValue = await fakeTimeouts( () => trackEvent( 'test-category', 'test-name', 'test-label', 'test-value' ) );
		expect( resolvedValue ).toBe( false );
```

assets/js/util/tracking/index.test.js
```diff
@@ -68,35 +76,76 @@ describe( 'trackEvent', () => {
...
+	it( 'resolves without warning when event_callback is called', async () => {
+		const dataLayer = {
+			[ DATA_LAYER ]: {
+				push: ( args ) => args[ 2 ].event_callback(),
+			},
+		};
+		const { trackEvent } = createTracking( { trackingEnabled: true }, dataLayer );
+
+		const consoleWarnSpy = jest.spyOn( global.console, 'warn' );
+		await fakeTimeouts( () => trackEvent( 'test-category', 'test-name', 'test-label', 'test-value' ) );
```
aaemnnosttv, line 132: We should also assert it is resolved with `true` here.

assets/js/components/activation/activation-app.js
```diff
@@ -43,15 +43,19 @@ export class ActivationApp extends Component {
 			buttonLabel = __( 'Go to Dashboard', 'google-site-kit' );
 		}
 
+		const onButtonClick = async ( event ) => {
```
aaemnnosttv, line 46: Ideally this would use `useCallback` to prevent `ActivationMain` from rerendering each time this component renders because `onButtonClick` will be different. Looks like this component is basically a functional component already so this would be a very straightforward refactor.

As a class component, I believe the alternative is to add a constructor and add the callback as a class method.
E.g.
```js
constructor( props ) {
    super( props );
    this.onButtonClick = this.onButtonClick.bind( this );
}
```
Not a big deal in this case since this component probably doesn't render often but changing it to be a functional component is trivial.

assets/js/util/tracking/createTrackEvent.js
```diff
@@ -32,17 +33,49 @@ export default function createTrackEvent( config, dataLayerTarget ) {
...
+				}
+				resolve();
+			};
+
+			// This timeout ensures a tracking event does not block the user
+			// event if it is not sent (in time).
+			// If this fails, it shouldn't reject the promise since event
+			// tracking should not result in user-facing errors. It will just
+			// trigger a console warning.
+			const failTimeout = setTimeout( () => {
```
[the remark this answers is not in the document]
felixarntz: I intentionally didn't put it there because `resolveOnce` ensures that it's only called once; if the timeout is hit where before the `event_callback` was called, `resolveOnce` should still only be run once due to its cancel logic in the beginning. We don't wanna trigger the console warning if the `event_callback` was actually called before the timeout.

assets/js/util/tracking/createTrackEvent.js
```diff
@@ -21,6 +21,7 @@ export default function createTrackEvent( config, dataLayerTarget ) {
 	 * @param {string} eventName The event category. Required.
 	 * @param {string} eventLabel The event category. Optional.
 	 * @param {string} eventValue The event category. Optional.
+	 * @return {Promise} Promise that always resolves.
 	 */
 	return function trackEvent( eventCategory, eventName, eventLabel = '', eventValue = '' ) {
```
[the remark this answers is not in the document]
felixarntz: Hmm, why that though? I'd think we only need that if were using `await` which we don't?

assets/js/util/tracking/createTrackEvent.js
```diff
@@ -32,17 +33,49 @@ export default function createTrackEvent( config, dataLayerTarget ) {
...
+			const resolveOnce = ( eventSent ) => {
+				if ( resolved ) {
+					return;
+				}
+				resolved = true;
+				if ( ! eventSent ) {
+					global.console.warn( `Tracking event "${ eventName }" (category "${ eventCategory }") took too long to fire.` );
+				}
+				resolve();
+			};
```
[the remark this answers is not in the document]
felixarntz: This is actually not to call `resolve` once as much as it is to only call the surrounding logic once - i.e. the warning shouldn't be triggered on the second call to `resolveOnce` (not part of the actual `resolve` function).

assets/js/util/tracking/createTrackEvent.js
```diff
@@ -32,17 +33,49 @@ export default function createTrackEvent( config, dataLayerTarget ) {
...
+			// tracking should not result in user-facing errors. It will just
+			// trigger a console warning.
+			const failTimeout = setTimeout( () => {
+				resolveOnce( false );
+			}, 1000 );
+
+			dataLayerPush( 'event', eventName, {
+				...eventData,
+				event_callback: () => {
+					clearTimeout( failTimeout );
```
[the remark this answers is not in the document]
felixarntz: See above for the reasoning, the warning isn't part of `resolve`.

assets/js/util/tracking/index.test.js
felixarntz: See similar comments in #1388, using `mockRestore` here is causing issues elsewhere.

assets/js/util/tracking/index.test.js
felixarntz: The promise actually doesn't resolve with a specific value - potential errors are handled internally by triggering the warning, since these shouldn't surface to the user in any way.

assets/js/util/tracking/index.test.js
felixarntz: See above, it doesn't resolve with a specific value.

assets/js/util/tracking/createTrackEvent.js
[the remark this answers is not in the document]
aaemnnosttv: In that case why not make a `warnOnce` function 😄 ?

You could simplify the implementation a bit using [`_.once`](https://lodash.com/docs/4.17.15#once)

```js
const warnOnce = once( global.console.warn )
```
Then `warnOnce` would only invoke the given function once per call to `trackEvent` since a new one would be created on each invocation.

Issue #1653, opened 2020-06-09
Add support for `dashboard-details` screen to entity detection


felixarntz:
## Feature Description

We recently introduced selectors on `core/site` to get information about the "entity" associated with the current context. An entity is basically a URL annotated with relevant WordPress information, e.g. whether it is a post, a term, the blog archive, a date archive etc. At the moment, only entities of type `post` (a single post) and `home` (the blog archive) are supported (for context, #174 is about expanding that at some point).

One area which is not yet supported by the entity detection mechanism on the PHP side is the details view for a URL within `googlesitekit-dashboard`. This is what this issue aims to solve: If you're on a details view, the entity detection mechanism should support detecting `post` and `home` entities from that as well.

A secondary change that should be made here is that `home` should be renamed to `blog`, as that is a more appropriate representation. While WordPress calls this "home" (e.g. `is_home()` function), it may not be the actual home page.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* When visiting `googlesitekit-dashboard` for a single URL (`permaLink` query parameter), `Context::get_reference_entity()` should support detecting that entity if it is a single post or the blog archive.
* Entities of type `home` should have their type changed to `blog`. Since this is not used anywhere yet, it's fine to just change.
* If the entity of type `blog` is the home page (i.e. there is no post associated with it), assume a title of `__( 'Home', 'google-site-kit' )`, similar to how WordPress labels it in e.g. its Menus UI.

## Implementation Brief

* Introduce method `Context::get_reference_entity_from_url( $url )`:
    * Replace the reference URL with the "local" `home_url()` to ensure local WP logic works with the URL (this is only relevant when the reference URL differs).
    * Use `url_to_postid` function (or WP.com equivalent where applicable).
        * If this doesn't return a result, check specifically whether the URL is for `get_option( 'page_for_posts' )`, since this particular case is not covered by `url_to_postid`.
    * If a post ID was found, return that post's entity object.
    * Otherwise, if the URL path is just empty or `/`, return the entity for type `blog` (see naming change, previously would have been `home`), via a new private `Context::create_entity_for_home_blog()`.
* In `Context::get_reference_entity()` and its `is_admin()` clause, check if we're on `page=googlesitekit-dashboard` with a `permaLink` query parameter. If so, call the new `Context::get_reference_entity_from_url( $url )` with the `permaLink` value.
[...]

Issue #1636, opened 2020-06-04
Expose web vitals metrics in PSI widget


felixarntz:
## Feature Description

With the launch of [Web Vitals](https://web.dev/vitals/), the PageSpeed Insights API now exposes the relevant field data in API responses for all sites where it is available. [Sample API request](https://www.googleapis.com/pagespeedonline/v5/runPagespeed?url=https%3A%2F%2Fweb.dev%2F&strategy=mobile)

We should expand the capabilities of the PageSpeed Insights widget to expose these metrics when available.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

**This issue is only about putting the functionality in place, the exact UI will be handled in a follow-up issue. The end result of this issue should essentially be the widget displaying everything that's needed in the most basic and "ugliest" way possible.** 😄 

* Web vitals metrics should be exposed in the PageSpeed Insights widget in both the dashboard as well as single page stats, **if available in the API response for both "mobile" and "desktop"**.
* There should be two tabs in the upper left for toggling "Lab Data" and "Field Data" (exact wording TBD).
* There should be a button group on the upper right for toggling "Mobile" and "Desktop".
* If there is no field data available in the API response (currently that is the majority), the "Lab Data" tab should be the default tab. Otherwise, the "Field Data" tab should be the default tab.
* "Lab Data" tab should for now display:
    * Total blocking time
    * Largest contentful paint
    * Cumulative layout shift
* "Field Data" tab should for now display:
    * First input delay
    * Largest contentful paint
    * Cumulative layout shift
* Every one of these values should be marked with either "Good", "Needs improvement" or "Poor", based on the value.
* The widget should have the common footer that says "Source: PageSpeed Insights", with a link.

## Implementation Brief

* Modify `Google\Site_Kit\Modules\PageSpeed_Insights::parse_data_response()` to return the full `$response` from the PSI API instead of just the `lighthouseResult` property. Document this change with a new `@since` annotation since technically this is a breaking change. Since likely nobody uses our API endpoint directly, we can probably get away with it, but should at least document it.
* Implement a new `DashboardPageSpeed` functional component in `assets/js/modules/pagespeed-insights/components/DashboardPageSpeed.js`, with tabs and UI per the ACs. The component should fully rely on the datastore, except for the `permaLink` reference below that should still come for the global (migrating that should be a separate effort).
[...]

PR #1657, opened 2020-06-09
Add support for entity detection to dashboard-details view.

felixarntz:
## Summary

Addresses issue #1653 

## Relevant technical choices

* Will need to be updated after #1636 is merged to update that usage of the legacy `global.googlesitekit.permaLink`.

## Checklist


includes/Context.php
```diff
@@ -190,6 +190,12 @@ public function get_reference_site_url() {
 	public function get_reference_entity() {
 		// If currently in WP admin, run admin-specific checks.
 		if ( is_admin() ) {
+			// Support specific URL stats being checked in Site Kit dashboard details view.
+			$entity_url_query_param = 'googlesitekit-dashboard' === $this->input()->filter( INPUT_GET, 'page' ) ? $this->input()->filter( INPUT_GET, 'permaLink' ) : '';
```
aaemnnosttv, line 194: I'd suggest breaking this onto separate lines for readability. We can also make it a little more secure by requiring `permaLink` to be a valid URL.
```suggestion
			$entity_url_query_param = 'googlesitekit-dashboard' === $this->input()->filter( INPUT_GET, 'page' )
				? $this->input()->filter( INPUT_GET, 'permaLink', FILTER_VALIDATE_URL )
				: '';
```

assets/js/googlesitekit/datastore/site/info.js
```diff
@@ -311,6 +315,34 @@ export const selectors = {
...
+	 * the same value as the `getReferenceSiteURL` selector.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {(string|undefined)} The current reference URL, or undefined if
+	 *                              not loaded yet.
+	 */
+	getCurrentReferenceURL: createRegistrySelector( ( select ) => () => {
+		const { currentEntityURL, referenceSiteURL } = select( STORE_NAME ).getSiteInfo() || {};
```
aaemnnosttv, line 335: I feel like maybe we should use the dedicated selectors for these, rather than reaching for them directly from the main site info selector. This is what the doc says above but that would be a bit more "guaranteed" if we used their respective selectors. It would still only invoke the underlying resolver once so I don't think there's a performance penalty to do so. What do you think?

felixarntz: Good point, that suggestion is also cleaner in not replicating the internal details of where to access the respective value from.

includes/Context.php
felixarntz: I opted for splitting into a separate if clause, I generally prefer these for readability over ternary unless the condition is really simple and line length remains short.

Issue #1650, opened 2020-06-09
AMP plugin comaptibility: Notice: is_amp_endpoint was called incorrectly


ernee:
## Bug Description

A [user in the support forums](https://wordpress.org/support/topic/issue-with-amp-plugin/) reported getting the following error when using the [AMP plugin](https://wordpress.org/plugins/amp/) and installing Site Kit on the wp-login page:

`Notice: is_amp_endpoint was called incorrectly. is_amp_endpoint() was called before the parse_query hook was called. Please see Debugging in WordPress for more information. (This message was added in version 0.4.2.) in /home/paiddbname/public_html/wp-includes/functions.php on line 5167`

After deleting Site Kit the error goes away. 

[Site Health Info
](https://wordpress.org/support/topic/issue-with-amp-plugin/#post-12913396)
## Additional Context

 - PHP Version: 
 - Plugin Version [e.g. 22]

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All the general logic in the `Assets` class should only be run if potentially relevant for Site Kit usage (i.e. if `current_user_can( Permissions::AUTHENTICATE )`) - otherwise it does nothing for Site Kit and could just introduce unintended side effects or performance or security leaks.
* Only the registration of Site Kit assets should continue happen everywhere.

## Implementation Brief

* In `Plugin`, move `Assets` instantiation into the `init` action so that the current user is set up already at this point. Also move the `Activation` instantiation in there since it relies on the `Assets` instance - that's fine since the hooks that it adds to all happen later anyway.
* In `Assets::register`, bail right after adding the asset registration action if `! current_user_can( Permissions::AUTHENTICATE )`.
* Move the call to `add_amp_dev_mode_attributes` out of `register_assets` and instead move it into the `Assets::register` method as well so that it only happens when relevant for Site Kit.
* Merge #1659.

## QA Brief

* 

## Changelog entry

* Do not run Site Kit assets logic when not applicable for the current user, avoiding unnecessary checks e.g. on the login screen.

westonruter:
Note how AMP plugin v1.5.3 implements `is_amp_endpoint()`: https://github.com/ampproject/amp-wp/blob/1.5.3/includes/amp-helper-functions.php#L569-L571

So it should short-circuit for `wp-login.php`. 

Take note that the logic has been revised in 1.6-alpha: https://github.com/ampproject/amp-wp/pull/4574

[...]

PR #1659, opened 2020-06-09
Do not run general asset logic if unnecessary for Site Kit

felixarntz:
## Summary

Addresses issue #1650

## Checklist


includes/Plugin.php
```diff
@@ -131,24 +131,26 @@ function () {
...
+				$options      = new Core\Storage\Options( $this->context );
+				$transients   = new Core\Storage\Transients( $this->context );
 				$user_options = new Core\Storage\User_Options( $this->context, get_current_user_id() );
 
 				$authentication = new Core\Authentication\Authentication( $this->context, $options, $user_options, $transients );
 				$authentication->register();
 
+				$permissions = new Core\Permissions\Permissions( $this->context, $authentication );
+				$permissions->register();
+
```
aaemnnosttv, line 147: Nitpick: the added variable here isn't necessary as nothing else depends on it.

includes/Plugin.php
```diff
@@ -158,6 +160,7 @@ function() use ( $options, $transients, $assets ) {
 				( new Core\Notifications\Notifications( $this->context, $options, $authentication ) )->register();
 				( new Core\Util\Debug_Data( $this->context, $options, $user_options, $authentication, $modules ) )->register();
 				( new Core\Admin\Standalone( $this->context ) )->register();
+				( new Core\Util\Activation( $this->context, $options, $assets ) )->register();
```
aaemnnosttv, line 163: This is too late for `Activation` to hook on to `googlesitekit_activation`, which is causing the activation notification not to show up.

felixarntz: I split the class into two so that the one class can take care of activation logic while the other one handles the notice. This also makes sense because only the latter requires the `Assets` instance, hence should be instantiated later.

tests/phpunit/integration/Core/Util/Activation_FlagTest.php
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	protected function network_activate_site_kit() {
+		add_filter(
+			'pre_site_option_active_sitewide_plugins',
+			function () {
+				return array( GOOGLESITEKIT_PLUGIN_BASENAME => true );
+			}
+		);
+	}
```
aaemnnosttv, line 93: This can be removed now that the same method exists on `TestCase`.

includes/Core/Util/Activation_Notice.php
```diff
@@ -0,0 +1,163 @@
...
+use Google\Site_Kit\Context;
+use Google\Site_Kit\Core\Admin\Notice;
+use Google\Site_Kit\Core\Assets\Assets;
+use Google\Site_Kit\Core\Util\Requires_Javascript_Trait;
+
+/**
+ * Class handling plugin activation.
+ *
+ * @since 1.0.0
+ * @since n.e.x.t Renamed from Activation.
```
aaemnnosttv, line 22: I feel like the `@since 1.0.0` doesn't really belong in this file, since it's a new class. The note about the class it came from in the "main" since annotation should be enough that "before this version it was handled in this other class". I'm not strongly against it, but it just seems odd that we would include a `@since` version on a class that didn't exist at the time.

includes/Core/Util/Activation_Notice.php
```diff
@@ -0,0 +1,163 @@
...
+		Activation_Flag $activation_flag,
+		Assets $assets = null
+	) {
+		$this->context         = $context;
+		$this->activation_flag = $activation_flag;
+
+		if ( ! $assets ) {
+			$assets = new Assets( $this->context );
+		}
+		$this->assets = $assets;
```
aaemnnosttv, line 73: This could be cleaned up for consistency with other constructors.

includes/Core/Util/Activation_Flag.php
```diff
@@ -1,106 +1,71 @@
...
-		Options $options = null,
-		Assets $assets = null
+		Options $options = null
 	) {
 		$this->context = $context;
 
 		if ( ! $options ) {
 			$options = new Options( $this->context );
 		}
 		$this->options = $options;
```
aaemnnosttv, line 60: This could be cleaned up for consistency with other constructors.

includes/Core/Util/Activation_Notice.php
aaemnnosttv: The other `@since 1.0.0` annotations in this file should be updated as well, no?

felixarntz: Oh yeah, right. Done!

PR #1662, opened 2020-06-10
Add new PSI dashboard widget with web vitals

aaemnnosttv:
## Summary

Addresses issue #1636

## Checklist


assets/js/modules/pagespeed-insights/datastore/constants.js
```diff
@@ -17,3 +17,12 @@
  */
 
 export const STORE_NAME = 'modules/pagespeed-insights';
+
+// Form ID for PageSpeed widget.
+export const FORM_DASH_WIDGET = 'pagespeedWidget';
```
felixarntz, line 22: May be worth making this more specific, to avoid potential conflicts, maybe `pagespeedDashboardWidget`? Or `webVitalsDashboardWidget`?

assets/js/modules/pagespeed-insights/components/DashboardPageSpeed.js
```diff
@@ -0,0 +1,137 @@
...
+
+	if ( ! reportMobile || ! reportDesktop || ! dataSrc ) {
+		return <ProgressBar />;
+	}
+
+	const reportData = strategy === STRATEGY_MOBILE ? reportMobile : reportDesktop;
+	const footerLinkHTML = sprintf(
+		/* translators: 1: link attributes, 2: translated service name */
+		__( 'Learn more details at <a %1$s>%2$s</a>', 'google-site-kit' ),
+		`href="${ addQueryArgs( 'https://developers.google.com/speed/pagespeed/insights/', { url } ) }" class="googlesitekit-cta-link googlesitekit-cta-link--external" target="_blank"`,
```
felixarntz, line 83: Not sure if that's possible in combination with `sprintf` here, but could we somehow use the `Link` component instead and have that generate the HTML? Not a big deal, but the attributes here would pretty much be handled by that.

assets/js/modules/pagespeed-insights/components/DashboardPageSpeed.test.js
```diff
@@ -0,0 +1,111 @@
...
+			new Promise( () => {} ), // Don't return a response.
+		);
+		const { queryByRole } = render( <DashboardPageSpeed />, { setupRegistry: setupNoReports } );
+
+		expect( queryByRole( 'progressbar' ) ).toBeInTheDocument();
+	} );
+
+	it( 'displays field data by default when available in both mobile and desktop reports', () => {
+		expect( fixtures.pagespeedMobile.loadingExperience ).toHaveProperty( 'metrics' );
+		expect( fixtures.pagespeedDesktop.loadingExperience ).toHaveProperty( 'metrics' );
```
felixarntz, line 49: May be unnecessary to assert something on our own fixtures here?

assets/js/modules/pagespeed-insights/components/DashboardPageSpeed.test.js
```diff
@@ -0,0 +1,111 @@
...
+	} );
+
+	it( 'has tabs for toggling the displayed data source', () => {
+		const { getByText } = render( <DashboardPageSpeed />, { setupRegistry } );
+
+		const labDataTabLink = getByText( /In the Lab/i );
+		expect( labDataTabLink ).not.toHaveClass( activeClass );
+		fireEvent.click( labDataTabLink );
+
+		expect( labDataTabLink ).toHaveClass( activeClass );
```
felixarntz, line 76: We could test here too that the "Field Data" tab is now no longer active I think.

assets/js/modules/pagespeed-insights/components/DashboardPageSpeed.test.js
```diff
@@ -0,0 +1,111 @@
...
+
+	it( 'has tabs for toggling the tested device', () => {
+		const { getByText } = render( <DashboardPageSpeed />, { setupRegistry } );
+
+		const desktopToggle = getByText( /desktop/i );
+		expect( desktopToggle ).not.toHaveClass( activeClass );
+
+		fireEvent.click( desktopToggle );
+
+		expect( desktopToggle ).toHaveClass( activeClass );
```
felixarntz, line 87: See above, we could test that "Mobile" is no longer active.

assets/js/modules/pagespeed-insights/components/FieldReportMetrics.js
```diff
@@ -0,0 +1,108 @@
...
+import { sanitizeHTML } from '../../../util';
+
+export default function FieldReportMetrics( { data } ) {
+	const learnMoreLink = (
+		<Link
+			href="https://web.dev/user-centric-performance-metrics/#in-the-field"
+			external
+			inherit
+			dangerouslySetInnerHTML={ sanitizeHTML(
+				__( 'Learn more<span class="screen-reader-text"> about field data.</span>', 'google-site-kit' ),
```
felixarntz, line 43: 👍

includes/Modules/PageSpeed_Insights.php
```diff
@@ -129,17 +129,15 @@ protected function create_data_request( Data_Request $data ) {
...
 	 *
 	 * @return mixed Parsed response data on success, or WP_Error on failure.
 	 */
 	protected function parse_data_response( Data_Request $data, $response ) {
-		switch ( "{$data->method}:{$data->datapoint}" ) {
-			case 'GET:pagespeed':
-				// TODO: Parse this response to a regular array.
-				return $response->getLighthouseResult();
+		switch ( "{$data->method}:{$data->datapoint}" ) { // phpcs:ignore Generic.CodeAnalysis.EmptyStatement.DetectedSwitch
 		}
```
felixarntz, line 141: We could just remove this I guess?

assets/js/modules/pagespeed-insights/components/DashboardPageSpeed.js
aaemnnosttv: I've done something similar with [`react-string-replace`](https://www.npmjs.com/package/react-string-replace) in the past. Not sure if we could do the same thing here since the output is translated but I agree, it's not the best, but a limitation of React.

assets/js/modules/pagespeed-insights/components/DashboardPageSpeed.test.js
aaemnnosttv: It's not intended to test the fixture but to provide context for the test. It saves you from looking into the fixture for context and would be a more clear failure if the fixture were updated in the future with data that didn't have it.

includes/Modules/PageSpeed_Insights.php
aaemnnosttv: I figure we'll add something to it in the future and the case was the only thing really to remove. We can remove the whole switch if you'd like.

assets/js/modules/pagespeed-insights/components/DashboardPageSpeed.js
felixarntz: Hmm not too big of a deal, let's keep as is for now.

assets/js/modules/pagespeed-insights/components/DashboardPageSpeed.js
```diff
@@ -45,8 +45,8 @@ import { sanitizeHTML } from '../../../util';
 const { useSelect, useDispatch } = Data;
 
 export default function DashboardPageSpeed() {
-	// TODO: remove legacy global fallback.
-	const permalink = global._googlesitekitLegacyData?.permaLink || global.googlesitekit?.permaLink;
+	// TODO: replace global with selector.
+	const permalink = global._googlesitekitLegacyData.permaLink;
```
felixarntz, line 49: 👍

Issue #1356, opened 2020-04-07
Chore: upgrade all package dependencies.


adamsilverstein:
## Task Description

* Update dependencies in package.json. 
* Update dependencies in composer.json.
* Verify build and codebase works correctly.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All our Composer and NPM packages should be updated to their latest versions.
* Functionality should not be affected.
* If there are certain packages that we cannot update without breakage or making severe changes, they should be discussed on a case-by-case basis.

## Implementation Brief

* All packages within `composer.json` and `package.json` should be updated to their latest versions as long as doing so does not result in site breakages - in which case this needs to be discussed.
* `composer.json` - run `composer outdated` - gives a list of packages which have updates. Incrementally update versions. Checking for breakages.
* `package.json` - run `npm outdated` -  - gives a list of dependency packages which have updates. Incrementally update versions. Checking for breakages .

## QA Brief

* 

## Changelog entry

* Update the majority of 3P dependencies to their latest versions.

aaemnnosttv:
IB ✅

eugene-manuilov:
After working on npm dependencies I stuck with a weird issue that blocked everything. I can't troubleshoot it too much because the error message is very vague and it looks like the issue happens inside of @wordpress/data package.

I would be grateful if @tofumatt or anyone from the team can look at it too. The latest commint in my branch (`enhancement/1356-upgrade-dependencies`) that works for me is 5daf1d4251639ebd7c37e3b2c59756dc9c4ff160. The next two commits are develop branch merging and updating a few php dependencies. So my guess is that something in the develop branch started causing this issue in conjunctions with npm updates.

Here is the screenshot of the issue that I see in my browser:

![Screenshot from 2020-06-15 16-16-15](https://user-images.githubusercontent.com/873049/84684980-21eb6a00-af42-11ea-88f7-014f1b951a97.png)

eugene-manuilov:
@felixarntz @aaemnnosttv @tofumatt i have prepared a few separate PRs to upgrade dependencies in chunks. Could you please review it and merge if possible? I need it to be merged to unblock another updates.

* Gulp tasks: https://github.com/google/site-kit-wp/pull/1737
* Babel and webpack: https://github.com/google/site-kit-wp/pull/1738
* Material packages: https://github.com/google/site-kit-wp/pull/1739 (depends on babel/webpack PR to be merged)
[...]

PR #1665, opened 2020-06-11
Enhancement/1356 upgrade dependencies

eugene-manuilov:
## Summary

Addresses issue #1356

## Checklist


tofumatt:
Oddly enough, it's any call to `useSelect` that triggers this error. I haven't been able to pin down _why_ it's happening, and looking at the `npm ls react` and `npm ls react-dom` it doesn't look like we have conflicting versions that would present a problem.

But removing https://github.com/google/site-kit-wp/blob/fd44651134b2239815a7915246e3d5dc01a48c22/assets/js/components/permissions-modal/index.js#L38 and https://github.com/google/site-kit-wp/blob/fd44651134b2239815a7915246e3d5dc01a48c22/assets/js/components/permissions-modal/index.js#L39 clears the error; all the other hooks are fine. I wonder if something particular to the datastore is causing the issue but I can't yet tell what it is.

I don't have much more to report on that particular issue, sorry. I don't think this is related to previous issues I saw that were red herring type errors with hooks.

felixarntz:
Looks like indeed the problem is still the `useSelect` hooks, despite downgrading `@wordpress/data` which we initially thought fixed it.

When I removed the two `useSelect` hooks from `PermissionsModal`, the error just occurred somewhere else (`DashboardSpeed`, where we're as of lately using hooks too) - so it's not specific to the component or a specific store...

Now the big question is which other updates could surface that? 🤔

felixarntz:
Closing this in favor of the more granular PRs.

Issue #1622, opened 2020-06-01
Expand `googlesitekit-modules` JS API to allow registering modules


felixarntz:
## Feature Description

Modules are currently registered in PHP and exposed to JS via the `core/modules/data/list` API endpoint. However, we will also need to register module data in JS, e.g. components to use as settings panels. Data registered in JS should be combined with the data registered in PHP, with JS taking precedence.

While the focus of allowing pure-JS module registration (no PHP) is not important at this point, we should build it pretty much with that in mind.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a scaffolded JS component `DefaultModuleSettings`, within `assets/js/googlesitekit/modules/components`. For now it can just return `null`, implementation should be in a follow-up issue.
* A new action `registerModule( slug, settings )` should be added to the `core/modules` datastore:
    * The following `settings` arguments should be supported for now:
        * `settings.name`: Optional. Module name. Default is the `slug`.
        * `settings.description`: Optional. Module description. Default empty string.
        * `settings.icon`: Optional. Module icon. Default empty string.
        * `settings.order`: Optional. Numeric indicator for module order. Default 10.
        * `settings.homepage`: Optional. Module homepage URL. Default empty string.
        * `settings.internal`: Optional. Whether the module is considered internal. Default false.
        * `settings.settingsComponent`: Optional. React component to render the settings panel. Default is the `DefaultModuleSettings` component.
    * This data should be registered in the existing store state property `modules`, being merged into the data for the respective module there, if already exists. In other words, before actually running the code register the module, the action should fetch server-side module data via `core/modules/data/list` automatically, so that after that everything can work fully client-side.
    * When merging, the above default values for `settings` should only be applied if there is _neither_ a client-side value _nor_ a server-side value for it. Note that these defaults should also apply if no client-side registration call is ever made. This will mostly be irrelevant for our current usage, however, e.g. the `settingsComponent` will never be registered in PHP (not possible), so that data should be filled with the default even if it was purely loaded from PHP.
* The existing `getModules` selector should be modified to return modules sorted by `order` ascending. If not present, 10 should be assumed.
* Both `getModules` and `getModule` should be modified to also include `settingsComponent`, which will technically be stored outside of the datastore (see how the `googlesitekit-widgets` datastore handles components related to datastore).
[...]

PR #1671, opened 2020-06-12
Feature/1622 js api allow registering modules

ryanwelcher:
## Summary

Addresses issue #1622 

## Checklist


assets/js/googlesitekit/modules/components/DefaultModuleSettings.js
```diff
@@ -0,0 +1,23 @@
+/**
```
tofumatt, line 1: Does this really need to be its own module? It's only being imported in one place and it's a no-op. I'd say it's fine to just pass `() => null` to https://github.com/google/site-kit-wp/pull/1671/files#diff-84bfda982eea3cfa46b67d1a68f5ffc7R182 and remove this file entirely.

assets/js/googlesitekit/modules/datastore/fixtures.json
```diff
@@ -8,7 +8,8 @@
     "active": true,
     "connected": true,
     "dependencies": [],
-    "dependants": []
+	"dependants": [],
```
tofumatt, line 11: Hardly a big deal, but looks like the indentation is off in here. 😅

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -154,6 +166,58 @@ const baseActions = {
...
+
+	/**
+	 * @param {string}          slug                         Module slug.
+	 * @param {Object}          settings                     Module settings.
+	 * @param {string}          [settings.name]              Optional. Module name. Default is the slug.
+	 * @param {string}          [settings.description]       Optional. Module description. Default empty string.
+	 * @param {string}          [settings.icon]              Optional. Module icon. Default empty string.
+	 * @param {number}          [settings.order]             Optional. Numeric indicator for module order. Default 10.
+	 * @param {string}          [settings.homepage]          Optional. Module homepage URL. Default empty string.
+	 * @param {boolean}         [settings.internal]          Optional. Whether the module is considered internal. Default false.
```
tofumatt, line 178: Should this be a public/documented param? Would we ever want modules to use this setting? Should we at least say it's not for 3rd-parties?

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -206,10 +296,46 @@ const baseSelectors = {
...
+	 * @since n.e.x.t
+	 * @private
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {(number|undefined)} The registry for the modules store.
+	 */
+	getModuleRegistryKey( state ) {
+		const { registryKey } = state;
+
+		return registryKey;
```
tofumatt, line 338: Instead of duplicating the functionality from `getWidgetRegistryKey` (which is essentially what this does), why not remove this selector/related actions and remove `getWidgetRegistryKey`—then combine them into one selector likr `getRegistryKey`. The point of the `getWidgetRegistryKey` code was to allow each registry to key data in globals. There's no reason to create _multiple_ registry keys across different sections of the app/datastore, so let's combine them instead of duplicating the functionality.

It's probably reasonable to move that code to `core/site` or something quite "core".

Sorry for not highlighting that in the IB, I can see how it was a bit vague and implied copying the existing implementation. I might've even thought that at first, but seeing as they're doing the same thing, let's consolidate these functions 🙂

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -206,10 +296,46 @@ const baseSelectors = {
 	 * @param {Object} state Data store's state.
 	 * @return {(Object|undefined)} Modules available on the site.
 	 */
-	getModules( state ) {
+	getModules: createRegistrySelector( ( select ) => ( state ) => {
 		const { modules } = state;
-
+		const registryKey = select( STORE_NAME ).getModuleRegistryKey();
+		if ( undefined !== modules ) {
```
tofumatt, line 302: Nitpick, but this might be a bit easier-to-read/follow if the conditional were the "early return". As-is, we `return modules;` after this `if` block, but it's by-definition `undefined`. Changing this section to be:

```js
if ( undefined === modules ) {
  return undefined;
}

// Continue with mappedModules code below...
```

assets/js/util/convert-array-to-keyed-object-map.js
```diff
@@ -0,0 +1,31 @@
...
+ */
+/**
+ * Converts an array of objects into an object map containing objects.
+ *
+ * @param {Array} arrayData       Array data to be converted to an Object
+ * @param {string} propertyKey  Object propert to be used for the property name in the main object.
+ *
+ * @return {Object} An object containing objects
+ */
+export function convertArrayListToKeyedObjectMap( arrayData, propertyKey ) {
```
tofumatt, line 26: This is nice, but doesn't lodash already have a utility like this? I'm usually a fan of using straightforward ES6 functions like this, but if we _already_ include lodash (and it doesn't look like we'll be removing it anytime soon), we can use that instead.

i think https://lodash.com/docs/4.17.15#keyBy does it. If we can use that, let's do that instead. But this looks fine if there's a reason we need to use it/can't use lodash. 🙂

assets/js/util/index.js
```diff
@@ -53,6 +53,8 @@ export * from './standalone';
 export * from './storage';
 export * from './i18n';
 export * from './helpers';
+export * from './sort-object-map-by-key';
```
tofumatt, line 56: Do we need to export these new functions? I thought the plan going forward was to keep util imports separated to make circular imports less likely and possibly do better chunk splitting. Since only new code references these files, we can scope the import to the specific files.

assets/js/util/sort-object-map-by-key.js
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Sorts an object map based on numberic order/priority key.
+ *
+ * @param {Object} objectList  Object containg objects to be sorted
+ * @param {string} priorityKey The object key to use for matching.
+ *
```
tofumatt, line 24: Should this empty line be here? 🤷

assets/js/util/sort-object-map-by-key.js
```diff
@@ -0,0 +1,37 @@
...
+ * @return {Array} An Array containing objects ordered by the priority key
+ */
+export function sortObjectMapByKey( objectList, priorityKey ) {
```
tofumatt, line 27: This is great, can you replace https://github.com/google/site-kit-wp/blob/ff9c5141800adbad9a1cf26ca09b291a57bd361a/assets/js/googlesitekit/widgets/datastore/widgets.js#L224 and https://github.com/google/site-kit-wp/blob/ff9c5141800adbad9a1cf26ca09b291a57bd361a/assets/js/googlesitekit/widgets/datastore/areas.js#L204 with this new function? 😄

assets/js/googlesitekit/modules/components/DefaultModuleSettings.js
ryanwelcher: Part of the AC was to stub out this component and have it be finished in a follow up ticket.

>There should be a scaffolded JS component DefaultModuleSettings, within assets/js/googlesitekit/modules/components. For now it can just return null, implementation should be in a follow-up issue.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -154,6 +166,57 @@ const baseActions = {
 
 		return { response, error };
 	},
+
+	/**
+	 * @param {string}          slug                         Module slug.
```
aaemnnosttv, line 171: This is missing the description and `@since` annotation.

assets/js/googlesitekit/datastore/site/registry-key.js
```diff
@@ -0,0 +1,96 @@
...
+			return {
+				...state,
+				registryKey,
+			};
+		}
+		default: {
+			return { ...state };
+		}
+	}
+};
```
tofumatt, line 73: Nitpick, but whitespace here would be nice 🙂 
```suggestion
};

```

assets/js/googlesitekit/datastore/site/registry-key.js
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Registry reducer.
+ *
+ * @since n.e.x.t
+ *
+ * @param {Object} state                      Data store's state.
+ * @param {Object} action                     Redux-style action.
+ * @param {(number|undefined)} action.payload The registryKey for a given store.
+ * @param {string}             action.type    The action type.
+ * @return {Object} Data store's state
```
tofumatt, line 58: I find these a bit more readable when aligned, but not a biggie!

```suggestion
 * @param {Object}             state          Data store's state.
 * @param {Object}             action         Redux-style action.
 * @param {(number|undefined)} action.payload The registryKey for a given store.
 * @param {string}             action.type    The action type.
 * @return {Object} Data store's state
```

assets/js/googlesitekit/datastore/site/registry-key.js
```diff
@@ -0,0 +1,96 @@
...
+const INITIAL_STATE = {
+	registryKey: undefined,
+};
+
+export const actions = {
+
+	/**
+	 * Sets the registryKey in the data store.
+	 *
+	 * @since n.e.x.t
```
tofumatt, line 36: Should have a newline after `@since`/`@private` sections.
```suggestion
	 * @since n.e.x.t

```

assets/js/googlesitekit/datastore/site/registry-key.js
```diff
@@ -0,0 +1,96 @@
...
+
+// Actions
+const SET_REGISTRY_KEY = 'SET_REGISTRY_KEY';
+
+const INITIAL_STATE = {
+	registryKey: undefined,
+};
+
+export const actions = {
+
```
tofumatt, line 32:
```suggestion
```

assets/js/googlesitekit/datastore/site/registry-key.test.js
```diff
@@ -0,0 +1,63 @@
...
+	describe( 'actions', () => {
+		describe( 'setRegistryKey', () => {
+			it( 'requires the registryKey param', () => {
+				expect( () => {
+					registry.dispatch( STORE_NAME ).setRegistryKey();
+				} ).toThrow( 'registryKey is required.' );
+			} );
+		} );
+	} );
+	describe( 'selectors', () => {
```
tofumatt, line 48: Breaking these up with a newline is usually what we do, I think it increases readability:

```suggestion

	describe( 'selectors', () => {
```

assets/js/googlesitekit/datastore/site/registry-key.test.js
```diff
@@ -0,0 +1,63 @@
...
+		it( 'returns undefined initially', async () => {
+			expect(
+				registry.select( STORE_NAME ).getRegistryKey()
+			).toBeUndefined();
+		} );
+		it( 'receives and sets registryKey', async () => {
```
tofumatt, line 54:
```suggestion

		it( 'receives and sets registryKey', async () => {
```

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -84,6 +96,7 @@ const BASE_INITIAL_STATE = {
 	// a module activation update, since the activation is technically complete
 	// before this data has been refreshed.
 	isAwaitingModulesRefresh: false,
+	registryKey: undefined,
```
tofumatt, line 99: `registryKey` should be stored and retrieved from `core/site` now, so this isn't needed.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -20,19 +20,33 @@
...
 import invariant from 'invariant';
+import { keyBy, sortBy } from 'lodash';
 
 /**
  * Internal dependencies
  */
 import API from 'googlesitekit-api';
 import Data from 'googlesitekit-data';
 import { STORE_NAME } from './constants';
+import { STORE_NAME as CORE_SITE_STORE_NAME } from '../../../googlesitekit/datastore/site';
```
tofumatt, line 31: We've usually imported these from the constants file (reduces possibility of circular imports), and in other files we've imported it as `CORE_SITE`:

```suggestion
import { STORE_NAME as CORE_SITE } from '../../../googlesitekit/datastore/site/constants';
```

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -45,9 +59,7 @@ const fetchGetModulesStore = createFetchStore( {
 		return {
 			...state,
 			isAwaitingModulesRefresh: false,
-			modules: modules.reduce( ( acc, module ) => {
-				return { ...acc, [ module.slug ]: module };
-			}, {} ),
+			modules: keyBy( sortBy( modules, [ ( { order } ) => order ] ), 'slug' ),
```
tofumatt, line 62: Is this doing the same thing? I find this new "shorthand" much more difficult to read. The original one is standard ES6 and a common pattern in ES6 reducer code. The new one requires knowledge of the lodash API and—being significantly "denser" by virtual of being a one-liner, implicit return function—takes more effort to parse.

I'd rather this return to the previous implementation unless there's a runtime benefit to using lodash.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -154,6 +167,58 @@ const baseActions = {
...
+	 * @param {React.Component} [settings.settingsComponent] React component to render the settings panel. Default is the DefaultModuleSettings component.
+	 * @return {Object} Redux-style action.
+	 */
+	*registerModule( slug, { settingsComponent = DefaultModuleSettings, ...settings } = {} ) {
+		invariant( slug, 'module slug is required' );
+
+		const registry = yield commonActions.getRegistry();
+		let registryKey = yield registry.select( CORE_SITE_STORE_NAME ).getRegistryKey();
+		if ( registryKey === undefined ) {
+			registryKey = Object.keys( ModuleComponents ).length + 1;
```
tofumatt, line 192: Shoot, I just realised an issue with this approach: we're declaring a new key based on the number of `ModuleComponents` registered—in this case `Object.keys( ModuleComponents ).length + 1;` might equal `1`, so we'll set this registry's key as `1`, but what if multiple registries have been registered with widgets first and there are is already a `1` and `2` key in the registry?

Either we need to use UUIDs here (not a bad idea really) or we need to also check that the registry key we create from this isn't already in use. Otherwise we could run into very edge-case issues with registries sharing a key.

tofumatt: (I'd recommend us creating UUIDs for the registry keys rather than our naive integer approach.)

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -197,7 +288,7 @@ const baseSelectors = {
 	 *   "dependencies": [
 	 *     "analytics"
 	 *   ],
-	 *   "dependants": []
```
tofumatt, line 200: Whoops, that's probably my fault, I get my English-es confused sometimes 😅

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -206,11 +297,31 @@ const baseSelectors = {
...
 
-		return modules;
-	},
+		// Return `undefined` if modules haven't been loaded yet.
+		if ( modules === undefined ) {
+			return undefined;
+		}
+
+		const registryKey = select( CORE_SITE_STORE_NAME ).getRegistryKey();
+		const sortedModules = sortBy( modules, [ ( { order } ) => order ] );
```
tofumatt, line 309: Why use `lodash` over JS built-in `.sort()` methods? I find these methods offer little extra convenience but add to the cognitive load of reading the code. Lodash's API isn't universally known.

If there's a benefit to this then fine to keep, but if this could be accomplished with:

```suggestion
		const sortedModules = modules.sort((a, b) => {
			return a.object > b.object ? 1 : -1;
		});
```

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -206,11 +297,31 @@ const baseSelectors = {
...
+				if ( ModuleComponents[ registryKey ][ module.slug ] ) {
+					moduleWithComponent.settingsComponent = ModuleComponents[ registryKey ][ module.slug ];
+				} else {
+					moduleWithComponent.settingsComponent = DefaultModuleSettings;
+				}
+			}
+
+			return moduleWithComponent;
+		} );
+		return keyBy( mappedModules, 'slug' );
```
tofumatt, line 323: Using this on its own is actually a bit more readable, for what it's worth. Maybe I just don't like lodash. 😅

assets/js/googlesitekit/modules/datastore/modules.test.js
```diff
@@ -36,9 +41,7 @@ import { STORE_NAME } from './constants';
 import FIXTURES from './fixtures.json';
 
 describe( 'core/modules modules', () => {
-	const fixturesKeyValue = FIXTURES.reduce( ( acc, module ) => {
-		return { ...acc, [ module.slug ]: module };
-	}, {} );
+	const fixturesKeyValue = keyBy( sortBy( FIXTURES, [ ( { order } ) => order ] ), 'slug' );
```
tofumatt, line 44: I think combining the `keyBy` and `sortBy` functions on one line increases the complexity of this single line too much. Maybe separating them out will help with readability:
```suggestion
	const sortedFixtures = sortBy( FIXTURES, [ ( { order } ) => order ] );
	const fixturesKeyValue = keyBy( sortedFixtures, 'slug' );
```

assets/js/googlesitekit/widgets/datastore/widgets.js
```diff
@@ -20,12 +20,14 @@
  * External dependencies
  */
 import invariant from 'invariant';
+import sortBy from 'lodash';
```
tofumatt, line 23: Shouldn't this be 
```suggestion
import { sortBy } from 'lodash';
```
?

tofumatt: I'm actually a bit surprised this didn't cause tests to fail 😢

assets/js/googlesitekit/modules/datastore/modules.js
ryanwelcher: Updated.

assets/js/googlesitekit/modules/datastore/modules.js
ryanwelcher: I've reverted this back.

assets/js/googlesitekit/modules/datastore/modules.js
ryanwelcher: modules is an object so using sortBy makes the code a little more concise. I'll add a comment to the code to be more clear.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -154,14 +169,92 @@ const baseActions = {
...
+	 * @param {React.Component} [settings.settingsComponent] React component to render the settings panel. Default is the DefaultModuleSettings component.
+	 * @return {Object} Redux-style action.
+	 */
+	*registerModule( slug, { settingsComponent = DefaultModuleSettings, ...settings } = {} ) {
+		invariant( slug, 'module slug is required' );
+
+		const registry = yield commonActions.getRegistry();
+		let registryKey = yield registry.select( CORE_SITE ).getRegistryKey();
+		if ( registryKey === undefined ) {
+			registryKey = uuidv4();
```
tofumatt, line 194: So, as @eugene-manuilov mentioned: symbols would give us UUIDs. But then I thought about it more after the call and they won't properly serialise to POJOs… take, for instance:

```js
const a = {};
const b = Symbol('key');
const c = Symbol('key');
a[b] = 1;
a[c] = 2;
print(a);
> { [Symbol(key)]: 1, [Symbol(key)]: 2 }
JSON.stringify(a)
> '{}'
```

Symbols aren't as good a fit for redux as using strings, so let's stick with the UUID approach.

package.json
```diff
@@ -133,6 +133,8 @@
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
     "svgxuse": "^1.2.6",
+    "tinymce": "^5.1.6",
```
tofumatt, line 136: Why was TinyMCE added? 😱

tofumatt:
@googlebot I consent.

assets/js/googlesitekit/datastore/site/registry-key.js
```diff
@@ -0,0 +1,97 @@
...
+	registryKey: undefined,
+};
+
+export const actions = {
+	/**
+	 * Sets the registryKey in the data store.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {(number|undefined)} registryKey The registryKey for a given store.
```
tofumatt, line 37: This should just be a string now, since we're using a UUID now, right?

```suggestion
	 * @param {string} registryKey The registryKey for a given store.
```

assets/js/googlesitekit/datastore/site/registry-key.js
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Registry reducer.
+ *
+ * @since n.e.x.t
+ *
+ * @param {Object}             state          Data store's state.
+ * @param {Object}             action         Redux-style action.
+ * @param {(number|undefined)} action.payload The registryKey for a given store.
+ * @param {string}             action.type    The action type.
+ * @return {Object} Data store's state
```
tofumatt, line 58: These args don't really make sense, because a reducer takes any number of actions. I don't think we normally do this, so these can really just be removed. Only the `state` and `action` args are needed:

```suggestion
 * @param {Object}             state          Data store's state.
 * @param {Object}             action         Redux-style action.
 * @return {Object} Data store's state
```

assets/js/googlesitekit/datastore/site/registry-key.js
```diff
@@ -0,0 +1,97 @@
...
+
+export const selectors = {
+	/**
+	 * Returns the registry key being used for a given store.
+	 *
+	 * @since n.e.x.t
+	 * @private
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {(number|undefined)} The registryKey for a given store.
```
tofumatt, line 83:
```suggestion
	 * @return {(string|undefined)} The registryKey for a given store. Returns `undefined` if the key has not yet been set.
```

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -16,23 +16,43 @@
...
+ * WordPress dependencies
+ */
+import { WPElement } from '@wordpress/element';
+
 /**
  * External dependencies
  */
 import invariant from 'invariant';
+import { keyBy, sortBy } from 'lodash';
+import { v4 as uuidv4 } from 'uuid';
```
tofumatt, line 29: Thanks, that's a lot more readable 👍

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -16,23 +16,43 @@
  * limitations under the License.
  */
 
+/**
```
tofumatt, line 19: Usually we put WordPress dependencies underneath external ones, and above "internal" ones.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -154,14 +174,92 @@ const baseActions = {
...
+	 */
+	*registerModule( slug, { settingsComponent = DefaultModuleSettings, ...settings } = {} ) {
+		invariant( slug, 'module slug is required' );
+
+		const registry = yield commonActions.getRegistry();
+		let registryKey = yield registry.select( CORE_SITE ).getRegistryKey();
+		if ( registryKey === undefined ) {
+			registryKey = uuidv4();
+			yield registry.dispatch( CORE_SITE ).setRegistryKey( registryKey );
+		}
```
felixarntz, line 201: Shouldn't this block better be internal to the `core/site` store? I think it would simplify the consuming code if `getRegistryKey` handled the not set case on its own. If it's not set, create it itself so that the selector always returns one.

assets/js/googlesitekit/modules/datastore/modules.js
felixarntz: How much does the `uuid` package increase bundle size? If there is a significant impact, we should reconsider something else - in real world usage the symbol approach would still work reliably I think?

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -154,14 +174,92 @@ const baseActions = {
...
+			order: 10,
+			homepage: null,
+			internal: false,
+			...settings,
+		};
+
+		return {
+			payload: { slug, settings: mergedModuleSettings },
+			type: REGISTER_MODULE,
+		};
```
felixarntz, line 227: This (with the corresponding reducer below) does not fully match the ACs:
* When registering a module in JS its data should be merged with existing data for the module. Modules are registered in PHP, and data registered in JS should be added to this potentially existing data, not completely replace it.
* This also means defaults cannot be set in here because then the reducer is unable to decide whether the value is just a default or whether it was actually passed with the registration.
* In other words, I believe we'll need to remove the default handling (except for the component) out of this action and into the reducer instead. In the reducer, the registered module object should basically be set to `{ ...defaults, ...( alreadyRegisteredSettings || {} ), ...newlyRegisteredSettings }`.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -154,14 +174,92 @@ const baseActions = {
...
+				},
+			};
+		}
+		case SET_MODULE_COMPONENT_KEY: {
+			const { registryKey } = payload;
+			return {
+				...state,
+				registryKey,
+			};
+		}
```
felixarntz, line 256: This doesn't seem to be used anymore now with the centralized registry approach. Remove?

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -154,14 +174,92 @@ const baseActions = {
...
 		return { response, error };
 	},
+
+	/**
+	 * Registers a module
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {string}          slug                         Module slug.
+	 * @param {Object}          settings                     Module settings.
```
felixarntz, line 184: This is optional too, so should be marked as such.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -154,14 +174,92 @@ const baseActions = {
...
+	 * @param {string}          [settings.name]              Optional. Module name. Default is the slug.
+	 * @param {string}          [settings.description]       Optional. Module description. Default empty string.
+	 * @param {string}          [settings.icon]              Optional. Module icon. Default empty string.
+	 * @param {number}          [settings.order]             Optional. Numeric indicator for module order. Default 10.
+	 * @param {string}          [settings.homepage]          Optional. Module homepage URL. Default empty string.
+	 * @param {WPElement}       [settings.settingsComponent] React component to render the settings panel. Default is the DefaultModuleSettings component.
+	 * @return {Object}                                      Redux-style action.
```
felixarntz, line 191: Indentation for `@return` should be decoupled from `@param` so this doesn't need to be indented all the way like this.

assets/js/googlesitekit/modules/datastore/modules.js
felixarntz: Nope, if this is wrong, it's on me :)

It is set to this on the server, so let's change it back for now since that's what will actually be returned. Let's rename it in a separate issue as there could be further implications on other code.

assets/js/googlesitekit/modules/datastore/modules.js
felixarntz: We actually had some discussions way back about using native JS over `lodash` where reasonably possible, so I'd prefer if we stuck with that approach. The sorting the `core/widgets` store currently has should be easy enough to use instead of `sortBy`. And `keyBy` we should be able to easily replace too?

assets/js/googlesitekit/modules/datastore/modules.test.js
```diff
@@ -36,9 +41,8 @@ import { STORE_NAME } from './constants';
 import FIXTURES from './fixtures.json';
 
 describe( 'core/modules modules', () => {
-	const fixturesKeyValue = FIXTURES.reduce( ( acc, module ) => {
-		return { ...acc, [ module.slug ]: module };
-	}, {} );
+	const sortedFixtures = sortBy( FIXTURES, [ ( { order } ) => order ] );
+	const fixturesKeyValue = keyBy( sortedFixtures, 'slug' );
```
felixarntz, line 45: See above, let's try to use native JS.

assets/js/googlesitekit/widgets/datastore/areas.js
```diff
@@ -198,9 +199,12 @@ export const selectors = {
...
 
-		return Object.values( areas ).filter( ( area ) => {
-			return contextAssignments[ contextSlug ] && contextAssignments[ contextSlug ].includes( area.slug );
-		} ).sort( ( areaA, areaB ) => areaA.priority - areaB.priority );
+		return sortBy(
+			Object.values( areas ).filter( ( area ) => {
+				return contextAssignments[ contextSlug ] && contextAssignments[ contextSlug ].includes( area.slug );
+			} ),
+			[ ( area ) => area.priority ]
+		);
```
felixarntz, line 207: See above, let's keep the original native JS code.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -154,14 +174,92 @@ const baseActions = {
 
 		return { response, error };
 	},
+
+	/**
+	 * Registers a module
```
felixarntz, line 179:
```suggestion
	 * Registers a module.
```

assets/js/googlesitekit/widgets/datastore/widgets.js
```diff
@@ -220,28 +220,22 @@ export const selectors = {
...
+				if ( WidgetComponents[ registryKey ] ) {
+					widgetWithComponent.component = WidgetComponents[ registryKey ][ widget.slug ];
+				}
 
-			return widgetWithComponent;
-		} );
+				return widgetWithComponent;
+			} ),
+			[ ( widget ) => widget.priority ]
+		);
```
felixarntz, line 237: See above, let's keep using native JS.

assets/js/googlesitekit/widgets/datastore/widgets.js
```diff
@@ -94,14 +97,11 @@ export const actions = {
 		invariant( Object.values( WIDGET_WIDTHS ).includes( width ), `Widget width should be one of: ${ WidgetWidthKeys }, but "${ width }" was provided.` );
 
 		const registry = yield commonActions.getRegistry();
-		let registryKey = yield registry.select( STORE_NAME ).getWidgetRegistryKey();
+		let registryKey = yield registry.select( CORE_SITE ).getRegistryKey();
```
felixarntz, line 100: +1 for migrating this to the new `getRegistryKey` too!

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -154,14 +174,92 @@ const baseActions = {
...
+		const mergedModuleSettings = {
+			slug,
+			name: slug,
+			description: null,
+			icon: null,
+			order: 10,
+			homepage: null,
+			internal: false,
+			...settings,
+		};
```
felixarntz, line 222: Thinking further about this data and what is part of a module, we should make sure that nobody can register values for the `active` and `connected` fields. Let's make sure these two are deleted from `settings` if somebody tries to pass them.

In the reducer though, we should give these two fields default values (`false`). This way they're always set, while they can only be modified via the server, not via registering on the client.

assets/js/util/sort-object-map-by-key.js
felixarntz: There shouldn't be empty lines between param and return, only above the param/return group (there should be a `@since` annotation in here, which _should_ be separated by a blank line).

assets/js/util/sort-object-map-by-key.js
felixarntz, line 27: It's a bit weird to have this accept an object and return an array, especially since in some instances above we're already passing an array to it I believe? Does this not fail as is?

Either way, I think we should either make this accept an array and return an array (transforming back and forth from object to array would be handled outside the function) or accept an object and return an object (transforming is handled completely by the function). Since we already use `Object.values` on these objects for other reasons too (e.g. filtering), I think it makes sense to simplify this function to just sort arrays. It also means the function fulfils only a single purpose which I think is cleaner.

assets/js/googlesitekit/widgets/datastore/widgets.js
```diff
@@ -220,28 +210,22 @@ export const selectors = {
...
-			return 0;
-		} ).map( ( widget ) => {
-			const widgetWithComponent = { ...widget };
-			if ( WidgetComponents[ registryKey ] ) {
-				widgetWithComponent.component = WidgetComponents[ registryKey ][ widget.slug ];
-			}
+		const sorted = sortObjectMapByKey(
+			Object.values( widgets ).filter( ( widget ) => {
+				return areaAssignments[ widgetAreaSlug ] && areaAssignments[ widgetAreaSlug ].includes( widget.slug );
+			} ).map( ( widget ) => {
```
felixarntz, line 218: See my above comment, here we're passing an array to `sortObjectMapByKey` which is not correct as it works now. However, I think we should simplify that function to deal with arrays, at which point this would become okay.

assets/js/googlesitekit/widgets/datastore/areas.js
```diff
@@ -198,9 +199,12 @@ export const selectors = {
 
 		const { areas, contextAssignments } = state;
 
-		return Object.values( areas ).filter( ( area ) => {
-			return contextAssignments[ contextSlug ] && contextAssignments[ contextSlug ].includes( area.slug );
-		} ).sort( ( areaA, areaB ) => areaA.priority - areaB.priority );
+		return sortObjectMapByKey(
+			Object.values( areas ).filter( ( area ) => {
```
felixarntz, line 203: See above, we're passing an array to the function which right now expects an object.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -206,11 +303,32 @@ const baseSelectors = {
...
-		return modules;
-	},
+		// Return `undefined` if modules haven't been loaded yet.
+		if ( modules === undefined ) {
+			return undefined;
+		}
+
+		const registryKey = select( CORE_SITE ).getRegistryKey();
+		// Sorting the modules object by order property.
+		const sortedModules = sortObjectMapByKey( modules, 'order' );
```
felixarntz, line 316: This is correct now, but when we change the function to work purely on arrays, we need to use `Object.values` here. Which makes sense anyway because below we need it in array format anyway.

assets/js/util/sort-by-property.js
```diff
@@ -17,19 +17,20 @@
...
+ * @since n.e.x.t
  *
+ * @param {Array} arrayData  Array containing objects to be sorted
+ * @param {string} property The object key to use for matching.
  * @return {Array} An Array containing objects ordered by the priority key
  */
-export function sortObjectMapByKey( objectList, priorityKey ) {
-	return Object.values( objectList ).sort( ( objectA, objectB ) => {
-		if ( objectA[ priorityKey ] > objectB[ priorityKey ] ) {
+export function sortByProperty( arrayData, property ) {
```
aaemnnosttv, line 28: Is there a reason why we created a utility for this instead of using [`_.sortBy`](https://lodash.com/docs/4.17.15#sortBy)? It can be used the same way so I'm just curious if there was a reason or other decision to use our own implementation.

felixarntz: We've been trying to use native JS implementations for simple things like this where it offers easy to use solutions so that we don't need to rely as much on `lodash`. But yeah, if we need to start splitting logic out into our own utility functions, I agree it becomes a bit less of a clear cut.

aaemnnosttv: Yeah that makes sense. There are several places in the older parts of the codebase where lodash `each`, `filter`,  or `map` are used that should probably use the native array methods as I'm not aware of any particular benefit there, other than maybe lodash offers better support with non-Arrays. I don't think we should be creating utilities though for something that is readily available (and likely much more thoroughly tested) via a dependency we already have and use.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -149,6 +178,51 @@ const baseActions = {
...
+	 * @param {number}    [settings.order]             Optional. Numeric indicator for module order. Default 10.
+	 * @param {string}    [settings.homepage]          Optional. Module homepage URL. Default empty string.
+	 * @param {WPElement} [settings.settingsComponent] React component to render the settings panel. Default is the DefaultModuleSettings component.
+	 * @return {Object} Redux-style action.
+	 */
+	*registerModule( slug, { settingsComponent = DefaultModuleSettings, ...settings } = {} ) {
+		invariant( slug, 'module slug is required' );
+
+		const registry = yield commonActions.getRegistry();
+		const registryKey = yield registry.select( CORE_SITE ).getRegistryKey();
```
felixarntz, line 201: @tofumatt Double-checking, is this okay to do? Can we `yield` a selector like this? I would think that in other cases where we need selector data to be available we've been using `waitFor*` controls - is that different here?

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -149,6 +178,51 @@ const baseActions = {
...
+	 * @param {string}    [settings.description]       Optional. Module description. Default empty string.
+	 * @param {string}    [settings.icon]              Optional. Module icon. Default empty string.
+	 * @param {number}    [settings.order]             Optional. Numeric indicator for module order. Default 10.
+	 * @param {string}    [settings.homepage]          Optional. Module homepage URL. Default empty string.
+	 * @param {WPElement} [settings.settingsComponent] React component to render the settings panel. Default is the DefaultModuleSettings component.
+	 * @return {Object} Redux-style action.
+	 */
+	*registerModule( slug, { settingsComponent = DefaultModuleSettings, ...settings } = {} ) {
+		invariant( slug, 'module slug is required' );
+
```
felixarntz, line 199: Something that just came to mind: We'll need to make sure server-side modules have been loaded (via `getModules` and its resolver) before running the logic below. Especially since these `registerModule` calls will likely happen early during app lifecycle.

See my other comment, we may need to explicitly wait for that? cc @tofumatt

assets/js/util/sort-by-property.js
felixarntz: @aaemnnosttv @ryanwelcher I think overall we should try to slowly eliminate `lodash` since I question its benefits with what latest native array and object methods offer.
More specifically to these two functions though, I somewhat question whether we even need utilities for something simple like this. For example `sortByProperty` was previously accomplished in one location with `arr.sort( (a, b) => a[ prop ] - b[ prop ] )`.

While doing that inline and not using a utility requires a tiny bit more code (although not even that in every case), I'd argue that code might still be easier to follow - especially since as soon as we add utility functions like this to the mix, the code starts mixing native array functions and those more specific functions. That may be a personal preference, but I find that code harder to read than _only_ native array functions or _only_ utility functions.

I don't have a strong preference on keeping these utility functions or removing them again and keeping that functionality inline. We can probably keep them since they'll cover a lot of what we'd use `lodash` for. I think my main guidance at this point is let's try to not use `lodash` unless there's a significant benefit.

assets/js/googlesitekit/modules/datastore/modules.js
tofumatt: This doesn't immediately strike me as something that should work, because `getRegistryKey()` is not a generator function and thus will immediately return. It will _seem_ like it works because `getRegistryKey()` returns a value, but if it were a selector with a resolver, this wouldn't wait until the resolver has completed—at least to my knowledge.

Good catch; I think this needs to be refactored to wait until `registryKey !== undefined`. (https://github.com/google/site-kit-wp/blob/develop/assets/js/modules/analytics/datastore/tags.js#L102 is an example of that sort of wait control, for reference.)

If this _does_ work and there's a test that covered it that'd be cool, but from my knowledge of `@wordpress/data` internals this shouldn't work, but either passes tests because it's very fast or we aren't actually testing it. It's possible that it's continuing on with `undefined` as the key and that's working.

assets/js/googlesitekit/modules/datastore/modules.js
felixarntz: Based on https://github.com/google/site-kit-wp/pull/1671#pullrequestreview-453736379, we'll probably need some internal `waitForModules` action with a control (similar like we have e.g. `waitForExistingTag`).

assets/js/googlesitekit/datastore/site/registry-key.js
```diff
@@ -0,0 +1,139 @@
...
+
+const INITIAL_STATE = {
+	registryKey: undefined,
+};
+
+export const actions = {
+	/**
+	 * Wait for the registryKey to be resolved.
+	 *
+	 * @since n.e.x.t
```
tofumatt, line 45: This could probably be safely marked as `@private`.

tofumatt: Actually, no, let's leave it public. It's a useful utility 😄

assets/js/googlesitekit/datastore/site/registry-key.test.js
```diff
@@ -0,0 +1,65 @@
...
+				} ).toThrow( 'registryKey is required.' );
+			} );
+		} );
+	} );
+
+	describe( 'selectors', () => {
+		it( 'generates a registryKey if undefined', async () => {
+			const firstCall = registry.select( STORE_NAME ).getRegistryKey();
+			expect( firstCall ).not.toBeUndefined();
+			expect( registry.select( STORE_NAME ).getRegistryKey() ).toEqual( firstCall );
```
tofumatt, line 53: Hmmm, this test should be failing if it _is_ generating a `registryKey`. I expect that `firstCall` here should be `undefined`, because we haven't yet set a registry key, and I'd expect we'd need to wait for `hasFinishedResolution('getRegistryKey')` then call this selector again.

```suggestion
			const firstCall = registry.select( STORE_NAME ).getRegistryKey();
			expect( firstCall ).toBeUndefined();

			await subscribeUntil( registry, () => registry
				.select( STORE_NAME )
				.hasFinishedResolution( 'getRegistryKey' )
			);

			expect( registry.select( STORE_NAME ).getRegistryKey() ).not.toBeUndefined();
```

I would expect this test to work.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -149,12 +193,111 @@ const baseActions = {
...
+		const registry = yield commonActions.getRegistry();
+		yield actions.waitForModules();
+		registry.select( STORE_NAME ).getModules();
```
tofumatt, line 217: What is the reason for this call to this selector? Its return value isn't being stored anywhere.

assets/js/googlesitekit/modules/datastore/modules.test.js
```diff
@@ -287,6 +288,31 @@ describe( 'core/modules modules', () => {
...
+			const moduleSlug = 'test-module';
+			const moduleSettings = {
+				name: 'Test Module',
+				order: 1,
+				description: 'A module for testing',
+				homepage: 'https://sitekit.withgoogle.com/',
+				icon: 'icon-name',
+			};
+
+			it( 'registers a module', async () => {
```
tofumatt, line 301: It would be good to add a test (or add to this test) confirming anything stored outside the redux store (in the `ModuleComponents` variable) is stored here as well.

assets/js/googlesitekit/datastore/site/registry-key.test.js
felixarntz: @tofumatt I've been wondering about this for a while. Maybe `@wordpress/data` has some logic built into it that somehow detects what the resolver does in some way? Maybe it'll fulfill immediately if there's no control involved (since it's actually not async then)? I mean this would actually be quite clever.

I think I had noticed this in a few other cases, and the current test passes (basically this selector never returns `undefined` for now). I think we should still write our logic defensively, supporting the case that this returns `undefined`, since we're not sure about the `@wordpress/data` internals yet, but maybe this is how they do it.

Issue #1649, opened 2020-06-08
Implement UI for new PSI widget


felixarntz:
## Feature Description

As a follow-up to #1636, this issue is about implementing the exact UI from the mock.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The new PSI widget implemented in #1636 should have UI as in the mock.
* Material components should be used where applicable.

## Implementation Brief

* Update the a newly created component from #1636 called `DashboardPageSpeed` at `assets/js/modules/pagespeed-insights/dashboard/dashboard-widget-web-vitals.js`. The component should: 
  * Display Web Vitals Metrics for both "In the Lab" and "In the Field", and "Mobile/Desktop" (See [Figma Mockup](https://www.figma.com/file/niZV5sVfeaE5sqnJAUgR3G/PageSpeed-Insights-for-10UP)).
  * Track the state of which tab is selected both "mobile/desktop" and "In the Lab/In the field" (states should be independent)
  * Pass the data retrieved from the PageSpeed Insights data store as props into `<LabReportMetrics>` and `<FieldReportMetrics>`  
  * Use a `<Layout>` component with no `header` prop to wrap:
    * `<header>` (see below)
    * a `<section>` which wraps `<LabReportMetrics>` or `<FieldReportMetrics>` (conditionally rendered based on the data source)

* Create a `<header>` inside `<Layout>` which should: 
    * Implement the "In the Lab" and "In the Field" tabs with `@material/react-tab-bar` to switch between the vitals for each scenario utilizing callbacks with `setDataSrcField` and `setDataSrcLab`
    * Use a re-usable mobile/desktop switcher component, called `DeviceSizeTabBar` (see below) which will trigger callbacks to call: `setStrategyMobile` and `setStrategyDesktop`

* The `LabReportMetrics`component should:
  * Accept `reportData` as a prop.
  * display the following items as `<div>` rows:
    * Row 1: "Lab data is useful for debugging performance issues, as it is collected in a controlled environment." and "Learn More" external arrow link.
    * Row 5: "View details at" and "PageSpeed Insights" external arrow link.
  * display the following metrics as rows in a `<table>`, styled to look the same as Rows 1 & 5 (this should have an invisible `<th>` which indiates the types of data):
    * Row 2: `Total Blocking Time` metric (for "In the Lab" only)
    * Row 3: `Largest Contentful Paint` metric (both)
[...]

PR #1679, opened 2020-06-16
Enhancement/1649 pagespeed insights widget ui

jqlee85:
## Summary

Addresses issue #1649 

## Relevant technical choices

 * Uses `@material/react-tab-bar` for "In the Lab" / "In the Field" selectors
 * Creates global `DeviceSizeTabBar` component, which also uses `@material/react-tab-bar`
 * Moves mobile/desktop switcher above "In the Lab" / "In the Field" selectors for very small screens

## Checklist


aaemnnosttv:
@googlebot I consent.

assets/js/modules/pagespeed-insights/components/FieldReportMetrics.js
```diff
@@ -49,6 +59,21 @@ export default function FieldReportMetrics( { data } ) {
...
+			inherit
+			dangerouslySetInnerHTML={ sanitizeHTML(
+				_x( 'PageSpeed Insights', 'Service name', 'google-site-kit' ),
+				{
+					ALLOWED_TAGS: [ 'a' ],
+					ALLOWED_ATTR: [ 'href', 'class', 'target' ],
+				}
+			) }
+		/>
+	);
```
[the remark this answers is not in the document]
jqlee85: @aaemnnosttv This was so it wouldn't be rendered in the "Field Data Unavailable" case. It could be kept in `DashboardPageSpeed` and conditionally rendered. But in terms of the HTML/CSS structure it made sense to me to have it in the `FieldReportMetrics` and `LabReportMetrics`.

assets/js/modules/pagespeed-insights/datastore/__fixtures__/index.js
```diff
@@ -17,5 +17,7 @@
  */
 
 export { default as pagespeedDesktop } from './pagespeed--desktop.json';
+export { default as pagespeedDesktopFieldDataUnavailable } from './pagespeed--desktop__field-data-unavailable.json';
```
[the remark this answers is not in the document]
jqlee85: Ah OK. Sounds good.

assets/js/components/DeviceSizeTabBar.js
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+/**
+ * External dependencies
+ */
+import Tab from '@material/react-tab';
+import TabBar from '@material/react-tab-bar';
+import PropTypes from 'prop-types';
+
+const DeviceSizeTabBar = ( { activeIndex, deviceSizes, handleDeviceSizeUpdate } ) => {
```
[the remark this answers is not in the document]
jqlee85: @aaemnnosttv Yeah, I guess it makes sense to apply it as a class, since this component could presumably be used elsewhere where it wouldn't warrant a shadow.

assets/js/components/DeviceSizeTabBar.js
```diff
@@ -0,0 +1,85 @@
...
+		} ),
+	),
+	handleDeviceSizeUpdate: PropTypes.func,
+};
+
+DeviceSizeTabBar.defaultProps = {
+	activeIndex: 0,
+	deviceSizes: [
+		{
+			index: 0,
```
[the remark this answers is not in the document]
jqlee85: :+1:

assets/js/modules/pagespeed-insights/components/ReportMetric.js
```diff
@@ -41,27 +41,29 @@ export default function ReportMetric( {
...
 				<div className="googlesitekit-pagespeed-report-metric__description">
 					{ description }
 				</div>
-			</div>
+			</td>
 
-			<div className={ `googlesitekit-pagespeed-report-metric--${ category }` }>
-				<div className="googlesitekit-pagespeed-report-metric__value">
-					{ displayValue }
+			<td className={ `googlesitekit-pagespeed-report-metric-value googlesitekit-pagespeed-report-metric--${ category.toLowerCase() }` }>
```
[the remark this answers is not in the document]
jqlee85: @aaemnnosttv I'll reformat this one and update this to use `classnames()`, that should make it cleaner.

assets/js/modules/pagespeed-insights/components/DashboardPageSpeed.js
```diff
@@ -77,61 +83,56 @@ export default function DashboardPageSpeed() {
...
+								<span className="mdc-tab__text-label">{ __( 'In the Lab', 'google-site-kit' ) }</span>
+							</Tab>
+							<Tab aria-label="in the field">
+								<span className="mdc-tab__text-label">{ __( 'In the Field', 'google-site-kit' ) }</span>
+							</Tab>
+						</TabBar>
+					</div>
+					<div className="googlesitekit-pagespeed-widget__device-size-tab-bar-wrapper">
+						<DeviceSizeTabBar
+							activeIndex={ strategy === STRATEGY_DESKTOP ? 1 : 0 }
```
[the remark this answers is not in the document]
jqlee85: @aaemnnosttv Makes sense!

assets/js/modules/pagespeed-insights/components/DashboardPageSpeed.test.js
```diff
@@ -27,7 +27,7 @@ import * as fixtures from '../datastore/__fixtures__';
 import fetchMock from 'fetch-mock';
 
 // TODO: update the active class.
-const activeClass = 'googlesitekit-cta-link--danger';
+const activeClass = 'mdc-tab--active';
```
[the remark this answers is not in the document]
jqlee85: :+1:

assets/js/modules/pagespeed-insights/components/DashboardPageSpeed.js
```diff
@@ -77,61 +83,56 @@ export default function DashboardPageSpeed() {
...
+			<Layout
+				className="googlesitekit-pagespeed-widget"
+			>
+				<header className="googlesitekit-pagespeed-widget__header">
+					<div className="googlesitekit-pagespeed-widget__data-src-tabs">
+						<TabBar
+							activeIndex={ dataSrc === DATA_SRC_FIELD ? 1 : 0 }
+							handleActiveIndexUpdate={ updateActiveTab }
+						>
+							<Tab aria-label="in the lab">
```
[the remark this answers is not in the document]
jqlee85: Makes sense, thanks!

assets/js/modules/pagespeed-insights/components/DashboardPageSpeed.js
```diff
@@ -77,61 +83,56 @@ export default function DashboardPageSpeed() {
...
-	const footerLinkHTML = sprintf(
-		/* translators: 1: link attributes, 2: translated service name */
-		__( 'View details at <a %1$s>%2$s</a>', 'google-site-kit' ),
-		`href="${ addQueryArgs( 'https://developers.google.com/speed/pagespeed/insights/', { url } ) }" class="googlesitekit-cta-link googlesitekit-cta-link--external" target="_blank"`,
-		_x( 'PageSpeed Insights', 'Service name', 'google-site-kit' )
-	);
+
+	// Update the active tab for "In the Lab" or "In The Field".
+	const updateActiveTab = ( dataSrcIndex ) => {
+		if ( dataSrcIndex === 0 ) {
```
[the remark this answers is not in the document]
jqlee85: @aaemnnosttv I think we'd have to wrap the `TabBar` component with another component in order to do this, or create another callback function which calls `updateActiveTab()`. Is it worth it to creating those to avoid using the index?

jqlee85:
@aaemnnosttv are you still seeing the blank tabs loading in Storybook? I haven't seen that happen, and haven't been able to reproduce.

aaemnnosttv:
> are you still seeing the blank tabs loading in Storybook? I haven't seen that happen, and haven't been able to reproduce.

@jqlee85 no, I fixed that - that's what the `WithNoPermalink` wrapper handles in Storybook although this will likely be removed shortly as the selector for what `_googlesitekitLegacyData.permaLink` provides will be merged soon anyways now. The problem is that since we check this global first, it was using this (which used a different URL from the storybook config data) so the reports we pre-received in the setup functions weren't used as it was requesting reports for a different URL. Since fetch is mocked to return an empty object for any unmatched request, this was what was being set for the report, and since the report object was no longer empty, it was not showing the loading state. Both `ReportMetrics` components return `null` if the provided report is missing necessary data which is why it was showing up with an empty tab content area.

jqlee85:
@aaemnnosttv Ah that makes sense. I've just pushed commits resolving conflicts after the addition of the new selector, and removing that `WithNoPermalink` wrapper.

.storybook/fetch-mocks.js
```diff
@@ -43,7 +43,7 @@ export function fetchMockSaveSettings() {
 
 // Catch-all mock to log any other requests.
 export function fetchMockCatchAll() {
-	fetchMock.mock( '*', ( url, opts ) => {
+	fetchMock.catch( ( url, opts ) => {
```
aaemnnosttv, line 46: This is the correct way to implement a catch-all fetch mock. The old mock which used a wildcard prevented other mocks from being added since it would match everything instead of anything which wasn't matched 😄

aaemnnosttv:
Thanks @jqlee85 - I've pushed a few more changes here and it should be ready for review once again now.

Summary of changes:
* Updated remaining usage of `_googlesitekitLegacyData.permaLink` to use new selectors (e.g. `DashboardSpeed` but also its presence in `*ReportMetrics` components)
* Extracted the footer link to a new `ReportDetailsLink` (also restored this to an earlier implementation to preserve the translated string as a whole as this this can't be concatenated))
* Extracted lab/field data learn more links to components in a new shared module  
These are exported from the same module because the underlying component they share is specific to those two links and so it isn't exported (maybe there's a generic `Link` variant in there but we can think about generalizing that later)
* Minor cleanup of a detail or two that was lost in the merge

assets/js/modules/pagespeed-insights/components/DataSourceLearnMoreLinks.js
```diff
@@ -0,0 +1,63 @@
...
+				{
+					ALLOWED_TAGS: [ 'span' ],
+					ALLOWED_ATTR: [ 'class' ],
+				}
+			) }
+		/>
+	);
+}
+
+export function FieldDataLearnMoreLink() {
```
tofumatt, line 47: I'm wary of exporting multiple components from the same file, as I know it can start to get a bit messy and harder-to-find components this way. But I get why it's happening here and it seems fine. Just something to keep an eye on I suppose, but this is fine 🙂

assets/js/modules/pagespeed-insights/components/FieldReportMetrics.js
```diff
@@ -75,30 +65,54 @@ export default function FieldReportMetrics( { data } ) {
...
-			/>
+		<div className="googlesitekit-pagespeed-insights-web-vitals-metrics">
+			<div className="googlesitekit-pagespeed-report__row googlesitekit-pagespeed-report__row--first">
+				<p>
+					{ __( 'Field data is useful for capturing true, real-world user experience - but has a more limited set of metrics.', 'google-site-kit' ) }
+					{ ' ' }
+					<FieldDataLearnMoreLink />
+				</p>
+			</div>
+			<table
```
tofumatt, line 76: I always get excited when it's time to actually use `<table>` tags legitimately, but they still feel like a red flag at first. I wonder if that visceral response will ever go away 😅

assets/js/modules/pagespeed-insights/components/ReportDetailsLink.js
```diff
@@ -0,0 +1,54 @@
...
+import { sanitizeHTML } from '../../../util';
+const { useSelect } = Data;
+
+export default function ReportDetailsLink() {
+	const referenceURL = useSelect( ( select ) => select( CORE_SITE ).getCurrentReferenceURL() );
+	const pagespeedInsightsURL = addQueryArgs( 'https://developers.google.com/speed/pagespeed/insights/', { url: referenceURL } );
+	const footerLinkHTML = sprintf(
+		/* translators: 1: link attributes, 2: translated service name */
+		__( 'View details at <a %1$s>%2$s</a>', 'google-site-kit' ),
+		`href="${ pagespeedInsightsURL }" class="googlesitekit-cta-link googlesitekit-cta-link--external" target="_blank"`,
```
tofumatt, line 39: Do we really need to be using `target="_blank"` here? It's a lousy experience on any device, especially mobile, it presents accessibility problems, and is really only defensible if the navigation action will lead to data loss (eg the user is completing a form we don't save and then navigate away with a link). Savvy users can _choose_ to open links in a new tab, but can't choose _not_ to when we set this attribute.

I don't think the link in this context does anything like that, right? If the user can click "back" after navigating to that link we should ditch the `target="_blank"` here.

assets/sass/utilities/_text-align.scss
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+.googlesitekit-plugin {
+
+	.googlesitekit-text-align-center {
+		text-align: center !important;
```
tofumatt, line 22: 😱 

I know sometimes when dealing with frameworks and giant libraries of SCSS we need to use `!important`, but I've always found it helpful to annotate _why_ it's needed/what's stopping us from omitting it. That makes refactoring later easier and gives context to the choice, rather than having them build up over time. What's the reason for this `!important` use? If you could add it in a comment here that'd be awesome 🙂

stories/module-pagespeed-insights-components.stories.js
```diff
@@ -31,14 +31,59 @@ import { STORE_NAME as CORE_SITE } from '../assets/js/googlesitekit/datastore/si
...
+				<Layout>
+					<DashboardPageSpeed />
+				</Layout>
+			</WithTestRegistry>
+		);
+	} )
+	.add( 'Dashboard widget (loading)', () => {
+		fetchMock.getOnce(
+			/^\/google-site-kit\/v1\/modules\/pagespeed-insights\/data\/pagespeed/,
+			new Promise( () => {} ) // Never returns a response for perpetual loading.
```
tofumatt, line 58: This comment is helpful but took me a second to parse. I think we could clarify it a bit:

```suggestion
			new Promise( () => {} ) // Never return a response, to keep the component in a perpetual "loading" state.
```

assets/js/modules/pagespeed-insights/components/DataSourceLearnMoreLinks.js
aaemnnosttv: I agree. It's not actually necessary anymore as the two will link to the same place so there will only be a single component 👍

assets/sass/utilities/_text-align.scss
aaemnnosttv: 😄 
It's simply a utility class, so it makes sense to use `!important` as it's intended to trump other rules and we can't rely on specificity alone since this selector has very little. We should always advocate for not using `important` in module or other non-utility styles but utilities are an exception I would say. The margin utility we have has this as well.

aaemnnosttv:
Thanks @felixarntz I was just about to ping you to take a look.

I'll take a look at the first two. The link was removed [per @marrrmarrr](https://www.figma.com/file/niZV5sVfeaE5sqnJAUgR3G?node-id=1:472#30196739) as a number of copy changes that were requested earlier today.

aaemnnosttv:
@felixarntz I've tracked down the opacity to be part of the `.mdc-tab__text-label` styles, so that is okay, it's just that the text label color is currently being set to #757575 which is itself a medium gray color, so it looks very light with the reduced opacity. You can currently see this in the tab bar on the settings page:

image

Using the default black for the text label color results in more clear text, even though it never appears completely black because of the opacity:
image

Here's how the PSI tabs look with the same tab styles
image

The light blue background is present for both, that's from the active/focus state.

I believe this change should only affect the settings page because those are the only other tabs I can think of. If you're okay with this change I'll update it as it will likely require VRT changes approved as well.

felixarntz:
@aaemnnosttv Sounds good, let's make the inactive text color more readable for both of these tab instances.

.storybook/fetch-mocks.js
jqlee85: Thanks for the catch!

assets/js/modules/pagespeed-insights/dashboard/dashboard-widget-speed.js
```diff
@@ -25,12 +25,17 @@ import { __ } from '@wordpress/i18n';
...
+import Data from 'googlesitekit-data';
 import DashboardModuleHeader from '../../../components/dashboard/dashboard-module-header';
 import DashboardPageSpeed from '../components/DashboardPageSpeed';
+import { STORE_NAME } from '../../../googlesitekit/datastore/site/constants';
+const { useSelect } = Data;
 
 function DashboardSpeed() {
-	const description = global.googlesitekit.permaLink ? __( 'How fast this page is.', 'google-site-kit' ) : __( 'How fast your home page is.', 'google-site-kit' );
+	const currentEntityURL = useSelect( ( select ) => select( STORE_NAME ).getCurrentEntityURL() );
+	const description = currentEntityURL
```
felixarntz, line 36: This technically works, however it may be safer to use both `getCurrentReferenceURL` and `getCurrentEntityURL` and then make this condition `currentReferenceURL === currentEntityURL`. Currently, this code only works based on the internals of `getCurrentReferenceURL`, so by checking both we would decouple that.

assets/sass/utilities/_text-align.scss
felixarntz: I would advise against the `!important` flag unless we actually run into an issue with it not being specific enough, at which point we could reconsider. When looking at our usage of it, it appears to be without it would work just as well, so let's remove the flag.

assets/js/modules/pagespeed-insights/components/ReportDetailsLink.js
felixarntz: This shouldn't be changed at this point. While I'm aware of the accessibility implications of this, we have these as external links elsewhere. At a minimum, even if we removed the `target="_blank"`, this should also have the external link indicator removed.

aaemnnosttv:
@googlebot 👋

Issue #716, opened 2019-10-25
Allow changing Analytics view name when creating a new view


ernee:
## Bug Description

Analytics new view creation duplicates default “All Web Site Data” view. No indication that the view was created, that the user needs to modify the view, or where to do it (on Analytics platform). Also referenced in #522 .

## Steps to reproduce

1. Go to the Analytics Module
2. Click on the View dropdown and select _Setup a new view_
3. Click on the button _Confirm Changes_
4. After the page refreshes, the new view named “All Website Data” is selected, which is the same as the default view in Analytics.
5. Navigating to the Analytics account, one gets a notification that there are “Redundant Views”

## Screenshots

**Duplicate All Website Data View in Site Kit**

![Duplicate All Website Data View in Site Kit](https://user-images.githubusercontent.com/8108643/67594652-ad955980-f719-11e9-8757-2758a65efc69.png)

**Duplicate All Website Data View in Google Analytics** 

![Duplicate All Web Site Data View in Analytics](https://user-images.githubusercontent.com/8108643/67594774-ef260480-f719-11e9-9cb9-b4cc98b984de.png)

**Redundant View Notification in Google Analytics** 

![Redundant View Notification in GA](https://user-images.githubusercontent.com/8108643/67594852-17156800-f71a-11e9-96cb-b4bca8e1250d.png)

## Additional Context

 - PHP Version: 7.1.30
 - OS: MacOS
 - Browser: Chrome
 - Plugin Version:  1.0.0-rc.1.1
 - Device: MacBook Air

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

If the user selects "Setup a new view": 
* display a field below the dropdown boxes where they can name the new view something different than the default ("All website data")
* display a message "You can make changes to this view (e.g. exclude URL query parameters) in Google Analytics.", without a link. 

Once the user creates the new view and saves it by clicking "configure analytics":
* in the Settings section display the same text ("You can make changes to this view (e.g. exclude URL query parameters) in Google Analytics.") and a deep link to the relevant part of GA FE where they can make more changes to the specific view (link would be in the format `https://analytics.google.com/analytics/web/?authuser={email}#/a[accountID]w[webPropertyID]p[profileID]/admin/view/settings`)

## Implementation Brief

* When the "Set up a new view" option is chosen during module set up or settings edit, render an additional text input for the new view name (likely right below the dropdown), which defaults to "All Web Site Data"
[...]

Issue #1654, opened 2020-06-09
Use component name for component file names


felixarntz:
## Feature Description

As decided recently, JavaScript components should have their respective file or folder names match the component name. For example, a component `DashboardPageSpeed` should be in either `DashboardPageSpeed.js` or `DashboardPageSpeed/index.js`.

Another change specific to Site Kit module components is that all components should be within `assets/js/modules/{module}/components` instead of `assets/js/modules/{module}`.

As part of this issue, (almost) all refactored components should be renamed/moved to match these standards. To not go completely overboard with this work and to ensure it doesn't touch too many files we may soon remove anyway, this issue should specifically focus on anything in `assets/js/googlesitekit` and `assets/js/modules`.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All module components should be found within `assets/js/modules/{module}/components`. Module directories that only have components should become subdirectories of that general components directory.
* All components within `assets/js/googlesitekit` and `assets/js/modules` should have their file names (or directory names, where applicable) changed to match the component name (plus `.js` of course 😄 ).

## Implementation Brief

* `npm install --save-dev eslint-plugin-filenames` in the repo
* Add `"filenames"` to `eslintrc.json`'s `plugins` section
* Add checks for all files in a `components/` folder with ESLint's overrides:
```json
    {
      "files": [
        "**/components/**/*.js"
      ],
      "rules": {
        "filenames/match-exported": [
          2,
          "pascal"
        ]
      }
    },
```
* Fix all _component_ names flagged by ESLint. Tested this and it looks like there shouldn't be any files flagged that aren't legitimate errors, so no overrides should be needed. If any overrides _are_ needed though (eg I missed something checking the ouput of `npm run lint:js`: feel free to add them 😄 

## QA Brief

* 

## Changelog entry

* N/A

swissspidy:
@felixarntz Have you considered enforcing this with ESLint? See https://github.com/selaux/eslint-plugin-filenames#matching-exported-values-match-exported for a possible solution

felixarntz:
@tofumatt IB sounds good overall, however I wonder whether we should do it the other way around and only apply this file name linting rule to files that are within a `components` folder? There are some components where this is currently not the case, and some non-components that are in a `components` folder, but we could add exceptions for those or then move them somewhere more appropriate.

tofumatt:
I've amended the overrides in the IB to use:

```json
[...]

PR #1680, opened 2020-06-16
Enhancement/716 change analytics view name

eugene-manuilov:
## Summary

Addresses issue #716

## Relevant technical choices

Added the ability to enter a name for a new Analytics view and displayed a link to the current view settings page.

## Checklist


assets/js/modules/analytics/common/profile-name.js
```diff
@@ -0,0 +1,83 @@
...
+import { STORE_NAME as CORE_FORMS } from '../../../googlesitekit/datastore/forms';
+import { TextField, HelperText, Input } from '../../../material-components';
+import { STORE_NAME, PROFILE_CREATE, FORM_SETUP } from '../datastore/constants';
+const { useSelect, useDispatch } = Data;
+
+export default function ProfileName() {
+	const propertyID = useSelect( ( select ) => select( STORE_NAME ).getPropertyID() );
+	const profiles = useSelect( ( select ) => select( STORE_NAME ).getProfiles( propertyID ) );
+	const profileID = useSelect( ( select ) => select( STORE_NAME ).getProfileID() );
+	const profileName = useSelect( ( select ) => select( CORE_FORMS ).getValue( FORM_SETUP, 'profileName' ) );
```
aaemnnosttv, line 38: Let's default this to `__( 'All Web Site Data', 'google-site-kit' )` as the field should be pre-filled with this when the user selects the option to "set up a new view". We'll probably need to use `useEffect` to set this in the form state on the fly.

This is a slightly better UX as well as the field label looks like a value when the field is empty.

I realize we have the default in place on the server side too; that's fine and we can keep that as the endpoint there should have that default as well - this is just about the UI.

assets/js/modules/analytics/datastore/profiles.js
```diff
@@ -98,12 +99,13 @@ const baseActions = {
 	 * @since 1.8.0
 	 *
 	 * @param {string} propertyID Google Analytics property ID.
+	 * @param {string} profileName The name for a new profile.
 	 * @return {Object} Object with `response` and `error`.
 	 */
-	*createProfile( propertyID ) {
+	*createProfile( propertyID, profileName = '' ) {
```
aaemnnosttv, line 105: After thinking about this a bit, let's make `profileName` required on the client side.

The reason is that if a user cleared the input for the profile name and hit submit, they would end up with a new view using the default (from the server) but would not see the warning about the duplicate view name if they already had one.

Apart from requiring the parameter, we should add `invariant` for it here, as well as adding logic for it into the `canSubmitChanges` selector, so that if `PROFILE_CREATE` is chosen and no `profileName` is set it returns `false`.

assets/js/modules/analytics/settings/settings-view.js
```diff
@@ -28,20 +28,38 @@ import Data from 'googlesitekit-data';
...
+
+	for ( let i = 0, len = strings.length; i < len; i++ ) {
+		uri += strings[ i ];
+		if ( i < values.length ) {
+			uri += encodeURIComponent( values[ i ] );
+		}
+	}
+
+	return uri;
+}
```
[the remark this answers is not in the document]
eugene-manuilov: > What do you think about using reduce instead? I find it's a bit easier to read than a for loop.

IMO for loop is the easiest construction to read and understand, especially for those who has never (or very rarely) used `reduce` function. But anyway, i'll rework that escapeURI tag function to use `reduce`.

assets/js/modules/analytics/common/profile-name.js
```diff
@@ -0,0 +1,83 @@
+/**
+ * Analytics Property Select component.
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/analytics/common/profile-name.js
```diff
@@ -0,0 +1,83 @@
...
+	const profiles = useSelect( ( select ) => select( STORE_NAME ).getProfiles( propertyID ) );
+	const profileID = useSelect( ( select ) => select( STORE_NAME ).getProfileID() );
+	const profileName = useSelect( ( select ) => select( CORE_FORMS ).getValue( FORM_SETUP, 'profileName' ) );
+
+	const { setValues } = useDispatch( CORE_FORMS );
+	const onChange = useCallback( ( { currentTarget } ) => {
+		setValues( FORM_SETUP, {
+			profileName: currentTarget.value,
+		} );
+	}, [ profileID ] );
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/analytics/common/profile-name.js
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Internal dependencies
+ */
+import Data from 'googlesitekit-data';
+import { STORE_NAME as CORE_FORMS } from '../../../googlesitekit/datastore/forms';
+import { TextField, HelperText, Input } from '../../../material-components';
+import { STORE_NAME, PROFILE_CREATE, FORM_SETUP } from '../datastore/constants';
+const { useSelect, useDispatch } = Data;
+
+export default function ProfileName() {
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/analytics/common/profile-name.js
```diff
@@ -0,0 +1,83 @@
...
+	const { setValues } = useDispatch( CORE_FORMS );
+	const onChange = useCallback( ( { currentTarget } ) => {
+		setValues( FORM_SETUP, {
+			profileName: currentTarget.value,
+		} );
+	}, [ profileID ] );
+
+	// bounce if an existing profile is selected
+	if ( profileID !== PROFILE_CREATE ) {
+		return false;
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/analytics/common/profile-name.js
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	let helperText;
+	let trailingIcon;
+
+	const existingProfile = profiles.find( ( item ) => item.name === profileName );
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/analytics/common/profile-name.js
```diff
@@ -0,0 +1,83 @@
...
+				<span className="screen-reader-text">
+					{ __( 'Warning', 'google-site-kit' ) }
+				</span>
+			</span>
+		);
+	}
+
+	return (
+		<div className="googlesitekit-analytics-profilename">
+			<TextField label="View Name" outlined helperText={ helperText } trailingIcon={ trailingIcon }>
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/analytics/settings/settings-view.js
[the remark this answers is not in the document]
eugene-manuilov: Updated.

includes/Modules/Analytics.php
```diff
@@ -1001,8 +1001,12 @@ function ( $order_def ) {
 						array( 'status' => 400 )
 					);
 				}
+				$profile_name = ! empty( $data['profileName'] ) ? trim( $data['profileName'] ) : '';
```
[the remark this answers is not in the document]

assets/js/modules/analytics/common/profile-name.js
eugene-manuilov: Updated.

assets/js/modules/analytics/datastore/profiles.js
eugene-manuilov: Updated.

assets/js/modules/analytics/util/validation.js
```diff
@@ -101,6 +101,16 @@ export function isValidProfileSelection( value ) {
 	return isValidNumericID( value );
 }
 
+/**
+ * Checks if the given profile name appears to be valid.
+ *
+ * @param {string} value Profile name to test
+ * @return {boolean} True if valid, otherwise false.
+ */
+export function isValidProfileName( value ) {
```
aaemnnosttv, line 110: 👍

assets/js/modules/analytics/datastore/settings.test.js
```diff
@@ -392,6 +415,7 @@ describe( 'modules/analytics settings', () => {
 				registry.dispatch( STORE_NAME ).receiveGetExistingTag( null );
 				registry.dispatch( STORE_NAME ).setSettings( validSettings );
 				registry.dispatch( STORE_NAME ).setProfileID( PROFILE_CREATE );
+				registry.dispatch( CORE_FORMS ).setValues( FORM_SETUP, { profileName: 'all web site data' } );
```
aaemnnosttv, line 418: Please also add a test for `canSubmitChanges` that it requires a valid `profileName` when creating a profile is selected.

assets/js/modules/analytics/datastore/profiles.js
```diff
@@ -162,8 +164,8 @@ const baseSelectors = {
 	 * @param {string} propertyID The Analytics Property ID to check for profile creation.
 	 * @return {boolean} `true` if creating a profile, `false` if not.
 	 */
-	isDoingCreateProfile: createRegistrySelector( ( select ) => ( state, propertyID ) => {
-		return select( STORE_NAME ).isFetchingCreateProfile( propertyID );
+	isDoingCreateProfile: createRegistrySelector( ( select ) => ( state, propertyID, profileName ) => {
+		return select( STORE_NAME ).isFetchingCreateProfile( propertyID, profileName );
```
aaemnnosttv, line 168: I understand why this was added here but I feel like the name isn't really important for this selector.
Unfortunately, it is needed to use the underlying `isFetching` selector. What we can do though instead is change this selector to return true if _any_ create profile request is in progress (we have a few other selectors which do a similar thing). 

E.g.
https://github.com/google/site-kit-wp/blob/5efc53aad0542751420381cbf8b522d10b5d5b50/assets/js/modules/analytics/datastore/accounts.js#L328-L334

Let's do the same here, since we could always technically select `isFetchingCreateProfile` if really needed.

eugene-manuilov: Updated.

assets/js/modules/analytics/datastore/settings.test.js
eugene-manuilov: Updated.

assets/js/modules/analytics/datastore/profiles.js
```diff
@@ -98,12 +100,14 @@ const baseActions = {
 	 * @since 1.8.0
 	 *
 	 * @param {string} propertyID Google Analytics property ID.
+	 * @param {string} profileName The name for a new profile.
 	 * @return {Object} Object with `response` and `error`.
 	 */
-	*createProfile( propertyID ) {
+	*createProfile( propertyID, profileName ) {
```
aaemnnosttv, line 106: @felixarntz @tofumatt what do you guys think about `profileName` here? Should it be a positional argument or in an object? I [made the case earlier that it should be required](https://github.com/google/site-kit-wp/pull/1680#discussion_r441443813) and I still think it should be, but it just seems odd to have a name as a positional parameter. What if we wanted to support customizing other details of a profile in the future? Obviously we wouldn't want to keep tacking on positional arguments. It's a bit odd here only because I think this is the first action where we've had to make a decision like this. Most other parameters for other actions and selectors are IDs.

tofumatt: Yeah, I agree: I think `*createProfile( propertyID, { name } = {} ) {` would be the more appropriate function signature here. Leaves room for customisation, for something that could easily get more arguments. 👍

aaemnnosttv: It would be `*createProfile( propertyID, { name } )` since the profile name is required 😄

felixarntz: Using an object sounds good to me here, and yes, the object should not have a default then 👍

assets/js/modules/analytics/common/profile-name.js
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * Internal dependencies
+ */
+import Data from 'googlesitekit-data';
+import { STORE_NAME as CORE_FORMS } from '../../../googlesitekit/datastore/forms';
+import { TextField, HelperText, Input } from '../../../material-components';
+import { STORE_NAME, PROFILE_CREATE, FORM_SETUP } from '../datastore/constants';
+const { useSelect, useDispatch } = Data;
+
+export default function ProfileNameTextField() {
```
aaemnnosttv, line 34: The file name here should be the same as the name of the component (i.e. `ProfileNameTextField.js`). This is more of a recent change so it isn't exactly obvious yet. We have an issue to update all the current usage separately. See #1654

assets/js/modules/analytics/common/profile-name.js
```diff
@@ -0,0 +1,85 @@
...
+				<span className="screen-reader-text">
+					{ __( 'Warning', 'google-site-kit' ) }
+				</span>
+			</span>
+		);
+	}
+
+	return (
+		<div className="googlesitekit-analytics-profilename">
+			<TextField label={ __( 'View Name', 'google-site-kit' ) } outlined helperText={ helperText } trailingIcon={ trailingIcon }>
```
aaemnnosttv, line 76: Please break these props onto multiple lines.

assets/js/modules/analytics/datastore/profiles.js
aaemnnosttv: @eugene-manuilov after checking with @tofumatt and @felixarntz, we've decided to change the second argument to `createProfile` to be an object which for now will only have `profileName` instead of the positional argument. See https://github.com/google/site-kit-wp/pull/1680#discussion_r442957195

Basically the signature would change like so:
```diff
- *createProfile( propertyID, profileName ) {
+ *createProfile( propertyID, { profileName } ) {
```
- where the second parameter would be an object that we destructure in the signature. As for docblocks, let's call the parameter `args` since it's required, and then `profileName` would be documented as `args.profileName`.

As for `argsToParams` this would destructure it out as well, and the rest would be the same (it wouldn't pass the whole object through in params).

`controlCallback` and `reducerCallback` would be unchanged since those take the result of `argsToParams`. This change should mostly be about updating the usage and tests.

eugene-manuilov: Updated.

assets/js/modules/analytics/common/profile-name.js
eugene-manuilov: Updated.

assets/js/modules/analytics/common/profile-name.js
eugene-manuilov: Updated.

felixarntz:
Solid work here, UI looks good to me!

PR #1683, opened 2020-06-17
Changed getauthentication to needsreauthentication in info store.

ShahAaron:
## Summary

Addresses issue #1559 

## Checklist

### DIFF
--- a/assets/js/googlesitekit/modules/create-info-store.js
+++ b/assets/js/googlesitekit/modules/create-info-store.js
@@ -76,10 +76,14 @@ export const createInfoStore = ( slug, {
 		 * @since n.e.x.t
 		 *
 		 * @param {boolean} reAuth The module activation status. Default is true.
-		 * @return {(string|undefined)} The admin reauthentication URL.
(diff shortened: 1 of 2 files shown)


assets/js/googlesitekit/modules/create-info-store.js
```diff
@@ -79,7 +79,7 @@ export const createInfoStore = ( slug, {
 		 * @return {(string|undefined)} The admin reauthentication URL.
 		 */
 		getAdminReauthURL: createRegistrySelector( ( select ) => ( state, reAuth = true ) => {
-			const { needsReauthentication } = select( CORE_USER ).getAuthentication() || {};
+			const needsReauthentication = select( CORE_USER ).needsReauthentication() || false;
```
felixarntz, line 82: Seeing this now, it actually needs another small tweak which is more important even than using the selector: If `needsReauthentication` is `undefined`, this shouldn't assume `false` - it should instead return `undefined`, because it means part of the dependency data hasn't loaded yet.

Issue #1610, opened 2020-05-29
Avoid extra click after granting OAuth scopes for a certain action


felixarntz:
## Feature Description

Follow-up to #1566: As the outcome there, after a user has returned from OAuth after granting the additional permissions, they end up on the screen they originally were on. However, at the moment they still have to re-trigger the same action (e.g. click on "Configure Analytics" if they had selected "Setup new property"), which isn't a great user experience.

For Analytics account provisioning, this experience is already in place because it was implemented there specifically. We should now look into how we can implement this with a more general approach, similar to how the `PermissionsModal` and permissions scope error handling itself is not coupled to any specific module or screen.

Since re-triggering the exact action is tied to the specific component that e.g. holds the button to click, it's unlikely that we can fully "automate" this without requiring any extra code per case. However, let's look for a solution that scales as much as possible, e.g. set a flag in the datastore automatically that the component could react on, re-triggering the action previously attempted (which resulted in the permissions modal and redirect).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `core/user` store should have a new selector `hasScope( scope )` which checks whether `getGrantedScopes()` includes the scope checked for.
* The specific `has*Scope()` selectors on `modules/analytics` should be removed and their usage should be replaced with calling the above `hasScope( scope )`.

## Implementation Brief

* Add a new `hasScope( scope: string )` registry selector to `assets/js/googlesitekit/datastore/user/authentication.js`
	- Returns `({boolean|undefined})` based on whether or not granted scopes are loaded yet  
	If loaded, returns `grantedScopes.includes( scope )`
	- Add Jest tests for selector behavior, similar to other selectors
* Refactor [Analytics-specific scope selectors](https://github.com/google/site-kit-wp/blob/develop/assets/js/modules/analytics/datastore/permissions.js) with new `hasScope` using scope constants
* Remove `assets/js/modules/analytics/datastore/permissions.js` and inclusion in combined store

## QA Brief

* 

## Changelog entry

* Add `hasScope( scope )` selector to `core/user` store, which allows checking whether the user has explicitly granted access to the respective scope.

aaemnnosttv:
@felixarntz this issue is a bit difficult to define as-is as the current AC is already met with the use-cases we need it for.

[...]

PR #1686, opened 2020-06-17
Change selectors to use hasScope()

tofumatt:
## Summary

Addresses issue #1610.

## Checklist


assets/js/googlesitekit/datastore/user/authentication.js
```diff
@@ -81,6 +81,24 @@ const baseSelectors = {
...
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @param {string} scope The scope constant to check for.
+	 * @return {(boolean|undefined)} `true` if scope is present; `false` if not.
+	 */
+	hasScope: createRegistrySelector( ( select ) => ( state, scope ) => {
+		const grantedScopes = select( STORE_NAME ).getGrantedScopes( state );
+
+		return ( grantedScopes !== undefined ) ? grantedScopes.includes( scope ) : undefined;
```
felixarntz, line 99: Nit-picking, but I think using a regular `if` clause for checking and returning undefined is easier to follow than this ternary one.

assets/js/googlesitekit/datastore/user/authentication.test.js
```diff
@@ -20,6 +20,7 @@
  * Internal dependencies
  */
 import API from 'googlesitekit-api';
+import { PROVISIONING_SCOPE, EDIT_SCOPE } from '../../../modules/analytics/datastore/constants';
```
felixarntz, line 23: Not a fan of importing constants specific to a module in this general core test. Since this is about testing the selector itself, we can just provide any arbitrary scope here, rather than rely on constants that come from a completely different "package".

Issue #1611, opened 2020-05-29
Modify apiFetch to only rely on server-preloaded data for the first request


felixarntz:
## Bug Description

We currently preload data for several API endpoints so that no requests on pageload are necessary:
* connection data for `core/site` store
* authentication data for `core/user` store
* module information for `core/modules` store
* settings for all `modules/*` stores with settings

There is a problem with that approach because the `apiFetch` middleware will always have these API requests resolve with this data, even later during the app lifecycle where things might have changed. For example, when you activate a module, `core/user` authentication data should be refetched. We don't use this anywhere yet so it's not critical, but in the current stage it will not work since it would still not issue an API request and instead rely on the server-preloaded data which at that point will be stale.

We should modify this behavior (e.g. by customizing the behavior of the middleware or potentially implementing our own version of it) so that it only uses the preloaded data for the initial request(s). It should just be one request, but in theory there could be a race condition where the data is requested multiple times. What would probably make sense is to "invalidate" the preloaded data after it is first used, potentially with some timeout of 500ms or so to cater for the race condition (maybe that's not even worth doing).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

### API preloaded data handling
* Server-preloaded data should only be used for `apiFetch` requests for the initial request(s). For subsequent requests, the data should be invalidated so that the actual API request will be fired.
    * Preloaded data should be invalidated as soon as it's used once for a request.
        * Theoretically this would be a problem if two similar requests were fired at about the same time, but the way that datastore resolvers work and the way we're using them make this practically irrelevant.
    * Since it could happen that no such request happens on pageload, we should set an arbitrary threshold of 1000ms (once we have a PR, let's test whether that value works well or whether we should adjust) after which preloaded data should get invalidated regardless of whether a request has been made that applies to it.
* Server-preloaded data should be used independently of whether a request uses caching or not (other than in the original IB and #1692); its only purpose is to avoid an API request right after pageload (which is common to be caused for all of the JS stores that rely on API requests where we preload data).

### Module store adjustments
* The `core/modules` store should, as part of the `setModuleActivation` action rely on the `core/user` store's authentication infrastructure instead of just refetching a current API response.
* Module data should be preloaded on the server via the `google-site-kit/v1/core/modules/data/list` REST endpoint.

### Custom preloading middleware
[...]

PR #1692, opened 2020-06-18
Bug/1611 api fetch server preloaded first request only

ryanwelcher:
## Summary

Addresses issue #1611

## Checklist


assets/js/googlesitekit/middleware/preloading.js
```diff
@@ -0,0 +1,83 @@
...
+import { addQueryArgs } from '@wordpress/url';
+
+/**
+ * Helper to remove the timestamp query param
+ *
+ * @param {string} uri The URI to remove the timestamp query param from.
+ *
+ * @return {string} Passed URI without the timestamp query param
+ */
+function removeTimestampQueryParam( uri ) {
```
aaemnnosttv, line 32: This doesn't seem to be necessary as `removeQueryArgs` from `@wordpress/url` works with a partial URL.

```
> var wpURL = require('@wordpress/url')
undefined
> wpURL.removeQueryArgs('foo?bar=abz', 'bar')
'foo?'
```

assets/js/googlesitekit/middleware/preloading.js
```diff
@@ -0,0 +1,83 @@
...
+		return result;
+	}, {} );
+
+	return ( options, next ) => {
+		const { parse = true } = options;
+		let uri = options.path;
+		let deleteCache = false;
+		if ( typeof options.path === 'string' ) {
+			const method = options.method || 'GET';
+			if ( options.path.match( /timestamp=[0-9]+/ ) ) {
```
aaemnnosttv, line 57: This should work as well 😄 
```suggestion
			if ( getQueryArg( options.path, 'timestamp' ) ) {
```

assets/js/googlesitekit/middleware/preloading.js
```diff
@@ -0,0 +1,83 @@
...
+function createPreloadingMiddleware( preloadedData ) {
+	const cache = Object.keys( preloadedData ).reduce( ( result, path ) => {
+		result[ getStablePath( path ) ] = preloadedData[ path ];
+		return result;
+	}, {} );
+
+	return ( options, next ) => {
+		const { parse = true } = options;
+		let uri = options.path;
+		let deleteCache = false;
```
aaemnnosttv, line 54: This variable is unnecessary as the cache should always be deleted for a matched path, the only difference is when it's deleted.

* If there's no cache for the request then it simply goes to the next middleware
* If there is a `timestamp` (i.e. `useCache: false`), then the cache for a matching path without the `timestamp` parameter should be deleted immediately, and the `next` should be called/returned (same as if there was no cache).
* If not, then the cache is used for the response and then deleted (like you have it)

assets/js/googlesitekit/middleware/preloading.test.js
```diff
@@ -0,0 +1,58 @@
...
+ * WordPress dependencies
+ */
+import { addQueryArgs } from '@wordpress/url';
+
+/**
+ * Internal dependencies
+ */
+import createPreloadingMiddleware from './preloading';
+
+describe( 'Preloading Middleware', () => {
```
aaemnnosttv, line 29: It would be nice if we could also add a test or two for this to show that it works as expected when used with `apiFetch` (doesn't need to do all the same things as the middleware test here but mostly to show it works in integration).

assets/js/googlesitekit/middleware/preloading.js
```diff
@@ -0,0 +1,83 @@
...
+		if ( typeof options.path === 'string' ) {
+			const method = options.method || 'GET';
```
aaemnnosttv, line 56: Let's cast `options.method` to uppercase if it exists, just to be safe. Should be simple with conditional chaining 👍

assets/js/googlesitekit/middleware/preloading.js
```diff
@@ -0,0 +1,83 @@
...
+		const split = paramSet.split( '=' );
+		if ( split[ 0 ] !== 'timestamp' ) {
+			return { ...acc, [ split[ 0 ] ]: split[ 1 ] };
+		}
+	}, {} );
+
+	return addQueryArgs( baseUrl, paramsObject );
+}
+
+function createPreloadingMiddleware( preloadedData ) {
```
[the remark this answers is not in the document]
ryanwelcher: Updated.

assets/js/googlesitekit/middleware/preloading.js
```diff
@@ -0,0 +1,83 @@
...
+ * WordPress dependencies.
+ */
+import { getStablePath } from '@wordpress/api-fetch/build/middlewares/preloading';
+import { addQueryArgs } from '@wordpress/url';
+
+/**
+ * Helper to remove the timestamp query param
+ *
+ * @param {string} uri The URI to remove the timestamp query param from.
+ *
```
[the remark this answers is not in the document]
ryanwelcher: This was removed in favor of` removeQueryArg`

assets/js/googlesitekit/middleware/preloading.js
```diff
@@ -0,0 +1,83 @@
+/**
```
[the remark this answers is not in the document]
ryanwelcher: I have moved the middleware file.

assets/js/googlesitekit/api/index.js
```diff
@@ -117,6 +118,11 @@ export const siteKitRequest = async ( type, identifier, datapoint, {
 			return value;
 		}
 	}
+	// If we're bypassing the cache on GET requests, add the timestamp parameter.
+	if ( false === useCacheForRequest && method === 'GET' ) {
+		// Add timestamp parameter to ensure that only the first request is preloaded.
```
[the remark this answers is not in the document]
ryanwelcher: Updated.

assets/js/util/refresh-authentication.js
```diff
@@ -30,7 +30,7 @@ export const refreshAuthentication = async () => {
 			requiredScopes,
 			grantedScopes,
 			unsatisfiedScopes = [],
-		} = await data.get( TYPE_CORE, 'user', 'authentication', { timestamp: Date.now() } );
```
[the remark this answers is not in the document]
ryanwelcher: Restored.

assets/js/googlesitekit/middleware/preloading.js
ryanwelcher: Updated.

assets/js/googlesitekit/middleware/preloading.js
ryanwelcher: Updated.

assets/js/googlesitekit/middleware/preloading.js
ryanwelcher: Updated.

assets/js/googlesitekit/middleware/preloading.js
ryanwelcher: Updated.

assets/js/googlesitekit/api/middleware/preloading.js
```diff
@@ -0,0 +1,73 @@
...
+	return ( options, next ) => {
+		const { parse = true } = options;
+		let uri = options.path;
+		if ( typeof options.path === 'string' ) {
+			const method = options.method?.toUpperCase() || 'GET';
+
+			// If the path contains the timestamp param, we remove the cache and return the next() call.
+			if ( getQueryArg( options.path, 'timestamp' ) ) {
+				uri = removeQueryArgs( options.path, 'timestamp' );
+				delete cache[ path ];
```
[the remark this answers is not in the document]
ryanwelcher: Thanks for catching this!

assets/js/googlesitekit/api/middleware/preloading.test.js
```diff
@@ -0,0 +1,193 @@
...
+		const preloadingMiddleware = createPreloadingMiddleware( {} );
+		const requestOptions = {
+			method: 'GET',
+			path: requestURI,
+		};
+		await preloadingMiddleware( requestOptions, next );
+		expect( next ).toHaveBeenCalled();
+	} );
+
+	it( 'returns a preloaded response only once for each preloaded request', async () => {
```
aaemnnosttv, line 115: This test seems redundant now as the "returns a preloaded response from multiple URIs" test already covers this.

assets/js/googlesitekit/api/middleware/preloading.test.js
```diff
@@ -0,0 +1,193 @@
...
+		};
+		const next = jest.fn();
+		const firstResponse = await preloadingMiddleware( requestOptions, next );
+		expect( firstResponse ).toEqual( preloadedData[ requestURI ].body );
+		expect( next ).not.toHaveBeenCalled();
+
+		const secondRequest = await preloadingMiddleware( requestOptions, next );
+		expect( secondRequest ).toBeUndefined();
+		expect( next ).toHaveBeenCalled();
+	} );
```
aaemnnosttv, line 133:
```suggestion
	} );
	
```

assets/js/googlesitekit/api/middleware/preloading.test.js
```diff
@@ -0,0 +1,193 @@
...
+	it( 'deletes a preloaded response from the cache when requested with a timestamp query paramater', async () => {
```
aaemnnosttv, line 134:
```suggestion
	it( 'deletes a preloaded response from the cache when requested with a timestamp query parameter', async () => {
```

assets/js/googlesitekit/api/middleware/preloading.test.js
```diff
@@ -0,0 +1,193 @@
...
+			apiFetch = require( '@wordpress/api-fetch' ).default;
+			apiFetch.use( createPreloadingMiddleware( preloadedData ) );
+		} );
+
+		afterEach( async () => {
+			// Invalidate the require cache for `api-fetch` so that it uses a fresh instance.
+			delete require.cache[ require.resolve( '@wordpress/api-fetch' ) ];
+		} );
+
+		it( 'returns a preloaded response when present.', async () => {
```
aaemnnosttv, line 167:
```suggestion
		it( 'returns a preloaded response when present', async () => {
```

assets/js/googlesitekit/api/middleware/preloading.test.js
```diff
@@ -0,0 +1,193 @@
...
+			fetchMock.any( '*', 200 );
+			const response = await apiFetch( {
+				method: 'GET',
+				path: requestURI,
+			} );
+			expect( response ).toEqual( preloadedData[ requestURI ].body );
+			expect( fetchMock ).not.toHaveFetched();
+		} );
+
+		it( 'returns an uncached response when a timestamp query parameter is present.', async () => {
```
aaemnnosttv, line 179:
```suggestion
		it( 'returns an uncached response when a timestamp query parameter is present', async () => {
```

assets/js/googlesitekit/api/middleware/preloading.test.js
```diff
@@ -0,0 +1,193 @@
...
+		body: {
+			connected: true,
+			resettable: true,
+			setupCompleted: true,
+		},
+	},
+
+	'google-site-kit/v1/core/user/authentication': {
+		headers: [],
+		body: {
```
aaemnnosttv, line 41: Let's avoid using real endpoints and response data in our tests (where it isn't relevant to the specific test).
The preloading middleware is only slightly different than the core middleware we based it on, and the only detail that is really Site Kit specific is the `timestamp` parameter handling. Apart from that, it will behave the same regardless of what endpoint it is used with.

It also somewhat compels us to maintain/update these if the endpoints or response formats were to change in the future to preserve consistency in the codebase. It also might give the impression to someone less familiar with the codebase that this middleware might only work this way for our endpoints, which isn't the case. Using generic "dummy" data intentionally de-emphasizes bits which are not important (perhaps only for making an assertion) from those that are to better convey intention 🙂 

I think using something like `test/path/a` and `test/path/b` would be just fine 👍

assets/js/googlesitekit/api/middleware/preloading.test.js
```diff
@@ -0,0 +1,193 @@
...
+				'https://www.googleapis.com/auth/webmasters',
+				'https://www.googleapis.com/auth/adsense.readonly',
+				'https://www.googleapis.com/auth/analytics.readonly',
+			],
+			unsatisfiedScopes: [],
+		},
+	},
+};
+
+describe( 'Preloading Middleware', () => {
```
aaemnnosttv, line 58:
```suggestion
describe( 'apiFetch.preloadingMiddleware', () => {
```

assets/js/googlesitekit/api/middleware/preloading.js
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * WordPress dependencies.
+ */
+import { getStablePath } from '@wordpress/api-fetch/build/middlewares/preloading';
+import { removeQueryArgs, getQueryArg } from '@wordpress/url';
+
+/**
+ * createPreloadingMiddleware
+ *
+ * Based on preloadMiddleware from from @wordpress/api-fetch, this middle is a single-use per-endpoint and provides cached
```
aaemnnosttv, line 28:
```suggestion
 * Based on preloadingMiddleware from @wordpress/api-fetch, this middleware is single-use per-endpoint and provides cached
```

assets/js/googlesitekit/api/middleware/preloading.js
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+/**
+ * WordPress dependencies.
+ */
+import { getStablePath } from '@wordpress/api-fetch/build/middlewares/preloading';
+import { removeQueryArgs, getQueryArg } from '@wordpress/url';
+
+/**
+ * createPreloadingMiddleware
```
aaemnnosttv, line 26:
```suggestion
 * Creates a preloadingMiddleware instance with the given preloaded data.
```

assets/js/googlesitekit/api/middleware/preloading.js
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * Based on preloadMiddleware from from @wordpress/api-fetch, this middle is a single-use per-endpoint and provides cached
+ * data for the first request only and any subsequent requests hit the server.
+ *
+ * @since n.e.x.t
+ *
+ * @param {Object} preloadedData Preloaded data paths.
+ * @return {Function} Function.
```
aaemnnosttv, line 34:
```suggestion
 * @return {Function} apiFetch middleware function.
```

Issue #1609, opened 2020-05-29
Migrate all modules to use `Module::get_datapoint_definitions()` and deprecate `Module::get_datapoint_services()`


felixarntz:
## Feature Description

#1566 introduced a new method `Module::get_datapoint_definitions()` that allows for more flexibility for a module to define its datapoint interface, surpassing the previous `Module::get_datapoint_services()`, which only allows a datapoint to be associated with a service, but no additional information. The new method has a base implementation that only exists for backward-compatibility though, in a perfect world it would be abstract.

As part of #1566 only the immediately necessary datapoints (some in Analytics, some in Tag Manager) were migrated to use `Module::get_datapoint_definitions()`, so this issue should migrate all other datapoints over, so that none of our modules any longer implements `Module::get_datapoint_services()`. Any code that currently calls `Module::get_datapoint_services()` should be modified to rely on `Module::get_datapoint_definitions()` (except the one call in `Module::get_datapoint_definitions()` that makes the backward-compatibility work).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All module PHP classes should implement `Module::get_datapoint_definitions()`, covering all datapoints that are currently covered in `Module::get_datapoint_services()`.
    * Note that `Module::get_datapoint_definitions()` also takes the request method into account, while `Module::get_datapoint_services()` didn't (another improvement here).
    * Since all datapoints currently covered by `Module::get_datapoint_services()` only provide the API service name they're using, the migration should be straightforward just providing the `service` key.
* `Module::get_datapoint_services()` shouldn't be implemented or used anymore anywhere, except in the `Module::get_datapoint_definitions()` base implementation.

#### For each `Module` sub-class
* Add a `protected function get_datapoint_definitions()` method (if not present)
* Add a `GET:{datapoint}` key to the returned array for all applicable entries in `get_datapoint_services`
* Add a `POST:{datapoint}` key to the returned array for all applicable entries in `get_datapoint_services`
* All added entries should be a minimum of `array( 'service' => {service} )` where `{service}` is the value defined in `get_datapoint_services`  
This key should always be set, even if empty
* Modules that already implement `get_datapoint_definitions` should be updated to no longer call the `parent::get_datapoint_definitions()`
    - Existing datapoint definitions should be updated to include the `service` key with their respective service
* The `get_datapoint_services` method should be removed
* References to `get_datapoint_services` should be updated to use `get_datapoint_definitions` and must account for different return value

[...]

PR #1705, opened 2020-06-22
Enhancement/1609 modules refactoring

eugene-manuilov:
## Summary

Addresses issue #1609

## Relevant technical choices

Refactored PHP modules to use `get_datapoint_definitions` approach instead of `get_datapoint_services`.

## Checklist


jest-puppeteer.config.js
```diff
@@ -0,0 +1,18 @@
+// https://github.com/smooth-code/jest-puppeteer/blob/master/packages/jest-environment-puppeteer/README.md#jest-puppeteerconfigjs
```
aaemnnosttv, line 1: Let's not override this completely. Instead, lets extend the default config and only add the parts we need:
https://github.com/WordPress/gutenberg/blob/7060b11b9b9efc00e099c9092fe520402d77ee4b/packages/scripts/config/puppeteer.config.js

We can import it from `@wordpress/scripts/config/puppeteer.config.js`

eugene-manuilov: Yep, updated.

includes/Core/Modules/Module.php
```diff
@@ -319,15 +320,15 @@ public function is_connected() {
...
 			}
 
-			if ( empty( $datapoint_services[ $datapoint ] ) ) {
+			if ( empty( $datapoint_services[ $datapoint_name ]['service'] ) ) {
 				continue;
 			}
 
-			if ( ! isset( $service_batches[ $datapoint_services[ $datapoint ] ] ) ) {
-				$service_batches[ $datapoint_services[ $datapoint ] ] = $this->google_services[ $datapoint_services[ $datapoint ] ]->createBatch();
+			if ( ! isset( $service_batches[ $datapoint_services[ $datapoint_name ]['service'] ] ) ) {
```
aaemnnosttv, line 327: Let's extract the service name to make this a bit more readable.
E.g.
```php
$datapoint_service = $datapoint_definitions[ "{$data->method}:{$data->datapoint}" ]['service'];
```

includes/Core/Modules/Module.php
```diff
@@ -388,18 +389,36 @@ public function is_connected() {
...
 	 * @since 1.9.0 No longer abstract.
+	 * @deprecated n.e.x.t
 	 *
 	 * @return array Associative array of $datapoint => $service_identifier pairs.
 	 */
 	protected function get_datapoint_services() {
+		_deprecated_function(
+			sprintf( '%s::get_datapoint_services', static::class ), // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+			'',
+			sprintf( '::get_datapoint_definitions', static::class ) // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
aaemnnosttv, line 419: * The first argument can simply be `__METHOD__`
* The second argument here should be `n.e.x.t`
* Missing placeholder in template string in the 3rd argument

This can be simplified a bit into a single line similar to this without `sprintf`
https://github.com/google/site-kit-wp/blob/3590c624a075e9d9cb28bc9bf6ee313dd7aba516/includes/Core/Authentication/Clients/OAuth_Client.php#L787

includes/Core/Modules/Module.php
```diff
@@ -388,18 +389,36 @@ public function is_connected() {
...
+
+		foreach ( array_keys( $definitions ) as $key ) {
+			$parts = explode( ':', $key );
+			$name  = end( $parts );
+			if ( ! empty( $name ) ) {
+				$keys[ $name ] = $name;
+			}
+		}
+
+		return array_values( $keys );
```
aaemnnosttv, line 403: Just a suggestion, but what do you think about doing it this way?
We can keep it your way as both should work, I prefer avoiding temporary variables when possible as long as it doesn't sacrifice readability.
```suggestion
return array_unique(
	array_map(
		function ( $key ) {
			// Prevent potential undefined index by ensuring second element is always set.
			list ( , $datapoint ) = explode( ':', $key, 2 ) + array( null, null );
			return $datapoint;
		},
		array_keys( $this->get_datapoint_definitions() )
	)
);
```

includes/Modules/AdSense.php
```diff
@@ -347,32 +347,34 @@ protected function amp_content_add_auto_ads( $content ) {
...
+			'GET:alerts'          => array( 'service' => 'adsense' ),
+			'GET:clients'         => array( 'service' => 'adsense' ),
+			'GET:urlchannels'     => array( 'service' => 'adsense' ),
+			'GET:earnings'        => array( 'service' => 'adsense' ),
+			'POST:connection'     => array( 'service' => '' ),
+			'POST:account-id'     => array( 'service' => '' ),
+			'POST:client-id'      => array( 'service' => '' ),
+			'POST:use-snippet'    => array( 'service' => '' ),
+			'POST:account-status' => array( 'service' => '' ),
+			'POST:setup-complete' => array( 'service' => '' ),
```
aaemnnosttv, line 377: Please alpha sort by datapoint name, and secondarily by method. This is how the cases are sorted in `create_data_request` and `parse_data_request` which makes it easier to navigate.

I won't comment on all `get_datapoint_definitions` methods for this but please do this for all of them.

includes/Modules/Analytics.php
```diff
@@ -595,30 +563,46 @@ protected function get_datapoint_services() {
...
+			'POST:tracking-disabled'           => array( 'service' => '' ),
+		);
 
-		$map['POST:create-account-ticket'] = array_merge(
-			$map['POST:create-account-ticket'],
-			array(
-				'scopes'                 => array( 'https://www.googleapis.com/auth/analytics.provision' ),
-				'request_scopes_message' => __( 'You’ll need to grant Site Kit permission to create a new Analytics account on your behalf.', 'google-site-kit' ),
-			)
+		$map['POST:create-account-ticket'] = array(
```
aaemnnosttv, line 590: Similar to my previous comment about sorting, let's always `return array(` in this method now that we aren't calling the parent and move these separate definitions to be inline in the main array in their respective places regardless if they have additional keys or not. Definitions with the extra keys can span multiple lines, the same as it is done here 👍

jest-puppeteer.config.js
```diff
@@ -0,0 +1,20 @@
...
+module.exports = {
+	launch: {
+		...( config.launch || {} ),
+		args: [
+			// https://peter.sh/experiments/chromium-command-line-switches/
+			'--disable-gpu',
+			'--no-sandbox',
+			'--lang=en-US',
+			'--disable-setuid-sandbox',
+			'--disable-dev-shm-usage',
```
aaemnnosttv, line 13: Are all of these arguments necessary? I know `--disable-dev-shm-usage` was the main one to add here but if the others aren't needed, let's not add them unless there's a good reason to.

jest-puppeteer.config.js
```diff
@@ -0,0 +1,20 @@
+const config = require( '@wordpress/scripts/config/puppeteer.config.js' );
```
aaemnnosttv, line 1: Please add a file header.

jest-puppeteer.config.js
```diff
@@ -0,0 +1,20 @@
...
+			'--no-sandbox',
+			'--lang=en-US',
+			'--disable-setuid-sandbox',
+			'--disable-dev-shm-usage',
+		],
+	},
+	// connect: {
+	// 	// https://github.com/puppeteer/puppeteer/blob/main/docs/api.md#puppeteerconnectoptions
+	// 	...( config.connect || {} ),
+	// },
```
aaemnnosttv, line 19: Please remove any unused/unnecessary bits.

package-lock.json
```diff
@@ -7453,6 +7453,12 @@
           "integrity": "sha512-GSmOT2EbHrINBf9SR7CDELwlJ8AENk3Qn7OikK4nFYAu3Ote2+JYNVvkpAEQm3/TLNEJFD/xZJjzyxg3KBWOzw==",
           "dev": true
         },
+        "prettier": {
+          "version": "npm:wp-prettier@1.19.1",
+          "resolved": "https://registry.npmjs.org/wp-prettier/-/wp-prettier-1.19.1.tgz",
+          "integrity": "sha512-mqAC2r1NDmRjG+z3KCJ/i61tycKlmADIjxnDhQab+KBxSAGbF/W7/zwB2guy/ypIeKrrftNsIYkNZZQKf3vJcg==",
+          "dev": true
+        },
```
aaemnnosttv, line 7461: Generally speaking, updates to this file should only happen if it's relevant to the issue. Since this issue is really PHP-specific, this should be left unchanged. I've seen this come up as well and this a simple change so you don't need to revert it; just something to keep in mind for the future 🙂

includes/Core/Modules/Module.php
```diff
@@ -285,7 +285,8 @@ public function is_connected() {
 				continue;
 			}
 
-			if ( ! isset( $datapoint_services[ $dataset->datapoint ] ) ) {
+			$datapoint_name = strtoupper( $dataset->method ) . ':' . $dataset->datapoint;
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

includes/Core/Modules/Module.php
```diff
@@ -263,7 +263,7 @@ public function is_connected() {
 
 		$restore_defer = $this->with_client_defer( true );
 
-		$datapoint_services = $this->get_datapoint_services();
+		$datapoint_services = $this->get_datapoint_definitions();
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

includes/Core/Modules/Module.php
eugene-manuilov: Updated.

includes/Core/Modules/Module.php
eugene-manuilov: Updated.

includes/Modules/AdSense.php
eugene-manuilov: Updated.

jest-puppeteer.config.js
eugene-manuilov: Updated.

jest-puppeteer.config.js
eugene-manuilov: Updated.

jest-puppeteer.config.js
```diff
@@ -0,0 +1,20 @@
+const config = require( '@wordpress/scripts/config/puppeteer.config.js' );
+
+// https://github.com/smooth-code/jest-puppeteer/blob/master/packages/jest-environment-puppeteer/README.md#jest-puppeteerconfigjs
+module.exports = {
+	launch: {
+		...( config.launch || {} ),
+		args: [
+			// https://peter.sh/experiments/chromium-command-line-switches/
```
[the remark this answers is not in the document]
eugene-manuilov: Updated without using lodash.merge.

jest-puppeteer.config.js
eugene-manuilov: Updated.

package.json
```diff
@@ -23,7 +23,7 @@
     "test:js:watch": "wp-scripts test-unit-js --config=tests/js/jest.config.js --watch",
     "pretest:e2e": "./bin/local-env/env-check.sh",
     "test:e2e": "WP_BASE_URL=http://localhost:9002 wp-scripts test-e2e --config=tests/e2e/jest.config.js",
-    "test:e2e:interactive": "npm run test:e2e -- --puppeteer-interactive",
+    "test:e2e:interactive": "PUPPETEER_HEADLESS=false npm run test:e2e",
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

package-lock.json
eugene-manuilov: Got it.

includes/Modules/Analytics.php

includes/Core/Modules/Module.php
eugene-manuilov: Nah... this approach is more cumbersome and is harder to read than just small `foreach` loop.

includes/Modules/AdSense.php
aaemnnosttv: Almost, I meant sorting by datapoint name and secondarily by method to match the sorting we have in the switch cases which follows the same logic:
image

There's nothing enforcing this, but we did have an issue to implement this for the data request methods previously so I think it makes sense to preserve the same for these as well.

includes/Core/Modules/Module.php
```diff
@@ -337,7 +339,7 @@ public function is_connected() {
 			} catch ( Exception $e ) {
 				// Set every result of this batch to the exception.
 				foreach ( $results as $key => $datapoint ) {
-					if ( ! is_string( $datapoint ) || ! isset( $datapoint_services[ $datapoint ] ) || $service_identifier !== $datapoint_services[ $datapoint ] ) {
+					if ( ! is_string( $datapoint ) || ! isset( $datapoint_definitions[ $datapoint_name ] ) || $service_identifier !== $datapoint_definitions[ $datapoint_name ]['service'] ) {
```
aaemnnosttv, line 342: This suggestion removes the [duplicate `isset` check](https://user-images.githubusercontent.com/1621608/86273152-7d6f5600-bbd8-11ea-9ddf-53f021fd4701.png) (not your addition 😄 ) and refactors with `$datapoint_service` 

```suggestion
					if ( ! is_string( $datapoint ) || $service_identifier !== $datapoint_service ) {
```

includes/Core/Modules/Module.php
```diff
@@ -285,7 +285,8 @@ public function is_connected() {
 				continue;
 			}
 
-			if ( ! isset( $datapoint_services[ $dataset->datapoint ] ) ) {
+			$datapoint_name = "{$dataset->method}:{$dataset->datapoint}";
```
aaemnnosttv, line 288: I originally suggested to inline this since there only seems to be 2 instances, but if you'd prefer to keep it as a variable that's okay, however let's use a name like `$definition_key` to not confuse it with `$dataset->datapoint` which is more like the "name" of the datapoint (e.g. `accounts` or `properties-profiles`).

eugene-manuilov: Updated.

includes/Modules/AdSense.php
eugene-manuilov: Ah... ok, updated. Should be good now.

includes/Core/Modules/Module.php
eugene-manuilov: Reworked it in another way because `$datapoint_service` is defined in another loop and we can't rely on it.

includes/Modules/AdSense.php
```diff
@@ -347,32 +347,34 @@ protected function amp_content_add_auto_ads( $content ) {
...
-			'notifications'  => '',
-			'tag-permission' => '',
-			'accounts'       => 'adsense',
-			'alerts'         => 'adsense',
-			'clients'        => 'adsense',
-			'urlchannels'    => 'adsense',
-			'earnings'       => 'adsense',
-			// POST.
-			'setup-complete' => '',
+			'GET:accounts'        => array( 'service' => 'adsense' ),
```
aaemnnosttv, line 358: `GET:accounts` should be after `GET:account-url` and before `GET:alerts`.

includes/Modules/Analytics.php
```diff
@@ -595,33 +563,44 @@ protected function get_datapoint_services() {
 	 * @return array Map of datapoints to their definitions.
 	 */
 	protected function get_datapoint_definitions() {
-		$map = parent::get_datapoint_definitions();
-
-		$map['POST:create-account-ticket'] = array_merge(
-			$map['POST:create-account-ticket'],
-			array(
+		return array(
+			'GET:accounts-properties-profiles' => array( 'service' => 'analytics' ),
```
aaemnnosttv, line 567: `GET:accounts-properties-profiles` should be after `POST:account-id`.

includes/Modules/Analytics.php
```diff
@@ -595,33 +563,44 @@ protected function get_datapoint_services() {
...
+				'service'                => 'analytics',
 				'scopes'                 => array( 'https://www.googleapis.com/auth/analytics.edit' ),
-				'request_scopes_message' => __( 'You’ll need to grant Site Kit permission to create a new Analytics view on your behalf.', 'google-site-kit' ),
-			)
+				'request_scopes_message' => __( 'You\'ll need to grant Site Kit permission to create a new Analytics property on your behalf.', 'google-site-kit' ),
+			),
+			'GET:internal-web-property-id'     => array( 'service' => '' ),
+			'POST:internal-web-property-id'    => array( 'service' => '' ),
+			'GET:goals'                        => array( 'service' => 'analytics' ),
+			'GET:profiles'                     => array( 'service' => 'analytics' ),
```
aaemnnosttv, line 591: `GET:profiles` should come after `POST:profile-id`.

includes/Modules/Search_Console.php
```diff
@@ -156,21 +156,18 @@ public function get_debug_fields() {
...
-			// GET.
-			'sites'           => 'webmasters',
-			'matched-sites'   => 'webmasters',
-			'searchanalytics' => 'webmasters',
-
-			// POST.
-			'site'            => 'webmasters',
+			'GET:matched-sites'   => array( 'service' => 'webmasters' ),
+			'GET:searchanalytics' => array( 'service' => 'webmasters' ),
+			'GET:sites'           => array( 'service' => 'webmasters' ),
```
aaemnnosttv, line 169: `GET:sites` should come after `POST:site`.

includes/Modules/Tag_Manager.php
```diff
@@ -444,17 +420,24 @@ protected function get_datapoint_services() {
...
 	 */
 	protected function get_datapoint_definitions() {
-		$map = parent::get_datapoint_definitions();
-
-		$map['POST:create-container'] = array_merge(
-			$map['POST:create-container'],
-			array(
+		return array(
+			'GET:accounts'               => array( 'service' => 'tagmanager' ),
+			'GET:accounts-containers'    => array( 'service' => 'tagmanager' ),
```
aaemnnosttv, line 425: These first two lines should come after `POST:account-id`.

tests/phpunit/includes/Core/Modules/FakeModule.php
```diff
@@ -108,9 +108,9 @@ public function register() {
 	 * @since 1.0.0
```
aaemnnosttv, line 108: This doc block should be updated as it still has the same one used for datapoint services which is not completely accurate.

includes/Modules/AdSense.php
eugene-manuilov: Updated.

includes/Modules/Analytics.php

includes/Modules/Analytics.php

includes/Modules/Search_Console.php
eugene-manuilov: Updated.

includes/Modules/Tag_Manager.php
eugene-manuilov: Updated.

tests/phpunit/includes/Core/Modules/FakeModule.php
eugene-manuilov: Updated.

includes/Core/Modules/Module.php
```diff
@@ -319,15 +320,16 @@ public function is_connected() {
...
 
-			if ( ! isset( $service_batches[ $datapoint_services[ $datapoint ] ] ) ) {
-				$service_batches[ $datapoint_services[ $datapoint ] ] = $this->google_services[ $datapoint_services[ $datapoint ] ]->createBatch();
+			if ( ! isset( $service_batches[ $datapoint_service ] ) ) {
+				$service_batches[ $datapoint_service ] = $this->google_services[ $datapoint_service ]->createBatch();
 			}
 
-			$service_batches[ $datapoint_services[ $datapoint ] ]->add( $request, $key );
+			$service_batches[ $datapoint_service ]->add( $request, $key );
 			$results[ $key ] = $datapoint;
```
felixarntz, line 333: Since we're now differentiating between the request method (see above `{$dataset->method}:{$dataset->datapoint}`), I think we should do that here as well, i.e. store both `$dataset->method` and `$datapoint` here so that we can reference that in the loop below for handling the exception.

Since the value here is only temporarily set and will always be overridden, changing the format doesn't matter in terms of backward compatibilty.

includes/Core/Modules/Module.php
```diff
@@ -337,11 +339,11 @@ public function is_connected() {
 			} catch ( Exception $e ) {
 				// Set every result of this batch to the exception.
 				foreach ( $results as $key => $datapoint ) {
-					if ( ! is_string( $datapoint ) || ! isset( $datapoint_services[ $datapoint ] ) || $service_identifier !== $datapoint_services[ $datapoint ] ) {
-						continue;
+					$get_service  = ! empty( $datapoint_definitions[ "GET:{$datapoint}" ] ) ? $datapoint_definitions[ "GET:{$datapoint}" ]['service'] : null;
+					$post_service = ! empty( $datapoint_definitions[ "POST:{$datapoint}" ] ) ? $datapoint_definitions[ "POST:{$datapoint}" ]['service'] : null;
+					if ( is_string( $datapoint ) && ( $service_identifier === $get_service || $service_identifier === $post_service ) ) {
```
felixarntz, line 344: See above, this should be changed accordingly. Or alternatively, if I'm not mistaken, the checks here other than `is_string` may be completely unnecessary? Since we already check further above whether there is a service for this datapoint and cancel otherwise?

includes/Core/Modules/Module.php
```diff
@@ -411,18 +426,7 @@ protected function get_datapoint_services() {
...
-			array_keys( $services ),
-			function ( $map, $datapoint ) use ( $services ) {
-				$map[ "GET:$datapoint" ]  = array( 'service' => $services[ $datapoint ] );
-				$map[ "POST:$datapoint" ] = array( 'service' => $services[ $datapoint ] );
-
-				return $map;
-			},
-			array()
-		);
+		return array();
```
felixarntz, line 429: Technically, I think this code should be maintained for backward-compatibility. But since it's unlikely anybody other than us is using this, I'm definitely up for discussing this. _If_ we get rid of the BC layer, then we should properly implement it all the way though, i.e. make `get_datapoint_definitions` abstract, because that's what it would have been if we had originally built it without BC consideration.

includes/Modules/Analytics.php
```diff
@@ -595,33 +563,44 @@ protected function get_datapoint_services() {
...
-			$map['POST:create-profile'],
-			array(
+				'request_scopes_message' => __( 'You\'ll need to grant Site Kit permission to create a new Analytics view on your behalf.', 'google-site-kit' ),
+			),
+			'POST:create-property'             => array(
+				'service'                => 'analytics',
 				'scopes'                 => array( 'https://www.googleapis.com/auth/analytics.edit' ),
-				'request_scopes_message' => __( 'You’ll need to grant Site Kit permission to create a new Analytics view on your behalf.', 'google-site-kit' ),
-			)
+				'request_scopes_message' => __( 'You\'ll need to grant Site Kit permission to create a new Analytics property on your behalf.', 'google-site-kit' ),
```
felixarntz, line 586: These messages should be reverted to their previous values using "You’ll" :)

Also, you sorted the values alphabetically here, so 👍  - while this is not really part of this issue, could you sort the `POST:create-profile` and `POST:create-property` clauses in the `switch` the same way? (They are currently wrong)

includes/Modules/Analytics.php
```diff
@@ -595,33 +563,44 @@ protected function get_datapoint_services() {
...
+			'GET:profile-id'                   => array( 'service' => '' ),
+			'POST:profile-id'                  => array( 'service' => '' ),
+			'GET:profiles'                     => array( 'service' => 'analytics' ),
+			'GET:properties-profiles'          => array( 'service' => 'analytics' ),
+			'GET:property-id'                  => array( 'service' => '' ),
+			'POST:property-id'                 => array( 'service' => '' ),
+			'GET:report'                       => array( 'service' => 'analyticsreporting' ),
+			'GET:tag-permission'               => array( 'service' => '' ),
+			'GET:tracking-disabled'            => array( 'service' => '' ),
+			'POST:tracking-disabled'           => array( 'service' => '' ),
```
felixarntz, line 600: As far as I can see `POST:tracking-disabled` does not actually exist. It was in the wrong place in the original array in regards to the method comments.

includes/Core/Modules/Module.php
eugene-manuilov: Yep, agree, updated.

includes/Core/Modules/Module.php
eugene-manuilov: Yes, updated using approach defined in the previous comment.

includes/Modules/Analytics.php
eugene-manuilov: Done.

includes/Modules/Analytics.php
eugene-manuilov: Ok, removed.

includes/Core/Modules/Module.php
eugene-manuilov: In my opinion, this method doesn't have to be abstract no matter we support BC or not. The purpose of this function is deterministic, the function is expected to have no side effects and not to be affected outside changes, plus no matter what happens it should return an array in all children implementations. Thus we can extract its default beheviour (which is returning an array) and put it in the parent class - exactly what we have here.

felixarntz: That's fair, there could be modules that don't define any datapoints anyway.

includes/Core/Modules/Module.php
```diff
@@ -319,29 +320,29 @@ public function is_connected() {
...
 				$batch_results = $batch->execute();
 			} catch ( Exception $e ) {
 				// Set every result of this batch to the exception.
-				foreach ( $results as $key => $datapoint ) {
-					if ( ! is_string( $datapoint ) || ! isset( $datapoint_services[ $datapoint ] ) || $service_identifier !== $datapoint_services[ $datapoint ] ) {
-						continue;
+				foreach ( $results as $key => $definition_key ) {
+					$datapoint_service = ! empty( $datapoint_definitions[ $definition_key ] ) ? $datapoint_definitions[ $definition_key ]['service'] : null;
+					if ( is_string( $definition_key ) && $service_identifier === $datapoint_service ) {
+						$results[ $key ] = $this->exception_to_error( $e, $definition_key );
```
aaemnnosttv, line 344: The second argument to `exception_to_error` is still expected to be the datapoint name (i.e. `$dataset->datapoint`). I think the key here would be more useful but we probably shouldn't change that in this issue.

Same below on `364`.

eugene-manuilov: Fixed.

Issue #1328, opened 2020-04-01
Implement existing tag detection in core/site datastore


felixarntz:
## Feature Description

We should have a mechanism to find an existing tag within the `core/site` datastore that replaces the current `getExistingTag()` utility function.

See #1317 for some related comments.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `core/site` store should have a new `getExistingTag( tagMatchers, options )` selector.
    * `tagMatchers` should be an array of tag matching regexes.
    * `options` should support for now support a single `urls` property which should be an array of URLs to scrape content from to run the regexes against. If not provided or empty, this should default to the `homeURL` or if AMP mode is "secondary" the `homeURL` and a single AMP post URL (similar to how the current `getExistingTag()` works.
    * Overall the selector should behave the same way as the existing utility function, e.g. pass `tagverify` query parameter etc.
* The `core/site` store should probably have an action like `fetchHTML( url )` to fetch the HTML markup string for a certain URL. The `getExistingTag()` resolver could then use that.
* The `getExistingTag()` selectors in the `modules/analytics` and `modules/adsense` stores (plus any other ones if introduced by then) should be modified to use the new `core/site` selector passing their respective `tagMatchers`. The logic around using the old `getExistingTag()` utility function should be removed.

## Implementation Brief

I think the API for the `getExistingTag()` selector should be a bit higher-level, exist on the module stores—not "core/site", and be supported by a private, lower-level `getHTMLforURL()` selector (that gets HTML for a particular URL and saves it in the store for future use).

* Add a `getExistingTag()` selector to the relevant module stores (`modules/analytics`, `modules/adsense`, and others). These will use `Data.select('core/site').getHTMLforURL( url )` under-the-hood to request the site URL, and then run the HTML source against matcher functions. (For now these will match regexes against the HTML source.). Each module will rely on `getExistingTagURLs()` to create a list of URLs to request and a `tagMatchers` array of `RegExp` in each modules' `utils.js` folder to determine if an existing tag is present. (Internally, something like `extractExistingTag( html, tagMatchers )` can be used to check for existing tags and return them if found.)
* The `getExistingTag()` selector should continue to add the query params `{tagverify: 1, timestamp: Date.now()}` to the URLs it uses, as in the existing approach (https://github.com/google/site-kit-wp/blob/af32da2ebb1fda7d69508efae68d69b71e8b34b8/assets/js/util/index.js#L521)
[...]

Issue #1655, opened 2020-06-09
Replace isDoingGet* selectors with relying on resolvers


felixarntz:
## Feature Description

We currently have several selectors in our datastores specifically to check whether an API call for a selector is in progress (these are always called `isDoingGet*`). This seems somewhat unnecessary since every datastore exposes selectors like `isResolving( selectorName )`, via Gutenberg core.

We should consider getting rid of all `isDoingGet*` selectors and instead rely on these built-in `isResolving` selectors, which is more applicable. The trade-off is that we would in a few more tests need to add calls to mark a resolver as completed (specifically Jest component tests). In the more dynamic cases like Storybook this should be covered by the logic we already have in our resolvers to not run the query if there already is data, but for Jest the resolvers need to be manually marked as finished.

We could think about removing all the boilerplate code to always check for data already being there in the resolver and instead become strict about requiring calls to `finishResolution` for pretty much every `receiveGet*` call - this would be a bit more consistent, but would actually make setting up test stores a lot more tedious, unless we combined this somehow. Potentially, `createFetchStore` could expose a `receiveAndResolveGet*( response, args )`, which calls `receiveGet*` with `response` and `argsToParams( ...args )` and `finishResolution( slug, args )`? This may be overkill, but worth discussing.

See https://github.com/google/site-kit-wp/pull/1388#discussion_r436806064 for some context on where this came up.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Whenever a `receiveGet*` action associated with a selector/resolver is called manually in the codebase (mostly in tests and Storybook, although there are a few cases in production code, such as in the `modules/analytics` store), the respective selector should also be marked as completed (via built-in action `finishResolution( selectorName, args )`).
    * This applies to almost all `receiveGet*` calls within tests and Storybook, although there are a few which require slightly different changes since the resolver does more than just run that single API request. The few production code cases likely require more specific inspection since they are special cases. For example, the Analytics `getAccounts` resolver should of course not mark itself as finished after calling `receiveGetAccounts`, since there is more logic to run first, and at the end it will be marked as finished anyway.
* In parallel to the above, all our resolvers should be updated to no longer check in the beginning if the data the resolver exists for is already set. Per the above, we should rely on explicitly marking selectors as resolved instead, so all of these somewhat double checks are no longer necessary.
* All usages of `isDoingGet*` should be migrated to using the built-in `isResolving` selectors.
[...]

PR #1721, opened 2020-06-29
Enhancement/1328 implement tag detection in core site datastore

jqlee85:
## Summary

Updates `getExistingTag` implementation.

Addresses issue #1328

## Relevant technical choices

Updates `getExistingTag` implementation to not use the old `getExistingTag()` utility, and instead use selectors in `Adsense`, `Analytics` and `Tagmanager` datastores, which use new utilities (`getExistingTagURLs`, `getHTMLForURL`, and `extractExistingTag`), and a `getHTMLForURL` selector in the `core/site` datastore.

## Checklist


assets/js/modules/tagmanager/datastore/existing-tag.js
```diff
@@ -27,15 +27,29 @@ import invariant from 'invariant';
...
 
 const fetchGetExistingTagStore = createFetchStore( {
 	baseName: 'getExistingTag',
-	controlCallback: () => getExistingTag( 'tagmanager' ),
+	controlCallback: async () => {
+		const existingTagURLs = await getExistingTagURLs( Data.select( CORE_SITE ) );
+		let tagFound = null;
+		for ( const url of existingTagURLs ) {
+			const html = await Data.select( CORE_SITE ).getHTMLForURL( url );
+			// TODO: html ends up being undefined here.
```
jqlee85, line 45: @aaemnnosttv From what I can tell, my main issue with the tests arises here, as this `html` ends up being `undefined`. I think the `getHTMLForURL()` selector initially returns `undefined`, since it doesn't have a value, so then it makes the fetch request to get the html for the given url.  So I think I need to wait until the selector has done all that, but I'm not sure how to go about that here within this `controlCallback`.

Also, is there a way to do `createRegistryControl` with these `controlCallbacks`? If I try to use it for the callback I get errors, which is why I'm using `Data.select` instead of the `registry.select`.

aaemnnosttv: > From what I can tell, my main issue with the tests arises here, as this `html` ends up being `undefined`. I think the `getHTMLForURL()` selector initially returns `undefined`, since it doesn't have a value, so then it makes the fetch request to get the html for the given url.  So I think I need to wait until the selector has done all that, but I'm not sure how to go about that here within this `controlCallback`.

If the selector uses a resolver to fulfill the value then it will return whatever the current value is - in this case `undefined` makes sense. Since selectors are synchronous you can't wait for anything in a selector. However, components that use this selector will update when the datastore updates (i.e. receives the HTML for a URL) so when they select it again it will use whatever the current state is at that time.

For this reason we generally support `undefined` as an expected value which always means that the data hasn't been loaded yet.

> Also, is there a way to do `createRegistryControl` with these `controlCallbacks`? If I try to use it for the callback I get errors, which is why I'm using `Data.select` instead of the `registry.select`.

Unfortunately not. That is something which was intentional in the design. If you do need a registry control however, you don't _need_ to use `createFetchStore`. It's possible we may update this to support using a registry control in the future if we have sufficient need for it but so far we haven't.

One reason why you may be having problems with tests may be from using `Data` directly. You can't use `Data` directly, especially in a datastore as we use many different registry instances in tests, none of which are instance returned by the `googlesitekit-data` import. `Data` should only be referenced directly for things like `useSelect`  or `combineStores`. The only exception to this are entrypoint files which are only really loaded in the browser/production.

assets/js/googlesitekit/datastore/site/html.js
```diff
@@ -0,0 +1,217 @@
...
+
+		return new Promise( ( resolve ) => {
+			const unsubscribe = registry.subscribe( () => {
+				if ( isHTMLForURLLoaded() ) {
+					unsubscribe();
+					resolve();
+				}
+			} );
+		} );
+	} ),
```
aaemnnosttv, line 140: I just found out that it is possible to do this using an existing method:
```suggestion
		return registry.__experimentalResolveSelect( STORE_NAME ).getHTMLForURL( url );
```

This also has the added benefit of returning the resolved value from the selector when it completes.

E.g.

```js
const existingTag = yield actions.waitForExistingTag();
```

I'm working on an issue to cover a larger refactoring with this so let's keep it to here for now.

assets/js/modules/adsense/datastore/__factories__/html-with-tag.js
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+function tagBodyHTML( adsenseID ) {
+	return `<script data-ad-client="${ adsenseID }" async src="https://pagead2.googlesyndication.com/pagead/js/adsbygoogle.js"></script>`;
```
aaemnnosttv, line 20: Let's make sure we're using the official snippet and placement for these.

assets/js/modules/adsense/datastore/tags.js
```diff
@@ -73,20 +57,101 @@ const fetchGetTagPermissionStore = createFetchStore( {
...
+			payload: { existingTag },
+			type: RECEIVE_EXISTING_TAG,
+		};
+	},
+	waitForExistingTag() {
+		return {
+			payload: {},
+			type: WAIT_FOR_EXISTING_TAG,
+		};
+	},
```
aaemnnosttv, line 93: Let's make sure any actions with controls are always generators that yield 👍  (see `getExistingTag` above too)

package-lock.json
```diff
@@ -11412,6 +11412,12 @@
           "integrity": "sha512-GSmOT2EbHrINBf9SR7CDELwlJ8AENk3Qn7OikK4nFYAu3Ote2+JYNVvkpAEQm3/TLNEJFD/xZJjzyxg3KBWOzw==",
           "dev": true
         },
+        "prettier": {
```
aaemnnosttv, line 11415: The lock file shouldn't have changed as no dependencies were added/updated/removed. Let's not update it unless there is a reason this is needed.

assets/js/modules/adsense/datastore/tags.js
```diff
@@ -73,20 +57,101 @@ const fetchGetTagPermissionStore = createFetchStore( {
...
 	tagPermissions: {},
 };
 
+const baseActions = {
+	getExistingTag() {
+		return {
+			payload: {},
+			type: GET_EXISTING_TAG,
+		};
+	},
```
felixarntz, line 76: Let's use `fetchGetExistingTag` and `FETCH_GET_EXISTING_TAG`, sticking to our conventions to always use "fetch" and "receive" for these kinds of internal actions.

felixarntz:
@jqlee85 @aaemnnosttv Note that the abstraction should also be able to receive some (optional) function to check for validity of the found tag. See #1760 where this first came up for Analytics, but generally is just as relevant for the other Site Kit modules. Maybe this could just be an optional `isValidTag` option.

assets/js/modules/adsense/datastore/tags.js
aaemnnosttv: I thought about that, but the `fetch` is really the responsibility of the `core/site` store, the module isn't really concerned with where it comes from, it only needs HTML to parse it from. It seems inaccurate to call a `FETCH_` action that doesn't itself perform a fetch.

felixarntz: While this action itself does not perform a `fetch`, technically a `fetch` is performed though. Per our naming conventions I don't think it makes sense to have anything but a public selector start with `get...`. Our "fetch" and "receive" prefixes indicate that these are internal actions that fetch and receive data from some URL, which applies here too - if we had not abstracted this out into a `getHTMLForURL` in `core/site`, the `fetch` request would happen here.

assets/js/modules/adsense/datastore/tags.js
jqlee85: @aaemnnosttv I updated most of these, but `fetchGetExistingTag` breaks if it's made into a generator. I think it's because we're yielding it inside the resolver.

assets/js/modules/adsense/datastore/index.js
```diff
@@ -51,11 +57,17 @@ const baseModuleStore = Modules.createModuleStore( 'adsense', {
 	adminPage: 'googlesitekit-module-adsense',
 } );
 
+const existingTagStore = createExistingTagStore( 'modules', 'adsense', {
+	tagMatchers,
+	isValidTag: isValidClientID,
+} );
```
felixarntz, line 63: I wonder whether we should handle this inside of the `tags.js` file? This is where the part of the store was defined before, and also `tags.test.js` tests the `getExistingTag` selector. So I think we should move this back into the `tags.js` file, it already has a `Data.combineStores` call anyway.

assets/js/modules/analytics/datastore/index.js
```diff
@@ -45,9 +50,15 @@ const baseModuleStore = Modules.createModuleStore( 'analytics', {
 	adminPage: 'googlesitekit-module-analytics',
 } );
 
+const existingTagStore = createExistingTagStore( 'modules', 'analytics', {
+	tagMatchers,
+	isValidTag: isValidPropertyID,
+} );
+
```
felixarntz, line 57: See above, let's move this to the `tags.js` file where the similar functionality was before. It's more of an implementation detail that it uses `createExistingTagStore`, so I think we should keep that where it originally was.

assets/js/modules/tagmanager/datastore/index.js
```diff
@@ -61,12 +67,18 @@ baseModuleStore = ( ( { actions, selectors, ...store } ) => {
 	};
 } )( baseModuleStore );
 
+const existingTagStore = createExistingTagStore( 'modules', 'tagmanager', {
+	tagMatchers,
+	isValidTag: isValidContainerID,
+} );
```
felixarntz, line 73: See above.

assets/js/modules/adsense/datastore/index.js
jqlee85: @felixarntz oh yeah, good call. It makes much more sense calling the store creation in the `tags.js` files. I'll move it in there.

assets/js/googlesitekit/datastore/site/html.test.js
```diff
@@ -0,0 +1,179 @@
...
+				await subscribeUntil( registry, () => registry
+					.select( STORE_NAME )
+					.hasFinishedResolution( 'getHTMLForURL', [ url ] )
+				);
+
+				expect( fetchMock ).not.toHaveFetched();
+				expect( selectedHTML ).toEqual( html );
+			} );
+
+			it( 'dispatches an error if the request fails', async () => {
```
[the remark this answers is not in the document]
jqlee85: Yep, makes sense. I've updated the controller to return the html if it's there (otherwise null), and this test to check that the value is set to null if the request comes back with undefined for the body.

assets/js/googlesitekit/datastore/site/html.js
```diff
@@ -0,0 +1,217 @@
...
+			timestamp: Date.now(),
+		};
+		const response = await fetch( addQueryArgs( url, fetchHTMLQueryArgs ), fetchHTMLOptions );
+		if ( ! response.ok ) {
+			throw {
+				code: response.statusText,
+				message: response.statusText,
+				data: { status: response.status },
+			};
+		}
```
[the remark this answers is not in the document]
jqlee85: :+1:

assets/js/modules/tagmanager/datastore/__factories__/html-with-tag.js
```diff
@@ -17,24 +17,22 @@
...
-<!-- Google Tag Manager (noscript) -->
-<noscript><iframe src="https://www.googletagmanager.com/ns.html?id=${ containerID }"
-height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
-<!-- End Google Tag Manager (noscript) -->
-`;
+	return `<!-- Google Tag Manager (noscript) -->
+		<noscript><iframe src="https://www.googletagmanager.com/ns.html?id=${ containerID }"
+		height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
+		<!-- End Google Tag Manager (noscript) -->
+	`;
```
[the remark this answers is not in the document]
jqlee85: @aaemnnosttv no other changes, just indented to match the indentation of of the return statement. Let me know if it should remain aligned to the beginning of the line.

assets/js/modules/tagmanager/datastore/index.js
```diff
@@ -66,7 +67,7 @@ const store = Data.combineStores(
 	accounts,
 	containers,
 	error,
-	existingTag,
+	tags,
```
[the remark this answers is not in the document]
jqlee85: Ah ok, makes sense :+1:

assets/js/modules/adsense/datastore/tags.test.js
```diff
@@ -53,6 +58,27 @@ describe( 'modules/adsense tags', () => {
 	} );
 
 	describe( 'selectors', () => {
+		describe( 'getExistingTag', () => {
+			it( 'uses a resolver to get tag', async () => {
+				const expectedTag = 'ca-pub-12345678';
```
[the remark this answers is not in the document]
jqlee85: Makes sense. I've simplified this just to see if it gets the correct tag.

assets/js/util/tag.js
```diff
@@ -103,3 +110,58 @@ export const extractTag = ( string, module ) => {
...
+	if ( AMP_MODE_SECONDARY === ampMode ) {
+		const ampPostURL = await apiFetch( { path: '/wp/v2/posts?per_page=1' } )
+			.then(
+				( posts ) => posts.slice( 0, 1 ).map(
+					( post ) => addQueryArgs( post.link, { amp: 1 } )
+				).pop()
+			);
+		if ( ampPostURL ) {
+			urls.push( ampPostURL );
+		}
```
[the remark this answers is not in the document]
jqlee85: Good call, I've updated the function and added a test.

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -208,11 +207,11 @@ const baseResolvers = {
 		let matchedProperty = registry.select( STORE_NAME ).getMatchedProperty();
 		// Only fetch accounts if there are none in the store.
 		if ( existingAccounts === undefined ) {
-			yield tagActions.waitForExistingTag();
+			yield registry.dispatch( STORE_NAME ).waitForExistingTag();
 			const existingTag = registry.select( STORE_NAME ).getExistingTag();
 			let existingTagPermission;
 			if ( existingTag ) {
-				yield tagActions.waitForTagPermission( existingTag );
+				yield registry.dispatch( STORE_NAME ).waitForTagPermission( existingTag );
```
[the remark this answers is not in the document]
jqlee85: Ah, yeah, I had changed this to fix an issue with a test, but the yield doesn't make sense here. I think it should just be dispatch.

aaemnnosttv:
@googlebot I consent.

assets/js/googlesitekit/datastore/site/html.js
```diff
@@ -0,0 +1,197 @@
...
+	/**
+	 * Resets the HTML for a given URL.
+	 *
+	 * @since n.e.x.t
+	 * @private
+	 *
+	 * @param {string} url URL for which the HTML should be reset.
+	 * @return {Object} Redux-style action.
+	 */
+	*resetHTMLForURL( url ) {
```
tofumatt, line 93: Should we maybe call this "clear" or something like that? "Reset" implies an initial value of something other than `null` or `undefined` to me. Just a thought.

assets/js/googlesitekit/data/create-existing-tag-store.js
```diff
@@ -0,0 +1,189 @@
...
+				const tagFound = extractExistingTag( html, tagMatchers );
+				if ( tagFound ) {
+					return tagFound;
+				}
+			}
+
+			return null;
+		} ),
+		[ WAIT_FOR_EXISTING_TAG ]: createRegistryControl( ( registry ) => () => (
+			registry.__experimentalResolveSelect( STORE_NAME ).getExistingTag()
```
felixarntz, line 113: I'm not sure where it was decided to use this function. It's currently causing one of the existing Jest tests to fail: https://travis-ci.com/github/google/site-kit-wp/jobs/363549318
Maybe it's because we don't resolve selectors when passing data to them (related: #1655)? cc @aaemnnosttv @tofumatt 

Either way, let's revert back to how this originally worked (in the Analytics store implementation from before) - with that tests all pass, and also maybe we shouldn't be using functions marked experimental in Gutenberg.

aaemnnosttv:
@felixarntz @tofumatt I made a few changes in response to the feedback.
* `isValidTag` is now a required argument - we always provide it and there isn't much value in the previous default so this is more consistent with our use
* `storeName` is now a required argument, and the function now takes a single object of args ([as originally suggested](https://github.com/google/site-kit-wp/pull/1721#pullrequestreview-446595791)) rather than requiring `type` and `identifier` unnecessarily, and adding tests around the store name generation.

assets/js/googlesitekit/data/create-existing-tag-store.js
```diff
@@ -0,0 +1,189 @@
...
+ * @param {Object}   options              Options to consider for the store.
+ * @param {Array}    options.tagMatchers  The tag matchers used to extract tags from HTML.
+ * @param {Function} [options.isValidTag] Optional. Function to detect whether a tag is valid or not.
+ * @param {string}   [options.storeName]  Optional. Store name to use. Default is '{type}/{identifier}'.
+ * @return {Object} The existing tag store object, with additional `STORE_NAME` and
+ * INITIAL_STATE` properties.
+ */
+export const createExistingTagStore = ( type, identifier, {
+	tagMatchers,
+	isValidTag = ( tag ) => typeof tag === 'string' && tag.length > 0,
```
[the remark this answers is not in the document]
aaemnnosttv: Given the right `tagMatchers`, I think this check should always pass and this default could be sufficient. It's mostly to prevent `receiving` an invalid tag directly in tests. I wouldn't mind making it required but Felix originally mentioned it as an optional argument (https://github.com/google/site-kit-wp/pull/1721#issuecomment-657629547) so this is consistent with that. This is a private utility and we pass `isValidTag` functions for all of our uses in modules, so the default isn't really used outside of the tests, so I think it makes sense to remove.

PR #1723, opened 2020-06-30
Fixed undefined profiles issue in the ProfileNameTextField component

eugene-manuilov:
## Summary

Addresses issue #716

## Relevant technical choices

Fixed an issue which appeared when the profiles variable was undefined.

## Checklist

### DIFF
--- a/assets/js/modules/analytics/common/ProfileNameTextField.js
+++ b/assets/js/modules/analytics/common/ProfileNameTextField.js
@@ -54,21 +54,23 @@ export default function ProfileNameTextField() {
 	let helperText;
 	let trailingIcon;
 
-	const existingProfile = profiles.some( ( { name } ) => name === profileName );
-	if ( existingProfile ) {
-		helperText = (
-			<HelperText persistent>
-				{ __( 'A view with this name already exists.', 'google-site-kit' ) }
-			</HelperText>
-		);
+	if ( Array.isArray( profiles ) && profiles.length ) {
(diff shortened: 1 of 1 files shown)


assets/js/modules/analytics/common/ProfileNameTextField.js
```diff
@@ -54,21 +54,23 @@ export default function ProfileNameTextField() {
 	let helperText;
 	let trailingIcon;
 
-	const existingProfile = profiles.some( ( { name } ) => name === profileName );
```
felixarntz, line 57: Nit-picking but we could probably simplify this PR to using an inline `Array.isArray( profiles )` here.

I think even better though it would be to return include `! Array.isArray( profiles )` (or `profiles === undefined` like we do elsewhere) in the clause above that returns `null`. The only reason that `profiles` would not be an array is that the data is not loaded yet, and for this kind of check we always have been using early returns, I think that clarifies it better.

eugene-manuilov: @felixarntz, we can't return if `profiles` is not an array because we need profiles just to determine whether or not we need to display a warning message when a user enters a duplicate name. I have updated my PR to have more elegant approach to determine if it is an existing profile or not.

felixarntz: Oh you're right, I missed that. Looks good now!

PR #1727, opened 2020-06-30
Add initial prototype for advanced GA events tracking

j-ka11:
## Summary

Ported prototype code into SiteKit
Addresses issue #1728 

## Relevant technical choices

Tracks events for common WP plugins

## Checklist


includes/Modules/Analytics/Measurement_Code_Injector.php
```diff
@@ -0,0 +1,108 @@
...
+				'/wp-content/plugins/' . $main_dir_path . '/assets/js/modules/analytics/advanced-tracking/measurement-event-tracking.js',
+				false,
+				null,
+				false
+			);
+			wp_localize_script(
+				'shirshu_inject_event_tracking',
+				'eventConfigurations',
+				$this->event_configurations
+			);
```
felixarntz, line 105: The overall approach for how to inject this code here should be slightly modified:
* Instead of loading an external file, this JavaScript should be inlined. While loading an actual JS file would generally be appropriate, the script needed here will be loaded in all users' frontends and it is fairly small - like all other Analytics `gtag` code it should be inlined.
* This can be done using `wp_add_inline_script( 'google_gtagjs', $script_code )` (see the `Analytics::enqueue_gtag_js()` method for reference). Since the script here is bigger than the very basic inline scripts already existing, you could use [PHP heredoc](https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc) to write the script into a variable and then pass that to `wp_add_inline_script`.
* Instead of injecting when `! Context::is_amp()` (which is not enough to check, since it only makes sense to inject the code when _all_ conditions for loading Site Kit's `gtag` script are met), we should directly rely on whether Site Kit already enqueued `google_gtagjs`. We can do so by calling [`wp_script_is( 'google_gtagjs' )`](https://developer.wordpress.org/reference/functions/wp_script_is/). Only if that is the case, this logic should run.
    * Since we now no longer need access to `Context::is_amp()`, the class property could be removed entirely.
* Since the `Analytics::enqueue_gtag_js()` method is hooked into WordPress at the default priority 10, the hook here should use a later priority (e.g. 11). This will need to be changed in the `add_action` call above.

includes/Modules/Analytics/Measurement_Code_Injector.php
```diff
@@ -0,0 +1,108 @@
...
+	 * Injector constructor.
+	 *
+	 * @param array   $active_plugins list of supported plugins site currently has activated.
+	 * @param Context $module_context context used to check for AMP.
+	 */
+	public function __construct( $active_plugins, $module_context ) {
+		$this->active_plugins       = $active_plugins;
+		$this->context              = $module_context;
+		$this->event_factory        = Measurement_Event_Factory::get_instance();
+		$this->event_configurations = $this->build_event_configurations();
```
felixarntz, line 60: This functionality only needs to run when the event tracking code actually gets injected, otherwise it is unnecessary, so for performance should be avoided. This could therefore be moved from the constructor into the `inject_event_tracking` method. Since this is the only place where it's used, we can also get rid of the class property.

includes/Modules/Analytics/Measurement_Code_Injector.php
```diff
@@ -0,0 +1,108 @@
...
+ */
+
+namespace Google\Site_Kit\Modules\Analytics;
+
+use Google\Site_Kit\Context;
+
+/**
+ * Injects Javascript based on the current active plugins
+ *
+ * Class Injector
```
felixarntz, line 18: Nit-picking, but like other class descriptions in Site Kit codebase this should rather just be a single line which describes the class (not starting with a verb) - e.g. "Class for injecting event tracking JavaScript based on currently active plugins.". The docblocks on other new classes should be adjusted similarly.

includes/Modules/Analytics/Measurement_Code_Injector.php
```diff
@@ -0,0 +1,108 @@
+<?php
+/**
+ * Class Google\Site_Kit\Modules\Analytics\Measurement_Code_Injector
```
felixarntz, line 3: Can we move this and all other new classes (except the main ~~`Shirshu_Class`~~ `Advanced_Tracking` class) into its own namespace `Google\Site_Kit\Modules\Analytics\Advanced_Tracking`? This way it's clearly separate from already existing code in the `Google\Site_Kit\Modules\Analytics` namespace.

includes/Modules/Analytics/Measurement_Event_Factory.php
```diff
@@ -0,0 +1,75 @@
...
+	 * Gets the instance for this class
+	 *
+	 * @return Measurement_Event_Factory
+	 */
+	public static function get_instance() {
+		if ( is_null( self::$instance ) ) {
+			self::$instance = new self();
+		}
+		return self::$instance;
+	}
```
felixarntz, line 44: See other comment, the singleton pattern should be avoided.

includes/Modules/Analytics/Measurement_Events/Measurement_Event.php
```diff
@@ -0,0 +1,174 @@
...
+// phpcs:disable Generic.Files.OneObjectStructurePerFile.MultipleFound
+
+namespace Google\Site_Kit\Modules\Analytics\Measurement_Events;
+
+/**
+ * Use to instantiate MeasurementEvent objects
+ *
+ * @class Measurement_Event_Builder
+ */
+class Measurement_Event_Builder {
```
felixarntz, line 20: While the builder approach can generally be useful, I think it's a bit overkill here. I think we should simplify the code to get rid of `Measurement_Event_Builder` and instead pass the `array $config` directly to `Measurement_Event` instances.

(Also, there are two classes in this file which should not be the case, but anyway I think we should remove the builder one.)

includes/Modules/Analytics/Measurement_Events/Measurement_Event.php
```diff
@@ -0,0 +1,174 @@
...
+	 * Measurement_Event constructor.
+	 *
+	 * @param Measurement_Event_Builder $builder The measurement event builder.
+	 */
+	public function __construct( Measurement_Event_Builder $builder ) {
+		$this->plugin_name    = $builder->get_plugin_name();
+		$this->event_category = $builder->get_category();
+		$this->event_action   = $builder->get_action();
+		$this->event_selector = $builder->get_selector();
+		$this->event_on       = $builder->get_on();
```
felixarntz, line 156: See my above comment, we can directly receive the `array $config` here instead. And then we don't need all these class properties, we could just have a single `$config` class property which we assign the passed value to after validating - this is much simpler than all the properties since we only need the entire `$config` later anyway.

The validation should either use `wp_parse_args` to fill up missing values in the associative array with defaults, or it should check for anything missing and throw an appropriate exception if so. We just need to make sure the `$config` in the class property is valid.

includes/Modules/Analytics/Measurement_Events/Measurement_Event.php
```diff
@@ -0,0 +1,174 @@
...
+	 * @return array
+	 */
+	public function jsonSerialize() {
+		return array(
+			'pluginName' => $this->plugin_name,
+			'category'   => $this->event_category,
+			'action'     => $this->event_action,
+			'selector'   => $this->event_selector,
+			'on'         => $this->event_on,
+		);
```
felixarntz, line 171: Based on the above comment, this can be simplified to just return `$this->config`.

includes/Modules/Analytics/Measurement_Events/Measurement_Event_List.php
```diff
@@ -0,0 +1,44 @@
+<?php
+/**
+ * Class Google\Site_Kit\Modules\Analytics\Measurement_Events\Measurement_Event_List
+ *
+ * @package   Google\Site_Kit
+ * @copyright 2019 Google LLC
```
felixarntz, line 6: Since all this code is new, this should say 2020 throughout the PR :)

includes/Modules/Analytics/Measurement_Events/WPForms_Event_List.php
```diff
@@ -0,0 +1,35 @@
...
+		$builder = Measurement_Event::create_builder(
+			array(
+				'pluginName' => 'WPForms',
+				'category'   => 'engagement',
+				'action'     => 'form_submit',
+				'selector'   => '.wpforms-submit-container button',
+				'on'         => 'click',
+			)
+		);
+		$this->add_event( $builder->build() );
```
felixarntz, line 33: Based on the suggested removal of `Measurement_Event_Builder`, this (and other plugin `*_Event_List` classes) can be simplified to just call `$this->add_event( new Measurement_Event( $config ) )`.

includes/Modules/Analytics/Measurement_Events/Measurement_Event.php
```diff
@@ -0,0 +1,174 @@
...
+	/**
+	 * Creates a measurement event builder for the plugin.
+	 *
+	 * @param Array $plugin The measurement event configuration.
+	 *
+	 * @return Measurement_Event_Builder
+	 */
+	public static function create_builder( $plugin ) {
+		return new Measurement_Event_Builder( $plugin );
+	}
```
felixarntz, line 144: See above, this should be removed.

includes/Modules/Analytics/Plugin_Detector.php
```diff
@@ -0,0 +1,52 @@
...
+
+	/**
+	 * Determines the user's current active plugins that Shirshu_Class supports
+	 *
+	 * @return array
+	 */
+	public function get_active_plugins() {
+		$active_plugins = array();
+		foreach ( $this->supported_plugins as $key => $function_name ) {
+			if ( defined( $function_name ) || function_exists( $function_name ) ) {
```
felixarntz, line 46: This feels odd. We should probably change the format of `$supported_plugins` to be an array containing e.g. associative arrays with keys `check_name` and `check_type`. Then `check_name` would be the actual name and `check_type` would be either `function` or `constant`. For the two latter, we should probably rely on constants to not having to write the string over and over again (error-prone): We could introduce `Plugin_Detector::TYPE_CONSTANT` and `Plugin_Detector::TYPE_FUNCTION`.

Then this method here would only make the single appropriate check instead of trying both which is a bit fuzzy.

includes/Modules/Analytics/Shirshu_Class.php
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	/**
+	 * Returns a list of plugins that Shirshu_Class supports
+	 *
+	 * @return array of strings
+	 */
+	protected function get_supported_plugins() {
+		return $this->supported_plugins;
+	}
```
felixarntz, line 99: This method isn't used anywhere as far as I can see, so it should be removed.

includes/Modules/Analytics/Shirshu_Class.php
```diff
@@ -0,0 +1,109 @@
...
+			'Contact Form 7'   => 'WPCF7_PLUGIN_DIR',
+			'Formidable Forms' => 'load_formidable_forms',
+			'Ninja Forms'      => 'NF_PLUGIN_DIR',
+			'WooCommerce'      => 'WC_PLUGIN_FILE',
+			'WPForms'          => 'WPFORMS_PLUGIN_DIR',
+			'WPForms Lite'     => 'WPFORMS_PLUGIN_DIR',
+		);
+		$this->plugin_detector   = new Plugin_Detector( $this->supported_plugins );
+		$this->get_active_plugins();
+	}
```
felixarntz, line 90: Revisiting. this class, I'm starting to think we should remove it entirely. All it does is detect the active plugins via the `Plugin_Detector` class. While I'm all for separating concerns, I think this is a bit too much here, especially since the list of supported plugins is actually tied to which `Measurement_Event_List` derived classes we implement.

I'd suggest to move this list of supported plugins and usage of `Plugin_Detector` into the `Measurement_Code_Injector` class and then remove this class. With this change, `Measurement_Code_Injector` would then become the new entry point for the advanced tracking functionality.

j-ka11: I can see where you're coming from. The original idea was to have a driver class that uses the functionality of the Plugin Detector and Code Injector, since those two classes perform separate tasks. However, the flow of things got messed up when we had to incorporate wordpress hooks. Knowing that this class won't be instantiated until after the plugins are loaded, could I try rearranging some things to make it closer to our original idea? If it still seems like overkill, we can definitely remove the class entirely.

felixarntz: Sure I'm open for alternatives, however one thing that should be avoided is running this logic to detect the active supported plugins when not needed. In other words, this probably needs to become dependent on the `Measurement_Code_Injector` class in some way, since that class has awareness of whether event tracking code should be injected or not (if Site Kit did not enqueue `google_gtagjs`, there is no point in running any advanced tracking logic).

j-ka11: Okay I will keep that in mind.

includes/Modules/Analytics/Measurement_Code_Injector.php
michael0525: For the string variable that holds the injected script using PHP heredoc, should I make the variable a static attribute of the class or initialize it as a local variable inside the inject_event_tracking() function?

includes/Modules/Analytics/Advanced_Tracking.php
```diff
@@ -0,0 +1,141 @@
...
+	 * @since n.e.x.t.
+	 * @var Plugin_Detector
+	 */
+	private $plugin_detector;
+
+	/**
+	 * Advanced_Tracking constructor.
+	 *
+	 * @param Plugin_Detector $plugin_detector optional plugin detector used for testing.
+	 * @since n.e.x.t.
```
felixarntz, line 56: Per doc conventions, `@since` should go above `@param` and `@return` docs, with one blank line in between.

includes/Modules/Analytics/Advanced_Tracking/Plugin_Detector.php
```diff
@@ -0,0 +1,80 @@
...
+	 */
+	public static $TYPE_CONSTANT = 'CONSTANT';
+
+	/**
+	 * The function check_type string for support_plugins array in the Advanced_Tracking class.
+	 *
+	 * @since n.e.x.t.
+	 * @var string
+	 */
+	public static $TYPE_FUNCTION = 'FUNCTION';
```
felixarntz, line 48: Constants shouldn't start in a `$`, that can be omitted.

includes/Modules/Analytics/Advanced_Tracking.php
```diff
@@ -0,0 +1,141 @@
...
+		$this->plugin_detector   = $plugin_detector;
+	}
+
+	/**
+	 * Registers functionality through WordPress hooks.
+	 *
+	 * @since n.e.x.t.
+	 */
+	public function register() {
+		add_action( 'wp_enqueue_scripts', array( $this, 'set_up_advanced_tracking' ), 11 );
```
felixarntz, line 100: As mentioned, let's make the hook callback method private, and then instead add a closure here which calls the private method.

includes/Modules/Analytics/Advanced_Tracking.php
```diff
@@ -0,0 +1,141 @@
...
+				'check_name' => 'WPFORMS_PLUGIN_DIR',
+				'check_type' => Plugin_Detector::$TYPE_CONSTANT,
+			),
+			array(
+				'name'       => 'WPForms Lite',
+				'check_name' => 'WPFORMS_PLUGIN_DIR',
+				'check_type' => Plugin_Detector::$TYPE_CONSTANT,
+			),
+		);
+		$this->plugin_detector   = $plugin_detector;
```
felixarntz, line 91: This should, if no `Plugin_Detector` instance is provided, instantiate one, otherwise this won't work.

includes/Modules/Analytics/Advanced_Tracking.php
```diff
@@ -0,0 +1,141 @@
...
+	 * @since n.e.x.t.
+	 */
+	public function set_up_advanced_tracking() {
+		if ( ! wp_script_is( 'google_gtagjs' ) ) {
+			return;
+		}
+
+		if ( null === $this->plugin_detector ) {
+			$this->plugin_detector = new Plugin_Detector( $this->supported_plugins );
+		}
```
felixarntz, line 115: See above, let's move this to the constructor, so that the property is always correctly set.

assets/js/modules/analytics/advanced-tracking/measurement-event-tracking.js
```diff
@@ -0,0 +1,14 @@
...
+	document.addEventListener( config.on, function( e ) {
+		if ( e.target.matches( thisConfig.selector ) ) {
+			//alert( 'Got an event called: '.concat( thisConfig.action ) );
+			//record event with gtag here
+		} else if ( e.target.matches( thisConfig.selector.concat( ' *' ) ) ) {
+			//alert( 'Got an event called: '.concat( thisConfig.action ) );
+			//record event with gtag here
+		}
+	}, true );
+}
```
felixarntz, line 14: This file needs to be removed.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Code_Injector.php
```diff
@@ -0,0 +1,79 @@
...
+	document.addEventListener( config.on, function( e ) {
+		if ( e.target.matches( thisConfig.selector ) ) {
+			alert( 'Got an event called: '.concat( thisConfig.action ) );
+			//record event with gtag here
+		} else if ( e.target.matches( thisConfig.selector.concat( ' *' ) ) ) {
+			alert( 'Got an event called: '.concat( thisConfig.action ) );
+			//record event with gtag here
+		}
+	}, true );
+}
```
felixarntz, line 51: We'll need to thoroughly test this code, also for cross-browser compatibility. This code will run in any environment we cannot control, so it needs to be more "old school" unfortunately. For example:
* The entire code should be wrapped in an anonymous function that is then called: `( function() { ... } )();`; this way it doesn't affect the global namespace in any way.
* It shouldn't use any more modern (e.g. ES6) JavaScript patterns, e.g. `let` and `const` should be replaced with `var`.
* The `matches` function may not always be present, so we'll need to use a [Polyfill like e.g. AMP does](https://github.com/ampproject/amphtml/blob/feb792fcdc7d328073300d083a516c18c87c1194/src/dom.js#L479).

includes/Modules/Analytics/Advanced_Tracking/Measurement_Code_Injector.php
```diff
@@ -0,0 +1,79 @@
...
+	 * The javascript code that is injected for event tracking.
+	 *
+	 * @since n.e.x.t.
+	 * @var string
+	 */
+	private static $inject_script = <<<INJECT_SCRIPT
+let config;
+for ( config of eventConfigurations ) {
+	const thisConfig = config;
+	document.addEventListener( config.on, function( e ) {
```
felixarntz, line 42: Not sure about this, is it more performance-efficient to include an event listener for every configuration, or is it more efficient to only include a single event listener for all of them together and then check whether any of them are matched? Would be good to research/test that.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Code_Injector.php
```diff
@@ -0,0 +1,79 @@
...
+		if ( e.target.matches( thisConfig.selector ) ) {
+			alert( 'Got an event called: '.concat( thisConfig.action ) );
+			//record event with gtag here
+		} else if ( e.target.matches( thisConfig.selector.concat( ' *' ) ) ) {
+			alert( 'Got an event called: '.concat( thisConfig.action ) );
+			//record event with gtag here
```
felixarntz, line 48: I think we should now start adding the actual `gtag` calls here. `gtag` will be present in the page since the Analytics script from Site Kit is present at this point.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Code_Injector.php
```diff
@@ -0,0 +1,79 @@
...
+	 * @since n.e.x.t.
+	 */
+	public function inject_event_tracking() {
+		wp_add_inline_script( 'google_gtagjs', self::$inject_script );
+
+		wp_localize_script(
+			'google_gtagjs',
+			'eventConfigurations',
+			$this->event_configurations
+		);
```
felixarntz, line 77: Instead of using `wp_localize_script()` (which is intended for providing translations), let's put the configuration right into the script we're injecting, using e.g. `$configurations = wp_json_encode( $this->event_configurations );` and then `var config = {$configurations};` - this way that also doesn't touch the global JS namespace. We don't need to have a class property for the JS.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Event_Factory.php
```diff
@@ -0,0 +1,60 @@
...
+				$event_list = new FormidableForms_Event_List();
+				break;
+			case 'Ninja Forms':
+				$event_list = new NinjaForms_Event_List();
+				break;
+		}
+		return $event_list;
+	}
+
+}
```
felixarntz, line 60: Instead of having this class and `switch` clause, we could simplify the code path by providing the class name for a plugin's `*_Event_List` within the `$supported_plugins` configuration. This entire class wouldn't be necessary then, and also it would keep the plugins definition coupled to having a class for it, which at the moment isn't immediately obvious.

includes/Modules/Analytics/Advanced_Tracking.php
```diff
@@ -0,0 +1,141 @@
...
+				'name'       => 'WPForms',
+				'check_name' => 'WPFORMS_PLUGIN_DIR',
+				'check_type' => Plugin_Detector::$TYPE_CONSTANT,
+			),
+			array(
+				'name'       => 'WPForms Lite',
+				'check_name' => 'WPFORMS_PLUGIN_DIR',
+				'check_type' => Plugin_Detector::$TYPE_CONSTANT,
+			),
+		);
```
felixarntz, line 90: See my other comment, here we could include an `event_list_class` entry for every plugin.

includes/Modules/Analytics/Advanced_Tracking.php
```diff
@@ -0,0 +1,141 @@
...
+
+		if ( null === $this->plugin_detector ) {
+			$this->plugin_detector = new Plugin_Detector( $this->supported_plugins );
+		}
+		$active_plugins = $this->plugin_detector->determine_active_plugins();
+
+		$event_factory              = new Measurement_Event_Factory();
+		$this->event_configurations = array();
+		foreach ( $active_plugins as $plugin_name ) {
+			$measurement_event_list = $event_factory->create_measurement_event_list( $plugin_name );
```
felixarntz, line 121: See above, here we would just instantiate the class from the plugin configuration directly.

includes/Modules/Analytics/Advanced_Tracking/Plugin_Detector.php
```diff
@@ -0,0 +1,80 @@
...
+		$active_plugins = array();
+		foreach ( $this->supported_plugins as $current_plugin ) {
+			if ( ( $current_plugin['check_type'] == self::$TYPE_CONSTANT &&
+					defined( $current_plugin['check_name'] ) ) ||
+					( $current_plugin['check_type'] == self::$TYPE_FUNCTION &&
+					function_exists( $current_plugin['check_name'] ) ) ) {
+				array_push( $active_plugins, $current_plugin['name'] );
+			}
+		}
+		return $active_plugins;
```
felixarntz, line 78: Maybe this method should be modified to return the entire plugin configuration for each, not just the name. This would make usage within `Advanced_Tracking` more straightforward, and potentially it's even more expected, since the return format will be the same format like the input.

With this, we could also simplify this code here to use `array_filter()` instead of a `foreach` loop with separate variable.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Measurement_Event.php
```diff
@@ -0,0 +1,54 @@
...
+
+	/**
+	 * Measurement_Event constructor.
+	 *
+	 * @since n.e.x.t.
+	 *
+	 * @param array $config The event's configuration.
+	 */
+	public function __construct( $config ) {
+		$this->config = $config;
```
felixarntz, line 40: As mentioned before, this should do some validation of `$config` - probably best to do strict validation by throwing an exception if one of the array arguments is missing or invalid.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Measurement_Event.php
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Class Google\Site_Kit\Modules\Analytics\Advanced_Tracking\Measurement_Events\Measurement_Event
+ *
+ * @package   Google\Site_Kit
+ * @copyright 2020 Google LLC
+ * @license   https://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
+ * @link      https://sitekit.withgoogle.com
+ */
+
+// phpcs:disable Generic.Files.OneObjectStructurePerFile.MultipleFound
```
felixarntz, line 11: This should be removed.

includes/Modules/Analytics/Advanced_Tracking/Plugin_Detector.php
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * @package   Google\Site_Kit\Modules\Analytics
+ * @copyright 2020 Google LLC
+ * @license   https://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
+ * @link      https://sitekit.withgoogle.com
+ */
+
+// phpcs:disable WordPressVIPMinimum.Constants.ConstantString.NotCheckingConstantName
+// phpcs:disable WordPress.NamingConventions.ValidVariableName.PropertyNotSnakeCase
+// phpcs:disable WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
```
felixarntz, line 13: PHPCS rules shouldn't be disabled. If a specific violation is flagged which is actually a false positive, let's use `phpcs:ignore` on the individual line.

includes/Modules/Analytics/Advanced_Tracking.php
```diff
@@ -0,0 +1,141 @@
...
+ * @link      https://sitekit.withgoogle.com
+ */
+
+namespace Google\Site_Kit\Modules\Analytics;
+
+use Google\Site_Kit\Modules\Analytics\Advanced_Tracking\Plugin_Detector;
+use Google\Site_Kit\Modules\Analytics\Advanced_Tracking\Measurement_Event_Factory;
+use Google\Site_Kit\Modules\Analytics\Advanced_Tracking\Measurement_Code_Injector;
+
+// phpcs:disable WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
```
felixarntz, line 17: See above, PHPCS rules shouldn't be disabled.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Measurement_Event_List.php
```diff
@@ -0,0 +1,51 @@
...
+
+	/**
+	 * Adds a measurement event to the measurement events array.
+	 *
+	 * @since n.e.x.t.
+	 *
+	 * @param Measurement_Event $e The measurement event object.
+	 */
+	protected function add_event( Measurement_Event $e ) {
+		array_push( $this->measurement_events, $e );
```
felixarntz, line 38: Better to use a proper variable name here, not `$e`. The only place where an abbreviated variable name like this is okay is e.g. for an `Exception`.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Event_Factory.php
```diff
@@ -0,0 +1,60 @@
...
+ */
+final class Measurement_Event_Factory {
+
+	/**
+	 * Instantiates a subclass of MeasurementEventList based on the given plugin name.
+	 *
+	 * @since n.e.x.t.
+	 *
+	 * @param string $plugin_name represents the plugin name to create a list of events for.
+	 * @return Measurement_Event_List
```
felixarntz, line 35: Missing return description.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Measurement_Event.php
```diff
@@ -0,0 +1,54 @@
...
+	public function __construct( $config ) {
+		$this->config = $config;
+	}
+
+	/**
+	 * Returns an associative event containing the event attributes.
+	 *
+	 * @since n.e.x.t.
+	 *
+	 * @return array
```
felixarntz, line 48: Missing return description. Those should exist everywhere; if applicable it can be very basic, e.g. here it could just say "The configuration in JSON-serializable format." or something like that.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Measurement_Event_List.php
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	/**
+	 * Gets the measurement events array.
+	 *
+	 * @since n.e.x.t.
+	 *
+	 * @return Measurement_Event[]
```
felixarntz, line 46: See above, missing return description. Please check the rest of the codebase for those too.

tests/phpunit/includes/Modules/MockPluginDetector.php
```diff
@@ -0,0 +1,58 @@
...
+	/**
+	 * Adds the specified plugin to active plugin list if not already added.
+	 *
+	 * @param string $plugin_name plugin to be added.
+	 */
+	public function add_active_plugin( $plugin_name ) {
+		if ( in_array( $plugin_name, $this->mock_active_plugins ) ) {
+			return;
+		}
+		array_push( $this->mock_active_plugins, $plugin_name );
```
felixarntz, line 44: Instead of using `array_push()`, it's preferable to use the brackets syntax (e.g. `$this->mock_active_plugins[] = $plugin_name;`). Can you adjust that throughout the PR please?

tests/phpunit/includes/Modules/MockPluginDetector.php
```diff
@@ -0,0 +1,58 @@
...
+
+class MockPluginDetector extends Plugin_Detector {
+
+	private $mock_active_plugins;
+
+	/**
+	 * MockPluginDetector constructor.
+	 */
+	public function __construct() {
+		$this->mock_active_plugins = array();
```
felixarntz, line 23: Based on my above comments for the `Plugin_Detector`, this will require some changes if it should return the array of the entire configuration objects. Shouldn't be too complicated though, we could just modify `MockPluginDetector::determine_active_plugins()` to filter the original input by the `$this->mock_active_plugins` property values.

tests/phpunit/integration/Modules/Analytics/AdvancedTrackingTest.php
```diff
@@ -0,0 +1,129 @@
...
+		}
+
+		$this->supported_plugins    = array(
+			'Contact Form 7',
+			'Formidable Forms',
+			'Ninja Forms',
+			'WooCommerce',
+			'WPForms',
+			'WPForms Lite',
+		);
```
felixarntz, line 48: This should use a similar configuration format like the original, otherwise it's not really accurate.

tests/phpunit/integration/Modules/Analytics/AdvancedTrackingTest.php
```diff
@@ -0,0 +1,129 @@
...
+
+		if ( ! wp_script_is( 'google_gtagjs' ) ) {
+			wp_enqueue_script( // phpcs:ignore WordPress.WP.EnqueuedResourceParameters.MissingVersion
+				'google_gtagjs',
+				'https://www.googletagmanager.com/gtag/js',
+				false,
+				null,
+				false
+			);
+		}
```
felixarntz, line 39: Let's move this into the respective tests since it is a side effect. You could create a protected utility method for it in this class so that you don't have to rewrite it every time. Another benefit of moving this out of `setUp()` is that for example, it should also be tested whether the conditional to set up advanced tracking works, which is based on whether or not this script is enqueued.

includes/Modules/Analytics/Advanced_Tracking.php
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	/**
+	 * Registers functionality through WordPress hooks.
+	 *
+	 * @since n.e.x.t.
+	 */
+	public function register() {
+		add_action( 'wp_enqueue_scripts', array( $this, 'set_up_advanced_tracking' ), 11 );
+	}
```
felixarntz, line 101: Something we haven't touched on yet, but talked about earlier, is that this PR also needs to support AMP - which to some degree is actually going to be easier because it only relies on a configuration object anyway (AMP does not support any custom JavaScript). Some initial pointers:
* Have a look add the [documentation for how to track `gtag` events with AMP](https://developers.google.com/analytics/devguides/collection/amp-analytics#measure_events).
* You can hook into Site Kit's `googlesitekit_amp_gtag_opt` filter which receives the entire `amp-analytics` configuration array. You can then alter its `triggers` value with the necessary event configurations data. Make sure to use the correct variable names (`event_name`, `event_category` etc).
    * We'd probably need some sort of utility method to transform an event configuration into the format that the `amp-analytics` config object expects.

includes/Modules/Analytics/Advanced_Tracking.php
j-ka11: What about when we call the function in the test?

includes/Modules/Analytics/Advanced_Tracking.php
michael0525: Solved.

includes/Modules/Analytics/Advanced_Tracking/Plugin_Detector.php
michael0525: Solved.

includes/Modules/Analytics/Advanced_Tracking.php
michael0525: Solved.

includes/Modules/Analytics/Advanced_Tracking.php
michael0525: Solved.

includes/Modules/Analytics/Advanced_Tracking.php
felixarntz: Our PHPUnit tests are mostly integration tests, so how we've been doing it is the following:
* Call `remove_all_actions( $hook_name )` / `remove_all_filters( $hook_name )`.
* Call `register()` so that the hook callback is registered.
* Manually run the hook `do_action( $hook_name, ... )` / `apply_filters( $hook_name, ... )`.

You can see how that works by looking at some other test classes in Site Kit. We also use this approach to test the `register()` methods themselves, to ensure they add the hook callbacks as expected.

tests/phpunit/includes/Modules/MockPluginDetector.php
j-ka11: Should this also be applied in Advanced_Tracking::set_up_advanced_tracking():134? Not sure what the key would be for a single event especially since some actions have multiple event objects.

includes/Modules/Analytics/Advanced_Tracking/Plugin_Detector.php
michael0525: `phpcs: ignore` seems only to ignore the warnings. In the plugin detector, however, it is an error that we are trying to ignore, which `phpcs:ignore` doesn't seem to work in this case. The error is: Constant name, as a string, should be used along with `defined()`.

includes/Modules/Analytics/Advanced_Tracking.php
michael0525: Resolved.

tests/phpunit/includes/Modules/MockPluginDetector.php
felixarntz: You don't need to know any keys, you can use `$this->event_configurations[] = $measurement_event`.

includes/Modules/Analytics/Advanced_Tracking/Plugin_Detector.php
felixarntz: `phpcs:ignore` should ignore both errors and warnings, can you double-check where you put it? It needs to be on the same or right above the line where the error is flagged.

tests/phpunit/includes/Modules/MockPluginDetector.php
michael0525: Resolved.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Code_Injector.php
michael0525: Resolved.

includes/Modules/Analytics/Advanced_Tracking.php
```diff
@@ -0,0 +1,175 @@
...
+ * Class for Advanced Tracking.
+ *
+ * @since n.e.x.t.
+ * @access private
+ * @ignore
+ */
+final class Advanced_Tracking {
+
+	/**
+	 * List of plugins SiteKit supports for event tracking.
```
felixarntz, line 33:
```suggestion
	 * List of plugins Site Kit supports for event tracking.
```
😄

includes/Modules/Analytics/Advanced_Tracking.php
```diff
@@ -0,0 +1,175 @@
...
+	 *
+	 * @since n.e.x.t.
+	 * @var array of strings
```
felixarntz, line 36: This shouldn't be a description, just the type `array`.

includes/Modules/Analytics/Advanced_Tracking.php
```diff
@@ -0,0 +1,175 @@
...
+	 * @var Plugin_Detector
+	 */
+	private $plugin_detector;
+
+	/**
+	 * Advanced_Tracking constructor.
+	 *
+	 * @since n.e.x.t.
+	 *
+	 * @param Plugin_Detector $mock_plugin_detector optional plugin detector used for testing.
```
felixarntz, line 61: All documentation descriptions should start with a capital letter, and if a parameter is optional, the description should say what's the default (e.g. append a "Default is a new instance.").

We should also use type hinting where possible (e.g. classes and interfaces), so we can add `Plugin_Detector` here so that it must be an instance of it. And the variable should just be called `$plugin_detector`, the mock is just what we pass to it in a test.

includes/Modules/Analytics/Advanced_Tracking.php
```diff
@@ -0,0 +1,175 @@
...
+				'event_config_list' => new Woocommerce_Event_List(),
+			),
+			'WPForms'          => array(
+				'check_name'        => 'WPFORMS_PLUGIN_DIR',
+				'check_type'        => Plugin_Detector::TYPE_CONSTANT,
+				'event_config_list' => new WPForms_Event_List(),
+			),
+		);
+		if ( null === $mock_plugin_detector ) {
+			$this->plugin_detector = new Plugin_Detector( $this->supported_plugins );
```
felixarntz, line 92: This feels off here, we shouldn't be passing constructor arguments to a class that only _potentially_ is instantiated here. Since the list of supported plugins is only needed by the one method `Plugin_Detector::determine_active_plugins`, it would be cleaner to remove the class property from there and instead make this a parameter on the method. This way it works as expected on both the default instance created here as well as one passed in the constructor.

includes/Modules/Analytics/Advanced_Tracking.php
felixarntz: This now contains an instance of the class, but it would be cleaner to only pass the _name_ of the class here. Creating an instance already runs all the initialization logic in that class, which may be unnecessary here since this code runs before knowing whether we need to add an advanced tracking snippet overall.

includes/Modules/Analytics/Advanced_Tracking.php
```diff
@@ -0,0 +1,175 @@
...
+	 * @since n.e.x.t.
+	 *
+	 * @param array $gtag_amp_opt gtag config options for AMP.
+	 * @return array $gtag_amp_opt gtag config options for AMP.
+	 */
+	private function set_up_advanced_tracking_amp( $gtag_amp_opt ) {
+		$this->configure_events();
+		foreach ( $this->event_configurations as $event_config ) {
+			$gtag_amp_opt['triggers'][ $event_config->get_amp_trigger_name() ] = $event_config->to_amp_config();
+		}
```
felixarntz, line 144: It would be a bit safer to add a check above the `foreach`, to set `$gtag_amp_opt['triggers']` to an empty array in case it's not set yet.

includes/Modules/Analytics/Advanced_Tracking.php
```diff
@@ -0,0 +1,175 @@
...
+		return $gtag_amp_opt;
+	}
+
+	/**
+	 * Creates list of event configurations.
+	 *
+	 * @since n.e.x.t.
+	 */
+	private function configure_events() {
+		$active_plugins = $this->plugin_detector->determine_active_plugins();
```
felixarntz, line 154: See above comment, here would be where we passed the list of supported plugins to the method.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Measurement_Event.php
```diff
@@ -0,0 +1,146 @@
...
+			'on'         => false,
+			'metadata'   => false,
+		);
+		foreach ( $config as $key => $value ) {
+			if ( array_key_exists( $key, $valid_keys ) ) {
+				if ( $valid_keys[ $key ] ) {
+					throw new \Exception( 'Duplicate configuration parameter: ' . $key );
+				} else {
+					$valid_keys[ $key ] = true;
+				}
```
felixarntz, line 70: It's not possible to have duplicate keys in PHP arrays, so this bit can be removed. We only need to throw an exception if `! array_key_exists`.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Measurement_Event.php
```diff
@@ -0,0 +1,146 @@
...
+	 * @throws \Exception Thrown when configuration value type is invalid.
+	 */
+	private function validate_config_value_types( $config ) {
+		foreach ( $config as $key => $value ) {
+			$value_type = gettype( $value );
+			if ( 'string' !== $value_type ) {
+				throw new \Exception( 'Invalid type [' . $value_type . '] for configuration paramter: ' . $key );
+			}
+		}
+	}
```
felixarntz, line 92: This would be better combined with the other method above, to avoid an extra `foreach` loop.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Measurement_Event_Pipe.php
```diff
@@ -0,0 +1,67 @@
...
+	private static function encode_measurement_event( $measurement_event ) {
+		$config       = $measurement_event->get_config();
+		$value_arr    = array();
+		$replace_keys = array();
+		foreach ( $config as $key => &$value ) {
+			if ( strpos( $value, 'function(' ) === 0 ) {
+				$value_arr[]    = $value;
+				$value          = '%' . $key . '%';
+				$replace_keys[] = '"' . $value . '"';
+			}
```
felixarntz, line 39: Nice! It's a bit hacky, but I like it.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Measurement_Event.php
```diff
@@ -0,0 +1,146 @@
...
+		$amp_config             = array();
+		$amp_config['selector'] = $this->config['selector'];
+		$amp_config['on']       = $this->config['on'];
+
+		$vars_config                   = array();
+		$vars_config['event_name']     = $this->config['action'];
+		$vars_config['event_category'] = $this->config['category'];
+
+		$amp_config['vars'] = $vars_config;
+		return $amp_config;
```
felixarntz, line 122: Clearly we don't get the metadata here for AMP. That's something I was already expecting to not be possible since we can't run custom JS in AMP, but I wonder whether they offer an alternative mechanism which maybe at least gives us limited abilities to determine things. I wonder if you can pass another selector somehow to get a value, or some kind of template variable that will be automatically replaced or so. May be worth looking into, although it's certainly possible we _may_ not find anything suitable.

includes/Modules/Analytics/Advanced_Tracking.php
michael0525: Resolved

includes/Modules/Analytics/Advanced_Tracking.php
michael0525: Resolved.

includes/Modules/Analytics/Advanced_Tracking.php
michael0525: Resolved.

includes/Modules/Analytics/Advanced_Tracking.php
j-ka11: This response is really to the comment above, but I am unable to add a response to that comment for some reason. 

Are you suggesting that the 'event_config_list' key in this array only contain the name of the class in string form rather than the actual instance? If that is the case, I think we would need the Measurement_Event_Factory class again to actually instantiate the lists.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Measurement_Event_Pipe.php
j-ka11: Awesome! 😃

includes/Modules/Analytics/Advanced_Tracking.php
michael0525: Resolved.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Measurement_Event.php
j-ka11: We'll look into it!

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/FormidableForms_Event_List.php
```diff
@@ -0,0 +1,47 @@
...
+		$event = new Measurement_Event(
+			array(
+				'pluginName' => 'Formidable Forms',
+				'category'   => 'engagement',
+				'action'     => 'form_submit',
+				'selector'   => '.frm_fields_container .frm_button_submit',
+				'on'         => 'click',
+				'metadata'   => <<<CALLBACK
+function( params, element ) {
+	var formId = element.closest('.frm_fields_container').querySelector('input[name="form_id"]').value;
```
felixarntz, line 37: `closest` may not be available in older browser (particularly IE). Typically in Site Kit we don't support older browsers, but since this code will run on the site's frontend for _any_ visitors, we'll need to.

Let's implement `closest` function in the original snippet in `Measurement_Code_Injector`, similar to: https://github.com/ampproject/amphtml/blob/1d29dfbc5ebb1bf89a77091afaa2a712457c1db6/src/dom.js#L297

includes/Modules/Analytics/Advanced_Tracking/Measurement_Code_Injector.php
```diff
@@ -0,0 +1,122 @@
...
+		$this->event_configurations = Measurement_Event_Pipe::encode_measurement_event_list( $event_configurations );
+		$this->inject_script        = <<<INJECT_SCRIPT
+( function() {
+    var eventConfigurations = {$this->event_configurations};
+	var config;
+	for ( config of eventConfigurations ) {
+		const thisConfig = config;
+		document.addEventListener( config.on, function( e ) {
+			var el = e.target;
+			var matcher = el.matches || el.webkitMatchesSelector || el.mozMatchesSelector || el.msMatchesSelector || el.oMatchesSelector;
```
felixarntz, line 57: Let's make this a function `matches` instead, similar to https://github.com/ampproject/amphtml/blob/1d29dfbc5ebb1bf89a77091afaa2a712457c1db6/src/dom.js#L479

Then below we can just call it directly instead of `matcher.call`.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Measurement_Event.php
```diff
@@ -0,0 +1,130 @@
...
+			'selector'   => false,
+			'on'         => false,
+			'metadata'   => false,
+		);
+		foreach ( $config as $key => $value ) {
+			if ( ! array_key_exists( $key, $valid_keys ) ) {
+				throw new \Exception( 'Invalid configuration parameter: ' . $key );
+			}
+			$value_type = gettype( $value );
+			if ( 'string' !== $value_type ) {
```
felixarntz, line 72: This can be simplified to `! is_string( $value )`. We don't need to include the given type in the error, saying that the type must be a string is more helpful.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/NinjaForms_Event_List.php
```diff
@@ -0,0 +1,49 @@
...
+		$event = new Measurement_Event(
+			array(
+				'pluginName' => 'Ninja Forms',
+				'category'   => 'engagement',
+				'action'     => 'form_submit',
+				'selector'   => 'div.nf-field-container.submit-container [type="button"]',
+				'on'         => 'click',
+				'metadata'   => <<<CALLBACK
+function( params, element ) {
+	var formIdLong = element.closest('.nf-form-cont').querySelector('.nf-form-title').id;
```
felixarntz, line 37: See above comment, this should use the new `closest( element, selector )` function instead, to work in older browsers too.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/WPForms_Event_List.php
```diff
@@ -0,0 +1,47 @@
...
+		$event = new Measurement_Event(
+			array(
+				'pluginName' => 'WPForms',
+				'category'   => 'engagement',
+				'action'     => 'form_submit',
+				'selector'   => '.wpforms-submit-container button',
+				'on'         => 'click',
+				'metadata'   => <<<CALLBACK
+function( params, element ) {
+	var formId = element.closest('.wpforms-submit-container').querySelector('input[name="wpforms[id]"]').value;
```
felixarntz, line 37: Flagging something here that we should do throughout all of these `metadata` callbacks: Let's code defensively. While chaining functions like this makes the code shorter, we should include checks to ensure these DOM functions return a result. For example, if `closest` returns `null`, right now it would throw a JavaScript error. And while that right now will probably never be the case, we don't control what these other plugins do so we need to put checks pretty much anywhere - it's a good practice either way.

Let's make this change throughout. If something returns a false-y value, stop proceeding and just return `params`. If anything, we could do a `console.error` in such a case, but if the DOM structure of a plugin changes, it must not result in an actual JS error.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Woocommerce_Event_List.php
```diff
@@ -0,0 +1,282 @@
...
+			)
+		);
+		$this->add_event( $event );
+
+		$event = new Measurement_Event(
+			array(
+				'pluginName' => 'Woocommerce',
+				'category'   => 'ecommerce',
+				'action'     => 'purchase',
+				'selector'   => '', /*'.woocommerce-page form.woocommerce-checkout'*/
```
felixarntz: Let's for now put the purchase button selector here and add a `TODO` to iterate on that?

includes/Modules/Analytics/Advanced_Tracking/Metadata_Collector.php
```diff
@@ -0,0 +1,238 @@
...
+use WC_Product;
+
+/**
+ * Class for collecting various event metadata.
+ *
+ * @since n.e.x.t.
+ * @access private
+ * @ignore
+ */
+final class Metadata_Collector {
```
felixarntz, line 22: This class appears to be entirely specific to WooCommerce, so it should be named appropriately. Also, this goes beyond the scope of this PR, let's think about this more advanced metadata gathering in a follow-up PR.

includes/Modules/Analytics/Advanced_Tracking/Plugin_Detector.php
```diff
@@ -0,0 +1,58 @@
...
+	 * @return array $active_plugins The list of active plugin configurations.
+	 */
+	public function determine_active_plugins( $supported_plugins ) {
+		return array_filter(
+			$supported_plugins,
+			function( $plugin_config ) {
+				return ( self::TYPE_CONSTANT === $plugin_config['check_type'] &&
+					defined( $plugin_config['check_name'] ) ) || // phpcs:ignore WordPressVIPMinimum.Constants.ConstantString.NotCheckingConstantName
+				( self::TYPE_FUNCTION === $plugin_config['check_type'] &&
+					function_exists( $plugin_config['check_name'] ) );
```
felixarntz, line 54: Let's split this into individual if clauses for better readability. The two constant conditions can be one if clause, if yes, return `true`. Then another if clause with the two function checks. And if none of them returned `true`, return `false` at the end.

tests/phpunit/integration/Modules/Analytics/AdvancedTrackingTest.php
```diff
@@ -0,0 +1,190 @@
...
+				params['event_category'] = thisConfig.category;
+				console.log(params);
+
+				gtag( 'event', thisConfig.action, params);
+			}
+		}, true );
+	}
+	}
+)();
+INJECT_SCRIPT;
```
felixarntz, line 169: We probably don't need to replicate the entire script content here, that would be painful to maintain. Either we should somehow take the base script from the production code class, but if that doesn't really work, we don't really need to test this. The functionality of building the configuration etc is much more substantial.

tests/phpunit/integration/Modules/Analytics/AdvancedTrackingTest.php
```diff
@@ -0,0 +1,190 @@
...
+				'event_config_list' => new WPForms_Event_List(),
+			),
+		);
+		$this->mock_plugin_detector = new MockPluginDetector();
+	}
+
+	/**
+	 * Tests if the sets of events in the active plugins are included in the tracking event configurations.
+	 */
+	public function test_event_configurations() {
```
felixarntz, line 76: Let's call all test methods based on the function/method they test. Additional suffixes are okay, but we should follow the pattern `test_{method_name}[_optional_suffixes_to_briefly_describe_the_test_if_there_are_multiple_for_the_same_method]` :)

tests/phpunit/integration/Modules/Analytics/AdvancedTrackingTest.php
```diff
@@ -0,0 +1,190 @@
...
+				'check_name'        => 'WPFORMS_PLUGIN_DIR',
+				'check_type'        => Plugin_Detector::TYPE_CONSTANT,
+				'event_config_list' => new WPForms_Event_List(),
+			),
+			'WPForms Lite'     => array(
+				'check_name'        => 'WPFORMS_PLUGIN_DIR',
+				'check_type'        => Plugin_Detector::TYPE_CONSTANT,
+				'event_config_list' => new WPForms_Event_List(),
+			),
+		);
```
felixarntz, line 69: Actually, let's make the `get_supported_plugins` method I mentioned above public. This way we don't need to replicate it over here.

includes/Modules/Analytics/Advanced_Tracking/Metadata_Collector.php
j-ka11: We were planning on adding the other plugins as well, we just haven't gotten there yet.

includes/Modules/Analytics/Advanced_Tracking.php
```diff
@@ -0,0 +1,181 @@
...
+		return $gtag_amp_opt;
+	}
+
+	/**
+	 * Creates list of event configurations.
+	 *
+	 * @since n.e.x.t.
+	 */
+	private function configure_events() {
+		$active_plugins = $this->plugin_detector->determine_active_plugins( $this->supported_plugins );
```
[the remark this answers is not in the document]
michael0525: This is the response to this comment and the last comment: The issue has been resolved.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Code_Injector.php
```diff
@@ -0,0 +1,122 @@
...
+				items.push( woocommerceOrderData.items[ itemName ] );
+			}
+			params['items'] = items;
+			params['event_category'] = 'ecommerce';
+
+		    console.log(params);
+
+			gtag( 'event', 'purchase', params );
+		}
+	});
```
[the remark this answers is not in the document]
michael0525: The issue has been resolved. Because this functionality is out of the scope of this PR, we have moved the issue out of the current PR and left a TODO comment.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Measurement_Event.php
michael0525: This issue has been resolved.

includes/Modules/Analytics/Advanced_Tracking/Plugin_Detector.php
michael0525: This issue has been resolved.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Code_Injector.php
```diff
@@ -0,0 +1,96 @@
...
+				params['event_category'] = thisConfig.category;
+				console.log(params);
+
+				gtag( 'event', thisConfig.action, params );
+			}
+		}, true );
+	}
+	}
+)();
+INJECT_SCRIPT;
```
j-ka11, line 74: TODO: We need to add support for non-standard cases for when events occur. The biggest issue at the moment is certain times we want an event to occur after a page load rather than on a button click on the earlier page so that we are able to gather the necessary metadata. One example of this is the `purchase` event: the `transaction_id` for an order does not exist until an order goes all the way through, so we can't record the event simply when the place order button is clicked.

These non-selector class of problems will need to be handled in a separate PR.

tests/phpunit/integration/Modules/Analytics/AdvancedTrackingTest.php
michael0525: Since the injected script from the production code class is basically a string, It would be difficult to extract base script from it.

tests/phpunit/integration/Modules/Analytics/AdvancedTrackingTest.php
michael0525: This issue has been resolved.

includes/Modules/Analytics/Advanced_Tracking.php
j-ka11: Won't this still instantiate the supported plugin configurations even if they are not active and we don't need them? A factory pattern could produce them dynamically and on demand.

tests/phpunit/integration/Modules/Analytics/AdvancedTrackingTest.php
michael0525: This issue has been resolved.

includes/Modules/Analytics/Advanced_Tracking/Metadata_Collector.php
felixarntz: @j-ka11 Yeah, but this also goes a little bit beyond the initial scope I think, let's limit this PR to handling events based on data which is available in the DOM originally and that we can get with selector queries.

This class here is already complex, and adding support for the other plugins will make that more so, so I think we should move this part out to the other branch for a follow-up PR.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Code_Injector.php
michael0525: This issue has been resolved.

tests/phpunit/integration/Modules/Analytics/AdvancedTrackingTest.php
michael0525: I have removed the unit test.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Measurement_Event.php
felixarntz: With the PHP approach we've been talking about this should be possible.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Measurement_Event.php
```diff
@@ -0,0 +1,129 @@
...
+	 * @throws \Exception Thrown when invalid keys or value type.
+	 */
+	private function validate_config( $config ) {
+		$valid_keys = array(
+			'pluginName' => false,
+			'category'   => false,
+			'action'     => false,
+			'selector'   => false,
+			'on'         => false,
+			'metadata'   => false,
```
felixarntz, line 65: Let's remove `metadata` since we're not using it anymore.

includes/Modules/Analytics/Advanced_Tracking.php
```diff
@@ -0,0 +1,204 @@
...
+				),
+				'WooCommerce'      => array(
+					'check_name'        => 'WC_PLUGIN_FILE',
+					'check_type'        => Plugin_Detector::TYPE_CONSTANT,
+					'event_config_list' => new Woocommerce_Event_List(),
+				),
+				'WPForms'          => array(
+					'check_name'        => 'WPFORMS_PLUGIN_DIR',
+					'check_type'        => Plugin_Detector::TYPE_CONSTANT,
+					'event_config_list' => new WPForms_Event_List(),
```
felixarntz, line 198: As discussed earlier, let's provide class names for `event_config_list` instead of class instances, and then instantiate them in the `configure_events` method. This way only the classes for active plugins are instantiated.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Code_Injector.php
```diff
@@ -0,0 +1,107 @@
...
+	    el.matches ||
+	    el.webkitMatchesSelector ||
+	    el.mozMatchesSelector ||
+	    el.msMatchesSelector ||
+	    el.oMatchesSelector;
+    if (matcher) {
+        return matcher.call(el, selector);
+    }
+    return false;
+}
```
felixarntz, line 61: This shouldn't be a global function, it should go within the enclosing function block so that it's only available to the code inside of it.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Code_Injector.php
```diff
@@ -0,0 +1,107 @@
...
+		const thisConfig = config;
+		document.addEventListener( config.on, function( e ) {
+			var el = e.target;
+			if ( matches(el, thisConfig.selector) || matches(el, thisConfig.selector.concat( ' *' )) ) {
+				alert( 'Got an event called: '.concat( thisConfig.action ) );
+
+				var params = {};
+				if ( "metadata" in thisConfig && null !== thisConfig.metadata ) {
+					params = thisConfig.metadata( params, el );
+				}
```
felixarntz, line 75: Let's remove this if clause for now, since it's no longer used, and we're probably not going to have this `metadata` key in the follow-up PR either, as discussed earlier.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Measurement_Event.php
```diff
@@ -0,0 +1,129 @@
...
+	 *
+	 * @param array $config The event's configuration.
+	 * @throws \Exception Thrown when config param is undefined.
+	 */
+	public function __construct( $config ) {
+		$this->validate_config( $config );
+
+		if ( ! array_key_exists( 'metadata', $config ) ) {
+			$config['metadata'] = null;
+		}
```
felixarntz, line 45: Let's remove this.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Measurement_Event.php
```diff
@@ -0,0 +1,129 @@
...
+			'selector'   => false,
+			'on'         => false,
+			'metadata'   => false,
+		);
+		foreach ( $config as $key => $value ) {
+			if ( ! array_key_exists( $key, $valid_keys ) ) {
+				throw new \Exception( 'Invalid configuration parameter: ' . $key );
+			}
+			if ( ! is_string( $value ) ) {
+				throw new \Exception( 'Invalid type [' . gettype( $value ) . '] for configuration paramter: ' . $key );
```
felixarntz, line 72: See my previous comment around this, no need for `gettype` usage. It's more helpful to just say in the warning that this configuration parameter should be a string. It doesn't really matter what the user provided instead.

Also, a typo here in "paramter" :)

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Measurement_Event_Pipe.php
```diff
@@ -0,0 +1,67 @@
...
+		}
+		$json_config = wp_json_encode( $config );
+		$json_config = str_replace( $replace_keys, $value_arr, $json_config );
```
felixarntz, line 42: Now that we're no longer passing a JS function, we can simplify this method to simply call `wp_json_encode` on `$config`, no more extra handling necessary.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Event_List_Factory.php
```diff
@@ -0,0 +1,103 @@
...
+				$event_list = new CF7_Event_List();
+				break;
+			case 'Formidable Forms':
+				$event_list = new FormidableForms_Event_List();
+				break;
+			case 'Ninja Forms':
+				$event_list = new NinjaForms_Event_List();
+				break;
+		}
+		return $event_list;
```
felixarntz, line 101: Hmm, maybe we had a misunderstanding here, but I thought we were going to keep the definitions in the configuration in `Advanced_Tracking::get_supported_plugins()`. Moving this here (like it first was) scatters the plugin-based configuration into two parts. Let's keep our architecture configuration-based, there shouldn't be any plugin-specific code outside of the configuration array and the plugin's own `*_Event_List` class.

I think we should revert https://github.com/google/site-kit-wp/pull/1727/commits/1d76ad15cf2a39b2c22d5cbb692e058224b8c515, the code before was closer to that approach. The only thing we should do then is change the configuration to not instantiate the classes but provide the class names (rename `event_config_list` key to `event_list_class`); the classes can then simply be instantiated by `Advanced_Tracking::configure_events`.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Measurement_Event.php
```diff
@@ -0,0 +1,123 @@
...
+			'action'     => false,
+			'selector'   => false,
+			'on'         => false,
+		);
+		foreach ( $config as $key => $value ) {
+			if ( ! array_key_exists( $key, $valid_keys ) ) {
+				throw new \Exception( 'Invalid configuration parameter: ' . $key );
+			}
+			if ( ! is_string( $value ) ) {
+				throw new \Exception( 'Configuration parameter type must be string' );
```
felixarntz, line 66:
```suggestion
				throw new \Exception( 'Configuration parameter ' . $key . ' must be a string' );
```

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Measurement_Event_Pipe.php
```diff
@@ -0,0 +1,55 @@
...
+			return null;
+		}
+		$result = '[';
+		foreach ( $event_configurations as $event ) {
+			$result = $result . self::encode_measurement_event( $event ) . ',';
+		}
+		$result = substr( $result, 0, -1 );
+		$result = $result . ']';
+		return $result;
+	}
```
felixarntz, line 53: This method can now be simplified to call `wp_json_encode`, and we can remove the `encode_measurement_event` method above.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Woocommerce_Event_List.php
```diff
@@ -0,0 +1,151 @@
...
+				'action'     => 'purchase',
+				'selector'   => '.woocommerce-page form.woocommerce-checkout',
+				'on'         => 'submit',
+			)
+		);
+		$this->add_event( $event );
+
+		$event = new Measurement_Event(
+			array(
+				'pluginName' => 'Woocommerce',
```
felixarntz, line 141: Let's change all these to be "WooCommerce" (capital C), same for the PHP class name.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Measurement_Event_Pipe.php
michael0525: This issue has been resolved by removing the Measurement_Event_Pipe class.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Woocommerce_Event_List.php
michael0525: This issue has been resolved.

includes/Modules/Analytics/Advanced_Tracking/Measurement_Events/Measurement_Event.php
michael0525: This issue has been resolved.

PR #1737, opened 2020-07-03
Upgrade gulp and rework gulp tasks.

eugene-manuilov:
## Summary

Addresses issue #1356

## Relevant technical choices

Upgrade gulp and its dependencies to the latest versions and reworked gulp tasks. Tasks that we don't use anymore have been removed. Other tasks have been refactored to follow gulp 4 approach and moved to the gulpfile.js folder (which is recommended by gulp itself).

## Checklist


gulpfile.js/copy.js
```diff
@@ -0,0 +1,26 @@
+/**
```
felixarntz, line 1: This and a few other of the JS files are missing the file header including license information, which need to be present.

.eslintignore
```diff
@@ -4,3 +4,4 @@
 /tests/backstop/html_report/*
 /php-scoper/*
 /vendor/*
+/gulpfile.js/*
```
felixarntz, line 7: This file shouldn't be excluded from linting. Let's either fix the violations, or alternatively we could also disable any JS doc-related rules from applying to any of the gulp files (not much point in documenting parameters etc as thoroughly as we do in production code).

gulpfile.babel.js
```diff
@@ -1,91 +0,0 @@
...
-/**
- * Gulp task to watch for file changes and run the associated processes.
- */
-gulp.task( 'watch', () => {
-	livereload.listen( { basePath: 'dist' } );
-	gulp.watch( './assets/sass/**/*.scss', [ 'build' ] );
-	gulp.watch( './assets/svg/**/*.svg', [ 'build' ] );
-	gulp.watch( './assets/js/*.js', [ 'build' ] );
-	gulp.watch( './assets/js/modules/**/*.js', [ 'build' ] );
-} );
```
felixarntz, line 50: A few tasks like these ones are missing from the new gulp file, can you clarify why you removed these? E.g. I can't find `build` and `watch`? I saw in your PR description you removed some, but would be great if you give a bit more context on why. Did you remove them because they're not used by any NPM scripts?

eugene-manuilov: @felixarntz those tasks have been removed because we don't use it anymore. Gulp tasks like `webpack`, `build`, `watch`, `local` and `browser-sync` aren't needed anymore. It has been replaced with webpack itself and webpack-cli commands.

gulpfile.js/copy.js
eugene-manuilov: Added.

.eslintignore
eugene-manuilov: Eslint issues have been fixed.

PR #1738, opened 2020-07-03
Upgrade babel and webpack dependencies

eugene-manuilov:
## Summary

Addresses issue #1356 

## Checklist


webpack.config.js
```diff
@@ -54,11 +54,14 @@ const rules = [
...
-						corejs: 2,
-					} ], '@babel/preset-react' ],
+				options: {
+					babelrc: false,
+					configFile: false,
+					cacheDirectory: true,
+					presets: [
+						'@wordpress/default',
+						'@babel/preset-react',
+					],
```
felixarntz, line 64: Shouldn't these options also include `@babel/preset-env`? Also I see you set `babelrc` to `false`, could we rely on our existing `.babelrc` file? Feels like some duplicate configuration otherwise.

eugene-manuilov: > Shouldn't these options also include @babel/preset-env?

It comes as part of the `@wordpress/default` preset, see [here](https://github.com/WordPress/gutenberg/blob/master/packages/babel-preset-default/index.js#L36).

> Also I see you set babelrc to false, could we rely on our existing .babelrc file? Feels like some **duplicate configuration** otherwise.

Duplicate configuration is exactly what I've tried to solve by moving configuration into webpack config and turning off babelrc there.

Currently, we have two points of configuration for build commands: in [webpack config](https://github.com/google/site-kit-wp/blob/develop/webpack.config.js#L57-L62) and [babelrc file](https://github.com/google/site-kit-wp/blob/develop/.babelrc) itself. It's not critical but would be better to have only one place where we configure babel for webpack builds. That's why I turned off babelrc explicitely in webpack config and updated `.babelrc` config to contain configuration required to run jest tests only.

.babelrc
```diff
@@ -1,12 +1,9 @@
 {
 	"presets": [
-		"@babel/env",
-		"@wordpress/default",
-		"@babel/preset-react"
+		"@babel/preset-env",
+		"@wordpress/default"
```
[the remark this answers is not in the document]
eugene-manuilov: @felixarntz after turning off babelrc in webpack config, jest became the only one who uses this `.babelrc` configuration and it didn't need `@babel/preset-react` to run tests. That's why it had been removed from this configuration.

PR #1740, opened 2020-07-03
Enahncement/1356 lint packages upgrade

eugene-manuilov:
## Summary

Addresses issue #1356 

## Checklist


tests/e2e/matchers/to-have-tracking.js
```diff
@@ -20,7 +20,6 @@
  * External dependencies
  */
 import { getDefaultOptions } from 'expect-puppeteer';
-import { Page, ElementHandle } from 'puppeteer';
```
felixarntz, line 23: Why was this removed here (also in the other file)? Shouldn't we import them so that the reference in the docs below is interpreted correctly?

.eslintrc.json
```diff
@@ -11,6 +11,8 @@
     "fetchMock": false
   },
   "rules": {
+    "jsdoc/no-undefined-types": "off",
+    "jsdoc/require-param": "off",
```
felixarntz, line 15: These two rules shouldn't be turned off, instead the violations should be fixed by importing the relevant types / adding the necessary param docs.

.stylelintrc
```diff
@@ -10,6 +10,7 @@
     "at-rule-no-unknown": null,
     "declaration-property-unit-whitelist": null,
     "function-url-quotes": "always",
+    "selector-id-pattern": null,
```
felixarntz, line 13: Same here, let's not globally ignore these types of issues because they should get flagged. In this case, all problems seem to be caused by `#google_dashboard_widget` which however is generated by WordPress. So let's instead annotate these specific lines to be ignore by stylelint.

tests/e2e/matchers/to-have-tracking.js
eugene-manuilov: > Why was this removed here (also in the other file)?

Eslint blames us that these are unused vars. So, i have decided to remove it (since technically speaking - it's not needed there) instead of disabling eslint check for that line.

> Shouldn't we import them so that the reference in the docs below is interpreted correctly?

Do we maintain publicly available documentation for this? Do we really need it to be interpreted correctly?

eugene-manuilov: Done.

.stylelintrc
eugene-manuilov: Ok, updated.

.eslintrc.json
eugene-manuilov: Ok, updated.

Issue #1392, opened 2020-04-14
Migrate plugin dashboard to use Widgets API infrastructure


felixarntz:
## Feature Description

After #1300, we should start using the Widgets API. Note that with the changes below we'll break all these screens by making them empty (there won't be any widgets initially).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The existing components for dashboard screen and the dashboard-details screen should be modified (let's not refactor these components; let's change them as little as possible, as much as needed for just this change here) to use the `WidgetAreasRenderer` to render widget areas. If possible, for now let's just add this component and not replace the existing "filtered components" currently being used to render the respective area. This will allow us to temporarily have both the old and new way available (we'll never publish any release with that state, but it'll potentially help building the new widgets and ensuring they look and behave similarly). Also keep in mind this is all happening in a branch separate from `develop`. The components should render widget areas for the given context as follows:
    * For the dashboard use context `dashboard`.
    * For the dashboard-details ("dashboard for a single URL") use context `pageDashboard`.
* Modify the `core/widgets` datastore module so that it automatically registers a few core widget areas for context `dashboard` (they should registered with prioritiy 1-5 in the order below):
    * `dashboardAllTraffic`
    * `dashboardSearchFunnel`
    * `dashboardPopularity`
    * `dashboardSpeed`
    * `dashboardEarnings`
* Modify the `core/widgets` datastore module so that it automatically registers a few core widget areas for context `dashboardDetails` (they should registered with prioritiy 1-4 in the order below):
    * `pageDashboardSearchFunnel`
    * `pageDashboardAllTraffic`
    * `pageDashboardTopQueries`
    * `pageDashboardSpeed`

## Implementation Brief

* Add feature flags and use them for this code:
  * Use [webpack-feature-flags-plugin](https://github.com/davolokh/webpack-feature-flags-plugin) to set our feature flags. This gives us tree-shaking dead code elimination (so we never ship that code to production until it's enabled), but allow us to reduce our usage of long-running feature branches.
[...]

PR #1745, opened 2020-07-07
Enhancement/1392 dashboard widget api

eugene-manuilov:
## Summary

Addresses issue #1392 

## Relevant technical choices

Added dashboard and pageDashboard widget areas behind feature flags.

## Checklist


assets/js/googlesitekit/widgets/datastore/areas.js
```diff
@@ -160,6 +164,47 @@ export const reducer = ( state, { type, payload } ) => {
...
 
 export const resolvers = {
+	/**
+	 * Defines default widget areas for a given context.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {string} contextSlug Widget context to get areas for.
+	 */
+	*getWidgetAreas( contextSlug ) {
```
felixarntz, line 174: This wasn't really defined in the ACs or IB, but the default widget areas (and later the default widgets) should be defined outside of the datastore. Let's implement this almost as if we were 3P ourselves, using our own API.

I suggest we create an `assets/js/googlesitekit/widgets/register-defaults.js` file with a `registerDefaults( registry )` function that dispatches `registerWidgetArea` and `assignWidgetArea` calls for both `CONTEXT_DASHBOARD` and `CONTEXT_PAGE_DASHBOARD`. This function could then be called from `assets/js/googlesitekit/widgets/index.js` with the `Data` registry.

assets/js/googlesitekit/widgets/datastore/areas.js
```diff
@@ -201,17 +246,7 @@ export const selectors = {
...
-				return 1;
-			}
-
-			if ( areaA.priority < areaB.priority ) {
-				return -1;
-			}
-
-			return 0;
-		} );
+		} ).sort( ( areaA, areaB ) => areaA.priority - areaB.priority );
```
felixarntz, line 249: 👍

assets/js/googlesitekit/widgets/datastore/constants.js
```diff
@@ -20,3 +20,6 @@ export const WIDGET_STYLES = { BOXES: 'boxes', COMPOSITE: 'composite' };
 export const WIDGET_WIDTHS = { QUARTER: 'quarter', HALF: 'half', FULL: 'full' };
 
 export const STORE_NAME = 'core/widgets';
+
+export const CONTEXT_DASHBOARD = 'dashboard';
+export const CONTEXT_PAGE_DASHBOARD = 'pageDashboard';
```
felixarntz, line 25: See my other comment, these constants should probably go into a file outside of the datastore.

It could also be worth defining constants for all widget areas themselves and exposing them on the `Widgets` object so that 3P developers can rely on them when registering their own widgets.

assets/js/components/dashboard-details/dashboard-details-app.js
```diff
@@ -45,6 +46,7 @@ class DashboardDetailsApp extends Component {
 		return (
 			<Fragment>
 				<Header />
+				{ featureFlags.widgets.pageDashboard.enabled && <WidgetContextRenderer slug="pageDashboard" /> }
```
felixarntz, line 49: This should rather go above the `<DashboardDetailsModules />` further below, which is the component that it eventually will replace.

eugene-manuilov: Updated.

assets/js/googlesitekit/widgets/datastore/areas.js
eugene-manuilov: Okay, updated it. Is it better now? I used Widgets API itself instead of using Data registery directly. Hope it's okay.

assets/js/googlesitekit/widgets/datastore/constants.js
eugene-manuilov: Ok, updated. I have also decided to rename area slugs not to contain context prefix (`AllTraffic` instead of `dashboardAllTraffic` and `pageDashboardAllTraffic`). It let's us easier define areas and assign it to multiple contexts. Does it work for you?

felixarntz: We had discussed that a while ago, so the widget areas should keep these long names because they should be unique. I can see why that might seem odd, but they should use the fully qualified names. When registering widgets, you only specify the widget area they should be added to, so therefore the widget areas cannot just be e.g. `allTraffic` since then that widget would be added to the widget area in both contexts, which is not what you might want.

assets/js/googlesitekit/widgets/register-defaults.js
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Defines default widget areas for a given context.
+ *
+ * @since n.e.x.t
+ *
+ * @param {Object} widgetsApi Widgets API.
+ */
+export function registerDefaults( widgetsApi ) {
+	const contexts = [ CONTEXT_DASHBOARD, CONTEXT_PAGE_DASHBOARD ];
```
felixarntz, line 39: This approach of relying on the `widgetsApi` object like 3P would looks good to me 👍 

We should probably name the variable `Widgets` though, which is the name we use elsewhere for these objects that will in production end up as e.g. `global.googlesitekit.widgets`.

assets/js/googlesitekit/widgets/register-defaults.js
```diff
@@ -0,0 +1,65 @@
...
+
+	widgetsApi.registerWidgetArea( AREA_ALL_TRAFFIC, {
+		title: 'All Triffic',
```
felixarntz, line 42: Typo here. Also, all of the strings here should be run through translation functions `__`. Probably best to copy all of them from their current places in the dashboard and the page dashboard ("dashboard-details").

assets/js/googlesitekit/widgets/default-areas.js
```diff
@@ -0,0 +1,31 @@
...
+export const AREA_POPULARITY = 'Popularity';
+export const AREA_SPEED = 'Speed';
+export const AREA_EARNINGS = 'Earnings';
+
+export default {
+	AREA_ALL_TRAFFIC,
+	AREA_SEARCH_FUNNEL,
+	AREA_POPULARITY,
+	AREA_SPEED,
+	AREA_EARNINGS,
```
felixarntz, line 30: See my other comment reply, these should be fully qualified including the context-specific prefix, so there should be 10 areas. The names should use camel-case (e.g. `dashboardAllTraffic`).

assets/js/googlesitekit/widgets/index.js
```diff
@@ -65,12 +69,12 @@ const Widgets = {
...
-	 * @param {(string|Array)}  [contextSlugs]     Optional. Widget context slug(s).
+	 * @param {(string|Array)}  [widgetAreaSlugs]  Optional. Widget area slug(s).
 	 */
-	registerWidget( slug, settings, contextSlugs ) {
+	registerWidget( slug, settings, widgetAreaSlugs ) {
 		dispatch( STORE_NAME ).registerWidget( slug, settings );
-		if ( contextSlugs ) {
-			dispatch( STORE_NAME ).assignWidgetArea( slug, contextSlugs );
+		if ( widgetAreaSlugs ) {
+			Widgets.assignWidget( slug, widgetAreaSlugs );
```
felixarntz, line 77: Good catch correcting those variable names 👍

assets/js/googlesitekit/widgets/register-defaults.js
```diff
@@ -0,0 +1,65 @@
...
+
+	widgetsApi.registerWidgetArea( AREA_SPEED, {
+		title: 'Page Speed and Experience',
+		subtitle: 'How fast your home page loads, how quickly people can interact with your content, and how stable your content is.',
+	}, contexts );
+
+	widgetsApi.registerWidgetArea( AREA_EARNINGS, {
+		title: 'Earnings',
+		subtitle: 'How much your site earns.',
+	}, contexts );
```
felixarntz, line 64: See above, we'll need 10 registrations here. Note that in a few cases the title and subtitle are also different, confirming these should be different widget areas for `dashboard` and `pageDashboard`.

assets/js/googlesitekit/widgets/datastore/constants.js
eugene-manuilov: Okay. Updated.

assets/js/googlesitekit/widgets/register-defaults.js
eugene-manuilov: Updated.

assets/js/googlesitekit/widgets/register-defaults.js
eugene-manuilov: Updated and translated static copy.

assets/js/googlesitekit/widgets/default-areas.js
eugene-manuilov: Updated.

assets/js/googlesitekit/widgets/register-defaults.js
eugene-manuilov: Updated.

assets/js/googlesitekit/widgets/register-defaults.js
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * @since n.e.x.t
+ *
+ * @param {Object} widgetsAPI Widgets API.
+ */
+export function registerDefaults( widgetsAPI ) {
+	widgetsAPI.registerWidgetArea( AREA_DASHBOARD_ALL_TRAFFIC, {
+		title: __( 'All Traffic', 'google-site-kit' ),
+		subtitle: __( 'How people found your site.', 'google-site-kit' ),
+		priority: 1,
```
eugene-manuilov, line 51: @felixarntz would it be better to use priorities multiple of ten for the core widget areas? Like 10, 20, 30 and so on, instead of 1, 2, 3... In this case if 3P wants to insert a new area between 20 an 30, they can use 25th priority.

felixarntz: @eugene-manuilov I thought about this, but the idea here is that for now our core widget areas should be higher priority than anything else, at least by default, that's why I went with these low numbers.

assets/js/googlesitekit-widgets.js
```diff
@@ -29,5 +29,10 @@ if ( global.googlesitekit.widgets === undefined ) {
 	global.googlesitekit.widgets = Widgets;
 }
 
+// register default widget areas
+document.addEventListener( 'DOMContentLoaded', () => {
+	registerDefaults( Widgets );
+} );
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

webpack.feature-flags.config.js
```diff
@@ -0,0 +1,10 @@
+exports.featureFlags = {
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

PR #1753, opened 2020-07-08
Add view name field to the setup form.

eugene-manuilov:
## Summary

Addresses issue #1754 

## Checklist

### DIFF
--- a/assets/js/modules/analytics/setup/setup-form.js
+++ b/assets/js/modules/analytics/setup/setup-form.js
@@ -41,6 +41,7 @@ import {
 	ExistingTagNotice,
 	ProfileSelect,
 	PropertySelect,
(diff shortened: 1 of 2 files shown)


assets/js/modules/analytics/setup/setup-form.js
```diff
@@ -99,6 +100,10 @@ export default function SetupForm( { finishSetup } ) {
 				<ProfileSelect />
 			</div>
 
+			<div className="googlesitekit-setup-module__inputs googlesitekit-setup-module__inputs--multiline">
+				<ProfileNameTextField />
+			</div>
```
felixarntz, line 105: Since this is the only item in this `div`, we should make sure it doesn't display if not necessary.

Issue #1651, opened 2020-06-09
AMP plugin compatibility: Analytics error when AMP Analytics is configured


ernee:
## Bug Description

When using the _Analytics_ section of the [AMP plugin](https://wordpress.org/plugins/amp/):

![AMP plugin Analytics](https://user-images.githubusercontent.com/8108643/84179054-3b6c4d80-aa3a-11ea-84ea-12d703703c10.png)

Attempting to connect Analytics via Site Kit causes the following error: 

![Site Kit Analytics error with AMP Analytics](https://user-images.githubusercontent.com/8108643/84179939-8e92d000-aa3b-11ea-80e7-2186727fd545.png)

Occurs when using all modes of the plugin: _Standard_,  _Transitional_, _Reader_ 

## Steps to reproduce

1. In the AMP plugin, configure the _AMP > Analytics_  section
2. Attempt to connect _Analytics_ in Site Kit
3. Go to the _Site Kit  > Analytics_ page 
4. See the Site Kit error

## Additional Context

 - PHP Version: 7.3.16
 - OS: MacOS
 - Browser: Chrome
 - Plugin Version: 1.9.0
 - Device: MacBook Air
 - AMP plugin version: 1.5.3

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be no errors during the Analytics setup (except the potentially expected user-facing error that they don't have access to the existing tag) when the existing tag extracted from a URL snippet is not a valid property ID.
* In such a case, the existing tag should be ignored, since Site Kit cannot assume anything here.

## Implementation Brief

* Return `null` in `hasExistingTag()` if an existing tag for Analytics has an invalid PropertyID.
* Merge #1760

## QA Brief

* 

## Changelog entry

* Fix bug where placing an invalid Analytics tag through another plugin could cause the Site Kit Analytics setup UI to break.

ryanwelcher:
@ernee I was able to reproduce this using the JSON in the example, however from what I've read in the [AMP analytics documentation](https://developers.google.com/analytics/devguides/collection/amp-analytics). I think that the issue may be that the `gtag_id` in the example contains a typo and is including a trailing `>` character. 

When I remove that character, I no longer get the Site Kit error but rather am correctly seeing the user-facing error that the ID is not associated with my account.

Can you confirm that the JSON is correct, please?

ernee:
[...]

PR #1760, opened 2020-07-10
Fix existing tag detection for invalid tags.

tofumatt:
## Summary

Addresses issue #1651.

If an invalid tag is found, the datastore (and components) proceed as if no tag was found.

## Checklist


assets/js/googlesitekit/data/create-settings-store.js
```diff
@@ -278,7 +278,7 @@ export const createSettingsStore = ( type, identifier, datapoint, {
 		 * @return {Object} Redux-style action.
 		 */
 		actions[ `set${ pascalCaseSlug }` ] = ( value ) => {
-			invariant( typeof value !== 'undefined', 'value is required.' );
+			invariant( typeof value !== 'undefined', `value is required for calls to set${ pascalCaseSlug }().` );
```
tofumatt, line 281: I added this to assist in debugging; because these method names are dynamically generated, this makes it a lot easier to discover which method is throwing the error, which at first made narrowing down the `invariant` call cumbersome.

assets/js/modules/analytics/datastore/tags.js
```diff
@@ -163,7 +163,14 @@ const baseSelectors = {
...
 
-		return existingTag;
+		if ( existingTag === undefined ) {
+			return undefined;
+		}
+
+		// It's possible to have an invalid accountID saved by something like the
+		// AMP plugin (see: https://github.com/google/site-kit-wp/issues/1651).
+		// Before returning a tag we should make sure it's valid.
+		return existingTag && isValidPropertyID( existingTag.propertyID ) ? existingTag : null;
```
aaemnnosttv, line 173: This won't work as-is as `existingTag` _is_ the detected property ID here, not an object 🙂 

```suggestion
		return existingTag && isValidPropertyID( existingTag ) ? existingTag : null;
```

tofumatt: Yeah, I caught that but tried this after getting a different error. Must've accidentally pushed this, but I'm still trying to sort out one last test failure 😅

aaemnnosttv:
@googlebot I consent.

Issue #1302, opened 2020-03-27
Migrate PageSpeed Insights content area in dashboard to use Widgets API


felixarntz:
## Feature Description

This is the first out of a few Site Kit module components to migrate to use the new Widgets API. Let's use this one as a test against our own API, to see whether it works as expected. We can expect some minor changes to the Widgets API itself to become necessary as part of working on this.

Further issues to migrate the other Site Kit modules' content areas will be opened once this one has been completed. Once this issue is ready, we should be able to consider our Widgets API stable.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

Refactor the PageSpeed Insights module directory per the following:
* Add a `js/modules/pagespeed-insights/components` folder for any new components.
* Re-implement the existing `DashboardSpeed` component as a standalone widget fully relying on the datastore.
    * Use the necessary components from `window.googlesitekit.widgets.components` (via importing from `Widgets.components` during development). If you need anything here that is not exposed on that object, maybe it should be - let's evaluate on a case-by-case basis.
    * Rely on existing datastores for any information you may need (like `modules/pagespeed-insights` and `core/site`).
    * The most outer component that will actually be passed as `settings.component` during widget registration should for example be able to determine whether it needs to render the actual widget content or the blue "Set up PageSpeed Insights" CTA via the datastore (this logic to differentiate is currently "hard-coded" in `js/modules/pagespeed-insights/index.js`). Again, if there is a selector missing here, let's discuss adding it to a proper place.
* The existing/old components here should be deleted, and the filters in the module's `index.js` files should be replaced with a call to `Widgets.registerWidget( 'pagespeedInsightsWebVitals', settings )`, and it should be assigned to render in the widget area `dashboardSpeed` and `pageDashboardSpeed`.

## Implementation Brief

As mentioned in #1392, this work should be done against `develop` with the use of feature flags.

* To enable access to WordPress user capabilities (needed by widget, and later by several other things too):
    * Modify `Permissions` class to add a filter to `googlesitekit_user_data`, which should amend the object passed to JS with a `permissions` object, that includes all primitive Site Kit capabilities as key and the result of `current_user_can( $cap )` as value.
    * Add selector `getCapabilities()` to `core/user` store, with a resolver that populates the initial store state with permissions from `global._googlesitekitUserData.permissions` and just returns that object (or `undefined` if not yet loaded).
[...]

Issue #1678, opened 2020-06-16
Add grid logic and additional styles for widgets and widget areas


felixarntz:
## Feature Description

While #1300 includes minimal styling for widget areas and widgets, there's quite a bit more that needs to be done, specifically to make the dynamic grid work as intended.

Every widget is registered with a width of `full`, `half` or `quarter`, and they should be organized within their widget area in a grid that accounts for this registration argument, while also optimizing for the available space as much as possible.

The other registration argument that matters for styling is each widget area's `style`.

For a better idea, this issue is about ensuring these two types of widget areas render correctly, similar to how they do today (while today all of it is hard-coded):

`composite` style widget area example (all widgets are within one box):
![Screenshot 2020-06-15 at 17 49 32](https://user-images.githubusercontent.com/3531426/84719731-f01dd780-af30-11ea-9492-e91a1110afa0.png)

`boxes` style widget area example (every widget is its own box):
![Screenshot 2020-06-15 at 17 49 54](https://user-images.githubusercontent.com/3531426/84719738-f4e28b80-af30-11ea-996a-0bb60848f878.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

### Overall widget area grid and styling
* The widget area should render as a grid (`mdc-layout-grid`).
* The widget area header should render as a grid cell that spans the full width (12 columns).
* If the widget area is of style `boxes`, each widget (per the `Widget` component) should receive white background and box-shadow as currently (e.g. Popularity widget area), as well as the current 24px padding (similar to `mdc-layout-grid` class). Every widget should appear as a cell within the widget area's main grid (see first bullet point).
* If the widget area is of style `composite`, the widgets wrapper element should receive white background and box-shadow as currently (e.g. Search Funnel widget area). In this case the widgets wrapper itself should render as yet another grid (`mdc-layout-grid`), giving it the same 24px padding from that, but also setting things up for inner widgets to appears as cells within that grid.

### Widget size and alignment
For context, on desktop the grid has 12 columns, on tablet it has 8 columns, on mobile it has 4 columns.

* Widget output should be inspected first: If a widget returns `null`, it should be skipped.
* Regardless of the widget area style, widgets should receive grid cell classes (via `WidgetRenderer` component) based on their registered `width` and how much space there is available):
    * For `full` width, the widget should span the full width on all viewports.
    * For `half` width, the widget should span half width on desktop and full width on tablet and smaller (since essentially the tablet is the half width).
[...]

PR #1766, opened 2020-07-13
Feature/1302 pagespeed insights widgets

eugene-manuilov:
## Summary

Addresses issue #1302

## Relevant technical choices

* `Widget` component has been updated to properly account for widget's with settings.
* `WidgetRenderer` component has been updated to pass `width` and `slug` params to the direct child (either widget component itself if wrapWidget is `false` or `Widget` wrapper otherwise).

## Checklist


assets/js/googlesitekit/datastore/user/permissions.js
```diff
@@ -106,6 +107,35 @@ export const selectors = {
...
+	 *
+	 * @param {Object} state Data store's state.
+	 * @param {string} capability Capability name to check.
+	 * @return {(boolean|undefined)} TRUE if the current user has this capability, otherwise FALSE. If capabilities ain't loaded yet, returns undefined.
+	 */
+	hasCapability( state, capability ) {
+		const { capabilities } = state;
+		if ( capabilities ) {
+			return !! capabilities[ capability ];
+		}
```
tofumatt, line 137: The docs say this returns `undefined` if `capabilities` aren't yet loaded, but there's no `return` statement after this `if` statement. We should explicitly return `undefined` rather than relying on the implicit `undefined` result of a function with no return value to improve readability.

assets/js/googlesitekit/datastore/user/permissions.js
```diff
@@ -27,6 +27,7 @@ const SET_PERMISSION_SCOPE_ERROR = 'SET_PERMISSION_SCOPE_ERROR';
 
 export const INITIAL_STATE = {
 	permissionError: null,
+	capabilities: global._googlesitekitUserData?.permissions || {},
```
tofumatt, line 30: Because this defaults to `{}` if `global._googlesitekitUserData?.permissions` is `false`-y, the checks below around `state.capabilities` always result in the `!! capabilities === true` path. This shouldn't have a default value of `{}` if there's the possibility that `global._googlesitekitUserData?.permissions` can be `undefined`/`false`/`null` legitimately.

assets/js/googlesitekit/widgets/components/Widget.js
```diff
@@ -22,21 +22,34 @@
...
 import { node, string } from 'prop-types';
 
-const Widget = ( { children, slug } ) => {
+/**
+ * Internal dependencies
+ */
+import { WIDGET_WIDTHS } from '../datastore/constants';
+
+function Widget( { children, slug, width } ) {
+	const classes = classnames( `googlesitekit-widget googlesitekit-widget--${ slug }`, {
```
tofumatt, line 31: Why move this up here? It's just that we tend to place `classnames` calls directly in the component's `className` attribute elsewhere, as it helps directly connect the classnames to the component they belong to. In this instance it's straightforward, but elsewhere doing this would make it more difficult to connect the classnames to a component.

assets/js/googlesitekit/widgets/components/WidgetRenderer.js
```diff
@@ -38,12 +38,17 @@ const WidgetRenderer = ( { slug } ) => {
...
-		: widgetComponent;
+	if ( ! wrapWidget ) {
+		return <Component slug={ slug } width={ width } />;
+	}
+
+	return (
+		<Widget slug={ slug } width={ width }>
+			<Component />
+		</Widget>
+	);
```
tofumatt, line 51: That's nicer to read, thanks 👍

assets/js/modules/pagespeed-insights/components/DashboardPageSpeedWidget.js
```diff
@@ -0,0 +1,63 @@
...
+	if ( ! pagespeedInsightsModule ) {
+		return null;
+	}
+
+	const { active, setupComplete } = pagespeedInsightsModule;
+	if ( ! active && ! setupComplete ) {
+		return <DashboardPageSpeedCTA />;
+	}
+
+	const { Widget } = global.googlesitekit.widgets.components;
```
tofumatt, line 44: We should import these with normal imports in our own code rather than using the global. (eg `import Widget from '../$WIDGET_PATH`;`

I just realised the IB mentions using the global directly, but we should nix that and import this component directly in our own code. Not sure why the IB said to do it this way… sorry about that.

assets/js/modules/pagespeed-insights/components/DashboardPageSpeedWidget.js
```diff
@@ -0,0 +1,63 @@
...
+		<Widget slug={ slug } width={ width }>
+			<DashboardPageSpeed />
+		</Widget>
+	);
+}
+
+DashboardPageSpeedWidget.propTypes = {
+	width: string,
+	slug: string,
+};
```
tofumatt, line 56:
```suggestion
DashboardPageSpeedWidget.propTypes = {
	width: string.isRequired,
	slug: string.isRequired,
};
```

assets/js/modules/pagespeed-insights/components/DashboardPageSpeedWidget.js
```diff
@@ -0,0 +1,63 @@
...
+
+DashboardPageSpeedWidget.defaultProps = {
+	width: '',
+	slug: '',
+};
```
tofumatt, line 61:
```suggestion
```
These don't seem like appropriate defaults. I'd remove them and set the props to be required, as neither should be blank.

assets/js/modules/pagespeed-insights/index.js
```diff
@@ -38,24 +41,10 @@ addFilter(
...
-	// Add to dashboard-details view.
-	addFilter(
-		'googlesitekit.DashboardDetailsModule',
-		'googlesitekit.PageSpeedInsights',
-		createAddToFilter( <DashboardSpeed /> ),
-		45
-	);
-}
+domReady( () => {
+	global.googlesitekit.widgets.registerWidget( 'pagespeedInsightsWebVitals', {
```
tofumatt, line 45: Let's use imports to reference these rather than globals as well, to make future maintenance easier. The IB is a bit ambiguous about global usage for this one, but we can use imports to get at this function instead and for our own code I think that's a better approach.

assets/js/googlesitekit/datastore/user/permissions.js
eugene-manuilov: Updated.

assets/js/googlesitekit/datastore/user/permissions.js
eugene-manuilov: Updated.

assets/js/googlesitekit/widgets/components/Widget.js
eugene-manuilov: Updated. Moved directly to the className attribute.

assets/js/modules/pagespeed-insights/components/DashboardPageSpeedWidget.js
eugene-manuilov: Updated.

assets/js/modules/pagespeed-insights/components/DashboardPageSpeedWidget.js
eugene-manuilov: Updated.

assets/js/modules/pagespeed-insights/components/DashboardPageSpeedWidget.js
eugene-manuilov: Updated.

assets/js/modules/pagespeed-insights/index.js
eugene-manuilov: Updated.

assets/js/modules/pagespeed-insights/components/DashboardPageSpeedCTA.js
```diff
@@ -6,23 +24,28 @@ import { __ } from '@wordpress/i18n';
...
+	const canManageOptions = useSelect( ( select ) => select( USER_STORE ).hasCapability( 'googlesitekit_manage_options' ) );
+
+	if ( ! pagespeedInsightsModule ) {
+		return null;
+	}
 
-	const { canManageOptions } = global._googlesitekitLegacyData.permissions;
+	const { active, setupComplete } = pagespeedInsightsModule;
 
 	if ( ! canManageOptions && ! setupComplete ) {
```
felixarntz, line 50: Shouldn't this be changed to `! canManageOptions || active && setupComplete`? If the user cannot manage options, we shouldn't show this, but if the module is already active and set up, we shouldn't show this either.

assets/js/modules/pagespeed-insights/index.js
```diff
@@ -20,14 +20,18 @@
...
 import { addFilter } from '@wordpress/hooks';
+import domReady from '@wordpress/dom-ready';
 
 /**
  * Internal dependencies
  */
 import './datastore';
-import { getModulesData, fillFilterWithComponent, createAddToFilter } from '../../util';
+import { fillFilterWithComponent } from '../../util';
+import Widgets from '../../googlesitekit/widgets';
```
felixarntz, line 30: This should _always_ be imported from `googlesitekit-wdgets` instead (similar to how we import `Data` from `googlesitekit-data`).

assets/js/googlesitekit/datastore/user/permissions.js
```diff
@@ -27,6 +27,7 @@ const SET_PERMISSION_SCOPE_ERROR = 'SET_PERMISSION_SCOPE_ERROR';
 
 export const INITIAL_STATE = {
 	permissionError: null,
+	capabilities: global._googlesitekitUserData?.permissions,
```
felixarntz, line 30: This works, but maybe we should use a resolver like for other data we take from globals? Would be a bit more decoupled.

assets/js/modules/pagespeed-insights/components/DashboardPageSpeedCTA.js
```diff
@@ -6,23 +24,28 @@ import { __ } from '@wordpress/i18n';
...
+const { useSelect } = Data;
 
-const PageSpeedInsightsCTA = () => {
-	const {
-		active,
-		setupComplete,
-	} = getModulesData()[ 'pagespeed-insights' ];
+function DashboardPageSpeedCTA() {
+	const pagespeedInsightsModule = useSelect( ( select ) => select( MODULES_STORE ).getModule( 'pagespeed-insights' ) );
+	const canManageOptions = useSelect( ( select ) => select( USER_STORE ).hasCapability( 'googlesitekit_manage_options' ) );
```
felixarntz, line 42: We should probably add the available Site Kit capabilities as constants to the `core/user` store so that they can be referenced. While this duplicates that list between PHP and JS, it's still cleaner than accessing via hard-coded string in components.

assets/js/googlesitekit/widgets/components/Widget.js
```diff
@@ -22,21 +22,32 @@
...
 
 Widget.defaultProps = {
 	children: undefined,
+	width: WIDGET_WIDTHS.QUARTER,
 };
 
 Widget.propTypes = {
 	children: node,
 	slug: string.isRequired,
+	width: string,
```
felixarntz, line 50: These sizes should be handled as part of #1678, `width` shouldn't be passed to the `Widget` component.

assets/js/modules/pagespeed-insights/index.legacy.js
```diff
@@ -37,7 +37,7 @@ if ( ! active || ! setupComplete ) {
 	addFilter(
 		'googlesitekit.DashboardModule',
 		'googlesitekit.PageSpeedInsights',
-		createAddToFilter( <PageSpeedInsightsCTA /> ),
+		createAddToFilter( <DashboardPageSpeedCTA /> ),
 		45
 	);
```
felixarntz, line 42: Actually, now that this component is refactored, let's move this filter to the regular `index.js` too, and then remove `index.legacy.js`.

assets/js/googlesitekit/datastore/user/permissions.js
eugene-manuilov: @felixarntz that's what i did at the beginning, but it turned out that resolver didn't work well in this case, so i decided to simplify it and put it straight into initial state.

assets/js/modules/pagespeed-insights/index.js
eugene-manuilov: Updated.

assets/js/modules/pagespeed-insights/components/DashboardPageSpeedCTA.js
eugene-manuilov: Ok, updated. However I have had to rename `PERMISSION_SCOPE_ERROR_CODE` into `ERROR_MISSING_REQUIRED_SCOPE` to escape confusions between error codes and permission names.

assets/js/googlesitekit/widgets/components/Widget.js
eugene-manuilov: Ok, widget width changes have been reverted.

assets/js/modules/pagespeed-insights/components/DashboardPageSpeedCTA.js
eugene-manuilov: @felixarntz i am not sure why we need to add `active` check here. That's how it works in the master right now:

https://github.com/google/site-kit-wp/blob/7f88b98c775446d2dcda7aafa8acde6d9a70bb23/assets/js/modules/pagespeed-insights/components/dashboard/PageSpeedInsightsCTA.js#L27

assets/js/modules/pagespeed-insights/index.legacy.js
eugene-manuilov: Done.

eugene-manuilov:
> Let's reorganize the PSI component files a little bit: All Dashboard* components should go into the pagespeed-insights/components/dashboard directory, and all other components that are currently in pagespeed-insights/components should go into pagespeed-insights/components/common.

Done.

> The pagespeed-insights/components/dashboard/util.js file should be moved to pagespeed-insights/util/index.js.

Done.

> One thing I just encountered now (man_facepalming ) which is wrong in the ACs: We cannot remove the old widget yet - that'll only be possible once we've migrated everything to the new widgets API. So let's bring back the original DashboardSpeed component, and name it LegacyDashboardSpeed. And then we'll also need to bring back the two filter hooks for the legacy component (let's annotate all these filters in index.js with TODO comments to remove once all widgets have been migrated.

Done.

assets/js/googlesitekit/datastore/user/permissions.js
eugene-manuilov: Ok, my bad. It should be good now. Reworked it to use a new resolver.

assets/js/modules/pagespeed-insights/components/DashboardPageSpeedCTA.js
felixarntz: @eugene-manuilov That's somewhat incorrect that it's there in `master`. It's also irrelevant though, since that component is only ever rendered if `! active || ! setupComplete`, which covers with what I suggested above. So we can:
* Either just use `! canManageOptions`, relying on this component only being used when `! active || ! setupComplete`.
* Or use `! canManageOptions || active && setupComplete`, which ensures the component itself never displays the CTA when the CTA would be irrelevant.

assets/js/googlesitekit/datastore/user/permissions.js
```diff
@@ -62,6 +71,21 @@ export const actions = {
...
+
+	/**
+	 * Sets user capabilities.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} capabilities User capabilities.
+	 * @return {Object} Redux-style action.
+	 */
+	setCapabilities( capabilities ) {
```
felixarntz, line 83: This should be called `receiveCapabilities` per our naming conventions, and marked as `@private`. We should only use `set*` for actions which we intend to have publicly used to set something, but this one here only exists to set data that comes from the server / API, it should never be called directly, except in our tests possibly.

For those types of actions that "receive" data from the server, let's always use `receive*` (and also `RECEIVE_*` for reducer etc).

assets/js/googlesitekit/datastore/user/permissions.js
```diff
@@ -84,13 +108,31 @@ export const reducer = ( state, { type, payload } ) => {
...
 
-export const resolvers = {};
+export const resolvers = {
+	/**
+	 * Fullfills user capabilities on initial request.
+	 *
+	 * @since n.e.x.t
+	 */
+	*getCapabilities() {
+		yield actions.setCapabilities( global._googlesitekitUserData?.permissions );
```
felixarntz, line 133: Let's be more strict here, if this object is not set, we should `global.console.error( 'Could not load core/user permissions.' )` and return early.

assets/js/googlesitekit/widgets/components/WidgetRenderer.js
```diff
@@ -39,11 +39,16 @@ const WidgetRenderer = ( { slug } ) => {
 
 	// Capitalize the "component" variable, as it is required by JSX.
 	const { component: Component, wrapWidget } = widget;
-	const widgetComponent = <Component slug={ slug } />;
 
-	return widgetComponent && wrapWidget
-		? <Widget slug={ slug }>{ widgetComponent }</Widget>
-		: widgetComponent;
+	if ( ! wrapWidget ) {
+		return <Component slug={ slug } />;
```
felixarntz, line 44: Oh I see we were originally passing `slug`, but that actually shouldn't be the case. Let's remove the prop here too and just render the component without props.

assets/js/googlesitekit/widgets/index.js
```diff
@@ -43,6 +43,27 @@ const Widgets = {
...
+	 * @since n.e.x.t
+	 */
+	contexts,
+
+	/**
+	 * Default areas for Site Kit widgets.
+	 *
+	 * @since n.e.x.t
+	 */
+	areas,
```
felixarntz, line 58: Let's not expose these defaults on the global object, we can import them directly where needed. Other plugins can register their own, and those wouldn't be available here, so we shouldn't set that precedence for now.

assets/js/googlesitekit/widgets/index.js
```diff
@@ -43,6 +43,27 @@ const Widgets = {
...
+
+	/**
+	 * Supported widths for Site Kit widgets.
+	 *
+	 * @since n.e.x.t
+	 */
+	widths: WIDGET_WIDTHS,
```
felixarntz, line 65: Let's expose this one at the same capitalized name `WIDGET_WIDTHS`. Also, while we're at it, we should expose `WIDGET_STYLES` here as well. Contrary to the default contexts and areas, these are constant lists which Site Kit fully controls, so it's more appropriate to expose here IMO.

assets/js/modules/pagespeed-insights/components/dashboard/DashboardPageSpeedCTA.js
```diff
@@ -6,23 +24,28 @@ import { __ } from '@wordpress/i18n';
...
+	const canManageOptions = useSelect( ( select ) => select( USER_STORE ).hasCapability( PERMISSION_MANAGE_OPTIONS ) );
+
+	if ( ! pagespeedInsightsModule ) {
+		return null;
+	}
 
-	const { canManageOptions } = global._googlesitekitLegacyData.permissions;
+	const { active, setupComplete } = pagespeedInsightsModule;
 
 	if ( ! canManageOptions && ! setupComplete ) {
```
felixarntz, line 50: See my previous comment, let's change this in one of the two ways. Either the component should fully take care of itself, or we should be able to assume it's only used if `! active || ! setupComplete` anyway.

assets/js/modules/pagespeed-insights/index.js
```diff
@@ -43,19 +48,40 @@ const {
...
 	addFilter(
 		'googlesitekit.DashboardDetailsModule',
 		'googlesitekit.PageSpeedInsights',
-		createAddToFilter( <DashboardSpeed /> ),
+		createAddToFilter( <LegacyDashboardSpeed /> ),
+		45
+	);
+}
+
+if ( ! active || ! setupComplete ) {
```
felixarntz, line 70: Nit-picking, but this could as well be in an `else` for the above.

assets/js/googlesitekit/datastore/user/permissions.js
eugene-manuilov: Updated.

assets/js/googlesitekit/datastore/user/permissions.js
eugene-manuilov: Updated.

assets/js/googlesitekit/widgets/components/WidgetRenderer.js
eugene-manuilov: Updated.

assets/js/googlesitekit/widgets/index.js
eugene-manuilov: Updated.

assets/js/googlesitekit/widgets/index.js
eugene-manuilov: Updated.

assets/js/modules/pagespeed-insights/components/DashboardPageSpeedCTA.js
eugene-manuilov: Ok, updated it to be `! canManageOptions || ( active && setupComplete )`.

assets/js/modules/pagespeed-insights/components/dashboard/DashboardPageSpeedCTA.js
eugene-manuilov: Ok, updated it to be `! canManageOptions || ( active && setupComplete )`.

assets/js/modules/pagespeed-insights/index.js
eugene-manuilov: Updated.

PR #1783, opened 2020-07-15
Improve stability of E2E tests

aaemnnosttv:
## Summary

Addresses issue #1821 

## Checklist


tests/e2e/specs/admin-tracking.test.js
```diff
@@ -34,8 +34,11 @@ import {
 
 async function toggleOptIn() {
 	await page.waitForSelector( '#googlesitekit-opt-in' );
-	await expect( page ).toClick( '#googlesitekit-opt-in' );
-	await page.waitForResponse( ( res ) => res.url().match( 'wp/v2/users/me' ) );
+	await pageWait();
```
felixarntz, line 37: Hmm, this feels hacky. Maybe good enough if it makes the test more stable, but why is timing even an aspect here, since below it should wait for these two things anyway?

tests/e2e/specs/modules/analytics/write-scope-requests.test.js
```diff
@@ -179,11 +179,11 @@ describe( 'Analytics write scope requests', () => {
 
 		// Click on confirm changes button and wait for permissions modal dialog.
 		await expect( page ).toClick( '.mdc-button', { text: /configure analytics/i } );
-		await page.waitForSelector( '.mdc-dialog__container' );
+		await page.waitForSelector( '.mdc-dialog--open' );
 
 		// Click on proceed button and wait for oauth request.
 		await Promise.all( [
-			expect( page ).toClick( '.mdc-dialog__actions .mdc-button', { text: /proceed/i } ),
+			expect( page ).toClick( '.mdc-dialog--open .mdc-button', { text: /proceed/i } ),
```
felixarntz, line 186: What's the reason to change the selector here? I'd imagine it would be consistently failing if it's wrong, how is `.mdc-dialog-open` more reliable?

tests/e2e/specs/admin-tracking.test.js
aaemnnosttv: It is a bit hacky and I'm not particularly happy about it either but my guess is that there is some race condition that is happening when attempting to click the element as soon as it appears. Perhaps some odd implementation detail of how our `Checkbox` component uses [`MDCCheckbox`](https://github.com/material-components/material-components-web/tree/v2.3.0/packages/mdc-checkbox) internally.  This one was particularly difficult to debug as even running the same tests on travis with `DEBUG_*` variables set caused it to consistently pass with no other changes. Let's run the tests a few times to make sure this doesn't come up again - it wouldn't be the first case where adding a wait like this fixed a failing test.

tests/e2e/specs/modules/analytics/write-scope-requests.test.js
aaemnnosttv: I would expect it to fail consistently as well but that's the problem we're trying to fix here 😄 

This should be the selector we wait for anyways according to the docs for the `mdc-dialog--open` class:

> Indicates that the dialog is open and visible.

I'm guessing this instability may be related to a race condition related to instantiation/transitions in the underlying [`MDCDialog`](https://github.com/material-components/material-components-web/blob/v2.3.0/packages/mdc-dialog/README.md#javascript-instantiation)

Issue #1326, opened 2020-04-01
Use webpack for all asset management


tofumatt:
## Feature Description

We use Webpack for the majority of our asset management (JS and SCSS), but there are still several `gulp` tasks for things like `imagemin` and SVG assets.

Having both `gulp` and `webpack` is redundant, and we should centralise our asset pipeline to use only webpack.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* `gulp` and related packages should be removed from the dependencies in `package.json`
* `gulp` tasks should be removed from the repo
* all asset processing should be done by `webpack`
* release task should be removed and use webpack/non-gulp tasks

## Implementation Brief

* Convert all tasks in https://github.com/google/site-kit-wp/blob/develop/gulpfile.babel.js/https://github.com/google/site-kit-wp/blob/develop/gulp-tasks/webpack.js to use webpack. This should have no/little impact on the scripts we run, but they shouldn't rely on `gulp` anymore.
* Remove all `gulp` tasks and dependencies
* The `release` task looks like the most involved one/the task without an obvious Webpack replacement (Imagemin and SVG manipulation should be straightforward): https://github.com/google/site-kit-wp/blob/bc8e51999da45896969e6a88a901f33fc4e9a478/gulp-tasks/zip.js, but converting that to a node+git library script should be straightforward.

Note that all scripts and build processes should be _cross-platform_ and should not be using shell scripts. See discussion in #1784.

It's unclear if changing from `gulp` to `webpack` will necessitate importing images/SVGs from JS files directly rather than referencing their direct URLs, but I suspect it will. So things like https://github.com/google/site-kit-wp/blob/develop/assets/js/util/svg-icon.js#L26 might need to be changed, but seemingly that's one of only a few places we reference SVGs...

## QA Brief

* This is a build/tooling-related change so we mainly need to make sure all images (not svg) continue to render correctly and everything should generally still work as before

## Changelog entry

* N/A

felixarntz:
I'd say while this makes sense, it should be quite low priority - there doesn't seem to be anything which this heavily affects.

aaemnnosttv:
IB ✅ 

We should consider using [`npm-run-all`](https://www.npmjs.com/package/npm-run-all) for better composite command composition as part of this too.

tofumatt:
[...]

PR #1784, opened 2020-07-15
Enhancement/1326 gulp to webpack

eugene-manuilov:
## Summary

Addresses issue #1326 

## Relevant technical choices

* This PR pins @wordpress/data package version to just `4.12.0` (instead of `^4.12.0`) to prevent unexpected upgrade to the latest version.
* It turns out that we don't need to mess with webpack config to process images and svg files, it can be done using regular cli commands provided by appropriate packages.
* Release command has been rewritten to be a bash script because it's mainly used in the GitHub Actions where we use Linux. It lets us to get rid of more dependencies and not to deal with self-written nodejs scripts.

## Checklist


eugene-manuilov:
@felixarntz @aaemnnosttv I need your opinion for this before I jump into revamping everything here. I worked on it in the scope of IB directives for this ticket which clearly define that `node`/`npm` approach for assets and a shell script for release are valid approaches:

> * Convert all tasks in ... to use webpack or **node/npm directly**. This should have no/little impact on the scripts we run, but they shouldn't rely on gulp anymore.
> * The release task looks like the most involved one/the task without an obvious Webpack replacement (Imagemin and SVG manipulation should be straightforward): ..., but converting that to either **a shell script** or a node+git library script should be straightforward.

I have already spent estimated time for this ticket and making additional changes will result in exceeding the limit. So need your evaluation whether this approach is sufficient or not.

Moving it webpack is nice but it is not necessary because there is little benefit in it. Furthermore it will require us to increase complexity of our webpack config, bring more dev dependencies and rework javascript files to import images in JS/SVG files directly into javascript using es6 which doesn't look nice in my opinion. Plus, additional benefit of the current approach in this PR is that it doesn't introduce changes in actual javascript files thus don't affect functionality itself.

Thanks in advance.

eugene-manuilov:
@aaemnnosttv / @tofumatt how do you want to process images with webpack loaded from PHP?

aaemnnosttv:
@eugene-manuilov can you give an example? I can't think of a case where we render an image outside of React 🤔

eugene-manuilov:
@aaemnnosttv 

https://github.com/google/site-kit-wp/blob/897c92e479c3577ceceb3355b8c58c5c7287c4e3/includes/Core/Admin/Screen.php#L126-L133

https://github.com/google/site-kit-wp/blob/897c92e479c3577ceceb3355b8c58c5c7287c4e3/includes/Modules/AdSense.php#L539-L551

aaemnnosttv:
Thanks @eugene-manuilov - I think that should ideally be replaced with a vector rather than a `png` - we could maybe even use https://developer.wordpress.org/resource/dashicons/#googleplus since `$icon_url` supports a Dashicon class name which looks identical at first glance 😄 
Otherwise, we could use one of the other two routes available to [`add_menu_page`](https://developer.wordpress.org/reference/functions/add_menu_page/)
> * Pass a base64-encoded SVG using a data URI, which will be colored to match the color scheme. This should begin with 'data:image/svg+xml;base64,'.
> * Pass 'none' to leave div.wp-menu-image empty so an icon can be added via CSS.

Any of these I think would be an improvement over the static image used currently.

Are there any other cases to consider?

eugene-manuilov:
> Any of these I think would be an improvement over the static image used currently.

Can you tell me which one of those you want me to do? Just don't want to rework it again.

> Are there any other cases to consider?

There is one more, but that one is easier.

aaemnnosttv:
> Can you tell me which one of those you want me to do? Just don't want to rework it again.

Sure @eugene-manuilov, using the base64-encoded SVG using a data URI seems like the best option. The only thing I'm not 100% sure about is if this is supported in WP 4.7. Let's check that and if it is, let's go with that. Otherwise, let's go with the CSS option which we've used in other places as well in a similar way: https://github.com/google/site-kit-wp/blob/897c92e479c3577ceceb3355b8c58c5c7287c4e3/assets/sass/components/global/_googlesitekit-text-field-icon.scss#L23

eugene-manuilov:
@aaemnnosttv @tofumatt what do you think about SVG? Currently, we merge all svg files into a single big svg file full of sprites and load that file. If we want to use webpack, we will need to load svg files separately, right? How do you think we should migrate it into webpack?

aaemnnosttv:
@eugene-manuilov Unless @tofumatt has any other ideas or suggestions, I think we can use [`@svgr/webpack`](https://www.npmjs.com/package/@svgr/webpack) which is what `create-react-app` uses. It gives you a named `ReactComponent` for each `svg` you import and is pretty nice in my experience. See https://create-react-app.dev/docs/adding-images-fonts-and-files/#adding-svgs

eugene-manuilov:
> I think we can use @svgr/webpack which is what create-react-app uses. 

@aaemnnosttv it will require to rework all occurrences of the SvgIcon component to load svg files directly instead of using the sprite. Is it okay?

aaemnnosttv:
@eugene-manuilov at first glance it seems like there are a fair number of these that would need replacing (~42) although in most cases this looks like a straightforward replacement and some of them will obviously be duplicates. There are a few instances however where the svg `id` to show is dynamic where there would be a little more to rework (e.g. `PageHeader` and `moduleIcon` in `assets/js/util/index.js`). Given that most of them seem easy, I'm still in favor of this change in general.  I can definitely help figure out the way forward with the ones that aren't straightforward though. Perhaps we should create a new issue just for these SVG changes as a prerequisite to finishing this one since this touches a bit more than just build configuration? Let's talk about it on our call shortly.

aaemnnosttv:
@eugene-manuilov as discussed, I'm creating a new issue for the svg related work now which will need to be done before this issue can be finished. It should be ready for definition soon.

eugene-manuilov:
Going to close this PR in favor of https://github.com/google/site-kit-wp/pull/2159

PR #1785, opened 2020-07-15
Bug/1611 new ib server preloaded data

ryanwelcher:
## Summary

New PR that builds on the work done in #1692 
Addresses issue #1611 

## Checklist


assets/js/googlesitekit/api/middleware/preloading.js
```diff
@@ -0,0 +1,70 @@
...
+	}, {} );
+
+	let middlewareHasRun = false;
+	return ( options, next ) => {
+		const { parse = true } = options;
+		const uri = options.path;
+		setTimeout( () => {
+			middlewareHasRun = true;
+		}, timeout );
+		if ( typeof options.path === 'string' && ! middlewareHasRun ) {
```
aaemnnosttv, line 49: Let's move the timeout flag variable check right to the top of the middleware and return `next( options )` immediately if the cache has expired (i.e. before we do anything with `options`). The `setTimeout` call can go immediately after that conditional. This way we aren't unnecessarily calling `setTimeout` on every subsequent API call after it's been invalidated 👍

assets/js/googlesitekit/api/middleware/preloading.js
```diff
@@ -0,0 +1,70 @@
...
+ * Based on preloadMiddleware from from @wordpress/api-fetch, this middle is a single-use per-endpoint and provides cached
+ * data for the first request only and any subsequent requests hit the server.
+ *
+ * @since n.e.x.t
+ *
+ * @param {Object} preloadedData Preloaded data paths.
+ * @param {number} timeout       Timeout value.
+ * @return {Function} Function.
+ */
+function createPreloadingMiddleware( preloadedData, timeout = 1000 ) {
```
aaemnnosttv, line 36: The added `timeout` parameter is a good idea for testing, but it's not necessary because Jest let's us do all kinds of cool things with its fake timers 😄  For testing, all we want to be able to do is skip the waiting when we say, which Jest allows us to do (see my previous comment). This can be inlined right in the `setTimeout` because the value doesn't need to be variable or used anywhere else.

assets/js/googlesitekit/api/middleware/preloading.test.js
```diff
@@ -0,0 +1,179 @@
...
+				'https://www.googleapis.com/auth/userinfo.email',
+				'https://www.googleapis.com/auth/siteverification',
+				'https://www.googleapis.com/auth/webmasters',
+				'https://www.googleapis.com/auth/adsense.readonly',
+				'https://www.googleapis.com/auth/analytics.readonly',
+			],
+			unsatisfiedScopes: [],
+		},
+	},
+};
```
aaemnnosttv, line 51: This looks like a slightly older version of the original PR than what was merged, but if you recall we opted to use generic URIs and response data rather than referencing something "real" that we use in Site Kit. You can see what we had before here: https://github.com/google/site-kit-wp/pull/1734/files#diff-028d3f0fd8e3ea4860e0f6e64db9861fL30-L41

Just to reiterate, generic references are preferred because they don't "need" to be maintained with the real code they reference, and using minimal data in supporting code makes the tests easier to read and understand.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -27,18 +27,24 @@ import invariant from 'invariant';
...
-	controlCallback: () => {
+	argsToParams: ( { useCache = true } = {} ) => {
+		return {
+			options: { useCache },
+		};
+	},
+	controlCallback: ( { options: { useCache } } ) => {
 		return API.get( 'core', 'modules', 'list', null, {
-			useCache: false,
+			useCache,
```
aaemnnosttv, line 47: This change shouldn't be necessary as we no longer need `useCache` to be variable here.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -149,17 +145,21 @@ const baseActions = {
 		const { response, error } = yield fetchSetModuleActivationStore.actions.fetchSetModuleActivation( slug, active );
 		if ( response?.success === true ) {
 			// Fetch (or re-fetch) all modules, with their updated status.
-			yield fetchGetModulesStore.actions.fetchGetModules();
+			yield fetchGetModulesStore.actions.fetchGetModules( { useCache: false } );
```
aaemnnosttv, line 148:
```suggestion
			yield fetchGetModulesStore.actions.fetchGetModules();
```

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -149,17 +145,21 @@ const baseActions = {
...
 		return { response, error };
 	},
 };
 
 export const baseControls = {
-	[ REFETCH_AUTHENICATION ]: () => {
-		return API.get( 'core', 'user', 'authentication', { timestamp: Date.now() }, { useCache: false } );
-	},
+	[ REFETCH_AUTHENICATION ]: createRegistryControl( ( { dispatch } ) => () => {
+		return dispatch( CORE_USER ).fetchGetAuthentication( { useCache: false } );
```
aaemnnosttv, line 161:
```suggestion
		return dispatch( CORE_USER ).fetchGetAuthentication();
```

assets/js/googlesitekit/api/middleware/preloading.test.js
```diff
@@ -0,0 +1,179 @@
...
+		};
+		const next = jest.fn();
+		const firstResponse = await preloadingMiddleware( requestOptions, next );
+		expect( firstResponse ).toEqual( preloadedData[ requestURI ].body );
+		expect( next ).not.toHaveBeenCalled();
+
+		const secondRequest = await preloadingMiddleware( requestOptions, next );
+		expect( secondRequest ).toBeUndefined();
+		expect( next ).toHaveBeenCalled();
+	} );
```
aaemnnosttv, line 128: Added a blank line between tests 👍 
```suggestion
	} );

```

assets/js/googlesitekit/api/middleware/preloading.test.js
```diff
@@ -0,0 +1,179 @@
...
+	it( 'returns an uncached response after the timeout has run', async () => {
```
aaemnnosttv, line 129: Let's reword the title of the test to be more consistent with the others that it "calls the next middleware" rather than returning a response since this middleware can only return preloaded responses.

assets/js/googlesitekit/api/middleware/preloading.js
```diff
@@ -0,0 +1,72 @@
...
+	let cacheHasExpired = false;
+	return ( options, next ) => {
+		const { parse = true } = options;
+		const uri = options.path;
+		if ( cacheHasExpired ) {
+			return next( options );
+		}
+		setTimeout( () => {
+			cacheHasExpired = true;
+		}, 1000 );
```
aaemnnosttv, line 50: Let's swap these around as `parse` and `uri` aren't relevant for the expiration logic.
```suggestion
		if ( cacheHasExpired ) {
			return next( options );
		}
		setTimeout( () => {
			cacheHasExpired = true;
		}, 1000 );
		
		const { parse = true } = options;
		const uri = options.path;
		
```

assets/js/googlesitekit/api/middleware/preloading.test.js
```diff
@@ -0,0 +1,163 @@
...
+			path: firstRequestURI,
+		};
+		const next = jest.fn();
+
+		const firstResponse = await preloadingMiddleware( requestOptions, next );
+		expect( firstResponse ).toEqual( preloadedData[ firstRequestURI ].body );
+		expect( next ).not.toHaveBeenCalled();
+
+		jest.runAllTimers();
+		// Confirm that responses after the timeout don't run hit the middleware.
```
aaemnnosttv, line 133: There's a bit of a typo in this sentence ("run hit"); how about this?
```suggestion
		// Confirm that invocations after the timeout ignore preloaded response cache.
```

Issue #1774, opened 2020-07-14
Add `getReport` selector to `modules/search-console` store


felixarntz:
There should be a `getReport` selector issuing an API request to our `GET:modules/search-console/data/searchanalytics` datapoint, which relies on https://developers.google.com/webmaster-tools/search-console-api-original/v3/searchanalytics/query.

While maintaining backward-compatibility, the datapoint should be adjusted to support a range of arguments closer to the original. At the same time, this should happen in a way where similar arguments are named the same way across different report datapoints (e.g. in Analytics or AdSense).

While not yet fully in its intended a state, a good reference here is the [existing `getReport` selector of the `modules/adsense` store](https://github.com/google/site-kit-wp/blob/develop/assets/js/modules/adsense/datastore/report.js). The one here should be approached similarly.

While the PageSpeed Insights report endpoint is not time-based and generally simpler than the ones here, it's good to have a look at that respective [selector implementation](https://github.com/google/site-kit-wp/blob/develop/assets/js/modules/pagespeed-insights/datastore/report.js) too.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `modules/search-console` JS store should expose a selector `getReport( options = {} )`.
    * Should use `createFetchStore` to issue an API request to the `GET:modules/search-console/data/searchanalytics` datapoint.
    * Supported arguments (should all be documented on the selector):
        * `startDate`: Required, unless `dateRange` is provided. Start date to query report data for  as `YYYY-mm-dd`.
        * `endDate`: Required, unless `dateRange` is provided. End date to query report data for as `YYYY-mm-dd`.
        * `dateRange`: Required, alternatively to `startDate` and `endDate`. A date range string such as 'last-28-days'.
        * `compareDateRanges`: Optional, only relevant with `dateRange`. Default `false`.
            * In the future, we will revisit date handling so that we'll do the mapping of start and end date parsing on the client and then use `startDate` and `endDate`. For now, our code should continue to use `dateRange` (and `compareDateRanges` where needed). `startDate` and `endDate` however should already work today, to allow for flexible queries regardless of Site Kit limitations.
        * `dimensions`: Optional. List of dimensions to group results by. See https://developers.google.com/webmaster-tools/search-console-api-original/v3/searchanalytics/query#dimensionFilterGroups.filters.dimension for possible values. Default empty array.
        * `url`: Optional. URL to get a report for only this URL. Default empty string.
[...]

PR #1805, opened 2020-07-17
Enhancement/1774 getreport selector

eugene-manuilov:
## Summary

Addresses issue #1774 

## Checklist


assets/js/modules/search-console/datastore/report.js
```diff
@@ -0,0 +1,116 @@
...
+			reports: {
+				...state.reports,
+				[ stringifyObject( options ) ]: report,
+			},
+		};
+	},
+	argsToParams: ( options ) => {
+		invariant( isPlainObject( options ), 'Options for Analytics report must be an object.' );
+		invariant( isValidDateRange( options ), 'Either date range or start/end dates must be provided for Analytics report.' );
+
```
felixarntz, line 52: We should have validation here for `dimensions` too (if provided).

assets/js/modules/search-console/datastore/report.js
```diff
@@ -0,0 +1,116 @@
...
+			...state,
+			reports: {
+				...state.reports,
+				[ stringifyObject( options ) ]: report,
+			},
+		};
+	},
+	argsToParams: ( options ) => {
+		invariant( isPlainObject( options ), 'Options for Analytics report must be an object.' );
+		invariant( isValidDateRange( options ), 'Either date range or start/end dates must be provided for Analytics report.' );
```
felixarntz, line 51: These strings should use "Search Console" :)

assets/js/modules/search-console/datastore/report.js
```diff
@@ -0,0 +1,116 @@
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object}         state                       Data store's state.
+	 * @param {Object}         options                     Options for generating the report.
+	 * @param {string}         options.startDate           Required, unless dateRange is provided. Start date to query report data for as YYYY-mm-dd.
+	 * @param {string}         options.endDate             Required, unless dateRange is provided. End date to query report data for as YYYY-mm-dd.
+	 * @param {string}         options.dateRange           Required, alternatively to startDate and endDate. A date range string such as 'last-28-days'.
+	 * @param {boolean}        [options.compareDateRanges] Optional. Only relevant with dateRange. Default false.
+	 * @param {Array.<string>} [options.dimensions]        Optional. List of dimensions to group results by. Default an empty array.
```
felixarntz, line 88: Let's add the link that is present in the ACs to this parameter description, so that developers can look up the possible values.

assets/js/modules/search-console/datastore/report.js
eugene-manuilov: Added.

assets/js/modules/search-console/datastore/report.js
eugene-manuilov: Updated :)

assets/js/modules/search-console/datastore/report.js
eugene-manuilov: Added stadnard jsodc @link

includes/Modules/Search_Console.php
```diff
@@ -186,23 +186,53 @@ protected function create_data_request( Data_Request $data ) {
...
+							$dimension = trim( $dimension );
+							if ( empty( $dimension ) ) {
+								return false;
+							}
+
+							return $dimension;
+						},
+						is_array( $dimensions ) ? $dimensions : explode( ',', $dimensions )
+					)
+				);
```
felixarntz, line 230: Let's put this into a `parse_report_dimensions` helper method with PHPUnit test coverage.

Also, we should maybe align this logic to be the same as https://github.com/google/site-kit-wp/pull/1810/files#r465934242? We could possibly put the utility method `parse_string_list` into the `Module` base class, what do you think?

eugene-manuilov: IMO we should avoid bringing new functionality straight into Module and its children classes because it's already like a Swiss Army knife. Such big classes hard to understand, test and work with. I wish we created more specialized classes that are responsible for a single thing and follow SOLID principles.

API requests are good canditates to be extracted into separate classes where each class is responsible for a single request handling. Then we would be able to extract common functionlity into parent classes and better cover it with unit tests.

Of course, it is out-of-scope for this ticket because it will require a lot of thinking and time to work out a good strategy for it. So, for now i'll do what you suggested and add that function to the parent Module class.

eugene-manuilov: Updated.

Issue #1775, opened 2020-07-14
Add `getReport` selector to `modules/analytics` store


felixarntz:
There should be a `getReport` selector issuing an API request to our `GET:modules/analytics/data/report` datapoint, which relies on https://developers.google.com/analytics/devguides/reporting/core/v4/rest/v4/reports/batchGet.

While maintaining backward-compatibility, the datapoint should be adjusted to support a range of arguments closer to the original. At the same time, this should happen in a way where similar arguments are named the same way across different report datapoints (e.g. in Search Console or AdSense).

While not yet fully in its intended a state, a good reference here is the [existing `getReport` selector of the `modules/adsense` store](https://github.com/google/site-kit-wp/blob/develop/assets/js/modules/adsense/datastore/report.js). The one here should be approached similarly.

While the PageSpeed Insights report endpoint is not time-based and generally simpler than the ones here, it's good to have a look at that respective [selector implementation](https://github.com/google/site-kit-wp/blob/develop/assets/js/modules/pagespeed-insights/datastore/report.js) too.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `modules/analytics` JS store should expose a selector `getReport( options = {} )`.
    * Should use `createFetchStore` to issue an API request to the `GET:modules/analytics/data/report` datapoint.
    * Supported arguments (should all be documented on the selector):
        * `startDate`: Required, unless `dateRange` is provided. Start date to query report data for  as `YYYY-mm-dd`.
        * `endDate`: Required, unless `dateRange` is provided. End date to query report data for as `YYYY-mm-dd`.
        * `dateRange`: Required, alternatively to `startDate` and `endDate`. A date range string such as 'last-28-days'.
        * `compareDateRanges`: Optional, only relevant with `dateRange`. Default `false`.
        * `multiDateRange`: Optional, only relevant with `dateRange`. Default `false`.
            * In the future, we will revisit date handling so that we'll do the mapping of start and end date parsing on the client and then use `startDate` and `endDate`. For now, our code should continue to use `dateRange` (and `compareDateRanges` and `multiDateRange` where needed). `startDate` and `endDate` however should already work today, to allow for flexible queries regardless of Site Kit limitations.
        * `metrics`: Required. List of metrics to query. See https://ga-dev-tools.appspot.com/dimensions-metrics-explorer/ for possible values. Alternatively to a string, each element can also be an object with 'expression' and 'alias' properties. At least one metric is required.
[...]

PR #1809, opened 2020-07-17
Enhancement/1775 analytics getreport

eugene-manuilov:
## Summary

Addresses issue #1775

## Relevant technical choices

* Added validation function to help  us validate different properties of the getReport options.
* Reworked `GET:report` handler to properly define dimenssions, metrics and date ranges.

## Checklist


assets/js/modules/analytics/datastore/report.js
```diff
@@ -0,0 +1,154 @@
...
+			invariant(
+				isValidDimensions( dimensions ),
+				'Dimensions for an Analytics report must be either a string, an array of strings, an object, an array of objects or a mix of strings and objects. If an object is used, it must have "name" property.',
+			);
+		}
+
+		if ( orderby ) {
+			invariant(
+				isValidOrders( orderby ),
+				'Orders for an Analytics report must be either an object or an array of objects where each object should have "fieldName" and "sortOrder" properties.',
```
tofumatt, line 84: These are all super-handy errors, nice! 👍

assets/js/modules/analytics/datastore/report.js
```diff
@@ -0,0 +1,154 @@
...
+	/**
+	 * Gets a Search Console report for the given options.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object}         state                     Data store's state.
+	 * @param {Object}         options                   Optional. Options for generating the report.
+	 * @param {string}         options.startDate         Required, unless dateRange is provided. Start date to query report data for as YYYY-mm-dd.
+	 * @param {string}         options.endDate           Required, unless dateRange is provided. End date to query report data for as YYYY-mm-dd.
+	 * @param {string}         options.dateRange         Required, alternatively to startDate and endDate. A date range string such as 'last-28-days'.
```
tofumatt, line 121:
```suggestion
	 * @param {string}         options.dateRange         Required, alternative to startDate and endDate. A date range string such as 'last-28-days'.
```

assets/js/util/report-validation.js
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * Validates data that can be either string or object of the certain type, or array of them.
+ *
+ * @since n.e.x.t
+ *
+ * @param {string|string[]|Object|Object[]} data The data to check.
+ * @param {Function} objectVerificationCallback The callback to verify an object.
+ * @return {boolean} TRUE if data is valid, otherwise FALSE.
+ */
+function isValidStringsOrObjects( data, objectVerificationCallback ) {
```
tofumatt, line 28: This isn't really a callback so much as a verification/check/test function. `verifyFunction` or `isValidObjectFunction` might work a bit better? But not a big deal.

assets/js/util/report-validation.js
```diff
@@ -0,0 +1,134 @@
...
+	if ( typeof data === 'string' ) {
+		return true;
+	}
+
+	if ( typeof data === 'object' && objectVerificationCallback.call( null, data ) ) {
```
tofumatt, line 33: What's the benefit of using `objectVerificationCallback.call(null, data)` here rather than `objectVerificationCallback(data)`?

assets/js/util/report-validation.js
```diff
@@ -0,0 +1,134 @@
...
+			const type = typeof data[ i ];
+			if ( type !== 'string' && ( type !== 'object' || ! objectVerificationCallback.call( null, data[ i ] ) ) ) {
+				return false;
+			}
+		}
+
+		return true;
+	}
+
+	return false;
```
tofumatt, line 48: We can use `every` for this check to tidy this up a bit:

```suggestion
	if ( Array.isArray( data ) ) {
		return data.every( ( item ) => {
			typeof item === 'string' || objectVerificationCallback( item )
		 } );
	}

	// Arguably this should fail/throw, because none of our allowed types were encountered.
	return false;
```

assets/js/util/report-validation.js
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * @param {string} dateRange The date range to check.
+ * @param {string} startDate The start date to check.
+ * @param {string} endDate The end date to check.
+ * @return {boolean} TRUE if either date range or start/end dates are valid, otherwise FALSE.
+ */
+export function isValidDateRange( dateRange, startDate, endDate ) {
+	const validStartDate = startDate && startDate.match( /^\d{4}-\d{2}-\d{2}$/ );
+	const validEndDate = endDate && endDate.match( /^\d{4}-\d{2}-\d{2}$/ );
+	const validDateRange = dateRange && dateRange.match( /^last-\d+-days$/i );
```
tofumatt, line 98: I don't think these should be case-insensitive, unless Google APIs and our own are fine with `LAST-30-DAYS`.

assets/js/util/report-validation.js
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * @since n.e.x.t
+ *
+ * @param {string} dateRange The date range to check.
+ * @param {string} startDate The start date to check.
+ * @param {string} endDate The end date to check.
+ * @return {boolean} TRUE if either date range or start/end dates are valid, otherwise FALSE.
+ */
+export function isValidDateRange( dateRange, startDate, endDate ) {
+	const validStartDate = startDate && startDate.match( /^\d{4}-\d{2}-\d{2}$/ );
```
tofumatt, line 96: When Site Kit stops working in the year 10,000, we'll be to blame 😉

assets/js/util/report-validation.js
```diff
@@ -0,0 +1,134 @@
...
+
+	if ( Array.isArray( orders ) ) {
+		for ( let i = 0; i < orders.length; i++ ) {
+			if ( typeof orders[ i ] !== 'object' || ! isValidOrder( orders[ i ] ) ) {
+				return false;
+			}
+		}
+
+		return true;
+	}
```
tofumatt, line 127: Again: this could be refactored to use `every` rather than the `for` loop, which is just a bit more concise and applicable given we aren't operating on the array but just verifying every element passes a test.

assets/js/util/report-validation.js
```diff
@@ -0,0 +1,134 @@
...
+ * Verifies that either date range or start and end dates are valid.
+ *
+ * @since n.e.x.t
+ *
+ * @param {string} dateRange The date range to check.
+ * @param {string} startDate The start date to check.
+ * @param {string} endDate The end date to check.
+ * @return {boolean} TRUE if either date range or start/end dates are valid, otherwise FALSE.
+ */
+export function isValidDateRange( dateRange, startDate, endDate ) {
```
tofumatt, line 95: Because not all arguments are required, we shouldn't use positional arguments here. A function signature like: 
```suggestion
export function isValidDateRange( { dateRange, startDate, endDate } = {} ) {
```

Would allow you to check `startDate` and `endDate` without needing to pass a bogus first argument to the function.

assets/js/util/report-validation.test.js
```diff
@@ -0,0 +1,185 @@
...
+} from './report-validation';
+
+describe( 'Reporting API validation', () => {
+	describe( 'isValidDateRange', () => {
+		it( 'should return TRUE if dateRange is valid only', () => {
+			expect( isValidDateRange( 'last-14-days', '', null ) ).toBeTruthy();
+		} );
+
+		it( 'should return TRUE if startDate and endDate are valid only', () => {
+			expect( isValidDateRange( '', '2020-01-01', '2020-04-05' ) ).toBeTruthy();
```
tofumatt, line 36: As mentioned, I think changing the function signature to use object args means we don't have to pass an argument to `dateRange`, which is clearer in use. Bonus: it makes it clear which is the start and end date.

assets/js/util/report-validation.js
eugene-manuilov: We will have a little time to fix it before Jan 1 10,000.

assets/js/util/report-validation.js
eugene-manuilov: Updated.

assets/js/util/report-validation.js
eugene-manuilov: In this particular case, there is no additional benefit of using one instead of another. Both approaches are valid. It's just a matter of personal preference. I have updated it to be as you want.

assets/js/util/report-validation.js
eugene-manuilov: Updated.

assets/js/util/report-validation.js
eugene-manuilov: Updated.

assets/js/util/report-validation.js
eugene-manuilov: Updated.

assets/js/modules/analytics/datastore/report.js
```diff
@@ -0,0 +1,154 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * External dependencies
+ */
+import invariant from 'invariant';
+import isPlainObject from 'lodash/isPlainObject';
```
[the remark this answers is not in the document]
eugene-manuilov: In theory, using `lodash/{...}` should let us have smaller bundles. Anyway this question is out of scope for this PR.

assets/js/util/report-validation.js
eugene-manuilov: Updated.

assets/js/util/report-validation.test.js
eugene-manuilov: Updated.

assets/js/modules/analytics/datastore/report.js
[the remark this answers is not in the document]
tofumatt: Ah, I figured webpack's tree-shaking would do the same thing, but maybe the way lodash's main imports work that doesn't.

assets/js/util/report-validation.test.js
```diff
@@ -0,0 +1,196 @@
...
+} from './report-validation';
+
+describe( 'Reporting API validation', () => {
+	describe( 'isValidDateRange', () => {
+		it( 'should return TRUE if dateRange is valid only', () => {
+			expect( isValidDateRange( {
+				dateRange: 'last-14-days',
+				startDate: '',
+				endDate: null,
+			} ) ).toBeTruthy();
```
tofumatt, line 36:
```suggestion
			expect( isValidDateRange( {
				dateRange: 'last-14-days',
			} ) ).toBeTruthy();
```

Reading the test, those extra args make it _seem_ as if they're required but that `dateRange` being valid means they're allowed to be invalid… but the test here really is just that "passing a valid `dateRange` returns `true".

While the test is true, it ends up testing for not _exactly_ what it says, or at least combines a test of "`dateRange`-only works" with "valid `dateRange` + invalid `startDate`/`endDate` works".

Arguably we probably should _only_ accept one arg or the other, as combining them isn't clear as to which one takes precedence.

But at least here we should limit the arguments to clarify the test. Happy if you want to add extra tests like this to show that `dateRange` takes precedence over `startDate`+`endDate` if all are supplied, which seems to be the case, but at least I think we should limit the input for this test.

eugene-manuilov: > Happy if you want to add extra tests like this to show that dateRange takes precedence over startDate+endDate if all are supplied...

The `isValidDate Range` function has nothing to do with precedence, it just checks that either date range or start/end dates are provided.

assets/js/modules/analytics/datastore/report.js
```diff
@@ -0,0 +1,151 @@
...
+	 * @param {string}         options.dateRange         Required, alternative to startDate and endDate. A date range string such as 'last-28-days'.
+	 * @param {boolean}        options.compareDateRanges Optional. Only relevant with dateRange. Default false.
+	 * @param {boolean}        options.multiDateRange    Optional. Only relevant with dateRange. Default false.
+	 * @param {Array.<string>} options.metrics           Required. List of metrics to query.
+	 * @param {Array.<string>} options.dimensions        Optional. List of dimensions to group results by. Default an empty array.
+	 * @param {Array.<Object>} options.orderby           Optional. An order definition object, or a list of order definition objects, each one containing 'fieldName' and 'sortOrder'. 'sortOrder' must be either 'ASCENDING' or 'DESCENDING'. Default empty array.
+	 * @param {string}         options.url               Optional. URL to get a report for only this URL. Default an empty string.
+	 * @param {number}         options.limit             Optional. Maximum number of entries to return. Default 1000.
```
felixarntz, line 125: Let's mark the optional parameters like e.g. `[options.dimensions]`.

assets/js/modules/analytics/datastore/report.js
```diff
@@ -0,0 +1,151 @@
...
+			return;
+		}
+
+		yield fetchGetReportStore.actions.fetchGetReport( options );
+	},
+};
+
+const baseSelectors = {
+	/**
+	 * Gets a Search Console report for the given options.
```
felixarntz, line 110:
```suggestion
	 * Gets an Analytics report for the given options.
```

assets/js/modules/analytics/datastore/report.js
```diff
@@ -0,0 +1,151 @@
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object}         state                     Data store's state.
+	 * @param {Object}         options                   Optional. Options for generating the report.
```
felixarntz, line 115: This is required, because some of the options are required basically. We should probably not use `{}` (contrary to the ACs, I just noticed it's in there) as default, but instead fail in here if `! isPlainObject( options )`.

assets/js/modules/analytics/datastore/report.js
```diff
@@ -0,0 +1,151 @@
...
+
+		invariant( isPlainObject( options ), 'Options for Analytics report must be an object.' );
+		invariant( isValidDateRange( options ), 'Either date range or start/end dates must be provided for Analytics report.' );
+
+		if ( metrics ) {
+			invariant(
+				isValidMetrics( metrics ),
+				'Metrics for an Analytics report must be either a string, an array of strings, an object, an array of objects or a mix of strings and objects. If an object is used, it must have "expression" and "alias" properties.',
+			);
+		}
```
felixarntz, line 69: Since `metrics` is required, this `invariant` should always run, not just if `metrics` is available.

assets/js/util/report-validation.js
```diff
@@ -0,0 +1,124 @@
...
+		return true;
+	}
+
+	if ( Array.isArray( data ) ) {
+		return data.every( ( item ) => typeof item === 'string' || ( typeof item === 'object' && verifyFunction( item ) ) );
+	}
+
+	// Arguably this should fail/throw, because none of our allowed types were encountered.
+	return false;
+}
```
felixarntz, line 43: +1 for putting these general validation functions into `assets/js/util`.

assets/js/util/report-validation.js
```diff
@@ -0,0 +1,124 @@
...
+ * Verifies that provided metrics match allowed values. Metrics can be a string,
+ * an array of string, an array of objects or mix of strings and objects. Objects
+ * must have "expression" and "alias" properties in order to be considered as valid.
+ *
+ * @since n.e.x.t
+ *
+ * @param {string|string[]|Object|Object[]} metrics The metrics to check.
+ * @return {boolean} TRUE if metrics are valid, otherwise FALSE.
+ */
+export function isValidMetrics( metrics ) {
```
felixarntz, line 55: Since this is specific to Analytics, I think we should either rename this function to `isValidAnalyticsMetrics` or move it to `assets/js/modules/analytics/util` (in there it could keep the current name since it would have the context from being in that folder).

The alternative here is we make this more generic. This means we would diverge a little bit from the original APIs in certain cases, but on the other hand it would be nice if these APIs would all work in a similar way when called via Site Kit - possibly makes it more trivial. On the other hand, if we diverge, reading the API docs could cause false assumptions on that it works the same way in Site Kit - what do you think @eugene-manuilov?

assets/js/util/report-validation.js
```diff
@@ -0,0 +1,124 @@
...
+ * Verifies provided dimensions to make sure it matches allowed values. It can be a string,
+ * array of strings, an object with "name" field, array of such objects or an array of strings
+ * and objects.
+ *
+ * @since n.e.x.t
+ *
+ * @param {string|string[]|Object|Object[]} dimensions The dimensions to check.
+ * @return {boolean} TRUE if dimensions are valid, otherwise FALSE.
+ */
+export function isValidDimensions( dimensions ) {
```
felixarntz, line 73: Same applies here, this is specific to Analytics.

includes/Modules/Analytics.php
```diff
@@ -854,23 +854,41 @@ protected function create_data_request( Data_Request $data ) {
 					return true;
 				};
 			case 'GET:report':
-				$date_range = $data['dateRange'] ?: 'last-28-days';
-
-				$dimensions = array_map(
-					function ( $name ) {
-						$dimension = new Google_Service_AnalyticsReporting_Dimension();
-						$dimension->setName( $name );
+				$request_args = array( 'page' => $data['url'] );
```
felixarntz, line 857: This works, but I think it'd be more clear to only set this key in an if clause, only if `! empty( $data['url'] )`.

includes/Modules/Analytics.php
```diff
@@ -854,23 +854,41 @@ protected function create_data_request( Data_Request $data ) {
 					return true;
 				};
 			case 'GET:report':
-				$date_range = $data['dateRange'] ?: 'last-28-days';
```
felixarntz, line 857: Let's add a check here to ensure `$data['metrics']` is not empty, and if it is return an error.

includes/Modules/Analytics.php
```diff
@@ -879,11 +897,18 @@ function ( $name ) {
...
+				$date_range = $data['dateRange'] ?: 'last-28-days';
+				$start_date = $data['startDate'];
+				$end_date   = $data['endDate'];
+				if ( ! strtotime( $start_date ) || ! strtotime( $end_date ) ) {
+					list( $start_date, $end_date ) = $this->parse_date_range(
 						$date_range,
 						$data['compareDateRanges'] ? 2 : 1
-					),
+					);
+				}
```
felixarntz, line 908: I think this code would be a bit more clear to read if we just checked first `strtotime( $data['startDate'] ) && strtotime( $data['endDate'] )` and then used an `else` for the date range logic.

includes/Modules/Analytics.php
```diff
@@ -905,44 +930,68 @@ function ( $date_range ) {
...
-					},
-					(array) $data['orderby']
-				);
-				$request->setOrderBys( $orderby );
+					$orderby = array_map(
+						function ( $order_def ) {
+							$order_def = array_merge(
+								array(
+									'fieldName' => '',
+									'sortOrder' => '',
```
felixarntz, line 980: Maybe we should specify "DESCENDING" here as default, to ensure this is never empty for a request?

assets/js/util/report-validation.js
eugene-manuilov: @felixarntz I like the idea to make it more generic, but unfortunately I agree with you that it could cause false assumptions and as a result it might confuse developers who will work with it and expect the same behavior from Site Kit. I'll move it to the analytics module.

assets/js/modules/analytics/datastore/report.js
eugene-manuilov: Updated.

assets/js/modules/analytics/datastore/report.js
eugene-manuilov: Updated.

assets/js/modules/analytics/datastore/report.js
eugene-manuilov: Updated.

assets/js/util/report-validation.js
eugene-manuilov: Updated.

assets/js/util/report-validation.js
eugene-manuilov: Updated.

includes/Modules/Analytics.php

includes/Modules/Analytics.php
eugene-manuilov: Updated.

includes/Modules/Analytics.php

includes/Modules/Analytics.php
eugene-manuilov: I am not sure that I get what you want. If I do it, we will have repeated code which is not what we need, isn't it? Or I miss something?

assets/js/modules/analytics/datastore/report.js
```diff
@@ -0,0 +1,144 @@
...
+				[ stringifyObject( options ) ]: report,
+			},
+		};
+	},
+	argsToParams: ( options ) => {
+		const {
+			metrics,
+			dimensions,
+			orderby,
+		} = options;
```
felixarntz, line 55: This needs to happen after the initial `invariant` check I think? Otherwise we're destructuring potentially `undefined`.

assets/js/modules/analytics/datastore/report.js
```diff
@@ -0,0 +1,144 @@
...
+	/**
+	 * Gets an Analytics report for the given options.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object}         state                       Data store's state.
+	 * @param {Object}         options                     Options for generating the report.
+	 * @param {string}         options.startDate           Required, unless dateRange is provided. Start date to query report data for as YYYY-mm-dd.
+	 * @param {string}         options.endDate             Required, unless dateRange is provided. End date to query report data for as YYYY-mm-dd.
+	 * @param {string}         options.dateRange           Required, alternative to startDate and endDate. A date range string such as 'last-28-days'.
```
felixarntz, line 111: Maybe this should be optional? Since it's only required as an alternative to `startDate` and `endDate`. Not sure, but I guess technically one of them is optional, and I think since this only here for BC, we should mark it as optional.

Alternatively, maybe we just shouldn't document the three legacy arguments at all? We need them for our own BC, but we don't really want other people to use them, so 🤷‍♂️ ? Previously mentioned in https://github.com/google/site-kit-wp/issues/1775#issuecomment-658236590

includes/Modules/Analytics.php
felixarntz: Oh I guess what I was meaning to say is: Let's only access `dateRange` in the if clause here. That should be the only place where it's needed, and it makes it more obvious when looking at the code that it is just a fallback.

assets/js/modules/analytics/datastore/report.js
eugene-manuilov: I would prefer to leave it here and we can't make it optional because it is required if other dates are not provided. If you still think we need to remove it or make it optional, let me know and i'll update it.

includes/Modules/Analytics.php
eugene-manuilov: It's also needed when `multiDateRange` is set. See here: https://github.com/google/site-kit-wp/blob/c52626a6b6346c1a92496f552e3b9895c6d030b6/includes/Modules/Analytics.php#L926-L928

So, we can't move it into the if clause.

assets/js/modules/analytics/datastore/report.js
eugene-manuilov: Good catch! Updated.

includes/Modules/Analytics.php
felixarntz: @eugene-manuilov I see, however `multiDateRange` is also a legacy argument that we should only support if `dateRange` is provided. So this should go into the clause for when `startDate` and `endDate` do not exist.

This means that we'll at some point need to introduce further arguments for the secondary date range to make this available in combination with the new `startDate` and `endDate` arguments. Maybe `compareStartDate` and `compareEndDate`? Anyway, let's not do this now since it needs more definition.

Can you move the `multiDateRange` bit accordingly? With that, this should be good to go.

eugene-manuilov: Yep, updated. I think it should be good now.

Issue #1208, opened 2020-03-04
call_user_func warnings raised when using wp-cli


majemedia:
## Bug Description

```
$ wp plugin list
Warning: call_user_func() expects parameter 1 to be a valid callback, array must have exactly two members in wp-includes/rest-api/class-wp-rest-server.php on line 1165
Warning: call_user_func() expects parameter 1 to be a valid callback, array must have exactly two members in wp-includes/rest-api/class-wp-rest-server.php on line 1165
```

```
$ wp plugin list --skip-plugins=google-site-kit
no warnings
```

## Steps to reproduce

1. Run `wp plugin list`
 - see warnings above plugin list table
1. Run `wp plugin list --skip-plugins=google-site-kit`
 - no warnings above plugin list table

## Additional Context

This seems to be environment specific and I can't duplicate on a server running PHP 7.4 (shrug)

- PHP CLI Version: `PHP 7.2.24-0ubuntu0.18.04.3 (cli)`
- PHP FPM Version: `v7.2.24-0ubuntu0.18.04.3`
- Server: `DISTRIB_DESCRIPTION="Ubuntu 18.04.4 LTS"`
- WP Cli Version: 2.4.0
- WP Core version: 5.3.2
- Google Site Kit: 1.4.0 (also reproduced on 1.3.1)

I DO have opcache enabled for fpm & cli on the affected server but when I turned it off I still got the warnings.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The REST API should not trigger a PHP warning due to Site Kit.
* Site Kit should provide REST API response schemas as expected by WordPress core.

## Implementation Brief

* Update `\Google\Site_Kit\Core\Modules\Modules::get_module_schema` method to return a closure that returns the schema itself.

## QA Brief

* Install and activate `woocommerce`
* With Site Kit active, run `wp plugin list`
* See no warnings above table output

## Changelog entry

* Fix PHP warning that could occur when retrieving REST API information via help endpoint.

lbonomo:
Hi.
I have the same problem, only together WooCommerce.
I see this "warning" when add a "new REST_Route()" with "schema". Example "... 'schema' => $this->get_module_schema()..." in line 583 of file google-site-kit/includes/Core/Modules/Modules.php

AgileIT:
Hi,
I have the same problem. Error with WooCommerce for each wp-cli request.
[...]

PR #1811, opened 2020-07-20
Update get_module_schema to return a function

aaemnnosttv:
## Summary

Addresses issue #1208 

## Relevant technical choices

* I was only able to reproduce the warnings with WooCommerce active, but confirmed that this fixes it.

## Checklist


includes/Core/Modules/Modules.php
```diff
@@ -806,68 +807,71 @@ private function prepare_module_data_for_response( Module $module ) {
 	 * Gets the REST schema for a module.
 	 *
 	 * @since 1.3.0
+	 * @since n.e.x.t Returns a function that returns the schema when invoked.
 	 *
-	 * @return array Module REST schema.
+	 * @return Closure Function which returns the Module REST schema.
```
felixarntz, line 812: This works, but why not keep this method as is and pass a closure that calls this to WP instead? I think it's more intuitive for a method called `get_module_schema` to actually return the schema directly.

aaemnnosttv: Yeah that makes more sense, I probably misunderstood the IB. I'll update it to be this way instead 👍

PR #1816, opened 2020-07-21
Add grid logic and SCSS.

tofumatt:
## Summary

Addresses issue #1678.

## Checklist


tofumatt:
So, this one is ready for a first pass, but I had to divert from the original plan a bit. It wasn't possible to introspect the components enough to tell if they render `null` or not without creating a `ref` that's forwarded to the Widget and checking it for a false-y value. This means we will require that Widgets accept a `ref`, but it also means they don't have to do things like dispatch an action to say they aren't rendering.

I also rendered a "hidden" div rather than `null` for now as, again, it was more straightforward.

There are a lot of tests here to cover the options, but let me know if you'd like to see more. I went with snapshots as it made creating the tests easy and kept weird queries out of them (it also gave a good sense of the output shape).

This one required a fair bit of rethinking from the original approach to get it all working, but I think it's in a good place. Please let me know if I missed anything on it.

assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js
```diff
@@ -57,9 +153,14 @@ const WidgetAreaRenderer = ( { slug } ) => {
...
 						return (
-							<Widget key={ widget.slug } slug={ widget.slug } />
+							<WidgetRenderer
+								className={ widgetClassNames[ i ] !== null ? classnames( widgetClassNames[ i ] ) : 'googlesitekit-widget-area--hidden' }
+								key={ widget.slug }
+								ref={ setRef( widget.slug ) }
+								slug={ widget.slug }
+							/>
 						);
 					} ) }
```
felixarntz, line 165: This way of rendering widgets is only correct for the "boxes" style. For the "composite" style, this widgets loop should be wrapped in another:
```
    <div className="mdc-layout-grid">
        <div className="mdc-layout-grid__inner">
            { widgets }
        </div>
    </div>
</div>
```

See "Overall widget area grid and styling" in IB.

assets/js/googlesitekit/widgets/components/WidgetRenderer.js
```diff
@@ -39,14 +44,15 @@ const WidgetRenderer = ( { slug } ) => {
 
 	// Capitalize the "component" variable, as it is required by JSX.
 	const { component: Component, wrapWidget } = widget;
-	const widgetComponent = <Component slug={ slug } />;
+	const widgetComponent = <Component className={ wrapWidget ? className : undefined } ref={ ref } slug={ slug } />;
 
-	return widgetComponent && wrapWidget
```
felixarntz, line 44: This shouldn't be removed, if the component returns `null`, it shouldn't be wrapped. We cannot rely on the approach where `Widget` returns `null` if it has no `children`, because `Widget` is not a mandatory component here (that approach only works for the `wrapWidget` approach).

felixarntz:
@tofumatt I've experimented with this for a while; the current approach of relying on using refs didn't appear to work correctly, and we [cannot rely on the `Widget` component counting its `children`](https://github.com/google/site-kit-wp/pull/1816/commits/af2410cb75720ac7c197539952f5e52f8dcd20bc) because not every component uses `Widget`.

I've implemented [another approach](https://github.com/google/site-kit-wp/pull/1816/commits/9d02dbd67186bf033ddc6e43c5837a6453d744b8), which is somewhat hacky, but closer to the IB and works as expected: By calling the component directly in `WidgetRenderer`, we can check whether it returns `null`. And then `WidgetRenderer` modifies state of `WidgetAreaRenderer` to let it know whether the widget has content or not. This isn't very clean at the moment, and I haven't added specific tests yet since I wanted to discuss first, but potentially we could go with this approach at least for now, since changing it will be possible at any time due to these components being fully internal. The `Widget` class, which is public, should not have to deal with any of that. Would be great if you could have a look, let's talk through this on Monday.

I've also found a few [issues with the class replacement logic which I fixed](https://github.com/google/site-kit-wp/pull/1816/commits/f1724d85174b552186aa35d53bbd22a5267a35f1), it's now working as per the ACs.

I've also made a few minor updates:
* [Register widget areas in their correct styles.](https://github.com/google/site-kit-wp/pull/1816/commits/88052b438b6fbdff9b1f97c6472648209bf97204)
* [Follow-up fixes for PageSpeed Insights widget.](https://github.com/google/site-kit-wp/pull/1816/commits/651954c5f2731395c8a6c91bf0ca6ab90bec4363) (see #1302)
* Misc CSS updates (https://github.com/google/site-kit-wp/pull/1816/commits/95f4fd7d7770844c0a601d6db70edfacf24f6c3c and https://github.com/google/site-kit-wp/pull/1816/commits/862526605807c3125a5ed4168be3e652dfffa3e9).

For testing all of that in the actual UI, what I did is add the following extra widget registrations to `assets/js/modules/pagespeed-insights/index.js`:
```js
	Widgets.registerWidget( 'halfWidth1', {
		component: () => 'Half Width 1',
		width: Widgets.WIDGET_WIDTHS.HALF,
		wrapWidget: true,
	}, [ AREA_DASHBOARD_SPEED, AREA_PAGE_DASHBOARD_SPEED ] );

	Widgets.registerWidget( 'nullWidget', {
		component: () => null,
		width: Widgets.WIDGET_WIDTHS.HALF,
		wrapWidget: true,
	}, [ AREA_DASHBOARD_SPEED, AREA_PAGE_DASHBOARD_SPEED ] );

	Widgets.registerWidget( 'quarterWidth1', {
		component: () => 'Quarter Width 1',
		width: Widgets.WIDGET_WIDTHS.QUARTER,
		wrapWidget: true,
	}, [ AREA_DASHBOARD_SPEED, AREA_PAGE_DASHBOARD_SPEED ] );

	Widgets.registerWidget( 'halfWidth2', {
		component: () => 'Half Width 2',
		width: Widgets.WIDGET_WIDTHS.HALF,
		wrapWidget: true,
	}, [ AREA_DASHBOARD_SPEED, AREA_PAGE_DASHBOARD_SPEED ] );

	Widgets.registerWidget( 'halfWidth3', {
		component: () => 'Half Width 3',
		width: Widgets.WIDGET_WIDTHS.HALF,
		wrapWidget: true,
	}, [ AREA_DASHBOARD_SPEED, AREA_PAGE_DASHBOARD_SPEED ] );

	Widgets.registerWidget( 'quarterWidth2', {
		component: () => 'Quarter Width 2',
		width: Widgets.WIDGET_WIDTHS.QUARTER,
		wrapWidget: true,
	}, [ AREA_DASHBOARD_SPEED, AREA_PAGE_DASHBOARD_SPEED ] );

	Widgets.registerWidget( 'quarterWidth3', {
		component: () => 'Quarter Width 3',
		width: Widgets.WIDGET_WIDTHS.QUARTER,
		wrapWidget: true,
	}, [ AREA_DASHBOARD_SPEED, AREA_PAGE_DASHBOARD_SPEED ] );

	Widgets.registerWidget( 'quarterWidth4', {
		component: () => 'Quarter Width 4',
		width: Widgets.WIDGET_WIDTHS.QUARTER,
		wrapWidget: true,
	}, [ AREA_DASHBOARD_SPEED, AREA_PAGE_DASHBOARD_SPEED ] );
```

Issue #1280, opened 2020-03-24
Change percentages should match service UI.


adamsilverstein:
## Bug Description

The "Change" percent displayed in Site Kit in the Analytics detail page does not match what is shown in the Analytics product UI. The time periods are correct, the calculations seem off.

* First discovered when working on https://github.com/google/site-kit-wp/issues/1202.

A related issue where these numbers may be swapped on the Search Console details page: https://github.com/google/site-kit-wp/issues/870

## Steps to reproduce

1. Go to the Analytics detail page on a site with moderate traffic.
2. Check reports for 28 days and 90 days.
3. Compare the percent change shown for Users, Sessions, Bounce Rate and Session Duration against the numbers shown in the Analytics UI.
4. Note the numbers do not match.

## Screenshots

**90 day view for my personal site:**
![image](https://user-images.githubusercontent.com/2676022/77459529-7b218280-6dc5-11ea-97d3-4130bce6fce0.png)

Users, Sessions, Bounce Rate and Session: 129.7, 128.9, 0 (green), 21.3

**same 90 days in Analytics UI:**
![image](https://user-images.githubusercontent.com/2676022/77459607-98eee780-6dc5-11ea-978a-72e46f069115.png)
 
Users, Sessions, Bounce Rate and Session: 133.6, 133.6, 0 (red), 19.5

## Additional Context

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The change percentages shown for Analytics data for a 90 day period comparison should match the percentages in Analytics UI (except Users, see #1681).

## Implementation Brief

* In `includes/Core/Modules/Module.php:parse_date_range` add a new `day_align` parameter that defaults to false. when true, the previous period's time range will be adjusted to match the closest period with the same days of the week (ie starting on the same day of the week).
* Compare the current period's end (yesterday)  day of the week to the calculated end of the previous date range. When they don't match, adjust the date...
```php
$previous_day_of_week = date( 'w', strtotime( $date_end ) );
$yesterday_day_of_week = date( 'w', strtotime( 'yesterday' ) );
if ( $day_align && $previous_day_of_week !== $yesterday_day_of_week ) { ...
```
* To adjust the date, offset the period to move it to the closest period with matching days. If the previous period is earlier in the week and less than 4 days away, move the period forward. If the previous period is later in the week and more than 4 days away, move the period forward. Otherwise, move the period backwards. Move the period by the difference in the days of the week between the two periods.

```php
[...]

PR #1822, opened 2020-07-22
Analytics: match days of week for previous period comparisons.

adamsilverstein:
## Summary

Addresses issue https://github.com/google/site-kit-wp/issues/1280

## Checklist


includes/Core/Modules/Module.php
```diff
@@ -570,11 +570,12 @@ private function make_data_request( Data_Request $data ) {
 	 * @param int    $offset     Days the range should be offset by. Default 1. Used by Search Console where
 	 *                           data is delayed by two days.
 	 * @param bool   $previous   Whether to select the previous period. Default false.
+	 * @param bool   $day_align  Whether to align the previous period days of the week to current period. Default false.
 	 *
 	 * @return array List with two elements, the first with the start date and the second with the end date, both as
 	 *               'Y-m-d'.
 	 */
-	protected function parse_date_range( $range, $multiplier = 1, $offset = 1, $previous = false ) {
+	public function parse_date_range( $range, $multiplier = 1, $offset = 1, $previous = false, $day_align = false ) {
```
aaemnnosttv, line 578: Let's not change the visibility here as that only seems to be done for testing purposes. While it's a little less clean, let's use `Reflection` to make the method accessible and invoke it in the test instead. See `\Google\Site_Kit\Tests\Core\Util\Activation_NoticeTest::assertAssetsEnqueued` for an example.

includes/Core/Modules/Module.php
```diff
@@ -587,6 +588,21 @@ protected function parse_date_range( $range, $multiplier = 1, $offset = 1, $prev
 		$start_date_offset = $offset + $number_of_days - 1;
 		$date_start        = gmdate( 'Y-m-d', strtotime( $start_date_offset . ' days ago' ) );
 
+		// Check the day of the week alignment.
+		$previous_day_of_week  = gmdate( 'w', strtotime( $date_end ) );
+		$yesterday_day_of_week = gmdate( 'w', strtotime( 'yesterday' ) );
```
aaemnnosttv, line 593: Why do we specifically use `yesterday` here? Should it not be relative to `$date_start` or at least include the `$offset` ?

tests/phpunit/integration/Modules/AnalyticsTest.php
```diff
@@ -153,6 +153,62 @@ public function test_get_datapoints() {
...
+	/**
+	 * @dataProvider data_parse_date_range
+	 */
+	public function test_parse_date_range( $period_requested, $previous_period_end_offset ) {
+		$analytics = new Analytics( new Context( GOOGLESITEKIT_PLUGIN_MAIN_FILE ) );
+
+		$result = $analytics->parse_date_range( 'last-' . $period_requested . '-days', 1, 1, true, true );
+		$previous_end = strtotime( $result[ 1 ] );
+		$yesterday_day_of_week = gmdate( 'w', strtotime( 'yesterday' ) );
+		$diff = round( ( strtotime( 'yesterday' ) - ( $period_requested * 86400 )  - $previous_end ) / 86400 );
```
aaemnnosttv, line 165: Let's use the WP constant instead of the literal here for more clarity.
```suggestion
		$diff = round( ( strtotime( 'yesterday' ) - ( $period_requested * DAY_IN_SECONDS )  - $previous_end ) / DAY_IN_SECONDS );
```

includes/Core/Modules/Module.php
adamsilverstein: Here we are comparing the day of the week for the last period vs the previous period. 

The periods are the same number of days, so it doesn't matter if we compare the first day or the last day of the periods - they should each align/be the same day of the week. 

The last period **always ends "yesterday"**, so it is easy to compute.

I'll add some inline docs to clarify the logic here.

includes/Core/Modules/Module.php
```diff
@@ -570,11 +570,12 @@ private function make_data_request( Data_Request $data ) {
 	 * @param int    $offset     Days the range should be offset by. Default 1. Used by Search Console where
 	 *                           data is delayed by two days.
 	 * @param bool   $previous   Whether to select the previous period. Default false.
+	 * @param bool   $day_align  Whether to align the previous period days of the week to current period. Default false.
```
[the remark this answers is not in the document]
adamsilverstein: Oh yep - fixed! I _may_ have been trying to avoid changing the doc block indentation.

includes/Core/Modules/Module.php
adamsilverstein: Aha, thanks for the tip!

tests/phpunit/integration/Modules/AnalyticsTest.php
adamsilverstein: Oh, right!

adamsilverstein:
@aaemnnosttv Thanks for the feedback which I have addressed. Please review.

includes/Core/Modules/Module.php
adamsilverstein: Adjusted

includes/Core/Modules/Module.php
aaemnnosttv: Thanks @adamsilverstein. I find the "last" vs "previous" rather confusing as to which one means the most recent and which means the second most recent 😄  I understand what it means here since it's documented but elsewhere  it isn't as obvious.

My understanding of the `$offset` here was that this was needed because not all services define the period in the same way so that offsetting by 1 day was necessary in some cases and not in others (or by 2 as mentioned in doc block). If the last period always ends "yesterday" for all services, regardless of `$offset` then I am confused what `$offset` is for.

includes/Core/Modules/Module.php
```diff
@@ -587,6 +588,25 @@ protected function parse_date_range( $range, $multiplier = 1, $offset = 1, $prev
 		$start_date_offset = $offset + $number_of_days - 1;
```
aaemnnosttv, line 588: This hasn't changed, but looks odd as `$offset` here already includes `$number_of_days` above for a previous period (from L583). That would mean `$start_date_offset` could be `($last_x_days * $multiplier) + ($last_x_days * $multiplier) + $offset` - is that correct?

includes/Modules/Analytics.php
```diff
@@ -919,7 +919,7 @@ function ( $dimension_def ) {
 					// When using multiple date ranges, it changes the structure of the response,
 					// where each date range becomes an item in a list.
 					if ( ! empty( $data['multiDateRange'] ) ) {
-						$date_ranges[] = $this->parse_date_range( $date_range, 1, 1, true );
+						$date_ranges[] = $this->parse_date_range( $date_range, 1, 1, true, true );
```
aaemnnosttv, line 922: The signature of this method is growing rather cryptic. It's probably beyond the scope of this issue to change, but I think it would be much more human-friendly if it took an array of arguments rather than a list of positional numbers and booleans. Can we open a new issue to address this?

tests/phpunit/integration/Modules/AnalyticsTest.php
```diff
@@ -153,6 +154,63 @@ public function test_get_datapoints() {
 		);
 	}
 
+	/**
+	 * @dataProvider data_parse_date_range
+	 */
+	public function test_parse_date_range( $period_requested, $previous_period_end_offset ) {
```
aaemnnosttv, line 160: Since the method belongs to the base `Module` and has not changed in the Analytics class (only how it's called), this test should move to `\Google\Site_Kit\Tests\Core\Modules\ModuleTest` respectively and use `FakeModule` for the instance. 

By doing this we can also add `parse_date_range` to that class to make it public so that using `Reflection` isn't necessary here. (I know that was my suggestion but I must have missed that there wasn't anything Analytics-specific in the implementation at the time - or it changed 😄 ). See `FakeModule::exception_to_error`.

tests/phpunit/integration/Modules/AnalyticsTest.php
```diff
@@ -153,6 +154,63 @@ public function test_get_datapoints() {
...
+		$previous_end          = strtotime( $result[1] );
+		$yesterday_day_of_week = gmdate( 'w', strtotime( 'yesterday' ) );
+		$diff                  = round( ( strtotime( 'yesterday' ) - ( $period_requested * DAY_IN_SECONDS ) - $previous_end ) / DAY_IN_SECONDS );
+
+		$previous_day_of_week  = gmdate( 'w', strtotime( $result[1] ) );
+		$yesterday_day_of_week = gmdate( 'w', strtotime( 'yesterday' ) );
+
+		$this->assertEquals( $previous_day_of_week, $yesterday_day_of_week );
+		$this->assertEquals( $previous_period_end_offset, $diff );
+
```
aaemnnosttv, line 174: Extra empty line at the end 🙂 
```suggestion
```

tests/phpunit/integration/Modules/AnalyticsTest.php
```diff
@@ -153,6 +154,63 @@ public function test_get_datapoints() {
...
+				2, // wed -> previous mon
+			),
+			array(
+				13,
+				1, // tues -> previous mon
+			),
+			array(
+				14,
+				0, // mon === mon
+			),
```
aaemnnosttv, line 210: Please add cases for `28`, and `90` days here as well to cover all of the ranges we use, but also because `90` is the odd one that raised this.

adamsilverstein:
> My understanding of the $offset here was that this was needed because not all services define the period in the same way so that offsetting by 1 day was necessary in some cases and not in others (or by 2 as mentioned in doc block). If the last period always ends "yesterday" for all services, regardless of $offset then I am confused what $offset is for.

Right, the period ends yesterday for all Analytics and Adsense queries, and two days ago for all Search console queries (although we should revisit that soon as search console data may now be fresher). Since the week-alignment for past period comparison logic only applies to Analytics, the yesterday logic applies, however this could include the offset logic as well if we discover that search console uses a similar alignment approach for past period comparisons (I haven't noticed that or checked carefully).

adamsilverstein:
>  I find the "last" vs "previous" rather confusing as to which one means the most recent and which means the second most recent

Same, open to better naming convention here. "previous" is clear, "past " is confusing - maybe "last"?

includes/Core/Modules/Module.php
adamsilverstein: This is correct, the math here is moving from the end of the period to the beginning of the period, so we need to add number_of_days (including the potential multiplier) again.

The logic works like this:
Say we are requesting a **previous** period, with an offset of 1 (default, ends yesterday) and a period of 7 days.
$offset is 1 + 7 = 8 - the period ends here - 8 days ago
the start of the period is then calculated:
$start_date_offset = $offset + $number_of_days - 1;
= 8 + 7 - 1 = 14 - the period starts 14 days ago. 14days ago - 8 days ago is the 7 day period that is previous to the last 7 days ending yesterday.

includes/Modules/Analytics.php
adamsilverstein: Agreed, I will open a separate issue for this.

tests/phpunit/integration/Modules/AnalyticsTest.php
adamsilverstein: Ok, I'll move this. Thanks for the tip.

aaemnnosttv:
> Since the week-alignment for past period comparison logic only applies to Analytics, the yesterday logic applies, however this could include the offset logic as well if we discover that search console uses a similar alignment approach for past period comparisons (I haven't noticed that or checked carefully).

Since this logic lives in the base `Module` date parsing method, there shouldn't be module-specific details about the implementation of the period; it sounds like that's what you're saying. It's okay if some modules will only ever call it with the `$weekday_align` parameter - that's totally fine, but we shouldn't assume that the period always ends "yesterday" only because that's true for the only modules that call the method with that parameter. How the service defines the periods for its data is up to that module/service and should affect how this method is called by that module, but the method should always work the same depending on how it's called, regardless of which service it's for. Does that make sense?

includes/Core/Modules/Module.php
adamsilverstein: That should have read 'the last period for analytics period always ends yesterday'. In any case, the offset would not change the day alignment logic since it would be added to both the end of the last and previous period, cancelling out in the math. So we can safely ignore offset for this section.

includes/Modules/Analytics.php
adamsilverstein: Created issue: https://github.com/google/site-kit-wp/issues/1877

includes/Core/Modules/Module.php
adamsilverstein: you are right about offset and how it is used. For the purposes of the day of the week comparison, we can safely ignore it though, since the previous period and the last period would  **both get the same offset**; the offset added to both the previous and past period so it cancels out when calculating if the days of the weeks align.

tests/phpunit/integration/Modules/AnalyticsTest.php
adamsilverstein: I moved the test to `tests/phpunit/integration/Core/Modules/ModuleTest.php` - that makes more sense.

> By doing this we can also add parse_date_range to that class to make it public so that using Reflection isn't necessary here.

@aaemnnosttv - By "that class" do you mean `Google\Site_Kit\Core\Modules`? That is where `parse_date_range` already is. So, should I change the visibility to public and remove the reflection in the test? I feel like these helpers should be private/protected unless we have a reason to expose them - what do you think @felixarntz?

tests/phpunit/integration/Modules/AnalyticsTest.php
adamsilverstein: Added.

adamsilverstein:
@aaemnnosttv This is ready for another review!

includes/Core/Modules/Module.php
```diff
@@ -575,7 +575,7 @@ private function make_data_request( Data_Request $data ) {
 	 * @return array List with two elements, the first with the start date and the second with the end date, both as
 	 *               'Y-m-d'.
 	 */
-	protected function parse_date_range( $range, $multiplier = 1, $offset = 1, $previous = false, $weekday_align = false ) {
+	public function parse_date_range( $range, $multiplier = 1, $offset = 1, $previous = false, $weekday_align = false ) {
```
aaemnnosttv, line 578: The visibility of this method shouldn't change here, it can however be changed in `FakeModule` like `\Google\Site_Kit\Tests\Core\Modules\FakeModule::exception_to_error`

tests/phpunit/integration/Modules/AnalyticsTest.php
aaemnnosttv: No sorry, I meant you can define the same method on `FakeModule` as `public` and simply call `parent::parse_date_range` with the same args so that it's callable without reflection for the test but doesn't need to be in the base `Module` class. Like this method https://github.com/google/site-kit-wp/blob/d68b6cedfcef57cb8cdb44e02dc00ca51b590182/tests/phpunit/includes/Core/Modules/FakeModule.php#L203-L205

includes/Core/Modules/Module.php
adamsilverstein: Ah, ok got it!

adamsilverstein: Fixed in [`476fd98` (#1822)](https://github.com/google/site-kit-wp/pull/1822/commits/476fd98d9ed7ad15f0d7d972a3a82d83bb875b98)

tests/phpunit/integration/Modules/AnalyticsTest.php
adamsilverstein: Got it, fixed in [`476fd98` (#1822)](https://github.com/google/site-kit-wp/pull/1822/commits/476fd98d9ed7ad15f0d7d972a3a82d83bb875b98)

adamsilverstein:
@aaemnnosttv Ready for review!

includes/Core/Modules/Module.php
felixarntz: Hmm as far as I can tell, I'm also thinking this should use the `$offset` instead of hard-coding "yesterday". "yesterday" works for Analytics, since for that the `$offset` is 1, but we should respect that parameter. If we wanted to use `$weekday_align` for e.g. Search Console where the `$offset` is 2, this wouldn't work correctly anymore, since the offset right now is only considered for the previous period's `$date_end`, but not the original period's end date because that is hard-coded to "yesterday".

adamsilverstein:
> Hmm as far as I can tell, I'm also thinking this should use the $offset instead of hard-coding "yesterday". "yesterday" works for Analytics, since for that the $offset is 1,

Right, this will only work for Analytics as written, I'll switch to using $offset + ' days ago' and add a test case to confirm.

includes/Core/Modules/Module.php
adamsilverstein: @felixarntz - I updated the code here to include the offset in calculations as well as adding test run's for offset = 2 for all cases.

adamsilverstein:
Code now handles and tests for $offset of 2. Ready for review.

adamsilverstein:
> Made one final update to clarify the test purpose and move the new variable more closely to the area where it's used to make the code a bit easier to follow. Also added a QA Brief, please make sure to provide these in the future.

Thank you! 

Will do - I'll make sure to add a QA brief for every ticket when moving to code review.

Issue #1681, opened 2020-06-16
Analytics user and user percent change should match Analytics dashboard


adamsilverstein:
## Bug Description

When viewing the Analytics data in Site Kit, the number shown for user change percent should match the number shown in the Google Analytics Dashboard for all reporting periods. Currently all numbers match **except** user change percent.

Previously discussed in https://github.com/google/site-kit-wp/issues/1280

## Steps to reproduce

1. Set up Analytics in Site Kit.
2. Compare the user change percentage change data shown in Analytics to the data shown in Site Kit

### Expected result
The numbers should match.

### Actual result
The numbers are different.

## Screenshots

<img width="1464" alt="7 days" src="https://user-images.githubusercontent.com/2676022/84824350-02554f80-afdd-11ea-8a30-c2987a538618.png">

![28-days](https://user-images.githubusercontent.com/2676022/84824357-05e8d680-afdd-11ea-95af-ce68aff3399e.png)

![90days](https://user-images.githubusercontent.com/2676022/84824363-08e3c700-afdd-11ea-86a2-e9fa04fafe37.png)

## Additional Context

 - PHP Version: 
 - Plugin Version [e.g. 22]

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The user percentage change in Site Kit should match the user percentage change displayed in the Analytics frontend.
    * This is most likely currently wrong for the same reason #1202 was implemented (see PR #1277) - that issue only resolved the problem for the overall number, but not the percentage. We can probably use the same request from there to get the percentage, by modifying https://github.com/google/site-kit-wp/pull/1277/files#diff-a75b3a0da6c74c4dc70fbf476e42f23cR437 to include the date range comparison.
    * This should be tested for any date range other than 90 days, which has its own separate problem (#1280).
* It should calculate the percentage change from the same API response that is used for the absolute users number.
* This should be fixed for:
    * Site Kit dashboard
    * Site Kit Analytics page
    * WP dashboard widget
    * `traffic-increase` publisher win
    * Admin-bar total users

## Implementation Brief

* In `assets/js/components/publisher-wins/index.js` swap the data point used:
   * Change `overviewReportDataDefaults` to `userReportDataDefaults` for the data callback with identifier `traffic-increase`

* In `assets/js/components/publisher-wins/traffic-increase.js` change user/percent calculations. Remove use of `calculateOverviewData` and instead extract the data from the new API call:

```js
[...]

PR #1823, opened 2020-07-22
Analytics: consistently use direct user query for user/visitor metric.

adamsilverstein:
## Summary

Addresses issue https://github.com/google/site-kit-wp/issues/1681

## Checklist


assets/js/modules/analytics/components/wp-dashboard/WPAnalyticsDashboardWidgetOverview.js
```diff
@@ -106,7 +143,33 @@ export default withData(
...
+			maxAge: getTimeInSeconds( 'day' ),
+			context: 'WPDashboard',
+			toState( state, { data } ) {
+				if ( ! state.directTotalUsers ) {
+					const directTotalUsers = get( data, '[0].data.totals[0].values[0]' );
+					const previousTotalUsers = get( data, '[0].data.totals[1].values[0]' );
+					return {
+						directTotalUsers,
+						previousTotalUsers,
+					};
```
aaemnnosttv, line 170: Let's use conditional chaining here too since it's all new. Also, (perhaps a bit of a stylistic nitpick) it might be a bit cleaner to include the values in the definition of the returned object rather than initializing variables right before the return. i.e.` directTotalUsers: data?.[0]?.data?.totals?.[0]?.values?.[0],`

If these values are accessed in multiple places repeatedly, it might be worth extracting a little `parseTotalUsersData` utility.

assets/js/modules/analytics/components/dashboard/AnalyticsDashboardWidgetTopLevel.js
```diff
@@ -258,12 +261,14 @@ export default withData(
 			},
 			priority: 1,
 			maxAge: getTimeInSeconds( 'day' ),
-			context: [ 'Dashboard' ],
+			context: 'Dashboard',
 			toState( state, { data } ) {
 				if ( ! state.directTotalUsers ) {
 					const directTotalUsers = get( data, '[0].data.totals[0].values[0]' );
+					const previousTotalUsers = get( data, '[0].data.totals[1].values[0]' );
```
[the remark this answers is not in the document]
adamsilverstein: 👍

assets/js/modules/analytics/components/wp-dashboard/WPAnalyticsDashboardWidgetOverview.js
adamsilverstein: Adjusted.

adamsilverstein:
Addressed feedback and will test data/rerun visual tests.

assets/js/modules/analytics/components/wp-dashboard/WPAnalyticsDashboardWidgetOverview.js
adamsilverstein: I fixed this instance. I like your `parseTotalUsersData` idea, we currently use this in four places so seems useful. I will add that.

adamsilverstein:
@aaemnnosttv The visual regression errors were caused by an incomplete data capture, thank you for catching those in your review!

I regenerated data for the wp-dashboard, site kit dashboard and analytics page and regenerated images which now look better. Please review.

ps. reviewing the new images, I noticed a slight column overlap we might want to address in a separate issue:

image

assets/js/modules/analytics/util/index.js
```diff
@@ -470,3 +471,16 @@ export const getTopPagesReportDataDefaults = () => {
 		limit: 10,
 	};
 };
+
+/**
+ * Returns the extracted total and past user data.
+ *
```
aaemnnosttv, line 477:
```suggestion
 *
 * @since n.e.x.t
 *
```

assets/js/modules/analytics/util/index.js
```diff
@@ -470,3 +471,16 @@ export const getTopPagesReportDataDefaults = () => {
...
+ * @param {Object} data The data returned from the Analytics API call.
```
aaemnnosttv, line 478:
```suggestion
 * @param {Array} data The data returned from the Analytics API call.
```

adamsilverstein:
> Regarding the VRT changes, I noticed that sessions and bounce rate stats changed in addition to Users

Although unrelated to the code changes, the VRT changes are expected. Because the data capture is per-page and I had to recapture data for the component changes here, the _data for all components on this page will change_ to current data values. While the numbers and charts have changed, we can verify that the numbers look reasonable, and the display looks fine.

adamsilverstein:
@aaemnnosttv thanks for the feedback and code improvements; moving this back to code review.

adamsilverstein:
> One thing I noticed (unrelated to this PR though) is that Site Kit was respecting a date range that was one day off from what Analytics showed me 

I will follow up and ask about this, hopefully we can match that cutoff time. The way our code currently works with calls like `strtotime( 'yesterday' )` I believe means the cutoff is related to the server time settings.

> The issue still needs a QA brief, to cover all the different areas where this has been changed and needs to be tested.

Will add.

adamsilverstein:
> One thing I noticed (unrelated to this PR though) is that Site Kit was respecting a date range that was one day off from what Analytics showed me

I found this information: the "day boundary" used for reports is determined by timezone ("Time zone country or territory") set in the view properties.

image

I checked the API response for the Analytics management API (Views->List) and the response does include the timezone:

image

I will create a follow up issue to adjust reporting here so our day boundary matches Analytics. 

Here are some initial details:

Analytics:

* Store the timezone for the user when they create or select a view.
* Update existing view record for already configured sites to grab the timezone.
* Update `parse_date_range` (and tests) to offset from the view timezone.

In addition we should check other services that use `parse_date_range` to ensure their day boundaries match Site Kit's. Some quick searches found:

* AdSense uses Pacific Time Zone (PTZ) by default, but users can change reporting timezones **at the account level** https://developers.google.com/adsense/management/timezone. The API returns timezone as part of the Account endpoint:

image

* Search console **timestamps data according to Pacific Daylight Time.** according to https://support.google.com/analytics/answer/1308626?hl=en implying the cutoff would be relative to Pacific daylight time (not respecting daylight savings time).

adamsilverstein:
@felixarntz Added QA brief on issue, moving this back to review (@cole10up please let me know if that is helpful and if you need help reproducing the "Traffic increase" win).

Issue #1818, opened 2020-07-22
Do not run ZIP and Storybook build workflows on PRs from forks


felixarntz:
## Bug Description

The GitHub actions for building ZIP files and building Storybook currently run on every PR, regardless of origin. This causes a problem for forks because usually those contributors do not have access to push to the wiki or to the `gh-pages` branch. Let's modify the GitHub actions so that they only run on PRs that come from the original `google/site-kit-wp` repo.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* GitHub actions to build and deploy ZIP files and build and deploy Storybook should only run on PRs from `google/site-kit-wp` branches, but not on PRs that come from forks.

## Implementation Brief

* Update job `if` conditions for the following jobs to prevent running when triggered by forks
    - `build-storybook`
    - `build-zips`
    - `remove-pr`
    - `remove-storybook`
* Include the following logic in the `if`
    - If the job already has an `if` condition, include this with `&&`  
    ```yml
    if: ( existing condition ) && github.event.pull_request.head.repo.fork == false
    ```
    - Otherwise, add the `if` to the job including the status function which is `success()` by default  
    ```yml
    if: success() && github.event.pull_request.head.repo.fork == false
    ```
* Deploy jobs depend on build jobs so they do not need added conditionals

Note: "When you use expressions in an if conditional, you may omit the expression syntax (`${{ }}`) because GitHub automatically evaluates the if conditional as an expression."

Reference: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif

**Suggested minor enhancements**
* Combine `build-zips` and `remove-pr` into a single `google-site-kit-zips.yml` workflow, similar to how the Storybook workflow contains `build`, `deploy`, and `remove` jobs
* Rename `build-storybook.yml` to `storybook.yml`
* Update workflow `name`s to be a user-facing name, e.g. "Storybook" or "Plugin Zip Files" as currently it has the same name as a job, which can be confusing when referencing a job id whether a job or a workflow is being referred to

## QA Brief

* 

## Changelog entry

* N/A

felixarntz:
IB ✅ 

The renaming/reorganizing suggestions sound good to me as well.

tofumatt:
QA ✅

Checks were skipped for a user without write permissions:

[...]

PR #1829, opened 2020-07-23
Enhancement/1818 pr storybook not for forks

eugene-manuilov:
## Summary

Addresses issue #1818 

## Relevant technical choices

* Instead of creating `google-site-kit-zips.yml` file, decided to name it as `zips.yml` to be consistent with `storybook.yml`.

## Checklist


.github/workflows/storybook.yml
```diff
@@ -14,8 +14,9 @@ on:
 
 jobs:
   build-storybook:
+    name: Build Storybook
     runs-on: ubuntu-latest
-    if: ${{ ! ( github.event_name == 'pull_request' && github.event.action == 'closed' ) }}
+    if: ${{ ! ( ( github.event_name == 'pull_request' && github.event.action == 'closed' ) || github.event.pull_request.head.repo.fork ) }}
```
aaemnnosttv, line 19: Let's remove the expression syntax here, both for consistency (the job below does not use it) but also to make this a bit easier to read.
```suggestion
    if: ! ( ( github.event_name == 'pull_request' && github.event.action == 'closed' ) || github.event.pull_request.head.repo.fork )
```

.github/workflows/zips.yml
```diff
@@ -0,0 +1,122 @@
...
+      - opened
+      - closed
+      - reopened
+      - synchronize
+
+jobs:
+  build-zips:
+    name: Build Zips
+    runs-on: ubuntu-latest
+    if: ${{ ! ( ( github.event_name == 'pull_request' && github.event.action == 'closed' ) || github.event.pull_request.head.repo.fork ) }}
```
aaemnnosttv, line 20: Same here, let's remove the expression syntax.
```suggestion
    if: ! ( ( github.event_name == 'pull_request' && github.event.action == 'closed' ) || github.event.pull_request.head.repo.fork )
```

.github/workflows/storybook.yml
eugene-manuilov: @aaemnnosttv i don't remember why exactly I needed to add `${{ }}` but it had been done on purpose. Let's try to remove it and see if it can work without expression syntax.

eugene-manuilov:
@aaemnnosttv ok, i have remembered why we can't remove `${{ }}` from `if` conditions where conditions start with `!`. Yaml parser treats values that start from `!` as special tags/functions and throws an error if it can't find appropriate handler for it. That's why we need to wrap those `if` conditions with expression syntax.

githubactions

aaemnnosttv:
@eugene-manuilov ah, okay that makes more sense now. How about we change them to be `false == ( conditions )` instead? I think that's a bit harder to miss than the `! ( conditions )` as well.

eugene-manuilov:
> @eugene-manuilov ah, okay that makes more sense now. How about we change them to be `false == ( conditions )` instead? I think that's a bit harder to miss than the `! ( conditions )` as well.

We can try it. Let' me update it.

Issue #1710, opened 2020-06-24
E2E Tests for WP nightly consistently fail


aaemnnosttv:
## Bug Description

Our E2E test job which runs on the nightly/trunk version of WordPress consistently fails with the same error (a timeout in the Search Console date range test).

This job is allowed to fail as it is running on a pre-release version of WordPress but the purpose of this job is also to address failures before the upcoming release if there are breaking changes to be addressed.

Example failing jobs:
* https://travis-ci.com/github/google/site-kit-wp/jobs/352990831
* https://travis-ci.com/github/google/site-kit-wp/jobs/352930516
* https://travis-ci.com/github/google/site-kit-wp/jobs/352855407

Note: the PHP tests job which fails running on WP nightly/trunk is failing for a different reason related to the nightly _PHP version_ used in that job, so this job is expected to continue to fail for the time being.

## Screenshots

![image](https://user-images.githubusercontent.com/1621608/85519223-d6a01e00-b609-11ea-9553-92be34ce22cd.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The **E2E Tests (WordPress nightly)** job should pass

## Implementation Brief

In WP 5.5, there is a change that causes `esc_url_raw( false )` to return an empty string instead of `false` (I'm not sure exactly where but it is reproducible). This causes references to `_googlesitekitLegacyData.permaLink` used in data requests to calculate a different cache key for the request which doesn't match what we've provided causing the test to fail because the component is in a "waiting for data" state.

```
# WP 5.4.2
wp> esc_url_raw( false )
=> bool(false)
wp> exit
www-data@2efd5ac3af93:~/wordpress$ wp core update --version=nightly
Updating to version nightly (en_US)...
Downloading update from https://wordpress.org/nightly-builds/wordpress-latest.zip...
Unpacking the update...
Cleaning up files...
No files found that need cleaning up.
Success: WordPress updated successfully.
www-data@2efd5ac3af93:~/wordpress$ wp shell
wp> esc_url_raw( false )
=> string(0) ""
```

* Update `permaLink` to always be `false` if not set https://github.com/google/site-kit-wp/blob/7442f4add1ba2b79449aabb2e4b5283949bf0f43/includes/Core/Assets/Assets.php#L816
* Merge #1837 

I also diffed `_googlesitekitLegacyData` between 5.4.2 and the current nightly and found that other than `timestamp` this was the only change, which is expected.

## QA Brief

[...]

Issue #1555, opened 2020-05-13
Fix deprecation notices on PHP nightly


felixarntz:
## Bug Description

Function `ReflectionParameter::isArray()` is deprecated, see e.g. https://travis-ci.com/github/google/site-kit-wp/jobs/333190714

For background information on how to replace usage, see https://github.com/php/php-src/pull/5209 (`ReflectionParameter::getType()`). We may need to keep both and use them depending on PHP version active.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be no PHP warnings generated by running our tests.
* Travis-CI build for PHP nightly and WordPress trunk should pass.

## Implementation Brief

* 

## QA Brief

* 

## Changelog entry

*

rahul2106:
Please assign the issue to me. Interested in starting my first contribution here.

aaemnnosttv:
Thanks for your enthusiasm @rahul2106. Unfortunately this issue is still being defined so it's not quite ready for execution yet.

aaemnnosttv:
@felixarntz not sure there is anything we can do about this right now. I tried upgrading PHPUnit to the latest supported version (v7) and it still raises deprecated warnings. See #1603 

I checked WordPress core travis configuration and the PHP8 job is added to the allowed failures, and currently failing.
- https://travis-ci.com/github/WordPress/wordpress-develop/builds/168543704
- https://travis-ci.com/github/WordPress/wordpress-develop/jobs/340875516#L2945

Moving this back to AC for now, but it's safe to say that this one isn't actionable right now.

felixarntz:
@aaemnnosttv Is this still relevant? We can probably close it.

PR #1837, opened 2020-07-28
Ensure permaLink is always `false` if not a URL

aaemnnosttv:
## Summary

Addresses issue #1710 

## Relevant technical choices

* In WP 5.5, there is a change that causes `esc_url_raw( false )` to return an empty string instead of `false`. This change ensures that `permaLink` is always `false` if it isn't set.

## Checklist


tofumatt:
PHP tests are still failing here, but the E2E ones are passing, so I think that's good. 👍

aaemnnosttv:
Thanks @tofumatt !

The PHP tests failure is due to an incompatibility with the latest version of PHP used there, see https://github.com/google/site-kit-wp/issues/1555#issuecomment-635878145

PR #1782, opened 2020-07-14
Fix/publisher wins broken

felixarntz:
## Summary

Addresses issue #1781

## Checklist

- [x] My code is tested and passes existing unit tests.
- [ ] My code has an appropriate set of unit tests which all pass.
- [x] My code is backward-compatible with WordPress 4.7 and PHP 5.6.
- [x] My code follows the [WordPress](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) coding standards.
- [x] My code has proper inline documentation.
- [x] I have signed the Contributor License Agreement (see <https://cla.developers.google.com/>).

Issue #1781, opened 2020-07-14
Fix broken publisher win notifications to show up again


felixarntz:
## Bug Description

Our publisher win notifications have been broken for several releases at least. I hadn't been seeing them for months, and earlier finally took the time to figure out what's going on (a bunch of problems).

This issue is to address these bugs so that these notifications show up again like they used to. Note that we're not going to refactor anything here at this point, publisher wins will remain legacy for the foreseeable future. Improving our mechanisms for notifications overall (which will include this) is likely somewhere down the line in late 2020 or early 2021. We're not going to add any new publisher win notifications until then.

TLDR: There are still very fundamental architectural issues with this code, but this is not going to be addressed soon. We'll need to keep in mind to consider these notifications when making changes to things like our API calls, datapoints, API caching etc. which is clearly not DRY but necessary for now. The good thing after investing a few hours into this code is I finally have a grasp of how these things (somehwat) work at the moment 😆 

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Publisher wins should show up on the dashboard again as expected:
    * On a fresh site with exactly 1 post, after activating Site Kit and authenticating you should see the notification "Congrats on your first post!".
    * On a fresh site with exactly 5 posts, after activating Site Kit and authenticating you should see the notification "Congrats on five published posts", accompanied by some Search Console stats.
    * On a site with more than 5 posts, after activating Site Kit and authenticating you should see the notification "Welcome Back!", accompanied by some Search Console stats.
    * With Analytics active, if your site's page views over the last month increased more than 10% compared to the month before, you should see the notification "Increased page views!", accompanied by some Analytics stats.
    * With Analytics active, if your site's unique visitors over the last month increased between 10% and 100% compared to the month before, you should see the notification "Congrats on more website visitors!", accompanied by some Analytics stats.
* All of these notifications should be dismissable and in this case they shouldn't show next time. (Note that only one of them will show at the same time.)

## Implementation Brief

* Fix duplicate JS action callback name `googlesitekit.dataLoadedGetNotifications` which was used by two different hooks. Both of these remove themselves after running once, however due to the duplicate name one would also remove the other. This prevented publisher win logic from running entirely.
[...]

Issue #297, opened 2019-07-17
Incorrect order of Clicks and Impressions data blocks + trends in admin bar vs single URL page


lilybonney:
Addressed by PR #892 

---

## Bug Description

The admin bar dropdown shows clicks then impressions, while the Site Kit dashboard (and WP dashboard) shows impressions then clicks - the latter makes more sense as impressions > clicks.

However, the trend stats underneath are not re-ordered, which results in showing the wrong trend for each stat:

How stats are currently displayed in the admin bar: 
![URL-admin](https://user-images.githubusercontent.com/13304791/68954120-a535b900-07c3-11ea-9d20-ca2b5420337b.png)

In the single URL detailed report, the order is reversed for clicks and impressions, but the trend numbers are not, resulting in contradictory reporting:  
![URL-dashboard](https://user-images.githubusercontent.com/13304791/68954137-a8c94000-07c3-11ea-9d1d-9d8ddc293bd2.png)

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The stats should be displayed in the same order both in the admin bar and in the single URL page: impressions followed by clicks
* The trend change stats should match the correct metric: impressions trend change for impressions, and clicks trend change for clicks. 

## Implementation Brief

* Swap the order of data blocks in `<SearchConsoleAdminbarWidgetOverview>` so that total impressions are shown first
* Correct references to change and sparkline data columns in `<SearchConsoleDashboardWidgetTopLevel>`
* Merge #892 

## Changelog entry

* Display impressions before clicks in admin bar as that is the commonly expected order.

felixarntz:
IB ✅

tofumatt:
QA ✅

Order is consistent in both screens, looks good! 👍

marrrmarrr:
👁👍

PR #1842, opened 2020-07-29
Bug/1781 publishing win fix

felixarntz:
## Summary

Addresses issue #1781 (follow-up to #1782)

## Relevant technical choices

* Ensure Search Console data extraction doesn't divide by 0. &rarr; If Search Console property is new for example, the list of rows will be empty, so we cannot blindly divide by `count`.
* Only display stats for `publishing-win` if they are relevant (i.e. if one of the values is not 0).

### Secondary improvements

* Display "Clicks" before "Impressions" like we do elsewhere (see #297).
* Move Search Console utilities to a more appropriate location.

## Checklist


assets/js/components/publisher-wins/publishing-win.js
```diff
@@ -50,25 +50,32 @@ const publishingWin = ( data, id ) => {
...
-				datapoint: averageCTR,
-				datapointUnit: '%',
-			},
-		];
+		// Only display stats if they are relevant. If everything is 0, it's likely
+		// because the Search Console property is entirely new and hasn't aggregated
+		// any data yet for the site.
+		if ( 0 !== parseInt( totalClicks ) ||
+			0 !== parseInt( totalImpressions ) ||
+			0 !== parseInt( averageCTR ) ) {
```
aaemnnosttv, line 58: `parseInt` doesn't default to base 10, so we should always call it with `10` as the second parameter to avoid unintended results.
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#Description

Issue #1700, opened 2020-06-19
Use hashes for all JS file names to avoid JS errors due to aggressive caching


jamesozzie:
## Bug Description

Using version 1.10.0 there have been separate reports of the Site Kit dashboard displaying a blank page. 

Console errors below:
```
load-scripts.php?c=0&load[chunk_0]=jquery-core,jquery-migrate,utils,moxiejs,plupload&ver=5.4.2:8 JQMIGRATE: Migrate is installed, version 1.4.1
googlesitekit-wp-dashboard.js:41 Uncaught TypeError: Cannot read property ‘active’ of undefined
at Object.<anonymous> (googlesitekit-wp-dashboard.js:41)
at Object.<anonymous> (googlesitekit-wp-dashboard.js:41)
at __webpack_require__ (googlesitekit-wp-dashboard.js:1)
at Object.<anonymous> (googlesitekit-wp-dashboard.js:41)
at __webpack_require__ (googlesitekit-wp-dashboard.js:1)
at Module.<anonymous> (googlesitekit-wp-dashboard.js:41)
at Module.<anonymous> (googlesitekit-wp-dashboard.js:41)
at __webpack_require__ (googlesitekit-wp-dashboard.js:1)
at a (googlesitekit-wp-dashboard.js:1)
at googlesitekit-wp-dashboard.js:1
```

WordPress support topics:
- https://wordpress.org/support/topic/sitekit-is-showing-blank-screen-after-server-migration/ (RESOLVED)
- https://wordpress.org/support/topic/blank-white-screen-23/  (RESOLVED)

Currently troubleshooting and awaiting further information from those impacted

Additional information:
- Dashboard loads as expected for one user when using browser incognito window
- Unable to reproduce so far in support
- Site Health info provided in [one](https://wordpress.org/support/topic/sitekit-is-showing-blank-screen-after-server-migration/) support topic 
- Similar support topics opened during the same version update (v.1.9.0 to 1.10.0). See #1701 and #1704 
---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All JS files Site Kit enqueues should have hashes in their file name to ensure the file name changes when its content changes and avoid "random" JS errors due to aggressive caching.
* File names should be generated in a way so that PHP logic to enqueue them can be aware of the actual names, without needing to inspect the actual files on the filesystem.
* The `ver` query parameter that WordPress adds by default should be omitted for those assets.

## Implementation Brief

[...]

PR #1846, opened 2020-07-29
Bug/1700 assets manifest

eugene-manuilov:
## Summary

Addresses issue #1700  

## Checklist


includes/Core/Assets/Assets.php
```diff
@@ -347,7 +347,7 @@ private function get_assets() {
 		$dependencies = array(
 			'googlesitekit-vendor',
 			'googlesitekit-commons',
-			'googlesitekit-base',
+			'googlesitekit-admin',
```
felixarntz, line 350: This shouldn't be changed to `admin`. Rather, let's rename the actual file `googlesitekit-admin.js` to `googlesitekit-base.js` which was the original plan here.

includes/Core/Assets/Assets.php
```diff
@@ -373,7 +373,7 @@ private function get_assets() {
 				)
 			),
 			new Script_Data(
-				'googlesitekit-base-data',
+				'googlesitekit-admin-data',
```
felixarntz, line 376: See above, this should remain `googlesitekit-base-data`.

includes/Core/Assets/Assets.php
```diff
@@ -433,23 +433,23 @@ private function get_assets() {
 			new Script(
 				'googlesitekit-activation',
 				array(
-					'src'          => $base_url . 'js/googlesitekit-activation.js',
+					'version'      => null,
```
felixarntz, line 436: Instead of setting `version` to `null` here throughout the file, let's handle this in the `Script` class automatically: If a file is in the manifest, no need to specify the version. If it is not in the manifest, we would specify the version.

includes/Core/Assets/Assets.php
```diff
@@ -515,51 +515,51 @@ private function get_assets() {
...
 					'dependencies' => array(
 						'googlesitekit-data',
 					),
 				)
 			),
 			// End JSR Assets.
 			new Script(
-				'googlesitekit-ads-detect',
+				'pagead2.ads',
```
felixarntz, line 526: Let's keep the `googlesitekit-` prefix here, the handle should still be scoped to Site Kit. What we can do instead to make it work is:
* Either add a special case for this one script hard-coded to the `Script` class.
* Or, if the handle does not exist in the manifest, strip the `googlesitekit-` prefix and try lookup again.

includes/Core/Assets/Script.php
```diff
@@ -53,15 +55,23 @@ public function __construct( $handle, array $args ) {
...
 	 */
-	public function register() {
+	public function register( Context $context ) {
 		if ( $this->args['fallback'] && wp_script_is( $this->handle, 'registered' ) ) {
 			return;
 		}
 
+		$src = $this->args['src'];
+		if ( class_exists( '\Google\Site_Kit\Core\Assets\Manifest' ) && isset( Manifest::$assets[ $this->get_handle() ] ) ) {
+			$src = $context->url( 'dist/assets/js/' . Manifest::$assets[ $this->get_handle() ] );
```
felixarntz, line 69: Based on my previous comment: If this condition is fulfilled, let's automatically specify `version` as `null` in the `wp_register_script` call below.

includes/Core/Assets/Assets.php
```diff
@@ -433,23 +433,23 @@ private function get_assets() {
 			new Script(
 				'googlesitekit-activation',
 				array(
-					'src'          => $base_url . 'js/googlesitekit-activation.js',
```
felixarntz, line 436: We should keep these `src` arguments around I'd say. It will be overridden internally by the `Script` class, and having them here makes it more obvious what file it actually is.

includes/Modules/AdSense.php
```diff
@@ -857,13 +857,11 @@ protected function setup_settings() {
...
 	 */
 	protected function setup_assets() {
-		$base_url = $this->context->url( 'dist/assets/' );
-
 		return array(
 			new Script(
 				'googlesitekit-modules-adsense',
 				array(
-					'src'          => $base_url . 'js/googlesitekit-modules-adsense.js',
+					'version'      => null,
```
felixarntz, line 864: See other comment, let's keep `src` in here and omit the `version` argument here. This can all be handled internally in the `Script` class.

tests/phpunit/integration/Core/Assets/AssetsTest.php
```diff
@@ -88,15 +88,15 @@ public function test_register() {
...
 
 		$assets = new Assets( new Context( GOOGLESITEKIT_PLUGIN_MAIN_FILE ) );
 
-		$assets->enqueue_asset( 'googlesitekit-base' );
+		$assets->enqueue_asset( 'googlesitekit-admin' );
 
-		$this->assertTrue( wp_script_is( 'googlesitekit-base', 'registered' ) );
-		$this->assertTrue( wp_script_is( 'googlesitekit-base', 'enqueued' ) );
+		$this->assertTrue( wp_script_is( 'googlesitekit-admin', 'registered' ) );
+		$this->assertTrue( wp_script_is( 'googlesitekit-admin', 'enqueued' ) );
```
felixarntz, line 99: See above, these should be reverted to using `base`.

tests/e2e/specs/modules/analytics/setup-proxy-no-account-no-tag.test.js
```diff
@@ -75,7 +75,7 @@ describe( 'setting up the Analytics module with no existing account and no exist
 		await expect( page ).toClick( '.googlesitekit-analytics__select-country' );
 		await expect( page ).toClick( '.mdc-menu-surface--open li', { text: /united kingdom/i } );
 
-		await expect( page ).toMatchElement( 'p', { text: /need to give Site Kit permission to create an Analytics account/i } );
+		await expect( page ).toMatchElement( 'p', { text: /need to give Site Kit permission to create an Analytics account/i, timeout: 750 } );
```
felixarntz, line 78: That seems unrelated - did that fix a random test failure you noticed while working on this issue?

includes/Core/Assets/Script.php
```diff
@@ -53,15 +55,23 @@ public function __construct( $handle, array $args ) {
...
+	 * @param Context $context Plugin context.
 	 */
-	public function register() {
+	public function register( Context $context ) {
 		if ( $this->args['fallback'] && wp_script_is( $this->handle, 'registered' ) ) {
 			return;
 		}
 
+		$src = $this->args['src'];
+		if ( class_exists( '\Google\Site_Kit\Core\Assets\Manifest' ) && isset( Manifest::$assets[ $this->get_handle() ] ) ) {
```
felixarntz, line 68: We should probably expect this class to exist here, otherwise our plugin is broken.
```suggestion
		if ( isset( Manifest::$assets[ $this->handle ] ) ) {
```

Also, based on my other comment, if this is not found, let's try stripping the `googlesitekit-` prefix and finding it from that.

includes/Core/Assets/Script.php
felixarntz, line 69:
```suggestion
			$src = $context->url( 'dist/assets/js/' . Manifest::$assets[ $this->handle ] );
```

includes/Core/Assets/Assets.php
eugene-manuilov: Updated.

includes/Core/Assets/Assets.php
eugene-manuilov: Updated.

includes/Core/Assets/Assets.php
eugene-manuilov: Updated.

includes/Core/Assets/Script.php
eugene-manuilov: Updated.

includes/Modules/AdSense.php
eugene-manuilov: Updated.

tests/phpunit/integration/Core/Assets/AssetsTest.php
eugene-manuilov: Updated.

includes/Core/Assets/Script.php
eugene-manuilov: Updated.

includes/Core/Assets/Script.php
eugene-manuilov: Updated.

includes/Core/Assets/Assets.php
eugene-manuilov: Reverted.

includes/Core/Assets/Assets.php
eugene-manuilov: Updated.

tests/e2e/specs/modules/analytics/setup-proxy-no-account-no-tag.test.js
eugene-manuilov: Yeah, that helped with a weird issue, but now it seems to be not needed anymore. Reverted.

eugene-manuilov:
@felixarntz updated.

.travis.yml
```diff
@@ -83,6 +83,7 @@ script:
   - |
     if [[ "$E2E" == "1" ]]; then
       npm run build:test || exit 1 # Build for tests.
+      docker run --rm -v "$PWD:/app" -v "$HOME/.cache/composer:/tmp/cache" composer install
```
aaemnnosttv, line 86: @eugene-manuilov Why was this added when the same command is already run for E2E tests above on line `47` in the `before_script`?

eugene-manuilov: We need to regenerate autoload map to include Manifest.php file generated after `npm run build:test`. Probably not the best approach to do it, but it works.

aaemnnosttv: Yeah, worst case we can run `composer run autoload-includes` to only regenerate the Site Kit includes without needing to run all of the prefixing again, but I think we can do better than that.

Issue #1862, opened 2020-07-30
Simplify logic to render Site Kit tags (Analytics, AdSense, Tag Manager) and add actions for loading them


felixarntz:
The logic for Site Kit to render its tags (done by the `Analytics`, `AdSense` and `Tag_Manager` module classes) is currently somewhat hard to follow, and because it renders tags in various formats (e.g. AMP and non-AMP) while still adding all of these hooks independently, there is some duplicate logic, and this has caused errors before where we e.g. changed a condition in one place but not another.

We should consolidate the logic to whether a certain tag needs to be rendered (regardless of AMP or non-AMP) into one place, and then fire an action like `googlesitekit_init_{module}_tag[_amp]`.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Every module that renders tags should have a single action hooked into `template_redirect`, and in that callback it should run the logic to determine whether a tag needs to be rendered, including checking for and determining the correct ID(s) (e.g. property ID / client ID / container ID).
* If the tag will be rendered, the current actions should be hooked in like they are now. However, all duplicate checks which are now handled by the `template_redirect` hook should be removed from the action callbacks. Furthermore, the tag rendering methods should be modified to receive the respective ID as parameter, having them passed from the logic above where they were already determined.
* If the tag will be rendered, an action `googlesitekit_{module}_init_tag[_amp]` should be fired, with the latter `_amp` suffix present or not based on whether this is an AMP request. The relevant ID should be passed to the action.

## Implementation Brief

* Implement approach based on ACs for `Analytics`, `AdSense`, `Tag_Manager` classes:
    * Add action to `template_redirect` in the `register` method.
    * Check all requirements for rendering the tag there.
    * Conditionally add hooks for non-AMP or AMP variant based on `Context::is_amp()`.
        * Update relevant methods used to print tags to have all checks removed, and have them receive the relevant tracking ID in a method parameter; then pass that parameter from the `register` method where it has been detected.
    * Fire new action hooks as defined in the ACs in the respective non-AMP or AMP clause, passing the tracking ID (property ID / client ID / container ID) to it.
* Merge #1863.

## QA Brief

* All Site Kit tags (Analytics, AdSense, Tag Manager) should be tested in detail to ensure that they are still placed in the frontend like before. For each of the three modules:
    * Regular site (no AMP).
    * AMP Standard Mode (with AMP plugin).
[...]

PR #1863, opened 2020-07-31
Enhancement/1862 simplify tag placement logic

felixarntz:
## Summary

Addresses issue #1862

## Relevant technical choices

* Consolidates logic for when to render a tag into a single place for each of the three modules, within the `wp` action, which runs before any HTML is being generated.
* Includes tests that the hooks are added properly via the respective class's `register` method (and its callback added to `wp`).
* Adds action hooks per the ACs.
* Fixes one unrelated issue for Tag Manager: For secondary AMP mode, the module should require both containers to be set to consider itself connected.

## Checklist


tests/phpunit/integration/Modules/Tag_ManagerTest.php
```diff
@@ -36,11 +36,150 @@ public function test_register() {
...
+		$this->assertFalse( has_action( 'amp_print_analytics' ) );
+		$this->assertFalse( has_action( 'wp_footer' ) );
+		$this->assertFalse( has_action( 'amp_post_template_footer' ) );
+		$this->assertFalse( has_filter( 'amp_post_template_data' ) );
+
+		$tagmanager->set_data( 'use-snippet', array( 'useSnippet' => true ) );
+		$tagmanager->set_data(
+			'container-id',
+			array(
+				'containerID'  => '9999999',
```
aaemnnosttv, line 66: `containerID` should be a valid `GTM-` string rather than an integer (unlike `internalContainerID`). I realize it probably doesn't matter for the logic used but we should be consistent in cases where we check with something that should work or not to use valid forms.

includes/Modules/AdSense.php
```diff
@@ -247,6 +267,24 @@ function() use ( $client_id ) {
...
+	 */
+	protected function print_amp_auto_ads( $client_id ) {
+		if ( $this->adsense_tag_printed ) {
+			return;
+		}
+
+		?>
+		<amp-auto-ads type="adsense" data-ad-client="<?php echo esc_attr( $client_id ); ?>"></amp-auto-ads>
+		<?php
+		$this->adsense_tag_printed = true;
```
aaemnnosttv, line 285: Minor detail, but I think it would be a bit cleaner to put this assignment right after the guard that checks it. We do it this way in several other places and it generally makes it a bit easier to read by separating the logic from output. The only reason I think we would want to set it afterwards was if there was the potential of an exception to be thrown during the render (potentially inside an output buffer), but that's not really the case here 🙂

includes/Modules/Analytics.php
```diff
@@ -282,28 +331,11 @@ public function get_debug_fields() {
...
-		}
-
-		$tracking_id = $this->get_data( 'property-id' );
-		if ( is_wp_error( $tracking_id ) ) {
-			return;
-		}
-
+	protected function enqueue_gtag_js( $tracking_id ) {
 		wp_enqueue_script( // phpcs:ignore WordPress.WP.EnqueuedResourceParameters.MissingVersion
 			'google_gtagjs',
```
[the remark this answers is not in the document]
felixarntz: I'd rather not make this change here since it's unrelated and technically breaking. We could open a follow-up issue for that.

includes/Modules/Analytics.php
```diff
@@ -133,10 +105,87 @@ function () {
...
+				if ( is_wp_error( $property_id ) || ! $property_id ) {
+					return;
+				}
+
+				// At this point, we know the tag should be rendered, so let's take care of it
+				// for AMP and non-AMP.
+				if ( $this->context->is_amp() ) {
+					$print_amp_gtag = function() use ( $property_id ) {
+						// This hook is only available in AMP plugin version >=1.3, so if it
+						// has already completed, do nothing.
```
aaemnnosttv, line 132: This is a bit off to read since this callback is used for multiple hooks now.

includes/Modules/Tag_Manager.php
```diff
@@ -85,45 +85,96 @@ final class Tag_Manager extends Module
...
-			// has already completed, do nothing.
-			if ( ! doing_action( 'amp_print_analytics' ) && did_action( 'amp_print_analytics' ) ) {
-				return;
-			}
+				// At this point, we know the tag should be rendered, so let's take care of it
+				// for AMP and non-AMP.
+				if ( $this->context->is_amp() ) {
+					$print_amp_gtm = function() use ( $container_id ) {
+						// This hook is only available in AMP plugin version >=1.3, so if it
+						// has already completed, do nothing.
```
aaemnnosttv, line 117: Same here, this callback is used for multiple actions so this comment isn't entirely accurate.

includes/Modules/Analytics.php
felixarntz: Good point, pushed another commit to change these.

aaemnnosttv: The only reason I think something like that would be necessary is if the AMP plugin printed the GA snippet itself and we needed to check to see if this action was already done to prevent duplicate output, but it looks like the output is only our own. So if that's the case, then this is less verbose and more consistent with the rest 👍

Issue #1743, opened 2020-07-06
Fix first admin usage and introduce concept of ownership


felixarntz:
While the goal of Site Kit is to treat all administrator users as equal in terms of their permissions, it is crucial to have the concept of ownership for certain parts of the plugin, e.g. to indicate that a certain administrator user originally configured it.

As part of this issue, no visual changes should be made yet, but the necessary infrastructure should be put in place to have awareness of owner users. In the future this will be surfaced like e.g.:
* Informing admins when editing module settings who originally set up the module (and therefore has access).
* Warning admins when editing key module settings set up by someone else that they will essentially take over as new module owner.
* Informing users who get Google API-related permission errors for whom they need to ask to get the necessary access.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `core/site` JS data store should expose a `getOwnerID` selector, which returns user ID (integer) of the first administrator that connected overall Site Kit.
    * `core/site/data/connection` should expose a new `ownerID` in API response (see below).
* The existing `googlesitekit_first_admin` option should be renamed to `googlesitekit_owner_id`, and it should be modified to always return an integer (i.e. `0` if there is none set).
    * An upgrade routine should rename the existing option in the database appropriately.
    * The logic to set the owner (currently happens on the fly when loading legacy JS data in `Authentication::inline_js_setup_data()`, which is a bad practice) should be moved to after having received an access token (`OAuth_Client::authorize_user()`), and of course it should only happen if the user can `Permissions::MANAGE_OPTIONS` and if there is no other user set yet, with the following caveat:
        * If there is already an owner set and it is a different user, check if that user still exists, and if they do, whether they can still `Permissions::MANAGE_OPTIONS`.  
        If one of these requirements is no longer satisfied, the current user should implicitly take over as the new Site Kit owner.
* Site Kit modules with editable settings which have Google API-side access implications should receive an `ownerID` (integer) sub setting:
    * `Analytics` (via account ID, property ID, profile ID)
    * `AdSense` (via account ID, client ID)
    * `TagManager` (via account ID, container ID, AMP container ID)
    * `Optimize` (via Optimize ID)
    * Whenever one of the above settings is **changed** for the respective module, the owner ID should be set to the current user ID.
    * The datastores of these modules should expose a `getOwnerID` selector.

## Implementation Brief

* Update `assets/js/googlesitekit/datastore/site/connection.js`
[...]

PR #1866, opened 2020-08-03
Enhancement/1743 ownership

eugene-manuilov:
## Summary

Addresses issue #1743

## Relevant technical choices

* Added `isOwner` property to the `_googlesitekitBaseData` data and updated tracking functionality to use that property instead of `isFirstAdmin`.

## Checklist


tests/phpunit/integration/Modules/AdSense/SettingsTest.php
```diff
@@ -59,6 +59,7 @@ public function test_get_default() {
 				'accountSetupComplete' => false,
 				'siteSetupComplete'    => false,
 				'useSnippet'           => true,
+				'ownerID'              => 0,
```
aaemnnosttv, line 62: We should also test that the `ownerID` is set and updated accordingly when the owned keys are added/changed.

Rather than duplicate these for all module settings tests that have owned keys, we should add a contract test  trait that we can `use` in the respective `ModuleSettings` sub-class test cases. See https://github.com/google/site-kit-wp/blob/develop/tests/phpunit/includes/Core/Modules/Module_With_Settings_ContractTests.php for an example of one of these traits.

tests/phpunit/integration/Core/Storage/Owner_IDTest.php
```diff
@@ -0,0 +1,69 @@
+<?php
+/**
+ * Class Google\Site_Kit\Core\Storage\Owner_IDTest
+ *
+ * @package   Google\Site_Kit
+ * @copyright 2019 Google LLC
```
aaemnnosttv, line 6:
```suggestion
 * @copyright 2020 Google LLC
```

assets/js/googlesitekit/datastore/site/connection.js
```diff
@@ -84,6 +84,20 @@ const baseSelectors = {
 		return connection;
 	},
 
+	/**
+	 * Gets owner ID.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {string|undefined} Owner ID if it exists, otherwise undefined.
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/googlesitekit/datastore/site/connection.test.js
```diff
@@ -36,6 +36,7 @@ describe( 'core/site connection', () => {
 		resettable: true,
 		setupCompleted: true,
 		hasConnectedAdmins: true,
+		ownerID: '123',
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

includes/Core/Authentication/Clients/OAuth_Client.php
```diff
@@ -681,6 +692,15 @@ public function authorize_user() {
 			exit();
 		}
 
+		$current_user_id  = get_current_user_id();
+		$current_owner_id = $this->owner_id->get();
+		if (
+			( empty( $current_owner_id ) && current_user_can( Permissions::MANAGE_OPTIONS ) ) ||
+			( ! empty( $current_owner_id ) && $current_owner_id != $current_user_id && ! user_can( $current_owner_id, Permissions::MANAGE_OPTIONS ) )
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

includes/Core/Modules/Module_With_Owner_Trait.php
```diff
@@ -0,0 +1,47 @@
...
+ * @access private
+ * @ignore
+ */
+trait Module_With_Owner_Trait {
+
+	/**
+	 * Owner_ID instance.
+	 *
+	 * @since n.e.x.t
+	 * @var Owner_ID|null
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

includes/Core/Storage/Owner_ID.php
```diff
@@ -0,0 +1,78 @@
...
+	/**
+	 * Gets the value of the setting.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return int Value set for the option, or registered default if not set.
+	 */
+	public function get() {
+		$owner_id = parent::get();
+		if ( false === $owner_id ) {
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

includes/Core/Storage/Owner_ID.php
```diff
@@ -0,0 +1,78 @@
...
+
+	/**
+	 * Gets the default value.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return int The default value.
+	 */
+	protected function get_default() {
+		return intval( get_option( 'googlesitekit_first_admin' ) );
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

includes/Core/Storage/Setting_With_Owned_Keys_Trait.php
```diff
@@ -0,0 +1,60 @@
...
+			10,
+			2
+		);
+
+		add_filter(
+			'pre_update_option_' . static::OPTION,
+			function( $value, $old_value ) use ( $keys ) {
+				if ( is_array( $value ) && is_array( $old_value ) ) {
+					foreach ( $keys as $key ) {
+						if ( isset( $value[ $key ] ) && isset( $old_value[ $key ] ) && $value[ $key ] !== $old_value[ $key ] ) {
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

tests/phpunit/integration/Core/Storage/Owner_IDTest.php
```diff
@@ -0,0 +1,69 @@
...
+	/**
+	 * Owner_ID object.
+	 *
+	 * @var Owner_ID
+	 */
+	private $owner_id;
+
+	/**
+	 * Set Up Test.
+	 */
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

tests/phpunit/integration/Core/Storage/Owner_IDTest.php
eugene-manuilov: Updated.

tests/phpunit/integration/Core/Storage/Owner_IDTest.php
```diff
@@ -0,0 +1,69 @@
...
+		$this->owner_id = new Owner_ID( $options );
+	}
+
+	/**
+	 * Test get() method.
+	 */
+	public function test_get() {
+		$this->assertEquals( 0, $this->owner_id->get() );
+
+		$this->owner_id->set( 1 );
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

tests/phpunit/integration/Core/Storage/Owner_IDTest.php
```diff
@@ -0,0 +1,69 @@
...
+		$this->assertEquals( 1, $this->owner_id->get() );
+	}
+
+	/**
+	 * Test set() method.
+	 */
+	public function test_set() {
+		$this->assertTrue( $this->owner_id->set( 1 ) );
+		$this->assertEquals( 1, $this->owner_id->get() );
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

tests/phpunit/integration/Core/Storage/Owner_IDTest.php
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	/**
+	 * Test has() method.
+	 */
+	public function test_has() {
+		$this->assertFalse( $this->owner_id->has() );
+
+		$this->owner_id->set( 1 );
+		$this->assertTrue( $this->owner_id->has() );
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

tests/phpunit/integration/Modules/AdSense/SettingsTest.php
eugene-manuilov: Updated, but decided not to use trait since tests for different modules rely on different keys.

aaemnnosttv: After looking over all the added tests here, I still think it makes sense to use an abstraction for these tests since the tests are essentially all the same with different keys. Using an abstraction will allow us to provide more robust coverage (i.e. test every key) with less code, and a single place to manage it.

The test trait can have an abstract `get_owned_keys` method which returns the list of keys which are owned. In order to use the test trait, those keys need to be provided. The non-owned keys would be an array diff of this array with the keys from the setting's default value (whatever comes back from `get()`, the keys will always be the same). One alternative to having this method on the test trait would be to add the `get_owned_keys` method to the `Module_With_Owned_Keys` interface and that would make them easily available in the test as well, although I'm not sure if @felixarntz would approve of this change.

Let's implement the `get_owned_keys` on the test trait first, and if Felix is okay with moving it to the interface, then that is a very simple refactor to make later.

includes/Core/Authentication/Clients/OAuth_Client.php
```diff
@@ -681,6 +692,11 @@ public function authorize_user() {
 			exit();
 		}
 
+		$current_user_id = get_current_user_id();
+		if ( $this->is_owner_must_be_changed( $this->owner_id->get(), $current_user_id ) ) {
+			$this->owner_id->set( $current_user_id );
+		}
```
aaemnnosttv, line 698: The placement here seems to be a bit arbitrary, and separates a check for the access token and the method that sets the access token which make sense to keep together. Let's move this down between the calls to `set_granted_scopes` and `refresh_profile_data`.

Also, we need to add a test for this to `OAuth_ClientTest`. Since the `test_authorize_user` method is already quite large, let's add a test for the new `should_update_owner_id` method directly. Ideally we wouldn't do that because it's private, but in this case I think it's actually cleaner.

tests/phpunit/integration/Modules/AdSense/SettingsTest.php
felixarntz: I think we can expose these keys via the `Setting_With_Owned_Keys_Trait`, and let's introduce a `Setting_With_Owned_Keys_Interface` for that purpose, which exposes the public `get_owned_keys` method.

aaemnnosttv: Thanks for your input @felixarntz – I mixed up the interface name above; +1 on adding a `Setting_With_Owned_Keys_Interface`

There is the `Module_With_Owner` interface introduced in this PR along with the `Module_With_Owner_Trait` but neither are actually used yet.

> Regarding the trait, we should probably also make the name of the `ownerID` key a parameter so that it can be controlled by the class using the trait.

Sounds good. How about we make it an optional parameter to `Setting_With_Owned_Keys_Trait::register_owned_keys` with a default value of `ownerID` since we don't have any cases where it needs to be different?

@eugene-manuilov to summarize:
* The relevant module sub-classes need to implement the `Module_With_Owner` interface and `use` the `Module_With_Owner_Trait`
    - A `\Google\Site_Kit\Tests\Core\Modules\Module_With_Owner_ContractTests` trait should be added for adding test coverage to a Module sub-class' test case, and `use`d by the respective `Module` sub-class' test cases
* A `Setting_With_Owned_Keys_Interface` should be added with `get_owned_keys() : string[]` and the same module settings classes that use the trait should implement this interface as well
    - The contract tests trait for `Setting_With_Owned_Keys_ContractTests` will now have access to the list of owned keys directly from the class under test via the `get_owned_keys` method (rather than putting this on the test trait itself as mentioned previously)
* `Setting_With_Owned_Keys_Trait::register_owned_keys` should accept an optional `$ownerID_key = 'ownerID'` as the second parameter.

includes/Core/Storage/Owner_ID.php
```diff
@@ -0,0 +1,96 @@
...
+	 * @return bool True on success, false on failure.
+	 */
+	public function set( $value ) {
+		return parent::set(
+			call_user_func(
+				$this->get_sanitize_callback(),
+				$value
+			)
+		);
+	}
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

includes/Core/Storage/Owner_ID.php
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * @package   Google\Site_Kit\Core\Storage
+ * @copyright 2020 Google LLC
+ * @license   https://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
+ * @link      https://sitekit.withgoogle.com
+ */
+
+namespace Google\Site_Kit\Core\Storage;
+
+use Google\Site_Kit\Core\Storage\Setting;
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

includes/Core/Storage/Owner_ID.php
```diff
@@ -0,0 +1,96 @@
...
+			call_user_func(
+				$this->get_sanitize_callback(),
+				$value
+			)
+		);
+	}
+
+	/**
+	 * Gets the default value using legacy "googlesitekit_first_admin" option. If
+	 * this option doesn't exist, this function will return 0.
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

includes/Core/Authentication/Clients/OAuth_Client.php
```diff
@@ -752,6 +768,31 @@ function( $scope ) {
...
+	/**
+	 * Determines whether the current owner ID must be changed or not.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param int $current_owner_id Current owner ID.
+	 * @param int $current_user_id Current user ID.
+	 * @return bool TRUE if owner needs to be changed, otherwise FALSE.
+	 */
+	private function is_owner_must_be_changed( $current_owner_id, $current_user_id ) {
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

includes/Core/Authentication/Clients/OAuth_Client.php
```diff
@@ -752,6 +768,31 @@ function( $scope ) {
...
+		if ( $current_owner_id == $current_user_id ) {
+			return false;
+		}
+
+		if ( ! empty( $current_owner_id ) && user_can( $current_owner_id, Permissions::MANAGE_OPTIONS ) ) {
+			return false;
+		}
+
+		if ( ! current_user_can( Permissions::MANAGE_OPTIONS ) ) {
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

tests/phpunit/integration/Core/Storage/Owner_IDTest.php
```diff
@@ -0,0 +1,77 @@
...
+	 *
+	 * @var Options
+	 */
+	private $options;
+
+	public function setUp() {
+		parent::setUp();
+
+		$this->options  = new Options( new Context( GOOGLESITEKIT_PLUGIN_MAIN_FILE ) );
+		$this->owner_id = new Owner_ID( $this->options );
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

tests/phpunit/integration/Core/Storage/Owner_IDTest.php
```diff
@@ -0,0 +1,77 @@
...
+		$this->assertTrue( is_int( $this->owner_id->get() ) );
+		$this->assertEquals( 0, $this->owner_id->get() );
+	}
+
+	public function test_set() {
+		$this->assertTrue( $this->owner_id->set( 1 ) );
+		$this->assertTrue( is_int( $this->options->get( Owner_ID::OPTION ) ) );
+		$this->assertEquals( 1, $this->options->get( Owner_ID::OPTION ) );
+
+		$this->owner_id->set( 'xxx' );
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

tests/phpunit/integration/Modules/AdSense/SettingsTest.php
eugene-manuilov: Added.

tests/phpunit/integration/Modules/AdSense/SettingsTest.php
```diff
@@ -106,6 +107,56 @@ public function test_legacy_options() {
...
+			$options[ $key ] = $value;
+			$settings->set( $options );
+
+			$options = get_option( Settings::OPTION );
+			$this->assertEquals( $user_id, $options['ownerID'] );
+		}
+	}
+
+	public function test_owner_id_is_not_set() {
+		$user_id = $this->factory()->user->create( array( 'role' => 'administrator' ) );
```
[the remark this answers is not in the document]
eugene-manuilov: Added.

tests/phpunit/integration/Modules/AdSense/SettingsTest.php
```diff
@@ -106,6 +107,56 @@ public function test_legacy_options() {
...
+			array(
+				'siteStatus' => 'old-site-status',
+			)
+		);
+
+		$settings = new Settings( new Options( new Context( GOOGLESITEKIT_PLUGIN_MAIN_FILE ) ) );
+		$settings->register();
+
+		$options = $settings->get();
+		$this->assertTrue( empty( $options['ownerID'] ) );
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

tests/phpunit/integration/Modules/Analytics/SettingsTest.php
```diff
@@ -231,6 +232,57 @@ public function test_adsense_linked_is_true_if_adsense_is_active_and_connected_o
...
+	}
+
+	public function test_owner_id_is_not_set() {
+		$user_id = $this->factory()->user->create( array( 'role' => 'administrator' ) );
+		wp_set_current_user( $user_id );
+
+		add_option(
+			Settings::OPTION,
+			array(
+				'useSnippet' => 'old-use-snippet',
```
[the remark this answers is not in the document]
eugene-manuilov: Irrelivant anymore.

tests/phpunit/integration/Modules/AdSense/SettingsTest.php
```diff
@@ -106,6 +107,56 @@ public function test_legacy_options() {
...
+	}
+
+	public function test_owner_id_is_not_set() {
+		$user_id = $this->factory()->user->create( array( 'role' => 'administrator' ) );
+		wp_set_current_user( $user_id );
+
+		add_option(
+			Settings::OPTION,
+			array(
+				'siteStatus' => 'old-site-status',
```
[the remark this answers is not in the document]
eugene-manuilov: Unnecessary anymore.

includes/Core/Authentication/Clients/OAuth_Client.php
eugene-manuilov: Updated.

tests/phpunit/includes/Core/Storage/Setting_With_Owned_Keys_ContractTests.php
```diff
@@ -0,0 +1,125 @@
...
+		$options_key = $testcase->get_option_name();
+		$fields      = $settings->get_owned_keys();
+		foreach ( $fields as $field ) {
+			delete_option( $options_key );
+
+			$options = $settings->get();
+			$testcase->assertEmpty( $options['ownerID'] );
+
+			$options[ $field ] = 'test-value';
+			$settings->set( $options );
```
aaemnnosttv, line 55: I thought this might be a problem due to differences in the `sanitize_callback`s for the different module settings, but apparently it works, so that works for me 😄

tests/phpunit/includes/Core/Modules/Module_With_Owner_ContractTests.php
```diff
@@ -0,0 +1,32 @@
...
+	abstract protected function get_module_with_owner();
+
+	public function test_owner_id() {
+		$testcase = $this->get_testcase();
+		$module   = $this->get_module_with_owner();
+
+		// By default, ownerID is 0.
+		$testcase->assertEquals( 0, $module->get_owner_id() );
+	}
+}
```
aaemnnosttv, line 32: Let's add another test or two here to cover a non-zero return value as well. To do that, we'll need to reach slightly beyond the `Module_With_Owner` interface alone, but that's okay as it's all within the context of the implicit contract.

In order to test this, we can assert that `$module->get_settings()` is an instance of `Setting_With_Owned_Keys_Interface`. We need to register the settings and then we can call `$settings->get_owned_keys()` and change one of them with the current user set to an admin with permissions. Then we can check that `$module->get_owner_id()` returns the expected user ID.

There isn't really a cleaner way to set the module's owner ID, but this makes the test a little less fragile to change in the future.

We don't need to be as thorough here and test each owned key and all the different possible combinations of permissions here as that's all well tested in the other tests. We just want to trigger the value to change and to ensure the new value comes back as expected.

includes/Core/Storage/Setting_With_Owned_Keys_Trait.php
```diff
@@ -24,14 +26,15 @@ trait Setting_With_Owned_Keys_Trait {
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @param array $keys An array of keys to check.
+	 * @param array  $keys An array of keys to check.
+	 * @param string $ownerID_key Optional. A key to use for storing owner ID setting.
```
aaemnnosttv, line 30: The descriptions should also be aligned; unfortunately PHPCS does not enforce this yet. Using the new name for the second parameter:
```suggestion
	 * @param array  $keys         An array of keys to check.
	 * @param string $owner_id_key Optional. The key to use for storing the owner ID.
```

aaemnnosttv (commented):
Almost done here, this should be the last few things to address here and then this should be good to go 👍  Looking good though!

includes/Core/Storage/Setting_With_Owned_Keys_Trait.php
```diff
@@ -24,14 +26,15 @@ trait Setting_With_Owned_Keys_Trait {
...
 	 */
-	protected function register_owned_keys( array $keys ) {
+	protected function register_owned_keys( array $keys, $ownerID_key = 'ownerID' ) { // phpcs:ignore WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

tests/phpunit/integration/Core/Authentication/Clients/OAuth_ClientTest.php
```diff
@@ -414,6 +417,47 @@ function ( Request $request ) {
 		$this->assertEquals( 'https://example.com/fresh.jpg', $profile['photo'] );
 	}
 
+	public function test_should_update_owner_id() {
+		$user_id_1 = $this->factory()->user->create( array( 'role' => 'administrator' ) );
+		$user_id_2 = $this->factory()->user->create( array( 'role' => 'administrator' ) );
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

includes/Core/Storage/Setting_With_Owned_Keys_Trait.php
eugene-manuilov: Updated.

tests/phpunit/integration/Core/Authentication/Clients/OAuth_ClientTest.php
```diff
@@ -414,6 +417,47 @@ function ( Request $request ) {
...
+		wp_set_current_user( $user_id_1 );
+
+		$context = new Context( GOOGLESITEKIT_PLUGIN_MAIN_FILE );
+		$options = new Options( new Context( GOOGLESITEKIT_PLUGIN_MAIN_FILE ) );
+		$client  = new OAuth_Client( $context, $options );
+
+		$class  = new \ReflectionClass( OAuth_Client::class );
+		$method = $class->getMethod( 'should_update_owner_id' );
+		$method->setAccessible( true );
+
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

tests/phpunit/includes/Core/Modules/Module_With_Owner_ContractTests.php
eugene-manuilov: Actually, i am not sure that it is going to work at all because Module_With_Owner_Trait uses Owner_ID setting that pulls owner ID from `googlesitekit_owner_id` and has nothing to do with ownerID from module settings. I think we don't need `Module_With_Owner_Trait` or we need to make it work differently.

aaemnnosttv: Ah, I see what you mean; I missed that in my initial review 😞 . The `Module_With_Owner_Trait::get_owner_id` method should return the `ownerID` from the module's settings as described in the IB – not the `googlesitekit_owner_id` which is the "site" owner and may be different than a module's owner. 

Regarding the `Module_With_Owner_Trait`, this is a bit complicated however considering the flexibility of the key used for holding the `ownerID`. `get_owner_id` essentially needs to return `$this->get_settings()->get()['ownerID']`.

@felixarntz I think perhaps we should remove the `$owner_id_key` parameter to `register_owned_keys` as it makes it complicated for other parts to account for this variable key and we don't really need it for the time being. What do you think?

assets/js/util/tracking/createTrackEvent.js
```diff
@@ -43,7 +43,7 @@ export default function createTrackEvent( config, dataLayerTarget ) {
 			event_label: eventLabel,
 			event_value: eventValue,
 			dimension1: referenceSiteURL,
-			dimension2: isFirstAdmin ? 'true' : 'false',
+			dimension2: isOwner ? 'true' : 'false',
 			dimension3: userIDHash,
 		};
```
felixarntz, line 48: We may need to make some changes because the dimension here is still expected to send whether the user is the "first admin", not whether they are an owner, which is slightly different. We may want to revisit whether this still even makes sense, but for now I think we're better off maintaining the original behavior here.

I suggest re-introducing the `isFirstAdmin` field specifically for `_googlesitekitBaseData` below and referencing that here. (**Only** here, everywhere else we should look at the new `isOwner` as is currently happening.)

assets/js/util/tracking/createTracking.js
```diff
@@ -6,7 +6,7 @@ import createEnableTracking from './createEnableTracking';
 import createTrackEvent from './createTrackEvent';
 
 const DEFAULT_CONFIG = {
-	isFirstAdmin: false,
+	isOwner: false,
```
felixarntz, line 9: See above, this should probably remain `isFirstAdmin`.

assets/js/util/tracking/index.js
```diff
@@ -22,7 +22,7 @@
 import createTracking from './createTracking';
 
 const {
-	isFirstAdmin,
+	isOwner,
```
felixarntz, line 25: See above.

assets/js/util/tracking/index.js
```diff
@@ -31,7 +31,7 @@ const {
 } = global._googlesitekitBaseData || {};
 
 const initialConfig = {
-	isFirstAdmin,
+	isOwner,
```
felixarntz, line 34: See above.

assets/js/util/tracking/index.test.js
```diff
@@ -72,7 +72,7 @@ describe( 'trackEvent', () => {
 			referenceSiteURL: 'https://www.example.com/',
 			userIDHash: 'a1b2c3',
 			trackingID: 'UA-12345678-1',
-			isFirstAdmin: true,
+			isOwner: true,
```
felixarntz, line 75: See above.

includes/Core/Authentication/Authentication.php
```diff
@@ -602,16 +604,8 @@ private function handle_oauth() {
...
-		$current_user_id = get_current_user_id();
-
-		// If no first admin is stored yet and the current user is one, consider them the first.
-		if ( ! $first_admin_id && current_user_can( Permissions::MANAGE_OPTIONS ) ) {
-			$first_admin_id = $current_user_id;
-		}
-		$data['isFirstAdmin'] = ( $current_user_id === $first_admin_id );
-		$data['splashURL']    = esc_url_raw( $this->context->admin_url( 'splash' ) );
-
+		$data['isOwner']             = $this->owner_id->get() === get_current_user_id();
```
felixarntz, line 607: See above: Here we should for now re-include `isFirstAdmin`, solely to keep the tracking logic behave like it did before. The ACs may have been somewhat unprecise here - the primary goal was to avoid actually _setting_ the owner here (persistently), which is happening in the similar logic below for `_googlesitekitLegacyData`.
```suggestion
		$data['isOwner']             = $this->owner_id->get() === get_current_user_id();
		$data['isFirstAdmin']        = $data['isOwner'] || ! $this->owner_id->get() && current_user_can( Permissions::MANAGE_OPTIONS );
```

includes/Core/Authentication/Authentication.php
```diff
@@ -676,15 +670,6 @@ private function inline_js_setup_data( $data ) {
...
 		}
 
-		// Flag the first admin user.
-		$first_admin_id  = (int) $this->first_admin->get();
-		$current_user_id = get_current_user_id();
-		if ( ! $first_admin_id && current_user_can( Permissions::MANAGE_OPTIONS ) ) {
-			$first_admin_id = $current_user_id;
-			$this->first_admin->set( $first_admin_id );
-		}
-		$data['isFirstAdmin'] = ( $current_user_id === $first_admin_id );
```
felixarntz, line 686: 👍  The most important thing to remove here was setting the actual option.

includes/Core/Authentication/Clients/OAuth_Client.php
```diff
@@ -859,6 +875,31 @@ public function get_proxy_setup_url( $access_code = '', $error_code = '' ) {
...
+		if ( ! empty( $current_owner_id ) && user_can( $current_owner_id, Permissions::MANAGE_OPTIONS ) ) {
+			return false;
+		}
+
+		if ( ! user_can( $user_id, Permissions::MANAGE_OPTIONS ) ) {
+			return false;
+		}
+
+		return true;
+	}
```
felixarntz, line 901: 👍

includes/Core/Storage/Owner_ID.php
```diff
@@ -0,0 +1,79 @@
+<?php
+/**
+ * Class Google\Site_Kit\Core\Storage\Owner_ID
```
felixarntz, line 3: Let's move this class to `Google\Site_Kit\Core\Authentication\Owner_ID`, similar to where the other authentication-related settings live.

I noticed the other day that the new `Has_Connected_Admins` class was also added to the `Storage` namespace, which was an oversight before, I don't think that should be the case. The `Storage` namespace solely contains base classes, and I think it makes sense to continue to do so.

includes/Core/Storage/Owner_ID.php
```diff
@@ -0,0 +1,79 @@
...
+	 *
+	 * @return int Value set for the option, or registered default if not set.
+	 */
+	public function get() {
+		$owner_id = parent::get();
+		if ( ! $this->has() ) {
+			$owner_id = $this->get_default();
+		}
+
+		return intval( $owner_id );
```
felixarntz, line 51: Couldn't we simplify this to:
```suggestion
		return (int) parent::get();
```

The default should be handled by WordPress already, via `register_setting`.

includes/Core/Storage/Owner_ID.php
```diff
@@ -0,0 +1,96 @@
...
+	/**
+	 * Gets the default value using legacy "googlesitekit_first_admin" option. If
+	 * this option doesn't exist, this function will return 0.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return int The default value.
+	 */
+	protected function get_default() {
+		return intval( get_option( 'googlesitekit_first_admin' ) );
```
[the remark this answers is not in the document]
felixarntz: @aaemnnosttv Good point:
```suggestion
		return (int) $this->options->get( 'googlesitekit_first_admin' );
```

includes/Core/Storage/Setting_With_Owned_Keys_Trait.php
```diff
@@ -0,0 +1,63 @@
...
+ */
+trait Setting_With_Owned_Keys_Trait {
+
+	/**
+	 * Registers hooks to determine an owner ID for a module.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array  $keys                   An array of keys to check.
+	 * @param string $owner_id_key Optional. The key to use for storing owner ID setting.
```
felixarntz, line 30:
```suggestion
	 * @param array  $keys         An array of keys to check.
	 * @param string $owner_id_key Optional. The key to use for storing owner ID setting. Default 'ownerID'.
```

includes/Core/Storage/Setting_With_Owned_Keys_Trait.php
```diff
@@ -0,0 +1,63 @@
...
+	 */
+	protected function register_owned_keys( array $keys, $owner_id_key = 'ownerID' ) {
```
felixarntz, line 32: Would it be possible to implement `get_owned_keys` as a public abstract method here and then override in the respective classes using the trait? Potentially we could simplify the `register_owned_keys` signature to only support the optional `$owner_id_key` argument, and it would rely on calling `get_owned_keys` as implemented by the class.

It would simplify using this trait.

includes/Modules/AdSense/Settings.php
```diff
@@ -88,6 +90,8 @@ public function register() {
 			)
 		);
 
+		$this->register_owned_keys( $this->get_owned_keys() );
```
felixarntz, line 93: See my last comment, if we changed that, here (and below) we would only need to call `$this->register_owned_keys()`.

includes/Modules/Analytics/Settings.php
```diff
@@ -127,6 +131,21 @@ function ( $option ) {
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array An array of keys for owned settings.
+	 */
+	public function get_owned_keys() {
+		return array(
+			'accountID',
+			'propertyID',
+			'profileID',
```
felixarntz, line 145: Nit-picking, this wasn't part of the ACs and technically shouldn't be a problem since it's tied to `propertyID`, but we probably should also include `internalWebPropertyID` here.

includes/Modules/Tag_Manager/Settings.php
```diff
@@ -41,6 +43,23 @@ public function register() {
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array An array of keys for owned settings.
+	 */
+	public function get_owned_keys() {
+		return array(
+			'accountID',
+			'containerID',
+			'ampContainerID',
```
felixarntz, line 61: See above, it may be worth including `internalContainerID` and `internalAMPContainerID` here too.

includes/Core/Modules/Module_With_Owner_Trait.php
```diff
@@ -0,0 +1,47 @@
...
+	 * @since n.e.x.t
+	 *
+	 * @return int Owner ID.
+	 */
+	public function get_owner_id() {
+		if ( is_null( $this->owner_id ) ) {
+			$this->owner_id = new Owner_ID( $this->options );
+		}
+
+		return $this->owner_id->get();
```
felixarntz, line 44: Per @aaemnnosttv's comment further down:
```suggestion
		if ( ! $this instanceof Module_With_Settings ) {
			return 0;
		}
		$settings = $this->get_settings()->get();
		if ( empty( $settings['ownerID'] ) ) {
			return 0;
		}
		return $settings['ownerID'];
```

tests/phpunit/includes/Core/Modules/Module_With_Owner_ContractTests.php
felixarntz: @aaemnnosttv I originally suggested introducing that key, but that's a fair point. In combination with my comment further above, this would furthermore allow us to simplify `register_owned_keys` as a parameter-less method, so +1 from my end.

includes/Core/Modules/Module_With_Owner_Trait.php
```diff
@@ -0,0 +1,47 @@
...
+ */
+trait Module_With_Owner_Trait {
+
+	/**
+	 * Owner_ID instance.
+	 *
+	 * @since n.e.x.t
+	 * @var Owner_ID
+	 */
+	protected $owner_id;
```
felixarntz, line 30: See other comments, this property can be removed.

includes/Core/Storage/Owner_ID.php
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/util/tracking/createTrackEvent.js
eugene-manuilov: Updated.

assets/js/util/tracking/createTracking.js
eugene-manuilov: Updated.

assets/js/util/tracking/index.js
eugene-manuilov: Updated.

assets/js/util/tracking/index.js
eugene-manuilov: Updated.

assets/js/util/tracking/index.test.js
eugene-manuilov: Updated.

includes/Core/Authentication/Authentication.php
eugene-manuilov: Updated.

includes/Core/Storage/Owner_ID.php
eugene-manuilov: Actually, we don't need to override `get` at all. Removed.

includes/Core/Storage/Setting_With_Owned_Keys_Trait.php
eugene-manuilov: Updated.

includes/Core/Storage/Setting_With_Owned_Keys_Trait.php
eugene-manuilov: Yep, we can do it. Updated.

includes/Modules/AdSense/Settings.php
eugene-manuilov: Updated.

includes/Modules/Analytics/Settings.php
eugene-manuilov: Added.

includes/Modules/Tag_Manager/Settings.php
eugene-manuilov: Added.

includes/Core/Modules/Module_With_Owner_Trait.php
eugene-manuilov: Yep, updated.

includes/Core/Modules/Module_With_Owner_Trait.php
eugene-manuilov: Removed.

includes/Core/Storage/Owner_ID.php
eugene-manuilov: Okay, moved Owner_ID and Has_Connected_Admins to the Authentication namespace.

tests/phpunit/includes/Core/Modules/Module_With_Owner_ContractTests.php
eugene-manuilov: Okay, it should be good now.

.storybook/config.js
```diff
@@ -86,7 +86,7 @@ global._googlesitekitBaseData = {
 	blogPrefix: 'wp_',
 	ampMode: false,
 	isNetworkMode: false,
-	isFirstAdmin: true,
```
felixarntz, line 89: This should probably be kept since it's currently included in `_googlesitekitBaseData`.

includes/Core/Authentication/Authentication.php
```diff
@@ -12,11 +12,12 @@
 
 use Google\Site_Kit\Context;
 use Google\Site_Kit\Core\Authentication\Clients\OAuth_Client;
+use Google\Site_Kit\Core\Authentication\Has_Connected_Admins;
+use Google\Site_Kit\Core\Authentication\Owner_ID;
```
felixarntz, line 16: Nit-picking, but these are not necessary because it's the same namespace.

includes/Core/Authentication/Owner_ID.php
```diff
@@ -0,0 +1,65 @@
...
+ * @access private
+ * @ignore
+ */
+class Owner_ID extends Setting {
+
+	/**
+	 * The option_name for this setting.
+	 */
+	const OPTION = 'googlesitekit_owner_id';
+
```
felixarntz, line 28: Maybe we should keep overriding the `get` method in this class, simply calling `parent::get()` and casting to `int`. WordPress saves this as a string in the DB no matter whether we cast it to an integer before, so I think overriding makes sense to ensure `get` returns an integer.

aaemnnosttv: I think you're right. I thought the sanitize callback was enough but I confirmed that if the value is loaded from the DB that it will return a string. I'll update method and tests to cover this.

tests/phpunit/includes/Core/Modules/Module_With_Owner_ContractTests.php
```diff
@@ -0,0 +1,71 @@
...
+		$testcase->assertEquals( 0, $module->get_owner_id() );
+
+		if ( $module instanceof Module_With_Settings ) {
+			$settings = $module->get_settings();
+
+			if ( $settings instanceof Setting ) {
+				$settings->register();
+			}
+
+			if ( $settings instanceof Setting_With_Owned_Keys_Interface ) {
```
aaemnnosttv, line 43: I realize the `instanceof` checks make the IDE happy here, but these contract tests should cause test failures if used for testing modules that don't implement these things. That's part of the benefit of these tests as they can enforce aspects of the contract which are not possible to do via PHP alone.

So rather than using `instanceof` with an `if` and nesting, let's simply assert `$testcase->assertInstanceOf( Module_With_Settings::class, $module );`. We can assume `get_settings` returns an instance of `Setting` since it returns a `Module_Settings`. Similarly, we can assert `$testcase->assertInstanceOf( Setting_With_Owned_Keys_Interface::class, $settings );` rather than nesting there as well, and now we have a nice flat method.

Again, it's not just about formatting, but that these tests should fail if the types are not implemented correctly, either because the test was used on a test case where it shouldn't be, or because something was done improperly and needs to be fixed for the test to pass 🙂

aaemnnosttv (commented):
@eugene-manuilov I pushed a few commits mostly because I updated a test to verify @felixarntz 's comments regarding the return value of `Owner_ID::get` potentially returning a non-integer which turned out to be true. I addressed the few other minor comments he had while I was at it.

I just have one remaining comment that I left for you regarding `Module_With_Owner_ContractTests` and then we can finally get this merged 😄

aaemnnosttv:
@googlebot I consent.

includes/Core/Authentication/Authentication.php
eugene-manuilov: Updated.

includes/Core/Authentication/Owner_ID.php
eugene-manuilov: Ok. it's updated.

tests/phpunit/includes/Core/Modules/Module_With_Owner_ContractTests.php
eugene-manuilov: Updated.

aaemnnosttv:
This is looking good, but VRT is failing consistently right now with one size from the "Global / Admin Bar" story

```
compare | ERROR { requireSameDimensions: false, size: isDifferent, content: 0.00%, threshold: 0.05% }: Global/Admin Bar google-site-kit_Global_Admin_Bar_0_document_0_small.png
```

It seems to be indicating that the sizes don't match rather than any difference in the content of the comparison.

I've tried running this locally and it passes for me so this is especially strange.

aaemnnosttv (approved):
Thanks for your patience on this one @eugene-manuilov 👍

Issue #1714, opened 2020-06-25
Improve messaging to connect versus configure Site Kit for site admins who have not originally done the Site Kit configuration


sonjaleix:
## Bug Description
I'm one of multiple admins on a Newspack site using Site Kit. The other admin has originally configured Site Kit and connected it to the Newspack site. When I now go to the Site Kit UI in the backend of WordPress, the messaging seems to indicate that Site Kit has not yet been set up – when in fact what I need to do is connect with my Google account to the site (this doesn't feel very intuitive and the messaging doesn't indicate what needs to happen).  

The disconnects and confusion I experience: 
A. The first time I (a site admin who was not the person who configured Site Kit) go to Site Kit, I see messaging that says "Configure Site Kit," and I become confused and worried that the site was never set up properly
B. I am afraid to click the button "Sign in with Google" because I fear I will connect the site with my personal Google Analytics account, not the official one for the publication
C) Even after I went through the flow to connect, I am confused because I can't see Analytics, which is because I hadn't been added to the publication's GA – this however makes sense.

## Steps to reproduce
1. Go to Site Kit UI screen in the WordPress backend
2. I land on the screen "Sign in with Google to configure Site Kit" (see below) – I'm worried Site Kit has never been configured
2. Click on "Sign in with Google"
3. I land on the screen "Welcome to Site Kit! Let's get you set up" (see below) – I'm worried I'm connecting the site to my personal Google Analytics account and not the publication's
4. Next screen has me "confirm my choices" (see below), which makes me wonder how this data is shared with the publication site
5. After setup is complete and I arrive back at the WordPress site I see a "Data error in Analytics" and some language around not having access – which is because I have not yet been added to the publication's Google analytics account. This makes sense to me since I'm aware of that.

## Screenshots
Step 1:
<img width="1424" alt="1" src="https://user-images.githubusercontent.com/4823744/85794941-ae411e00-b6f4-11ea-890a-a004049b63cc.png">
Step 3:
<img width="1123" alt="2" src="https://user-images.githubusercontent.com/4823744/85794956-b5682c00-b6f4-11ea-96bb-963824d090c1.png">
Step 4: 
<img width="473" alt="3" src="https://user-images.githubusercontent.com/4823744/85803171-d126fe80-b703-11ea-9a7e-fe10ed192773.png">
Step 5: 
[...]

PR #1871, opened 2020-08-04
Fix/1714 missing site info

aaemnnosttv:
## Summary

Addresses issue #1714 (follow up)

## Relevant technical choices

This PR fixes missing fields in the state received for and object returned by `getSiteInfo`, specifically `proxySetupURL` and `proxyPermissionsURL`. These values aren't used during E2E since they reference the external authentication service which is why they didn't cause the tests to fail previously.

This PR also updates the tests for `site/datastore/info.js` as well as `site/datstore/connection.js` to ensure that the key for each selector is present in the response from its "parent" selector. Previously this was a bit too easy to miss as it only asserted that the response matched what it was received with. The `connection` tests were DRYed up significantly by converting the existing individual (largely duplicate) tests into a `describe.each` similar to what was already done for the `info` tests and implemented in a similar way to also ensure that each selector's data was present in the parent object.

## Checklist


assets/js/googlesitekit/datastore/site/connection.test.js
```diff
@@ -157,183 +157,34 @@ describe( 'core/site connection', () => {
...
-				fetchMock.getOnce(
-					/^\/google-site-kit\/v1\/core\/site\/data\/connection/,
-					{ body: response, status: 500 }
-				);
+		describe.each( [
+			[ 'hasConnectedAdmins', 'hasConnectedAdmins' ],
+			[ 'isConnected', 'connected' ],
+			[ 'isResettable', 'resettable' ],
+			[ 'isSetupCompleted', 'setupCompleted' ],
+		] )( `%s`, ( selector, connectionKey ) => {
```
tofumatt, line 165: Very slick, without sacrificing readability in this case. I like it! 😄

Issue #1867, opened 2020-08-04
Analytics Top content is sorted by title and not by views


gmmedia:
## Bug Description

The Google Analytics **Top content** widget is sorted by the **title column** and not by the views column.

## Steps to reproduce

Same wrong sorting on the 3 pages:
* WordPress Dashboard widget
* Site Kit Dashboard
* Analytics

## Screenshots
![2020-08-04_074049](https://user-images.githubusercontent.com/7474702/89257311-3389ed80-d626-11ea-910d-50aed4cc3516.png)

## Additional Context

 - PHP Version: 7.3
 - OS: Win7
 - Browser: Chrome
 - Plugin Version: 1.13.0

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* For Analytics top content displays (and underlying queries), results should be sorted by descending views/pageviews. Top content appears in the following locations:

  * WordPress Dashboard - "Top content over the last 28 days"
  * Site Kit Dashboard - "Most popular content"
  * Analytics Module page - "Top content over the last [28] days"

## Implementation Brief

* In `includes/Modules/Analytics.php`, in the handler for `case 'GET:report'`, in the orderby handling immediately following the inline comment `// Order by.`:
  * remove `[0]` from the `if` condition block which checks whether `$orderby` is a numeric array or not:
```js
// When just object is passed we need to convert it to an array of objects.
if ( ! wp_is_numeric_array( $orderby[0] ) ) {
	$orderby = array( $orderby );
}
```
* Create a new protected method `parse_data_orderby` in the `includes/Modules/Analytics.php` that accepts an instance of the `Data_Request` class and returns `$orderby` result using algorithm from the `GET:report` handler.
  * Create unit tests for this method.
* Update `GET:report` handler to use the new method to get orderby settings.

## QA Brief

* Go to each of the following pages and check top content report, it should be sorted descendingly by the number of page views.
  * WordPress Dashboard widget
  * Site Kit Dashboard
  * Analytics

## Changelog entry

* Fix regression where Analytics top content wouldn't be sorted correctly by views.

adamsilverstein:
Thanks for the bug report @gmmedia.

I can confirm I am seeing this as well, top content is sorted by page title in version 1.9.0:


![image](https://user-images.githubusercontent.com/2676022/89323522-11eb2f00-d643-11ea-9bc5-14d033b20bc0.png)


In 1.8.1 I see pages forted by descending Views:

[...]

PR #1875, opened 2020-08-04
Bug/1867 analytics report order

eugene-manuilov:
## Summary

Addresses issue #1867

## Checklist

### DIFF
--- a/includes/Modules/Analytics.php
+++ b/includes/Modules/Analytics.php
@@ -1068,7 +1068,6 @@ final class Analytics extends Module
 	 * @since n.e.x.t
 	 *
 	 * @param array|null $orderby Data request orderby value.
-	 *
 	 * @return Google_Service_AnalyticsReporting_OrderBy[] An array of reporting orderby objects.
 	 */
 	protected function parse_reporting_orderby( $orderby ) {
--- a/includes/Modules/Analytics.php
+++ b/includes/Modules/Analytics.php
@@ -1071,8 +1071,7 @@ final class Analytics extends Module
 	 *
 	 * @return Google_Service_AnalyticsReporting_OrderBy[] An array of reporting orderby objects.
 	 */
-	protected function parse_data_orderby( Data_Request $data ) {
-		$orderby = $data['orderby'];
+	protected function parse_reporting_orderby( $orderby ) {
 		if ( empty( $orderby ) || ! is_array( $orderby ) ) {
 			return array();
 		}
--- a/includes/Modules/Analytics.php
+++ b/includes/Modules/Analytics.php
@@ -1063,7 +1063,7 @@ final class Analytics extends Module
 	}
 
 	/**
-	 * Parses orderby value of the data request.
+	 * Parses the orderby value of the data request into an array of reporting orderby object instances.
 	 *
 	 * @since n.e.x.t
 	 *
--- a/includes/Modules/Analytics.php
+++ b/includes/Modules/Analytics.php
@@ -1067,7 +1067,7 @@ final class Analytics extends Module
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @param array $orderby Data request object.
+	 * @param array|null $orderby Data request orderby value.
 	 *
 	 * @return Google_Service_AnalyticsReporting_OrderBy[] An array of reporting orderby objects.
 	 */
--- a/includes/Modules/Analytics.php
+++ b/includes/Modules/Analytics.php
@@ -1069,7 +1069,7 @@ final class Analytics extends Module
 	 *
 	 * @param Data_Request $data Data request object.
 	 *
-	 * @return array An array of parsed orders.
+	 * @return Google_Service_AnalyticsReporting_OrderBy[] An array of reporting orderby objects.
 	 */
 	protected function parse_data_orderby( Data_Request $data ) {
 		$orderby = $data['orderby'];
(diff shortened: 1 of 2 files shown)


includes/Modules/Analytics.php
```diff
@@ -1094,6 +1062,48 @@ function ( $order_def ) {
...
+						array(
+							'fieldName' => '',
+							'sortOrder' => 'DESCENDING',
+						),
+						(array) $order_def
+					);
+
+					if ( empty( $order_def['fieldName'] ) ) {
+						return null;
+					}
```
aaemnnosttv, line 1093: I know @felixarntz suggested adding the default `sortOrder` back in the previous PR but if we're going to filter out invalid orderbys then I think we should require both to be provided rather than just the `fieldName` and leave out any defaults. The merging there is mainly to ensure the structure to prevent non-existing key warnings, but also because orderbys used to be actual objects due to the way batch requests worked previously.
I'm happy to defer to @felixarntz on this, but in my opinion we should avoid any default here and require both or filter it out.

tests/phpunit/integration/Modules/AnalyticsTest.php
```diff
@@ -458,4 +461,50 @@ protected function get_module_with_screen() {
...
+		$order  = array(
+			'fieldName' => 'views',
+			'sortOrder' => 'ASCENDING',
+		);
+		$result = $reflected_parse_data_orderby_method->invoke( $analytics, new Data_Request( 'GET', null, null, 'report', array( 'orderby' => $order ) ) );
+		$this->assertTrue( is_array( $result ) );
+		$this->assertEquals( 1, count( $result ) );
+		$this->assertTrue( $result[0] instanceof Google_Service_AnalyticsReporting_OrderBy );
+		$this->assertEquals( $order['fieldName'], $result[0]->getFieldName() );
+		$this->assertEquals( $order['sortOrder'], $result[0]->getSortOrder() );
```
aaemnnosttv, line 486: Please use the string literals for the expected values rather than referencing the `$order` variable. They should be the same, but this makes the test easier to read and less likely to pass when it shouldn't if something is wrong in the input.

tests/phpunit/integration/Modules/AnalyticsTest.php
```diff
@@ -458,4 +461,50 @@ protected function get_module_with_screen() {
...
+		$result = $reflected_parse_data_orderby_method->invoke( $analytics, new Data_Request( 'GET', null, null, 'report', array( 'orderby' => $orders ) ) );
+		$this->assertTrue( is_array( $result ) );
+		$this->assertEquals( 2, count( $result ) );
+		$this->assertTrue( $result[0] instanceof Google_Service_AnalyticsReporting_OrderBy );
+		$this->assertEquals( $orders[0]['fieldName'], $result[0]->getFieldName() );
+		$this->assertEquals( $orders[0]['sortOrder'], $result[0]->getSortOrder() );
+		$this->assertTrue( $result[1] instanceof Google_Service_AnalyticsReporting_OrderBy );
+		$this->assertEquals( $orders[1]['fieldName'], $result[1]->getFieldName() );
+		$this->assertEquals( $orders[1]['sortOrder'], $result[1]->getSortOrder() );
+	}
```
aaemnnosttv, line 508: Let's also test that an orderby is not returned if it's `fieldName` is empty. Pending my previous suggestion, we should do the same if the `sortOrder` is also empty (currently this would never happen though since it has the default).

tests/phpunit/integration/Modules/AnalyticsTest.php
eugene-manuilov: Updated.

tests/phpunit/integration/Modules/AnalyticsTest.php
eugene-manuilov: I have added a test to check the case when `fieldName` is empty. Will udpate it if Felix agrees with your suggestion.

includes/Modules/Analytics.php
felixarntz: SGTM!

tests/phpunit/integration/Modules/AnalyticsTest.php
eugene-manuilov: Updated.

includes/Modules/Analytics.php
```diff
@@ -1094,6 +1062,50 @@ function ( $order_def ) {
 		throw new Invalid_Datapoint_Exception();
 	}
 
+	/**
+	 * Parses orderby value of the data request.
```
aaemnnosttv, line 1066:
```suggestion
	 * Parses the orderby value of the data request into an array of reporting orderby object instances.
```

includes/Modules/Analytics.php
```diff
@@ -1094,6 +1062,50 @@ function ( $order_def ) {
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $orderby Data request object.
```
aaemnnosttv, line 1070:
```suggestion
	 * @param array|null $orderby Data request orderby value.
```

includes/Modules/Analytics.php
```diff
@@ -1094,6 +1062,50 @@ function ( $order_def ) {
 		throw new Invalid_Datapoint_Exception();
 	}
 
+	/**
+	 * Parses the orderby value of the data request into an array of reporting orderby object instances.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array|null $orderby Data request orderby value.
+	 *
```
felixarntz, line 1071:
```suggestion
```

Issue #184, opened 2019-07-10
When you don't have enough data, and switch to 7 days and then back to larger data it shows an error that the data isn't available


bjackson27:
Noticed this on my site. I don't have any data from the last 7 days so when I switch to the last 7 days it rightly shows and error. When I switch back to 14, 28 or 90 days the error still displays.

Steps to recreate:
1) View dashboard (or module page) as normal
2) Switch to 7 day view - receive gathering stats message
3) Switch to 14, 28 or 90 day view - receive gathering stats message above the graph

Currently I get:
<img width="1243" alt="5e154a77a53dc159dd71b39ea1991787" src="https://user-images.githubusercontent.com/28811945/61000790-d12bc400-a323-11e9-9799-89876f805962.png">
I expect to get:
<img width="1310" alt="5e212abcb4c5fb8e7c92dff5cbfcc85a" src="https://user-images.githubusercontent.com/28811945/61000798-d557e180-a323-11e9-9e76-0cfa79d575cc.png">

Currently I get:
<img width="1248" alt="5e239c58c61eb62b904436db7d275509" src="https://user-images.githubusercontent.com/28811945/61000831-e9034800-a323-11e9-99cc-0e2d4a87a9eb.png">

I expect
<img width="1310" alt="5e2c4a6691ce5336456560b70247c020" src="https://user-images.githubusercontent.com/28811945/61000839-edc7fc00-a323-11e9-95c2-8f6d3aa7f08c.png">

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* "Gathering data" message should only be displayed for views where there's no data returned from the API.

## Implementation Brief

* Set `zeroData` in state any time new data is loaded.
* Merge #1876

## Changelog entry

* Fix bug where state of not having sufficient data for a Site Kit widget would sometimes incorrectly persist when switching the date range.

## QA Brief

* Test that switching between an empty view and a populated view does not display the error as pictured in https://github.com/google/site-kit-wp/issues/184#issuecomment-668780740
* This state can be hard to reproduce; if you want a copy of cached `sessionStorage` data to test it, ping me and I can send you some privately. The data I have been testing with needs to be anonymised so I don't want to post it here 😄

felixarntz:
Related: #447 

Potentially this will already fix it.

ThierryA:
Moving this to IB as it doesn't have an estimate

felixarntz:
[...]

PR #1876, opened 2020-08-04
Reset zeroData on new data requests.

tofumatt:
## Summary

Addresses issue #184.

## Checklist

### DIFF
--- a/assets/js/components/higherorder/withdata.js
+++ b/assets/js/components/higherorder/withdata.js
@@ -151,7 +151,7 @@ const withData = (
 				'googlesitekit.moduleDataReset',
 				'googlesitekit.moduleDataResetHandler',
 				() => {
-					this.setState( { data: false } );
+					this.setState( { data: false, zeroData: false } );
 				}
 			);
 
@@ -204,7 +204,6 @@ const withData = (
 					data: returnedData,
 					datapoint,
 					module: identifier,
-					zeroData: !! isDataZero( returnedData, datapoint, requestData ),
 				} );
 			};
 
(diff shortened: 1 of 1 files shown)


assets/js/components/higherorder/withdata.js
```diff
@@ -204,6 +204,7 @@ const withData = (
 					data: returnedData,
 					datapoint,
 					module: identifier,
+					zeroData: !! isDataZero( returnedData, datapoint, requestData ),
```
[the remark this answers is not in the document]
tofumatt: Ah, fair. Sorry, the code paths for those modules is sometimes I bit tough for me to follow, but that makes sense. Cheers! 👍

tofumatt:
All set for another review; made the change and it still worked a treat, so let's do the reset in the hook as mentioned 😄

assets/js/components/higherorder/withdata.js
```diff
@@ -151,7 +151,7 @@ const withData = (
 				'googlesitekit.moduleDataReset',
 				'googlesitekit.moduleDataResetHandler',
 				() => {
-					this.setState( { data: false } );
+					this.setState( { data: false, zeroData: false } );
```
aaemnnosttv, line 154: @tofumatt now that I see this in context, we're now resetting 2/3 of the keys in the `NewComponent` state - the other being `error`. It seems that `error` should probably also be reset - essentially resetting all of the component state when the date range changes. What do you think? cc: @felixarntz @adamsilverstein

tofumatt: I was thinking that too, but I was just a bit wary to change more than needed. That said, resetting it shouldn't cause any issues. I'll try it out locally, but I think that's a good call and would prevent future issues similar to this one. 👍

tofumatt: Just tested this locally and it worked fine without causing issues, so I've added `error: false` to the `setState` call here. Ready for another look!

Issue #174, opened 2019-07-10
Somewhere to view results for non-post pages (ex Home) within the admin


bjackson27:
Related PR  https://github.com/google/site-kit-wp/pull/73 as related pull-request where this was touched on.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Display a details page for non-post URLs:
    * term archives (categories, tags, custom taxonomies)
    * date archives (year, month, day)
    * author archives
* Display stats in the admin bar for any of the above URLs (as long as there is enough data returned from the API requests for that URL)

### Refactoring
* Implement new class `Util\Entity_Factory` with static methods (mainly to decouple and reduce code in `Context` which is already large and with the changes here would become significantly larger):
    * `from_context()`: Should have the current logic from `Context::get_reference_entity`, except the initial part where it looks as the `permaLink` query parameter, which is specific to Site Kit admin details and furthermore deals with the reference site URL, for which responsibility should be kept within `Context`.
    * `from_url( $url )`: Should have the current logic from `Context::get_reference_entity_from_url`, except the initial part where it replaces the reference site URL, for which responsibility should be kept within `Context`.
    * In other words, the new class should only deal with WordPress details, not Site Kit details. The "context" bit in the `from_context` method name refers to the actual WordPress context (via WP core functions), not Site Kit's `Context` class.
* `Context::get_reference_entity` should be reduced accordingly, so that it mostly relies on `Entity_Factory::from_context`, except for the Site Kit-related tweaks (see above).
* `Context::get_reference_entity_from_url` should be reduced accordingly, so that it mostly relies on `Entity_Factory::from_url`, except for the Site Kit-related tweaks (see above).
* Modify `Admin_Bar::is_active` so that it relies fully on `Context::get_reference_entity`:
    * Call `Context::get_reference_entity`. If return value is `null`, don't display.
    * Modify `Context::get_reference_entity` to only return an entity for a `WP_Post` if its post status is "publish". This is used in the admin bar, but also generally makes sense, since non-published content should not result in a reference entity being found for Site Kit purposes (only published content matters).
    * If the returned entity type is either "post" or "blog" and has an ID present, use the ID for the `Permissions::VIEW_POST_INSIGHTS` permissions check. In all other cases, for now the permissions check should be `Permissions::VIEW_DASHBOARD` (admin-only). In the future this can be expanded to e.g. also check term capabilities etc.
[...]

PR #1881, opened 2020-08-06
Enhancement/174 non post details

felixarntz:
## Summary

Addresses issue #174

## Relevant technical choices

* Introduces `Util\WP_Query_Factory` for creating a `WP_Query` object for any URL, "reverse-engineering" WP core logic, but actually mostly copying from `url_to_postid()` function and `WP::parse_request()`.
    * This is expensive logic, but it's only run on the single details page in the admin so that's okay.
* Restructures entity detection logic into `Util\Entity_Factory`, relying on WP core conditional logic and behavior as in e.g. `get_the_archive_title()` for entity titles.
* Adds support for non-post entities (via both of these classes), with substantial test coverage for all kinds of content.
* Modifies the admin bar logic to rely on entity detection, enabling the SK admin bar menu for all of these types of content.
* Gets rid of some legacy arguments.
* This PR does not cover paginated content for entity detection yet, which is okay, it can be another iteration and wasn't explicitly mentioned in the ACs. I've opened #1911 to implement that at some point. There are however already tests in place with `TODO` annotations to implement this correctly later.

## Checklist


tests/phpunit/integration/Core/Util/Entity_FactoryTest.php
```diff
@@ -0,0 +1,784 @@
...
+				$title = $this->extract_wp_entity_title_from_url_entity_title( $expected_entity->get_title() );
+				if ( isset( self::$term_names_to_ids[ $title ] ) ) {
+					$this->force_set_property( $expected_entity, 'id', (int) self::$term_names_to_ids[ $title ] );
+				}
+			} elseif ( 'user' === $expected_entity->get_type() ) {
+				$title = $this->extract_wp_entity_title_from_url_entity_title( $expected_entity->get_title() );
+				if ( isset( self::$user_display_names_to_ids[ $title ] ) ) {
+					$this->force_set_property( $expected_entity, 'id', (int) self::$user_display_names_to_ids[ $title ] );
+				}
+			}
```
aaemnnosttv, line 352: This seems unnecessary - does the data provider not have access to `self::$post_titles_to_ids` and others to set the ID in this way? If so, let's elaborate on the comment to explain why they're unknown to the provider.

tests/phpunit/integration/Core/Util/Entity_FactoryTest.php
```diff
@@ -0,0 +1,784 @@
...
+						'type'  => 'term',
+						'title' => 'Custom Taxonomy: Coffee',
+						'id'    => 0, // Filled via title lookup.
+					)
+				),
+			),
+		);
+	}
+
+	protected function assertEntity( Entity $expected, $actual ) {
```
aaemnnosttv, line 747: Nice 👍

tests/phpunit/integration/Core/Util/Entity_FactoryTest.php
```diff
@@ -0,0 +1,784 @@
...
+		add_filter(
+			'option_home',
+			function() {
+				return 'https://example.com';
+			}
+		);
+
+		// Run the actual query.
+		$query = new \WP_Query();
+		$query->query( $query_args );
```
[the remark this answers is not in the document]
aaemnnosttv: Fair enough 😄

tests/phpunit/integration/Core/Util/Entity_FactoryTest.php
aaemnnosttv: I found the reason why in [the docs](https://phpunit.readthedocs.io/en/7.0/writing-tests-for-phpunit.html#data-providers):
> All data providers are executed before both the call to the `setUpBeforeClass()` static method and the first call to the `setUp()` method. Because of that you can’t access any variables you create there from within a data provider. This is required in order for PHPUnit to be able to compute the total number of tests.

aaemnnosttv: After taking a closer look, we can avoid the need to mutate the `$expected_entity` in the test by returning a function from the data provider which returns the expected entity instead of returning it directly from the provider.
Then we only need to invoke the function in the test to get the expected entity; at which point it will have access to the full context of the `TestCase` so referencing the ID of the entity that was created during the set up is no problem.

E.g. this works:
(simplified)
```php
public function test_from_wp_query( array $query_args, \Closure $get_expected_entity ) {
	$expected_entity = $get_expected_entity();

	// ...

	// Run the actual query.
	$query = new \WP_Query();
	$query->query( $query_args );

	// ...

	$entity = Entity_Factory::from_wp_query( $query );

	if ( $expected_entity ) {
		$this->assertEntity( $expected_entity, $entity );
	} else {
		$this->assertNull( $entity );
	}
}

public function data_from_wp_query() {
	return array(
		'front page'                   => array(
			array(),
			function () {
				// The full context of the TestCase will be available when this is called.
				return new Entity(
					'https://example.com/',
					array(
						'type'  => 'post',
						'title' => 'Home',
						'id'    => self::$post_titles_to_ids['Home'],
					)
				);
			},
		),
		// ...
	);
}
```

I think we should opt for this approach to avoid complicated rewriting and keep the test method as simple as possible. It makes the provider method a little more complicated/verbose, but I think the tradeoff is definitely worth it.
Keep in mind that we could also use the `factory` to create a WP entity in the test from a provider-provided function like this too so they wouldn't have to all be created beforehand. Since the same enitites are referenced between other tests in this test case, I think it's okay to share them, just highlighting the other options and flexibility this has.

tests/phpunit/integration/Core/Util/Entity_FactoryTest.php
```diff
@@ -0,0 +1,765 @@
...
+
+	/**
+	 * @dataProvider data_from_wp_query
+	 *
+	 * @param array       $query_args      Query arguments.
+	 * @param Entity|null $expected_entity Expected entity, or null.
+	 */
+	public function test_from_wp_query( array $query_args, Entity $expected_entity = null ) {
+		global $wp_rewrite;
+
```
aaemnnosttv, line 291: This is currently unused in the method.
```suggestion
```

tests/phpunit/integration/Core/Util/Entity_FactoryTest.php
```diff
@@ -0,0 +1,765 @@
...
+	/**
+	 * Extracts the title of a WordPress entity (e.g. a post, term or user) from a Site Kit entity.
+	 *
+	 * Site Kit entities are based on URLs and therefore use the real title for those URLs, which often have a prefix
+	 * separated by the real entity title with ': '. This method strips out the prefix.
+	 *
+	 * @param string $title Site Kit entity title (potentially including ': ').
+	 * @return string Actual WordPress entity title.
+	 */
+	private function extract_wp_entity_title_from_url_entity_title( $title ) {
```
aaemnnosttv, line 758: This method becomes unnecessary as well once we no longer need to mutate the expected entity in the test 🙂

assets/js/googlesitekit-adminbar.js
```diff
@@ -50,19 +50,11 @@ export class GoogleSitekitAdminbar extends Component {
 	}
 
 	async handleMoreDetailsLink() {
-		const {
-			permaLink,
-			postID,
-			postType,
-			pageTitle,
-		} = global._googlesitekitLegacyData;
+		const { permaLink } = global._googlesitekitLegacyData;
```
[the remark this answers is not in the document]
aaemnnosttv: @felixarntz any reason why we should keep using the global here instead?

includes/Core/Util/Entity_Factory.php
```diff
@@ -0,0 +1,458 @@
...
+						$title = _x( 'Chats', 'post format archive title' );
+						break;
+				}
+				break;
+			default:
+				$tax   = get_taxonomy( $term->taxonomy );
+				$title = sprintf(
+					/* translators: %s: Taxonomy singular name. */
+					_x( '%s:', 'taxonomy term archive title prefix' ),
+					$tax->labels->singular_name
```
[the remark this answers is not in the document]
aaemnnosttv: Yeah, it just struck me as off – a specific comment for this line is not necessary I don't think. Any time we reference a section of code copied or based on core, it would be good to be extra clear as to whether it is a copy or adaptation, and where it starts and ends. This reduces the need for cross-referencing. Not a big deal though 👍

tests/phpunit/integration/Core/Util/Entity_FactoryTest.php
```diff
@@ -0,0 +1,784 @@
...
+		// Manually add public query vars and rewrite rules for post types and taxonomies because the latter were
+		// originally skipped due to an empty permalink structure.
+		foreach ( get_post_types( array(), 'objects' ) as $post_type ) {
+			self::fix_post_type_rewrite( $post_type );
+			$post_type->add_rewrite_rules();
+		}
+		foreach ( get_taxonomies( array(), 'objects' ) as $taxonomy ) {
+			self::fix_taxonomy_rewrite( $taxonomy );
+			$taxonomy->add_rewrite_rules();
+		}
```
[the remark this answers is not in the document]
aaemnnosttv: Good point 😄 

We could always call `( new static )->reset_post_types()` but that's a bit of a hack.
Oh well, not a big deal 👍

aaemnnosttv (commented):
Looking good @felixarntz - I think this is mostly good to go, just a few remaining comments and suggestions (mostly regarding the tests) including one or two from the original review which you may have missed.

assets/js/googlesitekit-adminbar.js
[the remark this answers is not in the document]
felixarntz: I changed this to a functional component using hooks. I'd rather avoid doing the "half" migration relying on `withSelect` - eventually we should try to make everything use functional components. I know we have migrated some components that were "out of scope" to make functional for the respective issue to use `withSelect`, but overall I'm thinking we should rather try to migrate to functional components...

* immediately (if easy enough)
* via a separate issue (if the component does more than render and have a single callback or so)

In this case, migrating to a functional component was trivial enough so I made that change.

assets/js/googlesitekit-adminbar.js
```diff
@@ -26,109 +26,92 @@ import './modules';
...
+	const dashboardURL = useSelect( ( select ) => select( CORE_SITE ).getAdminURL( 'googlesitekit-dashboard' ) );
+	const currentEntityURL = useSelect( ( select ) => select( CORE_SITE ).getCurrentEntityURL() );
+	const currentEntityTitle = useSelect( ( select ) => select( CORE_SITE ).getCurrentEntityTitle() );
 
+	const onMoreDetailsClick = useCallback( async () => {
+		const detailsURL = addQueryArgs( dashboardURL, { permaLink: currentEntityURL } );
 		await trackEvent( 'admin_bar', 'post_details_click' );
-		document.location = href;
-	}
+		document.location = detailsURL;
```
aaemnnosttv, line 56: Since we're refactoring this, let's use `assign` here as is our convention.
```suggestion
		document.location.assign( detailsURL );
```

tests/phpunit/integration/Core/Util/Entity_FactoryTest.php
```diff
@@ -286,8 +286,8 @@ public function test_from_url() {
 	 * @param array       $query_args      Query arguments.
 	 * @param Entity|null $expected_entity Expected entity, or null.
 	 */
-	public function test_from_wp_query( array $query_args, Entity $expected_entity = null ) {
-		global $wp_rewrite;
+	public function test_from_wp_query( array $query_args, \Closure $get_expected_entity ) {
```
aaemnnosttv, line 289: The doc block for the second `@param` still references `Entity|null`.

aaemnnosttv (commented):
Looks great, just a few minor details remain.

aaemnnosttv:
@googlebot I consent.

aaemnnosttv (approved):
@felixarntz I addressed the last few remaining minor feedback items to keep this moving since it's tagged with the upcoming release, so this should be good to go now 👍

assets/js/googlesitekit-adminbar.js
```diff
@@ -26,109 +26,92 @@ import './modules';
...
-				pageTitle,
-			}
-		);
+export const GoogleSitekitAdminbar = () => {
+	const dashboardURL = useSelect( ( select ) => select( CORE_SITE ).getAdminURL( 'googlesitekit-dashboard' ) );
+	const currentEntityURL = useSelect( ( select ) => select( CORE_SITE ).getCurrentEntityURL() );
+	const currentEntityTitle = useSelect( ( select ) => select( CORE_SITE ).getCurrentEntityTitle() );
 
+	const onMoreDetailsClick = useCallback( async () => {
+		const detailsURL = addQueryArgs( dashboardURL, { permaLink: currentEntityURL } );
```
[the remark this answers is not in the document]
felixarntz: I was thinking about what would happen if `currentEntityURL` was `undefined` or `null`. But it's probably okay as long as we keep `! currentEntityURL` in the if clause below to not render the component in that case. (Shouldn't be relevant anyway since that entire component should never be called in such a case, but better to handle here too.)

felixarntz:
@aaemnnosttv Thanks for pushing the final updates here, LGTM!

PR #1905, opened 2020-08-11
Update dependencies.

tofumatt:
## Summary

Addresses issue #2034

Upgrades dependencies, including an issue encountered with `withInstanceId` in WordPress.

## Checklist


tofumatt:
Good catch. I think that's actually an issue with our approach here: https://github.com/google/site-kit-wp/blob/89550bf07dcce177a09e38c1e03ce6abf9128a7c/assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js#L111

That likely needs to be refactored to not pass one component's `setState` to another. I think it might just be a matter of adding a new selector/action to the widget datastore that these components can use without requiring passing around the `setState`. Shouldn't be too hard.

tofumatt:
Looks like this is more work than expected, it's not just the Widget Renderer's passing setState that's the problem. While an odd approach, it should actually work. The issue is somewhere we have (functional) components that call `setState()` _during render_. This will lead to bugs and eventually break in future versions of React, so it's worth fixing, but it could be dependencies causing the issue. Unlikely, but I'm looking into it just a bit.

It will likely require pinning React to a pre-16.13 version (that's when this error was introduced) and filing a separate issue for that upgrade. It's still an issue, it's just that React only added the warnings for it in 16.13…

tofumatt:
Pinning the version is _still_ showing the error, and this seems to be a rabbit hole. I'm going to revisit this later…

tofumatt:
Update: we'll explore the `setState()` error separately (currently filing an issue), so this is ready for review with React pinned to `16.12` 😄

felixarntz (changes requested):
@tofumatt Jest tests and e2e tests are failing here, can you review those: https://travis-ci.com/github/google/site-kit-wp/builds/183766884

tofumatt:
Seems `@wordpress/data` is the culprit, but the [CHANGELOG isn't much help](https://github.com/WordPress/gutenberg/blob/master/packages/data/CHANGELOG.md), so I'll need to inspect what changes caused it to break…

tofumatt:
Upgrading `@wordpress/data` from `^4.12.0` to `^4.13.0` caused the bug, so seems like that's where to find the culprit...

felixarntz (changes requested):
@tofumatt There is a problem with the `Switch` modifications here, which surfaces e.g. when editing Analytics settings:
Screenshot 2020-09-16 at 18 47 35

Also, could you explain what the problem was with `withInstanceId`? You said there was a problem, but would be great to have a bit more detail on it and how your new approach fixes it :)

tofumatt:
Sure thing! I've fixed the issue with the `Switch` component; that's actually from us never supplying an `onChange` event and having the div above's `onClick` handle the change. For now we need to supply a no-op function there to silence the warning, though we could refactor that to use `<input onChange />` in the future, possibly. The `Switch` component seems a bit odd though, so I didn't want to delve too deep into it for now.

The `useInstanceId` errors I was seeing might've actually been related to some errors with the Switch components, or to another inconsistency, because they have disappeared. The issues before were Rules of Hooks violations but they weren't clear. It's possible my `package-lock.json` had some inconsistencies I didn't spot, I guess, but I've managed to revert that chance and stick with `@wordpress/compose` 👍🏻

tofumatt:
Argh, okay, something is off here. I'm not seeing this hook failures at all times or in the browser, but am with the Jest tests. I'm investigating that…

tofumatt:
Closed in favour of #2728, which also has issues…

Issue #1895, opened 2020-08-10
Official AMP plugin version 2.0(beta 3) -> Deprecated: AMP_Theme_Support::get_support_mode


maciejmackowiak:
## Bug Description
Hello,
We've started testing the beta version of the official AMP WP plugin and there is a tiny issue with Site Kit by Google:
```
ErrorException

User Deprecated: AMP_Theme_Support::get_support_mode is <strong>deprecated</strong> since version 2.0.0! Use AMP_Options_Manager::get_option instead.
```

Here is the culprit:
https://github.com/google/site-kit-wp/blob/1.13.1/includes/Context.php#L356

## Steps to reproduce

Install and activate AMP WP 2.0.0 beta 3
https://github.com/ampproject/amp-wp/releases/tag/2.0.0-beta3

## Additional Context

 - PHP Version: 7.4
 - Plugin Version 1.13.1
 - AMP WP Version: 2.0.0 beta 3

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* For AMP plugin version >= 2.0.0 (see https://github.com/ampproject/amp-wp/blob/develop/amp.php#L20), rely on `AMP_Options_Manager::get_option` instead of `AMP_Theme_Support::get_support_mode` to detect the AMP mode.
* Just like for the previous method, there should be checks in place to ensure the class and method exists before calling it.

## Implementation Brief

* Update `Context::get_amp_mode` method, it should depend on `AMP__VERSION` constant:
  * If `AMP__VERSION >= 2.0.0`, do the following:
    * Call `AMP_Options_Manager::get_option( 'theme_support' )` instead of `AMP_Theme_Support::get_support_mode()`.
    * Update `$exposes_support_mode` variable definition to use `method_exists( 'AMP_Options_Manager', 'get_option' )` instead of `method_exists( 'AMP_Theme_Support', 'get_support_mode' )`.
  * Otherwise (`AMP__VERSION < 2.0.0`) use existing approach to define `$exposes_support_mode` and to get support mode.

## QA Brief

* Install the latest AMP beta of version 2.0.0-beta3 (https://github.com/ampproject/amp-wp/releases/tag/2.0.0-beta3) on a WordPress site with Site Kit installed. There should be no error on the WordPress Admin dashboard, Analytics setup, or Site Kit Dashboard.
* Install a pre-2.0.0 version of AMP—it should also continue to work as before even with the new code.
[...]

PR #1906, opened 2020-08-11
Check for version 2.0 of AMP Plugin.

tofumatt:
## Summary

Addresses issue #1895.

## Relevant technical choices

I wasn't clear on _which_ page actually encountered this issue, so I was able to use the latest AMP plugin mentioned in the issue and not get an error, but _equally_ I couldn't get the mentioned error to appear anywhere. I'm not clear on which page the error occurs; it wasn't mentioned in the issues or the QA brief.

Still, code-wise this approach makes sense. Let me know how to test it in case I missed testing the actual issue, which I feel like I did 😓 

## Checklist


includes/Context.php
```diff
@@ -357,14 +357,25 @@ public function get_amp_mode() {
...
+			$exposes_support_mode = method_exists( 'AMP_Options_Manager', 'get_option' )
+				&& defined( 'AMP_Theme_Support::STANDARD_MODE_SLUG' )
+				&& defined( 'AMP_Theme_Support::TRANSITIONAL_MODE_SLUG' )
+				&& defined( 'AMP_Theme_Support::READER_MODE_SLUG' );
+		} else {
+			$exposes_support_mode = method_exists( 'AMP_Theme_Support', 'get_support_mode' )
+				&& defined( 'AMP_Theme_Support::STANDARD_MODE_SLUG' )
+				&& defined( 'AMP_Theme_Support::TRANSITIONAL_MODE_SLUG' )
+				&& defined( 'AMP_Theme_Support::READER_MODE_SLUG' );
+		}
```
felixarntz, line 372: Maybe the checks for the three constants would be better moved out of the if-else, into a separate variable since they're exactly the same.

tofumatt:
Ah, indeed, I'm seeing the error now. Looks like I had DEBUG disabled while I was testing something else and didn't enable it again 😅

includes/Context.php
tofumatt: I thought that too but wasn't sure if it'd be worth it. Two of us thinking the same thing says it's probably a good idea, so I'll do that 🙂

tofumatt:
Thanks; this is now tested and working a treat with both pre-2.0 and 2.0, including ones with/without beta version strings 😄

includes/Context.php
```diff
@@ -357,14 +357,32 @@ public function get_amp_mode() {
...
+		$amp_plugin_version       = AMP__VERSION;
+		$amp_exposes_support_mode = defined( 'AMP_Theme_Support::STANDARD_MODE_SLUG' )
 			&& defined( 'AMP_Theme_Support::TRANSITIONAL_MODE_SLUG' )
 			&& defined( 'AMP_Theme_Support::READER_MODE_SLUG' );
 
+		if ( strpos( $amp_plugin_version, '-' ) !== false ) {
+			$amp_plugin_version = explode( '-', $amp_plugin_version )[0];
+		}
+
+		$amp_plugin_version_2_or_higher = version_compare( $amp_plugin_version, '2.0.0', '>=' );
```
felixarntz, line 369: The `AMP__VERSION` constant _may_ not be defined. We should be super conservative and add existence checks for everything in logic like this which basically relies 100% on another plugin that we don't have control over, since it must never cause a fatal error.

I suggest we wrap the whole `$amp_plugin_version_2_or_higher` logic into a `defined( 'AMP__VERSION' )` if clause, and for the else just set it to `false` (_maybe_ a very old version of the plugin didn't have that constant).

includes/Context.php
```diff
@@ -357,14 +357,32 @@ public function get_amp_mode() {
...
+
+		if ( $amp_plugin_version_2_or_higher ) {
+			$exposes_support_mode = method_exists( 'AMP_Options_Manager', 'get_option' )
+				&& $amp_exposes_support_mode;
```
felixarntz, line 373:
```suggestion
			$exposes_support_mode = class_exists( 'AMP_Options_Manager' ) && method_exists( 'AMP_Options_Manager', 'get_option' )
				&& $amp_exposes_support_mode;
```

includes/Context.php
```diff
@@ -357,14 +357,32 @@ public function get_amp_mode() {
...
+		} else {
+			$exposes_support_mode = method_exists( 'AMP_Theme_Support', 'get_support_mode' )
+				&& $amp_exposes_support_mode;
+		}
+
 		if ( $exposes_support_mode ) {
 			// If recent version, we can properly detect the mode.
-			$mode = AMP_Theme_Support::get_support_mode();
+			if ( $amp_plugin_version_2_or_higher ) {
+				$mode = \AMP_Options_Manager::get_option( 'theme_support' );
```
felixarntz, line 382: Nit-picking, but we can remove the backslash from here and instead add a `use` statement for `AMP_Options_Manager` on top of the file.

tofumatt: Ah, sure thing 😄

includes/Context.php
```diff
@@ -357,14 +358,38 @@ public function get_amp_mode() {
...
 
-		$exposes_support_mode = method_exists( 'AMP_Theme_Support', 'get_support_mode' )
-			&& defined( 'AMP_Theme_Support::STANDARD_MODE_SLUG' )
-			&& defined( 'AMP_Theme_Support::TRANSITIONAL_MODE_SLUG' )
-			&& defined( 'AMP_Theme_Support::READER_MODE_SLUG' );
+		if ( defined( 'AMP__VERSION' ) ) {
+			$amp_plugin_version       = AMP__VERSION;
+			$amp_exposes_support_mode = defined( 'AMP_Theme_Support::STANDARD_MODE_SLUG' )
+				&& defined( 'AMP_Theme_Support::TRANSITIONAL_MODE_SLUG' )
+				&& defined( 'AMP_Theme_Support::READER_MODE_SLUG' );
```
felixarntz, line 365: This bit needs to remain outside of the clause, otherwise it could be undefined further below.

Issue #1184, opened 2020-02-27
Include link to support forums in JS error handler


aaemnnosttv:
## Feature Description

Currently when an error is raised in React it may trigger our `ErrorHandler` component (see #943) which in turn renders a `Notification` for the user (rather than a blank screen) with information about the error.

As a notification, we could also provide a `learnMoreURL` prop. If we were to include this, the most useful URL might be a GitHub issue search using the error message
```
https://github.com/google/site-kit-wp/issues?q=is%3Aissue+%22error-message-here%22
```

This should help the user to see if there is already an issue for their particular issue or not.

GitHub is the better place to send users in this case since a fatal error in any case is not likely something that will be easily fixable via support.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* When a JS error occurs, the error notification "Site Kit encountered an error" should include a link "Report this problem", which links to the WP support forums for the plugin.

## Implementation Brief

* Update [`ErrorHandler`](https://github.com/google/site-kit-wp/blob/develop/assets/js/components/ErrorHandler/index.js) component:
  * Replace `description` of the rendered `Notification` component to be a link to WP support forums (https://wordpress.org/support/plugin/google-site-kit/) with `Report this problem` text.
  * Add `{ error.message }` to the child `<pre>` tag right before `{ info.componentStack }`.
  * Add `style={ { overflow: 'auto' } }` attribute to the `<pre>` tag.
  * Add onclick handler to the `<pre>` tag with the following code to automatically select tag content when a user clicks on it:
    ```js
    const element = ...; // DOM element for <pre> tag

    const range = document.createRange();
    range.selectNodeContents(element);

    const selection = window.getSelection();
    selection.removeAllRanges();
    selection.addRange(range);
    ```

## QA Brief

* Add a code that throws an exception in one of the components that is rendered on the dashboard and build assets.
* Open the dashboard page, you should see that an error has occured, its details, a button to copy details to clipboard and a link to WordPress support forum.
* Click on the link, it should open WordPress support forum in a separate tab.
* Click on the "Copy to clipboard" buttton, it should change its icon and copy error details into your clipboard wrapped with backticks.

## Changelog entry

* Add copy-to-clipboard functionality and link to WordPress support forums to generic JavaScript error handler.

tofumatt:
Reading this over again: are we sure this is a good idea? I don't think it's a good UX to direct most users to our GitHub repository and have them check for potential errors or issues.

[...]

PR #1909, opened 2020-08-12
Enhancement/1184 support forums

eugene-manuilov:
## Summary

Addresses issue #1184

## Checklist

### DIFF
--- a/assets/js/components/ErrorHandler/index.js
+++ b/assets/js/components/ErrorHandler/index.js
@@ -1,4 +1,22 @@
 /**
+ * ErrorHandler component.
+ *
+ * Google Site Kit, Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
  * External dependencies
  */
 import PropTypes from 'prop-types';
@@ -56,7 +74,7 @@ class ErrorHandler extends Component {
 			</a>
 		);
 
-		const code = (
+		const errorInfo = (
 			// eslint-disable-next-line
 			<pre onClick={ this.errorClickHandler } style={ { overflow: 'auto' } }>
 				{ error.message + info.componentStack }
@@ -72,13 +90,11 @@ class ErrorHandler extends Component {
 				format="small"
 				type="win-error"
 			>
-				{ code }
+				{ errorInfo }
 			</Notification>
 		);
 	}
 }
-
-ErrorHandler.defaultProps = {};
 
 ErrorHandler.propTypes = {
 	/** @ignore */
--- a/assets/js/components/ErrorHandler/index.js
+++ b/assets/js/components/ErrorHandler/index.js
@@ -22,12 +22,23 @@ class ErrorHandler extends Component {
 			error: null,
 			info: null,
 		};
+
+		this.errorClickHandler = this.onErrorClick.bind( this );
 	}
 
 	componentDidCatch( error, info ) {
 		global.console.error( 'Caught an error:', error, info );
 
 		this.setState( { error, info } );
+	}
+
+	onErrorClick( e ) {
+		const range = document.createRange();
+		range.selectNodeContents( e.target );
+
+		const selection = global.getSelection();
+		selection.removeAllRanges();
+		selection.addRange( range );
 	}
 
 	render() {
@@ -39,16 +50,29 @@ class ErrorHandler extends Component {
 			return children;
 		}
 
+		const reportLink = (
+			<a href="https://wordpress.org/support/plugin/google-site-kit/" target="_blank" rel="noopener noreferrer">
+				{ __( 'Report this problem', 'google-site-kit' ) }
+			</a>
+		);
+
+		const code = (
+			// eslint-disable-next-line
+			<pre onClick={ this.errorClickHandler } style={ { overflow: 'auto' } }>
+				{ error.message + info.componentStack }
+			</pre>
+		);
+
 		return (
 			<Notification
 				id="googlesitekit-error"
 				title={ __( 'Site Kit encountered an error', 'google-site-kit' ) }
-				description={ <code>{ error.message }</code> }
+				description={ reportLink }
 				isDismissable={ false }
 				format="small"
 				type="win-error"
 			>
-				<pre>{ info.componentStack }</pre>
+				{ code }
 			</Notification>
 		);
 	}


assets/js/components/ErrorHandler/index.js
```diff
@@ -30,6 +50,15 @@ class ErrorHandler extends Component {
 		this.setState( { error, info } );
 	}
 
+	onErrorClick( e ) {
+		const range = document.createRange();
+		range.selectNodeContents( e.target );
+
+		const selection = global.getSelection();
+		selection.removeAllRanges();
+		selection.addRange( range );
```
aaemnnosttv, line 59: Let's attach this handler to a new "Copy error to clipboard" button and copy the text to the clipboard when the user clicks it. We can use [`document.execCommand( 'copy' )`](https://caniuse.com/#feat=mdn-api_document_execcommand_copy) to do this which has very good browser support, and then deselect the text

We can also wrap the copied text in single backticks (like WP does with Site Health information) so that we get better formatting in the forum when users use this to report an error this way.

assets/js/components/ErrorHandler/index.js
```diff
@@ -39,23 +68,34 @@ class ErrorHandler extends Component {
...
 				id="googlesitekit-error"
 				title={ __( 'Site Kit encountered an error', 'google-site-kit' ) }
-				description={ <code>{ error.message }</code> }
+				description={ reportLink }
 				isDismissable={ false }
 				format="small"
 				type="win-error"
 			>
-				<pre>{ info.componentStack }</pre>
+				{ errorInfo }
```
aaemnnosttv, line 93: Let's inline the `pre` tag here rather than extracting to a variable.

assets/js/components/ErrorHandler/index.js
```diff
@@ -39,23 +68,34 @@ class ErrorHandler extends Component {
...
+		const reportLink = (
+			<a href="https://wordpress.org/support/plugin/google-site-kit/" target="_blank" rel="noopener noreferrer">
+				{ __( 'Report this problem', 'google-site-kit' ) }
+			</a>
+		);
+
+		const errorInfo = (
+			// eslint-disable-next-line
+			<pre onClick={ this.errorClickHandler } style={ { overflow: 'auto' } }>
+				{ error.message + info.componentStack }
```
aaemnnosttv, line 80: Since this is a `pre` tag, let's make sure these don't potentially share the same line
```suggestion
				{ error.message }
				{ info.componentStack }
```

Also, 👍  for putting these together it's much better 🙂

aaemnnosttv (changes requested):
This is looking good, there's just a few small things to address and one potential enhancement.

Regarding the copy to clipboard button I suggested, I realize this is perhaps a little out of scope with respect to the original definition here, so if you feel strongly about this @eugene-manuilov or if @felixarntz objects, we can discuss this in a follow-up enhancement issue. However, I think there is sufficient benefit to include it as part of this issue (consistency with core, UX, and better quality reports in the forum) even if it takes a little more time.

Everything else should be fairly quick and easy to resolve 👍

assets/sass/utilities/_margin.scss
```diff
@@ -18,6 +18,10 @@
 
 .googlesitekit-plugin {
 
+	.googlesitekit-margin-right-1 {
+		margin-right: 0.5rem;
+	}
```
aaemnnosttv, line 23: This is a little tricky as we haven't really established a scale to use for utilities yet. We also have different units in use in different places and most of our margin styles appear to be using `px` currently.

I don't think we actually need this class though, we just need to add an `mdc-button__icon` to the icon and the spacing is handled for us 👍

package.json
```diff
@@ -18,7 +18,7 @@
     "test": "npm run build:production -- --display=none && bundlesize",
     "test:visualtest": "npm run build:static && start-server-and-test storybook http-get://localhost:9001 backstopjs",
     "test:visualapprove": "backstop approve --config=tests/backstop/config.js",
-    "test:analyze": "npm run build:production -- --env.analyze=true --json --progress --profile > /tmp/stats.json && cp ./dist/assets/js/*.js /tmp && webpack-bundle-analyzer /tmp/stats.json",
+    "test:analyze": "npm run remove-dist && mkdir -p ./dist/assets/js && webpack --mode=production --env.analyze=true --json --profile > ./dist/assets/js/stats.json && webpack-bundle-analyzer ./dist/assets/js/stats.json",
```
aaemnnosttv, line 21: This is beyond the scope of this issue; if there is a case for making these changes, let's open a new issue to discuss it 🙂

assets/js/components/ErrorHandler/index.js
```diff
@@ -6,13 +24,16 @@ import PropTypes from 'prop-types';
 /**
  * WordPress dependencies
  */
-import { Component } from '@wordpress/element';
+import { Component, createRef } from '@wordpress/element';
+import { Dashicon } from '@wordpress/components';
```
aaemnnosttv, line 28: I think this may be the reason the bundle size went over the threshold – we should prefer using `@wordpress/icons` over Dashicons though as dashicons are somewhat "legacy" as of WP 5.5.

After refactoring with the icon library, the bundle size is no longer an issue 👍

aaemnnosttv (commented):
Thanks @eugene-manuilov – this is pretty close but also includes a few things which shouldn't be changed as a part of this issue. Things like optimizing lodash imports are great, and we've been moving to using this syntax with the recent usage, and I agree it should be updated across the codebase but it really warrants its own issue as there are some related packages that should be removed at the same time, etc. Similarly, the changes to the build and webpack config should get a new issue if they're needed. In the future, please reach out when things like this come up as it would have been easier to bump the bundlesize threshold temporarily than to make extra changes that are ultimately not necessary 🙂

aaemnnosttv:
I've pushed a few commits here to address the previous feedback. I apologize if it seems like I'm taking over here but some of this was necessary to provide feedback and it didn't make sense to throw away, so it's a matter of efficiency.

* The changes that were done for the purpose of bundlesize tests have been reverted – let's revisit this in a new issue though as any improvements we can make to our filesizes are great 👍 
* The use of `Dashicons` from `@wordpress/components` has been replaced with `@wordpress/icons` which both solves the bundle size issue, but also gives us icons that are a bit closer to what Google uses for its copy buttons  
image  
image  
You can see all the icons in the WP library here: https://wordpress.github.io/gutenberg/?path=/story/icons-icon--library
* The copy to clipboard behavior has been simplified a bit and most of it has been extracted to a new utility
* I removed the timeout behavior of the copy button since the text it will copy never changes, so once it has been copied it will stay in that visual state, but it can still be copied multiple times if needed

Here's what it looks like
**Before copying**
sitekit 10uplabs com_wp-admin_admin php_page=googlesitekit-dashboard
**After copying**
sitekit 10uplabs com_wp-admin_admin php_page=googlesitekit-dashboard (1)
**The copied text**
```
`Something bad happened. 💣 (On purpose; ErrorComponent was used to simulate an error.)

    in ErrorComponent (created by DashboardApp)
    in DashboardApp (created by GoogleSitekitDashboard)
    in GoogleSitekitDashboard
    in RestoreSnapshots (created by Root)
    in ErrorHandler (created by Root)
    in Root`
```
Note the single backticks are how code blocks are formatted for the wp.org forum, so this is what it will look like there:
image
* Finally, some of the added CSS utilities were no longer necessary after adding the `mdc-button__icon` class on the icon as this takes care of the fill and the margin 👍

assets/js/components/button.js
```diff
@@ -82,9 +82,9 @@ class Button extends Component {
 				aria-controls={ ariaControls }
 				role={ 'a' === SemanticButton ? 'button' : undefined }
 			>
-				{ icon && icon }
+				{ icon }
 				<span className="mdc-button__label">{ children }</span>
-				{ trailingIcon && trailingIcon }
+				{ trailingIcon }
```
felixarntz, line 87: Why were these changed? I'm not saying this is necessarily a problem, but as far as I remember JSX behaves slightly different from regular JS logic for those types of comparisons - there were cases like this one before which resulted in unexpected behavior. Are we 100% sure removing these checks won't render something like `undefined` or `0` or `false` out as a string in the HTML?

assets/js/util/copy-to-clipboard.js
```diff
@@ -0,0 +1,35 @@
...
+ * Copies the given string of text to the clipboard.
+ *
+ * @since n.e.x.t
+ *
+ * @param {string} text Text to copy.
+ */
+export default function copyToClipboard( text ) {
+	const el = document.createElement( 'textarea' );
+	el.innerHTML = text;
+	document.body.appendChild( el );
```
felixarntz, line 29: Should this have some kind of CSS rule to be hidden? Theoretically it could become visible for a split second. May not be a big deal, but it could create a strange experience.

assets/sass/utilities/_overflow.scss
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+.googlesitekit-plugin {
+
+	.googlesitekit-overflow-auto {
+		overflow: auto;
+	}
```
felixarntz, line 23: Is this because of the textarea or why?

aaemnnosttv: I believe this is to allow the `pre` element to scroll horizontally when the content is wider than the element itself, so that it doesn't potentially stretch larger than its container. Can you confirm @eugene-manuilov ?

eugene-manuilov: Yes, that's correct.

assets/js/components/button.js
eugene-manuilov: @felixarntz it is not necessary because if trailingIcon is `undefined` it will be `null` (because it's a default value for trailingIcon) and will render nothing. If it is `false`, it will also render nothing too because react doesn't render an element if it's `false`. If you pass `0`, it will render `0` even with `{ trailingIcon && trailingIcon }` apporach because `0 && 0` always returns `0`.

assets/js/util/copy-to-clipboard.js
eugene-manuilov: @aaemnnosttv ^

aaemnnosttv: The short answer is I think this should be okay; it should happen too fast for it to ever be visible. FWIW, this is how it's done on Jurassic Ninja.

The long answer is that I looked into this a bit, and it's somewhat complicated. WP originally had a similar approach for use with the Site Health copy button, but replaced it with clipboardjs.com (see https://github.com/WordPress/WordPress/commit/65f7cdbce053c7ef0bd3356fc26432d379c1045d). This library boasts very good browser support, but isn't well-suited for copying arbitrary text, and also isn't well-suited for use in React due to added event listeners, and its `Clipboard` API instances that it creates. Not impossible but not ideal.

I think the need to hide the element is only relevant when it is persistent, like how WP had it implemented originally.

I've gone ahead and refactored the use here to use https://github.com/feross/clipboard-copy, which is a super-lightweight version of the same function, but it uses the browser's clipboard API if present first, which requires no temporary element, so I think this is the best solution for our needs.

assets/js/components/button.js
aaemnnosttv: @felixarntz I don't think there is anything JSX specific about this behavior. See the docs for it here: https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator

If I recall, I believe the problem that happened before was with settings where the default value was `'0'`. In JS, a string zero is considered a truthy value, so it will be output. E.g. 
image

The `&&` causes the result of the expression return the first falsy value, or the last truthy value, whichever comes first. 
image


In this case, I think @eugene-manuilov is correct to simplify this logic as it is unnecessarily duplicated.

Issue #1728, opened 2020-06-30
Track advanced plugin Analytics events for


felixarntz:
## Feature Description

For certain popular plugins, Site Kit's Analytics module should fire additional events if the respective plugin is active.

**This work should be done against a `feature/ga-advanced-tracking` branch.**

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Analytics should inject additional code to fire certain events as follows:
    * For WooCommerce, event category `ecommerce` and event actions:
        * `add_to_cart` (with item name)
        * `remove_from_cart` (with item name)
        * `checkout`
        * `review_cart`
        * `cart_contents`
        * `update_cart`
        * `product_details` (with item name)
        * `place_order`
        * `proceed_to_paypal`
    * For WP Forms, Contact Form 7, Gravity Forms, Ninja Forms, Formidable Forms, event category `engagement` and event actions:
        * `form_submit`
* Tracking of these events should work in both AMP and non-AMP.
* Extra JavaScript code included for non-AMP should be in the HTML source after the Analytics tag. It should not have any dependencies and rely on patterns established by the [AMP framework](https://github.com/ampproject/amphtml/blob/feb792fcdc7d328073300d083a516c18c87c1194/extensions/amp-analytics/0.1/analytics-root.js#L433) for performance-efficient selector matching.

### 3P event registration

* 3P plugins should be able to register their own events by hooking into a new action `googlesitekit_register_gtag_events`. The action should receive an instance of a `Plugin_Tracking_Events_Registry` class, which has a `register( array $config )` method.

## Implementation Brief

* See separate protoype and spreadsheet.

## QA Brief

* N/A

## Changelog entry

* Introduce API layer for advanced Analytics event tracking configurations from other plugins.

MiraSm:
If you release it, it will be a bomb. And will this plugin itself simply transmit events to GA, or place a datalayer in code?

OC2PS:
This issue is closed, but evidently the plugin still doesn't do enhanced ecommerce tracking...

jamesozzie:
Any update on this can be tracked via #3051, feel free to subscribe to that issue for further information.

PR #1916, opened 2020-08-14
Add Event List Registry for advanced GA event tracking

j-ka11:
## Summary
Implements an `Event_List_Registry` class that handles 3P event list registration.

Addresses issue #1728

## Relevant technical choices

Introduces an action with the tag `googlesitekit_analytics_register_event_lists` that passes the registry instance. 3P can pass their event list instance to the `register` method of the registry to register their event list.

## Checklist

### DIFF
--- a/includes/Modules/Analytics/Advanced_Tracking.php
+++ b/includes/Modules/Analytics/Advanced_Tracking.php
@@ -19,6 +19,7 @@ use Google\Site_Kit\Modules\Analytics\Advanced_Tracking\Measurement_Events\WooCo
 use Google\Site_Kit\Modules\Analytics\Advanced_Tracking\Measurement_Events\WPForms_Event_List;
 use Google\Site_Kit\Modules\Analytics\Advanced_Tracking\Plugin_Detector;
 use Google\Site_Kit\Modules\Analytics\Advanced_Tracking\Measurement_Code_Injector;
+use Google\Site_Kit\Modules\Analytics\Advanced_Tracking\Event_List_Registry;
 
 // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
 
@@ -64,6 +65,14 @@ final class Advanced_Tracking {
 	private $plugin_detector;
 
 	/**
+	 * Main class event list registry instance.
+	 *
+	 * @since n.e.x.t.
+	 * @var Event_List_Registry
+	 */
+	private $event_list_registry;
+
+	/**
 	 * Advanced_Tracking constructor.
 	 *
 	 * @since n.e.x.t.
@@ -76,6 +85,8 @@ final class Advanced_Tracking {
 		} else {
 			$this->plugin_detector = $plugin_detector;
 		}
+
+		$this->event_list_registry = new Event_List_Registry();
 	}
 
 	/**
@@ -157,6 +168,7 @@ final class Advanced_Tracking {
 			$plugin_event_list->register();
 			$this->plugin_event_lists[] = $plugin_event_list;
 		}
+		do_action( 'googlesitekit_analytics_register_event_list', $this->event_list_registry );
 	}
 
 	/**
@@ -171,6 +183,12 @@ final class Advanced_Tracking {
 				foreach ( $plugin_event_list->get_events() as $measurement_event ) {
 					$this->event_configurations[] = $measurement_event;
 				}
+			}
+		}
+
+		foreach ( $this->event_list_registry->get_active_event_lists() as $registry_event_list ) {
+			foreach ( $registry_event_list->get_events() as $measurement_event ) {
+				$this->event_configurations[] = $measurement_event;
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


includes/Modules/Analytics/Advanced_Tracking/Event_List_Registry.php
```diff
@@ -0,0 +1,70 @@
...
+	 *
+	 * @since n.e.x.t.
+	 *
+	 * @param Measurement_Event_List $event_list The third party event list to be registered.
+	 * @throws \Exception Thrown when $event_list is not an instance of Measurement_Event_List.
+	 */
+	public function register( Measurement_Event_List $event_list ) {
+		if ( ! $event_list instanceof Measurement_Event_List ) {
+			throw new \Exception( 'Event list must extend Measurement_Event_List.' );
+		}
```
felixarntz, line 52: Since this class is type-hinted in the method signature, we don't need to do this check here. If the provided value wasn't a `Measurement_Event_List`, it would already cause a PHP error before.

includes/Modules/Analytics/Advanced_Tracking/Event_List_Registry.php
```diff
@@ -0,0 +1,70 @@
...
+
+		$event_list->register();
```
felixarntz, line 54: I think this `register` call here is a side-effect which shouldn't happen here:
* This registry is only for storing registered event lists.
* The `register` method of each event list is for "registering" that event list in WordPress by integrating with its hooks.

Therefore I think it would be cleaner to remove this line from here and instead, below the `googlesitekit_analytics_register_event_lists` action above, iterate through all registered event lists and call `register` on them.

includes/Modules/Analytics/Advanced_Tracking/Event_List_Registry.php
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	/**
+	 * Gets the list of active event lists.
+	 *
+	 * @since n.e.x.t.
+	 *
+	 * @return Measurement_Event_List[] The list of active event lists.
+	 */
+	public function get_active_event_lists() {
```
felixarntz, line 65: Maybe shorten to more generic `get_all()`? Since this is only a registry for event lists, no need to duplicate that in the method name. We can keep it short like in `register()`.

includes/Modules/Analytics/Advanced_Tracking/Event_List_Registry.php
```diff
@@ -0,0 +1,70 @@
...
+	public function __construct() {
+		$this->active_event_lists = array();
+	}
+
+	/**
+	 * Registers a third party event lists.
+	 *
+	 * @since n.e.x.t.
+	 *
+	 * @param Measurement_Event_List $event_list The third party event list to be registered.
```
felixarntz, line 46: No need to mention "third party" here, it can be any event list - Site Kit could register one too (which is technically what's happening now).

includes/Modules/Analytics/Advanced_Tracking/Event_List_Registry.php
```diff
@@ -0,0 +1,70 @@
...
+	 * @since n.e.x.t.
+	 *
+	 * @return Measurement_Event_List[] The list of active event lists.
+	 */
+	public function get_active_event_lists() {
+		return $this->active_event_lists;
+	}
+
+
+}
```
felixarntz, line 70: No need for the blank lines above here.

includes/Modules/Analytics/Advanced_Tracking/Event_List_Registry.php
```diff
@@ -0,0 +1,70 @@
...
+ */
+class Event_List_Registry {
+
+	/**
+	 * The list of active event lists.
+	 *
+	 * @since n.e.x.t.
+	 * @var Measurement_Event_List[]
+	 */
+	private $active_event_lists;
```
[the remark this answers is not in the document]
michael0525: This issue has been resolved.

includes/Modules/Analytics/Advanced_Tracking.php
```diff
@@ -150,13 +153,17 @@ private function set_up_advanced_tracking_amp( $gtag_amp_opt ) {
...
-			$plugin_event_list->register();
-			$this->plugin_event_lists[] = $plugin_event_list;
+			add_action(
+				'googlesitekit_analytics_register_event_lists',
+				function( $event_list_registry ) use ( $plugin_event_list ) {
+					$event_list_registry->register( $plugin_event_list );
+				}
+			);
 		}
+		do_action( 'googlesitekit_analytics_register_event_lists', $this->event_list_registry );
```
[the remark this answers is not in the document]
michael0525: This issue has been resolved.

includes/Modules/Analytics/Advanced_Tracking/Event_List_Registry.php
michael0525: Resolved.

includes/Modules/Analytics/Advanced_Tracking/Event_List_Registry.php
michael0525: Resolved.

includes/Modules/Analytics/Advanced_Tracking/Event_List_Registry.php
michael0525: Resolved.

includes/Modules/Analytics/Advanced_Tracking/Event_List_Registry.php
michael0525: Resolved.

includes/Modules/Analytics/Advanced_Tracking/Event_List_Registry.php
michael0525: That's a good idea. Modified the code by the suggestion.

Issue #1917, opened 2020-08-14
Only query Search Console data for the current website


felixarntz:
## Bug Description

By default, queries to Search Console API include results for all URLs that are associated with the connected Search Console property. In most cases (e.g. URL properties), all these URLs are associated with the current site by definition - however, it isn't necessarily like that for domain properties, which are always for an entire second-level domain (i.e. may include URLs from other sites such as subdomains or subdirectory sites).

## Screenshots

* I have Site Kit set up on my personal main website, using a domain property for Search Console.
* However, I also have an entirely separate website on a subdomain where I post random cocktail pictures (which apparently is more popular in Search than my main site 😄 ).
* As you can see below, Search Console in Site Kit certainly includes data for that other site, which shouldn't be the case.

![Screenshot 2020-08-14 at 13 33 33](https://user-images.githubusercontent.com/3531426/90290457-c569d580-de32-11ea-938f-111b4fec70b8.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All Search Console API queries from Site Kit should only take URLs which are part of the current reference site into account.

## Implementation Brief

* In `Search_Console::create_search_analytics_data_request`, _always_ include a dimension filter with:
    * dimension `page`
    * operator `contains`
    * expression of the current reference site URL, without trailing slash
* If `page` argument is provided (as already possible before), combine that dimension filter with the above one into a single dimension filter group where both filters have to be satisfied. This ensures `page` cannot refer to a URL outside of the site, which should be expected behavior.
* Merge #1918.

## QA Brief

This is hard to manually replicate without sufficient time investment, unless you already have that setup. Therefore, feel free to go over that together with me (@felixarntz), as I have this kind of setup.

* Setup two separate WordPress sites where one is on a subdomain of the other.
* Set up a Search Console domain property for the second-level domain (has to be done manually, Site Kit will use domain properties if they exist, but cannot set them up).
* Put some content on the subdomain site and search for it in various ways in Google search so that Search Console over time picks this up. May take a couple days.
* Set up Site Kit on the second-level domain site.
* Without this issue fixed, queries would include data for the subdomain site too. After the fix for this issue, it should only include data for the second-level domain site as expected.

Apart from this, it should also be verified this doesn't change the Search Console data shown for the more common URL properties.

## Changelog entry

* Ensure Search Console data in Site Kit only includes data for the current site even when using a domain property.

aaemnnosttv:
[...]

PR #1918, opened 2020-08-14
Always query Search Console data only for the current site.

felixarntz:
## Summary

Addresses issue #1917

Screenshot after this fix (compared to the "incorrect" one in the issue description):
Screenshot 2020-08-14 at 14 03 14

It now shows only keywords relevant to my main site as expected.

## Checklist


includes/Modules/Search_Console.php
```diff
@@ -371,14 +371,28 @@ protected function create_search_analytics_data_request( array $args = array() )
 		if ( ! empty( $args['end_date'] ) ) {
 			$request->setEndDate( $args['end_date'] );
 		}
+
+		// Only include data for URLs that are part of the current site.
+		$filter = new Google_Service_Webmasters_ApiDimensionFilter();
```
aaemnnosttv, line 376: Although slightly more verbose, I think it would be preferable to name the filter variables according to the filter they apply for additional clarity - particularly because the same variable is currently used for both cases.
Something like `$scope_site_filter` for the first site-wide filter, and maybe `$scope_url_filter` for the page-specific one would be more clear?

includes/Modules/Search_Console.php
```diff
@@ -371,14 +371,28 @@ protected function create_search_analytics_data_request( array $args = array() )
...
+		$filter->setDimension( 'page' );
+		$filter->setOperator( 'contains' );
+		$filter->setExpression( esc_url_raw( untrailingslashit( $this->context->get_reference_site_url() ) ) );
```
aaemnnosttv, line 379: The `untrailingslashit` is redundant here since this is always applied for this method, and this is covered by a test:
https://github.com/google/site-kit-wp/blob/1c9f13947cba8d9953ca57b9e87d05c1993d212c/tests/phpunit/integration/ContextTest.php#L151-L153

```suggestion
		$filter->setExpression( esc_url_raw( $this->context->get_reference_site_url() ) );
```

includes/Modules/Search_Console.php
```diff
@@ -371,21 +373,42 @@ protected function create_search_analytics_data_request( array $args = array() )
 		if ( ! empty( $args['end_date'] ) ) {
 			$request->setEndDate( $args['end_date'] );
 		}
+
+		$filters = array();
+
+		// If domain property, limit data to URLs that are part of the current site.
+		if ( 0 === strpos( $property_id, 'sc-domain:' ) ) {
```
aaemnnosttv, line 380: I've noticed we have a number of similar conditionals in this module. I think it would make sense to extract this to a method or maybe even a `Search_Console_Property` class which could have a `get_type` method in addition to the `get_id` method. We can revisit this with a dedicated issue; it's not blocking here.

aaemnnosttv (approved):
Looks good!

Issue #1896, opened 2020-08-11
Migrate Analytics All Traffic dashboard widget to use Widgets API


felixarntz:
The current Analytics `AnalyticsAllTraffic` widget should be migrated to use the Widgets API (which is behind a feature flag), while maintaining the current version as legacy components. Those will be removed only once all widgets have been migrated so that we can do the entire switch.

Some general practices for migrating:
* The migration should happen by fully **re-implementing** the components: In other words, the existing components should keep their current functionality, they should **not** use the datastore. Those legacy components should be prefixed with `Legacy*`.
* A good starting point is probably to copy the components into new files and go from there, while renaming the old ones and leaving them untouched.
* For the refactored version, _only_ the actual widget components should be fully responsible for retrieving data from the datastore. If they rely on further `common` components for UI, these should receive such data via props.
* If there are UI parts that can be reused between the new and legacy components, those can be outsourced into `common` components, however as mentioned above **it is important that the legacy components keep using the legacy data API and are not switched to the datastore**. Only the new ones should be.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The [`AnalyticsAllTraffic`](https://github.com/google/site-kit-wp/blob/1.13.1/assets/js/modules/analytics/components/dashboard/AnalyticsAllTraffic.js) component, and its suplementary [`DashboardAcquisitionPieChart`](https://github.com/google/site-kit-wp/blob/1.13.1/assets/js/modules/analytics/components/dashboard/DashboardAcquisitionPieChart.js) and [`AnalyticsAllTrafficDashboardWidgetTopAcquisitionSources`](https://github.com/google/site-kit-wp/blob/1.13.1/assets/js/modules/analytics/components/dashboard/AnalyticsAllTrafficDashboardWidgetTopAcquisitionSources.js) components should be migrated to functional components using the datastore entirely, no globals and no legacy data API.
* Refactoring map:
    * `dashboard/AnalyticsAllTraffic.js` should become `dashboard/DashboardAllTrafficWidget.js` (refactored) and `dashboard/LegacyAnalyticsAllTraffic.js` (legacy). The refactored component should access the datastore for the main API data.
    * `dashboard/DashboardAcquisitionPieChart.js` should become `common/AcquisitionPieChart.js` (refactored) and `dashboard/LegacyDashboardAcquisitionPieChart.js` (legacy). The refactored component should receive its relevant data via props.
[...]

PR #1931, opened 2020-08-18
Enhancement/1896 all traffic widget

eugene-manuilov:
## Summary

Addresses issue #1896

## Checklist


assets/js/googlesitekit/data/transform-case.js
```diff
@@ -17,27 +17,40 @@
...
- * Transforms a camel-case name into its constant-case name.
+ * Transforms a camelCase name into its CONSTANT_CASE name.
  *
  * @since 1.10.0
  * @private
  *
- * @param {string} name Camel-case name to transform.
- * @return {string} constant-case name.
+ * @param {string} name camelCase name to transform.
+ * @return {string} CONSTANT_CASE name.
```
tofumatt, line 39: Nice changes that help clarify the usage 👍

assets/js/modules/analytics/components/common/AcquisitionPieChart.js
```diff
@@ -0,0 +1,101 @@
...
+				loadHeight={ 205 }
+			/>
+
+			{ source && (
+				<div className="googlesitekit-chart__source">
+					{ __( 'Source:', 'google-site-kit' ) }
+					{ ' ' }
+					<Link href={ getSiteKitAdminURL( 'googlesitekit-module-analytics' ) } inherit>
+						{ _x( 'Analytics', 'Service name', 'google-site-kit' ) }
+					</Link>
```
tofumatt, line 85: Technically we should be combining these strings and using something like `sprintf()` and `dangerouslySetInnerHTML()` with `sanizeHTML()` for this (see an example at https://github.com/google/site-kit-wp/blob/5c342cf4e251fd39e3dd70c9a7bb2a29ac13e771/assets/js/components/error-text.js#L40-L63).

Separating these two strings with a space and placing them in this order works fine in English, but that might not be the case for every locale. True, we lose the ability to use a `<Link>` component and need to use a regular `<a>` instead, but I think in this case it's fine and we can improve the localisation accuracy if we do.

We'd want something like:

```js
sprintf(
	/* translators: %1$s: URL to Analytics Module page in Site Kit Admin, %2$s: Analytics (Service Name) */
	__( 'Source: <a class="googlesitekit-cta-link googlesitekit-cta-link--inherit" href="%1$s">%2$s</a>', 'google-site-kit' ),
	getSiteKitAdminURL( 'googlesitekit-module-analytics' ),
	_x( 'Analytics', 'Service name', 'google-site-kit' ),
);
```

It's not the most lovely approach, but it's better for translations…

assets/js/modules/analytics/components/common/AcquisitionSources.js
```diff
@@ -0,0 +1,88 @@
...
+		},
+	];
+
+	const totalUsers = data[ 0 ].data.totals[ 0 ].values[ 1 ];
+	const dataMapped = data[ 0 ].data.rows.map( ( row, i ) => {
+		const percent = ( row.metrics[ 0 ].values[ 1 ] / totalUsers * 100 );
+
+		return [
+			row.dimensions[ 0 ],
+			<div key={ 'minichart-' + i } className="googlesitekit-table__body-item-chart-wrap">
```
tofumatt, line 63: Nitpick, but we can use template strings here:
```suggestion
			<div key={ `minichart-${ i }` } className="googlesitekit-table__body-item-chart-wrap">
```

assets/js/modules/analytics/components/common/AcquisitionSources.js
```diff
@@ -0,0 +1,88 @@
...
+				{ `${ percent.toFixed( 2 ) }% ` }
```
tofumatt, line 64: This is shown to users, right? If so it should be localised; percentages might want to be displayed differently in other locales.

(Also, is the trailing space here intentional?)

assets/js/util/when-active.js
```diff
@@ -0,0 +1,65 @@
...
+ * @param {Object}   options                     Options for enhancing function.
+ * @param {string}   options.moduleName          Name of a module to check.
+ * @param {Function} [options.fallbackComponent] Optional. Fallback component to render when the module is not active.
+ * @return {Function} Enhancing function.
+ */
+export default function whenActive( { moduleName, fallbackComponent = null } ) {
+	return ( wrappedComponent ) => {
+		const whenActiveComponent = ( props ) => {
+			// eslint-disable-next-line react-hooks/rules-of-hooks
+			const moduleInfo = useSelect( ( select ) => select( MODULES_STORE ).getModule( moduleName ) );
```
tofumatt, line 46: Why is this being disabled here? This component is getting created every time this code is run, so this points to it being better to create this as its own component then include it inside the HOC, if we stick with the HOC approach...

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget.js
```diff
@@ -0,0 +1,74 @@
...
+				</div>
+				<div className="mdc-layout-grid__cell mdc-layout-grid__cell--span-8-desktop mdc-layout-grid__cell--span-4-tablet mdc-layout-grid__cell--span-4-phone">
+					<AcquisitionSources data={ report } />
+				</div>
+			</div>
+		</div>
+	);
+}
+
+export default whenActive( { moduleName: 'analytics' } )( DashboardAllTrafficWidget );
```
tofumatt, line 74: Why use the HOC component here? I'd prefer we not introduce any new HOCs into the codebase, as we're using hooks going forward and HOCs are (in my view) more cognitive load to understand and work on.

Why not use the selector logic in the `whenActive` HOC here? https://github.com/google/site-kit-wp/pull/1931/files#diff-451a45872b732c4c542fba8c369ba0b2R44

I don't think it's worth abstracting that behaviour into its own HOC, as I find it harder-to-follow the abstracted approach (it's less direct and requires extra code) compared to a single `useSelect` in the component that, if `false` allows the component to return a fallback/`null`.

tofumatt (changes requested):
There are a few minor localisation changes to make, but they're straightforward 🙂 

The main issue with the PR as-implemented is the HOC for `whenActive`. I'd rather that HOC be removed and the `useSelect` logic be implemented directly in the component. Sure, it's technically less-abstracted, but I don't think there's a clear benefit to that abstraction… especially not yet, when it's only being used in one place.

Let's keep it simple for now and explore that sort of abstraction later, if needed.

assets/js/util/when-active.js
eugene-manuilov: It's disabled here because that eslint rule incorrectly detects the current use case for `useSelect` hook and thinks that it's called outside of a component which is not the case.

> This component is getting created every time this code is run, so this points to it being better to create this as its own component then include it inside the HOC, if we stick with the HOC approach...

That's exactly how it should work because we need to create a new component for every wrapping component.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget.js
eugene-manuilov: We need the HOC because it helps us to encapsulate the logic for detecting module status and let us get rid of duplicating the logic across a dozen of different widgets. In other words it lets us control when to display a widget without breaking DRY principle.

assets/js/modules/analytics/components/common/AcquisitionPieChart.js
eugene-manuilov: Yes, agree, it's a good improvement. Updated.

assets/js/modules/analytics/components/common/AcquisitionSources.js
eugene-manuilov: Updated.

assets/js/modules/analytics/components/common/AcquisitionSources.js
eugene-manuilov: Yep, updated it to be localizable.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget.js
tofumatt: The DRY you mention is pretty straightforward, and much easier to understand without the HOC obscuring the logic of the component. The main code you're abstracting is:

```js
const moduleInfo = useSelect( ( select ) => select( MODULES_STORE ).getModule( moduleName ) );

if ( ! moduleInfo ) {
    return null;
}

if ( ! moduleInfo.active || ! moduleInfo.connected ) {
    return <div>Fallback</div>;
}
```

Repeating that sort of code inside components/modules is very straightforward and I don't think worth adding the HOC. An issue with the HOC approach is that it moves some of the logic handling of a component _outside_ the body of the component, so understanding the component's flow and potential return values becomes more complex and spread-out. In particular, here, one needs to read:

* https://github.com/google/site-kit-wp/blob/166405fdfb075c0447d06352dbc37b0fafb3ed3c/assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget.js#L33-L72
* https://github.com/google/site-kit-wp/blob/166405fdfb075c0447d06352dbc37b0fafb3ed3c/assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget.js#L74
* https://github.com/google/site-kit-wp/blob/166405fdfb075c0447d06352dbc37b0fafb3ed3c/assets/js/util/when-active.js#L42-L65

to understand a reasonably simple/concise conditional.

Reducing repetition of the
```js
if ( ! moduleInfo ) {
    return null;
}

if ( ! moduleInfo.active || ! moduleInfo.connected ) {
    return <div>Fallback</div>;
}
```
code could be done with a hook that returned an object with `{ moduleInfo, activeAndConnected }`.

If you want to consolidate this code, I think a custom hook would be better.

But right now this is only being used in one place, and the fallback component you've implemented isn't being used. I don't think we should be preemptively optimising/DRYing up code that isn't yet being repeated. I think it would be better to see if, in future modules, we need to compare active-vs-connected or if checking for both always returns the same fallback. If we want non-null for lacking modules, etc. Let's abstract _after_ we've built a few of these components, as we don't know how they'll go yet. Given there wasn't anything in the IB about creating this abstracting and the selectors involved are only a few lines. In the case of this component:

```js
if ( ! moduleInfo ||  ! moduleInfo.active || ! moduleInfo.connected  ) {
    return null;
}
```

will be fine.

So I think we should remove the HOC approach and leave it in the component itself for now. If we do want to abstract in the future, let's prefer custom hooks over HOCs.

assets/js/modules/analytics/components/common/AcquisitionPieChart.js
```diff
@@ -0,0 +1,104 @@
...
+			<GoogleChart
+				data={ extractAnalyticsDataForTrafficChart( data ) }
+				options={ GOOGLE_CHART_PIE_SETTINGS }
+				chartType="pie"
+				id="overview-piechart"
+				loadHeight={ 205 }
+			/>
+
+			{ source && (
+				<div className="googlesitekit-chart__source" dangerouslySetInnerHTML={ { __html: sourceMessage } } />
```
tofumatt, line 89: In my previous comment I referenced code that used `sanitizeHTML()` here when injecting HTML from translation strings. I know the risk of them being corrupted is low, but let's protect against the value of `sourceMessage` containing malicious markup. 😄

assets/js/modules/analytics/components/common/AcquisitionSources.js
```diff
@@ -0,0 +1,92 @@
...
+
+	const totalUsers = data[ 0 ].data.totals[ 0 ].values[ 1 ];
+	const dataMapped = data[ 0 ].data.rows.map( ( row, i ) => {
+		const percent = ( row.metrics[ 0 ].values[ 1 ] / totalUsers * 100 );
+
+		return [
+			row.dimensions[ 0 ],
+			<div key={ `minichart-${ i }` } className="googlesitekit-table__body-item-chart-wrap">
+				{
+					/* translators: %s: acquisition source percentage */
```
tofumatt, line 65: Should this be:
```suggestion
					/* translators: %1$s: acquisition source percentage */
```
given the translation string has a numbered translation placeholder?

assets/js/modules/analytics/components/common/AcquisitionSources.js
```diff
@@ -0,0 +1,92 @@
...
+					sprintf( __( '%1$s%%', 'google-site-kit' ), percent.toFixed( 2 ) )
+				}
+				{ ' ' }
```
tofumatt, line 68: Is this still needed? Can you outline why with a comment if so?

tofumatt (changes requested):
Let's remove the HOC and abstraction for now, see my comment about it.

There's also an issue with the `dangerouslySetInnerHTML` prop but I left a comment about that too.

includes/Core/REST_API/REST_Routes.php
```diff
@@ -93,6 +93,33 @@ function( $server ) {
 				$this->register_routes( $server );
 			}
 		);
+
+		add_filter(
+			'do_parse_request',
+			function( $do_parse_request, $wp ) {
+				add_filter(
+					'query_vars',
+					function( $vars ) use ( $wp ) {
```
aaemnnosttv, line 102: Why the nested filters? Can we not simply filter `query_vars` and use a check to see if it's a REST request?
E.g. `defined( 'REST_REQUEST' ) && REST_REQUEST`

includes/Core/REST_API/REST_Routes.php
```diff
@@ -93,6 +93,33 @@ function( $server ) {
...
+						// Unsets standard public query vars to escape conflicts between WorPress core
+						// and Google Site Kit APIs which happen when WordPress incorrectly parses request
+						// arguments.
+						$namespace = rest_get_url_prefix() . '/' . self::REST_ROOT;
+						if ( substr( $wp->request, 0, strlen( $namespace ) ) === $namespace ) {
+							// List of variable names to remove from public query variables list.
+							$unset_vars = array( 'orderby' );
+							foreach ( $unset_vars as $unset_var ) {
+								unset( $vars[ array_search( $unset_var, $vars, true ) ] );
+							}
```
aaemnnosttv, line 112: This seems a but more complex than necessary. How about this?
```suggestion
							return array_values( array_diff( $vars, $unset_vars ) );
```

assets/js/util/when-active.js
aaemnnosttv: Perhaps it wouldn't error if using [`createHigherOrderComponent`](https://github.com/WordPress/gutenberg/blob/master/packages/compose/src/utils/create-higher-order-component/index.js)? WP uses several HOCs and we use their ruleset so it's a bit strange that it would raise this linting error.

assets/js/util/when-active.js
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * WordPress dependencies
+ */
+import { createElement } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import Data from 'googlesitekit-data';
+import { STORE_NAME as MODULES_STORE } from '../googlesitekit/modules/datastore/constants';
```
aaemnnosttv, line 28: Please use the same naming convention for the alias here as we do in other places 🙂 
```suggestion
import { STORE_NAME as CORE_MODULES } from '../googlesitekit/modules/datastore/constants';
```

aaemnnosttv (commented):
This is looking good, just leaving a few comments and questions of my own here.

includes/Core/REST_API/REST_Routes.php
eugene-manuilov: @aaemnnosttv we need to make sure that we turn off `orderby` only for our endpoints because otherwise we would break standard or 3rd party endpoints which is not what we need.

includes/Core/REST_API/REST_Routes.php
eugene-manuilov: @aaemnnosttv it seems to be easier but it makes more changes than we need. The task is to unset certain variables, your change changes `$vars` array in a whole. It can have a potential negative effect in the future when we come across a plugin that adds its own query vars using stringular keys ( `$query_vars['myplugin_var'] = 'myplugin_var';` ) and relies on those keys somewhere else. We can't be sure which edge cases might be there, so I would encourage not to change the query vars array and leave it as is.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget.js
eugene-manuilov: It's a very good point, @aaemnnosttv! Thanks for raising it.

tofumatt: I had considered the "return early" benefit of the HOC, but I think that should be thought of as a convenience, not a requirement. In the past we had selectors that would error or require particular arguments, but found that created problems like this—in the end the approach we've been taking is that selectors are always safe to call, and will just return `undefined` if not supplied with the right arguments. I don't _think_ we have any selectors left that are unsafe, but if we do: we should fix that. Even if not for this issue, they're in general tough to work with.

In terms of the performance hit of running many selectors… selectors are pretty fast to run so I don't think that's a huge deal. It _could_ matter if resolvers that make HTTP requests are involved, but if they're getting invalid/empty arguments (a bunch of `undefined` because a component's data isn't ready) then they should be designed to skip useless HTTP calls.

The main point of exiting early is neat, but there is a substantial cognitive overhead to higher-order-components compared to hooks. For newer developers hooks are easier to understand and HOCs can be a real pain when they're layered/composed.

From my point-of-view, there's no technical reason not to use hooks, and doing so means our components have a lower surface area of knowledge needed and are more consistent. I also think it's easier to understand a component's logic when its entirely encapsulated within the component. That's a big reason for using hooks over HOCs in general, so I'd prefer we use them here.

If there are reasons _right now_ that mean we can't—certain selectors that cause errors when run without arguments, for instance—then the HOC approach is fine, at least as a stopgap. But I don't see any reference to hooks being technically impossible to use, just the preference for HOCs and the "return early" approach. I understand that might be a slightly more efficient code path, but I think it's also a harder-to-follow code path. I'd rather our code be ever-so-slightly less optimised but more maintainable, and I think eschewing HOCs whenever possible gives us that 🙂 

-----

Regarding `isModuleConnected` though: yes, let's add it. Sounds good 👍 . I think doing so elsewhere/in a new issue might be easier, but whatever works for others 🙂

assets/js/util/when-active.js
eugene-manuilov: @aaemnnosttv unfortunately, it doesn't help.

assets/js/modules/analytics/components/common/AcquisitionPieChart.js
eugene-manuilov: Ah... Updated :)

assets/js/modules/analytics/components/common/AcquisitionSources.js
eugene-manuilov: Updated.

assets/js/modules/analytics/components/common/AcquisitionSources.js
eugene-manuilov: Looks like it is not. It was copied from the legacy component. Removed.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget.js
eugene-manuilov: @tofumatt I feel you are fighting with a non-existing problem. There is no cognitive overhead in HOCs. It's a very simple pattern that is widely used and much easier to understand than even hooks.

Can you tell exactly what makes it hard for you to understand it? A need to look into two files to understand how a component works? We use similar approach in datastores when we use higher-order functions (like createFetchStore) to generate different pieces of a datastore and it is fine for everyone.

Anyway, IB is approved and it states that we need to use whenActive HOC, so let's stick with it. If you want, create a new issue to replace whenActive with a hook. Ok?

eugene-manuilov: `isModuleConnected` is added.

tofumatt: After another look and some clarifications from @aaemnnosttv, the use of HOCs, particularly with selectors that have resolvers, is a good call here. Let's stick with using them, sounds good.

One of the main reasons to use them over hooks here is that hooks run with resolvers may call out to modules that aren't active, which would result in a lot of 404s at best and errors at worst. It might be worth adding another comment about that to clarify their usage, as we generally use render-props pattern or hooks over HOCs in the codebase. But I'm happy to stick with HOCs here 👍

assets/js/util/when-active.js
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * @param {Object}   options                     Options for enhancing function.
+ * @param {string}   options.moduleName          Name of a module to check.
+ * @param {Function} [options.fallbackComponent] Optional. Fallback component to render when the module is not active.
+ * @return {Function} Enhancing function.
+ */
+export default function whenActive( { moduleName, fallbackComponent = null } ) {
+	return ( wrappedComponent ) => {
+		const whenActiveComponent = ( props ) => {
+			// eslint-disable-next-line react-hooks/rules-of-hooks
```
tofumatt, line 45: Let's add a comment as to why this is fine to ignore here, just so future readers will have the context. 🙂

tofumatt (dismissed):
After another look and some clarifications from @aaemnnosttv, the use of HOCs, particularly with selectors that have resolvers, is a good call here. Let's stick with using them, sounds good.

One of the main reasons to use them over hooks here is that hooks run with resolvers may call out to modules that aren't active, which would result in a lot of 404s at best and errors at worst. It might be worth adding another comment about that to clarify their usage, as we generally use render-props pattern or hooks over HOCs in the codebase. But I'm happy to stick with HOCs here 👍 

----

I just have a few requests for comments to clarify why we're using the HOC approach and why we're ignoring the ESLint rule, but the approach here looks great to me. Thanks! 👍

assets/js/util/when-active.js
eugene-manuilov: Added.

assets/js/util/when-active.js
tofumatt: What about moving the `useSelect` above, outside the newly-created component. Technically, this would avoid the hooks violation and doesn't seem like a problem. Would that work? eg:

```js
export default function whenActive( { moduleName, fallbackComponent = null } ) {
	return ( wrappedComponent ) => {
		const isConnected = useSelect( ( select ) => select( CORE_MODULES ).isModuleConnected( moduleName ) );

		const whenActiveComponent = ( props ) => {
			// Return null if the module is not loaded yet or doesn't exist.
			if ( typeof isConnected === 'undefined' || isConnected === null ) {
				return null;
			}

[...]
```

eugene-manuilov: @tofumatt no, unfortunately, we can't do it too, it has to be in the `whenActiveComponent` because if we move it outside, we will definitely break the "rules of hooks" then.

tofumatt:
@googlebot I consent.

includes/Core/REST_API/REST_Routes.php
aaemnnosttv: @eugene-manuilov I see what you mean, but I don't think we need to be so cautious to protect plugins that are registering query vars the wrong way. `$public_query_vars`, as well as the filtered value of [`query_vars`](https://developer.wordpress.org/reference/hooks/query_vars/) is defined as a list of strings (`@var string[]`) so even though a plugin could do that, we shouldn't go out of our way to preserve it – especially since this is only changed for Site Kit REST requests.

tofumatt (approved):
Sweet, looks good. Let's :shipit:

includes/Core/REST_API/REST_Routes.php
eugene-manuilov: Okay, updated.

includes/Core/REST_API/REST_Routes.php
aaemnnosttv: I see what you mean now, it's only used to provide access to the `$wp` instance before the `query_vars` filter is run. I wasn't suggesting we would _only_ do it based on if it was a rest request or not and I think the way you have it now is fine, it's just a bit odd to see it used this way.

includes/Core/REST_API/REST_Routes.php
aaemnnosttv: Let's use [`$wp->remove_query_var( $name )`](https://developer.wordpress.org/reference/classes/wp/remove_query_var/) instead 👍

includes/Core/REST_API/REST_Routes.php
```diff
@@ -93,6 +93,33 @@ function( $server ) {
...
+			'do_parse_request',
+			function( $do_parse_request, $wp ) {
+				add_filter(
+					'query_vars',
+					function( $vars ) use ( $wp ) {
+						// Unsets standard public query vars to escape conflicts between WorPress core
+						// and Google Site Kit APIs which happen when WordPress incorrectly parses request
+						// arguments.
+						$namespace = rest_get_url_prefix() . '/' . self::REST_ROOT;
+						if ( substr( $wp->request, 0, strlen( $namespace ) ) === $namespace ) {
```
aaemnnosttv, line 107: This won't work for sites that are using plain permalinks as `$wp->request` is empty in this case. We can use this condition for cases where the site is using pretty permalinks, but for plain I'm not sure if there is a better option at this point in the request, and we may need to resort to a checking the `REQUEST_URI` but that's getting pretty hacky. Is there a way that you know of that we can reliably use for this case?

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -423,6 +423,36 @@ const baseSelectors = {
...
+	 * Returns `false` if the module exists but is either not active or not connected.
+	 * Returns `undefined` if state is still loading or if no module with that slug exists.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @param {string} slug  Module slug.
+	 * @return {(boolean|undefined)} TRUE when the module exists, is actvie and connected, otherwise FALSE; `undefined` if state is still loading or `null` if said module doesn't exist.
+	 */
+	isModuleConnected: createRegistrySelector( ( select ) => ( state, slug ) => {
```
aaemnnosttv, line 439: Please add a test for this selector 🙂

assets/js/modules/analytics/index.js
```diff
@@ -16,13 +16,21 @@
  * limitations under the License.
  */
 
+/**
+ * External dependencies
```
aaemnnosttv, line 20:
```suggestion
 * WordPress dependencies
```

aaemnnosttv (changes requested):
Looks good for the most part, just a few things to address – most notably the edge case with `query_vars` handling.

includes/Core/REST_API/REST_Routes.php
eugene-manuilov: We can fallback to `REQUEST_URI` if `$wp->request` is empty. Does it sound good to you? @felixarntz are you good with it?

includes/Core/REST_API/REST_Routes.php
eugene-manuilov: @aaemnnosttv we can't use `$wp->remove_query_var( $name )` here because everything that we remove with this function will be overriden with the results of the `query_vars` filter. Do you want me to revert your suggestion and use the initial foreach loop?

assets/js/googlesitekit/modules/datastore/modules.js
eugene-manuilov: Added.

includes/Core/REST_API/REST_Routes.php
felixarntz: @eugene-manuilov I think that's reasonable. Let's add a test for this if possible, for both pretty and ugly permalinks (relying on `rest_route` query parameter).

There is quite some complexity in this though to ensure we cover all instances across different setups, e.g. we'd also need to account for `$wp_rewrite->index` etc. Although this wouldn't exactly be "precise", I'm starting to think we should just see if the request path or query parameter `rest_route` contains `/google-site-kit/v1`?

aaemnnosttv: @eugene-manuilov this seems to be the only thing left to address here.

I think what we need to update this as follows:
- If `$wp->did_permalink`
    - it is a Site Kit REST request if `$wp->request` _contains_ `wp-json/google-site-kit/v1` (not necessarily starts with, so that it covers the `$wp_rewrite->index` case)
- If not If `$wp->did_permalink`
   - it is a Site Kit REST request if `$this->context->input->filter( INPUT_GET, 'rest_route' )` starts with `google-site-kit/v1`  
    The `rest_route` query param takes precedence in this case, even if other parameters are present (which shouldn't be the case) but WP only checks for this query var, so this is enough for us too.

eugene-manuilov: @aaemnnosttv updated it slightly different. Decided not to rely on `$wp->did_permalink` because it might be reset back to `false` in a certain situation.

includes/Core/REST_API/REST_Routes.php
```diff
@@ -79,6 +88,7 @@ public function __construct( Context $context, Authentication $authentication =
 			$modules = new Modules( $this->context, null, null, $this->authentication );
 		}
 		$this->modules = $modules;
+		$this->input   = new Input();
```
aaemnnosttv, line 91: We access this everywhere via the instance in `Context` (available via `$this->context->input`. This is so we can use `MutableInput` in tests and there is only one place to swap this out because the same context instance is shared in internal dependencies. Using a local instance is unnecessary and would make this more difficult. Please use `$this->context->input` instead 😄

aaemnnosttv (commented):
Changes look good, just the one comment.

includes/Core/REST_API/REST_Routes.php
eugene-manuilov: Updated :)

includes/Core/REST_API/REST_Routes.php
```diff
@@ -93,6 +93,40 @@ function( $server ) {
...
+						$unset_vars = ( $wp->request && stripos( $wp->request, trailingslashit( rest_get_url_prefix() ) . self::REST_ROOT ) !== false ) // Check regular permalinks.
+							|| ( empty( $wp->request ) && stripos( $this->context->input()->filter( INPUT_GET, 'rest_route' ), self::REST_ROOT ) !== false ); // Check plain permalinks.
+
+						if ( $unset_vars ) {
+							// List of variable names to remove from public query variables list.
+							return array_values(
+								array_diff(
+									$vars,
+									array(
+										'orderby',
```
felixarntz, line 116: I think for now this is good, but we may consider in the future to remove almost _any_ public query var that WordPress core sets here, to avoid potential similar conflicts with other query vars. I would imagine the only public query var we'll need to still keep would be `rest_route`.

assets/js/modules/analytics/components/common/AcquisitionSources.js
```diff
@@ -0,0 +1,91 @@
...
+	const totalUsers = data[ 0 ].data.totals[ 0 ].values[ 1 ];
+	const dataMapped = data[ 0 ].data.rows.map( ( row, i ) => {
+		const percent = ( row.metrics[ 0 ].values[ 1 ] / totalUsers * 100 );
+
+		return [
+			row.dimensions[ 0 ],
+			<div key={ `minichart-${ i }` } className="googlesitekit-table__body-item-chart-wrap">
+				{
+					/* translators: %1$s: acquisition source percentage */
+					sprintf( __( '%1$s%%', 'google-site-kit' ), percent.toFixed( 2 ) )
```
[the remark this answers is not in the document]
eugene-manuilov: @felixarntz unfortunately, not, because a eslint rule forces it to be ordered whe we have multiple placeholders (it considers `%%` as a placeholder too).

assets/js/googlesitekit/widgets/register-defaults.js
```diff
@@ -49,14 +49,14 @@ export function registerDefaults( widgetsAPI ) {
...
-		style: WIDGET_AREA_STYLES.COMPOSITE,
+		style: WIDGET_AREA_STYLES.BOXES,
 		priority: 1,
 	}, CONTEXT_DASHBOARD );
 
 	widgetsAPI.registerWidgetArea( AREA_DASHBOARD_SEARCH_FUNNEL, {
 		title: __( 'Search Funnel', 'google-site-kit' ),
 		subtitle: __( 'How your site appeared in Search results and how many visitors you got from Search.', 'google-site-kit' ),
-		style: WIDGET_AREA_STYLES.COMPOSITE,
+		style: WIDGET_AREA_STYLES.BOXES,
```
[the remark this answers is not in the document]
eugene-manuilov: Ah... yes! Updated.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget.js
```diff
@@ -0,0 +1,74 @@
...
+import ErrorText from '../../../../components/error-text';
+import AcquisitionPieChart from '../common/AcquisitionPieChart';
+import AcquisitionSources from '../common/AcquisitionSources';
+const { useSelect } = Data;
+
+function DashboardAllTrafficWidget() {
+	const { report, error } = useSelect( ( select ) => {
+		const store = select( STORE_NAME );
+		const args = {
+			...trafficSourcesReportDataDefaults,
```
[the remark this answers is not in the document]
eugene-manuilov: Yep, updated.

assets/js/modules/analytics/components/common/AcquisitionPieChart.js
```diff
@@ -0,0 +1,110 @@
...
+		4: { color: '#34CBE3' },
+		5: { color: '#82E88E' },
+	},
+	title: null,
+	width: '100%',
+};
+
+function AcquisitionPieChart( { data, source } ) {
+	if ( ! data ) {
+		return <PreviewBlock width="282px" height="282px" shape="circular" />;
```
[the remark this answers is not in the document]
eugene-manuilov: Yep, updated.

assets/js/modules/analytics/components/common/AcquisitionSources.js
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Internal dependencies
+ */
+import PreviewTable from '../../../../components/preview-table';
+import { getDataTableFromData, TableOverflowContainer } from '../../../../components/data-table';
+import MiniChart from '../../../../components/mini-chart';
+
+function AcquisitionSources( { data } ) {
+	if ( ! data ) {
+		return <PreviewTable rows={ 3 } rowHeight={ 50 } />;
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget.js
```diff
@@ -0,0 +1,74 @@
...
+		};
+
+		const url = select( CORE_SITE ).getCurrentEntityURL();
+		if ( url ) {
+			args.url = url;
+		}
+
+		return {
+			report: store.getReport( args ),
+			error: store.getErrorForSelector( 'getReport', [ args ] ),
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget.js
```diff
@@ -0,0 +1,74 @@
...
+	} );
+
+	if ( error ) {
+		return (
+			<div className="mdc-layout-grid__cell mdc-layout-grid__cell--span-12">
+				<ErrorText message={ error.message } />
+			</div>
+		);
+	}
+
```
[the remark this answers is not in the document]
eugene-manuilov: Yep, updated.

includes/Core/REST_API/REST_Routes.php
eugene-manuilov: Ok, resolving this for now.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget.js
```diff
@@ -67,30 +75,32 @@ function DashboardAllTrafficWidget() {
 	} );
 
 	if ( error ) {
```
felixarntz, line 94:
```suggestion
	if ( ! loading && error ) {
```

assets/js/modules/analytics/util/index.js
```diff
@@ -37,21 +37,21 @@ export { calculateOverviewData };
 export { default as parsePropertyID } from './parse-property-id';
 export * from './validation';
 
-export const extractAnalyticsDataForTrafficChart = ( reports ) => {
+export const extractAnalyticsDataForTrafficChart = ( reports, idx = 1 ) => {
```
felixarntz, line 40: Would be great to add a comment on what `idx` is and why it is here, as it may not be obvious to someone else lacking all that context.

assets/js/modules/analytics/components/common/AcquisitionSources.js
```diff
@@ -0,0 +1,137 @@
...
+				title: __( 'Sessions', 'google-site-kit' ),
+				/* translators: %s: date range */
+				tooltip: sprintf( __( 'Number of sessions users had on your website over last %s', 'google-site-kit' ), currentRange ),
+			},
+			{
+				title: __( 'Percentage', 'google-site-kit' ),
+				tooltip: __( 'Percentage of sessions', 'google-site-kit' ),
+			},
+		];
+	}
```
felixarntz, line 96: This logic somehow doesn't seem to work. When I look at the widget in the single URL details view, it still only shows the source and percentages like in the overall dashboard, not the five columns that it should as defined in the code here.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget.js
```diff
@@ -0,0 +1,124 @@
...
+		return {
+			loading: store.isResolving( 'getReport', [ args ] ),
+			report: store.getReport( args ),
+			reportArgs: args,
+			error: store.getErrorForSelector( 'getReport', [ args ] ),
+		};
+	} );
+
+	if ( error ) {
+		return getDataErrorComponent( _x( 'Analytics', 'Service name', 'google-site-kit' ), error.code, true, true, false, error );
```
felixarntz, line 95:
```suggestion
		return getDataErrorComponent( _x( 'Analytics', 'Service name', 'google-site-kit' ), error.message, true, true, false, error );
```

Since this is user-facing, it should show a message.

assets/js/modules/analytics/components/common/AcquisitionSources.js
eugene-manuilov: Hm... @felixarntz it works correctly on my end. Perhaps you forgot to re-build assets after pulling my changes? :slightly_smiling_face: Can you re-build assets and check again? 

See the storybook for this PR:

#### Dashboard
dashboard
https://google.github.io/site-kit-wp/storybook/pull/1931/?path=/story/analytics-module-components-dashboard-all-traffic-widget--loaded

#### Page Dashboard
page-dashboard
https://google.github.io/site-kit-wp/storybook/pull/1931/?path=/story/analytics-module-components-page-dashboard-all-traffic-widget--loaded

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget.js
eugene-manuilov: Good catch! Thanks!

assets/js/modules/analytics/util/index.js
eugene-manuilov: Updated.

assets/js/modules/analytics/components/common/AcquisitionSources.js
felixarntz: @eugene-manuilov It works on Storybook indeed, but not on the actual details view page, so there's must be something off contextually there.

Issue #1898, opened 2020-08-11
Migrate Analytics Unique Visitors / Bounce Rate / Goals dashboard widget to use Widgets API


felixarntz:
The current Analytics `AnalyticsDashboardWidgetTopLevel` widget should be migrated to use the Widgets API (which is behind a feature flag), while maintaining the current version as legacy components. Those will be removed only once all widgets have been migrated so that we can do the entire switch.

Some general practices for migrating:
* The migration should happen by fully **re-implementing** the components: In other words, the existing components should keep their current functionality, they should **not** use the datastore. Those legacy components should be prefixed with `Legacy*`.
* A good starting point is probably to copy the components into new files and go from there, while renaming the old ones and leaving them untouched.
* For the refactored version, _only_ the actual widget components should be fully responsible for retrieving data from the datastore. If they rely on further `common` components for UI, these should receive such data via props.
* If there are UI parts that can be reused between the new and legacy components, those can be outsourced into `common` components, however as mentioned above **it is important that the legacy components keep using the legacy data API and are not switched to the datastore**. Only the new ones should be.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The [`AnalyticsDashboardWidgetTopLevel`](https://github.com/google/site-kit-wp/blob/1.13.1/assets/js/modules/analytics/components/dashboard/AnalyticsDashboardWidgetTopLevel.js) component should be migrated to functional components using the datastore entirely, no globals and no legacy data API.
* Refactoring map:
    * `dashboard/AnalyticsDashboardWidgetTopLevel.js` should become `dashboard/DashboardUniqueVisitorsWidget.js` (refactored widget 1), `dashboard/DashboardBounceRateWidget.js` (refactored widget 2), `dashboard/DashboardGoalsWidget.js` (refactored widget 3), and `dashboard/LegacyAnalyticsDashboardWidgetTopLevel.js` (legacy) - i.e. this component should be split into three separate widget components for the refactoring. The refactored widget components should access the datastore for the main API data.
[...]

Issue #1824, opened 2020-07-23
Improve error message for insufficient Google service permissions


ThierryA:
## Feature Description

The current message displayed in a module for which a user doesn't have access to the given Google Service isn't really informative. We should clearly explain why it is as such and ideally give clear instruction on how to gain access (for example, who to reach out to which #1714 would make possible)

Current examples:
![Screenshot 2020-07-27 at 10 41 33](https://user-images.githubusercontent.com/3531426/88575108-336f6980-cff8-11ea-9054-c99cc130656e.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Error messages for missing permissions errors being returned from Google APIs (`error.data.reason` is `'insufficientPermissions'`) should inform the user on how to possibly resolve the issue:
     * If the current user can `list_users` (WordPress capability), the message should include the username of the module owner (see #1743) as point of reference.
    * Otherwise, it should instead refer to just "an administrator".
* Messages:
_Heading: Insufficient permissions in {moduleName}_
_Description: Your Google account does not have sufficient permissions for this {moduleName} account/property/view, so you won't be able to see stats from it on the Site Kit dashboard. This service was originally connected by {user} — you can contact them for more information._
    * `{moduleName}`: The module name (e.g. _Search Console_, _Analytics_)
    * `{user}`: Either _the administrator "xyz"_ (with username of module owner) or _an administrator_
    * To get the specific entity type, the error message should be inspected (they are not translated, so this is acceptable). The following specific entities should be supported:
        * For Search Console "this site" --> "this Search Console property"
        * For Analytics, "this account" / "this property" / "this profile" --> "this Analytics account/property/view"
        * Otherwise, the first sentence of the message should be altered to _Your Google account does not have sufficient permissions to access {moduleName} data._

## Implementation Brief

* Create an `isInsufficientPermissionsError(error)` function that checks `errorObj?.data?.reason === 'insufficientPermissions'`.
* Create an `getInsufficientPermissionsErrorDescription(error, moduleName)` function used to check an `errorObj` object (like the one found in https://github.com/google/site-kit-wp/blob/develop/assets/js/components/notifications/data-error.js#L49 or https://github.com/google/site-kit-wp/blob/517becda04242fc42b4a207bc9458b48 8f05cbe2/assets/js/components/StoreErrorNotice.js#L33) for `isInsufficientPermissionsError(errorObj)`. If it matches, return a modified error with an updated message:
[...]

PR #1938, opened 2020-08-20
Enhancement/1898 visitors goals bounce rate

eugene-manuilov:
## Summary

Addresses issue #1898

## Checklist


assets/js/modules/analytics/components/dashboard/DashboardBounceRateWidget.js
```diff
@@ -0,0 +1,126 @@
...
+
+		const url = select( CORE_SITE ).getCurrentEntityURL();
+		if ( url ) {
+			args.url = url;
+		}
+
+		const sparkDataArgs = {
+			...siteAnalyticsReportDataDefaults,
+			...args,
+		};
```
felixarntz, line 61: This should use hard-coded query args as specifically as possible instead of relying on one of the legacy data objects.

assets/js/modules/analytics/components/dashboard/DashboardBounceRateWidget.js
```diff
@@ -0,0 +1,126 @@
...
+			metrics: [ { expression: 'ga:bounceRate', alias: 'Bounce Rate' } ],
+			limit: 10,
+		};
+
+		return {
+			sparkData: store.getReport( sparkDataArgs ),
+			sparkDataError: store.getErrorForSelector( 'getReport', [ sparkDataArgs ] ),
+			bounceData: store.getReport( bounceDataArgs ),
+			bounceDataError: store.getErrorForSelector( 'getReport', [ bounceDataArgs ] ),
+		};
```
felixarntz, line 76: Why do we need two separate reports here? We should be able to get this information in a single request I think?

Also let's add `loading` fields here for the report(s), based on `store.isResolving`.

assets/js/modules/analytics/components/dashboard/DashboardBounceRateWidget.js
```diff
@@ -0,0 +1,126 @@
...
+		return (
+			<div className="mdc-layout-grid__cell mdc-layout-grid__cell--span-12">
+				<ErrorText message={ ( bounceDataError || sparkDataError ).message } />
+			</div>
+		);
+	}
+
+	if ( ! bounceData || ! sparkData ) {
+		return <PreviewBlock width="100%" height="202px" />;
+	}
```
felixarntz, line 89: This should use `loading` (see above) and be moved above the error clause.

assets/js/modules/analytics/components/dashboard/DashboardGoalsWidget.js
```diff
@@ -0,0 +1,138 @@
...
+
+		const url = select( CORE_SITE ).getCurrentEntityURL();
+		if ( url ) {
+			args.url = url;
+		}
+
+		const sparkDataArgs = {
+			...siteAnalyticsReportDataDefaults,
+			...args,
+		};
```
felixarntz, line 63: See above, the full object of arguments should be hard-coded here, requesting as little metrics as needed.

assets/js/modules/analytics/components/dashboard/DashboardGoalsWidget.js
```diff
@@ -0,0 +1,138 @@
...
+			metrics: [ { expression: 'ga:goalCompletionsAll', alias: 'Goal Completions' } ],
+			limit: 10,
+		};
+
+		return {
+			sparkData: store.getReport( sparkDataArgs ),
+			sparkDataError: store.getErrorForSelector( 'getReport', [ sparkDataArgs ] ),
+			goalsData: store.getReport( goalsDataArgs ),
+			goalsDataError: store.getErrorForSelector( 'getReport', [ goalsDataArgs ] ),
+			goals: store.getGoals(),
```
felixarntz, line 78: See above, this should include `loading`, and we should be able to use a single report request?

assets/js/modules/analytics/components/dashboard/DashboardGoalsWidget.js
```diff
@@ -0,0 +1,138 @@
...
+		return (
+			<div className="mdc-layout-grid__cell mdc-layout-grid__cell--span-12">
+				<ErrorText message={ ( goalsDataError || sparkDataError ).message } />
+			</div>
+		);
+	}
+
+	if ( ! goalsData || ! sparkData ) {
+		return <PreviewBlock width="100%" height="202px" />;
+	}
```
felixarntz, line 92: See above, this should rely on `loading` and be moved above the error clause.

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.js
```diff
@@ -0,0 +1,118 @@
...
+		return (
+			<div className="mdc-layout-grid__cell mdc-layout-grid__cell--span-12">
+				<ErrorText message={ ( visitorsDataError || sparkDataError ).message } />
+			</div>
+		);
+	}
+
+	if ( ! visitorsData || ! sparkData ) {
+		return <PreviewBlock width="100%" height="202px" />;
+	}
```
felixarntz, line 86: Same feedback as in the other components applies here.

assets/js/modules/analytics/components/dashboard/DashboardGoalsWidget.js
```diff
@@ -0,0 +1,138 @@
...
+	} = useSelect( ( select ) => {
+		const store = select( STORE_NAME );
+		const args = {
+			dateRange: select( CORE_USER ).getDateRange(),
+		};
+
+		const url = select( CORE_SITE ).getCurrentEntityURL();
+		if ( url ) {
+			args.url = url;
+		}
```
felixarntz, line 58: The goals widget is only relevant site-wide, so no current entity (`url` argument) should be considered here.

ryanwelcher:
@googlebot I consent.

assets/js/modules/analytics/index.js
```diff
@@ -64,4 +67,45 @@ domReady( () => {
...
+		{
+			component: DashboardBounceRateWidget,
+			width: Widgets.WIDGET_WIDTHS.QUARTER,
+			priority: 4,
+			wrapWidget: true,
+		},
+		[
+			AREA_DASHBOARD_ALL_TRAFFIC, // this is required to have this widget appear
+			AREA_PAGE_DASHBOARD_ALL_TRAFFIC,
+		],
```
felixarntz, line 109: These widget registrations are incorrect, actually already in the ACs (I've fixed that just now). They should be added to the respective `*_SEARCH_FUNNEL` widget areas instead.

ryanwelcher: @felixarntz I've made the changes indicated however, the bounce rate only appears if I add both areas to the registration:

```js
Widgets.registerWidget(
    'analyticsBounceRate',
     {
        component: DashboardBounceRateWidget,
        width: Widgets.WIDGET_WIDTHS.QUARTER,
        priority: 4,
        wrapWidget: true,
    },
    [
        AREA_DASHBOARD_SEARCH_FUNNEL,
        AREA_PAGE_DASHBOARD_SEARCH_FUNNEL,
    ],
);
```

felixarntz: @ryanwelcher I'm not sure how you mean, maybe this wasn't clear before: The Bounce Rate widget should _not_ be displayed on the dashboard, but only on the single URL details view (similar to how it is today). I just removed `AREA_DASHBOARD_SEARCH_FUNNEL` and it is working as expected for me.

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.js
```diff
@@ -0,0 +1,151 @@
...
+			dateRange: select( CORE_USER ).getDateRange(),
+		};
+
+		const url = select( CORE_SITE ).getCurrentEntityURL();
+		if ( url ) {
+			args.url = url;
+		}
+
+		const sparkDataArgs = {
+			compareDateRanges: 1,
```
felixarntz, line 61: Is this needed here? The spark data only displays results for the last 7/14/28/90 days, so we shouldn't need to request double the dates for any comparisons.

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.js
```diff
@@ -0,0 +1,151 @@
...
+					alias: 'Bounce Rate',
+				},
+				{
+					expression: 'ga:avgSessionDuration',
+					alias: 'Average Session Duration',
+				},
+				{
+					expression: 'ga:goalCompletionsAll',
+					alias: 'Goal Completions',
+				},
```
felixarntz, line 83: No need for all this, we only need `ga:users`.

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.js
```diff
@@ -0,0 +1,151 @@
...
+			],
+			limit: 180,
```
felixarntz, line 85: Let's get rid of this limit, it's not necessary and we don't use it anywhere else. There won't ever be more than 180 rows anyway (2*90 days).

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.js
```diff
@@ -0,0 +1,151 @@
...
+			...args,
+		};
+
+		const visitorsDataArgs = {
```
felixarntz, line 89: Maybe name this `args` and the above one `sparklineArgs`? To indicate this one is the main request, the other one is more supplementary. Also, let's add a comment above here: `// This request needs to be separate from the sparkline request because it would result in a different total if it included the ga:date dimension.`

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.js
```diff
@@ -0,0 +1,151 @@
...
+
+	if ( error ) {
+		return getDataErrorComponent( __( 'Analytics', 'google-site-kit' ), error.message );
+	}
+
+	if ( ( ! sparkData || ! sparkData.length ) && ( ! visitorsData || ! visitorsData.length ) ) {
+		return getNoDataComponent( __( 'Analytics', 'google-site-kit' ) );
+	}
+
+	const extractedAnalytics = extractAnalyticsDashboardSparklineData( sparkData );
```
felixarntz, line 121: As mentioned before, let's extract sparkline data manually here, only from `ga:users`, not relying on this legacy function.

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.js
```diff
@@ -0,0 +1,151 @@
...
+
+	const { previousTotalUsers, totalUsers } = parseTotalUsersData( visitorsData );
```
felixarntz, line 123: Let's not use this function and instead extract the values manually like it is done in the Bounce Rate widget above.

stories/module-analytics-components.stories.js
```diff
@@ -48,3 +60,135 @@ generateReportBasedWidgetStories( {
...
+	options: dashboardBounceRateWidgetArgs,
+	component: DashboardBounceRateWidget,
+} );
+
+generateReportBasedWidgetStories( {
+	moduleSlug: 'analytics',
+	datastore: STORE_NAME,
+	group: 'Analytics Module/Components/Page Dashboard/Bounce Rate Widget',
+	data: dashboardBounceRateWidgetData,
+	options: dashboardBounceRateWidgetArgs,
```
felixarntz, line 78: This should use page dashboard-specific data like in the All Traffic Storybook data above - otherwise it's just a duplicate.

stories/module-analytics-components.stories.js
```diff
@@ -48,3 +60,135 @@ generateReportBasedWidgetStories( {
...
+	variantCallbacks: {
+		Loaded: ( dispatch, data, options ) => {
+			dispatch( STORE_NAME ).receiveGetReport( data, { options } );
+			dispatch( STORE_NAME ).receiveGetGoals( goals );
+		},
+		'Data Unavailable': ( dispatch ) => {
+			dispatch( STORE_NAME ).receiveGetGoals( goals );
+		},
+	},
+} );
```
felixarntz, line 116: The Goals widget is only relevant site-wide, so no need for this page dashbaord-specific story.

stories/module-analytics-components.stories.js
```diff
@@ -48,3 +60,135 @@ generateReportBasedWidgetStories( {
...
+	datastore: STORE_NAME,
+	group: 'Analytics Module/Components/Page Dashboard/Unique Visitors Widget',
+	data: {
+		visitorData: dashboardUniqueVisitorsVisitorData,
+		sparkData: dashboardUniqueVisitorsSparkData,
+	},
+	options: {
+		visitorArgs: dashboardUniqueVisitorsVisitorArgs,
+		sparkArgs: dashboardUniqueVisitorsSparkArgs,
+	},
```
felixarntz, line 168: See above, this should use sample data for page dashboard, otherwise it's pretty much the same as above.

stories/utils/generate-widget-stories.js
```diff
@@ -81,27 +82,40 @@ export function generateReportBasedWidgetStories( {
 	group,
 	data,
 	options,
+	variantCallbacks = {},
```
felixarntz, line 85: I think we could simplify this function. The `variantCallbacks` introduces some complexity and there's a ton of similar code in the Storybook definition above that could be omitted. How about this alternative:
* Both `args.data` and `args.options` can also be arrays of multiple instances of data / options.
* If `args.options` is an array, `args.data` must be an array (of arrays) too. Their lengths must match each other.
* If any of this is not the case (invalid function call), throw an error.
* Otherwise, if they are both arrays, iterate through them and call `receiveGetReport`, `receiveError` for `getReport` etc. accordingly.

This way it's more automatic and avoids the need for writing all the callback functions above.

felixarntz (changes requested):
This is looking great overall, a few minor things to iterate on, plus I think the Storybook utility could be simplified.

assets/js/modules/analytics/components/dashboard/DashboardBounceRateWidget.js
```diff
@@ -0,0 +1,138 @@
...
+		if ( url ) {
+			args.url = url;
+		}
+
+		const dataArgs = {
+			multiDateRange: 1,
+			dimensions: 'ga:date',
+			metrics: [ { expression: 'ga:bounceRate', alias: 'Bounce Rate' } ],
+			...args,
+		};
```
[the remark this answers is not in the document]
ryanwelcher: Updated

assets/js/modules/analytics/components/dashboard/DashboardBounceRateWidget.js
```diff
@@ -0,0 +1,138 @@
...
+	// We only want half the date range, having `multiDateRange` in the query doubles the range.
+	for ( let i = Math.ceil( dataRows.length / 2 ); i < dataRows.length; i++ ) {
+		const { values } = dataRows[ i ].metrics[ 0 ];
+		const dateString = dataRows[ i ].dimensions[ 0 ];
+		const date = parseDimensionStringToDate( dateString );
+		sparkLineData.push( [
+			date,
+			values[ 0 ],
+		] );
+	}
```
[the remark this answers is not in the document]
ryanwelcher: 😄

assets/js/modules/analytics/components/dashboard/DashboardBounceRateWidget.js
```diff
@@ -0,0 +1,138 @@
...
+
+		const url = select( CORE_SITE ).getCurrentEntityURL();
+		if ( url ) {
+			args.url = url;
+		}
+
+		const dataArgs = {
+			multiDateRange: 1,
+			dimensions: 'ga:date',
+			metrics: [ { expression: 'ga:bounceRate', alias: 'Bounce Rate' } ],
```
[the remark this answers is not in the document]
ryanwelcher: Updated

assets/js/modules/analytics/components/dashboard/DashboardGoalsWidget.js
```diff
@@ -0,0 +1,142 @@
...
+		loading,
+		goals,
+	} = useSelect( ( select ) => {
+		const store = select( STORE_NAME );
+
+		const dataArgs = {
+			dateRange: select( CORE_USER ).getDateRange(),
+			multiDateRange: 1,
+			dimensions: 'ga:date',
+			metrics: [ { expression: 'ga:goalCompletionsAll', alias: 'Goal Completions' } ],
```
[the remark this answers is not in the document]
ryanwelcher: Updated

assets/js/modules/analytics/components/dashboard/DashboardGoalsWidget.js
```diff
@@ -0,0 +1,142 @@
...
+function DashboardGoalsWidget() {
+	const {
+		data,
+		error,
+		loading,
+		goals,
+	} = useSelect( ( select ) => {
+		const store = select( STORE_NAME );
+
+		const dataArgs = {
```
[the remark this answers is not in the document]
ryanwelcher: Updated

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.js
ryanwelcher: Updated.

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.js
ryanwelcher: Updated.

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.js
ryanwelcher: updated

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.js
ryanwelcher: Updated

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.js
ryanwelcher: updated

assets/js/modules/analytics/components/dashboard/DashboardGoalsWidget.js
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	const sparkLineData = [
+		[
+			{ type: 'date', label: 'Day' },
+			{ type: 'number', label: 'Bounce Rate' },
+		],
+	];
+
+	if ( data[ 0 ].data.rows ) {
```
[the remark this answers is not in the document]
ryanwelcher: With the change that we made to the Analytics API to include empty rows, this now shows a red line.

stories/module-analytics-components.stories.js
ryanwelcher: Updated.

stories/module-analytics-components.stories.js
ryanwelcher: Updated.

stories/module-analytics-components.stories.js
ryanwelcher: Updated.

stories/utils/generate-widget-stories.js
ryanwelcher: @felixarntz this approach will work as long as we are always expecting to dispatch `dispatch( datastore ).receiveGetReport( data, { options } );` Unfortunately, this isn't always the case.

For example the Goals widget needs to dispatch `dispatch( STORE_NAME ).receiveGetGoals( goals );` as well. 

Perhaps, we can extend this approach by only using `variantCallback` to hold additional functions to run after we have iterated over the data/options array?

ryanwelcher: @felixarntz I've implemented this in my commit here. This approach seems to work the best as we only need to populate the `additionalVariantCallbacks` objects when we deviate only needed to run `receiveGetReport`.

assets/js/modules/analytics/components/dashboard/DashboardBounceRateWidget.js
```diff
@@ -0,0 +1,139 @@
...
+			loading: store.isResolving( 'getReport', [ args ] ),
+		};
+	} );
+
+	if ( loading ) {
+		return <PreviewBlock width="100%" height="202px" />;
+	}
+
+	if ( error ) {
+		return getDataErrorComponent( __( 'Analytics', 'google-site-kit' ), error.message );
```
felixarntz, line 80: This will need to be updated after the recent changes in #1824, to pass the module slug as first parameter instead. Also it's missing to pass the `error` as last parameter (the signature of this legacy function is super terrible, but let's not worry about this now).

assets/js/modules/analytics/components/dashboard/DashboardGoalsWidget.js
```diff
@@ -0,0 +1,145 @@
...
+				{
+					expression: 'ga:goalCompletionsAll',
+					alias: 'Goal Completions',
+				},
+			],
+		};
+
+		return {
+			data: store.getReport( args ),
+			error: store.getErrorForSelector( 'getReport', [ args ] ),
```
felixarntz, line 66:
```suggestion
			error: store.getErrorForSelector( 'getReport', [ args ] ) || store.getErrorForSelector( 'getGoals', [] ),
```

assets/js/modules/analytics/components/dashboard/DashboardGoalsWidget.js
```diff
@@ -0,0 +1,145 @@
...
+			loading: store.isResolving( 'getReport', [ args ] ),
```
felixarntz, line 67:
```suggestion
			loading: store.isResolving( 'getReport', [ args ] ) || store.isResolving( 'getGoals', [] ),
```

assets/js/modules/analytics/components/dashboard/DashboardBounceRateWidget.js
```diff
@@ -0,0 +1,139 @@
...
+			],
+		};
+
+		const url = select( CORE_SITE ).getCurrentEntityURL();
+		if ( url ) {
+			args.url = url;
+		}
+
+		return {
+			args,
```
felixarntz, line 68: Why return this here? Seems unnecessary.

stories/utils/generate-widget-stories.js
```diff
@@ -81,27 +82,76 @@ export function generateReportBasedWidgetStories( {
...
+			if ( Array.isArray( options ) ) {
+				options.forEach( ( option, index ) => {
+					dispatch( datastore ).receiveGetReport( data[ index ], { options: option } );
+				} );
+			} else {
+				dispatch( datastore ).receiveGetReport( data, { options } );
+			}
+
+			// Run any additional callbacks if they exist.
+			if ( additionalVariantCallbacks ) {
```
felixarntz, line 119: Safer here:
```suggestion
			if ( additionalLoadingCallback ) {
```

stories/utils/generate-widget-stories.js
```diff
@@ -81,27 +82,76 @@ export function generateReportBasedWidgetStories( {
...
+				additionalLoadingCallback( dispatch, data, options );
+			}
 		},
 		'Data Unavailable': ( { dispatch } ) => {
-			dispatch( datastore ).receiveGetReport( [], { options } );
+			if ( Array.isArray( options ) ) {
+				options.forEach( ( option, index ) => {
+					const returnType = Array.isArray( options[ index ] ) ? [] : {};
```
felixarntz, line 126: This is somewhat off, we need to check the type of `data` here (which is the return value), not `options`, which is always an object and irrelevant for the response type.

felixarntz (changes requested):
Just a few minor things, I'll quickly update those.

Awesome work here @eugene-manuilov and @ryanwelcher! Especially amazing that we're already fixing two current bugs through this refactoring (see #1996) 🎉

felixarntz (approved):

PR #1936, opened 2020-08-19
Enhancement/1897 impressions and clicks widgets

eugene-manuilov:
## Summary

Addresses issue #1897

## Checklist

- [x] My code is tested and passes existing unit tests.
- [x] My code has an appropriate set of unit tests which all pass.
- [x] My code is backward-compatible with WordPress 4.7 and PHP 5.6.
- [x] My code follows the [WordPress](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) coding standards.
- [x] My code has proper inline documentation.
- [x] I have signed the Contributor License Agreement (see <https://cla.developers.google.com/>).

Issue #1899, opened 2020-08-11
Migrate Search Console Top Keywords dashboard widget to use Widgets API


felixarntz:
The current Search Console `DashboardWidgetPopularKeywordsTable` and `SearchConsoleDashboardWidgetKeywordTable` widgets should be migrated to use the Widgets API (which is behind a feature flag), while maintaining the current version as legacy components. Those will be removed only once all widgets have been migrated so that we can do the entire switch.

Some general practices for migrating:
* The migration should happen by fully **re-implementing** the components: In other words, the existing components should keep their current functionality, they should **not** use the datastore. Those legacy components should be prefixed with `Legacy*`.
* A good starting point is probably to copy the components into new files and go from there, while renaming the old ones and leaving them untouched.
* For the refactored version, _only_ the actual widget components should be fully responsible for retrieving data from the datastore. If they rely on further `common` components for UI, these should receive such data via props.
* If there are UI parts that can be reused between the new and legacy components, those can be outsourced into `common` components, however as mentioned above **it is important that the legacy components keep using the legacy data API and are not switched to the datastore**. Only the new ones should be.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The [`DashboardWidgetPopularKeywordsTable`](https://github.com/google/site-kit-wp/blob/1.13.1/assets/js/modules/search-console/components/dashboard/DashboardWidgetPopularKeywordsTable.js) and [`SearchConsoleDashboardWidgetKeywordTable`](https://github.com/google/site-kit-wp/blob/1.13.1/assets/js/modules/search-console/components/dashboard/SearchConsoleDashboardWidgetKeywordTable.js) component should be migrated to functional components using the datastore entirely, no globals and no legacy data API.
* Refactoring map:
    * `dashboard/DashboardWidgetPopularKeywordsTable.js` and `dashboard/SearchConsoleDashboardWidgetKeywordTable.js` should be combined into a single `dashboard/DashboardPopularKeywordsWidget.js` (refactored), plus `dashboard/LegacyDashboardWidgetPopularKeywordsTable.js` (legacy 1) and `dashboard/LegacySearchConsoleDashboardWidgetKeywordTable.js` (legacy 1). The refactored widget component should access the datastore for the main API data.
    * `dashboard/DashboardPopularity.js` should simply be renamed to `dashboard/LegacyDashboardPopularity.js`.
[...]

PR #1942, opened 2020-08-21
Enhancement/1899 top keywords widget

eugene-manuilov:
## Summary

Addresses issue #1899

**This needs to be reviewed after #1936 is merged into develop.**

## Checklist


ryanwelcher:
@googlebot I consent.

stories/module-search-console-components.stories.js
```diff
@@ -66,3 +69,21 @@ generateReportBasedWidgetStories( {
...
+	options: dashboardPopularKeyWordsWidgetArgs,
+	component: DashboardPopularKeywordsWidget,
+} );
+
+generateReportBasedWidgetStories( {
+	moduleSlug: 'search-console',
+	datastore: STORE_NAME,
+	group: 'Search Console Module/Components/Page Dashboard/Popular Keywords Widget',
+	data: dashboardPopularKeyWordsWidgetData,
+	options: dashboardPopularKeyWordsWidgetArgs,
```
felixarntz, line 87: This should use different data, which is more applicable to the page dashboard (e.g. `options` should include `url`).

assets/js/modules/search-console/components/dashboard/DashboardPopularKeywordsWidget.js
```diff
@@ -0,0 +1,142 @@
...
+
+		return {
+			data: store.getReport( args ),
+			error: store.getErrorForSelector( 'getReport', [ args ] ),
+			loading: store.isResolving( 'getReport', [ args ] ),
+			baseServiceURL: store.getServiceURL( {
+				path: '/performance/search-analytics',
+				query: {
+					resource_id: domain,
+					num_of_days: 28,
```
felixarntz, line 70: There's a quick win we can get here, linking to the actually relevant result if in the single URL view: If there is a current entity (`url` is set), let's also pass another query argument here which needs to be like `page: !${ url }`. The `!` ensures it will be looking for an exact URL match, you can try it out by visiting a URL like this in your Search Console property.

The current behavior ignores the current URL, which isn't as great, so let's fix this in the new widget.

assets/js/modules/search-console/index.js
```diff
@@ -84,4 +90,17 @@ domReady( () => {
 			AREA_PAGE_DASHBOARD_SEARCH_FUNNEL,
 		],
 	);
+	Widgets.registerWidget(
+		'searchConsolePopularKeywords',
+		{
+			component: DashboardPopularKeywordsWidget,
+			width: Widgets.WIDGET_WIDTHS.HALF,
+			priority: 1,
+			wrapWidget: true,
```
felixarntz, line 99: Since the table in the widget has its own padding etc, it should not use `wrapWidget: true` and instead pass `wrapWidget: false`. Then, the actual component above needs to take care of rendering the `Widget` component itself - for the regular case, where data is available, it should then add the class so that there's no padding on the widget itself. (For the other cases (error, loading, zero data), it can just render the `Widget` component as is, since for those cases the regular padding is needed.)

See the implementation of the new PageSpeed Insights widget to see an example for a widget without `wrapWidget: true`.

stories/module-search-console-components.stories.js
```diff
@@ -66,3 +69,21 @@ generateReportBasedWidgetStories( {
...
+	component: DashboardPopularKeywordsWidget,
+} );
```
felixarntz, line 89: Based on the above comment, in both of these tests, `wrapWidget: false` needs to be specified too for proper appearance within Storybook.

felixarntz (changes requested):
This looks great! Just a few things to iterate on.

felixarntz:
@eugene-manuilov @ryanwelcher Can one of you also add a QA Brief to the issue #1899? You can look at the previous similar ones for inspiration.

assets/js/modules/search-console/components/dashboard/DashboardPopularKeywordsWidget.js
```diff
@@ -0,0 +1,142 @@
...
+		];
+	} );
+
+	const options = {
+		hideHeader: false,
+		chartsEnabled: false,
+		links,
+	};
+
+	const dataTable = getDataTableFromData( dataMapped, headers, options );
```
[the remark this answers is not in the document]
ryanwelcher: Ticket created #2016.

assets/js/modules/search-console/components/dashboard/DashboardPopularKeywordsWidget.js
ryanwelcher: Updated.

stories/module-search-console-components.stories.js
ryanwelcher: @felixarntz I updated this. Just a note here that the data returned is the same shape both with and without the `url` parameter. I have created a `page` version of the fixture to keep things consistent - even though it is the same as the non-page version.

assets/js/modules/search-console/index.js
ryanwelcher: updated

stories/module-search-console-components.stories.js
ryanwelcher: updated

assets/js/modules/search-console/components/dashboard/DashboardPopularKeywordsWidget.js
```diff
@@ -0,0 +1,145 @@
...
+		},
+		{
+			title: __( 'Impressions', 'google-site-kit' ),
+			tooltip: __( 'Counted each time your content appears in search results', 'google-site-kit' ),
+		},
+	];
+
+	const links = [];
+	const dataMapped = data.map( ( row, i ) => {
+		const query = row.keys[ 0 ];
```
felixarntz, line 113: Just noticed while testing that this also needs the exclamation point in front to work correctly - otherwise these links go to the keyword overview, not the specific query.
```suggestion
		const query = `!${ row.keys[ 0 ] }`;
```

felixarntz (commented):
LGTM, just one follow-up thing I noticed (which is also wrong in the current implementation).

felixarntz (approved):
Awesome!

Issue #1900, opened 2020-08-11
Migrate Analytics Popular Pages dashboard widget to use Widgets API


felixarntz:
The current Analytics `AnalyticsDashboardWidgetPopularPagesTable` widget should be migrated to use the Widgets API (which is behind a feature flag), while maintaining the current version as legacy components. Those will be removed only once all widgets have been migrated so that we can do the entire switch.

Some general practices for migrating:
* The migration should happen by fully **re-implementing** the components: In other words, the existing components should keep their current functionality, they should **not** use the datastore. Those legacy components should be prefixed with `Legacy*`.
* A good starting point is probably to copy the components into new files and go from there, while renaming the old ones and leaving them untouched.
* For the refactored version, _only_ the actual widget components should be fully responsible for retrieving data from the datastore. If they rely on further `common` components for UI, these should receive such data via props.
* If there are UI parts that can be reused between the new and legacy components, those can be outsourced into `common` components, however as mentioned above **it is important that the legacy components keep using the legacy data API and are not switched to the datastore**. Only the new ones should be.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The [`AnalyticsDashboardWidgetPopularPagesTable`](https://github.com/google/site-kit-wp/blob/1.13.1/assets/js/modules/analytics/components/dashboard/AnalyticsDashboardWidgetPopularPagesTable.js) component should be migrated to a functional component using the datastore entirely, no globals and no legacy data API.
* Refactoring map:
    * `dashboard/AnalyticsDashboardWidgetPopularPagesTable.js` should become `dashboard/DashboardPopularPagesWidget.js` (refactored), and `dashboard/LegacyAnalyticsDashboardWidgetPopularPagesTable.js` (legacy). The refactored widget component should access the datastore for the main API data.
* The new `DashboardPopularPagesWidget` should return `null` if the Analytics module is not active. _Only_ otherwise Analytics data should be requested, i.e. this probably needs to go into a separate component `DashboardPopularPages`, since hooks cannot run conditionally (we use `*Widget` as top-level entry point while the un-suffixed version is essentially the main part of it, without any extra checks for active modules etc.).
* The new `DashboardPopularPagesWidget` component should be registered as a widget with priority 2, width `Widgets.WIDGET_WIDTHS.HALF`, and for the widget area `AREA_DASHBOARD_POPULARITY`.

#### Stage 1
[...]

Issue #1922, opened 2020-08-17
Links in "Most popular content" dashboard widget should point to Site Kit detail view


felixarntz:
The links in the Analytics "Most popular content" widget on the Site Kit dashboard and in the WP dashboard widget currently simply point to the page's URL in the frontend, which isn't very useful. They should instead point to the Site Kit detail page (`dashboard-details`) for that URL.

Note that the links should not be changed for the similar widget on the Analytics module page. On there, the links point to the Google Analytics service frontend, which should remain like that.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The links for the pages in the "Most popular content" Site Kit dashboard widget and the list in the WordPress dashboard widget should point to the respective Site Kit detail page:
    `{adminURL}/admin.php?page=googlesitekit-dashboard&permaLink={pageURL}`

## Implementation Brief

* Add an option to `getDataTableFromData` that allows admin URLs to be shown instead of site URLs. Because other tables are built from this function, it needs to be off by default.
* A proof of concept is located in #1956 
* An improvement might be to refactor `AnalyticsDashboardWidgetPopularPagesTable` as a functional component (should be straightforward) and use `const siteURL = useSelect( ( select ) => select( CORE_SITE ).getReferenceSiteURL() );` to get the siteURL instead of relying on the `global._googlesitekitLegacyData.admin.siteURL` as I did in the proof-of-concept PR above.

## QA Brief

* Clicking on a link's name in the main Site Kit dashboard in Analytics should now navigate to the Site Kit Analytics details page for that permalink. Clicking on the permalink below should still navigate to the frontend:

<img width="1200" alt="Screenshot 2020-08-26 at 19 44 16" src="https://user-images.githubusercontent.com/90871/91343915-1a173400-e7d5-11ea-9091-f9ab47baf57a.png">

* The Analytics dashboard functionality should not change and should still navigate to the external Analytics page:

<img width="1207" alt="Screenshot 2020-08-26 at 19 42 35" src="https://user-images.githubusercontent.com/90871/91343978-31562180-e7d5-11ea-8a31-c276d8fb177e.png">

## Changelog entry

* Modify title links in Popular Pages table in Site Kit dashboard and WordPress dashboard widget to point to the details view for the relevant URL.

felixarntz:
@tofumatt 
[...]

Issue #1859, opened 2020-07-30
Display errors correctly in module setup/setting flows


felixarntz:
## Bug Description

Currently, error handling in module setup flows is rudimentary, with only the last single error being displayed. The bigger problem though is that if an error occurs for the main request that is used to determine whether the setup should be in loading state, the loading bar will just keep going, and the error will never be displayed.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `createErrorStore` function should be expanded so that the store includes a new `getErrors()` selector, which returns an array of all errors in the store. It should also include a `hasErrors()` selector which returns a boolean based on `getErrors()`.
* There should be a new `ErrorNotice` component in `assets/js/components/ErrorNotice.js` which behaves similarly to the existing `StoreErrorNotice` component (i.e. including the logic for whether or not to return `null`), except that:
    * It receives an `error` object as prop instead of the `storeName`.
    * It displays the passed `error` instead of getting it from the store.
* The existing `StoreErrorNotice` should internally rely on `ErrorNotice`, simply passing the retrieved `error`.
* There should be a `StoreErrorNotices` component in `assets/js/components/StoreErrorNotices.js` which behaves similarly like the `StoreErrorNotice` component (see #1749), except that it displays potentially multiple errors, using the new `getErrors()` selector on the store passed. If a `shouldDisplayError` prop is provided, every error should be run through it (as well as the other checks like `isPermissionScopeError`) to see if it should be displayed. Then the component should display all errors using an `ErrorText` component for each.
    * Instead of replicating the logic which is now handled by the `ErrorNotice` above, it should render a `<ErrorNotice>` for every error.
* Every module's `ErrorNotice` component should be renamed to `ErrorNotices` and use the new `StoreErrorNotices`. All references should be updated to use that component.

#### `createErrorStore` additions
* Add a new `selectors.getErrors` which returns `Object.values( state.errors )` as well as `state.error` if set in a single array
    - Because `errors` may already contain `error`, the list should be reduced to a list of unique errors based on `error.code` and `error.message`
* `selectors.hasErrors` already exists, but it should be updated to use `getErrors` rather than the current logic which does not account for `state.error`

#### New components
* Create a new `ErrorNotice` component in `assets/js/components/ErrorNotice.js`
    - Should accept an optional `error` object as a prop
    - Everything else is the same as `StoreErrorNotice`, except the `error` is sourced from the prop, rather than the store
[...]

PR #1943, opened 2020-08-21
Enhancement/1900 popular pages widget

eugene-manuilov:
## Summary

Addresses issue #1900

## Checklist


assets/js/modules/analytics/components/dashboard/DashboardPopularPagesWidget.js
```diff
@@ -0,0 +1,120 @@
...
+	};
+
+	const dataTable = getDataTableFromData( dataMapped, headers, options );
+
+	return (
+		<Layout
+			className="googlesitekit-popular-content"
+			footer
+			footerCtaLabel={ _x( 'Analytics', 'Service name', 'google-site-kit' ) }
+			footerCtaLink={ ( analytics || {} ).homepage }
```
aaemnnosttv, line 110:
```suggestion
			footerCtaLink={ analytics?.homepage }
```

assets/js/modules/analytics/components/dashboard/DashboardPopularPagesWidget.js
```diff
@@ -0,0 +1,120 @@
...
+				<ErrorText message={ error.message } />
+			</div>
+		);
+	}
+
+	if ( ! data ) {
+		return <PreviewTable padding />;
+	}
+
+	if ( ! Array.isArray( data ) || ! data.length || ! Array.isArray( data[ 0 ].data.rows ) ) {
```
aaemnnosttv, line 71: We can condense this down into a single check with chaining 😄 
```suggestion
	if ( ! Array.isArray( data?.[ 0 ]?.data?.rows ) ) {
```

assets/js/modules/analytics/components/dashboard/DashboardPopularPagesWidget.js
```diff
@@ -0,0 +1,120 @@
...
+
+	if ( error ) {
+		return (
+			<div className="mdc-layout-grid__cell mdc-layout-grid__cell--span-12">
+				<ErrorText message={ error.message } />
+			</div>
+		);
+	}
+
+	if ( ! data ) {
```
aaemnnosttv, line 67: Maybe we should use the result of a `hasFinishedResolution` selector here for the report instead?

We return early if there's an error so the behavior should be the same, but we should be generally using resolution selectors for "is loading" logic I think.

assets/js/modules/analytics/components/dashboard/DashboardPopularPagesWidget.js
```diff
@@ -0,0 +1,120 @@
...
+			data: store.getReport( args ),
+			error: store.getErrorForSelector( 'getReport', [ args ] ),
+			analytics: select( CORE_MODULES ).getModule( 'analytics' ),
+		};
+	} );
+
+	if ( error ) {
+		return (
+			<div className="mdc-layout-grid__cell mdc-layout-grid__cell--span-12">
+				<ErrorText message={ error.message } />
```
aaemnnosttv, line 62: I feel like we should be using the generic `ErrorNotice` component here as defined in #1859. Even though this error is for the report specifically, it could potentially be a permission scope error. Also we don't have the `reconnectURL` prop here. If we don't want or need it, that's fine but not sure if this is acceptable or not. @felixarntz ?

aaemnnosttv (commented):
This looks good! Nothing really blocking I think but I left a few comments with suggestions and few questions for consideration, mostly for @felixarntz regarding the intended use here.

Will assign to Felix for a final pass.

assets/js/modules/analytics/components/dashboard/DashboardPopularPagesWidget.js
```diff
@@ -0,0 +1,120 @@
...
+			</div>
+		);
```
felixarntz, line 64: Let's use `getDataErrorComponent` like elsewhere.

assets/js/modules/analytics/components/dashboard/DashboardPopularPagesWidget.js
```diff
@@ -0,0 +1,120 @@
...
+		const store = select( STORE_NAME );
+		const args = {
+			...getTopPagesReportDataDefaults(),
+			dateRange: select( CORE_USER ).getDateRange(),
+		};
+
+		return {
+			data: store.getReport( args ),
+			error: store.getErrorForSelector( 'getReport', [ args ] ),
+			analytics: select( CORE_MODULES ).getModule( 'analytics' ),
```
felixarntz, line 55: Let's also include a `loading` property here like elsewhere.

assets/js/modules/analytics/components/dashboard/DashboardPopularPagesWidget.js
```diff
@@ -0,0 +1,120 @@
...
+		return (
+			<div className="mdc-layout-grid__cell mdc-layout-grid__cell--span-12">
+				<ErrorText message={ error.message } />
+			</div>
+		);
+	}
+
+	if ( ! data ) {
+		return <PreviewTable padding />;
+	}
```
felixarntz, line 69: See other PRs, this should instead rely on `loading` and be moved above the `error` clause.

felixarntz (changes requested):
Similar feedback like for the other PRs mostly. Let's add Storybook coverage too using the new utility.

ryanwelcher:
@googlebot I consent.

assets/js/modules/analytics/components/dashboard/DashboardPopularPagesWidget.js
```diff
@@ -0,0 +1,132 @@
...
+			loading: store.isResolving( 'getReport', [ args ] ),
+		};
+	} );
+
+	if ( loading ) {
+		return <PreviewTable padding />;
+	}
+
+	if ( error ) {
+		return getDataErrorComponent( __( 'Analytics', 'google-site-kit' ), error.message );
```
felixarntz, line 80: This will need to be updated after #1824.

assets/js/modules/analytics/datastore/__fixtures__/index.js
```diff
@@ -30,6 +30,8 @@ export { default as dashboardAllTrafficArgs } from './dashboard-all-traffic-widg
 export { default as dashboardAllTrafficData } from './dashboard-all-traffic-widget-data.json';
 export { default as pageDashboardAllTrafficArgs } from './page-dashboard-all-traffic-widget-args.json';
 export { default as pageDashboardAllTrafficData } from './page-dashboard-all-traffic-widget-data.json';
+export { default as pageDashboardPopularPagesArgs } from './dashboard-popular-pages-widget-args.json';
+export { default as pageDashboardPopularPagesData } from './dashboard-popular-pages-widget-data.json';
```
felixarntz, line 34: These should be prefixed with `dashboard`, not `pageDashboard`.

assets/js/modules/analytics/components/dashboard/DashboardPopularPagesWidget.js
```diff
@@ -0,0 +1,132 @@
...
+			numberFormat( row.metrics[ 0 ].values[ 0 ] ),
+		];
+	} );
+
+	const options = {
+		hideHeader: false,
+		chartsEnabled: false,
+		links,
+		showURLs: true,
+	};
```
felixarntz, line 113: We'll also need to add the change from #1922 to the new widget here, it should link to the single URL details view in the admin via `useAdminURLs`.

felixarntz (changes requested):
This looks great, just a few things to fix. Plus, this still needs a QA Brief on the issue.

assets/js/modules/analytics/components/dashboard/DashboardPopularPagesWidget.js
ryanwelcher: Updated.

assets/js/modules/analytics/datastore/__fixtures__/index.js
ryanwelcher: Updated.

assets/js/modules/analytics/components/dashboard/DashboardPopularPagesWidget.js
ryanwelcher: Updated.

assets/js/modules/analytics/components/dashboard/DashboardPopularPagesWidget.js
```diff
@@ -0,0 +1,133 @@
...
+				{
+					fieldName: 'ga:pageviews',
+					sortOrder: 'DESCENDING',
+				},
+			],
+			limit: 10,
+		};
+
+		return {
+			analytics: select( CORE_MODULES ).getModule( 'analytics' ),
```
felixarntz, line 68: Instead of the `analytics.homepage` below, we should rely on the `getServiceURL` selector of the `modules/analytics` store, similar like it is done in the Search Console Top Keywords widget in #1899. This way it works based on the datastore and does not rely on server-provided data (currently this link is not showing up as expected in Storybook because of that).

felixarntz (changes requested):

felixarntz (approved):

PR #1957, opened 2020-08-27
Bug/1700 manifest autload

eugene-manuilov:
## Summary

Addresses issue #1700

## Checklist


aaemnnosttv (approved):
LGTM 👍

aaemnnosttv:
The build is failing but it's unrelated to these changes it seems, but rather the new version of the AMP plugin that we install for E2E tests.

eugene-manuilov:
@aaemnnosttv so what's the plan? Do we need to wait till e2e are fixed or we can go ahead and merge this PR to unblock smoke testing?

aaemnnosttv:
@eugene-manuilov it can't be merged as long as checks are failing – GitHub won't allow it.

We can use the zip built from the action though to test with.

.github/workflows/zips.yml
```diff
@@ -31,9 +31,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-composer-
       - name: Composer Install
-        uses: php-actions/composer@v1
-        with:
-          args: install
+        run: composer install --no-interaction --no-suggest --no-progress
```
felixarntz, line 34: Why did this need to be updated?

eugene-manuilov: @felixarntz because `php-actions/composer` generates files owned by `root` user. It causes permission denied errors when we try to re-generate autoload map.

aaemnnosttv: @felixarntz I believe there was a permissions issue due to ownership differences between composer which runs in the action's container and the "local" composer since we now run composer during the zip command which uses the "local" composer binary and user permissions.

Issue #1902, opened 2020-08-11
Migrate AdSense Top Earning Pages dashboard widget to use Widgets API


felixarntz:
The current AdSense `DashboardAdSenseTopEarningPagesSmall` widget should be migrated to use the Widgets API (which is behind a feature flag), while maintaining the current version as legacy components. Those will be removed only once all widgets have been migrated so that we can do the entire switch.

Some general practices for migrating:
* The migration should happen by fully **re-implementing** the components: In other words, the existing components should keep their current functionality, they should **not** use the datastore. Those legacy components should be prefixed with `Legacy*`.
* A good starting point is probably to copy the components into new files and go from there, while renaming the old ones and leaving them untouched.
* For the refactored version, _only_ the actual widget components should be fully responsible for retrieving data from the datastore. If they rely on further `common` components for UI, these should receive such data via props.
* If there are UI parts that can be reused between the new and legacy components, those can be outsourced into `common` components, however as mentioned above **it is important that the legacy components keep using the legacy data API and are not switched to the datastore**. Only the new ones should be.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The [`DashboardAdSenseTopEarningPagesSmall`](https://github.com/google/site-kit-wp/blob/1.13.1/assets/js/modules/adsense/components/dashboard/DashboardAdSenseTopEarningPagesSmall.js) and [`AdSenseDashboardWidgetTopPagesTableSmall`](https://github.com/google/site-kit-wp/blob/1.13.1/assets/js/modules/analytics/components/dashboard/AdSenseDashboardWidgetTopPagesTableSmall.js) components should be migrated to a single functional component using the datastore entirely, no globals and no legacy data API.
* Refactoring map:
    * `dashboard/DashboardAdSenseTopEarningPagesSmall.js` and the Analytics `dashboard/AdSenseDashboardWidgetTopPagesTableSmall.js` components should be combined into a single `dashboard/DashboardTopEarningPagesWidget.js` (refactored), plus `dashboard/LegacyDashboardAdSenseTopEarningPagesSmall.js` (legacy 1) and Analytics `dashboard/LegacyAdSenseDashboardWidgetTopPagesTableSmall.js` (legacy 2). The refactored component should access the datastore for the main API data.
* The new `DashboardTopEarningPagesWidget` should perform checks via the datastore:
    * If the AdSense module is not active, return `null`.
    * If the Analytics module is not active, return `null`.
[...]

PR #1955, opened 2020-08-26
Enhancement/1901 adsense summary

eugene-manuilov:
## Summary

Addresses issue #1901

## Checklist

- [x] My code is tested and passes existing unit tests.
- [x] My code has an appropriate set of unit tests which all pass.
- [x] My code is backward-compatible with WordPress 4.7 and PHP 5.6.
- [x] My code follows the [WordPress](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) coding standards.
- [x] My code has proper inline documentation.
- [x] I have signed the Contributor License Agreement (see <https://cla.developers.google.com/>).

Issue #1901, opened 2020-08-11
Migrate AdSense Summary dashboard widget to use Widgets API


felixarntz:
The current AdSense `AdSenseDashboardMainSummary` widget should be migrated to use the Widgets API (which is behind a feature flag), while maintaining the current version as legacy components. Those will be removed only once all widgets have been migrated so that we can do the entire switch.

Some general practices for migrating:
* The migration should happen by fully **re-implementing** the components: In other words, the existing components should keep their current functionality, they should **not** use the datastore. Those legacy components should be prefixed with `Legacy*`.
* A good starting point is probably to copy the components into new files and go from there, while renaming the old ones and leaving them untouched.
* For the refactored version, _only_ the actual widget components should be fully responsible for retrieving data from the datastore. If they rely on further `common` components for UI, these should receive such data via props.
* If there are UI parts that can be reused between the new and legacy components, those can be outsourced into `common` components, however as mentioned above **it is important that the legacy components keep using the legacy data API and are not switched to the datastore**. Only the new ones should be.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The [`AdSenseDashboardMainSummary`](https://github.com/google/site-kit-wp/blob/1.13.1/assets/js/modules/adsense/components/dashboard/AdSenseDashboardMainSummary.js) component should be migrated to a functional component using the datastore entirely, no globals and no legacy data API.
* Refactoring map:
    * `dashboard/AdSenseDashboardMainSummary.js` should become `dashboard/DashboardSummaryWidget.js` (refactored) and `dashboard/LegacyAdSenseDashboardMainSummary.js` (legacy). The refactored component should access the datastore for the main API data.
    * `dashboard/DashboardEarnings.js` should simply be renamed to `dashboard/LegacyDashboardEarnings.js`.
* The new `DashboardSummaryWidget` should return `null` if the AdSense module is not active. _Only_ otherwise AdSense data should be requested, i.e. this probably needs to go into a separate component `DashboardSummary`, since hooks cannot run conditionally (we use `*Widget` as top-level entry point while the un-suffixed version is essentially the main part of it, without any extra checks for active modules etc.).
* The new `DashboardSummaryWidget` component should be registered as a widget with priority 1, width `Widgets.WIDGET_WIDTHS.HALF`, and for the widget area `AREA_DASHBOARD_EARNINGS`.

#### Stage 1
* Rename the following files and components accordingly:
[...]

PR #1963, opened 2020-08-27
Enhancement/1902 adsense top earning

eugene-manuilov:
## Summary

Addresses issue #1902

## Checklist


ryanwelcher:
@googlebot  I consent.

ryanwelcher:
@felixarntz @aaemnnosttv this looks to be ready for CR.

eugene-manuilov:
@ryanwelcher @aaemnnosttv @felixarntz this one is blocked by #1955. Once that PR is approved and merged, this one will be ready.

assets/js/modules/adsense/components/dashboard/DashboardTopEarningPagesWidget.js
```diff
@@ -72,22 +82,11 @@ function DashboardTopEarningPagesWidget() {
...
-	if ( ! data ) {
-		return (
-			<PreviewTable rows={ 5 } padding />
-		);
+		return getDataErrorComponent( __( 'Analytics and AdSense', 'google-site-kit' ), error.message );
 	}
 
-	if ( ! Array.isArray( data[ 0 ].data.rows ) ) {
-		return null;
+	if ( ! data || ! Array.isArray( data[ 0 ].data.rows ) ) {
```
aaemnnosttv, line 88:
```suggestion
	if ( ! Array.isArray( data?.[ 0 ]?.data?.rows ) ) {
```

assets/js/modules/adsense/components/dashboard/DashboardTopEarningPagesWidget.js
```diff
@@ -0,0 +1,140 @@
...
+		hideHeader: false,
+		chartsEnabled: false,
+		cap: 5,
+		links,
+	};
+
+	const dataTable = getDataTableFromData( dataMapped, headers, options );
+
+	return (
+		<Layout
```
felixarntz, line 123: The legacy `Layout` component normally shouldn't be used here, but since this is a data table that has some special formatting ramifications, let's keep it like that for now and cover it via #2022.

assets/js/modules/adsense/components/dashboard/DashboardTopEarningPagesWidget.js
```diff
@@ -0,0 +1,140 @@
...
+			return (
+				<Layout className="googlesitekit-top-earnings-pages" fill>
+					<AdSenseLinkCTA />
+				</Layout>
+			);
+		}
+		return getDataErrorComponent( 'adsense', error.message, false, false, false, error );
+	}
+
+	if ( ! data || ! data[ 0 ]?.data?.rows ) {
```
felixarntz, line 88:
```suggestion
	if ( ! data || ! data.length || ! data[ 0 ]?.data?.rows ) {
```
To avoid invalid access of `data[ 0 ]` after.

assets/js/modules/adsense/components/dashboard/DashboardTopEarningPagesWidget.js
```diff
@@ -0,0 +1,140 @@
...
+
+	if ( error ) {
+		// Specifically looking for string "badRequest"
+		if ( 'badRequest' === error?.data?.reason ) {
+			return (
+				<Layout className="googlesitekit-top-earnings-pages" fill>
+					<AdSenseLinkCTA />
+				</Layout>
+			);
+		}
```
felixarntz, line 84: Instead of having this (somewhat inaccurate) logic here, we should rely on the `getAdsenseLinked` selector of the `modules/analytics` store. If that returns `false`, let's return the `AdSenseLinkCTA` here.

Let's omit the `Layout` component here, the CTA itself should be enough.

assets/js/modules/adsense/components/dashboard/DashboardTopEarningPagesWidget.js
```diff
@@ -0,0 +1,140 @@
...
+		return getDataErrorComponent( 'adsense', error.message, false, false, false, error );
+	}
+
+	if ( ! data || ! data[ 0 ]?.data?.rows ) {
+		return getNoDataComponent( __( 'Analytics and AdSense', 'google-site-kit' ) );
```
felixarntz, line 89: Instead of this new string, let's use `_x( 'Analytics', 'Service name', 'google-site-kit' )` like elsewhere because that is where the data is coming from.

felixarntz (changes requested):
Some initial feedback here.

Since this looks like it's built on top of the PR for #1901 though, I'll continue review in more detail once that is merged.

assets/js/modules/adsense/components/dashboard/DashboardTopEarningPagesWidget.js
```diff
@@ -0,0 +1,138 @@
...
+	} );
+
+	if ( loading ) {
+		return (
+			<PreviewTable rows={ 5 } padding />
+		);
+	}
+
+	if ( error ) {
+		return getDataErrorComponent( 'adsense', error.message, false, false, false, error );
```
felixarntz, line 79: This should use `analytics` I think because the API request is made to Analytics.

stories/utils/generate-widget-stories.js
```diff
@@ -36,12 +36,12 @@ const { components: { Widget } } = Widgets;
...
- * @param {string} moduleSlug Module slug.
- * @param {string|null} url Current entity URL.
- * @param {Function} cb Callback for additional setup.
+ * @param {(string|Array)} moduleSlugs Module slug or slugs to activate.
+ * @param {string|null}   url          Current entity URL.
+ * @param {Function}     cb            Callback for additional setup.
  * @return {Function} A function to set up registry for widget stories.
  */
-function getSetupRegistry( moduleSlug, url, cb = () => {} ) {
+function getSetupRegistry( moduleSlugs, url, cb = () => {} ) {
```
felixarntz, line 44: I'd prefer to still keep this called `moduleSlug` just because that's the usual use-case (a single one).

stories/utils/generate-widget-stories.js
```diff
@@ -65,15 +77,16 @@ function getSetupRegistry( moduleSlug, url, cb = () => {} ) {
...
- * @param {Component} args.component               Widget component.
- * @param {boolean} args.wrapWidget                Whether to wrap in default <Widget> component. Default true.
+ * @param {Object}    args                              Widget arguments.
+ * @param {string}    args.moduleSlug                   Module slug.
+ * @param {string}    args.datastore                    Module datastore name.
+ * @param {string}    args.group                        Stories group name.
+ * @param {Array}     args.data                         Widget data.
+ * @param {Object}    args.options                      Arguments for report requests.
+ * @param {Object}    [args.additionalVariants]         Optional. Additional story variants.
+ * @param {Object}    [args.additionalVariantCallbacks] Optional. Additional custom callbacks to be run for each of the variants
```
felixarntz, line 87: With those being optional, they should be at the bottom of the list.

stories/utils/generate-widget-stories.js
```diff
@@ -82,6 +95,7 @@ export function generateReportBasedWidgetStories( {
 	group,
 	data,
 	options,
+	additionalVariants = {},
 	additionalVariantCallbacks = {},
```
felixarntz, line 99: The behavior here feels a bit confusing to me, maybe only because these two are named so similarly but do actually unrelated things. `additionalVariants` adds actual additional variants while `additionalVariantCallbacks` expands already existing variants with extra data. Furthermore I think it's odd that they don't support each other, e.g. I currently cannot provide an `additionalVariantsCallback` for one of the `additionalVariants`.

I suggest the following:
* Make `additionalVariants` a map of `variantName => variantFixtureData` pairs where `variantFixtureData` is an object that has `data` and `options`. The utility should then fill this data the same way as for the automatic existing variants.
* And then, add support for `additionalVariants` to `additionalVariantCallbacks` so that you can also provide extra logic for an additional variant in an additional variant callback.

felixarntz (changes requested):
Mostly LGTM, however I think there's room for iteration on the Storybook utility changes.

stories/utils/generate-widget-stories.js
```diff
@@ -65,15 +77,16 @@ function getSetupRegistry( moduleSlug, url, cb = () => {} ) {
...
+ * @param {Object} args                             Widget arguments.
+ * @param {string} args.moduleSlug                  Module slug.
+ * @param {string} args.datastore                   Module datastore name.
+ * @param {string} args.group                       Stories group name.
+ * @param {Array} args.data                         Widget data.
+ * @param {Object} args.options                     Arguments for report requests.
+ * @param {Component} args.component                Widget component.
+ * @param {boolean} args.wrapWidget                 Whether to wrap in default <Widget> component. Default true.
+ * @param {Array} [args.additionalVariants]         Optional. Additional story variants.
+ * @param {Array} [args.additionalVariantCallbacks] Optional. Additional custom callbacks to be run for each of the variants
```
felixarntz, line 89: Indentation here should be fixed, all parameter names and parameter descriptions should be aligned.

stories/utils/generate-widget-stories.js
felixarntz: @ryanwelcher I think the change you added here now makes it more complicated than it was before - why change `additionalVariants` to be an array? This makes the code unnecessarily complex. The format of `additionalVariantCallbacks` shouldn't have been changed, and `additionalVariants` should be a map (i.e. object) so that they can be easily looked up by name. Having the name as first value of an array is counter-intuitive I'd say, plus it requires iterating through an array in several places which can be avoided by using an object.

felixarntz (changes requested):
One indentation issue, but more importantly I think the utility for generating widget stories still needs an iteration, it's now more complicated than before IMO.

stories/utils/generate-widget-stories.js
```diff
@@ -158,6 +173,29 @@ export function generateReportBasedWidgetStories( {
 		},
 	};
 
+	// custom variants.
+	const customVariants = {};
+	Object.keys( additionalVariants ).forEach( ( name ) => {
+		const { data: variantData, options: variantOptions } = additionalVariants[ name ];
```
felixarntz, line 179: If `variantOptions` is an array, we should have here the same type of validation as happens for `options` further above.

stories/utils/generate-widget-stories.js
```diff
@@ -158,6 +173,29 @@ export function generateReportBasedWidgetStories( {
...
+		customVariants[ name ] = ( { dispatch } ) => {
+			if ( Array.isArray( variantOptions ) ) {
+				options.forEach( ( option, index ) => {
```
felixarntz, line 182: This needs to use `variantOptions`.

stories/utils/generate-widget-stories.js
```diff
@@ -65,15 +77,16 @@ function getSetupRegistry( moduleSlug, url, cb = () => {} ) {
...
- * @param {Component} args.component               Widget component.
- * @param {boolean} args.wrapWidget                Whether to wrap in default <Widget> component. Default true.
+ * @param {Object}    args                              Widget arguments.
+ * @param {string}    args.moduleSlug                   Module slug.
+ * @param {string}    args.datastore                    Module datastore name.
+ * @param {string}    args.group                        Stories group name.
+ * @param {Array}     args.data                         Widget data.
+ * @param {Object}    args.options                      Arguments for report requests.
+ * @param {Component} args.component                    Widget component.
+ * @param {boolean}   args.wrapWidget                   Whether to wrap in default <Widget> component. Default true.
```
felixarntz, line 87: Missed this before, should also be marked as optional as it is.

stories/utils/generate-widget-stories.js
```diff
@@ -158,6 +173,29 @@ export function generateReportBasedWidgetStories( {
...
+			if ( Array.isArray( variantOptions ) ) {
+				options.forEach( ( option, index ) => {
+					dispatch( datastore ).receiveGetReport( variantData[ index ], { options: option } );
+				} );
+			} else {
+				dispatch( datastore ).receiveGetReport( variantData, { options: variantOptions } );
+			}
+			const { [ name ]: callback } = additionalVariantCallbacks;
+			if ( callback ) {
+				callback( dispatch, data, options );
```
felixarntz, line 190: Nit-picking, but I think this could be simplified to
```suggestion
			if ( additionalVariantCallbacks[ name ] ) {
				additionalVariantCallbacks[ name ]( dispatch, data, options );
```

stories/utils/generate-widget-stories.js
```diff
@@ -36,9 +36,9 @@ const { components: { Widget } } = Widgets;
  *
  * @since 1.16.0
  *
- * @param {string} moduleSlug Module slug.
- * @param {string|null} url Current entity URL.
- * @param {Function} cb Callback for additional setup.
+ * @param {(string|Array)} moduleSlug Module slug or slugs to activate.
+ * @param {string|null}   url          Current entity URL.
+ * @param {Function}     cb            Callback for additional setup.
```
felixarntz, line 41: Incorrect indentation here (was overlooked before).

stories/module-adsense-components.stories.js
```diff
@@ -21,14 +21,18 @@
...
 import { STORE_NAME } from '../assets/js/modules/adsense/datastore';
+import { STORE_NAME as ANALYTICS_STORE } from '../assets/js/modules/analytics/datastore';
 import {
 	dashboardSummaryWidgetTodayData,
 	dashboardSummaryWidgetPeriodData,
 	dashboardSummaryWidget28DailyData,
 	dashboardSummaryWidgetTodayOptions,
 	dashboardSummaryWidgetPeriodOptions,
 	dashboardSummaryWidget28DailyOptions,
+	dashboardTopEarningPageWidgetArgs,
```
felixarntz, line 34: This should probably use `Options` instead of `Args` like elsewhere.

felixarntz (changes requested):
Almost LGTM, a few minor improvements to possibly make, only one blocking problem.

felixarntz (approved):
Great!

Issue #1691, opened 2020-06-18
Default selected view can be empty if view doesn't exist


aaemnnosttv:
## Bug Description

In #1404 we updated the logic in the Analytics setup and settings components to pre-select the default profile for a property if that property has a `defaultProfileId`.

This is a bit of an edge case, but if that profile is deleted, it will still be around for 30 days or so (GA refers to this as sending it to the "Trash Can"). While the deleted profile will not be returned by the API as a profile for the property, but the property will still references the deleted profile ID as its `defaultProfileId`. This causes the profile ID to be set to an ID that isn't a valid option in the list of choices, resulting in a blank selection.

## Steps to reproduce

1. In Google Analytics, go to Admin > Property Settings and select a Default View
1. Delete the view chosen as the default
1. In Site Kit, set up Analytics (or go to settings edit)
1. Select the account and property with the default view configured previously
1. See that the View is empty

## Screenshots

![image](https://user-images.githubusercontent.com/1621608/85038468-48521500-b18f-11ea-9b7f-cecc1f635f5f.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The default view for a property should only be selected if the view exists  
This should be fine as properties and profiles are all received before this selection happens

## Implementation Brief

* Update `selectProperty` selector to check that `profiles` array contains a profile with id that equals to `property.defaultProfileId` before calling `setProfileId` with a default value.
  * If such profile doesn't exist, skip setting the default value.

## QA Brief

* Follow the steps reproduce above and confirm that the View populates with "Set up a new View"

## Changelog entry

* Only select Analytics default view for the active property automatically if it still exists.

felixarntz:
IB ✅

tofumatt:
QA ✅

PR #1969, opened 2020-08-28
Bug/1691 default selected can be empty

ryanwelcher:
## Summary

Addresses issue #1691

## Checklist


aaemnnosttv (changes requested):
This looks like it'll do the trick! One suggestion regarding the implementation, otherwise please add a test for this for `selectProperty` 🙂 

Note: there are no tests for `selectProperty` yet which is a bit odd. I realize this is a bit out of scope here, but if you could add a few tests to cover this, it would be well worth the extra time needed.

assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -123,6 +123,67 @@ describe( 'modules/analytics properties', () => {
...
+					/^\/google-site-kit\/v1\/modules\/analytics\/data\/properties-profiles/,
+					{ body: fixtures.propertiesProfiles, status: 200 }
+				);
+				const propertyID = fixtures.propertiesProfiles.properties[ 0 ].id;
+				expect( registry.select( STORE_NAME ).getPropertyID() ).toBeUndefined();
+				registry.dispatch( STORE_NAME ).selectProperty( propertyID );
+				expect( registry.select( STORE_NAME ).getPropertyID() ).toBeUndefined();
+			} );
+
+			it( 'selects the property', async () => {
```
aaemnnosttv, line 144: Let's be a little more clear as to what is happening in this test as it asserts more than just the property selection 👍 
```suggestion
			it( 'selects the property and its default profile when set', async () => {
```

assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -123,6 +123,67 @@ describe( 'modules/analytics properties', () => {
...
+				const accountID = fixtures.propertiesProfiles.properties[ 0 ].accountId;
+				const propertyID = fixtures.propertiesProfiles.properties[ 0 ].id;
+				registry.dispatch( STORE_NAME ).setAccountID( accountID );
+				registry.dispatch( STORE_NAME ).selectProperty( propertyID );
+
+				await subscribeUntil( registry,
+					() => (
+						registry.select( STORE_NAME ).getProperties( accountID ) !== undefined
+					),
+				);
```
aaemnnosttv, line 158: We only need to wait for resolvers for selectors. For async actions, we can simply `await` the dispatch 🙂 

```suggestion
				await registry.dispatch( STORE_NAME ).selectProperty( propertyID );

```

assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -123,6 +123,67 @@ describe( 'modules/analytics properties', () => {
...
+				expect( registry.select( STORE_NAME ).getPropertyID() ).toMatch( propertyID );
+				expect( registry.select( STORE_NAME ).getProperties( accountID ) ).toEqual( fixtures.propertiesProfiles.properties );
+				expect( registry.select( STORE_NAME ).getInternalWebPropertyID() ).toEqual( fixtures.propertiesProfiles.properties[ 0 ].internalWebPropertyId );
+				expect( registry.select( STORE_NAME ).getProfileID() ).toEqual( fixtures.propertiesProfiles.properties[ 0 ].defaultProfileId );
+			} );
+
+			it( 'does not set the propertyID if default is not found', async () => {
```
aaemnnosttv, line 165: Do you mean the profileID? It's a bit unclear here as you're asserting that the `propertyID` is empty which would mean the property was not found in state.

We should also have a test for making sure that if `selectProperty` is called with a property ID of a property that _has_ a `defaultProfileId` set but no matching profile in state (because it may not exist), then the `profileID` should _not be set_ to the property's `defaultProfileId`. That may be what you're aiming for here but just want to make sure we have this covered as well.

assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -123,6 +123,67 @@ describe( 'modules/analytics properties', () => {
 				expect( properties ).toEqual( undefined );
 			} );
 		} );
+		describe( 'selectProperty', () => {
```
aaemnnosttv, line 126: Please keep a blank line between `describe` and `it` blocks 👍 
```suggestion

		describe( 'selectProperty', () => {
```

assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -123,6 +123,67 @@ describe( 'modules/analytics properties', () => {
...
+
+			it( 'does not set the propertyID if default is not found', async () => {
+				fetchMock.get(
+					/^\/google-site-kit\/v1\/modules\/analytics\/data\/properties-profiles/,
+					{ body: fixtures.propertiesProfiles, status: 200 }
+				);
+				fetchMock.getOnce(
+					/^\/google-site-kit\/v1\/modules\/analytics\/data\/profiles/,
+					{ body: fixtures.propertiesProfiles.profiles, status: 200 }
+				);
```
aaemnnosttv, line 173: Similar to our tests for selectors with resolvers, let's only mock the fetch requests in one test that hits them all (I think there's only these two anyways). All other tests for this action can then have the data received directly with the respective `receiveGet*` actions.

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -138,49 +138,51 @@ const baseActions = {
 	 *
 	 * @param {string} propertyID Property ID to select.
 	 * @param {string} [internalPropertyID] Internal property ID (if available).
+	 * @return {void}
```
aaemnnosttv, line 141: This is actually a [`Generator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator) instance now, so we can use this as the type.
```suggestion
	 * @return {Generator}
```

aaemnnosttv (changes requested):
Thanks @ryanwelcher ! This is looking good, just a few changes and suggestions to make, mostly around the tests. Let me know if you have any questions 👍

assets/js/modules/analytics/datastore/__fixtures__/select-property-property-profiles.json
```diff
@@ -0,0 +1,81 @@
+{
```
aaemnnosttv, line 1: Rather than create a new fixture variant here, I think we can use the existing `properties-profiles.json` and simply update the `defaultPropertyId` to be a non-existent profile ID (e.g. `99999`). We only need to do this on a copy of the fixture to ensure we're not interfering with other tests that use it.
e.g.
```js
const nonExistentProfileID = "99999";
const propertiesProfiles = {
	...fixtures.propertiesProfiles,
	properties: fixtures.propertiesProfiles.properties.map(
		( property ) => { ...property, defaultProfileId: nonExistentProfileID }
	)
};
// ...
expect( getProfiles().some( ( { id } ) => id === nonExistentProfileID ) ).toBe( false );
expect( getProfileID() ).not.toBe( nonExistentProfileID )
```

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -138,49 +138,50 @@ const baseActions = {
...
+				registry.dispatch( STORE_NAME ).setProfileID( property.defaultProfileId ); // Capitalization rule exception: defaultProfileId
+				return;
+			}
 
-		const matchedProfile = profiles.find( ( { webPropertyId } ) => webPropertyId === propertyID ) || { id: PROFILE_CREATE }; // Capitalization rule exception: webPropertyId
+			// Clear any profile ID selection in the case that selection falls to the getProfiles resolver.
+			registry.dispatch( STORE_NAME ).setProfileID( '' );
+			if ( profiles === undefined ) {
+				return; // Selection will happen in in getProfiles resolver.
+			}
```
aaemnnosttv, line 180: I think this should probably still happen before `getProfiles` is called to ensure that clearing the ID happens before the resolver runs.

assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -123,6 +123,52 @@ describe( 'modules/analytics properties', () => {
...
+				expect( () => {
+					registry.dispatch( STORE_NAME ).selectProperty();
+				} ).toThrow( 'A valid propertyID selection is required.' );
+			} );
+
+			it( 'returns if the accountID is not set', () => {
+				fetchMock.get(
+					/^\/google-site-kit\/v1\/modules\/analytics\/data\/properties-profiles/,
+					{ body: fixtures.propertiesProfiles, status: 200 }
+				);
```
aaemnnosttv, line 138: Please receive these directly rather than using a fetch mock for tests that don't test fetch actions or resolvers.

assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -123,6 +123,52 @@ describe( 'modules/analytics properties', () => {
...
+				expect( registry.select( STORE_NAME ).getPropertyID() ).toBeUndefined();
+				registry.dispatch( STORE_NAME ).selectProperty( propertyID );
+				expect( registry.select( STORE_NAME ).getPropertyID() ).toBeUndefined();
+			} );
+
+			it( 'selects the property and its default profile when set', async () => {
+				fetchMock.get(
+					/^\/google-site-kit\/v1\/modules\/analytics\/data\/properties-profiles/,
+					{ body: fixtures.propertiesProfiles, status: 200 }
+				);
```
aaemnnosttv, line 149: Please receive these directly rather than using a fetch mock for tests that don't test fetch actions or resolvers.

aaemnnosttv (commented):
Just a few remaining changes to address here. Should be fairly quick. Let me know if you have any questions!

aaemnnosttv (dismissed):
Thanks @ryanwelcher – this looks great now and is almost ready to merge, I just noticed a few silent rejected promise errors that are actually unrelated to your changes here that are being raised in the same file. I'll fix these up and then this will be good to go!

aaemnnosttv (approved):

aaemnnosttv:
@googlebot I consent.

Issue #1886, opened 2020-08-06
Event tracking timeout warning raised unnecessarily


aaemnnosttv:
## Bug Description

Currently in the Analytics module, we default to excluding logged in users from measurement for the selected view chosen during module setup.

Originally this would simply prevent the Analytics tracking snippet from being output when users were logged-in. In #944, we revised this to use the GA opt-out script (and respective AMP configuration) to opt-out of all GA measurement instead to ensure this setting was respected, regardless of where Google Analytics tracking was implemented (such as GTM).

In https://github.com/google/site-kit-wp/issues/1630, we updated our internal event tracking to use the `event_callback` parameter to ensure these events are being sent. In the event that the callback takes longer than 1 second to be invoked, we raise a warning about it. This is currently visible using the default GA settings when hovering the Site Kit admin bar menu:

> Tracking event "page_stats_view" (category "admin_bar") took too long to fire.

This is because the event is never tracked (as expected) due to the opt-out which is in effect via the GA module setting to exclude all logged in users. This is because our `trackEvent` function pushes the event onto the data layer regardless of the opt-out, and the callback is never invoked because of the opt-out.

So everything is working as expected, but we're raising an unnecessary warning in the console that could be easily prevented.

In the future, we may be able to change the logic used for excluding logged-in users from measurement in GA to use the [property ID specific method](https://developers.google.com/analytics/devguides/collection/analyticsjs/user-opt-out), but this wasn't an option previously as we didn't have a way to know what property ID was being used in GTM, although this is coming in #1381.

## Steps to reproduce

1. Set up Site Kit with Analytics configured using default settings
1. Be sure to opt into tracking on the Settings screen.
1. Navigate to a page on the front end that has data in Search Console or Analytics
1. Open the browser console
1. Hover the Site Kit admin bar item
1. See warning about event that took too long to fire

## Screenshots

![image](https://user-images.githubusercontent.com/1621608/89526639-b8286780-d7f0-11ea-98c5-dc1aba5e2694.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Internal event tracking should not raise warnings for events that took too long to fire when the GA opt-out is in effect.

## Implementation Brief

* Update `trackEvent` to return early if `global._gaUserPrefs?.ioo?()`
* Update tests for `trackEvent` accordingly

## QA Brief

1. Set up Site Kit with Analytics configured using default settings
[...]

PR #1970, opened 2020-08-28
Bug/1886 event tracking timeout warning

ryanwelcher:
## Summary

Addresses issue #1886

## Checklist


assets/js/util/tracking/createTrackEvent.js
```diff
@@ -32,6 +32,10 @@ export default function createTrackEvent( config, dataLayerTarget ) {
 			userIDHash,
 		} = config;
 
+		if ( global._gaUserPrefs?.ioo() ) {
```
aaemnnosttv, line 35: Even though this is rather unlikely, we should also include `ioo` in the chain to be extra safe.
```suggestion
		if ( global._gaUserPrefs?.ioo?.() ) {
```

aaemnnosttv (changes requested):
Looks good, just the one very small suggestion. Also, this should get an added test to the existing tests for `trackEvent`. We only really need to check that `dataLayer.push` is called when expected and that it is not called when the opt-out function is set and returns true (all other factors being equal).

This will require a few changes to the functions because currently the use of the global/window is nicely abstracted using the `dataLayerTarget` in `createTracking`. However, the opt-out function can _only_  be in this one place so it can't fit into the existing abstraction (even though we happen to reference the same object for the target). We'll need to add another parameter in these functions (where necessary) to pass the global object for dependency injection (see `getLocale` for an example) to use in tests for populating the presence of the opt-out function (rather than touching the real `global`). We'll need to add this to `createTracking` so that it can be passed through to `createTrackEvent` as well (an added new parameter to both functions).

I hope that doesn't sound too complicated – it should be fairly quick to do though so let me know if you have any questions 😄

aaemnnosttv (approved):
Super 👍

Issue #1978, opened 2020-08-31
Fix URL-based entity detection within the admin


felixarntz:
As found during testing #1896, entity detection may not work correctly in the admin when viewing stats for the homepage. After some investigation, I found that this is because `WP_Query` looks at `is_admin()` as part of the logic to determine whether the current context `is_home`. Because the Site Kit URL details view is loaded in the admin, `WP_Query::$is_home` will never be set to `true` as it should - this is why the logic works in the frontend (e.g. for the admin bar) and passes our tests, but not within the backend.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* URL-based entity detection for the home page should also work within the admin.

## Implementation Brief

There is a single `is_admin()` check in `WP_Query::parse_query()` which causes this problem - everywhere else (fortunately) `WP_Query` seems to not rely on contextual information.

* Introduce `Google\Site_Kit\Core\Util\WP_Context_Switcher_Trait` with static methods `with_frontend_context()` and `with_admin_context()`, which allows to temporarily change WordPress context between frontend and admin (via `$current_screen`) and returns a closure to restore the previous context.
    * Implement the behavior by overriding the `$current_screen` WordPress global.
    * Document that the closure to restore should always be called as soon as possible after the context switch.
* In `WP_Query_Factory::from_url` use `WP_Context_Switcher_Trait::with_frontend_context()` before calling `WP_Query::parse_query`. Then restore the original context.
* In `Entity_Factory::from_url` use `WP_Context_Switcher_Trait::with_frontend_context()` before calling `WP_Query::get_posts` (unfortunately this re-parses the query, there is no way to get around it). Then restore the original context.
* Merge #1979.

## QA Brief

1. Use a site that has some stats on Search Console / Analytics.
2. Go to your home page and open the Site Kit admin bar menu.
3. Click on More Details.
    * Prior to this fix, you should see the box that is supposed to contain the title be empty. The stats there incorrectly display stats for the entire site, not the URL you're looking for.
    * After this fix, you should see the home page title in the box, alongside the correct stats.

## Changelog entry

* Fix bug with URL-based entity detection where home page stats in Site Kit URL details view would not show up as expected.

aaemnnosttv:
@felixarntz the IB looks good for the most part, but I think we can simplify this a bit.

Similar to `Module::with_client_defer`, we only really need a single method `with_context( $hook_name )` which might look something like this:

```php
[...]

PR #1979, opened 2020-08-31
Fix/1978 url based entity detection

felixarntz:
## Summary

Addresses issue #1978

## Relevant technical choices

* This iterates on #1981, so that should be reviewed/merged first.
* Slightly different from the IB, I kept `WP_Context_Switcher` a class. With its solely static functionality, it didn't really make sense to have it as a trait that other classes inherit. Theoretically it could be made non-static, but our use-case for it are already static classes so that wouldn't work. I think the main win here in the IB iterations is the use of `with_*_context` which returns a closure to restore the original context, lmk if you see additional value in a trait over the static class here.

## Checklist

### DIFF
--- a/includes/Core/Util/WP_Context_Switcher.php
+++ b/includes/Core/Util/WP_Context_Switcher.php
@@ -1,0 +1,100 @@
+<?php
+/**
+ * Class Google\Site_Kit\Core\Util\WP_Context_Switcher
+ *
+ * @package   Google\Site_Kit\Core\Util
+ * @copyright 2020 Google LLC
+ * @license   https://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
+ * @link      https://sitekit.withgoogle.com
+ */
+
+namespace Google\Site_Kit\Core\Util;
+
+/**
+ * Class for temporarily switching WordPress context, e.g. from admin to frontend.
+ *
+ * @since n.e.x.t
+ * @access private
+ * @ignore
+ */
+class WP_Context_Switcher {
+	const CONTEXT_FRONT = 'front';
+	const CONTEXT_ADMIN = 'admin';
+
+	/**
+	 * Stack of original contexts switched from.
+	 *
+	 * @since n.e.x.t
+	 * @var array
+	 */
+	private static $context_stack = array();
+
+	/**
+	 * Switches WordPress context if necessary.
+	 *
+	 * Context is only switched if WordPress is not already in the given context. Context should only ever be switched
+	 * temporarily. Call {@see WP_Context_Switcher::restore_context()} as soon as possible after to restore the
+	 * original context.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $context WordPress context to switch to. Either 'front' or 'admin'.
+	 * @return boolean True if context was switched, false otherwise.
+	 */
+	public static function switch_context( $context ) {
+		global $current_screen;
+
+		switch ( $context ) {
+			case self::CONTEXT_ADMIN:
+				if ( is_admin() ) {
+					return false;
+				}
+				self::load_wp_screen_api();
+				self::$context_stack[] = $current_screen;
+				$current_screen        = \WP_Screen::get( 'index' ); // phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
+				return true;
+			case self::CONTEXT_FRONT:
+				if ( ! is_admin() ) {
+					return false;
+				}
+				self::load_wp_screen_api();
+				self::$context_stack[] = $current_screen;
+				$current_screen        = \WP_Screen::get( 'front' ); // phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
+				return true;
+		}
+
+		return false;
+	}
+
+	/**
+	 * Restores previous WordPress context if it has been switched.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return boolean True if context was restored, false otherwise.
+	 */
+	public static function restore_context() {
+		global $current_screen;
+
+		if ( empty( self::$context_stack ) ) {
+			return false;
+		}
+
+		$current_screen = array_pop( self::$context_stack ); // phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
+		return true;
+	}
+
+	/**
+	 * Loads the WordPress Screen API if needed.
+	 *
+	 * @since n.e.x.t
+	 */
+	private static function load_wp_screen_api() {
+		if ( class_exists( 'WP_Screen' ) && function_exists( 'get_current_screen' ) ) {
+			return;
+		}
+
+		require_once ABSPATH . 'wp-admin/includes/class-wp-screen.php';
+		require_once ABSPATH . 'wp-admin/includes/screen.php';
+	}
+}
--- a/includes/Core/Util/WP_Context_Switcher.php
+++ b/includes/Core/Util/WP_Context_Switcher.php
@@ -42,28 +42,6 @@ class WP_Context_Switcher {
 	}
 
 	/**
-	 * Switches to WordPress admin context if necessary.
-	 *
-	 * Context is only switched if WordPress is not already in admin context. Context should only ever be switched
-	 * temporarily. Call the returned closure as soon as possible after to restore the original context.
-	 *
-	 * @since n.e.x.t
-	 *
-	 * @return callable Closure that restores context and returns true if context was restored or false otherwise.
-	 */
-	public static function with_admin_context() {
-		global $current_screen;
-
-		if ( is_admin() ) {
-			return function() {
-				return false;
-			};
-		}
-
-		return self::switch_current_screen( 'index' );
-	}
-
-	/**
 	 * Switches the current WordPress screen via the given screen ID or hook name.
 	 *
 	 * @since n.e.x.t
--- a/includes/Core/Util/WP_Context_Switcher.php
+++ b/includes/Core/Util/WP_Context_Switcher.php
@@ -18,8 +18,6 @@ namespace Google\Site_Kit\Core\Util;
  * @ignore
  */
 class WP_Context_Switcher {
-	const CONTEXT_FRONT = 'front';
-	const CONTEXT_ADMIN = 'admin';
 
 	/**
 	 * Stack of original contexts switched from.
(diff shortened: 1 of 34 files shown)


felixarntz:
@aaemnnosttv I found another bug with entity detection causing tests to fail now https://travis-ci.com/github/google/site-kit-wp/jobs/379625031

It's because we don't correctly support 404s. I'll work on another issue to address that.

includes/Core/Util/WP_Context_Switcher.php
```diff
@@ -0,0 +1,124 @@
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return callable Closure that restores context and returns true if context was restored or false otherwise.
+	 */
+	private static function get_restore_closure() {
+		return static function() {
+			global $current_screen;
+
+			if ( empty( self::$context_stack ) ) { // phpcs:ignore WordPressVIPMinimum.Variables.VariableAnalysis.SelfInsideClosure
```
aaemnnosttv, line 102: I thought about possibly using a stack for this but I don't think there is a need for it. It shouldn't really matter in our current use as I don't think we have any nested context switching, but I think restoring in this way is riskier than simply restoring the previous screen as I originally suggested.

Re-setting the previous screen is an idempotent operation which guarantees that things are in fact back the way they were after the restore is called. Using a stack, every restore function is the same and is capable of restoring layers that it didn't "push".

E.g.
```php
$restore();
$restore();
$restore();
```

This is a contrived example of course, but just showing that this approach could introduce problems that wouldn't exist with the other. My [suggested approach](https://github.com/google/site-kit-wp/issues/1978#issuecomment-684014717) still supports nesting if indeed needed and is still stack-based of course (using the call stack), but is a bit more deterministic.

includes/Core/Util/WP_Context_Switcher.php
```diff
@@ -0,0 +1,124 @@
...
+		self::$context_stack[] = $current_screen;
+		$current_screen        = \WP_Screen::get( $screen_id ); // phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
+	}
+
+	/**
+	 * Gets closure to restore previous WordPress context if it has been switched.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return callable Closure that restores context and returns true if context was restored or false otherwise.
```
aaemnnosttv, line 96: I don't think the return value of the restore function is necessary – at least as a boolean. It should always restore the screen (it's simply setting the global which has zero cost), so the return value is not really useful as a success/failure if it always works 🙂 . If anything, it could return the restored screen instance, so touching `$current_screen` after that point wouldn't be necessary.

includes/Core/Util/WP_Context_Switcher.php
```diff
@@ -0,0 +1,124 @@
...
+	 */
+	public static function with_frontend_context() {
+		global $current_screen;
+
+		if ( ! is_admin() ) {
+			return function() {
+				return false;
+			};
+		}
```
aaemnnosttv, line 47: As mentioned in the other method, I think it would be better to replace the early return with a conditional around the `switch` method – the restore function would simply restore the current screen (essentially a no-op), unless something else had changed the screen, in which case this would restore it as expected.

I don't think it matters though in this case since `front` requests don't have screens. It only really matters in the admin case to preserve the current screen.

includes/Core/Util/WP_Context_Switcher.php
```diff
@@ -0,0 +1,124 @@
...
+	public static function with_admin_context() {
+		global $current_screen;
+
+		if ( is_admin() ) {
+			return function() {
+				return false;
+			};
+		}
+
+		self::switch_current_screen( 'index' );
```
aaemnnosttv, line 72: `index` is an alias for `dashboard` which I think is a bit more expressive to read as well.
```suggestion
		self::switch_current_screen( 'dashboard' );
```

aaemnnosttv (changes requested):
Using a class here makes sense due to the application of the `$context_stack` as you have it, however I think this makes it a bit more complicated than necessary and potentially less stable than what was originally discussed with the IB; see my comments below. There are a few other suggestions here too.

Aside from that, I think it may make more sense to implement as a trait as it would allow us to keep the methods protected since I don't think this is an API we want to expose publicly.

Additionally, this includes changes from #1980 where the IB may still change.

includes/Core/Util/WP_Context_Switcher.php
```diff
@@ -0,0 +1,124 @@
...
+	 * Switches to WordPress admin context if necessary.
+	 *
+	 * Context is only switched if WordPress is not already in admin context. Context should only ever be switched
+	 * temporarily. Call the returned closure as soon as possible after to restore the original context.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return callable Closure that restores context and returns true if context was restored or false otherwise.
+	 */
+	public static function with_admin_context() {
```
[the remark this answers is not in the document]
felixarntz: Sounds good, I'll remove it.

includes/Core/Util/WP_Context_Switcher.php
```diff
@@ -0,0 +1,124 @@
...
+		global $current_screen;
+
+		if ( is_admin() ) {
+			return function() {
+				return false;
+			};
+		}
```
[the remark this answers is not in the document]
felixarntz: I'd rather keep the check in this method since it technically is decoupled from switching the current screen. It wouldn't be a big deal to change, but I think `switch_current_screen` is more explicit as in "DO switch the current screen" whereas `with_*_context` is more implicit, so here the check to _maybe_ not actually switch context makes more sense IMO.

includes/Core/Util/WP_Context_Switcher.php
felixarntz: I think it makes sense to return this for knowing whether context was switched at all. `true|false` here doesn't indicate any sort of error - the method always behaves like it should, but the return value indicates whether a switch was even performed. Which is certainly more relevant for tests than production code, but I think it's worth having.

includes/Core/Util/WP_Context_Switcher_Trait.php
```diff
@@ -0,0 +1,68 @@
...
+	 *
+	 * @return callable Closure that restores context and returns true if context was restored or false otherwise.
+	 */
+	protected static function with_frontend_context() {
+		global $current_screen;
+
+		if ( ! is_admin() ) {
+			return function() {
+				return false;
+			};
```
aaemnnosttv, line 38: Not really a big deal, but if we're going to avoid switching the context in this case, we should probably still restore the current/original screen when this function is called. Even if the main function didn't change it, you could say this function should still behave the same way.

tests/phpunit/integration/ContextTest.php
```diff
@@ -191,7 +191,8 @@ public function test_get_reference_entity__in_admin_context() {
 		set_current_screen( 'toplevel_page_googlesitekit-dashboard' );
 		$_GET['page'] = 'googlesitekit-dashboard';
 		$this->assertNull( $context->get_reference_entity() );
-		$_GET['permaLink'] = home_url( '/non-existent-post/' );
+		// We can probably safely assume that 987654 is not a valid post ID in the test environment.
```
aaemnnosttv, line 194: A fairly safe assumption, but rather than assume, we can always add a quick assertion to make 💯  sure this is the case 😄 
```suggestion
		// We can probably safely assume that 987654 is not a valid post ID in the test environment, but let's make sure.
		$this->assertFalse( WP_Post::get_instance( 987654 ) );
```

aaemnnosttv (approved):
Great! I love how simple it is. Left a few comments but nothing blocking 👍

includes/Core/Util/WP_Context_Switcher_Trait.php
felixarntz: Hmm there would be nothing to restore. One of the main reasons for me to implement this logic to bail early in the first place is that we shouldn't load the extra WP admin files if it isn't needed.

tests/phpunit/integration/ContextTest.php
felixarntz: 👍 😂

includes/Core/Util/WP_Context_Switcher_Trait.php
aaemnnosttv: It would "restore" it back to whatever it was – nothing (i.e. `null`), or a `WP_Screen` with `id` of `front` if one was set.

I don't want to get into the weeds too much on this as it is definitely an edge case, but here's what I mean

```php
$current_screen = null; // A front end context

$restore_context =  self::with_frontend_context();

set_current_screen( 'dashboard' );

$restore_context(); // returns false

$current_screen instanceof WP_Screen; // true; a WP_Screen in admin context
```

I don't think this is relevant for our current use but just thinking about keeping the behavior consistent (not necessarily supporting something somewhere "doing it wrong").

felixarntz: Oh okay, that's fair. With the purpose to "restore" that makes sense, I was probably still more thinking about it from the original PR approach view which was closer to WP core's "switching things approach" where it only considers what was switched from originally but not any other unrelated "switches" of the same global.

includes/Core/Util/WP_Context_Switcher_Trait.php
```diff
@@ -27,18 +27,19 @@ trait WP_Context_Switcher_Trait {
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @return callable Closure that restores context and returns true if context was restored or false otherwise.
+	 * @return callable Closure that restores context.
 	 */
 	protected static function with_frontend_context() {
 		global $current_screen;
```
aaemnnosttv, line 33: This global is no longer referenced in this method.

includes/Core/Util/WP_Context_Switcher_Trait.php
```diff
@@ -47,22 +48,35 @@ protected static function with_frontend_context() {
...
+	 * Returns the closure to restore the current screen.
+	 *
+	 * Calling the closure will restore the `$current_screen` global to what it was set to at the time of calling
+	 * this method.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return callable Closure that restores context.
+	 */
+	private static function get_restore_closure() {
```
aaemnnosttv, line 71: This name is a bit too generic to use in a trait I think; even though it's private, it could still have a naming collision. How about something like this instead?
```suggestion
	private static function get_restore_current_screen_closure() {
```

aaemnnosttv (commented):
Looks great – just two final details and this should be good to go 👍

aaemnnosttv (approved):
👍 !

Issue #1838, opened 2020-07-28
Consistent index.js file in component directories


eugene-manuilov:
## Feature Description

After refactoring in #1654, it turned out that not all directories have sufficient index.js files. We need to create or update index.js files in all component directories to make it consistent accross all directories. A sufficient index.js exports all components in the folder using named exports. The following is the list of directories that need to be updated.

#### Directories with insufficient index files:

- `adsense/components/dashboard`
- `analytics/components/settings`
- `optimize/components/settings`
- `pagespeed-insights/components/settings`
- `search-console/components/settings`
- `tagmanager/components/common`

#### Directories without index files:

- `analytics/components/adminbar`
- `analytics/components/dashboard-details`
- `analytics/components/wp-dashboard`
- `pagespeed-insights/components/common`
- `pagespeed-insights/components/dashboard`
- `search-console/components/adminbar`
- `search-console/components/dashboard-details`
- `search-console/components/dashboard`
- `search-console/components/wp-dashboard`

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Every `modules/*/components` subdirectory should have an `index.js` file which exports all components in that directory.

## Implementation Brief

* Add test to ensure module index consistency (https://github.com/google/site-kit-wp/commit/323fc9703ee5b0be47608a726e0524e963a9e4a3)
* Add missing index modules and exports as identified by running `npm run test:js:watch -- --coverage=false assets/js/modules/index.test.js`
* Finish and merge #1954
	- The above PR only has the added test which will currently fail, all the index modules and exports just need to be added

## QA Brief

* `npm run test:js:watch -- --coverage=false assets/js/modules/index.test.js` should pass 😄 

## Changelog entry

* N/A

tofumatt:
IB ✅

tofumatt:
QA ✅

Created a folder with files that should cause a violation to be raised and both ESLint and the Jest test complained 👍🏻

PR #1982, opened 2020-09-01
Enhancement/1838 consistent module component indexes

caseymorrisus:
## Summary
Addresses issue [#1838](https://github.com/google/site-kit-wp/issues/1838)

## Relevant technical choices
Modifies test within [323fc97](https://github.com/google/site-kit-wp/commit/323fc9703ee5b0be47608a726e0524e963a9e4a3) to mock modules data using fixture to prevent error caused by `AnalyticsDashboardWidgetPopularPagesTable.renderLayout` being called within `withData` HOC params. 

## Checklist


assets/js/modules/index.test.js
```diff
@@ -0,0 +1,58 @@
...
+
+const getComponentNames = ( componentPath ) => fs.readdirSync( componentPath )
+	.filter( ( name ) => ! /^index|\.test\.js$/.test( name ) )
+	.map( ( name ) => name.replace( /\..*/, '' ) );
+
+describe( 'all modules', () => {
+	const fixturesKeyValue = convertArrayListToKeyedObjectMap( FIXTURES, 'slug' );
+
+	beforeAll( () => {
+		global._googlesitekitLegacyData = { modules: fixturesKeyValue };
```
aaemnnosttv, line 42: Rather than set this global here (which is a bit odd that this is needed, but I understand), let's fill out the existing global in the setup file instead since this seems to be about undefined keys in the object rather than the data itself.

https://github.com/google/site-kit-wp/blob/291eaa3c1fdaf3fad788b8e6c4077eb95b7e655e/tests/js/setup-globals.js#L27-L34

Also, this global has a bit more data in the structure than is present in the datastore fixture which represents the REST response from the `core/modules/data/list` datapoint so it might not be sufficient to simply drop-in.
image
Things like the module's `screenID` which is present in the setup file now aren't in the response fixture.

Since there isn't a complete fixture available for this value, let's just fill in the gaps where needed for now.

aaemnnosttv (changes requested):
Thanks @caseymorrisus – this looks great, just the one thing to change regarding the global we need to fill. Let me know if you have any questions 👍

Issue #1932, opened 2020-08-18
Add number of connected users to Site Health information


felixarntz:
Site Kit's Site Health information should include the number of connected users.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Site Kit's Site Health information should contain a row "Connected user count", following "User Status".
* It should display the number of users that currently have a Site Kit access token.

## Implementation Brief

* Add `Debug_Data::get_connected_user_count_field()`:
    * Query user count that have a SIte Kit access token set.
    * Use `label` "Connected user count", and specify the user count as `value`.
        * `debug` isn't necessary since it can use the same as `value`.
* Add `connected_user_count` key after `user_status` in `Debug_Data::get_fields()` and call the new `get_connected_user_count_field()` for it.

## QA Brief

* Set up Site Kit for multiple users.
* Confirm that the Connected user count is shown in Site Health and reflects the correct number of authenticated users.

## Changelog entry

* N/A

aaemnnosttv:
IB ✅

cole10up:
**Tested**

Set up Site Kit with multiple users.

Confirmed connected users count is shown in Site Health and reflects the correct number of authenticated users:

PR #1983, opened 2020-09-01
Enhancement/1932 add users to site health

ryanwelcher:
## Summary

Addresses issue #1932 

## Checklist

### DIFF
--- a/includes/Core/Util/Debug_Data.php
+++ b/includes/Core/Util/Debug_Data.php
@@ -147,27 +147,28 @@ class Debug_Data {
 	 */
 	protected function get_fields() {
 		$fields = array(
-			'version'         => array(
+			'version'              => array(
 				'label' => __( 'Version', 'google-site-kit' ),
 				'value' => GOOGLESITEKIT_VERSION,
 			),
-			'php_version'     => array(
+			'php_version'          => array(
 				'label' => __( 'PHP Version', 'google-site-kit' ),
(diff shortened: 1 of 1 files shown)


includes/Core/Util/Debug_Data.php
```diff
@@ -323,4 +324,26 @@ function ( Module_With_Debug_Fields $module ) {
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array
+	 */
+	private function get_connected_user_count_field() {
+		$users = new \WP_User_Query(
+			array(
+				// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_key
+				'meta_key' => 'wp_googlesitekit_access_token',
```
aaemnnosttv, line 339: A few small details here:
- For the key name, please use the constant `OAuth_Client::OPTION_ACCESS_TOKEN` rather than the literal value
- Regarding the `wp_` – we can't assume this prefix because it varies by table prefix and network mode, so we have `User_Options::get_meta_key` to handle this for us 🙂 

See `\Google\Site_Kit\Core\Authentication\Has_Connected_Admins::query_connected_admins` for an example, only the `meta_key` here needs to change though.

aaemnnosttv (changes requested):
Thanks @ryanwelcher, just one change to make to the query.

aaemnnosttv (approved):
Awesome!

Issue #1381, opened 2020-04-13
Modify Tag Manager module setup with Analytics guidance


felixarntz:
## Feature Description

If the Tag Manager container a user selects includes an Analytics tag, the user should be guided to deal with that properly.

**PR for this should be opened against `feature/ga-gtm-separation`.**

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

When the user selects a GTM container, Site Kit should check whether there's a GA tag configured for that container, via the `getLiveContainerVersion` selector.

Based on this detection, we need the following logic to happen:

### No GA tag in GTM, GA module not active

* Don't do anything (other than what's happening already).

### No GA tag in GTM, GA module is active

* Don't do anything (other than what's happening already).

### GA tag in GTM, GA module not active
**The changes in this section specifically should only apply to GTM set up, not GTM settings.**
* Recommend the user to also activate the GA module:  
	> Looks like you're already using Google Analytics within your Google Tag Manager configuration.
	> Activate the Google Analytics module in Site Kit to see relevant insights in your dashboard.
* Change the primary button for current functionality into a secondary button that says **"Complete setup without Analytics"**. Functionality should remain as-is.
* Add new primary submit button to say **"Continue to Analytics setup"**.  
When clicked, complete setup and activate GA module so that user gets redirected to OAuth consent screen for GA scopes and then GA module setup.

#### Property ID in GA tag in GTM and GA module match

* Inform the user:  
	> "Looks like you're using Google Analytics. Your Analytics property UA-1234567-1 is already set up in your Google Tag Manager configuration, so Site Kit will switch to using Google Tag Manager for Analytics."
* When user clicks button to complete setup, the GA module's `useSnippet` option should be set to `false`.

#### Property ID in GA tag in GTM and GA module match (GA snippet placed by other plugin/theme)

* Block the user from proceeding (since we can't prevent the existing tag from being output):  
	> "Looks like you're using Google Analytics within your Google Tag Manager configuration. [...]"

#### Property ID in GA tag in GTM and GA module don't match

* Block the user from proceeding:
	> "Looks like you're already using Google Analytics within your Google Tag Manager configuration. However, its Analytics property UA-1234568-1 is different from the Analytics property UA-1234567-1, which was previously selected in the plugin. You need to configure the same Analytics property in both places."

### How to deal with primary AMP

* Logic should be exactly the same as for non-AMP - just against the `ampContainerID` instead of the `containerID`.

[...]

PR #1984, opened 2020-09-02
Enhance Tag Manager setup with Analytics guidance

aaemnnosttv:
## Summary

Addresses issue #1381

This PR implements changes primarily around the various Setup components.

## Relevant technical choices

* Created new `SetupFormInstructions` component as a kind of higher-order version of `FormInstructions` which was used in both, but needs to include some additional instructions which are specific to setup
* Created a new `SetupFormSubmitButtons` component to encapsulate the logic around which buttons should show
* Introduces a new `withActive` helper for generating variants of the modules list fixtures
* Introduces a new `createStrictSelect` utility as a private internal helper used for creating a "strict" version of `registry.select`. It creates decorated selectors that will throw an error if `undefined` is returned (more info as to why it is this way in the comment) but this is very useful with the changes to `canSubmitChanges` validation
* Refactored the way `canSubmitChanges` works to use a validation function (mostly a bunch of `invariant` calls along with strict select). This way we can use the validation function in the test as well to see _why_ it returned `false` when it does to ensure we're seeing the result we expect for the right reason. This also makes the validation a bit cleaner as each check is mostly a single line `invariant` call rather than a full conditional with a return.
* Updated the validation for `forms.setValues` to better guard against a cryptic error if you forget the `formID` argument and only pass a single `data` object 😄 

## Checklist


aaemnnosttv:
@felixarntz @tofumatt marking this one ready for review to get some initial feedback. I'm expecting we'll probably want to add some tests, but should be ready for a first pass.

assets/js/modules/tagmanager/datastore/__factories__/utils.js
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Internal dependencies
+ */
+import { STORE_NAME } from '../constants';
+import {
+	buildLiveContainerVersionWeb,
+	buildLiveContainerVersionAMP,
+} from './builders';
+
+export const parseLiveContainerVersionIDs = ( { accountId, containerId, container: { publicId } }, callback ) => {
```
tofumatt, line 28: This function signature is a bit unwieldy, and I'm wondering about the `Id` vs `ID` capitalisation here versus elsewhere. Having `container` be its own variable and then separating out `publicId` inside the function would make this function's signature much easier to read. I think we should do that.

I think the capitalisation exception is due to the way the data is sent from the API, but in that case: let's explicitly mention it like we do [elsewhere](https://github.com/google/site-kit-wp/blob/develop/assets/js/modules/analytics/components/common/PropertySelect.js#L76).

tofumatt (changes requested):
Approach overall seems good, but there are a few things I think we can improve here:

1. Some naming of variables and function signatures could use some tidying up to improve readability
2. Maintaining the `onSubmit` behaviour of the setup form to work with keyboard-only uses and maintain a "primary" action for the form.
3. Adding tests for the `strictSelect` function.
4. A few smaller issues mentioned in the comments.

assets/js/modules/tagmanager/components/setup/SetupForm.js
```diff
@@ -16,64 +16,105 @@
...
-		if ( ! error ) {
+		} else if ( ! error ) {
 			setValues( FORM_SETUP, { autoSubmit: false } );
+			// If the property ID is set in GTM and Analytics is active,
+			// we disable the snippet output via Analyics to prevent duplicate measurement.
+			if ( gtmAnalyticsPropertyID && analyticsModuleActive ) {
+				dispatchAnalytics.setUseSnippet( false );
+				const saveAnalyticsSettings = await dispatchAnalytics.saveSettings();
+				if ( saveAnalyticsSettings.error ) {
+					receiveError( saveAnalyticsSettings.error );
```
felixarntz, line 78: We probably shouldn't introduce new usages of the legacy `receiveError` way. This error should exist in the `modules/analytics` store automatically, so we should probably reference it from there. We could create a custom `TagManagerSetupErrorNotice` component which looks at the two specific errors for `saveSettings` and `activeModule` (see below), and only displays the regular `StoreErrorNotice` component otherwise.

assets/js/modules/tagmanager/components/setup/SetupForm.js
```diff
@@ -16,64 +16,105 @@
...
+					return;
+				}
+			}
+			// If submitting with Analytics setup, and Analytics is not active,
+			// activate it, and navigate to its reauth/setup URL to proceed with its setup.
+			if ( submitMode === 'with_analytics_setup' && ! analyticsModuleActive ) {
+				setIsNavigating( true );
+				const activateAnalytics = await activateModule( 'analytics' );
+				if ( activateAnalytics.error ) {
+					receiveError( activateAnalytics.error );
```
felixarntz, line 88: See above.

assets/js/modules/tagmanager/components/setup/SetupForm.js
```diff
@@ -16,64 +16,105 @@
...
 		if ( autoSubmit && hasEditScope ) {
-			submitForm( { preventDefault: () => {} } );
+			submitForm( { submitMode } );
 		}
-	}, [ hasEditScope, autoSubmit, submitForm ] );
+	}, [ hasEditScope, autoSubmit, submitForm, submitMode ] );
+
+	// Form submit behavior now varies based on which button is clicked,
+	// so we no longer support the generic form submit event.
+	const preventDefaultFormSubmit = useCallback( ( event ) => event.preventDefault(), [] );
```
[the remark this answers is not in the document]
felixarntz: Agreed, we should be able to use the regular form submit event and detect the current `submitMode` from the form? Maybe we could have the two submit buttons with `name` attributes? This way only the one clicked should actually be sent, if I'm not mistaken.

assets/js/modules/tagmanager/components/setup/SetupForm.js
```diff
@@ -16,64 +16,105 @@
...
+				}
+			}
+			// If submitting with Analytics setup, and Analytics is not active,
+			// activate it, and navigate to its reauth/setup URL to proceed with its setup.
+			if ( submitMode === 'with_analytics_setup' && ! analyticsModuleActive ) {
+				setIsNavigating( true );
+				const activateAnalytics = await activateModule( 'analytics' );
+				if ( activateAnalytics.error ) {
+					receiveError( activateAnalytics.error );
+					return;
```
felixarntz, line 89: I think we'll need to call `setIsNavigating( false )` here to avoid a continuing progress bar in case of this error?

includes/Modules/Tag_Manager.php
```diff
@@ -95,6 +96,18 @@ final class Tag_Manager extends Module
...
+			'googlesitekit_apifetch_preload_paths',
+			function ( $paths ) {
+				$modules_routes = array(
+					// Preload Analytics settings in case it isn't active.
+					'/' . REST_Routes::REST_ROOT . '/modules/analytics/data/settings',
+				);
+
+				return array_merge( $paths, $modules_routes );
+			}
+		);
```
felixarntz, line 109: Is this needed or just an optimization? I wonder whether we can get along without adding it. How much does making the actual REST API request here affect the perceived speed?

stories/modules-tagmanager-setup.stories.js
```diff
@@ -162,6 +184,44 @@ storiesOf( 'Tag Manager Module/Setup/Secondary AMP', module )
...
+			webContainerVersion.container,
+			ampContainerVersion.container,
+		], { accountID } );
+		parseLiveContainerVersionIDs( webContainerVersion, ( { internalContainerID } ) => {
+			registry.dispatch( STORE_NAME ).receiveGetLiveContainerVersion( webContainerVersion, { accountID, internalContainerID } );
+		} );
+		parseLiveContainerVersionIDs( ampContainerVersion, ( { internalContainerID } ) => {
+			registry.dispatch( STORE_NAME ).receiveGetLiveContainerVersion( ampContainerVersion, { accountID, internalContainerID } );
+		} );
+
```
felixarntz, line 224: I think it would be good to add full Storybook coverage here, also for e.g. when there is an Analytics property in GTM and Analytics is already active.

We should probably also add the same coverage for AMP and non-AMP, currently you only cover primary and secondary AMP for the new Storybook items.

felixarntz (changes requested):
Awesome work so far! Just a few things in addition to what @tofumatt pointed out.

assets/js/modules/tagmanager/components/setup/SetupForm.js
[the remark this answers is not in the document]
aaemnnosttv: @felixarntz that would work in a traditional form submission where the form submit is triggering the navigation passing through its values in GET/POST. In this case, we need the choice to be set in the datastore before continuing.
It's true that we have access to the submitting `HTMLElement` in the form submit event so maybe we can use that. It seems somewhat less correct in terms of React to use the DOM for this though, so I'll see what we can do.

@tofumatt – regarding your point about the AC, it does explicitly say that "Continue to Analytics setup" should be the primary action in this case, so if we're going to change this that's okay with me, but just want to confirm this with @felixarntz.

I'll work on restoring the form submit handler rather than relying on the `onClick` handlers, and if we want to swap which button is primary/secondary for some case, that shouldn't be a big deal to change.

assets/js/modules/tagmanager/components/setup/SetupFormInstructions.js
```diff
@@ -0,0 +1,87 @@
...
+	if ( [ gtmAnalyticsPropertyID, hasMultipleAnalyticsPropertyIDs, analyticsPropertyID, analyticsModuleActive ].includes( undefined ) ) {
+		return <FormInstructions />;
+	}
+
+	// Multiple property IDs implies secondary AMP where selected containers don't reference the same Analytics property ID.
+	if ( hasMultipleAnalyticsPropertyIDs ) {
+		/* translators: %1$s: GTM Analytics property ID, %2$s: Analytics property ID */
+		const message = __( 'Looks like you’re already using Google Analytics within your Google Tag Manager configurations. However, the configured Analytics tags reference different property IDs. You need to configure the same Analytics property in both containers.', 'google-site-kit' );
+
+		return <ErrorText message={ sprintf( message, gtmAnalyticsPropertyID, analyticsPropertyID ) } />;
```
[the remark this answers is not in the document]
aaemnnosttv: Oops! You're right, I'll clean that up 👍

assets/js/modules/tagmanager/components/setup/SetupForm.js
```diff
@@ -16,64 +16,105 @@
...
 } from '../common';
 import StoreErrorNotice from '../../../../components/StoreErrorNotice';
+import SetupFormInstructions from './SetupFormInstructions';
+import SetupFormSubmitButtons from './SetupFormSubmitButtons';
 const { useSelect, useDispatch } = Data;
 
-export default function SetupForm( { finishSetup } ) {
-	const canSubmitChanges = useSelect( ( select ) => select( STORE_NAME ).canSubmitChanges() );
+export default function SetupForm( { finishSetup, setIsNavigating } ) {
+	const gtmAnalyticsPropertyID = useSelect( ( select ) => select( STORE_NAME ).getSingleAnalyticsPropertyID() );
```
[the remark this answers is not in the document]
aaemnnosttv: Sure – this is kind of how it was discussed in the IB, but I think there's room for improvement. Maybe `containerAnalyticsPropertyID` or `liveContainerAnalyticsPropertyID` ?

assets/js/modules/tagmanager/components/setup/SetupFormSubmitButtons.js
```diff
@@ -0,0 +1,90 @@
...
+				<Fragment>
+					<Button
+						onClick={ submitFormWithAnalytics }
+						disabled={ ! canSubmitChanges }
+					>
+						{ __( 'Continue to Analytics setup', 'google-site-kit' ) }
+					</Button>
+					<Link
+						className="googlesitekit-setup-module__sub-action"
+						onClick={ submitFormNoAnalytics }
```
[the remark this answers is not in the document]
aaemnnosttv: Changing the secondary action here to be the primary one is a change for @felixarntz and @marrrmarrr to decide, otherwise this is consistent with the current ACs. I'm not against it, but it's a substantial user-facing change 🙂 

Regarding the extraction for this into a separate component – while I agree that it isn't particularly reusable, you could say the same about all other `setup` or `settings` components because they're specific to that context. Our reusable components are essentially those in `common`. Whether the complexity justifies a separate component or not is a bit of a subjective preference I think. Previously the JSX for the submit portion of the form itself was about 3 lines with no additional logic; the extracted component is about 50 (more than 10x!). As it is, I think it's cleaner as two components but if we decide to keep the primary action consistent in all cases (like you're suggesting) then that may simplify the rendering logic for the buttons enough that the separation isn't as relevant.

assets/js/modules/tagmanager/components/setup/SetupForm.js
aaemnnosttv: Yeah I was wondering about the best way to handle the errors from the other store but that sounds like a good idea.

includes/Modules/Tag_Manager.php
aaemnnosttv: Yeah I wouldn't say it's needed, but avoids the extra request for settings during setup. I haven't measured the difference in perceived speed but it's worth comparing. This is in the IB though.

aaemnnosttv:
@felixarntz one thing I'm noticing is that there is an odd delay after selecting a container when the live container versions are loading where the submit button is disabled but there is no indication of a loading state.

You can see what I mean here:
Kapture 2020-09-04 at 13 39 26

My first thought is that we could add a mini progress bar to the right of the submit button(s) whenever live container versions are loading. This would require a bunch more selections in the `SetupForm` though so maybe it would be best extracted to its own component to encapsulate the selections and logic?

I realize this would also be a visual change, but the above is already a change that results in a poorer user experience IMO.

includes/Modules/Tag_Manager.php
aaemnnosttv: This doesn't appear to be doing anything after taking a closer look as the preloaded data is already expired by the time this is requested, I'll remove it.

assets/js/modules/tagmanager/components/setup/SetupFormSubmitButtons.js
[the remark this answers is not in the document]
aaemnnosttv: @tofumatt I've combined this component back into the `SetupForm` as it became a good deal simpler when refactoring to use the regular form submit handler again 👍

assets/js/googlesitekit/data/utils.js
```diff
@@ -20,6 +20,8 @@
  * External dependencies
  */
 import invariant from 'invariant';
+import mapValues from 'lodash/mapValues';
+import memize from 'memize';
```
aaemnnosttv, line 24: Note that while `memize` was added to `package.json`, it's already a dependency of `@wordpress/data` so this just makes it safe to use without really adding anything to the bundle.

assets/js/modules/tagmanager/components/setup/SetupForm.js
aaemnnosttv: I've refactored the function here to use a try/catch so we only need to set it back to `false` if any of the requests fail 🙂

assets/js/modules/tagmanager/components/setup/SetupForm.js
```diff
@@ -16,64 +16,124 @@
...
-		}
-		if ( ! error ) {
+	const dispatchAnalytics = useDispatch( MODULES_ANALYTICS );
+	const submitForm = useCallback( async ( { submitMode } = {} ) => {
+		const throwOnError = async ( func ) => {
+			const { error } = await func() || {};
+			if ( error ) {
+				throw error;
+			}
+		};
```
aaemnnosttv, line 71: This would obviously be useful elsewhere as well but it's defined here for now as a best practice for hook callbacks as there isn't a clear place for this to go just yet. Alternatively, creating something like `createStrictDispatch` could be useful to create decorated bound actions. The return value is our convention here so we can't magically make it work with everything.

assets/js/googlesitekit/data/utils.js
```diff
@@ -282,3 +283,35 @@ export const commonStore = {
...
+ * dependent selectors that have resolvers as many selectors are composed of
+ * higher-level selectors internally which is where a resolver is normally implemented.
+ *
+ * @since n.e.x.t
+ * @private
+ *
+ * @param {Function} select The registry.select function.
+ * @return {Function} The strict version of registry.select.
+ */
+export const createStrictSelect = ( select ) => ( storeName ) => {
```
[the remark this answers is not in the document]
aaemnnosttv: Tests added!

assets/js/modules/tagmanager/components/setup/SetupErrorNotice.js
```diff
@@ -0,0 +1,48 @@
...
+		useSelect( ( select ) => {
+			const settings = select( MODULES_ANALYTICS ).getSettings();
+			return select( MODULES_ANALYTICS ).getErrorForAction( 'saveSettings', [ settings ] );
+		} ),
+	].filter( Boolean );
+
+	if ( analyticsErrors.length ) {
+		return analyticsErrors.map(
+			( { message, reconnectURL } ) => <ErrorText key={ message } message={ message } reconnectURL={ reconnectURL } />
+		);
```
tofumatt, line 44: Is it okay to return an array from a component like this? I'd think you'd want to wrap this in a `<Fragment>` or the return value of this component will be an `Array`. I thought only components, strings, `false`, or `null` were valid return types.

```suggestion
		return (
			<Fragment>
				{analyticsErrors.map( ( { message, reconnectURL } ) => <ErrorText key={ message } message={ message } reconnectURL={ reconnectURL } />}
			</Fragment>
		);
```

seems like it'd do the trick.

assets/js/modules/tagmanager/components/setup/SetupForm.js
```diff
@@ -84,10 +144,37 @@ export default function SetupForm( { finishSetup } ) {
...
-				<Button disabled={ ! canSubmitChanges }>
-					{ __( 'Confirm & Continue', 'google-site-kit' ) }
-				</Button>
+				{ isSetupWithAnalytics && (
+					<Fragment>
+						<Button disabled={ ! canSubmitChanges }>
+							{ __( 'Continue to Analytics setup', 'google-site-kit' ) }
+						</Button>
+						{ /*
+						This "link" below will be rendered as a <button> but should not
```
tofumatt, line 153: 👍🏻

tofumatt (approved):
Looks great! A lot better with those changes 👍🏻 

I think there's a single comment suggestion I left about expanding `GTM` to `Tag Manager`, but it's not even a biggie. Looks good to me!

assets/js/modules/tagmanager/components/setup/SetupErrorNotice.js
aaemnnosttv: I had my doubts about this too but it [seems to work just fine](https://codesandbox.io/s/condescending-thompson-8yyw1?file=/src/App.js) 😄 

Apparently this was added in React v16:
https://reactjs.org/blog/2017/09/26/react-v16.0.html#new-render-return-types-fragments-and-strings

aaemnnosttv:
I still have cases to add for Storybook here but I think all of the rest of the feedback should be addressed now.

aaemnnosttv:
@felixarntz I've added new stories to storybook for Tag Manager setup cases for all 3 amp modes for the following:

* Container(s) with property ID, Analytics inactive
* Container(s) with property ID, Analytics active, ID match
* Container(s) with property ID, Analytics active, ID mismatch error (containers match, setting does not)
* Multiple property IDs, Analytics inactive (Secondary AMP only)
* Multiple property IDs, Analytics active (Secondary AMP only)

I _think_ that covers everything. Let me know if there's anything else you'd like to see.

assets/js/modules/tagmanager/components/setup/SetupFormInstructions.js
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	// Multiple property IDs implies secondary AMP where selected containers don't reference the same Analytics property ID.
+	if ( hasMultipleAnalyticsPropertyIDs ) {
+		const message = __( 'Looks like you’re already using Google Analytics within your Google Tag Manager configurations. However, the configured Analytics tags reference different property IDs. You need to configure the same Analytics property in both containers.', 'google-site-kit' );
+
+		return <ErrorText message={ message } />;
+	}
+
+	if ( analyticsModuleActive ) {
```
felixarntz, line 52:
```suggestion
	if ( analyticsModuleActive && singleAnalyticsPropertyID ) {
```

assets/js/modules/tagmanager/components/setup/SetupFormInstructions.js
```diff
@@ -0,0 +1,86 @@
...
+		// If the Analytics module is active, and selected containers reference a different property ID
+		// than is currently set in the Analytics module, display an error explaining why the user is blocked.
+		if ( singleAnalyticsPropertyID && singleAnalyticsPropertyID !== analyticsPropertyID ) {
```
felixarntz, line 55:
```suggestion
		if ( singleAnalyticsPropertyID !== analyticsPropertyID ) {
```

assets/js/modules/tagmanager/components/setup/SetupFormInstructions.js
```diff
@@ -0,0 +1,86 @@
...
+		// inform the user that GTM will take over outputting the tag/snippet.
+		if ( analyticsModuleActive && singleAnalyticsPropertyID && singleAnalyticsPropertyID === analyticsPropertyID ) {
+			/* translators: %s: Analytics property ID */
+			const message = __( 'Looks like you’re using Google Analytics. Your Analytics property %s is already set up in your Google Tag Manager configuration, so Site Kit will switch to using Google Tag Manager for Analytics.', 'google-site-kit' );
+			return (
+				<p>
+					{ sprintf( message, singleAnalyticsPropertyID ) }
+				</p>
+			);
+		}
```
felixarntz, line 72: With the above, this can essentially become the `else`, so the wrapping clause here can be removed.

assets/js/modules/tagmanager/components/setup/SetupMain.js
```diff
@@ -72,7 +67,7 @@ export default function SetupMain( { finishSetup } ) {
 	} else if ( isCreateAccount || ! accounts?.length ) {
 		viewComponent = <AccountCreate />;
 	} else {
-		viewComponent = <SetupForm finishSetup={ finishSetupAndNavigate } />;
+		viewComponent = <SetupForm finishSetup={ finishSetup } setIsNavigating={ setIsNavigating } />;
```
felixarntz, line 70: I'm not a fan of splitting this up here, it feels a bit counter-intuitive, especially when compared with the other module setup code, especially since `setIsNavigating` is a quite low-level implementation detail. I think we could do the following instead:
* Keep this component here unchanged. Whenever the setup is finished, `isNavigating` here should also be set, similar to today.
* In `SetupForm`, handle its own `isNavigating` local state and render a `ProgressBar` if applicable. This keeps the logic more separate between the two components.

felixarntz (changes requested):
Awesome work! Pretty much LGTM, just a bit of minor feedback, plus this one will need a QA Brief.

assets/js/modules/tagmanager/components/setup/SetupMain.js
aaemnnosttv: This sounds reasonable but the problem with that is that we currently set `isNavigating` to true at the beginning to prevent flashes between logic that would otherwise show a progress bar (even when successful – something that is visible today) and then only set it back to false if there's a problem. If we were to duplicate the progress bar rendering between components that seems more fragile to visual inconsistencies, when switching from one component to another (this may only happen for a split second or two but will likely happen under normal circumstances). For example, the vertical/horizontal alignment should be exactly the same between the two components (which is an odd coupling to have) but also it would look a bit odd for the position of the gradient/animation to snap to a different position due to rendering a different progress bar in the same place.

For this reason, I think it's preferable to keep the single progress bar in the parent. As far as sharing low-level functions between components, I'm not sure I agree with that entirely. `setIsNavigating` is just setting some state that a navigation is happening. This state happens to cause the progress bar to be rendered. If the shared state was `setProgressBarActive` – that would be much more low-level I think.

As an alternative, I think we can use form state in the datastore to avoid needing to pass this as a prop. That way `finishSetup` can remain the same implementation all the way through as well (decorating this is a bit confusing IMO even though I did that 😄 ). Let me try this and see what you think.

assets/js/modules/tagmanager/components/setup/SetupForm.js
```diff
@@ -16,61 +16,123 @@
...
+		const throwOnError = async ( func ) => {
+			const { error } = await func() || {};
+			if ( error ) {
+				throw error;
+			}
+		};
+		// We'll use form state to persist the chosen submit choice
+		// in order to preserve support for auto-submit.
+		// Set `inProgress` optimistically to avoid flashes of progress bar and content.
+		setValues( FORM_SETUP, { submitMode, inProgress: true } );
```
aaemnnosttv, line 75: @felixarntz I used `inProgress` here rather than `isNavigating` which we had previously as it is more accurate to what it represents. Even if we used something other than `ProgressBar`, this state would still be relevant.

assets/js/modules/tagmanager/components/setup/SetupMain.js
```diff
@@ -50,29 +50,23 @@ export default function SetupMain( { finishSetup } ) {
 	const isDoingGetAccounts = useSelect( ( select ) => select( STORE_NAME ).isDoingGetAccounts() );
 	const isDoingSubmitChanges = useSelect( ( select ) => select( STORE_NAME ).isDoingSubmitChanges() );
 	const hasResolvedAccounts = useSelect( ( select ) => select( STORE_NAME ).hasFinishedResolution( 'getAccounts' ) );
+	const inProgress = useSelect( ( select ) => select( CORE_FORMS ).getValue( FORM_SETUP, 'inProgress' ) );
```
felixarntz, line 53: How about calling this `submissionInProgress` or something slightly more precise? I think that would clarify the intent of this value better.

assets/js/modules/tagmanager/components/setup/SetupMain.js
```diff
@@ -50,29 +50,23 @@ export default function SetupMain( { finishSetup } ) {
...
 
 	// Set the accountID and containerID if there is an existing tag.
 	useExistingTagEffect();
 
-	// When `finishSetup` is called, flag that we are navigating to keep the progress bar going.
-	const [ isNavigating, setIsNavigating ] = useState( false );
-	const finishSetupAndNavigate = ( ...args ) => {
-		finishSetup( ...args );
-		setIsNavigating( true );
-	};
```
felixarntz, line 63: Not strictly necessary the way we have our code behave now, but I think it would be good to keep this functionality as it was before, just using `dispatch` to set `inProgress` in `core/forms` store instead. Then pass the full `finishSetupAndNavigate` to the `SetupForm` component below.

This way the `finishSetup` prop adds the progress bar on its own as expected when called, and `SetupForm` can still do its adjustments as needed. Due to this component and the progress bar now relying on a single shared value, despite potentially setting it in multiple places it wouldn't have any drawbacks regarding appearance.

felixarntz (commented):
LGTM! Two follow-up suggestions, but nothing critical.

assets/js/modules/tagmanager/components/setup/SetupMain.js
aaemnnosttv: Sounds good, I'll clarify it.

assets/js/modules/tagmanager/components/setup/SetupMain.js
aaemnnosttv: While this is more/less a harmless change, I'm having a hard time seeing why we would keep this over the simpler implementation. `finishSetup` in the `SetupForm` is only ever called within its `submitForm` function which handles the now `submitInProgress` form state as well as the form submission itself in all cases. Why would we modify this function in a way that wouldn't make a difference? If this function were called somewhere else, then I think this would make a lot more sense and would even be necessary, but I'm hesitant to make the change as-is. Maybe I'm missing something though – can you elaborate on why this would be good to keep?

aaemnnosttv:
Thanks @felixarntz – QA brief is added now, and feedback has been addressed, with a question about the one suggestion.

assets/js/modules/tagmanager/components/setup/SetupMain.js
felixarntz: I think keeping this in here too means that the `finishSetup` function is more self-contained. The other component that receives it as prop and calls it shouldn't _need_ to deal with `submitInProgress`, it's only a UX optimization that makes sense in the case of Tag Manager.

Yes, for our current implementation it is not important to have but I think it's more bulletproof to ensure that, no matter what the other component does, `finishSetup` would behave like it should.

aaemnnosttv: > I think keeping this in here too means that the `finishSetup` function is more self-contained. The other component that receives it as prop and calls it shouldn't _need_ to deal with `submitInProgress` [...]

In that sense, if we were to bring it back, I think it makes more sense to keep the local `isNavigating` state which is what `finishSetup` would set when called and add that back as an additional condition for showing the progress bar. It doesn't make sense for `finishSetup` to set `submitInProgress` because by the time it's called, the submit is basically done. Maybe we can update the `SetupForm` to always set `submitInProgress` to `false` when it's done, and we would rely on the state set when calling `finishSetup` to keep the progress bar going like it did before. I don't think there should be any flicker in that case as `submitInProgress` would still be `true` when `finishSetup` is called. What do you think?

> it's only a UX optimization that makes sense in the case of Tag Manager

I think the pattern here applies to other modules as well, but we can look at them as-needed.

aaemnnosttv: @felixarntz this has been updated as we talked about now. Just waiting for the build to finish – seeing some failures due to the Composer image failing to pull so it may need a few of them restarted. PHP and JS tests are all passing for me locally though.

Issue #1919, opened 2020-08-15
Use granular error handling in AdSense setup


felixarntz:
After #1814, let's start switching components over to use the more granular error handling. The now legacy `getError()` selector **without parameters** is only used in the `StoreErrorNotice` which is widely used and one other place in the AdSense setup flow. This issue is about resolving only that AdSense setup flow usage, but not any of the `StoreErrorNotice` usages, which we should look at separately.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be no usage of parameter-less `getError` selector call in the AdSense setup flow.

## Implementation Brief

* In the AdSense `SetupMain` component, remove `error` variable.
* Instead, introduce the following variables:
    * `accountsError`, relies on `getError( 'getAccounts' )`
    * `alertsError`, relies on `getError( 'getAlerts', [ accountID ] )`
* Modify signature of `determineAccountStatus` utility function to accepts `accountsError` and `alertsError` instead of `error`.
    * Replace the first `error` usage (if no accounts) with `accountsError`.
    * Replace the second `error` usage (if no alerts) with `alertsError`.
* Modify Jest tests and Storybook accordingly to account for the changes.

## QA Brief

* Running `npm run test:js` does not result in any errors
* All AdSense Setup states (loading, no account, multiple accounts, etc) work as expected within Storybook and in-app

## Changelog entry

* Improve accuracy of AdSense account status detection based on specific errors.

aaemnnosttv:
The IB looks good 👍 

One thought I have looking at this is that maybe it would make more sense to implement `determineAccountStatus` as a React hook (i.e. `useAccountStatus`) instead of a pure utility function, and let it `select` most of what it needs internally rather than requiring the parent component to select everything and provide so many arguments. It could still allow for and may even require parameters if necessary but just curious if you had considered this before.

felixarntz:
@aaemnnosttv I like that suggestion, it would simplify the components using it quite a bit. However, it would also increase the scope and estimate of this issue I'd imagine, so maybe let's consider that separately?

aaemnnosttv:
Yeah, I think it would be a different level of effort. Let's create a new issue for it as a potential enhancement for the future.

IB ✅

cole10up:
**Tested**

Ran npm run test:js successfully
![image](https://user-images.githubusercontent.com/57115907/94111476-fad1ed80-fdf8-11ea-92eb-691976b6f6ce.png)

[...]

Issue #1814, opened 2020-07-20
Implement more appropriate error handling in JS data stores and refactored components


felixarntz:
Currently Site Kit data stores simply store the error in an `error` property per store which has a few drawbacks:
* Only one error per store is saved.
* It is unclear where the error came from.
* It is unclear whether that error should be displayed and overrule regular app flow or whether it can be ignored.

In components, we currently use an `ErrorNotice` component which simply takes that single error and displays it. While this works _fairly_ well for module setup flows and settings being edited, it will show more obvious disadvantages as we're migrating dashboard widgets (e.g. PageSpeed Insights widget right now doesn't display any alternative content on error, it just keeps loading). The goal here should be that every widget knows whether (one of) its API queries resulted in a failure and that it then returns appropriate error content. Eventually such errors should use predefined components so that e.g. the widget area renderer can detect that and aggregate certain types of errors together - this way users would see a single error message instead of e.g. 4 times the same one.

Components in general also need to have awareness of errors. For example, loading indicators shouldn't keep loading forever when there was actually an error that will prevent the loading process from being completed. Related example: https://github.com/google/site-kit-wp/commit/d256c1065d91172bd9615682936804e5ddd61f88

Let's think about how we can make errors available to components in a more granular way.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

### Error Store

* There should be a `createErrorStore()` function which creates a partial store for errors. It should have:
    * action `receiveError( error, baseName, args )` (should be marked as private)
        * Theoretically, all three parameters are required, however due to legacy support for now only `error` should be required.
        * `baseName` should be documented with something like "This should be a selector or action name."
        * with all arguments provided (will be required once legacy part below no longer exists), the reducer should key the error by `baseName` and `args` (only one error per combination, i.e. an error for the exact same request would override the previous one)
        * if only `error` is provided, just store it in `error` state property like today; this part of the reducer should be annotated as legacy behavior with a TODO comment to remove it once all instances of that behavior have been removed
        * no new code should ever pass `error` only
    * selector `getError( baseName, args )` (should be marked as private)
        * Theoretically `baseName` should be required, but due to legacy support for now no parameters should be required.
        * `baseName` should be documented with something like "This should be a selector or action name."
        * if only `baseName` is provided, return the first error for that `baseName` (i.e. for any args)
[...]

PR #1988, opened 2020-09-04
Enhancement/1919 granular AdSense error handling

caseymorrisus:
## Summary
Addresses issue [#1919](https://github.com/google/site-kit-wp/issues/1919)

## Relevant technical choices
Removes use of generic `getError` call (no parameters) with `accountsError` and `alertsError` which are both passed into `determineAccountStatus`. Modifies existing `error` variable to return non-falsy value if either `accountsError` or `alertsError` exist as the underlying `ErrorNotice` component calls `errorToStatus` which parses accounts and alerts errors into strings (fallback to `ErrorText` with static string if not accounts or alerts error).

## Checklist


assets/js/modules/adsense/util/status.js
```diff
@@ -56,12 +57,13 @@ export const determineAccountStatus = ( {
...
 	alerts,
-	error,
+	accountsError,
+	alertsError,
 	previousAccountID,
 	previousClientID,
 } ) => {
 	if ( undefined === accounts || undefined === previousAccountID ) {
-		return accountsErrorToStatus( error );
+		return accountsErrorToStatus( accountsError );
```
aaemnnosttv, line 66: @felixarntz should this conditional be updated to perhaps only consider if there is an `accountsError` rather than checking `accounts` and `previousAccountID`?

assets/js/modules/adsense/util/status.js
```diff
@@ -81,7 +83,7 @@ export const determineAccountStatus = ( {
 	}
 
 	if ( undefined === alerts ) {
-		return alertsErrorToStatus( error );
+		return alertsErrorToStatus( alertsError );
```
aaemnnosttv, line 86: @felixarntz same with this conditional – should it check `alerts` or `alertsError` would be better?

assets/js/modules/adsense/components/setup/SetupMain.js
```diff
@@ -98,14 +98,17 @@ export default function SetupMain( { finishSetup } ) {
 	// Get additional information to determine account and site status.
 	const alerts = useSelect( ( select ) => select( STORE_NAME ).getAlerts( accountID ) );
 	const urlChannels = useSelect( ( select ) => select( STORE_NAME ).getURLChannels( accountID, clientID ) );
-	const error = useSelect( ( select ) => select( STORE_NAME ).getError() );
+	const accountsError = useSelect( ( select ) => select( STORE_NAME ).getError( 'getAccounts', [] ) );
+	const alertsError = useSelect( ( select ) => select( STORE_NAME ).getError( 'getAlerts', [ accountID ] ) );
+	const error = accountsError || alertsError;
```
aaemnnosttv, line 103: The IB says that `error` would be removed in this component, but it looks like perhaps we could still use this for cases where it is still used where it only seems to care "if there is any error". In that case, I think it would be slightly better to name it something like `hasError` and cast it to a boolean, but that's a bit of a nitpick.

I'm okay with keeping it if @felixarntz is.

aaemnnosttv (commented):
Thanks @caseymorrisus – the changes here look good! I have a few questions for @felixarntz here as he's the one who wrote these and is generally the most familiar with the AdSense module code. There's the one detail about the `error` variable which is potentially inconsistent with the IB but might be okay to keep as-is (again, pending Felix's decision).

assets/js/modules/adsense/components/setup/SetupMain.js
felixarntz: In principle, keeping `error` here like this is fine, however it's not fully accurate: With the new logic, we're looking up the two specific "expected errors" that are relevant for the `determineAccountStatus` function. However, right now any other "real error" won't be considered for the actual error rendering component further below.

Instead of this `error` variable, I think it would make sense to rely on the new `hasErrors` selector and write that into a `hasErrors` variable. This covers _any_ possible error in the AdSense store, similar to how it behaved before, but using the new selector from #1814.

assets/js/modules/adsense/util/status.js
felixarntz: I think it's okay as is - if both of the currently checked variables are defined, let's rely on that and ignore any error - although in practice there shouldn't be any `accountsError` in that circumstance anyway.

assets/js/modules/adsense/util/status.js
felixarntz: Same as above, let's keep it as is.

felixarntz (changes requested):
Mostly LGTM, just one thing to fix here so that _any_ error is still considered for displaying errors.

Also, I noticed one unrelated bug here which was not caught during #1814 - there is a `receiveError( undefined )` call which is now no longer valid and raises an invariant error. I've opened #2033 to address that, let's squeeze that into the sprint.

felixarntz (approved):
Great!

Issue #1858, opened 2020-07-30
Fix over-aggressive API requests to check whether Analytics and AdSense are linked


felixarntz:
## Bug Description

At the moment Site Kit determines the connection between Analytics and AdSense by making Analytics API requests with AdSense-specific metrics which are restricted and will result in an error, if AdSense is not connected.

These requests naturally happen for:
* The top earning pages widget on the Site Kit dashboard (`AdSenseDashboardWidgetTopPagesTableSmall`)
* The top earning pages widget on the AdSense page (`AnalyticsAdSenseDashboardWidgetTopPagesTable`)

However, a similar request is also manually triggered, separately from regular batch requests, **only** to check whether it returns an error (via `isAdsenseConnectedAnalytics()` utility function) for:
* The Analytics page overall (`AnalyticsDashboardWidget`)
* The AdSense page overall (`AdSenseDashboardWidget`)

The latter request is completely pointless: While it's supposed to prompt the user to connect AdSense and Analytics, this already happens in the two aforementioned widgets anyway if an error is returned, so we don't need to have an additional prompt (currently in the form of the `AdSenseDashboardOutro` component). Furthermore, this doesn't seem to work properly anyway (I see the prompts in the widgets, but I don't see the `AdSenseDashboardOutro` component being rendered).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

### Legacy fixes

* The `isAdsenseConnectedAnalytics()` utility function and all of its usages should be removed.
* The `AdSenseDashboardOutro` component and all of its usages should be removed.

### Future fixes

* The `modules/analytics` store should receive a selector `getAdsenseLinked()` and an action `setAdsenseLinked( linked )`. These should **not** be handled via the regular settings store (although this technically is a setting), because that setting is automatically handled on the server, and a client-side change should not cause the settings to be flagged as outdated. This selector and action should only affect the client.
    * The initial value for the state should come from the server though (i.e. via `getSettings().adsenseLinked`, where this is present), but it should then be stored (and possibly modified) separately.
* The resolver for `getReport` in the 'modules/analytics' store should be updated:
    * If `options.metrics` includes any `ga:adsense*` metrics, the resolver should inspect the results of the `fetchGetReport` call:
        * If there is a 400 error with the message starting in "Restricted metric(s):", it should `dispatch( 'modules/analytics' ).setAdsenseLinked( false )`.
[...]

Issue #1765, opened 2020-07-13
Refactor waitFor actions with asyncSelect


aaemnnosttv:
## Feature Description

Currently we have a number of common workarounds in place that would be better solved by existing (but not-so documented) solutions.

**Waiting for a selector to be resolved**

There are some cases where we always want the resolved value of a selector. In such cases we have added dedicated "wait" actions to be able to synchronously select the resolved value after.

This is what `registry.__experimentalResolveSelect` does. If the selector has a resolver, it returns a promise that returns the resolved value. This has the benefit of both waiting and returning the value we're waiting for.
Since it returns a promise, it is only usable in a control or in an async function within a component.

> A note on `__experimental` - This method was added back in 2019 and is since used in a handful of places in Gutenberg. There is also an open issue to audit experimental APIs in which this is mentioned as something which will very likely become stable. 
> Worst-case would be that we reimplement it from userland which is possible. Another reason this should be okay to use is that it is used by `@wordpress/data-controls` which is not marked as experimental.

To make this usable within an action or a resolver, we need to make this an action itself. This is one of the actions included in the [`@wordpress/data-controls`](https://github.com/WordPress/gutenberg/tree/master/packages/data-controls) package, which is a small package of helper actions+controls.

In `@wordpress/data-controls`, it includes a `select( storeName, selector, args[] )` action which uses `__experimentalResolveSelect` if the selector has a resolver and the "regular" synchronous select otherwise.
E.g.
```js
const profiles = yield select( STORE_NAME, 'getProfiles', [ propertyID ] ); 
```

The downsides of this in my opinion are:
- it confuses `select` with being asynchronous
- it has a less natural syntax (similar to `hasFinishedResolution` which is not great)

I would suggest a use like this:
```js
actions.asyncSelect = ( selectFn ) => {
	invariant( 'function' === typeof selectFn, 'selectFn must be a function.' );
	return {
		type: ASYNC_SELECT,
		payload: { selectFn },
	};
};
controls[ ASYNC_SELECT ] = createRegistryControl( ( registry ) => ( { payload: { selectFn } ) => {
	return selectFn( registry.__experimentalResolveSelect );
} );
```

> Similarly, a `syncSelect` action could be added to avoid the need for using `getRegistry` in most cases. The `data-controls` package has an action for this too, but it uses the same awkward syntax, but also has an odd name for the action itself: `__unstableSyncSelect`.

[...]

PR #1991, opened 2020-09-07
Fix aggressive Analytics-Adsense linked checks

aaemnnosttv:
## Summary

Addresses issue #1858

## Relevant technical choices

* The resolver for `getAdsenseLinked` uses a "private" action for getting Analytics settings asynchronously when resolving the value
* Adds a new utility for normalizing options passed to `getReport` – currently only handles `metrics` but can be extended in the future to include things like `dimensions` and `orderby`.
* Adds `memize` as a dependency – this was also added in the PR for #1984 and is a dependency of `@wordpress/data` this just makes it safe to use. It's also extremely small and has no dependencies of its own.

## Checklist


assets/js/modules/analytics/datastore/adsense.js
```diff
@@ -0,0 +1,103 @@
...
+			type: SET_ADSENSE_LINKED,
+		};
+	},
+};
+
+export const controls = {
+	// TODO: Refactor this once we have dedicated actions for this.
+	[ __ADSENSE_LINKED_GET_RESOLVED_SETTINGS ]: createRegistryControl( ( registry ) => () => {
+		return registry.__experimentalResolveSelect( STORE_NAME ).getSettings();
+	} ),
```
aaemnnosttv, line 51: Regarding the TODO here – see #1765

assets/js/modules/analytics/datastore/report.js
```diff
@@ -32,11 +33,13 @@ import { stringifyObject } from '../../../util';
...
 import { isValidDateRange, isValidOrders } from '../../../util/report-validation';
 import { isValidDimensions, isValidMetrics } from '../util/report-validation';
+import { actions as adsenseActions } from './adsense';
+import { normalizeReportOptions } from '../util/report-normalization';
 
 const fetchGetReportStore = createFetchStore( {
 	baseName: 'getReport',
 	controlCallback: ( { options } ) => {
-		return API.get( 'modules', 'analytics', 'report', options );
+		return API.get( 'modules', 'analytics', 'report', normalizeReportOptions( options ) );
```
aaemnnosttv, line 42: Normalization is applied late here rather than in `argsToParams` (as I initially thought) to not interfere with our calls to `receiveGetReport` which takes `params`.

assets/js/modules/analytics/datastore/report.js
```diff
@@ -56,6 +58,7 @@ const fetchGetReportStore = createFetchStore( {
 
 		const { metrics, dimensions, orderby } = options;
 
+		invariant( !! metrics, 'Requests must specify at least one metric for an Analytics report.' );
```
aaemnnosttv, line 61: The API requires at least one metric to be requested so we should enforce that here, rather than in `isValidMetrics` which more about the shape than request specific validation IMO.

assets/js/modules/analytics/datastore/report.js
```diff
@@ -92,7 +95,15 @@ const baseResolvers = {
 			return;
 		}
 
-		yield fetchGetReportStore.actions.fetchGetReport( options );
+		const { error } = yield fetchGetReportStore.actions.fetchGetReport( options );
+
+		if (
+			error?.code === 400 &&
+			error?.message.startsWith( 'Restricted metric' ) &&
+			normalizeReportOptions( options ).metrics.some( ( { expression } ) => /^ga:adsense/.test( expression ) )
```
felixarntz, line 103: There is something missing here:
> If `options.metrics` includes any `ga:adsense*` metrics, the resolver should inspect the results of the `fetchGetReport` call

The reason for inspecting `options.metrics` is that such requests should "either way" result in a call to `setAdsenseLinked`. The way you currently have it here, we could as well look at `error.message` to see if it includes any `ga:adsense*`, which would be simpler. However, that's not enough. The clause should be:
* If `options.metrics` includes any `ga:adsense*`
    * If there is a 400 Restricted metric error, `setAdsenseLinked( false )`.
    * Otherwise, `setAdsenseLinked( true )`.

felixarntz (changes requested):
Mostly LGTM, just two things. Also, can you add a QA Brief?

assets/js/modules/analytics/datastore/report.js
aaemnnosttv: `isValidMetrics` mostly checks the types/shape, it doesn't ensure that a non-empty value isn't provided. It is also a more useful error I think if no metrics were provided at all. Without this it could be called with `metrics: ''` and pass the validation.

Taking a closer look at `isValidStringsOrObjects`, it accepts any string (including empty), as well as an empty array, but as a generic utility, it isn't clear if these should be considered invalid or not. Also the `verifyFunction` is only called for objects. It might make more sense to split this into `areValidStrings` and `areValidObjects` functions. Then these aren't even really separate functions anymore, and you can have a simple `isValidCollection` function that looks something like this: `castArray( collection ).every( validateFn )`, but you still end up needing to validate the list length (whether or not an empty list should be valid). For that reason, maybe we should just normalize first here, ensure the list isn't empty and then validate each item in the list?

assets/js/modules/analytics/datastore/report.js
aaemnnosttv: Oops, yeah I missed that part. No problem 👍

assets/js/modules/analytics/datastore/report.js
```diff
@@ -54,8 +56,9 @@ const fetchGetReportStore = createFetchStore( {
 		invariant( isPlainObject( options ), 'Options for Analytics report must be an object.' );
 		invariant( isValidDateRange( options ), 'Either date range or start/end dates must be provided for Analytics report.' );
 
-		const { metrics, dimensions, orderby } = options;
+		const { metrics, dimensions, orderby } = normalizeReportOptions( options );
 
+		invariant( metrics.length, 'Requests must specify at least one metric for an Analytics report.' );
```
aaemnnosttv, line 61: @felixarntz this is normalized first now so we can rely on `metrics` being an array of strings or objects for this length check.

aaemnnosttv:
@felixarntz feedback should be addressed now, and QA brief added (let me know what you think of it since there may not be any user-facing changes here yet).

assets/js/modules/analytics/datastore/report.js
```diff
@@ -92,7 +95,12 @@ const baseResolvers = {
 			return;
 		}
 
-		yield fetchGetReportStore.actions.fetchGetReport( options );
+		const { error } = yield fetchGetReportStore.actions.fetchGetReport( options );
+
+		if ( normalizeReportOptions( options ).metrics.some( ( { expression } ) => /^ga:adsense/.test( expression ) ) ) {
+			const isRestrictedMetricError = error?.code === 400 && error?.message.startsWith( 'Restricted metric' );
```
felixarntz, line 101: This wasn't specified in the ACs, but seeing this I think it would be more accurate to check for any `ga:adsense*` in the error message as well. Otherwise, theoretically it could be that a request which includes an AdSense metric fails due to _another_ restricted metric, but would still (incorrectly) raise this flag of AdSense not being linked.

felixarntz (commented):
Just one additional thought here, I can quickly push that change. Regarding the QA Brief, that looks good to me, I just added one bullet point about checking that the old report request that was happening on every pageload is no longer run.

felixarntz:
@aaemnnosttv Can you look over the two commits I just added? Should be good afterwards from my end.

* There was a problem with metric validation, which is a broader issue, but not highly critical currently, so it's okay including that in the next release with this.
* I added another check to cover the case where theoretically a "Restricted metric" error could be returned due to a non-AdSense metric while requesting an AdSense metric. There was also one general problem with the new tests because the assertion for fetch calls was run too early.

felixarntz (approved):

assets/js/modules/analytics/util/report-validation.js
```diff
@@ -34,7 +34,9 @@ import { isValidStringsOrObjects } from '../../../util/report-validation';
 export function isValidMetrics( metrics ) {
 	return isValidStringsOrObjects( metrics, ( metric ) => {
 		const validExpression = metric.hasOwnProperty( 'expression' ) && typeof metric.expression === 'string';
-		const validAlias = metric.hasOwnProperty( 'alias' ) && typeof metric.alias === 'string';
+
+		// 'alias' is optional; if provided, it must be a string.
+		const validAlias = ! metric.hasOwnProperty( 'alias' ) || typeof metric.alias === 'string';
```
aaemnnosttv, line 39: 👍  although the doc block should be updated to reflect this as it currently contradicts this:
> Objects must have "expression" and "alias" properties in order to be considered as valid.

assets/js/modules/analytics/datastore/report.js
```diff
@@ -98,7 +98,7 @@ const baseResolvers = {
 		const { error } = yield fetchGetReportStore.actions.fetchGetReport( options );
 
 		if ( normalizeReportOptions( options ).metrics.some( ( { expression } ) => /^ga:adsense/.test( expression ) ) ) {
-			const isRestrictedMetricError = error?.code === 400 && error?.message.startsWith( 'Restricted metric' );
+			const isRestrictedMetricError = error?.code === 400 && error.message && error.message.startsWith( 'Restricted metric' ) && /ga:adsense/.test( error.message );
```
aaemnnosttv, line 101: This line is getting a bit long, why un-chain the message?
```suggestion
			const isRestrictedMetricError = error?.code === 400 && error?.message?.startsWith( 'Restricted metric' ) && /ga:adsense/.test( error.message );
```

assets/js/modules/analytics/datastore/report.test.js
```diff
@@ -135,33 +157,67 @@ describe( 'modules/analytics report', () => {
...
 				muteConsole( 'error' ); // fetch will trigger 400 error.
 				registry.select( STORE_NAME ).getReport( nonAdsenseOptions );
+				await untilResolved( registry, STORE_NAME ).getReport( nonAdsenseOptions );
+
 				expect( fetchMock ).toHaveFetchedTimes( 1 );
 
+				expect( registry.select( STORE_NAME ).getAdsenseLinked() ).toBe( true );
+			} );
+
+			it( 'sets adsenseLinked to true if a success response is returned for AdSense metrics', async () => {
```
aaemnnosttv, line 167:
```suggestion
			it( 'sets adsenseLinked to true if a successful response is returned for AdSense metrics', async () => {
```

assets/js/modules/analytics/datastore/report.test.js
```diff
@@ -135,33 +157,67 @@ describe( 'modules/analytics report', () => {
...
+				const nonAdsenseOptions = {
+					dateRange: 'last-28-days',
+					metrics: 'ga:adsenseRevenue',
+				};
```
aaemnnosttv, line 171: Probably just an oversight in the variable name here but these _are_ `AdsenseOptions`, no?

felixarntz (approved):

aaemnnosttv (commented):
Thanks for these additions @felixarntz – this looks good for the most part, there are just a few small details that could be corrected around documentation and naming. Otherwise just a suggestion or two.

assets/js/modules/analytics/datastore/report.js
felixarntz: I changed that because of the following clause, which shouldn't run the regular expression test against `undefined`.

felixarntz (approved):

Issue #87, opened 2019-06-30
Refresh button for PageSpeed card


amirsadeghian:
I believe the plugin is caching the page speed results. It would be nice to add a button to refresh the cache and re-run the page speed audit.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* In the Page Speed and Experience module, for the Lab data tab, add a "Run test again" link in the bottom left corner
* Clicking "Run test again" clears the currently cached results and re-runs the test
* "View details at PageSpeed Insights" is moved to the bottom right corner
* Since Field data accumulates over time as users visit the page, and there won't be significant differences, so this functionality is only necessary for the Lab data tab. 

## Implementation Brief

* Add a button in the PageSpeed Insights widget that invalidates the selector's resolution and the cached data for `API.get`.
* See #2000 for the basis of the approach recommended.

## QA Brief

* Navigate to a single page's Search Console page or the site's Site Kit Dashboard.
* Scroll to PageSpeed Insights
* Wait until data is loaded
* Refresh the page
* Scroll to PageSpeed Insights and notice data is cached. Click the "Refresh Data" button and watch it update.

## Changelog entry

* Show a button to refresh PageSpeed Insights report data in the widget.

ivankristianto:
@amirsadeghian 
Thanks for submitting this request. PSI data is cached in browser sessionStorage, so if you open in new tab you will get the new data.
Although I agree with you, having a button to re-fetch would be better.

amirsadeghian:
@ivankristianto , I am interested to contribute. Can I fork the project and add this enhancement and submit a pull request?
Thank you

ivankristianto:
@amirsadeghian yes feel free to forked and open a PR.
Also please read this contributing doc priorly: https://github.com/google/site-kit-wp/blob/develop/CONTRIBUTING.md

Note: Instead branch of and submit PR to `master`, branch of and submit the PR to `develop`. 
We will update that contributing doc to match later.

felixarntz:
@tofumatt While I generally agree with your point on not "surfacing" cache to the user, this is more specific to PSI here I'd argue: Since that API (to be precise, only the "Lab Data" part) essentially provides realtime data, it's important to allow users to get _super_ up-to-date data, which is not as relevant for other modules. For example, if a user makes a massive technical change to try to improve the site's performance, they should be able to see how that surfaces in the PSI widget right away. That doesn't apply to the other modules though which take time to gather metrics.

[...]

PR #2000, opened 2020-09-09
Add a "Force update" button to PageSpeed Insights.

tofumatt:
## Summary

Addresses issue #87.

## Relevant technical choices

This changes the `getReport` resolver to always run and the checks to be based on selector/resolver status rather than if there's any data in the reducer, so that invalidated resolvers and an API cache clear like in the PR will cause PageSpeed Inisghts to refresh.

## Checklist


assets/js/modules/pagespeed-insights/components/common/LabReportMetrics.js
```diff
@@ -25,18 +25,44 @@ import classnames from 'classnames';
...
 
 export default function LabReportMetrics( { data, error } ) {
+	const referenceURL = useSelect( ( select ) => select( CORE_SITE ).getCurrentReferenceURL() );
+
+	const { invalidateResolution } = useDispatch( STORE_NAME );
+	const updateReport = useCallback( async ( event ) => {
+		event.preventDefault();
+
+		// Invalidate the PageSpeed API request caches.
+		await API.invalidateCache( 'modules', 'pagespeed' );
```
aaemnnosttv, line 59: The second parameter here should be the module slug. `pagespeed` is actually the datapoint name for the report selector, which we want to keep. See the control for `getReport` in `assets/js/modules/pagespeed-insights/datastore/report.js`
```suggestion
		await API.invalidateCache( 'modules', 'pagespeed-insights', 'pagespeed' );
```

assets/js/modules/pagespeed-insights/components/common/LabReportMetrics.js
```diff
@@ -103,6 +129,12 @@ export default function LabReportMetrics( { data, error } ) {
 			</table>
 			<div className="googlesitekit-pagespeed-report__row googlesitekit-pagespeed-report__row--last">
 				<ReportDetailsLink />
+				<Button
+					href="#update"
```
aaemnnosttv, line 133: Why the `href` here? If we just want a normal `<button>` that won't submit, we can use `type="button"`.

aaemnnosttv (changes requested):
Looks good, just a small fix and one question. I also noticed that the storybook stories for the component need an update or two as well.

Kapture 2020-09-23 at 11 44 50

Compare with the [storybook for the current release](https://google.github.io/site-kit-wp/storybook/master/?path=/story/pagespeed-insights-module-components--dashboard-widget-loading) and you'll see what I mean. Note the loading state is not rendering properly in both, so bonus points if you can fix that easily 😄

assets/js/modules/pagespeed-insights/components/common/LabReportMetrics.js
tofumatt: Argh, that was a mistake, thanks! Didn't mean to commit it; non-`href` should be fine 👍🏻

tofumatt:
Storybook is all fixed, including the loading screen, so this should be set now. Thanks!

aaemnnosttv (commented):
Thanks @tofumatt – this looks good although there are a few details here which are not consistent with the AC. Particularly these:

- add a "Run test again" link in the bottom left corner
- "View details at PageSpeed Insights" is moved to the bottom right corner

The component currently looks like this on the lab tab:
image

The most obvious thing there is that the position is swapped; the button is said to be on the left but it is on the right.

Using some devtools magic, we can see what this would look like below

image

I think this looks better and is more consistent with other components by having the main action left-aligned. This may require other changes though for the details link to stay right aligned when there is no button rendered on the left (maybe an empty div is needed here to preserve the flex?).

Also "Run test again" is implemented as a button but the AC describes it as a link. I think this is more consistent with our other components as well as most actions are tied to buttons, although perhaps you could make the case that this is a "secondary" action so it shouldn't be a button.

Let's update the button positioning/alignment of the button and details link as well as double check with @felixarntz about the element type we should use for the action.

felixarntz:
@tofumatt @aaemnnosttv 
> Also "Run test again" is implemented as a button but the AC describes it as a link. I think this is more consistent with our other components as well as most actions are tied to buttons, although perhaps you could make the case that this is a "secondary" action so it shouldn't be a button.

Yes let's make this visually appear like a link. It shouldn't look like a button since it's not a primary action on the screen where it appears. Our buttons that actually look like buttons are intended for primary actions.

tofumatt:
Sounds good; updated and ready for another look! 👍🏻

assets/js/modules/pagespeed-insights/components/common/LabReportMetrics.js
```diff
@@ -128,10 +128,10 @@ export default function LabReportMetrics( { data, error } ) {
 				</tbody>
 			</table>
 			<div className="googlesitekit-pagespeed-report__row googlesitekit-pagespeed-report__row--last">
-				<ReportDetailsLink />
-				<Button onClick={ updateReport }>
+				<Link href="#update-pagespeed" onClick={ updateReport }>
```
aaemnnosttv, line 131: The `href` shouldn't be needed here. It should render as a `button` that looks like a link without it. Can we avoid adding this?

aaemnnosttv (commented):
Looks great @tofumatt – just the one question for you 🙂

assets/js/modules/pagespeed-insights/components/common/LabReportMetrics.js
tofumatt: Oh, sweet, done!

aaemnnosttv (approved):
Super 👍

PR #2006, opened 2020-09-10
Fix broken links issue on the Analytics page.

eugene-manuilov:
## Summary

Addresses issue #1922

## Checklist

### DIFF
--- a/assets/js/components/data-table.js
+++ b/assets/js/components/data-table.js
@@ -51,9 +51,6 @@ export const getDataTableFromData = ( data, headers, options ) => {
 	each( data, ( row, j ) => {
 		const cells = [];
 		const link = links && links[ j ];
-		const permaLink = link
-			? global._googlesitekitLegacyData.admin.siteURL + link
-			: false;
 
 		each( row, ( cell, i ) => {
 			// Replace (none) by direct.
@@ -75,7 +72,11 @@ export const getDataTableFromData = ( data, headers, options ) => {
 						? <div className="googlesitekit-table__body-item-content">
 							<Link
 								className="googlesitekit-table__body-item-link"
-								href={ useAdminURLs ? getSiteKitAdminURL( 'googlesitekit-dashboard', { permaLink } ) : permaLink }
+								href={
+									useAdminURLs
+										? getSiteKitAdminURL( 'googlesitekit-dashboard', { permaLink: global._googlesitekitLegacyData.admin.siteURL + link } )
+										: link
+								}
 								external={ ! useAdminURLs }
 								inherit
 							>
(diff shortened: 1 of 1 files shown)


assets/js/components/data-table.js
```diff
@@ -85,7 +86,7 @@ export const getDataTableFromData = ( data, headers, options ) => {
 							{ showURLs &&
 								<Link
 									className="googlesitekit-table__body-item-url"
-									href={ permaLink }
+									href={ link }
```
aaemnnosttv, line 89: This should maybe use a full URL here as `link` is a relative path is it not?

aaemnnosttv (commented):
Looks good to me, just the one question. Would be good to have @felixarntz take a look as well.

assets/js/components/data-table.js
felixarntz: Let's keep this as `link` for now which is how it originally behaved. We cannot assume anything about `link`, it could be an absolute or relative URL.

assets/js/components/data-table.js
```diff
@@ -75,7 +72,11 @@ export const getDataTableFromData = ( data, headers, options ) => {
 						? <div className="googlesitekit-table__body-item-content">
 							<Link
 								className="googlesitekit-table__body-item-link"
-								href={ useAdminURLs ? getSiteKitAdminURL( 'googlesitekit-dashboard', { permaLink } ) : permaLink }
+								href={
+									useAdminURLs
+										? getSiteKitAdminURL( 'googlesitekit-dashboard', { permaLink: global._googlesitekitLegacyData.admin.siteURL + link } )
```
felixarntz, line 77: This works in the way we currently use this function, but it could still cause issues. `link` is not necessarily a relative URL - for example on the Analytics module page, it is a URL to the Analytics FE.

To be safe here, we should **only** prefix `link` with the site URL if it is a relative URL (starting in `/`).

felixarntz (changes requested):
This technically works for our current codebase, but there's still a caveat here that I think we should address.

assets/js/components/data-table.js
eugene-manuilov: Ok, updated.

felixarntz (approved):

assets/js/components/data-table.js
```diff
@@ -62,6 +59,7 @@ export const getDataTableFromData = ( data, headers, options ) => {
 			}
 
 			const hiddenOnMobile = options && options.hideColumns && options.hideColumns.mobile.includes( i );
+			const permaLink = link[ 0 ] === '/' ? global._googlesitekitLegacyData.admin.siteURL + link : link;
```
felixarntz, line 62: `link` may be `undefined` here, which currently causes e2e tests to fail.
```suggestion
			const permaLink = link?.[ 0 ] === '/' ? global._googlesitekitLegacyData.admin.siteURL + link : link;
```

felixarntz (changes requested):

assets/js/components/data-table.js
eugene-manuilov: Haven't seen this and updated it slightly differently.

felixarntz (approved):

Issue #1761, opened 2020-07-10
Refactor Autocomplete (Get rid of componentWillMount and componentWillReceiveProps)


felixarntz:
## Bug Description

On the Site Kit dashboard, a console warning is currently triggered due to usage of legacy component lifecycle methods: https://fb.me/react-unsafe-component-lifecycles

<img width="1664" alt="Screenshot 2020-07-10 at 12 49 16" src="https://user-images.githubusercontent.com/3531426/87192863-36eebb00-c2ac-11ea-9caa-bb6358b039ae.png">

The warning is only emitted on the Site Kit dashboard (not on other Site Kit screens) so it has to be caused by something only in use there. I wasn't able to find any usage of these methods in our codebase, so I suspect it is something 3P-related. Let's find out the cause and see if we can upgrade the respective dependency so that this warning goes away.

Potentially related: #1356 

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The Site Kit dashboard should not emit a console warning about usage of legacy React component lifecycle methods.

## Implementation Brief

* Remove the `accessible-autocomplete`  package from package.json
* Install the `@reach/combobox` package
* Create a new functional component called `PostSearcherAutoSuggest`  in `js/components/PostSearcherAutoSuggest.js`
	* Build out the new component to provide the same functionality as the existing `Autocomplete` component - https://github.com/google/site-kit-wp/blob/develop/assets/js/components/post-searcher.js#L172
	* Follow the docs available here  https://reach.tech/combobox#server-side-search
	* Update as required to match existing styles and confirm that there are no unexpected visual changes via VRT.
* Refactor `PostSearcher`  in `assets/js/components/post-searcher.js`  to be a functional component and rename the file  to `PostSearcher.js` to match updated naming conventions
	* Replace the existing `Autocomplete` component with the new `PostSearcherAutoSuggest`
* Update e2e tests in `tests/e2e/specs/dashboard/search.test.js` to match any changes to selectors as needed.

## QA Brief

* The post search component on the Site Kit dashboard should continue to work as it does in the latest release
* Minor differences may be present in terms of functionality particularly regarding use of the keyboard as the replacement is designed for accessibility first, so while there may be differences here it should actually be an improvement

## Changelog entry

* Update post search input to use a better maintained and more accessible autocomplete library.

aaemnnosttv:
[...]

PR #2009, opened 2020-09-10
Fix missing spaces in autocomplete titles

felixarntz:
## Summary

Addresses issue #1761 (follow-up to #1890)

## Relevant technical choices

* Using `display: flex` on the autocomplete options causes [this problem](https://github.com/google/site-kit-wp/issues/1761#issuecomment-690778589) because the result values may be split into multiple `span` elements. If their content has trailing spaces, the flexbox styling will make it look like these spaces are not there.
* The same visual result can be achieved by specifying corresponding top and bottom padding for each autocomplete option instead of a fixed height - this way the text will still appear vertically centered.

## Checklist

### DIFF
--- a/assets/sass/components/global/_googlesitekit-autocomplete.scss
+++ b/assets/sass/components/global/_googlesitekit-autocomplete.scss
@@ -92,15 +92,11 @@
 }
 
 .autocomplete__option {
-	align-items: center;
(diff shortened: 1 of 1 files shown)


aaemnnosttv (approved):
Great!

Issue #2010, opened 2020-09-10
Analytics: gtag misconfigured for AMP causing _gl query parameter injection while navigating on origin


westonruter:
## Bug Description

I originally opened this as an AMP issue: https://github.com/ampproject/amphtml/issues/30176. However, it appears the issue is with how Site Kit is configuring `gtag`.

On my WordPress site which is AMP-first, I have the Google Site Kit plugin running with Google Analytics configured like this (as provided by Site Kit):

```html
<amp-analytics type="gtag" data-credentials="include">
    <script type="application/json">
        {"vars":{"gtag_id":"UA-123456-1","config":{"UA-123456-1":{"groups":"default","linker":{"domains":["weston.ruter.net"]}}}},"optoutElementId":"__gaOptOutExtension"}
    </script>
</amp-analytics>
```

From the homepage at https://weston.ruter.net/ if I then click on post for “Integrating with AMP Dev Mode in WordPress” which has the URL of https://weston.ruter.net/2019/09/24/integrating-with-amp-dev-mode-in-wordpress/ I am actually taken to a URL like:

```
https://weston.ruter.net/2019/09/24/integrating-with-amp-dev-mode-in-wordpress/?_gl=1~abcde5~
```

This happens for each internal link I click on my site. When I navigate around my site, each request includes a `_gl` query parameter, and then it gets stripped out with `history.replaceState()` when the page initializes (as far as I can tell). 

As I understand, this is in order to [measure customer journeys across domains](https://developers.google.com/gtagjs/devguide/linker). Nevertheless, I am not navigating across domains. I am navigating on my own domain. 

This `_gl` query parameter only be added links for pages that are served on the AMP Cache over to the origin domain. The injection of this query parameter when navigating around on the origin server is problematic for a few reasons:

1. It is distracting for users who care about what is in the Location bar. I see I am navigated to some URL with a long random string, and then it disappears. That's somewhat disconcerting.
2. It can interfere with full-page caching since no two page requests will have the same URL.
3. It also breaks service worker caching of navigation requests, unless you figure out logic to explicitly strip out the `_gl` parameter when caching a response or looking up a previously-cached response.

According to @zhouyx in https://github.com/ampproject/amphtml/issues/30176#issuecomment-690775882:

> I can see two issues here.
> 
[...]

PR #1203, opened 2020-03-03
Update AMP client ID usage

aaemnnosttv:
## Summary

Addresses issue #1160

## Relevant technical choices

* Updated `ampExperimentJSON` input in Optimize setup to be shown if `ampEnabled` instead of if `ampClientIDOptIn`

## Checklist

- [x] My code is tested and passes existing unit tests.
- [x] My code has an appropriate set of unit tests which all pass.
- [x] My code is backward-compatible with WordPress 4.7 and PHP 5.6.
- [x] My code follows the [WordPress](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) coding standards.
- [x] My code has proper inline documentation.
- [x] I have signed the Contributor License Agreement (see <https://cla.developers.google.com/>).

PR #2012, opened 2020-09-11
Remove analytics linker domain since we’re already on the same domain

kasparsd:
## Summary

Addresses issue #2010.

## Relevant technical choices

The Google Analytics gtag linker is necessary only to [track user visits between different top level domains](https://developers.google.com/gtagjs/devguide/linker). Since the plugin is aware only of the site it is installed on, we should default to _not_ including the domain name of the current site.

> The domain linker functionality enables two or more related sites on separate domains to be measured as one.

On requests served from the AMP cache the linker is enabled by default and associated with the canonical URL of the response [per docs](https://developers.google.com/gtagjs/devguide/amp):

> The capability to link to your canonical domain from the AMP cache is enabled by default.

This changeset reverts some of the code introduced in #1203.

## Checklist

### DIFF
--- a/includes/Modules/Analytics.php
+++ b/includes/Modules/Analytics.php
@@ -356,12 +356,6 @@ final class Analytics extends Module
 
 		$gtag_opt = array();
 
-		if ( $this->context->get_amp_mode() ) {
-			$gtag_opt['linker'] = array(
-				'domains' => array( $this->get_home_domain() ),
-			);
-		}
-
 		$anonymize_ip = $this->get_data( 'anonymize-ip' );
 		if ( ! is_wp_error( $anonymize_ip ) && $anonymize_ip ) {
 			// See https://developers.google.com/analytics/devguides/collection/gtagjs/ip-anonymization.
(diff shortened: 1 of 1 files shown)


adamsilverstein:
Hi @kasparsd! Thank you for your contribution here. I just spotted this issue.

This issue you linked to was fixed upstream in https://github.com/ampproject/amphtml/pull/32100.

Adding the local domain `origin.com` is desired if users link to a remote flow (for example a newsletter signup or cart checkout) and then return back to the `origin.com` site for confirmation. Including `origin.com` in the domains array lets analytics track the user session across both the origin and remote domains. 

The bug fix in AMP ensures that the linker attribute is never added when the user's window location matches the target domain, ie. a link to the site they are on. So, even though we add the local domain here in Site Kit, it will only be added to decorated links if the user is viewing the page from another domain (eg. embedded content), another example of where you might want the attribute added.

Finally, the AMP linker behavior now matches the behavior of the `gtag.js` non-AMP linker.

> On requests served from the AMP cache the linker is enabled by default and associated with the canonical URL of the response per docs:

Right, however this is added for different use cases: external linking and embedded content links. Given that, would it make sense to you if I close this issue as "won't fix"?

kasparsd:
> Given that, would it make sense to you if I close this issue as "won't fix"?

Thanks for the elaborate explanation @adamsilverstein! I'm glad this has been fixed upstream. Closing this as suggested.

Issue #1770, opened 2020-07-13
Update @wordpress/scripts to its latest version


felixarntz:
## Feature Description

The `@wordpress/scripts` dependency (currently used on 3.4.0) should be updated to its latest version. The newer versions include / depend on a new `@wordpress/jest-console` package though, which requires us to change many of our Jest tests to account for this change. See https://github.com/google/site-kit-wp/pull/1763#issuecomment-657267120 for more information.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* `@wordpress/scripts` should be updated to its latest version (newer than 3.4.0).
* Fix all unexpected console logging in Jest tests. Console logging in any test which is not expected should cause the test to fail.
* There should be no errors on Site Kit screens, and tests should pass (obviously).

## Implementation Brief

* Upgrade `@wordpress/scripts` to the latest version.
* Update javascript tests to explicitly call `expect( console ).toHaveErrored()`, `expect( console ).toHaveWarned()` or `expect( console ).toHaveLogged()` when a console output is expected.
* Remove `muteConsole` calls from all javascript tests.
* Remove `muteConsole` function from `tests/js/utils.js` file.

## QA Brief
* Running `npm run test:js` does not result in any errors or display any unexpected console logging (all console logs should fail tests)
* App functions normally without breakage

## Changelog entry

* N/A

felixarntz:
IB ✅ , however we should only add the console expect statements if they are actually expected. For cases where an error is logged because e.g. we forgot to intercept a request that we should have intercepted, that should be fixed by intercepting it.

aaemnnosttv:
@caseymorrisus would you please add a QA Brief here as well?

tofumatt:
QA  ✅

PR #2014, opened 2020-09-11
Enhancement/1770 update @wordpress/scripts to latest

caseymorrisus:
## Summary
Addresses issue [#1770](https://github.com/google/site-kit-wp/issues/1770)

## Relevant technical choices
- Replaces `muteConsole` in favor of explicitly calling `expect( console ).toHaveErrored()`, `expect( console ).toHaveWarned()`, or `expect( console ).toHaveLogged()` when a console output is expected. 
- Removes `muteConsole` function entirely (definition, use, and in comments).
- Fixes `act` errors thrown due to components updating internal state by wrapping with `waitFor`.
- Fill store state with minimum data to prevent unexpected fetches by resolvers.

## Checklist


assets/js/modules/analytics/components/settings/SettingsEdit.test.js
```diff
@@ -61,8 +61,11 @@ describe( 'SettingsEdit', () => {
 		};
 		const { registry } = render( <SettingsEdit />, { setupRegistry } );
 
-		expect( registry.select( STORE_NAME ).getAccountID() ).toBe( existingTag.accountID );
-		expect( registry.select( STORE_NAME ).getPropertyID() ).toBe( existingTag.propertyID );
-		expect( registry.select( STORE_NAME ).hasErrors() ).toBeFalsy();
+		waitFor( () => {
```
aaemnnosttv, line 64: Missing `await` – also, let's not put all the assertions into `waitFor`. We should only wait until the datastore state has stopped changing. It should be sufficient to have everything received ahead of time like it is now, so if that's still causing problems, we'll have to see what is changing in the store for it to complain about.

assets/js/modules/analytics/components/settings/SettingsMain.test.js
```diff
@@ -19,13 +19,31 @@
...
+import {
+	render,
+	fireEvent,
+	waitFor,
+	createTestRegistry,
+	unsubscribeFromAll,
+} from '../../../../../../tests/js/test-utils';
 import { STORE_NAME } from '../../datastore/constants';
 import { STORE_NAME as CORE_SITE } from '../../../../googlesitekit/datastore/site/constants';
+import { STORE_NAME as CORE_MODULE } from '../../../../googlesitekit/modules/datastore/constants';
```
aaemnnosttv, line 31:
```suggestion
import { STORE_NAME as CORE_MODULES } from '../../../../googlesitekit/modules/datastore/constants';
```

assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -96,8 +95,11 @@ describe( 'modules/analytics properties', () => {
 					{ body: response, status: 500 }
 				);
 
-				muteConsole( 'error' );
-				await registry.dispatch( STORE_NAME ).createProperty( accountID );
+				registry.dispatch( STORE_NAME ).createProperty( accountID );
+
+				await subscribeUntil( registry,
+					() => registry.select( STORE_NAME ).isDoingCreateProperty( accountID ) === false
+				);
```
aaemnnosttv, line 102: The previous version was correct here – async actions should simply be `await`-ed rather than using `subscribeUntil` as they return a Promise which resolves when the action is finished. Generally speaking, `subscribeUntil` is a low-level utility that shouldn't be needed in most tests. It's mostly used now in older tests where we didn't have an alternative yet.

assets/js/modules/optimize/components/common/UseSnippetInstructions.test.js
```diff
@@ -20,64 +20,71 @@
...
  */
 import UseSnippetInstructions from './UseSnippetInstructions';
-import { render } from '../../../../../../tests/js/test-utils';
+import {
+	createTestRegistry,
+	render,
+	unsubscribeFromAll,
+} from '../../../../../../tests/js/test-utils';
 import { STORE_NAME } from '../../datastore/constants';
 import { STORE_NAME as CORE_MODULE } from '../../../../googlesitekit/modules/datastore/constants';
```
aaemnnosttv, line 29:
```suggestion
import { STORE_NAME as CORE_MODULES } from '../../../../googlesitekit/modules/datastore/constants';
```

assets/js/modules/pagespeed-insights/components/dashboard/DashboardPageSpeed.test.js
```diff
@@ -59,12 +59,17 @@ const setupRegistryNoFieldDataDesktop = ( { dispatch } ) => {
...
 
 		const { queryByRole } = render( <DashboardPageSpeed />, { setupRegistry: setupRegistryNoReports } );
 
-		expect( queryByRole( 'progressbar' ) ).toBeInTheDocument();
+		await waitFor( () => {
+			expect( queryByRole( 'progressbar' ) ).toBeInTheDocument();
+		} );
+
+		expect( console ).toHaveErrored();
+		expect( console ).toHaveWarned();
```
aaemnnosttv, line 72: `waitFor` looks great here! Can you elaborate on why we would expect both an errors and warnings here? It isn't immediately obvious as to why we'd expect either since the previous test didn't.

assets/js/modules/tagmanager/components/common/AccountCreate.test.js
```diff
@@ -22,6 +22,7 @@
 import AccountCreate from './AccountCreate';
 import { fireEvent, render, waitFor, createTestRegistry, freezeFetch, muteFetch } from '../../../../../../tests/js/test-utils';
 import { STORE_NAME } from '../../datastore/constants';
+import { STORE_NAME as CORE_MODULE } from '../../../../googlesitekit/modules/datastore/constants';
```
aaemnnosttv, line 25:
```suggestion
import { STORE_NAME as CORE_MODULES } from '../../../../googlesitekit/modules/datastore/constants';
```

assets/js/modules/tagmanager/hooks/useExistingTagEffect.test.js
```diff
@@ -70,5 +68,7 @@ describe( 'useExistingTagEffect', () => {
 
 		expect( registry.select( STORE_NAME ).getContainerID() ).toBe( existingContainer.publicId );
 		expect( registry.select( STORE_NAME ).getInternalContainerID() ).toBe( existingContainer.containerId );
+		// Can't seem to prevent an error from updates happening outside of `act()` just for this part.
+		expect( console ).toHaveErrored();
```
aaemnnosttv, line 72: I think this was from me previously but it would be good to see if this is indeed still needed and see if we can remove the need for it (not worth spending a ton of time on though).

tests/js/setup-globals.js
```diff
@@ -43,6 +43,14 @@ global._googlesitekitLegacyData = {
 	},
 };
 
+// eslint-disable-next-line no-undef
```
aaemnnosttv, line 46: This exclusion doesn't seem to be necessary (maybe from during dev?).
```suggestion
```

tests/js/utils.js
```diff
@@ -74,25 +74,6 @@ export function WithTestRegistry( { children, callback, registry = createTestReg
...
- * API fetch errors or other things you expect to log to console but don't
- * want appearing in the jest output.
- *
- * @since 1.5.0
- * @private
- *
- * @param {string} type  Type of console to mute (one of: `'error'`, `'warn'`, `'log'`, `'info'`, or `'debug'`)
- * @param {number} times Number of times to mute console output perform resuming.
- */
-export const muteConsole = ( type = 'error', times = 1 ) => {
```
aaemnnosttv, line 90: I always thought `muteConsole` should cause a test failure if it was called and the expected kind of console method wasn't triggered so this is great 👍

aaemnnosttv (changes requested):
Thanks @caseymorrisus – this is looking really good for the most part, just a few things to clean up. A few of the comments apply to multiple places so try to address them holistically as not every occurrence has a comment. Let me know if you have any questions!

assets/js/modules/pagespeed-insights/components/dashboard/DashboardPageSpeed.test.js
caseymorrisus: After working through this with Evan, this was fixed by adding another `freezeFetch` call as the endpoint is called twice (once for desktop, the other for mobile).

assets/js/modules/tagmanager/hooks/useExistingTagEffect.test.js
caseymorrisus: This was fixed by importing `actHook` rather than `act` as `renderHook` is being used rather than `render`.

assets/js/modules/optimize/components/common/UseSnippetInstructions.test.js
felixarntz: This is still missing.

assets/js/modules/tagmanager/components/common/AccountCreate.test.js
felixarntz: This is still missing.

tests/js/setup-globals.js
felixarntz: This still needs to be removed.

felixarntz (changes requested):
Awesome work on cleaning these up! Just three things that were probably overlooked by accident from @aaemnnosttv's review.

felixarntz (approved):
Quickly committed the last few minor changes so that we can merge this. Great stuff!

Issue #2024, opened 2020-09-14
Simplify client side module registration


aaemnnosttv:
## Feature Description

While working on #1623, some improvements were discovered as to ways that we could simplify the way modules are registered on the client side.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* `actions.registerModule` should not use `getRegistryKey` or a separate object for `ModuleComponents` and can pass `settingsComponent` through to the reducer.
* `actions.registerModule` should no longer be a generator function and all asynchronous use should be converted to synchronous
* `actions.registerModule` should only pass the following keys through to its `payload.settings`:
	- `name`
	- `description`
	- `icon`
	- `order`
	- `homepage`
	- `settingsComponent` – only this key should have a default set in the action (`DefaultModuleSettings`), the rest should be `undefined` if not passed
* The modules datastore partial should have two new top-level keys in its state:
	- `clientDefinitions` – initial value of `{}`
	- `serverDefinitions` – initial value of `undefined`
* `fetchGetModulesStore.reducerCallback` should be updated to receive the modules into the new `serverDefinitions` instead of `modules`
* `selectors.getModules` should be updated to recursively/deep-merge its `clientDefinitions` over the `serverDefinitions` before returning a normalized, ~~sorted~~, object, similar to how it does now
	- if `serverDefinitions` are `undefined`, the selector should return `undefined`
	- `clientDefinitions` should take precedence over `serverDefinitions` but only for keys which are passed when registered (no defaults other than client-only settings)
* `resolvers.getModules` should not require changes
* Remove `waitForModules` and its control as no longer used
* Remove `modules` state as no longer used

#### Tests

* Update Jest tests for module registration – particularly `modules.test.js` to cover these changes

## Implementation Brief

* Merge #2026 

## QA Brief

* Review implementation changes and added tests to verify the AC is met

## Changelog entry

* Simplify module registration in JavaScript and only allow one registration call per module.

felixarntz:
IB ✅

tofumatt:
QA ✅

PR #2026, opened 2020-09-15
Simplify module registration

aaemnnosttv:
## Summary

Addresses issue #2024

## Relevant technical choices

* Removed sorting from return value of `getModules` – [see comment below](https://github.com/google/site-kit-wp/pull/2026#discussion_r488457023)
* Cleans up the tests a bit to be more consistent with others using `untilResolved` and awaiting async actions rather than using a selector

## Checklist

### DIFF
--- a/assets/js/googlesitekit/modules/datastore/modules.js
+++ b/assets/js/googlesitekit/modules/datastore/modules.js
@@ -19,6 +19,7 @@
 /**
  * External dependencies
  */
+import memize from 'memize';
 import merge from 'lodash/merge';
 import invariant from 'invariant';
 
@@ -53,6 +54,22 @@ const moduleDefaults = {
 	connected: false,
 	settingsComponent: DefaultModuleSettings,
 };
+
+const normalizeModules = memize(
+	( modules ) => Object.keys( modules )
+		.map( ( slug ) => {
+			return {
+				...moduleDefaults,
+				name: slug, // Ensure `name` is not empty.
+				...modules[ slug ],
+				slug,
+			};
+		} )
+		.sort( ( a, b ) => a.order - b.order )
+		.reduce( ( acc, module ) => {
+			return { ...acc, [ module.slug ]: module };
+		}, {} )
+);
 
 const fetchGetModulesStore = createFetchStore( {
 	baseName: 'getModules',
@@ -301,20 +318,7 @@ const baseSelectors = {
 		// but only for keys whose values are not `undefined`.
 		const modules = merge( {}, serverDefinitions, clientDefinitions );
 
-		return Object.keys( modules )
-			.map( ( slug ) => {
-				return {
-					...moduleDefaults,
-					name: slug, // Ensure `name` is not empty.
-					...modules[ slug ],
-					slug,
-				};
-			} )
-			.sort( ( a, b ) => a.order - b.order )
-			.reduce( ( acc, module ) => {
-				return { ...acc, [ module.slug ]: module };
-			}, {} )
-		;
+		return normalizeModules( modules );
 	},
 
 	/**
--- a/assets/js/googlesitekit/modules/datastore/modules.js
+++ b/assets/js/googlesitekit/modules/datastore/modules.js
@@ -257,10 +257,7 @@ const baseReducer = ( state, { type, payload } ) => {
 				...state,
 				clientDefinitions: {
 					...state.clientDefinitions,
-					[ slug ]: {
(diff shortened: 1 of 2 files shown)


assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -333,35 +290,27 @@ const baseSelectors = {
...
+		if ( serverDefinitions === undefined ) {
 			return undefined;
 		}
 
-		const registryKey = select( CORE_SITE ).getRegistryKey();
-		if ( registryKey === undefined ) {
-			return undefined;
-		}
-		// Sorting the modules object by order property.
-		const sortedModules = sortByProperty( Object.values( modules ), 'order' );
```
aaemnnosttv, line 349: I removed the sorting here intentionally as it doesn't make much sense to do when the return value is an object where order shouldn't really be significant. If we want to expose modules in a way where the order is significant, we should add a selector for getting modules as a list where that value would be sorted.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -236,47 +221,19 @@ export const baseControls = {
...
 			return {
 				...state,
-				modules: {
-					...existingModules,
+				clientDefinitions: {
+					...state.clientDefinitions,
 					[ slug ]: {
-						...defaults,
-						...( existingModules?.[ slug ] || {} ),
+						...( state.clientDefinitions[ slug ] || {} ),
```
felixarntz, line 236: Instead of this, now that we have this cleaner approach (where client and server are separated), let's also restrict module registration to be unique per slug. Similar to `registerWidget`, the reducer should trigger a console warning and not actually register the module if it is already registered on the client.

assets/js/googlesitekit/modules/datastore/modules.test.js
```diff
@@ -93,13 +90,7 @@ describe( 'core/modules modules', () => {
...
 				);
 
-				registry.dispatch( STORE_NAME ).activateModule( slug );
-
-				// Wait until this activation action has completed.
-				await subscribeUntil( registry, () => registry
-					.select( STORE_NAME )
-					.isDoingSetModuleActivation( slug ) === false
-				);
+				await registry.dispatch( STORE_NAME ).activateModule( slug );
```
felixarntz, line 93: Shouldn't we still wait here until module activation has been completed? There's the API call to refetch authentication data which is async.

assets/js/googlesitekit/modules/datastore/modules.test.js
```diff
@@ -143,13 +134,7 @@ describe( 'core/modules modules', () => {
...
 
 				muteConsole( 'error' );
-				registry.dispatch( STORE_NAME ).activateModule( slug );
-
-				// Wait until this activation action has completed.
-				await subscribeUntil( registry, () => registry
-					.select( STORE_NAME )
-					.isDoingSetModuleActivation( slug ) === false
-				);
+				await registry.dispatch( STORE_NAME ).activateModule( slug );
```
felixarntz, line 137: See above.

felixarntz (changes requested):
Mostly LGTM, just one thing I think we should change.

assets/js/googlesitekit/modules/datastore/modules.test.js
aaemnnosttv: We do wait – the action itself is async by association with those internal calls and will return a promise here that resolves when the action including its internal async calls are completed, similar to a resolver.

assets/js/googlesitekit/modules/datastore/modules.test.js
aaemnnosttv: Yep. Even though this action doesn't have a control of its own, we can still await it to wait for all of its internally `yield`ed actions (async or not) to be done.

assets/js/googlesitekit/modules/datastore/modules.test.js
felixarntz: Oops - I had totally missed the `await` 🤦‍♂️

assets/js/googlesitekit/modules/datastore/modules.js
felixarntz: @aaemnnosttv This sorting should be maintained in here. Sorting for objects _is_ relevant as we talked about a couple weeks back. Modules should be returned in their correct order by default so that when they're iterated over the order is correct automatically.

felixarntz (approved):
Great!

Issue #1923, opened 2020-08-17
Update source links to point to Google service FE appropriate locations


felixarntz:
Many of Site Kit's widgets include a source link at the bottom. However, most of them right now don't link to specific places relevant for the widget, so they aren't very helpful. Furthermore, sometimes these source links point to the Google service frontend, while other times (e.g. on the dashboard) they point to the respective Site Kit module page.

All source links should be updated to point to the respective Google service frontend, to the respective property/view/container/client/site data and, where possible, to a view that shows the same information that the Site Kit widget shows.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All links to change (see below) should use the respective `getServiceURL` datastore selector, to e.g. inherit the correct Google account indicator.
* The source links on the "Clicks", "Impressions", and "Top search queries for your site" Site Kit dashboard widgets (including single URL details), as well as on the "Top search queries over the last x days" Search Console module page widget, should point to:
    * `https://search.google.com/search-console/performance/search-analytics?resource_id={propertyID}&num_of_days={currentDateRangeDays}`
    * On the single URL details view, the URL should include another query parameter `&page=!{searchConsoleURL}` (see below)
* The source links on the "Unique Visitors from Search" Site Kit dashboard widget and the "Bounce Rate" Site Kit single URL details widget should point to:
    * `https://analytics.google.com/analytics/web/#/report/visitors-overview/a{accountID}w{internalWebPropertyID}p{profileID}/`
    * On the single URL details view, the URL should include a suffix `_r.drilldown=analytics.pagePath:{analyticsPagePath}/` (see below)
* The source links on the "All Traffic" widget area widget on Site Kit dashboard (including single URL details), as well as on Analytics module page, should point to:
    * `https://analytics.google.com/analytics/web/#/report/trafficsources-overview/a{accountID}w{internalWebPropertyID}p{profileID}/`
    * On the single URL details view, the URL should include a suffix `_r.drilldown=analytics.pagePath:{analyticsPagePath}/` (see below)
* The source links on the "Most popular content" Site Kit dashboard widget and the "Top content over the last 28 days" Analytics module page widget should point to:
    * `https://analytics.google.com/analytics/web/#/report/content-pages/a{accountID}w{internalWebPropertyID}p{profileID}/`

`{searchConsoleURL}` is the current URL, escaped for query parameters (e.g. slashes become "%2F").
[...]

PR #2030, opened 2020-09-15
Enhancement/1923 update links to point to google locations

ryanwelcher:
## Summary

Addresses issue #1923 

## Checklist


assets/js/modules/search-console/components/dashboard/LegacySearchConsoleDashboardWidgetTopLevel.js
```diff
@@ -37,123 +38,141 @@ import PreviewBlock from '../../../../components/preview-block';
...
+					}
+				/>
+			</div>
+			<div className="
 					mdc-layout-grid__cell
 					mdc-layout-grid__cell--align-bottom
 					mdc-layout-grid__cell--span-2-phone
 					mdc-layout-grid__cell--span-2-tablet
 					mdc-layout-grid__cell--span-3-desktop
 				">
```
felixarntz, line 154: This is now incorrectly indented.

assets/js/util/date-range.js
```diff
@@ -30,13 +30,17 @@ import { STORE_NAME as CORE_USER } from '../googlesitekit/datastore/user/constan
 /**
  * Gets the current dateRange string.
  *
- * @param {string} [dateRange] Optional. The date range slug.
+ * @param {string} [dateRange]      Optional. The date range slug.
+ * @param {boolean} [returnNumber]  Optional. If true, returns the number only.
  * @return {string} the date range string.
  */
-export function getCurrentDateRange( dateRange = getCurrentDateRangeSlug() ) {
+export function getCurrentDateRange( dateRange = getCurrentDateRangeSlug(), returnNumber = false ) {
```
felixarntz, line 37: Having boolean parameters to change function behavior is not a good practice because it violates the principle of functions being responsible for a single task, so the need for this indicates we need to split something out of here instead. We can introduce a function like `getCurrentDateRangeDayCount` or similar, which does this.

assets/js/modules/analytics/components/common/AcquisitionPieChart.js
```diff
@@ -87,7 +103,7 @@ function AcquisitionPieChart( { data, args, source } ) {
 						{
 							a: <Link
 								key="link"
-								href={ getSiteKitAdminURL( 'googlesitekit-module-analytics' ) }
+								href={ sourceURI }
```
felixarntz, line 106: This still needs to be marked as an external link.

assets/js/modules/analytics/components/dashboard/LegacyAnalyticsDashboardWidgetPopularPagesTable.js
```diff
@@ -20,90 +20,90 @@
...
+				{ children }
+			</Layout>
+		</div>
+	);
+};
 
-		if ( ! data || ! data.length ) {
-			return null;
-		}
+function LegacyAnalyticsDashboardWidgetPopularPagesTable( { data } ) {
```
felixarntz, line 60: Looks like you refactored this component, but didn't actually change the URL as required per the ACs.

assets/js/modules/analytics/components/dashboard/LegacyAnalyticsDashboardWidgetPopularPagesTable.js
```diff
@@ -20,90 +20,90 @@
...
+				footer
+				footerCtaLabel={ _x( 'Analytics', 'Service name', 'google-site-kit' ) }
+				footerCtaLink={ serviceURL }
+				fill
+			>
+				{ children }
+			</Layout>
+		</div>
+	);
+};
```
felixarntz, line 58: Is this used in multiple places? If so, we may need to add a prop or something for the service URL, now that it depends on the widget context itself.

felixarntz (changes requested):
A few things are missing from this PR:
* The same URL changes added to the Search Console Clicks and Impressions widgets are not present in the Popular Keywords widget (both new and legacy version, plus on Search Console module page).
* The URL changes for the Analytics All Traffic widget are not present in the legacy version on the single URL details view (see `AnalyticsDashboardDetailsWidgetTopAcquisitionSources`), as well as on the Analytics module page (see `AnalyticsDashboardWidget`).
* The URL changes required for the Analytics Popular Pages widget are missing entirely (both new and legacy version, plus Analytics module page).

Another thing that needs to be updated in general here is Storybook: Because all the new widgets now require additional data (to compose these URLs), that data needs to be provided in Storybook (right now these URLs contain lots of `undefined` strings in Storybook). I suggest the following:
* Update `generateReportBasedWidgetStories` to support an optional `registry` argument being passed. If provided, that registry should be passed to `WithTestRegistry` so that all modifications are applied on _that_ one instead of a new one.
* In the respective Storybook files where these widgets are set up, use `addDecorator` to create the basic version of that registry, which is then used for all the widgets. Should be used for general things like site information (via `provideSiteInfo`), user information (via `provideUserInfo`) and module-specific data (e.g. account, property, profile etc. for Analytics).

assets/js/modules/analytics/components/dashboard/LegacyAnalyticsDashboardWidgetPopularPagesTable.js
ryanwelcher: @felixarntz this is only used in this file, it is part of the `LegacyAnalyticsDashboardWidgetPopularPagesTable` but is also wrapping the `PreviewTable` component in the `withData` HOC.

ryanwelcher:
> * Update `generateReportBasedWidgetStories` to support an optional `registry` argument being passed. If provided, that registry should be passed to `WithTestRegistry` so that all modifications are applied on _that_ one instead of a new one.
> * In the respective Storybook files where these widgets are set up, use `addDecorator` to create the basic version of that registry, which is then used for all the widgets. Should be used for general things like site information (via `provideSiteInfo`), user information (via `provideUserInfo`) and module-specific data (e.g. account, property, profile etc. for Analytics).

@felixarntz as per the above, @aaemnnosttv and I discussed this approach and have concerns about using the `registry` parameter as it would used for all of the stories and we'll run into issues with the state being updated for all stories.
@aaemnnosttv is also concerned about this generator function being too large and complicated ( especially if we need to add more params ) and has asked me write stores for one of the widgets without the generator to get a sense of the effort.

assets/js/modules/analytics/components/dashboard-details/AnalyticsDashboardDetailsWidgetTopAcquisitionSources.js
```diff
@@ -19,71 +19,79 @@
...
-				<div className="
+const { useSelect } = Data;
+
+export default function AnalyticsDashboardDetailsWidgetTopAcquisitionSources() {
+	const serviceURL = useSelect( ( select ) => {
+		const accountID = select( STORE_NAME ).getAccountID();
+		const profileID = select( STORE_NAME ).getProfileID();
+		const internalWebPropertyID = select( STORE_NAME ).getInternalWebPropertyID();
+		return select( STORE_NAME ).getServiceURL(
+			{ path: `/report/trafficsources-overview/a${ accountID }w${ internalWebPropertyID }p${ profileID }/` }
```
felixarntz, line 43: This is used in the single URL details view as the name states (`dashboard-details`) and is missing the current entity URL path awareness, per the ACs.

assets/js/modules/analytics/components/dashboard/AnalyticsDashboardWidget.js
```diff
@@ -123,168 +127,148 @@ class AnalyticsDashboardWidget extends Component {
...
+							'mdc-layout-grid__cell',
+							'mdc-layout-grid__cell--span-12',
+							wrapperClass
+						) }>
+							<Layout
+								header
+								footer
+								/* translators: %s: date range */
+								title={ sprintf( __( 'Top content over the last %s', 'google-site-kit' ), currentDateRange ) }
+								headerCtaLink="https://analytics.google.com"
```
felixarntz, line 209: This also needs to use `topContentServiceURL`.

assets/js/modules/analytics/components/dashboard/AnalyticsDashboardWidget.js
```diff
@@ -123,168 +127,148 @@ class AnalyticsDashboardWidget extends Component {
...
+							'mdc-layout-grid__cell',
+							'mdc-layout-grid__cell--span-12',
+							wrapperClass
+						) }>
+							<Layout
+								header
+								footer
+								/* translators: %s: date range */
+								title={ sprintf( __( 'Top acquisition channels over the last %s', 'google-site-kit' ), currentDateRange ) }
+								headerCtaLink="https://analytics.google.com"
```
felixarntz, line 231: This also needs to use `topAcquisitionServiceURL`.

assets/js/modules/analytics/components/dashboard/AnalyticsDashboardWidget.js
```diff
@@ -123,168 +127,148 @@ class AnalyticsDashboardWidget extends Component {
...
+							<Layout
+								header
+								/* translators: %s: date range */
+								title={ sprintf( __( 'Audience overview for the last %s', 'google-site-kit' ), currentDateRange ) }
+								headerCtaLabel={ sprintf(
+									/* translators: %s: module name. */
+									__( 'See full stats in %s', 'google-site-kit' ),
+									_x( 'Analytics', 'Service name', 'google-site-kit' )
+								) }
+								headerCtaLink="http://analytics.google.com"
```
felixarntz, line 183: This piece is something that was missing from the ACs I just noticed. This should also link to `https://analytics.google.com/analytics/web/#/report/visitors-overview/a{accountID}w{internalWebPropertyID}p{profileID}/`, like the similar items on the main dashboard do.

assets/js/modules/analytics/components/dashboard/LegacyAnalyticsDashboardWidgetTopLevel.js
```diff
@@ -47,174 +46,199 @@ import {
...
+	const pathIDSegment = `a${ accountID }w${ internalWebPropertyID }p${ profileID }/`;
+	let pathPageSegment = '';
 
-		let goalCompletions = '',
-			goalCompletionsChange = '',
-			averageBounceRate = '',
-			averageBounceRateChange = '';
+	if ( url ) {
+		const parsedURL = new URL( url );
+		pathPageSegment += `_r.drilldown=analytics.pagePath:${ parsedURL.pathname.replace( /\//g, '~2F' ) }`;
```
felixarntz, line 112: Would be great to introduce a utility function to handle the URL path segment like that.

assets/js/modules/analytics/components/dashboard/LegacyDashboardAcquisitionPieChart.js
```diff
@@ -24,89 +24,102 @@ import PropTypes from 'prop-types';
...
-				4: { color: '#34CBE3' },
-				5: { color: '#82E88E' },
-			},
-			title: null,
-			width: '100%',
-		};
+	let path = `/report/trafficsources-overview/a${ accountID }w${ internalWebPropertyID }p${ profileID }/`;
+	if ( url ) {
+		const parsedURL = new URL( url );
+		path += `&_r.drilldown=analytics.pagePath:${ parsedURL.pathname.replace( /\//g, '~2F' ) }`;
```
felixarntz, line 55: See above, this should then use the new utility function.

felixarntz (changes requested):
@ryanwelcher This looks much better, a few things are still missing though:
* The footer link in the "Top Queries" widget in the single URL details view is not correct (see `DashboardDetailsWidgetKeywordsTable`).
* The individual links in the "Top Queries" widget in the single URL details view don't include the `page` parameter (see comment further below).
* The Storybook changes only include the Analytics Goals widget, but this should be added for all the widgets changed. Maybe you only did it for this one first to review the approach, which makes sense, but eventually this should be done for all the widgets that include these kinds of links.

Regarding the Storybook comment specifically (cc @aaemnnosttv): Let's go with the `registry` approach that I previously suggested, or something close to it. Passing a registry to this function doesn't make it much more complicated IMO, it's just exchanging the registry created there with a passed one. While this is a complex function, the reduction of code to write is pretty great, and furthermore. Regardless of that, it's out of scope of this issue to rewrite all these widgets using that function to use manual stories (which I'm not sure that's a good idea anyways), so let's keep using the utility function and add the optional `registry` argument to it.

aaemnnosttv:
@felixarntz 
> Let's go with the `registry` approach that I previously suggested, or something close to it. Passing a registry to this function doesn't make it much more complicated IMO, it's just exchanging the registry created there with a passed one.

The problem with this is that we don't want to share the same `registry` instance across multiple stories as each story should have an isolated context. If we share it, then there will be side-effects from switching between stories and you may end up with different results based on the order of stories viewed. The only places where we pass the `registry` instance to the `WithTestRegistry` component are where that instance is created in the decorator where the rest of the common setup is applied.

One of the problems that I saw when I reviewed this with @ryanwelcher recently was how the current function was quite complex with multiple setup functions being strung together and applied in different places. This is where it seemed like the function is doing too much and perhaps a less abstract approach would be easier to understand.

In short, I think we should avoid passing a `registry` to the story generator function for the reasons above and instead pass a setup function to apply the common setup that should be shared by all stories rather than providing a pre-setup registry. The only other setup that should happen is what's needed per-story; we shouldn't have additional callbacks which potentially call others and so on. That doesn't mean we need to use the `callback` prop on `WithTestRegistry` but if we don't use that then the registry should be created in the decorator so that each story gets a fresh instance.

felixarntz:
@aaemnnosttv 
> In short, I think we should avoid passing a `registry` to the story generator function for the reasons above and instead pass a setup function to apply the common setup that should be shared by all stories rather than providing a pre-setup registry.

Thanks for clarifying, that makes sense to me. @ryanwelcher Can you adjust to use the above approach supporting an optional setup function in the utility?

assets/js/modules/analytics/components/dashboard/DashboardBounceRateWidget.js
```diff
@@ -62,12 +69,15 @@ function DashboardBounceRateWidget() {
 		const url = select( CORE_SITE ).getCurrentEntityURL();
 		if ( url ) {
 			args.url = url;
+			const parsedURL = new URL( url );
+			path += `_r.drilldown=analytics.pagePath:${ parsedURL.pathname.replace( /\//g, '~2F' ) }`;
```
felixarntz, line 73: See above, this should use the new utility function.

stories/dashboard.stories.js
```diff
@@ -35,6 +36,14 @@ storiesOf( 'Dashboard', module )
 	.add( 'All Traffic', () => {
 		global._googlesitekitLegacyData = analyticsDashboardData;
 
+		const setupRegistry = ( { dispatch } ) => {
+			dispatch( ANALYTICS_STORE ).receiveGetSettings( {} );
+			dispatch( ANALYTICS_STORE ).setAccountID( '123456789' );
+			dispatch( ANALYTICS_STORE ).setPropertyID( '123456789' );
+			dispatch( ANALYTICS_STORE ).setInternalWebPropertyID( '123456789' );
+			dispatch( ANALYTICS_STORE ).setProfileID( '123456789' );
```
felixarntz, line 44: Nit-picking, but it would be slightly more appropriate to pass all of this using `receiveGetSettings`. The `set*` actions are public actions to change the respective setting value on the client, while `receiveGetSettings` is an internal action to populate what normally would be populated via a REST API response. Also, this way you can combine all of these into one action call. You can see the available Analytics settings for these other actions in the `Analytics\Settings` class in PHP, e.g. you need to pass `accountID` etc within the settings object.

The same applies to all similar code below.

stories/utils/generate-widget-stories.js
```diff
@@ -26,59 +26,22 @@ import { Component } from 'react';
...
+const { components: { Widget } } = Widgets;
 
 /**
  * Generates stories for a report based widget using provided data.
  *
  * @since 1.16.0
  *
  * @param {Object}    args                              Widget arguments.
- * @param {string}    args.moduleSlug                   Module slug.
+ * @param {Array}     args.moduleSlug                   List of modules to activate.
```
felixarntz, line 44: I'm okay making this require an array, but in that case it should also be renamed accordingly, to `moduleSlugs`. `moduleSlug` for an array variable doesn't really make sense.

stories/dashboard.stories.js
```diff
@@ -88,6 +99,14 @@ storiesOf( 'Dashboard', module )
 	.add( 'Search Funnel Analytics Inactive', () => {
 		global._googlesitekitLegacyData = analyticsDashboardData;
 
+		const setupRegistry = ( { dispatch } ) => {
+			dispatch( ANALYTICS_STORE ).receiveGetSettings( {} );
+			dispatch( ANALYTICS_STORE ).setAccountID( '123456789' );
+			dispatch( ANALYTICS_STORE ).setPropertyID( '123456789' );
+			dispatch( ANALYTICS_STORE ).setInternalWebPropertyID( '123456789' );
+			dispatch( ANALYTICS_STORE ).setProfileID( '123456789' );
```
felixarntz, line 107: Something missing here and below is populating the Search Console `propertyID` setting as well, otherwise these URLs are not correct. You can use `dispatch( SEARCH_CONSOLE_STORE ).receiveGetSettings( { propertyID: 'https://example.com/' } )` for that.

stories/dashboard.stories.js
```diff
@@ -160,15 +181,23 @@ storiesOf( 'Dashboard', module )
 				'Dashboard'
 			);
 		}, 250 );
-
+		const setupRegistry = ( { dispatch } ) => {
+			dispatch( ANALYTICS_STORE ).receiveGetSettings( {} );
+			dispatch( ANALYTICS_STORE ).setAccountID( '123456789' );
+			dispatch( ANALYTICS_STORE ).setPropertyID( '123456789' );
+			dispatch( ANALYTICS_STORE ).setInternalWebPropertyID( '123456789' );
+			dispatch( ANALYTICS_STORE ).setProfileID( '123456789' );
```
felixarntz, line 189: See above.

stories/module-analytics-components.stories.js
```diff
@@ -50,46 +51,67 @@ import {
...
+ * @param {wp.data.registry} registry Registry with all available stores registered.
+ */
+const setup = ( registry ) => {
+	const [ property ] = accountsPropertiesProfiles.properties;
+	// eslint-disable-next-line sitekit/camelcase-acronyms
+	registry.dispatch( STORE_NAME ).setAccountID( property.accountId );
+	// eslint-disable-next-line sitekit/camelcase-acronyms
+	registry.dispatch( STORE_NAME ).setInternalWebPropertyID( property.internalWebPropertyId );
+	// eslint-disable-next-line sitekit/camelcase-acronyms
+	registry.dispatch( STORE_NAME ).setProfileID( property.defaultProfileId );
```
felixarntz, line 68: See above, this could be combined to using the `receiveGetSettings` action.

stories/module-analytics-components.stories.js
```diff
@@ -50,46 +51,67 @@ import {
...
 } from '../assets/js/modules/analytics/datastore/__fixtures__';
 
+/**
+ * Defines some additional setup for all tests.
+ *
+ * @since n.e.x.t
+ *
+ * @param {wp.data.registry} registry Registry with all available stores registered.
+ */
+const setup = ( registry ) => {
```
felixarntz, line 61: A similar `setup` function should be added for all the Search Console components stories, which sets the Search Console `propertyID` setting (see comment further above), otherwise the URLs there are incorrect (omitting `property_id`).

felixarntz (changes requested):
This is almost good to go, a few last things are missing though.

stories/dashboard.stories.js
```diff
@@ -24,6 +24,7 @@ import PostSearcher from '../assets/js/components/PostSearcher';
 import { googlesitekit as analyticsDashboardData } from '../.storybook/data/wp-admin-admin.php-page=googlesitekit-module-analytics-googlesitekit';
 import { STORE_NAME as CORE_SITE } from '../assets/js/googlesitekit/datastore/site/constants';
 import { STORE_NAME as ANALYTICS_STORE } from '../assets/js/modules/analytics/datastore/constants';
+import { STORE_NAME as SEARCH_CONSOLE_STORE } from '../assets/js/modules/search-console/datastore/constants';
```
aaemnnosttv, line 27: @ryanwelcher looks like there was some inconsistency before, but would you please update these to use the same naming convention we use elsewhere for the store name alias?
`CORE_SITE` is correct above, but `ANALYTICS_STORE` should be `MODULES_ANALYTICS` and `SEARCH_CONSOLE_STORE` should be `MODULES_SEARCH_CONSOLE`.

felixarntz (changes requested):
Almost LGTM, there are two critical things though I just spotted now while testing.

felixarntz (approved):
Awesome work!

Issue #2049, opened 2020-09-17
Fix incorrect translation strings


felixarntz:
There are several important principles when it comes to providing translatable strings:
* Translation strings should always consist of full sentences that make sense on their own - they should never be composed out of concatenated strings.
* Variable values that need to be included in translation strings should always use placeholders like `%s`, never concatenation.
* Basically, no concatenation should ever be used around translation strings, unless it is to separate two individual sentences from each other to avoid an overly long translation string (which would have the drawback that a change in one sentence would also invalidate translations for all the other sentences).
* Translation strings should contain as little markup as possible - e.g. a string containing `<strong>%s</strong>` should instead place the `strong` tags as part of the placeholder value. Still, it should be included where it cannot be avoided (rather than splitting up the string, per the above points).
* If a translation string could mean multiple things, `_x` should be used to provide a context to briefly explain the intended purpose in as few words as possible.
* Translation strings should not include trailing spaces.

There are numerous occurrences of translation strings in Site Kit where these principles are violated, causing issues with localization. Furthermore, there are some instances where translator comments could be clarified.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Translatable strings in Site Kit should not violate the translation principles outlined on top of this issue.

## Implementation Brief

* Review translation strings across the plugin codebase to find violations of the above principles, and also look out for almost similar strings that should be combined into a single one.
* Fix the problems.
* Merge #2051.

## QA Brief

* General smoke testing, but in particular test the following:
    * Test whether the Site Kit dashboard and the Site Kit module pages still look like they did before.
    * Check that the publisher win notification for "Total Stats" (requires site with more than 5 published posts with Search Console data available at the point of new plugin activation) looks like before, including a "for month" suffix after the numbers.
    * Check that the WP dashboard widget renders blue boxes about Search Console and Analytics data being unavailable when using a site that doesn't have data on those services (e.g. a new site).

## Changelog entry

* Fix various translation strings to no longer violate localization best practices.

aaemnnosttv:
IB ✅

cole10up:
**Tested**

Installed, activated and setup latest SK.

Ran a series of regression tests around:

- Dashboard and modules, look and feel. **Pass**
![image](https://user-images.githubusercontent.com/57115907/94212744-dd992f80-fe89-11ea-9225-0c46205dbc24.png)

[...]

PR #2051, opened 2020-09-17
Fix incorrect translation strings

felixarntz:
## Summary

Addresses issue #2049

## Relevant technical choices

* Some almost similar strings were combined into one to avoid unnecessary work for translators.
* Some strings that were missing a context (particularly service and core web vital names which shouldn't be translated but still run through a translation function) were added that.
* The `DataBlock` component was modified so that the `period` prop has to include an `%s` placeholder to make translation strings work correctly.

## Checklist


assets/js/components/setup-wizard/wizard-steps.js
```diff
@@ -44,7 +44,7 @@ const STEPS = {
 		Component: WizardStepSearchConsoleProperty,
 	},
 	completeSetup: {
-		title: __( 'Finish', 'google-site-kit' ),
+		title: _x( 'Finish', 'action', 'google-site-kit' ),
```
aaemnnosttv, line 47: Is one word for the context sufficient here? "Finish" itself is a verb so I'm not sure "action" will be useful here.

Perhaps something like this?
```suggestion
		title: _x( 'Finish', 'complete module setup', 'google-site-kit' ),
```

assets/js/components/setup/compatibility-checks.js
```diff
@@ -174,29 +174,32 @@ export default class CompatibilityChecks extends Component {
...
 						dangerouslySetInnerHTML={ sanitizeHTML(
+							__( 'Looks like your site is having a technical issue with requesting data from Google services.', 'google-site-kit' ) +
+							'<br/>' +
 							sprintf(
 								/* translators: %1$s: Support Forum URL, %2$s: Error message */
-								__( 'Looks like your site is having a technical issue with requesting data from Google services.<br/>To get more help, ask a question on our <a href="%1$s">support forum</a> and include the text of the original error message:<br/>%2$s', 'google-site-kit' ),
+								__( 'To get more help, ask a question on our <a href="%1$s">support forum</a> and include the text of the original error message: %2$s', 'google-site-kit' ),
 								'https://wordpress.org/support/plugin/google-site-kit/',
-								error
+								'<br/>' + error
```
aaemnnosttv, line 183: Perhaps the translators comment should be updated to reflect that the placeholder value contains a new line?

Stylistically, perhaps we should use a template string here rather than `+` ? (just a thought)

E.g.
```js
dangerouslySetInnerHTML={ sanitizeHTML(
	`
	${ __( 'Looks like your site is having a technical issue with requesting data from Google services.', 'google-site-kit' ) }
	<br/>
	${ sprintf(
		/* translators: %1$s: Support Forum URL, %2$s: Error message */
		__( 'To get more help, ask a question on our <a href="%1$s">support forum</a> and include the text of the original error message: %2$s', 'google-site-kit' ),
		'https://wordpress.org/support/plugin/google-site-kit/',
		`<br/>${ error }`
	) }
	`,
	{
		ALLOWED_TAGS: [ 'a', 'br' ],
		ALLOWED_ATTR: [ 'href' ],
	}
) }
```

assets/js/modules/adsense/components/dashboard/AdSenseDashboardWidget.js
```diff
@@ -119,6 +119,24 @@ class AdSenseDashboardWidget extends Component {
...
 		const wrapperClass = ( loading || ! receivingData || zeroData ) ? 'googlesitekit-nodata' : '';
 
+		let moduleStatus;
+		let moduleStatusText;
+		if ( ! error && modulesData.adsense.setupComplete ) {
+			moduleStatus = 'connected';
+			moduleStatusText = sprintf(
+				/* translators: %s: module name. */
+				__( '%s is connected', 'google-site-kit' ),
+				_x( 'AdSense', 'Service name', 'google-site-kit' )
```
aaemnnosttv, line 129: Should we maybe use `module.name` here instead of translating it again?

assets/js/modules/adsense/components/settings/SettingsSetupIncomplete.js
```diff
@@ -41,9 +41,9 @@ export default function SettingsSetupIncomplete() {
 		<Fragment>
 			<AdBlockerWarning />
 
-			{ isPendingStatus && __( 'Site Kit has placed AdSense code on your site, ', 'google-site-kit' ) }
-			{ ! isPendingStatus && __( 'Setup incomplete: ', 'google-site-kit' ) }
-
+			{ isPendingStatus && __( 'Site Kit has placed AdSense code on your site:', 'google-site-kit' ) }
+			{ ! isPendingStatus && __( 'Setup incomplete:', 'google-site-kit' ) }
+			{ ' ' }
```
aaemnnosttv, line 46: This is concatenating here as it will either render

`Site Kit has placed AdSense code on your site:` + ` ` + [`check module page`](#)
or
`Setup incomplete:` + ` ` + [continue module setup](#)

Perhaps this would be a good use-case for [`createInterpolateElement`](https://make.wordpress.org/core/2020/07/17/introducing-createinterpolateelement/) ? That way the whole string with the markup could be translatable while allowing us to still use the `Link` component.

assets/js/modules/analytics/components/common/AcquisitionPieChart.js
```diff
@@ -66,11 +66,10 @@ function AcquisitionPieChart( { data, args, source } ) {
 
 	let sourceMessage = '';
 	if ( source ) {
-		sourceMessage = sprintf(
-			/* translators: %1$s: URL to Analytics Module page in Site Kit Admin, %2$s: Analytics (Service Name) */
-			__( 'Source: <a class="googlesitekit-cta-link googlesitekit-cta-link--external googlesitekit-cta-link--inherit" href="%1$s">%2$s</a>', 'google-site-kit' ),
+		sourceMessage = __( 'Source:', 'google-site-kit' ) + sprintf(
+			' <a class="googlesitekit-cta-link googlesitekit-cta-link--external googlesitekit-cta-link--inherit" href="%1$s">%2$s</a>',
```
aaemnnosttv, line 70: Here is another concatenation which should be updated. Even though `Source:` and the link are essentially separate, it isn't RTL friendly.

This would be another good use for `createInterpolateElement` I think so that we wouldn't need to include all of the classnames in the translation string (or extract them to a placeholder). That would allow us to use the `Link` component as well, rather than using its resulting HTML as a string which will be a pain to maintain if we were to make changes to the base `Link` component.

assets/js/modules/analytics/components/common/ProfileNameTextField.js
```diff
@@ -45,7 +45,7 @@ export default function ProfileNameTextField() {
 
 	useEffect( () => {
 		if ( ! profileName ) {
-			setValues( FORM_SETUP, { profileName: __( 'All Web Site Data', 'google-site-kit' ) } );
+			setValues( FORM_SETUP, { profileName: _x( 'All Web Site Data', 'default Analytics view name', 'google-site-kit' ) } );
```
aaemnnosttv, line 48: Much better 👌

assets/js/modules/analytics/components/dashboard/DashboardBounceRateWidget.js
```diff
@@ -113,7 +113,7 @@ function DashboardBounceRateWidget() {
 			className="overview-bounce-rate"
 			title={ __( 'Bounce Rate', 'google-site-kit' ) }
 			datapoint={ Number( averageBounceRate ).toFixed( 2 ) }
-			datapointUnit={ __( '%', 'google-site-kit' ) }
+			datapointUnit="%"
```
aaemnnosttv, line 116: See my previous comment in the `DataBlock` as to the problem with this.

aaemnnosttv (changes requested):
This looks pretty solid overall, although there are a few places where things need to be updated to avoid concatenation, or generally improve the way we're localizing a string.

The most substantial changes that seem to be needed are around the `DataBlock` component which both deals with number formatting (currently not using locale-specific format) as well as concatenation with some "units" – currently either `%` or `$`. I would imagine `%` is probably ubiquitous but wouldn't be surprised if it wasn't. `$` of course is not something we should assume anywhere though. We also can't assume which side of a value these things belong on. Where possible we should leverage the browser's ability to format numbers (including currencies) – I'm not sure if it supports percentages, but that would be great if it could.

Finally, there are a few of the more complicated strings which mix a string with a component (either as HTML or JSX) that look like great use-cases for `createInterpolateElement`. We haven't really used it yet, but from what I've seen, it really shines in this case as it allows us to keep all of the implementation details of the HTML out of the translation string, as well as allows us to use components as well.

All-in-all though, this is a big improvement across the board 👍

assets/js/components/setup-wizard/wizard-steps.js
felixarntz: I think technically "action" is sufficient because a context here is only needed to indicate that in this case it's a verb and thus separate from a potentially different translation string where "finish" could be a noun. But it doesn't hurt to be more specific.

assets/js/components/setup/compatibility-checks.js
felixarntz: I don't think that's necessary here, the translators shouldn't need to care about that.

assets/js/modules/adsense/components/settings/SettingsSetupIncomplete.js
felixarntz: I think we can keep this as is, IMO we should only need to use that if we need to break up a sentence into multiple pieces due to a component being used.

assets/js/components/data-block.js
```diff
@@ -124,7 +125,8 @@ class DataBlock extends Component {
 								/>
 							</span>
 							<span className="googlesitekit-data-block__value">
-								{ `${ numberFormat( Math.abs( change ) ) }${ changeDataUnit } ${ period }` }
+								{ period && sprintf( period, `${ numberFormat( Math.abs( change ) ) }${ changeDataUnit }` ) }
+								{ ! period && `${ numberFormat( Math.abs( change ) ) }${ changeDataUnit }` }
```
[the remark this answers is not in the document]
felixarntz: I've modified this by trying to format as a currency, which will solve the currency formatting issue (good catch!). I've also removed usage of hard-coded `$`s, that's a nice improvement we hadn't spotted before!

As far as the other potential units go, I think we shouldn't have a generic localized concatenation. Since we only have `%` for now, I don't think it's relevant - maybe we can revisit that part later.

assets/js/modules/adsense/components/dashboard/AdSenseDashboardWidget.js
felixarntz: That's a good point, however I'd say let's change that separately. There are a bunch of those references, and I think we could fix those throughout. IMO this makes for a good distinct follow-up issue only focused on that.

assets/js/components/setup-wizard/wizard-steps.js
aaemnnosttv: Good point; I don't really think of "finish" as a noun as well, so in that regard I suppose it makes more sense. I think the extra context would still be useful in some cases if it were to affect the conjugation although I don't have a concrete example in mind.

aaemnnosttv (commented):
The changes look good here, just the one question remaining.

aaemnnosttv (approved):
Awesome!

PR #2055, opened 2020-09-17
Refactor Error components to support multiple errors.

asvinb:
## Summary

Addresses issue #1859 

## Relevant technical choices
- Added new StoreErrorNotices which handles all errors compared to StoreErrorNotice which caters for a single error
- New tests added for create-error-store

## Checklist


asvinb:
@googlebot I fixed it.

assets/js/googlesitekit/data/create-error-store.test.js
```diff
@@ -0,0 +1,221 @@
...
+				expect( selectedError ).toMatchObject( ERROR_SAMPLE_2.error );
+			} );
+		} );
+
+		describe( 'getError', () => {
+			it( 'requires a baseName param', () => {
+				select.getError();
+
+				expect( store.getState().error ).toBeUndefined();
+			} );
```
aaemnnosttv, line 169: Similar to previous tests, this test should assert that the selector throws an error when `args` is provided, but `baseName` is empty.

assets/js/googlesitekit/data/create-error-store.test.js
```diff
@@ -0,0 +1,221 @@
...
+
+			it( 'returns the correct error', () => {
+				dispatch.receiveError( ERROR_SAMPLE.error, ERROR_SAMPLE.baseName, ERROR_SAMPLE.args );
+				dispatch.receiveError( ERROR_SAMPLE_2.error, ERROR_SAMPLE_2.baseName, ERROR_SAMPLE_2.args );
+
+				const error = select.getError( ERROR_SAMPLE_2.baseName, ERROR_SAMPLE_2.args );
+				expect( error ).toMatchObject( ERROR_SAMPLE_2.error );
+			} );
```
aaemnnosttv, line 177: We need to cover the following here:
* `state.error` is returned when `getError()` is called with no arguments
* The appropriate error is returned when called _with_ a `baseName` and `args`

assets/js/googlesitekit/data/create-error-store.test.js
```diff
@@ -0,0 +1,221 @@
...
+		} );
+
+		describe( 'getErrors', () => {
+			it( 'returns an empty array if there are no errors', () => {
+				const errors = select.getErrors();
+				expect( errors ).toEqual( [] );
+			} );
```
aaemnnosttv, line 184: 👍

assets/js/googlesitekit/data/create-error-store.test.js
```diff
@@ -0,0 +1,221 @@
...
+
+			it( 'checks if we have only 1 error', () => {
+				dispatch.receiveError( ERROR_SAMPLE.error, ERROR_SAMPLE.baseName, ERROR_SAMPLE.args );
+				const errors = select.getErrors();
+				expect( errors ).toHaveLength( 1 );
+			} );
```
aaemnnosttv, line 190: Generally speaking, selectors that "check" either return a boolean or throw, so the description is a little off here.

After selecting the errors, it's okay to check the length, but we can be a bit more thorough as we have the error object as well. We can actually test both with `expect( errors ).toEqual( [ error ] )` since there should only be one.

assets/js/googlesitekit/data/create-error-store.test.js
```diff
@@ -0,0 +1,221 @@
...
+
+			it( 'checks if we have more than 1 error', () => {
+				dispatch.receiveError( ERROR_SAMPLE.error, ERROR_SAMPLE.baseName, ERROR_SAMPLE.args );
+				dispatch.receiveError( ERROR_SAMPLE_2.error, ERROR_SAMPLE_2.baseName, ERROR_SAMPLE_2.args );
+				const errors = select.getErrors();
+				expect( errors ).toHaveLength( 2 );
```
aaemnnosttv, line 196: Here it's okay to check the length because we don't know what order the elements will come back in, but we should also assert that `errors` contains both errors. For that we should be able to use [`expect.arrayContaining`](https://jestjs.io/docs/en/expect#expectarraycontainingarray)

assets/js/googlesitekit/data/create-error-store.test.js
```diff
@@ -0,0 +1,221 @@
...
+			} );
+
+			it( 'checks if duplicate errors are removed', () => {
+				dispatch.receiveError( ERROR_SAMPLE.error, ERROR_SAMPLE.baseName, ERROR_SAMPLE.args );
+				dispatch.receiveError( ERROR_SAMPLE_2.error, ERROR_SAMPLE_2.baseName, ERROR_SAMPLE_2.args );
+				dispatch.receiveError( ERROR_SAMPLE_2.error, ERROR_SAMPLE_2.baseName, ERROR_SAMPLE_2.args );
+				dispatch.receiveError( ERROR_SAMPLE_2.error, ERROR_SAMPLE_2.baseName, ERROR_SAMPLE_2.args );
+				const errors = select.getErrors();
+				expect( errors ).toHaveLength( 2 );
+			} );
```
aaemnnosttv, line 206: This test actually doesn't test the functionality of the unique list handling because receiving the same error multiple times will only overwrite the same key in `state.errors` (it won't actually add it multiple times).

This is only relevant when receiving the _same error_ multiple times, with different args, like what we do for the legacy compatibility. You can see what I mean here: https://github.com/google/site-kit-wp/blob/f244c63925cd56d0bb03a048d0b387824c335fe5/assets/js/googlesitekit/data/create-fetch-store.js#L159-L165 

So the return here should be unique regardless of any `baseName`/`args` provided 🙂

assets/js/googlesitekit/data/create-error-store.test.js
```diff
@@ -0,0 +1,221 @@
...
+		} );
+
+		describe( 'hasErrors', () => {
+			it( 'returns false if there are no errors', () => {
+				expect( select.hasErrors() ).toBeFalsy();
```
aaemnnosttv, line 211:
```suggestion
				expect( select.hasErrors() ).toBe( false );
```

assets/js/googlesitekit/data/create-error-store.test.js
```diff
@@ -0,0 +1,221 @@
...
+			} );
+
+			it( 'returns true if there are errors', () => {
```
aaemnnosttv, line 214:
```suggestion
			it( 'returns true if there are any errors', () => {
```

assets/js/googlesitekit/data/create-error-store.test.js
```diff
@@ -0,0 +1,221 @@
...
+				dispatch.receiveError( ERROR_SAMPLE.error, ERROR_SAMPLE.baseName, ERROR_SAMPLE.args );
+				dispatch.receiveError( ERROR_SAMPLE_2.error, ERROR_SAMPLE_2.baseName, ERROR_SAMPLE_2.args );
+				expect( select.hasErrors() ).toBeTruthy();
```
aaemnnosttv, line 217: Let's make sure this returns true regardless of how the error is received to ensure this returns `true` even if it only receives the `error` with no basename/args as well as when there are others.

aaemnnosttv (changes requested):
Thanks @asvinb ! This looks good overall, there's just a few things to change. I left a number of comments for you to explain what's needed there as well as left a few suggestions for improvement in some places. The majority of the changes and comments are around the tests. Let me know if you have any questions.

aaemnnosttv:
@googlebot I consent.

felixarntz (changes requested):
Great work! Almost LGTM, just one comment around decoupling `ErrorNotice` from module-specific behavior.

assets/js/googlesitekit/data/create-error-store.js
```diff
@@ -119,7 +119,7 @@ export function createErrorStore() {
 				const newState = { ...state };
 				if ( baseName ) {
 					newState.errors = { ...( state.errors || {} ) };
-					for ( const key in Object.keys( newState.errors ) ) {
+					for ( const key in newState.errors ) {
```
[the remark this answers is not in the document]
aaemnnosttv: Yep 😕; found when adding tests.

assets/js/components/StoreErrorNotice.js
```diff
@@ -25,37 +25,30 @@ import PropTypes from 'prop-types';
...
+import ErrorNotice from '../components/ErrorNotice';
 import { STORE_NAME as CORE_MODULES } from '../googlesitekit/modules/datastore/constants';
-import { isPermissionScopeError, isInsufficientPermissionsError } from '../util/errors';
+import { isInsufficientPermissionsError } from '../util/errors';
 import { getInsufficientPermissionsErrorDescription } from '../util/insufficient-permissions-error-description';
-import ErrorText from '../components/error-text';
 const { useSelect } = Data;
 
-function StoreErrorNotice( { moduleSlug, storeName, shouldDisplayError } ) {
+export default function StoreErrorNotice( { storeName, shouldDisplayError, moduleSlug } ) {
```
aaemnnosttv, line 34: @felixarntz this component is no longer used at all – should we keep it around?

felixarntz: @aaemnnosttv I think it can be removed in that case.

assets/js/components/StoreErrorNotices.js
```diff
@@ -25,37 +25,34 @@ import PropTypes from 'prop-types';
...
-	let message = error.message;
-	if ( isInsufficientPermissionsError( error ) ) {
-		message = getInsufficientPermissionsErrorDescription( message, module );
-	}
-
-	return <ErrorText message={ message } reconnectURL={ error.data?.reconnectURL } />;
+	return errors
+		.map( ( error ) => {
+			if ( isInsufficientPermissionsError( error ) ) {
+				error.message = getInsufficientPermissionsErrorDescription( error.message, module );
```
felixarntz, line 41: We should probably destructure this into a new object, so that we don't modify the original `error` object.
```suggestion
				error = {
					...error,
					message: getInsufficientPermissionsErrorDescription( error.message, module )
				};
```

felixarntz (commented):
Just one small thing to address, otherwise LGTM.

felixarntz (approved):
Awesome!

Issue #1925, opened 2020-08-17
Move date range parsing logic and actual dates awareness to the client


felixarntz:
At the moment, the client fully relies on date ranges such as `last-28-days`. These are then passed to the server where they are parsed into actual start and end dates. There are cases where we need awareness of the actual dates on the client too though.

We've already updated the server-side layers to accept actual start and end dates via #1774, #1775, and #1776, so let's start moving towards that goal where we no longer require that date-parsing logic on the server and can instead just pass through dates to the respective Google API.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a new selector `getDateRangeDates( options = {} )` on the `core/user` store:
    * `options.offsetDays` is an integer that is the number of days to offset (e.g. if the API does not include fresh data for the last 1 or 2 days yet). Default should be `0`.
    * `options.compare` is a boolean for whether date ranges to compare should be included. Default should be `false`.
    * `options.weekDayAlign` is a boolean for whether the compared date range should be aligned for the weekdays. Only relevant if `options.compare` is `true`. Default should be `false`. See #1280 for reference on why that exists.
    * The selector should return an object that has `startDate` and `endDate` (format `Y-m-d`). If `options.compare` is true, the object should also include `compareStartDate` and `compareEndDate`. The compare dates should always come _before_ (i.e. if `startDate` is "2020-04-05", `compareEndDate` will definitely not be greater than "2020-04-04").
    * The selector logic should fully replicate the PHP logic from `Module::parse_date_range`, except:
        * It doesn't need `$previous`, since it will always include the previous dates if `options.compare` is `true`.
        * It doesn't need `$multiplier`, since that is only to query an extra period for APIs that don't support multiple date ranges. For those cases, it should be relied upon `options.compare` and then pass `compareStartDate` and `endDate` to the respective API in order to get the full range in one query.
    * Comprehensive Jest test coverage should be added.
* The Analytics `GET:report` datapoint in PHP should support receiving `compareStartDate` and `compareEndDate` arguments. If both of them are provided, they should make up a second date range object for the query.
    * This should only apply if the recently added `startDate` and `endDate` are also present, it shouldn't affect the "legacy" logic that relies on the `dateRange` argument.
* For now, no actual queries should be updated, this can happen in follow-up issues.

## Implementation Brief

* Add a new `selectors.getDateRangeDates` to `assets/js/googlesitekit/datastore/user/date-range.js`
[...]

Issue #1394, opened 2020-04-16
Analytics chart showing incorrect data/dates for Users


ryanwelcher:
## Bug Description

On the Analytics chart, the user count displayed when viewing a single date is actually showing data for the following day.

Additionally, the previous month is showing the same date as the selected range but the data is correct.

The label displayed that represents the previous range doesn't accurately reflect what is actually being seen. Instead of displaying `Previous Month`, perhaps the label should show the selected range such as  `Previous 90 Days`. Alternatively, we could mirror what is being done in the Analytics UI and use the actual date ranges.

Discovered while investigating #1280.

## Steps to reproduce

**Off by one data:**
1. Go to the Analytics screen in Site Kit
2. Hover over any date on the blue line to view User count for that day.
3. Compare the same date in Analytics account.

**Incorrect previous date:**
1. Go to the Analytics screen in Site Kit
2. Hover over any date on the blue line to view User count for that day.
3. Hover the same date on the dashed line
4. The dates for both ranges are the same but the User count is correctly displayed.
## Screenshots
**Off by one data**
![data-offset.png](https://images.zenhubusercontent.com/5e74b5396e1fdc722577f6d3/cc0b6b56-0536-47cb-a473-a8a40c3c20a4)

**Incorrect previous date:**
The previous date should read Apr 2, 2020 here.
![same-date.gif](https://images.zenhubusercontent.com/5e74b5396e1fdc722577f6d3/5db66506-6bb5-4c81-bf22-f75a216c7ece)
## Additional Context

 - PHP Version: 7.2
 - OS: Mac OS
 - Browser: Crome
 - Plugin Version: develop branch.
 - Device: MacBook Pro

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Analytics statistics should display the correct date-specific value for each date
* The Audience overview chart should label the "Previous month" series as "Previous period" (widget/area titles already have the duration)
* The series referenced when hovering an individual node in the "Previous month" series, should be labeled as "Previous period".

## Implementation Brief

* `In assets/js/modules/analytics/util/index.js` make the following changes:
    * Extract the year, month and day from `dateString`
    * Adjust the month to be zero-indexed.
    * Pass the values `new Date( year, month, day)` to get the correct date.
* Merge #1395 

## QA Brief
* Go to the Analytics screen in Site Kit
* Hover over any date on the blue line to view User count for that day.
[...]

PR #2073, opened 2020-09-18
Enhancement/1925 move date range parsing and actual dates to client

caseymorrisus:
## Summary

Addresses issue [#1925](https://github.com/google/site-kit-wp/issues/1925)

## Relevant technical choices

* Updates `Analytics::create_data_request` to return `array( compareStartDate, compareEndDate, startDate, endDate )` if both `$data['compareStartDate']` and `$data['compareEndDate']` are valid dates
* Adds file at `assets/js/googlesitekit/datastore/user/utils.js` which contains various date parsing utility functions to handle dates and date ranges.
  * `isValidDateInstance` returns `true` if the supplied `date` param constructs to a valid date instance using `new Date( date )`
  * `getDateString` parses the given `date` param into a date string with the format `YYYY-MM-DD`. Throws an error if the `date` param does not construct to a valid date instance.
  * `getPreviousDate` takes a `relativeDate` param and returns the previous date based on the `daysBefore` param. Throws an error if `relativeDate` is not a valid date string.
  * `getPreviousWeekDate` similar to `getPreviousDate` but shifts the resulting date so that it returns a date that falls on the same day of the week as the `relativeDate` param while preserving the `daysBefore` as a minimum. Throws an error if `relativeDate` is not a valid date string.
* Edits file at `assets/js/googlesitekit/datastore/user/date-range.js` to add new `getDateRangeDates` selector.
  * `getDateRangeDates` Returns the current date range as a list of date strings. Will throw an error if the current date range is invalid. If `compare: true` return will be `[ compareStartDate, compareEndDate, startDate, endDate, today ]` otherwise return will be `[ startDate, endDate, today ]`.

## Checklist


assets/js/googlesitekit/datastore/user/date-range.js
```diff
@@ -79,6 +84,50 @@ export const selectors = {
...
+	getDateRangeDates( state, {
+		offsetDays = 1,
+		compare = false,
+		weekDayAlign = false,
+	} = {} ) {
+		const dateRange = selectors.getDateRange( state );
+		const today = getDateString( new Date() );
+		const endDate = getPreviousDate( today, offsetDays );
+		const matches = dateRange.match( '-(.*)-' );
+		const numberOfDays = Number( matches ? matches[ 1 ] - 1 : 28 );
```
felixarntz, line 108: Why the `- 1` here? This should be exactly the days from the date range string. I think the `- 1` needs to go in the line below, where it's more clear on why this happens. If this is here, the alternative value afterwards would also need to be hard-coded to be `27` instead, but that overall is less understandable I'd argue that making the subtraction below where it matters.

assets/js/googlesitekit/datastore/user/date-range.js
```diff
@@ -79,6 +84,50 @@ export const selectors = {
...
+			const compareEndDate = getPreviousDate( startDate, 1 );
+			parsed.push( getPreviousDate( compareEndDate, numberOfDays ) );
+			parsed.push( compareEndDate );
+		}
+
+		if ( weekDayAlign ) {
+			const compareEndDate = getPreviousWeekDate( endDate, numberOfDays );
+			parsed.push( getPreviousDate( compareEndDate, numberOfDays ) );
+			parsed.push( compareEndDate );
+		}
```
felixarntz, line 122: I think this should be simplified:
* This should be wrapped in a single `if ( compare )` clause, because `weekdayAlign` is irrelevant otherwise.
* The assignment for `compareEndDate` is the only thing which is dependent on whether `weekdayAlign` or not.
* Let's assign these values to variables (i.e. add `compareStartDate`), no need for an array here (see other comment below).

assets/js/googlesitekit/datastore/user/date-range.js
```diff
@@ -79,6 +84,50 @@ export const selectors = {
...
+
+		return [
+			...parsed,
+			startDate,
+			endDate,
+			today,
+		];
```
felixarntz, line 129: This return value is off and doesn't match the ACs. It should be an **object** containing only `startDate` and `endDate`, or also `compareStartDate` and `compareEndDate` if `compare` is `true`.

It shouldn't include `today` or an array of the compare dates.

assets/js/googlesitekit/datastore/user/date-range.test.js
```diff
@@ -21,8 +21,10 @@
  */
 import { createTestRegistry } from '../../../../../tests/js/utils';
 import { STORE_NAME } from './constants';
+import { getPreviousDate, getDateString, getPreviousWeekDate } from './utils';
 
 describe( 'core/user date-range', () => {
+	const today = getDateString( new Date() );
```
felixarntz, line 27: Looking at the tests here, most of them will need to be updated due to the return value change required above. However, one other concern: The tests themselves rely on all the utility functions which are already tested elsewhere, and I feel like it would be more appropriate to test specific hard-coded input dates and result dates, relying on only `getDateRangeDates` calls instead.

How about the following idea:
* We add another optional `options.referenceDate` to `getDateRangeDates`, which if not provided will be set to "today" (i.e. what you have above).
* This argument wouldn't be used in production code, but is helpful for testing.
* This way all the tests could be updated to work with hard-coded dates so that they no longer rely on other utility functions, ensuring that the actual selector works as expected, regardless of the utility functions (and whether they may unexpectedly change).

assets/js/googlesitekit/datastore/user/utils.js
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * @since n.e.x.t
+ *
+ * @param {string} relativeDate Date string (YYYY-MM-DD) to subtract days from.
+ * @param {number} daysBefore   Number of days to subtract from relativeDate.
+ * @return {string}             The date string (YYYY-MM-DD) for the previous date.
+ */
+export const getPreviousDate = ( relativeDate = '', daysBefore ) => {
+	const dateArray = relativeDate.split( '-' );
+	const isValidRelativeDate = dateArray.length === 3 && isValidDateInstance( new Date( relativeDate ) );
```
felixarntz, line 72: It might make sense to move this to a `isValidDateString` function maybe?

assets/js/googlesitekit/datastore/user/utils.js
```diff
@@ -0,0 +1,104 @@
...
+	invariant( isValidRelativeDate, INVALID_DATE_INSTANCE_ERROR );
```
felixarntz, line 73: I'd prefer to use another error message here because this tests for a valid date string rather than a valid `Date` instance.

assets/js/googlesitekit/datastore/user/utils.js
```diff
@@ -0,0 +1,104 @@
...
+
+	const [ year, month, day ] = dateArray;
+	const date = new Date( year, month - 1, day );
```
felixarntz, line 76: Why is this splitting needed? Wouldn't `new Date( relativeDate )` be the same?

assets/js/googlesitekit/datastore/user/utils.js
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * @since n.e.x.t
+ *
+ * @param {string} relativeDate Date string (YYYY-MM-DD) to subtract days from.
+ * @param {number} daysBefore   Number of days to subtract from relativeDate.
+ * @return {string}             The date string (YYYY-MM-DD) for the previous date, shifted to fall on the same day of the week.
+ */
+export const getPreviousWeekDate = ( relativeDate, daysBefore ) => {
+	const date = new Date( relativeDate );
+	invariant( isValidDateInstance( date ), INVALID_DATE_INSTANCE_ERROR );
```
felixarntz, line 96: Maybe we should do the same validity check here as we do in `getPreviousDate` (since this also expects a date string of `YYYY-MM-DD` format)?

assets/js/googlesitekit/datastore/user/utils.js
```diff
@@ -0,0 +1,104 @@
...
+
+	const sameDay = daysBefore === 0;
+	const sameDayOfWeek = daysBefore % 7 === 0;
+	const addedDays = sameDay ? 7 : 7 - ( daysBefore % 7 );
+	const daysRemoved = daysBefore + ( sameDayOfWeek && ! sameDay ? 0 : addedDays );
```
felixarntz, line 101: The PHP logic for this is slightly different as far as I can tell:
* I don't think we need to check for `sameDay`. If `daysBefore` is 0 (which is not a normal use-case for us), I'd expect this to return the exact same date as the input date. So that would also simplify this code here.
* This code currently will only add more days to `daysBefore`, but in the PHP logic it has some logic where it may add or remove days, based on what "same" weekday date is closer. Looking at this now, I'm not really sure whether your approach here or the one used in PHP is actually correct - any thoughts @adamsilverstein? As far as I can tell the current PHP logic may result in the date range including a few days that are also part of the other date range to compare with, which isn't something that I think would be intentional?

assets/js/googlesitekit/datastore/user/utils.test.js
```diff
@@ -0,0 +1,202 @@
...
+		} );
+		it( 'should throw error if date supplied is invalid date', () => {
+			try {
+				getPreviousWeekDate( '2020-99-99', 1 );
+			} catch ( error ) {
+				expect( error.message ).toEqual( INVALID_DATE_INSTANCE_ERROR );
+			}
+		} );
+		it( 'should go back 7 days with daysBefore value of 0 and date of "2020-01-30"', () => {
+			expect( getPreviousWeekDate( '2020-01-30', 0 ) ).toEqual( '2020-01-23' );
```
felixarntz, line 172: See other comment, I'd expect this to return the exact same date if the "offset" is `0`.

assets/js/googlesitekit/datastore/user/utils.js
```diff
@@ -0,0 +1,104 @@
+/**
```
felixarntz, line 1: Instead of creating a single `utils.js` file, I think it would be a bit cleaner to create a `util` subdirectory and have one file per function. The tests could also then be split across files in the same way.

felixarntz (changes requested):
This is looking on the right track overall, there are a couple things to revisit though.

assets/js/googlesitekit/datastore/user/date-range.js
```diff
@@ -79,6 +84,50 @@ export const selectors = {
...
+	 * @param {boolean} options.compare      Set to true if date ranges to compare should be included. Default is: false.
+	 * @param {boolean} options.weekDayAlign Set to true if the compared date range should be aligned for the weekdays. Default is: false.
+	 * @return {Array.<string>}              List of date ranges in chronoligical order. [compareStart, compareEnd, start, end, today]
+	 */
+	getDateRangeDates( state, {
+		offsetDays = 1,
+		compare = false,
+		weekDayAlign = false,
+	} = {} ) {
+		const dateRange = selectors.getDateRange( state );
```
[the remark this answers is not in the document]
aaemnnosttv: @felixarntz we only use a registry selector to select the from the same store when that selector has a resolver to make sure it is triggered. If not, it's fine to call it like this 👍

assets/js/googlesitekit/datastore/user/date-range.js
```diff
@@ -79,6 +84,50 @@ export const selectors = {
...
+	 *
+	 * @param {Object}  state                The current data store's state.
+	 * @param {Object} 	options              Options parameter. Default is: {}.
+	 * @param {number} 	options.offsetDays   Number of days to offset. Default is: 0.
+	 * @param {boolean} options.compare      Set to true if date ranges to compare should be included. Default is: false.
+	 * @param {boolean} options.weekDayAlign Set to true if the compared date range should be aligned for the weekdays. Default is: false.
+	 * @return {Array.<string>}              List of date ranges in chronoligical order. [compareStart, compareEnd, start, end, today]
+	 */
+	getDateRangeDates( state, {
+		offsetDays = 1,
```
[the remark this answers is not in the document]
caseymorrisus: @felixarntz I can definitely change that but want to verify expected behavior on this part. The PHP version also has a default `offsetDays` value of `1`, this is an attempt to replicate the behavior of `Module::parse_date_range`. Having this default to `1` causes the `getDateRangeDates` method to exclude today in the date range whereas defaulting to `0` would include today (eg `getDateRangeDates()` with default `offsetDays` of `0` would return `['2020-09-22', '2020-09-23']` with a date range value of `'last-2-days'`). To sum it up, should today be included in the date range by default or should that be opt-in (explicit passing of `0`)?

assets/js/googlesitekit/datastore/user/utils.js
caseymorrisus: Unfortunately the date format we use (YYYY-MM-DD) doesn't play nicely when passed directly to the date constructor, it's off by 1 day (in my locale, on Chrome). This is due to the browser interpreting the string as UTC whereas my local timezone has a negative offset from UTC, putting me in the day before. Parsing dates by string is also browser specific, explicitly passing numbers gives consistent behavior. If we want to parse as strings directly, I'd recommend pulling in a library such as `date-fns`. 

`new Date('2020-09-23')` outputs `Tue Sep 22 2020 17:00:00 GMT-0700 (Pacific Daylight Time)` which is unexpected
`new Date(2020, 8, 23)` outputs `Wed Sep 23 2020 00:00:00 GMT-0700 (Pacific Daylight Time)` which is expected

If our dates were formatted using `YYYY-M-DD` this would work as expected (at least within Chrome as it isn't detecting a UTC date):
`new Date('2020-9-23')` outputs `Wed Sep 23 2020 00:00:00 GMT-0700 (Pacific Daylight Time)`.

I could always parse the string, but figured breaking into an array was an easy way to accomplish this across environments as well as giving us quick validation (verifying that array length is 3, otherwise date is invalid).

If helpful, I can include a summarized version of this info within a comment for future use.

assets/js/googlesitekit/datastore/user/utils.js
```diff
@@ -0,0 +1,104 @@
...
+ * Asserts whether a given date instance is valid or invalid.
+ *
+ * @since n.e.x.t
+ *
+ * @param {Date} date Date instance to be asserted against.
+ * @return {boolean}  True if the given date instance is valid.
+ */
+export const isValidDateInstance = ( date ) => {
+	return date instanceof Date && ! isNaN( date );
```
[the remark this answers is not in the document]
caseymorrisus: @aaemnnosttv The type coercion provided by global `NaN` is expected and is needed for this type of date validation to work. Global `isNaN` calls `valueOf` if the target is an object, which returns `NaN` if the date instance cannot provide a proper timestamp. 

`isNaN( new Date('invalid-date') )` outputs `true` as expected
`Number.isNaN( new Date('invalid-date') )` outputs `false` which is unexpected

If you're concerned about this, I can always explicitly do the same thing via `Number.isNaN( new Date('invalid-date').valueOf() )` but figured this was easier as the coercion is intended.

assets/js/googlesitekit/datastore/user/utils.js
adamsilverstein: > @adamsilverstein? As far as I can tell the current PHP logic may result in the date range including a few days that are also part of the other date range to compare with, which isn't something that I think would be intentional?

This is intentional and something I am pretty sure I saw in my testing, and what my internal ticket contacts told me to expect. It may not be expected (the whole aligning by day thing was surprising to me!), however I think this does match the current behavior of the Analytics dashboard.  If we are wrong it will show up in mismatched data (percent change) between Site Kit and Analytics.

The previous period range is adjusted to align with the last period range day of the week start, moving to the **closest** matching day. So if the last period starts on monday, the previous period should as well. if it starts on another day of the week, we shift to the closest monday - so if the day is tues, wed or thur, we move **back** to the previous monday - leaving a gap between the periods. if it is fri, sat or sunday, we move forward to the next monday - overlapping some days of the ranges. 

As it turns out, Site Kit's date ranges are all currently divisible by 7 - meaning the previous period will start on the same day - **except** the 90 day period which is off by one day. In this case, the logic will shift the previous period back one day, leaving a one day gap.

caseymorrisus: This is very helpful insight on my end, I will edit the JS logic to move dates forward when `weekDayAlign` is `true` (when applicable) as they currently only move back in dates (this will cause date range overlap, but that seems to be expected).

adamsilverstein: ps. Part of the reason the day alignment may make sense is that sites sometimes have distinct traffic patterns by day of the week (for example way more traffic/spikes on weekends). On a graph this will look like a series of humps, and when you display this against a previous period for comparison - you want those humps to line up in the graph!

assets/js/googlesitekit/datastore/user/date-range.js
caseymorrisus: @felixarntz The `- 1` is needed here because this is number used to determine the amount of days subtracted from the `endDate`. If we want a range that includes 2 days, we only subtract 1 day from the `endDate` to get the `startDate`. Number of days (subtracted from endDate) will always be 1 less than the range size. I could possibly rename `numberOfDays` to make this more clear but I'm unsure on the naming to do so. 

If it makes more sense I can add `- 1` to each use of `numberOfDays`, but will need to do that subtraction 3 times rather than once.

assets/js/googlesitekit/datastore/user/utils.js
aaemnnosttv: Relevant discussion where this came up in a previous issue:
- https://github.com/google/site-kit-wp/issues/1394#issuecomment-623342661
- https://github.com/google/site-kit-wp/issues/1394#issuecomment-637662727

As a result we created a `parseDimensionStringToDate` that uses the separate arguments for the `Date` constructor. Unfortunately that utility is specific to Analytics which provides the date as `YYYYMMDD`.

assets/js/googlesitekit/datastore/user/utils.js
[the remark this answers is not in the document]
aaemnnosttv: That makes sense now that I look more closely here, thanks for clarifying. In that case, let's add a comment to highlight that the use of this function is intentional and preferred.

assets/js/googlesitekit/datastore/user/utils.test.js
```diff
@@ -0,0 +1,202 @@
...
+
+import {
+	getDateString,
+	getPreviousDate,
+	getPreviousWeekDate,
+	isValidDateInstance,
+} from './utils';
+
+describe( 'core/user utils', () => {
+	const INVALID_DATE_INSTANCE_ERROR = 'Date param must construct to a valid date instance or be a valid date instance itself.';
```
aaemnnosttv, line 31: Rather than duplicating this for use in the test, let's export it from the source module.

assets/js/googlesitekit/datastore/user/utils.test.js
```diff
@@ -0,0 +1,202 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Internal dependencies
+ */
+
```
aaemnnosttv, line 22: Probably just an oversight but let's avoid adding blank lines between dependency group comments and imports 👍 
```suggestion
```

assets/js/googlesitekit/datastore/user/utils.test.js
```diff
@@ -0,0 +1,202 @@
...
+	isValidDateInstance,
+} from './utils';
+
+describe( 'core/user utils', () => {
+	const INVALID_DATE_INSTANCE_ERROR = 'Date param must construct to a valid date instance or be a valid date instance itself.';
+	describe( 'isValidDateInstance', () => {
+		it( 'should return true for valid date instance (no constructor params)', () => {
+			expect( isValidDateInstance( new Date() ) ).toEqual( true );
+		} );
+		it( 'should return true for valid date instance (valid constructor params)', () => {
```
aaemnnosttv, line 36: Please add blank lines between `it` calls for consistency with other tests and readability.
```suggestion
		} );

		it( 'should return true for valid date instance (valid constructor params)', () => {
```

assets/js/googlesitekit/datastore/user/utils.test.js
```diff
@@ -0,0 +1,202 @@
...
+		} );
+		it( 'should return false for Array', () => {
+			expect( isValidDateInstance( [] ) ).toEqual( false );
+		} );
+		it( 'should return false for Function', () => {
+			expect( isValidDateInstance( () => {} ) ).toEqual( false );
+		} );
+	} );
+
+	describe( 'getDateString ', () => {
```
aaemnnosttv, line 68: Trailing space in the name
```suggestion
	describe( 'getDateString', () => {
```

assets/js/googlesitekit/datastore/user/utils.test.js
```diff
@@ -0,0 +1,202 @@
...
+		const JANUARY = 0; // month is 0 indexed
+		const date = new Date( 2020, JANUARY, 1 );
+		const time = date.getTime();
+		const isoString = date.toISOString();
+		const gmtString = date.toGMTString();
+		const utcString = date.toUTCString();
+
+		it( `should throw error if no param is passed`, () => {
+			try {
+				getDateString( );
```
aaemnnosttv, line 78:
```suggestion
				getDateString();
```

aaemnnosttv (commented):
This is coming along well, just a few supporting comments and suggestions.

includes/Modules/Analytics.php
```diff
@@ -933,7 +933,13 @@ function ( $dimension_def ) {
 				$start_date  = $data['startDate'];
 				$end_date    = $data['endDate'];
 				if ( strtotime( $start_date ) && strtotime( $end_date ) ) {
-					$date_ranges[] = array( $start_date, $end_date );
+					$compare_start_date = $data['compareStartDate'];
+					$compare_end_date   = $data['compareEndDate'];
+					$date_ranges[]      = array( $start_date, $end_date );
+
+					if ( strtotime( $compare_start_date ) && strtotime( $compare_end_date ) ) {
+						$date_ranges = array_merge( array( $compare_start_date, $compare_end_date ), $date_ranges );
```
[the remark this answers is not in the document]
caseymorrisus: This has been modified to be in opposite order.

assets/js/googlesitekit/datastore/user/date-range.js
caseymorrisus: Return value has been modified as well as the related JSDoc comment.

assets/js/googlesitekit/datastore/user/date-range.test.js
caseymorrisus: `options.referenceDate` has been added & tests have been modified to use static dates.

assets/js/googlesitekit/datastore/user/utils.js
caseymorrisus: Added an additional error message via `INVALID_DATE_STRING_ERROR`.

assets/js/googlesitekit/datastore/user/utils.js
caseymorrisus: Removed this `invariant` as there's no possibility of any type errors within execution before getting to `getPreviousDate` (with a default set to `0` for `daysBefore`). `getPreviousDate` will throw  `INVALID_DATE_INSTANCE_ERROR` or `INVALID_DATE_STRING_ERROR` depending on the cause.

assets/js/googlesitekit/datastore/user/utils.js
caseymorrisus: @adamsilverstein Another helpful comment, I appreciate it! I went ahead and did a bit more research within the original issue (when this was added within PHP) as well as testing this functionality within the native Analytics UI.

@felixarntz Changes to modify the above logic to match native Analytics UI compare/day align logic has been pushed up.

assets/js/googlesitekit/datastore/user/date-range.js
caseymorrisus: @felixarntz I modified the logic here a little bit, it matches your expectations a bit closer on when to use `- 1`.

assets/js/googlesitekit/datastore/user/date-range.js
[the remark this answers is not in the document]
felixarntz: @caseymorrisus Sorry for the late reply, by default today should be included (i.e. `0`). The data should be as fresh as possible in theory, and the only reason the offset exists is to support the API implementation in practice which usually ranges between 1 or 2 days delayed. But using one of them as the default feels unintuitive to me.

includes/Modules/Analytics.php
```diff
@@ -933,7 +933,15 @@ function ( $dimension_def ) {
...
 				if ( strtotime( $start_date ) && strtotime( $end_date ) ) {
-					$date_ranges[] = array( $start_date, $end_date );
+					$compare_start_date = $data['compareStartDate'];
+					$compare_end_date   = $data['compareEndDate'];
+					$date_ranges[]      = array( $start_date, $end_date );
+
+					// When using multiple date ranges, it changes the structure of the response,
+					// where each date range becomes an item in a list.
+					if ( strtotime( $compare_start_date ) && strtotime( $compare_end_date ) ) {
+						array_push( $date_ranges, $compare_start_date, $compare_end_date );
```
felixarntz, line 943: This is not correct, the compare start and end dates should be the first and second element of its own array, and that array should then be the second item in `$date_ranges` (see code in the else clause below).

Also nit-picking, but let's not use `array_push( $date_ranges, ... )`, but rather `$date_ranges[] = ...`.
```suggestion
						$date_ranges[] = array( $compare_start_date, $compare_end_date );
```

felixarntz (changes requested):
Almost looking good, just a few minor comments, and one thing that needs to be fixed in PHP.

assets/js/googlesitekit/datastore/user/date-range.js
```diff
@@ -79,6 +84,49 @@ export const selectors = {
...
+	 */
+	getDateRangeDates( state, {
+		compare = false,
+		offsetDays = 1,
+		referenceDate = getDateString( new Date() ),
+		weekDayAlign = false,
+	} = {} ) {
+		const dateRange = selectors.getDateRange( state );
+		const endDate = getPreviousDate( referenceDate, offsetDays );
+		const matches = dateRange.match( '-(.*)-' );
```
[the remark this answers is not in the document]
caseymorrisus: Was fairly easy to add, did not cause any failing tests.

assets/js/googlesitekit/datastore/user/utils/get-previous-week-date.js
```diff
@@ -0,0 +1,45 @@
...
+ * `daysBefore` as a minimum.
+ *
+ * @since n.e.x.t
+ *
+ * @param {string} relativeDate Date string (YYYY-MM-DD) to subtract days from.
+ * @param {number} daysBefore   Number of days to subtract from relativeDate. Defaults to 0.
+ * @return {string}             The date string (YYYY-MM-DD) for the previous date, shifted to fall on the same day of the week.
+ */
+export const getPreviousWeekDate = ( relativeDate, daysBefore = 0 ) => {
+	const remainder = daysBefore % 7;
```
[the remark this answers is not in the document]
caseymorrisus: Makes sense to me, changed this to simplify logic.

assets/js/googlesitekit/datastore/user/utils/get-previous-week-date.test.js
```diff
@@ -0,0 +1,111 @@
...
+
+	it( 'should throw error if date supplied is invalid date', () => {
+		try {
+			getPreviousWeekDate( '2020-99-99', 1 );
+		} catch ( error ) {
+			expect( error.message ).toEqual( INVALID_DATE_STRING_ERROR );
+		}
+	} );
+
+	it( 'should go back to "2020-09-24" with daysBefore value of 0 and date of "2020-09-24"', () => {
```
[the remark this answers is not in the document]
caseymorrisus: Good call, definitely easier to parse (and less LOC) with this change.

felixarntz (approved):
Awesome! LGTM, just a few nit-picking things to potentially change.

caseymorrisus:
@felixarntz Latest requested changes have been pushed up.

felixarntz (approved):

Issue #2025, opened 2020-09-14
Do not show Tag Manager module in congratulations CTA notification


felixarntz:
The Congratulations notification for Site Kit includes a list of modules to activate (or whether they are already active). This list currently only includes modules that don't have any dependencies. Now that Tag Manager has been split out to no longer be dependent on Analytics, it shows up there too as expected - however it shouldn't as it's a fairly minor module that shouldn't be exposed here at this point.

![Screenshot 2020-09-14 at 13 42 57](https://user-images.githubusercontent.com/3531426/93136459-d3f51800-f690-11ea-8ecf-0a8ac1602dc5.png)

Let's simply modify the notification to only include information for the following modules:
* `search-console`
* `adsense`
* `analytics`
* `pagespeed-insights`

**PR for this should be opened against `feature/ga-gtm-separation`.**

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Only the following modules should be included in the "Congrats on completing..." notification:
    * `search-console`
    * `adsense`
    * `analytics`
    * `pagespeed-insights`

## Implementation Brief

* Introduce new optional prop `moduleSlugs` to `ModulesList` component. If provided, the modules to be rendered should be filtered against that list so that only those are included.
* In `DashboardSetupAlerts` component, pass the four slugs from the ACs to the `ModulesList` rendered.

## QA Brief

* Before this fix, after completing module setup for a module, the "Congrats on completing..." notification would include Tag Manager.
* After this fix, Tag Manager should no longer be included in that notification.

## Changelog entry

* N/A

aaemnnosttv:
@felixarntz this issue is related to https://github.com/google/site-kit-wp/issues/1148 which I raised a while back.

It's a bit different now of course that GTM will no longer depend on Analytics, but part of the idea there is that the CTA  is useful for showing other modules to activate (i.e. module discovery). I would suggest that we not _prevent_ Tag Manager from showing up there but implement a few rules to help better manage this area
- Only show a maximum of 4 modules to prevent it breaking to multiple rows on desktop
- Show inactive modules first and fill in the rest with active modules to always show 4, but prioritize inactive to streamline activating multiple
- Potentially use `module.order` as a secondary sort/prioritization
- Potentially always include the activated module to "reinforce" the messaging that it is now active/connected

[...]

PR #2089, opened 2020-09-22
Filter modules to be shown in notifications.

asvinb:
## Summary

Addresses issue #2025 

## Relevant technical choices
Adds a new moduleSlugs prop which is an array of module slugs which will be displayed in notifications. If moduleSlugs is undefined, all the modules will be displayed.

## Checklist

### DIFF
--- a/assets/js/components/modules-list.js
+++ b/assets/js/components/modules-list.js
@@ -81,12 +81,15 @@ class ModulesList extends Component {
 		const { moduleSlugs } = this.props;
 		const modulesData = getModulesData();
 
-		// Filter out internal modules and specific ones via the moduleSlugs prop.
-		const modules = Object.values( modulesData ).filter(
-			( module ) =>
-				! module.internal &&
-				( Array.isArray( moduleSlugs ) && moduleSlugs.length > 0 ? moduleSlugs.includes( module.slug ) : true )
-		);
+		// Filter specific modules
+		const moduleObjects = Array.isArray( moduleSlugs ) && moduleSlugs.length
+			? moduleSlugs
+				.filter( ( slug ) => modulesData.hasOwnProperty( slug ) )
+				.reduce( ( acc, slug ) => ( { ...acc, [ slug ]: modulesData[ slug ] } ), {} )
+			: modulesData;
+
+		// Filter out internal modules.
+		const modules = Object.values( moduleObjects ).filter( ( module ) => ! module.internal );
 
 		// Map of slug => name for every module that is active and completely set up.
 		const completedModuleNames = modules
--- a/assets/js/components/modules-list.js
+++ b/assets/js/components/modules-list.js
@@ -78,10 +78,15 @@ class ModulesList extends Component {
 	}
 
 	render() {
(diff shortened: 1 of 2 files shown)


aaemnnosttv:
@asvinb JS tests and lint jobs are failing, can you take a look?

assets/js/components/modules-list.js
```diff
@@ -78,10 +78,15 @@ class ModulesList extends Component {
...
+		const { moduleSlugs } = this.props;
 		const modulesData = getModulesData();
 
-		// Filter out internal modules.
-		const modules = Object.values( modulesData ).filter( ( module ) => ! module.internal );
+		// Filter out internal modules and specific ones via the moduleSlugs prop.
+		const modules = Object.values( modulesData ).filter(
+			( module ) =>
+				! module.internal &&
+				( Array.isArray( moduleSlugs ) && moduleSlugs.length > 0 ? moduleSlugs.includes( module.slug ) : true )
```
aaemnnosttv, line 88: This works, but is a bit much to have all on one line. It also checks the type an length of `moduleSlugs` on every iteration which is unnecessary. The `true` at the end of the ternary is also a bit confusing at first glance. Of course it doesn't make a huge difference with such a small list, but we can make it a bit easier to read by separating these things.

Rather than doing everything in the filter for every module, we can change how module objects are selected before filtering based on `moduleSlugs`:

```js
const moduleObjects = Array.isArray( moduleSlugs ) && moduleSlugs.length
    ? moduleSlugs
        .filter( ( slug ) => modulesData.hasOwnProperty( slug ) )
        .reduce( ( acc, slug ) => ( { ...acc, [ slug ]: modulesData[ slug ] } ), {} )
    : modulesData
```

This way `moduleObjects` come from `moduleSlugs` if it's passed as an array, otherwise it uses them all like before.

Then we only need to use `Object.values( moduleObjects )` instead in the existing line which would otherwise remain unchanged to filter out internal modules in all cases.

What do you think?

aaemnnosttv (commented):
Thanks @asvinb – One suggestion for you – let me know what you think?

assets/js/components/modules-list.js
asvinb: Makes sense @aaemnnosttv . I've updated the PR, can you take a look again?
Also I was thinking maybe we shouldn't check if the module is an internal one if moduleSlugs is not empty. The reason being if we have something in moduleSlugs, we intentionally want to display that module. So I was thinking whether we can do something like this:
`
if( Array.isArray( moduleSlugs ) && moduleSlugs.length ) {
    // Filter per slugs.
}
else {
   // Filter internal modules.
}
`

What do you think?

aaemnnosttv: > Also I was thinking maybe we shouldn't check if the module is an internal one if moduleSlugs is not empty

Good question – I had the same thought 😄  `internal` modules are those without any interface/settings (the only one right now is `site-verification` I think). So it wouldn't make sense to display an `internal` module in this context, but otherwise it's a very reasonable suggestion.

aaemnnosttv (approved):
👍

Issue #2033, opened 2020-09-15
Clear errors correctly when resetting


felixarntz:
## Bug Description

During the AdSense setup, reset logic is fired when the user returns to the screen after having been idle for more than 15 seconds. As part of the reset logic, a call is made to `receiveError( undefined )` (to clear the "global" error), which as of #1814 is no longer valid usage because that function now requires an error to be passed. See screenshot below:
![Screenshot 2020-09-15 at 12 11 27](https://user-images.githubusercontent.com/3531426/93254065-9e156980-f74c-11ea-8f71-b8a4f3879553.png)

As part of this issue, let's fix this invariant violation, but at the same time add a way to properly clear multiple errors at the same time via a `clearErrors( baseName )` - complementing similar selectors implemented via #1859. We should also update various `reset*` actions to include resetting related errors.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `createErrorStore` function should be expanded so that the returned store also includes a new action `clearErrors( baseName )`:
    * `baseName` should be an optional parameter.
    * If `baseName` is provided, all errors in the store that apply to `baseName` should be cleared.
    * If `baseName` is not provided, all errors in the entire store should be cleared (including the legacy `error` store property).
* The existing `clearError` should be modified to _also_ support the legacy `error` store property, similar to how `receiveError` does:
    * If `baseName` is provided, it should behave like it already does today.
    * If `baseName` is not provided, it should clear the `error` store property. A TODO annotation should be added here similar to the one in `receiveError` to remove this once all instances of the legacy behavior have been removed.
* The following store actions for resetting data should include calls to the new `clearErrors()` for the relevant `baseName` (selector name) used to originally obtain that data:
    * `modules/adsense` `resetAccounts`
    * `modules/adsense` `resetAlerts`
    * `modules/adsense` `resetClients`
    * `modules/adsense` `resetURLChannels`
    * `modules/analytics` `resetAccounts`
* The invalid `receiveError( undefined )` call in the AdSense `SetupMain` component should be replaced with a `clearError()` call which is now the "proper" way to clear the legacy error.

## Implementation Brief

* In `createErrorStore` store:
    * Modify `clearError` to clear the legacy `error` store property if no `baseName` is provided (including TODO annotation per the ACs).
    * Introduce `clearErrors` that either clears all store errors or all errors for a given `baseName`.
    * Add tests for the new behavior.
[...]

PR #2090, opened 2020-09-22
Bug/2033 clear errors with resetting

ryanwelcher:
## Summary

Addresses issue #2033 

## Checklist


assets/js/googlesitekit/data/create-error-store.js
```diff
@@ -92,12 +101,31 @@ export function createErrorStore() {
...
 				const { baseName, args } = payload;
-				const key = generateErrorKey( baseName, args );
+				const newState = { ...state };
+				if ( baseName ) {
+					const key = generateErrorKey( baseName, args );
+					newState.errors = { ...( state.errors || {} ) };
+					delete newState.errors[ key ];
+				} else {
+					// @TODO: remove it once all instances of the legacy behavior have been removed.
+					delete newState.error;
```
felixarntz, line 111: I don't think we should `delete` full state properties.
```suggestion
					newState.error = undefined;
```

assets/js/googlesitekit/data/create-error-store.js
```diff
@@ -92,12 +101,31 @@ export function createErrorStore() {
...
+			case CLEAR_ERRORS: {
+				const { baseName } = payload;
+				const newState = { ...state };
+				if ( baseName ) {
+					const key = generateErrorKey( baseName );
+					newState.errors = { ...( state.errors || {} ) };
+					delete newState.errors[ key ];
+				} else {
+					delete newState.errors;
+					delete newState.error;
```
felixarntz, line 126: See above.
```suggestion
					newState.errors = {};
					newState.error = undefined;
```

assets/js/googlesitekit/data/create-error-store.js
```diff
@@ -92,12 +101,31 @@ export function createErrorStore() {
...
+			}
 
-				return { ...state, errors };
+			case CLEAR_ERRORS: {
+				const { baseName } = payload;
+				const newState = { ...state };
+				if ( baseName ) {
+					const key = generateErrorKey( baseName );
+					newState.errors = { ...( state.errors || {} ) };
+					delete newState.errors[ key ];
```
felixarntz, line 123: This doesn't delete all errors for the `baseName`, it right now only deletes the error for `baseName` with no arguments. To actually delete all errors for `baseName`, we'll need to delete all entries in `newState.errors` where `key` is `baseName` or starts with `baseName`.

To make this bulletproof, we should probably update `generateErrorKey` to separate the `md5` bit with some distinct separator, e.g. `::`. This way we can check here as follows: `key === baseName || key.startsWith( baseName + '::' )`.

assets/js/modules/adsense/datastore/accounts.js
```diff
@@ -55,6 +55,8 @@ const baseActions = {
 			type: RESET_ACCOUNTS,
 		};
 
+		yield dispatch( STORE_NAME ).clearErrors( 'getAccounts' );
```
felixarntz, line 58: Based on the IB and https://github.com/google/site-kit-wp/issues/2033#issuecomment-693193545, these calls shouldn't use `dispatch`. This applies to the other changes below as well.

felixarntz (changes requested):
Only a few things to correct here. Let's wait until #1859 merge and then add extra tests for the behavior here.

assets/js/googlesitekit/data/create-error-store.js
```diff
@@ -92,12 +101,35 @@ export function createErrorStore() {
...
+			}
 
-				return { ...state, errors };
+			case CLEAR_ERRORS: {
+				const { baseName } = payload;
+				const newState = { ...state };
+				if ( baseName ) {
+					newState.errors = { ...( state.errors || {} ) };
+					for ( const key in Object.keys( newState.errors ) ) {
+						if ( key === baseName || key.startsWith( baseName ) ) {
```
felixarntz, line 123:
```suggestion
						if ( key === baseName || key.startsWith( `${ baseName }::` ) ) {
```

assets/js/googlesitekit/data/create-error-store.js
```diff
@@ -92,12 +101,35 @@ export function createErrorStore() {
...
+							delete newState.errors[ key ];
+						}
+					}
+				} else {
+					// @TODO: remove it once all instances of the legacy behavior have been removed.
```
felixarntz, line 128: This comment only applies to the `newState.error` assignment below, not to both assignments (the `errors` state property is the "correct" one to use while `error` is legacy).

felixarntz (changes requested):
Two small issues with this, but otherwise this is good to go!

felixarntz (approved):
Great!

Issue #2038, opened 2020-09-16
Scaffold new admin screen `googlesitekit-user-input`


felixarntz:
A new administration screen `googlesitekit-user-input` should be added, which should not be available in the admin menu but will be linked or redirected to under certain circumstances in the future.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* In PHP, a new Site Kit admin screen `googlesitekit-user-input` should be added.
    * It should not appear in the admin menu.
    * It should require capability `googlesitekit_manage_options` for access.
    * It should have title "User Input".
    * It should enqueue a new asset `googlesitekit-user-input` (see below).
    * It should render `<div id="js-googlesitekit-user-input" class="googlesitekit-page"></div>`, to be populated by JS.
* A new JS feature flag `userInput` should be added that is only enabled during `development`.
* A new JS asset with filename and handle `googlesitekit-user-input` should be added and registered in the PHP `Assets` class.
    * For now, this should just receive scaffolded code to initialize a separate React app similarly to e.g. in `googlesitekit-dashboard.js` (loading translations, wrapping in root etc.), and render a new component `UserInputApp` (to be put into `assets/js/components/user-input/UserInputApp.js`).
    * The `UserInputApp` component should for now be the following:
        * If `userInput` feature flag is active: A `div` containing an untranslatable string "TODO: Implement logic and UI.".
        * Otherwise: A `div` containing an untranslatable string like "Cheatin', huh?" 😄 

## Implementation Brief

* Add a new `Script` instance to the array returned by `Assets::get_assets`
    - `handle` of `googlesitekit-user-input`
    - `src` of `$base_url . 'js/googlesitekit-user-input.js'`
    - `dependencies` to use the default common `$dependencies`
* Add a new `Screen` instance to the array returned by `Screens::get_screens`
    - `slug` of `self::PREFIX . 'user-input'`
    - `args.title` of `User Input` (translatable)
    - `args.capability` of `Permissions::MANAGE_OPTIONS`
    - `args.enqueue_callback` that enqueues `googlesitekit-user-input`
    - `args.render_callback` should output the HTML to use for the React render target  
        ```html
        <div id="js-googlesitekit-user-input" class="googlesitekit-page"></div>
        ```
    - `args.parent_slug` can be omitted as this will be `null` by default and will not appear in the admin menu.  
[...]

PR #2102, opened 2020-09-24
Enhancement/2038 user input admin screen

ryanwelcher:
## Summary

Addresses issue #2038

## Checklist


.storybook/webpack.config.js
```diff
@@ -25,6 +31,13 @@ module.exports = async ( { config } ) => {
 	config.plugins = [
 		...config.plugins,
 		new MiniCssExtractPlugin(),
+		new FeatureFlagsPlugin(
+			flagsConfig,
+			{
+				modes: [ 'development', 'production' ],
+				mode: config.mode,
+			},
+		),
```
felixarntz, line 40: Just to clarify, why is this needed in the Storybook Webpack?

assets/js/components/user-input/UserInputApp.js
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export default function UserInputApp() {
+	return featureFlags.widgets.userInput.enabled ? ( <div>TODO: Implement logic and UI.</div> ) : ( <div>You need a higher level of permission.</div> );
```
felixarntz, line 20: Just to make this more readable and prepare for future expansion, it would be better to break this into multiple lines. I think we should have an if clause `if ( ! featureFlags.widgets.userInput.enabled )`, and in that return `__( 'Something went wrong.', 'google-site-kit' )`. This part will be removed once the feature goes to production. Then the TODO div can be the default return value.

includes/Core/Admin/Screens.php
```diff
@@ -383,6 +383,25 @@ private function get_screens() {
 			)
 		);
 
+		$screens[] = new Screen(
+			self::PREFIX . 'user-input',
+			array(
+				'title'            => __( 'User Input', 'google-site-kit' ),
+				'capability'       => Permissions::MANAGE_OPTIONS,
```
felixarntz, line 390: This also needs to pass the following so that the screen is not in the admin menu (per the ACs):
```suggestion
				'capability'       => Permissions::MANAGE_OPTIONS,
				'parent_slug'      => null,
```

stories/user-input.stories.js
```diff
@@ -0,0 +1,14 @@
...
+import { storiesOf } from '@storybook/react';
+
+/**
+ * Internal dependencies
+ */
+import UserInputApp from '../assets/js/components/user-input/UserInputApp';
+
+storiesOf( 'User Input', module )
+	.add( 'User Input Page', () => {
+		return <UserInputApp />;
```
felixarntz, line 13: While this component doesn't use the datastore yet since it's just scaffolded, let's already wrap this in a `WithDataRegistry` component, to which we can pass a default registry created via `createTestRegistry()`.

felixarntz (commented):
This is pretty much good to be merged, just a few nit-pick suggestions and one question.

felixarntz:
@ryanwelcher Can you also add a QA Brief for this one?

.storybook/webpack.config.js
ryanwelcher: @felixarntz this was added to allow access to the featureFlags. I have removed it and instead am setting the flag in the story.

includes/Core/Admin/Screens.php
ryanwelcher: @felixarntz my mistake here, the IB says:
> args.parent_slug can be omitted as this will be null by default and will not appear in the admin menu.

I took omitted to mean "not added to the array" -  I have addressed this.

felixarntz: @ryanwelcher Oh good point - I must have missed that when reviewing the IB.

stories/user-input.stories.js
ryanwelcher: @felixarntz I couldn't find any references to `WithDataRegistry`, so I am assuming that you meant `withTestRegistry` 😄

felixarntz: Oops, exactly 😆

felixarntz (approved):
Great!

Issue #1865, opened 2020-08-03
Move visual regression testing to GitHub Actions


aaemnnosttv:
## Feature Description

As a stepping stone towards moving our CI jobs to GitHub Actions (see #1779), a good first candidate to move is the VRT job. One benefit of moving this job in particular is that we can upload the generated HTML report that backstop generates as an artifact of the job for downloading/inspection later in the event of a failure where this is not readily available with Travis (possible but requires you bring your own storage and authentication).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The "Visual Regression" test job is removed from `.travis.yml`
* All references to the `$VRT` environment variable are removed from `.travis.yml`
* VRT should run on GitHub Actions as a check for all open PRs with a base branch of `develop` or `master` branches and must pass for a PR to be merged
* The generated VRT report HTML should be saved as an artifact if the job fails

#### Travis Config
* Remove job in Travis https://github.com/google/site-kit-wp/blob/6eb4c52e87d6b71f1d82ca15caeef5fe6c7d1801/.travis.yml#L115-L116
* Remove condition in `before_install` https://github.com/google/site-kit-wp/blob/6eb4c52e87d6b71f1d82ca15caeef5fe6c7d1801/.travis.yml#L41
* Remove conditional in `before_script` block https://github.com/google/site-kit-wp/blob/6eb4c52e87d6b71f1d82ca15caeef5fe6c7d1801/.travis.yml#L50-L52
* Remove conditional in `script` block https://github.com/google/site-kit-wp/blob/6eb4c52e87d6b71f1d82ca15caeef5fe6c7d1801/.travis.yml#L80-L82

#### GitHub Actions
* Create a new `.github/workflows/tests.yml` workflow file  
_See `zips.yml` for reference as to when the workflow should run (although we only need `pull_request`)_
* Create a `visual-regression` job with steps to run  
    _Most of this can be borrowed from [`storybook.yml`](https://github.com/google/site-kit-wp/blob/develop/.github/workflows/storybook.yml) since VRT starts Storybook as the target to use when it runs_
    - Checkout the repo
    - Install composer dependencies (use the same command as we had on Travis, with updated path for where the cache is located locally - although it should be the same)  
    ```
    docker run --rm -v "$PWD:/app" -v "$HOME/.cache/composer:/tmp/cache" composer install --no-scripts
    ```  
[...]

PR #2107, opened 2020-09-28
Enhancement/1865 vrt to GitHub actions

eugene-manuilov:
## Summary

Addresses issue #1865 

## Relevant technical choices

- Put VRT into its own workflow file instead of creating a file for all tests to be able to re-run VRT tests on failure.
- Managed to rework VRT not to rely on `johnpbloch/wordpress-core` dependency by importing required CSS files from the WordPress repository.
- Had to upgrade `backstopjs` to the latest version and as a result, to approve visual changes for ~42 screens (all of them just minor changes in text rendering caused by different versions of chrome).

## Checklist


.storybook/assets/sass/wp-admin.scss
```diff
@@ -0,0 +1,4 @@
+@import 'https://raw.githubusercontent.com/WordPress/WordPress/master/wp-admin/css/common.css';
+@import 'https://raw.githubusercontent.com/WordPress/WordPress/master/wp-admin/css/dashboard.css';
+@import 'https://raw.githubusercontent.com/WordPress/WordPress/master/wp-admin/css/edit.css';
+@import 'https://raw.githubusercontent.com/WordPress/WordPress/master/wp-admin/css/forms.css';
```
aaemnnosttv, line 4: This certainly feels nicer than installing all of WP via Composer just for VRT, but I don't really like the idea of using the latest styles for every build as it reduces the reproducibility/stability of these tests. Given the very strict nature of the comparisons, I think we should have a fixed version for use with our reference images. We could easily update these URLs to use a specific version instead, but then we would also be handling a dependency version outside of npm/composer. Is there a way we can use a sass variable for the WP version here and set that via config in `package.json`? We would probably need to tweak the webpack config that Storybook uses to provide that, but it sounds possible?

package.json
```diff
@@ -154,7 +154,7 @@
     "amphtml-validator": "^1.0.31",
     "babel-loader": "^8.1.0",
     "babel-plugin-lodash": "^3.3.4",
-    "backstopjs": "3.8.8",
+    "backstopjs": "^5.0.6",
```
aaemnnosttv, line 157: This is nice (with the exception of all of the reference images needing to be updated 😄 ) – can you elaborate on why this version change was necessary?

.storybook/config.js
```diff
@@ -1,47 +1,29 @@
...
-	removeFilter,
-	addAction,
-	doAction,
-	applyFilters,
-	removeAction,
-	removeAllFilters,
-} from '@wordpress/hooks';
+import { getQueryString, addQueryArgs } from '@wordpress/url';
+import { addFilter, removeFilter, addAction, doAction, applyFilters, removeAction, removeAllFilters } from '@wordpress/hooks';
+
```
aaemnnosttv, line 19: It looks like the import format was only changed from one-per-line to all on one line here. Can we keep it as it was so it's easier to see what changed? It's also more consistent with how we format multiple imports in other components.

aaemnnosttv (changes requested):
Looks good @eugene-manuilov! There are a few things to update, but mostly small things. I left a few questions for you as well.

Regarding the actual VRT run in GH Actions, it's adding a huge number of lines to the job log from the webpack build where  this was not the case on Travis:

https://github.com/google/site-kit-wp/runs/1177468808?check_suite_focus=true#step:7:5270
https://travis-ci.com/github/google/site-kit-wp/jobs/390737195#L528-L543

Can you take a look and see if we can preserve that? It would also be nice if we could keep the nicer colored logs like we had on Travis, but not required.

tests/backstop/config.js
```diff
@@ -32,6 +32,9 @@ module.exports = {
 	asyncCompareLimit: 50,
 	debug: false,
 	debugWindow: false,
+	// Use a custom command template to make sure it works correctly in the GitHub actions environment.
+	// The only difference between the original dockerCommandTemplate and this one is that there is no -t flag in the currect template.
```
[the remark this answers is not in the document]
eugene-manuilov: @aaemnnosttv take a look at the backstopjs [documentation](https://github.com/garris/BackstopJS#using-docker-for-testing-across-different-environments):

> Tip: to run BackstopJS in Docker in an environment where the output is piped (e.g. CI server or an IDE's output window), remove the -t parameter (change the default to "docker run --rm -i --mount...)

If we don't remove `-t`, then the job will fail with exit code `1` because it doesn't support TTY and can't allocate a pseudo-tty. It also answers your question of why we don't have colors.

package.json
eugene-manuilov: `3.8.8` doesn't have `dockerCommandTemplate` option which is important for us.

.storybook/config.js
eugene-manuilov: > It's also more consistent with how we format multiple imports in other components.

We don't have consistency even in the original version of this file (compare imports from `@wordpress/i18n` and other WP dependencies). There is a nice eslint rule which can help us to make it consistent: `object-curly-newline`. We can use it like to restrict how many properties we allow to import using a single-line approach and when to start using multiline. For example, something like this:

```
"object-curly-newline": [ "error", { "ImportDeclaration": "always", "minProperties": 4 } ],
```
> Can we keep it as it was so it's easier to see what changed?

I'll revert.

.storybook/assets/sass/wp-admin.scss
aaemnnosttv: I just found this https://github.com/webpack-contrib/sass-loader#additionaldata which looks like it would do the trick.

What do you think about adding something like a `config.storybook_wordpress_version` key to `package.json` and then referencing this as a `process.env.npm_package_config_storybook_wordpress_version` for the above sass variable?

eugene-manuilov: Yes, it's a solid plan. I have implemented it.

.github/workflows/visual-regression.yml
```diff
@@ -0,0 +1,44 @@
+name: Visual Regresion Testing
+
+on:
+  pull_request:
+    types:
```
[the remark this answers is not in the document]
eugene-manuilov: Added.

.storybook/config.js
eugene-manuilov: Updated.

tests/backstop/config.js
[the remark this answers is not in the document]
aaemnnosttv: Thanks for clarifying. I wonder what the difference is with Travis since they both ran it in Docker? 🤔

aaemnnosttv:
Thanks @eugene-manuilov, this is looking really nice.

Two things remain:
- There's a merge conflict to resolve with the `package-lock.json`
- Can you trigger a failing build for VRT so we can verify that the logic and behavior there works as expected? I think simply reverting a few/all of the reference images that failed would do the trick, and then once verified we can revert that change back to the current passing/approved versions. This is somewhat of a QA type of task, but since this affects our builds and will go into effect as soon as it's merged, it's a bit more important that we make sure everything is working as expected beforehand.

eugene-manuilov:
@aaemnnosttv

> There's a merge conflict to resolve with the package-lock.json

The merge conflict is solved.

> Can you trigger a failing build for VRT so we can verify that the logic and behavior there works as expected?

See here: https://github.com/google/site-kit-wp/actions/runs/276741491 - this is a failed run before approving changed images. Is it enough for you?

aaemnnosttv:
> See here: https://github.com/google/site-kit-wp/actions/runs/276741491 - this is a failed run before approving changed images. Is it enough for you?

@eugene-manuilov yes, this is great, thanks!

It looks like there are a few lint issues to fix which is causing the Lint, JS, and E2E jobs to fail. Once fixed I think this will be good to go 👍 

---

Looks like you're working on this right now, just one remaining eslint error to fix:

```
google-site-kit/postcss.config.js
  20:7  error  `importUrl` violates naming rules  sitekit/camelcase-acronyms
  26:2  error  `importUrl` violates naming rules  sitekit/camelcase-acronyms

✖ 2 problems (2 errors, 0 warnings)
```

eugene-manuilov:
@aaemnnosttv ok, should be good now. I have reverted package-lock.json changes and re-installed dependencies to regenerate lock file. Now everything seems fine.

aaemnnosttv:
@googlebot I consent.

aaemnnosttv:
Was able to fix the crazy verbose output when storybook was building by passing a few different flags to an alternate command to use when running VRT:

image

The relevant options to `start-storybook` and their descriptions:
```
  --ci                          CI mode (skip interactive prompts, don't open browser)
  --quiet                       Suppress verbose build output
  --no-version-updates          Suppress update check
```

This reduced the size of the job log by **over 5.7k lines** 😄 

It will also cause the browser window to no longer be opened when starting VRT locally, which can be annoying sometimes. These changes only affect VRT, so using `npm run storybook` will still work as usual (it just won't bug you about new versions of Storybook now 😄 ).

aaemnnosttv (approved):
🎉

Issue #1985, opened 2020-09-03
Update proxy permissions URL to pass through backend


aaemnnosttv:
## Feature Description

In #1857, we changed the way the proxy setup URL works to be similar to the `connect` and `disconnect` URLs by using a WP "admin action" URL. This allows us to do additional permission and capability checks before sending the user to the proxy. It also exposes less sensitive information to the client than is necessary.

Since the above issue is already very large, it makes sense to split out this portion to update the permissions URL to function in a similar way as this URL also contains some of the same information such as proxy credentials.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a new method `Authentication::get_proxy_permissions_url()` which returns a URL similar to `Authentication::get_proxy_setup_url()` but for a different action `Google_Proxy::ACTION_PERMISSIONS` (and corresponding nonce).
* There should be a callback hooked into that new admin action which checks that the current user can `authenticate`, whether the proxy is in use, and only then it should redirect to the real proxy permissions URL (`OAuth_Client::get_proxy_permissions_url()`).
* Any links in Site Kit UI should be updated to point to the new "proxied" proxy permissions URL, and the real proxy permissions URL should not be exposed in any client-side code.
* There should be a new `admin_action_googlesitekit_proxy_permissions` action handler which when called, qualifies the current user before redirecting them to the `OAuth_Client::get_proxy_permissions_url`
* The `proxyPermissionsURL` provided to the client should be in the form:
`{admin_url}?action=googlesitekit_proxy_permissions&nonce={nonce}`

## Implementation Brief

* Define a new `Google_Proxy::ACTION_PERMISSIONS` class constant for the `googlesitekit_proxy_permissions` literal
* Add a new `Authentication::get_proxy_permissions_url` method, essentially returning:  
    ```php
    return add_query_args(
        array(
            'action' => Google_Proxy::ACTION_PERMISSIONS,
            'nonce'  => wp_create_nonce( Google_Proxy::ACTION_PERMISSIONS )
        ),
        admin_url( 'index.php' )
    );
    ```
- Update `proxyPermissionsURL` in `Authentication::inline_js_base_data` to use the new `get_proxy_permissions_url` instead so that the client always sends users through the WP backend to get there.
* In `Authentication::register`, add a new action on **`'admin_action_' . Google_Proxy::ACTION_PERMISSIONS`** which calls a new private method `handle_proxy_permissions`:
[...]

PR #2109, opened 2020-09-28
Enhancement/1985 update proxy permissions url

ryanwelcher:
## Summary

Addresses issue #1985 

## Checklist


includes/Core/Authentication/Authentication.php
```diff
@@ -443,6 +451,47 @@ public function get_disconnect_url() {
 		);
 	}
 
+	/**
+	 * Gets the proxy permission url
```
felixarntz, line 455: _All_ documentation should end in a period. (unless it already has exclamation mark, question mark etc of course). Also, let's use correct capitalization of abbreviations.

includes/Core/Authentication/Authentication.php
```diff
@@ -443,6 +451,47 @@ public function get_disconnect_url() {
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string Proxy permission url.
```
felixarntz, line 459: See above.

includes/Core/Authentication/Authentication.php
```diff
@@ -443,6 +451,47 @@ public function get_disconnect_url() {
...
+
+		$nonce = $this->context->input()->filter( INPUT_GET, 'nonce' );
+		if ( ! wp_verify_nonce( $nonce, Google_Proxy::ACTION_PERMISSIONS ) ) {
+			wp_die( esc_html__( 'An error has occurred. Please reload the page and try again.', 'google-site-kit' ) );
+		}
+
+		wp_safe_redirect( $this->get_oauth_client()->get_proxy_permissions_url() );
+		exit;
+
+	}
```
felixarntz, line 493: Rather than putting these two methods at an arbitrary point in the class, let's add them below where the related `handle_sync_site_fields` and `get_proxy_setup_url` methods are.

includes/Core/Authentication/Authentication.php
```diff
@@ -443,6 +451,47 @@ public function get_disconnect_url() {
...
+	 *
+	 * @since n.e.x.t
+	 */
+	private function handle_proxy_permissions() {
+		if ( ! current_user_can( Permissions::AUTHENTICATE ) ) {
+			wp_die( esc_html__( 'You need a higher level of permission.', 'google-site-kit' ) );
+		}
+
+		if ( ! $this->credentials->using_proxy() ) {
+			wp_die( esc_html__( 'You need a higher level of permission.', 'google-site-kit' ) );
```
felixarntz, line 482: This is not an accurate message, there is no permission issue here. There is already a similar message for the same condition in the `handle_sync_site_fields` method which we can reuse: _Site Kit is not configured to use the authentication proxy._

includes/Core/Authentication/Authentication.php
```diff
@@ -443,6 +451,47 @@ public function get_disconnect_url() {
...
+	}
+
+	/**
+	 * Handles proxy permissions
+	 *
+	 * @since n.e.x.t
+	 */
+	private function handle_proxy_permissions() {
+		if ( ! current_user_can( Permissions::AUTHENTICATE ) ) {
+			wp_die( esc_html__( 'You need a higher level of permission.', 'google-site-kit' ) );
```
felixarntz, line 478:
```suggestion
			wp_die( esc_html__( 'You have insufficient permissions to manage Site Kit permissions.', 'google-site-kit' ) );
```

includes/Core/Authentication/Authentication.php
```diff
@@ -443,6 +451,47 @@ public function get_disconnect_url() {
...
+		}
+
+		if ( ! $this->credentials->using_proxy() ) {
+			wp_die( esc_html__( 'You need a higher level of permission.', 'google-site-kit' ) );
+		}
+
+		$nonce = $this->context->input()->filter( INPUT_GET, 'nonce' );
+		if ( ! wp_verify_nonce( $nonce, Google_Proxy::ACTION_PERMISSIONS ) ) {
+			wp_die( esc_html__( 'An error has occurred. Please reload the page and try again.', 'google-site-kit' ) );
```
felixarntz, line 487: This is not helpful, it should say what's wrong. We already have this in several places and it should be the same here: _Invalid nonce._

includes/Core/Authentication/Authentication.php
```diff
@@ -443,6 +451,47 @@ public function get_disconnect_url() {
...
+		if ( ! current_user_can( Permissions::AUTHENTICATE ) ) {
+			wp_die( esc_html__( 'You need a higher level of permission.', 'google-site-kit' ) );
+		}
+
+		if ( ! $this->credentials->using_proxy() ) {
+			wp_die( esc_html__( 'You need a higher level of permission.', 'google-site-kit' ) );
+		}
+
+		$nonce = $this->context->input()->filter( INPUT_GET, 'nonce' );
+		if ( ! wp_verify_nonce( $nonce, Google_Proxy::ACTION_PERMISSIONS ) ) {
```
felixarntz, line 486: Can we move the nonce check up to be the first check? That's the most basic one - if the nonce is invalid, clearly the request is not authorized. This is also how the setup handler works.

includes/Core/Authentication/Authentication.php
```diff
@@ -25,6 +25,7 @@
 use WP_REST_Request;
 use WP_REST_Response;
 use Exception;
+use OAuth;
```
felixarntz, line 28: Why was this added here? Not a valid class I believe, and it's unused.

felixarntz (changes requested):
Mostly LGTM, just a few minor things to iterate on, mostly around messages.

includes/Core/Authentication/Authentication.php
```diff
@@ -443,6 +450,23 @@ public function get_disconnect_url() {
...
+	 */
+	private function get_proxy_permissions_url() {
+		return add_query_arg(
+			array(
+				'action' => Google_Proxy::ACTION_PERMISSIONS,
+				'nonce'  => wp_create_nonce( Google_Proxy::ACTION_PERMISSIONS ),
+			),
+			admin_url( 'index.php' )
+		);
+	}
```
felixarntz, line 468: See my previous review, this method should also be moved to the end of the class where the similar proxy setup methods are located.

felixarntz (commented):
Just one (non-critical) thing that was missed from the previous review. I'll quickly make that update.

felixarntz (approved):

Issue #2111, opened 2020-09-29
AdSense body snippet code is being stripped from AMP singular content types when Yoast plugin is active


jamesozzie:
## Bug Description

For AMP users when using Site Kit to place the AdSense snippet and with Yoast active the body snippet is being removed from the markup for singular content types (pages and posts), resulting in no ads being served. 

This occurs with the latest version of Site Kit, the AMP plugin and Yoast. Initially reported in the [AMP support forums ](https://wordpress.org/support/topic/adsense-ads-and-embedded-tweets-are-a-no-show-after-installing-amp/page/2/#post-13468341) and just today a similar topic arose today in the [Site Kit forums](https://wordpress.org/support/topic/google-site-kit-not-serving-ads-on-amp/#post-13467602). 

Reproducible in support, see below:

![issue](https://user-images.githubusercontent.com/41326532/94557104-03418280-0256-11eb-9a78-4f656a38f737.gif)

Reports (Site Health info visible in trix):
- [AMP Support Topic](https://wordpress.org/support/topic/adsense-ads-and-embedded-tweets-are-a-no-show-after-installing-amp/page/2/#post-13468341) | http://www.thickaccent.com | [SH info](https://docs.google.com/forms/d/1JTQmphDlk9YgHpDoFRbHXpkDqv8-QmzTT7MDR4AstyQ/edit#response=ACYDBNiQRVvRfneeqvIbqHdWEgcxIxFH4vmJJh1cQWsNCY6TW6pptiENR0OXiPrm5ZI0dH4)
- [Site Kit Support Topic](https://wordpress.org/support/topic/google-site-kit-not-serving-ads-on-amp/#post-13467602) | https://nerdwork.space/ | [Site Health info](https://docs.google.com/forms/d/1lhjns2K7c0Ny7ME4gxTcsnYtNUtcMr8BvFpPQi23Y1E/edit?gxids=7628#response=ACYDBNjtLzRAWw9MTJLueNxvtgUhV2dd2VrRL6eg2L-vl4zhSLT_0UcTtTzYI3fFIhoCTU0)

A possible similar topic arrived in the [AMP forums](https://wordpress.org/support/topic/amp-adsense-auto-ads-not-working-on-mobile/#post-13468354), with the code being stripped out **on mobile**. Awaiting additional information but looks similar (same theme others reported, Site Kit placed AdSense snippet and AMP plugin user with Yoast indicators)

## Steps to reproduce
[...]

PR #2114, opened 2020-09-29
AdSense: only add AMP auto-ads via main loop `the_content` filter.

adamsilverstein:
## Summary

Addresses issue https://github.com/google/site-kit-wp/issues/2111

## Checklist


adamsilverstein:
Note: ideally we should have tests that verify the correct tag is output in each of the three AMP modes, for the latest of each major version of the AMP plugin (1.x.x and 2.x.x).

ryanwelcher:
@googlebot I consent.

tests/e2e/matchers/to-have-amp-auto-ads-tag.js
```diff
@@ -0,0 +1,32 @@
+/**
```
aaemnnosttv, line 1: Please add a file header 🙂

tests/e2e/plugins/apply-content-fliters.php
```diff
@@ -0,0 +1,21 @@
...
+ * Author URI:  https://opensource.google.com
+ *
+ * @package   Google\Site_Kit
+ * @copyright 2020 Google LLC
+ * @license   https://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
+ * @link      https://sitekit.withgoogle.com
+ */
+
+
+// Applying the filters outside of the loop breaks the AdSense snippet with AMP enabled.
```
aaemnnosttv, line 15: This comment is no longer accurate as of this PR, correct?
I think it would be good to reword this to be specific to this being a previous bug.
It would also be good to include a link to the issue here as well for additional context.

tests/e2e/specs/modules/adsense/setup-new-user.test.js
```diff
@@ -345,4 +346,59 @@ describe( 'setting up the AdSense module', () => {
...
+				} );
+			};
+		} );
+
+		afterAll( async () => {
+			await deactivatePlugin( 'amp' );
+			await deactivatePlugin( 'e2e-tests-apply-content-filters' );
+		} );
+
+		it( 'it output in primary mode', async () => {
```
aaemnnosttv, line 393:
```suggestion
		it( 'is output in primary mode', async () => {
```

tests/e2e/specs/modules/adsense/setup-new-user.test.js
```diff
@@ -345,4 +346,59 @@ describe( 'setting up the AdSense module', () => {
...
+			await setAMPMode( 'primary' );
+			await proceedToAdsenseSetup();
+			await expect( '/hello-world' ).toHaveAMPAutoAdsTag();
+		} );
+		it( 'it output in secondary mode', async () => {
```
aaemnnosttv, line 398:
```suggestion
		it( 'is output in secondary mode', async () => {
```

tests/e2e/specs/modules/adsense/setup-new-user.test.js
```diff
@@ -345,4 +346,59 @@ describe( 'setting up the AdSense module', () => {
...
+		afterAll( async () => {
+			await deactivatePlugin( 'amp' );
+			await deactivatePlugin( 'e2e-tests-apply-content-filters' );
+		} );
+
+		it( 'it output in primary mode', async () => {
+			await setAMPMode( 'primary' );
+			await proceedToAdsenseSetup();
+			await expect( '/hello-world' ).toHaveAMPAutoAdsTag();
+		} );
```
aaemnnosttv, line 397:
```suggestion
		} );

```

tests/e2e/specs/modules/adsense/setup-new-user.test.js
```diff
@@ -345,4 +346,59 @@ describe( 'setting up the AdSense module', () => {
...
 	} );
+
+	describe( 'AMP <amp-auto-ads> tag', () => {
+		beforeAll( async () => {
+			await activatePlugin( 'amp' );
+			await activatePlugin( 'e2e-tests-apply-content-filters' );
+		} );
+
+		beforeEach( () => {
+			datapointHandlers.accounts = ( request ) => {
```
aaemnnosttv, line 357: [Per @felixarntz 's comment](https://github.com/google/site-kit-wp/issues/2111#issuecomment-707305831) it is a bit inefficient to go through the AdSense setup for each AMP mode just to check the tag output.

How about we do the following:
* add a `setupAdSense` utility, similar to `setupAnalytics` in `tests/e2e/utils/setup-analytics.js` since these tests aren't concerned with the module setup at all
* Move this test to a new `amp-auto-ads-tag.test.js` file in the same directory
* Use the new utility to setup AdSense in the `beforeAll` instead of going through setup for each test

tests/phpunit/integration/Modules/AdSenseTest.php
```diff
@@ -187,6 +187,7 @@ public function test_block_on_consent_amp( $enabled ) {
 
 	/**
 	 * @dataProvider block_on_consent_provider
+	 * @group ryan
```
aaemnnosttv, line 190:
```suggestion
```

tests/phpunit/integration/Modules/AdSenseTest.php
```diff
@@ -204,6 +205,10 @@ public function test_block_on_consent_amp_content( $enabled ) {
 			add_filter( 'googlesitekit_adsense_tag_amp_block_on_consent', '__return_true' );
 		}
 
+		// We need to fake the global to allow the hook to add the tag.
+		global $wp_query;
+		$wp_query->in_the_loop = true;
```
aaemnnosttv, line 210: While this is necessary to make this test pass, let's add another test for this specifically to test that the tag is also not added when not in the loop.

bin/local-env/install-wordpress.sh
```diff
@@ -113,6 +113,7 @@ else
 	wp plugin update amp --minor --quiet
 fi
 
+
```
aaemnnosttv, line 116:
```suggestion
```

includes/Modules/AdSense.php
```diff
@@ -132,6 +132,10 @@ function() use ( $client_id ) {
 					add_filter( // For AMP Reader, and AMP Native and Transitional (as fallback).
 						'the_content',
 						function( $content ) use ( $client_id ) {
+							// Only run for the primary application the `the_content` filter.
```
aaemnnosttv, line 135:
```suggestion
							// Only run for the primary application of the `the_content` filter.
```

aaemnnosttv (changes requested):
Thanks @ryanwelcher – this is looking pretty good, just a few changes to make and minor things to clean up. The main thing being a change to the way the E2E tests run to be a bit more efficient.

tests/phpunit/integration/Modules/AdSenseTest.php
ryanwelcher: 🤦

tests/e2e/matchers/to-have-amp-auto-ads-tag.js
```diff
@@ -0,0 +1,50 @@
...
+ * @since n.e.x.t
+ *
+ * @param {string} path The URL path of the current site to check.
+ * @return {Object} Matcher result.
+ */
+export async function toHaveAMPAutoAdsTag( path ) {
+	const result = {};
+
+	const page = await browser.newPage();
+	await page.goto( createURL( path, 'amp' ) );
```
aaemnnosttv, line 36: Since the `amp` query param only applies to secondary AMP modes (traditional at least), let's pass that as part of the `path` when calling rather than use this version of the URL unconditionally.

That way we could also test that `expect( pathWithoutAMP ).not.toHaveAMPAutoAdsTag()` when AMP is active and configured to assert that the non-AMP version/URL of the page doesn't render it.

tests/e2e/plugins/apply-content-filters.php
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @package   Google\Site_Kit
+ * @copyright 2020 Google LLC
+ * @license   https://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
+ * @link      https://sitekit.withgoogle.com
+ */
+
+
+/**
+ * Applying the filters outside of the loop causes the <amp-auto-ads> not be rendered on the single template
```
aaemnnosttv, line 16:
```suggestion
 * Applying the `the_content` filter outside of the loop used to cause the <amp-auto-ads> not be rendered on the single template.
```

tests/phpunit/integration/Modules/AdSenseTest.php
aaemnnosttv: This test is yet to be added it seems.

tests/e2e/specs/modules/adsense/amp-auto-ads-tag.test.js
```diff
@@ -0,0 +1,55 @@
...
+
+	it( 'is output in primary mode', async () => {
+		await setAMPMode( 'primary' );
+		await expect( '/hello-world' ).toHaveAMPAutoAdsTag();
+	} );
+
+	it( 'is output in secondary mode', async () => {
+		await setAMPMode( 'secondary' );
+		await expect( '/hello-world' ).toHaveAMPAutoAdsTag();
+	} );
```
aaemnnosttv, line 54: As previously mentioned, let's also test here that the auto ads tag is _not_ output on the web version (not `?amp`) when in secondary AMP mode. 

For some reason it won't let me suggest on this file, but how about this for clarity?
* `is output in secondary mode in AMP context` _(existing test)_
* `is not output in secondary mode in web context` _(test to add)_

tests/e2e/specs/modules/adsense/amp-auto-ads-tag.test.js
```diff
@@ -0,0 +1,55 @@
...
+ * Internal dependencies
+ */
+import {
+	setAMPMode,
+	setupAdSense,
+} from '../../../utils';
+
+describe( 'AMP <amp-auto-ads> tag', () => {
+	beforeAll( async () => {
+		await setupAdSense();
```
aaemnnosttv, line 37: Looks great! Sooo much cleaner! 😄

aaemnnosttv (commented):
Looking really good, just a few small things left to address and then this should be all set!

tests/e2e/matchers/to-have-amp-auto-ads-tag.js
ryanwelcher: @aaemnnosttv I originally had this in place however, the `?amp` was getting escaped so i went this route. I'll add a second optional boolean param that will add `amp` if `true`

tests/phpunit/integration/Modules/AdSenseTest.php
```diff
@@ -225,6 +229,45 @@ public function block_on_consent_provider() {
...
+
+		// Confirm that the tag is not added if we're not in the loop.
+		$output = apply_filters( 'the_content', 'test content' );
+		$this->assertNotContains( 'data-ad-client="ca-pub-12345678"', $output );
+
+		// We need to fake the global to allow the hook to add the tag.
+		global $wp_query;
+		$wp_query->in_the_loop = true;
+
+		// Confirm that the tag is added if we're not in the loop.
```
aaemnnosttv, line 255: This comment is a bit off, as it says we're still not in the loop.
```suggestion
		// Confirm that the tag is added when in the loop.
```

tests/phpunit/integration/Modules/AdSenseTest.php
```diff
@@ -225,6 +229,45 @@ public function block_on_consent_provider() {
...
+		$output = apply_filters( 'the_content', 'test content' );
+		$this->assertContains( 'data-ad-client="ca-pub-12345678"', $output );
+	}
+
+	public function amp_auto_ads_tag_in_the_loop_provider() {
```
aaemnnosttv, line 260: Super nitpick, but we recently opted to use the same naming as WP core for data providers which uses (albeit inconsistently) a `data_` prefix rather than a `_provider` suffix.
```suggestion
	public function data_amp_auto_ads_tag_in_the_loop() {
```

tests/phpunit/integration/Modules/AdSenseTest.php
```diff
@@ -225,6 +229,45 @@ public function block_on_consent_provider() {
 		);
 	}
 
+	/**
+	 * @dataProvider amp_auto_ads_tag_in_the_loop_provider
```
aaemnnosttv, line 233:
```suggestion
	 * @dataProvider data_amp_auto_ads_tag_in_the_loop
```

tests/e2e/specs/modules/adsense/amp-auto-ads-tag.test.js
```diff
@@ -45,11 +46,16 @@ describe( 'AMP <amp-auto-ads> tag', () => {
...
 	it( 'is output in primary mode', async () => {
 		await setAMPMode( 'primary' );
-		await expect( '/hello-world' ).toHaveAMPAutoAdsTag();
+		await expect( createURL( '/hello-world' ) ).toHaveAMPAutoAdsTag();
 	} );
 
 	it( 'is output in secondary mode', async () => {
 		await setAMPMode( 'secondary' );
-		await expect( '/hello-world' ).toHaveAMPAutoAdsTag();
+		await expect( createURL( '/hello-world', 'amp' ) ).toHaveAMPAutoAdsTag();
```
aaemnnosttv, line 54: Originally I was thinking the query could be passed through in the first argument to `createURL` but it gets URL encoded, so this is fine since there isn't a great way to pass the query separately. It's possible to pass through as an argument to the matcher itself, but `toHaveAMPAutoAdsTag( 'amp' )` would be weird.

aaemnnosttv (approved):
Super!

Issue #1878, opened 2020-08-05
Refactor svgs to use React components via svg import


aaemnnosttv:
## Feature Description

Currently our build process transforms our svgs into a sprite which is then used by a `SvgIcon` component to render them. This is problematic to moving our asset pipeline over to be fully-managed by Webpack (#1326), and svg rendered in this way lacks benefits such as CSS styling.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Webpack should be configured to use [`@svgr/webpack`](https://www.npmjs.com/package/@svgr/webpack) for loading svgs
* All svg rendered within React apps should use a `ReactComponent` from the imported svg file itself
    - See https://create-react-app.dev/docs/adding-images-fonts-and-files/#adding-svgs
* The `SvgIcon` component should be removed
* All svg sprite-related dependencies and build steps should be removed
* Current [React svg components](https://github.com/google/site-kit-wp/tree/897c92e479c3577ceceb3355b8c58c5c7287c4e3/assets/js/components/icons) should be moved to svg files and replaced with the import

## Implementation Brief

* Install `@svgr/webpack` as a dev dependency.
* Update webpack config to use `@svgr/webpack` loader for `*.svg` files.
* Remove `Assets::svg_sprite` method.
* Update PageSpeed Insights module icon:
  * replace image icon with its svg version.
  * use CSS to create disable state for the icon.
* Refactor `PageHeader` component:
  * Rework it to be functional component.
  * Update `icon` property to be `PropTypes.node`.
  * Update icon rendering logic to render an icon passed via property directly instead of using `SvgIcon` component.
  * Replace conditional part with `${ props.children }`.
  * Remove `iconWidth`, `iconHeight` and `iconID` props.
  * Update `AdSenseDashboardWidget`, `AnalyticsDashboardWidget` and `GoogleSitekitSearchConsoleDashboardWidget` components to pass correct icon component to `PageHeader`.
  * Remove `googlesitekit.showDateRangeSelector-${ iconID }` filter and extract conditional part for date range into a new component `PageHeaderDateRange`.
  * Update `AnalyticsDashboardWidget` and `GoogleSitekitSearchConsoleDashboardWidget` components to pass an istance of `PageHeaderDateRange` component as a child for `PageHeader` component.
  * Remove `googlesitekit.showDateRangeSelector-${ slug }` filters from `analytics/index.legacy.js` and `search-console/index.legacy.js` files.
[...]

PR #2115, opened 2020-09-30
Enhancement/1878 refactor SVGs to use direct imports

caseymorrisus:
## Summary

Addresses issue [#1878](https://github.com/google/site-kit-wp/issues/1878)

## Relevant technical choices

- Configured Webpack, Storybook, & Jest to use `@svgr/webpack`
- Replaced all occurrences of `SvgIcon` in favor of direct imports
- Removed sprite related code (`Assets::svg_sprite`)
- Refactored `PageHeader` to be functional component as well as removing icon conditional logic in favor of passing icon directly.
- Conditional logic to show date ranges within `PageHeader` refactored with new `PageHeaderDateRange` component that is passed as a child.
- Refactor `DashBoardSplashModule` to be a functional component and render icons directly.
- Removed unused assets (`icon-pagespeed.png` & `SvgIcon`)
- Update `moduleIcon` function to use direct imports as well as remove unused `blockedByParentModule` param as this logic is now handled elsewhere.
- Use `import SvgName from 'svg-name.svg'` rather than `import { ReactComponent as SvgName } from 'svg-name.svg'` to avoid using `url-loader` which causes issues with data URIs within SCSS.

## Checklist


assets/svg/pagespeed-insights.svg
```diff
@@ -0,0 +1,39 @@
...
+                        <path id="_Path_12" data-name="&lt;Path&gt;" d="M132.86 79.06l-30.17 48.65a12.51 12.51 0 0 1-23.21-5.93v.48a12.52 12.52 0 0 0 23.24 6.5l30.14-48.66a2 2 0 0 0 .39-1.22 1.91 1.91 0 0 0-.08-.54 2.31 2.31 0 0 1-.31.72z" style="isolation:isolate" opacity=".2" fill="#212121"/>
+                        <path id="_Path_13" data-name="&lt;Path&gt;" d="M132.82 76.58a2 2 0 0 1 .43 1.26 2.06 2.06 0 0 1-.39 1.22l-30.17 48.65a12.53 12.53 0 0 1-18.84 3.05l28.08 28.07h59.47a12.53 12.53 0 0 0 12.54-12.54v-18.61z" fill="url(#linear-gradient)"/>
+                    </g>
+                    <path id="_Path_14" data-name="&lt;Path&gt;" d="M183.93 12.54A12.53 12.53 0 0 0 171.39 0H12.58A12.53 12.53 0 0 0 0 12.54v133.73a12.53 12.53 0 0 0 12.54 12.54h158.85a12.53 12.53 0 0 0 12.54-12.54z" fill="url(#radial-gradient)"/>
+                </g>
+            </g>
+        </g>
+    </g>
+</svg>
```
aaemnnosttv, line 39: I cleaned this up a bit manually – can you check and see that it looks the same?
```suggestion
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 183.95 158.81">
  <defs>
    <radialGradient id="radial-gradient" cx="1360.51" cy="10958.05" r="458.83" gradientTransform="matrix(.5 0 0 .5 -675.69 -5476.24)" gradientUnits="userSpaceOnUse">
      <stop offset="0" stop-color="#fff" stop-opacity=".1"/>
      <stop offset="1" stop-color="#fff" stop-opacity="0"/>
    </radialGradient>
    <linearGradient id="linear-gradient" x1="1572.36" y1="11164.11" x2="1691.74" y2="11283.49" gradientTransform="matrix(.5 0 0 .5 -675.69 -5476.24)" gradientUnits="userSpaceOnUse">
      <stop offset="0" stop-color="#212121" stop-opacity=".2"/>
      <stop offset="1" stop-color="#212121" stop-opacity="0"/>
    </linearGradient>
  </defs>
  <path d="M183.93 25.08v121.19a12.53 12.53 0 01-12.54 12.54H12.58A12.53 12.53 0 010 146.27V25.08z" fill="#e1e1e1"/>
  <path d="M12.58 0h158.81a12.53 12.53 0 0112.54 12.54v12.54H0V12.54A12.53 12.53 0 0112.58 0z" fill="#c2c2c2"/>
  <path d="M171.39 0H12.58A12.57 12.57 0 000 12.54v1A12.58 12.58 0 0112.58 1h158.81a12.59 12.59 0 0112.54 12.54v-1A12.58 12.58 0 00171.39 0z" fill-opacity=".2" fill="#fff"/>
  <path fill="#212121" fill-opacity=".1" d="M0 25.08h183.93v1H0z"/>
  <path d="M91.91 62.69a66.76 66.76 0 00-66.76 66.86c0 1.4-.09 3.14 0 4.18h29.13a36.26 36.26 0 010-4.18 37.64 37.64 0 0157.36-32l21-21a66.62 66.62 0 00-40.73-13.86z" fill="#4285f4"/>
  <path d="M132.64 76.52l-21 21a37.59 37.59 0 0117.9 32 36.26 36.26 0 010 4.18h29.27c.08-1 0-2.78 0-4.18a66.73 66.73 0 00-26.17-53z" fill="#f44336"/>
  <circle fill="#eee" cx="16.76" cy="12.54" r="4.18"/>
  <circle fill="#eee" cx="33.48" cy="12.54" r="4.18"/>
  <path fill="#212121" fill-opacity=".1" d="M171.41 157.76H12.54A12.57 12.57 0 010 145.23v1a12.57 12.57 0 0012.54 12.54h158.87a12.58 12.58 0 0012.59-12.5v-1a12.58 12.58 0 01-12.54 12.53z"/>
  <path d="M132.86 79.06a2.06 2.06 0 00.39-1.22 2.1 2.1 0 00-2.09-2.09 2 2 0 00-1.28.46l-45.57 35.07a12.53 12.53 0 1018.37 16.41l30.17-48.65z" fill="#9e9e9e"/>
  <path d="M84.31 112.33l45.57-35.09a2.06 2.06 0 011.28-.46 2.09 2.09 0 012 1.55 2 2 0 00-2-2.59 2 2 0 00-1.28.46l-45.57 35.07a12.5 12.5 0 00-4.87 9.92v.56a12.49 12.49 0 014.87-9.42z" style="isolation:isolate" opacity=".2" fill="#fff"/>
  <path d="M132.86 79.06l-30.17 48.65a12.51 12.51 0 01-23.21-5.93v.48a12.52 12.52 0 0023.24 6.5l30.14-48.66a2 2 0 00.39-1.22 1.91 1.91 0 00-.08-.54 2.31 2.31 0 01-.31.72z" style="isolation:isolate" opacity=".2" fill="#212121"/>
  <path d="M132.82 76.58a2 2 0 01.43 1.26 2.06 2.06 0 01-.39 1.22l-30.17 48.65a12.53 12.53 0 01-18.84 3.05l28.08 28.07h59.47a12.53 12.53 0 0012.54-12.54v-18.61z" fill="url(#linear-gradient)"/>
  <path d="M183.93 12.54A12.53 12.53 0 00171.39 0H12.58A12.53 12.53 0 000 12.54v133.73a12.53 12.53 0 0012.54 12.54h158.85a12.53 12.53 0 0012.54-12.54z" fill="url(#radial-gradient)"/>
</svg>
```

stories/page-header.stories.js
```diff
@@ -26,10 +27,7 @@ storiesOf( 'Global', module )
 			<p>
 				<PageHeader
 					title="Module Page Title with Icon"
-					icon
-					iconWidth="23"
-					iconHeight="26"
-					iconID="analytics"
+					icon={ <AnalyticsIcon width={ 23 } height={ 26 } className="googlesitekit-page-header__icon" /> }
```
aaemnnosttv, line 30: See above.

tests/js/svgrMock.js
```diff
@@ -0,0 +1,5 @@
+export default function() {
+	return <div />;
```
aaemnnosttv, line 2: Should we maybe return `<svg />` instead?

tests/js/svgrMock.js
```diff
@@ -0,0 +1,5 @@
+export default function() {
```
aaemnnosttv, line 1: Please add a file header 😄

.storybook/webpack.config.js
```diff
@@ -55,9 +55,22 @@ module.exports = async ( { config } ) => {
...
+		options: {
+			// strip width & height to allow manual override using props
+			dimensions: false,
+		},
+	} );
+
 	config.module.rules.push(
 		{
-			test: /\.(png|woff|woff2|eot|ttf|svg|gif)$/,
+			test: /\.(png|woff|woff2|eot|ttf|gif)$/,
```
aaemnnosttv, line 73: Maybe we should export these in a way that we can use them from the main webpack config file rather than duplicating the rules? We already share a few things with it.

.storybook/webpack.config.js
```diff
@@ -55,9 +55,22 @@ module.exports = async ( { config } ) => {
 		},
 	);
 
+	// exclude existing svg rule created by storybook before pushing custom rule
+	const fileLoaderRule = config.module.rules.find( ( rule ) => rule.test && rule.test.test( '.svg' ) );
+	fileLoaderRule.exclude = /\.svg$/;
```
aaemnnosttv, line 60: Why would we exclude the rule? Maybe we should just remove it so we only use the one going forward? Just curious if there is a reason why we would want/need it now that we're doing the same outside of storybook.

aaemnnosttv (changes requested):
This looks really nice @caseymorrisus 👍 

A few minor changes to make, mostly around formatting in a few places. I also left a few questions and suggestions.

There are a few items from the AC that are still left to be done (to be fair, these were missing in the IB somehow 😄 ):
* All svg sprite-related dependencies and build steps should be removed  
_i.e. `gulp-svgstore`, `gulpfile.js/svgstore.js`, and other related tasks/dependencies which are no longer needed_
* Current [React svg components](https://github.com/google/site-kit-wp/tree/897c92e479c3577ceceb3355b8c58c5c7287c4e3/assets/js/components/icons) should be moved to svg files and replaced with the import

assets/js/components/page-header.js
```diff
@@ -21,98 +21,69 @@
...
+		`;
 
-		/**
-		 * Filter whether to show a date range selector.
-		 *
-		 * Modules can op into the date range selection feature by returning true.
-		 */
-		const showDateRangeSelector = applyFilters( `googlesitekit.showDateRangeSelector-${ iconID }`, false );
+	// Determine whether the details cell should display.
+	const hasDetails = '' !== status || typeof props.children !== 'undefined';
```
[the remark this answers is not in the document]
caseymorrisus: I can agree with that and will change this.

I will say that setting children directly as `null` or `false` seems like an anti-pattern unless you're intentionally wanting to bypass an unset prop conditional. I like keeping `null` and `undefined` values distinct as they have different meanings IMO. Checking an `undefined` type for `children` directly allows you to test components in isolation (no children rendered, to assert on wrapper elements) via `<Component children={ null } />` vs supplying a dummy component `<Component children={ () => {} } />`. I don't feel too strongly about it either way.

Changing this to the following to keep the return type of `hasDetails` a boolean. Without, it will have the return type of `false | React.Node`. A bit nit picky as `React.Node` is truthy, but may fail if someone later expects a boolean and operates on the value with a method that does not use type coercion or has unexpected coercion (handles React components in a special way). 

```js
const hasDetails = '' !== status || Boolean( children ); // returns true | false
```

assets/js/util/index.js
```diff
@@ -573,28 +577,33 @@ export const decodeHTMLEntity = ( str ) => {
...
-		iconComponent = <img src={ global._googlesitekitLegacyData.admin.assetsRoot + 'images/icon-pagespeed.png' } width={ width } alt="" className={ useClass } />;
+	switch ( module ) {
+		case 'adsense':
+			return <AdSenseIcon width={ width } height={ height } className={ useClass } />;
+		case 'analytics':
+			return <AnalyticsIcon width={ width } height={ height } className={ useClass } />;
+		case 'optimize':
+			return <OptimizeIcon width={ width } height={ height } className={ useClass } />;
+		case 'pagespeed-insights':
+			return <PageSpeedInsightsIcon width={ undefined } height={ height } className={ useClass } />;
```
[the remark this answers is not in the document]
caseymorrisus: Width is no longer needed, it originally fixed a visual regression test but a later change fixed it elsewhere. Using a map vs a switch statement (since all props are equal) is a good refactor, making that change as well as returning `null` vs throwing.

.storybook/webpack.config.js
caseymorrisus: Storybook automatically adds a rule for SVGs in the background that conflicts with the rule we try to use with our SVGR config. This is required for SVGs within Storybook to match behavior within our standard application.

assets/svg/device-size-desktop-icon.svg
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg name="google-sitekit-device-size-desktop" width="23" height="17" viewBox="0 0 23 17" fill="none" xmlns="http://www.w3.org/2000/svg">
```
aaemnnosttv, line 2: Does the `xml` tag at the top make a difference at all? If not maybe we should just remove it since it seems others don't have it.

I see the `name` was in the previous component as well but looks like it isn't used anywhere; perhaps we should remove this as well?

These comments apply to `assets/svg/device-size-mobile-icon.svg` too.

gulpfile.js/index.js
```diff
@@ -62,7 +61,6 @@ exports.imagemin = function( cb ) {
  * Gulp task to minify and combine svg's.
  */
 exports.svg = gulp.series(
-	svgstore,
 	svgmin,
```
aaemnnosttv, line 64: Is this task still relevant? I would think the `gulp svg` command should probably be unnecessary now as svgs should be entirely handled by Webpack. I mentioned svgstore before because it was clearly related to the sprite creation but again, svgs should be fully built/managed by Webpack going forward.

assets/js/util/index.js
```diff
@@ -573,28 +586,26 @@ export const decodeHTMLEntity = ( str ) => {
...
  * @param {string}  width                 The icon width.
  * @param {string}  height                The icon height.
  * @param {string}  useClass              Class string to use for icon.
  *
  * @return {HTMLImageElement}             <img> tag with module icon.
  */
-export function moduleIcon( module, blockedByParentModule, width = '33', height = '33', useClass = '' ) {
-	if ( ! global._googlesitekitLegacyData ) {
+export function moduleIcon( module, width = '33', height = '33', useClass = '' ) {
+	if ( ! global._googlesitekitLegacyData || ! moduleMap.hasOwnProperty( module ) ) {
```
aaemnnosttv, line 596: This condition should no longer require `_googlesitekitLegacyData` since it isn't actually a dependency anymore 😄 
```suggestion
	if ( ! moduleMap.hasOwnProperty( module ) ) {
```

---

I'm starting to think this should simply be a `ModuleIcon` component, since it's basically a functional component now.
How about this:
```js
function ModuleIcon( { slug, className, width = 33, height = 33 } ) { ... }
// or even better
function ModuleIcon( { slug, width = 33, height = 33, ...props } ) { ... }
// and spread the rest into the inner component
```
That would make it much clearer to use as well rather than numeric positional parameters for width and height and `useClass` instead of `className` 😄 

There are only a handful of instances of this so it would be a simple refactor to apply as well.

assets/js/components/page-header.js
```diff
@@ -121,7 +92,7 @@ PageHeader.propTypes = {
 
 PageHeader.defaultProps = {
 	title: '',
-	icon: false,
+	icon: null,
 	iconID: '',
```
aaemnnosttv, line 96: `iconID` should no longer be needed 😄

assets/js/components/page-header.js
```diff
@@ -21,98 +21,69 @@
...
+			</div>
+		</header>
+	);
 }
 
 PageHeader.propTypes = {
 	title: PropTypes.string,
-	icon: PropTypes.bool,
+	icon: PropTypes.node,
 	iconID: PropTypes.string,
```
aaemnnosttv, line 86: Should no longer be needed.

aaemnnosttv (commented):
Thanks @caseymorrisus ! A few more comments and suggestions for you but this is looking good 👍

assets/svg/device-size-desktop-icon.svg
caseymorrisus: @aaemnnosttv The XML declaration should not make a difference (unless we're using non-standard encoding) in-browser as the parser will strip this out before rendering. However, this may make a difference if you try to use the SVG in a non-browser environment (opening the SVG in a program to edit) which is why I try to XML declarations in unless we have a copy of the original file (Sketch/Illustrator file). I'd rather keep SVGs closer to source and process them on build (can minify SVGs via Webpack/SVGO to remove any extra attributes or declarations). As far as consistency goes (without the 2 newly added SVGs) 20 have XML declarations and 13 do not, adding the XML declaration follow consistency a bit more. Let me know how you feel about this, if we're going with no XML declarations I think we should remove them from the 20 SVGs that currently have them to avoid future confusion.

gulpfile.js/index.js
caseymorrisus: @aaemnnosttv Now that I'm looking more at this (and considering a previous comment where I mention minifying), should we be adding feature parity to Webpack that Gulp used to have? We previously had an SVG minifier setup within our build step but the original issue makes no mention of re-creating the minifying logic in Webpack, only that the previous setup should be removed.

assets/js/util/index.js
```diff
@@ -573,28 +586,26 @@ export const decodeHTMLEntity = ( str ) => {
...
-	if ( blockedByParentModule ) {
-		iconComponent = <SvgIcon id={ `${ module }-disabled` } width={ width } height={ height } className={ useClass } />;
-	} else if ( 'pagespeed-insights' === module ) {
-		iconComponent = <img src={ global._googlesitekitLegacyData.admin.assetsRoot + 'images/icon-pagespeed.png' } width={ width } alt="" className={ useClass } />;
-	}
-
-	return iconComponent;
+	return (
+		<ModuleIconComponent
+			width={ module === 'pagespeed-insights' ? undefined : width }
```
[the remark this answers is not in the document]
caseymorrisus: @aaemnnosttv I think accepting the visual change is the fix for this one as all of the other icons behave similarly (width of `33px` with a variable height) whereas the PageSpeed icon had a width of `38.22px`


**"Broken" PageSpeed Icon, previous implementation (38.22px width)**
image

**PageSpeed Icon with matching behavior (33px width)**
image

gulpfile.js/index.js
caseymorrisus: Related gulp tasks were removed but I'm keeping this convo unresolved to discuss Webpack feature parity with the previous Gulp setup.

assets/js/util/index.js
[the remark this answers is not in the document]
aaemnnosttv: @caseymorrisus this looks okay to me but since this is a user-facing change I'd like to get @felixarntz to weigh in on whether or not this is okay to change.

I think `moduleIcon`s should have consistent dimensions, regardless of which icon is shown. I think the reason the PSI icon is shown larger is to compensate for it's content being shorter vertically which gives the appearance of it looking smaller than others. In this case, you could argue that perhaps the `SetupModule` component (I think it is) that calls `moduleIcon` for PSI specifically could call it with the larger width to compensate, but I think `moduleIcon` itself should be consistent.

assets/svg/device-size-desktop-icon.svg
aaemnnosttv: > I'd rather keep SVGs closer to source and process them on build

That sounds like a good approach to me.

> As far as consistency goes (without the 2 newly added SVGs) 20 have XML declarations and 13 do not, adding the XML declaration follow consistency a bit more. Let me know how you feel about this, if we're going with no XML declarations I think we should remove them from the 20 SVGs that currently have them to avoid future confusion.

While I agree it would be good to have consistency here, it's not really in the scope of this issue to change so I'd prefer to revisit that in a separate issue (feel free to create that one if you'd like?).

Since the existing components didn't have the xml tag, let's not include them in the svg files for now since it seems it isn't required.

gulpfile.js/index.js
aaemnnosttv: Good question! It looks like SVGO is included with `@svgr/webpack`, we may just need to configure it
```
├─┬ @svgr/webpack@5.4.0
│ └─┬ @svgr/plugin-svgo@5.4.0
│   └── svgo@1.3.2
```
https://react-svgr.com/docs/configuration-files/

I don't think we need to go to any extreme lengths to preserve 100% parity with what we had before, but establishing the configuration file with any necessary/sensible values would be great 👍

assets/js/components/module-icon.js
```diff
@@ -0,0 +1,63 @@
...
+	analytics: AnalyticsIcon,
+	optimize: OptimizeIcon,
+	'pagespeed-insights': PageSpeedInsightsIcon,
+	'search-console': SearchConsoleIcon,
+	tagmanager: TagManagerIcon,
+};
+
+export default function ModuleIcon( { slug, width = 33, height = 33, ...props } ) {
+	if ( ! moduleMap.hasOwnProperty( slug ) ) {
+		return;
```
aaemnnosttv, line 45: This should return `null` now that it's a component.
```suggestion
		return null;
```

caseymorrisus: Good catch, returning `undefined` is a no go.

assets/js/util/index.js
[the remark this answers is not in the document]
caseymorrisus: @aaemnnosttv I had the same thought process when comparing the Analytics icon & PageSpeed icon side by side, but if you compare them to the other icons they're all similar in size, that's my bad for not providing a better screenshot. The Analytics icon is the outlier, if anything it should be the element to get a 1-off style to match the other 3 icons.

See this screenshot below (hid the Analytics section) to see how the icons compare with each other when they all have matching behavior (no conditionals):

image

gulpfile.js/index.js
caseymorrisus: Looks like we were using the default SVGO config with `removeViewbox` set to `false`, this is pretty easy to mimic with a config file. This has been added to `.svgo.yml`

stories/page-header.stories.js
```diff
@@ -26,10 +27,13 @@ storiesOf( 'Global', module )
...
 					title="Module Page Title with Icon"
-					icon
-					iconWidth="23"
-					iconHeight="26"
-					iconID="analytics"
+					icon={
+						<AnalyticsIcon
+							className="googlesitekit-page-header__icon"
+							width={ 23 }
+							height={ 26 }
```
felixarntz, line 34: Why are these values `23` and `26` while further below they are both `33`? Shouldn't this be square here as well?

stories/splash.stories.js
```diff
@@ -76,7 +80,7 @@ storiesOf( 'Splash', module )
 							mdc-layout-grid__cell--span-3-desktop
 						">
 							<DashboardSplashModule
-								icon="pagespeed"
+								icon={ <PageSpeedInsightsIcon height="33" /> }
```
felixarntz, line 83: This is missing the `width` I assume?

assets/js/components/page-header.js
```diff
@@ -121,8 +91,7 @@ PageHeader.propTypes = {
 
 PageHeader.defaultProps = {
 	title: '',
-	icon: false,
-	iconID: '',
+	icon: null,
```
felixarntz, line 94: Nice!

assets/js/components/module-icon.js
```diff
@@ -0,0 +1,63 @@
...
+const moduleMap = {
+	adsense: AdSenseIcon,
+	analytics: AnalyticsIcon,
+	optimize: OptimizeIcon,
+	'pagespeed-insights': PageSpeedInsightsIcon,
+	'search-console': SearchConsoleIcon,
+	tagmanager: TagManagerIcon,
+};
+
+export default function ModuleIcon( { slug, width = 33, height = 33, ...props } ) {
```
felixarntz, line 43: This component is good to have for now, but I think in the future these icons should come directly from `registerModule` in JS where they should be passed as the `icon` property, so that the icons are really coming from the module files and not hardcoded into the JS core of Site Kit as they are here.

assets/js/components/module-icon.js
felixarntz: Looks like this wasn't addressed yet.

assets/js/components/settings/settings-module.js
```diff
@@ -275,7 +276,7 @@ class SettingsModule extends Component {
 											googlesitekit-heading-4
 											googlesitekit-settings-module__title
 										">
-											{ moduleIcon( slug, false, '24', '26', 'googlesitekit-settings-module__title-icon' ) }
+											<ModuleIcon slug={ slug } width={ 24 } height={ 26 } className="googlesitekit-settings-module__title-icon" />
```
felixarntz, line 279: Same question as before (also applies to several other locations): Shouldn't the dimensions for module icons always be square?

felixarntz (commented):
Some questions here, mostly LGTM.

assets/js/util/index.js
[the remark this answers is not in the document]
felixarntz: I think the new format is okay, let's keep the module icon widths consistent. I had some related questions in my PR review, like why aren't the dimensions for the module icons always square?

stories/page-header.stories.js
caseymorrisus: I'm not 100% sure on the original motivation as this is matching the previous implementation (without, VR tests fail). I will note that that setting values here as `23, 26` would be the same as setting it to `26, 26` as the height is the largest value and SVGs automatically scale the smaller value to fit without stretching (unlike images when setting both).

stories/splash.stories.js
caseymorrisus: This is a leftover before the consistency fix was implemented, good catch.

assets/js/components/settings/settings-module.js
caseymorrisus: @felixarntz Answered this in a previous comment but figured I'd add thoughts here as well. I'm curious if setting non-square values is a leftover from when the icons were images or having the assumption that SVGs scale the same way that images do. As all of the icons are rectangles, setting them with equal width/height values (if they were images) would result in visual distortion as the images would scale up to fill the space whereas SVGs would not scale up the smaller value. I tried to find the commit where these values were added but they point to the initial commit. I also tried to find discussion about the topic by searching the repository for `moduleIcon` but it doesn't return any results.

assets/js/components/page-header.js
```diff
@@ -121,8 +91,7 @@ PageHeader.propTypes = {
...
 	iconWidth: '',
 	iconHeight: '',
```
aaemnnosttv, line 96:
```suggestion
```

assets/js/components/page-header.js
```diff
@@ -121,8 +91,7 @@ PageHeader.propTypes = {
 
 PageHeader.defaultProps = {
 	title: '',
-	icon: false,
-	iconID: '',
+	icon: null,
 	iconWidth: '',
```
aaemnnosttv, line 95: Two more leftovers to remove here as well 😄 
```suggestion
```

aaemnnosttv (commented):
This is nearly good to go, just a handful of legacy props to clean up that were left over from before.

stories/page-header.stories.js
aaemnnosttv: @felixarntz I think it's preferable to keep the dimensions consistent with what were used before for now. I'll open a new issue as a follow up here to cover this, as well as the refactor to provide the icon during module registration.

aaemnnosttv (approved):

stories/page-header.stories.js
aaemnnosttv: @felixarntz see https://github.com/google/site-kit-wp/issues/2151

Issue #1995, opened 2020-09-08
Fix loading state logic ignoring errors in components


felixarntz:
This started as part of #1859, but was since separated out to keep issue scope more manageable. Some module components (mostly setup/settings-related) currently have the problem that the logic for displaying the overall `ProgressBar` depends on data being returned from an API - however, if an error occurs during an API request, that data will never be populated, keeping the component in an infinite loading state.

Instead of relying on selector-based data to exist, it should be relied upon whether the respective resolver has finished, which will work for both the success and error case.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* For the condition whether to display a progress bar, JS components that are used in module setup or module settings must not rely on selectors returning `undefined` which have a resolver. Whenever a selector has a resolver, such logic should rely on `hasFinishedResolution` for the selector instead.

## Implementation Brief

* Update conditionals used for early return with `<ProgressBar>` when data selected from the datastore `=== undefined` to use `! hasFinishedResolution` selector instead (only for selectors with a resolver).  
    Note, not all selectors have a dedicated selector but use a parent selector which has a resolver. In this case, the "closest" resolver in the selector chain should be checked.  
    Also, special care should be taken to ensure bugs are not introduced by changing the conditional to return when something is `undefined`. It will now be possible for these selected values to be `undefined` (such as in the event of an error) beyond the progress bar conditional, so we should pay extra attention to ensure potential type errors are handled.  
    - `assets/js/modules/adsense/components/common/AccountSelect.js`
    - `assets/js/modules/adsense/components/common/SiteSteps.js`
    - `assets/js/modules/adsense/components/common/UserProfile.js`
    - `assets/js/modules/adsense/components/settings/SettingsEdit.js`
    - `assets/js/modules/adsense/components/setup/SetupMain.js`
    - `assets/js/modules/analytics/components/common/AccountSelect.js`
    - `assets/js/modules/analytics/components/common/AccountCreate/index.js`
    - `assets/js/modules/tagmanager/components/common/AccountCreate.js`
    - `assets/js/modules/tagmanager/components/common/AccountSelect.js`
    - `assets/js/modules/tagmanager/components/common/ContainerSelect.js`
* Similar to the last bullet, some conditionals used in the same way are missing this check and are not checking `undefined` so they only need the same resolver check added  
    _You can see an example of this done correctly in `assets/js/modules/analytics/components/settings/SettingsEdit.js`_
[...]

PR #2121, opened 2020-10-01
Bug/1995 fix loading state

eugene-manuilov:
## Summary

Addresses issue #1995

## Relevant technical choices

- Removed progress bar from `assets/js/modules/adsense/components/common/SiteSteps.js` because we don't need it there.

## Checklist


assets/js/modules/adsense/components/settings/SettingsForm.js
```diff
@@ -31,11 +31,17 @@ import {
 	ErrorNotice,
 	UseSnippetSwitch,
 } from '../common';
+import ProgressBar from '../../../../components/progress-bar';
 const { useSelect } = Data;
 
 export default function SettingsForm() {
 	const clientID = useSelect( ( select ) => select( STORE_NAME ).getClientID() );
 	const existingTag = useSelect( ( select ) => select( STORE_NAME ).getExistingTag() );
+	const hasResolvedGetExistingTag = useSelect( ( select ) => select( STORE_NAME ).hasFinishedResolution( 'getExistingTag' ) );
```
felixarntz, line 40: Nitpick, but it would probably make sense to combine these like you did above (since they belong to the same selector call in a way).

assets/js/modules/analytics/components/common/ProfileSelect.js
```diff
@@ -51,7 +53,7 @@ export default function ProfileSelect() {
 		}
 	}, [ profileID ] );
 
-	if ( isLoadingAccounts || isLoadingProperties || isLoadingProfiles ) {
+	if ( ! hasResolvedAccounts || ( accountID && ! hasResolvedProperties ) || ! hasResolvedProfiles ) {
```
felixarntz, line 56: Why the `accountID` check in the second part of the if clause? I'd think if we add that, it would make sense to do the same for `accountID` and `propertyID` in the third segment since those are arguments to `getProfiles`.

assets/js/modules/analytics/components/common/ProfileSelect.test.js
```diff
@@ -37,11 +37,20 @@ apiFetchMock.mockImplementation( ( ...args ) => {
...
-	dispatch( STORE_NAME ).setPropertyID( webPropertyId ); // eslint-disable-line sitekit/camelcase-acronyms
+	const {
+		id,
+		webPropertyId: propertyID, // eslint-disable-line sitekit/camelcase-acronyms
+		accountId: accountID, // eslint-disable-line sitekit/camelcase-acronyms
+	} = fixtures.propertiesProfiles.profiles[ 0 ];
+
+	dispatch( STORE_NAME ).finishResolution( 'getAccounts', [] );
+	dispatch( STORE_NAME ).finishResolution( 'getProperties', [ accountID ] );
+	dispatch( STORE_NAME ).finishResolution( 'getProperties', [ '0' ] );
```
felixarntz, line 48: Why is that here? Seems like there's something else we need to fix instead... Also, I think we should call `receiveGetAccounts` and `receiveGetProperties` here with applicable fixture values too. While they're technically not used in the component, not passing that data while at the same time marking the resolver resolution as finished causes a state that shouldn't exist.

assets/js/modules/analytics/components/common/ProfileSelect.test.js
```diff
@@ -50,6 +59,10 @@ const setupRegistryWithExistingTag = ( { dispatch } ) => {
 		propertyID: fixtures.accountsPropertiesProfiles.profiles[ 0 ].webPropertyId, // eslint-disable-line sitekit/camelcase-acronyms
 	};
 	const { id } = fixtures.propertiesProfiles.profiles[ 0 ];
+
+	dispatch( STORE_NAME ).finishResolution( 'getAccounts', [] );
+	dispatch( STORE_NAME ).finishResolution( 'getProperties', [ existingTag.accountID ] );
+	dispatch( STORE_NAME ).finishResolution( 'getProfiles', [ existingTag.accountID, existingTag.propertyID ] );
```
felixarntz, line 65: See above, let's also populate data for these selectors.

assets/js/modules/analytics/components/common/ProfileSelect.test.js
```diff
@@ -60,7 +73,13 @@ const setupRegistryWithExistingTag = ( { dispatch } ) => {
 const setupEmptyRegistry = ( { dispatch } ) => {
 	const accountID = fixtures.accountsPropertiesProfiles.profiles[ 0 ].accountId; // eslint-disable-line sitekit/camelcase-acronyms
 	const propertyID = fixtures.accountsPropertiesProfiles.profiles[ 0 ].webPropertyId; // eslint-disable-line sitekit/camelcase-acronyms
+
+	dispatch( STORE_NAME ).finishResolution( 'getAccounts', [] );
+	dispatch( STORE_NAME ).finishResolution( 'getProperties', [ accountID ] );
+	dispatch( STORE_NAME ).finishResolution( 'getProfiles', [ accountID, propertyID ] );
```
felixarntz, line 79: See above.

assets/js/modules/analytics/components/common/PropertySelect.js
```diff
@@ -50,10 +50,10 @@ export default function PropertySelect() {
 		return data;
 	} );
 
-	const hasExistingTag = useSelect( ( select ) => select( STORE_NAME ).hasExistingTag() );
-	const hasGTMPropertyID = useSelect( ( select ) => !! select( MODULES_TAGMANAGER ).getSingleAnalyticsPropertyID() );
 	const propertyID = useSelect( ( select ) => select( STORE_NAME ).getPropertyID() );
 	const hasResolvedAccounts = useSelect( ( select ) => select( STORE_NAME ).hasFinishedResolution( 'getAccounts' ) );
+	const hasExistingTag = useSelect( ( select ) => select( STORE_NAME ).hasExistingTag() );
+	const hasGTMPropertyID = useSelect( ( select ) => !! select( MODULES_TAGMANAGER ).getSingleAnalyticsPropertyID() );
```
felixarntz, line 56: Is there a specific reason you moved these? Just curious since I'd imagine this doesn't change anything?

assets/js/modules/tagmanager/components/common/AMPContainerSelect.test.js
```diff
@@ -50,6 +50,8 @@ describe( 'AMPContainerSelect', () => {
 		registry.dispatch( STORE_NAME ).setAccountID( accountID );
 		registry.dispatch( STORE_NAME ).receiveGetAccounts( [ account ] );
 		registry.dispatch( STORE_NAME ).receiveGetContainers( [ ...webContainers, ...ampContainers ], { accountID } );
+		registry.dispatch( STORE_NAME ).finishResolution( 'getAccounts', [] );
+		registry.dispatch( STORE_NAME ).finishResolution( 'getContainers', [ accountID ] );
```
felixarntz, line 54: Maybe we can group each `finishResolution` call with the corresponding `receiveGet...` call? That would make the connection more clear when reading the code.

assets/js/modules/tagmanager/components/common/AMPContainerSelect.test.js
```diff
@@ -70,6 +72,8 @@ describe( 'AMPContainerSelect', () => {
 		registry.dispatch( STORE_NAME ).setAccountID( accountID );
 		registry.dispatch( STORE_NAME ).receiveGetAccounts( [ account ] );
 		registry.dispatch( STORE_NAME ).receiveGetContainers( containers, { accountID } );
+		registry.dispatch( STORE_NAME ).finishResolution( 'getAccounts', [] );
+		registry.dispatch( STORE_NAME ).finishResolution( 'getContainers', [ accountID ] );
```
felixarntz, line 76: See above (same applies to the rest of this file).

assets/js/modules/tagmanager/components/common/AMPContainerSelect.test.js
```diff
@@ -135,6 +147,8 @@ describe( 'AMPContainerSelect', () => {
 		const accountID = account.accountId; // eslint-disable-line sitekit/camelcase-acronyms
 		registry.dispatch( STORE_NAME ).receiveGetAccounts( [ account ] );
 		registry.dispatch( STORE_NAME ).setAccountID( accountID );
+		registry.dispatch( STORE_NAME ).finishResolution( 'getAccounts', [] );
+		registry.dispatch( STORE_NAME ).finishResolution( 'getContainers', [ accountID ] );
```
felixarntz, line 151: Why is this marked as finished here? I'd think this should explicitly _not_ be finished here, to trigger the loading state in the UI.

assets/js/modules/tagmanager/components/common/AMPContainerSelect.test.js
```diff
@@ -149,6 +163,8 @@ describe( 'AMPContainerSelect', () => {
 		registry.dispatch( STORE_NAME ).receiveGetAccounts( [ account ] );
 		registry.dispatch( STORE_NAME ).receiveGetContainers( containers, { accountID } );
 		registry.dispatch( CORE_SITE ).receiveSiteInfo( { ampMode: AMP_MODE_PRIMARY } );
+		registry.dispatch( STORE_NAME ).finishResolution( 'getAccounts', [] );
+		registry.dispatch( STORE_NAME ).finishResolution( 'getContainers', [ accountID ] );
```
felixarntz, line 167: See above comment regarding grouping.

assets/js/modules/tagmanager/components/common/AMPContainerSelect.test.js
```diff
@@ -162,6 +178,8 @@ describe( 'AMPContainerSelect', () => {
 		registry.dispatch( STORE_NAME ).receiveGetAccounts( [ account ] );
 		registry.dispatch( STORE_NAME ).receiveGetContainers( containers, { accountID } );
 		registry.dispatch( CORE_SITE ).receiveSiteInfo( { ampMode: AMP_MODE_SECONDARY } );
+		registry.dispatch( STORE_NAME ).finishResolution( 'getAccounts', [] );
+		registry.dispatch( STORE_NAME ).finishResolution( 'getContainers', [ accountID ] );
```
felixarntz, line 182: See above.

assets/js/modules/tagmanager/components/common/ContainerSelect.js
```diff
@@ -43,12 +43,16 @@ export default function ContainerSelect( {
...
+		hasResolvedAccounts: select( STORE_NAME ).hasFinishedResolution( 'getAccounts' ),
+	} ) );
+
 	const accountID = useSelect( ( select ) => select( STORE_NAME ).getAccountID() );
 	const hasExistingTag = useSelect( ( select ) => select( STORE_NAME ).hasExistingTag() );
-	const isLoadingContainers = useSelect( ( select ) => select( STORE_NAME ).isDoingGetContainers( accountID ) );
+	const hasResolvedContainers = useSelect( ( select ) => select( STORE_NAME ).hasFinishedResolution( 'getContainers', [ accountID ] ) );
 
-	if ( accounts === undefined || containers === undefined || isLoadingContainers ) {
+	if ( accounts === undefined || ! hasResolvedAccounts || containers === undefined || ! hasResolvedContainers ) {
```
felixarntz, line 55: I think the `undefined` checks still need to be removed from here.

assets/js/modules/tagmanager/components/common/WebContainerSelect.test.js
```diff
@@ -50,6 +50,8 @@ describe( 'WebContainerSelect', () => {
 		registry.dispatch( STORE_NAME ).setAccountID( accountID );
 		registry.dispatch( STORE_NAME ).receiveGetAccounts( [ account ] );
 		registry.dispatch( STORE_NAME ).receiveGetContainers( [ ...webContainers, ...ampContainers ], { accountID } );
+		registry.dispatch( STORE_NAME ).finishResolution( 'getAccounts', [] );
+		registry.dispatch( STORE_NAME ).finishResolution( 'getContainers', [ accountID ] );
```
felixarntz, line 54: See above.

felixarntz (changes requested):
This mostly looks solid, I mostly have a few follow-up questions and some minor suggestions.

assets/js/modules/adsense/components/settings/SettingsForm.js
eugene-manuilov: Updated.

assets/js/modules/analytics/components/common/PropertySelect.js
eugene-manuilov: I don't remember why it has been moved. Reverted to its original state.

assets/js/modules/tagmanager/components/common/AMPContainerSelect.test.js
eugene-manuilov: Yep, makes sense. Updated.

assets/js/modules/tagmanager/components/common/AMPContainerSelect.test.js
eugene-manuilov: Updated.

assets/js/modules/tagmanager/components/common/AMPContainerSelect.test.js
eugene-manuilov: Good catch! Thanks! Updated to fix this issue.

assets/js/modules/tagmanager/components/common/AMPContainerSelect.test.js
eugene-manuilov: Updated.

assets/js/modules/tagmanager/components/common/AMPContainerSelect.test.js
eugene-manuilov: Updated.

assets/js/modules/tagmanager/components/common/ContainerSelect.js
eugene-manuilov: Yep, updated.

assets/js/modules/tagmanager/components/common/WebContainerSelect.test.js
eugene-manuilov: Updated.

assets/js/modules/analytics/components/common/ProfileSelect.js
eugene-manuilov: Ok, looks like we don't need it anymore, but in the beginning, I had a situation when profiles select got loaded faster than properties select and having `accountID` check here helped to solve that race condition.

I have removed `accountID`, so it should be good now.

assets/js/modules/analytics/components/common/ProfileSelect.test.js
eugene-manuilov: It is required for the `should be disabled when in the absence of an valid account or property ID.` test. I have updated that test to finish this resolution there and added fixture values too.

assets/js/modules/analytics/components/common/ProfileSelect.test.js
eugene-manuilov: Updated.

assets/js/modules/analytics/components/common/ProfileSelect.test.js
eugene-manuilov: Updated.

assets/js/modules/analytics/components/common/ProfileSelect.js
eugene-manuilov: Actually, there was an issue when nothing was selected on the setup page. I have updated ProfileSelect and PropertySelect components to check whether fetching is started to prevent progress bars from appearing unexpectedly. Also added a new story to the Analytics setup storybook.

assets/js/modules/analytics/components/common/ProfileSelect.js
```diff
@@ -34,13 +34,41 @@ import { trackEvent } from '../../../../util';
...
+		hasResolvedProfiles,
+	} = useSelect( ( select ) => {
+		const data = {
+			accountID: select( STORE_NAME ).getAccountID(),
+			propertyID: select( STORE_NAME ).getPropertyID(),
+			profiles: [],
+			hasStartedFetchingProperties: false,
+			hasResolvedProperties: false,
+			hasStartedFetchingProfiles: false,
+			hasResolvedProfiles: false,
```
felixarntz, line 56: In this case, wouldn't we be able to use the `isResolving` meta selector instead? Then we could have e.g. `isResolvingProperties` which I believe would do the same as `hasStartedFetchingProperties && ! hasResolverProperties`?

assets/js/modules/analytics/components/common/PropertySelect.js
```diff
@@ -35,15 +35,22 @@ import { trackEvent } from '../../../../util';
...
+		accountID,
+		properties,
+		hasStartedFetchingProperties,
+		hasResolvedProperties,
+	} = useSelect( ( select ) => {
 		const data = {
 			accountID: select( STORE_NAME ).getAccountID(),
 			properties: [],
+			hasStartedFetchingProperties: false,
 			hasResolvedProperties: false,
```
felixarntz, line 48: See above, we might be able to simplify this via `isResolving`.

assets/js/modules/analytics/components/common/PropertySelect.js
```diff
@@ -64,7 +71,7 @@ export default function PropertySelect() {
 		}
 	}, [ propertyID ] );
 
-	if ( ! hasResolvedAccounts || ( accountID && ! hasResolvedProperties ) ) {
+	if ( ! hasResolvedAccounts || ( hasStartedFetchingProperties && ! hasResolvedProperties ) ) {
```
felixarntz, line 74: Why is this extra check for whether resolving has already started necessary? If there's nothing selected, shouldn't the resolver still have to be resolved either way (it would be the resolver for e.g. a value of `undefined`).

felixarntz (commented):
A few follow-up questions here, but this should be almost good to go.

assets/js/modules/analytics/components/common/ProfileSelect.js
eugene-manuilov: Ah... yes, you are right. We can use `isResolving` here. Updated.

assets/js/modules/analytics/components/common/PropertySelect.js
eugene-manuilov: Updated to use `isResolving`.

assets/js/modules/analytics/components/common/PropertySelect.js
eugene-manuilov: I have updated it to use `isResolving` which is more accurate and strictly identifies when we are fetching something or not. It helps us to show the progress bar only when an actual process of fetching is happening.

felixarntz (approved):
Awesome!

Issue #2123, opened 2020-10-02
Enhance simulating AMP context in PHP tests


aaemnnosttv:
## Feature Description

Currently, a number of features and behaviors are AMP-specific and rely on logic implemented in the `Context` class for determining the AMP context and mode.

In PHP tests, this has been less than ideal to handle and requires us mocking a dummy object with stubbed responses to `is_amp` method calls. Because classes under test still need to be instantiated with a real `Context` instance, we need to do that first and then force-set the private `context` property with the dummy instance. This is problematic because any internally instantiated classes will have the `Context` instance that was passed in the constructor rather than the fake one.

This is all overly complicated simply because `Context` is defined as `final` which it has been since the initial release and prevents us from using a simple sub-class for use in tests. This is unnecessarily restrictive as nearly our entire PHP codebase is intended to be private so anyone who might be extending this class would be "doing it wrong" so we shouldn't need to protect developers from taking risks that they shouldn't.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `\Google\Site_Kit\Context` class should no longer be declared as `final`
* Tests requiring AMP-specific context should use an instance of `Context` which is fixed to return AMP-specific values for relevant methods
    * `::is_amp()` should always return `true` unconditionally
    * `::get_amp_mode()` should always return `primary` or `secondary` using class constants on the `Context` class
* AMP mode-specific instances should be used rather than generic/configurable instances
* All existing workarounds for faking AMP context in PHPUnit tests should be updated to use the new instances

## Implementation Brief

* Merge #2125

### Test Coverage

* This issue is about refactoring for tests; no additional coverage is necessary

### Visual Regression Changes

* N/A

## QA Brief

* PHPUnit tests should all pass
* There should be no instances of a mocked class of `MockClass` used in place of `Context` in tests

## Changelog entry

* N/A

felixarntz:
IB ✅ 😄

aaemnnosttv:
@felixarntz I've updated the ACs here to be less implementation-specific and accommodate the changes we agreed on in #2125 while preserving the original purpose of the issue.

tofumatt:
QA ✅

PR #2125, opened 2020-10-02
Enhance/2123 test amp context

aaemnnosttv:
## Summary

Addresses issue #2123 

## Relevant technical choices

* Replaced usage of mocked `Context` for `is_amp` to return `false` with normal `Context` instances as these will always return `false` since our tests do not load the AMP plugin even if it is present

## Checklist


tests/phpunit/includes/FakeAMPContextPrimary.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+namespace Google\Site_Kit\Tests;
+
+use Google\Site_Kit\Context;
+
+/**
+ * Class for faking a primary AMP context.
+ */
+class FakeAMPContextPrimary extends Context {
```
felixarntz, line 18: How about `FakePrimaryAMPContext` for the name instead? I think that's more clear because the class is still a `Context`.

tests/phpunit/includes/FakeAMPContextSecondary.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+namespace Google\Site_Kit\Tests;
+
+use Google\Site_Kit\Context;
+
+/**
+ * Class for faking a secondary AMP context.
+ */
+class FakeAMPContextSecondary extends Context {
```
felixarntz, line 18: See above, I think this would be better as `FakeSecondaryAMPContext`.

felixarntz (commented):
LGTM, just two suggestions for more clear naming.

aaemnnosttv:
@felixarntz I'm not too attached to a particular name, but I kind of like how these are together 😄 
image

If `Context` at the end is important to you, then we could name it slightly differently. How about:
- `FakeAMPContext_Primary`
- `FakeAMPContext_Secondary`

I kind of like that better, or we could do:
* `FakeContext_AMPPrimary`
* `FakeContext_AMPSecondary`

OR we could use `Context` as a namespace segment 😄 
* `Google\Site_Kit\Tests\Context\FakeAMP_Primary`
* `Google\Site_Kit\Tests\Context\FakeAMP_Secondary`
That may not be as clear in use though.

As a single class name, I prefer the first of these 3 alternatives but we can go with whichever one you'd like, including your original suggestion; just offering a few other ideas here.

felixarntz:
@aaemnnosttv How about naming them like I suggested but also putting them into a `Context` sub-namespace like you suggested? I think this would provide most clarity both when looking at the files and when looking at a simple class name reference in a file (`*Context` makes it clear that this is _a_ `Context`).

aaemnnosttv:
@felixarntz That sounds good to me.

One other option that I didn't think of previously is to use the mock builder (again) but we can actually mock `Context` directly now that it isn't `final` and we'd essentially get the same thing without the need for named classes 🤔 

I still like the idea of having a ready-to-use AMP-specific context instances, but we could also create this from a trait or from a method on `TestCase` . E.g. `::get_amp_primary_context() : Context`

I think that's all the possible variants to consider 😄  – let me know which way you'd like to go and I'll update accordingly 👍

felixarntz:
@aaemnnosttv 
> I still like the idea of having a ready-to-use AMP-specific context instances, but we could also create this from a trait or from a method on `TestCase` . E.g. `::get_amp_primary_context() : Context`

How about both of that? :) I like the idea of the classes, but instead of directly referencing the class names everywhere we could use methods like you're saying on `TestCase`. This way we don't couple all our test code to these class names and could for example easily replace it with a `MockBuilder` approach if we wanted to in the future.

aaemnnosttv:
> I like the idea of the classes, but instead of directly referencing the class names everywhere we could use methods like you're saying on `TestCase`. This way we don't couple all our test code to these class names and could for example easily replace it with a `MockBuilder` approach if we wanted to in the future.

@felixarntz I agree; updated! 😄

tests/phpunit/integration/Modules/AdSenseTest.php
```diff
@@ -80,13 +75,8 @@ public function test_register_template_redirect_amp() {
 	}
 
 	public function test_register_template_redirect_non_amp() {
-		$context      = new Context( GOOGLESITEKIT_PLUGIN_MAIN_FILE );
-		$mock_context = $this->getMockBuilder( 'MockClass' )->setMethods( array( 'is_amp', 'input' ) )->getMock();
-		$mock_context->method( 'input' )->will( $this->returnValue( $context->input() ) );
-		$mock_context->method( 'is_amp' )->will( $this->returnValue( false ) );
-
+		$context = new Context( GOOGLESITEKIT_PLUGIN_MAIN_FILE );
```
felixarntz, line 78: Maybe still worth introducing some kind of "no AMP" `Context` mock in `TestCase` because this here is somewhat implicit and ties into the real behavior of `Context` which could be considered somewhat error-prone. But nothing critical for now.

felixarntz (approved):
Great!

Issue #2001, opened 2020-09-10
Show error on single URL details view if no entity is matched


felixarntz:
Before we implemented proper entity detection in Site Kit (see #174), the single URL details view worked for any URL part of the site, or even URLs not part of the site for certain services that don't incorporate permission management (such as PageSpeed Insights).

However, now the single URL details view relies on entity detection to find out more information of the "entity" based on the given URL. This means that, if a single URL details view is accessed, it can and should only provide meaningful results if an entity is detected.

This issue should modify the single URL details view accordingly so that, if no current entity is available, a corresponding message is displayed, instead of still attempting to show data (for the new widgets, doing so would fall back to showing site-wide data, which would be just as irrelevant on that screen).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* On the single URL details view, if there is no current entity (e.g. if the `getCurrentEntityURL` selector of the `core/site` store returns `null`), the screen should show a corresponding message instead of all the regular UI:
    * The "Detailed Page Stats" heading and the back link should be kept.
    * The date range selector should not be there.
    * The white box that would normally have the page title and URL should include the following message: _It looks like the URL %s is not part of this site, therefore there is no data available to display._ (`%s` should be the URL taken from the `permaLink` query parameter)
    * Everything below that box should not be displayed.

## Implementation Brief

* Extract the inner Entity-dependent elements from `DashboardDetailsApp` into a new `DashboardDetailsEntityView` component in `assets/js/components/dashboard-details/`
    - Starting from the `div` containing `<DateRangeSelector />`
    - Ending with `<DashboardDetailsModules />`, inclusive
* Add a new `DashboardDetailsEntityNotFoundView` component as a variant (but not copy of) of `DashboardDetailsEntityView`
    - renders with the following title/url portion https://github.com/google/site-kit-wp/blob/f244c63925cd56d0bb03a048d0b387824c335fe5/assets/js/components/dashboard-details/dashboard-details-app.js#L85-L109
        - The `h3` and `Link` components should be replaced by the message in the AC rendered as plain text (i.e. `<p>`):  
        _It looks like the URL %s is not part of this site [...]_
        - The URL in this block should be sourced from the full URL passed in the `permaLink` query parameter of `global.location.href`  
* Update `DashboardDetailsApp` to conditionally render the appropriate component based on the value of `getCurrentEntityURL`
    - This URL is provided from the server response so there is no resolver to wait for – it will be available immediately
[...]

PR #2127, opened 2020-10-05
Bug/2001 error on single url

eugene-manuilov:
## Summary

Addresses issue #2001

## Checklist

### DIFF
--- a/assets/js/components/dashboard-details/DashboardDetailsEntityHeaderContainer.js
+++ b/assets/js/components/dashboard-details/DashboardDetailsEntityHeaderContainer.js
@@ -1,0 +1,52 @@
+/**
+ * DashboardDetailsEntityHeaderContainer component.
+ *
+ * Site Kit by Google, Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * External dependencies
+ */
+import PropTypes from 'prop-types';
+
+/**
+ * Internal dependencies
+ */
+import Link from '../link';
+import Layout from '../layout/layout';
+
+export default function DashboardDetailsEntityHeaderContainer( { children, url } ) {
+	return (
+		<div className="mdc-layout-grid__cell mdc-layout-grid__cell--span-12">
+			<Layout>
+				<div className="mdc-layout-grid">
+					<div className="mdc-layout-grid__inner">
+						<div className="mdc-layout-grid__cell mdc-layout-grid__cell--span-12">
+							{ children }
+							<Link href={ url } inherit external>
+								{ url }
+							</Link>
+						</div>
+					</div>
+				</div>
+			</Layout>
+		</div>
+	);
+}
+
+DashboardDetailsEntityHeaderContainer.propTypes = {
+	url: PropTypes.string.isRequired,
+	children: PropTypes.node.isRequired,
+};
--- a/assets/js/components/dashboard-details/DashboardDetailsEntityHeaderContainer.js
+++ b/assets/js/components/dashboard-details/DashboardDetailsEntityHeaderContainer.js
@@ -27,7 +27,7 @@ import PropTypes from 'prop-types';
 import Link from '../link';
 import Layout from '../layout/layout';
 
-export default function DashboardDetailsEntityHeaderContainer( { children, url } ) {
+export default function DashboardDetailsEntityHeaderContainer( { children, permalink } ) {
 	return (
 		<div className="mdc-layout-grid__cell mdc-layout-grid__cell--span-12">
 			<Layout>
@@ -35,8 +35,8 @@ export default function DashboardDetailsEntityHeaderContainer( { children, url }
 					<div className="mdc-layout-grid__inner">
 						<div className="mdc-layout-grid__cell mdc-layout-grid__cell--span-12">
 							{ children }
-							<Link href={ url } inherit external>
-								{ url }
+							<Link href={ permalink } inherit external>
+								{ permalink }
 							</Link>
 						</div>
 					</div>
@@ -47,6 +47,6 @@ export default function DashboardDetailsEntityHeaderContainer( { children, url }
 }
 
 DashboardDetailsEntityHeaderContainer.propTypes = {
-	url: PropTypes.string.isRequired,
+	permalink: PropTypes.string.isRequired,
 	children: PropTypes.node.isRequired,
 };
--- a/assets/js/components/dashboard-details/DashboardDetailsEntityHeaderContainer.js
+++ b/assets/js/components/dashboard-details/DashboardDetailsEntityHeaderContainer.js
@@ -24,10 +24,9 @@ import PropTypes from 'prop-types';
 /**
  * Internal dependencies
  */
-import Link from '../link';
 import Layout from '../layout/layout';
 
-export default function DashboardDetailsEntityHeaderContainer( { children, permalink } ) {
+export default function DashboardDetailsEntityHeaderContainer( { children } ) {
 	return (
 		<div className="mdc-layout-grid__cell mdc-layout-grid__cell--span-12">
 			<Layout>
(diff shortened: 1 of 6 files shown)


stories/dashboard-details.stories.js
```diff
@@ -0,0 +1,71 @@
...
+
+storiesOf( 'Dashboard Details', module )
+	.add( 'Existing Entity', () => {
+		global.featureFlags = {
+			widgets: {
+				pageDashboard: {
+					enabled: false,
+				},
+			},
+		};
```
aaemnnosttv, line 41: We should be mindful to prevent stories from having side-effects that persist between stories.

Ideally, globals would be reset between each story like we do in other tests but I'm not sure we can do that globally with Storybook. We have some stories that do this already using decorators, or manually in the Story function but I think we can handle this a bit better with a wrapper component.

Let's introduce a wrapper component for forcing specific global values before rendering and restoring the previous values on unmount.

Essentially something like this:

```js
function WithGlobals( { children, globals } ) {
	const previousValues = _.pick( global, globals );
	const restoreGlobals = useCallback( () => Object.assign( global, previousValues ), [] );

	useEffect( () => {
		Object.assign( global, globals )
		return () => restoreGlobals(),
	}, []);

	return children;
}
```

Alternatively, maybe something like this might be a bit better:

```js
function WithEffects( { children, setUp = () => {}, tearDown = () => {} } ) {
	useEffect( () => {
		setUp()
		return tearDown,
	}, []);

	return children;
}
// and then WithGlobals could be based on this
function WithGlobals( { children, globals } ) {
	const previousValues = _.pick( global, globals );
	const restoreGlobals = useCallback( () => Object.assign( global, previousValues ), [] );

	return <WithEffects
		setUp={ () => Object.assign( global, globals ) }
		tearDown={ restoreGlobals }
		children={ children }
	/>
}
```

That way components needing more complex setup/teardown could use the lower-level `WithEffects`.

What do you think?

assets/js/components/dashboard-details/DashboardDetailsEntityNotFoundView.js
```diff
@@ -0,0 +1,62 @@
...
+import { __, sprintf } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import { sanitizeHTML } from '../../util';
+import DashboardDetailsEntityHeaderContainer from './DashboardDetailsEntityHeaderContainer';
+
+export default function DashboardDetailsEntityNotFoundView( { permalink } ) {
+	const currentEntityURL = permalink || ( new URL( global.location.href ) ).searchParams.get( 'permaLink' );
```
aaemnnosttv, line 36: Please use `getQueryArg` from `@wordpress/url` instead.

assets/js/components/dashboard-details/DashboardDetailsEntityNotFoundView.js
```diff
@@ -0,0 +1,62 @@
...
+	);
+}
+
+DashboardDetailsEntityNotFoundView.propTypes = {
+	permalink: PropTypes.string,
+};
+
+DashboardDetailsEntityNotFoundView.defaultProps = {
+	permalink: '',
+};
```
aaemnnosttv, line 62: We don't really need `defaultProps` because the default is handled in the component.

aaemnnosttv (changes requested):
This looks good, there are just a few things to address and suggestions to consider.

assets/js/components/dashboard-details/DashboardDetailsEntityNotFoundView.js
eugene-manuilov: Ok, removed.

assets/js/components/dashboard-details/DashboardDetailsEntityNotFoundView.js
eugene-manuilov: Updated.

stories/dashboard-details.stories.js
```diff
@@ -0,0 +1,71 @@
...
+				currentEntityTitle: 'Test Page',
+				currentEntityURL: 'https://example.com/test-page',
+				siteName: 'My Site Name',
+			} );
+
+			dispatch( CORE_USER ).receiveGetAuthentication( {
+				authenticated: true,
+				requiredScopes: [],
+				grantedScopes: [],
+			} );
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

stories/dashboard-details.stories.js
eugene-manuilov: I like this idea, but for some reason, I can't make it work. `useEffect` hook isn't being triggered and I can't figure out why :/

aaemnnosttv: I'll take a closer look. I didn't test my suggestion there but it seems like it should work.

tofumatt: I'm seeing a similar issue, and I'm not yet sure why. It might be worth exploring that in another issue to not block this—maybe storing the feature flags in context or something less fragile could be a good solution going forward…

tofumatt: Ah yes, of course 😄  The issue is `<DashboardDetailsApp />` will render before the `<WithGlobals>` function. It's not exactly that the `useEffect` _doesn't run_, but that the innermost components will get rendered first, and thus error before the `global.featureFlags` variables are set. If you replace the approach with something that defines the globals before rendering, but _removes_ them on unmount, you can see it works:

```js
function ResetFeatureFlagsOnUnmount( { children } ) {
	useEffect( () => {
		return () => {
			delete global.featureFlags;
		};
	}, [] );

	return children;
}

storiesOf( 'Dashboard Details', module )
	.add( 'Existing Entity', () => {
			provideUserAuthentication( registry );
			provideSiteInfo( registry, {
				currentEntityTitle: 'Test Page',
				currentEntityURL: 'https://example.com/test-page',
			} );
		};
		global.featureFlags = {
			widgets: {
				pageDashboard: {
					enabled: false,
				},
			},
		};

		return (
			<ResetFeatureFlagsOnUnmount>
				<WithTestRegistry callback={ setupRegistry }>
					<DashboardDetailsApp />
				</WithTestRegistry>
			</ResetFeatureFlagsOnUnmount>
		);
	} )
	.add( 'Not Found Entity', () => (
		<DashboardDetailsEntityNotFoundView permalink="https://example.com/test-page" />
	) );
```

will work as expected:



-----

https://imkev.dev/react-rendering-order is a handy blog post to understand component rendering. It's a bit counter-intuitive, but explains why this is happening 😄

I think having a better way of dealing with globals is good, though maybe it means we should be injecting featureFlags into context to make them easier to mock. We should do that in a separate issue though, so I think this PR is good-to-go.

aaemnnosttv: Thanks @tofumatt – I think this might be good to add to the global decorator so we don't have to do it for every story, what do you think?
https://github.com/google/site-kit-wp/blob/71c25b424182cfef6ea115ae1c77b17bdf488a6a/.storybook/config.js#L105-L110

assets/js/components/dashboard-details/DashboardDetailsEntityHeaderContainer.js
```diff
@@ -0,0 +1,52 @@
...
+	return (
+		<div className="mdc-layout-grid__cell mdc-layout-grid__cell--span-12">
+			<Layout>
+				<div className="mdc-layout-grid">
+					<div className="mdc-layout-grid__inner">
+						<div className="mdc-layout-grid__cell mdc-layout-grid__cell--span-12">
+							{ children }
+							<Link href={ permalink } inherit external>
+								{ permalink }
+							</Link>
```
felixarntz, line 40: This link shouldn't be there for the `DashboardDetailsEntityNotFoundView`, so it should be removed from here and become part of `children` instead - per the ACs and IB the message should replace both the title and URL, the URL is already part of the message and should not be linked.

stories/dashboard-details.stories.js
felixarntz: I think adding a Storybook entry for this goes a bit too far here - the UI here is not new, so it's strictly not responsibility of this PR to add it. Because of the involved complexity, both with the above discussion as well as we'd need to provide data for all the widgets to avoid the load flashing, I think it should be removed from this PR.

stories/dashboard-details.stories.js
```diff
@@ -0,0 +1,58 @@
...
+		};
+
+		return (
+			<WithTestRegistry callback={ setupRegistry }>
+				<DashboardDetailsApp />
+			</WithTestRegistry>
+		);
+	} )
+	.add( 'Not Found Entity', () => (
+		<DashboardDetailsEntityNotFoundView permalink="https://example.com/test-page" />
```
felixarntz, line 56: Let's wrap this in a `WithTestRegistry` and display the full `DashboardDetailsApp`. This _is_ new UI, and providing the necessary datastore data for this to work should be trivial.

felixarntz (changes requested):
Looks good for the most part, just a few things.

felixarntz:
I quickly fixed the remaining pieces, and I added basic functionality to the `core/site` store so that we can centrally handle the `permaLink` query parameter (see https://github.com/google/site-kit-wp/pull/2127/commits/1717eb9f9f7dd484e44c21a318f83ce8c49f5c2a). Merging to get the release built.

felixarntz (approved):

Issue #2074, opened 2020-09-18
Allow registering module setup components


felixarntz:
Every Site Kit module's setup component should be registered via a new (optional) `setupComponent` argument which contains the component for the setup.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `registerModule` action in the `core/modules` store should receive a `setupComponent` argument that is optional and by default is set to a new `DefaultModuleSetup` component.
* The `DefaultModuleSetup` should essentially render the skeleton that all existing setup components have, but nothing more (it's for now just an "unused" fallback but it should exist):
    * It should render its output in a `<div className="googlesitekit-setup-module googlesitekit-setup-module--{moduleSlug}" />`.
    * It should include a `<h2 className="googlesitekit-heading-3 googlesitekit-setup-module__title" />` containing the module title.
    * It should include a `<form className="googlesitekit-{moduleSlug}-setup__form" onSubmit={ finishSetup } />`, containing only the `<div className="googlesitekit-setup-module__action" />` with the "Confirm & Continue" button.
* The `SetupWrapper` component should be updated to rely on registered module setup components:
    * It should be refactored to a functional component using hooks (except for relying on `global._googlesitekitLegacyData.setup.moduleToSetup` which shouldn't be refactored yet as part of this).
    * Instead of the `googlesitekit.ModuleSetup-${ slug }` filter, it should rely on the registered module setup component via the `core/modules` store.
    * The `onSettingsPage` and `isEditing` props should be removed from here.
* All Site Kit modules that currently add their settings pages via the `googlesitekit.ModuleSetup-${ slug }` should instead rely on the `registerModule` action from the `core/modules` store, passing only the `setupComponent` for now.

## Implementation Brief

* Create `DefaultModuleSetup.js` in `js/googlesitekit/modules/components` and import it to `datastore/modules.js`
* Add the implementation for `DefaultModuleSetup`details as described in the AC section.
* Add the new `setupComponent` argument to the `registerModule` action with `DefaultModuleSetup` as the default value.
* Refactor the `SetupWrapper` in `assets/js/components/setup/setup-wrapper.js` to a functional component using hooks:
    * `getSiteKitAdminURL`  ==> `select('core/site').getAdminURL()`
* Follow AC for the remainder of the items that require refactoring.
* Update the modules to register their settings pages via the data store rather than the filter in the `index.js` files at the root of each module in`assets/js/modules`

## QA Brief

[...]

Issue #1623, opened 2020-06-01
Enable settings view and settings edit component registration


felixarntz:
## Feature Description

We need to actually allow registration and usage of `SettingsView` and `SettingsEdit` components for modules.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `registerModule` action of the `core/modules` store should be modified:
    * The `settingsComponent` argument should be removed.
    * Instead, optional `settingsViewComponent` and `settingsEditComponent` arguments should be introduced, which should be React components for the module's view and edit component respectively. There should not be any default (i.e. `undefined`).
* The `DefaultModuleSettings` component should be removed.
* Every module that currently uses the `googlesitekit.ModuleSettingsDetails-{$slug}` filter should instead include a dispatch call to `registerModule`, passing its `SettingsView` and `SettingsEdit` component as applicable (some modules have both, some modules have just the view, other modules may have neither).
* A new functional hook-based component `SettingsRenderer` should be introduced:
    * It should receive `slug` (string), `isOpen` (bool), and `isEditing` (bool) props.
    * It should basically mirror the implementation of today's `SettingsMain` component of every module.
    * It should assume `modules/${ slug }` as the module's store name for now.
    * It should retrieve the module's `settingsViewComponent` and `settingsEditComponent` from `core/modules` datastore's `getModule` selector.
* The `SettingsModule` component should be modified:
    * It should be wrapped in the `withSelect` HOC, to receive the registered module's data via the datastore. For now, it should only pass the `hasSettings` prop to the inner component, based on whether there is a `moduleData.settingsEditComponent` (i.e. `hasSettings = !! moduleData.settingsEditComponent`).
    * The `googlesitekit.ModuleSettingsDetails-{$slug}` filter should be removed, and instead it should render the above new `SettingsRenderer` component.
    * The component should **not be further refactored, or as little as possible**.
* The now unused `ModuleSettingsDetails` component should be removed.

## Implementation Brief

* Move `assets/js/modules/analytics/components/settings/SettingsMain.js` to `assets/js/components/settings/SettingsRenderer.js`
    - Rename to `SettingsRenderer.js`
    - Update it to accept a `slug` prop and use a dynamic `STORE_NAME` defined as `modules/${ slug }` (see https://github.com/google/site-kit-wp/issues/1623#issuecomment-708328766)  
    _`STORE_NAME` should be changed to `storeName` now as it is no longer a constant_
[...]

PR #2129, opened 2020-10-05
Enhancement/2074 allow registering module setup components

caseymorrisus:
## Summary

Addresses issue [#2074](https://github.com/google/site-kit-wp/issues/2074)

## Relevant technical choices

* Created `DefaultModuleSetup` component
* Added `registerModule` action with `setupComponent` argument with `DefaultModuleSetup` as the default value
* Refactored `SetupWrapper` to be a functional component and grab module data from registry via `getModule`
* Update modules to register components via `registerModule` rather than using a filter via `googlesitekit.ModuleSetup-${ slug }`
* Fixed stories to account for modules coming from the store

## Checklist


caseymorrisus:
Tests pass locally, failing on server due to flaky tests therefore I am moving this out of draft as I cannot manually re-run Travis CI jobs without dummy commits. Whoever ends up reviewing this will need to re-run Travis to get checks to pass.

assets/js/components/setup/setup-wrapper.js
```diff
@@ -24,58 +24,45 @@ import { delay } from 'lodash';
...
-			</Fragment>
-		);
-	}
-}
-
-class SetupWrapper extends Component {
-	constructor( props ) {
-		super( props );
+export default function SetupWrapper() {
+	const settingsPageURL = useSelect( ( select ) => select( CORE_SITE ).getAdminURL() );
```
eugene-manuilov, line 43:
```suggestion
	const settingsPageURL = useSelect( ( select ) => select( CORE_SITE ).getAdminURL( 'googlesitekit-settings' ) );
```

assets/js/components/setup/setup-wrapper.js
```diff
@@ -94,75 +81,66 @@ class SetupWrapper extends Component {
...
 	}
 
-	render() {
-		const { currentModule } = this.state;
-		const setupModule = SetupWrapper.loadSetupModule( currentModule );
-		const settingsPageURL = getSiteKitAdminURL(
-			'googlesitekit-settings',
-			{}
-		);
+	const setupModule = loadSetupModule();
```
eugene-manuilov, line 84: The `loadSetupModule` function isn't needed anymore. Please, render the `SetupComponent` component directly in the returning markup.

assets/js/components/setup/setup-wrapper.js
```diff
@@ -24,58 +24,45 @@ import { delay } from 'lodash';
...
+				finishSetup={ finishSetup }
 			/>
 		);
 	}
 
 	/**
 	 * When module setup done, we redirect the user to Site Kit dashboard.
 	 */
-	static finishSetup() {
+	function finishSetup() {
```
eugene-manuilov, line 65: Update this function to be created with `useCallback` and `[]` dependencies to avoid unnecessary re-renders.

assets/js/components/setup/setup-wrapper.js
```diff
@@ -94,75 +81,66 @@ class SetupWrapper extends Component {
...
+													mdc-layout-grid__cell
+													mdc-layout-grid__cell--span-2-phone
+													mdc-layout-grid__cell--span-4-tablet
+													mdc-layout-grid__cell--span-6-desktop
 												">
-													{ __( 'Connect Service', 'google-site-kit' ) }
-												</p>
-												{ setupModule }
+												<Link
+													id={ `setup-${ moduleToSetup }-cancel` }
```
eugene-manuilov, line 123: Please, use `module.slug` instead of `moduleToSetup`:

```suggestion
													id={ `setup-${ module.slug }-cancel` }
```

eugene-manuilov (changes requested):
Nice work, @caseymorrisus! Left a few comments for you. Could you, please, address it?

assets/js/googlesitekit/modules/components/DefaultModuleSetup.js
```diff
@@ -0,0 +1,57 @@
...
+					</Button>
+				</div>
+			</form>
+		</div>
+	);
+}
+
+DefaultModuleSetup.propTypes = {
+	finishSetup: PropTypes.func.isRequired,
+	module: PropTypes.shape( {
```
eugene-manuilov, line 53: @caseymorrisus Let's make `module` required and remove its default value.

eugene-manuilov (changes requested):
@caseymorrisus LGTM! Just left a final suggestion, let's update it and I'll approve this PR.

assets/js/googlesitekit/modules/components/DefaultModuleSetup.js
caseymorrisus: @eugene-manuilov I'm not confident that we should do this as the `module` prop is a passthrough from a `getModule` selector which has a return type of `Maybe(Object)` as it returns `undefined` if state is loading or the requested module does not exist. If this selector returns `undefined` as it says it will, this will throw `Uncaught TypeError: Cannot read property 'slug' of undefined` unexpectedly.

eugene-manuilov: Ah... yes, you are right. Let's reject my request.

eugene-manuilov (approved):

eugene-manuilov:
@caseymorrisus  I have approved this PR, but could you please resolve merge conflicts?

eugene-manuilov (approved):

assets/js/googlesitekit/modules/components/DefaultModuleSetup.js
```diff
@@ -0,0 +1,57 @@
...
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import Button from '../../../components/button';
+
+export default function DefaultModuleSetup( { finishSetup, module = {} } ) {
```
felixarntz, line 34: I think we should use minimal props here like we usually do and only pass the `moduleSlug` which is sufficient to identify the module. Then this component can look up the module object to get `module.name` from the datastore.

assets/js/googlesitekit/modules/components/DefaultModuleSetup.js
```diff
@@ -0,0 +1,57 @@
...
+		</div>
+	);
+}
+
+DefaultModuleSetup.propTypes = {
+	finishSetup: PropTypes.func.isRequired,
+	module: PropTypes.shape( {
+		name: PropTypes.string,
+		slug: PropTypes.string,
+	} ),
```
felixarntz, line 56: See above. Passing a slug is more straightforward here, I think it's a bit odd to expect an object here, especially since the shape provided here is not the full module object - it shouldn't depend too much on the implementation of the component, let's require the module slug, from there we can get anything else we could possibly need via the datastore.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -221,6 +223,7 @@ const baseActions = {
 	 * @param {number}    [settings.order]             Optional. Numeric indicator for module order. Default 10.
 	 * @param {string}    [settings.homepage]          Optional. Module homepage URL. Default empty string.
 	 * @param {WPElement} [settings.settingsComponent] React component to render the settings panel. Default is the DefaultModuleSettings component.
+	 * @param {WPElement} [settings.setupComponent]    React component to render the setuppanel. Default is the DefaultModuleSetup component.
```
felixarntz, line 226:
```suggestion
	 * @param {WPElement} [settings.setupComponent]    React component to render the setup panel. Default is the DefaultModuleSetup component.
```

assets/js/modules/adsense/index.js
```diff
@@ -113,4 +108,10 @@ domReady( () => {
 			AREA_DASHBOARD_EARNINGS,
 		],
 	);
+	Modules.registerModule(
+		'adsense',
+		{
+			setupComponent: SetupMain,
+		},
+	);
```
felixarntz, line 116: Super-nitpick, but maybe we should put the `registerModule` call above the widget registrations, just because module registration is the more general step. Of course this doesn't technically change anything, but without module registration widget registration is mostly useless (most widgets only are active if their respective module is registered and active).

This also applies to the other similar files.

stories/module-adsense-setup.stories.js
```diff
@@ -77,351 +61,299 @@ const accountCompleteSettings = {
...
-		const setupRegistry = ( registry ) => {
-			registry.dispatch( STORE_NAME ).receiveGetSettings( defaultSettings );
-			registry.dispatch( STORE_NAME ).receiveIsAdBlockerActive( false );
-		};
-
-		return <Setup callback={ setupRegistry } />;
+	.addDecorator( ( storyFn ) => {
+		global._googlesitekitLegacyData.setup.moduleToSetup = 'adsense';
+		const registry = createTestRegistry();
+		registry.dispatch( CORE_MODULES ).receiveGetModules( [
```
felixarntz, line 67: I think it would be worth switching this and similar calls to using the new `provideModules` utility function.

stories/module-analytics-setup.stories.js
```diff
@@ -63,254 +47,235 @@ function usingGenerateGTMAnalyticsPropertyStory( args ) {
...
+	.addDecorator( ( storyFn ) => {
+		global._googlesitekitLegacyData.setup.moduleToSetup = 'analytics';
+		const registry = createTestRegistry();
+		registry.dispatch( CORE_MODULES ).receiveGetModules( [
+			{
+				slug: 'analytics',
+				active: true,
+				connected: true,
+			},
+		] );
```
felixarntz, line 69: See above.

stories/module-optimize-setup.stories.js
```diff
@@ -61,40 +44,46 @@ function Setup( props ) {
...
+	.addDecorator( ( storyFn ) => {
+		const registry = createTestRegistry();
+		global._googlesitekitLegacyData.setup.moduleToSetup = 'optimize';
+		registry.dispatch( CORE_MODULES ).receiveGetModules( [
+			{
+				slug: 'optimize',
+				active: true,
+				connected: true,
+			},
+		] );
```
felixarntz, line 56: See above.

stories/module-optimize-setup.stories.js
```diff
@@ -61,40 +44,46 @@ function Setup( props ) {
...
 
-		const setupRegistry = ( { dispatch } ) => {
-			dispatch( CORE_MODULES ).receiveGetModules( analyticsFixture );
-			dispatch( MODULES_ANALYTICS ).setUseSnippet( true );
-			dispatch( STORE_NAME ).receiveGetSettings( {} );
-		};
+		return storyFn( registry );
+	} )
+	.add( 'Start', ( registry ) => {
+		registry.dispatch( CORE_MODULES ).receiveGetModules( analyticsFixture );
```
felixarntz, line 64: This is somewhat incorrect (or at least confusing) because it will set the modules so that only Analytics exists and is active, overriding the data provided in the decorator that makes Optimize the only available and active module. Since Optimize requires Analytics to be active, we should probably remove all these clauses and instead in the decorator use `provideModules` with both `analytics` _and_ `optimize`.

stories/modules-tagmanager-setup.stories.js
```diff
@@ -61,7 +46,18 @@ function Setup( props ) {
...
 	.addDecorator( ( storyFn ) => {
+		global._googlesitekitLegacyData.setup.moduleToSetup = 'tagmanager';
 		const registry = createTestRegistry();
+		registry.dispatch( CORE_MODULES ).receiveGetModules( [
+			{
+				slug: 'tagmanager',
+				active: true,
+				connected: true,
+			},
+		] );
```
felixarntz, line 57: See above.

stories/modules-tagmanager-setup.stories.js
```diff
@@ -21,37 +21,22 @@
  */
```
felixarntz, line 21: Similar to AdSense, something here seems to mess up several Tag Manager setup stories, e.g.:
Screenshot 2020-10-12 at 15 44 19

stories/utils/generate-gtm-analytics-property-story.js
```diff
@@ -48,9 +48,9 @@ export function generateGTMAnalyticsPropertyStory( {
 	setUp = () => {},
 } ) {
 	return () => {
-		setUp();
-
 		const setupRegistry = ( registry ) => {
+			setUp( registry );
```
felixarntz, line 52: While resolving a merge conflict which was around this logic, I moved the `setUp` function in here so that it can run logic on the registry, which was needed (`registerModule` is registry-specific, while the previous filter-based mechanism wasn't).

assets/js/components/setup/setup-wrapper.js
```diff
@@ -96,77 +68,75 @@ class SetupWrapper extends Component {
 		delay( function() {
 			global.location.replace( redirectURL );
 		}, 500, 'later' );
+	}, [] );
```
felixarntz, line 71: I think here we should pass `moduleToSetup` from above as callback dependency and then use it in the callback, instead of relying on the same global again, which may in theory cause inconsistent callbacks.

assets/js/components/setup/setup-wrapper.js
```diff
@@ -24,53 +24,25 @@ import { delay } from 'lodash';
...
-			<FilteredModuleSetup
-				finishSetup={ SetupWrapper.finishSetup }
-				onSettingsPage={ false }
-				isEditing={ true }
-			/>
-		);
-	}
+export default function SetupWrapper() {
+	const settingsPageURL = useSelect( ( select ) => select( CORE_SITE ).getAdminURL( 'googlesitekit-settings' ) );
+	const { moduleToSetup } = global._googlesitekitLegacyData.setup;
```
felixarntz, line 44: This was explicitly not required as part of the ACs, but maybe it would be worth moving this legacy dependency out of this component and require it as a prop instead (`moduleSlug`). This would for example simplify the Storybook configuration so that we no longer need to mess with the global there which is error-prone.

We could instead do the `global._googlesitekitLegacyData.setup.moduleToSetup` lookup in the components using `SetupWrapper` and pass the value as prop, so that at least this component is fully refactored (let's not worry about the other ones):
* `DashboardSplashApp`
* `GoogleSitekitDashboard`
* `GoogleSitekitModule`

felixarntz (changes requested):
The production code mostly LGTM, but this seems to break some Tag Manager stories.

caseymorrisus:
@felixarntz Failed CI tests have been confirmed to pass (multiple times) locally. Please re-run tests when re-reviewing to verify.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -230,6 +233,7 @@ const baseActions = {
 		order,
 		homepage,
 		settingsComponent = DefaultModuleSettings,
+		setupComponent = DefaultModuleSetup,
```
felixarntz, line 236: Revisiting this, we may want to diverge from the ACs here in that there should not be any default value for the `setupComponent` argument. Per #1623, this will be changed for the `settingsComponent` replacements as well. Providing a default here prevents us from checking whether the module has a setup component attached or not - which will come in handy in the future, e.g. when we need to determine whether we need to link to the module setup screen or the regular module page (linking to the setup screen when there is no setup component would not make sense). There is currently some hard-coded exception in place for PageSpeed Insights (which doesn't require a setup) for that very reason - which we can eliminate by checking whether the module has a setup component.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -221,6 +223,7 @@ const baseActions = {
 	 * @param {number}    [settings.order]             Optional. Numeric indicator for module order. Default 10.
 	 * @param {string}    [settings.homepage]          Optional. Module homepage URL. Default empty string.
 	 * @param {WPElement} [settings.settingsComponent] React component to render the settings panel. Default is the DefaultModuleSettings component.
+	 * @param {WPElement} [settings.setupComponent]    React component to render the setup panel. Default is the DefaultModuleSetup component.
```
felixarntz, line 226: Based on my above comment, this could then say "Default none.".

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -59,6 +60,7 @@ const moduleDefaults = {
 	order: 10,
 	icon: null,
 	settingsComponent: DefaultModuleSettings,
+	setupComponent: DefaultModuleSetup,
```
felixarntz, line 63: See above, the default should be `undefined`.

assets/js/components/setup/setup-wrapper.js
```diff
@@ -96,77 +68,79 @@ class SetupWrapper extends Component {
...
 	}
 
-	render() {
-		const { currentModule } = this.state;
-		const setupModule = SetupWrapper.loadSetupModule( currentModule );
-		const settingsPageURL = getSiteKitAdminURL(
-			'googlesitekit-settings',
-			{}
-		);
+	const { setupComponent: SetupComponent } = module;
```
felixarntz, line 77: Based on my above comment, we could either remove the `DefaultModuleSetup` component entirely (it strictly isn't needed), or we could use it here internally to render it if `SetupComponent` is `undefined`. Since this screen would in the future never be linked to if there is no `setupComponent`, it may also be okay just rendering nothing in that case.

felixarntz (commented):
@caseymorrisus Against the ACs, this is good to go. However, based on some changes in the approach we decided on for the related #1623, I suggested some changes here (see specific comments). @aaemnnosttv What do you think about my alternative approach suggested here?

assets/js/googlesitekit/modules/datastore/modules.js
aaemnnosttv: Yeah, I think that makes sense. A `DefaultModuleSetup` component should be capable of setting up the module, but setup is very module-specific so I don't think it's possible to have such a thing in our case as we can't really assume anything about how every module works to have a generic setup component.

`setupComponent` can default to `null` if not passed and is what I have in #1623 for settings components so far (almost ready for review, but you can see it here: https://github.com/google/site-kit-wp/pull/2202/files#diff-51079bbb133e20fa794c0e1fd5fd30b551dda818e0cfe0ba4162502670cf133a)

assets/js/googlesitekit/modules/datastore/modules.js
aaemnnosttv: I chose to go with `null` for the settings components to indicate that it is intentionally empty as we do elsewhere rather than `undefined` which is more of an initial state or unresolved value. Perhaps we're not consistent with this, but both values should work for our needs. If you would prefer to use `undefined` instead, that's an easy change to make for me.

felixarntz (approved):
Awesome work!

PR #1971, opened 2020-08-28
Enhancement/317 add date filters for adsense

ryanwelcher:
## Summary

Addresses issue #317

## Checklist

- [x] My code is tested and passes existing unit tests.
- [x] My code has an appropriate set of unit tests which all pass.
- [x] My code is backward-compatible with WordPress 4.7 and PHP 5.6.
- [x] My code follows the [WordPress](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) coding standards.
- [x] My code has proper inline documentation.
- [x] I have signed the Contributor License Agreement (see <https://cla.developers.google.com/>).

Issue #317, opened 2019-07-21
Add date filters for relevant metrics in AdSense dashboard


lilybonney:
## Feature Description

In the Search Console and Analytics dashboard there is a date filter:

![Date Filter](https://user-images.githubusercontent.com/52471748/61595572-b9402400-abf0-11e9-8325-ea153697a3b6.png)

However, this is not present in the AdSense dashboard screen. The "Estimated Earnings" widget doesn't need date filter, but the other two statistics on this screen would benefit from it.

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a date picker on top of the AdSense module page, similar like it is on e.g. the Analytics module page.
* The "Estimated Earnings" widget should be removed entirely.
* The first "Performance over previous 28 days" should be updated:
    * Title should change to "Performance over the last 28 days" (suffix should be based on current date range like in e.g. Analytics module page), for consistency with date picker and similar headings in e.g. Analytics module page.
    * There should be two additional columns in the widget:
        * The "Earnings" value for the current period should be the first column, before "Page RPM".
        * The "Page CTR" value for the current period should be the fourth column, after "Impressions" (the AdSense metric is called `PAGE_VIEWS_CTR`).
    * The values (i.e. the respective queries) should be based on the currently selected date range.
    * "Advanced Settings" should change to "See full stats in AdSense".
* The second "Performance over previous 28 days" should be updated:
    * Title should change to "Performance by page over the last 28 days" (suffix should be based on current date range like in e.g. Analytics module page).
    * The values (i.e. the respective queries) should be based on the currently selected date range.
    * "Advanced Settings" should change to "See full stats in AdSense".

## Implementation Brief
* In `assets/js/modules/adsense/index.legacy.js` add the `googlesitekit.showDateRangeSelector-${ slug }` to return true to display the `DateRangeSelector` component in the AdSense page.
* Wrap the `AdSenseDashboardWidget` with the `withSelect` HOC from `Data` that  adds the `dateRange` prop using the `getDateRange` selector from `core/user`
* import the `getCurrentDateRange`  to use for displaying the date range in other components.
* Remove the `<Layout />` component from  `AdSenseDashboardWidget` component that represents the Estimated earnings section.
* Update the titles for each remaining `Layout` components to reflect date range as per the AC.
* Refactor the `AdSensePerformanceWidget` to be a functional component.
[...]

PR #2135, opened 2020-10-06
Enhancement/317 add date filters redux

ryanwelcher:
## Summary
This PR is meant to replace #1971.

Addresses issue #317 

## Relevant technical choices
In order to not refactor `<AdSensePerformaceWidget />` to use hook, an edge case was added to the legacy data API to return data using the` prev-**-days` date range. 

## Checklist


assets/js/modules/adsense/components/dashboard/AdSenseDashboardWidget.js
```diff
@@ -182,17 +183,6 @@ class AdSenseDashboardWidget extends Component {
...
-								<Layout
-									header
-									title={ __( 'Estimated earnings', 'google-site-kit' ) }
-									headerCtaLabel={ __( 'Advanced Settings', 'google-site-kit' ) }
-									headerCtaLink={ homepage }
-								>
-									<AdSenseEstimateEarningsWidget
-										handleDataError={ this.handleDataError }
-										handleDataSuccess={ this.handleDataSuccess }
-									/>
```
felixarntz, line 194: While this component and its usage should be removed, the removal of passing the `handleDataError` and `handleDataSuccess` props still causes the same problem here that I mentioned in the previous PR: If the AdSense API returns only 0 numbers, this `handleDataError` needs to be triggered, otherwise `handleDataSuccess` needs to be triggered. I assume because this was no longer happening, this is why you removed `loading` above, which shouldn't be the case though; we need to maintain this behavior. What you should see if you have only 0s in your AdSense data is a prompt informing you that you might need to wait to get impressions instead of the regular screen UI - this is what this logic is used for.

Since the `handleDataSuccess` and `handleDataError` props are automatically handled in the `withData` HOC that is also used for an AdSense API request in the `AdSensePerformanceWidget`, you should be able to pass the props to that component below.

felixarntz (changes requested):
This looks solid for the most part, however this PR still breaks one important thing: When your AdSense data is only 0s (see `isAdSenseDataZero` function), you shouldn't see the regular screen UI, but instead only this:
Screenshot 2020-10-07 at 15 08 23

assets/js/modules/adsense/components/dashboard/AdSensePerformanceWidget.js
```diff
@@ -71,23 +72,38 @@ class AdSensePerformanceWidget extends Component {
...
-				change: ! isUndefined( prev28Days.totals ) ? prev28Days.totals[ 2 ] : 0,
+				datapoint: readableLargeNumber( currentRangeData.totals[ 2 ] ),
+				change: ( ! isUndefined( prevRangeData.totals ) ) ? changeToPercent( prevRangeData.totals[ 2 ], currentRangeData.totals[ 2 ] ) : 0,
+				changeDataUnit: '%',
+			},
+			{
+				className: 'googlesitekit-data-block--impression',
+				title: __( 'Page CTR', 'google-site-kit' ),
+				/* translators: %s: percentage value. */
+				datapoint: sprintf( _x( ' %1$s%%', 'AdSense performance Page CTA percentage', 'google-site-kit' ), currentRangeData.totals[ 3 ] * 100 ),
```
ryanwelcher, line 105: @felixarntz just wanted to call this out. The formatting of numbers followed by % signs can be different per language. For example, English is 50% while many other languages would display as 50 %.

felixarntz: @ryanwelcher Thanks for pointing this out. Let's open a follow-up issue to address this throughout the plugin. We have many cases where we pass the unit independently, so this would involve a bit more work.

includes/Modules/AdSense.php
```diff
@@ -810,16 +810,12 @@ private function date_range_to_dates( $date_range ) {
...
-					gmdate( 'Y-m-d', strtotime( '8 days ago' ) ),
-				);
+			case 'prev-14-days':
 			case 'prev-28-days':
-				return array(
-					gmdate( 'Y-m-d', strtotime( '56 days ago' ) ),
-					gmdate( 'Y-m-d', strtotime( '29 days ago' ) ),
-				);
+			case 'prev-90-days':
+				return $this->parse_date_range( $date_range, 1, 1, true );
```
felixarntz, line 818: Nice!

felixarntz (approved):
Awesome!

assets/js/modules/adsense/components/dashboard/AdSensePerformanceWidget.js
ryanwelcher: @felixarntz I've opened #2200 to track this.

Issue #2021, opened 2020-09-14
Only render widget area when there are widgets


felixarntz:
There is one minor thing still missing in the new widget API, which is correct handling of widget areas without widgets. At the moment, widget areas via the new API and components are always rendered, regardless of whether they actually include active widgets. Widget areas should only be rendered if they include active widgets.

This should both consider whether a widget area has widgets overall (this should be obvious), but also whether it has active widgets (i.e. if all widgets of an area return `null`, the widget area should be omitted as well).

Follow-up to #1300 and #1678.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A widget area via the new API should only be rendered if there are actual widgets to display inside it (i.e. **active widgets**).

## Implementation Brief

Edit `WidgetAreaRenderer` within `js/googlesitekit/widgets/components/WidgetAreaRenderer.js`

* Return `null` if `widgets.length === 0`
* Refactor `activeWidgets` to be handled directly within `WidgetAreaRenderer` without passing `activeWidgets` and `setActiveWidgets` to `WidgetRenderer`
  * Replace use of `useState`, set value of `activeWidgets` using `widgets.filter( isActiveWidget )`
    * See comment for `isActiveWidget` filter implementation: https://github.com/google/site-kit-wp/issues/2021#issuecomment-698894816
* Return `null` if `activeWidgets.length === 0`
* Modify `widgetClassNames` to account for `activeWidgets` being an array rather than an object
  * Use `activeWidgets.length` within `[].fill( ... )` call
  * Iterate over `activeWidgets` via `activeWidgets.forEach`
    * Can safely remove conditional & early return that checks if active widget as array only includes active widgets
* Modify `widgetOutput` to map over `activeWidgets` rather than `widgets`
* Modify `widgetOutput`, remove `activeWidgets` and `setActiveWidgets` props from `WidgetRenderer` component

Edit `WidgetRenderer` within `js/googlesitekit/widgets/components/WidgetRenderer.js`

* Remove use and logic associated with `activeWidgets` & `setActiveWidgets`, there are 2 conditionals that call `setActiveWidgets` with early `null` returns, these are no longer necessary as non-active widgets are not rendered.

## QA Brief

* Activate Site Kit in a development build without the AdSense module active.
* Prior to this fix, the **new** widget areas would include an empty "Earnings" area.
* After this fix, the "Earnings" area should not be present because it doesn't include actual content.
[...]

PR #2138, opened 2020-10-07
Enhancement/2021 only render widget area when active widgets exist

caseymorrisus:
## Summary

Addresses issue [#2021](https://github.com/google/site-kit-wp/issues/2021)

## Relevant technical choices

Followed IB except that returning `null` when `widgets.length === 0` is not needed as the earliest call site of this is the same location where we can assert against `activeWidget.length === 0`. This is to avoid conditionally calling the `useMemo` hook.

## Checklist


caseymorrisus:
Test failures appear to be unrelated (same flaky failures in other PRs), all tests pass locally. Whoever reviews this will need to re-run Travis CI jobs to verify checks pass as I do not have access to re-run them myself.

assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js
```diff
@@ -155,16 +154,18 @@ const WidgetAreaRenderer = ( { slug } ) => {
 		}
 
 		return classNames;
-	}, [ widgets, activeWidgets ] );
+	}, [ activeWidgets ] );
+
+	if ( activeWidgets.length === 0 ) {
```
eugene-manuilov, line 159: @caseymorrisus let's move this conditional up to come after `const activeWidgets = widgets.filter( isActiveWidget );`.

eugene-manuilov (changes requested):
@caseymorrisus this mostly looks good, just added a few comments for you to address.

assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js
```diff
@@ -155,16 +154,18 @@ const WidgetAreaRenderer = ( { slug } ) => {
 		}
 
 		return classNames;
-	}, [ widgets, activeWidgets ] );
+	}, [ activeWidgets ] );
```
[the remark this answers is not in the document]
caseymorrisus: @eugene-manuilov Good catch, `useMemo` is not relevant here because we're creating a new array on every render so we're unable to check dependencies based on referential equality. If we were able to abstract the `activeWidgets` filtering logic into a selector we'd be able to use the `useMemo` optimization as `useSelector` returns a cached result which allows for a referential equality check (returns the same reference if unchanged). Unfortunately how we check for a null render causes issues with hooks rendering order (throws error) so this is not an option.

assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js
```diff
@@ -106,19 +106,18 @@ const WidgetAreaRenderer = ( { slug } ) => {
...
 	const widgets = useSelect( ( select ) => select( STORE_NAME ).getWidgets( slug ) );
 
-	// State handled by WidgetRenderer instances, based on whether the widget
-	// renders content or `null`.
-	const [ activeWidgets, setActiveWidgets ] = useState( {} );
+	// Verify that widgets are active (do not render null)
+	const isActiveWidget = ( widget ) =>
+		widgets.find( ( item ) => item.slug === widget.slug ) &&
+		typeof widget.component === 'function' &&
+		widget.component( {} );
```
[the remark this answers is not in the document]
caseymorrisus: @eugene-manuilov This is a good concern to call out as this will fail if presented with a class component. Using `createElement` directly is unnecessary, we could avoid the import by using JSX directly (this gets changed to createElement under the hood, it's syntatical sugar) via `<widget.component />`. Unfortunately the return value of `createElement` is an object which does not contain the return value of the component, which is what we need to assert against.

assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js
```diff
@@ -106,19 +101,22 @@ const WidgetAreaRenderer = ( { slug } ) => {
...
-			if ( ! activeWidgets[ widget.slug ] ) {
-				return;
-			}
+	const activeWidgets = widgets.filter( isActiveWidget );
 
+	if ( activeWidgets.length === 0 ) {
+		return null;
+	}
+
+	const getWidgetClassNames = () => {
```
eugene-manuilov, line 116: @caseymorrisus let's extract this function:

```js
const getWidgetClassNames = ( activeWidgets ) => {
    ...
};

const WidgetAreaRenderer = ( { slug } ) => {
    ...
    const widgetClassNames = getWidgetClassNames( activeWidgets );
    ...
};
```

eugene-manuilov (changes requested):
@caseymorrisus LGTM, there is just one suggestion and I'll approve this PR.

assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js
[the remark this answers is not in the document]
eugene-manuilov: Yep, that's right. Thanks for updating it!

assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js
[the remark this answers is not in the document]
eugene-manuilov: You are right, @caseymorrisus. What about using [`renderToStaticMarkup()`](https://reactjs.org/docs/react-dom-server.html#rendertostaticmarkup)? It will render a widget and return a string that we can check in the condition.

assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js
```diff
@@ -106,19 +102,22 @@ const WidgetAreaRenderer = ( { slug } ) => {
 	const widgetArea = useSelect( ( select ) => select( STORE_NAME ).getWidgetArea( slug ) );
 	const widgets = useSelect( ( select ) => select( STORE_NAME ).getWidgets( slug ) );
 
-	// State handled by WidgetRenderer instances, based on whether the widget
-	// renders content or `null`.
-	const [ activeWidgets, setActiveWidgets ] = useState( {} );
+	// Verify that widgets are active (do not render null)
+	const isActiveWidget = ( widget ) =>
+		widgets.some( ( item ) => item.slug === widget.slug ) &&
+		typeof widget.component === 'function' &&
```
eugene-manuilov, line 108: @caseymorrisus we don't need this check since we use `renderToStaticMarkup` now.

eugene-manuilov (changes requested):

assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js
caseymorrisus: @eugene-manuilov I think the check for a function here is just as relevant as before (used to throw if `widget.component` was not a function via `undefined is not a function`).

With the new changes, if `widget.component` were not a component, it would throw here with the error: `Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: <INVALID_ELEMENT_TYPE>` as `renderToStaticMarkup` expects a component as an argument. Checking the type prevents this error from being a possibility (can't find a typedef that makes me confident that `widget.component` is always defined/required, seems possible to be undef).

eugene-manuilov: Ok, good.

eugene-manuilov (changes requested):
@caseymorrisus looks like there are some js tests fail now. Could you please check it on your end and fix issues?

eugene-manuilov (approved):

assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js
```diff
@@ -102,69 +97,74 @@ const resizeClasses = ( classNames, counter ) => {
...
+	const widgetArea = useSelect( ( select ) => select( STORE_NAME ).getWidgetArea( slug ) );
+	const widgets = useSelect( ( select ) => select( STORE_NAME ).getWidgets( slug ) );
+
+	// Verify that widgets are active (do not render null)
+	const isActiveWidget = ( widget ) =>
+		widgets.some( ( item ) => item.slug === widget.slug ) &&
+		typeof widget.component === 'function' &&
+		widget.component.prototype.render
+			? new widget.component( {} ).render()
+			: widget.component( {} );
```
felixarntz, line 152: I think we can move this function out of the component, it doesn't need to be declared inside the component itself. Furthermore, for readability I'd suggest not using shorthand syntax for this one because it spans multiple lines and conditionals, so I think it makes it unnecessarily hard to digest.

felixarntz (changes requested):
Great work! The code itself is LGTM, just one minor thing I think we could improve. What would be great to add here is 1-2 Jest tests to ensure this works as expected, e.g.:
1. Render widget area with a single widget which returns `null`, ensuring that the widget area markup isn't rendered.
2. Render widget area with multiple widgets, one of which returns `null`, ensuring that the widget area renders without that widget and grid classes are applied as if that inactive widget didn't exist.

Maybe 2. is a bit overkill, but at least we should have a basic test case covering something like 1.

assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js
caseymorrisus: Abstracted the component out and put individual conditionals into variables to increase readability.

felixarntz (approved):
Awesome!

Issue #2096, opened 2020-09-23
Merge Analytics & Tag Manager decoupling branch into `develop`


felixarntz:
The `feature/ga-gtm-separation` branch should be merged into `develop`.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Tag Manager should no longer depend on Analytics being active. See #1380.
* There should be Analytics-related guidance in the Tag Manager setup flow if an Analytics property is found in the Tag Manager container. See #1381.
* There should be Tag Manager-related guidance in the Analytics setup flow if Tag Manager is active and an Analytics property is found in the Tag Manager container. See #1382.
* Tag Manager should still not be included in the list of modules to activate in the "Congratulations" notifications because it is a more advanced module. See #2025.

## Implementation Brief

* Open PR from `feature/ga-gtm-separation` against `develop`.

## QA Brief

Generally already covered via dependency issues #1380, #1381, #1382, #2025. Still, check these flows at a high level:
* Set up (in Tag Manager frontend) a container that contains an Analytics tag, with the same Analytics property ID you intend to use for Site Kit.
* Test 1: Activate Tag Manager before Analytics. Select the container with the Analytics tag. See prompt informing about Analytics, and continue to Analytics setup. You should get redirected to OAuth consent screen and afterwards the Analytics setup. There, the property that you have in your Tag Manager container should be force-selected. Complete setup.
    * Afterwards, go to Analytics settings and ensure it says "Snippet is not inserted" (because Tag Manager handles it).
* Test 2: Activate Analytics before Tag Manager. Select the property which you have set up in your Tag Manager container. Complete setup. Then, activate Tag Manager. Select the container with the Analytics tag. You should see a notice informing that you already have an Analytics tag setup and that it matches, so that the plugin will use Tag Manager instead of the Analytics snippet from now on. Complete setup.
    * Same as in test 1: Afterwards, go to Analytics settings and ensure it says "Snippet is not inserted" (because Tag Manager handles it).

## Changelog entry

* N/A

felixarntz:
Setting the estimate to 7 mostly due to QA (even though the individual pieces were previously QAd already). Code review should be minimal because all of this has already been reviewed and merged into the feature branch.

aaemnnosttv:
IB ✅  😄

cole10up:
[...]

PR #2140, opened 2020-10-07
Decouple Tag Manager from Analytics

aaemnnosttv:
## Summary

Addresses issue #2096

## Checklist


felixarntz (changes requested):
There's one minor UX flaw here that we should fix: When first setting up Tag Manager with a container that has an Analytics property, and then continuing to the Analytics setup from there, the error that there is a conflicting Analytics property will briefly flash up (for a split second):
Screenshot 2020-10-07 at 17 01 44 (2)

This is likely because at the end of the flow we activate the Analytics module, and of course then the initially set up property won't match. It shouldn't even get to displaying anything like that in the Tag Manager flow though and instead stay "navigating" until the user gets to the Analytics setup.

aaemnnosttv:
@googlebot I consent.

aaemnnosttv:
@felixarntz the previous UX issue is good now, the only remaining issue is the reauthURL not requesting the added scopes.

felixarntz:
@aaemnnosttv The problem with the reauth URL was indeed with timing:
* Module activation triggers a refresh of the user authentication data, which includes required scopes and comparing to the user's available scopes.
* This data is a dependency to the module's reauth URL, which will therefore generally be updated accordingly, _however_: Because the URL is retrieved at the root of the component in a `useSelect`, at the time of executing the callback it is in the state of before the module activation. **It needs to be fresh from right after module activation instead.**
* For this kind of required behavior, it makes sense to provide the data from the action response so that the callback can use it. This can be accomplished with a registry control that selects the data.

See https://github.com/google/site-kit-wp/pull/2140/commits/42d4babeae82a11213e56cc426c91f8272aeb838.

felixarntz (approved):

Issue #2087, opened 2020-09-21
Provide mini API for cookie consent plugins to integrate with


felixarntz:
There are a myriad of WordPress plugins that implement a cookie banner or other way of consent in the frontend of a WordPress sites. Some of these plugins also include functionality to integrate with certain plugins so that features that require the user's consent are not added until that consent is granted. Site Kit should expose a simple layer for such plugins to prevent Site Kit's Google service snippets from being rendered on a page if the user has not granted consent.

While theoretically for Analytics it would be sufficient to rely on the existing `googlesitekit_is_tracking_disabled` filter, there should be a broader solution that prevents output of those snippets entirely.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Filters `googlesitekit_{$slug}_tag_blocked` should be fired before potentially rendering a Google service tag snippet in the frontend for non-AMP, where `$slug` is the service name (`adsense`, `analytics`, `tagmanager`). The filterable parameter should be a boolean (default `false`).
    * All non-AMP tag snippet logic should only be run if the filter value is `false`.
* Filters `googlesitekit_{$slug}_tag_amp_blocked` should be fired before potentially rendering a Google service tag snippet in the frontend for AMP, where `$slug` is the service name (`adsense`, `analytics`, `tagmanager`). The filterable parameter should be a boolean (default `false`).
    * All AMP tag snippet logic should only be run if the filter value is `false`.
* [Since AMP includes a standardized layer for blocking snippets based on a consent](https://amp.dev/documentation/components/amp-consent/#basic-blocking-behaviors), more fine-grained support on the client-side is available, and it should be exposed via a server-side `googlesitekit_{$slug}_tag_amp_block_on_consent` filter, where `$slug` is the service name (`adsense`, `analytics`, `tagmanager`). The filterable parameter should be `false` by default, but can also be `true` or one of the strings `_till_responded`, `_till_accepted`, or `_auto_reject`.
    * If the filter value is `true` or one of the three strings, the AMP component tags should receive a `data-block-on-consent` attribute (if one of the three strings, with that value).
* For non-AMP, there is no client-side standardization, but there should still be a server-side `googlesitekit_{$slug}_tag_block_on_consent` filter, where `$slug` is the service name. The filterable parameter should be `false` by default but can also be `true`.
[...]

PR #2148, opened 2020-10-08
Add cookie consent API for modules with blockable tags

aaemnnosttv:
## Summary

Addresses issue #2087

## Relevant technical choices

* ~~For tag blocking, the non-amp variant of the filter is used as the default value of the amp-specific filter that way both filters do not need to be used to block both (there shouldn't be a case where only one would be intended to be blocked, but individual control is still possible by using both filters so this is just a convenience).~~ **reverted**
* A dedicated test was added for the blockable tags trait which offers thorough coverage of the behavior of the return values in each case. Module-specific tests are less rigorous as a result and mostly concerned with the presence of the block on consent attributes rather than specific values.

## Checklist


includes/Modules/AdSense.php
```diff
@@ -259,10 +269,14 @@ protected function output_adsense_script( $client_id ) {
 		// If we haven't completed the account connection yet, we still insert the AdSense tag
 		// because it is required for account verification.
 		?>
-<script async src="//pagead2.googlesyndication.com/pagead/js/adsbygoogle.js"></script> <?php // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript ?>
+<script
+	async
+	src="//pagead2.googlesyndication.com/pagead/js/adsbygoogle.js" <?php // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript ?>
+		<?php echo $this->get_tag_block_on_consent_attribute(); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>
```
aaemnnosttv, line 275: Not sure why but PHPCS insists on this having 2 indents rather than 1.

includes/Modules/AdSense.php
```diff
@@ -259,10 +269,14 @@ protected function output_adsense_script( $client_id ) {
...
+></script>
 <script>
 (adsbygoogle = window.adsbygoogle || []).push({
-google_ad_client: "<?php echo esc_attr( $client_id ); ?>",
+google_ad_client: '<?php echo esc_js( $client_id ); ?>',
```
aaemnnosttv, line 279: Changed this to use the proper escaping function for use in JS.

includes/Modules/AdSense.php
```diff
@@ -44,8 +44,14 @@
...
-final class AdSense extends Module implements Module_With_Screen, Module_With_Scopes, Module_With_Settings, Module_With_Assets, Module_With_Debug_Fields, Module_With_Owner {
-	use Module_With_Screen_Trait, Module_With_Scopes_Trait, Module_With_Settings_Trait, Module_With_Assets_Trait, Module_With_Owner_Trait;
+final class AdSense extends Module
+	implements Module_With_Screen, Module_With_Scopes, Module_With_Settings, Module_With_Assets, Module_With_Debug_Fields, Module_With_Owner {
+	use Module_With_Assets_Trait;
+	use Module_With_Blockable_Tags_Trait;
+	use Module_With_Owner_Trait;
+	use Module_With_Scopes_Trait;
+	use Module_With_Screen_Trait;
+	use Module_With_Settings_Trait;
```
aaemnnosttv, line 54: Changed updated modules to use individual `use` statements for traits (unfortunately we can't do the same for interfaces), which makes them a bit easier to read/scan.

includes/Modules/Analytics.php
```diff
@@ -340,9 +348,11 @@ public function get_debug_fields() {
 	 * @param string $property_id Analytics property ID to use in the snippet.
 	 */
 	protected function enqueue_gtag_js( $property_id ) {
+		$gtag_src = "https://www.googletagmanager.com/gtag/js?id=$property_id";
+
 		wp_enqueue_script( // phpcs:ignore WordPress.WP.EnqueuedResourceParameters.MissingVersion
 			'google_gtagjs',
-			'https://www.googletagmanager.com/gtag/js?id=' . esc_attr( $property_id ),
+			$gtag_src,
```
aaemnnosttv, line 355: Escaping the `$property_id` isn't necessary as the whole URL runs through `esc_url` internally.

includes/Modules/Analytics.php
```diff
@@ -411,6 +421,28 @@ protected function enqueue_gtag_js( $property_id ) {
...
+				if ( 'google_gtagjs' !== $handle ) {
+					return $tag;
+				}
+
+				return str_replace(
+					array(
+						"<script src='$gtag_src", // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
+						"<script src=\"$gtag_src\"", // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
+						"<script type='text/javascript' src='$gtag_src", // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
+						"<script type=\"text/javascript\" src=\"$gtag_src", // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
```
aaemnnosttv, line 438: The presence of the `type` attribute depends on the current theme's declared support for `html5` so we have to account for both cases here. Both quote types are supported as well just to be safe.

Also, the reason this matches the `src` at all is because secondary extra/inline scripts associated with the handle are also included in `$tag` so we only want to change the main script.

includes/Core/Assets/Assets.php
```diff
@@ -97,6 +97,15 @@ public function register() {
...
 		add_action( 'wp_enqueue_scripts', $register_callback );
 
+		add_filter(
+			'script_loader_tag',
+			function( $tag, $handle ) {
+				return $this->add_async_defer_attribute( $tag, $handle );
+			},
+			10,
+			2
+		);
```
felixarntz, line 107: 👍

includes/Core/Modules/Module_With_Blockable_Tags_Trait.php
```diff
@@ -0,0 +1,123 @@
...
+		/* @var Module $this The module instance. */
+
+		/**
+		 * Filters whether or not the tag should be blocked from rendering.
+		 *
+		 * @since n.e.x.t
+		 *
+		 * @param bool $blocked Whether or not the tag output is suppressed. Default: false.
+		 */
+		$blocked = (bool) apply_filters( "googlesitekit_{$this->slug}_tag_blocked", false );
```
felixarntz, line 39: Right now this filter runs for AMP too, it should only be run for non-AMP. This similar to how we have other filters like `*_init_tag` and `*_init_tag_amp` that run on non-AMP vs AMP respectively.

includes/Core/Modules/Module_With_Blockable_Tags_Trait.php
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	/**
+	 * Gets the list of allowed block on consent values.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string[] Block on consent attribute values.
+	 */
+	protected function get_allowed_block_on_consent_values() {
```
felixarntz, line 116: Maybe rename to `get_allowed_amp_block_on_consent_values`? We could also add a `@see` annotation in the method description that links to https://amp.dev/documentation/components/amp-consent/#advanced-predefined-consent-blocking-behaviors.

includes/Core/Modules/Module_With_Blockable_Tags_Trait.php
```diff
@@ -0,0 +1,123 @@
...
+
+		/**
+		 * Filters whether the tag requires user consent before loading.
+		 *
+		 * @since n.e.x.t
+		 *
+		 * @param bool $blocked Whether or not the tag requires user consent to load. Default: false.
+		 */
+		if ( apply_filters( "googlesitekit_{$this->slug}_tag_block_on_consent", false ) ) {
+			return 'type="text/plain" data-block-on-consent';
```
felixarntz, line 73: This should be prefixed by a space, so that it fits right into the tag where we call it (if nothing is returned, no whitespace is inserted).

includes/Core/Modules/Module_With_Blockable_Tags_Trait.php
```diff
@@ -0,0 +1,123 @@
...
+		 * @param bool $blocked Whether or not the tag requires user consent to load. Default: false.
+		 */
+		$block_on_consent = apply_filters( "googlesitekit_{$this->slug}_tag_amp_block_on_consent", false );
+
+		if ( in_array( $block_on_consent, $this->get_allowed_block_on_consent_values(), true ) ) {
+			return sprintf( 'data-block-on-consent="%s"', $block_on_consent );
+		}
+
+		if ( filter_var( $block_on_consent, FILTER_VALIDATE_BOOLEAN ) ) {
+			return 'data-block-on-consent';
```
felixarntz, line 103: See above, these should be prefixed by a space.

includes/Modules/AdSense.php
felixarntz: I think we can keep this in a single line. This will also ensure we don't add extra whitespace. The opening `<?php` tag for conditionally adding the attribute can be right after the previous closing `"`, as it should add an initial space as needed.

includes/Modules/AdSense.php
```diff
@@ -285,7 +299,11 @@ protected function print_amp_auto_ads( $client_id ) {
 		$this->adsense_tag_printed = true;
 
 		?>
-		<amp-auto-ads type="adsense" data-ad-client="<?php echo esc_attr( $client_id ); ?>"></amp-auto-ads>
+<amp-auto-ads
+	type="adsense"
+	data-ad-client="<?php echo esc_attr( $client_id ); ?>"
+		<?php echo $this->get_tag_amp_block_on_consent_attribute(); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>
+></amp-auto-ads>
```
felixarntz, line 306: See above, we could keep this in a single line.

includes/Modules/AdSense.php
```diff
@@ -353,7 +371,12 @@ protected function amp_content_add_auto_ads( $content, $client_id ) {
 		}
 
 		$this->adsense_tag_printed = true;
-		return '<amp-auto-ads type="adsense" data-ad-client="' . esc_attr( $client_id ) . '"></amp-auto-ads> ' . $content;
+
+		return sprintf(
+			'<amp-auto-ads type="adsense" data-ad-client="%s" %s></amp-auto-ads> ',
```
felixarntz, line 376: See above:
```suggestion
			'<amp-auto-ads type="adsense" data-ad-client="%s"%s></amp-auto-ads> ',
```

includes/Modules/Analytics.php
```diff
@@ -411,6 +421,33 @@ protected function enqueue_gtag_js( $property_id ) {
...
+						"<script src='$gtag_src'", // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
+						"<script src=\"$gtag_src\"", // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
+						"<script type='text/javascript' src='$gtag_src'", // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
+						"<script type=\"text/javascript\" src=\"$gtag_src\"", // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
+					),
+					array( // `type` attribute intentionally excluded in replacements.
+						"<script $block_on_consent_attrs src='$gtag_src'", // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
+						"<script $block_on_consent_attrs src=\"$gtag_src\"", // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
+						"<script $block_on_consent_attrs src='$gtag_src'", // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
+						"<script $block_on_consent_attrs src=\"$gtag_src\"", // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
```
felixarntz, line 444: See above
```suggestion
						"<script{$block_on_consent_attrs} src='$gtag_src'", // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
						"<script{$block_on_consent_attrs} src=\"$gtag_src\"", // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
						"<script{$block_on_consent_attrs} src='$gtag_src'", // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
						"<script{$block_on_consent_attrs} src=\"$gtag_src\"", // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
```

includes/Modules/Analytics.php
```diff
@@ -467,11 +504,15 @@ protected function print_amp_gtag( $property_id ) {
...
-		<amp-analytics type="gtag" data-credentials="include">
-			<script type="application/json">
-				<?php echo wp_json_encode( $gtag_amp_opt_filtered ); ?>
-			</script>
-		</amp-analytics>
+<amp-analytics
+	type="gtag"
+	data-credentials="include"
+		<?php echo $this->get_tag_amp_block_on_consent_attribute(); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>
+>
```
felixarntz, line 511: See above.

includes/Modules/Tag_Manager.php
```diff
@@ -308,18 +315,20 @@ public function get_debug_fields() {
...
-					j = d.createElement( s ), dl = l != 'dataLayer' ? '&l=' + l : '';
-				j.async = true;
-				j.src = 'https://www.googletagmanager.com/gtm.js?id=' + i + dl;
-				f.parentNode.insertBefore( j, f );
-			} )( window, document, 'script', 'dataLayer', '<?php echo esc_js( $container_id ); ?>' );
-		</script>
-		<!-- End Google Tag Manager -->
+<!-- Google Tag Manager added by Site Kit -->
+<script
+		<?php echo $this->get_tag_block_on_consent_attribute(); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>
```
felixarntz, line 320: See above.

includes/Modules/Tag_Manager.php
```diff
@@ -371,13 +384,17 @@ protected function print_amp_gtm( $container_id ) {
...
-				<?php echo wp_json_encode( $gtm_amp_opt ); ?>
-			</script>
-		</amp-analytics>
-		<!-- End Google Tag Manager -->
+<!-- Google Tag Manager added by Site Kit -->
+<amp-analytics
+	config="<?php echo esc_url( "https://www.googletagmanager.com/amp.json?id=$container_id" ); ?>"
+	data-credentials="include"
+		<?php echo $this->get_tag_amp_block_on_consent_attribute(); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>
+>
```
felixarntz, line 392: See above.

felixarntz (changes requested):
This mostly looks good, just one important thing to change (filter execution), and a lot of nitpicks: I'd prefer to avoid introducing extra whitespace for these scripts in the frontend, it's not a big deal if these lines get a bit long in PHP - or we can use variables to shorten them.

In the ACs I actually prefixed the block on consent attributes with a space, but somehow GitHub makes it look like they aren't there - otherwise that intent might have been more clear.

aaemnnosttv:
@felixarntz all feedback should be addressed now. I used `printf` for outputting tags to minimize whitespace on the frontend while preserving line-specific phpcs-exclusions which was the main reason I went for the multiline approach before. There's even less whitespace now than before! 😄 

Let me know if there's anything else 👍

includes/Core/Modules/Module_With_Blockable_Tags_Trait.php
```diff
@@ -0,0 +1,123 @@
...
+	protected function is_tag_blocked() {
+		/* @var Module $this The module instance. */
+
+		if ( $this->context->is_amp() ) {
+			/**
+			 * Filters whether or not the AMP tag should be blocked from rendering.
+			 *
+			 * @since n.e.x.t
+			 *
+			 * @param bool $blocked Whether or not the tag output is suppressed. Defaults to the value of the 'googlesitekit_{slug}_tag_blocked' filter.
```
felixarntz, line 38:
```suggestion
			 * @param bool $blocked Whether or not the tag output is suppressed. Default: false.
```

includes/Core/Modules/Module_With_Blockable_Tags_Trait.php
```diff
@@ -0,0 +1,123 @@
...
+	 */
+	protected function get_tag_amp_block_on_consent_attribute() {
+		/* @var Module $this The module instance. */
+
+		/**
+		 * Filters whether the tag requires user consent before loading.
+		 *
+		 * @since n.e.x.t
+		 *
+		 * @param bool $blocked Whether or not the tag requires user consent to load. Default: false.
```
felixarntz, line 92:
```suggestion
		 * @param bool|string $blocked Whether or not the tag requires user consent to load. Alternatively, this can also be one of
		 *                             the special string values '_till_responded', '_till_accepted', or '_auto_reject'. Default: false.
```

felixarntz (approved):
Awesome! Just two documentation pieces that should be corrected.

felixarntz (approved):

Issue #1817, opened 2020-07-21
Tag Manager error during new container setup: Error: Returned an error response for your request 


jamesozzie:
## Bug Description

When using the plugin to setup a new Tag Manager container the below error appears when clicking on "Confirm & Continue" 
`Error: Returned an error response for your request.` This seems to occur only when trying to create a new container for the second time (after disconnecting the Tag Manager module)

![image](https://user-images.githubusercontent.com/41326532/88068827-a005e680-cb68-11ea-8bfb-c94d1a2c7643.png)

## Steps to reproduce

1. Start the Tag Manager connection process
2. Select the option to "Setup a new container", either web or AMP
3. Click on "Confirm & Continue"
4. Error appears

## Additional Context

- ~~ModSecurity enabled on this particular site.~~ Issue occurs with ModSecurity also disabled
- Setup proceeds when existing containers are selected in both dropdowns only (when using paired AMP)
- No security, caching or optimization plugins active
- Reproducible on differing sites
- Only seems to occur after attempting to create an container within Site Kit on an account whereby you've already used SK to create a container (ie. You can't create a new container for the second time on the same account)

<details><summary>Site Health info</summary>

```
### wp-core ###

version: 5.4.2
site_language: en_US
user_language: en_US
permalink: /%postname%/
https_status: true
user_registration: 0
default_comment_status: open
multisite: false
user_count: 5
dotorg_communication: true

### wp-paths-sizes ###

wordpress_path: /var/www/vhosts/beijing.ie/httpdocs
wordpress_size: 96.22 MB (100894666 bytes)
uploads_path: /var/www/vhosts/beijing.ie/httpdocs/wp-content/uploads
uploads_size: 37.40 MB (39214146 bytes)
themes_path: /var/www/vhosts/beijing.ie/httpdocs/wp-content/themes
themes_size: 19.17 MB (20096230 bytes)
plugins_path: /var/www/vhosts/beijing.ie/httpdocs/wp-content/plugins
plugins_size: 231.01 MB (242230756 bytes)
database_size: 84.15 MB (88235503 bytes)
total_size: 467.94 MB (490671301 bytes)

### wp-active-theme ###

name: Twenty Seventeen (twentyseventeen)
version: 2.2 (latest version: 2.3)
author: the WordPress team
author_website: https://wordpress.org/
parent_theme: none
[...]

PR #2149, opened 2020-10-08
Bug/1817 new container setup

eugene-manuilov:
## Summary

Addresses issue #1817

## Relevant technical choices

* It turned out that there are two files in the `util` folder of the tagmanager module that contain duplicate functions from the `validation.js` file. So, I have removed that in favor of functions from the `validation.js` file.

## Checklist


assets/js/modules/tagmanager/components/common/ContainerNameTextField.js
```diff
@@ -0,0 +1,76 @@
...
+		? <HelperText persistent>{ __( 'A container with this name already exists.', 'google-site-kit' ) }</HelperText>
+		: undefined;
+
+	let trailingIcon;
+	if ( helperText ) {
+		trailingIcon = (
+			<span className="googlesitekit-text-field-icon--warning">
+				<span className="screen-reader-text">
+					{ __( 'Warning', 'google-site-kit' ) }
+				</span>
```
tofumatt, line 58: I don't think this screen-reader text is appropriate. The text "Warning" will appear at the end of the textfield's… label, presumably? But it will be quite out of context—as I understand it, this will read: `A container with this name already exists. Warning`. Even if the warning prepended the `helperText`, without a colon it won't be correct: `Warning A container with this name already exists.`. And that won't translate well.

Best to remove the icon's screen-reader text and have it be a visual-only icon. Augmenting the `helperText` to look something like: `<span className="screen-reader-text">Warning: </span>A container with this name already exists.` would be better, but I don't think it's even necessary here.

assets/js/modules/tagmanager/components/common/WebContainerNameTextField.js
```diff
@@ -0,0 +1,63 @@
...
+	} );
+
+	const siteName = useSelect( ( select ) => select( CORE_SITE ).getSiteName() );
+	const isSecondaryAMP = useSelect( ( select ) => select( CORE_SITE ).isSecondaryAMP() );
+	const containerName = useSelect( ( select ) => select( CORE_FORMS ).getValue( FORM_SETUP, 'containerName' ) );
+
+	const { setValues } = useDispatch( CORE_FORMS );
+	useEffect( () => {
+		if ( ! containerName ) {
+			setValues( FORM_SETUP, { containerName: siteName || global.location.hostname } );
```
tofumatt, line 48: These fallbacks to `location.hostname` are there for the rare cases that `select( CORE_SITE )` is unpopulated, to prevent errors, correct? I'm assuming we'd rarely run into this in normal use, so just curious really.

assets/js/modules/tagmanager/components/settings/SettingsForm.js
```diff
@@ -19,17 +19,42 @@
...
+			? <AMPContainerNameTextField />
+			: null;
+
+		containerNames = (
+			<div className="googlesitekit-setup-module__inputs">
+				{ webContainerName }
+				{ ampContainerName }
+			</div>
+		);
+	}
```
tofumatt, line 56: The scoping of these variables and the assignments I found a bit tough to follow here. I don't notice a big readability difference with replacing this block with

```js
const showContainerName = containerID === CONTAINER_CREATE || ampContainerID === CONTAINER_CREATE;
```

… and replacing `{ containerNames }` below with:

```js
{ showContainerName && (
	<div className="googlesitekit-setup-module__inputs">
		{ containerID === CONTAINER_CREATE && <WebContainerNameTextField /> }
		{ ampContainerID === CONTAINER_CREATE && <AMPContainerNameTextField /> }
	</div>
) }
```

Just a thought, let me know what you think 🙂

assets/js/modules/tagmanager/components/setup/SetupForm.js
```diff
@@ -128,6 +132,24 @@ export default function SetupForm( { finishSetup } ) {
...
+			? <AMPContainerNameTextField />
+			: null;
+
+		containerNames = (
+			<div className="googlesitekit-setup-module__inputs">
+				{ webContainerName }
+				{ ampContainerName }
+			</div>
+		);
+	}
```
tofumatt, line 151: Similar to my above, I think we could improve readability here a bit.

Though given this code looks exactly the same as the one in the above component, seems like we could extract this logic out into its own component and use it here and in `assets/js/modules/tagmanager/components/settings/SettingsForm.js`.

assets/js/modules/tagmanager/util/index.js
```diff
@@ -1,4 +1,21 @@
+/**
+ * Utilities.
```
tofumatt, line 2:
```suggestion
 * Utilities for the Tag Manager module.
```

tofumatt (changes requested):
Looks good to me! There's one bit that needs changing (the icon screen reader text). The only other comment was about tweaks to the readability of the `SettingsForm.js`/`SetupForm.js` code—I think the best thing to do there would be to extract out a small `ContainerNames.js` component there as the code looks totally duplicated.

After that this should be good to go 👍🏻

assets/js/modules/tagmanager/components/common/WebContainerNameTextField.js
eugene-manuilov: @tofumatt yep, you are right. It replicates the logic from the `Tag_Manager.php` class.

assets/js/modules/tagmanager/components/common/ContainerNameTextField.js
eugene-manuilov: Ok, removed screen reader text.

assets/js/modules/tagmanager/components/settings/SettingsForm.js
eugene-manuilov: I like the idea of using a small component for it. Updated.

assets/js/modules/tagmanager/components/setup/SetupForm.js
eugene-manuilov: Updated to use an external component.

tofumatt (approved):
Awesome 🙂  👍🏻

stories/modules-tagmanager-settings.stories.js
```diff
@@ -133,6 +133,15 @@ storiesOf( 'Tag Manager Module/Settings', module )
...
 		return <Settings isOpen={ true } isEditing={ true } registry={ registry } />;
 	} )
+	.add( 'Edit, with "Set up a new container"', ( registry ) => {
+		const webContainerVersion = fixtures.liveContainerVersions.web.gaWithVariable;
+		const accountID = webContainerVersion.accountId; // eslint-disable-line sitekit/camelcase-acronyms
+
+		registry.dispatch( STORE_NAME ).receiveGetAccounts( fixtures.accounts );
+		registry.dispatch( STORE_NAME ).receiveGetContainers( [], { accountID } );
+
+		return <Settings isOpen={ true } isEditing={ true } registry={ registry } />;
```
tofumatt, line 143: Hah, just noticed that here (and elsewhere in this file) we're using `isOpen={ true }` rather than just `isOpen`. Hardly a big deal, but I'm surprised ESLint didn't warn us—I thought a lot of ESLint rules prefer boolean props to omit the `={ true }` part. 🤷🏻‍♂️

eugene-manuilov: I have updated these stories to use just `isOpen` and `isEditing`.

assets/js/modules/tagmanager/datastore/settings.test.js
```diff
@@ -597,24 +670,54 @@ describe( 'modules/tagmanager settings', () => {
...
+					registry.dispatch( STORE_NAME ).receiveGetContainers( containers, { accountID } );
+
 					registry.dispatch( STORE_NAME ).setContainerID( CONTAINER_CREATE );
 					registry.dispatch( STORE_NAME ).setInternalContainerID( '' );
 					registry.dispatch( STORE_NAME ).setAMPContainerID( CONTAINER_CREATE );
 					registry.dispatch( STORE_NAME ).setInternalAMPContainerID( '' );
+					registry.dispatch( CORE_FORMS ).setValues( FORM_SETUP, {
+						containerName: 'Sitekit',
+						ampContainerName: 'Sitekit AMP',
+					} );
```
aaemnnosttv, line 720: These changes look good so far, but it was only updated to show what is supported. Please add/update a few cases to show the changes to what is now required similar to other tests where `canSubmitChanges` is `false`.

- Creating a web container requires a unique container name
- Creating web and AMP containers requires unique container names
- Creating an AMP container requires a unique container name

assets/js/modules/tagmanager/util/validation.js
```diff
@@ -75,6 +75,57 @@ export function isValidContainerID( containerID ) {
...
+ */
+export function getNormalizedContainerName( containerName ) {
+	let sanitizedContainerName = containerName;
+
+	// Remove any leading or trailing whitespace.
+	sanitizedContainerName = sanitizedContainerName.trim();
+	// Must not start with an underscore.
+	sanitizedContainerName = sanitizedContainerName.replace( /^_+/, '' );
+	// Convert accents to basic characters to prevent them from being stripped.
+	sanitizedContainerName = sanitizedContainerName.normalize( 'NFD' ).replace( /[\u0300-\u036f]/g, '' );
```
aaemnnosttv, line 94: Nice!

assets/js/modules/tagmanager/util/validation.js
```diff
@@ -75,6 +75,57 @@ export function isValidContainerID( containerID ) {
...
+ *
+ * @since n.e.x.t
+ *
+ * @param {string}         containerName Container name to test.
+ * @param {Array.<Object>} containers    Available containers.
+ * @return {boolean} True if unique, otherwise false.
+ */
+export function isUniqueContainerName( containerName, containers ) {
+	const normalizedContainerName = getNormalizedContainerName( containerName );
+	return ! Array.isArray( containers ) || ! containers.some( ( { name } ) => name === normalizedContainerName );
```
aaemnnosttv, line 126: Would you please add a few tests for this?

I think the logic is slightly off here
```suggestion
	return Array.isArray( containers ) && ! containers.some( ( { name } ) => name === normalizedContainerName );
```

assets/js/modules/tagmanager/util/validation.js
```diff
@@ -75,6 +75,57 @@ export function isValidContainerID( containerID ) {
...
 
+/**
+ * Gets normalized container name.
+ *
+ * @since n.e.x.t
+ *
+ * @param {string} containerName Container name to normalize.
+ * @return {string} Normalized container name.
+ */
+export function getNormalizedContainerName( containerName ) {
```
aaemnnosttv, line 86: While this is related to validation, I don't think it belongs in this module.

Let's put it in its own module, or a general `utils/container` module.

Please add tests for this function as well. You can probably borrow most of it from the current tests in PHP – see `container_name_provider`.

stories/modules-tagmanager-settings.stories.js
```diff
@@ -118,20 +118,29 @@ storiesOf( 'Tag Manager Module/Settings', module )
...
 	} )
 	.add( 'Edit, with "Set up a new account"', ( registry ) => {
 		registry.dispatch( STORE_NAME ).setAccountID( ACCOUNT_CREATE );
 		registry.dispatch( STORE_NAME ).receiveGetAccounts( fixtures.accounts );
 		registry.dispatch( STORE_NAME ).receiveGetSettings( defaultSettings );
 
-		return <Settings isOpen={ true } isEditing={ true } registry={ registry } />;
+		return <Settings isOpen isEditing registry={ registry } />;
+	} )
+	.add( 'Edit, with "Set up a new container"', ( registry ) => {
```
aaemnnosttv, line 136: Please also add a case to show the error/warning state when a container by the same name exists.

stories/modules-tagmanager-setup.stories.js
```diff
@@ -96,6 +96,15 @@ storiesOf( 'Tag Manager Module/Setup', module )
 
 		return <Setup registry={ registry } />;
 	} )
+	.add( 'Set up a new container', ( registry ) => {
```
aaemnnosttv, line 99: Same as before - let's add a case for the "container already exists" scenario.

assets/js/modules/tagmanager/util/index.js
```diff
@@ -1,4 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export * from './validation';
 export { default as getContainers } from './get-containers';
-export { default as isValidAccountID } from './is-valid-account-id';
-export { default as isValidContainerID } from './is-valid-container-id';
```
aaemnnosttv, line 3: Thanks! Not sure why these were left around 🤔

assets/js/modules/tagmanager/components/common/ContainerNameTextField.js
```diff
@@ -0,0 +1,69 @@
...
+	const onChange = useCallback( ( { currentTarget } ) => {
+		setValues( FORM_SETUP, { [ formFieldID ]: currentTarget.value } );
+	}, [ formFieldID ] );
+
+	const helperText = containerName && ! isUniqueContainerName( containerName, containers )
+		? <HelperText persistent>{ __( 'A container with this name already exists.', 'google-site-kit' ) }</HelperText>
+		: undefined;
+
+	const trailingIcon = helperText
+		? <span className="googlesitekit-text-field-icon--warning" />
```
aaemnnosttv, line 53: In contrast to Analytics, I think this should probably show an error icon here instead. Unlike Analytics, the user will be blocked from submitting changes without a unique name for the new container.

assets/js/modules/tagmanager/components/common/AMPContainerNameTextField.js
```diff
@@ -0,0 +1,58 @@
...
+		return select( STORE_NAME ).getAMPContainers( accountID );
+	} );
+
+	const siteName = useSelect( ( select ) => select( CORE_SITE ).getSiteName() );
+	const ampContainerName = useSelect( ( select ) => select( CORE_FORMS ).getValue( FORM_SETUP, 'ampContainerName' ) );
+
+	const { setValues } = useDispatch( CORE_FORMS );
+	useEffect( () => {
+		if ( ! ampContainerName ) {
+			setValues( FORM_SETUP, { ampContainerName: `${ siteName || global.location.hostname } AMP` } );
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/tagmanager/components/common/AMPContainerNameTextField.js
```diff
@@ -0,0 +1,58 @@
...
+		}
+	}, [] );
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/tagmanager/components/common/AMPContainerNameTextField.js
```diff
@@ -0,0 +1,58 @@
...
+import { useEffect } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import Data from 'googlesitekit-data';
+import { STORE_NAME, FORM_SETUP } from '../../datastore/constants';
+import { STORE_NAME as CORE_SITE } from '../../../../googlesitekit/datastore/site/constants';
+import { STORE_NAME as CORE_FORMS } from '../../../../googlesitekit/datastore/forms';
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/tagmanager/components/common/AMPContainerNameTextField.js
```diff
@@ -0,0 +1,58 @@
...
+	const { setValues } = useDispatch( CORE_FORMS );
+	useEffect( () => {
+		if ( ! ampContainerName ) {
+			setValues( FORM_SETUP, { ampContainerName: `${ siteName || global.location.hostname } AMP` } );
+		}
+	}, [] );
+
+	return (
+		<ContainerNameTextField
+			formFieldID="ampContainerName"
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/tagmanager/components/common/ContainerNameTextField.js
```diff
@@ -0,0 +1,69 @@
...
+
+export default function ContainerNameTextField( { label, containers, formFieldID } ) {
+	const containerName = useSelect( ( select ) => select( CORE_FORMS ).getValue( FORM_SETUP, formFieldID ) );
+
+	const { setValues } = useDispatch( CORE_FORMS );
+	const onChange = useCallback( ( { currentTarget } ) => {
+		setValues( FORM_SETUP, { [ formFieldID ]: currentTarget.value } );
+	}, [ formFieldID ] );
+
+	const helperText = containerName && ! isUniqueContainerName( containerName, containers )
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/tagmanager/components/common/ContainerNameTextField.js
```diff
@@ -0,0 +1,69 @@
...
+ * Internal dependencies
+ */
+import Data from 'googlesitekit-data';
+import { STORE_NAME as CORE_FORMS } from '../../../../googlesitekit/datastore/forms';
+import { TextField, HelperText, Input } from '../../../../material-components';
+import { FORM_SETUP } from '../../datastore/constants';
+import { isUniqueContainerName } from '../../util';
+const { useSelect, useDispatch } = Data;
+
+export default function ContainerNameTextField( { label, containers, formFieldID } ) {
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/tagmanager/components/common/ContainerNameTextField.js
```diff
@@ -0,0 +1,69 @@
...
+		? <HelperText persistent>{ __( 'A container with this name already exists.', 'google-site-kit' ) }</HelperText>
+		: undefined;
+
+	const trailingIcon = helperText
+		? <span className="googlesitekit-text-field-icon--warning" />
+		: undefined;
+
+	return (
+		<div className="googlesitekit-tagmanager-containername">
+			<TextField label={ label } outlined helperText={ helperText } trailingIcon={ trailingIcon }>
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/tagmanager/components/common/ContainerNameTextField.js
```diff
@@ -0,0 +1,69 @@
...
+				<Input value={ containerName } onChange={ onChange } />
+			</TextField>
+		</div>
+	);
+}
+
+ContainerNameTextField.propTypes = {
+	label: PropTypes.string.isRequired,
+	formFieldID: PropTypes.string.isRequired,
+	containers: PropTypes.arrayOf( PropTypes.object ),
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/tagmanager/components/common/ContainerNames.js
```diff
@@ -0,0 +1,42 @@
...
+	const ampContainerID = useSelect( ( select ) => select( STORE_NAME ).getAMPContainerID() );
+
+	if ( containerID !== CONTAINER_CREATE && ampContainerID !== CONTAINER_CREATE ) {
+		return null;
+	}
+
+	return (
+		<div className="googlesitekit-setup-module__inputs">
+			{ containerID === CONTAINER_CREATE && <WebContainerNameTextField /> }
+			{ ampContainerID === CONTAINER_CREATE && <AMPContainerNameTextField /> }
```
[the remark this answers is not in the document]
eugene-manuilov: Added these checks to text field components itself, but I don't think we need to remove it from here because it helps us to prevent unnecessary selects.

assets/js/modules/tagmanager/datastore/settings.js
```diff
@@ -221,16 +228,36 @@ export const validateCanSubmitChanges = ( select ) => {
...
+
 	// Note: these error messages are referenced in test assertions.
 	invariant( ! isDoingSubmitChanges(), 'cannot submit changes while submitting changes' );
 	invariant( haveSettingsChanged(), 'cannot submit changes if settings have not changed' );
-	invariant( isValidAccountID( getAccountID() ), 'a valid accountID is required to submit changes' );
+	invariant( isValidAccountID( accountID ), 'a valid accountID is required to submit changes' );
+
+	const containerID = getContainerID();
+	if ( containerID === CONTAINER_CREATE ) {
+		const containerName = select( CORE_FORMS ).getValue( FORM_SETUP, 'containerName' );
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/tagmanager/datastore/settings.js
```diff
@@ -221,16 +228,36 @@ export const validateCanSubmitChanges = ( select ) => {
...
+		invariant( isValidContainerName( containerName ), `a container name is invalid` );
+
+		const webContainers = getWebContainers( accountID );
+		invariant( isUniqueContainerName( containerName, webContainers ), `a container with "${ containerName }" name already exists` );
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/tagmanager/datastore/settings.test.js
```diff
@@ -387,7 +441,17 @@ describe( 'modules/tagmanager settings', () => {
 				} );
 
 				it( 'supports creating a web container', () => {
+					const { buildAndReceiveWebAndAMP } = createBuildAndReceivers( registry );
```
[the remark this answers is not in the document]
eugene-manuilov: I don't think we should move it into the top-level `beforeEach` because not all tests need it.

assets/js/modules/tagmanager/util/validation.js
eugene-manuilov: The logic is fine, it just accounts for a case when containers aren't loaded yet and `containers` value is undefined.

assets/js/modules/tagmanager/components/common/ContainerNameTextField.js
eugene-manuilov: Updated.

assets/js/modules/tagmanager/util/validation.js
eugene-manuilov: Updated.

assets/js/modules/tagmanager/datastore/settings.test.js
eugene-manuilov: Added.

stories/modules-tagmanager-settings.stories.js
eugene-manuilov: Added.

stories/modules-tagmanager-setup.stories.js
eugene-manuilov: Added.

assets/js/modules/tagmanager/datastore/settings.test.js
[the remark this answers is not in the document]
aaemnnosttv: Not all tests call `buildAndReceiveWebAndAMP` - I'm not suggesting we call that for all tests, but that we call `createBuildAndReceivers( registry )` (like we do in `tagmanager/datastore/versions.test.js`) for each which is not an expensive call and would save a few extra lines. It's not a big deal though.

assets/js/modules/tagmanager/util/validation.js
aaemnnosttv: That makes more sense, thanks for clarifying.

Not sure if you considered this before, but I think this might be better implemented as a private selector in `containers.js`. Then we could simply have `select(...).isUniqueContainerName( name )` and it would select all containers internally and return `undefined` if containers were not loaded yet for the current account. That way it would be less prone to improper use and more accurate in regards to the behavior in `canSubmitChanges` (an `undefined` return value from the strict selector would throw). What do you think?

assets/js/modules/tagmanager/components/common/AMPContainerNameTextField.js
```diff
@@ -0,0 +1,70 @@
...
+
+	let ampContainerName = siteName;
+
+	if ( ! ampContainerName ) {
+		try {
+			const url = new URL( referenceSiteURL );
+			ampContainerName = url.hostname;
+		} catch {
+		}
+	}
```
aaemnnosttv, line 49: How about this?

```suggestion
	if ( ! ampContainerName && isURL( referenceSiteURL ) ) {
		ampContainerName = new URL( referenceSiteURL ).hostname;
	}
```

Same for `WebContainerNameTextField`.

aaemnnosttv (commented):
Thanks @eugene-manuilov - this one is pretty close, just a few remaining points to address.

assets/js/modules/tagmanager/components/common/AMPContainerNameTextField.js
```diff
@@ -0,0 +1,70 @@
...
+import { STORE_NAME, FORM_SETUP, CONTAINER_CREATE } from '../../datastore/constants';
+import { STORE_NAME as CORE_SITE } from '../../../../googlesitekit/datastore/site/constants';
+import { STORE_NAME as CORE_FORMS } from '../../../../googlesitekit/datastore/forms/constants';
+import ContainerNameTextField from './ContainerNameTextField';
+const { useSelect, useDispatch } = Data;
+
+export default function AMPContainerNameTextField() {
+	const ampContainerID = useSelect( ( select ) => select( STORE_NAME ).getAMPContainerID() );
+	const siteName = useSelect( ( select ) => select( CORE_SITE ).getSiteName() );
+	const initialAMPContainerName = useSelect( ( select ) => select( CORE_FORMS ).getValue( FORM_SETUP, 'ampContainerName' ) );
```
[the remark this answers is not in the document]
eugene-manuilov: Ok, removed it completely because we don't really need it.

assets/js/modules/tagmanager/components/common/WebContainerNameTextField.js
```diff
@@ -0,0 +1,72 @@
...
+import { STORE_NAME as CORE_SITE } from '../../../../googlesitekit/datastore/site/constants';
+import { STORE_NAME as CORE_FORMS } from '../../../../googlesitekit/datastore/forms/constants';
+import ContainerNameTextField from './ContainerNameTextField';
+const { useSelect, useDispatch } = Data;
+
+export default function WebContainerNameTextField() {
+	const containerID = useSelect( ( select ) => select( STORE_NAME ).getContainerID() );
+	const siteName = useSelect( ( select ) => select( CORE_SITE ).getSiteName() );
+	const isSecondaryAMP = useSelect( ( select ) => select( CORE_SITE ).isSecondaryAMP() );
+	const initialContainerName = useSelect( ( select ) => select( CORE_FORMS ).getValue( FORM_SETUP, 'containerName' ) );
```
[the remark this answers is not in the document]
eugene-manuilov: Removed it completely because we don't really need it.

assets/js/modules/tagmanager/components/common/ContainerNames.js
```diff
@@ -0,0 +1,42 @@
...
+export default function ContainerNames() {
+	const containerID = useSelect( ( select ) => select( STORE_NAME ).getContainerID() );
+	const ampContainerID = useSelect( ( select ) => select( STORE_NAME ).getAMPContainerID() );
+
+	if ( containerID !== CONTAINER_CREATE && ampContainerID !== CONTAINER_CREATE ) {
+		return null;
+	}
+
+	return (
+		<div className="googlesitekit-setup-module__inputs">
```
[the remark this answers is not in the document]
eugene-manuilov: The `googlesitekit-setup-module__inputs--multiline` class makes it look worse. I have added a new class that removes the top margin.

assets/js/modules/tagmanager/components/common/AMPContainerNameTextField.js
eugene-manuilov: Updated.

assets/js/modules/tagmanager/util/validation.js
eugene-manuilov: I like how it is implemented now. The `isUniqueContainerName` function is intended for validation, not for selecting something from the store. So, I don't see it in the datastore itself.

assets/js/modules/tagmanager/util/validation.test.js
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Internal dependencies
+ */
+import { isUniqueContainerName } from './validation';
+
+describe( 'tag manager / utils / validation', () => {
+	describe( 'isUniqueContainerName', () => {
+		it( 'should return TRUE when "containers" arg is not an array yet', () => {
+			expect( isUniqueContainerName( 'test container', undefined ) ).toBe( true );
```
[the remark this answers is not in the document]
eugene-manuilov: As replied in another comment, I don't really like the idea to make it a selector because it is not a selector.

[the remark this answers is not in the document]
aaemnnosttv: I don't see that reply, can you point me to it?

assets/js/modules/tagmanager/components/common/WebContainerNameTextField.js
```diff
@@ -0,0 +1,66 @@
...
+	const referenceSiteURL = useSelect( ( select ) => select( CORE_SITE ).getReferenceSiteURL() );
+
+	let containerName = siteName;
+	if ( ! containerName && isURL( referenceSiteURL ) ) {
+		containerName = new URL( referenceSiteURL ).hostname;
+	}
+
+	const { setValues } = useDispatch( CORE_FORMS );
+	useEffect( () => {
+		setValues( FORM_SETUP, { containerName } );
```
aaemnnosttv, line 49: We should only call `setValues` here if the `containerName` is empty.

Because this happens unconditionally, the storybook stories for the warning don't show it because this effect is likely overriding it on render.

assets/js/modules/tagmanager/util/validation.test.js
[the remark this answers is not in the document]
eugene-manuilov: Here: https://github.com/google/site-kit-wp/pull/2149#discussion_r508536091

aaemnnosttv (commented):
@eugene-manuilov this is pretty close. There are a few merge conflicts to address as well as my comment below about the `useEffect`.

I also couldn't find your comment about why you felt using a selector for the unique container name was not a good idea.

There is one other issue here but I think we can leave that for a follow-up issue which is that in secondary AMP, if the user tries to create a new container for both using the same name, it won't show this as an error, but we should prevent them from submitting in this case because it will result in an error as well. Again, this is out of scope from the original issue here so I'll open a follow up for it.

assets/js/modules/tagmanager/components/common/WebContainerNameTextField.js
eugene-manuilov: Ah... my bad... I got it now :doh:... Updated.

assets/js/modules/tagmanager/util/validation.js
aaemnnosttv: It's not _bad_ the way it is now, my suggestion was that I think it may make more sense as a selector than a utility function. The main reason why I think that is because it depends on all the containers in datastore for the current account. Not only that, but it wouldn't make sense to use otherwise – there's no other case where we would need to check if a container name was unique among a different set of containers. So rather than putting the burden on the consumer to get the `accountID`, and select all containers properly to provide to the utility, that can all be done in a single selector.

Then the usage becomes much more simple:
```diff
diff --git a/assets/js/modules/tagmanager/components/common/ContainerNameTextField.js b/assets/js/modules/tagmanager/components/common/ContainerNameTextField.js
index ce5155479..206efb937 100644
--- a/assets/js/modules/tagmanager/components/common/ContainerNameTextField.js
+++ b/assets/js/modules/tagmanager/components/common/ContainerNameTextField.js
@@ -34,22 +34,18 @@ import Data from 'googlesitekit-data';
 import { STORE_NAME, FORM_SETUP } from '../../datastore/constants';
 import { STORE_NAME as CORE_FORMS } from '../../../../googlesitekit/datastore/forms/constants';
 import { TextField, HelperText, Input } from '../../../../material-components';
-import { isUniqueContainerName } from '../../util';
 const { useSelect, useDispatch } = Data;
 
 export default function ContainerNameTextField( { label, name } ) {
-	const containers = useSelect( ( select ) => {
-		const accountID = select( STORE_NAME ).getAccountID();
-		return select( STORE_NAME ).getContainers( accountID );
-	} );
 	const containerName = useSelect( ( select ) => select( CORE_FORMS ).getValue( FORM_SETUP, name ) );
+	const isUniqueContainerName = useSelect( ( select ) => select( STORE_NAME ).isUniqueContainerName( containerName ) );
 
 	const { setValues } = useDispatch( CORE_FORMS );
 	const onChange = useCallback( ( { currentTarget } ) => {
 		setValues( FORM_SETUP, { [ name ]: currentTarget.value } );
 	}, [ name ] );
 
-	const helperText = containerName && ! isUniqueContainerName( containerName, containers )
+	const helperText = containerName && ! isUniqueContainerName
 		? <HelperText persistent>{ __( 'A container with this name already exists.', 'google-site-kit' ) }</HelperText>
 		: undefined;
```

The utility function itself could still exist as it is, the selector by the same name would simply use it internally.

aaemnnosttv (commented):
@eugene-manuilov this all looks good to me, nice work!

I still think there's good reason for using a selector for the unique container name check. See [my response above](https://github.com/google/site-kit-wp/pull/2149#discussion_r510788151) for my reasoning and let me know what you think.

assets/js/modules/tagmanager/util/validation.js
eugene-manuilov: @aaemnnosttv I understand what you suggest and my opinion remains the same: I don't like the idea of putting validator functions into a datastore because it is out of scope for datastores. Putting all together breaks the "single responsibility" principle.

Following your logic, we would also need to move `isValidAccountID`, `isValidContainerID`, `isValidInternalContainerID`, and other validation functions into datastore as well because they all rely on data from a datastore and it would be easier to select just `isValidAccountID` selector than `getAccountID() + isValidAccountID()`.

I think consistency is more important than usability in this case.

eugene-manuilov:
@aaemnnosttv i have left an answer to your comment. If you want let's have a call instead of ping-ponging this ticket to each other.

assets/js/modules/tagmanager/util/validation.js
aaemnnosttv: > Following your logic, we would also need to move `isValidAccountID`, `isValidContainerID`, `isValidInternalContainerID`, and other validation functions into datastore as well because they all rely on data from a datastore and it would be easier to select just `isValidAccountID` selector than selecting an account ID with `getAccountID()` and calling `isValidAccountID()` function.

No, that's not the same. We use those to validate provided arguments/props in many places where the source of the data is unknown – they aren't only used to validate a value which is already in the datastore.

This isn't a validation function, it's a selector 😄  I'm not suggesting we add `isValidContainerName` as a selector on the store, that would definitely be out of place. This would only be asking the store if it has a container with the given name; that's it. Since this value is only relevant in the context of all containers for the currently selected account, I think it makes sense as a selector, again particularly because it isn't useful with different/unknown values for `containers`; we specifically care about all of the containers for the current account.

> I think consistency is more important than usability in this case.

I don't think it's a tradeoff between the two. In a way, it becomes more consistent as a selector than only as a utility, but again, the two aren't mutually exclusive, we can have both 😄

aaemnnosttv: @eugene-manuilov – After thinking about this a bit more, this shouldn't block the PR from being merged, but I do think it is an enhancement we should apply in a follow-up issue. Another issue is needed anyways regarding the edge case I mentioned previously for secondary AMP, so we can continue the discussion in the definition of that issue.

assets/js/modules/tagmanager/datastore/settings.test.js
```diff
@@ -342,6 +391,12 @@ describe( 'modules/tagmanager settings', () => {
 		} );
 
 		describe( 'canSubmitChanges', () => {
```
aaemnnosttv, line 393: Please add cases here for ensuring that a valid container name is required for each AMP mode.

assets/js/modules/tagmanager/components/common/ContainerNameTextField.js
```diff
@@ -0,0 +1,77 @@
...
+		? <HelperText persistent>{ __( 'A container with this name already exists.', 'google-site-kit' ) }</HelperText>
+		: undefined;
+
+	const trailingIcon = helperText
+		? <span className="googlesitekit-text-field-icon--error" />
+		: undefined;
+
+	return (
+		<div className="googlesitekit-tagmanager-containername">
+			<TextField
```
aaemnnosttv, line 62: Let's also add the `mdc-text-field--error` class if `!containerName || ! isUniqueContainerName`

assets/js/modules/tagmanager/util/validation.js
```diff
@@ -75,6 +76,34 @@ export function isValidContainerID( containerID ) {
...
+ *
+ * @param {string} containerName Container name to test.
+ * @return {boolean} True if valid, otherwise false.
+ */
+export function isValidContainerName( containerName ) {
+	return typeof containerName === 'string' && getNormalizedContainerName( containerName ).length > 0;
+}
+
+/**
+ * Checks if the given cotnainer name is unique accross account containers.
```
aaemnnosttv, line 92:
```suggestion
 * Checks if the given container name is unique across account containers.
```

aaemnnosttv (changes requested):
Thanks @eugene-manuilov – just a few minor details left to address for the most part; see my comments below.

Most importantly though, I noticed this item which is not yet implemented from the AC's:
> The new behavior should be covered by end-to-end tests for both setup and settings contexts

Rather than adding E2E for this, I think a simpler option would be to add Jest tests for Tag Manager's `SettingsEdit` and `SetupForm` components to ensure that the container name field is displayed when expected, and otherwise not displayed. These tests don't exist yet so they'd need to be created, but only these cases would be needed for now. Since I originally wrote the AC's and this should only shrink the scope I'll update this item only to keep the effort needed here to a minimum.

assets/js/modules/tagmanager/util/validation.js
eugene-manuilov: Sounds good @aaemnnosttv

assets/js/modules/tagmanager/components/common/ContainerNameTextField.js
eugene-manuilov: Added.

assets/js/modules/tagmanager/datastore/settings.test.js
eugene-manuilov: It's already added there or you want to add something else?

aaemnnosttv: Hmm, I see the cases where we test for a _unique_ container name, but not where we test for a valid container name which is different (i.e. checks for invalid characters). If it is there and I'm not seeing it, please point me to it! 😄

assets/js/modules/tagmanager/components/setup/SetupMain.test.js
```diff
@@ -0,0 +1,168 @@
+/**
+ * Tag Manager Settings Edit component tests.
```
aaemnnosttv, line 2:
```suggestion
 * Tag Manager Setup Main component tests.
```

aaemnnosttv (commented):
This looks great, thanks @eugene-manuilov.

Still one question remaining here that looks like you may have missed from earlier https://github.com/google/site-kit-wp/pull/2149#discussion_r512034611

Also, I realize that I meant to suggest `SettingsForm` and `SetupForm` for the components to test rather than `SettingsEdit` and `SetupForm`. The tests you added are fine but might be requiring a bit more data to be received/selected due to being higher-level. If you want to change them to be specific to the form components, that would be great but I won't ask you to do it 😄  

Otherwise this is very close. Thanks for your patience.

assets/js/modules/tagmanager/datastore/settings.test.js
eugene-manuilov: Ok, added a few tests that check cases when a container name is invalid.

eugene-manuilov:
> If you want to change them to be specific to the form components, that would be great but I won't ask you to do it.

No, thanks. I don't want to change it.

aaemnnosttv (approved):
Great, thanks!

Issue #2092, opened 2020-09-22
Merge advanced event tracking API


felixarntz:
The advanced Analytics event tracking API from #1728 which is currently in the `feature/ga-advanced-tracking` should be merged into `develop`. Only the API itself should be included though, not any of the 3P plugin-specific code, which is more of a sample implementation and would preferably be implemented by the respective plugins.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Site Kit should offer a 3P API to register Analytics event tracking configurations.
    * The API should be configuration-based in a way that works in AMP and non-AMP. In other words, the actual configurations should not include any JavaScript code.
    * For AMP, it should be possible to transform the configuration into a JSON object of `amp-analytics` triggers. For non-AMP, a similar behavior should be achieved by injecting some minimal JavaScript code in the frontend that iterates through the configuration.
    * The JavaScript injected into the frontend for non-AMP should be _very_ backward-compatible, i.e. don't rely on recent browser features etc.
* No 3P plugin-specific code should be included in the Site Kit codebase.
* Test coverage for the new API should be expanded where possible.

## Implementation Brief

* Remove all 3P plugin-specific code from the `feature/ga-advanced-tracking` branch.
* Update branch against `develop`.
* Add extra test coverage.
* Complete and merge #2117.

## QA Brief

* Verify that the new 3P API satisfies the requirements to track Analytics events based on a configuration.

## Changelog entry

* N/A

aaemnnosttv:
IB ✅

eugene-manuilov:
QA :x:. The generated javascript file doesn't work on the frontend because it is built as a jsonp-based bundle.

eugene-manuilov:
Created a PR against `master` branch that fixed javascript issues: https://github.com/google/site-kit-wp/pull/2152

felixarntz:
After #2152, the code snippet added to the frontend works as expected. QA ✅

PR #2152, opened 2020-10-09
Enhancement/2092 advanced analytics tracking fixes

eugene-manuilov:
## Summary

Addresses issue #2092

## Checklist

### DIFF
--- a/assets/js/analytics-advanced-tracking.js
+++ b/assets/js/analytics-advanced-tracking.js
@@ -32,11 +32,7 @@ import setUpAdvancedTracking from './analytics-advanced-tracking/set-up-advanced
  * @param {Object} metadata Additional event metadata to send, or `null`.
  */
 function sendEvent( action, metadata ) {
-	/* eslint-disable no-restricted-globals */
-	if ( window.gtag ) {
-		window.gtag( 'event', action, metadata );
-	}
-	/* eslint-enable */
+	window.gtag( 'event', action, metadata || undefined ); // eslint-disable-line no-restricted-globals
 }
 
 const events = window._googlesitekitAnalyticsTrackingData || []; // eslint-disable-line no-restricted-globals
--- a/assets/js/analytics-advanced-tracking.js
+++ b/assets/js/analytics-advanced-tracking.js
@@ -32,11 +32,11 @@ import setUpAdvancedTracking from './analytics-advanced-tracking/set-up-advanced
  * @param {Object} metadata Additional event metadata to send, or `null`.
  */
 function sendEvent( action, metadata ) {
-	if ( ! metadata ) {
-		window.gtag( 'event', action ); // eslint-disable-line no-restricted-globals
-	} else {
-		window.gtag( 'event', action, metadata ); // eslint-disable-line no-restricted-globals
+	/* eslint-disable no-restricted-globals */
+	if ( window.gtag ) {
+		window.gtag( 'event', action, metadata );
 	}
+	/* eslint-enable */
 }
 
 const events = window._googlesitekitAnalyticsTrackingData || []; // eslint-disable-line no-restricted-globals
--- a/assets/js/analytics-advanced-tracking.js
+++ b/assets/js/analytics-advanced-tracking.js
@@ -21,7 +21,6 @@
 /**
  * Internal dependencies
  */
-import SITEKIT_ANALYTICS_ADVANCED_TRACKING_EVENTS from 'analytics-advanced-tracking-events';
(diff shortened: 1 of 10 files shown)


assets/js/analytics-advanced-tracking.js
```diff
@@ -33,11 +32,10 @@ import setUpAdvancedTracking from './analytics-advanced-tracking/set-up-advanced
  * @param {Object} metadata Additional event metadata to send, or `null`.
  */
 function sendEvent( action, metadata ) {
-	if ( ! metadata ) {
-		global.gtag( 'event', action );
-	} else {
-		global.gtag( 'event', action, metadata );
-	}
+	window.gtag( 'event', action, metadata || undefined ); // eslint-disable-line no-restricted-globals
```
felixarntz, line 35: Why not `global` like elsewhere?

eugene-manuilov: @felixarntz it prevents webpack module concatenation optimization if we use global variables.

assets/js/analytics-advanced-tracking/set-up-advanced-tracking.js
```diff
@@ -48,7 +48,7 @@ export default function setUpAdvancedTracking( eventConfigurations, sendEvent )
 
 	return () => {
 		toRemove.forEach( ( listenerArgs ) => {
-			document.removeEventListener( ...listenerArgs );
+			document.removeEventListener.apply( document, listenerArgs );
```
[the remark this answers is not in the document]
eugene-manuilov: There are no technical differences, but if we use the old way to invoke the function with an array of arguments, babel won't try to inject a bunch of functions to transpile spread operation into an approach supported by old browsers.

eugene-manuilov:
@felixarntz with all these changes I have managed to reduce the bundle size for advanced Analytics events tracking from 2.8kb to 1.6kb (~43%).

aaemnnosttv (dismissed):
This works well for me although I'm seeing another problem which is that there is a difference in the number of triggers generated for web tracking than for AMP. This is because duplicate events may be added by a hook which runs more than once for the same hook.

I tested this locally using an `Event_List` that was included with tests that hooks 2 events onto `the_post` action. With the TwentyTwenty theme and a single post on the home page, it was run 5 times, which added a total of 10 triggers in "web mode".

For AMP however, event triggers are keyed with a hash based on the action, event, and selector. This resulted in only 2 triggers being registered when loading the same page in an AMP context.

It seems like we should see consistent results in the number of triggers/eventListeners configured regardless of AMP mode.

Ideally I think the unique handling would be done when registering the Event. Currently, this doesn't happen until the events are injected which means many duplicate `Event` instances may be accumulated unnecessarily.

@felixarntz what do you think about keying the added/registered `Events` in the array in the same way for both? That way if a duplicate `Event` is registered, it simply overwrites the previous in the array, and we can always get them all as a list with `array_values` similar to how we handle errors in the datastore.

felixarntz:
@aaemnnosttv Good point, keying the events in the same way like in AMP makes sense.

aaemnnosttv:
@felixarntz what do you think about adding a `Event::get_hash` method?

Then we can change `add_event` like so:

```php
	protected function add_event( Event $event ) {
		$this->events[ $event->get_hash() ] = $event;
	}

// alternatively, perhaps we just hash it in `add_event`?

	protected function add_event( Event $event ) {
		$hash = md5( wp_json_encode( $event->jsonSerialize() ) ); 
		$this->events[ $hash ] = $event;
	}
```

Then the injectors only need to iterate over the map of events which will already have the hash for each, and the injector can be more specific to passing the list along to the frontend.

aaemnnosttv:
@googlebot I consent.

aaemnnosttv:
@felixarntz I can confirm that these changes result in only the 2 expected event triggers in AMP and non-amp now:

image

image

includes/Modules/Analytics/Advanced_Tracking/AMP_Config_Injector.php
```diff
@@ -37,11 +37,9 @@ public function inject_event_configurations( $gtag_amp_opt, $events ) {
 			$gtag_amp_opt['triggers'] = array();
 		}
 
-		foreach ( $events as $event ) {
+		foreach ( $events as $amp_trigger_key => $event ) {
```
felixarntz, line 40: Strictly saying, we should maybe call this `$identifier` or `$key`, it' not specific for AMP. But not a big deal.

felixarntz (approved):
Awesome! Just one super-minor comment, but nothing blocking, so it's good to merge.

PR #2159, opened 2020-10-12
Enhancement/1326 images via webpack

eugene-manuilov:
## Summary

Addresses issue #1326

## Checklist


webpack.config.js
```diff
@@ -188,7 +189,33 @@ const webpackConfig = ( mode ) => {
...
-				rules,
+				rules: [
+					...rules,
+					{
+						test: /\.(png|jpg)$/i,
+						use: [
+							{
+								loader: 'file-loader',
+								options: {
+									name: '[name].[ext]',
```
aaemnnosttv, line 200: Since we're loading these via webpack now, can we add a hash to the filename as well? While not as critical as JS, they face the same potential browser caching problems. It seems like this would be an easy change, no?

bin/release
```diff
@@ -60,34 +64,61 @@ function getGit() {
...
-	// Remove any existing file by the same name.
-	del.sync( path.resolve( __dirname, `../${ filename }` ) );
+	const cp = ( what ) => fse.copy( resolve( root, what ), resolve( pluginDir, what ) );
+
+	// Remove and re-create release directory.
+	await fse.remove( releaseDir );
+	await fse.ensureDir( pluginDir );
 
-	// eslint-disable-next-line no-console
-	console.log( `Creating ${ filename }` );
```
aaemnnosttv, line 88: I think this would be good to keep since the name is dynamic.

bin/release
```diff
@@ -60,34 +64,61 @@ function getGit() {
...
+	const pluginDir = resolve( releaseDir, 'google-site-kit' );
 	const filename = generateFilename();
 
-	// Remove any existing file by the same name.
-	del.sync( path.resolve( __dirname, `../${ filename }` ) );
+	const cp = ( what ) => fse.copy( resolve( root, what ), resolve( pluginDir, what ) );
+
+	// Remove and re-create release directory.
+	await fse.remove( releaseDir );
+	await fse.ensureDir( pluginDir );
```
aaemnnosttv, line 102: We also need to remove the target zip file if one already exists for the same path, do we not? Or will the zip process overwrite an existing file if present?

assets/js/components/notifications/dashboard-setup-alerts.js
```diff
@@ -29,6 +29,7 @@ import { applyFilters } from '@wordpress/hooks';
 import { getQueryParameter, getModulesData } from '../../util';
 import Notification from './notification';
 import ModulesList from '../modules-list';
+import rocket from '../../../images/rocket.png';
```
aaemnnosttv, line 32: Let's name the import in a consistent way to indicate that it is an image, similar to what we do with svg icons.

```suggestion
import rocketImage from '../../../images/rocket.png';
```

The same applies to other image imports as well.

aaemnnosttv (changes requested):
Thanks @eugene-manuilov – this looks great overall, just a few minor changes to address, as well as a few questions and suggestions.

assets/js/googlesitekit-base.js
```diff
@@ -26,6 +26,9 @@ import {
 	clearWebStorage,
 } from './util/standalone';
 
+// Load image for the parent menu item.
+import '../images/logo-g_white_small.png';
```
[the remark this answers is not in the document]
eugene-manuilov: It's used in PHP as a menu icon: https://github.com/google/site-kit-wp/blob/3571133f387d089c50796748e918b3c50c6136b3/includes/Core/Admin/Screen.php#L126-L133

bin/release
```diff
@@ -60,34 +64,61 @@ function getGit() {
...
+	// Remove release folder at the end.
+	await fse.remove( releaseDir );
+}
 
-	return gulp.src( 'release/google-site-kit/**', { base: 'release/' } )
-		.pipe( zip( filename ) )
-		.pipe( gulp.dest( '.' ) );
-};
+makeRelease().catch( ( error ) => {
+	process.stderr.write( `${ error }\n` );
```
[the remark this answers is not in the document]
eugene-manuilov: Console functions are prohibited by our eslint rules and my vscode blames me for using it. `process.stderr` is what `console.error` uses to output a message.

package.json
```diff
@@ -5,21 +5,28 @@
...
     "build": "npm run build:production",
-    "build:dev": "npm run remove-dist && npm run build:static && webpack --mode=development --debug --devtool cheap-source-map --output-pathinfo",
-    "build:production": "npm run remove-dist && npm run build:static && webpack --mode=production",
+    "prebuild:dev": "npm run remove-dist",
+    "build:dev": "webpack --mode=development --debug --devtool cheap-source-map --output-pathinfo",
+    "prebuild:production": "npm run remove-dist",
+    "build:production": "webpack --mode=production",
     "build:test": "npm run build:production -- --include-tests",
-    "build:static": "gulp imagemin",
+    "prebuild:storybook": "npm run remove-dist --silent",
```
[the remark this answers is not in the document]
eugene-manuilov: `--silent` is an NPM flag that suppresses NPM related output. It helps to avoid displaying unnecessary information generated by a subcommand.

For example, this is what you see when you don't use `--silent` for a subcommand:

without--silent

And here is how it looks like when you use `--silent` for a subcommand:

with--silent

Also it worth mentioning that `--silent` doesn't affect non-NPM related output, so storybook logs will be shown even if we use `--silent` flag.

webpack.config.js
eugene-manuilov: No, we can't add hashes because it will break images in the storybook and VRT tests.

bin/release
```diff
@@ -60,34 +64,61 @@ function getGit() {
...
+ */
+async function makeRelease() {
+	const root = dirname( __dirname );
+	const releaseDir = resolve( root, 'release' );
+	const pluginDir = resolve( releaseDir, 'google-site-kit' );
 	const filename = generateFilename();
 
-	// Remove any existing file by the same name.
-	del.sync( path.resolve( __dirname, `../${ filename }` ) );
+	const cp = ( what ) => fse.copy( resolve( root, what ), resolve( pluginDir, what ) );
```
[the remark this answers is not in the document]
eugene-manuilov: It's supposed to target the `pluginDir`. Updated the `cp` function to use `path` instead of `what`.

bin/release
```diff
@@ -60,34 +64,61 @@ function getGit() {
...
+	await fse.ensureDir( pluginDir );
 
-	// eslint-disable-next-line no-console
-	console.log( `Creating ${ filename }` );
+	// Copy files to the release directory.
+	await cp( 'readme.txt' );
+	await cp( 'google-site-kit.php' );
+	await cp( 'dist/assets' );
+	await cp( 'includes' );
+	await cp( 'third-party' );
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

bin/release
eugene-manuilov: `fs.createWriteStream` uses the `w` flag by default which truncates a file if it exists. Anyway added one more `fse.remove` call to remove the file if it exists.

assets/js/components/notifications/dashboard-setup-alerts.js
eugene-manuilov: Updated.

bin/release
eugene-manuilov: Added back.

bin/release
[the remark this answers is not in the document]
eugene-manuilov: Ok, updated it to be `console.error`

assets/js/googlesitekit-base.js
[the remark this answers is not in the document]
aaemnnosttv: Yeah, I saw the comment there but it wasn't immediately obvious what the import does.
Is it because the image won't exist in the `dist` directory if it isn't imported via JS?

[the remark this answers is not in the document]
eugene-manuilov: Yep, if we don't import it here, it won't be loaded by webpack and won't be copied to the dist folder.

webpack.config.js
aaemnnosttv: I understand changes in storybook could break VRT, but changing the filename should have no affect on VRT by itself.

I think the only thing we would need to do is update `stories/notifications.stories.js` to import the images as well like they are in the app now, no?

It's not a big deal, but just trying to understand the obstacles to doing this, and we should try to keep our use in Storybook consistent with the app.

bin/release
aaemnnosttv: Good to know, thanks 👍

bin/release
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/env node
+
+/**
+ * Gulp zip task.
```
aaemnnosttv, line 4: This no longer uses Gulp 😄

bin/release
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * Site Kit by Google, Copyright 2019 Google LLC
```
aaemnnosttv, line 6:
```suggestion
 * Site Kit by Google, Copyright 2020 Google LLC
```

package.json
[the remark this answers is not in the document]
aaemnnosttv: Let's not add the `--silent` flags for now. I agree it's a bit cleaner output but unless there's a reason why this is important to do I'd prefer we don't hide the commands that run by default. We can always pass `--silent` to the command itself at run time.

aaemnnosttv (commented):
Just a few final very minor tweaks and then this should be good to go 👍

package.json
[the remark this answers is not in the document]
eugene-manuilov: Updated.

bin/release
eugene-manuilov: Updated.

webpack.config.js
eugene-manuilov: I haven't looked at it too deeply but storybook puts all images into a different folder. So it will require more time to figure out how to configure it correctly. If you still want it to be implemented, I would suggest creating a new ticket for it.

aaemnnosttv (approved):
Awesome! 🎉

Issue #2093, opened 2020-09-22
Fix accessibility issue for links opening in a new tab


felixarntz:
## Bug Description

Any links that open in a new tab should be marked with an "external link" icon (which mostly is the case now already), but also with a screen reader-only text that says "(opens in a new tab)". Related: #1684 

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All external links in Site Kit should be annotated with an `aria-label` attribute containing `_x( '(opens in a new tab)', 'screen reader text', 'google-site-kit' )`
	* An external link that receives an `aria-label` attribute should have the new tab text appended to it
	* An external link that does not receive a `aria-label` attribute and has string `children` should append the new tab text to the `children` for the `aria-label` attribute
* Any existing links that use the `<span class="screen-reader-text">` element in the children to truncate the link text between visible and hidden portions should be updated to use separate translation strings where the full version is used as the `aria-label` and previously truncated version as the children

## Implementation Brief

* Update `assets/js/components/link.js` to set an `aria-label` attribute based on the logic in the AC
* Update usage of `Link` component with `<span class="screen-reader-text">` to use an `aria-label` attribute instead

<details>
<summary>Original IB</summary>

In the `link (https://github.com/google/site-kit-wp/blob/develop/assets/js/components/link.js)` component (`assets/js/components/link.js`):

- Add the following WordPress dependencies:
  - `import { _x } from '@wordpress/i18n';`
  - `import { VisuallyHidden } from '@wordpress/components';`

- In the render function, after outputting the { children } , conditionally display the "(opens in a new tab)" string. 
`
{ external && (
					<VisuallyHidden as="span" className="screen-reader-text">
						{ _x( '(opens in a new tab)', 'screen reader text', 'google-site-kit' ) }
					</VisuallyHidden>
				) }
`

</details>

## QA Brief
- Inspect any external link using Chrome devtools. They are easily recognisable by having the external icons.
- The anchor tag should have the aria-label attribute with the value of the text within the link and the following text: "(opens in a new tab)"
- No visual regressions should be noted.

## Changelog entry

* Fix accessibility issue with links opening in a new tab by annotating them with screen reader text informing about it.

aaemnnosttv:
IB looks good for the most part, although I think it would be a good idea to use a dedicated component for screen reader text. Gutenberg has a `VisuallyHidden` component in `@wordpress/components` which is really flexible. It would be great if we could use that.

[...]

PR #2163, opened 2020-10-13
Enhancement/2093 accessibility label

asvinb:
## Summary

Addresses issue #2093 

## Relevant technical choices

Use the `VisuallyHidden` component from `@wordpress/components` to render label for screen readers, informing users when links open in a new tab. However since we are overwriting the props of the VisuallyHidden component, as per @tofumatt 's [suggestion](https://github.com/google/site-kit-wp/pull/2163#discussion_r504030464), for now, we can use a span tag with the screen-reader-text class name.
However when adding the span with the screen-reader-text class name within links, it introduced some visual regressions. Thus we opted to use aria-label instead.

## Checklist


assets/js/components/link.js
```diff
@@ -73,6 +75,12 @@ class Link extends Component {
 				{ ...extraProps }
 			>
 				{ children }
+
+				{ external && (
+					<VisuallyHidden as="span" className="screen-reader-text">
```
tofumatt, line 80: Using WordPress's `VisuallyHidden` is a `div` and not a `span` makes it less useful (as we'll frequently want visually-hidden text to accompany other text inside a block-level element that shouldn't hold a `div`, like here or in an `li`/`a` tag).

That's not the end-of-the-world, but combined with the custom `className` used here, there's really no reason to use `VisuallyHidden` over a `span`. It communicates intent, but it's not actually `VisuallyHidden` on its own at all.

From the looks of it, we have many other `span` elements using `screen-reader-text`, which isn't even defined in our CSS anyway. Importing `VisuallyHidden` here buys us nothing but inflates the size of our bundle, so let's replace `<VisuallyHidden as="span" className="screen-reader-text">` with a `<span className="screen-reader-text">` for now.

If there's demand for it we can make a follow-up issue to add our own `VisuallyHidden` component that's lighter-weight than Gutenberg's and will suit our CSS/usage. I've filed #2165 to track this.

tofumatt (changes requested):
Looks like our IB didn't take into account we'd have to override both Gutenberg defaults to work with our setup, so let's revert to `span` tags for now.

I've filed #2165 to track our own `VisuallyHidden` component, which I think will make more sense.

assets/js/components/link.js
asvinb: Makes sense @tofumatt . I've updated the PR. Can you check again?

aaemnnosttv (dismissed):
LGTM 👍

aaemnnosttv:
@tofumatt can you take another look and see if anything else is needed here on your end?

aaemnnosttv (changes requested):
@asvinb – this is failing VRT now due to changes where the output is slightly different.

Can you take a look and see if there's anything reasonable we can do there?

assets/js/components/link.js
```diff
@@ -73,6 +74,12 @@ class Link extends Component {
 				{ ...extraProps }
 			>
 				{ children }
+
+				{ external && (
+					<span className="screen-reader-text">
+						{ _x( '(opens in a new tab)', 'screen reader text', 'google-site-kit' ) }
```
tofumatt, line 80: Tested this in a screenreader and it didn't quite work.

We need either:

```suggestion
						{ _x( ' (opens in a new tab)', 'screen reader text', 'google-site-kit' ) }
```

or 

```suggestion
						{ ' ' }{ _x( '(opens in a new tab)', 'screen reader text', 'google-site-kit' ) }
```

This is because otherwise React will remove the whitespace and text like:

```js
<Link className="googlesitekit-help-link" href={ url } external>Need help?</Link>
```

will output to: `Need help(opens in a new tab)`.

So there needs to be whitespace here to separate the text from the parens. I get caught by React's clever whitespace removal sometimes 😅 

I'd prefer using the separate whitespace, as it makes it easier for translators to understand and this approach already has some i18n downfalls, but is good enough for now.

After that this is good to go.

tofumatt (changes requested):
Just one small tweak needed.

assets/js/components/setup/compatibility-checks.js
```diff
@@ -122,35 +122,39 @@ export default class CompatibilityChecks extends Component {
...
+				labelHTML: __( 'Configure', 'google-site-kit' ),
+				a11yLabel: __( 'the helper plugin', 'google-site-kit' ),
 				href: configureURL,
 				external: false,
 			};
 		}
 		return {
-			labelHTML: __( 'Learn how<span class="screen-reader-text"> to install and use the helper plugin</span>', 'google-site-kit' ),
+			labelHTML: __( 'Learn how', 'google-site-kit' ),
+			a11yLabel: __( 'to install and use the helper plugin', 'google-site-kit' ),
```
felixarntz, line 149: Translatable sentences should never be split or concatenated in any way - they need to be maintained in one string like they were before. For this purpose there is `createInterpolateElement`, would be great if you had a look at that.

assets/js/modules/adsense/components/setup/SetupAccountNoClient.js
```diff
@@ -46,14 +45,13 @@ export default function SetupAccountNoClient() {
...
-							ALLOWED_ATTR: [ 'class' ],
-						}
-					) }
-				/>
+				>
+					{ __( 'Learn more', 'google-site-kit' ) }
+					{ ' ' }
+					<span className="screen-reader-text">
+						{ __( 'about updating your AdSense account', 'google-site-kit' ) }
+					</span>
```
felixarntz, line 53: See above, this also shouldn't be split.

assets/js/modules/adsense/components/setup/SetupAccountCreate.js
```diff
@@ -97,14 +96,13 @@ export default function SetupAccountCreate() {
...
-							ALLOWED_ATTR: [ 'class' ],
-						}
-					) }
-				/>
+				>
+					{ __( 'Learn more', 'google-site-kit' ) }
+					{ ' ' }
+					<span className="screen-reader-text">
+						{ __( 'about adding a user to an existing AdSense account', 'google-site-kit' ) }
+					</span>
```
felixarntz, line 104: See above.

assets/js/modules/analytics/components/common/AnonymizeIPSwitch.js
```diff
@@ -66,14 +65,13 @@ export default function AnonymizeIPSwitch() {
...
-							ALLOWED_ATTR: [ 'class' ],
-						}
-					) }
-				/>
+				>
+					{ __( 'Learn more', 'google-site-kit' ) }
+					{ ' ' }
+					<span className="screen-reader-text">
+						{ __( 'about IP anonymization.', 'google-site-kit' ) }
+					</span>
```
felixarntz, line 73: See above.

assets/js/modules/pagespeed-insights/components/common/MetricsLearnMoreLink.js
```diff
@@ -25,21 +25,19 @@ import { __ } from '@wordpress/i18n';
...
-					ALLOWED_ATTR: [ 'class' ],
-				}
-			) }
-		/>
+		>
+			{ __( 'Learn more', 'google-site-kit' ) }
+			{ ' ' }
+			<span className="screen-reader-text">
+				{ __( 'how metrics are measured.', 'google-site-kit' ) }
+			</span>
```
felixarntz, line 40: See above.

felixarntz (changes requested):
Keep in mind that translatable strings should never be partial sentences that are concatenated or combined in any way - this won't work with the majority of languages.

assets/js/components/setup/compatibility-checks.js
asvinb: Thanks for the info @felixarntz . Code has been updated to use `createInterpolateElement` :)

assets/js/modules/adsense/components/setup/SetupAccountNoClient.js
asvinb: Updated to use `createInterpolateElement`

assets/js/modules/adsense/components/setup/SetupAccountCreate.js
asvinb: Updated to use `createInterpolateElement`

assets/js/modules/analytics/components/common/AnonymizeIPSwitch.js
asvinb: Updated to use `createInterpolateElement`

assets/js/modules/pagespeed-insights/components/common/MetricsLearnMoreLink.js
asvinb: Updated to use `createInterpolateElement`

asvinb:
@aaemnnosttv Can you take another look. As per our conversation, we are now using aria-label instead of having a span with the screen-reader-text class name since it introduced some visual regressions.
cc @felixarntz

assets/js/components/link.js
```diff
@@ -25,9 +25,24 @@ import classnames from 'classnames';
...
+	getArialLabel() {
+		const {
+			children,
+			external,
+		} = this.props;
+
+		let label;
+		if ( external && typeof children === 'string' ) {
+			label = `${ children } ${ _x( '(opens in a new tab)', 'screen reader text', 'google-site-kit' ) }`;
+		}
```
aaemnnosttv, line 41: This should respect an [`aria-label` prop](https://reactjs.org/docs/accessibility.html#wai-aria) if passed, and only use `children` as a fallback (see other comments below).

I.e.
- if `aria-label`, append the new tab text
- if no `aria-label`, but string `children`, append the new tab text to `children`
- if no `aria-label` or string `children` return the new tab text

assets/js/components/setup/compatibility-checks.js
```diff
@@ -122,27 +122,47 @@ export default class CompatibilityChecks extends Component {
 
 		if ( ! installed && installURL ) {
 			return {
-				labelHTML: __( 'Install<span class="screen-reader-text"> the helper plugin</span>', 'google-site-kit' ),
+				labelHTML: createInterpolateElement(
+					__( 'Install<span> the helper plugin</span>', 'google-site-kit' ),
+					{
+						span: <span className="screen-reader-text" />,
+					}
+				),
```
aaemnnosttv, line 130: Since we're using `aria-label` for the full context of the element's label, I don't think `createInterpolateElement` is necessary here anymore.

Instead, let's pass through separate strings for `children` and `aria-label` then these objects can be simply spread into the `Link` component since they're essentially objects of props now.

E.g.
```suggestion
				children: __( 'Install', 'google-site-kit' ),
				'aria-label': __( 'Install the helper plugin', 'google-site-kit' ),
```

assets/js/modules/adsense/components/setup/SetupAccountCreate.js
```diff
@@ -97,14 +96,16 @@ export default function SetupAccountCreate() {
...
-						{
-							ALLOWED_TAGS: [ 'span' ],
-							ALLOWED_ATTR: [ 'class' ],
-						}
-					) }
-				/>
+				>
+					{
+						createInterpolateElement(
+							__( 'Learn more<span> about adding a user to an existing AdSense account</span>', 'google-site-kit' ),
```
aaemnnosttv, line 102: Same for uses like this, let's use `aria-label` for the full sentence (without html), and leave the children here to only the partial text.

aaemnnosttv (changes requested):
Thanks @asvinb this is looking good, just a few details to address and then I think this should be good to go. I'm going to update the ACs to reflect the change to use `aria-label` as discussed.

assets/js/components/link.js
asvinb: @aaemnnosttv Updated. Can you check?

assets/js/components/setup/compatibility-checks.js
asvinb: @aaemnnosttv Updated. Can you check?

assets/js/modules/adsense/components/setup/SetupAccountCreate.js
asvinb: @aaemnnosttv Updated. Can you check?

assets/js/components/Link.js
```diff
@@ -35,13 +40,29 @@ function Link( {
...
+
+		if ( ariaLabel === '' && external && typeof children === 'string' ) {
+			label = children;
+		}
+
+		if ( external ) {
+			label += ` ${ newTabLabel }`;
+		}
+
+		return label;
```
aaemnnosttv, line 63: This could be simplified by checking `external` first and if not, return `ariaLabel` right away since there's no benefit to duplicating the string text content as the `aria-label` otherwise. Then the two current `external` checks could be removed.

assets/js/components/setup/compatibility-checks.js
```diff
@@ -122,35 +122,39 @@ export default class CompatibilityChecks extends Component {
 
 		if ( ! installed && installURL ) {
 			return {
-				labelHTML: __( 'Install<span class="screen-reader-text"> the helper plugin</span>', 'google-site-kit' ),
+				ariaLabel: __( 'Install the helper plugin', 'google-site-kit' ),
+				label: __( 'Install', 'google-site-kit' ),
```
aaemnnosttv, line 126: I mentioned this before but why not use `children` instead and spread into the `Link` below since these would all be props now? Not a big deal, just curious if there is a reason to do it this way instead.

aaemnnosttv (commented):
@asvinb this looks good for the most part and is just about good to go. One suggested change for you and another question regarding a previous suggestion that I made which I saw you implemented differently.

assets/js/components/Link.js
asvinb: @aaemnnosttv We should cater for the scenario where both, ariaLabel is provided and the link is external. So, even if the ariaLabel is supplied, if it's external, we should append the `(opens in a new tab) ` text to the passed on ariaLabel prop. Whereas what you're suggesting is that is that we just check whether we check for external, if not, return ariaLabel.

assets/js/components/setup/compatibility-checks.js
asvinb: @aaemnnosttv Makes sense :) I didn't know about the children props :(

assets/js/components/Link.js
aaemnnosttv: @asvinb I think you misunderstood my comment. I agree we should support a provided `ariaLabel` if the link is `external` as well. My point was that the function could return early if `external === false` and then the remaining conditionals (for when `external` is true) could be simplified, and the second conditional would no longer be needed and the `newTabLabel` would be unconditionally appended.

Something like this:
```js
let label = ariaLabel;

if ( ! external ) {
	return label;
} else if ( typeof children === 'string' ) {
	label = label || children;
}

return `${ label } ${ _x( '(opens in a new tab)', 'screen reader text', 'google-site-kit' ) }`;
```

asvinb: Ah now I see what you meant. Updated @aaemnnosttv

assets/js/components/Link.js
```diff
@@ -35,13 +40,29 @@ function Link( {
 	caps,
 	danger,
 	disabled,
+	ariaLabel,
```
aaemnnosttv, line 43: Sorry @asvinb but I made a mistake here, and the prop for this should actually be `aria-label` as React doesn't treat these the same way as other html attributes (and actually warns about passing it as camel-case):

> Note that all `aria-*` HTML attributes are fully supported in JSX. Whereas most DOM properties and attributes in React are camelCased, these attributes should be hyphen-cased (also known as kebab-case, lisp-case, etc) as they are in plain HTML

I realize it's passed to the anchor this way but using different names for the prop could result in this one being overridden by a `aria-label` prop by the spread of `extraProps`.

```suggestion
	'aria-label': ariaLabel,
```
This would need to be updated elsewhere too but should be a straightforward rename.

assets/js/components/Link.js
```diff
@@ -90,6 +110,7 @@ Link.propTypes = {
 	caps: PropTypes.bool,
 	danger: PropTypes.bool,
 	disabled: PropTypes.bool,
+	ariaLabel: PropTypes.string,
```
aaemnnosttv, line 113: I think we can remove this rather than rename it since it isn't really a custom prop but an HTML attribute we pass through.
```suggestion
```

assets/js/components/setup/compatibility-checks.js
```diff
@@ -122,35 +122,38 @@ export default class CompatibilityChecks extends Component {
 
 		if ( ! installed && installURL ) {
 			return {
-				labelHTML: __( 'Install<span class="screen-reader-text"> the helper plugin</span>', 'google-site-kit' ),
+				ariaLabel: __( 'Install the helper plugin', 'google-site-kit' ),
```
aaemnnosttv, line 125:
```suggestion
				'aria-label': __( 'Install the helper plugin', 'google-site-kit' ),
```

assets/js/components/setup/compatibility-checks.js
```diff
@@ -122,35 +122,38 @@ export default class CompatibilityChecks extends Component {
...
+				children: __( 'Install', 'google-site-kit' ),
 				href: installURL,
 				external: false,
 			};
 		}
 		if ( installed && ! active && activateURL ) {
 			return {
-				labelHTML: __( 'Activate<span class="screen-reader-text"> the helper plugin</span>', 'google-site-kit' ),
+				ariaLabel: __( 'Activate the helper plugin', 'google-site-kit' ),
```
aaemnnosttv, line 133:
```suggestion
				'aria-label': __( 'Activate the helper plugin', 'google-site-kit' ),
```

assets/js/components/setup/compatibility-checks.js
```diff
@@ -122,35 +122,38 @@ export default class CompatibilityChecks extends Component {
...
+				children: __( 'Activate', 'google-site-kit' ),
 				href: activateURL,
 				external: false,
 			};
 		}
 		if ( installed && active && configureURL ) {
 			return {
-				labelHTML: __( 'Configure<span class="screen-reader-text"> the helper plugin</span>', 'google-site-kit' ),
+				ariaLabel: __( 'Configure the helper plugin', 'google-site-kit' ),
```
aaemnnosttv, line 141:
```suggestion
				'aria-label': __( 'Configure the helper plugin', 'google-site-kit' ),
```

assets/js/components/setup/compatibility-checks.js
```diff
@@ -122,35 +122,38 @@ export default class CompatibilityChecks extends Component {
...
+				children: __( 'Configure', 'google-site-kit' ),
 				href: configureURL,
 				external: false,
 			};
 		}
 		return {
-			labelHTML: __( 'Learn how<span class="screen-reader-text"> to install and use the helper plugin</span>', 'google-site-kit' ),
+			ariaLabel: __( 'Learn how to install and use the helper plugin', 'google-site-kit' ),
```
aaemnnosttv, line 148:
```suggestion
			'aria-label': __( 'Learn how to install and use the helper plugin', 'google-site-kit' ),
```

assets/js/modules/adsense/components/setup/SetupAccountCreate.js
```diff
@@ -97,14 +96,10 @@ export default function SetupAccountCreate() {
...
 					external
-					dangerouslySetInnerHTML={ sanitizeHTML(
-						__( 'Learn more<span class="screen-reader-text"> about adding a user to an existing AdSense account</span>', 'google-site-kit' ),
-						{
-							ALLOWED_TAGS: [ 'span' ],
-							ALLOWED_ATTR: [ 'class' ],
-						}
-					) }
-				/>
+					ariaLabel={ __( 'Learn more about adding a user to an existing AdSense account', 'google-site-kit' ) }
```
aaemnnosttv, line 99:
```suggestion
					aria-label={ __( 'Learn more about adding a user to an existing AdSense account', 'google-site-kit' ) }
```

assets/js/modules/adsense/components/setup/SetupAccountNoClient.js
```diff
@@ -46,14 +45,10 @@ export default function SetupAccountNoClient() {
...
 					external
-					dangerouslySetInnerHTML={ sanitizeHTML(
-						__( 'Learn more<span class="screen-reader-text"> about updating your AdSense account</span>', 'google-site-kit' ),
-						{
-							ALLOWED_TAGS: [ 'span' ],
-							ALLOWED_ATTR: [ 'class' ],
-						}
-					) }
-				/>
+					ariaLabel={ __( 'Learn more about updating your AdSense account', 'google-site-kit' ) }
```
aaemnnosttv, line 48:
```suggestion
					aria-label={ __( 'Learn more about updating your AdSense account', 'google-site-kit' ) }
```

assets/js/modules/analytics/components/common/AnonymizeIPSwitch.js
```diff
@@ -66,14 +65,10 @@ export default function AnonymizeIPSwitch() {
...
 					inherit
-					dangerouslySetInnerHTML={ sanitizeHTML(
-						__( 'Learn more<span class="screen-reader-text"> about IP anonymization.</span>', 'google-site-kit' ),
-						{
-							ALLOWED_TAGS: [ 'span' ],
-							ALLOWED_ATTR: [ 'class' ],
-						}
-					) }
-				/>
+					ariaLabel={ __( 'Learn more about IP anonymization.', 'google-site-kit' ) }
```
aaemnnosttv, line 68:
```suggestion
					aria-label={ __( 'Learn more about IP anonymization.', 'google-site-kit' ) }
```

assets/js/modules/pagespeed-insights/components/common/MetricsLearnMoreLink.js
```diff
@@ -25,21 +25,16 @@ import { __ } from '@wordpress/i18n';
...
 			inherit
-			dangerouslySetInnerHTML={ sanitizeHTML(
-				__( 'Learn more<span class="screen-reader-text"> how metrics are measured.</span>', 'google-site-kit' ),
-				{
-					ALLOWED_TAGS: [ 'span' ],
-					ALLOWED_ATTR: [ 'class' ],
-				}
-			) }
-		/>
+			ariaLabel={ __( 'Learn more how metrics are measured.', 'google-site-kit' ) }
```
aaemnnosttv, line 35:
```suggestion
			aria-label={ __( 'Learn more how metrics are measured.', 'google-site-kit' ) }
```

aaemnnosttv (commented):
Thanks for your patience on this one @asvinb – I think this is the last change or two needed (mostly around the prop name used for the label) which should be very quick to make. I left suggestions on everything so you can apply them all together if nothing else.

asvinb:
> Thanks for your patience on this one @asvinb – I think this is the last change or two needed (mostly around the prop name used for the label) which should be very quick to make. I left suggestions on everything so you can apply them all together if nothing else.

@aaemnnosttv I made the suggested updates. Can you check?

aaemnnosttv (approved):
👍 👍 👍

aaemnnosttv:
@felixarntz @tofumatt would you please take a quick look since you requested changes previously?

assets/js/components/Link.js
```diff
@@ -35,13 +40,26 @@ function Link( {
...
 	// so if disabled we force it to be a button.
 	const isAnchor = href && ! disabled;
 	const SemanticLink = isAnchor ? 'a' : 'button';
 
+	const getAriaLabel = () => {
+		let label = ariaLabel;
+
+		if ( ! external ) {
+			return label;
+		} else if ( typeof children === 'string' ) {
```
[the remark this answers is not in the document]
tofumatt: (It seems like it might be fine, I'm just not sure 🙂 )

tofumatt (changes requested):
At first I thought this has a potential issue with the ACs, resulting in the aria-labels not actually being rendered.

But it turns out that the `aria-label` is always being used directly as supplied when provided. Which is good, and means this component is actually behaving as-expected, but to me in an unpredictable way:

The output for the link element is:

```js
		<SemanticLink
			className={ classnames(
				'googlesitekit-cta-link',
				className,
				{
					'googlesitekit-cta-link--arrow': arrow,
					'googlesitekit-cta-link--external': external,
					'googlesitekit-cta-link--inverse': inverse,
					'googlesitekit-cta-link--back': back,
					'googlesitekit-cta-link--small': small,
					'googlesitekit-cta-link--inherit': inherit,
					'googlesitekit-cta-link--caps': caps,
					'googlesitekit-cta-link--danger': danger,
					'googlesitekit-cta-link--disabled': disabled,
				},
			) }
			href={ isAnchor ? href : undefined }
			target={ isAnchor && external ? '_blank' : undefined }
			rel={ external ? 'noopener noreferrer' : undefined }
			disabled={ disabled }
			aria-label={ getAriaLabel() }
			// This causes the aria-label to always be overwritten, rendering the above function moot if an aria-label prop is supplied.
			{ ...extraProps }
		>
			{ children }
		</SemanticLink>
```

That works, but it's quite _unexpected behaviour_ to me reading the code.

If we have special logic to determine a prop's output, it should appear _after_ the prop-spreading we do on components like `SemanticLink`.

I'd say it's best to refactor this a _bit_ more so the `getAriaLabel()` function is _always_ what is output into `aria-label`.

assets/js/components/Link.js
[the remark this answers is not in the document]
aaemnnosttv: > But in this case, if `children` is a string it will override the ARIA label, even when one was provided.

That's not the case, because the very next line has `label || children` 😄 

I suppose we could change the conditional but I thought it was actually easier to read like that but apparently not!

`children` can be a string (or a function or anything else), so this works as expected in my testing. I'll add a few tests here that I wrote to check this so we don't need to worry about it going forward 🙂

assets/js/components/Link.js
```diff
@@ -63,6 +81,7 @@ function Link( {
 			target={ isAnchor && external ? '_blank' : undefined }
 			rel={ external ? 'noopener noreferrer' : undefined }
 			disabled={ disabled }
+			aria-label={ getAriaLabel() }
```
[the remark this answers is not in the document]
aaemnnosttv: Generally I agree with your thoughts here, the only thing that justifies the function is that you would either need to mutate the prop when appending (frowny face) or create another temporary variable for the value which is the route taken within the function. I think it's a little cleaner to keep it encapsulated there but not a big deal either way.

aaemnnosttv:
@tofumatt – regarding [your comment above](url)

> This causes the aria-label to always be overwritten, rendering the above function moot if an aria-label prop is supplied.

This isn't true since the `aria-label` prop is destructured out of `props` it just may not seem that way due to the necessary alias. The `extraProps` which are spread into the component afterward only contain the rest of the props which weren't destructured out so it's both safe to spread after as well as consistent with how the other props are applied.

I'll add a few tests that I wrote to check a few of your uncertainties that would be good to include anyways.

Let's get any remaining details clarified on a call if needed since this one has taken quite a bit longer than originally anticipated.

aaemnnosttv:
@googlebot I consent.

tofumatt (approved):
After another look through, and a chat with @aaemnnosttv, this actually looks good. Let's 🚢

Issue #2046, opened 2020-09-16
Implement new UI for splash screen


felixarntz:
The splash screen should receive the new UI design previously created.

<img width="717" alt="Screenshot 2020-09-15 at 17 18 14" src="https://user-images.githubusercontent.com/3531426/93277748-75569980-f777-11ea-9224-82ada15fc978.png">

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* If the `featureFlags.userInput` is enabled, the Site Kit splash screen UI should be modified to look like [this](https://www.figma.com/file/K4MXDvLRULqRRz2zpMlh5e/SiteKit---Phase-1%3A-Onboarding?node-id=1098%3A17463). (Otherwise, the current UI should be displayed.)
* The copy for all cases except the `else` case in the current `SetupProxy` component should be maintained, except that the button should _always_ say "Sign in with Google". Only for the `else` case the copy from the new design should be used.
    * Let's also fix a somewhat off description here for the `revoked` case, it shouldn't say `click "Start Setup"`, it should say `click "Sign in with Google"`, because there is no "Start Setup" button.

## Implementation Brief

 
In the [SetupUsingProxy component:](https://github.com/google/site-kit-wp/blob/develop/assets/js/components/setup/setup-proxy.js)
 - add the `classnames` package as external dependency. `import classnames from 'classnames';`
 - import the `SvgIcon` component from ../../util. `import { SvgIcon, trackEvent } from '../../util';`
 - set `startSetupText` as a constant which will have the following value: "**Sign in with Google**" and remove the `startSetupText` variable from the conditions.
 - update the description for the "revoked" condition to: "**Site Kit will no longer have access to your account. If you’d like to reconnect Site Kit, click "_Sign in with Google_" below to generate new credentials.**"
 - update the title and description for the else condition to the following:
 -- title: "**Set up Site Kit**"
 -- description: "**Get insights about how people find and use your site, how to improve and monetize your content, directly in your WordPress dashboard**"
 - add the following markup for the walking person image just after the `.mdc-layout-grid__inner` element. The image should be previously be exported as a SVG in the `/assets/svg/` folder.
```
{ featureFlags.widgets.userInput.enabled && (
		<div
			className="
				googlesitekit-wizard-progress__icon
				mdc-layout-grid__cell
				mdc-layout-grid__cell--span-12-tablet
				mdc-layout-grid__cell--span-6-desktop
			"
		>
[...]

PR #2164, opened 2020-10-13
Feature/2046 splash ui

asvinb:
## Summary

Addresses issue #2046 

## Relevant technical choices

Conditionally set layout specific class names if in development mode to show the updated UI with the image.

## Checklist


assets/js/components/setup/setup-proxy.js
```diff
@@ -75,28 +77,24 @@ function SetupUsingProxy() {
 
 	let title;
 	let description;
-	let startSetupText;
+	const startSetupText = __( 'Sign in with Google', 'google-site-kit' );
```
eugene-manuilov, line 80: Let's move it to where it's used in the markup.

eugene-manuilov (changes requested):
LGTM, there is just one thing to improve:

assets/js/components/setup/setup-proxy.js
asvinb: Updated @eugene-manuilov

eugene-manuilov (approved):

asvinb:
@eugene-manuilov Just FYI, VR tests failing because of an image not loading.

assets/sass/components/dashboard/wizard/_googlesitekit-wizard-progress.scss
```diff
@@ -37,4 +37,28 @@
 			padding: $grid-gap-desktop 0 0 0;
 		}
 	}
+
+	.googlesitekit-wizard-progress__content {
+		margin: 0 auto;
+		max-width: 465px;
```
felixarntz, line 43: Just curious, why did you choose this maximum width specifically? Feels a bit like an arbitrary value, maybe this should be in a variable somewhere?

assets/sass/components/dashboard/wizard/_googlesitekit-wizard-progress.scss
```diff
@@ -37,4 +37,28 @@
 			padding: $grid-gap-desktop 0 0 0;
 		}
 	}
```
felixarntz, line 39: The rules in this file almost all apply to the legacy setup flow which is only visible when _not_ using the proxy (only relevant for development or staging sites). The `SetupProxy` component only uses `.googlesitekit-wizard-progress`, which is some legacy remainder. We don't need to change that at this point, but let's rather decouple the new classes from that and make them based on `.googlesitekit-setup__*` (and put them into `_googlesitekit-setup.scss`).

felixarntz (changes requested):
Mostly LGTM, just 2 comments and one other thing: The margin above and below the opt-in checkbox is a bit small and looks odd, quite different from the mocks. Can you make that margin be the same `1rem` like what is around the description?

assets/sass/components/dashboard/wizard/_googlesitekit-wizard-progress.scss
asvinb: @felixarntz So the max width is as per the provided designs. If you look at the dashboard setup here: https://www.figma.com/file/f07vatCPHzubuJoXkgd5RU/SiteKit-Phase-1_-Onboarding?node-id=1098%3A17677 , the content is centered with a max width of 465px. I replaced it by a variable though. Let me know if that works for you or do you think we should get rid of the max width or shall we follow the design closely?

assets/sass/components/dashboard/wizard/_googlesitekit-wizard-progress.scss
asvinb: Sounds good @felixarntz . I moved the new class names. Can you check again?

asvinb:
> Mostly LGTM, just 2 comments and one other thing: The margin above and below the opt-in checkbox is a bit small and looks odd, quite different from the mocks. Can you make that margin be the same `1rem` like what is around the description?

I added some more breathing room around the opt-in checkbox. Let me know if it's ok now. I specifically added the margin around the opt-in checkbox only if we are showing the person icon since I wanted to leave everything as is if `featureFlags.widgets.userInput.enabled` is false.

felixarntz (approved):
Awesome!

Issue #842, opened 2019-11-12
Silence expected console errors during E2E tests


tofumatt:
Bug Description

The E2E tests produce a lot of "junk" output when running because they have intentional/expected errors that result in `console.error` calls. Jest forwards all console output to the terminal, which results in really noisy test runs where things _look_ like they're not working when in fact they are (see screenshot below).

We should silence expected console output (such as `Failed to load resource: the server responded with a status`) so only relevant errors make it to the console when running the tests.

## Steps to reproduce

1. Run the E2E tests (`npm run test:e2e`)
2. Observe noisy console output

## Screenshots

<img width="673" alt="Screenshot 2019-11-12 at 17 36 18" src="https://user-images.githubusercontent.com/90871/68695436-f26f1c00-0572-11ea-9f24-9e1991c8b1ab.png">

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All e2e tests that result in error logs in the console should be updated so that the console is muted for the specific test.
* The console should _not_ be muted generally throughout all tests, as that would increase our chance of missing problems in a test. It therefore should be done on case-by-case basis.

## Implementation Brief

* Add `'@wordpress/jest-console'` into the `setupFilesAfterEnv` array in `tests/e2e/jest.config.js`
* Update all debugging `console` calls in `config/bootstrap.js` to use `console.debug` instead
* Update E2E tests with the necessary assertions to `expect( console ).` `toHaveErrored`, `toHaveWarned`, `toHaveLogged`.
	- Any of the above messages which are not necessarily expected, but are not relevant can be filtered out and ignored by adding cases for them in the `observeConsoleLogging` function in `bootstrap.js`  
	E.g. React notices, cookie notices, or anything which is more/less unrelated to our tests ([one example where this may be relevant for WP](https://github.com/WordPress/gutenberg/blob/5ced7f1ab18d6019da01f2cb812697ddde8837b7/packages/e2e-tests/config/setup-test-framework.js#L139-L145))

## QA Brief

* E2E tests should produce no extra output when run as shown above (unless there is a legitimate error or failed assertion)

## Changelog entry

* N/A

felixarntz:
@tofumatt I believe this is no longer relevant after #1770? I'll close this for now, but feel free to reopen if I missed something.

felixarntz:
Oh I just noticed this about e2e tests 😆

felixarntz:
IB ✅

[...]

PR #2167, opened 2020-10-13
Add jest console for E2E tests

ivankruchkoff:
## Summary

Addresses issue #842 

## Relevant technical choices

* Most cases only needed batch requests to be stubbed and nothing else
* Removed unconditional debug logging of HTTP responses with 400+ status codes (can still be enabled via `DEBUG_` env vars)
* Added a new `flag-mode` option to build commands to force a particular feature-flag mode regardless of the webpack mode used. This is particularly useful for running e2e tests with the build in watch mode which being a dev build would normally include features which aren't ready to pass e2e yet.

image

## Checklist


ivankruchkoff:
Here's one of these examples of a dual warning, occurring once, and not the other:

Pass 1: https://travis-ci.com/github/google/site-kit-wp/jobs/399050056#L1306

```
● Optimize Activation › prompts to insert your Optimize ID when Analytics snippet is enabled
    expect(jest.fn()).not.toHaveWarned(expected)
    Expected mock function not to be called but it was called with:
    ["Error caught during combinedGet JSHandle@object"]


```
Pass 2: https://travis-ci.com/github/google/site-kit-wp/jobs/399038101#L1294

```
 Optimize Activation › prompts to insert your Optimize ID when Analytics snippet is enabled
    expect(jest.fn()).toHaveWarned(expected)
    Expected mock function to be called.
      91 | 
      92 | 		await finishOptimizeSetup();
    > 93 | 		expect( console ).toHaveWarned();
         | 		                  ^
      94 | 	} );

```

Which compares with what I've been experiencing locally too:
Pass 1 Locally:
```
Optimize Activation › prompts to insert your Optimize ID when Analytics snippet is disabled, with extra instructions

    expect(jest.fn()).not.toHaveWarned(expected)

    Expected mock function not to be called but it was called with:
    ["Error caught during combinedGet JSHandle@object"]


```
Pass 2 Locally:
```
    expect(jest.fn()).toHaveWarned(expected)

    Expected mock function to be called.

      110 |
      111 | 		await finishOptimizeSetup();
    > 112 | 		expect( console ).toHaveWarned();
```

One idea to working around this is to silence errors such as: 
`Error caught during combinedGet JSHandle@object from surfacing a warning/error
`
by adding it to https://github.com/google/site-kit-wp/pull/2167/files#diff-cd20a88b9a0d8ff93f4ebc81383c60b53312bb8aed35909f2ae80ec6094dc924R180

Any other suggestions?
@tofumatt / @aaemnnosttv

aaemnnosttv:
@googlebot I consent.

@ivankruchkoff the problem in this case is related to unrelated requests which weren't currently being handled properly in the environment for this test. In this case it was unhandled batch requests, which are irrelevant for all module set up cases.

It's also worth noting that you will see additional errors when running a development build as we have some additional components behind a feature flag in this case. In the past I usually ran E2E locally with the dev build but we can't do that anymore due to its affect on changing what components are rendered.

I pushed one change here to add this for the Optimize test you referenced but most tests should already have request interception in place, so in most cases you'll only need to update it. Keep in mind it can only be used once per test (you can't call it twice) and must handle all requests.

Hope this helps, let me know if you have any other questions here 😄

ivankruchkoff:
@aaemnnosttv I think applying this useRequestInterception to one test is causing it to fail:
Added here:
https://github.com/google/site-kit-wp/blob/0025d01f891db7a0471b589ceaced774668647be/tests/e2e/specs/modules/pagespeed-insights/activation.test.js#L21

Timed out in three environments:
https://travis-ci.com/github/google/site-kit-wp/jobs/400498054#L1403
https://travis-ci.com/github/google/site-kit-wp/jobs/400498053#L1207
https://travis-ci.com/github/google/site-kit-wp/jobs/400498052#L1059


Also, unrelated to the issue above, I'm seeing some unreliability in tests:
Timeouts:
https://travis-ci.com/github/google/site-kit-wp/jobs/400483220#L1019

Passing in one env, failing in another, intermittently:
https://travis-ci.com/github/google/site-kit-wp/jobs/400498054#L1216
https://travis-ci.com/github/google/site-kit-wp/jobs/400498053#L1180

assets/js/components/data/index.js
```diff
@@ -199,7 +199,7 @@ const dataAPI = {
 		} ).then( ( results ) => {
 			each( results, ( result, key ) => {
 				if ( ! keyIndexesMap[ key ] ) {
-					console.error( 'data_error', 'unknown response key ' + key ); // eslint-disable-line no-console
+					console.debug( 'data_error', 'unknown response key ' + key ); // eslint-disable-line no-console
```
aaemnnosttv, line 202: I've downgraded the severity of the logging here as this is only really relevant for debugging and not something which is really likely to ever happen under normal conditions.

It also happens regularly in E2E for tests where we return a mocked response for a batch request since only uncached request keys will be requested on subsequent calls.

tests/e2e/config/bootstrap.js
```diff
@@ -167,11 +167,31 @@ function observeConsoleLogging() {
 			text.startsWith( 'Failed to decode downloaded font:' ) ||
 			text.startsWith( 'OTS parsing error:' ) ||
 			text.includes( 'Download the React DevTools for a better development experience' ) ||
-			text.includes( 'https://fb.me/react-unsafe-component-lifecycles' )
+			text.includes( 'Can\'t perform a React state update on an unmounted component' ) ||
+			text.includes( 'https://fb.me/react-unsafe-component-lifecycles' ) ||
+			text.includes( 'https://fb.me/react-strict-mode-' )
```
aaemnnosttv, line 172: Some of these are triggered by Gutenberg itself and are good to know about but not something we should be expecting in our tests and by extension require our tests to fail if they weren't called.

tests/e2e/config/bootstrap.js
```diff
@@ -296,7 +316,7 @@ beforeAll( async () => {
 		if ( res.status() > 399 ) {
 			const req = res.request();
 			// eslint-disable-next-line no-console
-			console.warn( res.status(), req.method(), req.url() );
+			console.debug( res.status(), req.method(), req.url() );
```
aaemnnosttv, line 319: Note, the `pageerror` above is kept on `console.error` since these should require a manual exception as they generally are not expected.

https://github.com/google/site-kit-wp/blob/c5e2acfef02e82ecf5752e52368bf90949249e57/tests/e2e/config/bootstrap.js#L302-L304

webpack.config.js
```diff
@@ -208,7 +213,7 @@ const webpackConfig = ( mode ) => {
 					flagsConfig,
 					{
 						modes: [ 'development', 'production' ],
-						mode,
+						flagMode, // Default: mode; override with --flagMode={mode}
```
aaemnnosttv, line 216: This is particularly useful for E2E development as a way to use a development build with `watch` for quick iteration, without causing tests to fail due to extra development-only features.

tests/e2e/config/bootstrap.js
```diff
@@ -167,11 +167,31 @@ function observeConsoleLogging() {
 			text.startsWith( 'Failed to decode downloaded font:' ) ||
 			text.startsWith( 'OTS parsing error:' ) ||
 			text.includes( 'Download the React DevTools for a better development experience' ) ||
-			text.includes( 'https://fb.me/react-unsafe-component-lifecycles' )
+			text.includes( 'Can\'t perform a React state update on an unmounted component' ) ||
```
tofumatt, line 170: Minor nitpick, but 
```suggestion
			text.includes( "Can't perform a React state update on an unmounted component" ) ||
```
is a bit nicer and usually how we deal with quote marks in strings 😅

tests/e2e/config/bootstrap.js
```diff
@@ -167,11 +167,31 @@ function observeConsoleLogging() {
...
+			text.startsWith( 'data_error unknown response key' ) ||
+			text.includes( 'No triggers were found in the config. No analytics data will be sent.' )
+		) {
+			return;
+		}
+
+		// As of WordPress 5.3.2 in Chrome 79, navigating to the block editor
+		// (Posts > Add New) will display a console warning about
+		// non - unique IDs.
+		// See: https://core.trac.wordpress.org/ticket/23165
```
tofumatt, line 190: ❤️ for including the core ticket here 👍🏻

tofumatt (approved):
This looks sooooo much better, thanks! 🎉 

I'm happy with this and think we should  🚢 

Now real errors in our E2E tests will be MUCH easier to spot.

felixarntz (approved):
Awesome work!

Issue #1541, opened 2020-05-11
Dashboard menu item conflict with WP Engine hosted sites


jamesozzie:
## Bug Description

As reported in the [WordPress support forums](https://wordpress.org/support/topic/messes-with-the-wordpress-admin-menu/) with Site Kit activated the WP Engine settings tab is placed below the dashboard tab. 

Without Site Kit active the WP Engine settings tab is placed top of the admin dashboard tabs. 

Unknown what parameters are passed with WP Engine `add_menu_page`, it's a hosted placed entry, not an installed plugin. 

## Steps to reproduce

1. Confirm tab placement of WP Engine settings tab (with WP Engine hosting)
2. Install and active Site Kit
3. Check placement of WP Engine settings tab

## Screenshots

![image](https://user-images.githubusercontent.com/41326532/81550263-e4f6eb00-9377-11ea-97d9-69334a378ca0.png)

 
---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Site Kit's PHP logic that currently moves the WordPress dashboard and the Site Kit menu as the top two items of the overall WordPress admin menu should be updated as follows:
    * The WP dashboard menu item should not be moved (and not be assumed to be at the first position).
    * The Site Kit menu item should always appear right after the WP dashboard menu item, wherever that is.

## Implementation Brief

* Update the hook for the `menu_order` filter in the `includes/Core/Admin/Screens.php` file to do the following:
  * Return early if either `googlesitekit-dashboard` or `index.php` doesn't exist in the list.
  * Move the `googlesitekit-dashboard` slug to be right after `index.php` in the menu order list.
  * Make sure other slug positions are not modified.
* Update tests in the `ScreensTest::test_menu_order` method to account for a case when there might be a third-party-page above the dashboard.

Use this snippet as an example:

  ```php
  function( array $menu_order ) {
      $new_order = array();
      foreach ( $menu_order as $index => $item ) {
        if ( $item !== self::PREFIX . 'dashboard' ) ) {
          $new_order[] = $item;
        }

        if ( 'index.php' === $item ) {
          $new_order[] = self::PREFIX . 'dashboard';
        }
      }

      return $new_order;
  }
  ```

## QA Brief

* Either try out using WP Engine / Bluehost account, or
* Add and activate bluehost plugin https://github.com/bluehost/bluehost-wordpress-plugin, or WP Engine mu-plugins folder.
[...]

PR #2170, opened 2020-10-14
Bug/1541 plugin menu order

ivankruchkoff:
## Summary

Addresses issue #1541 

## Checklist

### DIFF
--- a/includes/Core/Admin/Screens.php
+++ b/includes/Core/Admin/Screens.php
@@ -154,12 +154,20 @@ final class Screens {
 			function( array $menu_order ) {
 				// Move the Site Kit dashboard menu item to be one after the index.php item if it exists.
 				$dashboard_index = array_search( 'index.php', $menu_order, true );
-				$sitekit_index   = array_search( self::PREFIX . 'dashboard', $menu_order, true );
+
+				$sitekit_index = false;
+				foreach ( $menu_order as $key => $value ) {
+					if ( substr( $value, 0, strlen( self::PREFIX ) ) === self::PREFIX ) {
+						$sitekit_index = $key;
+						$sitekit_value = $value;
+					}
+				}
+
 				if ( false === $dashboard_index || false === $sitekit_index ) {
 					return $menu_order;
 				}
 				unset( $menu_order[ $sitekit_index ] );
-				array_splice( $menu_order, $dashboard_index + 1, 0, self::PREFIX . 'dashboard' );
+				array_splice( $menu_order, $dashboard_index + 1, 0, $sitekit_value );
 				return $menu_order;
 			}
 		);
--- a/includes/Core/Admin/Screens.php
+++ b/includes/Core/Admin/Screens.php
@@ -152,14 +152,15 @@ final class Screens {
 		add_filter(
 			'menu_order',
 			function( array $menu_order ) {
-				$new_order = array();
-				foreach ( $menu_order as $index => $item ) {
-					if ( 'index.php' === $item || 0 === strpos( $item, self::PREFIX ) ) {
-						$new_order[] = $item;
-						unset( $menu_order[ $index ] );
-					}
+				// Move the sitekit-dashboard menu item to be one after the index.php item if it exists.
+				$dashboard_index = array_search( 'index.php', $menu_order, true );
+				$sitekit_index   = array_search( self::PREFIX . 'dashboard', $menu_order, true );
+				if ( false === $dashboard_index || false === $sitekit_index ) {
+					return $menu_order;
 				}
-				return array_values( array_merge( $new_order, $menu_order ) );
+				unset( $menu_order[ $sitekit_index ] );
+				array_splice( $menu_order, ++$dashboard_index, 0, self::PREFIX . 'dashboard' );
+				return $menu_order;
 			}
 		);
 	}
--- a/includes/Core/Admin/Screens.php
+++ b/includes/Core/Admin/Screens.php
@@ -152,14 +152,14 @@ final class Screens {
 		add_filter(
 			'menu_order',
 			function( array $menu_order ) {
-				// Move the sitekit-dashboard menu item to be one after the index.php item if it exists.
+				// Move the Site Kit dashboard menu item to be one after the index.php item if it exists.
 				$dashboard_index = array_search( 'index.php', $menu_order, true );
 				$sitekit_index   = array_search( self::PREFIX . 'dashboard', $menu_order, true );
 				if ( false === $dashboard_index || false === $sitekit_index ) {
 					return $menu_order;
 				}
 				unset( $menu_order[ $sitekit_index ] );
-				array_splice( $menu_order, ++$dashboard_index, 0, self::PREFIX . 'dashboard' );
+				array_splice( $menu_order, $dashboard_index + 1, 0, self::PREFIX . 'dashboard' );
 				return $menu_order;
 			}
 		);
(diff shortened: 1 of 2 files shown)


aaemnnosttv (changes requested):
Thanks @ivankruchkoff ! This looks great, just a few very minor tweaks. The main thing is the changes to the tests. Let me know if you have any questions 👍

includes/Core/Admin/Screens.php
```diff
@@ -152,14 +152,15 @@ function() {
 		add_filter(
 			'menu_order',
 			function( array $menu_order ) {
-				$new_order = array();
-				foreach ( $menu_order as $index => $item ) {
-					if ( 'index.php' === $item || 0 === strpos( $item, self::PREFIX ) ) {
-						$new_order[] = $item;
-						unset( $menu_order[ $index ] );
-					}
+				// Move the sitekit-dashboard menu item to be one after the index.php item if it exists.
```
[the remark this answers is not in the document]
ivankruchkoff: Updated

includes/Core/Admin/Screens.php
```diff
@@ -152,14 +152,15 @@ function() {
...
+				$dashboard_index = array_search( 'index.php', $menu_order, true );
+				$sitekit_index   = array_search( self::PREFIX . 'dashboard', $menu_order, true );
+				if ( false === $dashboard_index || false === $sitekit_index ) {
+					return $menu_order;
 				}
-				return array_values( array_merge( $new_order, $menu_order ) );
+				unset( $menu_order[ $sitekit_index ] );
+				array_splice( $menu_order, ++$dashboard_index, 0, self::PREFIX . 'dashboard' );
```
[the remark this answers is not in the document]
ivankruchkoff: Updated the index

tests/phpunit/integration/Core/Admin/ScreensTest.php
```diff
@@ -108,11 +108,19 @@ public function test_menu_order() {
...
+			$menu_order
+		);
+
 		$this->screens->register();
 
 		// Imitate WordPress core running these filters.
 		if ( apply_filters( 'custom_menu_order', false ) ) {
-			$menu_order = apply_filters( 'menu_order', $menu_order );
+			$menu_order            = apply_filters( 'menu_order', $menu_order );
+			$overridden_menu_order = apply_filters( 'menu_order', $overridden_menu_order );
```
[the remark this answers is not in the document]
ivankruchkoff: Added a data provider, added the third case, and gave a one liner for each test case in the data provider.

ivankruchkoff:
Addressed all CR feedback.

aaemnnosttv (commented):
Thanks @ivankruchkoff – this looks great 👍 

There are a few comments below about the data provider which would be good for you to review. They're so small that it doesn't make sense to send back here though so I'll apply those changes and we can get this merged 😄

aaemnnosttv:
@googlebot I consent.

aaemnnosttv (dismissed):
Awesome!

includes/Core/Admin/Screens.php
```diff
@@ -152,14 +152,15 @@ function() {
...
 			function( array $menu_order ) {
-				$new_order = array();
-				foreach ( $menu_order as $index => $item ) {
-					if ( 'index.php' === $item || 0 === strpos( $item, self::PREFIX ) ) {
-						$new_order[] = $item;
-						unset( $menu_order[ $index ] );
-					}
+				// Move the Site Kit dashboard menu item to be one after the index.php item if it exists.
+				$dashboard_index = array_search( 'index.php', $menu_order, true );
+				$sitekit_index   = array_search( self::PREFIX . 'dashboard', $menu_order, true );
```
felixarntz, line 157: I think we cannot rely on `self::PREFIX . 'dashboard'` here - it was only using `self::PREFIX` before for a reason, which is in certain cases the Site Kit menu is not available under the `dashboard` slug but the `dashboard-splash` slug.

We may need to change this to a more dynamic search which finds the entry that starts with `self::PREFIX` instead.

tests/phpunit/integration/Core/Admin/ScreensTest.php
```diff
@@ -99,30 +99,73 @@ public function test_removal_of_admin_notices_outside_sitekit( $hookname ) {
...
 
-	public function test_menu_order() {
-		$menu_order = array(
-			'index.php',
-			'third-party-plugin',
-			'edit.php',
-			'options-general.php',
-			'googlesitekit-dashboard',
+
+	public function menu_order_provider() {
```
felixarntz, line 103: Let's add a test case here that uses another `googlesitekit-`-prefixed value, e.g. `googlesitekit-dashboard-splash`.

felixarntz, line 103: Nit-picking, but WP core (and also most of Site Kit) always names the data provider methods `data_*` similarly to the `test_*` method name (i.e. in this case `data_menu_order`).

felixarntz (changes requested):
The overall approach LGTM, but there's one thing we need to maintain from the way it was before.

tests/phpunit/integration/Core/Admin/ScreensTest.php
aaemnnosttv: I noticed that too (although they have many providers that don't use this convention as well), and we have both in our codebase (I'm pretty sure I've always used `_provider` as a suffix) but I'm happy to standardize on something.

ivankruchkoff: Looked into it, we're at **31 dataProviders**. **17** with **_provider**, **14** with **data_** and this.
```
[/google-site-kit/tests/phpunit/integration]$ ack dataProvider|grep -v vendor|wc -l
      31
[/google-site-kit/tests/phpunit/integration]$ $ ack dataProvider|grep -v vendor|grep _provider |wc -l
      17
[/google-site-kit/tests/phpunit/integration]$ $ ack dataProvider|grep -v vendor|grep data_ |wc -l
      14
```

includes/Core/Admin/Screens.php
ivankruchkoff: Updated the search to check for self::PREFIX

tests/phpunit/integration/Core/Admin/ScreensTest.php
ivankruchkoff: Added the fourth test case.

ivankruchkoff: Updated to `data_menu_order`.

includes/Core/Admin/Screens.php
```diff
@@ -152,14 +152,23 @@ function() {
...
-				foreach ( $menu_order as $index => $item ) {
-					if ( 'index.php' === $item || 0 === strpos( $item, self::PREFIX ) ) {
-						$new_order[] = $item;
-						unset( $menu_order[ $index ] );
+				// Move the Site Kit dashboard menu item to be one after the index.php item if it exists.
+				$dashboard_index = array_search( 'index.php', $menu_order, true );
+
+				$sitekit_index = false;
+				foreach ( $menu_order as $key => $value ) {
+					if ( substr( $value, 0, strlen( self::PREFIX ) ) === self::PREFIX ) {
```
felixarntz, line 160: Maybe simpler would be:
```suggestion
					if ( strpos( $value, self::PREFIX ) === 0 ) {
```

includes/Core/Admin/Screens.php
```diff
@@ -152,14 +152,23 @@ function() {
...
+						$sitekit_index = $key;
+						$sitekit_value = $value;
```
felixarntz, line 162: I think we should only find the first item here, similar to how `array_search` would do it.
```suggestion
						$sitekit_value = $value;
						break;
```

felixarntz (approved):
Awesome!

Issue #2150, opened 2020-10-08
Enforce CTA acronym to be capitalized and fix violations


felixarntz:
Via #1493, ESLint now flags violations of our rule that acronyms should be capitalized. The list of the acronyms to check is manually maintained based on what we're generally using. One acronym missing from there is `CTA`, which we in some places use incorrectly as `Cta` within variable and prop names.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* `CTA` should be one of the acronyms required to be capitalized by the `camelcase-acronyms` ESLint rule.
* All violations to that rule should be fixed, including prop renamings throughout the codebase where needed.

## Implementation Brief

* Add `CTA` as to list of `acronyms` variable in `packages/eslint-plugin/rules/camelcase-acronyms.js`.
* Run ESLint throughout the codebase.
* Fix all flagged violations (e.g. by renaming `footerCtaLink` to `footerCTALink`).

### Test Coverage

* 

### Visual Regression Changes

* N/A

## QA Brief

- No visual changes should occur as a result of these changes.

## Changelog entry

* N/A

aaemnnosttv:
IB ✅

ShubhamPalriwala:
I'd like to resolve this issue if it is open !

tofumatt:
QA ✅

Tried a variable with `Cta` and it failed, as expected 👍🏻

PR #2176, opened 2020-10-15
Rename cta acronym to CTA.

asvinb:
## Summary

Addresses issue #2150 

## Relevant technical choices

- Rename all cta occurrences in the code by CTA
- Update eslint rules to include CTA acronym

## Checklist


assets/js/components/layout/layout.js
```diff
@@ -62,15 +62,15 @@ class Layout extends Component {
 				{ header &&
 					<LayoutHeader
 						title={ title }
-						ctaLabel={ headerCtaLabel }
-						ctaLink={ headerCtaLink }
+						CTALabel={ headerCTALabel }
+						CTALink={ headerCTALink }
```
aaemnnosttv, line 66: The prop names for `LayoutHeader` shouldn't change here because the acronym capitalization only affects the letters/word when it falls in the middle of a variable where the first character would already be capitalized per normal camel case style.

Any variables that _start_ with, or are an acronym by themselves should remain lowercase. E.g. `url` or `ampMode` are correct forms of these variables containing acronyms.

Also, it looks like the case wasn't changed in the `LayoutHeader` and `LayoutFooter` components to match, but again – the prop names shouldn't change, just making an observation here.

assets/js/components/layout/layout.js
```diff
@@ -62,15 +62,15 @@ class Layout extends Component {
...
 					/>
 				}
 				{ children }
 				{ footer &&
 					<LayoutFooter
-						ctaLabel={ footerCtaLabel }
-						ctaLink={ footerCtaLink }
+						CTALabel={ footerCTALabel }
+						CTALink={ footerCTALink }
```
aaemnnosttv, line 73: Same as above for `LayoutFooter`.

aaemnnosttv (changes requested):
Thanks @asvinb - just a few minor corrections to make and then this should be good to go 👍

aaemnnosttv:
I've opened a new issue to avoid this in the future: https://github.com/google/site-kit-wp/issues/2195

assets/js/components/layout/layout.js
asvinb: This is what happens when you just blindly do a search and replace 🤦 .
Thanks for catching that @aaemnnosttv . This has been updated

assets/js/components/layout/layout.js
asvinb: Updated @aaemnnosttv

asvinb:
Looking into the failed VRT

aaemnnosttv (approved):
LGTM 👍

PR #2202, opened 2020-10-16
Refactor settings with registered components

aaemnnosttv:
## Summary

Addresses issue #1623

## Checklist


assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -58,18 +58,18 @@ const moduleDefaults = {
...
-				...moduleDefaults,
-				name: slug, // Ensure `name` is not empty.
-				...modules[ slug ],
-				slug,
-			};
+			const module = { ...modules[ slug ], slug };
+			// Fill any `undefined` values with defaults.
+			defaults( module, { name: slug }, moduleDefaults );
+
+			return module;
```
aaemnnosttv, line 72: Note this fixes a minor bug where defaults would be overwritten by an `undefined` value which will still take precedence when spreading into an object.

assets/js/modules/adsense/index.js
```diff
@@ -88,6 +85,14 @@ addFilter(
 );
 
 domReady( () => {
+	Data.dispatch( CORE_MODULES ).registerModule(
+		'adsense',
+		{
+			settingsEditComponent: SettingsEdit,
+			settingsViewComponent: SettingsView,
+		}
+	);
```
felixarntz, line 94: Note that #2074 introduces `Modules.registerModule`, which should be used here. To make this smoother, can you add that as part of here too (just copy https://github.com/google/site-kit-wp/pull/2129/files#diff-813cf127a74b2b233ac59519257a5d7c4079deb9707cca867e4c7a5e06d16f40R39) and use it throughout?

assets/js/util/helpers.js
```diff
@@ -65,14 +65,14 @@ export const createAddToFilter = ( NewComponent ) => {
...
- * 	addFilter( `googlesitekit.ModuleSettingsDetails-${slug}`,
- * 		'googlesitekit.AdSenseModuleSettingsDetails',
- * 		fillFilterWithComponent( AdSenseSettings, {
- * 			onSettingsPage: true,
- * 		} ) );
+ * 	addFilter(
+ * 		`googlesitekit.someFilter`,
+ * 		'googlesitekit.someNamespace',
+ * 		fillFilterWithComponent( SomeComponent, extraProps )
+ * );
```
felixarntz, line 75: Good catch 👍

felixarntz (changes requested):
Mostly LGTM, just some minor feedback.

assets/js/components/settings/SettingsRenderer.js
```diff
@@ -25,18 +25,21 @@ import { useEffect } from '@wordpress/element';
...
+import { STORE_NAME as CORE_MODULES } from '../../googlesitekit/modules/datastore/constants';
 const { useSelect, useDispatch } = Data;
+const NullComponent = () => null;
+
+export default function SettingsRenderer( { slug, isOpen, isEditing } ) {
+	const storeName = `modules/${ slug }`;
+	const isDoingSubmitChanges = useSelect( ( select ) => select( storeName )?.isDoingSubmitChanges?.() );
+	const haveSettingsChanged = useSelect( ( select ) => select( storeName )?.haveSettingsChanged?.() );
+	const SettingsEdit = useSelect( ( select ) => select( CORE_MODULES ).getModule( slug )?.settingsEditComponent ) || NullComponent;
+	const SettingsView = useSelect( ( select ) => select( CORE_MODULES ).getModule( slug )?.settingsViewComponent ) || NullComponent;
```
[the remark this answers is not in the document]
aaemnnosttv: > Does `?.` work on `null` values like it does on `undefined`? Otherwise we'd run into problems here.

Yes; from the docs: 
image

> I think this would be easier to read by assigning only the raw selector return value into a `const module`, and then we could take care of the `settingsEditComponent` and `settingsViewComponent` from there.

I'll update it to use a single select instead 👍

assets/js/components/settings/SettingsRenderer.js
```diff
@@ -25,18 +25,21 @@ import { useEffect } from '@wordpress/element';
...
-	const haveSettingsChanged = useSelect( ( select ) => select( STORE_NAME ).haveSettingsChanged() );
 	// Rollback any temporary selections to saved values if settings have changed and no longer editing.
-	const { rollbackSettings } = useDispatch( STORE_NAME );
+	const { rollbackSettings } = useDispatch( storeName ) || {};
 	useEffect( () => {
-		if ( haveSettingsChanged && ! isDoingSubmitChanges && ! isEditing ) {
+		if ( rollbackSettings && haveSettingsChanged && ! isDoingSubmitChanges && ! isEditing ) {
 			rollbackSettings();
 		}
 	}, [ haveSettingsChanged, isDoingSubmitChanges, isEditing ] );
```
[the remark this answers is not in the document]
aaemnnosttv: I think generally, yes but this is now it was before. I only added the check to make sure the action exists before calling it. I'll add it to the array though since there should be no harm in that, but in this case it shouldn't matter.

assets/js/modules/adsense/index.js
aaemnnosttv: Sounds good, updated 👍

felixarntz (approved):
Awesome!

Issue #2023, opened 2020-09-14
Implement search box widget in Site Kit dashboard


felixarntz:
The last widget missing to complete the Site Kit dashboard with the new Widget API is the search box:

**Without Analytics active:**
![Screenshot 2020-09-14 at 12 42 43](https://user-images.githubusercontent.com/3531426/93130788-e3bc2e80-f687-11ea-8767-4d640302f047.png)

**With Analytics active:**
![Screenshot 2020-09-14 at 12 40 34](https://user-images.githubusercontent.com/3531426/93130808-e9b20f80-f687-11ea-81a1-918676a1eb23.png)

Since this widget is not specific to any module, it should be part of the core widgets API layer as a default.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The "Search for individual page or post information" widget should be re-implemented in `assets/js/googlesitekit/widgets/components/URLSearchWidget.js`. (The old widget should be left untouched and kept in place for the current/old UI.)
* The widget should match the functionality and behavior as of today. It should **not** use the legacy `Layout` component though (related #2022).
* The widget should be registered in the `registerDefaults` function into `AREA_DASHBOARD_POPULARITY`, with:
    * priority `100`
    * width `WIDGET_WIDTHS.HALF` (this is acceptable for now; it currently renders in half or full width based on other active widgets, but that shouldn't be a concern for now)
    * wrapWidget based on what makes most sense for implementation

## Implementation Brief

* Add a new `URLSearchWidget` component in `assets/js/googlesitekit/widgets/components/URLSearchWidget.js`
    - Based on the current/legacy implementation in `assets/js/components/PostSearcher.js`
        - Should not have Analytics-specific styling or logic internally
    - Should use the new wrapper component/strategy to be introduced in #2022 rather than the legacy `<Layout>` component internally
* Register the widget in `assets/js/googlesitekit/widgets/register-defaults.js`
    * Area: `AREA_DASHBOARD_POPULARITY`
    * Priority: `100`
    * Width: `WIDGET_WIDTHS.HALF`
    * Title: `Search for individual page or post information`
    * Subtitle: none
    * `wrapWidget` (dependent on results of #2022)

This issue should depend on #2022 to avoid the need for the legacy `Layout`.

#### Tests

* Add Storybook story/stories for the new widget `stories/dashboard.stories.js`
[...]

PR #2207, opened 2020-10-19
Feature/2023 implement search box widget

caseymorrisus:
## Summary

Addresses issue [#2023](https://github.com/google/site-kit-wp/issues/2023)

## Checklist


stories/dashboard.stories.js
```diff
@@ -98,6 +99,23 @@ storiesOf( 'Dashboard', module )
...
+	.add( 'URL Search Widget', () => {
+		const setupRegistry = ( { dispatch } ) => {
+			dispatch( CORE_SITE ).receiveSiteInfo( {
+				usingProxy: true,
+				referenceSiteURL: 'http://example.com',
+				adminURL: 'http://example.com/wp-admin',
+				timezone: 'America/Detroit',
+				siteName: 'My Site Name',
+			} );
+		};
```
eugene-manuilov, line 111: We can just use `provideSiteInfo` helper:

```suggestion
		const setupRegistry = ( registry ) => {
			provideSiteInfo( registry );
		};
```

eugene-manuilov (changes requested):
Almost LGTM, added a few suggestions.

caseymorrisus:
@eugene-manuilov Added updates based on PR feedback. VRT test failed but does not seem to be related.

assets/js/googlesitekit/widgets/components/URLSearchWidget.js
```diff
@@ -0,0 +1,88 @@
...
+	const [ canSubmit, setCanSubmit ] = useState( false );
+	const [ match, setMatch ] = useState( {} );
+
+	const detailsURL = useSelect( ( select ) => select( CORE_SITE ).getAdminURL( 'googlesitekit-dashboard', { permaLink: match?.permalink } ) );
+
+	const onClick = useCallback( () => {
+		if ( match?.permalink ) {
+			global.location.assign( detailsURL );
+		}
+	}, [ detailsURL, match ] );
```
eugene-manuilov, line 46: @caseymorrisus we can't use objects as a dependency, let's use permalink's value instead:

```suggestion
	}, [ detailsURL, match?.permalink ] );
```

eugene-manuilov (changes requested):
@caseymorrisus it's almost ready from my perspective. There is just one thing that we need to update. Could you please check it?

assets/js/googlesitekit/widgets/components/URLSearchWidget.js
caseymorrisus: @eugene-manuilov I think it's fine to use an object as a dependency here and avoid variable assignment. Deps for hooks use referential equality comparison for objects. If we were comparing 2 completely different objects this would be an issue but as we're comparing an object to a possible future version of itself, if it is equal, the reference will also be equal. Match is also only set as a whole object meaning that any change would also modify the reference.

eugene-manuilov: Ok, sounds good then!

eugene-manuilov (dismissed):

assets/js/googlesitekit/widgets/register-defaults.js
```diff
@@ -108,4 +109,12 @@ export function registerDefaults( widgetsAPI ) {
 		style: WIDGET_AREA_STYLES.BOXES,
 		priority: 4,
 	}, CONTEXT_PAGE_DASHBOARD );
+
+	widgetsAPI.registerWidget( AREA_PAGE_DASHBOARD_POPULARITY, {
+		title: __( 'Search for individual page or post information', 'google-site-kit' ),
+		style: WIDGET_AREA_STYLES.BOXES,
+		priority: 100,
+		width: WIDGET_WIDTHS.HALF,
+		component: URLSearchWidget,
```
felixarntz, line 118: This should pass `wrapWidget: false`, otherwise the `Widget` manually rendered by the component will be wrapped in another `Widget`.

felixarntz (changes requested):
Code-wise this mostly looks good (except the few comments), however there seems to be something off here, as I cannot get this to show up in my dashboard at all. I only see the legacy component, but not the new component among the other widgets. 🤔

assets/js/googlesitekit/widgets/register-defaults.js
```diff
@@ -110,11 +111,10 @@ export function registerDefaults( widgetsAPI ) {
...
-	widgetsAPI.registerWidget( AREA_PAGE_DASHBOARD_POPULARITY, {
-		title: __( 'Search for individual page or post information', 'google-site-kit' ),
-		style: WIDGET_AREA_STYLES.BOXES,
+	widgetsAPI.registerWidget( WIDGET_URL_SEARCH, {
 		priority: 100,
 		width: WIDGET_WIDTHS.HALF,
 		component: URLSearchWidget,
-	}, CONTEXT_PAGE_DASHBOARD );
+		wrapWidget: false,
+	}, [ AREA_DASHBOARD_POPULARITY, AREA_PAGE_DASHBOARD_POPULARITY ] );
```
felixarntz, line 119: Oh now I see the confusion here - the ACs were wrong. This should only be available for `AREA_DASHBOARD_POPULARITY`, similarly how it is today. I've fixed that in the ACs, where it was incorrectly referring to `AREA_PAGE_DASHBOARD_POPULARITY`.

felixarntz (changes requested):

felixarntz (approved):
Great! Quickly fixed the last feedback from above.

Issue #2110, opened 2020-09-28
Fix Search Console deep links for domain properties to filter only current site


felixarntz:
## Bug Description

#1917 fixed data for Search Console domain properties to only be displayed for the current site. However, now Search Console deep links to the service frontend are still not considering this, resulting in potentially inconsistent information. Following up on #1923, if the site is using a Search Console domain property, all Search Console URLs deep links (except those that are already for specific URLs) should be updated to include an extra argument to only include URLs that are part of the site.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All deep links to `https://search.google.com/u/1/search-console/performance/search-analytics` in Site Kit widgets (both new and legacy) should be updated to include an extra query parameter `page=*{siteURL}` where `{siteURL}` is the reference site URL, without trailing slash at the end.
    * The query parameter should only be added if the site uses a Search Console domain property (starting in `sc-domain:`).
    * The exception to this are deep links for single URL details view where that `page` query parameter is already added based on the current entity URL.

## Implementation Brief

* Add an `isDomainProperty()` selector to `modules/search-console` data store that offers a boolean return value to check if a property is a Domain Property:
```js
isDomainProperty: createRegistrySelector( ( select ) => () => {
	const domain = select( STORE_NAME ).getPropertyID();

	return domain && domain.startsWith( 'sc-domain:' );
} ),
```
* Add relevant tests for the above selector and include it in the Search Console data store.
* Use `useSelect( () => select('core/site').getReferenceSiteURL() )` to get the `referenceSiteURL`, then remove the trailing slash from it if needed. The rest of the IB assumes it's stored in a `referenceSiteURL` variable.
* Add a <code>page: \`*${ referenceSiteURL }\`</code> param to the `query` param for all `getServiceURL()` selector calls with URLs (`path` params) using `/performance/search-analytics`, _if_ `select( 'modules/search-console' ).isDomainProperty() === true`. (And as long as they don't have an existing `page` param, see the ACs.)

## QA Brief

* Install Site Kit and set it up with a site that uses a Search Console domain property and has Search Console data.
* Search Console deep link URLs (in both new widgets, via development build, as well as legacy widgets) should be updated with the new `page=*$REFERENCE_SITE_URL` query params, unless they are for a specific URL (in which case they, as before, should have a `page=!$SPECIFIC_URL` query param).

## Changelog entry

* Fix Search Console deep links to point to correct locations in case of a domain property as well.

[...]

PR #2229, opened 2020-10-22
Add `page` param to deep links for SC domain properties

kostyalmm:
## Summary

Addresses issue #2110 

## Relevant technical choices

1. Adds a `isDomainProperty` selector to `search-console` module.  Relevant tests added also.
2. When the domain property is a Search Domain property, add `page` param to the Search Console deep link URL.

## Checklist

### DIFF
--- a/assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
+++ b/assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
@@ -38,12 +38,20 @@ const { useSelect } = Data;
 const DashboardDetailsWidgetKeywordsTable = () => {
 	const propertyID = useSelect( ( select ) => select( STORE_NAME ).getPropertyID() );
 	const url = useSelect( ( select ) => select( CORE_SITE ).getCurrentEntityURL() );
+	const isDomainProperty = useSelect( ( select ) => select( STORE_NAME ).isDomainProperty() );
+	let referenceSiteURL = useSelect( ( select ) => select( CORE_SITE ).getReferenceSiteURL() );
+	if ( referenceSiteURL.endsWith( '/' ) ) {
+		// remove the trailing slash
+		referenceSiteURL = referenceSiteURL.slice( 0, -1 );
+	}
 	const footerCTALinkArgs = {
 		resource_id: propertyID,
 		num_of_days: getCurrentDateRangeDayCount(),
 	};
 	if ( url ) {
 		footerCTALinkArgs.page = `!${ url }`;
+	} else if ( isDomainProperty ) {
+		footerCTALinkArgs.page = `*${ referenceSiteURL }`;
 	}
 	const footerCTALink = useSelect( ( select ) => select( STORE_NAME ).getServiceURL( {
 		path: '/performance/search-analytics',
--- a/assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
+++ b/assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
@@ -32,6 +32,7 @@ import Layout from '../../../../components/layout/layout';
 import { STORE_NAME } from '../../datastore/constants';
 import { STORE_NAME as CORE_SITE } from '../../../../googlesitekit/datastore/site/constants';
 import { getCurrentDateRangeDayCount } from '../../../../util/date-range';
+import { unTrailingSlashIt } from '../../../../util';
 
 const { useSelect } = Data;
 
@@ -39,18 +40,14 @@ const DashboardDetailsWidgetKeywordsTable = () => {
 	const propertyID = useSelect( ( select ) => select( STORE_NAME ).getPropertyID() );
 	const url = useSelect( ( select ) => select( CORE_SITE ).getCurrentEntityURL() );
 	const isDomainProperty = useSelect( ( select ) => select( STORE_NAME ).isDomainProperty() );
-	let referenceSiteURL = useSelect( ( select ) => select( CORE_SITE ).getReferenceSiteURL() );
-	if ( referenceSiteURL.endsWith( '/' ) ) {
-		// remove the trailing slash
-		referenceSiteURL = referenceSiteURL.slice( 0, -1 );
-	}
+	const referenceSiteURL = useSelect( ( select ) => unTrailingSlashIt( select( CORE_SITE ).getReferenceSiteURL() ) );
 	const footerCTALinkArgs = {
 		resource_id: propertyID,
 		num_of_days: getCurrentDateRangeDayCount(),
 	};
 	if ( url ) {
 		footerCTALinkArgs.page = `!${ url }`;
-	} else if ( isDomainProperty ) {
+	} else if ( isDomainProperty && referenceSiteURL ) {
 		footerCTALinkArgs.page = `*${ referenceSiteURL }`;
 	}
 	const footerCTALink = useSelect( ( select ) => select( STORE_NAME ).getServiceURL( {
--- a/assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
+++ b/assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
@@ -32,6 +32,7 @@ import Layout from '../../../../components/layout/layout';
 import { STORE_NAME } from '../../datastore/constants';
 import { STORE_NAME as CORE_SITE } from '../../../../googlesitekit/datastore/site/constants';
 import { getCurrentDateRangeDayCount } from '../../../../util/date-range';
+import { untrailingslashit } from '../../../../util';
 
 const { useSelect } = Data;
 
@@ -42,8 +43,15 @@ const DashboardDetailsWidgetKeywordsTable = () => {
 		resource_id: propertyID,
 		num_of_days: getCurrentDateRangeDayCount(),
 	};
+	const isDomainProperty = useSelect( ( select ) => select( STORE_NAME ).isDomainProperty() );
+	const referenceSiteURL = useSelect( ( select ) => {
+		return untrailingslashit( select( CORE_SITE ).getReferenceSiteURL() );
+	} );
 	if ( url ) {
 		footerCTALinkArgs.page = `!${ url }`;
+	}
+	if ( isDomainProperty && referenceSiteURL ) {
+		footerCTALinkArgs.page = `*${ referenceSiteURL }`;
 	}
 	const footerCTALink = useSelect( ( select ) => select( STORE_NAME ).getServiceURL( {
 		path: '/performance/search-analytics',
--- a/assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
+++ b/assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
@@ -49,8 +49,7 @@ const DashboardDetailsWidgetKeywordsTable = () => {
 	} );
 	if ( url ) {
 		footerCTALinkArgs.page = `!${ url }`;
-	}
-	if ( isDomainProperty && referenceSiteURL ) {
+	} else if ( isDomainProperty && referenceSiteURL ) {
 		footerCTALinkArgs.page = `*${ referenceSiteURL }`;
 	}
 	const footerCTALink = useSelect( ( select ) => select( STORE_NAME ).getServiceURL( {
--- a/assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
+++ b/assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
@@ -32,23 +32,18 @@ import Layout from '../../../../components/layout/layout';
 import { STORE_NAME } from '../../datastore/constants';
 import { STORE_NAME as CORE_SITE } from '../../../../googlesitekit/datastore/site/constants';
 import { getCurrentDateRangeDayCount } from '../../../../util/date-range';
-import { unTrailingSlashIt } from '../../../../util';
 
 const { useSelect } = Data;
 
 const DashboardDetailsWidgetKeywordsTable = () => {
 	const propertyID = useSelect( ( select ) => select( STORE_NAME ).getPropertyID() );
 	const url = useSelect( ( select ) => select( CORE_SITE ).getCurrentEntityURL() );
-	const isDomainProperty = useSelect( ( select ) => select( STORE_NAME ).isDomainProperty() );
-	const referenceSiteURL = useSelect( ( select ) => unTrailingSlashIt( select( CORE_SITE ).getReferenceSiteURL() ) );
 	const footerCTALinkArgs = {
 		resource_id: propertyID,
 		num_of_days: getCurrentDateRangeDayCount(),
 	};
 	if ( url ) {
 		footerCTALinkArgs.page = `!${ url }`;
-	} else if ( isDomainProperty && referenceSiteURL ) {
-		footerCTALinkArgs.page = `*${ referenceSiteURL }`;
 	}
 	const footerCTALink = useSelect( ( select ) => select( STORE_NAME ).getServiceURL( {
 		path: '/performance/search-analytics',
--- a/assets/js/modules/search-console/components/dashboard/DashboardClicksWidget.js
+++ b/assets/js/modules/search-console/components/dashboard/DashboardClicksWidget.js
@@ -28,7 +28,7 @@ import Data from 'googlesitekit-data';
 import { STORE_NAME } from '../../datastore/constants';
 import { STORE_NAME as CORE_SITE } from '../../../../googlesitekit/datastore/site/constants';
 import { STORE_NAME as CORE_USER } from '../../../../googlesitekit/datastore/user/constants';
-import { extractForSparkline } from '../../../../util';
+import { extractForSparkline, untrailingslashit } from '../../../../util';
 import { trackEvent } from '../../../../util/tracking';
 import { extractSearchConsoleDashboardData } from '../../util';
 import whenActive from '../../../../util/when-active';
@@ -47,6 +47,8 @@ function DashboardClicksWidget() {
 
 		const propertyID = store.getPropertyID();
 		const url = select( CORE_SITE ).getCurrentEntityURL();
+		const isDomainProperty = select( STORE_NAME ).isDomainProperty();
+		const referenceSiteURL = untrailingslashit( select( CORE_SITE ).getReferenceSiteURL() );
 
 		const args = {
 			dimensions: 'date',
@@ -61,6 +63,9 @@ function DashboardClicksWidget() {
 		if ( url ) {
 			args.url = url;
 			serviceBaseURLArgs.page = `!${ url }`;
+		}
+		if ( isDomainProperty && referenceSiteURL ) {
+			serviceBaseURLArgs.page = `*${ referenceSiteURL }`;
 		}
 
 		return {
--- a/assets/js/modules/search-console/components/dashboard/DashboardClicksWidget.js
+++ b/assets/js/modules/search-console/components/dashboard/DashboardClicksWidget.js
@@ -63,8 +63,7 @@ function DashboardClicksWidget() {
 		if ( url ) {
 			args.url = url;
 			serviceBaseURLArgs.page = `!${ url }`;
-		}
-		if ( isDomainProperty && referenceSiteURL ) {
+		} else if ( isDomainProperty && referenceSiteURL ) {
 			serviceBaseURLArgs.page = `*${ referenceSiteURL }`;
 		}
 
--- a/assets/js/modules/search-console/components/dashboard/DashboardImpressionsWidget.js
+++ b/assets/js/modules/search-console/components/dashboard/DashboardImpressionsWidget.js
@@ -29,7 +29,7 @@ import { STORE_NAME } from '../../datastore/constants';
 import { STORE_NAME as CORE_SITE } from '../../../../googlesitekit/datastore/site/constants';
 import { STORE_NAME as CORE_USER } from '../../../../googlesitekit/datastore/user/constants';
 import { extractSearchConsoleDashboardData } from '../../util';
-import { extractForSparkline } from '../../../../util';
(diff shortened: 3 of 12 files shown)


assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
```diff
@@ -38,12 +38,20 @@ const { useSelect } = Data;
 const DashboardDetailsWidgetKeywordsTable = () => {
 	const propertyID = useSelect( ( select ) => select( STORE_NAME ).getPropertyID() );
 	const url = useSelect( ( select ) => select( CORE_SITE ).getCurrentEntityURL() );
+	const isDomainProperty = useSelect( ( select ) => select( STORE_NAME ).isDomainProperty() );
+	let referenceSiteURL = useSelect( ( select ) => select( CORE_SITE ).getReferenceSiteURL() );
+	if ( referenceSiteURL.endsWith( '/' ) ) {
```
tofumatt, line 43: This is likely where the E2E tests are failing: note that https://github.com/google/site-kit-wp/blob/50ffe5232a0046632f78bf50f88a0ee615eee5d2/assets/js/googlesitekit/datastore/site/info.js#L323 can return `undefined`, which means this can possibly execute `undefined.endsWith( '/' )`.

That's the error I'm seeing in Travis: https://travis-ci.com/github/google/site-kit-wp/jobs/404472155#L1773

```suggestion
	if ( referenceSiteURL && referenceSiteURL.endsWith( '/' ) ) {
```

would do the trick 😄

kostyalmm: @tofumatt I actually tried that but the build still failed. So I have reverted that commits a few hrs back.
With the logic you proposed, I saw this [build error](https://travis-ci.com/github/google/site-kit-wp/jobs/403824791#L1773)

As I understand, I have to load `siteInfo` to the test environment. I don't really know how to do that. I tried searching code for similar use cases but couldn't find anything. Can you please help me with it :)

assets/js/modules/search-console/components/dashboard/LegacyDashboardWidgetPopularKeywordsTable.js
```diff
@@ -37,18 +37,29 @@ import {
 } from '../../util';
 import { getCurrentDateRangeDayCount } from '../../../../util/date-range';
 import { STORE_NAME } from '../../datastore/constants';
+import { STORE_NAME as CORE_SITE } from '../../../../googlesitekit/datastore/site';
```
tofumatt, line 40: This was the culprit of the failing tests and it not loading properly. We recently removed these exports from the main module and moved them to use `/constants`. Looks like this one needs updating, and then it's good to go 😄 
```suggestion
import { STORE_NAME as CORE_SITE } from '../../../../googlesitekit/datastore/site/constants';
```

kostyalmm: @tofumatt it seems to have worked 😃 . But the test still fails because of timeouts. Should I be worried about that? My other PR fails for the same reason. 🤔

aaemnnosttv: @kostyalmm It looks like it was just one of the 3 E2E jobs so I restarted it. These can be a bit unstable, but usually they all fail if there's something actually wrong. In rare cases, there is a WP version-related failure (since this is different between jobs) but you can usually tell by the error there too because the unstable tests always fail in the same way.

assets/js/modules/search-console/components/dashboard/GoogleSitekitSearchConsoleDashboardWidget.js
```diff
@@ -58,13 +59,23 @@ const GoogleSitekitSearchConsoleDashboardWidget = () => {
 	const [ loading, setLoading ] = useState( true );
 	const dateRange = useSelect( ( select ) => select( CORE_USER ).getDateRange() );
 	const propertyID = useSelect( ( select ) => select( STORE_NAME ).getPropertyID() );
+	const isDomainProperty = useSelect( ( select ) => select( STORE_NAME ).isDomainProperty() );
+	let referenceSiteURL = useSelect( ( select ) => select( CORE_SITE ).getReferenceSiteURL() );
+	if ( referenceSiteURL.endsWith( '/' ) ) {
+		// remove the trailing slash
+		referenceSiteURL = referenceSiteURL.slice( 0, -1 );
+	}
```
felixarntz, line 67: Since we're using this in multiple places now, I think it would make sense to add a simple `untrailingslashit` utility function, almost similar to what WP core has in PHP. To keep our code simple, let's in that function only strip the trailing slash and even do the check if the passed value is a string. This way we can just run the return value of the `getReferenceSiteURL` selector through it immediately, i.e. we can use a `const` for `referenceSiteURL` above.

felixarntz (changes requested):
Overall LGTM, however I think we can simplify this a bit by introducing a utility function for removing the trailing slash.

assets/js/modules/search-console/components/dashboard/GoogleSitekitSearchConsoleDashboardWidget.js
kostyalmm: I added this function to `assets/js/util/sanitize.js` as it seemed fitting to me. Please let me know if you have a better place in mind.

assets/js/modules/search-console/datastore/service.js
```diff
@@ -53,6 +55,19 @@ export const selectors = {
 		}
 		return addQueryArgs( baseURI, queryArgs );
 	} ),
+
+	/**
+	 * Checks whether the property is a Search Console Domain Property.
+	 *
+	 * @since 1.18.0
```
aaemnnosttv, line 62: Minor but important detail: new `@since` annotations need to always use the literal placeholder `n.e.x.t` since the release version is always unknown. This is already inaccurate though as the most recent release is `1.19.0`, so this wouldn't be released before `1.20.0` but still the placeholder should be used instead as the actual version is applied when the release is prepared.

assets/js/util/sanitize.js
```diff
@@ -8,3 +8,25 @@ export const sanitizeHTML = ( unsafeHTML, domPurifyConfig = {} ) => {
...
+
+/**
+ * Takes a URL string, removes the trailing stash if any and returns it.
+ *
+ * @since 1.18.0
+ * @private
+ *
+ * @param {string} url The URL with or without trailing slash.
+ * @return {string|null} The URL string after removing the trailing slash.
+ */
```
aaemnnosttv, line 20: A few points here:
* This utility shouldn't be URL-specific. Like it's PHP counterpart, it's a basic string utility.
* The `@since` annotation should use the `n.e.x.t` placeholder
* I feel like the `T` maybe shouldn't be capitalized, but not a big deal

assets/js/util/sanitize.js
```diff
@@ -8,3 +8,25 @@ export const sanitizeHTML = ( unsafeHTML, domPurifyConfig = {} ) => {
...
+	if ( typeof url !== 'string' ) {
+		return null;
+	}
+
+	if ( url.endsWith( '/' ) ) {
+		// remove the trailing slash
+		return url.slice( 0, -1 );
+	}
+
+	return url;
```
aaemnnosttv, line 31: Similar to the PHP function, I think this utility should remove any number of trailing slashes rather than assuming there is only one.

Similarly, I think this can be reduced to a simple one-liner:
```suggestion
	return string?.replace?.( /\/+$/, '' );
```

In this case, it would return `undefined` if the given input is not a string, which could have a default handled with `|| null` or `|| ''`, but I'm okay with it returning `undefined` as well since this is an unexpected value.

assets/js/util/sanitize.test.js
```diff
@@ -29,3 +29,19 @@ describe( 'sanitizeHTML', () => {
...
+
+	it( 'should change a URL with trailing slash', () => {
+		const url = 'http://example.org/';
+		expect( unTrailingSlashIt( url ) ).toEqual( 'http://example.org' );
+	} );
+
+	it( 'should return null if the parameter is not a string', () => {
+		expect( unTrailingSlashIt( 1 ) ).toBeNull();
+	} );
+} );
```
aaemnnosttv, line 47: Let's add a case or two here to show that it removes more than one consecutive slash at the end if present. E.g. `http://` -> `http:`

aaemnnosttv (changes requested):
This looks pretty good to me, just a few comments regarding some minor details that were missed, mostly around documentation, as well as a few suggestions and thoughts around the new utility.

assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
```diff
@@ -32,18 +32,23 @@ import Layout from '../../../../components/layout/layout';
...
 const DashboardDetailsWidgetKeywordsTable = () => {
 	const propertyID = useSelect( ( select ) => select( STORE_NAME ).getPropertyID() );
 	const url = useSelect( ( select ) => select( CORE_SITE ).getCurrentEntityURL() );
+	const isDomainProperty = useSelect( ( select ) => select( STORE_NAME ).isDomainProperty() );
+	const referenceSiteURL = useSelect( ( select ) => unTrailingSlashIt( select( CORE_SITE ).getReferenceSiteURL() ) );
 	const footerCTALinkArgs = {
 		resource_id: propertyID,
 		num_of_days: getCurrentDateRangeDayCount(),
 	};
 	if ( url ) {
```
tofumatt, line 48: The `getCurrentEntityURL()` selector [will use `getReferenceSiteURL()` as a fallback](https://github.com/google/site-kit-wp/blob/5e51067e3a4bc93f7cd5419d798f6ee2d0f56713/assets/js/googlesitekit/datastore/site/info.js#L351-L353), so I don't see how this check could ever be `false` but have the `else if` be `true`.

assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
```diff
@@ -32,18 +32,23 @@ import Layout from '../../../../components/layout/layout';
...
 		footerCTALinkArgs.page = `!${ url }`;
+	} else if ( isDomainProperty && referenceSiteURL ) {
+		footerCTALinkArgs.page = `*${ referenceSiteURL }`;
```
tofumatt, line 51: Seems like this needs to be a separate `if` statement, because this `else if` can't ever be truthy as I understand the docs for the `getCurrentEntityURL()` selector.

```suggestion
	}
	
	if ( isDomainProperty && referenceSiteURL ) {
		footerCTALinkArgs.page = `*${ referenceSiteURL }`;
```

It would be worth adding a component test to ensure the CTA link here is correct when `isDomainProperty()` returns `true`.

assets/js/util/sanitize.js
```diff
@@ -8,3 +8,25 @@ export const sanitizeHTML = ( unsafeHTML, domPurifyConfig = {} ) => {
...
+};
```
tofumatt, line 32: I would expect this to remove _all_ trailing slashes, not just the last one. The [WordPress function removes all slashes](https://developer.wordpress.org/reference/functions/untrailingslashit), so we should probably match that functionality.

assets/js/util/sanitize.test.js
```diff
@@ -29,3 +29,19 @@ describe( 'sanitizeHTML', () => {
 		} );
 	} );
 } );
+
+describe( 'unTrailingSlashIt', () => {
+	it( 'should not change a URL without trailing slash', () => {
+		const url = 'http://example.org';
+		expect( unTrailingSlashIt( url ) ).toEqual( 'http://example.org' );
+	} );
```
tofumatt, line 37: Defining the constant here doesn't add much to the test, I would say it actually impedes readability here. Let's make these all one-liners to make them easier to scan 🙂 
```suggestion
	it( 'should not change a URL without trailing slash', () => {
		expect( unTrailingSlashIt( 'http://example.org' ) ).toEqual( 'http://example.org' );
	} );
```

assets/js/util/sanitize.test.js
```diff
@@ -29,3 +29,19 @@ describe( 'sanitizeHTML', () => {
...
+
+	it( 'should change a URL with trailing slash', () => {
+		const url = 'http://example.org/';
+		expect( unTrailingSlashIt( url ) ).toEqual( 'http://example.org' );
+	} );
+
+	it( 'should return null if the parameter is not a string', () => {
+		expect( unTrailingSlashIt( 1 ) ).toBeNull();
+	} );
```
tofumatt, line 46: Let's add an extra test for extra trailing slashes, just to be complete:

```suggestion
	it( 'should return null if the parameter is not a string', () => {
		expect( unTrailingSlashIt( 1 ) ).toBeNull();
	} );
	
	it( 'should remove multiple trailing slashes, like the WordPress PHP function', () => {
		expect( unTrailingSlashIt( 'http://example.org/foo// ) ).toEqual( 'http://example.org/foo' );
	} );
```

assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
```diff
@@ -32,18 +32,23 @@ import Layout from '../../../../components/layout/layout';
...
 import { getCurrentDateRangeDayCount } from '../../../../util/date-range';
+import { unTrailingSlashIt } from '../../../../util';
 
 const { useSelect } = Data;
 
 const DashboardDetailsWidgetKeywordsTable = () => {
 	const propertyID = useSelect( ( select ) => select( STORE_NAME ).getPropertyID() );
 	const url = useSelect( ( select ) => select( CORE_SITE ).getCurrentEntityURL() );
+	const isDomainProperty = useSelect( ( select ) => select( STORE_NAME ).isDomainProperty() );
+	const referenceSiteURL = useSelect( ( select ) => unTrailingSlashIt( select( CORE_SITE ).getReferenceSiteURL() ) );
```
tofumatt, line 43: Not a big deal, but if we're going to wrap the result of selectors, lets make it not a one-liner to improve readability. This is pretty dense code. How about:
```suggestion
	const referenceSiteURL = useSelect( ( select ) => {
		return unTrailingSlashIt( select( CORE_SITE ).getReferenceSiteURL() );
	} );
```

tofumatt (changes requested):
I think there's a bug with this implementation—were you able to verify that the `footerCTALinkArgs.page` value changed with this code?

I think it needs some adjustments before it'll work.

assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
kostyalmm: I looked at the IB again and found that we don't even need to add the `page` param in this component.

assets/js/modules/search-console/components/dashboard/GoogleSitekitSearchConsoleDashboardWidget.js
```diff
@@ -58,13 +60,21 @@ const GoogleSitekitSearchConsoleDashboardWidget = () => {
 	const [ loading, setLoading ] = useState( true );
 	const dateRange = useSelect( ( select ) => select( CORE_USER ).getDateRange() );
 	const propertyID = useSelect( ( select ) => select( STORE_NAME ).getPropertyID() );
+	const isDomainProperty = useSelect( ( select ) => select( STORE_NAME ).isDomainProperty() );
+	const referenceSiteURL = useSelect( ( select ) => {
+		return unTrailingSlashIt( select( CORE_SITE ).getReferenceSiteURL() );
+	} );
```
felixarntz, line 66: This can be in a single line like other simple `useSelect` usages.

assets/js/modules/search-console/components/dashboard/LegacyDashboardWidgetPopularKeywordsTable.js
```diff
@@ -37,18 +37,27 @@ import {
...
+import { STORE_NAME as CORE_SITE } from '../../../../googlesitekit/datastore/site/constants';
 const { useSelect } = Data;
 
 const LegacyDashboardWidgetPopularKeywordsTable = ( props ) => {
 	const { data } = props;
 	const domain = useSelect( ( select ) => select( STORE_NAME ).getPropertyID() );
+	const isDomainProperty = useSelect( ( select ) => select( STORE_NAME ).isDomainProperty() );
+	const referenceSiteURL = useSelect( ( select ) => {
+		return unTrailingSlashIt( select( CORE_SITE ).getReferenceSiteURL() );
+	} );
```
felixarntz, line 49: See above.

assets/js/modules/search-console/datastore/service.js
```diff
@@ -53,6 +55,19 @@ export const selectors = {
 		}
 		return addQueryArgs( baseURI, queryArgs );
 	} ),
+
+	/**
+	 * Checks whether the property is a Search Console Domain Property.
```
felixarntz, line 60: Maybe slightly more accurate would be:
```suggestion
	 * Checks whether the Search Console property is a domain property.
```
(it _always_ is a Search Console property)

assets/js/util/sanitize.js
```diff
@@ -8,3 +8,14 @@ export const sanitizeHTML = ( unsafeHTML, domPurifyConfig = {} ) => {
...
+/**
+ * Takes a string, removes the trailing stash if any and returns it.
+ *
+ * @since n.e.x.t
+ * @private
+ *
+ * @param {string} string A string with or without trailing slash.
+ * @return {string|null} The string after removing the trailing slash.
+ */
+export const unTrailingSlashIt = ( string ) => string?.replace?.( /\/+$/, '' );
```
felixarntz, line 21: I think it would make sense to rename this to `untrailingslashit`. It's a familiar function name for WordPress developers and using a different style of naming here makes it a bit confusing. Alternatively, we could use a completely different name like `removeTrailingSlash`, but I think since we're in WordPress, using that "familiar" name makes sense.

felixarntz (changes requested):
Only some minor feedback on the existing implementation.

However, the change is missing from a couple of areas where it is required as well:
* `DashboardDetailsWidgetKeywordsTable`
* `DashboardClicksWidget`
* `DashboardImpressionsWidget`
* `DashboardPopularKeywordsWidget`
* `LegacySearchConsoleDashboardWidgetKeywordTable`
* `LegacySearchConsoleDashboardWidgetTopLevel`

See ACs:
> The exception to this are deep links for single URL details view where that `page` query parameter is already added based on the current entity URL.

While all of the above components _conditionally_ add that query parameter based on whether there is a current entity URL, for all of these there needs to be an `else` clause that conditionally adds the `page` query parameter in case of a domain property like already happening for the other components touched in this PR.

assets/js/modules/search-console/components/dashboard/GoogleSitekitSearchConsoleDashboardWidget.js
kostyalmm: @tofumatt recommended this style in favor of readability. [see comment](https://github.com/google/site-kit-wp/pull/2229#discussion_r512850815)

assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
felixarntz: @tofumatt That's only happening in the `getCurrentReferenceURL` selector, but not the `getCurrentEntityURL` selector which is being used here.

assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
```diff
@@ -42,9 +43,16 @@ const DashboardDetailsWidgetKeywordsTable = () => {
...
 		num_of_days: getCurrentDateRangeDayCount(),
 	};
+	const isDomainProperty = useSelect( ( select ) => select( STORE_NAME ).isDomainProperty() );
+	const referenceSiteURL = useSelect( ( select ) => {
+		return untrailingslashit( select( CORE_SITE ).getReferenceSiteURL() );
+	} );
 	if ( url ) {
 		footerCTALinkArgs.page = `!${ url }`;
 	}
+	if ( isDomainProperty && referenceSiteURL ) {
```
felixarntz, line 53: This needs to be an `else if` for the above, otherwise it would override the more specific URL for the current entity, which is against the ACs: If there is a specific URL to show results for, there's no point in limiting to a certain URL prefix anyway since the URL would have that by definition and be more specific.

assets/js/modules/search-console/components/dashboard/DashboardClicksWidget.js
```diff
@@ -62,6 +64,9 @@ function DashboardClicksWidget() {
 			args.url = url;
 			serviceBaseURLArgs.page = `!${ url }`;
 		}
+		if ( isDomainProperty && referenceSiteURL ) {
```
felixarntz, line 67: See above, this needs to be an `else if` for the `if ( url )` check. This also applies to all similar cases throughout this PR.

felixarntz (changes requested):
One thing is still incorrect here, but after that's fixed it should be good to go.

assets/js/modules/search-console/components/dashboard/DashboardImpressionsWidget.js
```diff
@@ -61,6 +63,8 @@ function DashboardImpressionsWidget() {
 		if ( url ) {
```
felixarntz, line 63: There's a problem here that I just noticed, which was missed during code review and QA of #1923: The above value for `num_of_days` is invalid (e.g. `28 days` instead of expected `28`), and the call to `getCurrentDateRange` is incorrect (it doesn't support a second parameter).

Unrelated to this PR, but we should fix it.

assets/js/util/sanitize.test.js
```diff
@@ -29,3 +29,21 @@ describe( 'sanitizeHTML', () => {
 		} );
 	} );
 } );
+
+describe( 'unTrailingSlashIt', () => {
```
felixarntz, line 33: Let's give this the correct name too:
```suggestion
describe( 'untrailingslashit', () => {
```

felixarntz (commented):
Pretty much LGTM, just one minor thing and one problem that however isn't caused by this PR.

felixarntz (approved):
Great!

I'll address the final two details in a follow-up commit (can't push to your fork).

Issue #2039, opened 2020-09-16
Implement base JS components for `googlesitekit-user-input` screen


felixarntz:
Following up on the new `googlesitekit-user-input` admin screen introduced in #2038, JS components for its UI should be implemented and used on the screen. The actual UI design will be handled in a follow-up issue.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* JS components for the new `googlesitekit-user-input` admin screen from #2038 should be implemented which cover the necessary functionality for that screen, and then they should be incorporated into the `UserInputApp` entry component from #2038. Let's not worry about exact appearance as part of this issue (that can happen afterwards in #2040). Still, please have a look at the [UI mocks](https://www.figma.com/file/K4MXDvLRULqRRz2zpMlh5e/SiteKit---Phase-1%3A-Onboarding?node-id=1098%3A16956) to get an idea of what it will look like later, as it should help clarify the functionality too.
* The overall app should consist of the regular Site Kit header (existing component) and a main content wrapper.
* The following UI sections should be present as main content:
    * A header saying _Customize Site Kit to match your goals_ and a short subtitle beside it: _Get metrics and suggestions that are specific to your site by telling Site Kit more about your site_
    * Components for the five user input questions. These should use common components wherever possible. Please refer to the UI mocks for more information on copy for each. The questions are (setting ID: _question_ (type, scope)):
        1. `role`: _Which best describes your team/role in relation to this site?_ (single choice, per user)
        2. `postFrequency`: _How often do you create new post for this site?_ (single choice, per user)
        3. `goals`: _What are the goals of this site?_ (multiple choice, entire site)
        4. `helpNeeded`: _What do you need help most with for this site?_ (multiple choice, entire site)
        5. `searchTerms`: _To help us identify opportunities for yoursite, enter the top three search terms that you’d like to show up for_ (list of free-text keywords separated by comma, entire site)
    * A component for the overview of all questions and the provided answers, with the ability to go back and edit, or submit ([see mock](https://www.figma.com/file/K4MXDvLRULqRRz2zpMlh5e/SiteKit---Phase-1%3A-Onboarding?node-id=1098%3A17156)).
* High-level functionality should be as follows:
    * For retrieving and setting current values for the questions, the selectors and actions from #2037 should be used. Every setting is stored as an array of string values here, i.e. for the single choice questions, just use index 0. For the multiple choice questions, check if the option is in the array of values.
[...]

Issue #2040, opened 2020-09-16
Implement UI design for new `googlesitekit-user-input` admin screen


felixarntz:
After #2039, this issue is about actually implementing the UI design for the `googlesitekit-user-input` admin screen (also see #2038) previously created using CSS (and minor alterations of the new components as necessary).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The UI for the `googlesitekit-user-input` screen should be implemented per the [designs](https://www.figma.com/file/K4MXDvLRULqRRz2zpMlh5e/SiteKit---Phase-1%3A-Onboarding?node-id=1098%3A16956), using the React components implemented as part of #2039.
    * Screen header
    * Steps 1-5
        * The respective next step should show in lower opacity below, as far as the screen allows for it. In other words, there should be no extra scrolling beyond the current step, the lower-opacity next step should only be displayed as some kind of "teaser". On viewports where the step itself fills more than the viewport size, this next step "teaser" wouldn't be displayed.
    * Final overview/confirmation (after the 5th step)
* The changes should mostly be CSS, plus some minor DOM changes in the HTML generated by the components from #2039. It should be possible to rely on the existing components overall though.

## Implementation Brief

- Create a new folder named `user-input` in `/assets/sass/components/`.
- Within the newly created folder, create a file named: `_googlesitekit-user-input.scss`.
- Above file should contain all the CSS to style the user input screens as per the designs. Existing styles should be re-used as much as possible.

## QA Brief

The components should be styled as per the designs across all breakpoints.

## Changelog entry

* Implement UI design for `googlesitekit-user-input` screen.

felixarntz:
IB ✅

cole10up:
### QA ❌

### **Questions**
Development using Chrome on the left, Figma on the right. Note - many of the items found or in question are Chrome specific.

Browsers Tested:

- Chome
- Firefox
- Edge
- Safari
- iOS
- Android

<details>
 <summary>Question 1:</summary>

![image](https://user-images.githubusercontent.com/57115907/99481533-2eb82380-290f-11eb-9d14-b1272452908a.png)

Items to note:
- '1 out of 5' doesn't match the styling in the Figma design both color and size are wrong **using Chrome browser**
</details> 

<details>
 <summary>Question 2:</summary>

![image](https://user-images.githubusercontent.com/57115907/99482254-4f34ad80-2910-11eb-9bbe-87031a07a56a.png)

[...]

PR #2232, opened 2020-10-22
Feature/2039 user input screen base components

eugene-manuilov:
## Summary

Addresses issue #2039

## Relevant technical choices

* Added helper components for the grid from the material UI library.

## Checklist


assets/js/components/user-input/UserInputQuestionnaire.js
```diff
@@ -0,0 +1,154 @@
...
+	}, [ activeSlugIndex ] );
+
+	const back = goTo.bind( null, 1 );
+
+	return (
+		<Fragment>
+			<ProgressBar
+				height={ 0 }
+				indeterminate={ false }
+				progress={ ( activeSlugIndex + 1 ) / questions.length }
```
felixarntz, line 62: Currently, this is not fully correct as it also takes the `preview` step into account, but that shouldn't affect the progress bar progress. Based on my previous comment, if this relies on really only the `questions` (not `steps`, see previous comment), then it would work as expected.

assets/js/components/user-input/UserInputSelectOptions.js
```diff
@@ -0,0 +1,141 @@
...
+import Data from 'googlesitekit-data';
+import { STORE_NAME as CORE_USER } from '../../googlesitekit/datastore/user/constants';
+import Radio from '../radio';
+import Checkbox from '../checkbox';
+import { Cell, Input, TextField } from '../../material-components';
+const { useSelect, useDispatch } = Data;
+
+export default function UserInputSelectOptions( { slug, options, max } ) {
+	const values = useSelect( ( select ) => select( CORE_USER ).getUserInputSetting( slug ) || [] );
+	const [ other, setOther ] = useState( values.filter( ( value ) => ! options[ value ] )[ 0 ] || '' );
```
felixarntz, line 42: Nice! It's great that the `other` value is stored as the actual response itself - so by comparing to the available predefined answers we know whether it's an "other" one.

assets/js/components/user-input/util/constants.js
```diff
@@ -0,0 +1,56 @@
...
+	retaining_visitors: __( 'Retaining visitors, turning them into loyal readers or customers', 'google-site-kit' ),
+	improving_performance: __( 'Improving speed and performance', 'google-site-kit' ),
+	finding_new_topics: __( 'Finding new topics to write about that connect with my audience', 'google-site-kit' ),
+	growing_audience: __( 'Growing my audience', 'google-site-kit' ),
+	expanding_business: __( 'Expanding my business into new cities, states or markets', 'google-site-kit' ),
+	generating_revenue: __( 'Generating more revenue', 'google-site-kit' ),
+	help_better_rank: __( 'Help my content rank in a better position in Google search results', 'google-site-kit' ),
+	understanding_content_performance: __( 'Understanding which content is performing best', 'google-site-kit' ),
+	encourage_to_post: __( 'Encouragement to post more frequently', 'google-site-kit' ),
+};
```
felixarntz, line 56: I don't think these can be constants because this way the translation strings are attempted to be translated immediately which can cause timing problems when the translation data is only loaded a bit later. For registering widgets for example, we therefore wrapped the calls into a `domReady` callback. I think for this scenario here, we could instead use functions instead such as `getUserInputAnswers( questionSlug )` so that they are only translated once needed.

assets/js/googlesitekit/datastore/user/user-input-settings.js
```diff
@@ -59,9 +54,13 @@ const fetchSaveUserInputSettingsStore = createFetchStore( {
 // Actions
 const SET_USER_INPUT_SETTINGS = 'SET_USER_INPUT_SETTINGS';
 const SET_USER_INPUT_SETTING = 'SET_USER_INPUT_SETTING';
+const START_SAVING_USER_SETTINGS = 'START_SAVING_USER_SETTINGS';
+const FINISH_SAVING_USER_SETTINGS = 'FINISH_SAVING_USER_SETTINGS';
```
felixarntz, line 58: I don't think we need these and all the new logic around them. The fetch store automatically creates `START_FETCH_SAVE_USER_INPUT_SETTINGS` and `FINISH_FETCH_SAVE_USER_INPUT_SETTINGS` to check whether settings are being saved.

With your original approach where the display of this screen conditionally includes the "Congrats" notification, you probably needed the changes you added here, but since this isn't necessary I don't think we need those separate indicators.

assets/js/googlesitekit/datastore/user/user-input-settings.js
```diff
@@ -172,12 +201,63 @@ const baseSelectors = {
...
+	 * Determines whether or not user input settings have been saved.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {boolean} TRUE if saving has finished, otherwise FALSE.
+	 */
+	hasFinishedSavingInputSettings( state ) {
+		return !! state.hasFinishedSavingInputSettings;
+	},
```
felixarntz, line 260: See above, I think we don't need these selectors.

assets/js/material-components/layout/Cell.js
```diff
@@ -0,0 +1,57 @@
...
+
+export default function Cell( { className, sm, md, lg, children } ) {
+	return (
+		<div className={ classnames(
+			className,
+			'mdc-layout-grid__cell',
+			{
+				[ `mdc-layout-grid__cell--span-${ lg }-desktop` ]: lg,
+				[ `mdc-layout-grid__cell--span-${ md }-tablet` ]: md,
+				[ `mdc-layout-grid__cell--span-${ sm }-phone` ]: sm,
```
felixarntz, line 33: I think we should add another prop to this component, there is another size that can be passed which is the "default" column size to use, not specific to a certain device width: It will result in a CSS class like `mdc-layout-grid__cell--span-${ size }` which right now isn't covered by this component.

I suggest to also clarify the naming of the existing props. How about:
* `size`
* `smSize`
* `mdSize`
* `lgSize`

assets/js/material-components/layout/index.js
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export { default as Cell } from './Cell';
+export { default as Row } from './Row';
+export { default as Grid } from './Grid';
```
felixarntz, line 21: Nice!

assets/sass/components/user-input/googlekitekit-user-input-module-header.scss
```diff
@@ -0,0 +1,267 @@
...
+			padding: 35px;
+
+			@include shadow;
+
+			& h1 {
+				margin-bottom: 0;
+			}
+		}
+	}
+}
```
felixarntz, line 267: Great work on already adding all the CSS here - although most of that is technically part of #2040. But that clearly reduces the necessary work there, we'll probably only need to make some minor UI refinements there which I won't address in here.

includes/Core/Admin/Screens.php
```diff
@@ -398,7 +398,7 @@ private function get_screens() {
 			array(
 				'title'            => __( 'User Input', 'google-site-kit' ),
 				'capability'       => Permissions::MANAGE_OPTIONS,
-				'parent_slug'      => null,
+				'parent_slug'      => defined( 'SHOW_USER_INPUT_SETTINGS_PAGE' ) ? Screen::MENU_SLUG : null,
```
felixarntz, line 401: This should be removed, it should be `null`. We can always manually enter the URL for testing.

includes/Core/REST_API/REST_Routes.php
```diff
@@ -165,20 +166,65 @@ function( $pre, $args, $url ) {
...
+									'scope'      => $values['scope'],
+									'answeredBy' => $answered_by,
+								);
+							} else {
+								$user_settings[ $key ] = array(
+									'values'     => ! empty( $body[ $key ] ) ? $body[ $key ] : array(),
+									'scope'      => $values['scope'],
+									'answeredBy' => $user_id,
+								);
+							}
```
felixarntz, line 207: I think it's a great idea to introduce the `answeredBy` concept here which will result in a more clear user experience rather than relying on just the site's Site Kit `owner`. However, the Site Kit Service API currently has no plan to support this, so patching over it in this temporary HTTP request interception logic isn't really feasible unless we decide on an approach that will also work in the same way with the real API endpoint once it exists.

Let's leave this as is for now, relying on the user ID, and we will discuss whether that can be implemented in the API layer as well. It's certainly a good proof of concept to have here.

includes/Core/REST_API/REST_Routes.php
```diff
@@ -165,20 +166,65 @@ function( $pre, $args, $url ) {
...
+						update_option( 'googlesitekit_temp_userinput_' . $user_id, $user_settings, 'no' );
 					}
 				}
 
-				$user_input = get_option( 'googlesitekit_temp_userinput', $defaults );
+				$user_input    = array();
+				$user_settings = get_option( 'googlesitekit_temp_userinput_' . $user_id, array() );
+				$site_settings = get_option( 'googlesitekit_temp_userinput_sitewide', array() );
+
+				foreach ( $defaults as $key => $values ) {
```
felixarntz, line 219: Since you added `answeredBy` to the expected API response, we should probably add this key to each value in `$defaults` as well so that it's consistent. It can just be `0` there.

assets/js/components/user-input/UserInputPreview.js
```diff
@@ -0,0 +1,128 @@
...
+
+export default function UserInputPreview( { goTo } ) {
+	const settings = useSelect( ( select ) => select( CORE_USER ).getUserInputSettings() );
+	const {
+		hasStartedSavingInputSettings,
+		hasFinishedSavingInputSettings,
+	} = useSelect( ( select ) => ( {
+		hasStartedSavingInputSettings: select( CORE_USER ).hasStartedSavingInputSettings(),
+		hasFinishedSavingInputSettings: select( CORE_USER ).hasFinishedSavingInputSettings(),
+	} ) );
```
felixarntz, line 56: Instead of this, let's use the automatically available `isFetchingSaveUserInputSettings` selector.

assets/js/components/user-input/UserInputPreview.js
```diff
@@ -0,0 +1,128 @@
...
+
+	const { saveUserInputSettings } = useDispatch( CORE_USER );
+	const submitChanges = useCallback( () => {
+		saveUserInputSettings( settings );
```
felixarntz, line 60: Per the ACs, this should redirect to the Site Kit dashboard on success (we can inspect the action's return value for that). If successful, let's set an internal state flag `isNavigating` to `true`, to keep the progress bar below active, even though technically the save process has already completed. We use this approach elsewhere too when we redirect to another URL, since the progress bar should just keep going until the user is on the other URL.

If an error occurs, we should then set `isNavigating` to `false` again.

assets/js/components/user-input/UserInputPreview.js
```diff
@@ -0,0 +1,128 @@
...
+	}, [ settings ] );
+
+	return (
+		<div className="googlesitekit-user-input__preview">
+			<Row>
+				<Cell>
+					{ hasStartedSavingInputSettings && ! hasFinishedSavingInputSettings && (
```
felixarntz, line 67: See above, this should be `isFetchingSaveUserInputSettings || isNavigating`.

assets/js/components/user-input/UserInputPreview.js
```diff
@@ -0,0 +1,128 @@
...
+						<ProgressBar />
+					) }
+					{ ! hasStartedSavingInputSettings && (
```
felixarntz, line 70: See above, this should be the opposite of the new above clause.

assets/js/components/user-input/UserInputPreview.js
```diff
@@ -0,0 +1,128 @@
...
+								</Cell>
+							</Row>
+
+							<div className="googlesitekit-user-input__buttons">
+								<UserInputQuestionNotice />
+								<div>
+									<Button text onClick={ goTo.bind( null, 1 ) }>{ __( 'Back', 'google-site-kit' ) }</Button>
+									<Button onClick={ submitChanges }>{ __( 'Submit', 'google-site-kit' ) }</Button>
+								</div>
+							</div>
```
felixarntz, line 117: I think it would be great to add some error handling to this component, for the case that the API request to save unexpectedly fails. We can use the `core/user` store's `getErrorForAction( 'saveUserInputSettings', [] )` selector, and if there is one, display it via an `ErrorNotice`. Maybe close to the buttons would work well.

felixarntz (changes requested):
Awesome work here! 🎉  This looks super solid and almost already covers even #2040. I have a couple of important feedback points, but this is really close to what we want.

assets/js/components/user-input/UserInputQuestionWrapper.js
```diff
@@ -0,0 +1,121 @@
...
+	const values = useSelect( ( select ) => select( CORE_USER ).getUserInputSetting( slug ) || [] );
+	const scope = useSelect( ( select ) => select( CORE_USER ).getUserInputSettingScope( slug ) );
+	const author = useSelect( ( select ) => select( CORE_USER ).getUserInputSettingAuthor( slug ) );
+
+	return (
+		<div className={ classnames(
+			'googlesitekit-user-input__question',
+			{
+				'googlesitekit-user-input__active-question': isActive,
+				'googlesitekit-user-input__next-question': ! isActive,
```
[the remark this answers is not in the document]
eugene-manuilov: Agree, updated!

assets/js/components/user-input/UserInputQuestionnaire.js
```diff
@@ -0,0 +1,154 @@
...
+	USER_INPUT_ANSWERS_GOALS,
+	USER_INPUT_ANSWERS_HELP_NEEDED,
+	USER_INPUT_ANSWERS_POST_FREQUENCY,
+	USER_INPUT_ANSWERS_ROLE,
+} from './util/constants';
+
+export default function UserInputQuestionnaire() {
+	const [ activeSlug, setActiveSlug ] = useState( 'role' );
+
+	const questions = [ 'role', 'postFrequency', 'goals', 'helpNeeded', 'searchTerms', 'preview' ];
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/components/user-input/UserInputQuestionnaire.js
eugene-manuilov: Updated.

assets/js/components/user-input/UserInputQuestionWrapper.js
```diff
@@ -0,0 +1,121 @@
...
+						<div className="googlesitekit-user-input__buttons">
+							{ back && (
+								<Button text onClick={ back }>
+									{ __( 'Back', 'google-site-kit' ) }
+								</Button>
+							) }
+							{ next && (
+								<Button
+									onClick={ next }
+									disabled={ values.filter( ( value ) => value.trim().length > 0 ).length !== max }
```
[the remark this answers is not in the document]
eugene-manuilov: Ok, updated to require at least one answer.

assets/js/components/user-input/UserInputQuestionnaire.js
```diff
@@ -0,0 +1,154 @@
...
+	const activeSlugIndex = questions.indexOf( activeSlug );
+
+	const next = useCallback( () => {
+		setActiveSlug( questions[ activeSlugIndex + 1 ] );
+	}, [ activeSlugIndex ] );
+
+	const goTo = useCallback( ( steps = 1 ) => {
+		setActiveSlug( questions[ activeSlugIndex - steps ] );
+		global.scrollTo( 0, 0 );
+	}, [ activeSlugIndex ] );
```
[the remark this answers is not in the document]
eugene-manuilov: Works for me. Updated.

assets/js/components/user-input/util/constants.js
eugene-manuilov: Updated.

assets/js/material-components/layout/Cell.js
eugene-manuilov: Looks good to me. Updated.

assets/js/components/user-input/UserInputApp.js
```diff
@@ -19,12 +19,81 @@
...
+							<Row>
+								<Cell>
+									<ProgressBar />
+								</Cell>
+							</Row>
+						</Grid>
+					) }
+					{ hasFinishedSavingInputSettings && (
+						<UserInputCongrats />
+					) }
```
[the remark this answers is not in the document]
eugene-manuilov: Ok, removed the congrats notification and added automatic redirect.

assets/js/components/user-input/UserInputApp.js
```diff
@@ -19,12 +19,81 @@
...
+					{ hasFinishedGettingInputSettings && ! hasFinishedSavingInputSettings && (
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/components/user-input/UserInputPreview.js
eugene-manuilov: Updated.

assets/js/googlesitekit/datastore/user/user-input-settings.js
eugene-manuilov: Yep, makes sense, removed.

assets/js/components/user-input/UserInputPreview.js
eugene-manuilov: `isFetchingSaveUserInputSettings` isn't enough for the preview screen. It's much easier to use hasStarted* and hasFinished* flags.

assets/js/components/user-input/UserInputPreview.js
eugene-manuilov: Ok, updated. But decided to stick with `hasStartedSavingInputSettings` and `hasFinishedSavingInputSettings` because it helps to better indicate the status of the process than `isNavigating`

assets/js/components/user-input/UserInputPreview.js
eugene-manuilov: Decided to stick with `hasStartedSavingInputSettings` and `hasFinishedSavingInputSettings` because it better helps to identify the status of the process.

assets/js/components/user-input/UserInputPreview.js
eugene-manuilov: Updated. See above.

assets/js/googlesitekit/datastore/user/user-input-settings.js
eugene-manuilov: It is still needed because it helps to identify when we need to redirect a user to the dashboard.

includes/Core/Admin/Screens.php
eugene-manuilov: Updated.

eugene-manuilov:
@felixarntz this is ready for another review. I haven't removed hasStarted*/hasFinished* properties because it's more convenient to use it IMO.

assets/js/components/user-input/UserInputQuestionWrapper.js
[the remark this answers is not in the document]
felixarntz: Confirmed by @marrrmarrr that this should mean **at least**. The wording is a bit confusing, but wording will be revisited in a follow-up issue.

assets/js/googlesitekit/datastore/user/user-input-settings.js
felixarntz: I think this should be handled based on the submit user action, see my above comment. So these two selectors can be removed.

assets/js/material-components/layout/Cell.js
```diff
@@ -0,0 +1,69 @@
...
+		PropTypes.arrayOf( PropTypes.node ),
+	] ).isRequired,
+};
+
+Cell.defaultProps = {
+	className: '',
+	size: 0,
+	smSize: 4,
+	mdSize: 8,
+	lgSize: 12,
```
felixarntz, line 68: I think it makes sense for all these defaults to be `0` (i.e. they won't be used). The default behavior shouldn't add all these classes, it should be more like an "opt-in" approach.

felixarntz (changes requested):
This is almost good to go, however some follow-up feedback from the previous review which I think is worth modifying.

includes/Core/REST_API/REST_Routes.php
eugene-manuilov: Added.

assets/js/material-components/layout/Cell.js
eugene-manuilov: Updated.

eugene-manuilov:
@felixarntz IMO moving that redirect into a datastore action is not a great idea because it is a side effect and it shouldn't be there. It will make datastore actions harder to test and will affect the behavior of user input settings on the Settings page which is not what we need, right?

felixarntz:
@eugene-manuilov Maybe I worded it the wrong way, I wasn't meaning to add the redirect to the datastore action. My suggestion is to add the redirect to the `useCallback` callback for `submitChanges` in the component, i.e. pass an `async` function to it that:
1. Sets a component-internal `isNavigating` state to `true`.
2. Awaits the datastore action.
3. Inspects the results to see if there was an error or if it completed successfully.
4. On error, set `isNavigating` to `false` again to stop the progress bar from showing up "infinitely". On success, do the redirect (keep `isNavigating` as `true` because it should visually keep loading until the user has actually been redirected).

assets/js/components/user-input/UserInputPreview.js
```diff
@@ -0,0 +1,142 @@
...
+export default function UserInputPreview( { back, goTo } ) {
+	const dashboardURL = useSelect( ( select ) => select( CORE_SITE ).getAdminURL( 'googlesitekit-dashboard' ) );
+	const settings = useSelect( ( select ) => select( CORE_USER ).getUserInputSettings() );
+	const {
+		hasStartedSavingInputSettings,
+		hasFinishedSavingInputSettings,
+		error,
+	} = useSelect( ( select ) => ( {
+		hasStartedSavingInputSettings: select( CORE_USER ).hasStartedSavingInputSettings(),
+		hasFinishedSavingInputSettings: select( CORE_USER ).hasFinishedSavingInputSettings(),
```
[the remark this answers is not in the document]
eugene-manuilov: I see what you mean now. Your approach is better. I have removed it from the PR.

assets/js/components/user-input/UserInputPreview.js
```diff
@@ -0,0 +1,142 @@
...
+	} ) );
+
+	const { saveUserInputSettings } = useDispatch( CORE_USER );
+	const submitChanges = useCallback( saveUserInputSettings, [] );
+
+	useEffect( () => {
+		if ( hasFinishedSavingInputSettings && ! error ) {
+			global.location.assign( dashboardURL );
+		}
+	}, [ dashboardURL, hasFinishedSavingInputSettings, error ] );
```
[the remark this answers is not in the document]
eugene-manuilov: I have reworked it to use your approach. It's better than those selectors. Sorry for the misunderstanding.

assets/js/components/user-input/UserInputPreview.js
```diff
@@ -0,0 +1,142 @@
...
+		USER_INPUT_ANSWERS_HELP_NEEDED,
+		USER_INPUT_ANSWERS_POST_FREQUENCY,
+		USER_INPUT_ANSWERS_ROLE,
+	} = getUserInputAnwsers();
+
+	return (
+		<div className="googlesitekit-user-input__preview">
+			<Row>
+				<Cell>
+					{ ( hasStartedSavingInputSettings || hasFinishedSavingInputSettings ) && ! error && (
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/components/user-input/UserInputPreview.js
```diff
@@ -0,0 +1,142 @@
...
+						<ProgressBar />
+					) }
+					{ ( ( ! hasStartedSavingInputSettings && ! hasFinishedSavingInputSettings ) || error ) && (
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/googlesitekit/datastore/user/user-input-settings.js
```diff
@@ -133,10 +136,27 @@ export const baseReducer = ( state, { type, payload } ) => {
...
+				hasFinishedSavingInputSettings: false,
+			};
+		}
+		case 'FINISH_FETCH_SAVE_USER_INPUT_SETTINGS': {
+			return {
+				...state,
+				hasStartedSavingInputSettings: false,
+				hasFinishedSavingInputSettings: true,
+			};
+		}
```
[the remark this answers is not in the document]
eugene-manuilov: Removed.

assets/js/components/user-input/UserInputPreview.js
eugene-manuilov: Updated.

assets/js/components/user-input/UserInputPreview.js
eugene-manuilov: Updated to use `isFetchingSaveUserInputSettings`.

assets/js/googlesitekit/datastore/user/user-input-settings.js
eugene-manuilov: Yep, you are right. I got it now. removed these selectors.

eugene-manuilov:
I got it now. :doh: I have updated it to work as you described. Could you please take another look at it?

assets/js/components/user-input/UserInputPreview.js
```diff
@@ -0,0 +1,141 @@
...
+	} = getUserInputAnwsers();
+
+	return (
+		<div className="googlesitekit-user-input__preview">
+			<Row>
+				<Cell lgSize={ 12 } mdSize={ 8 } smSize={ 4 }>
+					{ ( isSavingSettings || isNavigating ) && ! error && (
+						<ProgressBar />
+					) }
+					{ ( ( ! isSavingSettings && ! isNavigating ) || error ) && (
```
felixarntz, line 80: I don't think the `error` check is needed in either of these. If there is an error, per the behavior both `isSavingSettings` and `isNavigating` will always be `false`. So I think we can omit that.

felixarntz (commented):
This is good to go, just one minor comment but nothing blocking. Let me know what you think.

assets/js/components/user-input/UserInputPreview.js
eugene-manuilov: Yes, you are right. Removed it.

felixarntz (approved):
Awesome work here!

Issue #2182, opened 2020-10-15
Add actions and selectors for centrally checking module settings submission state


felixarntz:
The `core/modules` store should provide a central management layer for handling submission of module settings (which is generally handled in each module's own store, see #2136). Note while #2136 has the goal of every module store providing these actions and selectors, that issue is _not_ a dependency. The code here should work regardless of whether the module store has the action/selector or not.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `core/modules` store should receive a registry selector `canSubmitChanges( slug )` that checks for store `modules/${ slug }` for whether it exists and has a `canSubmitChanges` selector. If yes, it should return that one's result. Otherwise, it should return `false`.
* The `core/modules` store should receive a registry selector `isDoingSubmitChanges( slug )` that checks for store `modules/${ slug }` for whether it exists and has a `isDoingSubmitChanges` selector. If yes, it should return that one's result. Otherwise, it should return `false`.
* The `core/modules` store should receive an action `submitChanges( slug )` that via a registry control checks for whether store `modules/${ slug }` exists and whether it has a `submitChanges` action. If yes, it should dispatch that action and return the result. Otherwise it should return an object with a single `error` property, similar to how the `submitChanges` action (see #2136) would if an error happened.

## Implementation Brief

* Add a new file to the `core/modules` datastore: `assets/js/googlesitekit/modules/datastore/changes.js`. The actions/selectors/etc. described below should be added to this file.
* Add a new `canSubmitChanges( slug )` registry selector that returns <code>!! select( \`modules/${ slug }\` ).canSubmitChanges && select( \`modules/${ slug }\` ).canSubmitChanges();</code>.
* Add a new `isDoingSubmitChanges( slug )` registry selector that returns <code>!! select( \`modules/${ slug }\` ).isDoingSubmitChanges && select( \`modules/${ slug }\` ).isDoingSubmitChanges();</code>.
* Add a new `submitChanges( slug )` action with an associated registry control that:
  * Checks for the existence of <code>registry.dispatch( \`modules/${ slug }\` ).submitChanges</code>
  * If the action exists: dispatches <code>registry.dispatch( \`modules/${ slug }\` ).submitChanges()</code> and returns its result.
* If the action doesn't exist, return an object with the shape: <code>{ error: \`'modules/${ slug }' does not have a submitChanges() action.\` }</code>

### Test Coverage

[...]

PR #2265, opened 2020-10-24
Enhancement/2182 centrally check module settings

ivankruchkoff:
## Summary

Addresses issue #2182

## Relevant technical choices
Named file as settings.js rather than changes.js per https://github.com/google/site-kit-wp/issues/2182#issuecomment-710115194.

## Checklist


assets/js/googlesitekit/modules/datastore/settings.js
```diff
@@ -0,0 +1,81 @@
...
+	/**
+	 * Submits all changes for a module.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {string} slug Slug for module store.
+	 * @return {Object} Module's submitChanges response object if it exists, otherwise object with `error` property if it doesn't.
+	 */
+	*submitChanges( slug ) {
+		invariant( slug, 'slug is required.' );
```
aaemnnosttv, line 40: I'm not sure if this is still an issue, due to a recent change in the way the tests run, but using `invariant` in a generator like this wouldn't cause the test to fail due to the asynchronous nature of how these run in the datastore.

The best way to check here would be to add a test to dispatch `submitChanges` without a slug and without `await` and expect it to throw the invariant message. E.g. https://github.com/google/site-kit-wp/blob/c042f538e0e6e97de9d9eac7453d2f01b435f52e/assets/js/googlesitekit/data/create-settings-store.test.js#L77-L80

If that doesn't work, then we simply need to change `submitChanges` here to not be a generator, but return one. E.g. https://github.com/google/site-kit-wp/blob/c042f538e0e6e97de9d9eac7453d2f01b435f52e/assets/js/googlesitekit/data/create-fetch-store.js#L177-L186
I thought we had more examples of this but I couldn't find one. It may be that they were refactored and this technique was no longer necessary.

assets/js/googlesitekit/modules/datastore/settings.js
```diff
@@ -0,0 +1,81 @@
...
+		const registry = yield Data.commonActions.getRegistry();
+		if ( !! registry.dispatch( `modules/${ slug }` ) && !! registry.dispatch( `modules/${ slug }` ).submitChanges ) {
```
aaemnnosttv, line 42: This is a good use case for conditional chaining
```suggestion
		if ( !! registry.dispatch( `modules/${ slug }` )?.submitChanges ) {
```

assets/js/googlesitekit/modules/datastore/settings.js
```diff
@@ -0,0 +1,81 @@
...
+	 * Checks whether changes are currently being submitted for a module.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {string} slug Slug for module store.
+	 * @return {Object} Module's submitChanges response object if it exists, otherwise object with `error` property if it doesn't.
+	 */
+	isDoingSubmitChanges: createRegistrySelector( ( select ) => ( state, slug ) => {
+		invariant( slug, 'slug is required.' );
+		return !! select( `modules/${ slug }` ) && !! select( `modules/${ slug }` ).isDoingSubmitChanges && select( `modules/${ slug }` ).isDoingSubmitChanges();
```
aaemnnosttv, line 60: Same here 😄 
```suggestion
		return !! select( `modules/${ slug }` )?.isDoingSubmitChanges?.();
```

assets/js/googlesitekit/modules/datastore/settings.js
```diff
@@ -0,0 +1,81 @@
...
+	 * Checks whether we can submit changes for a module.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {string} slug Slug for module store.
+	 * @return {boolean} Whether the module supports submitting changes.
+	 */
+	canSubmitChanges: createRegistrySelector( ( select ) => ( state, slug ) => {
+		invariant( slug, 'slug is required.' );
+		return !! select( `modules/${ slug }` ) && !! select( `modules/${ slug }` ).canSubmitChanges && select( `modules/${ slug }` ).canSubmitChanges();
```
aaemnnosttv, line 73:
```suggestion
		return !! select( `modules/${ slug }` )?.canSubmitChanges?.();
```

assets/js/googlesitekit/modules/datastore/settings.test.js
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Internal dependencies
+ */
+import Data from 'googlesitekit-data';
+import Modules from 'googlesitekit-modules';
+import { STORE_NAME } from './constants';
+import { createTestRegistry } from '../../../../../tests/js/utils';
+
+describe( 'core/modules store changes', () => {
```
aaemnnosttv, line 27: The describe here should mirror the structure for consistency

```suggestion
describe( 'core/modules settings', () => {
```

assets/js/googlesitekit/modules/datastore/settings.test.js
```diff
@@ -0,0 +1,98 @@
...
+	let registry;
+	const slug = 'test-module';
+	const nonExistentModuleSlug = 'not-module';
+	const moduleStoreName = `modules/${ slug }`;
+	let moduleCanSubmitChanges = false;
+	let submittingChanges = false;
+
+	beforeAll( () => {
+	} );
```
aaemnnosttv, line 36: Please remove any unused setup/teardown.

assets/js/googlesitekit/modules/datastore/settings.test.js
```diff
@@ -0,0 +1,98 @@
...
+
+	afterAll( () => {
+	} );
+
+	afterEach( () => {
+	} );
+
+	describe( 'selectors', () => {
+		describe( 'submitting changes', () => {
```
aaemnnosttv, line 73: The describe blocks at this level should mirror the selector name, and then there would be an adjacent `describe` for each selector.
```suggestion
		describe( 'isDoingSubmitChanges', () => {
```

assets/js/googlesitekit/modules/datastore/settings.test.js
```diff
@@ -0,0 +1,98 @@
...
+
+			it( 'can submit changes', () => {
+				expect( registry.select( STORE_NAME ).canSubmitChanges( slug ) ).toBe( false );
+				moduleCanSubmitChanges = true;
+				expect( registry.select( STORE_NAME ).canSubmitChanges( slug ) ).toBe( true );
+
+				expect( registry.select( STORE_NAME ).canSubmitChanges( nonExistentModuleSlug ) ).toBe( false );
+			} );
+
+			it( 'does submit changes', async () => {
```
aaemnnosttv, line 90: `submitChanges` is an action rather than a selector, so it belongs in a separate `actions` describe block which is usually defined before the selectors. See any of the other datastore tests if you're unsure about the structure.

assets/js/googlesitekit/modules/datastore/settings.test.js
```diff
@@ -0,0 +1,98 @@
...
+				const expectedError = { error: `'modules/${ nonExistentModuleSlug }' does not have a submitChanges() action.` };
+				expect( await registry.dispatch( STORE_NAME ).submitChanges( nonExistentModuleSlug ) ).toEqual( expectedError );
+
+				expect( await registry.dispatch( STORE_NAME ).submitChanges( slug ) ).toBeTruthy();
```
aaemnnosttv, line 94: I'm not sure what the return value is here since this action only dispatches a dummy action.

To control the actual return value, this function should be a generator which yields an action that has an associated control. The return from the control function is what the return value would be here (just like with the real action) so let's use that instead.

assets/js/googlesitekit/modules/datastore/settings.test.js
```diff
@@ -0,0 +1,98 @@
...
+		describe( 'submitting changes', () => {
+			it( 'is submitting changes', async () => {
+				expect( registry.select( STORE_NAME ).isDoingSubmitChanges( nonExistentModuleSlug ) ).toBe( false );
+
+				expect( registry.select( STORE_NAME ).isDoingSubmitChanges( slug ) ).toBe( false );
+				submittingChanges = true;
+				expect( registry.select( STORE_NAME ).isDoingSubmitChanges( slug ) ).toBe( true );
+			} );
+
+			it( 'can submit changes', () => {
```
aaemnnosttv, line 82: Regarding the test names, I think we can word it a little differently to better describe what the actions and selectors do. This applies to all of the cases here, but something like "it proxies the selector call to the module with the given slug" would be a bit more descriptive and useful for someone who didn't know what it does. Even if the name for this test might be the same for each selector/action, it would still be scoped per-selector. Because of the structure of the tests, each test is already in the context of a specific selector/action so it's best to use the name to explain more about what's happening. Again, there are lots of examples to reference from other stores, but here's a bit more of a [comprehensive example](https://github.com/google/site-kit-wp/blob/6beb2381961962da4cfeedde0a2d027e95128662/assets/js/modules/tagmanager/datastore/versions.test.js) from one of the partials that I worked on recently.

assets/js/googlesitekit/modules/datastore/settings.test.js
```diff
@@ -0,0 +1,98 @@
...
+			return submittingChanges;
+		};
+
+		registry = createTestRegistry();
+		registry.registerStore( moduleStoreName, Data.combineStores(
+			storeDefinition,
+			{
+				actions: {
+					submitChanges,
+				},
```
aaemnnosttv, line 57: Per my previous comment:
```suggestion
				},
				controls: {
					DUMMY_ACTION: () => submitChangesResult,
				},
```

aaemnnosttv (changes requested):
Thanks @ivankruchkoff – this looks like a great start, just a few things to polish up, mostly for consistency with other data stores.

Let me know if you have any questions 👍

assets/js/googlesitekit/modules/datastore/settings.js
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Internal dependencies
+ */
+import Data from 'googlesitekit-data';
+const { createRegistrySelector } = Data;
+
+/**
+ * External dependencies
+ */
+import invariant from 'invariant';
```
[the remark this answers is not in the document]
ivankruchkoff: Updated and addressed.

assets/js/googlesitekit/modules/datastore/settings.js
ivankruchkoff: Approach 1 won't work.
With approach 2, it feels somewhat less correct to change the implementation so as to accommodate for the testing scenario, as it becomes less readable and thus less maintable, what if we instead:

`*submitChanges( slug ) {
		slug = slug ? slug : '';`

Which will correctly error on submitChanges `await registry.dispatch( STORE_NAME ).submitChanges()` with the following `{ error: "'modules/' does not have a submitChanges() action." };`

assets/js/googlesitekit/modules/datastore/settings.js
ivankruchkoff: Addressed.

assets/js/googlesitekit/modules/datastore/settings.js
ivankruchkoff: Addressed.

assets/js/googlesitekit/modules/datastore/settings.js
ivankruchkoff: Addressed.

assets/js/googlesitekit/modules/datastore/settings.js
aaemnnosttv: > With approach 2, it feels somewhat less correct to change the implementation so as to accommodate for the testing scenario, as it becomes less readable and thus less maintable, what if we instead

It's not entirely for testing, and we don't actually need to split the two up as much as is done in the create fetch store example. This will work just as well:
```js
submitChanges( slug ) {

	return ( function* () {
		// yield ...
		// yield ...
	} )();
```

To your point however, in this case I think it would be fine to omit the invariant call and let it return the error as you suggested. I don't think it's necessary to have the fallback value though.

ivankruchkoff: Updated, and removed the fallback.

assets/js/googlesitekit/modules/datastore/settings.test.js
ivankruchkoff: Addressed.

assets/js/googlesitekit/modules/datastore/settings.test.js
ivankruchkoff: Addressed.

assets/js/googlesitekit/modules/datastore/settings.test.js
ivankruchkoff: Addressed.

assets/js/googlesitekit/modules/datastore/settings.test.js
ivankruchkoff: Addressed.

assets/js/googlesitekit/modules/datastore/settings.test.js
ivankruchkoff: Addressed.

assets/js/googlesitekit/modules/datastore/settings.test.js
ivankruchkoff: Addressed.

assets/js/googlesitekit/modules/datastore/settings.test.js
```diff
@@ -0,0 +1,104 @@
...
+				controls: {
+					[ DUMMY_ACTION ]: () => controlReturn,
+				},
+			}
+		) );
+	} );
+
+	describe( 'actions', () => {
+		describe( 'submitChanges', () => {
+			it( 'it proxies the selector call to the module with the given slug', async () => {
```
aaemnnosttv, line 70: This was probably just an oversight but the `it` shouldn't be part of the string. Other tests have this as well and should be updated.

Also, `submitChanges` is an action so the wording needs to be adapted for accuracy here.
```suggestion
			it( 'proxies the dispatched action to the module with the given slug', async () => {
```

assets/js/googlesitekit/modules/datastore/settings.test.js
```diff
@@ -0,0 +1,104 @@
...
+				const expectedError = { error: `'modules/${ nonExistentModuleSlug }' does not have a submitChanges() action.` };
+				expect( await registry.dispatch( STORE_NAME ).submitChanges( nonExistentModuleSlug ) ).toEqual( expectedError );
+
+				expect( await registry.dispatch( STORE_NAME ).submitChanges( nonExistentModuleSlug ) ).toEqual( expectedError );
```
aaemnnosttv, line 74: These are duplicate calls and assertions here.

assets/js/googlesitekit/modules/datastore/settings.js
```diff
@@ -0,0 +1,81 @@
...
+	/**
+	 * Submits all changes for a module.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {string} slug Slug for module store.
+	 * @return {Object} Module's submitChanges response object if it exists, otherwise object with `error` property if it doesn't.
+	 */
+	*submitChanges( slug ) {
+		slug = slug ? slug : '';
```
aaemnnosttv, line 40: After testing this with the latest develop included, the `invariant` call works as expected here, so let's restore that.
```suggestion
		invariant( slug, 'slug is required.' );
```

assets/js/googlesitekit/modules/datastore/settings.test.js
```diff
@@ -0,0 +1,104 @@
...
+
+				const expectedErrorNoSlug = { error: "'modules/' does not have a submitChanges() action." };
+				expect( await registry.dispatch( STORE_NAME ).submitChanges() ).toEqual( expectedErrorNoSlug );
```
aaemnnosttv, line 77: As mentioned in a previous comment, the invariant call works as expected now in the action.

After restoring that, let's extract this to a new test similar to the one I linked previously where we assert that submitChanges throws the expected invariant message when called without a `slug`.

The test that checks `modules/` in the error here can be removed here since that would no longer be possible.

assets/js/googlesitekit/modules/datastore/settings.test.js
```diff
@@ -0,0 +1,104 @@
...
+	describe( 'selectors', () => {
+		describe( 'isDoingSubmitChanges', () => {
+			it( 'it proxies the selector call to the module with the given slug', async () => {
+				expect( registry.select( STORE_NAME ).isDoingSubmitChanges( nonExistentModuleSlug ) ).toBe( false );
+
+				expect( registry.select( STORE_NAME ).isDoingSubmitChanges( slug ) ).toBe( false );
+				submittingChanges = true;
+				expect( registry.select( STORE_NAME ).isDoingSubmitChanges( slug ) ).toBe( true );
+			} );
+		} );
```
aaemnnosttv, line 93: Super nitpick, but please add a blank line between adjacent `describe` and `it` calls. No biggie though.
```suggestion
		} );

```

aaemnnosttv (commented):
Looking good @ivankruchkoff – there are a few changes requested from before that are still yet to be applied. Apart from those I left a few comments on the remaining bits to address which should be fairly quick.

assets/js/googlesitekit/modules/datastore/settings.test.js
ivankruchkoff: Updated

assets/js/googlesitekit/modules/datastore/settings.js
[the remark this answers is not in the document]
ivankruchkoff: Updated.

assets/js/googlesitekit/modules/datastore/settings.js
ivankruchkoff: Updated to include the invariant with generator in the fn.

assets/js/googlesitekit/modules/datastore/settings.test.js
ivankruchkoff: Updated

assets/js/googlesitekit/modules/datastore/settings.test.js
ivankruchkoff: Updated.

assets/js/googlesitekit/modules/datastore/settings.test.js
```diff
@@ -0,0 +1,105 @@
...
+	} );
+
+	describe( 'actions', () => {
+		describe( 'submitChanges', () => {
+			it( 'proxies the dispatched action to the module with the given slug', async () => {
+				const expectedError = { error: `'modules/${ nonExistentModuleSlug }' does not have a submitChanges() action.` };
+				expect( await registry.dispatch( STORE_NAME ).submitChanges( nonExistentModuleSlug ) ).toEqual( expectedError );
+
+				expect( await registry.dispatch( STORE_NAME ).submitChanges( slug ) ).toBe( controlReturn );
+			} );
```
aaemnnosttv, line 75:
```suggestion
			} );

```

assets/js/googlesitekit/modules/datastore/settings.test.js
```diff
@@ -0,0 +1,106 @@
...
+				expect( () => {
+					registry.dispatch( STORE_NAME ).submitChanges();
+				} ).toThrow();
+			} );
+		} );
+	} );
+
+	describe( 'selectors', () => {
+		describe( 'isDoingSubmitChanges', () => {
+			it( 'it proxies the selector call to the module with the given slug', async () => {
```
aaemnnosttv, line 87:
```suggestion
			it( 'proxies the selector call to the module with the given slug', async () => {
```

assets/js/googlesitekit/modules/datastore/settings.test.js
```diff
@@ -0,0 +1,106 @@
...
+				expect( registry.select( STORE_NAME ).isDoingSubmitChanges( nonExistentModuleSlug ) ).toBe( false );
+
+				expect( registry.select( STORE_NAME ).isDoingSubmitChanges( slug ) ).toBe( false );
+				submittingChanges = true;
+				expect( registry.select( STORE_NAME ).isDoingSubmitChanges( slug ) ).toBe( true );
+			} );
+		} );
+
+		describe( 'canSubmitChanges', () => {
+			it( 'it proxies the selector call to the module with the given slug', () => {
```
aaemnnosttv, line 97:
```suggestion
			it( 'proxies the selector call to the module with the given slug', () => {
```

aaemnnosttv (approved):
👍

Issue #2191, opened 2020-10-15
Refactor `ResetButton` component


felixarntz:
The `ResetButton` component should be refactored to a functional component using the datastore.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `ResetButton` component should be refactored to a functional component using the datastore and hooks.
* Its file should be renamed so that the file name matches the component name, and all import references should be updated accordingly.
* There should be a working Storybook story for the button under "Global".

## Implementation Brief

* Rename `assets/js/components/reset-button.js` to `assets/js/components/ResetButton.js`. Update all imports accordingly, or use VSCode's feature to do it automatically:
<img width="663" alt="Screenshot 2020-10-16 at 22 44 12" src="https://user-images.githubusercontent.com/90871/96311316-1e0b4980-1001-11eb-8429-af1a7d4d4115.png">

* Refactor the `ResetButton` to use hooks. The main bit of refactoring will be around `componentDidMount`/`componentWillUnmount`. Something like:

```js
const [ dialogActive, setDialogActive ] = useState( false );
const handleCloseModal = useCallback( ( event ) => {
    if ( 27 === event.keyCode ) {
        setDialogActive(false);
    }
}, [ setDialogActive ] );
useEffect( () => {
    global.addEventListener( 'keyup', handleCloseModal, false );

    return global.removeEventListener( 'keyup', handleCloseModal );
}, [ handleCloseModal ] )
```

* Change `splashURL` to use `useSelect( ( select ) => select( CORE_SITE ).getAdminURL( 'googlesitekit-splash' ) );` instead of the current `global._googlesitekitBaseData.splashURL`.
* Remove `postResetURL` from state and make it a constant in the component, as it does not change.
* Refactor `handleUnlinkConfirm()` to use `dispatch( 'core/site' ).reset()`
* There are additional, small refactors needed, but they're mostly trivial/obvious ones resulting from a class-to-functional component refactor.
* Add a new Storybook story to `stories/ResetButton.js` under `storiesOf( 'Global', module ).add( 'ResetButton', () => {} )`, demonstrating the reset button and its modal dialog in use.

### Test Coverage

* Adding test coverage for the `ResetButton` is welcome, largely around the handling of modal clicks if appropriate.

### Visual Regression Changes

* There should be no visual changes resulting from this refactoring.

## QA Brief

[...]

PR #2268, opened 2020-10-26
Refactor ResetButton component

kostyalmm:
## Summary

Addresses issue #2191 

## Relevant technical choices

- Refactors the `ResetButton` component to a functional component.
- Adds story for the component

## Checklist


assets/js/components/ResetButton.test.js
```diff
@@ -0,0 +1,109 @@
...
+			fireEvent.click( document.querySelector( '.mdc-dialog--open .mdc-dialog__cancel-button' ) );
+			expect( document.querySelector( '.mdc-dialog--open' ) ).toBeNull();
+		} );
+
+		it( 'should close the modal on pressing escape key', async () => {
+			fireEvent.keyUp( global, { keyCode: 27 } );
+			expect( document.querySelector( '.mdc-dialog--open' ) ).toBeNull();
+		} );
+
+		it( 'should close the modal on clicking Reset and ', async () => {
```
tofumatt, line 73: I think the name of this test is incomplete or has an extra "and" 🙂

assets/js/components/ResetButton.test.js
```diff
@@ -0,0 +1,109 @@
...
+				return {
+					getItem( key ) {
+						return store[ key ];
+					},
+					clear() {
+						store = {};
+					},
+				};
+			}() );
+			Object.defineProperty( global, 'localStorage', { value: localStorageMock } );
```
tofumatt, line 96: We include `jest-localstorage-mock` in our tests already, so I don't think you need to do this here 🙂 

See: https://github.com/google/site-kit-wp/blob/5e51067e3a4bc93f7cd5419d798f6ee2d0f56713/assets/js/googlesitekit/api/cache.test.js#L123 for an example of it in-use.

assets/js/components/ResetButton.test.js
```diff
@@ -0,0 +1,109 @@
...
+		} );
+
+		it( 'should close the modal on clicking Reset and ', async () => {
+			const response = true;
+			fetchMock.postOnce(
+				/^\/google-site-kit\/v1\/core\/site\/data\/reset/,
+				{ body: JSON.stringify( response ), status: 200 },
+			);
+
+			delete global.location;
```
tofumatt, line 80: Why do we delete the location here as part of the test?

tofumatt (changes requested):
Great start, but looks like there are some improvements we can make to the tests and also to the component. Let me know what you think about my comments and feel free to assign me to the code review once it's ready again 🙂

assets/js/components/ResetButton.js
```diff
@@ -0,0 +1,103 @@
...
+	useEffect( () => {
+		global.addEventListener( 'keyup', handleCloseModal, false );
+
+		return () => global.removeEventListener( 'keyup', handleCloseModal );
+	}, [ handleCloseModal ] );
+
+	const handleUnlinkConfirm = async () => {
+		resetCoreSite();
+		clearWebStorage();
+		handleDialog();
```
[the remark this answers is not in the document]
kostyalmm: I think the right UX will be to show a loader in place of reset button until the page redirects. What do you think?
Otherwise we can just keep on showing the dialog and redirect the page.

assets/js/components/ResetButton.js
```diff
@@ -0,0 +1,103 @@
...
+	const handleCloseModal = useCallback( ( event ) => {
+		if ( 27 === event.keyCode ) {
+			setDialogActive( false );
+		}
+	}, [ dialogActive ] );
+
+	useEffect( () => {
+		global.addEventListener( 'keyup', handleCloseModal, false );
+
+		return () => global.removeEventListener( 'keyup', handleCloseModal );
```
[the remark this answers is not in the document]
kostyalmm: > **Note**
> 
> React guarantees that setState function identity is stable and won’t change on re-renders. This is why it’s safe to omit from the useEffect or useCallback dependency list.

According to [this](https://reactjs.org/docs/hooks-reference.html#usestate), `setDialogActive` will always be the same as `setState` is always same. This means the `useCallback` value (`handleCloseModal`) will never change. In turn, the `useEffect` will never run except component mount and unmount. 

What I think should be instead here is this (ignore comments and console logs for now):
```js
const handleCloseModal = useCallback( ( event ) => {
	if ( 27 === event.keyCode ) {
		// Only close the modal if the "Escape" key is used.
		setDialogActive( false );
	}
	// eslint-disable-next-line no-console
	console.log( 'event callback called' );
// as it depends on nothing, an empty dependency array
}, [ ] );

useEffect( () => {
	if ( dialogActive ) {
                // only when the dialog is shown, add the event listener
		// eslint-disable-next-line no-console
		console.log( 'adding listener' );
		global.addEventListener( 'keyup', handleCloseModal, false );
	}

	return () => {
		global.removeEventListener( 'keyup', handleCloseModal );
		// eslint-disable-next-line no-console
		console.log( 'removing listener' );
	};
// will be run when the dialog is shown or hidden
}, [ dialogActive ] );
```

In this case, the listener is removed when the dialog is closed and added when the dialog is opened. What do you think 🤔 ?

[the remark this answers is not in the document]
tofumatt: Yeah, I actually thought the `setState` function wouldn't change, but had it in there for completeness. But I suspected `useCallback` with no arguments will work.

I think you can remove the `[]` argument altogether, but otherwise looks good.

-----

Regarding having `useEffect` update when `dialogActive` changes: that's fine, but your proposal isn't quite right because technically every time it changes and `dialogActive` is `true`, _another_ event handler will be attached as I understand it.

I forget if attaching the same handler to an event twice will actually cause it to be run twice, but we don't need that.

(In your suggestion, the event listener will persist until the component unmounts.)

Basically, we want:

1. mount the component
2. define the callback
3. attach the `handleCloseModal` handlers

and then everything just stays as-is until we unmount, when the handler should be unmounted.

But I guess removing the listener when the dialog disappears is actually a good call, so let's do that:

```js
const handleCloseModal = useCallback( ( event ) => {
	if ( 27 === event.keyCode ) {
		// Only close the modal if the "Escape" key is used.
		setDialogActive( false );
	}
} );

useEffect( () => {
	if ( dialogActive ) {
		global.addEventListener( 'keyup', handleCloseModal, false );

        return () => {
            global.removeEventListener( 'keyup', handleCloseModal );
        };
	} else {
        // Remove the event listener when the dialog is removed; there's no need
        // to have it attached when it won't be used.
        global.removeEventListener( 'keyup', handleCloseModal );
    }
}, [ dialogActive ] );
```

assets/js/components/ResetButton.js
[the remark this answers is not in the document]
tofumatt: I had a look and the existing UX definitely just closes the reset dialog and then redirects. It happens pretty quickly on a decent internet connection, but it isn't great.

I don't think showing a generic loading animation is a good indication of what's _currently happening_, because it would remove "Reset Site Kit" with "…" (some loading animation) so you might not _know_ what's loading/resetting. Changing the button to be disabled with the text "Site Kit is resetting" would be nice, but it's a UX change unrelated to the original IB and too much to add to this issue.

I think it'd be nicer if the dialog didn't _close_ on `handleUnlinkConfirm` though. But I'll file a follow-up issue to discuss all of this; for now let's leave it as-is 😄

tofumatt (dismissed):
Looks good, thanks for the changes 👍🏻

assets/js/components/ResetButton.test.js
```diff
@@ -0,0 +1,98 @@
...
+				writable: true,
+			} );
+
+			await act( async () => {
+				fireEvent.click( document.querySelector( '.mdc-dialog--open .mdc-button--danger' ) );
+				await subscribeUntil( registry, () => registry.select( CORE_SITE ).isDoingReset() === false );
+			} );
+
+			expect( fetchMock ).toHaveFetchedTimes( 1 );
+			expect( document.querySelector( '.mdc-dialog--open' ) ).toBeNull();
```
aaemnnosttv, line 93:
```suggestion
			expect( document.querySelector( '.mdc-dialog--open' ) ).not.toBeInTheDocument();
```

assets/js/components/ResetButton.test.js
```diff
@@ -0,0 +1,98 @@
...
+			expect( localStorage.clear ).toHaveBeenCalled();
+			expect( global.location.href ).toBe( 'http://example.com/admin.php?page=googlesitekit-splash&notification=reset_success' );
```
aaemnnosttv, line 95: While this is correct, we can make this test a bit more resilient to changes that shouldn't cause it to fail, such as the order of query args in the URL. For this we have a custom matcher `toMatchQueryParameters`
```suggestion
			expect( global.location.pathname ).toBe( '/wp-admin/admin.php' );
			expect( global.location.href ).toMatchQueryParameters( {
				page: 'googlesitekit-splash',
				notification: 'reset_success',
			} );
```
Also, it looks like `wp-admin` is missing here which is due to the way `adminURL` is received above.

assets/js/components/ResetButton.test.js
```diff
@@ -0,0 +1,98 @@
...
+import ResetButton from './ResetButton';
+import { subscribeUntil } from '../../../tests/js/utils';
+
+describe( 'ResetButton', () => {
+	let registry;
+	const adminURL = 'http://example.com/';
+
+	beforeEach( () => {
+		registry = createTestRegistry();
+		registry.dispatch( CORE_SITE ).receiveSiteInfo( { adminURL } );
```
aaemnnosttv, line 33: Please use `provideSiteInfo` from `tests/js/utils.js` instead using the default values.

stories/reset-button.stories.js
```diff
@@ -0,0 +1,45 @@
...
+import ResetButton from '../assets/js/components/ResetButton';
+import { STORE_NAME as CORE_SITE } from '../assets/js/googlesitekit/datastore/site/constants';
+import { WithTestRegistry } from '../tests/js/utils';
+
+storiesOf( 'Global', module )
+	.add( 'Reset Button', () => {
+		const setupRegistry = ( { dispatch } ) => {
+			dispatch( CORE_SITE ).receiveSiteInfo( {
+				adminURL: '#',
+			} );
```
aaemnnosttv, line 36: Please use `provideSiteInfo` instead 🙂

stories/settings.stories.js
```diff
@@ -103,11 +104,19 @@ storiesOf( 'Settings', module )
 		global._googlesitekitLegacyData.admin.clientID = '123456789-xxx1234ffghrrro6hofusq2b8.apps..com';
 		global._googlesitekitLegacyData.admin.clientSecret = '••••••••••••••••••••••••••••';
 
+		const setupRegistry = ( { dispatch } ) => {
+			dispatch( CORE_SITE ).receiveSiteInfo( {
+				adminURL: '#',
+			} );
```
aaemnnosttv, line 110: See above.

assets/js/components/ResetButton.test.js
```diff
@@ -0,0 +1,98 @@
...
+			} );
+
+			await act( async () => {
+				fireEvent.click( document.querySelector( '.mdc-dialog--open .mdc-button--danger' ) );
+				await subscribeUntil( registry, () => registry.select( CORE_SITE ).isDoingReset() === false );
+			} );
+
+			expect( fetchMock ).toHaveFetchedTimes( 1 );
+			expect( document.querySelector( '.mdc-dialog--open' ) ).toBeNull();
+			expect( localStorage.clear ).toHaveBeenCalled();
```
aaemnnosttv, line 94: Please add an assertion for `sessionStorage` as well.

aaemnnosttv (changes requested):
Thanks @kostyalmm – this is looking pretty good, just a few minor changes to make. I left a number of comments below, most with suggestions. Several of them are duplicates so it may look like there are more than there really are. Thanks 👍

assets/js/components/ResetButton.js
```diff
@@ -0,0 +1,107 @@
...
+
+	const toggleDialogActive = () => {
+		setDialogActive( ! dialogActive );
+	};
+
+	return (
+		<Fragment>
+			<Link
+				className="googlesitekit-reset-button"
+				onClick={ () => setDialogActive( true ) }
```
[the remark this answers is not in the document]
kostyalmm: @aaemnnosttv 
IMHO As explained in [this post](https://kentcdodds.com/blog/usememo-and-usecallback#why-is-usecallback-worse),  using `useCallback` here with `[]` dependency is not worth it and will be less efficient than it already is.
If you still feel that a `useCallback` will be a better thing to do here, I will be happy to make the change 😊

assets/js/components/ResetButton.js
```diff
@@ -0,0 +1,106 @@
...
+import { Fragment, useState, useEffect, useCallback, createInterpolateElement } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import Data from 'googlesitekit-data';
+import { clearWebStorage } from '../util';
+import Dialog from './dialog';
+import Link from './Link';
+import Modal from './modal';
```
felixarntz, line 32: This will need to be updated after #2321 is merged.

assets/js/components/ResetButton.js
```diff
@@ -0,0 +1,106 @@
...
+		global.location.assign( postResetURL );
+	}, [ reset, postResetURL ] );
+
+	const toggleDialogActive = useCallback( () => {
+		setDialogActive( ! dialogActive );
+	}, [ dialogActive ] );
+
+	return (
+		<Fragment>
+			<Link
```
aaemnnosttv, line 80: Not sure why this was a `Link` before, but let's make it a `<Button>` instead, including `type="button"` so it doesn't potentially trigger a form submit when clicked.

assets/js/components/ResetButton.js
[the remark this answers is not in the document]
aaemnnosttv: Thanks for raising this. I've read that article before but it's been a while. I find the [article by Ryan Florence](https://reacttraining.com/blog/react-inline-functions-and-performance/) which is linked at the bottom a bit more useful. This is worth its own issue for Site Kit to re-evaluate our use of our performance related optimizations.

For this component specifically, it shouldn't be necessary. I suppose for now it's more of a question of not using in some places while using it everywhere else.

So, let's update it to use `useCallback` for now just for consistency, but I'll open an issue to discuss performance optimizations going forward 👍

assets/js/components/ResetButton.test.js
aaemnnosttv: @kostyalmm – you resolved the question here without a reply; would you please clarify?

assets/js/components/ResetButton.test.js
```diff
@@ -0,0 +1,98 @@
...
+	it( 'should render passed children inside of the button', () => {
+		const { container } = render( <ResetButton>Test Value</ResetButton>, { registry } );
+		expect( container ).toHaveTextContent( 'Test Value' );
+		expect( document.querySelector( '.mdc-dialog--open' ) ).toBeNull();
+	} );
+
+	describe( 'after click', () => {
+		let container;
+		beforeEach( () => {
+			( { container } = render( <ResetButton />, { registry } ) );
```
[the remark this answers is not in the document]
aaemnnosttv: @kostyalmm

assets/js/components/ResetButton.test.js
aaemnnosttv: @kostyalmm

aaemnnosttv (changes requested):
@kostyalmm – looks like there are a few comments left from previous rounds which were resolved without changes or a response. Please leave conversations to be resolved by the reviewer, thanks 👍

assets/js/components/ResetButton.test.js
kostyalmm: @aaemnnosttv I had removed this statement earlier and the code worked without it. But with the latest(a couple of days ago) requested changes, I had to reintroduce this. I had forgotten about this comment from earlier. I'll take care of such cases moving forward.

assets/js/components/ResetButton.test.js
kostyalmm: I've added it. It is on the next line.

kostyalmm: I have added this already. Please see the next line.

assets/js/components/ResetButton.test.js
kostyalmm: Why I reintroduced it:
For some reason if I just reassign `global.location` and set its `assign` property to a jest mock function, it does not work properly.
I tried doing it without the `delete`, but the mock function was not being called at all.
After researching online, I noticed that all the people are deleting the `global.location` object for the same use case. When I deleted it, the test worked fine.
I am not aware of the reason why this happens, maybe some reference issues, but deleting the location object solves it.

assets/js/components/ResetButton.js
kostyalmm: `Button`:
 image

`Link`(This is how it is right now):
 image

Do you want me to change this?

assets/js/components/ResetButton.test.js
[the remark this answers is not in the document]
kostyalmm: I reverted this earlier, seems that the pre commit hook is changed this automatically.

assets/js/components/ResetButton.js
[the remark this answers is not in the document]
tofumatt: Filed #2347 to discuss.

assets/js/components/ResetButton.js
aaemnnosttv: Oh, no you're right 😄  my mistake, I was thinking of the button that is shown in the modal when you click that 👍 

Please keep as-is, thank you 🙂

assets/js/components/ResetButton.test.js
aaemnnosttv: Thanks for the context @kostyalmm – I took a look into this and found that it has to do with restrictions introduced in a recent version of jsdom.

You are correct that we can redefine the `location` global, we just can't change it. I found we can keep everything else like so:

```js
	Object.defineProperty( global, 'location', {
		writable: true,
		value: {
			...global.location,
			assign: jest.fn(),
		},
	} );
```

**Let's keep the location mocking in this test for now** and we can come back to this with a follow-up issue since this should probably be baked into the test setup in a top-level `beforeAll`.

One thing we should do here is backup the original "real" `location` and restore it at the end so that other tests aren't affected by this since it is not reset between tests as it is.

assets/js/components/ResetButton.js
```diff
@@ -0,0 +1,110 @@
...
+import Link from './Link';
+import { STORE_NAME as CORE_SITE } from '../googlesitekit/datastore/site/constants';
+
+const { useSelect, useDispatch } = Data;
+
+function ResetButton( { children } ) {
+	const postResetURL = useSelect( ( select ) => select( CORE_SITE ).getAdminURL( 'googlesitekit-splash', { notification: 'reset_success' } ) );
+
+	const [ dialogActive, setDialogActive ] = useState( false );
+	// const postResetURL = addQueryArgs( splashURL,  );
```
aaemnnosttv, line 41:
```suggestion
```

assets/js/components/ResetButton.js
```diff
@@ -0,0 +1,110 @@
...
+
+	const handleCloseModal = useCallback( ( event ) => {
```
aaemnnosttv, line 43: Let's move the definition of this function into the `useEffect` below since that's the only place it's used.

aaemnnosttv (commented):
Thanks @kostyalmm – just a few last things to address and then this should be good to go 👍

aaemnnosttv (approved):
LGTM, thanks!

Issue #2136, opened 2020-10-06
Ensure every module store has a canSubmitChanges selector and submitChanges action


felixarntz:
So far all Site Kit module stores that support editing settings via their settings panel have a `canSubmitChanges` selector, which checks whether the current client-side settings are valid to be saved, as well as a `submitChanges` action, which actually submits the changes persistently with the server. While these two have been added to stores by convenience so far, going forward it should be ensured that _every_ module store (per using the `createModuleStore` utility) has this selector and action respectively.

Related: #2108

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a new utility function `createSubmitChangesStore( options )` in `assets/js/googlesitekit/modules/create-submit-changes-store.js`:
    * Accepts an optional argument `submitChanges`, which needs to be an asynchronous function which receives `registry` as only parameter. The default should be `() => { return {}; }`.
    * Accepts an optional argument `validateCanSubmitChanges`, which needs to be a function which receives `registry.select` as first parameter and `state` as second parameter. The default should be an empty `() => {}`.
    * The function should return a (partial) store with:
        * A `submitChanges` action and related reducers, behaving exactly like existing `submitChanges` implementations in module stores (e.g. `modules/analytics`). The `SUBMIT_CHANGES` control should be a registry control that simply calls the provided `submitChanges` function. That function should then return a result object, with an `error` property if the process failed.
        * A `canSubmitChanges` selector, which is the `safeSelector` created via `createValidationSelector` (see #2108) with the provided `validateCanSubmitChanges` function. That function should then return `true` or `false`.
        * An `isDoingSubmitChanges` selector, which should look at whether the `isDoingSubmitChanges` flag is set in state or not. It should return `true` or `false`.
* `createModuleStore` should be expanded to also include the new optional `canSubmitChanges` and `submitChanges` arguments and based on them include the partial store from `createSubmitChangesStore` in the returned module store. The defaults here should be based on the existing usages and on whether the module has settings:
    * If `settingSlugs` is not set (i.e. there are no settings for the module), the defaults should be the same no-op defaults like in `createSubmitChangesStore`.
    * If `settingSlugs` is set (i.e. there is a settings store as part of the module), the defaults should be:
		**`submitChanges`:**
		```js
		async( registry ) => {
			if ( registry.select( storeName ).haveSettingsChanged() ) {
				const { error } = await registry.dispatch( storeName ).saveSettings();
				if ( error ) {
[...]

PR #2278, opened 2020-10-27
Enhancement/2136 submit changes changes

eugene-manuilov:
## Summary

Addresses issue #2136

## Checklist


assets/js/googlesitekit/modules/create-submit-changes-store.js
```diff
@@ -0,0 +1,201 @@
...
+
+	const controls = {
+		[ SUBMIT_CHANGES ]: submitChanges,
+	};
+
+	const resolvers = {};
+
+	const {
+		safeSelector: canSubmitChanges,
+		dangerousSelector: validateSubmitChanges,
```
felixarntz, line 136: This one should be exposed as `__dangerousCanSubmitChanges` like before to make it obvious it shouldn't be used externally.

assets/js/googlesitekit/modules/create-submit-changes-store.js
```diff
@@ -0,0 +1,201 @@
...
+export function createSubmitChangesStore( {
+	storeName,
+	submitChanges = () => () => ( {} ),
+	validateCanSubmitChanges = () => {},
+} = {} ) {
+	invariant( storeName, 'storeName is required.' );
+
+	const initialState = {
+		hasStartedSubmittingChanges: false,
+		hasFinishedSubmittingChanges: false,
```
felixarntz, line 63: Let's remove these two - we only need `isDoingSubmitChanges` like in the existing implementations.

assets/js/googlesitekit/modules/create-submit-changes-store.js
```diff
@@ -0,0 +1,201 @@
...
+ *
+ * @param {Object}   args                            Arguments for creating the submitChanges store.
+ * @param {string}   args.storeName                  Datastore slug.
+ * @param {Function} [args.submitChanges]            Optional. Callback function to issue the submit changes request. Will be used inside the submit changes control.
+ * @param {Function} [args.validateCanSubmitChanges] Optional. A helper function to validate that settings can be submitted.
+ * @return {Object} Partial store object with properties 'actions', 'controls', 'reducer', 'resolvers', and 'selectors'.
+ */
+export function createSubmitChangesStore( {
+	storeName,
+	submitChanges = () => () => ( {} ),
```
felixarntz, line 56: I think this is incorrect? It should be a function that returns the empty object, not a function that returns a function.

assets/js/googlesitekit/modules/create-submit-changes-store.js
```diff
@@ -0,0 +1,201 @@
...
+		/**
+		 * Submits all changes currently present in the client, persisting them on the server.
+		 *
+		 * @since n.e.x.t
+		 *
+		 * @return {Object} Empty object on success, object with `error` property on failure.
+		 */
+		*submitChanges() {
+			const { dispatch } = yield Data.commonActions.getRegistry();
+			dispatch( storeName ).clearError( 'submitChanges', [] );
```
felixarntz, line 77: We should only dispatch this action if it's available on the store. We can use a similar approach as in `createFetchStore`, where the direct action creators are called.

assets/js/googlesitekit/modules/create-submit-changes-store.js
```diff
@@ -0,0 +1,201 @@
...
+				payload: {},
+			};
+
+			const result = yield {
+				type: SUBMIT_CHANGES,
+				payload: {},
+			};
+
+			if ( result.error ) {
+				yield dispatch( storeName ).receiveError( result.error, 'submitChanges', [] );
```
felixarntz, line 90: See above, same applies here.

assets/js/googlesitekit/modules/create-submit-changes-store.js
```diff
@@ -0,0 +1,201 @@
...
+		},
+	};
+
+	const reducer = ( state, { type } ) => {
+		switch ( type ) {
+			case START_SUBMIT_CHANGES: {
+				return {
+					...state,
+					hasStartedSubmittingChanges: true,
+					hasFinishedSubmittingChanges: false,
```
felixarntz, line 108: See above, let's remove these (also below).

assets/js/googlesitekit/modules/create-submit-changes-store.js
```diff
@@ -0,0 +1,201 @@
...
+
+	const {
+		safeSelector: canSubmitChanges,
+		dangerousSelector: validateSubmitChanges,
+	} = createValidationSelector( validateCanSubmitChanges );
+
+	const selectors = {
+		canSubmitChanges,
+		validateSubmitChanges,
+		...createStateSelectors( Object.keys( initialState ) ),
```
felixarntz, line 142: I think this abstraction goes a bit too far - especially since we only need a single state property, and for the benefit of better documentation, let's write out the selector `isDoingSubmitChanges` instead of doing this.

assets/js/googlesitekit/data/utils.js
```diff
@@ -348,3 +348,30 @@ export function createValidationSelector( validate ) {
...
+ *
+ * @param {Array.<string>} properties Array of property names.
+ * @return {Object} An object with created selectors.
+ */
+export function createStateSelectors( properties ) {
+	return properties.reduce( ( accum, property ) => ( {
+		...accum,
+		[ property ]: createStateSelector( property ),
+	} ), {} );
+}
```
felixarntz, line 377: I don't think we should have/use these, see my previous comment.

assets/js/googlesitekit/modules/create-submit-changes-store.js
```diff
@@ -0,0 +1,201 @@
...
+			}
+
+			default: {
+				return state;
+			}
+		}
+	};
+
+	const controls = {
+		[ SUBMIT_CHANGES ]: submitChanges,
```
felixarntz, line 129: See ACs and https://github.com/google/site-kit-wp/issues/2136#issuecomment-707262704, this should be using `createRegistryControl` and from there call `submitChanges`, passing the registry to that function. The function passed as `submitChanges` should not have to call `createRegistrySelector` itself.

assets/js/googlesitekit/modules/create-submit-changes-store.js
```diff
@@ -0,0 +1,201 @@
...
+ * Creates a default submitChanges control function.
+ *
+ * @since n.e.x.t
+ *
+ * @param {string} slug      Module slug.
+ * @param {string} storeName Datastore slug.
+ * @return {Function} Control function to submit changes.
+ */
+export function makeDefaultSubmitChanges( slug, storeName ) {
+	return createRegistryControl( ( { select, dispatch } ) => async () => {
```
felixarntz, line 165: See above, this function shouldn't include the call to `createRegistryControl`.

assets/js/googlesitekit/modules/create-submit-changes-store.js
```diff
@@ -0,0 +1,201 @@
...
+		const strictSelect = createStrictSelect( select );
+		const {
+			haveSettingsChanged,
+			isDoingSubmitChanges,
+		} = strictSelect( storeName );
+
+		invariant( ! isDoingSubmitChanges(), INVARIANT_DOING_SUBMIT_CHANGES );
+		invariant( haveSettingsChanged(), INVARIANT_SETTINGS_NOT_CHANGED );
+	};
+}
```
felixarntz, line 201: `makeDefaultSubmitChanges` and `makeDefaultCanSubmitChanges` are defaults only for module stores that have settings, so I think these functions would be more suitable to place within `create-module-store.js` instead.

assets/js/googlesitekit/modules/create-module-store.js
```diff
@@ -79,12 +88,19 @@ export const createModuleStore = ( slug, {
...
+			validateCanSubmitChanges: validateCanSubmitChanges || makeDefaultCanSubmitChanges( storeName ),
+		} );
+
 		// to prevent duplication errors during combining stores, we don't need to combine
 		// Data.commontStore here since settingsStore already uses commonActions and commonControls
 		// from the Data.commonStore.
 		combinedStore = Data.combineStores(
 			notificationsStore,
 			settingsStore,
+			submitChangesStore,
```
felixarntz, line 103: The `else` clause (where `settingSlugs` is `undefined`) is missing the `submitChangesStore`. Per the ACs, this partial store should be part of _every_ module store, regardless of whether or not it has settings. Only the defaults should differ (in case of no settings, rely on the regular defaults from `createSubmitChangesStore`).

assets/js/modules/analytics/datastore/settings.js
```diff
@@ -55,137 +53,52 @@ export const INVARIANT_INVALID_PROFILE_NAME = 'a valid profile name is required
...
-
-		if ( result.error ) {
-			yield registry.dispatch( STORE_NAME ).receiveError( result.error, 'submitChanges', [] );
-		}
-
-		yield {
-			payload: {},
-			type: FINISH_SUBMIT_CHANGES,
-		};
+export const submitChanges = createRegistryControl( ( { select, dispatch } ) => async () => {
```
felixarntz, line 56: See above comment, this shouldn't include `createRegistryControl`. Also applies elsewhere throughout the PR.

assets/js/modules/adsense/datastore/settings.test.js
```diff
@@ -253,11 +253,11 @@ describe( 'modules/adsense settings', () => {
 				expect( registry.select( STORE_NAME ).canSubmitChanges() ).toBe( true );
 
 				registry.dispatch( STORE_NAME ).setAccountID( '0' );
-				expect( () => registry.select( STORE_NAME ).__dangerousCanSubmitChanges() )
+				expect( () => registry.select( STORE_NAME ).validateSubmitChanges() )
```
felixarntz, line 256: See above comment, the previous name should be kept here.

felixarntz (changes requested):
This is overall on a good path, but there are a couple things here that diverge from the ACs, plus some other specific feedback.

assets/js/googlesitekit/modules/create-submit-changes-store.js
eugene-manuilov: Updated.

assets/js/googlesitekit/modules/create-submit-changes-store.js
eugene-manuilov: Yes, it's a typo. Updated.

assets/js/googlesitekit/modules/create-submit-changes-store.js
eugene-manuilov: Updated.

assets/js/googlesitekit/modules/create-submit-changes-store.js
eugene-manuilov: Updated.

assets/js/googlesitekit/modules/create-module-store.js
eugene-manuilov: Updated.

assets/js/modules/adsense/datastore/settings.test.js
eugene-manuilov: Updated.

assets/js/googlesitekit/modules/create-submit-changes-store.js
eugene-manuilov: Removed.

assets/js/googlesitekit/data/utils.js
eugene-manuilov: Removed.

assets/js/googlesitekit/modules/create-submit-changes-store.js
eugene-manuilov: Updated.

assets/js/googlesitekit/modules/create-submit-changes-store.js
eugene-manuilov: Updated.

assets/js/googlesitekit/modules/create-submit-changes-store.js
eugene-manuilov: Updated.

assets/js/modules/analytics/datastore/settings.js
eugene-manuilov: Updated.

assets/js/googlesitekit/modules/create-submit-changes-store.js
eugene-manuilov: @felixarntz I have moved it to the `create-settings-store.js` since it is settings related functionality. Does it work for you?

felixarntz: Sounds good!

felixarntz (approved):
Awesome!

Issue #2188, opened 2020-10-15
Refactor `AnalyticsInactiveCTA` component


felixarntz:
The `AnalyticsInactiveCTA` component should be refactored to a functional component using the datastore.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `AnalyticsInactiveCTA` component should be refactored to a functional component using the datastore and hooks.
    * It shouldn't use any legacy utility functions or globals anymore to get data.
    * The only legacy function that's okay to keep is `showErrorNotification` since there's no replacement yet.
* Its file should be renamed so that the file name matches the component name, and all import references should be updated accordingly.
* There should be a working Storybook story for the button under "Global".

## Implementation Brief

* Change the static `AnalyticsInactiveCTA.setupAnalyticsClick` method to be a normal async callback function named `activateModule` that looks like:

```js
		const { activateModule } = useDispatch( MODULES_STORE );

		const activateModule = async () => {
			const { error, response } = await activateModule( 'analytics' );

			if ( ! error ) {
				global.location = response.moduleReauthURL;
			} else {
				showErrorNotification( GenericError, {
					id: 'analytics-setup-error',
					title: __( 'Internal Server Error', 'google-site-kit' ),
					description: error.message,
					format: 'small',
					type: 'win-error',
				} );
			}
		};
```

* Change `const { canManageOptions } = global._googlesitekitLegacyData.permissions;` to use hooks `const canManageOptions = useSelect( ( select ) => select( USER_STORE ).hasCapability( PERMISSION_MANAGE_OPTIONS ) );` where `USER_STORE` is from `import { STORE_NAME as USER_STORE, PERMISSION_MANAGE_OPTIONS } from '../googlesitekit/datastore/user/constants';`
* Change the class-based component to be a function-based component, with the function being largely based off the existing `AnalyticsInactiveCTA`'s `render()` method.
 Add a new Storybook story to `stories/module-analytics-cta.stories.js` under storiesOf( 'Global', module ).add( 'AnalyticsInactiveCTA', () => {} ), demonstrating the CTA, possibly including an error state if straightforward to implement.

### Test Coverage

* Add test coverage for the click actions for this component, both when the error state occurs and when the response is valid and there's a redirect.

### Visual Regression Changes

* There should be no visual changes as a result of this refactor.

## QA Brief

[...]

PR #2281, opened 2020-10-28
Refactor AnalyticsInactiveCTA and add new story for the same.

kostyalmm:
## Summary

Addresses issue #2188 

## Relevant technical choices

* Refactor `AnalyticsInactiveCTA` component to functional component
* Add story in Storybook for `AnalyticsInactiveCTA`

## Checklist


kostyalmm:
In contrast to the IB for the issue, the name of the click handler method is kept `setupAnalyticsClick` as `activateModule` is already defined (it is the name of an action from `core/module`). 
Any suggestions for a better name are welcomed.

assets/js/components/AnalyticsInactiveCTA.js
```diff
@@ -24,64 +24,56 @@ import PropTypes from 'prop-types';
...
-		return (
-			<CTA
-				title={ title }
-				description={ description }
-				onClick={ AnalyticsInactiveCTA.setupAnalyticsClick }
-				ctaLabel={ ctaLabel }
-			/>
-		);
+	if ( ! canManageOptions ) {
+		return <Fragment />;
```
aaemnnosttv, line 66: Please return `null` for components that should render nothing instead.

assets/js/components/AnalyticsInactiveCTA.js
```diff
@@ -91,6 +83,9 @@ AnalyticsInactiveCTA.propTypes = {
 };
 
 AnalyticsInactiveCTA.defaultProps = {
+	title: __( 'Learn more about what visitors do on your site.', 'google-site-kit' ),
+	description: __( 'Connect with Google Analytics to see unique visitors, goal completions, top pages and more.', 'google-site-kit' ),
+	ctaLabel: __( 'Set up Analytics', 'google-site-kit' ),
```
aaemnnosttv, line 88: See above.

stories/analytics-inactive-cta.stories.js
```diff
@@ -0,0 +1,34 @@
+/**
```
aaemnnosttv, line 1: Please add the a file header 🙂

assets/js/components/AnalyticsInactiveCTA.js
```diff
@@ -24,64 +24,56 @@ import PropTypes from 'prop-types';
...
-	activateOrDeactivateModule,
-	getReAuthURL,
 	showErrorNotification,
 } from '../util';
-import { refreshAuthentication } from '../util/refresh-authentication';
-import data from './data';
+import Data from 'googlesitekit-data';
 import CTA from './notifications/cta';
 import GenericError from './notifications/generic-error';
+import { STORE_NAME as USER_STORE, PERMISSION_MANAGE_OPTIONS } from '../googlesitekit/datastore/user/constants';
```
aaemnnosttv, line 39: I realize this was named this way in the IB but please use the same convention we use for store name aliases everywhere else.
```suggestion
import { STORE_NAME as CORE_USER, PERMISSION_MANAGE_OPTIONS } from '../googlesitekit/datastore/user/constants';
```

stories/analytics-inactive-cta.stories.js
```diff
@@ -0,0 +1,34 @@
...
+ */
+import { storiesOf } from '@storybook/react';
+
+/**
+ * Internal dependencies
+ */
+import AnalyticsInactiveCTA from '../assets/js/components/AnalyticsInactiveCTA';
+import {
+	PERMISSION_MANAGE_OPTIONS,
+	STORE_NAME as CORE_USER,
```
aaemnnosttv, line 12: Like here ✅  😄

aaemnnosttv (changes requested):
Thanks @kostyalmm ! This is off to a good start. Just a few changes to make here otherwise this looks pretty close.

assets/js/components/AnalyticsInactiveCTA.js
```diff
@@ -24,64 +24,60 @@ import PropTypes from 'prop-types';
...
-	static async setupAnalyticsClick() {
-		try {
-			await activateOrDeactivateModule( data, 'analytics', true );
+const { useSelect, useDispatch } = Data;
 
-			await refreshAuthentication();
+function AnalyticsInactiveCTA( {
+	title = __( 'Learn more about what visitors do on your site.', 'google-site-kit' ),
+	description = __( 'Connect with Google Analytics to see unique visitors, goal completions, top pages and more.', 'google-site-kit' ),
+	ctaLabel = __( 'Set up Analytics', 'google-site-kit' ),
```
felixarntz, line 47: I think writing out prop destructuring in parameters makes the code unnecessarily hard to read, unless it's for very minimal cases (e.g. 1-3 props without defaults). Generally, let's destructure props in a `const` assignment within a component - this is also an improvement because this way the variables are explicitly `const`.

In this particular case, it can furthermore cause actual problems because translation strings shouldn't be default parameter values - they need to be translated when the function is actually called, otherwise we might still run into the same timing issues @aaemnnosttv was mentioning before.

assets/js/components/AnalyticsInactiveCTA.js
```diff
@@ -24,64 +24,60 @@ import PropTypes from 'prop-types';
...
+	const { activateModule } = useDispatch( CORE_MODULES );
 
-			// Redirect to ReAuthentication URL
-			global.location = getReAuthURL( 'analytics', true );
-		} catch ( err ) {
+	const onSetupAnalytics = useCallback( async () => {
+		const { error, response } = await activateModule( 'analytics' );
+
+		if ( ! error ) {
+			global.location = response.moduleReauthURL;
```
felixarntz, line 55: This is still not correct from the previous review, we should be using `global.location.assign()`.

felixarntz (changes requested):
Mostly LGTM, however there are still two small problems here.

assets/js/components/AnalyticsInactiveCTA.js
aaemnnosttv: @felixarntz – using the translation strings here as JS default values is safe because these aren't evaluated until the function is called, unlike React `Component.defaultProps` which are evaluated when the module is loaded.
It's essentially the same thing you're suggesting (moving to a destructure with `const { } = props`), just inline.

No problems with that suggestion, but just clarifying that using default values here doesn't have the same risk as when defined outside of the component 👍

aaemnnosttv (approved):
LGTM 👍

aaemnnosttv:
@felixarntz can you have a quick look since you requested changes as well?

This should still wait to be merged until develop is cut for the release.

felixarntz (approved):
Great!

Issue #2180, opened 2020-10-15
Add `getModuleDependantNames` and `getModuleDependencyNames` selectors


felixarntz:
The `core/modules` store should receive new selectors that expose the list of module names that are dependencies / dependants of a given module.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `core/modules` store should have a new selector `getModuleDependencyNames( slug )`:
    * It should return an array of module names that are `dependencies` of the module identified by the given `slug`.
    * If modules haven't loaded yet, it should return `undefined`.
    * If the module doesn't exist or doesn't have any dependencies, it should return an empty array.

* The core/modules store should have a new selector `getModuleDependantNames( slug )`:
    * It should return an array of module names that are `dependants` of the module identified by the given `slug`.
    * If modules haven't loaded yet, it should return `undefined`.
    * If the module doesn't exist or doesn't have any dependants, it should return an empty array.

## Implementation Brief

* In  `assets/js/googlesitekit/modules/datastore/modules.js`, add the new `getModuleDependencyNames` selector to the `baseSelectors` object.
* The selector should leverage `createRegistrySelector`  and accept `slug` as a parameter ( see `isModuleActive` as an example )
* Retrieve the module via `geModule(slug)`
	* If the returned module is not loaded ( `undefined` ), return `undefined`
	* If the module is not found( `null` ) return an empty array
	* If there are `dependencies`, return an array of module names based on the slugs in the array, looking up the module by its slug in the store.
* 
* In  `assets/js/googlesitekit/modules/datastore/modules.js`, add the new `getModuleDependantNames` selector to the `baseSelectors` object.
* The selector should leverage `createRegistrySelector`  and accept `slug` as a parameter ( see `isModuleActive` as an example )
* Retrieve the module via `geModule(slug)`
	* If the returned module is not loaded ( `undefined` ), return `undefined`
	* If the module is not found( `null` ) return an empty array
	* If there are `dependants`, return an array of module names based on the slugs in the array, looking up the module by its slug in the store.

### Test Coverage

* Add related tests to `modules.test.js`

### Visual Regression Changes

* 

## QA Brief

* 

## Changelog entry

* Add selectors `getModuleDependencyNames` and `getModuleDependantNames` to `core/modules` store.

felixarntz:
[...]

PR #2293, opened 2020-10-30
Enhancement/2180 get module dependency selectors

ivankruchkoff:
## Summary

Addresses issue #2180 

## Relevant technical choices
Went for a more KISS approach and less DRY. We could abstract both calls to call a function which takes a `slug` and a `key` where key would be one of dependency/dependant as they are almost identical in functionality.
Went with getModule per IB rather than getModules() and calculating everything based on that.

## Checklist

### DIFF
--- a/assets/js/googlesitekit/modules/datastore/modules.js
+++ b/assets/js/googlesitekit/modules/datastore/modules.js
@@ -416,7 +416,7 @@ const baseSelectors = {
 
 		// Module is found, return the names of the dependencies
 		const modules = select( STORE_NAME ).getModules();
-		return module.dependencies.map( ( dependencySlug ) => modules[ dependencySlug ].name );
+		return module.dependencies.map( ( dependencySlug ) => modules[ dependencySlug ].name || dependencySlug );
 	} ),
 
 	/**
@@ -445,7 +445,7 @@ const baseSelectors = {
 			return [];
 		}
 
-		// Module is found, return the names of the dependencies
+		// Module is found, return the names of the dependants
 		const modules = select( STORE_NAME ).getModules();
 		return module.dependants.map( ( dependantSlug ) => modules[ dependantSlug ].name );
 	} ),
--- a/assets/js/googlesitekit/modules/datastore/modules.js
+++ b/assets/js/googlesitekit/modules/datastore/modules.js
@@ -389,6 +389,68 @@ const baseSelectors = {
 	} ),
 
 	/**
+	 * Gets module dependency names by slug.
+	 *
+	 * Returns a list of modules that depend on this module.
+	 * Returns `undefined` if state is still loading or if said module doesn't exist.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @param {string} slug  Module slug.
+	 * @return {(Array|undefined)} An array of dependency module names; `undefined` if state is still loading.
+	 */
+	getModuleDependencyNames: createRegistrySelector( ( select ) => ( state, slug ) => {
+		const module = select( STORE_NAME ).getModule( slug );
+
+		// Return `undefined` if module with this slug isn't loaded yet.
+		if ( module === undefined ) {
+			return undefined;
+		}
+
+		// A module with this slug couldn't be found; return `[]` to signify the
+		// "not found" state.
+		if ( module === null ) {
+			return [];
+		}
+
+		// Module is found, return the names of the dependencies
+		const modules = select( STORE_NAME ).getModules();
+		return module.dependencies.map( ( dependencySlug ) => modules[ dependencySlug ].name );
+	} ),
+
+	/**
+	 * Gets module dependant names by slug.
+	 *
+	 * Returns a list of modules on which this module depends.
+	 * Returns `undefined` if state is still loading or if said module doesn't exist.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @param {string} slug  Module slug.
+	 * @return {(Array|undefined)} An array of dependant module names; `undefined` if state is still loading.
+	 */
+	getModuleDependantNames: createRegistrySelector( ( select ) => ( state, slug ) => {
+		const module = select( STORE_NAME ).getModule( slug );
+
+		// Return `undefined` if module with this slug isn't loaded yet.
+		if ( module === undefined ) {
+			return undefined;
+		}
+
+		// A module with this slug couldn't be found; return `[]` to signify the
+		// "not found" state.
+		if ( module === null ) {
+			return [];
+		}
+
+		// Module is found, return the names of the dependencies
+		const modules = select( STORE_NAME ).getModules();
+		return module.dependants.map( ( dependantSlug ) => modules[ dependantSlug ].name );
+	} ),
+
+	/**
 	 * Checks a module's activation status.
 	 *
 	 * Returns `true` if the module exists and is active.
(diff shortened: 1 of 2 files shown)


assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -388,6 +388,68 @@ const baseSelectors = {
...
+
+		// A module with this slug couldn't be found; return `[]` to signify the
+		// "not found" state.
+		if ( module === null ) {
+			return [];
+		}
+
+		// Module is found, return the names of the dependencies
+		const modules = select( STORE_NAME ).getModules();
+		return module.dependencies.map( ( dependencySlug ) => modules[ dependencySlug ].name );
```
eugene-manuilov, line 419: We need to filter out unavailable modules before mapping it to make sure we don't try to get a name from `undefined`. 

Plus, I think we need to use the module slug if the module name is empty: `modules[ dependencySlug ].name || dependencySlug`. It's not needed for the current modules, but since we are going to allow 3rd party developers to register their own modules, there might be a case when a module is registered with an empty name.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -388,6 +388,68 @@ const baseSelectors = {
...
+		}
+
+		// A module with this slug couldn't be found; return `[]` to signify the
+		// "not found" state.
+		if ( module === null ) {
+			return [];
+		}
+
+		// Module is found, return the names of the dependencies
+		const modules = select( STORE_NAME ).getModules();
```
eugene-manuilov, line 418: Returned modules might be `undefined` which means it is not loaded yet. We need to return undefined in this case:
```suggestion
		const modules = select( STORE_NAME ).getModules();
		if ( modules === undefined ) {
			return undefined;
		}
```

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -388,6 +388,68 @@ const baseSelectors = {
...
```
eugene-manuilov, line 449: Same here, let's return undefined if modules aren't loaded yet.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -388,6 +388,68 @@ const baseSelectors = {
...
+		return module.dependants.map( ( dependantSlug ) => modules[ dependantSlug ].name );
```
eugene-manuilov, line 450: Same here, let's filter out unavailable modules (`modules[ dependantSlug ] === undefined`) and use slugs as a fallback if the name is empty.

assets/js/googlesitekit/modules/datastore/modules.test.js
```diff
@@ -543,6 +543,102 @@ describe( 'core/modules modules', () => {
 			} );
 		} );
 
+		describe( 'getModuleDependencyNames', () => {
```
eugene-manuilov, line 546: @ivankruchkoff let's combine `getModuleDependencyNames` and `getModuleDependantNames` into one [`describe.each`](https://jestjs.io/docs/en/api#describeeachtablename-fn-timeout) block:

```js
describe.each( [
	[ 'getModuleDependencyNames', 'dependencies' ],
	[ 'getModuleDependantNames', 'dependants' ],
] )( '%s', ( selector, collectionName) => {
	...

	it( 'returns dependency module names when modules are loaded', async () => {
		fetchMock.getOnce(
			/^\/google-site-kit\/v1\/core\/modules\/data\/list/,
			{ body: FIXTURES, status: 200 }
		);
		const slug = 'optimize';
		registry.select( STORE_NAME )[ selector ]( slug );

		// Wait for loading to complete.
		await untilResolved( registry, STORE_NAME ).getModules();

		const namesLoaded = registry.select( STORE_NAME )[ selector ]( slug );

		expect( fetchMock ).toHaveFetchedTimes( 1 );
		expect( namesLoaded ).toMatchObject( fixturesKeyValue[ slug ][ collectionName ].map( ( moduleSlug ) => fixturesKeyValue[ moduleSlug ].name ) );
	} );

	...
} );
```

eugene-manuilov (changes requested):
@ivankruchkoff it looks good to me. Found a few minor issues and added a small suggestion. Could you please take a look at it?

assets/js/googlesitekit/modules/datastore/modules.js
ivankruchkoff: Modules can't be undefined at this point due to https://github.com/google/site-kit-wp/blob/fdc6a94a3951234c3b836f685d9ec65fcf37f8bf/assets/js/googlesitekit/modules/datastore/modules.js#L378

assets/js/googlesitekit/modules/datastore/modules.js
ivankruchkoff: I don't think module properties can be undefined per https://github.com/google/site-kit-wp/blob/fdc6a94a3951234c3b836f685d9ec65fcf37f8bf/assets/js/googlesitekit/modules/datastore/modules.js#L71

assets/js/googlesitekit/modules/datastore/modules.js
ivankruchkoff: We do (line 439), per above comment ( https://github.com/google/site-kit-wp/blob/fdc6a94a3951234c3b836f685d9ec65fcf37f8bf/assets/js/googlesitekit/modules/datastore/modules.js#L374 )

assets/js/googlesitekit/modules/datastore/modules.test.js
ivankruchkoff: Combined.

assets/js/googlesitekit/modules/datastore/modules.js
eugene-manuilov: Ah... yes, you are right. Maybe we should add a comment which says that modules are already resolved at this point, shouldn't we?

assets/js/googlesitekit/modules/datastore/modules.test.js
eugene-manuilov: Awesome, thanks!

assets/js/googlesitekit/modules/datastore/modules.js
eugene-manuilov: @ivankruchkoff i mean a case when a 3rd part plugin registers its own module with dependencies like this: `[ 'analytics', 'foo', 'bar' ]`. Then, your code will find the Analytics module and will throw an error for `foo` because `modules['foo']` will result in `undefined`.

assets/js/googlesitekit/modules/datastore/modules.js
ivankruchkoff: Addressed.

eugene-manuilov: Awesome! Thanks!

assets/js/googlesitekit/modules/datastore/modules.js
eugene-manuilov: Awesome! Thanks.

assets/js/googlesitekit/modules/datastore/modules.js
ivankruchkoff: Addressed.

assets/js/googlesitekit/modules/datastore/modules.js
ivankruchkoff: Addressed.

eugene-manuilov (approved):
Awesome! LGTM!

felixarntz (approved):
Great!

Issue #2295, opened 2020-10-30
Update Composer dependencies to their latest versions


felixarntz:
We haven't updated Composer dependencies in a while, so let's do that again. Particularly the Google API libraries should be kept up to date.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All Composer dependencies, primarily the Google API libraries, should be up to date.

## Implementation Brief

* Add `google/apiclient-services` to our `composer.json` explicitly, since we actually require it directly. It's more than a dependency of `google/apiclient`.
* Run `composer update`.
* Test some basic behavior of the plugin and fix potential failures, e.g. if Google API classes are no longer loading.
* Merge #2296.

### Test Coverage

* N/A

### Visual Regression Changes

* N/A

## QA Brief

* This should mostly be covered implicitly by regular smoke testing.
* However, let's do some extra testing to QA this issue specifically. We should visit all of the following areas and make sure they don't result in a blank page or some other (fairly obvious) error:
    * The Site Kit dashboard
    * The Site Kit dashboard details view for a single URL
    * The Site Kit widget in the WordPress dashboard
    * The Site Kit admin bar menu in the frontend
    * All 4 module setup flows (AdSense, Analytics, Optimize, Tag Manager)

## Changelog entry

* N/A

aaemnnosttv:
IB ✅

tofumatt:
QA ✅

PR #2296, opened 2020-10-30
Update/2295 composer dependencies

felixarntz:
## Summary

Addresses issue #2295

## Relevant technical choices

* The Google API Client now uses PSR-4 filenames instead of PSR-0. Backward-compatibility is maintained via class aliases, see https://github.com/googleapis/google-api-php-client/releases/tag/v2.8.0. We don't need to update our class names used (especially because they're even used by the Google API Services), but in order to make the aliases work, another manual PHP Scoper replacement is necessary for the `aliases.php` file.
* The WordPress VIP Coding Standards now rely on an external package for variable analysis (see https://github.com/Automattic/VIP-Coding-Standards/releases/tag/2.2.0), so our PHPCS configuration that referenced it needed to be updated.

## Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,6 +18,7 @@
     "ext-json": "*",
     "ext-openssl": "*",
     "google/apiclient": "^2.0",
+    "google/apiclient-services": "~0.1",
     "guzzlehttp/guzzle": "^5.3.3"
   },
   "replace": {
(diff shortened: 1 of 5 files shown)


aaemnnosttv (approved):
LGTM 👍

PR #2300, opened 2020-10-30
Bug/1995 fix race conditions in the analytics datastore.

eugene-manuilov:
## Summary

Addresses issue #1995

## Checklist

### DIFF
--- a/assets/js/modules/analytics/datastore/accounts.js
+++ b/assets/js/modules/analytics/datastore/accounts.js
@@ -121,10 +121,13 @@ const baseActions = {
 		invariant( isValidAccountSelection( accountID ), 'A valid accountID is required to select.' );
 
 		const registry = yield Data.commonActions.getRegistry();
-		registry.dispatch( STORE_NAME ).setProfileID( '' );
-		registry.dispatch( STORE_NAME ).setInternalWebPropertyID( '' );
-		registry.dispatch( STORE_NAME ).setPropertyID( '' );
-		registry.dispatch( STORE_NAME ).setAccountID( accountID );
+
+		registry.dispatch( STORE_NAME ).setSettings( {
+			accountID,
+			internalWebPropertyID: '',
+			propertyID: '',
+			profileID: '',
+		} );
 
 		if ( ACCOUNT_CREATE === accountID ) {
 			return;
--- a/assets/js/modules/analytics/datastore/accounts.js
+++ b/assets/js/modules/analytics/datastore/accounts.js
@@ -121,10 +121,10 @@ const baseActions = {
 		invariant( isValidAccountSelection( accountID ), 'A valid accountID is required to select.' );
 
 		const registry = yield Data.commonActions.getRegistry();
-		registry.dispatch( STORE_NAME ).setAccountID( accountID );
-		registry.dispatch( STORE_NAME ).setPropertyID( '' );
-		registry.dispatch( STORE_NAME ).setInternalWebPropertyID( '' );
 		registry.dispatch( STORE_NAME ).setProfileID( '' );
+		registry.dispatch( STORE_NAME ).setInternalWebPropertyID( '' );
+		registry.dispatch( STORE_NAME ).setPropertyID( '' );
+		registry.dispatch( STORE_NAME ).setAccountID( accountID );
 
 		if ( ACCOUNT_CREATE === accountID ) {
 			return;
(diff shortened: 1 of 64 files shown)


assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -121,10 +121,10 @@ const baseActions = {
 		invariant( isValidAccountSelection( accountID ), 'A valid accountID is required to select.' );
 
 		const registry = yield Data.commonActions.getRegistry();
-		registry.dispatch( STORE_NAME ).setAccountID( accountID );
-		registry.dispatch( STORE_NAME ).setPropertyID( '' );
-		registry.dispatch( STORE_NAME ).setInternalWebPropertyID( '' );
 		registry.dispatch( STORE_NAME ).setProfileID( '' );
+		registry.dispatch( STORE_NAME ).setInternalWebPropertyID( '' );
+		registry.dispatch( STORE_NAME ).setPropertyID( '' );
+		registry.dispatch( STORE_NAME ).setAccountID( accountID );
```
felixarntz, line 127: It might help to add a clarifying comment here on that this order matters and why. Alternatively, this might be even a bit more error-proof by using the `setSettings` action and passing an object (with `accountID`, `propertyID`, `internalWebPropertyID`) to set all three values at the same time.

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -151,6 +151,8 @@ const baseActions = {
 				return;
 			}
 
+			// Clear any profile ID selection in the case that selection falls to the getProfiles resolver.
+			registry.dispatch( STORE_NAME ).setProfileID( '' );
 			registry.dispatch( STORE_NAME ).setPropertyID( propertyID );
 
 			if ( PROPERTY_CREATE === propertyID ) {
```
felixarntz, line 158: That's unrelated, but shouldn't there also be a call to `setInternalWebPropertyID`, making it empty, for this case? Otherwise, when selecting to create a new property, any previous internal web property ID will still be in state, potentially causing problems.

felixarntz (commented):
Overall LGTM, just two points of feedback.

assets/js/modules/analytics/datastore/accounts.js
eugene-manuilov: Yep, `setSettings` is better. I have updated it to use that action.

assets/js/modules/analytics/datastore/properties.js
eugene-manuilov: Yes, it was a bit confusing. I have updated it to be more properly organized. Does it look good to you?

felixarntz: This looks better, setting it all at the same time. However, my one point above about the `internalWebProperty` not being set here could still cause problems IMO. But that's not actually in the scope of this issue, so it's okay.

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -151,10 +151,12 @@ const baseActions = {
 				return;
 			}
 
-			registry.dispatch( STORE_NAME ).setPropertyID( propertyID );
-
 			if ( PROPERTY_CREATE === propertyID ) {
-				registry.dispatch( STORE_NAME ).setProfileID( PROFILE_CREATE );
+				registry.dispatch( STORE_NAME ).setSettings( {
+					propertyID,
+					profileID: PROFILE_CREATE,
```
felixarntz, line 157: I think it would make sense to also set `internalWebPropertyID` to `''` here. Otherwise, it might keep its previous value (let's say a property was selected before, but then the user chooses "Create new property"), but since it always should depend on `propertyID`, that would be incorrect.

felixarntz (approved):
LGTM for solving this issue!

assets/js/modules/analytics/datastore/properties.js
eugene-manuilov: Yes, you are right. I have added internalWebPropertyID here as well.

felixarntz:
@eugene-manuilov There's a failing Jest test here that may be unexpected, assigning back to you.

eugene-manuilov:
@felixarntz looks like my changes in the `properties.js` aren't necessary and cause those test errors. I have reverted it and left just the required changes. It's ready for review now.

felixarntz (approved):
Great!

Issue #2210, opened 2020-10-20
`receiveError` and `clearError` should require `args`


felixarntz:
## Bug Description

Following up on #1814, #1859, and #2033, there appears to be one thing that was previously missed or not considered: Currently it is valid to call the `receiveError` and `clearError` actions with a `baseName` parameter while not passing an `args` parameter. This should **not** be allowed because:
* `receiveError` and `clearError` apply to one **specific** error. Only passing `baseName` doesn't identify a specific error (which is different for `clearErrors`, where only passing `baseName` is expected because it will clear all errors from that `baseName`).
* Either `args` needs to be required, or at least `args` should default to `[]`, to make cases where the selector or action doesn't even support any arguments more intuitive. However I think it makes sense to be explicit and always require `args` (for parameter-less actions/selectors an empty array needs to be passed).

Note that the legacy behavior, where `receiveError` and `clearError` also work without `baseName` and `args` entirely, should not be affected by this change - we'll need to maintain this until all legacy usage has been phased out (not part of this issue).

Related: #2174 (which would have been avoided, had we implemented this in the first place)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The partial store returned from `createErrorStore` needs to be updated:
    * `receiveError` should either require only `error` or it should require all three parameters. In other words, if `baseName` is provided, `args` must be provided as well (and it must be an array).
    * `clearError` should either require no parameters or it should require all two parameters. In other words, if `baseName` is provided, `args` must be provided as well (and it must be an array).
* The tests that check required parameters etc should be adjusted accordingly.
* All usages of `receiveError` and `clearError` must be reviewed to ensure that they are:
    * either legacy usage (neither passing `baseName` nor `args`), which should not be updated here
    * or passing both `baseName` and `args`

## Implementation Brief

* Add a check for `args` in `receiveError` after https://github.com/google/site-kit-wp/blob/af48c90e7eb72a700fd15e951491731f08758c4c/assets/js/googlesitekit/data/create-error-store.js#L47:

```js
if ( baseName ) {
    invariant(
        args && Array.isArray( args ),
        'args is required (and must be an array) when baseName is specified.'
    );
}
```

* The same check should be added to https://github.com/google/site-kit-wp/blob/af48c90e7eb72a700fd15e951491731f08758c4c/assets/js/googlesitekit/data/create-error-store.js#L65
[...]

PR #2307, opened 2020-11-02
Bug/2210 errors should require args

eugene-manuilov:
## Summary

Addresses issue #2210

## Checklist

### DIFF
--- a/assets/js/googlesitekit/data/create-settings-store.js
+++ b/assets/js/googlesitekit/data/create-settings-store.js
@@ -169,13 +169,17 @@ export const createSettingsStore = ( type, identifier, datapoint, {
 		 */
 		*saveSettings() {
 			const registry = yield Data.commonActions.getRegistry();
-			registry.dispatch( STORE_NAME ).clearError( 'saveSettings', [] );
+
+			const { clearError, receiveError } = registry.dispatch( STORE_NAME );
+			if ( clearError ) {
+				clearError( 'saveSettings', [] );
+			}
 
 			const values = registry.select( STORE_NAME ).getSettings();
 			const { response, error } = yield fetchSaveSettingsStore.actions.fetchSaveSettings( values );
-			if ( error ) {
+			if ( error && receiveError ) {
 				// Store error manually since saveSettings signature differs from fetchSaveSettings.
-				registry.dispatch( STORE_NAME ).receiveError( error, 'saveSettings', [] );
+				receiveError( error, 'saveSettings', [] );
 			}
 
 			return { response, error };
--- a/assets/js/googlesitekit/data/create-settings-store.js
+++ b/assets/js/googlesitekit/data/create-settings-store.js
@@ -169,12 +169,13 @@ export const createSettingsStore = ( type, identifier, datapoint, {
 		 */
 		*saveSettings() {
 			const registry = yield Data.commonActions.getRegistry();
-			const values = registry.select( STORE_NAME ).getSettings();
+			registry.dispatch( STORE_NAME ).clearError( 'saveSettings', [] );
 
+			const values = registry.select( STORE_NAME ).getSettings();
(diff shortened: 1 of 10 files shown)


assets/js/googlesitekit/data/create-settings-store.js
```diff
@@ -169,12 +169,13 @@ export const createSettingsStore = ( type, identifier, datapoint, {
...
 			const registry = yield Data.commonActions.getRegistry();
-			const values = registry.select( STORE_NAME ).getSettings();
+			registry.dispatch( STORE_NAME ).clearError( 'saveSettings', [] );
 
+			const values = registry.select( STORE_NAME ).getSettings();
 			const { response, error } = yield fetchSaveSettingsStore.actions.fetchSaveSettings( values );
 			if ( error ) {
 				// Store error manually since saveSettings signature differs from fetchSaveSettings.
-				registry.dispatch( STORE_NAME ).receiveError( error, 'saveSettings' );
+				registry.dispatch( STORE_NAME ).receiveError( error, 'saveSettings', [] );
```
felixarntz, line 178: There's a problem here which is unrelated to this issue, but was missed previously when introducing the dynamic error functionality: `receiveError` here should only be called if it's available on the store.

We can assume that that is the case in all other areas where we defined specific partial stores, but this here is a reusable function, and there could be a store that calls `createSettingsStore` but does not call `createErrorStore` for the same store - then this would fail. So we should first check if that action exists on the store and only call it then. Or we could do it the same way as in https://github.com/google/site-kit-wp/blob/1.19.0/assets/js/googlesitekit/data/create-fetch-store.js#L43, that way we don't need to do a check.

assets/js/googlesitekit/data/create-settings-store.js
```diff
@@ -169,12 +169,13 @@ export const createSettingsStore = ( type, identifier, datapoint, {
 		 */
 		*saveSettings() {
 			const registry = yield Data.commonActions.getRegistry();
-			const values = registry.select( STORE_NAME ).getSettings();
+			registry.dispatch( STORE_NAME ).clearError( 'saveSettings', [] );
```
felixarntz, line 172: See above comment, we should only call this if the store has it.

assets/js/googlesitekit/data/create-settings-store.test.js
```diff
@@ -52,7 +54,12 @@ describe( 'createSettingsStore store', () => {
 			settingSlugs: [ 'isSkyBlue' ],
 			registry,
 		} );
-		registry.registerStore( storeDefinition.STORE_NAME, storeDefinition );
+
+		registry.registerStore( storeDefinition.STORE_NAME, Data.combineStores(
+			storeDefinition,
+			createErrorStore(),
+		) );
+
```
felixarntz, line 62: I think this change can be reverted after my above comments are addressed. Potentially we could add an extra Jest test though to ensure that the error store actions are called _if_ the store has the error partial store.

felixarntz (changes requested):
This is overall LGTM, however there's a problem overlooked earlier, which I think makes sense to fix as part of this.

assets/js/googlesitekit/data/create-settings-store.test.js
eugene-manuilov: Yep, we can just revert it now. Updated.

assets/js/googlesitekit/data/create-settings-store.js
eugene-manuilov: Updated.

assets/js/googlesitekit/data/create-settings-store.js
eugene-manuilov: Updated.

felixarntz (approved):
Great!

Issue #2311, opened 2020-11-03
Delete site from authentication service upon uninstallation


felixarntz:
When the plugin is uninstalled, it should delete the site from the authentication service (via an available endpoint there). Note that this is only possible if the site is currently connected (has site ID and site secret).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* An uninstall hook should be added that, when the plugin is being uninstalled, issues a `POST` request to the authentication service proxy URL with path `/o/oauth2/delete-site/`.
    * The request should only be sent if the site is currently connected.
    * The request should include URL-encoded body parameters `site_id` and `site_secret`.

**Note that this endpoint is not yet live. Before this gets tested and merged, it should be confirmed that it has been published.**

## Implementation Brief

* Add a public method `unregister_site()` to `Google\Site_Kit\Core\Authentication\Google_Proxy`. (#2044 should add the respective counterpart `register_site()` to that class.) This method should send a `POST` request to Site Kit Service as described in the ACs.
* Add an `uninstall.php` file which, if the `WP_UNINSTALL_PLUGIN` constant is there, simply loads the main plugin file to bootstrap the plugin, and after that triggers a parameter-less action `googlesitekit_uninstallation`.
* Implement class `Google\Site_Kit\Core\Util\Uninstallation`:
    * Implement a private `uninstall()` method which calls the new `Google_Proxy::unregister_site()` method. Wrap it in a `try catch` statement, just to ensure it won't cause a fatal error blocking the user from uninstalling.
    * Implement a public `register()` method which hooks the private `uninstall()` method into `googlesitekit_uninstallation` using a closure.
* Instantiate the new `Uninstallation` class in the root of `Plugin::register` (so that it runs "immediately"), and call its `register` method.
* Merge #2317.

### Test Coverage

* 

### Visual Regression Changes

* 

### Plugin side testing
* Install [Log HTTP Requests](https://wordpress.org/plugins/log-http-requests/)
    - Log found under (wp admin) > _Tools_ > _Log HTTP Requests_
* Install Site Kit and go through the setup flow.
* Verify that deleting the plugin triggers a request to the authentication service's `/o/oauth2/delete-site/` route  
_Currently only implemented on the staging proxy – so if not using this, a 404 is expected, otherwise it should be a 200 response_

### Proxy side testing
* Verify that, after setting up the plugin, deleting it marks the site as deleted on the authentication service.
[...]

PR #2317, opened 2020-11-04
Unregister site on proxy on uninstallation of plugin

felixarntz:
## Summary

Addresses issue #2311

## Checklist

### DIFF
--- a/includes/Core/Util/Uninstallation.php
+++ b/includes/Core/Util/Uninstallation.php
@@ -1,0 +1,99 @@
+<?php
+/**
+ * Class Google\Site_Kit\Core\Util\Uninstallation
+ *
+ * @package   Google\Site_Kit\Core\Util
+ * @copyright 2020 Google LLC
+ * @license   https://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
+ * @link      https://sitekit.withgoogle.com
+ */
+
+namespace Google\Site_Kit\Core\Util;
+
+use Google\Site_Kit\Context;
+use Google\Site_Kit\Core\Storage\Options;
+use Google\Site_Kit\Core\Storage\Encrypted_Options;
+use Google\Site_Kit\Core\Authentication\Credentials;
+use Google\Site_Kit\Core\Authentication\Google_Proxy;
+
+/**
+ * Utility class for handling uninstallation of the plugin.
+ *
+ * @since n.e.x.t
+ * @access private
+ * @ignore
+ */
+class Uninstallation {
+
+	/**
+	 * Plugin context.
+	 *
+	 * @since n.e.x.t
+	 * @var Context
+	 */
+	private $context;
+
+	/**
+	 * Options object.
+	 *
+	 * @since n.e.x.t
+	 * @var Options
+	 */
+	private $options = null;
+
+	/**
+	 * Constructor.
+	 *
+	 * This class and its logic must be instantiated early in the WordPress
+	 * bootstrap lifecycle because an uninstallation hook runs decoupled from
+	 * regular action hooks like 'init'.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Context $context Plugin context.
+	 * @param Options $options Optional. Option API instance. Default is a new instance.
+	 */
+	public function __construct(
+		Context $context,
+		Options $options = null
+	) {
+		$this->context = $context;
+		$this->options = $options ?: new Options( $this->context );
+	}
+
+	/**
+	 * Registers functionality through WordPress hooks.
+	 *
+	 * @since n.e.x.t
+	 */
+	public function register() {
+		add_action(
+			'googlesitekit_uninstallation',
+			function() {
+				$this->uninstall();
+			}
+		);
+	}
+
+	/**
+	 * Runs necessary logic for uninstallation of the plugin.
+	 *
+	 * If connected to the proxy, it will issue a request to unregister the site.
+	 *
+	 * @since n.e.x.t
+	 */
+	private function uninstall() {
+		$credentials = new Credentials( new Encrypted_Options( $this->options ) );
+
+		if ( ! $credentials->has() || ! $credentials->using_proxy() ) {
+			return;
+		}
+
+		$google_proxy = new Google_Proxy( $this->context );
+		try {
+			$google_proxy->unregister_site( $credentials );
+		} catch ( Exception $e ) { // phpcs:ignore Generic.CodeAnalysis.EmptyStatement.DetectedCatch
+			// Just avoid this from being thrown.
+		}
+	}
+}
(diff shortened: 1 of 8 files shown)


felixarntz:
@adamsilverstein @aaemnnosttv This can currently be tested against the staging version of the authentication proxy.

tests/phpunit/integration/Core/Util/UninstallationTest.php
```diff
@@ -0,0 +1,77 @@
...
+		add_filter( 'pre_http_request', array( $this, 'record_request_url_and_cause_error' ), 10, 3 );
+		do_action( 'googlesitekit_uninstallation' );
+
+		// Assert no HTTP request was made.
+		$this->assertNull( $this->last_request_url );
+	}
+
+	public function record_request_url_and_cause_error( $preempt, $args, $url ) {
+		$this->last_request_url = $url;
+		return new WP_Error( 'outgoing_request_blocked', 'Outgoing request blocked for testing.' );
```
aaemnnosttv, line 75: The WP_Error is unnecessary as outgoing requests are already configured to be blocked for PHPUnit as it is, so I would return `$preempt` here instead since the specific response isn't used.

tests/phpunit/integration/Core/Authentication/Google_ProxyTest.php
```diff
@@ -82,23 +88,72 @@ public function test_get_user_fields_for_network_administrator() {
...
+		$this->expect_request_response(
+			$google_proxy->url( Google_Proxy::OAUTH2_DELETE_SITE_URI ),
+			$expected_error_response,
+			400
+		);
+
+		// Ensure exception with correct message is thrown for error response.
+		try {
+			$google_proxy->unregister_site( $credentials );
+			$this->assertEquals( $expected_error_response['error'], null );
```
aaemnnosttv, line 144: Better to fail explicitly here if this line isn't expected to be reached.
```suggestion
			$this->fail( 'Expected an exception to be thrown when unregistering the site' );
```

aaemnnosttv (commented):
Overall looks good to me. Left a few comments but nothing major.

adamsilverstein:
@felixarntz I tested this and it worked well, I could see that once the plugin was deleted from within the wp-admin UI, it was marked uninstalled in the Site Kit service ("proxy"). 

One somewhat related issue I noticed is that when I deleted the plugin (and WordPress warned me "Are you sure you want to delete Site Kit and it's Data?"), my site Kit settings were not actually reset. 
 When I re-installed the plugin and activated, my configuration was still active which seemed unexpected.

tests/phpunit/integration/Core/Authentication/Google_ProxyTest.php
```diff
@@ -146,21 +201,40 @@ public function test_exchange_site_code() {
...
+
+	/**
+	 * Adds a 'pre_http_request' filter that will ensure the request for the
+	 * given URL will yield a specific response.
+	 *
+	 * @param string $request_url   Request URL to modify response for.
+	 * @param array  $response_data Response data to return for the request. Will be JSON-encoded.
+	 * @param int    $response_code Optional. Response status code to return. Default 200.
+	 */
+	private function expect_request_response( $request_url, array $response_data, $response_code = 200 ) {
```
aaemnnosttv, line 224: The name of this method is a little confusing as it reads a bit like an assertion when in fact it's setting up a mocked response for a request. Something like `mock_http_request` would be a bit more clear I think, but something else without `expect` would work too.

It would be nice to add something like this to the base test case at some point as well but that needs a bit more thought still.

tests/phpunit/integration/Core/Util/UninstallationTest.php
```diff
@@ -0,0 +1,77 @@
...
+		$this->fake_site_connection();
+		add_filter( 'pre_http_request', array( $this, 'record_request_url_and_cause_error' ), 10, 3 );
+		do_action( 'googlesitekit_uninstallation' );
+
+		// Assert no HTTP request was made.
+		$this->assertNull( $this->last_request_url );
+	}
+
+	public function record_request_url_and_cause_error( $preempt, $args, $url ) {
+		$this->last_request_url = $url;
```
aaemnnosttv, line 74: Rather than recording the last URL, how about it records a boolean as to whether or not the `OAUTH2_DELETE_SITE_URI` URL was requested? That way it isn't dependent on being the last request recorded and the assertions can be simply assert `true|false`.

uninstall.php
```diff
@@ -0,0 +1,20 @@
...
+ * @package   Google\Site_Kit
+ * @copyright 2020 Google LLC
+ * @license   https://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
+ * @link      https://sitekit.withgoogle.com
+ */
+
+// Prevent execution from directly accessing the file.
+if ( ! defined( 'WP_UNINSTALL_PLUGIN' ) ) {
+	exit;
+}
```
aaemnnosttv, line 14: Good call! 👍

includes/Core/Util/Uninstallation.php
```diff
@@ -0,0 +1,100 @@
...
+	/**
+	 * Plugin context.
+	 *
+	 * @since n.e.x.t
+	 * @var Context
+	 */
+	private $context;
+
+	/**
+	 * Options object.
```
aaemnnosttv, line 38: Super nitpick but this is how it's defined most places.
```suggestion
	 * Options instance.
```

includes/Core/Util/Uninstallation.php
```diff
@@ -0,0 +1,100 @@
...
+	 *
+	 * @since n.e.x.t
+	 * @var Options
+	 */
+	private $options = null;
```
aaemnnosttv, line 43:
```suggestion
	private $options;
```

includes/Core/Util/Uninstallation.php
```diff
@@ -0,0 +1,100 @@
...
+
+	/**
+	 * Constructor.
+	 *
+	 * This class and its logic must be instantiated early in the WordPress
+	 * bootstrap lifecycle because an uninstallation hook runs decoupled from
+	 * regular action hooks like 'init'.
```
aaemnnosttv, line 50: This is mostly still true, but may be slightly inaccurate as the uninstall is no longer based on a WP uninstall hook.

includes/Core/Util/Uninstallation.php
```diff
@@ -0,0 +1,100 @@
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Context $context Plugin context.
+	 * @param Options $options Optional. Option API instance. Default is a new instance.
```
aaemnnosttv, line 55:
```suggestion
	 * @param Options $options Optional. Options instance. Default is a new instance.
```

aaemnnosttv (commented):
This looks pretty good @felixarntz – nothing really blocking here.

I left a few comments mostly around the tests with a few suggested changes to make, as well as a few nitpicks around the documentation, but again – nothing critical.

aaemnnosttv (approved):
Super, thanks!

Issue #2245, opened 2020-10-22
Introduce utility functions `sumObjectListValue` and `averageObjectListValue`


felixarntz:
There should be new utility functions `sumObjectListValue` and `averageObjectListValue` that iterate over an array of objects (or alternatively an array of arrays) and calculate the sum or average respectively over one value in those objects.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a function `sumObjectListValue( list, fieldName )` in `assets/js/util/sum-object-list-value.js`:
    * It should iterate over the array `list`. It should be expected that each value in list is an object or array.
    * For every item `item` in `list` it should add `item[ fieldName ]` (if set) to a variable.
    * It should return the resulting numeric value.
* There should be a function `averageObjectListValue( list, fieldName )` in `assets/js/util/average-object-list-value.js`:
    * It should iterate over the array `list`. It should be expected that each value in list is an object or array.
    * For every item `item` in `list` it should add `item[ fieldName ]` (if set) to a variable.
    * Afterwards it should divide the sum by the number of items in `list`.
    * It should return the resulting numeric value. Based on the values aggregated, if they are integers, the return value should also be an integer, i.e. rounded so that it doesn't have any decimal point. Otherwise, it should be returned as floating point value as it is.
* Move `extractForSparkline` into `assets/js/util/extract-for-sparkline.js`, and add an optional `indexColumn` parameter which defaults to `0`:
    * Use `row[ indexColumn ]` instead of `row[ 0 ]`.
    * If `column` contains dots, split the string and traverse along `row` to that property/key (e.g. if `column` is `metrics.0.values.0`, look at `row[ 'metrics' ][ 0 ][ 'values' ][ 0 ]`).
    * If `indexColumn` contains dots, split the string and traverse along `row` to that property/key (e.g. if `indexColumn` is `keys.0`, look at `row[ 'keys' ][ 0 ]`).
* The Search Console `DashboardClicksWidget` and `DashboardImpressionsWidget` components should be updated to use the new functions instead of relying on the legacy function `extractSearchConsoleDashboardData`:
    * Split `data` into first half (main values) and second half (values to compare to).
    * Call `sumObjectList` for both halves with `clicks` or `impressions` respectively for `fieldName`.
    * Then call `changeToPercent` to compare both totals, similarly to how it's currently done in the legacy function `extractSearchConsoleDashboardData`.
[...]

PR #2325, opened 2020-11-04
Enhancement/2245 introduce new util functions

ryanwelcher:
## Summary

Addresses issue #2245 

## Checklist


assets/js/util/test/extractForSparkline.js
```diff
@@ -114,8 +114,128 @@ const valuesToTest = [
...
+						'616',
+					],
+				},
+			],
+		},
+	];
+
+	it( 'returns data from an index passed as a string', () => {
+		expect( extractForSparkline( data, 0, 'metrics.0.values.0' ) ).toStrictEqual( [ [ '1469', '' ], [ '616', 0 ] ] );
+		expect( extractForSparkline( data, 1, 'dimensions.1' ) ).toStrictEqual( [ [ '/', '' ], [ '/site-kit-tests/', 0 ] ] );
```
aaemnnosttv, line 239: The `0` and `1` here as the second argument to `extractForSparkline` is the path to the y-axis value. In this case, neither exists which results in an odd looking result.

The y value from the first pair is an empty string because the first row is treated differently but we can ignore that for now. The y value from the second pair is `0` for both again because neither row has a property of `0` so the fallback value is `0` for both.

Let's update the data used to have a more generic structure and update the test show the relationship between the data and the path arguments more clearly.

e.g.
```js
const data = [
	{ key: 'foo', value: 123 },
	{ key: 'bar', value: 456 },
]
extractForSparkline( data, 'value', 'key' ) // [ [ 'foo', 123 ], [ 'bar', 456 ] ]
```
Another test to show that object paths `a.b.c.` are supported for both arguments as well would be great

assets/js/util/test/extractForSparkline.js
```diff
@@ -114,8 +114,128 @@ const valuesToTest = [
...
+];
+
 describe( 'extractForSparkline', () => {
 	it.each( valuesToTest )( 'for start date %s and end date %s should returns %s', ( data, column, expected ) => {
 		expect( extractForSparkline( data, column ) ).toStrictEqual( expected );
 	} );
+
+	it.each( columnIndexValuesToTest )( 'for start date %s and columnIndex %s it should return %s', ( data, columnIndex, expected ) => {
+		expect( extractForSparkline( data, 0, columnIndex ) ).toStrictEqual( expected );
+	} );
```
aaemnnosttv, line 207:
```suggestion
	} );

```

assets/js/modules/search-console/components/dashboard/DashboardImpressionsWidget.js
```diff
@@ -88,14 +89,35 @@ function DashboardImpressionsWidget() {
 		return getNoDataComponent( _x( 'Search Console', 'Service name', 'google-site-kit' ) );
 	}
 
-	const { totalImpressions, totalImpressionsChange, dataMap } = extractSearchConsoleDashboardData( data );
+	// const { totalImpressions, totalImpressionsChange, dataMap } = extractSearchConsoleDashboardData( data );
```
aaemnnosttv, line 92: Please clean this up 🙂

assets/js/modules/search-console/components/dashboard/DashboardClicksWidget.js
```diff
@@ -87,15 +88,32 @@ function DashboardClicksWidget() {
...
 
+	// Sparkline data needs headers and dates formatted as MM/DD
+	const sparklineData = [
+		[
+			{ type: 'string', label: 'Day' },
+			{ type: 'number', label: 'Clicks' },
+		],
+		...extractForSparkline( latestData, 'clicks', 'keys.0' ).map( ( row ) => {
+			const date = new Date( row[ 0 ] );
+			return [ date.getMonth() + 1 + '/' + date.getUTCDate(), row[ 1 ] ];
```
aaemnnosttv, line 108: Let's move the comment closer to what it's referring to (applies elsewhere too)
```suggestion
			// Sparkline data needs headers and dates formatted as MM/DD
			return [ date.getMonth() + 1 + '/' + date.getUTCDate(), row[ 1 ] ];
```

assets/js/modules/search-console/components/dashboard/DashboardImpressionsWidget.js
```diff
@@ -88,14 +89,35 @@ function DashboardImpressionsWidget() {
...
+	// Split the data in two chunks.
+	const half = Math.floor( data.length / 2 );
+	const latestData = data.slice( half );
+	const olderData = data.slice( 0, half );
+
+	const totalImpressions = sumObjectListValue( latestData, 'impressions' );
+	const totalOlderImpressions = sumObjectListValue( olderData, 'impressions' );
+	const totalImpressionsChange = changeToPercent( totalOlderImpressions, totalImpressions );
+
+	// Sparkline data needs headers and dates formatted as MM/DD
```
aaemnnosttv, line 103: See above

aaemnnosttv (changes requested):
Thanks @ryanwelcher – this is looking good overall, just a few things to clean up and tweak in a few places. Let me know if you have any questions 👍

assets/js/modules/search-console/components/dashboard/DashboardImpressionsWidget.js
ryanwelcher: I did this and didn't push the commit. 🤦

assets/js/modules/search-console/components/dashboard/DashboardClicksWidget.js
```diff
@@ -87,15 +88,33 @@ function DashboardClicksWidget() {
...
+	const half = Math.floor( data.length / 2 );
+	const latestData = data.slice( half );
+	const olderData = data.slice( 0, half );
 
-	const { totalClicks, totalClicksChange, dataMap } = extractSearchConsoleDashboardData( data );
+	const totalClicks = sumObjectListValue( latestData, 'clicks' );
+	const totalOlderClicks = sumObjectListValue( olderData, 'clicks' );
+	const totalClicksChange = changeToPercent( totalOlderClicks, totalClicks );
 
+	// Sparkline data needs headers and dates formatted as MM/DD
```
aaemnnosttv, line 100: We can remove this now that it has been moved below as requested before.
```suggestion
```

aaemnnosttv (commented):
LGTM, just the one tiny thing missed from the last review and a merge conflict to resolve now 😄

aaemnnosttv:
@googlebot I consent.

assets/js/util/average-object-list-value.js
```diff
@@ -0,0 +1,52 @@
...
+ * @param {Array}  list      Array of objects or arrays.
+ * @param {string} fieldName The name of the field to be averaged.
+ *
+ * @return {number} The average.
+ */
+export default function averageObjectListValue( list, fieldName ) {
+	// Filter the list to remove any entries that don't have fieldName so our length is correct for division.
+	const filteredList = list.filter( ( item ) => {
+		return Object.keys( item ).includes( fieldName );
+	} );
```
aaemnnosttv, line 38: The list shouldn't be filtered to a smaller set as this would affect the total length and thus the average. It's expected that every item would have the same shape in the list. If for some reason the `fieldName` isn't present we should use `0` instead.

aaemnnosttv (changes requested):
One thing to change for the added utility for averaging which is not in the ACs.

assets/js/modules/search-console/components/dashboard/DashboardClicksWidget.js
```diff
@@ -87,15 +88,33 @@ function DashboardClicksWidget() {
...
+
+	const sparklineData = [
+		[
+			{ type: 'string', label: 'Day' },
+			{ type: 'number', label: 'Clicks' },
+		],
+		...extractForSparkline( latestData, 'clicks', 'keys.0' ).map( ( row ) => {
+			const date = new Date( row[ 0 ] );
+			// Sparkline data needs headers and dates formatted as MM/DD
+			return [ date.getMonth() + 1 + '/' + date.getUTCDate(), row[ 1 ] ];
```
tofumatt, line 108: This is pretty minor but using string interpolation would better-signal what this item in the array is meant to be:

```suggestion
			return [ `${ date.getMonth() + 1 }/${ date.getUTCDate() }`, row[ 1 ] ];
```

I dunno, maybe it's just me, but I always prefer string interpolation over concat operations for readability.

assets/js/modules/search-console/components/dashboard/DashboardImpressionsWidget.js
```diff
@@ -88,14 +89,33 @@ function DashboardImpressionsWidget() {
...
+
+	const sparklineData = [
+		[
+			{ type: 'string', label: 'Day' },
+			{ type: 'number', label: 'Clicks' },
+		],
+		...extractForSparkline( latestData, 'impressions', 'keys.0' ).map( ( row ) => {
+			const date = new Date( row[ 0 ] );
+			// Sparkline data needs headers and dates formatted as MM/DD
+			return [ date.getMonth() + 1 + '/' + date.getUTCDate(), row[ 1 ] ];
```
tofumatt, line 109: Same here re: string interpolation. Let me know if you disagree, that's fine 😄 
```suggestion
			return [ `${ date.getMonth() + 1 }/${ date.getUTCDate() }`, row[ 1 ] ];
```

assets/js/util/average-object-list-value.js
```diff
@@ -0,0 +1,51 @@
...
+import mean from 'lodash/mean';
+
+/**
+ * Returns the average of a given fieldName in a list.
+ *
+ * @since n.e.x.t
+ *
+ * @param {Array}  list      Array of objects or arrays.
+ * @param {string} fieldName The path name of the field to be averaged.
+ * @return {number} The average.
```
tofumatt, line 32:
```suggestion
 * @return {number} The average (mean) value of every item in the list.
```

assets/js/util/sum-object-list-value.js
```diff
@@ -0,0 +1,35 @@
...
+import sumBy from 'lodash/sumBy';
+
+/**
+ * Returns the sum of a given fieldName in a list.
+ *
+ * @since n.e.x.t
+ *
+ * @param {Array}  list      Array of objects or arrays.
+ * @param {string} fieldName The path name to the field to be summed.
+ * @return {number} The sum.
```
tofumatt, line 31:
```suggestion
 * @return {number} The sum of all fields in the list.
```

assets/js/util/test/extractForSparkline.js
```diff
@@ -114,8 +114,119 @@ const valuesToTest = [
 	],
 ];
 
+const columnIndexValuesToTest = [
```
tofumatt, line 117: These are pretty tough to parse and understand, especially because the data is a bit far away from the test. I understand the efficiency in use `it.each` here, but we need to make sure our tests are readable and parsable in the future if we need to change them or diagnose errors. That's why I prefer more verbose tests that don't follow DRY always, because they're easier to reason about. Test code doesn't need to be DRY as much.

That said, I think this is fine to leave as-is, but could you leave comments in the data indicating where the input starts and where the result is? That way someone can look at this `columnIndexValuesToTest` variable and see what the result should be. That'll make this easier to maintain.

tofumatt (changes requested):
Looks good to me, my only comments are around readability of the test data and some documentation tweaks really.

If @aaemnnosttv's cool with it happy to make those changes myself, because this is otherwise good.

aaemnnosttv:
Those suggestions sound good to me @tofumatt -you're welcome to make those changes if you'd like 👍

tofumatt:
@googlebot I consent.

tofumatt (approved):

aaemnnosttv (approved):
👍 👍

PR #2327, opened 2020-11-05
Fix typo in uninstall.php

aaemnnosttv:
## Summary

Addresses issue #2311 

## Relevant technical choices

- Confirmed this fixes the error
- Updated QA brief on issue

## Checklist

### DIFF
--- a/uninstall.php
+++ b/uninstall.php
@@ -14,7 +14,7 @@ if ( ! defined( 'WP_UNINSTALL_PLUGIN' ) ) {
 }
 
 // Load plugin main file to bootstrap infrastructure and add hooks.
-require_once dirname( __FILE ) . '/google-site-kit.php';
+require_once dirname( __FILE__ ) . '/google-site-kit.php';
 
 // Fire action to trigger uninstallation logic.
 do_action( 'googlesitekit_uninstallation' );


uninstall.php
```diff
@@ -14,7 +14,7 @@
 }
 
 // Load plugin main file to bootstrap infrastructure and add hooks.
-require_once dirname( __FILE ) . '/google-site-kit.php';
+require_once dirname( __FILE__ ) . '/google-site-kit.php';
```
eugene-manuilov, line 17: @aaemnnosttv maybe we should use just `__DIR__`?
```suggestion
require_once __DIR__ . '/google-site-kit.php';
```

aaemnnosttv: Yeah, the result there is the same and a little cleaner so generally I would opt for that as well, but given that this needs to have maximum compatibility with older PHP versions (just like `google-site-kit.php`) I think it's marginally better to keep the older style here. Given that I think `__DIR__` was introduced in PHP 5.3, it's not really a big concern, but I'm guessing this is probably what @felixarntz was thinking here.

eugene-manuilov: Ok, not a big deal though. We can use `dirname`.

eugene-manuilov (approved):

felixarntz (approved):
OMG 🤦‍♂️  Thanks for fixing this!

PR #2328, opened 2020-11-05
Tweaks to the user input styles to match Figma designs.

asvinb:
## Summary

Addresses issue #2040

## Relevant technical choices

Tweaks to the existing CSS to match what we have in the designs.

## Checklist


felixarntz (changes requested):
There are a few things off with the styling here in comparison with the Figma file:
* Vertical spacing between the radio buttons/checkboxes is too much. In the Figma file they are 48px tall each (20px + 2 * 14px), but this design makes them 80px tall (40px + 2 * 20px), causing a lot of extra whitespace.
* The answers column should start at the same vertical level as the question number in the left column (e.g. "1 out of 5"), but in this design they start as the same level as the actual question, resulting in extra whitespace above.
* One thing that's missing overall is limiting the screen area to only the current question and whatever is visible of the next question (this is described in the ACs): The user shouldn't be able to scroll vertically to see all questions. They should only be able to scroll far enough to see all of the current question. If the screen height is large enough, they should then be able to see "a teaser" of the next question, but not be able to scroll further.

asvinb:
@felixarntz 

1. Vertical spacing fixed. It was using the same spacing as on small screens
2. Question number alignment fixed
3. For the vertical scroll, I tried to go with a pure CSS solution, let me know if that makes sense to you. So, on small screens, only the active question is displayed since there is no reason to show the next question because the screen estate is limited. This is applicable till the desktop breakpoint. The on desktop bp (>= 960px), the next question following the current one is displayed along with a mask which fades to WordPress' gray background color. Thus the user can only see a teaser of the next question. So instead of "If the screen height is large enough, they should then be able to see "a teaser" of the next question, but not be able to scroll further.", we can change that to "On large screens, they should then be able to see "a teaser" of the next question, but not be able to scroll further." This eliminates the need to add some custom JS to check for the screen height and current question, to decide whether or not there is enough space to show the teaser of the next question. Let me know what you think.

felixarntz (approved):
Awesome - looks solid now! I like the CSS-only approach, good call on that.

Issue #2254, opened 2020-10-23
Introduce `core/site` selector `getDeveloperPluginState`


felixarntz:
There should be a new selector `getDeveloperPluginState` in the `core/site` store which returns information about whether the Site Kit developer plugin is installed etc.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `core/site` store should receive a new selector `getDeveloperPluginState`.
    * The selector should have a resolver that via `createFetchStore` calls the existing `core/site/data/developer-plugin` REST datapoint and from the response populates its data.
    * The selector should return the response data object, or `undefined` if not loaded yet.

## Implementation Brief

* Create a new `assets/js/googlesitekit/datastore/site/developer-plugin.js` file:
  * Use `createFetchStore` to create a partial store for the `core/site/data/developer-plugin` REST endpoin:
    * Use `getDeveloperPluginState` as a base name;
    * Add a control callback function that sends a GET request to the `developer-plugin` endpoint;
    * Add a reducer callback function that saves received data under the `developerPluginState` property;
    * Assign created partial store to the `fetchGetDeveloperPluginState` variable.
  * Define the new `getDeveloperPluginState` resolver that calls `fetchGetDeveloperPluginState` action of the `fetchGetDeveloperPluginState` variable if there is no developer plugin data in the store yet;
  * Define the new `getDeveloperPluginState` selector that returns a value from  the `developerPluginState` property of the site datastore state;
  * Export the new partial datastore.
* Import developer plugin partial store in the `assets/js/googlesitekit/datastore/site/index.js` file and it into `Data.combineStores()` function call.

### Test Coverage

* Create jest tests that cover the new selector functionality.

### Visual Regression Changes

* Not needed.

## QA Brief

* Check that the new selector is implemented and correctly added to the site datastore.

## Changelog entry

* N/A

felixarntz:
IB ✅

tofumatt:
QA ✅

PR #2333, opened 2020-11-06
Enhancement/2254 get developer plugin state selector

ryanwelcher:
## Summary

Addresses issue #2254 

## Checklist


assets/js/googlesitekit/datastore/site/developer-plugin.test.js
```diff
@@ -0,0 +1,151 @@
...
+
+		describe( 'receiveGetDeveloperPluginState', () => {
+			it( 'requires the response param', () => {
+				expect( () => {
+					registry.dispatch( STORE_NAME ).receiveGetDeveloperPluginState();
+				} ).toThrow( 'response is required.' );
+			} );
+
+			it( 'receives and sets developer plugin state', async () => {
+				await registry.dispatch( STORE_NAME ).receiveGetDeveloperPluginState( responseDeveloperPluginState );
```
aaemnnosttv, line 74: Receive actions are synchronous so the `await` is unnecessary here.
```suggestion
			it( 'receives and sets developer plugin state', () => {
				registry.dispatch( STORE_NAME ).receiveGetDeveloperPluginState( responseDeveloperPluginState );
```

assets/js/googlesitekit/datastore/site/developer-plugin.test.js
```diff
@@ -0,0 +1,151 @@
...
+				);
+
+				const initialDeveloperPluginState = registry.select( STORE_NAME ).getDeveloperPluginState();
+				// The developer plugin state will be its initial value while the developer plugin state is fetched.
+				expect( initialDeveloperPluginState ).toEqual( undefined );
+				await subscribeUntil( registry,
+					() => (
+						registry.select( STORE_NAME ).getDeveloperPluginState() !== undefined
+					),
+				);
```
aaemnnosttv, line 98: Please use `untilResolved` instead.

assets/js/googlesitekit/datastore/site/developer-plugin.test.js
```diff
@@ -0,0 +1,151 @@
...
+
+			it( 'does not make a network request if data is already in state', async () => {
+				registry.dispatch( STORE_NAME ).receiveGetDeveloperPluginState( responseDeveloperPluginState, {} );
+
+				const developerPluginState = registry.select( STORE_NAME ).getDeveloperPluginState();
+
+				await subscribeUntil( registry, () => registry
+					.select( STORE_NAME )
+					.hasFinishedResolution( 'getDeveloperPluginState' )
+				);
```
aaemnnosttv, line 119: Please use `untilResolved` instead 🙂

assets/js/googlesitekit/datastore/site/developer-plugin.test.js
```diff
@@ -0,0 +1,151 @@
...
+
+				const developerPluginState = registry.select( STORE_NAME ).getDeveloperPluginState();
+
+				expect( fetchMock ).toHaveFetchedTimes( 1 );
+				expect( developerPluginState ).toEqual( responseDeveloperPluginState );
+
+				const developerPluginStateSelect = registry.select( STORE_NAME ).getDeveloperPluginState();
+				expect( fetchMock ).toHaveFetchedTimes( 1 );
+
+				expect( developerPluginStateSelect ).toEqual( developerPluginState );
```
aaemnnosttv, line 108: These lines seem to be mostly testing the same thing as the following test "does not make a network request if data is already in state" and can be removed.

assets/js/googlesitekit/datastore/site/developer-plugin.test.js
```diff
@@ -0,0 +1,151 @@
...
+					/^\/google-site-kit\/v1\/core\/site\/data\/developer-plugin/,
+					{ body: response, status: 500 }
+				);
+
+				registry.select( STORE_NAME ).getDeveloperPluginState();
+				await subscribeUntil( registry,
+					// TODO: We may want a selector for this, but for now this is fine
+					// because it's internal-only.
+					() => registry.select( STORE_NAME ).isFetchingGetDeveloperPluginState() === false,
+				);
```
aaemnnosttv, line 141: Please use `untilResolved` instead 🙂

assets/js/googlesitekit/datastore/site/developer-plugin.test.js
```diff
@@ -0,0 +1,151 @@
...
+} from 'tests/js/utils';
+import { STORE_NAME } from './constants';
+
+describe( 'core/site developer plugin state', () => {
+	const responseDeveloperPluginState = {
+		active: true,
+		installed: true,
+		activateURL: 'http://something.test',
+		installURL: 'http://something.test',
+		configureURL: 'http://something.test',
```
aaemnnosttv, line 37: Although it doesn't make a difference as currently used, let's make these 3 URLs different since they would never be the same and this could result in an assertion passing when it shouldn't.
Also, super-nitpick but please use `example.com` in tests unless otherwise needed.
 ```suggestion
		activateURL: 'http://example.com/activate',
		installURL: 'http://example.com/install',
		configureURL: 'http://example.com/configure',
```

**Review by aaemnnosttv — CHANGES_REQUESTED** [review]
Thanks @ryanwelcher – this looks great, just a few quick changes to address regarding the tests; overall very nice though 👌

File: assets/js/googlesitekit/datastore/site/developer-plugin.test.js (hunk 5 above)
Comment (ryanwelcher) [reply assets/js/googlesitekit/datastore/site/developer-plugin.test.js:141]: @aaemnnosttv using `untilResolved` here fired `isFetchingGetDeveloperPluginState is not a function` errors.

Comment (aaemnnosttv) [reply assets/js/googlesitekit/datastore/site/developer-plugin.test.js:131]: `untilResolved` is specific to selectors with resolvers. The above `isFetchingGetDeveloperPluginState` is a selector without a resolver which we don't need to use here. We only need to wait for `getDeveloperPluginState` to be resolved just like in the other tests.
```suggestion
				await untilResolved( registry, STORE_NAME ).getDeveloperPluginState();
```

aaemnnosttv:
@googlebot I consent.

aaemnnosttv (approved):
Super!

Issue #2130, opened 2020-10-05
Implement API to check module activation requirements


felixarntz:
Site Kit modules may have certain requirements that need to be fulfilled before they can be activated. For example, AdSense requires that no AdBlocker is currently enabled. There should be a general way to centrally handle these requirement checks.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

### API
* When registering a module in JavaScript, it should be possible to provide a function that checks whether all requirements for using the module are fulfilled. The default should be a function that returns `true`.
    * The function should be able to return an error object with what specifically is missing.
* There should be a selector on the `core/modules` store that returns whether a given module can be activated, based on the new module registration argument. This function should return a boolean.
    * Before calling the registration argument function, it should first check whether all the module's dependencies are activated. The error message in such a case should be `You need to set up %1$s to gain access to %2$s` (translatable of course).
* There should be another selector on the `core/modules` store that returns the specific error why a given module cannot be activated, or `null` otherwise.

### UI / legacy component migration
* In the UI in Site Kit Settings where a module can be activated, the new datastore selectors should be relied upon.
    * The `googlesitekit.SetupModuleShowLink` filter should be removed.
    * Instead of `blockedByParentModule`, the button should be disabled solely based on the return value from the new `core/modules` selector, which per the above includes the dependency logic.
    * The button link should always remain as `Set up %s`, even when it's disabled. Instead, the potential error message should be displayed instead of the `ModuleSettingsWarning` component, with similar UI as the current only usage (AdSense module's `AdBlockerWarning` component) should be.
* Similar as above, all usages of `ModuleSettingsWarning` should be replaced with relying on the new selector and displaying an error from there. The `ModuleSettingsWarning` component should be removed entirely.
* The `SetupModule` component should be refactored to a functional component and the file should be renamed to match the component name. It should then be moved into `assets/js/components/settings`.

### Usage update
(The only module currently using this is AdSense, so this is all that needs to be migrated.)
* The AdSense module should make use of the new registration argument to register a function that returns an error when an AdBlocker is detected. The error message should be `Ad blocker detected, you need to disable it in order to set up AdSense.`
* The AdSense module's usage of the `googlesitekit.SetupModuleShowLink` and `googlesitekit.ModuleSettingsWarning` filters should be removed, as it is now covered by the new module registration argument above.

### API
* Add function argument called `checkRequirements` to the `registerModule` action with a default value `() => true`.
	* This function should return an error object or true.
[...]

PR #2335, opened 2020-11-09
Enhancement/2130 module activation requirements api

ivankruchkoff:
## Summary

Addresses issue #2130

## Relevant technical choices
Resolver runs once and is cached for any set of args. Modules `canActivateModule` resolver needs `getModule` / `getModules` to have already resolved prior to completion. Latest version of `data.select` uses resolveSelect ( see https://developer.wordpress.org/block-editor/packages/packages-data-controls/#select and https://github.com/WordPress/gutenberg/pull/17558 ) 
To workaround this, we make our resolver call with a second parameter that we use for "cache busting" purposes which allows us to re-run the resolver.
We have an issue to update `@wordpress/data` in #1769 which will avoid this workaround.

## Checklist


assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -293,6 +338,30 @@ const baseReducer = ( state, { type, payload } ) => {
 			};
 		}
 
+		case RECEIVE_CHECK_REQUIREMENTS_ERROR: {
+			const checkRequirementsResults = { ...state.checkRequirementsResults };
+
+			for ( const [ slug, error ] of Object.entries( payload ) ) {
+				checkRequirementsResults[ slug ] = error;
+			}
```
felixarntz, line 346: See above, I don't think this action needs to be able to receive multiple errors - we only need to store a single error per module.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -309,6 +378,54 @@ const baseResolvers = {
...
+		// A module with this slug couldn't be found return null
+		if ( module === null ) {
+			return null;
+		}
+
+		const inactiveModules = [];
+
+		module.dependencies.forEach( ( dependencySlug ) => {
+			const dependedentModule = registry.select( STORE_NAME ).getModule( dependencySlug );
+			if ( ! dependedentModule.active ) {
```
felixarntz, line 405:
```suggestion
			if ( ! dependedentModule?.active ) {
```

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -309,6 +378,54 @@ const baseResolvers = {
...
+
+		// If we have inactive dependencies, there's no need to check if we can
+		// activate the module until the dependencies have been activated.
+		if ( inactiveModules.length ) {
+			const formatter = new Intl.ListFormat( getLocale(), { style: 'long', type: 'conjunction' } );
+
+			/* translators: Error message text. 1: A flattened list of module names. 2: A module name. */
+			const errorMessage = sprintf( __( 'You need to set up %1$s to gain access to %2$s.', 'google-site-kit' ), formatter.format( inactiveModules ), module.name );
+
+			yield baseActions.receiveCheckRequirementsError( { [ slug ]: errorMessage } );
```
felixarntz, line 418: Apologies that this wasn't obvious from the ACs, but when we talk about "errors" here, this should be error **objects** like we use elsewhere in the plugin: Every error should be an object with (at minimum) `code` and `message` properties.

Also based on my above comment, let's change usage here to pass `slug` and `error` separately.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -545,6 +662,21 @@ const baseSelectors = {
 		// update.
 		return state.isAwaitingModulesRefresh;
 	} ),
+
+	canActivateModule( state, slug ) {
+		const moduleRequirements = state.checkRequirementsResults?.[slug];
```
felixarntz, line 667: I don't think this is necessary here:
```suggestion
		const moduleRequirements = state.checkRequirementsResults[ slug ];
```

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -545,6 +662,21 @@ const baseSelectors = {
 		// update.
 		return state.isAwaitingModulesRefresh;
 	} ),
+
+	canActivateModule( state, slug ) {
```
felixarntz, line 666: Selectors like this should be documented.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -545,6 +662,21 @@ const baseSelectors = {
...
+		if ( moduleRequirements === undefined ) {
+			return undefined;
+		}
+
+		return moduleRequirements === true;
+	},
+
+	getCheckRequirementsStatus( state, slug ) {
+		invariant( slug, 'slug is required.' );
+		return state.checkRequirementsResults[ slug ];
```
felixarntz, line 678: This should only be returned if it is an error object. If it's not set, `undefined` should be returned, or if it's `true`, `null` should be returned (there is no error).

Maybe it would be better to name this selector `getCheckRequirementsError` to make its purpose more clear?

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -309,6 +378,54 @@ const baseResolvers = {
...
+			const errorMessage = sprintf( __( 'You need to set up %1$s to gain access to %2$s.', 'google-site-kit' ), formatter.format( inactiveModules ), module.name );
+
+			yield baseActions.receiveCheckRequirementsError( { [ slug ]: errorMessage } );
+		} else {
+			const checkResult = yield Data.commonActions.await( module.checkRequirements() );
+
+			if ( checkResult === true ) {
+				yield baseActions.receiveCheckRequirementsSuccess( slug );
+			} else {
+				yield baseActions.receiveCheckRequirementsError( checkResult );
```
felixarntz, line 425: If not `true`, `checkResult` should be expected to be an error object instead of a map that contains the module slug (which would be redundant to provide anyway). In other words, in combination with my other comment above this line would need to be as follows:
```suggestion
				yield baseActions.receiveCheckRequirementsError( slug, checkResult );
```
(where `checkResult` is an `error`)

Via https://github.com/google/site-kit-wp/issues/2130#issuecomment-724225862, this should be modified to rely on errors being thrown instead of returned.

assets/js/modules/adsense/index.js
```diff
@@ -84,6 +75,12 @@ domReady( () => {
 			settingsEditComponent: SettingsEdit,
 			settingsViewComponent: SettingsView,
 			setupComponent: SetupMain,
+			checkRequirements: () => {
+				const isAdBlockerActive = select( STORE_NAME ).isAdBlockerActive();
+				const errorMessage = __( 'Ad blocker detected, you need to disable it in order to set up AdSense.', 'google-site-kit' );
+
+				return ! isAdBlockerActive ? true : { adsense: errorMessage };
```
felixarntz, line 82: See above, this should return `true` or an error object.

assets/js/components/settings/SetupModule.js
```diff
@@ -0,0 +1,140 @@
...
+	};
+
+	// @TODO: Resolver only runs once per set of args, so we are working around
+	// this to rerun after modules are loaded.
+	// Once #1769 is resolved, we can remove the call to getModules,
+	// and remove the !! modules cache busting param.
+	const modules = useSelect( ( select ) => select( CORE_MODULES ).getModules() );
+	const canActivateModule = useSelect( ( select ) => select( CORE_MODULES ).canActivateModule( slug, !! modules ) );
+	const requirementsStatus = useSelect( ( select ) => select( CORE_MODULES ).getCheckRequirementsStatus( slug, !! modules ) );
+	const errorMessage = canActivateModule ? null : requirementsStatus;
```
felixarntz, line 87: Theoretically it's sufficient here to only call `getCheckRequirementsStatus` (or `getCheckRequirementsError`) as it only returns a truthy value if that value is an error object. So the return value should be `null` anyway if `canActivateModule` is `true`.

assets/js/components/settings/SetupModule.js
```diff
@@ -0,0 +1,140 @@
...
+			</h3>
+			<p className="googlesitekit-settings-connect-module__text">
+				{ description }
+			</p>
+
+			{ errorMessage &&
+				<div
+					className={ classnames( 'googlesitekit-settings-module-warning', 'googlesitekit-settings-module-warning--modules-list' ) } >
+					<ErrorIcon height="20" width="23" /> { errorMessage }
+				</div>
```
felixarntz, line 118: Let's instead put this markup into `ModuleSettingsWarning` - also that component can contain the call to `getCheckRequirementsStatus` based on the received `slug`. This way it's more self-contained, and it would return `null` if there is no error.

felixarntz (changes requested):

assets/js/components/settings/SetupModule.js
ivankruchkoff: I think that's the intention, however the way the IB is written, there is no resolver for `getCheckRequirementsStatus` selector. I'm guessing we want to use the same resolver as in `canActivateModule` so that we can call either of the functions independently of the other. Currently to `getCheckRequirementsStatus` we need to `canActivateModule` first.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -258,6 +267,42 @@ const baseActions = {
...
+
+	/**
+	 * Receives the check requirements error map for specified modules modules.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} errorMap Map of errors of the format { slug: error }.
+	 * @return {Object} Action for RECEIVE_CHECK_REQUIREMENTS_ERROR.
+	 */
+	receiveCheckRequirementsError( errorMap ) {
```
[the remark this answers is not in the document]
ivankruchkoff: Updated.

assets/js/googlesitekit/modules/datastore/modules.js
ivankruchkoff: Updated.

assets/js/modules/adsense/index.js
ivankruchkoff: Updated.

assets/js/components/settings/SetupModule.js
ivankruchkoff: Updated, I also added the commented second param `context`, as I have a feeling that will impact how we FE render the message.

assets/js/googlesitekit/modules/datastore/modules.js
ivankruchkoff: Updated.

assets/js/components/settings/SetupModule.js
ivankruchkoff: I've updated this to reuse the resolver in https://github.com/google/site-kit-wp/pull/2335/commits/4f7fa81b2283e8f53ee7ad8da73760e7d0e5aa59

ivankruchkoff: With the updated call to the resolver I could update the logic to just call getCheckRequirementsStatus.

assets/js/googlesitekit/modules/datastore/modules.js
ivankruchkoff: Updated.

felixarntz (changes requested):
There are still a few things off here, particularly we should always use WP core-like objects for errors like we do elsewhere, i.e. they need to be objects with `code` and `message` properties (plus optional properties like `data`).

There are a couple other areas here where things need to be cleaned up, like old behavior, plus missing documentation.

assets/js/components/notifications/module-settings-warning.js
```diff
@@ -1,16 +1,35 @@
...
+const { useSelect } = Data;
+
+/*
  * A single module. Keeps track of its own active state and settings.
  */
-class ModuleSettingsWarning extends Component {
-	render() {
+export default function ModuleSettingsWarning( {
+	slug,
+	// context,
```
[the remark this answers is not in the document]
ivankruchkoff: Updated and removed. I assumed we'd need context param for some styling / classes.

assets/js/components/settings/SetupModule.js
```diff
@@ -0,0 +1,133 @@
...
+					googlesitekit-subheading-1
+					googlesitekit-settings-connect-module__title
+				">
+				{ name }
+			</h3>
+			<p className="googlesitekit-settings-connect-module__text">
+				{ description }
+			</p>
+
+			<ModuleSettingsWarning slug={ slug } context="modules-list" />
```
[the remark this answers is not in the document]
ivankruchkoff: Removed.

stories/settings.stories.js
```diff
@@ -91,8 +92,11 @@ storiesOf( 'Settings', module )
 		global._googlesitekitLegacyData.canAdsRun = true;
 		global._googlesitekitLegacyData.modules.analytics.setupComplete = false;
 		global._googlesitekitLegacyData.modules.adsense.active = false;
+		const setupRegistry = ( { dispatch } ) => {
+			dispatch( CORE_MODULES ).registerModule( 'adsense' )
+		};
```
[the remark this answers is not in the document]
ivankruchkoff: Resolves one test failure in https://github.com/google/site-kit-wp/actions/runs/357287610 as adsense wouldn't pass canActivateModule in https://github.com/google/site-kit-wp/pull/2335/files/2cf936a644ebd1f18295c73ce2753625542c8849#diff-7700ce4c015949d94e5e350df03d24bccd10e05eb0ebbe3922ec9dd72d406affR92 I don't think we need all the args for that particular test scenario.

assets/js/components/notifications/module-settings-warning.js
```diff
@@ -1,16 +1,35 @@
...
+} ) {
+	// @TODO: Resolver only runs once per set of args, so we are working around
+	// this to rerun after modules are loaded.
+	// Once #1769 is resolved, we can remove the call to getModules,
+	// and remove the !! modules cache busting param.
+	const modules = useSelect( ( select ) => select( CORE_MODULES )?.getModules() );
+	const requirementsStatus = useSelect( ( select ) => select( CORE_MODULES )?.getCheckRequirementsStatus( slug, !! modules ) );
+
+	if ( [ null, undefined ].includes( requirementsStatus ) ) {
```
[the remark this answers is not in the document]
ivankruchkoff: Updated.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -293,6 +341,29 @@ const baseReducer = ( state, { type, payload } ) => {
...
+		case RECEIVE_CHECK_REQUIREMENTS_ERROR: {
+			const checkRequirementsResults = { ...state.checkRequirementsResults };
+
+			const { slug, error } = payload;
+			checkRequirementsResults[ slug ] = error;
+
+			return {
+				...state,
+				checkRequirementsResults,
+			};
```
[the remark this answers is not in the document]
ivankruchkoff: Updated.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -309,7 +380,56 @@ const baseResolvers = {
...
+
+		// If we have inactive dependencies, there's no need to check if we can
+		// activate the module until the dependencies have been activated.
+		if ( inactiveModules.length ) {
+			const formatter = new Intl.ListFormat( getLocale(), { style: 'long', type: 'conjunction' } );
+
+			/* translators: Error message text. 1: A flattened list of module names. 2: A module name. */
+			const errorMessage = sprintf( __( 'You need to set up %1$s to gain access to %2$s.', 'google-site-kit' ), formatter.format( inactiveModules ), module.name );
+
+			yield baseActions.receiveCheckRequirementsError( slug, errorMessage );
```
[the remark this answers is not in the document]
ivankruchkoff: Updated.

assets/js/modules/adsense/index.js
```diff
@@ -84,6 +75,14 @@ domReady( () => {
 			settingsEditComponent: SettingsEdit,
 			settingsViewComponent: SettingsView,
 			setupComponent: SetupMain,
+			checkRequirements: () => {
+				const isAdBlockerActive = select( STORE_NAME ).isAdBlockerActive();
+				if ( ! isAdBlockerActive ) {
+					return;
+				}
+
+				throw __( 'Ad blocker detected, you need to disable it in order to set up AdSense.', 'google-site-kit' );
```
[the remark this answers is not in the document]
ivankruchkoff: Updated.

assets/js/components/notifications/module-settings-warning.js
```diff
@@ -1,16 +1,35 @@
...
 		return null;
 	}
-}
 
-export default withFilters( 'googlesitekit.ModuleSettingsWarning' )( ModuleSettingsWarning );
+	return (
+		<div
+			className={ classnames( 'googlesitekit-settings-module-warning', 'googlesitekit-settings-module-warning--modules-list' ) } >
+			<ErrorIcon height="20" width="23" /> { requirementsStatus }
```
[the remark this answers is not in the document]
ivankruchkoff: Updated.

assets/js/components/notifications/module-settings-warning.js
```diff
@@ -1,16 +1,35 @@
...
+export default function ModuleSettingsWarning( {
+	slug,
+	// context,
+} ) {
+	// @TODO: Resolver only runs once per set of args, so we are working around
+	// this to rerun after modules are loaded.
+	// Once #1769 is resolved, we can remove the call to getModules,
+	// and remove the !! modules cache busting param.
+	const modules = useSelect( ( select ) => select( CORE_MODULES )?.getModules() );
+	const requirementsStatus = useSelect( ( select ) => select( CORE_MODULES )?.getCheckRequirementsStatus( slug, !! modules ) );
```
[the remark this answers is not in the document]
ivankruchkoff: How about this for an alternate approach: https://github.com/google/site-kit-wp/pull/2335/commits/ae94aa2cb24dd5fe5c948d862c21557139ab2039#diff-51079bbb133e20fa794c0e1fd5fd30b551dda818e0cfe0ba4162502670cf133aR382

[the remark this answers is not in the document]
aaemnnosttv: @felixarntz @ivankruchkoff there is a bit of an odd situation here that we haven't run into yet where we essentially need to wait for the module to be registered which can currently happen at any time. For our internal modules, these are all registered on `domReady` and that seems to work just fine to run before all the datastore stuff runs that depends on it like this. This may be something we can improve in a separate issue, but no changes to this are needed here – we just need to wait for modules to be loaded.

While some resolver invalidation may be necessary (depending on what happens in the resolver), the important thing to add here is a `resolveSelect` call to `getModules` for both `canActivateModule` and `getCheckRequirementsStatus` resolvers (they currently share the same resolver) but that way we can essentially wait for `getModule(s)` to return actual data and all modules should be registered at this point. This currently requires that we would use the new `await` utility action with `__experimentalResolveSelect` but this should be fine since this API is now stable in newer versions of `@wordpress/data`.

assets/js/components/settings/SetupModule.js
```diff
@@ -0,0 +1,133 @@
...
+			setIsSaving( false );
+		}
+	};
+
+	// @TODO: Resolver only runs once per set of args, so we are working around
+	// this to rerun after modules are loaded.
+	// Once #1769 is resolved, we can remove the call to getModules,
+	// and remove the !! modules cache busting param.
+	const modules = useSelect( ( select ) => select( CORE_MODULES )?.getModules() );
+	const canActivateModule = useSelect( ( select ) => select( CORE_MODULES )?.canActivateModule( slug, !! modules ) );
```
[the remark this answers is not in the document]
ivankruchkoff: Updated.

assets/js/components/notifications/module-settings-warning.js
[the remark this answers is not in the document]
ivankruchkoff: Added the call to `__experimentalResolveSelect` in https://github.com/google/site-kit-wp/pull/2335/commits/dd9f2e780e3d5e1df3c35f124af89cb83fdc8f47

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -258,6 +267,45 @@ const baseActions = {
...
 	},
+
+	/**
+	 * Receives the check requirements error map for specified modules modules.
+	 *
+	 * @since n.e.x.t
+	 * @private
+	 *
+	 * @param {string} slug  Module slug.
+	 * @param {string} error Error message.
```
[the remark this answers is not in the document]
felixarntz: Still needs to be addressed.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -258,6 +267,45 @@ const baseActions = {
...
+	 * @return {Object} Action for RECEIVE_CHECK_REQUIREMENTS_ERROR.
+	 */
+	receiveCheckRequirementsError( slug, error ) {
+		invariant( slug, 'slug is required' );
+		invariant( error, 'error is required' );
```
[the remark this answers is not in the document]
felixarntz: Still needs to be addressed.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -309,7 +380,56 @@ const baseResolvers = {
...
+
+		// Return `undefined` if module with this slug isn't loaded yet.
+		if ( module === undefined ) {
+			return undefined;
+		}
+
+		// A module with this slug couldn't be found return null
+		if ( module === null ) {
+			return null;
+		}
```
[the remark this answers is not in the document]
felixarntz: Still needs to be addressed.

assets/js/googlesitekit/modules/datastore/modules.js
felixarntz: Still needs to be addressed.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -545,6 +665,22 @@ const baseSelectors = {
...
+		const moduleRequirements = state.checkRequirementsResults?.[slug];
+
+		if ( moduleRequirements === undefined ) {
+			return undefined;
+		}
+
+		return moduleRequirements === true;
+	},
+
+	getCheckRequirementsStatus( state, slug ) {
```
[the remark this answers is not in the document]
felixarntz: Still needs to be addressed.

assets/js/modules/adsense/index.js
```diff
@@ -84,6 +76,18 @@ domReady( () => {
...
+			checkRequirements: () => {
+				const isAdBlockerActive = select( STORE_NAME ).isAdBlockerActive();
+				if ( ! isAdBlockerActive ) {
+					return;
+				}
+
+				throw {
+					code: ERROR_CODE_ADBLOCKER_ACTIVE,
+					message: __( 'Ad blocker detected, you need to disable it in order to set up AdSense.', 'google-site-kit' ),
+					data: null,
```
felixarntz, line 88: Nit-pick, but the `data` property could also be omitted when `null`. It only matters when it contains something meaningful. We intentionally don't validate any specific schema (except possibly `code` and `message`) for these since they come from the WP REST API and `data` could be anything.

assets/js/util/errors.js
```diff
@@ -24,6 +24,8 @@ import isPlainObject from 'lodash/isPlainObject';
 // Export codes and reasons.
 export const ERROR_CODE_MISSING_REQUIRED_SCOPE = 'missing_required_scopes'; // When scopes are missing.
 export const ERROR_REASON_INSUFFICIENT_PERMISSIONS = 'insufficientPermissions';
+export const ERROR_CODE_INSUFFICIENT_MODULE_DEPENDENCIES = 'insufficient_module_dependencies';
+export const ERROR_CODE_ADBLOCKER_ACTIVE = 'adblocker_active';
```
felixarntz, line 28: I think it'd be more appropriate to define these in the areas where they're used. The first one could go into `assets/js/googlesitekit/modules/datastore/constants.js`, the latter into a new `assets/js/modules/adsense/constants.js`?

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -307,9 +379,60 @@ const baseResolvers = {
 
 		if ( ! existingModules ) {
 			yield fetchGetModulesStore.actions.fetchGetModules();
+			yield registry.dispatch( STORE_NAME ).invalidateResolutionForStoreSelector( 'canActivateModule' );
+			yield registry.dispatch( STORE_NAME ).invalidateResolutionForStoreSelector( 'getCheckRequirementsStatus' );
```
felixarntz, line 383: Following up on the previous comments, these invalidations shouldn't happen here for `getModules`. They need to happen when a new module is registered in JS, which is the action that _may_ change the outcome of these selectors by registering the `checkRequirements` function.

In other words, we need to move this to `registerModule`. In `registerModule`, we can make it more specific too, to avoid invalidating "irrelevant" resolvers: We only need to invalidate resolutions for the selector for the specifically registered module, i.e. `yield registry.dispatch( STORE_NAME ).invalidateResolution( 'canActivateModule', [ slug ] );` and `yield registry.dispatch( STORE_NAME ).invalidateResolution( 'getCheckRequirementsStatus', [ slug ] );`.

felixarntz (changes requested):
@ivankruchkoff This mostly looks good now functionally, just what we discussed earlier this morning needs to be updated.

Also there are several outstanding comments from previous reviews that haven't been addressed. In the future, please address those comments before sending back to code review, _unless_ that code depends on some outstanding discussion to be resolved (which I don't think is the case here though). In case you don't address specific CR feedback, please leave a comment on why.

assets/js/modules/adsense/index.js
ivankruchkoff: I was going to leave data off, except we have this util function that implies that it's a necessary param. https://github.com/google/site-kit-wp/blob/develop/assets/js/util/errors.js#L40 Unless we're not considering this error obj to be a WPError.

ivankruchkoff:
@felixarntz I was looking at the PR view ( https://github.com/google/site-kit-wp/pull/2335 ) which didn't show some of the comments you left in your review. Taking a look now.
Screen Shot 2020-11-16 at 9 03 53 PM

assets/js/googlesitekit/modules/datastore/modules.js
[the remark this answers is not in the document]
ivankruchkoff: Addressed, the previous comment doesn't surface in the PR view.

assets/js/googlesitekit/modules/datastore/modules.js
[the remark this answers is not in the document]
ivankruchkoff: Addressed, that feedback didn't surface in the PR view.

assets/js/googlesitekit/modules/datastore/modules.js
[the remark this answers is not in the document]
ivankruchkoff: Addressed.

assets/js/googlesitekit/modules/datastore/modules.js
ivankruchkoff: Addressed.

assets/js/googlesitekit/modules/datastore/modules.js
ivankruchkoff: Addressed.

assets/js/googlesitekit/modules/datastore/modules.js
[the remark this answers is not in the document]
ivankruchkoff: Renamed and commented.

assets/js/util/errors.js
ivankruchkoff: That was my initial thought, but saw that there was already an established approach and didn't want to deviate from it.

ivankruchkoff: Addressed.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -309,7 +380,56 @@ const baseResolvers = {
...
 	},
+
+	*canActivateModule( slug ) {
+		const registry = yield Data.commonActions.getRegistry();
+		const module = registry.select( STORE_NAME ).getModule( slug );
+
+		// We use yield dataControls.resolveSelect in favour of registry.select
+		// to avoid a race condition where the module hasn't loaded yet.
+		// @TODO after #1769, we can do this.
+		// const module = yield dataControls.resolveSelect( STORE_NAME, 'getModule', slug );
```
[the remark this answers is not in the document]
ivankruchkoff: Resolved, except this file is the _'core/modules'_ store.

assets/js/googlesitekit/modules/datastore/modules.js
ivankruchkoff: Are we sure that we want to do this in registerModule action? Since the module is only registered (for the purpose of getModule() selector) after the reducer REGISTER_MODULE completes, it's only after that point that we should be invalidating. In the registerModule would be too early.

assets/js/modules/adsense/index.js
felixarntz: Good call, makes sense! 👍

assets/js/util/errors.js
felixarntz: That's a valid approach. I think the difference is that the error codes currently in there are more general and they're also used by the same general error codebase, while the new ones here are more specific and apply only to another specific area.

assets/js/googlesitekit/modules/datastore/modules.js
felixarntz: Good point, we cannot do it the way that action is currently implemented. However, the call needs to happen in `registerModule` because that is where a `checkRequirements` function may be registered which would alter the results. We should be able to implement that though without too much effort:
* Make `registerModule` a generator.
* `yield` the `REGISTER_MODULE` action data.
* This way we can _afterwards_ in the `registerModule` action invalidate the resolutions for these selectors and the given `slug`.

You can check the `setModuleActivation` implementation in the same file for a fairly similar approach (the only difference is that `REGISTER_MODULE` doesn't return anything).

ivankruchkoff: Unfortunately it's still too soon, I've changed it to a generator, added the resolver invalidation and added some quick test code to `getCheckRequirementsError` for the slug, with the expectation that `optimize` will give us this error:
```json
{
    "code": "insufficient_module_dependencies", 
    "message": "You need to set up Analytics to gain access to Optimize.", 
    "data": {
        "inactiveModules": [
            "Analytics"
        ]
    }
}
```

```js

		yield {
			payload: {
				settings,
				slug,
			},
			type: REGISTER_MODULE,
		};

		yield registry.dispatch( STORE_NAME ).invalidateResolution( 'canActivateModule', [ slug ] );
		yield registry.dispatch( STORE_NAME ).invalidateResolution( 'getCheckRequirementsError', [ slug ] );
		console.log( slug ); // eslint-disable-line no-console
		// eslint-disable-next-line no-undef
		console.log( googlesitekit.data.stores[ 'core/modules' ].getSelectors().getCheckRequirementsError( slug ) ); // eslint-disable-line no-console
```

However all of the errors are still undefined.

ivankruchkoff: On second thought, the likelihood of running one of those resolvers immediately is very small (or zero?) so I pushed up the generator change in https://github.com/google/site-kit-wp/pull/2335/commits/206ffa353a511394bcbcfb49016f73d4000a8ace

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -225,9 +232,10 @@ const baseActions = {
 	 * @param {WPComponent} [settings.settingsEditComponent] Optional. React component to render the settings edit panel. Default none.
 	 * @param {WPComponent} [settings.settingsViewComponent] Optional. React component to render the settings view panel. Default none.
 	 * @param {WPComponent} [settings.setupComponent]        Optional. React component to render the setup panel. Default none.
-	 * @return {Object} Redux-style action.
+	 * @param {Function}    [settings.checkRequirements]     Optional. Function to check requirements for the module. Throws a WP error object for error or returns on success.
+	 * @return {void}
```
felixarntz, line 236: `@return` should be omitted in case of `void`.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -248,16 +257,63 @@ const baseActions = {
...
+	 * @private
+	 *
+	 * @param {string} slug  Module slug.
+	 * @param {Object} error WordPress Error object containing code, message and data properties.
+	 * @return {Object} Action for RECEIVE_CHECK_REQUIREMENTS_ERROR.
+	 */
+	receiveCheckRequirementsError( slug, error ) {
+		invariant( slug, 'slug is required' );
+		invariant( error, 'error is required' );
+		invariant( isPlainObject( error ), 'error must be an object' );
```
felixarntz, line 291: Nit-pick, but we could put this into just one `invariant` call.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -248,16 +257,63 @@ const baseActions = {
...
+	 * Receives the check requirements success for a module.
+	 *
+	 * @since n.e.x.t
+	 * @private
+	 *
+	 * @param {string} slug Success for a module slug.
+	 * @return {Object} Action for RECEIVE_CHECK_REQUIREMENTS_SUCCESS.
+	 */
+	receiveCheckRequirementsSuccess( slug ) {
+		invariant( slug, 'module slug is required' );
```
felixarntz, line 308: Let's just say `slug is required` according to the variable name, like above too.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -545,6 +667,47 @@ const baseSelectors = {
...
+	 * Returns `undefined` if slug can not be found in state.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @param {string} slug  Module slug.
+	 * @return {(boolean|undefined)} Can activate module status; `undefined` if state is still loading or if no module with that slug exists.
+	 */
+	canActivateModule( state, slug ) {
+		const moduleRequirements = state.checkRequirementsResults[ slug ];
```
felixarntz, line 685: It's not strictly necessary (just since we don't do that everywhere else), but since you've added an `invariant` for requiring `slug` in the `getCheckRequirementsError` selector, I'd suggest to do that here too.

stories/settings.stories.js
```diff
@@ -91,8 +92,11 @@ storiesOf( 'Settings', module )
 		global._googlesitekitLegacyData.canAdsRun = true;
 		global._googlesitekitLegacyData.modules.analytics.setupComplete = false;
 		global._googlesitekitLegacyData.modules.adsense.active = false;
+		const setupRegistry = ( { dispatch } ) => {
+			dispatch( CORE_MODULES ).registerModule( 'adsense' )
```
felixarntz, line 96: Missing semicolon here, this makes linting fail on Travis-CI.

felixarntz (changes requested):
Just some minor last feedback, more crucial though it is that some Jest tests are failing right now, partly around the new functionality and partly because `registerModule` was modified to become a generator. So we'll need to update the failing tests to account for this change.

stories/settings.stories.js
ivankruchkoff: Addressed.

assets/js/googlesitekit/modules/datastore/modules.js
ivankruchkoff: Addressed.

assets/js/googlesitekit/modules/datastore/modules.js
ivankruchkoff: Addressed.

assets/js/googlesitekit/modules/datastore/modules.js
ivankruchkoff: Updated.

assets/js/googlesitekit/modules/datastore/modules.js
ivankruchkoff: Addressed.

ivankruchkoff: This needs to be added to #2355 as we currently have: 
```
     218:2  error  Missing JSDoc @return declaration    jsdoc/require-returns
     218:2  error  Missing JSDoc @returns for function  valid-jsdoc
```
Per https://github.com/google/site-kit-wp/blob/develop/.eslintrc.json#L31

felixarntz (approved):
Awesome work!

stories/settings.stories.js
```diff
@@ -91,8 +92,11 @@ storiesOf( 'Settings', module )
 		global._googlesitekitLegacyData.canAdsRun = true;
 		global._googlesitekitLegacyData.modules.analytics.setupComplete = false;
 		global._googlesitekitLegacyData.modules.adsense.active = false;
+		const setupRegistry = ( { dispatch } ) => {
+			dispatch( CORE_MODULES ).registerModule( 'adsense' );
+		};
 		return (
-			<WithTestRegistry>
+			<WithTestRegistry callback={ setupRegistry }>
```
felixarntz, line 99: I think we'll need to modify the `setupRegistry` implementation here to set up modules correctly in the datastore, using the `provideModules` function, passing the actual dependencies that these modules have.

The current story here right now does not reflect what the same UI actually looks like in the product - the store state here is basically the loading state, which we shouldn't use. Outside of loading, it shouldn't be possible for a module to show up with the grayed-out, disabled UI while not having a `ModuleSettingsWarning`. I believe the reason this happens is because the datastore state here isn't set up at all (except the AdSense module registration), which means that `canActivateModule` will return `undefined`, causing the area to be grayed-out - while at the same time there is no error.

felixarntz (changes requested):
Unfortunately there's still one issue here, purely around the Storybook setup. We'll need to populate the store for the affected story with a "realistic" state - right now the store is mostly empty, returning data as if everything is loading.

stories/settings.stories.js
ivankruchkoff: I've gone through and added the modules to the stories test with https://github.com/google/site-kit-wp/commit/97e937d57b49991d18b9f7f86426dce009c3b1e3#diff-49d6554e5e787508efbc6a4b5dc1910eb74664f8ad5fa2c81226b7f811b7bbd4R96 and updated screenshots for a few of the tests in https://github.com/google/site-kit-wp/commit/764fece807fb977ca13c1818b3df4f3a667922ee

ivankruchkoff: The storybook completes fine, but I also noticed that in `Notifications > Module Setup Complete`, we're getting some errors thrown as we're using `provideModules` there too, but not a corresponding `registerModule`, which means no corresponding `checkRequirements()` on the module.

aaemnnosttv: @ivankruchkoff we still need to add the default `checkRequirements` to [`moduleDefaults`](https://github.com/google/site-kit-wp/pull/2335/files#diff-51079bbb133e20fa794c0e1fd5fd30b551dda818e0cfe0ba4162502670cf133a) in `assets/js/googlesitekit/modules/datastore/modules.js`.

felixarntz (approved):
Awesome work!

I've made some final updates here:
* Use the datastore to activate the module in the refactored version of `SetupModule`.
* Get rid of some unused/unnecessary `SetupModule` props and use prop-types for it.
* Update Storybook story for "Connect more Services" to be a scenario which would also be possible during real Site Kit usage.
* Minor Storybook fixes.

Issue #2314, opened 2020-11-03
Add user's Site Kit capabilities to Site Health report.


adamsilverstein:
## Feature Description

Add user capability information to the Site Kit section of the Site Health screen in order to assist debugging issues where users reports insufficient capabilities to use Site Kit (eg. they get an error: `You don't have permissions to authenticate with Site Kit.`). 

Site Kit uses a set of [primitive capabilities](https://github.com/google/site-kit-wp/blob/782d3697679bc478ebd5ca56dc0a935df07342a5/includes/Core/Permissions/Permissions.php#L24-L33) which are mapped to WordPress core capabilities. On the site health screen, we should add a `user_capabilities` section with a line for whether the user has each capability:  
`[Capability Name]: current_user_can( [CAPABILITY] ); `

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The Site Kit section of the Site Health screen should include a "User Capabilities" entry that lists all Site Kit capabilities (the primitive capabilities declared in the `Permissions` class) and whether the current user has them or not.
* Visually this should look similar to the existing "Required scopes" list, with each capability accompanied by an emoji icon.

## Implementation Brief

* Add a new private function `get_capabilities_field` to the `Core\Util\Debug_Data` class:
  * It should return an array that contains the following elements:
    * `label` equal to `User Capabilities`
    * `value` - an associative array where each key is a capability slug defined in the `Core\Permissions\Permissions` class and value is either ✅ or ⭕ depending on whether the current user has that capability or not (see [get_required_scopes_field](https://github.com/google/site-kit-wp/blob/782d3697679bc478ebd5ca56dc0a935df07342a5/includes/Core/Util/Debug_Data.php#L310-L324) for reference).
* Add `capabilities` element to the `fields` array in the `get_fields` method. It should get a result from the `get_capabilities_field` function call.

### Test Coverage

* Not needed

### Visual Regression Changes

* Nothing should be changed on the plugin pages, but a new section should be added to the site health page.

## QA Brief

* Go to the Site Health page and open `Info` tab. You should see capabilities information in the site kit section.

## Changelog entry

* Include current user's Site Kit capabilities in Site Health report.

felixarntz:
IB ✅

wpdarren:
### QA Update: Failed  ❌ 

**Fail:**

[...]

PR #2340, opened 2020-11-09
Enhancement/2314 add to site health

ryanwelcher:
## Summary

Addresses issue #2314 

## Checklist


includes/Core/Util/Debug_Data.php
```diff
@@ -347,4 +349,26 @@ function ( Module_With_Debug_Fields $module ) {
...
+
+	/**
+	 * Gets capabilities for the current user.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array
+	 */
+	private function get_capabilities_field() {
+		$user_data = apply_filters( 'googlesitekit_user_data', array() );
```
aaemnnosttv, line 362: While this works and requires minimal changes to implement, I don't think we should couple the check here to the filter for user data.

Instead, let's extract the function in `Permissions` which is attached to that filter to a new public method that returns the map of `$permission => $current_user_can` (i.e. the result of the `array_combine` call). Then `Permissions` can be updated to call this method in its callback for the filter.

In this method, we can then simply instantiate a new `Permissions` instance and call the new method – perhaps something like `check_all_for_current_user` would be a good name. The new method should go right after the definition for `Permissions::register` since it's public. The `Debug_Data` class already has all the necessary dependencies for instantiating `Permissions` so nothing extra would be needed for that.

includes/Core/Util/Debug_Data.php
```diff
@@ -347,4 +349,26 @@ function ( Module_With_Debug_Fields $module ) {
...
+
+
+	/**
+	 * Gets capabilities for the current user.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array
+	 */
+	private function get_capabilities_field() {
```
aaemnnosttv, line 361: A bit nitpicky but please move the definition of this method to go after `get_required_scopes_field` and before `get_module_fields` for consistency with the others.

aaemnnosttv (changes requested):
Overall this looks pretty good. One change to make regarding how the permissions are retrieved and the other is purely organizational. It should be fairly straightforward to implement so please let me know if you have any questions about any of it.

includes/Core/Permissions/Permissions.php
```diff
@@ -193,6 +179,28 @@ function( array $allcaps ) {
 		);
 	}
 
+	/**
+	 * Check permissions for current user.
+	 *
```
aaemnnosttv, line 184:
```suggestion
	 *
	 * @since n.e.x.t
	 *
```

aaemnnosttv (commented):
Just the missing annotation, otherwise looks great! I noticed the VRT is failing but that seems to be unrelated, so that will probably have to be addressed separately.

aaemnnosttv (approved):
Great!

Issue #2242, opened 2020-10-22
Introduce AdSense utility function `isZeroReport` and use it in new widgets


felixarntz:
There should be a new AdSense utility function `isZeroReport`, similar to the legacy `isDataZeroAdSense` function but without some of the legacy details. The new function should be used in all refactored AdSense widgets for the respective conditional which takes care of returning the "no-data" component.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new function `isZeroReport` should be added in `assets/js/modules/adsense/util/is-zero-report.js`:
    * It should expect a `report` object like the one coming from the `modules/adsense` `getReport` selector.
    * If `report` is `undefined`, it should return `undefined` (to account for loading state).
    * Otherwise, it should return `true` if `report?.rows` is not set or empty or if `report?.totals` is not set or empty. It should also return `true` if none of the values in `report.totals` is greater than 0.
    * Otherwise it should return `false` (i.e. there _is_ data).
* The `DashboardSummaryWidget` component should use the above function instead of manual checks, for all reports `today`, `period`, `daily`.

## Implementation Brief

* Create a new `assets/js/modules/adsense/util/is-zero-report.js` file;
* Add a new `isZeroReport` function to the `is-zero-report.js` file that satisfies requirements defined in the AC section;
* Add `export * from './is-zero-report';` to the `assets/js/modules/adsense/util/index.js` file;
* Update the `DashboardSummaryWidget` component to use the new `isZeroReport` function instead of manual checks, for all reports.

### Test Coverage

* Create a new `assets/js/modules/adsense/util/is-zero-report.test.js` file;
* Add tests to check the `isZeroReport` function.

### Visual Regression Changes

* No visual changes should occur as a result of these changes.

## QA Brief

* Check the `DashboardSummaryWidget` component to make sure it uses the new `isZeroReport` function for all reports.
* There should be no visual changes on the dashboard page.

## Changelog entry

* Improve logic to determine whether an AdSense API report response is "empty", via new `isZeroReport` function for AdSense.

felixarntz:
IB ✅

caseymorrisus:
QA ✅

PR #2354, opened 2020-11-11
Enhancement/2242 is zero report

eugene-manuilov:
## Summary

Addresses issue #2242 

## Relevant technical choices

* AC says that the function should return `undefined` if `undefined` has been passed to the function, but it doesn't work well because it will bypass zero data check and cause javascript errors. So, I have updated the function to return `true` when `undefined` is passed.

## Checklist


assets/js/modules/adsense/util/is-zero-report.js
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Checks whether the report data is empty or not.
+ *
+ * @since n.e.x.t
+ *
+ * @param {Object} report Report data.
+ * @return {boolean} TRUE if the report has no data, otherwise FALSE.
+ */
+export function isZeroReport( report ) {
+	if ( typeof report === 'undefined' ) {
```
tofumatt, line 28: Very minor, but we've been tending toward `report === undefined` instead of using `typeof`. I know back in the day `typeof` was more used, but no one redefines `undefined` in ES6+ and it's a bit cleaner to read.

```suggestion
	if ( report === undefined ) {
```

assets/js/modules/adsense/util/is-zero-report.js
```diff
@@ -0,0 +1,50 @@
...
+		return true;
```
tofumatt, line 29: You mentioned this is the reason you're returning `true` instead of `undefined` here, but by doing so we can't differentiate between a report without data and a report that is not-yet-loaded.

The idea behind returning `undefined` is components could know that the report wasn't yet loaded so show a "loading" UI instead of a "zero day" UI. Since we might want to show different things for each state, I think returning `undefined` here gives us more flexibility.

Of course, that will mean adding a `report === undefined` state to the `DashboardSummaryWidget`. I've left a comment there about that, but I think it's worth doing so this check can be more flexible.

assets/js/modules/adsense/components/dashboard/DashboardSummaryWidget.js
```diff
@@ -83,7 +83,7 @@ function DashboardSummaryWidget() {
 		return getDataErrorComponent( 'adsense', error.message, false, false, false, error );
 	}
 
-	if ( ! today?.totals && ! period?.totals && ! daily?.totals ) {
+	if ( isZeroReport( today ) && isZeroReport( period ) && isZeroReport( daily ) ) {
```
tofumatt, line 86: I'm curious why `today`, `period`, and/or `daily` would be `undefined` here given `loading` indicates that they're still loading. I guess that `isResolving` selector used to load `loading` might not return `true` on first render here?

If so, I think it's best to change the `if ( loading )` check above to check for `if ( loading || today === undefined || period === undefined || daily === undefined )`.

I think it's better not to have undefined reports (eg ones that are loading) show a zero data state, as that's different to loading and could be confusing.

assets/js/modules/adsense/util/is-zero-report.test.js
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+/**
+ * Internal dependencies.
+ */
+import { isZeroReport } from './is-zero-report';
+
+describe( 'isZeroReport', () => {
+	it.each( [
+		[ 'undefined', undefined ],
```
tofumatt, line 26: As mentioned, I think this one should return `undefined`.

tofumatt (changes requested):
I think we should account for `undefined` reports in that component better and use a separate `undefined` response from `isZeroData`. Because I don't think we should be conflating a loading report with one that is empty, but that's what this ends up doing if an `undefined` report is being loaded past the `if ( loading ) {` check in `DashboardSummaryWidget`.

assets/js/modules/adsense/util/is-zero-report.js
eugene-manuilov: Ok, updated to return `undefined` when `undefined` is passed.

assets/js/modules/adsense/util/is-zero-report.test.js
eugene-manuilov: Updated.

assets/js/modules/adsense/components/dashboard/DashboardSummaryWidget.js
eugene-manuilov: @tofumatt it happened because of the wrong order of selectors call. `isResolving` selectors must be called after `getReport` ones. I have updated it in the component and now everything works correctly.

assets/js/modules/adsense/util/is-zero-report.js
```diff
@@ -0,0 +1,50 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * Checks whether the report data is empty or not.
+ *
+ * @since n.e.x.t
+ *
+ * @param {Object} report Report data.
+ * @return {boolean} TRUE if the report has no data, otherwise FALSE.
```
tofumatt, line 25:
```suggestion
 * @return {boolean|undefined} TRUE if the report has no data, otherwise FALSE. `undefined` if the `report` passed is undefined.
```

assets/js/modules/adsense/components/dashboard/DashboardSummaryWidget.js
tofumatt: Ah, awesome, good catch.

tofumatt (approved):
Awesome, thanks for the update, that looks great 👍🏻

tofumatt (approved):

felixarntz (approved):
Great!

Issue #2251, opened 2020-10-22
Allow widgets to register a fallback width


felixarntz:
When registering a widget, a `width` argument has to be provided (either `quarter`, `half` or `full`, indicating the width on desktop, which from there is inferred for smaller viewport widths). While this works for the majority of cases, there is at least one existing widget (see #2023) that requires two different widths in order to replicate the current UI behavior.

Widgets should be able to optionally include a `fallbackWidth` argument, supporting the same possible values. The widget renderering mechanism will try to use the regular `width`, but if this causes gaps in the UI, it will check for `fallbackWidth` and whether that would make the grid behave more nicely.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* `registerWidget` should support a new optional argument `fallbackWidth`. If provided, it must be one of the available `WIDGET_WIDTHS`.
* The logic to layout the widget grid in `WidgetAreaRenderer` should be updated:
    * If the current widget's `width` is too much so that it will go into the next row (`counter > 12`):
        * First, test whether the widget has a `fallbackWidth` set and if so whether with that `fallbackWidth` it would still fit into the current row. If yes, rely on the counter value and CSS classes based on the `fallbackWidth` instead.
        * If not, check if the **previous** widget has a `fallbackWidth` set and if so whether that `fallbackWidth` would fill the remaining space in the current row that would otherwise get empty. If yes, replace the counter value and CSS classes of the previous widget based on the `fallbackWidth` values instead.
    * After iterating through all the widgets, but _before the final resizing logic (`counter === 9`), check if the last entire row is not full (`counter < 12`). If so, run the same logic described in the second bullet point above, checking the last widget's `fallbackWidth` and if possible to fill the remaining space alter its counter value and CSS classes.
* The two following widgets should both be updated to specify `half` as primary width and `full` as secondary width:
    * `urlSearch`
    * `searchConsolePopularKeywords`

## Implementation Brief

[...]

PR #2357, opened 2020-11-12
Enhancement/2251 widgets fallback size

eugene-manuilov:
## Summary

Addresses issue #2251

## Relevant technical choices

* The implementation is slightly different from what is requested in the IB, but it also works and does the thing a bit more "elegantly".

## Checklist


assets/js/googlesitekit/widgets/util/get-widget-class-names.js
```diff
@@ -0,0 +1,170 @@
...
+		if ( i + 1 === activeWidgets.length ) {
+			// We need to check whether we have a size that can fit into the row and if so, try to get it.
+			const hasSizeThatCanFitIntoRow = sizes.some( ( [ widgetWidth ] ) => widgetWidth <= 12 );
+			if ( hasSizeThatCanFitIntoRow ) {
+				// Sort available sizes to have the descending order.
+				sizes = sizes.sort( ( [ a ], [ b ] ) => b - a );
+				// Filter out only those sizes that fit into the current row.
+				sizes = sizes.filter( ( [ widgetWidth ] ) => widgetWidth <= 12 );
+			}
+		}
```
felixarntz, line 131: This approach doesn't address the requirement outlined in the ACs to try to fit the last widget _in a row_ into the row - the logic you have here only does something to the last widget _in the entire widget area_. Furthermore it doesn't take the current value of `counter` into account, without which I think it's impossible to address the issue.

To explain a bit more:
* The counter goes up for each widget until it reaches 12, then it goes back to 0 (meaning a new row will be started).
* If a widget increases the counter to a value greater than 12, the widget is too wide. **This is when we need to add new logic here to try to either fit this widget in or try to fill up the remaining space otherwise:**
    * If the widget has a fallback width that fits (`counter` would with that width remain <= 12), use that width. More specifically, use the first width in the array of widths for which that condition applies (if any).
    * Otherwise (the current widget "just doesn't fit"), it will for sure go into the next row (as it already would with the current logic in `develop`). In this case, check the previous widget whether it has an alternative width specified which would allow it to fill up the remaining space (i.e. increase `counter` to the maximum value <= 12).

This logic should happen whenever we reach the end of a row, and as well when we reach the last widget in the entire area.

assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js
```diff
@@ -26,122 +26,19 @@ import PropTypes from 'prop-types';
  * Internal dependencies
  */
 import Data from 'googlesitekit-data';
-import { STORE_NAME, WIDGET_WIDTHS, WIDGET_AREA_STYLES } from '../datastore/constants';
+import { STORE_NAME, WIDGET_AREA_STYLES } from '../datastore/constants';
 import WidgetRenderer from './WidgetRenderer';
+import { getWidgetClassNames } from '../util';
```
felixarntz, line 31: Good idea to move this to a separate file.

assets/js/modules/search-console/index.js
```diff
@@ -25,7 +25,7 @@ import domReady from '@wordpress/dom-ready';
  * Internal dependencies
  */
 import Modules from 'googlesitekit-modules';
-import Widgets from 'googlesitekit-widgets';
+import { registerWidget, WIDGET_WIDTHS } from 'googlesitekit-widgets';
```
felixarntz, line 28: Nit-pick, but we always include the full objects for our "external" imports, let's keep that.

stories/widgets.stories.js
```diff
@@ -131,3 +169,31 @@ storiesOf( 'Global/Widgets', module )
...
+		</WithTestRegistry>
+	) )
+	.add( 'Irregular sizes', ( registry ) => (
+		<WithTestRegistry registry={ registry }>
+			{ createWidgetAreas(
+				registry,
+				[ QUARTER, QUARTER, HALF, [ QUARTER, FULL ] ],
+				[ [ HALF, FULL ], QUARTER, QUARTER ],
+				[ QUARTER, [ FULL, HALF ], QUARTER ],
+				[ QUARTER, QUARTER, [ HALF, FULL ] ],
```
felixarntz, line 195: The overall implementation of these stories looks good, however a few "key sizes" to test some behavior is missing (those are outlined in the IB):
* 3 quarter widgets (tests that expanding them equally to fill the full width works, see the `counter === 9` logic)
* 1 half widget, 1 quarter widget that has half as secondary width, 1 full widget (tests that filling up remaining space in a row with the fallback width works, see my above comment where the previous widget should be checked to find the maximum width that is <= 12)

stories/widgets.stories.js
```diff
@@ -131,3 +169,31 @@ storiesOf( 'Global/Widgets', module )
...
+			) }
+		</WithTestRegistry>
+	) )
+	.add( 'Irregular sizes', ( registry ) => (
+		<WithTestRegistry registry={ registry }>
+			{ createWidgetAreas(
+				registry,
+				[ QUARTER, QUARTER, HALF, [ QUARTER, FULL ] ],
+				[ [ HALF, FULL ], QUARTER, QUARTER ],
+				[ QUARTER, [ FULL, HALF ], QUARTER ],
```
felixarntz, line 194: This test case is great to have, I wanna point this one out particularly because right now it is not showing up the way it should. Per the ACs and my other comment above, the following should happen here:
1. Render quarter widget (1).
2. Notice that the full widget (2) doesn't fit into the row (`counter` would be 15), hence try to see if the widget (2) has a smaller size that fits - which it does, so use the half width version of (2) instead.
3. Render the quarter widget (3), which will now actually even fit into the same row.

felixarntz (changes requested):
The logic here doesn't satisfy the ACs, I left some more details to hopefully clarify further.

assets/js/modules/search-console/index.js
eugene-manuilov: Reverted.

stories/widgets.stories.js
eugene-manuilov: Added these variations to the storybook story.

stories/widgets.stories.js
eugene-manuilov: It should be fixed now after amending the algorithm.

assets/js/googlesitekit/widgets/util/get-widget-class-names.js
eugene-manuilov: @felixarntz, I have updated my approach to work for every row where an alternative size can be used. So, I think we should be good now.

> Furthermore it doesn't take the current value of `counter` into account, without which I think it's impossible to address the issue.

It does take it into account and makes all calculations based on the current value of the `counter`.

assets/js/googlesitekit/widgets/util/get-widget-class-names.js
```diff
@@ -0,0 +1,189 @@
...
+
+	activeWidgets.forEach( ( widget, i ) => {
+		// Get available sizes for the current widget to select the most appropriate width for the current row.
+		let sizes = getWidgetSizes( counter, widget );
+
+		if (
+			// If it is the last widget in the current row.
+			i + 1 === activeWidgets.length ||
+			// Or the next widget can't fit into the current row anyway, then we can try to use alternative sizes.
+			getWidgetSizes( sizes.sort( ascending )[ 0 ][ 0 ], activeWidgets[ i + 1 ] ).filter( fitIntoRow ).length === 0
```
felixarntz, line 140: Nice! 👍

felixarntz (approved):
Awesome work, +1 figuring out all the logic needed here!

Just a few minor comments that would be great to address/discuss.

assets/js/googlesitekit/widgets/util/get-widget-class-names.js
```diff
@@ -0,0 +1,189 @@
...
+}
+
+/**
+ * Gets an array of sizes that widget can take accounting the current counter value.
+ *
+ * @since n.e.x.t
+ *
+ * @param {number} counter The current counter.
+ * @param {Object} widget  Widget object.
+ * @return {Array.<Array.<number, string>>} Sizes array.
```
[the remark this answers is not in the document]
eugene-manuilov: Yes, I like the idea to return an array of objects. Updated.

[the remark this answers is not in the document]
felixarntz: The documentation here needs to be updated now.

felixarntz (approved):
Awesome work!

Issue #2156, opened 2020-10-12
Received errors not cleared when calling a second time for some actions


aaemnnosttv:
## Bug Description

Several actions manually receive errors for their actions but do not clear them when called again. This results in the potential for an error message to persist in the interface when it shouldn't.

As an example, if an error is returned from the control for the `SUBMIT_CHANGES` action, each datastore receives the error into the store under the `submitChanges` base name. This error is displayed properly on the front end.

![image](https://user-images.githubusercontent.com/1621608/95724481-19e8d000-0c7f-11eb-91d1-546c454e337c.png)

However, after submitting again, the error is not cleared as is shown above.

Example of problem in GTM:

https://github.com/google/site-kit-wp/blob/3a5b6d9871598791fedf478042f0f150510a474b/assets/js/modules/tagmanager/datastore/settings.js#L62-L85

Other actions that currently exhibit this bug include:
- `saveUserInputSettings`
- `saveSettings`
- Analytics' `createAccount`
- Analytics' `getAccounts`
- Analytics' `getProperties`

[See `createFetchStore` for correct error action use](https://github.com/google/site-kit-wp/blob/3a5b6d9871598791fedf478042f0f150510a474b/assets/js/googlesitekit/data/create-fetch-store.js#L136-L174)

## Steps to reproduce

This is easiest to test using the bug in https://github.com/google/site-kit-wp/issues/1817

1. Activate Tag Manager and create a new container
1. Attempt to create another container for the same account (it will fail because GTM requires all containers in the account to have a unique name)
    - You should see the error on the screen
1. In a new tab, change your site title under Settings > General
1. Back in the GTM setup/settings, submit the form again to create a new container
1. The request will succeed but the error will still be shown

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Error actions used within the datastore should be yielded directly rather than using `registry.dispatch()` which should only be used when interacting with the data stores externally (e.g. Storybook)
* All actions that manually call `receiveError` must call `clearError` for the same error before an inner (potentially async) action is triggered that may return the error that would be received

## Implementation Brief

* Updating the errors to `yield` the error action directly by importing from `assets/js/googlesitekit/data/create-error-store.js` and using `yield createErrorStore.actions.receiveError();`
[...]

PR #2359, opened 2020-11-13
Clear received errors before retrying an action

kostyalmm:
## Summary

Addresses issue #2156

## Relevant technical choices

* Add clearError in relevant places.
* Use `createErrorStore` instead of `registry.dispatch` inside of store

## Checklist


assets/js/modules/tagmanager/datastore/settings.js
```diff
@@ -88,6 +88,11 @@ export async function submitChanges( { select, dispatch } ) {
 	// This action shouldn't be called if settings haven't changed,
 	// but this prevents errors in tests.
 	if ( select( STORE_NAME ).haveSettingsChanged() ) {
+		// Clear any errors that may have been generated while creating container.
+		await dispatch( STORE_NAME ).clearErrors( 'createContainer' );
+		// Clear any error (in state.error) that legacy code may have reported.
+		await dispatch( STORE_NAME ).clearError();
```
kostyalmm, line 94: This solves the bug on the issue.

kostyalmm (commented):
Most of this issue was already done earlier in #2136 and the existing IB is outdated now.
@tofumatt , @felixarntz can you confirm that this issue is solved after these changes?

assets/js/modules/tagmanager/datastore/settings.js
aaemnnosttv: @kostyalmm would you please provide a bit more detail as to what specifically this is solving?.

assets/js/modules/tagmanager/datastore/settings.test.js
```diff
@@ -210,6 +210,24 @@ describe( 'modules/tagmanager settings', () => {
 					expect( console ).toHaveErrored();
 				} );
 
+				it( 'clears errors from past dispatches', async () => {
```
aaemnnosttv, line 213: I'm not sure this is needed since each of its inner actions is a fetch request which will clear its own errors before making the request again.

aaemnnosttv (changes requested):
Thanks @kostyalmm – there are a few things left to address here, see my comments below. A few questions as well where it wasn't clear why something had been changed. In the future, please provide a bit more detail so reviewers can better understand your thought process.

Regarding #2136, I'm going to look at that one now as it seems there are a few things missed on that one as well. Keep an eye on this one for changes, although all the items below are still valid.

assets/js/modules/tagmanager/datastore/settings.js
kostyalmm: The problem mentioned in the issue seems to be fixed already before I took the issue. Can you confirm?
I was confused and made up another (unrealistic) flow altogether on my own 😓 . 
This change is not required, reverted it back.

assets/js/googlesitekit/data/create-fetch-store.js
```diff
@@ -142,7 +142,7 @@ export const createFetchStore = ( {
 			type: START_FETCH,
 		};
 
-		yield clearError( baseName, args );
+		yield clearErrors( baseName );
```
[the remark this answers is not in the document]
kostyalmm: Not required. See [comment](https://github.com/google/site-kit-wp/pull/2359#discussion_r524941058)

tofumatt (approved):
Just had a look through this and comparing it with the ACs/IBs—for the places this is still relevant it looks like everything that needed updating was changed. As mentioned: some of this code changed after #2136.

Looks like this should cover everything needed and the points raised by @aaemnnosttv have been resolved.

aaemnnosttv (approved):
👍 👍

Issue #2027, opened 2020-09-15
Cancel previous workflow runs on newer refs


aaemnnosttv:
## Feature Description

Our GitHub Action workflows currently run until completed or fail for every commit they are triggered by.

In most cases however, it is usually undesirable for a workflow to continue running once another workflow for the same branch has started. In some cases, this can even cause a race condition where git-based deploy jobs fail due to upstream changes cause a push to fail. This can result in deployed assets not reflecting the latest state for their respective PR/branch or cause deployed artifacts to be left in the target repo longer than they should be due to a similar conflict in the prune/remove workflow job.

Travis has similar functionality available to automatically cancel builds for a branch/PR when a newer commit is pushed as it essentially "invalidates" the previous run. There's little benefit in letting the jobs finish as we have a limited number of concurrent jobs that can run. This is only manageable via the [settings for the repository for Travis](https://travis-ci.com/github/google/site-kit-wp/settings), but we should enable these as well:
![image](https://user-images.githubusercontent.com/1621608/93198371-9dcd9c00-f755-11ea-9202-b87d10011b56.png)

For GitHub Actions, we can leverage the [`cancel-workflow-action`](https://github.com/marketplace/actions/cancel-workflow-action) to get similar behavior for our actions. (It's not quite as good as Travis' as it requires the job to run which in turn requires availability in the queue)

This (in combination with the settings above for Travis and the recently added `fast_finish` configuration) will help us to run our builds as efficiently as possible within our limits for maximum concurrent jobs/builds.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* GitHub Action workflows for queued/pending, or in progress runs should be automatically cancelled when a newer commit is pushed to the same branch

## Implementation Brief

* Merge #2361 

## QA Brief

* Review the [Actions](https://github.com/google/site-kit-wp/actions) log for the repository
* Check workflows for VRT, Plugin Zips, and Storybook to verify cancelled workflows are present (shown by the gray octagon with ❕ inside) - e.g.  
![image](https://user-images.githubusercontent.com/1621608/99103292-09ae7280-25e8-11eb-9d67-8439ad3eac22.png)

## Changelog entry

* N/A

felixarntz:
IB ✅

felixarntz:
@aaemnnosttv Not high priority to finish within this sprint, but since you already implemented it, and if this is good to go, we can just add Sprint 36 to it.

[...]

PR #2361, opened 2020-11-13
Add auto-cancellation workflow

aaemnnosttv:
## Summary

Addresses issue #2027

## Relevant technical choices

- Used dedicated workflow for cancelling all relevant workflows rather than updating individual flows and allows for cancelling a workflow before it runs

**Testing**
- [Triggered workflow, cancelled by subsequent commit](https://github.com/google/site-kit-wp/pull/2361/checks?check_run_id=1397056795)
- [Cancel workflow run](https://github.com/google/site-kit-wp/runs/1397062484?check_suite_focus=true)

<details>
<summary><strong>Screenshots</strong></summary>

image
image
image
image

</details>

## Checklist


.github/workflows/cancel.yml
```diff
@@ -0,0 +1,29 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 3
+    steps:
+      # See https://github.com/marketplace/actions/cancel-workflow-action#advanced
+      - uses: styfle/cancel-workflow-action@0.6.0
+        env:
+          # See https://api.github.com/repos/google/site-kit-wp/actions/workflows for reference.
+          WID_SB: 1979909 # storybook.yml
+          WID_VRT: 2751946 # visual-regression.yml
+          WID_ZIPS: 1979910 # zips.yml
```
felixarntz, line 26: According to the docs, this also supports workflow names - wouldn't this be more intuitive to use?

felixarntz (approved):
@aaemnnosttv One comment here, but nothing blocking.

.github/workflows/cancel.yml
aaemnnosttv: Looking at it again, that would be okay too. I originally read the docs for that to mean you could use the actual workflow `name` which I wouldn't want but it's the file name which is much more stable.

I guess it depends on how GH treats the workflow ID in relation to the file name. I would expect the ID to remain unchanged even if the workflow changed completely.

After taking a closer look at this, I found the [API documentation](https://developer.github.com/v3/actions/workflow-runs/#list-workflow-runs) for listing workflow runs which is what this uses (via octokit). 

> List all workflow runs for a workflow. You can also replace `:workflow_id` with `:workflow_file_name`. For example, you could use `main.yml`. You can use parameters to narrow the list of results. For more information about using parameters, see Parameters.

So that means we wouldn't need to use the full `.github/workflows/zips.yml` path which is what I was thinking it would require.

File names it is then? 😄

aaemnnosttv: Updated to use filenames instead. It's much cleaner and doesn't require referencing something external to update. Also, the ability to use a filename instead is a feature provided by the underlying GitHub API ([example](https://api.github.com/repos/google/site-kit-wp/actions/workflows/storybook.yml/runs)) rather than the extension so this is much better I think 👍

felixarntz (approved):

Issue #2310, opened 2020-11-02
Mark user input settings state flag as completed upon completion


felixarntz:
Following up on #2039 and #2042, the `googlesitekit_user_input_state` flag option should be set to `completed` once user input settings are submitted and complete.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `googlesitekit_user_input_state` flag option should be set to `completed` once user input settings have been saved on the proxy and are set for everything.
* The [logic currently used to check for whether settings have been completed](https://github.com/google/site-kit-wp/pull/2215/files#diff-fd8b5c88edabac5e94ebf3751d5ea639f149dec14a1296a74b6890617a92bf2dR1220-R1228) should be moved to some reusable method which can be used both after syncing values with the proxy as well as in the `admin_init` check when the flag option is empty.

## Implementation Brief

* Add a new public function `get_user_input_state` to the `Core\Authentication\Authentication` class. It should return the private `$user_input_state` property.
* Add a new public function `are_settings_empty` to the `Core\Util\User_Input_Settings` class. It should get the current settings using the `get_settings` function if settings aren't provided via arguments and return TRUE if there is at least one setting with empty `values`, otherwise, return FALSE (see [here](https://github.com/google/site-kit-wp/pull/2215/files#diff-fd8b5c88edabac5e94ebf3751d5ea639f149dec14a1296a74b6890617a92bf2dR1220-R1228) for reference).
* Update the `verify_user_input_settings` method of the `Authentication` class to use the new `are_settings_empty` method of the `$this->user_input_settings` instance instead of hardcoded logic.
* Update the `set_settings` function of the `User_Input_Settings` class to set settings state flag after successfully saving user input settings on the proxy server.

### Test Coverage

* Add appropriate tests to the `Tests\Core\Util\User_Input_SettingsTest` class to check the new `are_settings_empty` method and the new behavior of the `set_settings` method.

### Visual Regression Changes

* No visual changes should be visible on the plugin pages.

## QA Brief

* Check that the `Authentication` class has a new method to get user input state instance and it uses the `are_settings_empty` function to determine whether there are empty settings.
* Check that the `User_Input_Settings` class has a new static method and it sets user input state flag after successful submission.
* Make sure PHPUnit tests the new functionality of the `User_Input_Settings` class.

## Changelog entry

* Mark user input state as completed once user has submitted their information.

felixarntz:
@eugene-manuilov 
[...]

PR #2362, opened 2020-11-13
Enhancement/2310 mark user settings state flag

eugene-manuilov:
## Summary

Addresses issue #2310

## Checklist


tofumatt (dismissed):
Looks good to me. The bit about the `null` return value in `are_settings_empty ` is well-documented—I appreciated it especially as it wasn't mentioned in the IB. 🙂 

Thanks for that 👍🏻

includes/Core/Authentication/Authentication.php
```diff
@@ -1202,29 +1207,23 @@ private function get_proxy_permissions_url() {
...
-			|| ! $this->is_authenticated()
-			|| ! $this->credentials()->has()
-			|| ! $this->credentials->using_proxy()
+			empty( $this->user_input_state->get() )
+			&& $this->is_authenticated()
+			&& $this->credentials()->has()
+			&& $this->credentials->using_proxy()
 		) {
-			return;
-		}
```
felixarntz, line 1219: Nit-picking, but why turn this condition around? I think where possible we should use early returns to avoid deep nesting. Of course, the nesting here is only one level deep, but I think sticking to this pattern overall makes sense and avoids unnecessary code changes when later indenting differently, making code review unnecessarily harder.

felixarntz (approved):
Looks good!

@eugene-manuilov Just one comment, but nothing blocking.

includes/Core/Authentication/Authentication.php
eugene-manuilov: I thought it would be easier to read if I combined it. Do you want me to revert it?

felixarntz: Thanks, but not a big deal. Just overall I think we should prefer early returns over nesting where possible.

Issue #2137, opened 2020-10-06
Rely on submitChanges and canSubmitChanges in settings components


felixarntz:
Following #2136, every module store now has a `submitChanges` action and a `canSubmitChanges` selector. The legacy `SettingsModule` and `SettingsModules` components should start making use of that.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `SettingsModule` (singular) component should be modified:
    * It should receive a `canSubmitChanges` prop via the `withSelect` HOC by selecting `canSubmitChanges( slug )` from the `core/modules` store.
    * The "Confirm Changes" button should be marked as disabled if either `isSavingModule` (like today) or `! canSubmitChanges`.
    * The component should also receive a `submitChanges` function prop via the `withDispatch` HOC from the `core/modules` store.
    * The component should **not be further refactored, or as little as possible**.
    * The `handleEdit` call with `'confirm'` should be expanded to pass the `submitChanges` prop as the third argument.
* The `SettingsModules` (plural) component should be modified to receive `submitChanges` as an optional third parameter to the `handleButtonAction` function, which is only relevant if `action === 'confirm'`
    - Instead of running the `googlekit.SettingsConfirmed` filter, it should call the given `submitChanges( slug )` (if present).
    - If the `submitChanges` call resolves to an object that has an `error` property, the promise should be rejected with that `error` to flow through the `catch` logic, otherwise it should be resolved normally as a successful completion.
    * The component should **not be further refactored, or as little as possible**.
* From every module's `SettingsEdit` component, the `useEffect` usages that toggle the disabled state of the "Confirm Changes" button and that hook in the `googlekit.SettingsConfirmed` filter should be removed, as they are now covered by the above automatically.

## Implementation Brief

* Edit the `assets/js/components/settings/settings-module.js` file:
  * Update `withSelect` HOC to get a `canSubmitChanges` value using the `canSubmitChanges( slug )` selector from the `core/modules` store;
  * Update the `Confirm Changes` button to be switched to disabled state when the `isSavingModule` variable is true or the `canSubmitChanges` property is false;
  * Import [`compose` ](https://github.com/WordPress/gutenberg/tree/master/packages/compose#api)function from the `@wordpress/compose` package and use it to compose HOCs for the `SettingsModule` components;
[...]

PR #2363, opened 2020-11-15
Use submitChange store in SettingsModule.

kostyalmm:
## Summary

Addresses issue #2137

## Relevant technical choices

* Use `withDispatch` in `SettingsModule`.
* Use Submit Changes store in Settings module.

## Checklist

### DIFF
--- a/assets/js/components/settings/settings-module.js
+++ b/assets/js/components/settings/settings-module.js
@@ -222,6 +222,14 @@ class SettingsModule extends Component {
 			}
 		}
 
+		// Set button action based on state.
+		let buttonActionName = 'cancel';
+		let buttonAction;
+		if ( hasSettings && setupComplete ) {
+			buttonActionName = 'confirm';
+			buttonAction = submitChanges;
+		}
+
 		return (
 			<Fragment>
 				{ active ? (
@@ -366,7 +374,7 @@ class SettingsModule extends Component {
 											{ isEditing[ moduleKey ] || isSavingModule ? (
 												<Fragment>
 													<Button
-														onClick={ () => handleEdit( moduleKey, hasSettings && setupComplete ? 'confirm' : 'cancel', hasSettings && setupComplete ? submitChanges : null ) }
+														onClick={ () => handleEdit( moduleKey, buttonActionName, buttonAction ) }
 														disabled={ isSavingModule || ! canSubmitChanges }
 														id={ hasSettings && setupComplete ? `confirm-changes-${ slug }` : `close-${ slug }` }
 													>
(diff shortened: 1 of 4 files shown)


assets/js/components/settings/settings-module.js
```diff
@@ -363,8 +366,8 @@ class SettingsModule extends Component {
 											{ isEditing[ moduleKey ] || isSavingModule ? (
 												<Fragment>
 													<Button
-														onClick={ () => handleEdit( moduleKey, hasSettings && setupComplete ? 'confirm' : 'cancel' ) }
-														disabled={ isSavingModule }
+														onClick={ () => handleEdit( moduleKey, hasSettings && setupComplete ? 'confirm' : 'cancel', hasSettings && setupComplete ? submitChanges : null ) }
```
tofumatt, line 369: The multiple ternaries on the same line as arguments is quite tough to read here. I know it wasn't in the IB but I think this set of params would benefit from being defined above, before the JSX render as variables that can be passed to `handleEdit`. As-is, this is too hard to read—even before it was crowded, but I think the extra argument makes it worth it to define these params above and use a more concise call to `handleEdit` here.

assets/js/components/settings/settings-modules.js
```diff
@@ -82,31 +81,16 @@ class SettingsModules extends Component {
...
-			}
-			modulePromise.then( () => {
-				// Clears session and local storage on every successful setting.
-				clearWebStorage();
-
-				this.setState( {
-					isSaving: false,
-					error: false,
-				} );
+			const { error: err } = await submitChanges( module );
```
tofumatt, line 91: Why define this as `err`? `error` is a fine variable name here and means the code will be a bit less complicated 🙂

tofumatt (changes requested):
Looks good, just a few readability tweaks I think worth making, let me know what you think. After that feel free to assign it back to me and we should be good to merge. 🙂

tofumatt (approved):

assets/js/components/settings/settings-modules.js
```diff
@@ -82,45 +81,43 @@ class SettingsModules extends Component {
...
+						error: {
+							errorCode: error.code,
+							errorMsg: error.message,
+						},
+					} );
 				}
 				this.setState( {
 					isSaving: false,
-					error,
 				} );
```
felixarntz, line 108: Let's not make more `setState` calls here than needed, we can change this code to always make a single one, which is also what the previous implementation did. Since `isSaving` is always set to `false`, let's put that into both `setState` calls in the clause above and remove the call from here instead.

felixarntz (dismissed):
Looks good! Just one point of minor improvement, but nothing blocking.

felixarntz (approved):

Issue #2248, opened 2020-10-22
Refactor `TableOverflowContainer` component


felixarntz:
The `TableOverflowContainer` component should be refactored to become a functional component using hooks, and it should be moved to its own component file.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `TableOverflowContainer` component should be refactored to a functional component using hooks.
* It should be moved out of `assets/js/components/data-table.js` into its own file `assets/js/components/TableOverflowContainer.js`.
* All import references should be updated accordingly.
* A global Storybook story should be added for `TableOverflowContainer`, with some manually created table with any rows and columns inside - it should be a fairly large table to demonstrate the overflow capacity (e.g. 30 rows, 15 columns or something like that).

## Implementation Brief

* Create a new `assets/js/components/TableOverflowContainer.js` file that exports `TableOverflowContainer` functional component;
* Update all from `data-table.js` file to use the new `TableOverflowContainer.js` file to import `TableOverflowContainer` component.
* Replicate the logic from the legacy component in the new `TableOverflowContainer` component:
  * Use the `useState` hook to define `isScrolling` state property;
  * Use the `useEffect` hook to add an event listener for the `resize` event.
* Remove the legacy `TableOverflowContainer` component from the `assets/js/components/data-table.js` file.
* Add a new story to the "global" section of the storybook for the new `TableOverflowContainer` component (see requirements in the AC section).

### Test Coverage

* Not needed

### Visual Regression Changes

* Not needed

## QA Brief

* Check that all imports of the `TableOverflowContainer` component use the new file.
* Check that the new component replicates the logic from the legacy one.
* Check that there is no visual regression caused by the new component.

## Changelog entry

* N/A

felixarntz:
IB ✅

caseymorrisus:
QA ✅

PR #2364, opened 2020-11-15
Refactor TableOverflowContainer and add stories for same.

kostyalmm:
## Summary

Addresses issue #2248

## Relevant technical choices

* Add `TableOverflowContainer.js`
* Move the refactored `TableOverflowContainer` from `data-table.js` to the created file.
* Fix imports.
* Add story for the same.

## Checklist

### DIFF
--- a/assets/js/components/TableOverflowContainer.js
+++ b/assets/js/components/TableOverflowContainer.js
@@ -20,35 +20,33 @@
  * External dependencies
  */
 import PropTypes from 'prop-types';
+import { debounce } from 'lodash';
+import classnames from 'classnames';
 
 /**
  * WordPress dependencies
  */
 import { useEffect, useRef, useState } from '@wordpress/element';
 
-/**
- * Internal dependencies
- */
-import { debounce } from 'lodash';
-import classnames from 'classnames';
-
 function TableOverflowContainer( { children } ) {
 	const [ isScrolling, setScrolling ] = useState( false );
 	const scrollRef = useRef();
 
 	useEffect( () => {
-		updateFadeOnScroll();
+		setIsScrolling();
 
-		const resize = debounce( function() {
-			updateFadeOnScroll();
-		}, 100 );
+		const resize = debounce( setIsScrolling, 100 );
 
 		global.addEventListener( 'resize', resize );
 
 		return () => global.removeEventListener( 'resize', resize );
 	}, [] );
 
-	const updateFadeOnScroll = () => {
+	const setIsScrolling = () => {
+		if ( ! scrollRef.current ) {
+			return;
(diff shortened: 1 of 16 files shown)


assets/js/components/TableOverflowContainer.js
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * WordPress dependencies
+ */
+import { useEffect, useRef, useState } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import { debounce } from 'lodash';
+import classnames from 'classnames';
```
tofumatt, line 33: These aren't internal dependencies, so they should be moved above to the "External dependencies" section.

assets/js/components/TableOverflowContainer.js
```diff
@@ -0,0 +1,78 @@
...
+			updateFadeOnScroll();
+		}, 100 );
+
+		global.addEventListener( 'resize', resize );
+
+		return () => global.removeEventListener( 'resize', resize );
+	}, [] );
+
+	const updateFadeOnScroll = () => {
+		const { scrollLeft, scrollWidth, offsetWidth } = scrollRef.current;
```
tofumatt, line 52: It's possibly for `scrollRef.current` to be `null`, isn't it? I suppose it's unlikely given the logic flow of this being called in a `useEffect` hook, but I think it might be best to protect against `scrollRef.current` being null, just-in-case.

assets/js/components/TableOverflowContainer.js
```diff
@@ -0,0 +1,78 @@
...
+		const resize = debounce( function() {
+			updateFadeOnScroll();
+		}, 100 );
+
+		global.addEventListener( 'resize', resize );
+
+		return () => global.removeEventListener( 'resize', resize );
+	}, [] );
+
+	const updateFadeOnScroll = () => {
```
tofumatt, line 51: I know this name is inherited from the old component, but it's a bit of an odd name as it's run on resize and scroll, and also seems to be signalling that it's updating a "fade" which is the gradient appearing on scroll/resize.

Let's use a more descriptive name, like `calculateScrollStatus` or `setIsScrolling`—that better reflects what this does and doesn't tie it to a particular action like scrolling.

tofumatt (changes requested):
Looks good 👍🏻 

Just a few minor tweaks/comments; let me know what you think about my naming suggestions and the `.current` safety checks. Feel free to assign this review back to me after you've addressed the feedback 🙂

tofumatt (approved):
Awesome 👍🏻

felixarntz (approved):
Awesome!

Issue #2243, opened 2020-10-22
Introduce Analytics utility function `isZeroReport` and use it in new widgets


felixarntz:
There should be a new Analytics utility function `isZeroReport`, similar to the legacy `isDataZeroForReporting` function but without some of the legacy details. The new function should be used in all refactored Analytics widgets for the respective conditional which takes care of returning the "no-data" component.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new function `isZeroReport` should be added in `assets/js/modules/analytics/util/is-zero-report.js`:
    * It should expect a `report` object like the one coming from the `modules/analytics` `getReport` selector.
    * If `report` is `undefined`, it should return `undefined` (to account for loading state).
    * Otherwise, it should return `true` if `report?.[ 0 ]?.data?.rows` is not set or empty or if `report?.[ 0 ]?.data?.totals?.[ 0 ]` is not set or empty. It should also return `true` if none of the values in `report[ 0 ].data.totals[ 0 ]` is greater than 0.
    * Otherwise it should return `false` (i.e. there _is_ data).
* The following (widget) components should use the above function instead of manual checks, for all reports:
    * `DashboardAllTrafficWidget`
    * `DashboardBounceRateWidget`
    * `DashboardGoalsWidget`
    * `DashboardPopularPagesWidget`
    * `DashboardUniqueVisitorsWidget`
    * `DashboardTopEarningPagesWidget` (AdSense)

## Implementation Brief

* Create a new `assets/js/modules/analytics/util/is-zero-report.js` file;
* Add a new `isZeroReport` function to the `is-zero-report.js` file that satisfies requirements defined in the AC section;
* Add `export * from './is-zero-report';` to the `assets/js/modules/analytics/util/index.js` file;
* Update components listed in the AC to use the new `isZeroReport` function instead of `isDataZeroForReporting` or `! data || ! data.length`.

### Test Coverage

* Create a new `assets/js/modules/analytics/util/is-zero-report.test.js` file;
* Add tests to check the `isZeroReport` function.

### Visual Regression Changes

* No visual changes should occur as a result of these changes.

## QA Brief

* Check the components defined in the AC to make sure they use the new `isZeroReport` function.
* There should be no visual changes on the dashboard page.

## Changelog entry

* Improve logic to determine whether an Analytics API report response is "empty", via new `isZeroReport` function for Analytics.

felixarntz:
IB ✅

[...]

PR #2369, opened 2020-11-16
isZeroReport utility function for Analytics

benbowler:
## Summary

Add a new utility for Analytics widgets to detect empty data and show the empty component in this case.

Addresses issue #2243 

## Relevant technical choices

I've added the new helper to this file `assets/js/modules/analytics/util/is-zero-report.js` with tests here `assets/js/modules/analytics/util/is-zero-report.test.js` before applying it to each template that loads this report type.

## Checklist


assets/js/modules/analytics/util/is-zero-report.test.js
```diff
@@ -0,0 +1,86 @@
...
+				data: {
+					rows: [
+						{}, {}, {},
+					],
+					totals: [
+						{
+							values: [
+								'928',
+								'928',
+							],
```
aaemnnosttv, line 76: Please add another case that has values with `0` and a positive number to show that it doesn't require all values to be non-zero.

assets/js/modules/analytics/util/is-zero-report.test.js
```diff
@@ -0,0 +1,86 @@
...
+			{
+				data: {
+					rows: [
+						{}, {}, {},
+					],
+					totals: [
+						{
+							values: [
+								'928',
+								'928',
```
aaemnnosttv, line 75: Bit of a nitpick here but let's make it a little more clear that these are dummy values and that there is no significance to them being the same number.
```suggestion
								'123',
								'234',
```

assets/js/modules/analytics/util/is-zero-report.test.js
```diff
@@ -0,0 +1,86 @@
...
+						{}, {}, {},
+					],
+					totals: [
+						{
+							values: [
+								'0',
+								'0',
+							],
+						},
+					],
```
aaemnnosttv, line 56: Not to say that we should mirror our real/expected usage in tests (we shouldn't) but we should cover the cases that we use which aren't all covered here yet.
* Multiple items in `totals` (most of the time we just have 1, but some have 2, depending on the number of date ranges requested – 2 will be more common in the future)  
This also relates to my previous comment about whether or not we should consider values from other totals or only the first one (pending input from @felixarntz) but is worth including here either way.
* A single value in the list of `values` (depends on the number of metrics requested, but we usually only request one per report it seems)

aaemnnosttv (changes requested):
Thanks @benbowler – this is off to a good start, just a few things to address – mostly little things and a few tests to add/tweak. There's one comment where I've asked for clarification from @felixarntz but that shouldn't block the rest of it from being addressed.

assets/js/modules/analytics/util/is-zero-report.js
```diff
@@ -0,0 +1,48 @@
...
+export function isZeroReport( report ) {
+	if ( report === undefined ) {
+		return undefined;
+	}
+
+	if ( ! report?.[ 0 ]?.data?.rows?.length || ! report?.[ 0 ]?.data?.totals?.[ 0 ]?.values?.length ) {
+		return true;
+	}
+
+	const sumOfTotals = report[ 0 ].data.totals.reduce( function( totalSum, values ) {
```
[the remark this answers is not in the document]
benbowler: I've updated to check the first totals result with 'Array.some' now based on the AB and will wait for input from @felixarntz on expanding to all items.

assets/js/modules/analytics/util/is-zero-report.test.js
benbowler: I've added in tests for multiple totals objects and single, multiple and mixed 0 and non-0 values.

assets/js/modules/analytics/util/is-zero-report.js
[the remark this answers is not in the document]
felixarntz: @benbowler @aaemnnosttv Good point, let's expand the check to cover all `report[ 0 ].data.totals`, not just the first one (there will only be a maximum of two date ranges per how the API works fyi). If the previous date range has data, it should indeed be an indicator that this is not a "zero data" state.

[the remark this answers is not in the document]
benbowler: Thanks Felix, I'll get this updated and ready for CR.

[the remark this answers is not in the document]
benbowler: I've updated the function to test all totals values and simplified it to a single line return.

benbowler:
Updated based on feedback.

assets/js/modules/analytics/util/is-zero-report.test.js
```diff
@@ -0,0 +1,196 @@
...
+		[ 'a string', 'test' ],
+		[ 'an empty object', {} ],
+		[ 'an object without rows or totals', [ { data: {} } ] ],
+		[ 'an object with empty rows', [ { data: { rows: [] } } ] ],
+		[ 'an object with empty totals', [ { data: { totals: [] } } ] ],
+	] )( 'returns true when %s is passed', ( _, report ) => {
+		expect( isZeroReport( report ) ).toBe( true );
+	} );
+
+	it( 'returns true id all totals are 0', () => {
```
aaemnnosttv, line 42: Oops!
```suggestion
	it( 'returns true if all totals are 0', () => {
```

assets/js/modules/analytics/util/is-zero-report.js
```diff
@@ -0,0 +1,48 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * Checks whether the report data is empty or not.
+ *
+ * @since n.e.x.t
+ *
+ * @param {Object} report Report data object.
+ * @return {boolean | undefined} Returns undefined if in the loading state, true if the report has no data or missing data, otherwise false.
```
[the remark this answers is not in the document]
aaemnnosttv: This is yet to be addressed.

assets/js/modules/analytics/util/is-zero-report.test.js
```diff
@@ -0,0 +1,196 @@
...
+								'123',
+							],
+						},
+					],
+				},
+			},
+		];
+
+		expect( isZeroReport( reportSingleTotal ) ).toBe( false );
+	} );
```
aaemnnosttv, line 194: Please split the assertions in the last two tests here into separate tests. That is, those two currently have 3 expect calls which work fine as-is, but would be a bit more useful in the case of a failure to have separately with descriptive names for each.

aaemnnosttv (commented):
This one is super close @benbowler – just a few final small things left to address and then this should be good to go 👍

assets/js/modules/analytics/util/is-zero-report.test.js
benbowler: Done.

assets/js/modules/analytics/util/is-zero-report.test.js
benbowler: Both done and pushed.

aaemnnosttv: Thanks @benbowler – it looks like you only split the last test, can you do it for the previous one as well? Apologies if this wasn't clear from my last comment.

aaemnnosttv (approved):
LGTM 👍

Issue #2244, opened 2020-10-22
Introduce Search Console utility function `isZeroReport` and use it in new widgets


felixarntz:
There should be a new Search Console utility function `isZeroReport`, similar to the legacy `isDataZeroSearchConsole` function but without some of the legacy details. The new function should be used in all refactored Search Console widgets for the respective conditional which takes care of returning the "no-data" component.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new function `isZeroReport` should be added in `assets/js/modules/search-console/util/is-zero-report.js`:
    * It should expect a `report` object like the one coming from the `modules/search-console` `getReport` selector.
    * If `report` is `undefined`, it should return `undefined` (to account for loading state).
    * Otherwise, it should return `true` if `report` is not an array or empty. It should also return `true` if none of the values in `report` (`clicks`, `impressions`, `ctr`, `position`) is greater than 0. The function should _not_ use the legacy function `extractSearchConsoleDashboardData`.
    * Otherwise it should return `false` (i.e. there _is_ data).
* The following (widget) components should use the above function instead of manual checks, for all reports:
    * `DashboardClicksWidget`
    * `DashboardImpressionsWidget`
    * `DashboardPopularKeywordsWidget`

## Implementation Brief

* Create a new `assets/js/modules/search-console/util/is-zero-report.js` file;
* Add a new function `isZeroReport` to the `is-zero-report.js` file that satisfies requirements defined in the AC section;
* Add `export * from './is-zero-report';` to the `assets/js/modules/search-console/util/index.js` file;
* Update `DashboardClicksWidget`, `DashboardImpressionsWidget`, and `DashboardPopularKeywordsWidget` widgets to use the new function instead of manually checking `! data || ! data.length`.

### Test Coverage

* Create a new `assets/js/modules/search-console/util/is-zero-report.test.js` file;
* Add tests to check the `isZeroReport` function.

### Visual Regression Changes

* No visual changes should occur as a result of these changes.

## QA Brief

* Check `DashboardClicksWidget`, `DashboardImpressionsWidget`, and `DashboardPopularKeywordsWidget` widgets to make sure they use the new `isZeroReport` function;
* There should be no visual changes on the dashboard page.

## Changelog entry

* Improve logic to determine whether a Search Console API report response is "empty", via new `isZeroReport` function for Search Console.

felixarntz:
IB ✅

tofumatt:
[...]

PR #2370, opened 2020-11-16
Implement isZeroReport for Search Console widgets.

benbowler:
## Summary

Add a new utility for Search Console widgets to detect empty data and show the empty component in this case.

Addresses issue #2244

## Relevant technical choices

I created the isZeroReport function in the `assets/js/modules/search-console/util/is-zero-report.js` file and exported it in the main utils for these widgets. Then I applied this function in every widget where search console data is displayed.

## Checklist


aaemnnosttv (changes requested):
Thanks @benbowler ! This is looking good with just a few changes to address.

Most of the feedback here is the same as for your other issue for the same utility for another module. In this case it's reasonable that these would be so similar and you probably worked on them around the same time. Likewise, it would make sense to address feedback that applies to both proactively whenever that comes up, although I would say it's pretty rare 😄 

Let me know if you have any other questions!

aaemnnosttv (approved):
Super, thanks!

Issue #2194, opened 2020-10-15
Refactor `ModulesList` component


felixarntz:
The `ModulesList` component should be refactored to a functional component using the datastore.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `ModulesList` component should be refactored to a functional component using the datastore and hooks.
    * It shouldn't use any legacy utility functions or globals anymore to get data.
    * The only legacy function that's okay to keep is `showErrorNotification` since there's no replacement yet.
* The logic handling the case where a module cannot be activated because of a missing dependency module should be removed - it should be unreachable anyway because such modules are already filtered out before.
* Its file should be renamed so that the file name matches the component name, and all import references should be updated accordingly.
* There should be a working Storybook story for the button under "Global".

## Implementation Brief

* Rename `assets/js/components/modules-list.js` to `assets/js/components/ModulesList.js`;
* Rename modules list import in the `assets/js/components/notifications/dashboard-setup-alerts.js` file to use the new file name;
* Edit the `assets/js/components/ModulesList.js` file:
  * Add a new `ModulesList` function that does the following:
    * Uses the `useSelect` hook to pull `modules` from the modules datastore;
    * Uses the `useState` hook to define `isSaving` state property;
    * Uses the `useCallback` hook to define `setupModuleClick` callback that replicates the logic defined in the legacy method:
      * Dispatch the `activateModule` action and grab `moduleReauthURL` from its response;
      * Call the `trackEvent` function like its done in the `activateOrDeactivateModule` function;
      * Redirect the browser to the `modlueReauthURL` received from the `activateModule` actions;
      * Don't call the `refreshAuthentication` function since it is handled automatically by module activation now.
    * Returns markup similar to the one generated by the legacy `render` method but with new variables received from hooks;
  * Remove the legacy `ModulesList` class;
* Add a new story to the "global" section of the storybook that renders the modules list component.

### Test Coverage

* Not needed

### Visual Regression Changes

* Not needed

## QA Brief

* Disconnect the Site Kit on your site if it has been already connected;
* Sign in again;
* Check the following:
  * [ ] Make sure you see `Congrats on completing the setup for Site Kit!` notification on the dashboard with a list of modules;
  * [ ] The modules list look the same as it looked before;
  * [ ] You can activate a module by clicking on the `Connect Service` link.

## Changelog entry

* N/A

aaemnnosttv:
[...]

Issue #2043, opened 2020-09-16
Display dashboard notification when user input is required


felixarntz:
Returning users that are already connected to Site Kit and have updated to the new version which exposes the user input flow should be prompted with a notification on the dashboard about the new feature and related capabilities. See #2042 for related logic.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a notification "Customize Site Kit to match your goals", which is displayed on the Site Kit dashboard if the user needs to still fill out the user input questions.
* The notification should look like [this](https://www.figma.com/file/f07vatCPHzubuJoXkgd5RU/SiteKit---Phase-1_-Onboarding?node-id=1098%3A16859).
* The notification should be displayed if:
    * ...the `featureFlags.userInput` is enabled (see #2038).
    * ...the JS `core/user` store's `getUserInputState()` selector returns anything but `completed`.
    * ...the notification is not stored as dismissed on the client (see below).
* The CTA button should be a link that points to the `googlesitekit-user-input` admin screen (see #2038).
* The Dismiss link should dismiss the notification client-side with a timeout of 3 hours.
* For storing the dismissal, let's not worry about a new (proper) API, we should be able to rely on the existing client-side code that stores notification dismissals in `sessionStorage`.

## Implementation Brief

* UserInputSettings notification:
  * Create `assets/js/components/notifications/UserInputSettings.js` file and appropriate functional component;
  * Implement the user input settings notification component using design provided in the AC;
    * Use `Notification` component with the following properties:
      * `dismissExpires` should be set to `10800` (3 * 60 * 60 seconds = 3 hours)
      * `ctaLink` should use a URL received from the `getAdminURL( 'googlesitekit-user-input' )` selector from the `core/site` datastore.
  * Don't render the notification (return `null`) if `getUserInputState()` selector from the `core/user` datastore returns `completed`.
* In the `assets/js/components/notifications/index.js` file:
  * Add a new condition that checks if `featureFlags.userInput.enabled` is truthy and adds `UserInputSettings` notification to the `googlesitekit.ErrorNotification` filter like its done for `DashboardAuthScopesAlert`.

## QA Brief

**To test, a development build must be run locally – release zips will not work.**

1. To verify this notification hasn't already been dismissed, clear your browser session storage.
2. Dispatch an action via browser console: `window.googlesitekit.data.dispatch( 'core/user' ).receiveUserInputState( 'missing' )`
[...]

PR #2375, opened 2020-11-17
Refactor ModulesList component and add stories for the same.

kostyalmm:
## Summary

Addresses issue #2194

## Relevant technical choices

* Refactor ModulesList component to functional component
* Add story for the same
* Update imports

## Checklist


package-lock.json
```diff
@@ -8515,11 +8515,6 @@
       "integrity": "sha512-YJhbp0sz3egFFKl3BcCNPQKzuGFOP4PACcsifhK6ROGnJUW9ViYLuLybQ9GQZm7Zejy3tkGuiXYMq3GiyGkU4g==",
       "dev": true
     },
-    "@types/mime-types": {
```
eugene-manuilov, line 8518: @kostyalmm seems like `package-lock.json` has been updated without dependencies change in the `package.json` file. Could you please revert changes in the `package-lock.json`?

assets/js/components/ModulesList.js
```diff
@@ -0,0 +1,174 @@
+/**
+ * ModulesList component.
+ *
+ * Site Kit by Google, Copyright 2019 Google LLC
```
eugene-manuilov, line 4:
```suggestion
 * Site Kit by Google, Copyright 2020 Google LLC
```

assets/js/components/ModulesList.js
```diff
@@ -0,0 +1,174 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * External dependencies
+ */
+import { map } from 'lodash';
```
eugene-manuilov, line 22:
```suggestion
import map from 'lodash/map';
```

assets/js/components/ModulesList.js
```diff
@@ -0,0 +1,174 @@
...
+		} catch ( err ) {
+			showErrorNotification( GenericError, {
+				id: 'setup-module-error',
+				title: __( 'Internal Server Error', 'google-site-kit' ),
+				description: err.message,
+				format: 'small',
+				type: 'win-error',
+			} );
+		}
+	}, [ activateModule ] );
```
eugene-manuilov, line 73: @kostyalmm it misses the following from the IB:

```
Call the trackEvent function like its done in the activateOrDeactivateModule function;
```

assets/js/components/ModulesList.js
```diff
@@ -0,0 +1,174 @@
...
+			.filter( ( slug ) => modulesData.hasOwnProperty( slug ) )
+			.reduce( ( acc, slug ) => ( { ...acc, [ slug ]: modulesData[ slug ] } ), {} )
+		: modulesData;
+
+	// Filter out internal modules.
+	const modules = Object.values( moduleObjects ).filter( ( module ) => ! module.internal );
+
+	// Map of slug => name for every module that is active and completely set up.
+	const completedModuleNames = modules
+		.filter( ( module ) => module.active && module.setupComplete )
```
eugene-manuilov, line 87: Modules, received from the `core/modules` datastore, don't have `setupComplete` property. Use `connected` instead.

assets/js/components/ModulesList.js
```diff
@@ -0,0 +1,174 @@
...
+
+	return (
+		<div className="googlesitekit-modules-list">
+			{ map( sortedModules, ( module ) => {
+				let blockedByParentModule = false;
+				let parentBlockerName = '';
+				const {
+					slug,
+					name,
+					active: isConnected,
```
eugene-manuilov, line 107: `active` means that the module is active but not connected (a user hasn't configured it yet). We need to use the `connected` property to determine if the module connected or not.

eugene-manuilov (changes requested):
Hi @kostyalmm! Thanks for implementing it. I have found some issues that need to be addressed before we can merge it. Could you please address it?

assets/js/components/ModulesList.js
```diff
@@ -0,0 +1,180 @@
...
+
+	return (
+		<div className="googlesitekit-modules-list">
+			{ map( sortedModules, ( module ) => {
+				let blockedByParentModule = false;
+				let parentBlockerName = '';
+				const {
+					slug,
+					name,
+					connected,
```
eugene-manuilov, line 113: @kostyalmm  sorry for the incorrect suggestion at the beginning, but here we need to rely on the `active` property instead of `connected`.

eugene-manuilov (changes requested):
Thanks, @kostyalmm! Mostly LGTM. Just found one thing to change.

assets/js/components/ModulesList.js
kostyalmm: Are you sure? 🤔 `connected` seems to be the correct property here according to me as the [old component has](https://github.com/google/site-kit-wp/blob/develop/assets/js/components/modules-list.js#L119):
```
setupComplete: isConnected,
```

By this logic, `connected` seems like the right property. What do you think?

eugene-manuilov: @kostyalmm as far as I understand it, `connected` means that the module is activated and the user has configured which is not what we need here. We need to know whether a module is activated and if it's not, let the user activate it by clicking on the "Connect" link.

@aaemnnosttv am I right?

aaemnnosttv: @eugene-manuilov `setupComplete` for the legacy module format is equal to `module.active && module.connected`:
https://github.com/google/site-kit-wp/blob/27d4df7bbfbf6c7e59310ee9338e2a6155870268/includes/Core/Modules/Modules.php#L159

So in this case, it's not a 1-to-1 equivalent logically but it is the closest thing for `setupComplete`.

With that said, the consuming logic would _also_ need to check for `active` as well whereas before it was included in the definition of `setupComplete`.

cc @kostyalmm

eugene-manuilov: Thanks, @aaemnnosttv! @kostyalmm let's add `active` and use `( active && connected )` to determine whether a module is connected or not.

eugene-manuilov (approved):
LGTM! Thanks, @kostyalmm!

stories/modules-list.stories.js
```diff
@@ -0,0 +1,43 @@
...
+ */
+import ModulesList from '../assets/js/components/ModulesList';
+import { WithTestRegistry } from '../tests/js/utils';
+import { STORE_NAME as CORE_MODULES } from '../assets/js/googlesitekit/modules/datastore/constants';
+import FIXTURES from '../assets/js/googlesitekit/modules/datastore/fixtures';
+
+storiesOf( 'Global', module )
+	.add( 'ModulesList', () => {
+		const setupRegistry = ( { dispatch } ) => {
+			dispatch( CORE_MODULES ).receiveGetModules( FIXTURES );
```
aaemnnosttv, line 35: Please use the `provideModules` utility instead.

stories/notifications.stories.js
```diff
@@ -11,24 +11,35 @@ import { __ } from '@wordpress/i18n';
...
-			dismiss={ __( 'OK, Got it!', 'google-site-kit' ) }
-			format="large"
-			type="win-success"
-		>
-			<ModulesList />
-		</Notification>
-	) )
+	.add( 'Module Setup Complete', () => {
+		const setupRegistry = ( { dispatch } ) => {
+			dispatch( CORE_MODULES ).receiveGetModules( FIXTURES );
```
aaemnnosttv, line 24: See above.

stories/notifications.stories.js
```diff
@@ -11,24 +11,35 @@ import { __ } from '@wordpress/i18n';
...
+		};
+
+		return (
+			<WithTestRegistry callback={ setupRegistry } >
+				<Notification
+					id="notification-id"
+					title={ __( 'Congrats on completing the setup for Analytics!', 'google-site-kit' ) }
+					handleDismiss={ () => {} }
+					winImage={ `${ global._googlesitekitLegacyData.admin.assetsRoot }images/rocket.png` }
```
aaemnnosttv, line 33: Please update the URL, using an imported image to be consistent with other stories as this recently changed.

stories/notifications.stories.js
```diff
@@ -11,24 +11,35 @@ import { __ } from '@wordpress/i18n';
...
+		const setupRegistry = ( { dispatch } ) => {
+			dispatch( CORE_MODULES ).receiveGetModules( FIXTURES );
+		};
+
+		return (
+			<WithTestRegistry callback={ setupRegistry } >
+				<Notification
+					id="notification-id"
+					title={ __( 'Congrats on completing the setup for Analytics!', 'google-site-kit' ) }
+					handleDismiss={ () => {} }
```
aaemnnosttv, line 32: Please remove as this isn't a prop of `Notification`.

assets/js/components/ModulesList.js
```diff
@@ -0,0 +1,182 @@
...
+	/**
+	 * Handle setup module click event.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @param {string} slug Module slug.
+	 */
+	const handleSetupModule = useCallback( async ( slug ) => {
+		try {
+			const { response } = await activateModule( slug );
```
aaemnnosttv, line 61: The `try/catch` here is no longer working as it was before as neither `activateModule` or `trackEvent` will throw. Instead, `error` should come from the return value of `activateModule`, just as it does with our other async actions.

aaemnnosttv (changes requested):
Thanks @kostyalmm ! There's still a few changes to make here as the new component is still too close to the old implementation. In particular, there is a substantial amount of code to remove/update related to the module dependencies which was to be removed per the ACs.

Regarding the changes to the VRT reference images – the new files look identical, can you explain why this was necessary? The refactoring here shouldn't have required updating these images.

assets/js/components/ModulesList.js
```diff
@@ -0,0 +1,182 @@
...
+				// Check if required modules are active.
+				if ( 0 < dependencies.length ) {
+					dependencies.forEach( ( requiredModule ) => {
+						if ( completedModuleNames[ requiredModule ] ) {
+							blockedByParentModule = true;
+							parentBlockerName = completedModuleNames[ requiredModule ];
+						}
+					} );
+				}
+
```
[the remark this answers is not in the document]
kostyalmm: I removed the filter but I kept the `blockedByParentModule `. Let me explain why:
As I understand, the `ModulesList` component will show a list of modules that can be activated. The modules which can't be activated because their dependency is not activated, should show a greyed out message asking to activate the dependency first. 

I made a few changes to how `blockedByParentModule` was set and the surrounding code. 

I'd like to request you to pull the latest changes, start storybook and go to the `ModulesList` story. Once you go you will see all the modules including `TagManager`  and `Optimize` since `provideModules` activates `Analytics` by default.
Now go to `fixtures.json` and set `active` for `Analytics` to false, wait for storybook to reload, and you will see that `TagManager` and `Optimize` can't be activated and shows the desired message. 

Do let me know if this is the desired requirement for this Component. If not, then I will remove `blockedByParentModule ` as you suggested above. In that case I have two questions:
1. Will the modules with *unactivated* dependencies be shown by default?
2. Will the aforementioned modules show the greyed out message asking to activate `dependency`? Or will they be allowed to be activated unconditionally?
3. Will the aforementioned modules be not shown?

stories/notifications.stories.js
kostyalmm: It seems that image imports can't be done like 
```js
winImage={ global._googlesitekitLegacyData.admin.assetsRoot + rocketImage }
```

as it results in a src 
```
http://localhost:9001/assets/static/media/rocket.3e64a093.png
```
which gives a 404.

Am I doing something wrong here? The other image files are referenced directly by the file name in the same `Notifications` story scope.

kostyalmm:
Regarding the VRT changes, the change is there because earlier the `ModulesList` component was using `getModulesData` which in turn used `global._googlesitekitLegacyData`, this has been now changed to use the datastore. In the changes before the requested changes were made by @aaemnnosttv , the stories were supplied with `FIXTURES` which had different data.

After making the requested changes, there are still difference in the VRT because it now shows all the modules rather than showing the ones which don't have any dependency. <= See [this comment](https://github.com/google/site-kit-wp/pull/2375#discussion_r528721724) for more info

stories/notifications.stories.js
aaemnnosttv: Oh sorry, my mistake. This changed recently in the PR for #2043 and it just hasn't been merged yet but I thought it had. You're right, this should be as you have it now 👍

assets/js/components/ModulesList.js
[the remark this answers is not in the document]
aaemnnosttv: Regarding my comment above, I wasn't suggesting we remove the filter, I was only suggesting we update that part (as it's relevant to the ACs) to be a bit more clear either by using separate variables or inline it perhaps to not have a variable called "sorted" which also filters. 

Thanks for the additional context. It's important to understand that the issue was defined specifically with the removal of `blockedByParentModule` behavior as mentioned in the AC because it is not possible to work otherwise with the current/legacy version of the component. To do otherwise means that we would be introducing a behavioral change to the component where it is not called for and not currently present. 

While this may have been a bug and I can appreciate your desire to fix it, a conflict with the AC should be raised with @felixarntz as soon as possible. As it is, I think the issue is clear about what should be done but we can certainly wait for @felixarntz to give the final word.

.storybook/storybook-data.js
```diff
@@ -177,7 +191,6 @@ module.exports = [
 			options: {
 				hierarchyRootSeparator: '|',
 				hierarchySeparator: {},
-				delay: 1000,
```
aaemnnosttv, line 180: This was probably a mistake but please don't change other VRT scenarios.

.storybook/storybook-data.js
```diff
@@ -167,6 +167,20 @@ module.exports = [
 			},
 		},
 	},
+	{
+		id: 'global--modules-list',
+		kind: 'Global',
+		name: 'Modules List',
+		story: 'Modules List',
```
aaemnnosttv, line 174: This issue doesn't call for a new VRT scenario to be added since this component is already covered by other existing scenarios, it should only have the added Storybook story.

package-lock.json
```diff
@@ -10834,11 +10834,6 @@
           "resolved": "https://registry.npmjs.org/prelude-ls/-/prelude-ls-1.2.1.tgz",
           "integrity": "sha512-vkcDPrRZo1QZLbn5RLGPpg/WmIQ65qoWWhcGKf/b5eplkkarX0m9z8ppCat4mlOqUsWpyNuYgO3VRyrYHSzX5g=="
         },
-        "prettier": {
```
aaemnnosttv, line 10837: Please don't include changes to the `package-lock.json` for issues which don't include dependency changes.

aaemnnosttv (commented):
> After making the requested changes, there are still difference in the VRT because it now shows all the modules rather than showing the ones which don't have any dependency.

I believe it should continue to only show modules which have no dependencies as before but let's wait for @felixarntz to confirm. See [my comment](https://github.com/google/site-kit-wp/pull/2375#discussion_r528823006) on the thread above. There should be no visual changes needed to the existing scenarios let's see how to reconcile those rather than updating references and remove the added scenario.

assets/js/components/ModulesList.js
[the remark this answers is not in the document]
felixarntz: @kostyalmm @aaemnnosttv Indeed the ability of showing modules with inactive dependencies greyed out is legacy behavior, which used to be the case, but was intentionally changed a while ago. Unfortunately the code back then wasn't fully removed which is probably what created the confusion here. The way things currently look after your latest commits should be good.

assets/js/components/ModulesList.js
```diff
@@ -0,0 +1,149 @@
...
+		? moduleSlugs
+			.filter( ( slug ) => modulesData.hasOwnProperty( slug ) )
+			.reduce( ( acc, slug ) => ( { ...acc, [ slug ]: modulesData[ slug ] } ), {} )
+		: modulesData;
+
+	// Filter out internal modules and remove modules with dependencies.
+	const modules = Object.values( moduleObjects ).filter( ( module ) => ! module.internal && 0 === module.dependencies.length );
+
+	// Sort modules and exclude those that required other modules to be set up.
+	// Logic still in place below in case we want to add blocked modules back.
```
aaemnnosttv, line 90: This comment is no longer relevant.
```suggestion
```

aaemnnosttv (commented):
LGTM, just one tiny thing to clean up from the old component.

assets/js/components/ModulesList.js
```diff
@@ -0,0 +1,146 @@
...
+		: modulesData;
+
+	// Filter out internal modules and remove modules with dependencies.
+	const modules = Object.values( moduleObjects ).filter( ( module ) => ! module.internal && 0 === module.dependencies.length );
+	const sortedModules = modules
+		.sort( ( module1, module2 ) => module1.sort - module2.sort );
+
+	return (
+		<div className="googlesitekit-modules-list">
+			{ sortedModules.map( ( module ) => {
```
aaemnnosttv, line 93:
```suggestion
	const modules = Object.values( moduleObjects )
		.filter( ( module ) => ! module.internal && 0 === module.dependencies.length )
		.sort( ( module1, module2 ) => module1.sort - module2.sort );

	return (
		<div className="googlesitekit-modules-list">
			{ modules.map( ( module ) => {
```

aaemnnosttv (approved):
LGTM 👍

Issue #2289, opened 2020-10-29
Inform users on pre-5.0 versions of WP about limitations


felixarntz:
Site Kit should display a notification to users that are on pre-5.0 versions of WordPress and inform them that they are missing out on certain features, due to those older WordPress versions not offering the necessary infrastructure.

An example of that is JS translations support (see #2171).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The compatibility checks during setup should include a new "check" that displays a (non-blocking) warning if the site's WP version is lower than 5.0.
* The message should have the following copy: _Looks like you're using a version of WordPress that's older than 5.0. You can still install and use Site Kit, but some of its features might not work (for example translations)._

## Implementation Brief

* Add the current WordPress version to the `$data` array in the `inline_js_base_data` method of the `Authentication` class;
* Add a new check to the checks array in the `assets/js/components/setup/compatibility-checks.js` file It should read WP version from the `global._googlesitekitBaseData` object and throw `ERROR_OLD_VERSION` error if the current version is lower than 5;
* Update the `renderError` function of the `CompatibilityChecks` component to render a message for `ERROR_OLD_VERSION` error. Use the message from the AC.

### Test Coverage

* Not needed.

### Visual Regression Changes

* A new warning notice should appear on the splash screen.

## QA Brief

* Set up a new instance of WordPress 4.9 or older;
* Install and activate the Google Site Kit plugin;
* Go to the splash screen and make sure you see the new warning notice that contains a message from the AC.

## Changelog entry

* Display informational warning before setup about limited support for WordPress versions before 5.0.

felixarntz:
@eugene-manuilov The IB here is not quite correct: The version check should be incorporated as part of the compatibility checks which happen on the splash screen, see `CompatibilityChecks` component.

In this case, we don't need to run any logic for the check, but it should still be implemented as part of that component, also visually. An admin notice would not be the same visual outcome. In order to get the WordPress version in JS, we could do something simple as exposing it with the site info on the global and e.g. add a `getWordPressVersion()` selector to the `core/site` store.

eugene-manuilov:
@felixarntz got it. Unfortunately, we can't use `getWordPressVersion` selector because of the way how other checks are configured right now. I have updated IB to use `global._googlesitekitBaseData` directly to get the WordPress version. Let me know if it works.

felixarntz:
[...]

PR #2376, opened 2020-11-17
Enhancement/2289 pre5 0 limitations

eugene-manuilov:
## Summary

Addresses issue #2289

## Checklist


includes/Core/Authentication/Authentication.php
```diff
@@ -660,8 +660,11 @@ private function handle_oauth() {
 	 * @return array Filtered $data.
 	 */
 	private function inline_js_base_data( $data ) {
+		global $wp_version;
+
 		$data['isOwner']             = $this->owner_id->get() === get_current_user_id();
 		$data['isFirstAdmin']        = $data['isOwner'] || ( ! $this->owner_id->get() && current_user_can( Permissions::MANAGE_OPTIONS ) );
+		$data['isWP5.0+']            = version_compare( $wp_version, '5.0.0', '>=' );
```
aaemnnosttv, line 667: This is a bit unnecessarily specific. Why not expose the version to the client in a way that's just as easy to use but more flexible? We do something similar in E2E as it is where one of the utilities requires WP5 as well. I think using a similar strategy here would be just as easy and more valuable. See https://github.com/google/site-kit-wp/blob/aa6cdd0792898920dd626dd71b2d722663a50756/tests/e2e/mu-plugins/e2e-rest-core-version.php#L27-L34

Perhaps the key could be `wpVersion` with a shape like `{ full, major, minor }` ?

Then we would only need to check `wpVersion.major < 5` 😄 

Also note we don't need to touch the `$wp_version` global to get the same value.

assets/js/components/setup/compatibility-checks.js
```diff
@@ -42,6 +42,7 @@ const ERROR_FETCH_FAIL = 'check_fetch_failed';
 const ERROR_TOKEN_MISMATCH = 'setup_token_mismatch';
 const ERROR_GOOGLE_API_CONNECTION_FAIL = 'google_api_connection_fail';
 const ERROR_AMP_CDN_RESTRICTED = 'amp_cdn_restricted';
+const ERROR_OLD_VERSION = 'old_wordpress';
```
aaemnnosttv, line 45: The name here isn't that important but I think we can be a bit more clear with the intent with a more self-documenting name here since "old" is a bit vague.

How about this?
```suggestion
const ERROR_WP_PRE_V5 = 'wp_pre_v5';
```

assets/js/components/setup/compatibility-checks.js
```diff
@@ -158,61 +165,67 @@ export default class CompatibilityChecks extends Component {
...
+					<p>
+						{ ! installed && __( 'Looks like this may be a staging environment. If so, you’ll need to install a helper plugin and verify your production site in Search Console.', 'google-site-kit' ) }
+						{ installed && __( 'Looks like this may be a staging environment and you already have the helper plugin. Before you can use Site Kit, please make sure you’ve provided the necessary credentials in the Authentication section and verified your production site in Search Console.', 'google-site-kit' ) }
+						{ ' ' }
+						<Link
+							{ ...this.helperCTA() }
+							inherit
+						/>
+					</p>
+				);
```
aaemnnosttv, line 178: It looks like the formatting of the markup here was just updated to be wrapped with the `()` like we do elsewhere (I see `Fragment`s removed below as well) but just wanted to confirm there are no other significant changes here that might be missed, other than the addition of the new case?

assets/js/components/setup/compatibility-checks.test.js
```diff
@@ -84,6 +84,9 @@ describe( 'CompatibilityChecks', () => {
 	it( 'should make API requests to "setup-checks, health-checks and AMP Project test URL', async () => {
 		const token = 'test-token-value';
 
+		global._googlesitekitBaseData = global._googlesitekitBaseData || {};
+		global._googlesitekitBaseData[ 'isWP5.0+' ] = true;
```
aaemnnosttv, line 88: We should add the definition to `.storybook/config.js` as well.

aaemnnosttv (changes requested):
Thanks @eugene-manuilov, this is looking good just a few small things to change, and one question for you.

assets/js/components/setup/compatibility-checks.js
eugene-manuilov: Yes, I just fixed the formatting. Nothing else has changed.

includes/Core/Authentication/Authentication.php
eugene-manuilov: Yep, i like it. Updated.

assets/js/components/setup/compatibility-checks.test.js
eugene-manuilov: We don't need it anymore after I have reworked it using your suggestion.

felixarntz (approved):
Great!

Issue #2331, opened 2020-11-05
Expose date picker in Site Kit header bar and make it sticky


felixarntz:
A long-standing UX issue with the plugin has been the date pickers on the various screens since they are only shown on top of the page, which makes it cumbersome to adjust the date when looking at metrics further down on the page. This issue aims at improving this.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The Site Kit header bar should be modified so that it's sticky, i.e. that it remains at the top of the screen when scrolling down. It should remain right below the WordPress admin bar in case the admin bar is sticky too.
    * Note that this might require some detailed look at admin bar scroll behavior across various viewports.
* On all pages where there is a date-picker, that date picker should be displayed as part of the Site Kit header bar (enabled via React component prop):
    * Site Kit dashboard
    * single URL details view
    * AdSense module screen
    * Analytics module screen
    * Search Console module screen
* The date picker UI and the new UI overall should look like in [this Figma file](https://www.figma.com/file/XxCBceEOfDsljYv9DNaJdS/Version-1---Sticky-header%2C-no-animation?node-id=0%3A1).

## Implementation Brief

- Using the file: `/assets/sass/components/global/_googlesitekit-header.scss`
  - Update styles for `.googlesitekit-header` to use `position: sticky;` with the `top` value taking into account the WordPress admin bar across all breakpoints.
- Using the file: `/assets/js/components/Header.js`
  - Update component to accept the `children` prop which will be outputted before `UserMenu`
- Using the file: `/assets/js/components/dashboard/dashboard-app.js`
  - Move the `DateRangeSelector` component to being a prop for the `Header` component
- Using the file: `/assets/js/components/dashboard-details/dashboard-details-app.js`
  - Import the `DateRangeSelector` component and use it as a prop for the `Header` component only if `currentEntityURL` is defined.
- Using the file: `/assets/js/components/dashboard-details/DashboardDetailsEntityView.js`
  - Remove the `DateRangeSelector` component
- Using the file: `/assets/js/modules/adsense/components/dashboard/AdSenseDashboardWidget.js`
  - Move the `PageHeaderDateRange` component to be a prop for the `Header` component
- Using the file: `/assets/js/modules/analytics/components/dashboard/AnalyticsDashboardWidget.js`
  - Move the `PageHeaderDateRange` component to be a prop for the `Header` component
[...]

PR #2379, opened 2020-11-17
Sticky header date picker

benbowler:
## Summary

I brought the `DateRangeSelector` component into the app-wide header rather than in the individual widgets as well as making the header sticky during scroll.
Addresses issue #2331  

## Relevant technical choices

I initially used the existing Select component for this dropdown but this caused issues with the fixed position header - the MenuList component failed to scroll with the page. For this reason, I replicated the user-menu dropdown custom menu which now scrolls correctly when open.

## Checklist


assets/js/components/DateRangeSelector.js
```diff
@@ -19,40 +19,92 @@
...
  * Internal dependencies
  */
 import Data from 'googlesitekit-data';
-import { Option, Select } from '../material-components';
+import DateRangeIcon from '../../svg/date-range.svg';
+import Menu from './menu';
 import { getAvailableDateRanges } from '../util/date-range';
 import { STORE_NAME as CORE_USER } from '../googlesitekit/datastore/user/constants';
+import Button from './button';
+import { map } from 'lodash';
```
tofumatt, line 33: We should probably have ESLint rules for this, but we group "external" dependencies up in the "External dependencies" group.

assets/js/components/DateRangeSelector.js
```diff
@@ -19,40 +19,92 @@
...
-	const onChange = useCallback( ( index, item ) => {
-		setDateRange( item.dataset.value );
-	}, [ ranges ] );
+
+	const [ menuOpen, toggleMenu ] = useState( false );
+	const menuButtonRef = useRef();
+	const menuRef = useRef();
+
+	useEffect( () => {
+		const handleMenuClose = ( e ) => {
```
tofumatt, line 46: Can we use `event` here instead? Single-letter-variable names are always harder to read. We can spare the bytes 😉

assets/js/components/DateRangeSelector.js
```diff
@@ -19,40 +19,92 @@
...
 
 function DateRangeSelector() {
 	const ranges = Object.values( getAvailableDateRanges() );
 	const dateRange = useSelect( ( select ) => select( CORE_USER ).getDateRange() );
 	const { setDateRange } = useDispatch( CORE_USER );
-	const onChange = useCallback( ( index, item ) => {
-		setDateRange( item.dataset.value );
-	}, [ ranges ] );
+
+	const [ menuOpen, toggleMenu ] = useState( false );
```
tofumatt, line 41: Let's call this `setMenuOpen` rather than `toggleMenu`. A function named `toggleMenu()` implies to me that it'll switch it from the current state to the inverse state, but that's not what this does 🙂

assets/js/components/DateRangeSelector.js
```diff
@@ -19,40 +19,92 @@
...
+			global.removeEventListener( 'mouseup', handleMenuClose );
+			global.removeEventListener( 'keyup', handleMenuClose );
+		};
+	}, [] );
+
+	const handleMenu = useCallback( () => {
+		toggleMenu( ! menuOpen );
+	}, [ menuOpen ] );
+
+	const handleMenuItemSelect = useCallback( ( index, e ) => {
```
tofumatt, line 71: Let's use `event` here as well instead of `e`.

assets/js/components/DateRangeSelector.js
```diff
@@ -19,40 +19,92 @@
...
+			setDateRange( ranges[ index ].slug );
+			toggleMenu( false );
+		}
+	}, [ handleMenu ] );
+
+	const currentDateRangeLabel = ranges.reduce( ( acc, range ) => {
+		return range.slug === dateRange ? range : acc;
+	}, {} ).label;
+
+	const menuItems = map( ranges, 'label' );
```
tofumatt, line 85: Instead of using `lodash.map`, let's use `Array.map` instead.

Historically there was a lot of lodash use in the app but moving forward we're preferring ES6 built-ins when possible. Lodash has some fancy helpers that are worth using, but for stuff like `map` I think `ranges.map( (range) => range.label )` is more immediately intuitive and doesn't require knowing the Lodash API—or at least as much of it 🙂

assets/js/components/DateRangeSelector.js
```diff
@@ -19,40 +19,92 @@
...
+				aria-controls="date-range-selector-menu"
+			>
+				{ currentDateRangeLabel }
+			</Button>
+			<Menu
+				ref={ menuRef }
+				menuOpen={ menuOpen }
+				menuItems={ menuItems }
+				onSelected={ handleMenuItemSelect }
+				id="date-range-selector-menu" />
```
tofumatt, line 106: Not a big deal, and I'm sad ESLint doesn't care, but we usually stick the self-closing `/>` on a newline.

assets/js/components/Header.js
```diff
@@ -31,7 +36,7 @@ import ErrorNotification from '../components/notifications/error-notification';
 import { STORE_NAME as CORE_USER } from '../googlesitekit/datastore/user/constants';
 const { useSelect } = Data;
 
-const Header = () => {
+const Header = ( { children } ) => {
```
tofumatt, line 39: The IB/issue doesn't mention any VRT changes expected, but I see this in all the headers:

Screenshot 2020-11-18 at 13 40 17

The dropdown indicator is gone. I can't seem to access the Figma file to see the mocks—is that intentional? It seems odd to remove the indication that it's clickable, but then it's similar to the style on google.com so I guess it's fine. Just making sure.

stories/header-date-selector.stories.js
```diff
@@ -0,0 +1,42 @@
+/**
+ * External dependencies
+ */
+import { storiesOf } from '@storybook/react';
+/**
```
tofumatt, line 5: Very minor, but a newline here is usually our style.

stories/header-date-selector.stories.js
```diff
@@ -0,0 +1,42 @@
...
```
tofumatt, line 1: It'd be good to add the Site Kit header here. Eg:

Screenshot 2020-11-18 at 13 52 30

tofumatt (changes requested):
Just a few questions and comments. A few small things to tweak but mostly wondering about the VRT changes.

Was that old arrow pointing down in the VRT headers because there was no date range text but the select still rendered? Eg: was the old VRT wrong? Or is the change intentional? If so, seems fine to me, but I just can't confirm using the figma link.

Let me know and then assign this one back to me for another review once the tweaks are made, should be good to go after that. 🙂

assets/js/components/DateRangeSelector.js
benbowler: Yes updated both `e` vars

assets/js/components/DateRangeSelector.js
benbowler: Yes I like this terminology.

assets/js/components/DateRangeSelector.js
benbowler: Ok, that makes sense. I've switched to `Array.map`

assets/js/components/Header.js
benbowler: I'll send you a screenshot in Slack, the new designs remove this arrow and increase the size of the artwork to 36px square.

assets/js/components/DateRangeSelector.js
```diff
@@ -19,40 +19,93 @@
...
-				</Option>
-			) ) }
-		</Select>
+		<div className="googlesitekit-date-range-selector">
+			<Button
+				ref={ menuButtonRef }
+				className="googlesitekit-header__date-range-selector-menu mdc-button--dropdown"
+				text
+				onClick={ handleMenu }
+				icon={ ( <DateRangeIcon width="18" height="20" /> ) }
```
tofumatt, line 94:
```suggestion
				icon={ <DateRangeIcon width="18" height="20" /> }
```

tofumatt (dismissed):

tofumatt (dismissed):

assets/svg/date-range.svg
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="18px" height="20px" viewBox="0 0 18 20" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <title>date-range</title>
+    <g id="Page-1" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="date-range" fill="#5B5B61">
+            <path d="M6,9 L4,9 L4,11 L6,11 L6,9 Z M10,9 L8,9 L8,11 L10,11 L10,9 Z M14,9 L12,9 L12,11 L14,11 L14,9 Z M16,2 L15,2 L15,0 L13,0 L13,2 L5,2 L5,0 L3,0 L3,2 L2,2 C0.89,2 0.01,2.9 0.01,4 L0,18 C0,19.1 0.89,20 2,20 L16,20 C17.1,20 18,19.1 18,18 L18,4 C18,2.9 17.1,2 16,2 Z M16,18 L2,18 L2,7 L16,7 L16,18 Z" id="Shape"></path>
+        </g>
+    </g>
+</svg>
```
aaemnnosttv, line 9: There is a bit of unnecessary structure here but it looks like most of this is already being cleaned up by SVGO during import

image

However, we should change the `fill` to `fill="currentColor"` to inherit the current `color` rather than hardcoding it. Then just make sure the `color` for that is set correctly via CSS, if it isn't already.

stories/header-date-selector.stories.js
```diff
@@ -0,0 +1,61 @@
...
+storiesOf( 'Global', module )
+	.add( 'Plugin Header with Date Selector', () => {
+		const setupRegistry = ( { dispatch } ) => {
+			dispatch( CORE_SITE ).receiveSiteInfo( {
+				usingProxy: true,
+				proxySetupURL: 'https://sitekit.withgoogle.com/site-management/setup/',
+				proxyPermissionsURL: 'https://sitekit.withgoogle.com/site-management/permissions/',
+				referenceSiteURL: 'http://example.com',
+				siteName: 'My Site Name',
+			} );
```
aaemnnosttv, line 42: Please use `provideSiteInfo` instead (defaults should be fine here I think)

stories/header-date-selector.stories.js
```diff
@@ -0,0 +1,61 @@
...
+			dispatch( CORE_USER ).receiveGetAuthentication( {
+				authenticated: true,
+				requiredScopes: [],
+				grantedScopes: [],
+			} );
```
aaemnnosttv, line 47: Please use `provideUserAuthentication` instead 😄

stories/header-date-selector.stories.js
```diff
@@ -0,0 +1,61 @@
...
+			<WithTestRegistry callback={ setupRegistry }>
+				<Header>
+					<DateRangeSelector />
+				</Header>
+			</WithTestRegistry>
+		);
+	}, {
+		options: {
+			delay: 3000, // Wait for image to load.
+		},
```
aaemnnosttv, line 60: Some other stories have this but I don't think it's actually needed anymore since we wait for images to load before signaling Backstop that it's ready to go.

stories/header-date-selector.stories.js
```diff
@@ -0,0 +1,61 @@
+/**
+ * Header with Date Selector stories.
```
aaemnnosttv, line 2: I don't think separate stories are needed for this, let's combine them with `stories/header.stories.js` and we can add a decorator for receiving common setup into the registry instead. Look for usages of `addDecorator` in stories for an example 👍

aaemnnosttv (changes requested):
Thanks @benbowler – this is looking really nice.

First and foremost, the E2E tests are all failing – it seems like it's looking for `mdc-select`-specific classes which are no longer present since the dropdown was reimplemented to not  use that, so it should be fairly straightforward to fix.

Looking at this in the browser, there is a bit of a disparity between the new date range dropdown and the user menu as shown below:

| Date range dropdown | User Menu |
| --- | --- |
| image | image |

Note the positioning of the menu – I would expect these to be consistent. It's also a bit odd for the menu to cover the button when being shown, so let's make the date range's dropdown match the positioning of the user menu relative to the button.

I also noticed a bit of odd stacking in the tablet range before the user menu is collapsed.
image
image

Would be nice if we could improve this (perhaps collapsing the user menu sooner?) but not a big deal I think as it's a bit of an edge case.

assets/js/components/Header.js
```diff
@@ -66,4 +73,8 @@ const Header = () => {
 	);
 };
 
+Header.propTypes = {
+	children: PropTypes.element,
```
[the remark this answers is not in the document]
benbowler: Done.

assets/js/components/DateRangeSelector.js
```diff
@@ -19,40 +19,93 @@
...
-			onEnhancedChange={ onChange }
-			value={ dateRange }
-		>
-			{ ranges.map( ( { slug, label } ) => (
-				<Option key={ slug } value={ slug }>
-					{ label }
-				</Option>
-			) ) }
-		</Select>
+		<div className="googlesitekit-date-range-selector">
```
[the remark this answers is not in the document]
benbowler: Good shout. I've managed to remove a good few lines after working through each class and seeing what's needed. I still need to reference `.googlesitekit-dropdown-menu` in the file because this ticket included reducing the font-weight on the dropdown label and increasing the image size, both of which I have to overwrite here.

assets/sass/components/global/_googlesitekit-header.scss
```diff
@@ -51,6 +62,60 @@
...
 
+	.mdc-button__label {
+		font-weight: $fw-primary-normal;
+	}
+
+	.googlesitekit-date-range-selector {
+		display: inline-block;
+		margin-right: 8px;
+		overflow: visible;
+		position: relative;
```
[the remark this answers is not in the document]
benbowler: Removed in refactor for above.

assets/svg/date-range.svg
benbowler: Updated.

stories/header-date-selector.stories.js
benbowler: Updated along with the existing header story.

stories/header-date-selector.stories.js
benbowler: Updated along with the existing header component.

stories/header-date-selector.stories.js
benbowler: Combined using `addDecorator`.

stories/header-date-selector.stories.js
benbowler: Removed.

assets/sass/components/global/_googlesitekit-header.scss
```diff
@@ -51,6 +62,60 @@
 		}
 	}
 
+	.mdc-button__label {
+		font-weight: $fw-primary-normal;
+	}
+
+	.googlesitekit-date-range-selector {
```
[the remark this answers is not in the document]
benbowler: Since I've simplified the styles and removed all styles from the mdc selector styles all of the changes are specific changes to match the native elements to the Figma designs specifically for the header. Let me know if you still feel this file needs to be split.

benbowler:
Ready for you to take a look @aaemnnosttv, my E2E test chance looks good so far on Travis. All other comments resolved and answered inline.

aaemnnosttv:
Thanks @benbowler, I'll take a look shortly. It seems that there is a merge conflict now, would you please resolve this first?

benbowler:
> Thanks @benbowler, I'll take a look shortly. It seems that there is a merge conflict now, would you please resolve this first?

I've fixed the merge conflict with the renamed `DashboardApp` component.

aaemnnosttv:
@benbowler looks like [VRT is failing now](https://github.com/google/site-kit-wp/pull/2379/checks?check_run_id=1447248272), can you take a look?

_Users_aaemnnosttv_Downloads_vrt-report%20(1)_html_report_index html

benbowler:
I've updated the visual tests based on my new breakpoint changes.

stories/header.stories.js
```diff
@@ -1,39 +1,66 @@
...
+import DateRangeSelector from '../assets/js/components/DateRangeSelector';
+import { createTestRegistry, provideSiteInfo, provideUserAuthentication, WithTestRegistry } from '../tests/js/utils';
+
+function Setup( props ) {
+	return (
+		<WithTestRegistry { ...props }>
+			<Header>
+				{ props.children }
+			</Header>
+		</WithTestRegistry>
```
aaemnnosttv, line 37: Let's not use this abstraction since we want to see how the `Header` component is used in each story and it isn't worth extracting a component just for the test registry so please inline below.

assets/sass/components/global/_googlesitekit-header.scss
```diff
@@ -51,7 +66,68 @@
...
+		@media (min-width: $bp-wpAdminBarTablet) {
+			margin-right: 24px;
+		}
+
+		svg {
+			margin-right: 8px;
+		}
+
+		.mdc-button--dropdown {
+			padding-right: 25px;
```
aaemnnosttv, line 103: For values where we aren't referencing a variable, can you please annotate or clarify why we're using that value? We should always use existing variables where available and potentially add them where it makes sense over using an arbitrary value which is difficult to maintain.

assets/sass/components/global/_googlesitekit-header.scss
```diff
@@ -51,7 +66,68 @@
...
+
+			.mdc-button__label {
+
+				@media (max-width: $bp-xsmallOnly) {
+					display: none;
+				}
+			}
```
aaemnnosttv, line 110: According to the design this shouldn't be hidden in mobile

image

image

aaemnnosttv (commented):
Thanks @benbowler – this one is pretty close, just a few things left to address.

assets/svg/date-range.svg
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
benbowler: This naming came from MaterialUI Icons (which I know we're not using directly) because there is a lot of different calendar icons with very specific uses, this one is specifically for a calendar with data range:

Screenshot 2020-11-25 at 10 15 03

assets/sass/components/global/_googlesitekit-header.scss
benbowler: Switched to variables in almost all cases, added additional comments.

assets/svg/date-range.svg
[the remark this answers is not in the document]
aaemnnosttv: Oh wow, well in that case, that's what it should be! 😄  Thanks for clarifying 👍

aaemnnosttv (approved):
Great, thanks!

Issue #2358, opened 2020-11-12
Investigate high instance of `profile_id_not_set` errors.


adamsilverstein:
## Bug Description

Our new error logging has revealed an unexpectedly high rate of errors from the `POST:modules/analytics/data/report` endpoint with the message/code: `Analytics view ID not set. (code: profile_id_not_set)`. (Note that profile ID is called a "View" in the Analytics UI).

This happens at this [code point](https://github.com/google/site-kit-wp/blob/edafc3d19bc305f2d829fa50f83df63088e65858/includes/Modules/Analytics.php#L846).

This lookup happens for every data call in Analytics, since the profile ID is required for those queries. We need to determine what condition causes the profile ID to not be set, and avoid making queries in that case.

## Steps to reproduce

1. Activate Analytics, **do not complete setup**
1. Navigate to Site Kit dashboard
1. Observe response to `google-site-kit/v1/data/` includes many responses with the `profile_id_not_set` error code
1. Observe a "WP Error in data response" warning in the console for each

## Screenshots

![image](https://user-images.githubusercontent.com/1621608/99418521-ca439700-2903-11eb-9a2f-4e5f01ddbbec.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Site Kit should never issue a REST API request to `modules/analytics/data/report` when the Analytics module is not **connected** (it is not sufficient to check only for whether the module is activated). In other words, it should not be possible to run into a `profile_id_not_set` error through regular usage of the plugin.

## Implementation Brief

* Skip requests for modules which have not completed setup in `withData`
* Finish and merge #2382 (mostly adding a few more tests)

### Test Coverage

* Add basic test coverage for `withData` functionality, including the setup incomplete component rendering, and request skipping

### Visual Regression Changes

* N/A

## QA Brief

* The PR for this issue changes the way batch requests are handled which only affects legacy components
* Test with a production build
* Batch requests for modules which have incomplete setup should be skipped now (does not affect others in the same batch)
	- This means warnings in the console should no longer appear for these as well
* Test dashboard widgets on main Site Kit dashboard, WP dashboard, and module dashboard pages continue to display as before
	- Note: module dashboards are not reachable when the module's setup is not yet completed so this should only affect the first two
* Test with different modules in various states of active, active + setup complete, and active + setup incomplete
    - Analytics and AdSense are the only two modules which can be activated and require setup that have dashboard widgets that make requests

## Changelog entry

[...]

PR #2382, opened 2020-11-17
Skip batch requests for unconnected modules

aaemnnosttv:
## Summary

Addresses issue #2358 

## Relevant technical choices

* Checks module setup before no data since skipped requests will always have no data and `setupComplete` is available  on page load
* Added basic tests for `withData`

## Checklist


assets/js/components/higherorder/withdata.js
```diff
@@ -219,34 +221,32 @@ const withData = (
...
-						 * Request data for the context.
-						 */
-						addFilter( `googlesitekit.module${ acontext }DataRequest`,
-							`googlesitekit.data${ acontext }`, ( moduleData ) => {
-								data.callback = ( returnedData ) => {
-									handleReturnedData( returnedData, data );
-								};
-								moduleData.push( data );
-								return moduleData;
-							} );
```
aaemnnosttv, line 235: This was previously duplicated for both the array and non-array cases. It's been consolidated into a single block using `castArray`.

felixarntz (changes requested):
Technically this works fine, however it results in some odd problem that now the "All Traffic" widget (legacy) displays the "Complete activation" CTA three times instead of just once for the entire widget. This is likely due to some detail in that component and how it makes its requests, but we should fix it to avoid a regression due to this fix.

Screenshot 2020-11-18 at 11 34 01

aaemnnosttv:
@felixarntz it had to do specifically with how the `LegacyAnalyticsAllTraffic` component rendered the error/incomplete-setup component separately and had separate logic for hiding the inner components in that case (`.googlesitekit-nodata`) which was slightly different than what was used to select the CTA component which made it possible for them to both show. I've updated that component to conditionally render the inner components now rather than simply hiding with CSS. Now that requests for modules with incomplete setup are no longer made, that means that components rendered `withData` that only request the single module and use the `handleDataError` or `handleDataSuccess` callbacks, these will never be called because they rely on a data response. I don't think this should really cause a problem since it only affects the dashboard which is easy to check. The module dashboard isn't accessible because it will just load the module setup, so I don't foresee this "sneaking up on us" later somewhere else.

image

As an aside, I also made a small tweak to `withData` so that it no longer hides the wrapped component names and shows a [proper display name](https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging) for an HOC, which is really refreshing.

image

felixarntz (approved):
Great! Looks like this also fixed another (even more severe) issue on the single URL details view: In there, the "All Traffic" widget used to stay in loading state forever, due to a JS error (this is even broken in the last release, and probably has been for a while 😱 ):
Screenshot 2020-11-18 at 15 40 30

This PR fixes it so the CTA shows there instead. (It does currently show twice, which _may_ be worth improving, but I'd rather ignore that for now, since what we have here is a huge improvement. Let's just get it into the release as is.)

Issue #2238, opened 2020-10-22
Remove legacy wizard CSS classes from proxy splash screen UI


felixarntz:
The default Site Kit splash UI (using the proxy) currently uses some CSS classes that are a remainder of the previous legacy setup wizard (which is only displayed when _not_ using the proxy). The default screen should be cleaned up to only use classes related to `googlesitekit-setup` instead of `googlesitekit-wizard` (the latter are the old classes).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The splash UI (using the proxy, i.e. `SetupUsingProxy` component) should not use any CSS classes start with `googlesitekit-wizard`. Instead it should use classes starting in `googlesitekit-setup`. New CSS classes may be added as needed.
* The splash UI furthermore should no longer use the class `googlesitekit-setup__footer` around its main UI (that doesn't make any sense, it's not a footer).
* Existing CSS rules should not be modified, unless it's certain that won't change anything in another screen.
* The appearance should not look different from how it is now.

## Implementation Brief
- Using the `/assets/js/components/setup/setup-proxy.js`
  - Update [class name]( https://github.com/google/site-kit-wp/blob/develop/assets/js/components/setup/setup-proxy.js#L118) from `googlesitekit-wizard` to `googlesitekit-setup` 
  - Remove the [container](https://github.com/google/site-kit-wp/blob/develop/assets/js/components/setup/setup-proxy.js#L124) with the class name `googlesitekit-setup__footer` since it does not do anything.
  - Update [class name](https://github.com/google/site-kit-wp/blob/develop/assets/js/components/setup/setup-proxy.js#L123) from `googlesitekit-wizard-progress` to `googlesitekit-setup__splash` 
- In `/assets/sass/components/setup/_googlesitekit-setup.scss`, add the following styles:
```
.googlesitekit-setup__splash {
	border-bottom: 1px solid #e6e6e6;
	padding-left: 10px;
	padding-right: 10px;
}

.googlesitekit-setup__splash .googlesitekit-setup__title {
	font-family: "Google Sans", "Helvetica Neue", Helvetica, Arial, sans-serif;
	margin: .67em 0;
}
```

### Test Coverage

* 

### Visual Regression Changes

* 

## QA Brief

* Disconnect and reset the plugin
* Viewing the setup page - the layout should be visually identical to the pre-commit version.  
_Check for both default proxy and legacy (GCP) splash screens_

## Changelog entry

* N/A

[...]

PR #2387, opened 2020-11-18
Enhancement/2238 remove legacy wizard styles

benbowler:
## Summary

Re-structured the setup page CSS with clearer class names and removed irrelevant footer wrapper.
Addresses issue #2238

## Relevant technical choices

Removed the irrelevant footer wrapper and split the class names out so this screen can be styled independently to the main app screens.

## Checklist


assets/js/modules/search-console/util/index.js
```diff
@@ -25,6 +25,7 @@ import { each } from 'lodash';
  * Internal dependencies
  */
 import { changeToPercent, readableLargeNumber } from '../../../util';
+export * from './is-zero-report';
```
tofumatt, line 28: This feels a _bit_ dangerous to me, because we're letting one file control another's exports. It's not a big deal, but given it's only one function anyway, I'd rather us be explicit.

That said: these big utils files can sometimes make for annoying circular imports or big files. I'm fine with this, but I'm also fine if we import from a specific file. I don't think we're really good about sticking to one way, so I get why you did this 😄

Maybe let's just use an explicit import here instead of `*`?

assets/js/modules/search-console/util/is-zero-report.js
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Checks whether the report data is valid.
```
tofumatt, line 20:
```suggestion
 * Checks whether the report data has any valid contents or is empty.
```

tofumatt (changes requested):
Looks good to me—just a few minor things and then this is good-to-go. Just ping me for another quick review after that 🙂

assets/js/modules/search-console/util/index.js
benbowler: Switched to explicit export of isZeroReport within utils

benbowler:
Hey @tofumatt, I'd incorrectly based this branch so it had some changes from other work, sorry about that. I've updated it now to include only these CSS changes. Best. Ben

tofumatt (approved):

assets/sass/components/setup/_googlesitekit-setup.scss
```diff
@@ -107,6 +107,17 @@
...
 		border-top: 1px solid $c-border-light;
 	}
+
+	.googlesitekit-setup__splash {
+		border-bottom: 1px solid #e6e6e6;
+		padding-left: 10px;
+		padding-right: 10px;
+	}
+
+	.googlesitekit-setup__splash .googlesitekit-setup__title {
```
aaemnnosttv, line 117: This should be simplified per @felixarntz in https://github.com/google/site-kit-wp/issues/2238#issuecomment-723280630

aaemnnosttv (changes requested):
Almost good to go, just a few quick details to address.

assets/sass/components/setup/_googlesitekit-setup.scss
benbowler: Updated.

benbowler:
Updated based on feedback.

aaemnnosttv (approved):
Nice 👍

PR #2388, opened 2020-11-18
Tweak user input notification to match Figma designs more closely.

asvinb:
## Summary

Addresses issue #2043 

## Relevant technical choices
Update the CSS for the user input notification to match the Figma designs closely.

## Checklist


assets/js/components/notifications/UserInputSettings.js
```diff
@@ -54,6 +54,7 @@ export default function UserInputSettings() {
 			ctaLabel={ __( 'Let’s go', 'google-site-kit' ) }
 			dismiss={ __( 'Remind me later', 'google-site-kit' ) }
 			winImage={ global._googlesitekitLegacyData.admin.assetsRoot + personSitImage }
+			className="googlesitekit-user-input__notification"
```
aaemnnosttv, line 57: Let's move this up next to the `id` as we usually have `className` as one of the first props.

aaemnnosttv (commented):
This is looking good @asvinb but it looks like it could use some adjustment for consistency with Figma.

The design shows that the image should still be present at 922px but it disappears at this width in Storybook
image
image

Also note how the text content has a max-width of what seems to be ~50% in this viewport where as it is spanning the whole notification without the image.

I'm noticing as well that the copy used in tablet does not match that in desktop so I'm guessing all viewports weren't updated when that changed, but we should check with @marrrmarrr.

asvinb:
> This is looking good @asvinb but it looks like it could use some adjustment for consistency with Figma.
> 
> The design shows that the image should still be present at 922px but it disappears at this width in Storybook
> image
> image
> 
> Also note how the text content has a max-width of what seems to be ~50% in this viewport where as it is spanning the whole notification without the image.
> 
> I'm noticing as well that the copy used in tablet does not match that in desktop so I'm guessing all viewports weren't updated when that changed, but we should check with @marrrmarrr.

@aaemnnosttv About the 922px, I do think that the designs should align with the existing breakpoints that we have:
https://github.com/google/site-kit-wp/blob/develop/assets/sass/config/_variables.scss#L150-L154
I wouldn't introduce a new breakpoint at 922px only for this particular notification when we have an existing breakpoint at 960px :) Else we'll potentially end up with hundreds of breakpoints :)

aaemnnosttv:
> About the 922px, I do think that the designs should align with the existing breakpoints that we have:

@asvinb Agreed! Although I wasn't suggesting we add or change any breakpoints. The design in Figma shows the image above from under the Tablet section which ends at 960px. With Mobile below it ending at 600px, this is consistent with the current variables you referenced for breakpoints. I was only pointing out how the current component does not show the image at all when the design shows it in tablet viewport. It should only be hidden below tablet as is shown in the design 🙂 

> Also note how the text content has a max-width of what seems to be ~50% in this viewport where as it is spanning the whole notification without the image.

I just wanted to highlight this again as it seems that both tablet and desktop split the content 50/50 with text and image. The image doesn't fill the full half, but occupies the space, the main difference being the "cropped" view of the image in desktop.

assets/js/components/notifications/UserInputSettings.js
asvinb: Updated @aaemnnosttv

asvinb:
> > About the 922px, I do think that the designs should align with the existing breakpoints that we have:
> 
> @asvinb Agreed! Although I wasn't suggesting we add or change any breakpoints. The design in Figma shows the image above from under the Tablet section which ends at 960px. With Mobile below it ending at 600px, this is consistent with the current variables you referenced for breakpoints. I was only pointing out how the current component does not show the image at all when the design shows it in tablet viewport. It should only be hidden below tablet as is shown in the design 🙂
> 
> > Also note how the text content has a max-width of what seems to be ~50% in this viewport where as it is spanning the whole notification without the image.
> 
> I just wanted to highlight this again as it seems that both tablet and desktop split the content 50/50 with text and image. The image doesn't fill the full half, but occupies the space, the main difference being the "cropped" view of the image in desktop.

@aaemnnosttv Image is now showing up for tablets > 600px. Following our conversation we had, I'll open a new ticket about updating the Notification component to be more flexible by adding a new layout prop which can cater for different kind of layouts, such as 50/50 or 75/25, etc

assets/js/components/notifications/UserInputSettings.js
aaemnnosttv: This is still at the bottom here 🙂

aaemnnosttv (commented):
This looks good, just the one tiny thing that was missed.

aaemnnosttv:
@googlebot I consent.

Issue #2283, opened 2020-10-29
Implement Congrats dashboard notification for submitting user input settings


felixarntz:
A new Site Kit dashboard notification should be implemented for when a user has successfully submitted their user input flow data.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new notification should be implemented using the `Notification` component:
    * Its layout should look similar to the screenshot below.
    * It should use the image from the screenshot below as SVG (SVG is also pasted below).
    * The copy for the heading should be: _Congrats! You set your site goals._
    * The copy for the content should be: _Now Site Kit will begin showing you suggestions how to add more metrics to your dashboard that are relevant specifically to you, based on the goals you shared._
* The notification should be displayed in the Site Kit dashboard when a certain query parameter indicating that the user input flow was completed is present - i.e. when the user is redirected to the Site Kit dashboard after completing the user input flow screen (see #2039), that query parameter should be included.
* The notice should be hooked in in a similar way like the authentication success notification is currently displayed. In other words, this should use filters etc. Refactoring is a non-goal for this issue.

### Screenshot of similar-looking notification (different content)
<img width="640" alt="Screenshot 2020-10-29 at 09 35 39" src="https://user-images.githubusercontent.com/3531426/97604410-b2df5100-19ca-11eb-94e4-0f7a82e1804e.png">

### SVG of icon above
```
<svg width="79" height="89" viewBox="0 0 79 89" fill="none" xmlns="http://www.w3.org/2000/svg">
<circle cx="39.5" cy="49.3164" r="39.5" fill="#FEF7E0"/>
[...]

PR #2313, opened 2020-11-03
Display dashboard notification when user input is required, adds User…

caseymorrisus:
## Summary

Addresses issue [#2043](https://github.com/google/site-kit-wp/issues/2043)

## Checklist

- [x] My code is tested and passes existing unit tests.
- [x] My code has an appropriate set of unit tests which all pass.
- [x] My code is backward-compatible with WordPress 4.7 and PHP 5.6.
- [x] My code follows the [WordPress](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) coding standards.
- [x] My code has proper inline documentation.
- [x] I have added a QA Brief on the issue linked above.
- [x] I have signed the Contributor License Agreement (see <https://cla.developers.google.com/>).

PR #2397, opened 2020-11-19
Add user input success notification and add story.

kostyalmm:
## Summary

Addresses issue #2283

## Relevant technical choices

* Add `UserInputSuccessNotification` component.
* Add story for the same.

## Checklist


assets/js/components/notifications/UserInputSuccessNotification.js
```diff
@@ -0,0 +1,41 @@
...
+ */
+import Notification from '../notifications/notification';
+import UserInputSuccessImage from '../../../svg/congrats.svg';
+
+export default function UserInputSuccessNotification() {
+	return (
+		<Notification
+			id="user-input-success"
+			title={ __( 'Congrats! You set your site goals.', 'google-site-kit' ) }
+			description={ __( 'Based on your goals, now Site Kit will begin showing you suggestions how to add more metrics to your dashboard that are relevant specifically to you', 'google-site-kit' ) }
```
tofumatt, line 35: I think this should have a full-stop at the end.
```suggestion
			description={ __( 'Based on your goals, now Site Kit will begin showing you suggestions how to add more metrics to your dashboard that are relevant specifically to you.', 'google-site-kit' ) }
```

Looking through the code, usage of punctuation isn't 100% consistent, but most instances of `<Notification />` have a `description` prop that is a full sentence. Especially when the title is a full sentence, it'd make sense that this one is too.

I know it wasn't in the IB, but I think that was a small oversight.

assets/js/components/notifications/UserInputSuccessNotification.js
```diff
@@ -0,0 +1,41 @@
...
+			smallImage={ <UserInputSuccessImage /> }
+			dismiss={ __( 'OK, Got it!', 'google-site-kit' ) }
```
tofumatt, line 37:
```suggestion
			dismiss={ __( 'OK, got it!', 'google-site-kit' ) }
```

(The screenshot shows a lowercase "g", which is grammatically correct here 🙂 )

assets/js/components/notifications/notification.js
```diff
@@ -302,7 +302,10 @@ class Notification extends Component {
 								mdc-layout-grid__cell
 								mdc-layout-grid__cell--span-1
 							">
-								<img className="googlesitekit-publisher-win__small-image" alt="" src={ smallImage } />
+								{
+									typeof smallImage === 'string'
+										? <img className="googlesitekit-publisher-win__small-image" alt="" src={ smallImage } /> : smallImage
+								}
```
tofumatt, line 308: I'm not a huge fan of this overloading of the prop; accepting very different types of arguments for props like this can be confusing, and wasn't part of the IB. I understand wanting to import the SVG directly, but the IB suggests using `${global._googlesitekitLegacyData.admin.assetsRoot}${userInputSuccessImage}` as the `src`/`smallImage` value here. It'd be great to refactor this to accept components directly in the future, but let's do that all-at-once rather than this hybrid approach that makes the API less clear.

tofumatt (changes requested):
There are a few minor text changes suggested, but mainly the issue with this approach is the SVG icon component rather than path for `smallImage`. The IB mentions using a path to the asset, which I think is the best approach for now.

If you want to file a code quality issue for `<Notification />` to accept nodes instead of strings we could do that, though I think that component in general warrants enough refactoring that we were thinking of replacing it later.

For now I think it's best to provide the SVG path as an `src` value rather than changing the `Notification` component.

assets/js/components/notifications/notification.js
tofumatt: @kostyalmm mentioned this on Slack:

> About this PR comment, the issue is that the svg files are not hosted to the server like images are.
Exists: http://sitekit.10uplabs.com/wp-content/plugins/google-site-kit/dist/assets/images/rocket.png
Does not Exist: http://sitekit.10uplabs.com/wp-content/plugins/google-site-kit/dist/assets/svg/adsense.svg
Everywhere else SVGs are used as nodes and the import statement returns a node instead of the path, that is why I decided to overload the Notification param.

So it was my bad thinking SVGs were available as directly-referenced files. In that case, I think this is fine 🙂

tofumatt (dismissed):

assets/js/components/notifications/notification.js
felixarntz: While this works, I think we should at least give it a second thought. I left a related comment here: https://github.com/google/site-kit-wp/pull/2313#discussion_r529021339

We should decide on a single approach for how we want to handle image URLs vs components like SVGs. @caseymorrisus @kostyalmm Would be great if you could find a common approach here. For both `smallImage` and `winImage` props (and in general), do we want to allow overloading this so that you can pass a component (SVG) or a string (image URL)? Or do we want to have separate props for the two alternatives? IMO the latter would be a cleaner API, but let's discuss. We should definitely implement this the same way between here and #2313.

assets/js/components/notifications/UserInputSuccessNotification.js
```diff
@@ -0,0 +1,41 @@
...
+ */
+import Notification from '../notifications/notification';
+import UserInputSuccessImage from '../../../svg/congrats.svg';
+
+export default function UserInputSuccessNotification() {
+	return (
+		<Notification
+			id="user-input-success"
+			title={ __( 'Congrats! You set your site goals.', 'google-site-kit' ) }
+			description={ __( 'Based on your goals, now Site Kit will begin showing you suggestions how to add more metrics to your dashboard that are relevant specifically to you.', 'google-site-kit' ) }
```
felixarntz, line 35: The text is slightly off against the ACs, missing the last clause.

felixarntz (changes requested):
Mostly LGTM, however one copy change required, and we need to decide how we're gonna deal with https://github.com/google/site-kit-wp/pull/2397/files#r529025540.

assets/js/components/user-input/UserInputPreview.js
```diff
@@ -57,7 +57,7 @@ export default function UserInputPreview( { back, goTo } ) {
 		setIsNavigating( true );
 		const response = await saveUserInputSettings();
 		if ( ! response.error ) {
-			global.location.assign( dashboardURL );
+			global.location.assign( `${ dashboardURL }&notification=user_input_success` );
```
felixarntz, line 60: I think it'd be more safe to use `addQueryArgs` here.

felixarntz (changes requested):

assets/js/components/notifications/notification.js
kostyalmm: @felixarntz @tofumatt @caseymorrisus I suggest we use two separate props like:
```js
smallImageSrc: PropTypes.string
smallImage: PropTypes.elementType

winImageSrc: PropTypes.string
winImage: PropTypes.elementType
```
Using `PropTypes.elementType` or `WPComponent` so that we can pass additional props to the SVGs if we want to.
What do you guys think?

aaemnnosttv: @felixarntz 👍  on adding svg support.

@kostyalmm 

> Using `PropTypes.elementType` or `WPComponent` so that we can pass additional props to the SVGs if we want to.

This is a good point – the svg loader we have strips the width and height from the source svgs to allow them to be passed as props (not sure why it does that really but it does). That means that even if we wanted to accept a component type as the prop here for an svg, it would have to be wrapped for svg components to be able to specify dimensions. E.g. `winImage={ ( props ) => <SomeSVG width="300" height="200" { ...props } />  }`
So it would be a bit cleaner to pass an element, but less flexible if we wanted to support passing class names, etc to the image element from within the `Notification`.

It would be nice if we had an image loader that loaded images in the same way we load SVG so that the usage was compatible. I realize they have some differences with the way sizing works but might be nice.

caseymorrisus: @felixarntz @aaemnnosttv @kostyalmm  Related to this conversation, I added an `ImageSVG` prop to the `Notification` component, can be found [here](https://github.com/google/site-kit-wp/blob/0ba1c62e482e89c0c40d8f263eb443a6a38247c6/assets/js/components/notifications/notification.js#L394).

As far as explicitly passing src props, I think we should stick with passing components or elements unless we're working directly within a custom image component (then a src prop is more valid).

kostyalmm: @caseymorrisus I agree. 
About the `ImageSVG`, I think this should be renamed to `WinImageSVG` or simply `WinImage` as using `ImageSVG` will create confusion since there is `winImage` and `smallImage`.

caseymorrisus: @kostyalmm This is a good callout, I originally had the assumption that `smallImage` was an alternate variant of `winImage` (meaning no need to distinguish for SVGs as they scale without loss of quality) but this is incorrect.

I'll modify my PR to call this `WinImageSVG` to make this more clear which area of the notification this is affecting.

felixarntz: @kostyalmm This PR still needs to be updated per the above conversation to add a `SmallImageSVG` prop and use that, instead of overloading the `smallImage` prop (similar like how in the other PR there will be a `WinImageSVG`).

felixarntz (changes requested):
@kostyalmm The one change from the previous discussion is still outstanding, see https://github.com/google/site-kit-wp/pull/2397#discussion_r533061609

.storybook/storybook-data.js
```diff
@@ -368,6 +368,19 @@ module.exports = [
...
+		name: 'User Input Success Notification',
+		story: 'User Input Success Notification',
+		parameters: {
+			fileName: './stories/notifications.stories.js',
+			options: {
+				hierarchyRootSeparator: '|',
+				hierarchySeparator: {},
+			},
+		},
+	},
```
felixarntz, line 383: @kostyalmm This is not really necessary to add a Storybook story, it adds the story to the visual regression tests which would be a separate requirement. It's not a blocker that you added this here, but going forward please only add new Storybook stories to Storybook, not the VRT. We'll have to revisit which stories we include in VRT overall since there is a lot of legacy bloat there, while other important things are missing.

Just FYI 👍

felixarntz (approved):
Awesome! Just one point of feedback, but no blocker for merging.

PR #2404, opened 2020-11-19
Restore CSS hack for keeping components mounted

aaemnnosttv:
## Summary

Addresses issue #2358

PR targets `master` for inclusion in the release. @cole10up and @adamsilverstein confirmed that this change fixed the original problem they found.

## Relevant technical choices

This PR undoes a change where inner `withData`-wrapped components in `LegacyAnalyticsAllTraffic` would be conditionally rendered. This caused a problem when an error component was caused to be rendered which would unmount the inner components. When the date range changed, and the error component related state was cleared, the inner `withData` connected components rendered an infinite loading state because they seem to have missed the timing for `withData` to hook into the batch request for the context if the requests for that batch are already cached. There may be a bit more nuance to it but that's the gist of it, and likely why this part was hidden with CSS previously.

## Checklist

### DIFF
--- a/assets/js/modules/analytics/components/dashboard/LegacyAnalyticsAllTraffic.js
+++ b/assets/js/modules/analytics/components/dashboard/LegacyAnalyticsAllTraffic.js
@@ -115,41 +115,42 @@ class LegacyAnalyticsAllTraffic extends Component {
 					/>
 				</div>
 
-				{ errorComponent || (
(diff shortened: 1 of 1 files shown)


felixarntz (approved):
Hiding the component doesn't LGTM, but since that's old and we shouldn't spend much time on improving, LGTM 😆

Issue #2247, opened 2020-10-22
Fix "Maximum call stack size exceeded" errors from Google charts library


felixarntz:
## Bug Description

Occasionally, any Google charts on the dashboard get replaced by "Maximum call stack size exceeded" errors. This happens in both the legacy content and the widgets API-driven content, so it's unrelated to that.

**Legacy UI:**
![Screenshot 2020-10-22 at 13 12 23](https://user-images.githubusercontent.com/3531426/96925120-85ddec00-1468-11eb-87ce-cab73a22bdd6.png)

**Widget API-driven UI:**
![Screenshot 2020-10-22 at 13 12 34](https://user-images.githubusercontent.com/3531426/96925133-8bd3cd00-1468-11eb-9b56-a27d6ff98bac.png)

It seems that this error happens primarily after returning to a previously opened tab with the Site Kit dashboard open. I'm not 100% about this assessment, but it has definitely _never_ happened to me on pageload (or more precisely, initial API response), always a bit later.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Google charts shouldn't occasionally be replaced by "Maximum call stack size exceeded" errors.

## Implementation Brief

This appears to be caused by us loading the chart library multiple times, best I could tell from similar errors encountered elsewhere. In the `develop` branch, the charts are loaded several times on the dashboard:

<img width="816" alt="Screenshot 2020-11-20 at 01 20 06" src="https://user-images.githubusercontent.com/90871/99743897-8ea8f880-2ace-11eb-8542-d25f94a0c7a9.png">

This seems intermittent (a race condition based on other instances of `GoogleChart`'s component/loading state).

* Ensure only one Google Charts loader script tag is loaded.
* Ensure `google.charts.load()` is only called once, and any remaining Charts instances wait for it to be ready.
* See #2408 for a proof-of-concept. (Rough, and could use improvements and possible datastore usage.)

Essentially, we should only be inserting one script tag into the DOM to load the Charts code and ensuring `google.charts.load()` is only called once.

Currently the proof-of-concept does this inside the component and by using the `wait-promise` module to check if `google.charts && google.charts.visualizations` is available. It could be worth storing the Google Charts loading state in the data store instead and providing a selector, as the existing code is not very easy-to-follow.

If no data store is to be used, we can tidy up the existing PoC (#2408) with better variable naming and reducing complexity where possible.

### Test Coverage

[...]

PR #2408, opened 2020-11-20
Fix Google Charts stack exceeded errors.

tofumatt:
## Summary

Addresses issue #2247.

## Relevant technical choices

Used a global variable to prevent issues if multiple entrypoints used Google Charts. It's less pretty, but more reliable for something like Charts that have global state.

## Checklist


felixarntz:
As previously noticed, #2445 appears to already fix the underlying bug, but it's probably worth making this improvement too. Also, checking without the #2445 (or now `develop`) branch merged into this still shows the error gets fixed by this too, so +1!

Two different bug fixes for the same bug is better than one bugfix! 🎉 🤣

felixarntz (approved):
Awesome!

Issue #2143, opened 2020-10-07
Allow registering module icons in JS


felixarntz:
See https://github.com/google/site-kit-wp/pull/2115#discussion_r501283031, the `registerModule` action in the `core/modules` store should allow registering a module icon.

As part of this issue, we should also define criteria that all module icons should adhere to.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The optional `icon` property in the `registerModule` action of `core/modules` should be changed to not be a string, but a React component type (e.g. `AdSenseIcon`).
    - There should be no default, i.e. we shouldn't assume that every module has an icon.
    - The `icon` should be passed through with the module's settings into the reducer
* All six Site Kit modules in JS should include passing their `icon` in the `registerModule` call.
* There should be a new `getModuleIcon( slug )` selector on the `core/modules` store which returns the registered icon component for the given module slug, `null` if there is no icon for the module, or `undefined` if the module has not been registered.
* The `ModuleIcon` component should be updated to source its inner icon component from the `getModuleIcon` selector only
* All module icons should be square or render as squares without distortion when given the same `width` and `height`
* `ModuleIcon` should be updated to accept a `size` prop instead of `width` and `height` which should be passed as both `width` and `height` to the inner icon component
     - All use of `ModuleIcon` should be updated accordingly

## Implementation Brief

* Change the `icon` argument documentation to be a `WPComponent` https://github.com/google/site-kit-wp/blob/07d95eddab386c3c5b9f8d1dad1f123d7cb33dae/assets/js/googlesitekit/modules/datastore/modules.js#L222. As the component can be stored in the data store, no code needs to be updated to support storing the component.
* Remove https://github.com/google/site-kit-wp/blob/07d95eddab386c3c5b9f8d1dad1f123d7cb33dae/assets/js/components/ModuleIcon.js#L44-L46
* Replace https://github.com/google/site-kit-wp/blob/07d95eddab386c3c5b9f8d1dad1f123d7cb33dae/assets/js/components/ModuleIcon.js#L48 with `const ModuleIconComponent = useSelect( ( select ) => select( CORE_MODULES ).getModuleIcon( slug ) );`
[...]

PR #2415, opened 2020-11-23
Allow registering module icons, adjust stories and tests

kostyalmm:
## Summary

Addresses issue #2143

## Relevant technical choices

* Add `icon` property to the Module object.
* Add `getModuleIcon` selector.
* Use the selector in `ModuleIcon` component
* Wrap relevant stories in `WithTestRegistry` and use `provideModules` methods.
* Alter `provideModules` test utility to add `icon` to the fixtures before adding them to the store.

## Checklist


kostyalmm:
I understand that the issue says no change for the VRT but one change will come up due to the change [here](https://github.com/google/site-kit-wp/pull/2415/files#diff-6d33cbec4183cae98c82ae1d28df73b2c5e5c0c0987187e73446c1e33c33d2ceL289)
The `width` and `height` props were different which can't be the case now after using `size`

assets/js/components/ModuleIcon.js
```diff
@@ -24,40 +24,28 @@ import PropTypes from 'prop-types';
...
-	adsense: AdSenseIcon,
-	analytics: AnalyticsIcon,
-	optimize: OptimizeIcon,
-	'pagespeed-insights': PageSpeedInsightsIcon,
-	'search-console': SearchConsoleIcon,
-	tagmanager: TagManagerIcon,
-};
+const { useSelect } = Data;
+
+export default function ModuleIcon( { slug, size = 33, ...props } ) {
```
felixarntz, line 32: Maybe use `defaultProps` instead?

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -388,6 +388,35 @@ const baseSelectors = {
...
+	 * Gets a specific module icon by slug.
+	 *
+	 * Returns a specific module icon by its slug.
+	 * Returns `null` if state is still loading or if said module doesn't exist or doesn't have an icon.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @param {string} slug  Module slug.
+	 * @return {(WPComponent|null|undefined)} A specific module's icon; `undefined` if state is still loading or if said module doesn't exist; `null` if it doesn't have an icon.
```
felixarntz, line 401: The explanation here is not exactly correct based on the code below: `undefined` is only returned if state is still loading. `null` is returned if the module doesn't exist or if it doesn't have an icon.

tests/js/utils.js
```diff
@@ -204,13 +210,23 @@ export const provideModules = ( registry, extraData = [] ) => {
...
 	const moduleSlugs = coreModulesFixture.map( ( { slug } ) => slug );
 	const modules = coreModulesFixture
 		.map( ( module ) => {
+			const icon = moduleIconMap[ module.slug ];
 			if ( extraModules[ module.slug ] ) {
-				return { ...module, ...extraModules[ module.slug ] };
+				return { ...module, ...extraModules[ module.slug ], icon };
 			}
-			return { ...module };
+			return { ...module, icon };
```
felixarntz, line 229: While this technically works, it doesn't really fit here, since `receiveGetModules` has the purpose of receiving module data from the API, i.e. PHP. JS-specific data (like `icon` is) should be registered via `registerModule`. They'll also be stored separately internally in the datastore (`serverDefinitions` vs `clientDefinitions`), and we should stick to that here.

I suggest we introduce a `provideModuleRegistrations` or similar test utility function which calls `registerModule` for every one of the 6 modules by default, with the expected data, allowing for custom overrides to be optionally passed (basically similarly to how this `provideModules` works.

Tests or Storybook stories that require JS module data to be registered should then use that utility function.

felixarntz (changes requested):
Mostly LGTM, just a few points of feedback.

tests/js/utils.js
kostyalmm: @felixarntz I have implemented the suggested test utility.

tests/js/utils.js
```diff
@@ -190,6 +196,49 @@ export const provideUserInfo = ( registry, extraData = {} ) => {
...
+		const icon = moduleIconMap[ module.slug ];
+		let moduleData = { ...module, icon };
+		if ( extraModules[ module.slug ] ) {
+			moduleData = { ...moduleData, ...extraModules[ module.slug ] };
+		}
+		delete moduleData.slug;
+		registry.dispatch( CORE_MODULES ).registerModule( module.slug, { ...moduleData } );
+	} );
+
+	registry.dispatch( CORE_MODULES ).receiveGetModules( modules );
```
felixarntz, line 239: The implementation here is slightly off:
* It should only call `registerModule` for all six modules and not deal with populating server-side module data (`receiveGetModules`).
* Therefore it also shouldn't rely on the `coreModulesFixture`, which is only responsible for server-side module data. Basically, the main part of this function would be to replicate all six `registerModule` calls that in the production code are found within the `assets/js/modules/*/index.js` files into this single utility function, in order to allow us to use the same module registration data in Storybook that the plugin also uses in production.
* In other words, we should probably have for each one of the six modules an entry here with all arguments that are passed to the real `registerModule` call, and combine that base data with `extraData` for that module, if passed, before calling `registerModule`.

felixarntz (changes requested):
LGTM, only the new `provideModuleRegistrations` utility is slightly off in what it does. It might be worth discussing this further, let me know if you have any questions/feedback.

tests/js/utils.js
kostyalmm: I made the changes as requested. I had to call the `receiveGetModules` otherwise a number of stories would have to be overwritten. Now I am calling `provideModuleRegistrations` followed by `provideModules` which works as expected.

tests/js/utils.js
```diff
@@ -190,6 +196,30 @@ export const provideUserInfo = ( registry, extraData = {} ) => {
...
+ */
+export const provideModuleRegistrations = ( registry, extraData = [] ) => {
+	const moduleIconMap = {
+		adsense: AdsenseIcon,
+		analytics: AnalyticsIcon,
+		optimize: OptimizeIcon,
+		'pagespeed-insights': PagespeedInsightsIcon,
+		'search-console': SearchConsoleIcon,
+		tagmanager: TagManagerIcon,
+	};
```
felixarntz, line 216: Following-up on my previous comment: While this works for this particular issue, I meant that it would make sense to provide the complete **real** module registration args here for each of the six modules, which include not only the icon, but also a couple of JS components. I'm gonna add these changes here for now.

felixarntz (changes requested):
Mostly LGTM, just one Storybook-specific feedback. We'll also need to call `provideModuleRegistrations` in a few other places where now the icons are missing in Storybook, e.g. in all the module-specific settings stories.

felixarntz (approved):
Good to go now. My additional commits address various broken Storybook stories that were lacking some setup, some of which were already broken before in `develop`.

felixarntz (approved):

PR #2427, opened 2020-11-24
Fix stylistic regression to proxy splash screen

aaemnnosttv:
## Summary

Addresses issue #2238 

This PR fixes a regression caused by some styles being overridden by new styles intended for use with feature flag only.

## Relevant technical choices

* Preserves vertical margin on `googlesitekit-opt-in` and `googlesitekit-setup__description` classes in production mode as present in current release
* Preserves user-input mode specific styles using `googlesitekit-setup__content` class  
_Ideally this would be handled in a more holistic way but this is beyond the scope of this issue_
* Adds stories to "Setup / Using Proxy" for both user-input and non-user-input enabled variants

## Checklist


felixarntz (changes requested):
While this fixes the actual issue, there is a problem with the new Storybook stories: I don't think it's expected that it shows the staging environment warning in its "default" story state - if we want a case with that, we should rather have an extra story.

Furthermore the new setup design if a bit off when such a warning shows up, with the icon and the text overlapping:
Screenshot 2020-11-30 at 20 05 42

aaemnnosttv:
@felixarntz there seems to have been something different between when you last looked at this and when I got it it as these stories were all broken seemingly due to other changes that were merged since. I've fixed that with some additions to this branch that resets all the feature flags for every story, so each story only needs to enable whatever it needs using some new story utilities.

So the story wasn't rendering the "not ready for Site Kit" error that you were seeing, but it's possible you were seeing this side-effect from viewing another story first which sets `global._googlesitekitLegacyData.setup.isSiteKitConnected = false;`

Regarding the CSS issue you pointed out, I took a look at improving this as well but didn't get that far. One thing to point out though is that it may look worse than it really is due to the 100px of width which is lost by the padding added by `.googlesitekit-plugin-preview` to all stories.

With padding:
image

Without padding
image

These screenshots were taken right at the point before the grid collapses into a column so this would be the closest they get. There's also another 36px or so that would be added by the collapsed admin menu that would cause this point to be reached a little sooner.

For screens like this, I think it would be better to use a WP admin wrapper rather than some arbitrary padding (which is more for centering I think). I'll open an issue to consider this.

Take another look and if you still feel the style needs adjustment let's see if @asvinb can help address it.

aaemnnosttv:
@felixarntz I pushed a few more updates which are admittedly a bit out of scope here but rather necessary (and a bit overdue). Mainly, I added code to reset Site Kit globals before every story, and removed a bunch of unnecessary imports/globals that were still there from a long time ago when we were sharing these globals from WP. We can extract some/most of this out to a new issue if you'd like but this was the best way to unblock Asvin without creating more of the same kind of problems as before with global state mutations leaking between stories.

asvinb:
@googlebot I consent.

assets/sass/components/setup/_googlesitekit-setup.scss
```diff
@@ -127,14 +128,21 @@
 	margin-right: 16px;
 }
 
-.googlesitekit-wizard {
+.googlesitekit-setup__warning {
+	align-items: center;
+	display: flex;
```
aaemnnosttv, line 133: @asvinb are there `mdc-layout-*` classes we can use for these? 🤔

assets/sass/components/setup/_googlesitekit-setup.scss
```diff
@@ -127,14 +128,21 @@
...
+
+	svg {
+		width: 34px;
+	}
```
aaemnnosttv, line 137: @asvinb Is this necessary? It seems like this should already be the case from the component here: https://github.com/google/site-kit-wp/blob/160b5d1393cee4cce60574827564b5b02cd8b78e/assets/js/components/legacy-notifications/warning.js#L28

assets/sass/components/setup/_googlesitekit-setup.scss
asvinb: @aaemnnosttv Using the `mdc-layout-*` classes are overkilled for this since the `mdc-layout-*` is using the grid layout. The grid gap is set to 0 on certain breakpoints, resulting in the overlapping icon and when the content for a cell is greater than the cell itself. For eg. the warning icon.

assets/sass/components/setup/_googlesitekit-setup.scss
asvinb: Yes it's necessary @aaemnnosttv , to prevent the SVG from shrinking as we resize the browser. Since we are using flex, everything will remain on the same line

asvinb:
@aaemnnosttv As mentioned in Slack,  adjusting the grid might take longer and we need to think about it, because we are using all the out of the box grid classes from material for this section. So we'll either be overwriting material's CSS or having some custom class names.

felixarntz (approved):
Awesome!

Issue #2261, opened 2020-10-23
Handle global error notifications via `googlesitekit-api`


felixarntz:
The legacy data API includes some special error handling that sets global notifications. Some similar behavior was already implemented in `googlesitekit-api` relying on the datastore. This issue is exposing the new behavior behind a feature flag. It is furthermore about separating the related pieces from the rest of the notifications which is mostly legacy and should remain untouched.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `assets/js/components/notifications` folder should be renamed to `assets/js/components/legacy-notifications`. All import references should be updated accordingly.
* A new `assets/js/components/notifications` folder should be created.
* The components `AuthError` and `DashboardAuthScopesAlert` should be moved from the `legacy-notifications` folder to the `notifications` folder. The latter should be renamed to `UnsatisfiedScopesAlert`. Again, all import references should be updated accordingly.
* A new functional component `ErrorNotifications` should be added in `assets/js/components/notifications`. It should simply render both `AuthError` and `UnsatisfiedScopesAlert`.
* A new feature flag `storeErrorNotifications` should be introduced. It should be `true` for development builds, `false` for production builds.
* The `Header` component in `assets/js/components` should be modified: If the new `storeErrorNotifications` flag is `true`, the component should render the new `ErrorNotifications` component _right above_ the legacy `ErrorNotification` component.

## Implementation Brief

* Rename `assets/js/components/notifications` to `assets/js/components/legacy-notifications`, use your IDE for this so you can update references.
* Make directory `assets/js/components/notifications`
* Move `assets/js/components/legacy-notifications/AuthError.js` to `assets/js/components/notifications/AuthError.js`
* Move `assets/js/components/legacy-notifications/DashboardAuthScopesAlert.js` to `assets/js/components/notifications/UnsatisfiedScopesAlert.js`
* In `assets/js/components/notifications/UnsatisfiedScopesAlert.js`, rename the default function export `export default function UnsatisfiedScopesAlert()`
* Update all references from `DashboardAuthScopesAlert.js UnsatisfiedScopesAlert` to `UnsatisfiedScopesAlert.js UnsatisfiedScopesAlert`
* Update all references for all components in `assets/js/components/legacy-notifications` to the new location (previously: `assets/js/components/notifications`)
* Create `assets/js/components/notifications/ErrorNotifications.js`, it will look like this:

```js
import { Fragment } from '@wordpress/element';

[...]

PR #2434, opened 2020-11-25
Handle global error notifications via `googlesitekit-api`

kostyalmm:
## Summary

Addresses issue #2261

## Relevant technical choices
* Move `components/notifications` to `components/legacy-notifications`
* Move `DashboardAuthScopesAlert ` and `UnsatisfiedScopesAlert ` to `components/notifications`
* Introduce `components/notifications/ErrorNotifications.js` and add it to `Header`
* Add `storeErrorNotifications` flag
* Update imports and stories

## Checklist


aaemnnosttv:
@kostyalmm – would you please resolve the conflicting files?

stories/header.stories.js
```diff
@@ -37,13 +37,17 @@ storiesOf( 'Global', module )
...
+		global.featureFlags = { storeErrorNotifications: { enabled: true } };
+
 		return (
 			<WithTestRegistry registry={ registry }>
 				<Header />
 			</WithTestRegistry>
 		);
 	} )
 	.add( 'Plugin Header with Date Selector', ( registry ) => {
+		global.featureFlags = { storeErrorNotifications: { enabled: true } };
```
aaemnnosttv, line 49: We need a better way to work with feature flags for Storybook, but for now, let's only set the one flag rather than overwrite the entire object. (Same above)

This is really easy using `lodash.set` – e.g. https://github.com/google/site-kit-wp/pull/2427/files#diff-b784d442e62f68075556b9485703ee43928f2ec379430fbfb7983001aa9cbae1R40

stories/setup.stories.js
```diff
@@ -19,6 +19,7 @@ storiesOf( 'Setup', module )
 		global._googlesitekitLegacyData.setup.isVerified = false;
 		global._googlesitekitLegacyData.setup.hasSearchConsoleProperty = false;
 		global._googlesitekitLegacyData.permissions.canSetup = true;
+		global.featureFlags = { storeErrorNotifications: { enabled: true } };
```
aaemnnosttv, line 22: See above.

aaemnnosttv (changes requested):
Thanks @kostyalmm – this looks good, just a few things to tweak and then this should be good to go.

aaemnnosttv (approved):
LGTM 👍

aaemnnosttv:
@kostyalmm – The todo above indicates that the QA Brief was added but none is there on the issue; would you please add one?

kostyalmm:
Sorry I forgot to add QA brief. But while making a brief I noticed that this notification will be shown twice with these changes(once from `ErrorNotifications` and then again by `withFilter`):

visual change:
image

react component tree:


Do you want me to remove the logic which adds this filter? Or do we want to keep it like this in context of this issue?

aaemnnosttv:
@kostyalmm – I believe the intention is that there would be duplication, but we can confirm with @felixarntz.

@felixarntz – this is expected as references to the legacy notices would have been updated to use the new components which are also rendered in the feature-guarded component, correct?

felixarntz:
@aaemnnosttv @kostyalmm Yes, the duplicate is expected because one is triggered from the new API, the other one from the legacy API. Because the new one is behind a feature flag though, it doesn't affect the production build, where it would still come from the legacy API only.

felixarntz (approved):
Great!

I've updated one instance where the setup for the feature flags in Storybook were missing, causing the `DashboardDetailsApp` stories to break. I've also changed the Storybook feature flags to be set to `false`, since Storybook should reflect the current production state unless it's explicitly a story for a feature hidden behind a feature flag. It's not great at all that we have to manually set up feature flags in Storybook right now though - #2452 should resolve that.

Issue #2344, opened 2020-11-10
Webpack dist grows with every change in watch mode


aaemnnosttv:
## Bug Description

In #1700 we updated our build process to write content hashes into our webpack bundle filenames to avoid conflicts with aggressive browser/cdn/plugin caching. While this works well, it currently doesn't clean up after itself when running in watch mode.

This causes unnecessary files to accumulate, which aside from the wasted disk space, increases CPU usage noticeably when making changes as webpack is both generating new files, and the editor is indexing them.

## Steps to reproduce

1. Run `npm run watch`
1. Modify `assets/js/googlesitekit-activation.js` and save (the file is arbitrary – any bundle that's built with a hash in the file name will work)
1. Notice there are now duplicate entries in `dist/assets/js` for this module

```
-rw-r--r--   1 aaemnnosttv  staff   615K Nov 10 22:48 googlesitekit-activation.19dfb03d188a9fe4e5d9.js
-rw-r--r--   1 aaemnnosttv  staff   436K Nov 10 22:48 googlesitekit-activation.19dfb03d188a9fe4e5d9.js.map
-rw-r--r--   1 aaemnnosttv  staff   615K Nov 10 22:48 googlesitekit-activation.5a32dc60289c284f93fe.js
-rw-r--r--   1 aaemnnosttv  staff   436K Nov 10 22:48 googlesitekit-activation.5a32dc60289c284f93fe.js.map
```

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Only the current version of a bundle should be present in the `dist` directory, regardless of which mode Webpack is running in (watch or not)

## Implementation Brief

* Edit the main `wepback.config.json` file:
  * Change output.filename to be `mode == 'production' ? '[name].[contenthash].js' : '[name].js'`

### Test Coverage

* Not needed

### Visual Regression Changes

* No visual changes are expected as a result of this change

## QA Brief

* Run `npm run watch` command and check that generated assets don't have hashes in filenames.
* Build production assets with `npm run build:prod` and make sure that generated assets have hashes in filenames.

## Changelog entry

* N/A

aaemnnosttv:
@eugene-manuilov here's the issue I mentioned last week. cc: @felixarntz @tofumatt

felixarntz:
IB ✅

eugene-manuilov:
QA :heavy_check_mark: Webpack works as expected: uses hashes for the production build and builds assets without hashes in the development mode.

Issue #2316, opened 2020-11-03
Implement feature flags support in PHP


felixarntz:
Site Kit already uses feature flags in JS (powered by Webpack) to enable certain features only for development builds. We should add support to achieve something similar in PHP.

Note that due to the way JS is bundled by Webpack, the JS code that is behind feature flags won't even be included in a production build. This isn't possible in PHP, but that is acceptable. We already ship a bunch of PHP code that will only be used by future features - the goal of this enhancement is to be able to easily test PHP features before enabling them in a production build.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

### Infrastructure
* The `webpack.feature-flags.config.js` file should be turned into a `feature-flags.json` file (note JSON), since it's no longer purely Webpack-specific.
* Webpack should generate a JSON file `config.json` that is added within `/dist`, which simply includes the following code:  `{ "flagMode":"BUILDMODE" }` (where `BUILDMODE` is either `production` or `development`).
* There should be a `Google\Site_Kit\Core\Util\Feature_Flags` class that allows detecting whether specific feature flags are enabled.
    * The class should have a public static method `enabled( string $flag )` that returns a boolean whether the feature flag is active, by checking whether the requested feature flag's required build mode (see below) matches the current build mode (see below). Callers should be able to pass strings like `userInput` or `widgets.dashboard`.
    * The class should have a private method that reads the `feature-flags.json` file if it exists. Its content should be flattened into a map where the keys are the full `$flag` qualifiers (e.g. `widgets.dashboard`) and the values are the mode for which the feature flag is enabled.
    * The class should have a private method that returns the current flag mode. If the above `dist/config.json` file exists, it should be parsed, and the `flagMode` value should be used. If it doesn't exist or parsing fails, `production` should be assumed. That value should then be filterable via `googlesitekit_flag_mode` (which can be modified e.g. for testing).

#### Example usage
```php
if ( Feature_Flags::enabled( 'userInput' ) ) {
    // Code that only runs when feature flag is enabled.
}
```

### Initial use-case
(originally part of #2133)
* When redirecting the user to the authentication service in PHP, if the `userInput` feature flag is enabled in PHP (via the above infrastructure), the `googlesitekit_user_input_state` user option should be set to `required` if it's not already set to `completed`. Furthermore, if it's not already set to `completed`, the redirect should happen including a query parameter `mode` set to `user_input`, which will cause the copy and UI to be slightly modified to cater for the change in the flow.

[...]

PR #2435, opened 2020-11-25
Introduce foundation for feature flags in PHP

aaemnnosttv:
## Summary

Addresses issue #2316

## Relevant technical choices

* Also supports an array of modes, just like `webpack-feature-flags-plugin`
* Introduced a `JSON_File` class to abstract the filesystem access but mainly to prevent unnecessary filesystem reads every time a feature flag is referenced (only loads once, on-demand)
    - Will be useful for other issues in the future, such as enhancements to `Manifest` (#2344)
* Since the JSON file implements `ArrayAccess` - `Feature_Flags` is able to use normal arrays internally for tests to decouple it from the JSON class
* Updates base `TestCase` to reset `Feature_Flags` (mode+config) before and after all test cases as the staticly stored values will persist between tests

## Checklist


includes/Core/Util/Feature_Flags.php
```diff
@@ -0,0 +1,140 @@
...
+		$feature_modes = array_reduce(
+			$feature_path,
+			function ( $value, $key ) {
+				if ( isset( $value[ $key ] ) ) {
+					return $value[ $key ];
+				}
+				return null;
+			},
+			$this->features
+		);
```
felixarntz, line 118: I think it'd be great to flatten the JSON structure internally once to get an array where the keys are the full identifiers including periods. This will make lookups more efficient because we wouldn't need to flatten on every `is_feature_enabled` call.

includes/Plugin.php
```diff
@@ -239,6 +242,14 @@ public static function load( $main_file ) {
 			return false;
 		}
 
+		$config = new JSON_File( GOOGLESITEKIT_PLUGIN_DIR_PATH . 'dist/config.json' );
+		Feature_Flags::set_instance(
+			new Feature_Flags(
+				new JSON_File( GOOGLESITEKIT_PLUGIN_DIR_PATH . 'feature-flags.json' ),
+				$config['flagMode']
```
felixarntz, line 249: Right now, this value may be empty (in production we won't ship the `config.json` file for now), which shouldn't be the case, as it would then bypass the `production` default and run the `googlesitekit_flag_mode` filter with an empty value. Let's make this more stable, we should pass `production` if `$config['flagMode']` is empty.

includes/Core/Util/JSON_File.php
```diff
@@ -0,0 +1,139 @@
...
+use JsonSerializable;
+
+/**
+ * Class for handling access to JSON files.
+ *
+ * @since n.e.x.t
+ * @access private
+ * @ignore
+ */
+class JSON_File implements ArrayAccess, JsonSerializable {
```
felixarntz, line 24: Nice!

includes/Core/Util/Feature_Flags.php
```diff
@@ -0,0 +1,140 @@
...
+	/**
+	 * Checks if the given feature is enabled in the current mode on the main instance.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $feature Feature key path to check.
+	 * @return bool
+	 */
+	public static function enabled( $feature ) {
+		return static::get_instance()->is_feature_enabled( $feature );
```
felixarntz, line 80: When called too early, this could cause a fatal. Maybe we should cater for that and return `false` and possibly a `_doing_it_wrong`.

includes/Core/Util/Feature_Flags.php
```diff
@@ -0,0 +1,140 @@
...
+
+	/**
+	 * Gets the main feature flags instance.
+	 *
+	 * @since n.e.x.t
+	 * @internal
+	 *
+	 * @return Feature_Flags
+	 */
+	public static function get_instance() {
```
felixarntz, line 55: I'm not sure an instance here is needed. While that's usually our preferred approach, we only need the static `enabled` method here. We should still be able to write integration tests via some internals static `set_config` or `set_mode` methods.

I'd prefer if we went truly static here which would also allow us to get rid of some unnecessary code.

felixarntz (changes requested):
I have a few points of feedback on the implementation. Most importantly though the last section of the ACs is missing, which is to implement the first actual usage of this.

includes/Core/Util/Feature_Flags.php
aaemnnosttv: This doesn't flatten out the entire array of features, it only "drills down" to get the value of the specific feature that was requested so it's only following one "path" through the object. This allows us to support multiple values for `enabled` which is also supported by the webpack plugin that this configuration was based on. Otherwise, we'd have to make certain assumptions about the configuration in order to flatten it (e.g. the value is always assigned to a "enabled" key, even though that isn't required).

If you're concerned about the cost, it would probably be simpler and more efficient to memoize the flattening of `$feature => $value` on-the-fly here rather than flattening the entire structure all at once which would require recursively iterating down all paths (used or not). In reality, I think it's already simple enough that such optimization isn't necessary, but I'm happy to go whichever way you'd like with it 🙂

felixarntz: Oh that's a good point actually, I missed that. We shouldn't require people to pass a feature path like `widgets.dashboard.enabled`, the feature path should be `widgets.dashboard`. We can define the convention for how this is declared in the JSON ourselves, so I'm okay assuming that the value is always assigned to an `enabled` key.

We could also get rid of the `enabled` nesting level and put the value right with the actual feature name, but I wouldn't wanna change this now since it would also require changing it everywhere in the JS codebase. Let's alter this code here so that we don't need to pass `enabled`, as that shouldn't be part of the feature name / feature path.

tests/phpunit/includes/TestCase.php
```diff
@@ -11,15 +11,38 @@
...
+			self::$featureFlagsConfig = new JSON_File( GOOGLESITEKIT_PLUGIN_DIR_PATH . 'feature-flags.json' );
+		}
+
+		self::reset_feature_flags();
+	}
+
+	public static function tearDownAfterClass() {
+		parent::tearDownAfterClass();
+		self::reset_feature_flags();
+	}
```
aaemnnosttv, line 39: I chose these so that classes could do their own setup/teardown for each test but it wouldn't leak between test cases. The JSON file is only instantiated and read once as well.

felixarntz (approved):
Awesome work! I've updated the QA Brief to also include the requirement to land on the `googlesitekit-user-input` screen at the end, which is the goal of why the user input state needs to be set to `required`.

Issue #2383, opened 2020-11-17
Fix unreliable logic to update site URLs registered on authentication service


felixarntz:
#1857 made a significant improvement for supporting sites which have changed their URL, by detecting whether the URL has changed compared to the URL that the site had been set up with. Users will be automatically disconnected from those sites, and once they reconnect, an extra request will be made to the authentication service to update the URLs.

However, there are some scenarios where this logic can be bypassed, for example:
1. Site Kit is used on a staging site.
2. The site is then being launched to production on another URL, maintaining the database (so that site ID and site secret are still there).
3. However, the host provides automatic functionality that helps with the migration, replacing all old staging URLs with the new URL in the database, which unfortunately will also replace the URL that was intentionally stored as "original URL" by Site Kit.
4. The result is that Site Kit does not detect that the URL has changed because it also changed in the option to compare with.

We should make the flow for updating the authentication service URLs as needed more reliable, which is what this issue is about. We should keep the local URL detection itself as is, since disconnecting the user with an explanation is still a more seamless UX. This issue should solely focus on introducing the ability to change the authentication service URLs even if no local change has been detected. This can be accomplished by introducing new route functionality on the authentication service which allows a connected site to access its currently stored information.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new method `Google_Proxy::fetch_site_fields( Credentials $credentials )` should be introduced:
    * It should return early if there are no credentials.
    * It should issue a similar request as `Google_Proxy::sync_site_fields` does, however **only** passing the `site_id` and `site_secret` body parameters.
    * On success, it should `json_decode` the response as associative array and return it. On error, it should return the `WP_Error`.
* A new method `Google_Proxy::are_site_fields_synced( Credentials $credentials )` should be introduced:
    * It should call the above new "remote" `fetch_site_fields` method, as well as the existing "local" `get_site_fields` method.
    * It should return the error if `fetch_site_fields` returned an error.
    * It should return `true` if all keys in the local site fields are also in the remote site fields with the same values. Otherwise it should return `false`.
* The `Authentication::handle_sync_site_fields` method should be modified: The condition to call `Google_Proxy::sync_site_fields` should be modified to check if the above `Google_Proxy::are_site_fields_synced` returns `false`. In other words, every time a user connects, we should check the current site info from the proxy and update it if needed.
[...]

PR #2439, opened 2020-11-26
Fix unreliable logic to update site URLs registered on authentication service

benbowler:
## Summary

Addresses issue #2383

## Relevant technical choices

Although PHP allows comparison of two arrays using `==` I chose to compare by looping through the get_site_fields and matching each to the fetched values so that we can match for type as well as value while still ignoring the order of the items in the array.

```
foreach ( $get_site_fields as $key => $site_field ) {
	if ( ! array_key_exists( $key, $fetch_site_fields ) || $fetch_site_fields[ $key ] !== $site_field ) {
		return false;
	}
}
```

## Checklist


eugene-manuilov (changes requested):
Hi @benbowler. Is this ready for review? I don't see the following requirement from IB to be implemented. Perhaps you forgot to push your changes?

```
Update the Authentication::handle_sync_site_fields method to call $this->google_proxy->sync_site_fields( ... ) only if $this->google_proxy->are_site_fields_synced( ... ) returns FALSE;
```

benbowler:
You're right @eugene-manuilov, I've updated that function as per the IB.

includes/Core/Authentication/Authentication.php
```diff
@@ -1152,7 +1152,7 @@ private function handle_sync_site_fields() {
 			wp_die( esc_html__( 'Site Kit is not configured to use the authentication proxy.', 'google-site-kit' ) );
 		}
 
-		if ( $this->disconnected_reason->get() === Disconnected_Reason::REASON_CONNECTED_URL_MISMATCH ) {
+		if ( ! $this->google_proxy->are_site_fields_synced( $this->credentials ) ) {
```
eugene-manuilov, line 1155: @benbowler let's explicitly define that we need to sync site fields when the `are_site_fields_synced` method returns `FALSE` because it can also return an instance of the `WP_Error` class.

```suggestion
		if ( $this->google_proxy->are_site_fields_synced( $this->credentials ) === false ) {
```

includes/Core/Authentication/Google_Proxy.php
```diff
@@ -95,6 +96,75 @@ public function get_site_fields() {
 		);
 	}
 
+	/**
+	 * Fetch site fields
+	 *
+	 * @since x.x.x
```
eugene-manuilov, line 102:
```suggestion
	 * @since n.e.x.t
```

includes/Core/Authentication/Google_Proxy.php
```diff
@@ -95,6 +96,75 @@ public function get_site_fields() {
...
+			);
+		}
+
+		return $response_data;
+	}
+
+	/**
+	 * Are site fields synced
+	 *
+	 * @since x.x.x
```
eugene-manuilov, line 144:
```suggestion
	 * @since n.e.x.t
```

tests/phpunit/integration/Core/Authentication/Google_ProxyTest.php
```diff
@@ -43,6 +44,134 @@ public function test_get_site_fields() {
 		);
 	}
 
+	public function test_fetch_site_fields() {
```
eugene-manuilov, line 47: @benbowler let's also add a case when proxy returns an error to make sure that `fetch_site_fields` returns an appropriate WP_Error instance.

tests/phpunit/integration/Core/Authentication/Google_ProxyTest.php
```diff
@@ -43,6 +44,134 @@ public function test_get_site_fields() {
...
+		// Ensure the request was made with the proper URL and body parameters.
+		$this->assertEquals( $google_proxy->url( Google_Proxy::OAUTH2_SITE_URI ), $pre_url );
+		$this->assertEquals( 'POST', $pre_args['method'] );
+		$this->assertEqualSetsWithIndex(
+			array(
+				'site_id',
+				'site_secret',
+			),
+			array_keys( $pre_args['body'] )
+		);
```
eugene-manuilov, line 146: @benbowler no need to assert functionality from the `fetch_site_fields` method. It is already checked in its own test.

tests/phpunit/integration/Core/Authentication/Google_ProxyTest.php
```diff
@@ -43,6 +44,134 @@ public function test_get_site_fields() {
...
+		$this->mock_http_request(
+			$google_proxy->url( Google_Proxy::OAUTH2_SITE_URI ),
+			$mock_non_matching_response
+		);
+		$failure_response_data = $google_proxy->are_site_fields_synced( $credentials );
+
+		// Ensure matching response array returns true.
+		$this->assertEquals( $success_response_data, true );
+		// Ensure WP_Error response returns WP_Error.
+		$this->assertEquals( is_wp_error( $error_response_data ), true );
```
eugene-manuilov, line 170: @benbowler let's use `$this->assertWPErrorWithMessage`, defined in the `tests/phpunit/includes/TestCase.php` file, to better evaluate the error object.

eugene-manuilov (changes requested):
Thanks, @benbowler. It's a good start. Added some minor comments to address before I can approve it.

benbowler:
Thanks @eugene-manuilov, all updated.

includes/Core/Authentication/Google_Proxy.php
```diff
@@ -95,6 +95,72 @@ public function get_site_fields() {
...
+	 * Fetch site fields
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Credentials $credentials Credentials instance.
+	 * @return array|WP_Error The response as an associative array or WP_Error on failure.
+	 */
+	public function fetch_site_fields( Credentials $credentials ) {
+		if ( ! $credentials->has() ) {
+			return null;
```
eugene-manuilov, line 108: @benbowler let's return an error if we don't have credentials.

includes/Core/Authentication/Google_Proxy.php
```diff
@@ -95,6 +95,72 @@ public function get_site_fields() {
...
+				'site_id'     => $creds['oauth2_client_id'],
+				'site_secret' => $creds['oauth2_client_secret'],
+			),
+		);
+
+		$response = wp_remote_post( $this->url( self::OAUTH2_SITE_URI ), $request_args );
+
+		if ( is_wp_error( $response['body'] ) ) {
+			return $response['body'];
+		}
```
eugene-manuilov, line 124: We need to check the response object itself, not its body:

```suggestion
		if ( is_wp_error( $response ) ) {
			return $response;
		}
```

eugene-manuilov (changes requested):
Thanks, @benbowler! Just two more minor things to address and I'll approve it.

includes/Core/Authentication/Google_Proxy.php
benbowler: This does not work because the $reponse is a standard array, the WP_Error or other results are returned in the body of the request. In the initial PR I fixed this by testing `is_wp_error($raw_body)`, we can do this or test the `$reponse['body']` let me know which you prefer.

```
var_dump($response);
```

```
array(5) {
  ["headers"]=>
  array(0) {
  }
  ["body"]=>
  object(WP_Error)#3664 (2) {
    ["errors"]=>
    array(1) {
      ["test_message"]=>
      array(1) {
        [0]=>
        string(12) "test_message"
      }
    }
    ["error_data"]=>
    array(0) {
    }
  }
  ["response"]=>
  array(2) {
    ["code"]=>
    int(200)
    ["message"]=>
    string(2) "OK"
  }
  ["cookies"]=>
  array(0) {
  }
  ["http_response"]=>
  NULL
}
```

eugene-manuilov: @benbowler see the documentation for the [`wp_remote_post`](https://developer.wordpress.org/reference/functions/wp_remote_post/) function:

```
Return #
(array|WP_Error) The response or WP_Error on failure.
```

benbowler:
Hey @eugene-manuilov, I've updated the WP_Error for oauth_credentials_not_exist. For the other request, it's not possible to check the raw `$response` array `is_wp_error`, I've left a comment above and reverted to my original implementation of checking `$raw_body` `is_wp_error`.

benbowler:
Updated.

benbowler:
@eugene-manuilov

includes/Core/Authentication/Google_Proxy.php
```diff
@@ -95,6 +95,72 @@ public function get_site_fields() {
...
+	 * Fetch site fields
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Credentials $credentials Credentials instance.
+	 * @return array|WP_Error The response as an associative array or WP_Error on failure.
+	 */
+	public function fetch_site_fields( Credentials $credentials ) {
+		if ( ! $credentials->has() ) {
+			return new WP_Error( 'oauth_credentials_not_exist' );
```
eugene-manuilov, line 108: @benbowler, please, add `use WP_Error;` to the top of the class to make it work here.

includes/Core/Authentication/Google_Proxy.php
```diff
@@ -95,6 +95,72 @@ public function get_site_fields() {
...
+			),
+		);
+
+		$response = wp_remote_post( $this->url( self::OAUTH2_SITE_URI ), $request_args );
+
+		if ( is_wp_error( $response ) ) {
+			return $response;
+		}
+
+		$raw_body = wp_remote_retrieve_body( $response );
```
eugene-manuilov, line 126: @benbowler let's also make sure that the response status code is 200:

```php
if ( wp_remote_retrieve_response_code( $response ) != 200 ) {
    return new WP_Error( ... );
}

$raw_body = wp_remote_retrieve_body( $response );
```

tests/phpunit/integration/Core/Authentication/Google_ProxyTest.php
```diff
@@ -43,6 +44,121 @@ public function test_get_site_fields() {
...
+			},
+			10,
+			3
+		);
+
+		// Force WP_Error response from as http requests are blocked.
+		$this->mock_http_request(
+			$google_proxy->url( 'http://example.com' ),
+			null
+		);
```
eugene-manuilov, line 74: @benbowler let's better use the real URL for it. You can create another helper function to test errors. Something like this:

```php
private function mock_http_failure( $request_url, $response_error ) {
    add_filter(
        'pre_http_request',
        function( $response, $parsed_args, $url ) use ( $request_url, $response_error ) {
            if ( $url === $request_url ) {
                return new WP_Error( $response_error );
            } else {
                return $response;
            }
        },
        10,
        3
    );
}
```

eugene-manuilov (changes requested):
Thanks, @benbowler. Found a few more things to improve. Could you please check it?

eugene-manuilov (dismissed):
LGTM! Thanks, @benbowler!

includes/Core/Authentication/Google_Proxy.php
```diff
@@ -95,6 +96,76 @@ public function get_site_fields() {
...
+
+		$response = wp_remote_post( $this->url( self::OAUTH2_SITE_URI ), $request_args );
+
+		if ( is_wp_error( $response ) ) {
+			return $response;
+		}
+
+		if ( wp_remote_retrieve_response_code( $response ) !== 200 ) {
+			return new WP_Error( 'non_200_response_code' );
+		}
```
felixarntz, line 129: This is somewhat unprecise, as the `WP_Error` returned would not provide information on the real error here - which may be okay as a last resort, but we should try to use more specific errors. For example, below you run the check whether the response contains an `error` key, which is possible, and in that case the response status code would not be 200. So we'd want to in such a case return the code in `error` rather than a less precise error that the response didn't come back with a 200.

I think we should just remove this clause completely. If the entire request fails, that would be caught by the `WP_Error` from `wp_remote_post`, and if the request comes back with an error response, we should be okay relying on the `error` key in the response.

tests/phpunit/integration/Core/Authentication/Google_ProxyTest.php
```diff
@@ -218,20 +334,24 @@ public function test_exchange_site_code() {
...
 				}
 
+				if ( is_array( $response_data ) ) {
+					$response_data = json_encode( $response_data );
+				}
+
 				return array(
 					'headers'       => array(),
-					'body'          => json_encode( $response_data ),
+					'body'          => $response_data,
```
felixarntz, line 354: Why was this changed? I don't see any usage where `$response_data` is not an array, and I think it's better to keep the signature consistent, especially if there's no reason to change it.

tests/phpunit/integration/Core/Authentication/Google_ProxyTest.php
```diff
@@ -244,4 +364,26 @@ function ( $preempt, $args, $url ) use ( $request_url, $response_data, $response
 			3
 		);
 	}
+
+	/**
+	 * Adds a 'pre_http_request' filter that will ensure the request for the
+	 * given URL will return WP_Error.
+	 *
+	 * @param string $request_url   Request URL to modify response for.
+	 * @param (WP_Error)  $response_error Response WP_Error to return.
```
felixarntz, line 373: This shouldn't be in parentheses, and the parameter docs should be aligned.
```suggestion
	 * @param string   $request_url    Request URL to modify response for.
	 * @param WP_Error $response_error Response WP_Error to return.
```

felixarntz (changes requested):
Great! Just some minor feedback on things that could be improved.

Normally I wouldn't block merging on such changes, but since this is still awaiting for the authentication service side to be launched, I think it's worth making these minor enhancements.

tests/phpunit/integration/Core/Authentication/Google_ProxyTest.php
benbowler: In a previous commit, I updated this function to allow tests to pass a WP_Error via $response_data before splitting it out later into `mock_http_failure`.

I've added this type back in. 👌

includes/Core/Authentication/Google_Proxy.php
benbowler: Great, updated and pushed in the most recent push.

tests/phpunit/integration/Core/Authentication/Google_ProxyTest.php
```diff
@@ -229,9 +345,13 @@ function ( $preempt, $args, $url ) use ( $request_url, $response_data, $response
 					return $preempt;
 				}
 
+				if ( is_array( $response_data ) ) {
+					$response_data = json_encode( $response_data );
+				}
```
felixarntz, line 350: The check here is also no longer necessary, since it'll always be an array. We can unconditionally `json_encode` it like before.

felixarntz (approved):
Awesome! Logic on our end is now pushed to the staging version of the authentication service, and I've updated the QA Brief with the related information for testing.

Issue #2334, opened 2020-11-06
Flush browser/session cache when plugin upgraded.


adamsilverstein:
## Feature Description

On the Admin->Plugins screen after the user upgrades Site Kit to the latest version, we should flush any session data stored in the user's browser (ie. sessionStorage). Otherwise this cached data may not work correctly when they return to Site Kit (for example if the expected cached data or data schema has changed). 

Question: when the plugin is upgraded in the background (eg auto-updates), should we still endeavor to flush local/browser caches? This would still be straightforward if the plugin version were stored in sessionStorage and compared with each load. This situation seems much less likely to be encountered by actual users though.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Site Kit `sessionStorage` and `localStorage` items from a previous release should not be used for the current version after upgrading  

## Implementation Brief

* Edit the webpack config file:
  * Import `DefinePlugin` from the `webpack` package;
  * Add a new instance of the `DefinePlugin` class to the plugins list of the main configuration (with all javascript files):
    * The constructor should accept an object with `${ global.GOOGLESITEKIT_VERSION || '' }` property that contains the plugin version read from the `google-site-kit.php` file and appended with an underscore (something like `1.20.0_`). Use the following snippet to read version from the PHP file:
      ```js
      fs.readFileSync( path.resolve( __dirname, 'google-site-kit.php' ) )?.match( /(?<!GOOGLESITEKIT_VERSION',\s+')\d+\.\d+\.\d+(?=')/ig )?.[0] || ''
      ```
* Update the `STORAGE_KEY_PREFIX` constant in the `assets/js/googlesitekit/api/cache.js` file to get the following value instead of what it has right now:
  ```js
  `googlesitekit_${ global.GOOGLESITEKIT_VERSION || '' }`
  ```
  `DefinePlugin` will update it to be `'googlesitekit_1.20.0_'` during compilation and we will get versioned prefixes for the session storage.
* Update the `STORAGE_KEY_PREFIX` constant in the `assets/js/components/data/cache.js` file using the same approach - it should have the following value:
  ```js
  `googlesitekit_legacy_${ global.GOOGLESITEKIT_VERSION || '' }`
  ```

### Test Coverage

* Not needed

### Visual Regression Changes

* No visual changes are expected

## QA Brief

* Make sure session storage uses prefixes with the current version of the plugin to store data;
* Check that browser caching keeps working as before.

## Changelog entry

[...]

PR #2441, opened 2020-11-26
Flush browser/session cache when plugin upgraded.

benbowler:
## Summary

Addresses issue #2334

## Relevant technical choices

The optional chaining operator did not work in the webpack file so I had to simplify the version reading code.

## Checklist


assets/js/components/data/cache.js
```diff
@@ -27,7 +27,7 @@ import { cloneDeep } from 'lodash';
 import { getStorage } from '../../util/storage';
 import { stringifyObject } from '../../util/stringify';
 
-export const STORAGE_KEY_PREFIX = 'googlesitekit_legacy_';
+export const STORAGE_KEY_PREFIX = `googlesitekit_legacy_${ global.GOOGLESITEKIT_VERSION || '' }`;
```
eugene-manuilov, line 30: Unfortunately, looks like DefinePlugin can't replace things in string templates. So this doesn't work and we need to update it to the following:

```suggestion
export const STORAGE_KEY_PREFIX = 'googlesitekit_' + ( global.GOOGLESITEKIT_VERSION || '' );
```

assets/js/googlesitekit/api/cache.js
```diff
@@ -24,7 +24,7 @@
  * @since 1.5.0
  * @private
  */
-export const STORAGE_KEY_PREFIX = 'googlesitekit_';
+export const STORAGE_KEY_PREFIX = `googlesitekit_${ global.GOOGLESITEKIT_VERSION || '' }`;
```
eugene-manuilov, line 27: Same here:
```suggestion
export const STORAGE_KEY_PREFIX = 'googlesitekit_' + ( global.GOOGLESITEKIT_VERSION || '' );
```

webpack.config.js
```diff
@@ -262,6 +262,9 @@ const webpackConfig = ( env, argv ) => {
 						return content;
 					},
 				} ),
+				new DefinePlugin( {
+					GOOGLESITEKIT_VERSION: fs.readFileSync( path.resolve( __dirname, 'google-site-kit.php' ), 'utf8' ).match( /(?<!GOOGLESITEKIT_VERSION',\s+')\d+\.\d+\.\d+(?=')/ig )[ 0 ] || '',
```
eugene-manuilov, line 266: A few things here:

1. The key should be `global.GOOGLESITEKIT_VERSION`
1. The regular expression doesn't seem to work and returns incorrect data. We probably need to rework it a little bit and use another approach to extract the version from the PHP file. Leaving it to you to figure out the correct way to do it.
1. The value should be appended with `_`.
1. The value should be JSON encoded using the `JSON.stringify()` function.

eugene-manuilov (changes requested):
Hey @benbowler! I have tested it on my local and it doesn't seem to work at all. Have you tested it on your end before submitting it for review? Left some comments that we need to address before we can merge it.

webpack.config.js
benbowler: Hey @eugene-manuilov,

1. Updated.
2. For me the regex does work, let me know if you can't see it passing through collectly.
3. Done.
4. Done.

eugene-manuilov: @benbowler the regular expression selects a version from the `GOOGLESITEKIT_PHP_MINIMUM` constant instead of `GOOGLESITEKIT_VERSION`. The following is what I see in the generated `dist/assets/js/googlesitekit-api.xxxx.js` file. As you can see it uses `5.6.0` instead of `1.21.0`.

```js
/**
 * Prefix used for all Site Kit keys.
 *
 * Anything not using this key should not be touched by this library.
 *
 * @since 1.5.0
 * @private
 */
var STORAGE_KEY_PREFIX = 'googlesitekit_' + "5.6.0_";
var defaultOrder = ['sessionStorage', 'localStorage'];
var storageBackend;
var storageOrder = [].concat(defaultOrder);
```

benbowler: Yes I see, updated.

eugene-manuilov: Awesome! Thanks!

eugene-manuilov (dismissed):

webpack.config.js
```diff
@@ -140,6 +140,11 @@ const resolve = {
 	modules: [ projectPath( '.' ), 'node_modules' ],
 };
 
+// Get the app version from the google-site-kit.php file - optional chaining operator not supported here
+const googleSiteKitFile = fs.readFileSync( path.resolve( __dirname, 'google-site-kit.php' ), 'utf8' );
+const googleSiteKitVersion = googleSiteKitFile.match( /(?<='GOOGLESITEKIT_VERSION',\s+')\d+.\d+.\d+(?=')/ig );
+const GOOGLESITEKIT_VERSION = googleSiteKitVersion ? googleSiteKitVersion[ 0 ] + '_' : '';
```
aaemnnosttv, line 146: This results in the version having an underscore appended, which is only necessary in the `STORAGE_KEY_PREFIX` – it shouldn't be part of the version itself. I realize this was part of the IB but perhaps that detail was overlooked in the review. Let's keep the version to only include that which was parsed from the plugin file and add the underscore to the storage prefixes.

Also, we already have a function for parsing the plugin version in `bin/release` so it would be nice if we could share that in both places, since they both run in a node context but there isn't really a good way to do that right now. I'll open an issue to address this in the future.

aaemnnosttv (changes requested):
Thanks @benbowler – this is very close, just the one detail that seems to have been missed from the IB review which should be slightly different.

webpack.config.js
benbowler: No problem, updated.

aaemnnosttv (approved):
LGTM 👍

Issue #2237, opened 2020-10-22
Pass actual dates in API calls in Search Console widgets


felixarntz:
The Search Console `dashboard` and `pageDashboard` widgets should be migrated to pass `startDate` and `endDate` instead of `dateRange` with their API requests.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The following widgets should pass `startDate` and `endDate` (using the `core/user` `getDateRangeDates` selector) instead of `dateRange` to their `getReport` calls:
    * `DashboardPopularKeywordsWidget`
* The following widgets should pass `startDate` and `endDate` (using the `core/user` `getDateRangeDates` selector) instead of `dateRange` and `compareDateRanges` to their `getReport` calls:
    * `DashboardClicksWidget`
    * `DashboardImpressionsWidget`
* The dates sent should match (compare the API responses) what the server would have previously computed from the passed `dateRange` (and related) parameter (e.g. depending on the `offsetDays` value passed to `getDateRangeDates`).

## Implementation Brief

* Update the `DashboardPopularKeywordsWidget` component to use the `getDateRangeDates` selector from the user datastore to pull `startDate` and `endDate` dates. These dates must be set as `startDate` and `endDate` properties in the report args object instead of just `dateRange`.
* Update `DashboardClicksWidget` and `DashboardImpressionsWidget` components to use the `getDateRangeDates` selector from the user datastore to pull `compareStartDate` and `compareEndDate` dates using `compare: true` option. These dates must be set as `startDate` and `endDate` properties in the report args instead of just `dateRange`, plus `compareDateRanges: true` argument should be removed.

### Test Coverage

* Not needed.

### Visual Regression Changes

* Not expected.

## QA Brief

* Make sure the aforementioned widgets use the new date range dates instead of ranges.

## Changelog entry

* Move date awareness for Search Console API requests to the client, allowing for more contextual date information to be available.

felixarntz:
@eugene-manuilov IB mostly ✅ , however I think it'd be worth adding that when calling `getDateRangeDates` with `compare: true`, it will also return `compareStartDate` and `compareEndDate` (which are the compare period, before the main period). To make that work, we'll need to pass `compareStartDate` as `startDate` to the API, so that the results basically encompass both ranges in one.

eugene-manuilov:
@felixarntz yep, updated.

felixarntz:
[...]

Issue #2236, opened 2020-10-22
Pass actual dates in API calls in Analytics widgets


felixarntz:
The Analytics `dashboard` and `pageDashboard` widgets should be migrated to pass `startDate` and `endDate` instead of `dateRange` (as well as `compareStartDate` and `compareEndDate` where applicable) with their API requests.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The following widgets should pass `startDate` and `endDate` (using the `core/user` `getDateRangeDates` selector) instead of `dateRange` to their `getReport` calls:
    * `DashboardAllTrafficWidget`
    * `DashboardPopularPagesWidget`
* The following widgets should pass `startDate`, `endDate`, `compareStartDate`, and `compareEndDate` (using the `core/user` `getDateRangeDates` selector) instead of `dateRange` and `multiDateRange` to their `getReport` calls:
    * `DashboardBounceRateWidget`
    * `DashboardGoalsWidget`
    * `DashboardUniqueVisitorsWidget`
* The dates sent should match (compare the API responses) what the server would have previously computed from the passed `dateRange` (and related) parameter (e.g. depending on the `offsetDays` value passed to `getDateRangeDates`).

#### `DashboardAllTrafficWidget`, `DashboardPopularPagesWidget`

* Select `getDateRangeDates` as `dateRangeDates`
	- with `offsetDays: 1`
* Update `args`
	- Remove `dateRange`
	- Merge with selected `dateRangeDates`

#### `DashboardBounceRateWidget`, `DashboardGoalsWidget`

* Select `getDateRangeDates` as `dateRangeDates`
	- with `offsetDays: 1`
	- with `compare: true`
	- with `weekdayAlign: true`
* Update `args`
	- Remove `dateRange`
	- Merge with selected `dateRangeDates`

#### `DashboardUniqueVisitorsWidget`

* Update `commonArgs`
	- Remove `dateRange`
	- Set `url` unconditionally (only has effect if not empty)
* Select `getDateRangeDates` as `sparkRangeDates`
	- with `offsetDays: 1`
* Update `sparklineArgs`
	- Merge with selected `sparkRangeDates`
* Select `getDateRangeDates` as `dateRangeDates`
	- with `offsetDays: 1`
	- with `compare: true`
	- with `weekdayAlign: true`
* Update `args`
	- Merge with selected `dateRangeDates`

### Test Coverage

* No changes necessary

### Visual Regression Changes

[...]

PR #2447, opened 2020-11-27
Pass startDate and endDate in API calls in Search Console widgets.

benbowler:
## Summary

Pass `startDate` and `endDate` in API calls in Search Console widgets.
Addresses issue #2237

## Checklist


benbowler:
I've updated each of the three widgets to use the separate `startDate` and `endDate` or `compareStartDate` and `compareEndDate` and tested the range matches the previous value passed via `dateRange`.

tofumatt (dismissed):

assets/js/modules/search-console/components/dashboard/DashboardClicksWidget.js
```diff
@@ -52,10 +52,11 @@ function DashboardClicksWidget() {
 		const isDomainProperty = select( STORE_NAME ).isDomainProperty();
 		const referenceSiteURL = untrailingslashit( select( CORE_SITE ).getReferenceSiteURL() );
 
+		const dateRange = select( CORE_USER ).getDateRangeDates( { compare: true } );
 		const args = {
 			dimensions: 'date',
-			compareDateRanges: true,
-			dateRange: select( CORE_USER ).getDateRange(),
+			startDate: dateRange.compareStartDate,
+			endDate: dateRange.compareEndDate,
```
felixarntz, line 59: The IB here is actually slightly incorrect, I've missed that before. My comment https://github.com/google/site-kit-wp/issues/2237#issuecomment-728529776 has the intended behavior though: The `compare*Date` values represent a secondary date range (in addition to `startDate` and `endDate`) that occurs right before that main date range. However the Search Console API doesn't support passing two date ranges, so what we'll need to do is pass it as follows to encompass the entirety of the two date ranges:
```suggestion
			// Combine both date ranges into one single date range.
			startDate: dateRange.compareStartDate,
			endDate: dateRange.endDate,
```

assets/js/modules/search-console/components/dashboard/DashboardClicksWidget.js
```diff
@@ -52,10 +52,11 @@ function DashboardClicksWidget() {
 		const isDomainProperty = select( STORE_NAME ).isDomainProperty();
 		const referenceSiteURL = untrailingslashit( select( CORE_SITE ).getReferenceSiteURL() );
 
+		const dateRange = select( CORE_USER ).getDateRangeDates( { compare: true } );
```
felixarntz, line 55: Another thing missing from the IB (🤦‍♂️ ) is that we'll need to pass the `offsetDays` argument to match the behavior of how the date range is parsed in PHP (where the offset is 1). The default in the JS selector is `0`, so we'll need to pass that as `1`.

assets/js/modules/search-console/components/dashboard/DashboardImpressionsWidget.js
```diff
@@ -52,10 +52,11 @@ function DashboardImpressionsWidget() {
 		const isDomainProperty = select( STORE_NAME ).isDomainProperty();
 		const referenceSiteURL = untrailingslashit( select( CORE_SITE ).getReferenceSiteURL() );
 
+		const dateRange = select( CORE_USER ).getDateRangeDates( { compare: true } );
```
felixarntz, line 55: See above.

assets/js/modules/search-console/components/dashboard/DashboardImpressionsWidget.js
```diff
@@ -52,10 +52,11 @@ function DashboardImpressionsWidget() {
...
 		const args = {
 			dimensions: 'date',
-			compareDateRanges: true,
-			dateRange: select( CORE_USER ).getDateRange(),
+			startDate: dateRange.compareStartDate,
+			endDate: dateRange.compareEndDate,
```
felixarntz, line 59: See above.

assets/js/modules/search-console/components/dashboard/DashboardPopularKeywordsWidget.js
```diff
@@ -52,8 +52,11 @@ function DashboardPopularKeywordsWidget() {
 	} = useSelect( ( select ) => {
 		const store = select( STORE_NAME );
 		const domain = store.getPropertyID();
+
+		const dateRange = select( CORE_USER ).getDateRangeDates();
```
felixarntz, line 56: See above, here we'll also need to pass `offsetDays: 1`.

felixarntz (changes requested):
Against the IB this looks good, however I just noticed two critical (but small) things were missing in the IB, see comments below. Would be great if you could address these.

benbowler:
Morning @felixarntz, I've updated these widgets based on your comments. 👍

assets/js/modules/search-console/components/dashboard/DashboardClicksWidget.js
```diff
@@ -52,10 +52,12 @@ function DashboardClicksWidget() {
 		const isDomainProperty = select( STORE_NAME ).isDomainProperty();
 		const referenceSiteURL = untrailingslashit( select( CORE_SITE ).getReferenceSiteURL() );
 
+		const dateRange = select( CORE_USER ).getDateRangeDates( { compare: true, offsetDays: 1 } );
```
felixarntz, line 55: Can you add a `DATE_RANGE_OFFSET` constant to `assets/js/modules/search-console/datastore/constants.js` and use that in all these locations? This way we put the offset for this module in one place, similarly how it was done for AdSense and Analytics in #2459.

felixarntz (changes requested):
@benbowler So far this looks good, however as mentioned before there is some outstanding work based on what we found via #2459: We'll need to make sure the Storybook stories for `DashboardClicksWidget`, `DashboardImpressionsWidget` and `DashboardPopularKeywordsWidget` continue to show up as before. In order to do that, we'll need to use mock request data that matches the new `startDate` and `endDate` arguments passed to the API. We can use the reference date functionality of the `core/user` store introduced as part of #2459 to work against a fixed reference date so that Storybook (and its mock data) is not dependent on the current date.

This makes the work here a bit more involved, so let's see if we need to adjust the estimate. The same underlying requirements apply to #2236, so we'll need to consider it there as well.

Let me know if you have questions on this, we could also jump on a quick call if necessary.

assets/js/modules/search-console/components/dashboard/DashboardClicksWidget.js
benbowler: Updated and I've added the value as 2 until #2522 is implemented.

benbowler:
Hey @felixarntz, I've tested through and each Search Console story is now working.

stories/module-search-console-components.stories.js
```diff
@@ -63,8 +562,518 @@ generateReportBasedWidgetStories( {
...
+				'2020-08-25',
+			],
+			position: 0,
+		},
+	],
+	referenceDate: '2020-08-27',
+	options: {
+		dimensions: 'date',
+		startDate: '2020-07-01',
+		endDate: '2020-08-25',
```
felixarntz, line 1075: I think the `url` arg is missing here - it's needed since this is the `pageDashboard` version of the widget, i.e. there is a current entity set.

stories/module-search-console-components.stories.js
```diff
@@ -83,8 +1602,518 @@ generateReportBasedWidgetStories( {
...
```
felixarntz, line 2115: See above, the `url` arg needs to be added here.

stories/module-search-console-components.stories.js
```diff
@@ -104,8 +2230,106 @@ generateReportBasedWidgetStories( {
...
+			position: 4.25,
+		},
+	],
+	referenceDate: '2020-08-27',
+	options: {
+		startDate: '2020-07-01',
+		endDate: '2020-08-25',
+		dimensions: 'query',
+		limit: 10,
+		url: 'https://www.sitekit.com/',
```
felixarntz, line 2331: This URL needs to be the current entity URL to match the actual URL used. We might need some extra set up here therefore.

felixarntz (changes requested):
There's still a few last things off with the Storybook config, but we're almost there.

stories/module-search-console-components.stories.js
```diff
@@ -104,8 +2230,106 @@ generateReportBasedWidgetStories( {
...
+			keys: [
+				'sitekit test',
+			],
+			position: 4.25,
+		},
+	],
+	referenceDate: '2020-08-27',
+	options: {
+		startDate: '2020-07-01',
+		endDate: '2020-08-25',
```
felixarntz, line 2328: Actually, the main problem here is this: The request for `DashboardPopularKeywordsWidget` doesn't use `compare: true`, so the date range passed here is incorrect, which makes this story render in an empty state. The `startDate` needs to be `2020-07-29`.

felixarntz (changes requested):

felixarntz (approved):
LGTM!

@benbowler Please review the above comments as they'll be relevant for #2236 as well potentially. I've addressed them in https://github.com/google/site-kit-wp/pull/2447/commits/b272f8fe505ba4c876b1ec1a72030ae5706f6c56 to get this merged. Summarizing:
1. The additions of `url` should be there for all "Page Dashboard" stories, since the "Page Dashboard" is always for a specific URL.
2. One `options` definition here was technically incorrect, which was likely just an oversight.

PR #2449, opened 2020-11-27
Correct the sticky positioning of the header on mobile screen widths.

benbowler:
## Summary

Addresses issue #2331

## Checklist

### DIFF
--- a/assets/sass/components/global/_googlesitekit-header.scss
+++ b/assets/sass/components/global/_googlesitekit-header.scss
@@ -21,18 +21,18 @@
 	@include googlesitekit-inner-padding;
 	@include shadow;
 	background-color: $c-base;
+	left: 0;
+	position: sticky;
+	right: 0;
+	top: 0;
 	z-index: 12;
-
-	// only become sticky when WordPress's #wpadminbar becomes sticky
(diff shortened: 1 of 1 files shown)


assets/sass/components/global/_googlesitekit-header.scss
```diff
@@ -21,20 +21,20 @@
...
-		top: 46px;
-	}
-
 	@media (min-width: $bp-wpAdminBarTablet) {
 		top: 32px;
 	}
 
+	@media (min-width: $bp-tablet) {
+		top: 46px;
+	}
```
aaemnnosttv, line 36: This is the correct rule but should come before the previous media query as it is currently resulting in a gap in larger viewports.
image

image

aaemnnosttv (changes requested):
This is almost good to go @benbowler – it seems to just need a tweak to the ordering of the media queries – please double-check that though 👍

assets/sass/components/global/_googlesitekit-header.scss
benbowler: Good spot.

aaemnnosttv (approved):
Super, thanks!

Issue #2200, opened 2020-10-16
The formatting of numbers followed by % signs can be different per language


ryanwelcher:
## Feature Description

Depending on the language, a number may or may not have a space preceding a % symbol. For example, English is 50% while many other languages would display as 50 %.

Anywhere Site Kit displays a number followed by a % symbol should be [marked for translation](https://github.com/google/site-kit-wp/blob/develop/assets/js/modules/adsense/components/dashboard/AdSensePerformanceWidget.js#L105)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All percent and currency numbers, shown on the admin pages, should be formatted according to the current locale. Special attention should be paid to the languages that use other formatting conventions for numbers and punctuations (like in France) or RTL languages.
* When the user changes languages in the settings, the formatting number should also change accordingly.

## Implementation Brief

* Add a new more Site Kit-specific opinionated version of `numberFormat`, which has the same signature and internally wraps `numberFormat`, but provides defaults according to Site Kit formatting standards.
* Add a new possible value for style, which is `metric`. In this case, the new function should rely on the existing `readableLargeNumber` to format the number. This `metric` value should also be the default `style` for the new function.
* Allow a shorthand for the `options` parameter: Passing `%` should format the value as percentage, passing `s` should format the value as seconds / duration, passing any other string should be interpreted as if this string was a currency code. This allows simple usages like `numFmt( 0.23, '%' )`, `numFmt( 182, 's' )` or `numFmt( 9.99, 'USD' )`.
* For `percent`, the function should expect a number between 0 and 1, like common elsewhere and `numberFormat` as well. Anywhere where we actually use full percentage numbers outside (between 0 and 100), those need to be divided by 100 when passed to the formatting function.
* Here is a draft of that function:
```js
function numFmt( number, options = {} ) { // Name TBD.
	// Expand shorthand values for options.
	if ( '%' === options ) {
		options = { style: 'percent' }
	} else if ( 's' === options ) {
		options = { style: 'seconds' }
	} else if ( typeof options === 'string' ) {
		options = { style: 'currency', currencyCode: options }
	}
	// Note: `metric` is our custom, default style.
	const { style = 'metric' } = options;
​
	if ( 'metric' === style ) {
		return readableLargeNumber( number );
	}
​
	if ( 'seconds' === style ) {
		return prepareSecondsForDisplay( number );
	}
​
	return numberFormat( number, options );
}
```
[...]

Issue #1921, opened 2020-08-17
Display graph of AdSense key metrics over time in AdSense module page


felixarntz:
As a follow-up to #317, the "Performance over the last 28 days" widget on the AdSense module page, which right now only contains aggregate numbers for Earnings, Page RPM, and Impressions, should be expanded to also include a graph for this information over time.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The "Performance over the last 28 days" widget on the AdSense module page should include graphs for the 3 different metrics and behave similarly like "Audience overview for the last 28 days" widget on the Analytics module page (including the #1836 fix).
    * Each of the four metrics can be clicked, showing the respective graph below.
    * Each graph should respect the currently selected date range.
    * By default, the "Earnings" graph should be active.

## Implementation Brief

<details>
<summary><strong>Context/Reference</strong></summary>

The Analytics audience overview dashboard widget is composed of 2 separate components

* The [top component](https://github.com/google/site-kit-wp/blob/21464e98e574d80721aaf4551096e9acd915bf2f/assets/js/modules/analytics/components/dashboard/AnalyticsDashboardWidgetOverview.js) shows the aggregate values and controls the active plots on the chart below it  
![image](https://user-images.githubusercontent.com/1621608/97967586-1819b480-1dc6-11eb-8844-e5f23728d13e.png)
* The [bottom component](https://github.com/google/site-kit-wp/blob/21464e98e574d80721aaf4551096e9acd915bf2f/assets/js/modules/analytics/components/dashboard/AnalyticsDashboardWidgetSiteStats.js) shows the chart with line graphs for the active plots
![image](https://user-images.githubusercontent.com/1621608/97967628-27006700-1dc6-11eb-8c4e-3e12d283e719.png)
* Both components are implemented as class-components using the legacy `withData` HOC, using a total of 3 different queries per date range
    - `overviewReportDataDefaults`
    - `userReportDataDefaults`
    - `siteAnalyticsReportDataDefaults`
* The two components are rendered as children of the main dashboard widget for the module  
https://github.com/google/site-kit-wp/blob/21464e98e574d80721aaf4551096e9acd915bf2f/assets/js/modules/analytics/components/dashboard/AnalyticsDashboardWidget.js#L198-L209

</details>

#### Overview

[...]

PR #2451, opened 2020-11-28
Add numFmt, update the percentage in locale methods.

kostyalmm:
## Summary

Addresses issue #2200

## Relevant technical choices

* Add `numFmt` method to utils.

## Checklist


assets/js/util/i18n.js
```diff
@@ -38,6 +38,40 @@ export const numberFormat = ( number, options = {} ) => {
...
+			maximumFractionDigits: 2,
+			style: 'percent',
+			...options,
+		} );
+	}
+
+	try {
+		return numberFormat( Math.abs( number ), { style: 'currency', currency: unit, ...options } );
+	} catch ( e ) {
+		return `${ numberFormat( Math.abs( number ) ) }${ unit }`;
```
felixarntz, line 71:
```suggestion
		return `${ numberFormat( Math.abs( number ), options ) }${ unit }`;
```

assets/js/util/index.js
```diff
@@ -145,7 +146,7 @@ export const readableLargeNumber = ( number, currencyCode = false ) => {
 	}
 
 	if ( currencyCode ) {
-		return numberFormat( number, { style: 'currency', currency: currencyCode } );
+		return numberFormatWithUnit( number, currencyCode );
```
felixarntz, line 149: I think we should enhance this function so that the second parameter is not `currencyCode`, but `unit` in general. This way it can just internally rely on any unit that it is passed, and if there is a unit call `numberFormatWithUnit`.

Looking at the code for all the numbers we display in the plugin UI in JS again, I think we should update the `DataBlock` component to always run `readableLargeNumber( datapoint, datapointUnit )` and `readableLargeNumber( change, changeDataUnit )` instead of what we have now. Then, all usages of the 4 props should be updated to only pass through raw unformatted numbers and their units in the separate props where applicable. Many areas currently pass the `datapoint` prop as an already formatted value including a unit, which is counterproductive and prone to inconsistent formatting. Let's change that so that `readableLargeNumber` becomes our central formatting function that works regardless of whether there's a unit present or not.

assets/js/components/data-block.js
```diff
@@ -87,11 +87,7 @@ class DataBlock extends Component {
 		if ( change ) {
 			// If changeDataUnit is given, try using it as currency first, otherwise add it as suffix.
 			if ( changeDataUnit ) {
-				try {
-					changeFormatted = numberFormat( Math.abs( change ), { style: 'currency', currency: changeDataUnit } );
-				} catch ( e ) {
-					changeFormatted = `${ numberFormat( Math.abs( change ) ) }${ changeDataUnit }`;
-				}
+				changeFormatted = numberFormatWithUnit( Math.abs( change ), changeDataUnit );
```
felixarntz, line 90: See above, let's make this
```suggestion
				changeFormatted = readableLargeNumber( Math.abs( change ), changeDataUnit );
```

assets/js/components/data-block.js
```diff
@@ -123,7 +119,7 @@ class DataBlock extends Component {
 						{ title }
 					</h3>
 					<div className="googlesitekit-data-block__datapoint">
-						{ `${ datapoint }${ datapointUnit }` }
+						{ datapointUnit ? numberFormatWithUnit( datapoint, datapointUnit ) : datapoint }
```
felixarntz, line 122: See above, let's make this
```suggestion
						{ readableLargeNumber( datapoint, datapointUnit ) }
```

All usages of the `datapoint` prop should be updated to pass unformatted values. In cases where the current formatting (which should be removed) includes a unit, that unit should be passed as `datapointUnit` prop.

assets/js/modules/adsense/components/dashboard/AdSensePerformanceWidget.js
```diff
@@ -103,7 +103,7 @@ class AdSensePerformanceWidget extends Component {
 				className: 'googlesitekit-data-block--impression',
 				title: __( 'Page CTR', 'google-site-kit' ),
 				/* translators: %s: percentage value. */
-				datapoint: sprintf( _x( ' %1$s%%', 'AdSense performance Page CTA percentage', 'google-site-kit' ), numberFormat( currentRangeData.totals[ 3 ] * 100, { maximumFractionDigits: 2 } ) ),
+				datapoint: sprintf( _x( ' %1$s', 'AdSense performance Page CTA percentage', 'google-site-kit' ), numberFormatWithUnit( currentRangeData.totals[ 3 ] * 100, '%', { maximumFractionDigits: 2 } ) ),
```
felixarntz, line 106: There's no need for this translation string anymore now, let's remove it. Based on the other comments and changes, this can simply become
```suggestion
				datapoint: currentRangeData.totals[ 3 ] * 100,
```
The percentage should be passed in the `datapointUnit` prop.

felixarntz (changes requested):
Looking at the code here in more detail, I think we should alter the approach here a bit:
* Let's update `readableLargeNumber` so that it becomes the central formatting function for numbers, regardless of whether or not they have a unit. It's already almost there, we just need to change the `currencyCode` parameter to become `unit`.
* Let's fix the `DataBlock` component so that it centrally handles formatting. All its props should require **unformatted** numbers, so all usages where formatted numbers (potentially with units) are passed to the props should be updated. If a unit is included, that should be passed via the corresponding prop (in most cases `datapointUnit`).
* Naming is potentially a bit confusing, and we could argue that what `numberFormatWithUnit` does could also be incorporated into `readableLargeNumber` directly, making it clear that the latter is the canonical formatting function. Potentially we could have `numberFormatWithUnit` as its own function but not export it. It would also make sense to move `readableLargeNumber` to `assets/js/util/i18n.js` where it would live alongside other number formatting functions that it relies on.

The above are some fairly broad changes, but I think they get us much closer to something that makes sense, with consistent number formatting and _one_ formatting function to use everywhere. Maybe some of the function names are slightly off, but I'd say let's worry about that later.

cc @aaemnnosttv

felixarntz (commented):
@kostyalmm @aaemnnosttv Just so we don't forget, we'll also need to apply some related changes to the new AdSense `getSiteStatsDataForGoogleChart` function after #1921.

kostyalmm:
@felixarntz shall I change `DataBlock`'s `datapoint` to `number`? 
If I do this then almost everything works fine but these two scenarios:
* Time in seconds which is formatted by `prepareSecondsForDisplay`
* Traffic increase win notification story which has the datapoint value as `xx,xxx` 
So shall I keep `datapoint` as `any(string,number)`?

Also `readableLargeNumber` seems a bit of an awkward name, what if we change it to `readableNumber` as it handles not only large but all the numbers. 
What do you think?

felixarntz:
@kostyalmm I've updated the IB for #2200 for the new requirements. I think the new approach is quite different from the original one, so it might be worth opening a completely new PR and close this one - I could see how iterating on this one is more confusing than helpful. But I'll leave that call up to you to make.

> shall I change `DataBlock`'s `datapoint` to `number`?

Yes, I've incorporated that in the new IB as well.

Please review the new IB and let me know if that sounds good to you. Let's not worry about naming too much for now: Let's keep `readableLargeNumber` named as is, and let's name the new function `numFmt`, acknowledging that both names might be changed, either in the future or even before merge. But that's fine, the naming replacement would be trivial.

Due to the change in IB here, I'm assuming this issue will go beyond the original estimate. Would be great if you could make a new estimate based on the current IB and also consider how much time you've already spent on this so that we can have a better idea.

cc @aaemnnosttv @eclarke1

kostyalmm:
@felixarntz can you check the VRT reference changes and approve if this is fine. I can revert it to the way it was before but that will require tinkering with a lot of places that set these `datapoint` values.

I would have to pass `ts` as a unit type, which I introduced [here](https://github.com/google/site-kit-wp/pull/2451/files#diff-35e7b36727377afdfe72372f2fdc85d4d74f6a423a820ad573c77b3c07a1341bR207-R210), everywhere it comes up in the stories' stacktrace. These are the occurences in the stories alone, other places (which are not in stories and may go undetected) where a big number comes will be passed through `numFmt` which in turn will make it go through `readableLargeNumber`.

So are we okay to pass all the numbers (unless explicitly mentioned) through `readableLargeNumber`?

felixarntz:
@kostyalmm The new shorthand `ts` shouldn't be necessary, also it would be an odd name because it's not a unit (these numbers don't have a unit affix). `numFmt` wraps `numberFormat`, so you should be able to call it the same way in the calls where this type of format is needed, passing `style: 'decimal'` within the options.

(`decimal` is the default of `numberFormat`, our own default is `metric` though. So as soon as you pass a `style` other than `metric`, it will go through `numberFormat` as usual.)

assets/js/util/index.js
```diff
@@ -185,20 +172,68 @@ export const readableLargeNumber = ( number, currencyCode = false ) => {
...
- * @param {string} [options.locale] Locale to use for formatting. Defaults to current locale used by Site Kit.
- * @return {string} The formatted number.
+ * @param {number|string}            number    The number to format.
+ * @param {string}                   [unit]    The unit for the number.
+ * @param {Intl.NumberFormatOptions} [options] Formatting options.
+ * @return {string} The formatted number with unit.
  */
-export const numberFormat = ( number, options = {} ) => {
-	const { locale = getLocale(), ...formatOptions } = options;
+export const numFmt = ( number, unit, options = {} ) => {
```
felixarntz, line 184: @kostyalmm See the IB, the signature should be `numFmt( number, options = {} )` in order to match the familiar `numberFormat` function. The `unit` you have here should instead simply be a shorthand version of `options`, to be able to call the function in simpler ways for the most common cases.

Please change this to the signature with the two parameters similar to the preliminary code in the IB.

felixarntz (commented):
I see this isn't ready for review yet, but just saw this one thing here and thought it's worth flagging already.

assets/js/util/index.js
kostyalmm: will we be overloading the `options` parameter as `{object|string}`? That seemed a bit odd to me.
What will happen when we want to use percentage and also pass additional option such as `maximumFractionDigits`?
I made this change keeping this in mind. Please let me know if I should revert this to the IB.

kostyalmm:
> @kostyalmm The new shorthand `ts` shouldn't be necessary, also it would be an odd name because it's not a unit (these numbers don't have a unit affix). `numFmt` wraps `numberFormat`, so you should be able to call it the same way in the calls where this type of format is needed, passing `style: 'decimal'` within the options.
> 
> (`decimal` is the default of `numberFormat`, our own default is `metric` though. So as soon as you pass a `style` other than `metric`, it will go through `numberFormat` as usual.)

Do you mean that I should pass `decimal` everywhere we want thousand-separated numbers?

assets/js/util/index.js
felixarntz: Yes, the `string` is just a shorthand, but yes it should be overloaded in that sense. Since it's just a shorthand, if you wanted to provide more details, you could still pass an object instead with `style: 'percent'` and other options in that case.

felixarntz:
@kostyalmm 
> Do you mean that I should pass `decimal` to `numFmt` everywhere we want thousand-separated numbers?

Yes that's correct, basically `numFmt( number, { style: 'decimal' } )`.

kostyalmm:
Added `decimal` as a custom style as we need to show decimal(thousand separated) numbers in `DataBlock` as well where only `datapointUnit` can be passed (can't pass `{ style: 'decimal' }`).

assets/js/modules/search-console/components/dashboard/DashboardPopularKeywordsWidget.js
```diff
@@ -117,8 +117,8 @@ function DashboardPopularKeywordsWidget() {
 		links[ i ] = addQueryArgs( baseServiceURL, { query: `!${ query }` } );
 		return [
 			query,
-			numberFormat( row.clicks ),
-			numberFormat( row.impressions ),
+			numFmt( row.clicks ),
+			numFmt( row.impressions ),
```
felixarntz, line 121: These need to use `decimal` style.

assets/js/modules/search-console/components/dashboard/LegacyDashboardWidgetPopularKeywordsTable.js
```diff
@@ -87,8 +91,8 @@ const LegacyDashboardWidgetPopularKeywordsTable = ( props ) => {
 		links[ i ] = addQueryArgs( baseServiceURL, { query: `!${ query }` } );
 		return [
 			query,
-			numberFormat( row.clicks ),
-			numberFormat( row.impressions ),
+			numFmt( row.clicks ),
+			numFmt( row.impressions ),
```
felixarntz, line 95: These need to use `decimal` style.

assets/js/modules/search-console/components/dashboard/LegacySearchConsoleDashboardWidgetKeywordTable.js
```diff
@@ -87,8 +87,8 @@ const LegacySearchConsoleDashboardWidgetKeywordTable = ( props ) => {
 		links[ i ] = addQueryArgs( baseServiceURL, { query: `!${ query }` } );
 		return [
 			query,
-			numberFormat( row.clicks ),
-			numberFormat( row.impressions ),
+			numFmt( row.clicks ),
+			numFmt( row.impressions ),
```
felixarntz, line 91: These need to use `decimal` style.

assets/js/modules/search-console/util/index.js
```diff
@@ -58,23 +58,15 @@ function reduceSearchConsoleData( rows ) {
...
-
-	totalClicks = readableLargeNumber( totalClicks );
-	totalImpressions = readableLargeNumber( totalImpressions );
-
 	// Do not divide by zero.
-	const averageCTR = count > 0 ? ( totalCTR / count * 100 ).toFixed( 1 ) : '0.0';
+	const averageCTR = count > 0 ? +( totalCTR / count * 100 ).toFixed( 1 ) : 0.0;
 	const averageCTRRaw = count > 0 ? totalCTR / count : 0.0;
-	const averagePosition = count > 0 ? ( totalPosition / count ).toFixed( 1 ) : '0.0';
+	const averagePosition = count > 0 ? +( totalPosition / count ).toFixed( 1 ) : 0.0;
```
felixarntz, line 64: Why these changes?

assets/js/util/index.js
```diff
@@ -185,20 +172,70 @@ export const readableLargeNumber = ( number, currencyCode = false ) => {
...
+		number /= 100;
+	} else if ( 's' === options ) {
+		formatOptions = {
+			style: 'seconds',
+		};
+	} else if ( 'decimal' === options ) {
+		// Node: `decimal` is our custom style used to return numbers with thousand separators
+		formatOptions = {
+			style: 'decimal',
+		};
```
felixarntz, line 210: Let's remove this and pass `{ style: 'decimal' }` everywhere it's needed instead. The shorthands here are intended for "units" like %, s etc. Actual styles should be passed in the more appropriate `options` object format.

felixarntz (changes requested):
This is becoming more and more of a painful issue/PR because of all the legacy code it touches. Nevertheless, I see great potential here to clean up a lot of mess, so let's stick to it even if it increases the estimate further.

Let's do the following:
* Modify `numFmt` so that it expects percentages to be passed as value between 0 and 1.
* Change usage of `changeToPercent` to a new `calculateChange` wherever applicable to get the value between 0 and 1 instead of full percentage.
* Modify `readableLargeNumber` to call default `numberFormat` for values lower than 1000.
* Remove `decimal` as shorthand for `numFmt`, as it's not a unit.

Let's do these things first, and then review all usages carefully. It will require some extra work, but afterwards we'll eventually have a consistent API and consistent number formatting.

General guideline for whether to use `numFmt( number )` (metric) or `numFmt( number, { style: 'decimal' } )` (decimal):
* Numbers standing on their own (e.g. _all_ usages of `DataBlock` should use 'metric'.
* Numbers in data tables (e.g. data prepared for a data table) should use 'decimal'.

Apologies for the ongoing changes in requirements here @kostyalmm, but I think we can make a big change here by getting this completely right. Can you make the foundational changes mentioned above and update the usages? I'll then double-review those to minimize the chance of _unexpected_ visual breakage.

I'll remove this from the current release so that we have some extra time to spot unexpected changes before pushing this in the following release on January 18.

kostyalmm:
@felixarntz please go through the latest changes once. I updated the code to use `calculateChange` and it looks fine.

I just had to do `/100` in a couple of places as the data from the API is the actual percentage and not the change.

I am not quite sure about this suggestion https://github.com/google/site-kit-wp/pull/2451/files#diff-4543d16612613d83992c7efd3700beb3813157a53a4dbf8b26dd90cdc42ee973R102-R109. If we send empty string or null object, the number will be formatted using `metric` style which we intend. I think it is better off as `numFmt( datapoint, datapointUnit )` and the rest will be handled accordingly by the function. What do you think? I made the change anyways for you to look into it.

I will write/update tests once you approve the overall approach here.

assets/js/util/index.js
```diff
@@ -315,16 +182,29 @@ export const getDaysBetweenDates = ( dateStart, dateEnd ) => {
  * @param {number} current  The current value.
  * @return {(number|string)} The percent change.
  */
-export const changeToPercent = ( previous, current ) => {
+export const changeToPercent = ( previous, current ) =>
+	( calculateChange( previous, current ) * 100 ).toFixed( 1 );
```
felixarntz, line 186: Since it's now no longer used, let's remove this function entirely now.

felixarntz (changes requested):
This is good to go production-code-wise IMO - I've fixed a few last broken formats, the failing JS tests, and a JS error in a few extra commits.

A few remaining things to do here:
1. Let's remove `changeToPercent` since it's no longer needed.
2. Let's add test coverage for `calculateChange`.

After that, I think this is good to be merged. @aaemnnosttv Can you have a look as well?

assets/js/modules/search-console/util/index.js
```diff
@@ -58,25 +58,15 @@ function reduceSearchConsoleData( rows ) {
...
-
-	totalClicks = readableLargeNumber( totalClicks );
-	totalImpressions = readableLargeNumber( totalImpressions );
-
 	// Do not divide by zero.
-	const averageCTR = count > 0 ? ( totalCTR / count * 100 ).toFixed( 1 ) : '0.0';
-	const averageCTRRaw = count > 0 ? totalCTR / count : 0.0;
-	const averagePosition = count > 0 ? ( totalPosition / count ).toFixed( 1 ) : '0.0';
+	const averageCTR = count > 0 ? totalCTR / count : 0.0;
+	const averagePosition = count > 0 ? +( totalPosition / count ).toFixed( 1 ) : 0.0;
```
aaemnnosttv, line 63: We were doing this before but I don't think we should be truncating or otherwise pre-formatting the value here as this should be handled by the formatting function. Unless we specifically care about doing the math this way, we should leave formatting to the utilities.

assets/js/util/i18n.js
```diff
@@ -19,22 +19,189 @@
...
+	}
+	const results = {};
+	results.hours = Math.floor( seconds / 60 / 60 );
+	results.minutes = Math.floor( ( seconds / 60 ) % 60 );
+	results.seconds = Math.floor( seconds % 60 );
+
+	const returnString =
+		( results.hours ? results.hours + 'h ' : '' ) +
+		( results.minutes ? results.minutes + 'm ' : '' ) +
+		( results.seconds ? results.seconds + 's ' : '' );
```
aaemnnosttv, line 55: @felixarntz Perhaps beyond the scope here, but the formatting of time should use `numberFormat` (or `numFmt`) with `style: 'unit', unitDisplay: 'narrow'` and `unit: timeUnit` where `timeUnit` is one of `second`, `minute`, or `hour`.

This would result in the same output, but properly localized to use the correct abbreviations for the unit of time.

Unfortunately there doesn't seem to be a single function in `Intl` for formatting an amount of time in this way using multiple units so the way they are concatenated should probably be translatable as well.

assets/js/util/i18n.js
```diff
@@ -19,22 +19,189 @@
...
+ * @return {string} The formatted number with unit.
+ */
+export const numFmt = ( number, options = {} ) => {
+	// Cast parsable values to numeric types.
+	number = isFinite( number ) ? number : Number( number );
+
+	if ( ! isFinite( number ) ) {
+		// eslint-disable-next-line no-console
+		console.warn( 'Invalid number', number, typeof number );
+		return null;
```
aaemnnosttv, line 151: Do we want to return `null` here? The return tag shows `string` as the only return type.

We should probably change `number` to `0` in this case and let the function continue to format it according to the options.

assets/js/util/i18n.js
```diff
@@ -19,22 +19,189 @@
...
+	} else if ( 's' === options ) {
+		formatOptions = {
+			style: 'seconds',
+		};
+	} else if ( !! options && typeof options === 'string' ) {
+		formatOptions = {
+			style: 'currency',
+			currency: options,
+		};
+	} else if ( typeof options === 'object' ) {
```
aaemnnosttv, line 171: We should probably use `isPlainObject` here to avoid complication with the `typeof null === 'object'` bug.

assets/js/util/i18n.js
```diff
@@ -19,22 +19,189 @@
...
+	// Expand shorthand values for units.
+	if ( '%' === options ) {
+		formatOptions = {
+			style: 'percent',
+			maximumFractionDigits: 2,
+		};
+	} else if ( 's' === options ) {
+		formatOptions = {
+			style: 'seconds',
+		};
```
aaemnnosttv, line 165: I'm not sure this is really accurate. `s` for seconds is a unit, but I would expect it to be formatted as `X sec` or `Xs` (as formatted by `{ style: 'unit', unit: 'second' }`, but this is instead formatting as a human-readable _duration_ from the total given seconds by reducing the number in seconds to the equivalent number of larger units of time where applicable.

If we need to support a shorthand for this as a unit, I'm not sure there is an actual unit we can use to accurately represent this. Perhaps we should remove the concept of a "unit" from the second parameter (`percent` isn't actually a unit either).

That's all to say, I think it's okay to have a shorthand for this, but the actual (custom) `style` should probably be `duration` which would be easier to understand in its usage than `s` as well IMO.

assets/js/util/i18n.js
```diff
@@ -19,22 +19,189 @@
...
+	}
+
+	try {
+		return numberFormat( number, formatOptions );
+	} catch ( e ) {
+		if ( !! options && typeof options === 'string' ) {
+			// if the formatting doesn't work, return the concatenated options(assuming as a custom unit) and number
+			return `${ number }${ options }`;
+		}
+	}
```
aaemnnosttv, line 193: @felixarntz do we know what conditions where the fallback would actually be used here? I know it's the fallback but simply appending the unit isn't very i18n-friendly, however making a translatable string for generally appending an unknown unit to a number isn't i18n-friendly either.

At a minimum, we should at least return `numberFormat( number )` (with no options) to apply basic formatting to the number which shouldn't fail.

Also in this code block:
* `( e )` can be removed since it isn't used
* If `options` is falsy or not a string then the function would return `undefined`

assets/js/util/index.js
```diff
@@ -315,16 +182,29 @@ export const getDaysBetweenDates = ( dateStart, dateEnd ) => {
...
+
+/**
+ * Calculates the change between two values.
+ *
+ * @since 1.0.0
```
aaemnnosttv, line 191:
```suggestion
 * @since n.e.x.t
```

assets/js/util/test/numFmt.js
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * Sets the Site Kit locale to a (legacy) global variable.
+ *
+ * @since 1.7.0
+ * @private
+ *
+ * @param {string} langCode The locale to set Site Kit to use. E.g. `en-US` or `de-DE`.
+ * @return {Object} Site Kit configuration object.
+ */
```
aaemnnosttv, line 32: I don't think we need to document internal test-only functions which aren't exported. Also the `@since` is inaccurate as it is 😄

assets/js/util/test/numFmt.js
```diff
@@ -0,0 +1,122 @@
...
+const setupGoogleSiteKit = ( langCode ) => {
+	return global._googlesitekitLegacyData = {
+		locale: langCode,
+	};
```
aaemnnosttv, line 36: This would be better to use `lodash.set` to only change the `locale` rather than overwriting the entire object.

assets/js/util/test/numFmt.js
```diff
@@ -0,0 +1,122 @@
...
+};
+
+// Unlike normal space. ASCII 32
+// This `NO-BREAK SPACE` or HTML's `&nbsp;` is used by Intl.NumberFormat between unit and number. ASCII `194 160`
+const NO_BREAK_SPACE = ' ';
+
+describe( 'numFmt', () => {
+	afterEach( () => {
+		global._googlesitekitLegacyData = null;
```
aaemnnosttv, line 45: This would then only need to reset the `locale` rather than overwrite the whole object.

aaemnnosttv (dismissed):
Thanks @kostyalmm – leaving my comments below per @felixarntz 's request; overall looking like a great improvement though!

assets/js/modules/adsense/components/dashboard/DashboardSummaryWidget.js
```diff
@@ -132,7 +133,8 @@ function DashboardSummaryWidget() {
 					<DataBlock
 						className="overview-adsense-earnings"
 						title={ __( 'Total Earnings', 'google-site-kit' ) }
-						datapoint={ readableLargeNumber( period.totals[ 0 ], currencyCode ) }
+						datapoint={ period.totals[ 0 ] }
+						datapointUnit={ currencyCode }
```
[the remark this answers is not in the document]
felixarntz: Yeah I think that's out of scope here. Probably at some point we can refactor this as a better version, but let's keep it as is for now.

assets/js/modules/analytics/components/dashboard/DashboardPopularPagesWidget.js
```diff
@@ -106,7 +106,7 @@ function DashboardPopularPagesWidget() {
 
 		return [
 			title,
-			numberFormat( row.metrics[ 0 ].values[ 0 ] ),
+			numFmt( row.metrics[ 0 ].values[ 0 ], { style: 'decimal' } ),
```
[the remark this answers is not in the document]
felixarntz: I think using `numFmt` everywhere makes sense as that's our preferred function. Since it supports everything that `numberFormat` does (and more), I think we should use it.

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.js
```diff
@@ -132,15 +132,15 @@ function DashboardUniqueVisitorsWidget() {
 	}
 
 	const { totals } = visitorsData[ 0 ].data;
-	const totalUsers = totals[ 0 ].values;
-	const previousTotalUsers = totals[ 1 ].values;
-	const totalUsersChange = changeToPercent( previousTotalUsers, totalUsers );
+	const totalUsers = totals[ 0 ].values[ 0 ];
+	const previousTotalUsers = totals[ 1 ].values[ 0 ];
```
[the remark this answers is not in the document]
felixarntz: Maybe something else changed? That _does_ look odd to me, but it appeared to work correctly in my testing. @kostyalmm can you check why you made this change and what it relates to?

assets/js/modules/analytics/components/dashboard/LegacyAnalyticsDashboardWidgetTopLevel.js
```diff
@@ -149,7 +148,7 @@ function LegacyAnalyticsDashboardWidgetTopLevel( { data, requestDataToState } )
 				<DataBlock
 					className="overview-total-users"
 					title={ __( 'Unique Visitors from Search', 'google-site-kit' ) }
-					datapoint={ readableLargeNumber( totalUsers ) }
+					datapoint={ !! totalUsers ? totalUsers : 0 }
```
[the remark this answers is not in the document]
felixarntz: The value for the first split-second is `false` which is not valid against the new `PropTypes` definition for `DataBlock`. Since this is a legacy component, let's not bother improving this further, just disallowing `false` will fix that error.

assets/js/util/i18n.js
felixarntz: Maybe worth a follow-up issue?

assets/js/util/i18n.js
felixarntz: Using `duration` for the `style` sounds good to me.

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.js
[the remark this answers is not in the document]
kostyalmm: Prior to these changes, the following was happening
```
{ `${ datapoint }${ datapointUnit }` } ==> { `${ [ 10 ] }${ '%' }` } ==> 10%
``` 
and 
```
numberFormat( Math.abs( change ), {} ) ==> numberFormat( Math.abs( [ -1 ] ), {} ) ==> numberFormat( 1, {} )
```

If we pass the array after these changes, passing array will probably work fine but there will be violation of propTypes. That is why I sent the first element instead of the array.

assets/js/modules/analytics/components/dashboard/LegacyAnalyticsDashboardWidgetPopularPagesTable.js
```diff
@@ -91,7 +91,7 @@ function LegacyAnalyticsDashboardWidgetPopularPagesTable( { data } ) {
 
 		return [
 			title,
-			numberFormat( row.metrics[ 0 ].values[ 0 ] ),
+			numFmt( row.metrics[ 0 ].values[ 0 ], { style: 'decimal' } ),
```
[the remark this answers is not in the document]
kostyalmm: If we don't send the `decimal` style, the number will be displayed in the `metric` style(default style).
For example, `10000` will be displayed as `10K` but in tables we want it to show as `10,000` 

p.s. https://github.com/google/site-kit-wp/pull/2451#pullrequestreview-554014113

assets/js/modules/search-console/util/index.js
kostyalmm: @aaemnnosttv, I've kept this as this logic demands only one fraction digit. i.e 1.99 should be displayed as 1.9% and not 1.99%.
In the new `numFmt` implementation the default number of fraction digits is 2 therefore we need this logic. If you can approve that having 2 fraction digits is fine in terms of the UI then I can go ahead and remove this logic.

assets/js/util/i18n.js
kostyalmm: sounds good. About "unit" being ambiguous, shall we change `datapointUnit` and `changeUnit` to `datapointStyle` and `changeStyle`? Style can either be a custom definition or a unit 🤔

assets/js/components/MiniChart.js
```diff
@@ -57,9 +62,10 @@ function MiniChart( { index, percent } ) {
 		width: 28,
 	};
 
+	const percent = +( change * 100 ).toFixed( 1 );
```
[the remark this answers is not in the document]
kostyalmm: This statement is used just to get a percentage number with 1 fraction digit. This will be consumed by google chats library. I don't think we need to produce a translated string here.

assets/js/components/MiniChart.js
```diff
@@ -57,9 +62,10 @@ function MiniChart( { index, percent } ) {
...
 	const data = [
 		[ 'source', 'percent' ],
-		[ '', +percent ],
+		[ '', percent ],
 		[ '', ( 100 - percent ) ],
```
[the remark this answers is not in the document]
kostyalmm: this is used by `global.google.visualization.arrayToDataTable`, I don't think we can do anything about this.

assets/js/components/data-block.js
```diff
@@ -100,6 +100,15 @@ class DataBlock extends Component {
 			}
 		}
 
+		let datapointFormatted = datapoint;
+		if ( datapoint ) {
+			if ( datapointUnit ) {
+				datapointFormatted = numFmt( datapoint, datapointUnit );
+			} else {
+				datapointFormatted = numFmt( datapoint );
+			}
```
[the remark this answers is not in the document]
kostyalmm: That's true actually.

assets/js/modules/analytics/components/dashboard/LegacyAnalyticsDashboardWidgetPopularPagesTable.js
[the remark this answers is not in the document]
aaemnnosttv: Yeah, I meant that `decimal` is already the default style which would be used if this line was unchanged and continued to use `numberFormat`. Felix would prefer to use the one function though so this is fine. See https://github.com/google/site-kit-wp/pull/2451#discussion_r547820756

assets/js/modules/search-console/util/index.js
aaemnnosttv: I would think that the default number of fractional digits would be okay here but I think that would be a change @felixarntz would have to approve.

My point still stands however that we should not be manipulating the numbers here to achieve a specific formatted result. If this number should be formatted in a specific way which is different than the default, then it should be possible to pass this via props to pass different `options` to `numFmt`.

felixarntz: I think we should update the `metric` format to allow for a maximum of 1 fraction digit. This maintains the formatting as present, and it also matches the rest of the behavior of that function (where there's never more than 1 digit).

assets/js/util/i18n.js
felixarntz: @kostyalmm You updated the shorthand here to `duration`, but what we meant was to update the `style` to `duration`. The shorthand should be kept as `s` - which is more "unit"-like.

assets/js/modules/search-console/util/index.js
kostyalmm: @felixarntz the style in question here is `percent` and not `metric`. There probably isn't any issue with number of fraction digits for `metric` or `decimal` style. 
The problem here is related to the UI issues we might face in the concerned components if we use default behavior of `DataBlock`'s `datapoint` formatting logic i.e. percentage can have maximum of 2 fraction digits.
We have these options here:
* We can either send a number with only one fraction digit here 
* allow overloading `datapointUnit`/`changeUnit` props to accept `object` which will be passed to `numFmt`.
* be fine with 2 fraction digits.

assets/js/util/i18n.js
felixarntz: I don't think we need this. I believe the try catch was added here because in the original logic where this kinda came from, we were trying to format with `currency` style and if that failed using the fallback of concatenating. But now that we catch all other special shorthands before (`%` and `s`), we should be good without this. It will also show potential errors on our end explicitly.

assets/js/modules/search-console/util/index.js
felixarntz: @kostyalmm The value in question here is the `averagePosition` (not the `averageCTR` which is also shown in the diff), and that one doesn't pass `datapointUnit`, i.e. uses `metric`. So I think that is correct. The `averageCTR` uses a regular percentage formatting and doesn't have the `toFixed` part (as expected).

felixarntz (approved):
Awesome work! I think this is good to be merged. Let's keep an eye out for any formatting oddities that may come up from this. But with almost 2 weeks until the release, we should have enough time to spot any.

Issue #2299, opened 2020-10-30
Introduce `ActivateModuleCTA` and `CompleteModuleActivationCTA` components


felixarntz:
The `AnalyticsInactiveCTA`, `DashboardPagespeedCTA`, as well as the legacy `getSetupIncompleteComponent` components are fairly similar in behavior. Let's introduce generic components:
* `ActivateModuleCTA` which suggests to activate a specific module and has a button to activate it
* `CompleteModuleActivationCTA` which prompts to complete activation of a specific module and has a button to continue the module setup

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new `ActivateModuleCTA` component should be introduced:
    * It should receive the module slug as required `slug` prop.
    * It should also support optional string props for `title` and `description`. The defaults should be the following:
        * Title: _Activate %s_ (where `%s` is the module name)
        * Description: value of the module's `description` field if not empty, or alternatively _%s module needs to be configured_ (where `%s` is the module name)
    * A CTA label _Set up %s_ (where `%s` is the module name) should be passed to the inner `Link` component.
    * The functionality and UI should be similar to the existing `AnalyticsInactiveCTA` component.
* A new `CompleteModuleActivationCTA` component should be introduced:
    * It should receive the module slug as required `slug` prop.
    * It should also support optional string props for `title` and `description`. The defaults should be the following:
        * Title: _Complete %s activation_ (where `%s` is the module name)
        * Description: _%s module setup needs to be completed_ (where `%s` is the module name)
    * A CTA label _Complete setup_ should be passed to the inner `Link` component.
    * A CTA aria-label _Complete %s setup_ (where `%s` is the module name) should be passed to the inner `Link` component.
    * The functionality and UI should be similar to the existing `AnalyticsInactiveCTA` component, except that the CTA button should not activate the module, but link to the module's "reauth URL" (to complete the setup flow).
* The `AnalyticsInactiveCTA` component should be simplified to simply rely on the new `ActivateModuleCTA` component with custom props where needed.
* The `DashboardPageSpeedCTA` component should be simplified to simply rely on either the `ActivateModuleCTA` or the `CompleteModuleActivationCTA` component based on whether the module is not connected or not active, only passing the `slug` prop to them.
* The `getSetupIncompleteComponent` function should be modified to use the `CompleteModuleActivationCTA` component, passing `module` as `slug` prop to it.

## Implementation Brief
- Using `assets/js/components`
[...]

Issue #2271, opened 2020-10-26
Allow registering module store name


felixarntz:
Right now there are a couple of places in the codebase where a module's store name is assumed to be `modules/{slug}`. While this is the default value and we exclusively rely on it and don't have modules with custom store names, it _is_ (and should be) allowed to be different. Therefore we should allow the store names to be registered.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `registerModule` action of the `core/modules` store should support an optional `storeName` argument to provide the module store's name.
    * The value should not have a default.
    * If no value is specified, it should be assumed that the module does not have its own datastore.
* Anywhere in the codebase where currently `modules/{slug}` is assumed to be the store name (except of course where this value is used as the default when creating a module store), it should instead be relied upon the `storeName` module registration value. If there is none, the functionality should be assumed to not exist and, where applicable and possible, use sane fallback values.

## Implementation Brief

* In `assets/js/googlesitekit/modules/datastore/modules.js`
	* Add `storeName: null` to [`moduleDefaults`](https://github.com/google/site-kit-wp/blob/fedee50421b4b8bcdefc4ed47bb49bdda3b39794/assets/js/googlesitekit/modules/datastore/modules.js#L49-L64)
	* add `storeName` to the objects passed as the second parameter and to the `settings` object return in the action payload.
	* Add `selectors.getModuleStoreName` as a registry selector that accepts a `slug`
		- selects `module = getModule( slug )`
			* returns `undefined` if the `module` is `undefined` (not resolved yet)
			* returns `module?.storeName || null`
* Update references to `modules/${ slug }` – for tests just update to something like `tests/` instead as it isn't particularly significant, but to ensure all previous uses are removed
	- `SettingsRenderer` + test
		- use `core/modules` for the store name to select `isDoingSubmitChanges` (passing through the `slug`) instead of selecting from the module store now that core selectors are in place
		- update `haveSettingsChanged` with a fallback value of `false` in case the module does not have a settings store
			- A module without settings cannot have changed settings so this works without changing any dependent logic
	- Modules `createInfoStore` in `assets/js/googlesitekit/modules/create-info-store.js` + test
		- Make `storeName` a required argument, require with `invariant`
		- Rename `options` to `args`, mark others as optional args in doc block
	- `createModuleStore` + test
		- Make `storeName` a required argument, require with `invariant`
		- Rename `options` to `args`, mark others as optional args in doc block
[...]

PR #2460, opened 2020-12-01
Introduce ActivateModuleCTA and CompleteModuleActivationCTA components.

asvinb:
## Summary

Addresses issue #2299 

## Checklist


aaemnnosttv:
@googlebot I consent.

assets/js/components/ActivateModuleCTA.js
```diff
@@ -0,0 +1,100 @@
...
+			title={ moduleTitle }
+			description={ moduleDescription }
+			onClick={ onCTAClick }
+			ctaLabel={ moduleCTALabel }
+		/>
+	);
+};
+
+ActivateModuleCTA.propTypes = {
+	slug: PropTypes.string.isRequired,
```
felixarntz, line 97: Can you add `title` and `description` here as well?

assets/js/components/AnalyticsInactiveCTA.js
```diff
@@ -24,58 +24,24 @@ import PropTypes from 'prop-types';
...
-import Data from 'googlesitekit-data';
-import CTA from './legacy-notifications/cta';
-import GenericError from './legacy-notifications/generic-error';
-import { STORE_NAME as CORE_USER, PERMISSION_MANAGE_OPTIONS } from '../googlesitekit/datastore/user/constants';
-import { STORE_NAME as CORE_MODULES } from '../googlesitekit/modules/datastore/constants';
-
-const { useSelect, useDispatch } = Data;
-
+import ActivateModuleCTA from './ActivateModuleCTA';
 function AnalyticsInactiveCTA( props ) {
```
felixarntz, line 33: Nit-pick, but let's keep an empty line between imports and the main code (like the function here).

assets/js/components/AnalyticsInactiveCTA.js
```diff
@@ -24,58 +24,24 @@ import PropTypes from 'prop-types';
...
-	}
 
 	return (
-		<CTA
+		<ActivateModuleCTA
+			slug="analytics"
 			title={ title }
 			description={ description }
-			onClick={ onSetupAnalytics }
 			ctaLabel={ ctaLabel }
```
felixarntz, line 45: This is not a valid prop of `ActivateModuleCTA` and therefore should be removed.

assets/js/components/AnalyticsInactiveCTA.js
```diff
@@ -24,58 +24,24 @@ import PropTypes from 'prop-types';
...
-import { STORE_NAME as CORE_MODULES } from '../googlesitekit/modules/datastore/constants';
-
-const { useSelect, useDispatch } = Data;
-
+import ActivateModuleCTA from './ActivateModuleCTA';
 function AnalyticsInactiveCTA( props ) {
 	const {
 		title = __( 'Learn more about what visitors do on your site.', 'google-site-kit' ),
 		description = __( 'Connect with Google Analytics to see unique visitors, goal completions, top pages and more.', 'google-site-kit' ),
 		ctaLabel = __( 'Set up Analytics', 'google-site-kit' ),
```
felixarntz, line 37: This should be removed as well, since it's not actually relevant. Please also check any usages of `AnalyticsInactiveCTA` to remove the prop if passed.

assets/js/components/CompleteModuleActivationCTA.js
```diff
@@ -0,0 +1,88 @@
...
+			description={ moduleDescription }
+			onClick={ onCTAClick }
+			ctaLabel={ moduleCTA }
+			aria-label={ moduleAriaLabel }
+		/>
+	);
+};
+
+CompleteModuleActivationCTA.propTypes = {
+	slug: PropTypes.string.isRequired,
```
felixarntz, line 85: See above, let's add `title` and `description` here too.

assets/js/components/legacy-notifications/cta.js
```diff
@@ -22,66 +22,58 @@
...
-					{ 'googlesitekit-cta--error': error }
-				) }
+const CTA = ( {
+	title,
+	description,
+	ctaLink,
+	ctaLabel,
+	error,
+	onClick,
+	'aria-label': ariaLabel,
```
felixarntz, line 37: Maybe better to use `ariaLabel` for the prop name too here? I see that is already the case for `Link`, but it seems a bit unnecessary to me. What do you think @aaemnnosttv?

assets/js/components/CompleteModuleActivationCTA.js
```diff
@@ -0,0 +1,88 @@
...
+ */
+import CTA from './legacy-notifications/cta';
+import Data from 'googlesitekit-data';
+import { STORE_NAME as CORE_USER, PERMISSION_MANAGE_OPTIONS } from '../googlesitekit/datastore/user/constants';
+import { STORE_NAME as MODULES_STORE } from '../googlesitekit/modules/datastore/constants';
+const { useSelect } = Data;
+
+const CompleteModuleActivationCTA = ( { slug, title, description } ) => {
+	const module = useSelect( ( select ) => select( MODULES_STORE ).getModule( slug ) );
+	const adminReauthURL = useSelect( ( select ) => select( `modules/tagmanager` ).getAdminReauthURL() );
```
felixarntz, line 41: This is incorrect, it needs to use `modules/${ slug }` for the store name (pending #2271).

assets/js/components/higherorder/withdata.test.js
```diff
@@ -131,13 +139,13 @@ describe( 'withData', () => {
...
 
-		const { container, queryByTestID } = render( <WrappedComponent /> );
+		const { container, queryByTestID } = render( <WrappedComponent />, { setupRegistry } );
 
 		collectModuleData( context );
 
 		expect( queryByTestID( 'test-component' ) ).not.toBeInTheDocument();
 		expect( container.querySelector( '.googlesitekit-cta__title' ) ).toHaveTextContent( 'Test Module activation' );
-		expect( container.querySelector( '.googlesitekit-cta__description' ) ).toHaveTextContent( 'Test Module module needs to be configured' );
+		expect( container.querySelector( '.googlesitekit-cta__description' ) ).toHaveTextContent( 'Test Module module setup needs to be' );
```
felixarntz, line 148: Let's use the full description here (add "completed"), this is a bit odd otherwise.

assets/js/components/higherorder/withdata.test.js
```diff
@@ -167,7 +175,7 @@ describe( 'withData', () => {
...
```
felixarntz, line 178: See above.

assets/js/components/legacy-notifications/cta.js
```diff
@@ -22,66 +22,58 @@
...
+		}
+	</div>
+);
 
 CTA.propTypes = {
 	title: PropTypes.string.isRequired,
 	description: PropTypes.oneOfType( [ PropTypes.string, PropTypes.node ] ),
 	ctaLink: PropTypes.string,
 	ctaLabel: PropTypes.string,
+	ariaLabel: PropTypes.string,
```
felixarntz, line 76: I think this is not in line with how you named the prop above, where you called it `aria-label`. Depends on the other comment what we should go for, but right now it's out of sync either way.

assets/js/modules/adsense/index.js
```diff
@@ -73,6 +73,7 @@ domReady( () => {
 	Modules.registerModule(
 		'adsense',
 		{
+			name: 'AdSense',
```
felixarntz, line 76: Why add the module `name` here? This is part of the server-side module information and shouldn't be registered in JS. If you added this due to tests failing, let's find another way to make these work. The `provideModules` utility function (via the `receiveGetModules` datastore action) should set this information up correctly.

assets/js/modules/analytics/index.js
```diff
@@ -46,6 +46,7 @@ domReady( () => {
 	Modules.registerModule(
 		'analytics',
 		{
+			name: 'Analytics',
```
felixarntz, line 49: See above, this shouldn't be added.

assets/js/modules/optimize/index.js
```diff
@@ -33,6 +33,7 @@ domReady( () => {
 	Modules.registerModule(
 		'optimize',
 		{
+			name: 'Optimize',
```
felixarntz, line 36: See above.

assets/js/modules/pagespeed-insights/components/dashboard/DashboardPageSpeedCTA.js
```diff
@@ -51,39 +45,32 @@ function DashboardPageSpeedCTA() {
...
-				onClick={ handleSetUpClick }
-			/>
+			{
+				( ! connected && active )
+					? (
+						<CompleteModuleActivationCTA
+							title={ __( 'Complete PageSpeed Insights activation.', 'google-site-kit' ) }
+							description={ description }
+							slug="pagespeed-insights"
+							ctaLabel={ __( 'Complete activation', 'google-site-kit' ) }
```
felixarntz, line 62: This is not a valid prop of `CompleteModuleActivationCTA` and needs to be removed.

assets/js/modules/pagespeed-insights/components/dashboard/DashboardPageSpeedCTA.js
```diff
@@ -51,39 +45,32 @@ function DashboardPageSpeedCTA() {
...
+						/>
+					)
+					: (
+						<ActivateModuleCTA
+							title={ __( 'Activate PageSpeed Insights.', 'google-site-kit' ) }
+							description={ description }
+							slug="pagespeed-insights"
+							ctaLabel={ __( 'Activate PageSpeed Insights', 'google-site-kit' ) }
```
felixarntz, line 70: See above, not a valid prop of `ActivateModuleCTA`.

assets/js/modules/pagespeed-insights/components/dashboard/DashboardPageSpeedCTA.js
```diff
@@ -51,39 +45,32 @@ function DashboardPageSpeedCTA() {
...
-					? __( 'Complete activation', 'google-site-kit' )
-					: __( 'Activate PageSpeed Insights', 'google-site-kit' )
-				}
-				onClick={ handleSetUpClick }
-			/>
+			{
+				( ! connected && active )
+					? (
+						<CompleteModuleActivationCTA
+							title={ __( 'Complete PageSpeed Insights activation.', 'google-site-kit' ) }
```
felixarntz, line 59: Let's remove this prop from here too since we can rely on the `CompleteModuleActivationCTA` default to get a similar result.

assets/js/modules/pagespeed-insights/components/dashboard/DashboardPageSpeedCTA.js
```diff
@@ -51,39 +45,32 @@ function DashboardPageSpeedCTA() {
...
+							description={ description }
+							slug="pagespeed-insights"
+							ctaLabel={ __( 'Complete activation', 'google-site-kit' ) }
+						/>
+					)
+					: (
+						<ActivateModuleCTA
+							title={ __( 'Activate PageSpeed Insights.', 'google-site-kit' ) }
```
felixarntz, line 67: See above, let's remove this prop since the `ActivateModuleCTA` default produces a similar result. No need for duplicate code/strings here.

assets/js/modules/pagespeed-insights/index.js
```diff
@@ -71,6 +71,7 @@ domReady( () => {
 	Modules.registerModule(
 		'pagespeed-insights',
 		{
+			name: 'PageSpeed Insights',
```
felixarntz, line 74: See above.

assets/js/modules/search-console/index.js
```diff
@@ -42,6 +42,7 @@ domReady( () => {
 	Modules.registerModule(
 		'search-console',
 		{
+			name: 'Search Console',
```
felixarntz, line 45: See above.

assets/js/modules/tagmanager/index.js
```diff
@@ -33,6 +33,7 @@ domReady( () => {
 	Modules.registerModule(
 		'tagmanager',
 		{
+			name: 'Tag Manager',
```
felixarntz, line 36: See above.

assets/js/modules/pagespeed-insights/components/dashboard/DashboardPageSpeedCTA.js
```diff
@@ -51,39 +45,32 @@ function DashboardPageSpeedCTA() {
...
-				description={ __( 'Google PageSpeed Insights gives you metrics about performance, accessibility, SEO and PWA.', 'google-site-kit' ) }
-				ctaLink={ '#' }
-				ctaLabel={ ! connected && active
-					? __( 'Complete activation', 'google-site-kit' )
-					: __( 'Activate PageSpeed Insights', 'google-site-kit' )
-				}
-				onClick={ handleSetUpClick }
-			/>
+			{
+				( ! connected && active )
```
felixarntz, line 56: Let's simplify this code a bit, avoiding the `if else` in JSX: We can have a single if clause `! active` that returns the `ActivateModuleCTA` with an early return. And then afterwards (due to the logic before this will mean the module is active but not connected) return the `CompleteModuleActivationCTA`.

felixarntz (changes requested):
The new components overall look good, but there are a few problems that need to be addressed here, plus some minor feedback.

assets/js/components/ActivateModuleCTA.js
asvinb: Sure, added @felixarntz

assets/js/components/AnalyticsInactiveCTA.js
asvinb: Updated @felixarntz

assets/js/components/AnalyticsInactiveCTA.js
asvinb: 👍  Sorry it remained here during my first iterations.

assets/js/components/AnalyticsInactiveCTA.js
asvinb: Updated @felixarntz

assets/js/components/CompleteModuleActivationCTA.js
asvinb: Updated @felixarntz

assets/js/components/legacy-notifications/cta.js
asvinb: @felixarntz I tried to keep it consistent with what we have in `Link` but happy to change. Let me know.

assets/js/components/CompleteModuleActivationCTA.js
asvinb: My bad, some leftover when testing/debugging e2e tests.

assets/js/components/higherorder/withdata.test.js
asvinb: Updated @felixarntz

assets/js/components/higherorder/withdata.test.js
asvinb: Updated @felixarntz

assets/js/components/legacy-notifications/cta.js
asvinb: Updated @felixarntz

assets/js/modules/adsense/index.js
asvinb: For whatever reason when I was working on that, the name had the same value as the slug. I removed the name from here and looks to be fine now. Thanks for letting me know.

assets/js/modules/analytics/index.js
asvinb: Updated @felixarntz

assets/js/modules/optimize/index.js
asvinb: Updated @felixarntz

assets/js/modules/pagespeed-insights/components/dashboard/DashboardPageSpeedCTA.js
asvinb: Updated @felixarntz

assets/js/modules/pagespeed-insights/components/dashboard/DashboardPageSpeedCTA.js
asvinb: Updated @felixarntz

assets/js/modules/pagespeed-insights/components/dashboard/DashboardPageSpeedCTA.js
asvinb: Updated @felixarntz

assets/js/modules/pagespeed-insights/components/dashboard/DashboardPageSpeedCTA.js
asvinb: Updated @felixarntz

assets/js/modules/pagespeed-insights/index.js
asvinb: Updated @felixarntz

assets/js/modules/search-console/index.js
asvinb: Updated @felixarntz

assets/js/modules/tagmanager/index.js
asvinb: Updated @felixarntz

assets/js/modules/pagespeed-insights/components/dashboard/DashboardPageSpeedCTA.js
asvinb: @felixarntz Not sure if we can do an early return here since the components are within 
```
```
Let me know if you want the logic to be extracted to a function where we can do an early return instead?

assets/js/components/legacy-notifications/cta.js
felixarntz: @aaemnnosttv @tofumatt What do you think? Is it a good practice to use `aria-label` as actual prop for the `aria-label` attribute? Or should we use `ariaLabel` as prop name?

Depending on the decision, we might need to update this PR and also adjust this in the `Link` component.

felixarntz (commented):
Great! Mostly LGTM, just two follow-up comments.

asvinb:
> Great! Mostly LGTM, just two follow-up comments.

@felixarntz I have simplified the if statement, checking only for `! active`. Earlier in the code, we have another check for `active  && connected` which returns null. Which means that the else condition to render the `CompleteModuleActivationCTA` component will only be met when active && ! connected.

assets/js/components/legacy-notifications/cta.js
aaemnnosttv: We should use `aria-label` IMO as [all `aria-*` attributes are supported by React](https://reactjs.org/docs/accessibility.html#wai-aria). That way we don't have to worry about which form to use, or whether the component passes extra props through as React will warn if you try to set `ariaLabel` on an HTML element in JSX:
image

Within the component, I think it's fine to alias it to camel case like we do in `Link`, but between components I think we should always use `aria-label` for consistency.

tofumatt: Agreed, we should use `aria-label` as the actual prop name everywhere for consistency. It's what I would expect as a React developer—and it's good to be consistent. I wouldn't expect to use `ariaLabel` as it's an HTML element and I know React has explicit support for `aria-*` props.

felixarntz (approved):
Awesome!

Issue #2459, opened 2020-12-01
Fix JS date range parsing logic for AdSense widgets


felixarntz:
#2235 moved responsibility for parsing AdSense date ranges from the server to the client. However, one thing was missed during definition of the issue, which was considering the day offset for AdSense API requests. On the server-side it is `1`, but now on the client-side it is `0`. The original behavior should be maintained so this will need to be updated.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The refactored AdSense widgets should use a day offset of `1` when calling the `getDateRangeDates` selector. The value should not be hard-coded in the widgets, but it should come from an API-specific constant.
* Since the offsets to use are API-specific and one of the widgets uses the AdSense API while the other uses the Analytics API, the two offsets should come from different sources.

## Implementation Brief

* Introduce a `DATE_RANGE_OFFSET` constant in `assets/js/modules/adsense/datastore/constants.js` with value 1.
* Introduce a `DATE_RANGE_OFFSET` constant in `assets/js/modules/analytics/datastore/constants.js` with value 1.
* In the AdSense `DashboardSummaryWidget`, update the `getDateRangeDates` call to pass the argument `offsetDays: DATE_RANGE_OFFSET` with the AdSense-specific constant.
* In the AdSense `DashboardTopEarningPagesWidget`, update the `getDateRangeDates` call to pass the argument `offsetDays: DATE_RANGE_OFFSET` with the Analytics-specific constant.

### Test Coverage

* N/A

### Visual Regression Changes

* The Storybook stories for `AdSense Module/Components/Dashboard/Summary Widget` and `AdSense Module/Components/Dashboard/Top Earning Pages Widget` are currently broken (likely since #2235). In order to fix this, we'll need to set up the mock data in Storybook correctly so that report data is provided for the actual dates based on the current date range.
* Since we don't want to deal with the actual date in Storybook, we need to have a way to centrally provide an alternative reference date:
    * Add internal action `setReferenceDate` to `core/user` (in `assets/js/googlesitekit/datastore/user/date-range.js`) and store it in a `referenceDate` state property. Its initial value should be the `YYYY-MM-DD` string of today. This action should only be used for testing to replace the real "today" with a fixed reference date.
    * Add selector `getReferenceDate` to `core/user`.
    * Update `getDateRangeDates` to use the store's referenceDate, unless the `referenceDate` argument is provided.
    * Add related test coverage.
[...]

PR #2465, opened 2020-12-01
Fix AdSense widget JS date parsing and Storybook

felixarntz:
## Summary

Addresses issue #2459

## Checklist


assets/js/googlesitekit/datastore/user/date-range.js
```diff
@@ -112,16 +146,20 @@ export const selectors = {
...
 		compare = false,
 		offsetDays = 0,
-		referenceDate = getDateString( new Date() ),
+		referenceDate,
 		weekDayAlign = false,
 	} = {} ) {
+		if ( ! referenceDate ) {
+			referenceDate = state.referenceDate;
+		}
+
```
aaemnnosttv, line 162: I'm fairly certain this can be simplified like so:
```suggestion
		referenceDate = state.referenceDate,
		weekDayAlign = false,
	} = {} ) {
```
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#Earlier_parameters_are_available_to_later_default_parameters

assets/js/googlesitekit/datastore/user/date-range.js
```diff
@@ -140,6 +178,19 @@ export const selectors = {
...
+	 * Returns the current reference date, typically today.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state The current data store's state.
+	 * @return {string} The current reference date as YYYY-MM-DD.
+	 */
+	getReferenceDate( state ) {
+		const { referenceDate } = state;
+		return referenceDate;
```
aaemnnosttv, line 192: Not a big deal but the destructuring here is unnecessary.
```suggestion
		return state.referenceDate;
```

assets/js/googlesitekit/datastore/user/date-range.test.js
```diff
@@ -44,6 +45,21 @@ describe( 'core/user date-range', () => {
...
+				expect( () => {
+					registry.dispatch( STORE_NAME ).setReferenceDate();
+				} ).toThrow( 'Date string is required.' );
+			} );
+
+			it( 'should set the reference date', () => {
+				const someReferenceDate = '2020-09-12';
+
+				registry.dispatch( STORE_NAME ).setReferenceDate( someReferenceDate );
+				expect( registry.select( STORE_NAME ).getReferenceDate() ).toEqual( someReferenceDate );
```
aaemnnosttv, line 60: Generally I prefer avoiding testing the state directly in tests, but for most actions I think it's preferable to avoid mixing potential logic in the selector, which has its own tests – especially since we usually need to use the action in the selector test.

```suggestion
				expect( registry.stores[ STORE_NAME ].store.getState().referenceDate ).toEqual( someReferenceDate );
```

assets/js/googlesitekit/datastore/user/date-range.test.js
```diff
@@ -211,5 +227,22 @@ describe( 'core/user date-range', () => {
...
 			} );
 		} );
+
+		describe( 'getReferenceDate', () => {
+			it( 'should return the reference date once set', () => {
+				const someReferenceDate = '2020-08-04';
+
+				registry.dispatch( STORE_NAME ).setReferenceDate( someReferenceDate );
+				expect( registry.select( STORE_NAME )
+					.getReferenceDate() )
```
aaemnnosttv, line 237: This is a bit of an odd break here 😄 
```suggestion
				expect( registry.select( STORE_NAME ).getReferenceDate() )
```

assets/js/googlesitekit/datastore/user/date-range.test.js
```diff
@@ -211,5 +227,22 @@ describe( 'core/user date-range', () => {
...
+					.toEqual( someReferenceDate );
+			} );
+
+			it( 'should return current date when no reference date is set', () => {
+				expect( registry.select( STORE_NAME )
+					.getReferenceDate() )
```
aaemnnosttv, line 243:
```suggestion
				expect( registry.select( STORE_NAME ).getReferenceDate() )
```

stories/module-adsense-components.stories.js
```diff
@@ -24,30 +24,130 @@ import DashboardSummaryWidget from '../assets/js/modules/adsense/components/dash
...
+			startDate: '2020-09-01',
+			endDate: '2020-09-12',
+		},
 	],
 	options: [
-		dashboardSummaryWidgetTodayOptions,
-		dashboardSummaryWidgetPeriodOptions,
-		dashboardSummaryWidget28DailyOptions,
+		{
+			dateRange: 'today',
```
aaemnnosttv, line 123: This should be replaced with start and end dates, correct?

stories/module-adsense-components.stories.js
```diff
@@ -24,30 +24,130 @@ import DashboardSummaryWidget from '../assets/js/modules/adsense/components/dash
...
+			metrics: [
+				'EARNINGS',
+				'PAGE_VIEWS_RPM',
+				'IMPRESSIONS',
+			],
+		},
+		{
+			// getDateRangeDates( { offsetDays: 1 }) for 'last-28-days' and '2020-09-12'.
```
aaemnnosttv, line 131: Can we use the selector here instead of this comment now that we can provide the reference date to it?

stories/module-adsense-components.stories.js
```diff
@@ -24,30 +24,130 @@ import DashboardSummaryWidget from '../assets/js/modules/adsense/components/dash
...
+			startDate: '2020-08-15',
+			endDate: '2020-09-11',
+			metrics: [
+				'EARNINGS',
+				'PAGE_VIEWS_RPM',
+				'IMPRESSIONS',
+			],
+		},
+		{
+			dateRange: 'this-month',
```
aaemnnosttv, line 141: Same question here as above.

stories/module-adsense-components.stories.js
```diff
@@ -57,15 +157,262 @@ generateReportBasedWidgetStories( {
...
+							'15.2164',
+							'825',
+						],
+					},
+				],
+			},
+		},
+	],
+	options: {
+		// getDateRangeDates( { offsetDays: 1 }) for 'last-28-days' and '2020-09-12'.
```
aaemnnosttv, line 375: Can we use the selector here instead?

aaemnnosttv (commented):
This is looking good, just a few comments and questions. Also, would you please add a QA Brief on the issue?

stories/module-adsense-components.stories.js
felixarntz: I didn't wanna replace these just yet since they rely on special AdSense logic we don't have elsewhere. `today` would be trivial to do, but e.g. `this-month` would require another function or some extra logic in the existing function. I'd rather do that via a separate follow-up issue.

stories/module-adsense-components.stories.js
felixarntz: See above.

stories/module-adsense-components.stories.js
felixarntz: I guess technically we're still calling the selector like it is in the comment. Providing the reference date was already possible before, but we need a _centrally managed_ reference date for Storybook - that's why the new state, action and selector were added.

stories/module-adsense-components.stories.js
felixarntz: See above.

stories/module-adsense-components.stories.js
aaemnnosttv: Got it, just got a bit confused by the abstraction here 👍

aaemnnosttv (approved):
Super!

stories/module-adsense-components.stories.js
felixarntz: @aaemnnosttv I've opened #2477 for the above.

PR #2471, opened 2020-12-03
Add util function for flattening lists.

ivankruchkoff:
## Summary

Addresses issue #

## Checklist


assets/js/util/i18n.js
```diff
@@ -38,6 +39,37 @@ export const numberFormat = ( number, options = {} ) => {
...
+ * @param {Array}  list             The list to flatten.
+ * @param {Object} [options]        Formatting options.
+ * @param {string} [options.locale] Locale to use for formatting. Defaults to current locale used by Site Kit.
+ * @param {string} [options.style]  Length of the formatted message. Defaults to long.
+ * @param {string} [options.type]   Type of list. Defaults to 'conjunction' (A, B, and C).
+ *                                  Also available 'disjunction' (A, B, or C)
+ *                                  Also available 'unit' (5 pounds, 12 ounces)
+ * @return {string} The flattened list.
+ */
+export const listFlatten = ( list, options = {} ) => {
```
aaemnnosttv, line 57: I think it is a bit of a misnomer to name as `listFlatten` as this could easily be confused with functionality performed by [`_.flatten`](https://lodash.com/docs/4.17.15#flatten) which is not the case – this is more of an i18n-friendly `Array.join`.

Let's call it `listFormat` for consistency with our `numberFormat` utility which wraps `Intl.NumberFormat`.

Similarly, this utility shouldn't impose any opinionated defaults (it mostly provides the locale and passes through the rest to a new formatter and calls `format`.

I think we should have a separate `listJoin` utility if we want to have something which has more opinionated defaults.

As for the fallback, I'd rather we actually do away with that and polyfill `Intl.ListFormat` since this is likely a problem that we'll run into in other cases for other `Intl` APIs so it's worth looking into a bit more and address the root cause.

assets/js/util/i18n.js
```diff
@@ -20,6 +20,7 @@
  * External dependencies
  */
 import { get } from 'lodash';
+import { __ } from '@wordpress/i18n';
```
aaemnnosttv, line 23: This belongs under "WordPress dependencies".

aaemnnosttv (changes requested):
Not sure if this is ready for review or not but left a few comments below.

felixarntz:
@ivankruchkoff Which issues does this PR apply to? Can you please connect it as applicable?

ivankruchkoff:
> @ivankruchkoff Which issues does this PR apply to? Can you please connect it as applicable?

#2486 I've added the link.

assets/js/components/settings/settings-module.js
```diff
@@ -42,7 +42,7 @@ import {
 	activateOrDeactivateModule,
 	getReAuthURL,
 	showErrorNotification,
-	getModulesData,
+	getModulesData, listFormat,
```
felixarntz, line 45: Nitpick, but let's not mix a single `import` statement to use both one entry per line vs all in the same line. This one should be moved to a new line like the other ones in the same import group are.

felixarntz (commented):
Just one minor piece of feedback, but overall LGTM!

felixarntz (approved):
Great!

Issue #2041, opened 2020-09-16
Implement settings block for user input UI


felixarntz:
A new settings box should be added in the Site Kit > Settings > Admin Settings tab, which exposes the same UI that is part of the preview UI for the user input screen (see #2038, #2039, #2040). It should reuse components from #2039.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* In the Site Kit settings "Admin Settings" tab, another box should be added, as the very first box (above "Plugin Status").
* If the current user has not yet completed the user input flow, it should have the same content as the notification from #2043 (without the dismiss button.
* If the current user has already completed the user input flow, it should display the part of the preview UI of the user input screen that shows the questions with their answers (see screenshot below). The "Edit" links should link to the `googlesitekit-user-input` screen, with query parameters that will ensure the corresponding question is shown initially.
* Components implemented in #2039 and #2043 should be reused for the UI, splitting them out and adding props to tweak behavior as necessary (the UI can be reused, but some button/link behavior needs to be different).

### Screenshot of relevant section of User Input screen
<img width="707" alt="Screenshot 2020-10-29 at 10 08 02" src="https://user-images.githubusercontent.com/3531426/97607785-ae1c9c00-19ce-11eb-89ee-597d4d780d2f.png">

## Implementation Brief

* Add another box section above https://github.com/google/site-kit-wp/blob/3a6d747e3e27795c7de81a896bb1969ac7bdcaa4/assets/js/components/settings/settings-admin.js#L35 in `assets/js/components/settings/settings-admin.js`
  * **If the user input flow has not been completed:** use the contents rendered inside the notification in #2043, but without a "dismiss" button as it isn't a notification here. The main part of that UI should be a component that can be reused for this to work, so it should either be implemented that way or refactored in this ticket.
  * **If the user input flow has been completed:** resuse component from #2039 to show the questions and their existing answers (see the provided mockup screenshots), with edit links that link to the User Input URLs to the relevant answer so the user can edit that question. The User Input app should respond to a URL provided that points to: 1. edit mode 2. the specific question
[...]

PR #2476, opened 2020-12-03
Add User Input block to Admin Settings, use query param for state, and add redirect URL.

kostyalmm:
## Summary

Addresses issue #2041

## Relevant technical choices

* Add `UserInputSettings` notification to Admin Settings when the user input is incomplete.
* Add `UserInputPreview` to Admin Setting when the user input is completed.
* Hide `UserInputPreview` footer in Admin Settings.
* Add `useQueryString` hook to use the query param as state
* Add a `question` query param state to `UserInputQuestionnaire`
* Add a `redirect_url` query param to `UserInputQuestionnaire`
* Go to `${ userInputURL }&question=${ question }` when clicked on edit button in `AdminSettings`.
* After clicking submit, go to the Admin Settings using the `redirect_url` query param.

Reason why `pbeshai/use-query-params` wasn't used: We need to wrap the app in a Context for this to work. `react-use` doesn't have the ability to use query param as state and using hash would have resulted in problem with methods like `addQueryArgs` (it converts # to %23).

## Checklist


assets/sass/components/settings/_googlesitekit-settings-user-input.scss
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+.googlesitekit-settings-user-input {
+
+	.googlesitekit-user-input__preview {
+		padding: 25px !important;
+	}
+}
```
eugene-manuilov, line 24: Just start your rule from `.googlesitekit-plugin` and you won't need `!important`:

```suggestion
.googlesitekit-plugin {
    .googlesitekit-settings-user-input {
        .googlesitekit-user-input__preview {
            padding: 25px;
        }
    }
}
```

assets/js/hooks/useQueryString.js
```diff
@@ -0,0 +1,77 @@
...
+		},
+		[ key ]
+	);
+
+	return [ value, onSetValue ];
+}
+
+const setQueryStringWithoutPageReload = ( qsValue ) => {
+	const newURL = `${ global.location.pathname }${ qsValue }`;
+	global.history.pushState( { path: newURL }, '', newURL );
```
eugene-manuilov, line 53: @kostyalmm use `replaceState` instead of `pushState` to eliminate browser history pollution.

assets/js/components/settings/settings-admin.js
```diff
@@ -19,20 +19,68 @@
...
+import { STORE_NAME as CORE_USER } from '../../googlesitekit/datastore/user/constants';
 import Layout from '../layout/layout';
 import OptIn from '../optin';
 import VisuallyHidden from '../VisuallyHidden';
 import ResetButton from '../ResetButton';
+import UserInputPreview from '../user-input/UserInputPreview';
+import {
+	USER_INPUT_QUESTION_GOALS, USER_INPUT_QUESTION_HELP_NEEDED,
+	USER_INPUT_QUESTION_POST_FREQUENCY,
+	USER_INPUT_QUESTION_ROLE, USER_INPUT_QUESTION_SEARCH_TERMS,
```
eugene-manuilov, line 38: Please, use one import per line if you expand named imports into multiple lines.

eugene-manuilov (changes requested):
Thanks, @kostyalmm! I have added some comments that need to be addressed before I can approve this PR. Could you please look at it?

assets/js/components/user-input/UserInputQuestionnaire.js
```diff
@@ -38,17 +39,43 @@ import {
...
+				return i;
+			}
+		}
+	} );
+
+	useEffect( () => {
+		if ( activeSlugIndex > answeredUntilIndex ) {
+			setActiveSlug( steps[ answeredUntilIndex ] );
+		}
+	}, [ answeredUntilIndex ] );
```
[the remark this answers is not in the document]
kostyalmm: So if someone tries accessing `question=5` through the URL and questions 1-4 are not filled out, this logic will put them on `question=1`.

Also if someone is on question 5, and then hits reload, we want them to re fill the form from state as it is stored in the local state.

Unless and until, questions until `n` are filled, the user should not be able to see/fill `n+1`th question.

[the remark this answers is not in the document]
eugene-manuilov: Ok, thanks!

assets/js/components/settings/settings-admin.js
```diff
@@ -19,20 +19,70 @@
...
+import UserInputPreview from '../user-input/UserInputPreview';
+import {
+	USER_INPUT_QUESTION_GOALS,
+	USER_INPUT_QUESTION_HELP_NEEDED,
+	USER_INPUT_QUESTION_POST_FREQUENCY,
+	USER_INPUT_QUESTION_ROLE,
+	USER_INPUT_QUESTION_SEARCH_TERMS,
+} from '../user-input/util/constants';
+import { STORE_NAME as CORE_SITE } from '../../googlesitekit/datastore/site/constants';
+import { addQueryArgs } from '@wordpress/url';
```
eugene-manuilov, line 44: @kostyalmm, please, move it to the WordPress dependencies section.

assets/js/components/settings/settings-admin.js
```diff
@@ -19,20 +19,70 @@
...
+	USER_INPUT_QUESTION_SEARCH_TERMS,
+];
 
 const SettingsAdmin = () => {
+	const { isUserInputCompleted, userInputURL } = useSelect( ( select ) => {
+		return {
+			isUserInputCompleted: select( CORE_USER ).getUserInputState() === 'completed',
+			userInputURL: select( CORE_SITE ).getAdminURL( 'googlesitekit-user-input' ),
+		};
+	} );
```
eugene-manuilov, line 62:
```suggestion
    const { isUserInputCompleted, userInputURL } = useSelect( ( select ) => ( {
        isUserInputCompleted: select( CORE_USER ).getUserInputState() === 'completed',
        userInputURL: select( CORE_SITE ).getAdminURL( 'googlesitekit-user-input' ),
    } ) );
```

assets/js/hooks/useQueryString.js
```diff
@@ -0,0 +1,69 @@
...
+			setValue( newValue );
+			setQueryStringValue( key, newValue );
+		},
+		[ key ]
+	);
+
+	return [ value, onSetValue ];
+}
+
+const setQueryStringWithoutPageReload = ( qsValue ) => {
```
eugene-manuilov, line 46: @kostyalmm, could you please convert `setQueryStringWithoutPageReload`, `getQueryStringValue` and `setQueryStringValue` functions to be regular functions (not arrow ones) and add jsdoc to it.

eugene-manuilov (changes requested):
Awesome work, @kostyalmm! Just added a few stylistic requests and I will approve it.

assets/js/hooks/useQueryString.js
kostyalmm: I actually ended up removing those methods and putting them inside the hook with `useCallback`. Do you want me to add the jsDoc even now? or can we skip that?

orders_export.csv
```diff
@@ -0,0 +1,11 @@
+Name,Email,Financial Status,Paid at,Fulfillment Status,Fulfilled at,Accepts Marketing,Currency,Subtotal,Shipping,Taxes,Total,Discount Code,Discount Amount,Shipping Method,Created at,Lineitem quantity,Lineitem name,Lineitem price,Lineitem compare at price,Lineitem sku,Lineitem requires shipping,Lineitem taxable,Lineitem fulfillment status,Billing Name,Billing Street,Billing Address1,Billing Address2,Billing Company,Billing City,Billing Zip,Billing Province,Billing Country,Billing Phone,Shipping Name,Shipping Street,Shipping Address1,Shipping Address2,Shipping Company,Shipping City,Shipping Zip,Shipping Province,Shipping Country,Shipping Phone,Notes,Note Attributes,Cancelled at,Payment Method,Payment Reference,Refunded Amount,Vendor,Outstanding Balance,Employee,Location,Device ID,Id,Tags,Risk Level,Source,Lineitem discount,Tax 1 Name,Tax 1 Value,Tax 2 Name,Tax 2 Value,Tax 3 Name,Tax 3 Value,Tax 4 Name,Tax 4 Value,Tax 5 Name,Tax 5 Value,Phone,Receipt Number
```
eugene-manuilov, line 1: Do we need this file? Probably it has been committed by accident, hasn't it?

assets/js/hooks/useQueryString.js
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * @param {string}        key          The query param key to be used.
+ * @param {(string|null)} initialValue The initial value for the query param to be used.
+ * @return {Array} The getter and setter for the query param state.
+ */
+function useQueryString( key, initialValue = null ) {
+	const getQueryStringValue = useCallback( () => {
+		const urlParams = new URLSearchParams( global.location.search );
+		return urlParams.get( key );
+	}, [ global.location.search ] );
```
eugene-manuilov, line 37: Looks like we don't need this function and can rework `useState` initial value to be something like this:

```js
const [ value, setValue ] = useState(
    new URLSearchParams( global.location.search ).get( key ) || initialValue,
);

```

assets/js/hooks/useQueryString.js
```diff
@@ -0,0 +1,58 @@
...
+
+	const setQueryStringValue = useCallback( ( value ) => {
+		const urlParams = new URLSearchParams( global.location.search );
+		urlParams.set( key, value );
+		const newURL = `${ global.location.pathname }?${ urlParams.toString() }`;
+		global.history.replaceState( { path: newURL }, '', newURL );
+	}, [ global.location.pathname, global.location.search ] );
```
eugene-manuilov, line 44: Same here, I don't think we need this callback function. We can incorporate it into the `onSetValue`:

```js
    const onSetValue = useCallback( ( newValue ) => {
        setValue( newValue );

        const urlParams = new URLSearchParams( global.location.search );
        urlParams.set( key, newValue );

        const path = `${ global.location.pathname }?${ urlParams.toString() }`;
        global.history.replaceState( { path }, '', path );
}, [ key ] );
```

eugene-manuilov (changes requested):

assets/js/hooks/useQueryString.js
eugene-manuilov: No, we don't need jsdoc then. :+1:

assets/js/hooks/useQueryString.js
kostyalmm: Made the changes. Also I have deleted your previous comment 😅, it was a sensitive file so didn't want it to have a reference. 😅

eugene-manuilov (approved):
Awesome! Thanks!

eugene-manuilov (changes requested):
@kostyalmm! Could you please fix broken js tests?

eugene-manuilov (dismissed):
Thanks, @kostyalmm!

assets/js/hooks/useQueryString.js
```diff
@@ -0,0 +1,51 @@
...
+ * Uses a location query param as a state.
+ *
+ * @since n.e.x.t
+ *
+ * @param {string}        key          The query param key to be used.
+ * @param {(string|null)} initialValue The initial value for the query param to be used.
+ * @return {Array} The getter and setter for the query param state.
+ */
+function useQueryString( key, initialValue = null ) {
+	const [ value, setValue ] = useState( new URLSearchParams( global.location.search ).get( key ) || initialValue );
```
aaemnnosttv, line 34: Please use [`getQueryArg`](https://github.com/WordPress/gutenberg/tree/master/packages/url#getQueryArg) instead.

assets/js/hooks/useQueryString.js
```diff
@@ -0,0 +1,51 @@
...
+ */
+import { useState, useCallback } from '@wordpress/element';
+
+/**
+ * Uses a location query param as a state.
+ *
+ * @since n.e.x.t
+ *
+ * @param {string}        key          The query param key to be used.
+ * @param {(string|null)} initialValue The initial value for the query param to be used.
```
aaemnnosttv, line 30: This should be marked as an optional parameter.
```suggestion
 * @param {string} [initialValue] Optional. The initial value for the query param to be used. Default: `null`.
```

assets/js/hooks/useQueryString.js
```diff
@@ -0,0 +1,51 @@
...
+ * @return {Array} The getter and setter for the query param state.
+ */
+function useQueryString( key, initialValue = null ) {
```
aaemnnosttv, line 33: Can we name this `useQueryArg` instead for consistency with the `(add|get)QueryArg(s)` functions that we already use in many places? I think that would be a slightly more accurate name as well since we're getting the value of a single query arg rather than the entire query string.

assets/js/hooks/useQueryString.js
```diff
@@ -0,0 +1,51 @@
...
+	const [ value, setValue ] = useState( new URLSearchParams( global.location.search ).get( key ) || initialValue );
+	const onSetValue = useCallback(
+		( newValue ) => {
+			setValue( newValue );
+
+			const urlParams = new URLSearchParams( global.location.search );
+			urlParams.set( key, newValue );
+
+			const newURL = `${ global.location.pathname }?${ urlParams.toString() }`;
```
aaemnnosttv, line 42: Please use [`addQueryArgs`](https://github.com/WordPress/gutenberg/tree/master/packages/url#addQueryArgs) instead: e.g.
```js
const newURL = addQueryArgs( global.location.href, { [ key ]: newValue } )
```

assets/js/hooks/useQueryString.test.js
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * Internal dependencies
+ */
+import { renderHook, actHook } from '../../../tests/js/test-utils';
+import useQueryString from './useQueryString';
+
+describe( 'useQueryString', () => {
+	let oldLocation;
+	let oldHistory;
+	const historyPushMock = jest.fn();
```
aaemnnosttv, line 28: Let's name this to match the method more closely. Otherwise it might be confused as a mock for `history.pushState`.

```suggestion
	const historyReplaceStateMock = jest.fn();
```

assets/js/hooks/useQueryString.test.js
```diff
@@ -0,0 +1,69 @@
...
+		actHook( () => {
+			setQuery( 'prod' );
+		} );
+
+		const updatedURL = 'path?page=prod';
+		expect( historyPushMock ).toHaveBeenCalledWith( {
+			path: updatedURL,
+		}, '', updatedURL );
+	} );
+} );
```
aaemnnosttv, line 69: Let's add another test here to make sure that `setQuery` URL-encodes the value (done automatically by `addQueryArgs`).

assets/sass/components/settings/_googlesitekit-settings-user-input.scss
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+.googlesitekit-plugin {
+
+	.googlesitekit-settings-user-input {
+
+		.googlesitekit-user-input__preview {
+			padding: 25px !important;
```
aaemnnosttv, line 24: Is there not a scss variable we can use for the value instead? It looks close to `$grid-gap-desktop` – if so, it may also need to use `$grid-gap-phone` in smaller viewports.

Also, why `!important` ?

assets/js/components/settings/settings-admin.js
```diff
@@ -19,20 +19,68 @@
...
+import UserInputSettings from '../notifications/UserInputSettings';
+const { useSelect } = Data;
+
+const questions = [
+	USER_INPUT_QUESTION_ROLE,
+	USER_INPUT_QUESTION_POST_FREQUENCY,
+	USER_INPUT_QUESTION_GOALS,
+	USER_INPUT_QUESTION_HELP_NEEDED,
+	USER_INPUT_QUESTION_SEARCH_TERMS,
+];
```
aaemnnosttv, line 54: This array is duplicated with the same array in `UserInputQuestionnaire` https://github.com/google/site-kit-wp/blob/878a304b92fd13a642894c6938657e1e1932c5f9/assets/js/components/user-input/UserInputQuestionnaire.js#L47-L53

This risks these two being out of sync if the array changes in one place and not the other.

Let's extract this array into the user input constants and reference that in both places rather than duplicating it in both.

aaemnnosttv (changes requested):
Thanks @kostyalmm – this is off to a good start, just a few points to address.

assets/js/components/user-input/UserInputPreview.js
```diff
@@ -58,7 +58,12 @@ export default function UserInputPreview( { back, goTo } ) {
 		setIsNavigating( true );
 		const response = await saveUserInputSettings();
 		if ( ! response.error ) {
-			global.location.assign( addQueryArgs( dashboardURL, { notification: 'user_input_success' } ) );
+			let [ sanitizedRedirectURL, hash ] = ( redirectURL ?? dashboardURL ).split( '#' );
+			sanitizedRedirectURL = addQueryArgs( sanitizedRedirectURL, { notification: 'user_input_success' } );
+			if ( hash ) {
+				sanitizedRedirectURL += `#${ hash }`;
+			}
```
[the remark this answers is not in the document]
kostyalmm: As you can see [here](https://github.com/WordPress/gutenberg/blob/master/packages/url/src/add-query-args.js#L23), the `addQueryArgs` does not care about the `hash` at all. 
It takes the url, appends query arg to the url. 
What happens in the case of hash is this:
if 
```js
url = http://example.com/path?argument=value#hash
```
then 
```js
addQueryArgs(url, {test: "test"}) = "http://example.com/path?argument=value%23hash&test=test"
```

The `#` gets sandwiched into a query arg itself. What do you suggest I do here? Splitting seems to be the best thing as we need to get rid of the hash before adding query args to it and then add the hash later.

assets/js/components/user-input/UserInputQuestionnaire.js
```diff
@@ -38,17 +39,43 @@ import {
...
 		USER_INPUT_QUESTION_SEARCH_TERMS,
 	];
+	const steps = [ ...questions, 'preview' ];
+
+	const [ activeSlug, setActiveSlug ] = useQueryString( 'question', steps[ 0 ] );
+	const [ redirectURL ] = useQueryString( 'redirect_url' );
+
+	const activeSlugIndex = steps.indexOf( activeSlug );
+
+	if ( activeSlugIndex === -1 ) {
```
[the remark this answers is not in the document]
kostyalmm: We need the index later in the component, that is why it is done like this.

assets/sass/components/settings/_googlesitekit-settings-user-input.scss
kostyalmm: The `googlesitekit-user-input__preview` class has a hardcoded value of `padding: 50px 23px;` that is why I used `25px`. I used the variables instead, please check.

also removed the `!important`, it was not required 😅

assets/js/components/user-input/UserInputPreview.js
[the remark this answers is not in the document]
aaemnnosttv: I see what you mean @kostyalmm . It works fine when the hash is before the query string but is not handled when it is after which seems to be the correct place when both are present 🤔 

This seems to be a known bug in the package: https://github.com/WordPress/gutenberg/issues/16655

I feel like the safest way is to use `URL` rather than assume the hash is in a particular spot. This seems to be the direction the PR for that issue is taking as well.

For this specific case, I think we can avoid using `addQueryArgs` since we're simply adding a specific, known parameter which also doesn't require encoding.

```js
const url = new URL( redirectURL || dashboardURL );
// Here we don't use `addQueryArgs` due to a bug with how it handles hashes
// See https://github.com/WordPress/gutenberg/issues/16655
url.searchParams.set( 'notification', 'user_input_success' );
const finalURL = url.toString();
```

assets/js/hooks/useQueryArg.js
```diff
@@ -0,0 +1,50 @@
...
+import { useState, useCallback } from '@wordpress/element';
+import { addQueryArgs, getQueryArg } from '@wordpress/url';
+
+/**
+ * Uses a location query param as a state.
+ *
+ * @since n.e.x.t
+ *
+ * @param {string} key            The query param key to be used.
+ * @param {string} [initialValue] Optional. The initial value for the query param to be used. Default: `null`.
```
aaemnnosttv, line 31: On second thought, I think `undefined` is correct for the initial value here since a `null` value will result in the query variable being added with no value – e.g. `?foo=`, where `undefined` is more accurate to the key not existing.

```suggestion
 * @param {string} [initialValue] Optional. The initial value for the query param to be used.
```

assets/js/hooks/useQueryArg.js
```diff
@@ -0,0 +1,50 @@
...
+ */
+function useQueryArg( key, initialValue ) {
+	const [ value, setValue ] = useState( getQueryArg( global.location.href, key ) || initialValue );
+
+	const onSetValue = useCallback(
+		( newValue ) => {
+			setValue( newValue );
+
+			const newURL = addQueryArgs( global.location.href, { [ key ]: newValue } );
+			global.history.replaceState( { path: newURL }, '', newURL );
```
aaemnnosttv, line 42: I'm not sure the first argument (`stateObj`) is needed here. Using `path` for a full URL isn't really accurate either.

Let's pass this as `null` for now since there isn't anything we currently need this for.
```suggestion
			global.history.replaceState( null, '', newURL );
```

assets/sass/components/settings/_googlesitekit-settings-user-input.scss
```diff
@@ -0,0 +1,32 @@
...
+
+	.googlesitekit-settings-user-input {
+		padding-bottom: 0;
+
+		.googlesitekit-user-input__preview {
+			padding: $grid-gap-desktop;
+
+			@media (max-width: $bp-mobileOnly) {
+				padding: $grid-gap-phone;
+			}
```
aaemnnosttv, line 29: This should be the other way around – mobile styles first and then media query with `min-width` for larger viewports 🙂  Almost all of our styles are this way.

assets/js/components/user-input/util/constants.js
```diff
@@ -27,6 +27,14 @@ export const USER_INPUT_QUESTION_GOALS = 'goals';
 export const USER_INPUT_QUESTION_HELP_NEEDED = 'helpNeeded';
 export const USER_INPUT_QUESTION_SEARCH_TERMS = 'searchTerms';
 
+export const QUESTIONS = [
```
aaemnnosttv, line 30: Let's name this a little more specifically, for consistency with the others but also to make it a bit more clear what it is.

```suggestion
export const USER_INPUT_QUESTIONS_LIST = [
```

aaemnnosttv (commented):
Thanks @kostyalmm – a few additional comments to address but this is looking good!

assets/js/components/user-input/UserInputQuestionnaire.js
```diff
@@ -19,36 +19,56 @@
...
+	} );
+
+	useEffect( () => {
+		if ( answeredUntilIndex === -1 ) {
+			return;
+		}
+		if ( activeSlugIndex > answeredUntilIndex ) {
+			setActiveSlug( steps[ answeredUntilIndex ] );
+		}
+	}, [ answeredUntilIndex ] );
```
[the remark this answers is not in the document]
kostyalmm: I did not add this as I thought that the answers are set in the local state which will make the `answeredUntilIndex` a constant for the lifetime of the component. But as I saw, `getUserInputSettings` returns an object which has the questions answered in realtime therefore added `activeSlugIndex`. Thanks for the catch 👍

assets/js/hooks/useQueryArg.js
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+/**
+ * WordPress dependencies
+ */
+import { useState, useCallback } from '@wordpress/element';
+import { addQueryArgs, getQueryArg } from '@wordpress/url';
+
+/**
+ * Uses a location query param as a state.
```
tofumatt, line 26: I'm not sure what "as a state" here means. How about:

```suggestion
 * Uses a location query param as a variable in a component.
```

assets/js/hooks/useQueryArg.js
```diff
@@ -0,0 +1,50 @@
...
+ * @since n.e.x.t
+ *
+ * @param {string} key            The query param key to be used.
+ * @param {string} [initialValue] Optional. The initial value for the query param to be used.
+ * @return {Array} The getter and setter for the query param state.
+ */
+function useQueryArg( key, initialValue ) {
+	const [ value, setValue ] = useState( getQueryArg( global.location.href, key ) || initialValue );
+
+	const onSetValue = useCallback(
```
tofumatt, line 37: This shouldn't need `useCallback`, so let's remove it.

assets/js/hooks/useQueryArg.js
```diff
@@ -0,0 +1,50 @@
...
+		( newValue ) => {
+			setValue( newValue );
+
+			const newURL = addQueryArgs( global.location.href, { [ key ]: newValue } );
+			global.history.replaceState( null, '', newURL );
```
tofumatt, line 42: Nice 👍🏻

assets/js/hooks/useQueryArg.js
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Uses a location query param as a state.
+ *
+ * @since n.e.x.t
+ *
+ * @param {string} key            The query param key to be used.
+ * @param {string} [initialValue] Optional. The initial value for the query param to be used.
+ * @return {Array} The getter and setter for the query param state.
+ */
+function useQueryArg( key, initialValue ) {
```
tofumatt, line 34: I'd be inclined to add a third argument for dependency injection here, rather than the global mocking done in the test below. Something like `function useQueryArg( key, initialValue, _global = global ) { ` or something. What do you think?

assets/js/hooks/useQueryArg.test.js
```diff
@@ -0,0 +1,79 @@
...
+		oldHistory = global.history;
+		delete global.location;
+		delete global.history;
+		global.location = {
+			href: 'http://example.com/path?page=demo',
+		};
+		global.history = {
+			replaceState: historyReplaceStateMock,
+		};
+	} );
```
tofumatt, line 41: As mentioned above—it might be a bit tidier to allow for the `global` variable in the hook to be injected for testing purposes rather than doing this global mocking dance.

assets/js/hooks/useQueryArg.test.js
```diff
@@ -0,0 +1,79 @@
...
+
+	it( 'should push a new state to history when setQuery is called', () => {
+		const { result: { current: [ , setQuery ] } } = renderHook( ( ) => useQueryArg( 'page' ) );
+
+		actHook( () => {
+			setQuery( 'prod' );
+		} );
+
+		const updatedURL = 'http://example.com/path?page=prod';
+		expect( historyReplaceStateMock ).toHaveBeenCalledWith( null, '', updatedURL );
```
tofumatt, line 66: Hmm, I noticed the nested `const` object declaration in the function above and find it hard to read, but this is an even denser line. I think it's hard to follow `const { result: { current: [ , setQuery ] } }` especially, but the whole line is using so much shorthand that it reduces readability to me. I think this, and other functions in this file, can be expanded to allow for better readability.

Since these are tests we don't need to be too concise or even worry about DRY—the main aim of the tests should be the utmost readability and understanding at a glance to discern attention, even if it means they're a bit longer as to be clearer 🙂 

```suggestion
		const { result } = renderHook( ( ) => useQueryArg( 'page' ) );
		const [ , setQuery ] = result.current;

		actHook( () => {
			setQuery( 'prod' );
		} );

		const updatedURL = 'http://example.com/path?page=prod';
		expect( historyReplaceStateMock ).toHaveBeenCalledWith( null, '', updatedURL );
```

tofumatt (changes requested):
This looks really good to me. I think mostly small tweaks are needed here. Mostly around test readability and the `useCallback` hook usage.

Nice 👍🏻

assets/js/hooks/useQueryArg.js
kostyalmm: This actually is a great idea. Thanks :)

assets/js/components/settings/settings-admin.js
```diff
@@ -21,18 +21,50 @@
...
+			{ featureFlags.userInput.enabled && (
+				<Cell size={ 12 }>
+					<Layout>
+						{ isUserInputCompleted ? (
+							<div className="googlesitekit-module-page googlesitekit-settings-user-input">
+								<UserInputPreview goTo={ goTo } noFooter />
+							</div>
+						) : (
+							<UserInputSettings isDimissable={ false } />
+						) }
```
tofumatt, line 64: In the future I'd prefer we not use ternaries in JSX and use something like:

```suggestion
						{ isUserInputCompleted && (
							<div className="googlesitekit-module-page googlesitekit-settings-user-input">
								<UserInputPreview goTo={ goTo } noFooter />
							</div>
						) }
						{ ! isUserInputCompleted && <UserInputSettings isDimissable={ false } /> }
```

tofumatt (approved):
Looks great! Thanks for the updates. 🙂

aaemnnosttv (approved):
LGTM 👍

assets/js/components/user-input/UserInputPreview.js
```diff
@@ -58,7 +57,11 @@ export default function UserInputPreview( { back, goTo } ) {
 		setIsNavigating( true );
 		const response = await saveUserInputSettings();
 		if ( ! response.error ) {
-			global.location.assign( addQueryArgs( dashboardURL, { notification: 'user_input_success' } ) );
+			const url = new URL( redirectURL || dashboardURL );
+			// Here we don't use `addQueryArgs` due to a bug with how it handles hashes
+			// See https://github.com/WordPress/gutenberg/issues/16655
+			url.searchParams.set( 'notification', 'user_input_success' );
```
felixarntz, line 63: We don't need to add this query parameter for the alternative redirect, it should only apply to the default `dashboardURL`. That query parameter isn't used elsewhere.

That way we can simplify this to only `addQueryArgs` to the default URL like before (if no `redirectURL` is passed), and use `redirectURL` as is.

felixarntz (changes requested):
Awesome work here so far! Just a few pieces of feedback, one unrelated to this PR, but worth addressing while working on this.

I also think we might wanna add a heading to the new box (see https://github.com/google/site-kit-wp/issues/2041#issuecomment-744765020), but that's pending so don't worry about it for now.

benbowler:
@aaemnnosttv @tofumatt @felixarntz Are you happy for me to pick up this ticket from here to get it through CR? I'm working related work that it blocked by this.

benbowler:
Also, I got wires crossed and have re-implemented this ticket using `useHash` as part of my work on #2499 - checkout branch `enhancement/2041-settings-block-user-input-ui-incorrect-ticket` - it seems super clean to do it this way.

eugene-manuilov:
@googlebot I consent.

felixarntz (approved):
Apologies for the delay on reviewing this one.

The only thing missing here were the title and description as required via https://github.com/google/site-kit-wp/issues/2041#issuecomment-745357833. I fixed that in f64fafbccc3040b508e8b491d7b2c4621471fd8e, and also fixed the UI a little bit by using the grid instead of manually adding paddings via CSS.

Good for merge now!

Issue #2075, opened 2020-09-18
Update WordPress dashboard widget to functional component using datastore


felixarntz:
The WordPress dashboard widget for Site Kit should be refactored to become a functional component relying on the datastore. It should not rely on the Widget API at this point and not be extensible in any way. All sub-components of the dashboard should be part of "Site Kit core" (i.e. in `assets/js/components`), even those relying on certain modules (`assets/js/modules` should only contain files that are relevant for the respective module JS entrypoints, i.e. register things in Site Kit core).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `WPDashboardMain` component should be renamed to `WPDashboardApp`.
* A feature flag `widgets.wpDashboard` should be introduced. (While this won't for now use the actual Widgets API, it still underlies several similar requirements.)
* Individual components (using hooks and datastore exclusively) should be added in `assets/js/components/wp-dashboard` for all components that are added via the `googlesitekit.WPDashboardHeader` or `googlesitekit.WPDashboardModule` filter:
    * `WPDashboardUniqueVisitors` and `WPDashboardSessionDuration` (from legacy component `WPAnalyticsDashboardWidgetOverview`)
    * `WPDashboardClicks` and `WPDashboardImpressions` (from legacy component `WPSearchConsoleDashboardWidgetOverview`)
    * See the existing `Dashboard*Widget` components which will have most of the code for these already included, so a lot is available for copy-paste.
* If the above feature flag is enabled, `WPDashboardApp` should render the above new components right above the legacy `WPDashboardModules` component, in the same order in which they're displayed now.
    * Like before, the components that have their legacy versions filtered in via `googlesitekit.WPDashboardHeader` should be rendered within the `div.googlesitekit-wp-dashboard-stats` wrapper, the other ones (via `googlesitekit.WPDashboardModule`) after that container.
    * If the Analytics module is not active and connected, instead of the `WPDashboardUniqueVisitors` and `WPDashboardSessionDuration` components, the `AnalyticsInactiveCTA` should be displayed, similarly to today.
* `WPDashboardModules` should be renamed to `LegacyWPDashboardModules`.
* `WPDashboardModule` should be renamed to `LegacyWPDashboardModule`.
* `WPDashboardHeader` should be renamed to `LegacyWPDashboardHeader`.
* All files in `assets/js/components/wp-dashboard` should have names that match the component name that they export.

## Implementation Brief

[...]

PR #2482, opened 2020-12-04
Update WP dashboard widgets to functional components using datastore.

kostyalmm:
## Summary

Addresses issue #2075

## Relevant technical choices

* Rename `WPDashboardMain` to `WPDashboardApp`
* Create `WPDashboardUniqueVisitors`
* Create `WPDashboardSessionDuration`
* Create `WPDashboardClicks`
* Create `WPDashboardImpressions`

## Checklist


assets/js/components/wp-dashboard/wp-dashboard-unique-visitors.js
```diff
@@ -0,0 +1,90 @@
+/**
+ * WPDashboardUniqueVisitors component.
+ *
+ * Site Kit by Google, Copyright 2019 Google LLC
```
tofumatt, line 4: This might as well be updated to 2020, but really we should just search + replace them all to 2021 in a few weeks so 🤷🏻‍♂️

assets/js/components/wp-dashboard/wp-dashboard-session-duration.js
```diff
@@ -0,0 +1,86 @@
...
+
+	const {
+		averageSessionDuration,
+		averageSessionDurationChange,
+	} = calculateOverviewData( data );
+
+	return (
+		<DataBlock
+			className="googlesitekit-wp-dashboard-stats__data-table overview-average-session-duration"
+			title={ __( 'Avg. Time on Page', 'google-site-kit' ) }
```
tofumatt, line 78: I see this was this way in the IB, and is already written this way in the existing app (https://github.com/google/site-kit-wp/blob/develop/assets/js/modules/analytics/components/wp-dashboard/WPAnalyticsDashboardWidgetOverview.js#L115), but it's a shame because it's not very accessible (no `<def>` tag) and it isn't especially clear to translators what the abbreviation is.

Sure, in this case it's pretty easy to infer from context, but it might be worth improving this either now (if it isn't a UI issue to expand "Avg." to "Average" or later in a follow-up issue if it is...

assets/js/components/wp-dashboard/wp-dashboard-impressions.js
```diff
@@ -0,0 +1,94 @@
...
+		totalImpressions,
+		totalImpressionsChange,
+	} = processedData;
+
+	if ( loading ) {
+		return <PreviewBlock width="48%" height="92px" />;
+	}
+
+	if ( error ) {
+		trackEvent( 'plugin_setup', 'search_console_error', error.message );
```
tofumatt, line 75: This should be in a `useEffect` so that it runs only when `error` changes (and when `error` is truth-y).

As it stands this will track an error event on every render (anytime props change). That's not ideal as it could lead to extra events. We'd want something more like:

```js
useEffect(() => {
  if (error) {
    trackEvent( 'plugin_setup', 'search_console_error', error.message );
  }
}, [error]);
```

assets/js/components/wp-dashboard/wp-dashboard-unique-visitors.js
```diff
@@ -0,0 +1,90 @@
...
+import DataBlock from '../data-block';
+import { parseTotalUsersData, userReportDataDefaults } from '../../modules/analytics/util';
+const { useSelect } = Data;
+
+const WPDashboardUniqueVisitors = () => {
+	const {
+		data,
+		loading,
+		error,
+	} = useSelect( ( select ) => {
```
tofumatt, line 45: Earlier these were written all on one line, which I think is nicer:

```suggestion
	const { data, loading, error } = useSelect( ( select ) => {
```

We should at least be consistent, and I think the single line is legible so let's go with that 😃

assets/js/components/wp-dashboard/wp-dashboard-session-duration.js
```diff
@@ -0,0 +1,86 @@
...
+import ReportError from '../ReportError';
+import calculateOverviewData from '../../modules/analytics/util/calculateOverviewData';
+const { useSelect } = Data;
+
+const WPDashboardSessionDuration = () => {
+	const {
+		data,
+		loading,
+		error,
+	} = useSelect( ( select ) => {
```
tofumatt, line 43: Earlier these were written all on one line, which I think is nicer:

```suggestion
	const { data, loading, error } = useSelect( ( select ) => {
```

We should at least be consistent, and I think the single line is legible so let's go with that 😃

assets/js/components/wp-dashboard/wp-dashboard-app.js
```diff
@@ -0,0 +1,78 @@
...
+						<WPDashboardImpressions />
+						<WPDashboardClicks />
+						{ ( analyticsModuleActive && analyticsModuleConnected ) ? (
+							<Fragment>
+								<WPDashboardUniqueVisitors />
+								<WPDashboardSessionDuration />
+							</Fragment>
+						) : (
+							<AnalyticsInactiveCTA />
+						) }
```
tofumatt, line 69: Let's avoid ternaries for big sections of component markup like this wherever we can. It's harder to read and I think makes it less easier to read about the "if case" for something rendering or not.

```suggestion
						{ analyticsModuleActive && analyticsModuleConnected && (
							<Fragment>
								<WPDashboardUniqueVisitors />
								<WPDashboardSessionDuration />
							</Fragment>
						) }
						{ ( ! analyticsModuleActive || ! analyticsModuleConnected ) && <AnalyticsInactiveCTA /> }
```

That means we tie each section we render to a specific conditional—if either condition needs to change it can do so independently. It's what we've tried to use in general throughout the app and I think it's worth doing here too.

(Sidenote, I realise `( ! analyticsModuleActive || ! analyticsModuleConnected )` could be more concisely expressed with `! ( analyticsModuleActive && analyticsModuleConnected )` but the former is a bit clearer and I don't think that will lead to any bugs. 🙂 )

assets/js/components/wp-dashboard/wp-dashboard-app.js
```diff
@@ -0,0 +1,78 @@
...
+
+	return (
+		<div className="googlesitekit-wp-dashboard">
+			<div className="googlesitekit-wp-dashboard__cta">
+				<Link className="googlesitekit-wp-dashboard__cta-link" href={ dashboardURL }>
+					{ __( 'Visit your Site Kit Dashboard', 'google-site-kit' ) }
+				</Link>
+			</div>
+			<div className="googlesitekit-wp-dashboard-stats googlesitekit-wp-dashboard-stats--fourup">
+				{ wpDashboardEnabled && (
```
tofumatt, line 58:
```suggestion
				{ featureFlags.widgets.wpDashboard.enabled && (
```

assets/js/components/wp-dashboard/wp-dashboard-app.js
```diff
@@ -0,0 +1,78 @@
...
+import AnalyticsInactiveCTA from '../AnalyticsInactiveCTA';
+import { STORE_NAME as CORE_SITE } from '../../googlesitekit/datastore/site/constants';
+import { STORE_NAME as CORE_MODULES } from '../../googlesitekit/modules/datastore/constants';
+const { useSelect } = Data;
+
+const WPDashboardApp = () => {
+	const dashboardURL = useSelect( ( select ) => select( CORE_SITE ).getAdminURL( 'googlesitekit-dashboard' ) );
+	const analyticsModuleActive = useSelect( ( select ) => select( CORE_MODULES ).isModuleActive( 'analytics' ) );
+	const analyticsModuleConnected = useSelect( ( select ) => select( CORE_MODULES ).isModuleConnected( 'analytics' ) );
+	const wpDashboardEnabled = featureFlags.widgets.wpDashboard.enabled;
```
tofumatt, line 44: This variable is declared here only to be used once, immediately below. I don't think we need to explicitly set a local variable for this value; elsewhere we use `featureFlags.foo.enabled` variables directly in components. It makes them a bit easier to `grep` for as well.

```suggestion
	const wpDashboardEnabled = featureFlags.widgets.wpDashboard.enabled;
```

tofumatt (changes requested):
Looks good, but I think there are minor code styling things to fix and there's one issue with the `trackEvent` call being triggered on re-render that we should fix.

assets/js/components/wp-dashboard/wp-dashboard-session-duration.js
kostyalmm: If I change it to `Average`, at least I don't see any issue.

image
vs
image

Let me know if I should change it 😃

assets/js/components/wp-dashboard/wp-dashboard-unique-visitors.js
kostyalmm: I think we should open an issue for the same.

assets/js/components/wp-dashboard/wp-dashboard-clicks.js
```diff
@@ -0,0 +1,88 @@
...
+
+		const commonArgs = {
+			dateRange: select( CORE_USER ).getDateRange(),
+		};
+
+		const args = {
+			dimensions: 'date',
+			compareDateRanges: true,
+			...commonArgs,
+		};
```
tofumatt, line 50: Huh, I just noticed this one too. Why assign the `commArgs` variable here? It's not used anywhere except to be placed immediately into the dictionary below. Let's add it to the regular `args` variable instead:
```suggestion
		const args = {
			compareDateRanges: true,
			dateRange: select( CORE_USER ).getDateRange(),
			dimensions: 'date',
		};
```

assets/js/components/wp-dashboard/wp-dashboard-impressions.js
```diff
@@ -0,0 +1,100 @@
...
+
+		const commonArgs = {
+			dateRange: select( CORE_USER ).getDateRange(),
+		};
+
+		const args = {
+			dimensions: 'date',
+			compareDateRanges: true,
+			...commonArgs,
+		};
```
tofumatt, line 51: Again, `commonArgs` isn't used elsewhere here, so let's simplify this:

```suggestion
		const args = {
			compareDateRanges: true,
			dateRange: select( CORE_USER ).getDateRange(),
			dimensions: 'date',
		};
```

assets/js/components/wp-dashboard/wp-dashboard-impressions.js
```diff
@@ -0,0 +1,100 @@
...
+
+	useEffect( () => {
+		if ( error ) {
+			trackEvent( 'plugin_setup', 'search_console_error', error.message );
+		}
+	}, [ error ] );
+
+	if ( ! data || ! data.length ) {
+		return null;
+	}
```
tofumatt, line 68: Why/when would this be true? This is what the `isZeroData` is here to check, but otherwise we should return the `loading` state, correct?

Is `loading` not set correctly on first render? I'd expect `loading` to be `true` on first render based on the selectors here. If that's the case, this should be removed (that may require moving the `if ( loading ) { return <PreviewBlock width="48%" height="92px" />; }` conditional to here.

assets/js/components/wp-dashboard/wp-dashboard-app.js
```diff
@@ -0,0 +1,76 @@
...
+import { STORE_NAME as CORE_SITE } from '../../googlesitekit/datastore/site/constants';
+import { STORE_NAME as CORE_MODULES } from '../../googlesitekit/modules/datastore/constants';
+const { useSelect } = Data;
+
+const WPDashboardApp = () => {
+	const dashboardURL = useSelect( ( select ) => select( CORE_SITE ).getAdminURL( 'googlesitekit-dashboard' ) );
+	const analyticsModuleActive = useSelect( ( select ) => select( CORE_MODULES ).isModuleActive( 'analytics' ) );
+	const analyticsModuleConnected = useSelect( ( select ) => select( CORE_MODULES ).isModuleConnected( 'analytics' ) );
+
+	if ( ! dashboardURL ) {
```
tofumatt, line 45: This is a loading state, correct? Might be a bit nicer to be very explicit here—`dashboardURL === undefined` instead of `! dashboardURL`—I don't expect this should ever be `false` or `null`, just not yet loaded (even then, we tend to load it with API prefetch data, but the check is good to have).

Still, might be nice if it were a bit clearer as to _why_ nothing is rendering yet.

assets/js/components/wp-dashboard/wp-dashboard-app.js
```diff
@@ -0,0 +1,76 @@
...
+						{ analyticsModuleActive && analyticsModuleConnected && (
+							<Fragment>
+								<WPDashboardUniqueVisitors />
+								<WPDashboardSessionDuration />
+							</Fragment>
+						) }
+						{ ( ! analyticsModuleActive || ! analyticsModuleConnected ) && <AnalyticsInactiveCTA /> }
+					</Fragment>
+				) }
+			</div>
```
tofumatt, line 70: Should this feature flag check be _inside_ the `<div>` here? Why render the container div if it isn't set? Bonus: that removes the need for one of the Fragments:

```suggestion
			{ featureFlags.widgets.wpDashboard.enabled && (
				<div className="googlesitekit-wp-dashboard-stats googlesitekit-wp-dashboard-stats--fourup">
					<WPDashboardImpressions />
					<WPDashboardClicks />
					{ analyticsModuleActive && analyticsModuleConnected && (
						<Fragment>
							<WPDashboardUniqueVisitors />
							<WPDashboardSessionDuration />
						</Fragment>
					) }
					{ ( ! analyticsModuleActive || ! analyticsModuleConnected ) && <AnalyticsInactiveCTA /> }
				</div>
			) }
```

assets/js/components/wp-dashboard/wp-dashboard-clicks.js
```diff
@@ -0,0 +1,88 @@
...
+			loading: ! store.hasFinishedResolution( 'getReport', [ args ] ),
+		};
+	} );
+
+	if ( loading ) {
+		return <PreviewBlock width="48%" height="92px" />;
+	}
+
+	if ( error ) {
+		trackEvent( 'plugin_setup', 'search_console_error', error.message );
```
tofumatt, line 64: Looks like this component has the same issue with `trackEvent`—it should be placed in a `useEffect` so it's only called once.

assets/js/components/wp-dashboard/wp-dashboard-session-duration.js
```diff
@@ -0,0 +1,82 @@
...
+		const store = select( STORE_NAME );
+
+		const commonArgs = {
+			dateRange: select( CORE_USER ).getDateRange(),
+		};
+
+		const args = {
+			...overviewReportDataDefaults,
+			...commonArgs,
+		};
```
tofumatt, line 49: Let's remove `commArgs` here as well and place the `dateRange` into `args` directly as with other components.

assets/js/components/wp-dashboard/wp-dashboard-session-duration.js
tofumatt: My immediate thought was yes, but I did consider consistency here. I had a look at what the actual Analytics page does and it's also "Avg." where "Average" would totally work:

Screenshot 2020-12-08 at 16 18 20

Let's stick with "Avg." for now and see about proposing an upstream fix then mirroring the site's change. I think the consistency in labelling is useful so we'll stick with the (admittedly less-good) one for now. 😅

assets/js/components/wp-dashboard/wp-dashboard-unique-visitors.js
```diff
@@ -0,0 +1,86 @@
...
+		const store = select( STORE_NAME );
+
+		const commonArgs = {
+			dateRange: select( CORE_USER ).getDateRange(),
+		};
+
+		const args = {
+			...userReportDataDefaults,
+			...commonArgs,
+		};
```
tofumatt, line 51: Let's remove `commonArgs` here as well.

tofumatt (changes requested):
Looks good, a few more cleanups, an issue with loading state, and another `trackEvent` `useEffect` are needed but then we're good. 🙂

tofumatt (dismissed):
Looks great, thanks 👍🏻

assets/js/components/wp-dashboard/wp-dashboard-clicks.js
```diff
@@ -0,0 +1,90 @@
...
+import { STORE_NAME as CORE_USER } from '../../googlesitekit/datastore/user/constants';
+import { extractSearchConsoleDashboardData, isZeroReport } from '../../modules/search-console/util';
+import { trackEvent } from '../../util/tracking';
+import DataBlock from '../data-block';
+import PreviewBlock from '../PreviewBlock';
+import ReportError from '../ReportError';
+import ReportZero from '../ReportZero';
+const { useSelect } = Data;
+
+const WPDashboardClicks = () => {
```
aaemnnosttv, line 39: The file name needs to be updated to match here as well.

assets/js/components/wp-dashboard/wp-dashboard-clicks.js
```diff
@@ -0,0 +1,90 @@
...
+			dateRange: select( CORE_USER ).getDateRange(),
+			dimensions: 'date',
+			compareDateRanges: true,
+		};
+
+		return {
+			data: store.getReport( args ),
+			error: store.getErrorForSelector( 'getReport', [ args ] ),
+			loading: ! store.hasFinishedResolution( 'getReport', [ args ] ),
+		};
```
aaemnnosttv, line 53: Please split this into separate `useSelect` calls (one for each `data`, `error`, and `loading` – and maybe one more for `args`) as is our convention going forward.

assets/js/components/wp-dashboard/wp-dashboard-clicks.js
```diff
@@ -0,0 +1,90 @@
...
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { useEffect } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import Data from 'googlesitekit-data';
+import { STORE_NAME } from '../../modules/search-console/datastore/constants';
```
aaemnnosttv, line 29: I think @tofumatt called this out (or similar) before but this should be aliased to indicate the store it refers to since this component is not specific to Search Console.

```suggestion
import { STORE_NAME as MODULES_SEARCH_CONSOLE } from '../../modules/search-console/datastore/constants';
```

assets/js/components/wp-dashboard/wp-dashboard-impressions.js
```diff
@@ -0,0 +1,90 @@
...
+ * WordPress dependencies
+ */
+import { useEffect } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import Data from 'googlesitekit-data';
+import { STORE_NAME } from '../../modules/search-console/datastore/constants';
```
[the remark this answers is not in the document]
aaemnnosttv: I knew I had seen it somewhere 🙂  let's update this as mentioned previously as it is our convention to do so any time we're not in the context of a specific module's store. I agree with @tofumatt that we should export these as constants rather than relying on this convention, but let's update it anyways for consistency. 

```suggestion
import { STORE_NAME as MODULES_SEARCH_CONSOLE } from '../../modules/search-console/datastore/constants';
```

assets/js/components/wp-dashboard/wp-dashboard-impressions.js
```diff
@@ -0,0 +1,90 @@
...
+			dimensions: 'date',
+			compareDateRanges: true,
+		};
+
+		return {
+			data: store.getReport( args ),
+			error: store.getErrorForSelector( 'getReport', [ args ] ),
+			loading: ! store.hasFinishedResolution( 'getReport', [ args ] ),
+		};
+	} );
```
aaemnnosttv, line 54: This should be updated to use individual `useSelect` calls as above.

assets/js/components/wp-dashboard/wp-dashboard-session-duration.js
```diff
@@ -0,0 +1,78 @@
...
+			dateRange: select( CORE_USER ).getDateRange(),
+			...overviewReportDataDefaults,
+		};
+
+		return {
+			data: store.getReport( args ),
+			loading: ! store.hasFinishedResolution( 'getReport', [ args ] ),
+			error: store.getErrorForSelector( 'getReport', [ args ] ),
+		};
+	} );
```
aaemnnosttv, line 52: Please use separate `useSelect` calls.

assets/js/components/wp-dashboard/wp-dashboard-session-duration.js
```diff
@@ -0,0 +1,78 @@
...
+ */
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import Data from 'googlesitekit-data';
+import { prepareSecondsForDisplay } from '../../util';
+import DataBlock from '../data-block';
+import { STORE_NAME } from '../../modules/analytics/datastore/constants';
```
aaemnnosttv, line 30:
```suggestion
import { STORE_NAME as MODULES_ANALYTICS } from '../../modules/analytics/datastore/constants';
```

assets/js/components/wp-dashboard/wp-dashboard-unique-visitors.js
```diff
@@ -0,0 +1,82 @@
...
+			dateRange: select( CORE_USER ).getDateRange(),
+			...userReportDataDefaults,
+		};
+
+		return {
+			data: store.getReport( args ),
+			loading: ! store.hasFinishedResolution( 'getReport', [ args ] ),
+			error: store.getErrorForSelector( 'getReport', [ args ] ),
+		};
+	} );
```
aaemnnosttv, line 54: Please use separate `useSelect` calls.

assets/js/components/wp-dashboard/wp-dashboard-unique-visitors.js
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import Data from 'googlesitekit-data';
+import { STORE_NAME } from '../../modules/analytics/datastore/constants';
```
aaemnnosttv, line 28:
```suggestion
import { STORE_NAME as MODULES_ANALYTICS } from '../../modules/analytics/datastore/constants';
```

stories/wp-dashboard.stories.js
```diff
@@ -40,12 +40,14 @@ import WPAnalyticsDashboardWidgetTopPagesTable from '../assets/js/modules/analyt
...
 import { STORE_NAME as CORE_USER } from '../assets/js/googlesitekit/datastore/user/constants';
 import { WithTestRegistry } from '../tests/js/utils';
+import { disableFeature } from './utils/features';
 
 storiesOf( 'WordPress', module )
 	.add( 'WordPress Dashboard', () => {
 		global._googlesitekitLegacyData = wpDashboardData;
 		global._googlesitekitLegacyData.admin.assetsRoot = ''; // See .storybook/config.js
 		global._googlesitekitLegacyData.canAdsRun = true;
+		disableFeature( 'widgets.wpDashboard' );
```
aaemnnosttv, line 50: This call shouldn't be necessary as all features are to be disabled as part of the setup for each story (this is a more recent change so you probably haven't seen it yet).

Please add it here instead 🙂  https://github.com/google/site-kit-wp/blob/b9e590c436de83d1b198bcbf5ffc6c70b86f2c70/.storybook/config.js#L39-L45

aaemnnosttv (changes requested):
Thanks @kostyalmm ! There are still a few items to address here; see my comments below.

Particularly this item from the ACs applies to several files (I didn't comment on all of them):

> All files in `assets/js/components/wp-dashboard` should have names that match the component name that they export

Otherwise there are a few instances where there is some "deprecated" style used that we shouldn't copy over from the legacy implementations as well as a few missing aliases.

assets/js/components/wp-dashboard/WPDashboardSessionDuration.js
```diff
@@ -0,0 +1,74 @@
...
+import calculateOverviewData from '../../modules/analytics/util/calculateOverviewData';
+const { useSelect } = Data;
+
+const WPDashboardSessionDuration = () => {
+	const dateRange = useSelect( ( select ) => select( CORE_USER ).getDateRange() );
+
+	const args = {
+		dateRange,
+		...overviewReportDataDefaults,
+	};
```
aaemnnosttv, line 44: I don't think we should be using these old args anymore (see #2236 , #2237) but that may be outside of the scope of this ticket.

@felixarntz what do you think?

assets/js/components/wp-dashboard/WPDashboardUniqueVisitors.js
```diff
@@ -0,0 +1,78 @@
...
+import { parseTotalUsersData, userReportDataDefaults } from '../../modules/analytics/util';
+const { useSelect } = Data;
+
+const WPDashboardUniqueVisitors = () => {
+	const dateRange = useSelect( ( select ) => select( CORE_USER ).getDateRange() );
+
+	const args = {
+		dateRange,
+		...userReportDataDefaults,
+	};
```
aaemnnosttv, line 46: See above

assets/js/components/wp-dashboard/WPDashboardImpressions.js
```diff
@@ -0,0 +1,86 @@
...
+import PreviewBlock from '../PreviewBlock';
+import ReportError from '../ReportError';
+import ReportZero from '../ReportZero';
+const { useSelect } = Data;
+
+const WPDashboardImpressions = () => {
+	const dateRange = useSelect( ( select ) => select( CORE_USER ).getDateRange() );
+
+	const args = {
+		dateRange,
```
aaemnnosttv, line 43: Should this use start/end dates instead?

assets/js/components/wp-dashboard/WPDashboardClicks.js
```diff
@@ -0,0 +1,86 @@
...
+import PreviewBlock from '../PreviewBlock';
+import ReportError from '../ReportError';
+import ReportZero from '../ReportZero';
+const { useSelect } = Data;
+
+const WPDashboardClicks = () => {
+	const dateRange = useSelect( ( select ) => select( CORE_USER ).getDateRange() );
+
+	const args = {
+		dateRange,
```
aaemnnosttv, line 43: Should this use start/end dates instead?

aaemnnosttv (commented):
LGTM, just one question for @felixarntz regarding the scope of this issue.

assets/js/components/wp-dashboard/WPDashboardApp.js
```diff
@@ -0,0 +1,74 @@
...
+					<WPDashboardClicks />
+					{ analyticsModuleActive && analyticsModuleConnected && (
+						<Fragment>
+							<WPDashboardUniqueVisitors />
+							<WPDashboardSessionDuration />
+						</Fragment>
+					) }
+					{ ( ! analyticsModuleActive || ! analyticsModuleConnected ) && <AnalyticsInactiveCTA /> }
+				</div>
+			) }
```
felixarntz, line 68: Looks like we need to switch the order here, the two Analytics widgets come before the two Search Console widgets in the current verrsion of the WP dashboard widget, so that should be maintained.

assets/js/components/wp-dashboard/WPDashboardClicks.js
felixarntz: @kostyalmm Indeed, we should no longer pass `dateRange` (and the related `compareDateRanges` and `multiDateRange`) arguments as they are legacy. Please rely on the `getDateRangeDates` selector in the same way that https://github.com/google/site-kit-wp/blob/develop/assets/js/components/adminbar/AdminBarClicks.js does.

assets/js/components/wp-dashboard/WPDashboardClicks.js
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	if ( isZeroReport( data ) ) {
+		return <ReportZero moduleSlug="search-console" />;
+	}
+
+	const {
+		totalClicks,
+		totalClicksChange,
+	} = extractSearchConsoleDashboardData( data );
```
felixarntz, line 73: We should also no longer rely on this legacy function but use a more specific calculation via `sumObjectListValue` and `calculateChange` - same as in https://github.com/google/site-kit-wp/blob/develop/assets/js/components/adminbar/AdminBarClicks.js.

assets/js/components/wp-dashboard/WPDashboardImpressions.js
felixarntz: @kostyalmm Same here, we should use `getDateRangeDates` like in https://github.com/google/site-kit-wp/blob/develop/assets/js/components/adminbar/AdminBarImpressions.js.

assets/js/components/wp-dashboard/WPDashboardImpressions.js
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	if ( isZeroReport( data ) ) {
+		return <ReportZero moduleSlug="search-console" />;
+	}
+
+	const {
+		totalImpressions,
+		totalImpressionsChange,
+	} = extractSearchConsoleDashboardData( data );
```
felixarntz, line 73: See above, let's use the new functions instead, the same way as https://github.com/google/site-kit-wp/blob/develop/assets/js/components/adminbar/AdminBarImpressions.js does.

assets/js/components/wp-dashboard/WPDashboardSessionDuration.js
felixarntz: @kostyalmm See above, this should use `getDateRangeDates` for passing the dates and also manually pass the other arguments instead of relying on the legacy `overviewReportDataDefaults`. You can probably copy most of that from https://github.com/google/site-kit-wp/blob/develop/assets/js/components/adminbar/AdminBarSessions.js, except that the expression should be `ga:avgSessionDuration` and the alias `Average Session Duration`.

assets/js/components/wp-dashboard/WPDashboardSessionDuration.js
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	if ( error ) {
+		return <ReportError moduleSlug="analytics" error={ error } />;
+	}
+
+	const {
+		averageSessionDuration,
+		averageSessionDurationChange,
+	} = calculateOverviewData( data );
```
felixarntz, line 61: Same here, we shouldn't use the legacy `calculateOverviewData`, but instead more specifically extract the values from the response and `calculateChange` - same as in https://github.com/google/site-kit-wp/blob/develop/assets/js/components/adminbar/AdminBarSessions.js.

assets/js/components/wp-dashboard/WPDashboardSessionDuration.js
```diff
@@ -0,0 +1,74 @@
...
+
+	return (
+		<DataBlock
+			className="googlesitekit-wp-dashboard-stats__data-table overview-average-session-duration"
+			title={ __( 'Avg. Time on Page', 'google-site-kit' ) }
+			datapoint={ prepareSecondsForDisplay( averageSessionDuration ) }
```
felixarntz, line 67: After #2200 merged, this no longer works - we need to pass the value directly and instead pass "s" as `datapointUnit` for correct formatting. Right now the widget just displays 0.

assets/js/components/wp-dashboard/WPDashboardUniqueVisitors.js
felixarntz: @kostyalmm See above, let's use `getDateRangeDates` to pass actual dates and specify all arguments manually instead of relying on the legacy function. It needs to be pretty much the same as https://github.com/google/site-kit-wp/blob/develop/assets/js/components/adminbar/AdminBarUniqueVisitors.js.

assets/js/components/wp-dashboard/WPDashboardUniqueVisitors.js
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	if ( error ) {
+		return <ReportError moduleSlug="analytics" error={ error } />;
+	}
+
+	const {
+		totalUsers,
+		previousTotalUsers,
+	} = parseTotalUsersData( data );
```
felixarntz, line 63: Same as in the other components, let's not use this legacy function and instead extract the values manually, same as in https://github.com/google/site-kit-wp/blob/develop/assets/js/components/adminbar/AdminBarUniqueVisitors.js.

assets/js/components/wp-dashboard/WPDashboardUniqueVisitors.js
```diff
@@ -0,0 +1,78 @@
...
+
+	const totalUsersChange = calculateChange( previousTotalUsers, totalUsers );
+
+	return (
+		<DataBlock
+			className="googlesitekit-wp-dashboard-stats__data-table overview-total-users"
+			title={ __( 'Total Unique Visitors', 'google-site-kit' ) }
+			datapoint={ readableLargeNumber( totalUsers ) }
```
felixarntz, line 71: After #2200 this is broken, the call to `readableLargeNumber` should be removed as all formatting happens inside `DataBlock`.

felixarntz (changes requested):
@kostyalmm I fixed the merge conflicts after the recent major changes that got merged.

Overall the code looks solid, however there are a bunch of legacy functions still used that should no longer be leveraged, no `dateRange` arguments should be passed, and a few things need to be fixed now that #2200 is merged.

Most of it should be simple though because you can pretty much copy-paste similar code from the admin bar components.

assets/js/components/wp-dashboard/WPDashboardClicks.js
```diff
@@ -28,21 +28,29 @@ import { useEffect } from '@wordpress/element';
...
+		offsetDays: DATE_RANGE_OFFSET,
+	} ) );
 
 	const args = {
-		dateRange,
+		startDate: compareStartDate,
+		endDate,
 		dimensions: 'date',
-		compareDateRanges: true,
+		url,
```
felixarntz, line 53: We must not pass a `url` here, the widget is always for the entire site, not a specific URL.

assets/js/components/wp-dashboard/WPDashboardImpressions.js
```diff
@@ -28,21 +28,29 @@ import { __ } from '@wordpress/i18n';
...
+		offsetDays: DATE_RANGE_OFFSET,
+	} ) );
 
 	const args = {
-		dateRange,
+		startDate: compareStartDate,
+		endDate,
 		dimensions: 'date',
-		compareDateRanges: true,
+		url,
```
felixarntz, line 53: See above.

assets/js/components/wp-dashboard/WPDashboardSessionDuration.js
```diff
@@ -25,27 +25,38 @@ import { __ } from '@wordpress/i18n';
...
+		...dateRangeDates,
+		dimensions: 'ga:date',
+		limit: 10,
+		metrics: [
+			{
+				expression: 'ga:avgSessionDuration',
+				alias: 'Average Session Duration',
+			},
+		],
+		url,
```
felixarntz, line 54: See above, the `url` parameter must be removed again.

assets/js/components/wp-dashboard/WPDashboardUniqueVisitors.js
```diff
@@ -25,29 +25,36 @@ import { __ } from '@wordpress/i18n';
...
-		...userReportDataDefaults,
+	const reportArgs = {
+		...dateRangeDates,
+		metrics: [
+			{
+				expression: 'ga:users',
+				alias: 'Total Users',
+			},
+		],
+		url,
```
felixarntz, line 52: See above.

assets/js/components/wp-dashboard/WPDashboardApp.js
```diff
@@ -0,0 +1,86 @@
...
+
+	return (
+		<div className="googlesitekit-wp-dashboard">
+			<div className="googlesitekit-wp-dashboard__cta">
+				<Link className="googlesitekit-wp-dashboard__cta-link" href={ dashboardURL }>
+					{ __( 'Visit your Site Kit Dashboard', 'google-site-kit' ) }
+				</Link>
+			</div>
+			{ featureFlags.widgets.wpDashboard.enabled && (
+				<div className="googlesitekit-wp-dashboard-stats googlesitekit-wp-dashboard-stats--fourup">
```
felixarntz, line 59: We shouldn't need the new class `googlesitekit-wp-dashboard-stats--fourup` here, but reuse the existing styling since we visually need to replicate the old widget anyway.

assets/js/components/wp-dashboard/WPDashboardApp.js
```diff
@@ -0,0 +1,86 @@
...
+						<Cell mdSize={ 4 } lgSize={ 6 }>
+							{ ! analyticsModuleActive && (
+								<AnalyticsInactiveCTA />
+							) }
+
+							{ ( analyticsModuleActive && ! analyticsModuleConnected ) && (
+								<CompleteModuleActivationCTA slug="analytics" />
+							) }
+						</Cell>
+					) }
```
felixarntz, line 70: It looks like the CTA needs to actually _follow_ the regular `DataBlock`s, so this needs to be moved below. Furthermore, the CTA needs to be wrapped in a `div.googlesitekit-wp-dashboard-stats__cta`. We need to follow the DOM structure of the existing widget here.

felixarntz (changes requested):
There are a few problems here where the visual appearance clearly differs from the current experience.

felixarntz (approved):
I've addressed the last feedback here, this now looks the same way as the current version of the widget. We missed one part of the widget entirely (already during definition), for which I opened #2610 to address as a follow-up.

Issue #2421, opened 2020-11-24
Add support for querying Analytics reports with dimension filters


felixarntz:
Currently it is not possible to use dimension filters when requesting an Analytics report. This should be implemented as it is needed for the new Analytics All Traffic widget (see #2420).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `GET:report` datapoint in the Analytics class should be expanded to support an optional `dimensionFilters` argument:
    * The argument should be specifiable in a simple way, as a map of `dimensionName => dimensionValue` pairs. They should then be parsed into the necessary `Google_Service_AnalyticsReporting_DimensionFilter` (see also https://developers.google.com/analytics/devguides/reporting/core/v4/rest/v4/reports/batchGet#DimensionFilter) structure required for the Google API request internally, with operator `EXACT`. All entries in the map should then be added as a new `Google_Service_AnalyticsReporting_DimensionFilterClause` (see also https://developers.google.com/analytics/devguides/reporting/core/v4/rest/v4/reports/batchGet#DimensionFilterClause), with operator `AND`.
    * This dimension filter clause then needs to be combined with the existing dimension filter clause based on the `page` query parameter (see `Analytics::create_analytics_site_data_request`).
* The `getReport` selector of the `modules/analytics` datastore should be enhanced accordingly, so that the new `dimensionFilters` argument is supported. The argument should also be documented as one of the selector's `options`.

## Implementation Brief

* Update the `Google\Site_Kit\Modules\Analytics` class:
  * Check if the `$data['dimensionFilters']` isn't empty the handler for the `GET:report` endpoint and copy over its values into the appropriate `dimensionFilters` property of the `$request_args` variable. Please, make sure you to filter out dimensions that are not listed in the `$request_args['dimensions']` array;
  * Update the `create_analytics_site_data_request` method to populate dimension filters if `$args['dimensionFilters']` property isn't empty. See AC for additional details.
* Update the report datastore in the `assets/js/modules/analytics/datastore/report.js` file:
  * Add `dimensionFilters` validator to the `validateParams` function of the `fetchGetReportStore` store. It should check whether `dimensionFilters` is a map of dimension names and values and all dimension names are also defined in the `dimensions` option.
  * Update jsdoc for the `getReport` selector to list `dimensionFilters` argument of the options paramter.

### Test Coverage

* Adjust existing tests if any are failing.

### Visual Regression Changes

* Not expected.

## QA Brief

[...]

PR #2485, opened 2020-12-06
Enhancement/2421 query analytics reports dimension filters

ivankruchkoff:
## Summary

Addresses issue #2421 

## Checklist


assets/js/modules/analytics/util/report-validation.js
```diff
@@ -57,3 +57,38 @@ export function isValidDimensions( dimensions ) {
...
+
+	if ( Array.isArray( dimensions ) ) {
+		dimensions.forEach( ( singleDimension ) => parseDimension( singleDimension, validDimensions ) );
+	} else {
+		parseDimension( dimensions, validDimensions );
+	}
+
+	// Ensure every dimensionFilter key corresponds to a valid dimension and ensure every value is a string.
+	return Object.keys( dimensionFilters ).every( ( dimensionFilterKey ) => validDimensions.include( dimensionFilterKey ) ) &&
+		Object.values( dimensionFilters ).every( ( dimensionFilterValue ) => typeof dimensionFilterValue === 'string' && dimensionFilterValue.length );
```
eugene-manuilov, line 93: @ivankruchkoff I think we don't need to verify that all values are non-empty strings because there might be a case when we want to pass numeric values for certain dimensions. So, let's change it to check only keys.

assets/js/modules/analytics/util/report-validation.js
```diff
@@ -57,3 +57,38 @@ export function isValidDimensions( dimensions ) {
...
+	};
+
+	if ( Array.isArray( dimensions ) ) {
+		dimensions.forEach( ( singleDimension ) => parseDimension( singleDimension, validDimensions ) );
+	} else {
+		parseDimension( dimensions, validDimensions );
+	}
+
+	// Ensure every dimensionFilter key corresponds to a valid dimension and ensure every value is a string.
+	return Object.keys( dimensionFilters ).every( ( dimensionFilterKey ) => validDimensions.include( dimensionFilterKey ) ) &&
```
eugene-manuilov, line 92: I believe we can simplify it to this:

```suggestion
	return Object.keys( dimensionFilters ).every( validDimensions.include, validDimensions ) &&
```

includes/Modules/Analytics.php
```diff
@@ -962,6 +962,33 @@ function ( $dimension_def ) {
...
 				}
 
+				$dimension_filters          = $data['dimensionFilters'];
+				$dimension_filter_instances = array();
+				if ( ! empty( $dimension_filters ) && ( is_string( $dimension_filters ) || is_array( $dimension_filters ) ) ) {
+					if ( is_string( $dimension_filters ) ) {
+						$dimension_filters = explode( ',', $dimension_filters );
+					} elseif ( is_array( $dimension_filters ) && ! wp_is_numeric_array( $dimension_filters ) ) { // If single object is passed.
+						$dimension_filters = array( $dimension_filters );
+					}
```
eugene-manuilov, line 972: @ivankruchkoff `$data['dimensionFilters']` should always be an associated array (see your `isValidDimensionFilters` function), so we don't need this checks. If the `dimensionFilters` is not an array, we can ignore it.

includes/Modules/Analytics.php
```diff
@@ -1379,14 +1418,19 @@ protected function create_analytics_site_data_request( array $args = array() ) {
 			$request->setDateRanges( array( $date_range ) );
 		}
 
+		if ( ! empty( $args['dimension_filters'] ) ) {
+			$request->setDimensions( (array) $args['dimension_filters'] );
+		}
+
```
eugene-manuilov, line 1424: @ivankruchkoff do we really need to set dimensions here using `$args['dimension_filters']`? We already set dimensions using `$args['dimensions']` a few steps above.

includes/Modules/Analytics.php
```diff
@@ -1379,14 +1418,19 @@ protected function create_analytics_site_data_request( array $args = array() ) {
...
 			$dimension_filter->setDimensionName( 'ga:pagePath' );
 			$dimension_filter->setOperator( 'EXACT' );
 			$args['page'] = str_replace( trim( $this->context->get_reference_site_url(), '/' ), '', esc_url_raw( $args['page'] ) );
 			$dimension_filter->setExpressions( array( rawurldecode( $args['page'] ) ) );
-			$dimension_filter_clause = new Google_Service_AnalyticsReporting_DimensionFilterClause();
-			$dimension_filter_clause->setFilters( array( $dimension_filter ) );
+			$dimension_filters_clauses[] = $dimension_filter;
+			$dimension_filter_clause     = new Google_Service_AnalyticsReporting_DimensionFilterClause();
+			$dimension_filter_clause->setFilters( $dimension_filters_clauses );
 			$request->setDimensionFilterClauses( array( $dimension_filter_clause ) );
```
eugene-manuilov, line 1434: @ivankruchkoff we need to move this out of the `if ( ! empty( $args['page'] ) )` condition to make sure that dimension filters work even if the page is not empty.

assets/js/modules/analytics/datastore/report.js
```diff
@@ -72,6 +72,13 @@ const fetchGetReportStore = createFetchStore( {
 			);
 		}
 
+		if ( dimensionFilters ) {
+			invariant(
+				isValidDimensionFilters( dimensionFilters, dimensions ),
+				'Dimensions filters must be an object where the keys are valid Dimensions and values are non-zero length strings.',
```
eugene-manuilov, line 78: We probably need to adjust this message since we expect values not to be just non-zero strings.

eugene-manuilov (changes requested):
Thanks, @ivankruchkoff! I have found a few things that we need to address before I can approve this PR. Could you please take a look at it?

includes/Modules/Analytics.php
ivankruchkoff: Removed.

assets/js/modules/analytics/util/report-validation.js
ivankruchkoff: Addressed

includes/Modules/Analytics.php
ivankruchkoff: Addressed.

assets/js/modules/analytics/datastore/report.js
ivankruchkoff: Addressed.

includes/Modules/Analytics.php
ivankruchkoff: Perhaps I was reading too literally into AC ```This dimension filter clause then needs to be combined with the existing dimension filter clause based on the page query parameter (see Analytics::create_analytics_site_data_request).```

ivankruchkoff: Addressed.

eugene-manuilov (dismissed):
Awesome! Thanks, @ivankruchkoff!

aaemnnosttv (changes requested):
Thanks @ivankruchkoff – this is looking good overall, just a few changes to address.

assets/js/modules/analytics/datastore/report.js
```diff
@@ -57,7 +57,7 @@ const fetchGetReportStore = createFetchStore( {
 		invariant( isPlainObject( options ), 'Options for Analytics report must be an object.' );
 		invariant( isValidDateRange( options ), 'Either date range or start/end dates must be provided for Analytics report.' );
 
-		const { metrics, dimensions, orderby } = normalizeReportOptions( options );
+		const { metrics, dimensions, dimensionFilters, orderby } = normalizeReportOptions( options );
```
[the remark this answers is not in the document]
ivankruchkoff: I've add a function to normalize dimensions in https://github.com/google/site-kit-wp/pull/2485/commits/a439154ad17f08504a06fb79068d10ff48fe7df8 however not sure what the dimensionFilters possible unnormalized values would be as it's a new param we're introducing. Any thoughts @aaemnnosttv ?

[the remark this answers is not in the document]
ivankruchkoff: The best I could think of is https://github.com/google/site-kit-wp/pull/2485/commits/e623d0317ebea23a001043f817233df3fbe02dd1

assets/js/modules/analytics/util/report-validation.js
```diff
@@ -57,3 +57,37 @@ export function isValidDimensions( dimensions ) {
...
+/**
+ * Verifies provided dimensionFilters to make sure they match allowed values found in dimensions.
+ *
+ * @since n.e.x.t
+ *
+ * @param {Object}                               dimensionFilters The dimension filters to check.
+ * @param {string|string[]|Object|Object[]|null} dimensions       The dimensions to check.
+ * @return {boolean} TRUE if dimension filters are valid, otherwise FALSE.
+ */
+export function isValidDimensionFilters( dimensionFilters, dimensions ) {
```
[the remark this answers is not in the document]
ivankruchkoff: Added in https://github.com/google/site-kit-wp/pull/2485/commits/df1676361aa88f1cbff68dce9e8b72a783a15bed

assets/js/modules/analytics/datastore/report.js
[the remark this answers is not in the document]
aaemnnosttv: Re: normalizing `dimensionFilters`, I was mostly referring to the type since I'm not exactly sure what the possible values can be here, but if the types are limited then I think what you have is good 👍

includes/Modules/Analytics.php
```diff
@@ -962,6 +962,27 @@ function ( $dimension_def ) {
 					}
 				}
 
+				$dimension_filters          = $data['dimensionFilters'];
+				$dimension_filter_instances = array();
+				if ( ! empty( $dimension_filters ) && is_array( $dimension_filters ) ) {
+					foreach ( $request_args['dimensions'] as $dimension ) {
+						$dimension_name = $dimension->getName();
+						if ( isset( $dimension_filters[ $dimension_name ] ) ) {
```
aaemnnosttv, line 970: Let's avoid the additional nesting here by using `continue` if the key isn't set.

aaemnnosttv (commented):
Thanks @ivankruchkoff – this is very close, just a few final things to touch up then this should be good to go.

assets/js/modules/analytics/util/report-normalization.js
```diff
@@ -51,3 +53,26 @@ const normalizeMetrics = ( metrics ) => {
...
+
+const normalizeDimensionFilters = ( dimensionFilters ) => {
+	const normalizedFilters = {};
+	if ( isPlainObject( dimensionFilters ) ) {
+		for ( const [ key, value ] of Object.entries( dimensionFilters ) ) {
+			if ( [ 'string', 'number', 'boolean' ].includes( typeof value ) ) {
+				normalizedFilters[ key ] = value;
+			}
+		}
+	}
```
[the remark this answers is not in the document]
ivankruchkoff: Updated.

includes/Modules/Analytics.php
ivankruchkoff: Updated.

assets/js/modules/analytics/util/report-validation.test.js
```diff
@@ -64,6 +64,35 @@ describe( 'Analytics Reporting API validation', () => {
 		} );
 	} );
 
+	describe( 'isValidDimensionFilters', () => {
+		it( 'should return TRUE if a valid object is passed with a valid dimension', () => {
+			expect( isValidDimensionFilters( {
+				test: 'foo',
+			}, [ {
+				name: 'test',
+			} ] ) ).toBeTruthy();
```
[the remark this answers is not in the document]
ivankruchkoff: Updated.

includes/Modules/Analytics.php
```diff
@@ -1335,27 +1358,30 @@ function ( Google_Service_Analytics_Account $account ) {
...
 	 * @param array $args {
 	 *     Optional. Additional arguments.
 	 *
-	 *     @type array  $dimensions List of request dimensions. Default empty array.
-	 *     @type string $start_date Start date in 'Y-m-d' format. Default empty string.
-	 *     @type string $end_date   End date in 'Y-m-d' format. Default empty string.
-	 *     @type string $page       Specific page URL to filter by. Default empty string.
-	 *     @type int    $row_limit  Limit of rows to return. Default 100.
+	 *     @type array  $dimensions        List of request dimensions. Default empty array.
+	 *     @type array  $dimension_filters List of dimension filters for the specified request dimensions. Default empty array.
```
aaemnnosttv, line 1367: Let's add the type here as it isn't quite clear that this argument requires `Google_Service_AnalyticsReporting_DimensionFilter` instances.

```suggestion
	 *     @type Google_Service_AnalyticsReporting_DimensionFilter[]  $dimension_filters List of dimension filter instances for the specified request dimensions. Default empty array.
```

includes/Modules/Analytics.php
```diff
@@ -962,6 +962,29 @@ function ( $dimension_def ) {
...
+						if ( ! isset( $dimension_filters[ $dimension_name ] ) ) {
+							continue;
+						}
+
+						$dimension_value  = $dimension_filters[ $dimension_name ];
+						$dimension_filter = new Google_Service_AnalyticsReporting_DimensionFilter();
+						$dimension_filter->setDimensionName( $dimension_name );
+						$dimension_filter->setOperator( 'EXACT' );
+						$dimension_filter->setExpressions( array( $dimension_value ) );
+						$dimension_filter->setCaseSensitive( false );
```
aaemnnosttv, line 979: Let's remove this since the default is `false` and it isn't mentioned in the AC.
```suggestion
```

includes/Modules/Analytics.php
```diff
@@ -1385,8 +1420,12 @@ protected function create_analytics_site_data_request( array $args = array() ) {
 			$dimension_filter->setOperator( 'EXACT' );
 			$args['page'] = str_replace( trim( $this->context->get_reference_site_url(), '/' ), '', esc_url_raw( $args['page'] ) );
 			$dimension_filter->setExpressions( array( rawurldecode( $args['page'] ) ) );
+			$dimension_filters_clauses[] = $dimension_filter;
```
aaemnnosttv, line 1423: This is incorrect – only the `Google_Service_AnalyticsReporting_DimensionFilterClause` should be added to this list, not the `Google_Service_AnalyticsReporting_DimensionFilter` instance (as it is in the current version of the file on the base branch).

includes/Modules/Analytics.php
```diff
@@ -1385,8 +1420,12 @@ protected function create_analytics_site_data_request( array $args = array() ) {
...
+		}
+
+		if ( ! empty( $dimension_filters_clauses ) ) {
 			$dimension_filter_clause = new Google_Service_AnalyticsReporting_DimensionFilterClause();
-			$dimension_filter_clause->setFilters( array( $dimension_filter ) );
+			$dimension_filter_clause->setFilters( $dimension_filters_clauses );
```
aaemnnosttv, line 1428: This clause should remain independent as it is in develop now, but `setFilters` should be called with `array( $dimension_filter )` like it was before rather than the array of clauses.

includes/Modules/Analytics.php
```diff
@@ -1385,8 +1420,12 @@ protected function create_analytics_site_data_request( array $args = array() ) {
...
 			$request->setDimensionFilterClauses( array( $dimension_filter_clause ) );
```
aaemnnosttv, line 1429: This is where the array of clauses should be set 😄 , although this condition should remain separate from the conditions that define the clauses themselves.
```suggestion
			$request->setDimensionFilterClauses( $dimension_filters_clauses );
```

aaemnnosttv (commented):
Thanks @ivankruchkoff ! This is pretty close, just a few details to square away on the PHP side 👍

includes/Modules/Analytics.php
ivankruchkoff: Yes, you're correct.

includes/Modules/Analytics.php
ivankruchkoff: That's right, changed it back.

includes/Modules/Analytics.php
```diff
@@ -1335,27 +1357,30 @@ function ( Google_Service_Analytics_Account $account ) {
...
-	 *     @type string $start_date Start date in 'Y-m-d' format. Default empty string.
-	 *     @type string $end_date   End date in 'Y-m-d' format. Default empty string.
-	 *     @type string $page       Specific page URL to filter by. Default empty string.
-	 *     @type int    $row_limit  Limit of rows to return. Default 100.
+	 *     @type array  $dimensions        List of request dimensions. Default empty array.
+	 *     @type Google_Service_AnalyticsReporting_DimensionFilter[]  $dimension_filters List of dimension filter instances for the specified request dimensions. Default empty array.
+	 *     @type string $start_date        Start date in 'Y-m-d' format. Default empty string.
+	 *     @type string $end_date          End date in 'Y-m-d' format. Default empty string.
+	 *     @type string $page              Specific page URL to filter by. Default empty string.
+	 *     @type int    $row_limit         Limit of rows to return. Default 100.
```
aaemnnosttv, line 1370: These need to be aligned now after the `@type` change for `$dimension_filters`. Odd that PHPCS doesn't flag it - it must be this "inner" declaration isn't a linted the same way. No worries.

```suggestion
	 *     @type array                                               $dimensions        List of request dimensions. Default empty array.
	 *     @type Google_Service_AnalyticsReporting_DimensionFilter[] $dimension_filters List of dimension filter instances for the specified request dimensions. Default empty array.
	 *     @type string                                              $start_date        Start date in 'Y-m-d' format. Default empty string.
	 *     @type string                                              $end_date          End date in 'Y-m-d' format. Default empty string.
	 *     @type string                                              $page              Specific page URL to filter by. Default empty string.
	 *     @type int                                                 $row_limit         Limit of rows to return. Default 100.
```

includes/Modules/Analytics.php
```diff
@@ -1385,9 +1419,12 @@ protected function create_analytics_site_data_request( array $args = array() ) {
 			$dimension_filter->setOperator( 'EXACT' );
 			$args['page'] = str_replace( trim( $this->context->get_reference_site_url(), '/' ), '', esc_url_raw( $args['page'] ) );
 			$dimension_filter->setExpressions( array( rawurldecode( $args['page'] ) ) );
+		}
+
+		if ( ! empty( $dimension_filters_clauses ) ) {
 			$dimension_filter_clause = new Google_Service_AnalyticsReporting_DimensionFilterClause();
 			$dimension_filter_clause->setFilters( array( $dimension_filter ) );
-			$request->setDimensionFilterClauses( array( $dimension_filter_clause ) );
+			$request->setDimensionFilterClauses( $dimension_filters_clauses );
```
aaemnnosttv, line 1427: This is still not quite right (no better place to leave this comment although this specific line _is_ correct).

The `$dimension_filter_clause` immediately above belongs in the preceding conditional where the filter is defined.

This conditional should only have the call to `$request->setDimensionFilterClauses()`.

aaemnnosttv (commented):
Almost there, just a few small but important details left to fix. I'll push a few commits to address these quickly.

aaemnnosttv:
@googlebot I consent.

aaemnnosttv (approved):
Super!

includes/Modules/Analytics.php
```diff
@@ -1368,6 +1393,15 @@ protected function create_analytics_site_data_request( array $args = array() ) {
 		$request->setIncludeEmptyRows( true );
 		$request->setViewId( $profile_id );
 
+		$dimension_filter_clauses = array();
+		if ( ! empty( $args['dimension_filters'] ) ) {
+			$dimension_filters       = $args['dimension_filters'];
+			$dimension_filter_clause = new Google_Service_AnalyticsReporting_DimensionFilterClause();
+			$dimension_filter_clause->setFilters( array( $dimension_filters ) );
```
felixarntz, line 1400: I believe this is incorrect? `$dimension_filters` should already be passed as the array required here, so it must not be wrapped again.

includes/Modules/Analytics.php
```diff
@@ -1368,6 +1393,15 @@ protected function create_analytics_site_data_request( array $args = array() ) {
...
+			$dimension_filter_clause->setOperator( 'AND' );
+			$dimension_filter_clauses[] = $dimension_filter_clause;
+		}
+
```
felixarntz, line 1404: Nitpick, but maybe move this a bit further down to be right above the `if ( ! empty( $args['page'] ) ) {` clause? This way all logic touching dimension filter clauses is located close together.

includes/Modules/Analytics.php
```diff
@@ -962,6 +962,28 @@ function ( $dimension_def ) {
 					}
 				}
 
+				$dimension_filters          = $data['dimensionFilters'];
+				$dimension_filter_instances = array();
+				if ( ! empty( $dimension_filters ) && is_array( $dimension_filters ) ) {
+					foreach ( $request_args['dimensions'] as $dimension ) {
```
felixarntz, line 968: This is not correct. `dimensionFilters` and `dimensions` should be completely decoupled, the arguments are not related, unless I'm missing something. To validate/normalize the `dimensionFilters` argument, you should only need to look at `dimensionFilters`, not `dimensions`.

* `dimensionFilters` ensures only results that satisfy a certain dimension criteria are included in the response.
* `dimensions` modifies the response format to group results by certain dimensions.

assets/js/modules/analytics/util/report-validation.js
```diff
@@ -57,3 +57,21 @@ export function isValidDimensions( dimensions ) {
...
+/**
+ * Verifies provided dimensionFilters to make sure they match allowed values found in dimensions.
+ *
+ * @since n.e.x.t
+ *
+ * @param {Object}   dimensionFilters The dimension filters to check.
+ * @param {Object[]} dimensions       The dimensions to check.
+ * @return {boolean} TRUE if dimension filters are valid, otherwise FALSE.
+ */
+export function isValidDimensionFilters( dimensionFilters, dimensions ) {
```
felixarntz, line 70: See my other comment, `dimensionFilters` and `dimensions` arguments should not have any relation. `dimensionFilters` are a standalone argument.

felixarntz (changes requested):
I believe there's something off in the implementation here. Why is `dimensionFilters` validation/normalization somehow based on the passed `dimensions`? These are separate arguments with entirely different purposes. Looks like I went over the IB too quickly 🤦‍♂️ 

cc @aaemnnosttv @eugene-manuilov

ivankruchkoff:
Updating, per your comments @felixarntz. The IB had:
```
 It should check whether dimensionFilters is a map of dimension names and values and all dimension names are also defined in the dimensions option.
```

includes/Modules/Analytics.php
ivankruchkoff: Updated.

includes/Modules/Analytics.php
ivankruchkoff: Updated.

includes/Modules/Analytics.php
ivankruchkoff: Updated.

assets/js/modules/analytics/util/report-validation.js
ivankruchkoff: Updated.

assets/js/modules/analytics/util/report-normalization.js
```diff
@@ -51,3 +53,26 @@ const normalizeMetrics = ( metrics ) => {
...
+const normalizeDimensions = ( dimensions ) => {
+	return castArray( dimensions )
+		.map(
+			( dimension ) => typeof dimension === 'string'
+				? { name: dimension }
+				: dimension
+		)
+		.filter( ( dimension ) => isPlainObject( dimension ) )
+	;
+};
```
felixarntz, line 66: The normalization here makes sense, but this causes a bug visible on the dashboard, in `DashboardPopularPagesWidget`, where (for whatever odd reason 😆 ) we're passing multiple values as a comma-separated string. I'll update that component to pass the values as array of strings which is more appropriate.

assets/js/modules/analytics/util/report-normalization.js
```diff
@@ -51,3 +53,26 @@ const normalizeMetrics = ( metrics ) => {
...
+	const normalizedFilters = {};
+	if ( isPlainObject( dimensionFilters ) ) {
+		for ( const [ key, value ] of Object.entries( dimensionFilters ) ) {
+			if ( typeof value === 'string' ) {
+				normalizedFilters[ key ] = value;
+			}
+		}
+	}
+	return normalizedFilters;
+};
```
felixarntz, line 78: I don't think we need a normalization function for dimension filters. The reason they exist for dimensions and metrics is that those arguments support multiple different ways of being specified (e.g. string, array of strings, array of objects). For dimension filters though, for now we only support a map of scalar values - so let's remove `normalizeDimensionFilters`. We only need validation here.

assets/js/modules/analytics/util/report-validation.js
```diff
@@ -57,3 +57,18 @@ export function isValidDimensions( dimensions ) {
...
+ * @since n.e.x.t
+ *
+ * @param {Object} dimensionFilters The dimension filters to check.
+ * @return {boolean} TRUE if dimension filters are valid, otherwise FALSE.
+ */
+export function isValidDimensionFilters( dimensionFilters ) {
+	// Ensure every dimensionFilter key corresponds to a valid dimension.
+	return Object.keys( dimensionFilters ).every(
+		( dimension ) => typeof dimensionFilters[ dimension ] === 'string'
+	);
```
felixarntz, line 73: We shouldn't require dimension values to be a string. They just need to be scalar values (probably we could check whether it's either a string or a number). Something more important to cover in the validation here though is to ensure that a map of `dimensionName => dimensionValue` is passed. The keys here actually need to be strings. For example someone shouldn't be able to pass an array of values.

assets/js/modules/analytics/datastore/report.js
```diff
@@ -57,20 +57,23 @@ const fetchGetReportStore = createFetchStore( {
...
-		if ( dimensions ) {
-			invariant(
-				isValidDimensions( dimensions ),
-				'Dimensions for an Analytics report must be either a string, an array of strings, an object, an array of objects or a mix of strings and objects. If an object is used, it must have "name" property.',
-			);
-		}
+		invariant(
+			isValidDimensions( dimensions ),
+			'Dimensions for an Analytics report must be either a string, an array of strings, an object, an array of objects or a mix of strings and objects. If an object is used, it must have "name" property.',
+		);
```
felixarntz, line 71: Why did you remove the if clause here? I think that's an unrelated change.

assets/js/modules/analytics/datastore/report.js
```diff
@@ -57,20 +57,23 @@ const fetchGetReportStore = createFetchStore( {
...
+
+		invariant(
+			isValidDimensionFilters( dimensionFilters ),
+			'Dimension filters must be an object where the values are strings.',
```
felixarntz, line 75: See above comment, let's update the messaging here to be more appropriate, e.g.
```suggestion
			'Dimension filters must be a map of dimension names as keys and dimension values as values.',
```

felixarntz (changes requested):
Mostly LGTM, a few minor follow-up things to address.

assets/js/modules/analytics/datastore/report.js
ivankruchkoff: Yes, unrelated, thanks for noticing it.

assets/js/modules/analytics/util/report-normalization.js
ivankruchkoff: We initially had the normalisation to pass dimensions to the `isValidDimensionFilters`, which we don't need now, but it's a good idea to keep it as it makes it cleaner to have it normalised.

assets/js/modules/analytics/util/report-validation.js
ivankruchkoff: I figured dimension values would be string due to https://developers.google.com/analytics/devguides/reporting/core/v4/rest/v4/reports/batchGet#DimensionFilter ( expressions[] are 'string' )

ivankruchkoff: I've added a check for both string and number per your comment, and checking that keys are string.

felixarntz (approved):
Awesome!

PR #2489, opened 2020-12-07
Switch `core/user/data/user-input-settings` REST endpoint to use remote API.

benbowler:
Switch `core/user/data/user-input-settings` REST endpoint to use remote API. 

## Summary

Addresses issue #

## Checklist


includes/Core/Util/User_Input_Settings.php
```diff
@@ -126,19 +126,27 @@ function( $setting ) {
...
-		$user_input_settings_url  = $this->authentication->get_google_proxy()->url( Google_Proxy::USER_INPUT_SETTINGS_URI );
+		$user_input_settings_url = $this->authentication->get_google_proxy()->url( Google_Proxy::USER_INPUT_SETTINGS_URI );
+		$creds                   = $this->authentication->credentials()->get();
+
 		$user_input_settings_args = array(
 			'headers' => array(
 				'Authorization' => 'Bearer ' . $this->authentication->get_oauth_client()->get_access_token(),
+				'Content-Type'  => 'application/json',
+			),
+			'body'    => array(
```
adamsilverstein, line 137: The body should be JSON encoded here. Eg. `'body'    => wp_json_encode( array(`

includes/Core/Util/User_Input_Settings.php
```diff
@@ -126,19 +126,27 @@ function( $setting ) {
...
 
 		if ( ! empty( $settings ) ) {
-			$user_input_settings_args['headers']['Content-Type'] = 'application/json';
-			$user_input_settings_args['body']                    = wp_json_encode( $settings );
+			$user_input_settings_args['body']['settings']       = $settings;
+			$user_input_settings_args['body']['client_user_id'] = get_current_user_id();
 		}
 
-		$response = wp_remote_post( $user_input_settings_url, $user_input_settings_args );
+		$response = wp_remote_post( $user_input_settings_url, wp_json_encode( $user_input_settings_args ) );
```
adamsilverstein, line 148: The args should not be json encoded here, only the body.

```suggestion
		$response = wp_remote_post( $user_input_settings_url, $user_input_settings_args );
```

includes/Core/Util/User_Input_Settings.php
```diff
@@ -126,19 +126,27 @@ function( $setting ) {
...
+				'site_secret' => $creds['oauth2_client_secret'],
 			),
 		);
 
 		if ( ! empty( $settings ) ) {
-			$user_input_settings_args['headers']['Content-Type'] = 'application/json';
-			$user_input_settings_args['body']                    = wp_json_encode( $settings );
+			$user_input_settings_args['body']['settings']       = $settings;
+			$user_input_settings_args['body']['client_user_id'] = get_current_user_id();
 		}
```
adamsilverstein, line 147: before sending the request, the body should be JSON encoded. ie add:
`$user_input_settings_args['body'] = wp_json_encode( $user_input_settings_args['body'] );`

includes/Core/Util/User_Input_Settings.php
```diff
@@ -126,19 +126,27 @@ function( $setting ) {
...
+				'site_id'     => $creds['oauth2_client_id'],
+				'site_secret' => $creds['oauth2_client_secret'],
 			),
 		);
 
 		if ( ! empty( $settings ) ) {
-			$user_input_settings_args['headers']['Content-Type'] = 'application/json';
-			$user_input_settings_args['body']                    = wp_json_encode( $settings );
+			$user_input_settings_args['body']['settings']       = $settings;
+			$user_input_settings_args['body']['client_user_id'] = get_current_user_id();
```
adamsilverstein, line 145: The settings endpoint currently expects `client_user_id` as a string (because we plan to support non-WordPress systems that may not use an integer user id). To make this work, cast the ID as a string. cc: @felixarntz please confirm that sounds right? The type sent must match what the endpoint expects or we get a service side "unmarshaling" error.
```suggestion
			$user_input_settings_args['body']['client_user_id'] = (string) get_current_user_id();
```

adamsilverstein:
Nice work @benbowler - this is progressing well! Here are some testing results...

I did some testing on this branch and noticed a couple of tiny code issues and added comments inline. Once these were fixed locally, I was able to get thru the setup flow and save the settings. Success!

image

Next, I reset Site Kit and started over. When I got to the settings stage, the settings I had saved earlier were all set correctly. 

**Question**: after a reset, should all settings be reset? If not, should I go thru the settings flow again?
  
* I don't see settings under Site Kit -> Settings, I assume that part isn't complete yet?

I also tested the user/site settings part by creating multiple admin users and running thru the setup with each of them. When I chose a new Google user id to log into during the setup flow for the new user, I noted that the user questions (1 & 2) were not pre-filled, while the site questions (3, 4 & 5) were pre-filled. The page also showed a note indicating another user, by their wp email:

image

I feel like **this should use `user_nicename`** because users are more likely to know other user's wp names than their emails.

Next I logged in as a separate WordPress user, but with an existing Google ID already used by another account on the site. In this case, the settings for the user part was pre-filled which was a little confusing. This is because the actual user settings storage is currently **based on the users Google ID**. So the API responds with the settings the user stored, including that user's ID, and the UI displays those settings:

image

This is currently expected behavior because storage is keyed to the Google user ID not the WordPress user ID, however as a user I found this confusing - i would expect to be able to set my own (wp) user settings. I would suggest one of two possible solutions:

1. Add the `client_user_id` to the storage key service side, so two WordPress users using the same Google ID would result in distinct keys. 
2. Keep things the way they are service side - instead improve messaging/help users understand. eg something like "These settings were previously set by [WordPress Nicename] and are tied to your account [Google account]."

Option 1 feels like a better solution and concerns around spoofing the `client_user_id` should be ameliorated by the user's Google ID requirement. What do you think @felixarntz?

includes/Core/Util/User_Input_Settings.php
felixarntz: Ah that's correct, good catch. Did you try leaving out the call to `wp_json_encode` entirely? I'm wondering whether the WP request mechanism would automatically JSON-encode the body with `Content-Type` set to `application/json`. Probably not, but worth trying.

includes/Core/Util/User_Input_Settings.php
felixarntz: Yes, that's correct. I missed that before because I was thinking of URL-encoded parameters where it would become a string anyways. But yeah I guess we need to cast this to a string here, although it's a bit unintuitive. Let's go with that for now, maybe in the future we can make the service more resilient to support an integer and cast it to a string on its own.

felixarntz:
@adamsilverstein Some responses:
> after a reset, should all settings be reset? If not, should I go thru the settings flow again?

No, these should persist regardless of reset. That is the main reason we store them on the service - they should not be coupled to the state of the settings on the WordPress site.

> I feel like this should use **`user_nicename`** because users are more likely to know other user's wp names than their emails.

That might make sense, but maybe then the user name would be even better (because `user_nicename` may not be unique). Regardless, I think that's out of scope of this issue, we can revisit once we do a final pass once the entire implementation is complete. Let's not change it here.

> This is currently expected behavior because storage is keyed to the Google user ID not the WordPress user ID, however as a user I found this confusing - i would expect to be able to set my own (wp) user settings.

Let's discuss whether we should change it on our side. It's not critical IMO because it's not the most common scenario, although of course it _does_ exist. I think we should either fix on our side or leave it as is entirely potentially. I think the plugin behavior shouldn't be affected.

@benbowler Can you address @adamsilverstein's other code review feedback from above? You should also be able to test it then against the staging version of the authentication service.

includes/Core/Util/User_Input_Settings.php
adamsilverstein: Without json encoding the body, I got an "unmarshaling" error. I don't think we have encoding for body in core's `WP_Http` class [request](https://github.com/WordPress/wordpress-develop/blob/8eda4984e1d409a97b260c52b3485defccae495d/src/wp-includes/class-http.php#L149)

includes/Core/Util/User_Input_Settings.php
```diff
@@ -126,19 +126,27 @@ function( $setting ) {
...
 		}
 
-		$response = wp_remote_post( $user_input_settings_url, $user_input_settings_args );
+		$response = wp_remote_post( $user_input_settings_url, wp_json_encode( $user_input_settings_args ) );
+
 		if ( is_wp_error( $response ) ) {
 			return $response;
 		}
```
felixarntz, line 152: I cannot mark the code I actually intend to here since you actually didn't touch it, but it's worth improving because right now the error messaging here is not great because it doesn't provide any details that may be present in the API response.
1. Can you move the response body decoding (lines 163-164) above the check for the response code (lines 154-155)?
2. Then, in that clause, add another nested `if ( is_array( $settings ) && ! empty( $settings['error'] ) )`. If that is set (it's an error code), return a new `WP_Error` where you use that value as error code and something like `sprintf( __( 'User input settings request failed with error: %s', 'google-site-kit' ), $settings['error'] )` as message.

This way you'll be able to tell from the client side _what_ actually went wrong in most cases.

Let's keep the currently used unprecise `WP_Error` only for when we cannot get more details from the API response like I described above.

felixarntz (changes requested):

felixarntz:
@benbowler You can now test this PR against the staging version of the authentication service. Before doing that though, please address all the review comments from @adamsilverstein and myself as they'll already fix a couple things that we ran into with testing.

includes/Core/Util/User_Input_Settings.php
benbowler: Thanks for these comments adam, I've worked down each below.

includes/Core/Util/User_Input_Settings.php
benbowler: Updated and I confirm this works correctly.

includes/Core/Util/User_Input_Settings.php
benbowler: This caused me to get a `{"error":"missing authentication token"}` error repeatedly in testing. I now have it working.

includes/Core/Util/User_Input_Settings.php
benbowler: Added.

includes/Core/Util/User_Input_Settings.php
benbowler: Hey Felix, I've modified this based on your suggestions to [surface errors from the API more clearly](https://github.com/google/site-kit-wp/pull/2489/files#diff-f2d89ed2b380a2e93a3a19fd41ec17471f836fcf9b2ffc8c0bc44ae06ab672ddR162-R169).

benbowler:
Thanks for your detailed testing and feedback @adamsilverstein + @felixarntz, I've completed the changes and can verify the user input survey now runs and saves for me using the staging endpoint.

includes/Core/Util/User_Input_Settings.php
felixarntz: This looks solid, the only thing worth improving would be to use the `error` code as error code for the `WP_Error` as well, as that provides more insight than a generic `user_input_settings_request`.

felixarntz (approved):
Awesome!

Issue #2272, opened 2020-10-26
Document functions on public googlesitekit.modules object


felixarntz:
The functions on the public API object `googlesitekit.modules` should be documented.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All functions on the public API object `googlesitekit.modules` should be documented:
    * `activateModule`
    * `deactivateModule`
    * `registerModule`

## Implementation Brief

	/**
	 * Registers a module.
	 *
	 * @since 1.13.0
	 * @since 1.20.0  Introduced the ability to register settings and setup components.
	 * @since n.e.x.t Introduced the ability to add a checkRequirements function.
	 *
	 * @param {string}      slug                             Module slug.
	 * @param {Object}      [settings]                       Optional. Module settings.
	 * @param {string}      [settings.name]                  Optional. Module name. Default is the slug.
	 * @param {string}      [settings.description]           Optional. Module description. Default empty string.
	 * @param {string}      [settings.icon]                  Optional. Module icon. Default empty string.
	 * @param {number}      [settings.order]                 Optional. Numeric indicator for module order. Default 10.
	 * @param {string}      [settings.homepage]              Optional. Module homepage URL. Default empty string.
	 * @param {WPComponent} [settings.settingsEditComponent] Optional. React component to render the settings edit panel. Default none.
	 * @param {WPComponent} [settings.settingsViewComponent] Optional. React component to render the settings view panel. Default none.
	 * @param {WPComponent} [settings.setupComponent]        Optional. React component to render the setup panel. Default none.
	 * @param {Function}    [settings.checkRequirements]     Optional. Function to check requirements for the module. Throws a WP error object for error or returns on success.
	 */

* Edit the `assets/js/googlesitekit/modules/index.js` adding a new docblock comment to each of these three functions:
    * `activateModule`
      * Add a title and single-line description of this function to add a module to the store.
      * Add `@since`, `@param` and `@return` descriptions for each based on the [docblock of the internal function](https://github.com/google/site-kit-wp/blob/c279f1b156952de53aafbc0961ffbc0e502fa1e6/assets/js/googlesitekit/modules/datastore/modules.js#L143-L154).
    * `deactivateModule`
      * Add a title and single-line description of this function to deactivate a module in the store.
[...]

Issue #2413, opened 2020-11-23
Rename `WPComponent` arguments to use capitalized names


felixarntz:
Any arguments or props that require a `WPComponent` / `ComponentType` should be updated to have a capitalized name, to indicate the required argument type better (e.g. `SettingsEditComponent` instead of `settingsEditComponent`). This will eliminate the need to rename arguments before rendering them.

See https://github.com/google/site-kit-wp/issues/2381#issuecomment-730196523 for context.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All arguments accepting `WPComponent`s should receive a capitalized name instead of starting with a lowercase letter.
    * In `registerModule` of `core/modules` store.
    * In `registerWidget` of `core/widgets` store.
* The documentation of `core/widgets` should be updated to reference `WPComponent` as required type instead of `React.Component`.
* The `registerWidget` function exposed on `googlesitekit.widgets` should have its documentation updated accordingly.
* All usage of these arguments should be updated to use the new names, i.e. renaming them to render them will no longer be necessary (since they'll already be capitalized).

## Implementation Brief

* Rename all the arguments of type `WPComponent` to be capitalized, with updated documentation in the following:
	* `registerModule` in `assets/js/googlesitekit/modules/datastore/modules.js`
	* `registerWidget` in `assets/js/googlesitekit/widgets/datastore/widgets.js`
* Use `WPComponent` instead of `React.Component` in `assets/js/googlesitekit/widgets/index.js`'s `registerWidget`. Also update the renamed arguments from the point above.
* Search for the usage of the arguments and rename them accordingly. Also remove any renaming done(capitalizing name when using) in the usage. 

	Example:
	In `assets/js/components/setup/setup-wrapper.js`, rename the usage to `SetupComponent` also update the destructure statement like
	```diff
	- const { setupComponent: SetupComponent } = module;
	+ const { SetupComponent } = module;
	```

### Test Coverage

* No additional tests required. Update tests if and when required to accommodate renamed arguments.

### Visual Regression Changes

* No visual changes.

## QA Brief

* The site kit module setup must look and work exactly as before
* Widgets must look the same. The stories for widgets must look the same as well.

## Changelog entry

* Rename arguments across various datastore actions requiring a `WPComponent` type to use capitalized names to better indicate a `WPComponent` (and not a `WPElement`) is required.

felixarntz:
IB ✅

tofumatt:
[...]

PR #2493, opened 2020-12-07
Document functions on public googlesitekit.modules object.

benbowler:
## Summary

Document functions on public googlesitekit.modules object.
Addresses issue #2272

## Relevant technical choices

N/A

## Checklist


tofumatt (dismissed):
Looks good to me, thanks!

assets/js/googlesitekit/modules/index.js
```diff
@@ -30,12 +30,59 @@ import './datastore';
...
+	 * @param {string}      [args.settings.description]           Optional. Module description. Default empty string.
+	 * @param {string}      [args.settings.icon]                  Optional. Module icon. Default empty string.
+	 * @param {number}      [args.settings.order]                 Optional. Numeric indicator for module order. Default 10.
+	 * @param {string}      [args.settings.homepage]              Optional. Module homepage URL. Default empty string.
+	 * @param {WPComponent} [args.settings.settingsEditComponent] Optional. React component to render the settings edit panel. Default none.
+	 * @param {WPComponent} [args.settings.settingsViewComponent] Optional. React component to render the settings view panel. Default none.
+	 * @param {WPComponent} [args.settings.setupComponent]        Optional. React component to render the setup panel. Default none.
+	 * @param {Function}    [args.settings.checkRequirements]     Optional. Function to check requirements for the module. Throws a WP error object for error or returns on success.
+	 * @return {Function}   Generator Function with updated instance of the module registry.
```
aaemnnosttv, line 84: In contrast with the underlying action creator, the return value of a dispatched action (bound action creator) is always a `Promise`.
See https://github.com/WordPress/gutenberg/tree/master/packages/data#dispatch

assets/js/googlesitekit/modules/index.js
```diff
@@ -30,12 +30,59 @@ import './datastore';
...
+	/**
+	 * Activates Module
+	 *
+	 * Activates a module in the modules store based on the provided args.slug parameter.
+	 *
+	 * @since 1.8.0
+	 *
+	 * @param {Object} [args]      Module arguments.
+	 * @param {string} [args.slug] Slug of the module to activate.
+	 * @return {Object} Object with `{response, error}`. On success, `response.moduleReauthURL`
```
aaemnnosttv, line 42: See previous comment. This will always return a `Promise` – the object is the value that the promise will be settled with.

assets/js/googlesitekit/modules/index.js
```diff
@@ -30,12 +30,59 @@ import './datastore';
...
+	/**
+	 * Deactivates Module
+	 *
+	 * Deactivates a module in the modules store based on the provided args.slug parameter.
+	 *
+	 * @since 1.8.0
+	 *
+	 * @param {Object} [args]      Module arguments.
+	 * @param {string} [args.slug] Slug of the module to activate.
+	 * @return {Object} Object with `{response, error}`.
```
aaemnnosttv, line 58: Same here 😄

assets/js/googlesitekit/modules/index.js
```diff
@@ -30,12 +30,59 @@ import './datastore';
 
 const Modules = {
 	createModuleStore,
+	/**
+	 * Activates Module
+	 *
+	 * Activates a module in the modules store based on the provided args.slug parameter.
```
aaemnnosttv, line 36: Let's use the same short and long descriptions as the internal functions.

assets/js/googlesitekit/modules/index.js
```diff
@@ -30,12 +30,59 @@ import './datastore';
...
+	 *
+	 * @since 1.8.0
+	 *
+	 * @param {Object} [args]      Module arguments.
+	 * @param {string} [args.slug] Slug of the module to activate.
```
aaemnnosttv, line 41: Even though the function passes through all arguments to the action creator, we should document the arguments as if they were passed individually (this should arguably be changed in the function signature as well but that's outside of the scope here).

```suggestion
	 * @param {string} slug Slug of the module to activate.
```

Also, `slug` is not an optional parameter to the action but it is documented as an optional argument here using the bracket syntax.

Please do the same for the others here as well, (documenting individually – no `args.` prefix – and matching optional – only `registerModule` has optional `settings`)

aaemnnosttv (changes requested):
Thanks @benbowler – just a few changes to make here and then this should be good to go 👍

assets/js/googlesitekit/modules/index.js
benbowler: I had to pass args.slug to satisfy ESLint.

benbowler: Otherwise updated.

benbowler:
Updated based on your feedback @aaemnnosttv

assets/js/googlesitekit/modules/index.js
```diff
@@ -30,12 +30,54 @@ import './datastore';
...
 	createModuleStore,
+	/**
+	 * Activates a module on the server.
+	 *
+	 * Activate a module (based on the slug provided).
+	 *
+	 * @since 1.8.0
+	 *
+	 * @param {any}    args      Module action arguments.
+	 * @param {string} args.slug Slug of the module to activate.
```
aaemnnosttv, line 41: @felixarntz we should probably change the signature of the functions here to match the underlying action creator rather than spreading the args through which is forcing us to document them differently.
What do you think?

assets/js/googlesitekit/modules/index.js
```diff
@@ -30,12 +30,54 @@ import './datastore';
...
+	 * @param {Object}      args                         Module action arguments.
+	 * @param {string}      args.slug                    Module slug.
+	 * @param {string}      [args.name]                  Optional. Module name. Default is the slug.
+	 * @param {string}      [args.description]           Optional. Module description. Default empty string.
+	 * @param {string}      [args.icon]                  Optional. Module icon. Default empty string.
+	 * @param {number}      [args.order]                 Optional. Numeric indicator for module order. Default 10.
+	 * @param {string}      [args.homepage]              Optional. Module homepage URL. Default empty string.
+	 * @param {WPComponent} [args.settingsEditComponent] Optional. React component to render the settings edit panel. Default none.
+	 * @param {WPComponent} [args.settingsViewComponent] Optional. React component to render the settings view panel. Default none.
+	 * @param {WPComponent} [args.setupComponent]        Optional. React component to render the setup panel. Default none.
```
aaemnnosttv, line 77: Sorry @benbowler but it looks like these were recently updated after #2413 being merged 😄 
https://github.com/google/site-kit-wp/blob/80c932da36bc35fae7637015e4ba31099979d543/assets/js/googlesitekit/modules/datastore/modules.js#L230-L235

Also, note that these are named `settings` in the action docs. I understand the way the args are named below forces us to use the same name here but it's actually kind of cheating – the real `args` below is an _array_ of all arguments passed to the function but we're documenting it above as an object. I would suggest we replace the `args` spreading with the real args instead but let's wait and see what @felixarntz says about that.

assets/js/googlesitekit/modules/index.js
felixarntz: I agree, let's match the function signatures. Spreading them through is rather detrimental here.

assets/js/googlesitekit/modules/index.js
```diff
@@ -30,12 +30,54 @@ import './datastore';
...
+	 *
+	 * Activate a module (based on the slug provided).
+	 *
+	 * @since 1.8.0
+	 *
+	 * @param {any}    args      Module action arguments.
+	 * @param {string} args.slug Slug of the module to activate.
+	 * @return {Promise} The dispatched action.
+	 */
 	activateModule: ( ...args ) => {
```
felixarntz, line 44: See above comment, let's change this to `( slug )` and use the same documentation as for the store action parameter.

assets/js/googlesitekit/modules/index.js
```diff
@@ -30,12 +30,54 @@ import './datastore';
...
+	 *
+	 * Deactivate a module (based on the slug provided).
+	 *
+	 * @since 1.8.0
+	 *
+	 * @param {any}    args      Module action arguments.
+	 * @param {string} args.slug Slug of the module to activate.
+	 * @return {Promise} The dispatched action.
+	 */
 	deactivateModule: ( ...args ) => {
```
felixarntz, line 58: See above, let's change this to `( slug )`.

assets/js/googlesitekit/modules/index.js
```diff
@@ -30,12 +30,54 @@ import './datastore';
...
+	 * @param {string}      [args.icon]                  Optional. Module icon. Default empty string.
+	 * @param {number}      [args.order]                 Optional. Numeric indicator for module order. Default 10.
+	 * @param {string}      [args.homepage]              Optional. Module homepage URL. Default empty string.
+	 * @param {WPComponent} [args.settingsEditComponent] Optional. React component to render the settings edit panel. Default none.
+	 * @param {WPComponent} [args.settingsViewComponent] Optional. React component to render the settings view panel. Default none.
+	 * @param {WPComponent} [args.setupComponent]        Optional. React component to render the setup panel. Default none.
+	 * @param {Function}    [args.checkRequirements]     Optional. Function to check requirements for the module. Throws a WP error object for error or returns on success.
+	 * @return {Promise} The dispatched action.
+	 */
 	registerModule: ( ...args ) => {
```
felixarntz, line 81: See above, let's change this to `( slug, settings )` and copy over the parameter docs from the action.

felixarntz (changes requested):
Sorry for the delay review here, but I agree with @aaemnnosttv's comments that we should bring the function signatures here in line with the underlying action creators. This way we can copy-paste most of the documentation and don't have to rewrite it.

assets/js/googlesitekit/modules/index.js
benbowler: Updated, we'll need to make sure these keep in sync when the actions are updated in future.

assets/js/googlesitekit/modules/index.js
benbowler: Great, matching the function signatures makes is much clearer and more simple for me.

tofumatt (approved):
Looks good to me.

felixarntz (approved):
Great!

Issue #2498, opened 2020-12-07
Ship flag mode file with Site Kit development build ZIP


felixarntz:
## Bug Description

The `dist/config.json` file determines whether the current build is a production or development build, via #2316. Intentionally, that file is not included in the production ZIP bundle because for that ZIP file we can through that avoid additional file reading and JSON parsing logic to run on every pageload. However, we didn't consider development ZIP bundles in that thought: Those ZIP files are currently missing the file too, but in there it's necessary for PHP to know that this is a development build.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `release` script should be updated: If the current build is not a production build, it should also copy the `dist/config.json` file so that it is included in the bundle ZIP.

## Implementation Brief

* Read the `flagMode` from the `dist/config.json` in the release script
* If the `flagMode !== 'production'`, include the `config.json` in the zip, at the same path
* Merge https://github.com/google/site-kit-wp/pull/2500

### Test Coverage

* None needed

### Visual Regression Changes

* N/A

## QA Brief

* Run `npm run release-zip` – inspect the resulting zip file to ensure `dist/config.json` is **not** present
* Run `npm run dev-zip` – inspect the resulting zip file to ensure `dist/config.json` **is present**

## Changelog entry

* N/A

felixarntz:
IB ✅

cole10up:
### QA ✅

Confirmed dist/config.json is **not present** in release-zip
![image](https://user-images.githubusercontent.com/57115907/101570543-80425400-398b-11eb-9293-db6c8045d463.png)

Confirmed dist/config.json is **present** in dev-zip
![image](https://user-images.githubusercontent.com/57115907/101570616-ac5dd500-398b-11eb-8ee6-3f5cb480ca4d.png)

Sending to approved

PR #2500, opened 2020-12-07
Include dist/config.json for non-production zips

aaemnnosttv:
## Summary

Addresses issue #2498 

## Checklist

### DIFF
--- a/bin/release
+++ b/bin/release
@@ -125,7 +125,7 @@ async function makeRelease() {
 		'production' !== flagMode ? cp( 'dist/config.json' ) : []
 	) );
 
-	// Archvie the release directory.
+	// Archive the release directory.
 	const archive = archiver( 'zip', { zlib: { level: 9 } } );
 	archive.pipe( fse.createWriteStream( filename ) );
 	archive.directory( pluginDir, 'google-site-kit' );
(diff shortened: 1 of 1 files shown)


bin/release
```diff
@@ -116,9 +121,11 @@ async function makeRelease() {
 		cp( 'dist/assets' ),
 		cp( 'includes' ),
 		cp( 'third-party' ),
-	] );
+	].concat(
+		'production' !== flagMode ? cp( 'dist/config.json' ) : []
+	) );
```
felixarntz, line 126: Is the empty array here correct? Maybe this should rather be `undefined`? If the `flagMode` isn't "production" then we don't need to concatenate anything.

Alternatively we could make this a bit more intuitive to read and put the array of promises into a variable, and then add the additional entry only if needed, rather than always concatenating (potentially with nothing).

felixarntz (commented):
One question on the implementation here.

bin/release
aaemnnosttv: Yes, if you concatenate anything other than an array, it appends that to the array, otherwise it joins the two arrays into a new one. We did the same thing before with some of the `Select` options I recall, but no problem to change. I'll update 👍 

image

felixarntz (approved):
Great!

Issue #2221, opened 2020-10-20
Add user locale query parameter in help links to `support.google.com`.


adamsilverstein:
## Feature Description

When linking to help articles on `support.google.com`, Site Kit should set the language/locale query parameter (`hl`) based on the user's WordPress profile language setting (`get_user_locale`).

The plugin links to help articles on [https://support.google.com/](https://support.google.com/) in numerous places:

![image](https://user-images.githubusercontent.com/2676022/96652022-49d54a80-12f3-11eb-9c8b-7b0a0f7996e4.png)

Some of these are hard coded to use the english version (`&hl=en`). Instead, all should use `hl=[USER_LOCALE]`.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All external help links to `support.google.com` should include the current user's locale as the `hl` (language) query parameter.
* Links to `support.google.com` in JS should be handled via a `getGoogleSupportURL` selector of the `core/site` store. It should receive an object with optional `path`, `query` and `hash` arguments. The selector should automatically add the current locale as `hl` query parameter to the URL returned.
* All usages of links pointing to `support.google.com` should use the `getGoogleSupportURL` selector.

## Implementation Brief

* Add `getGoogleSupportURL` to `assets/js/googlesitekit/datastore/site/info.js`.
	* It should receive an object with optional `path`, `query` and `hash` arguments.
	* Add `query` and build the `url` object with the `hl` parameter using `getLocale()` from `assets/js/util/i18n.js`.
	  ```js
	  const url = new URL( addQueryArgs( 'https://support.google.com', { ...query, hl: getLocale() } ) )
      ```
	* Add the hash `url.hash = hash || ''` 
	* Return `url.toString()`
* Use the added selector in the following files using the `useSelect` hook:
	* `SetupAccountCreate.js`
	* `SetupAccountNoClient.js`
	* `AdSenseLinkCTA.js`
	* `AnonymizeIPSwitch.js`
	* `DashboardGoalsWidget.js`
	* `LegacyAnalyticsDashboardWidgetTopLevel.js`
	* `OptimizeIDFieldInstructions.js`
	* `UseSnippetInstructions.js`
	* Anywhere else if missed here
* Remove the translation function call on the URL like the one present in `UseSnippetInstructions.js` . We don't need to translate the URL anymore since we are adding the `hl`(locale) parameter.

### Test Coverage

[...]

PR #2515, opened 2020-12-10
Introduce new getGoogleSupportURL selector for Google support URLs wh…

asvinb:
…ich include the user locale.

## Summary

Addresses issue #2221 

## Checklist


assets/js/googlesitekit/datastore/site/info.test.js
```diff
@@ -140,6 +140,84 @@ describe( 'core/site site info', () => {
 			} );
 		} );
 
+		describe( 'getGoogleSupportURL', () => {
+			it( 'returns null if no arguments are supplied', async () => {
+				await registry.dispatch( STORE_NAME ).receiveSiteInfo( { ...baseInfo, ...entityInfo } );
+
+				const supportURL = registry.select( STORE_NAME ).getGoogleSupportURL();
+				expect( supportURL ).toEqual( null );
+			} );
```
eugene-manuilov, line 149: @asvinb looks like these tests are identical and we can convert them into table tests. So, instead of having multiple tests, we need to convert it into one test with multiple incoming args:

```javascript
it.each( [
    [
        'returns null if no arguments are supplied',
        undefined,
        null,
    ],
    [
       'returns the path with the user locale',
        { path: 'https://support.google.com/analytics/answer/1032415' },
        'https://support.google.com/analytics/answer/1032415?hl=en-US',
    ],
    ...
] )( '%s', async ( _, in, out ) => {
    await registry.dispatch( STORE_NAME ).receiveSiteInfo( { ...baseInfo, ...entityInfo } );
    const supportURL = registry.select( STORE_NAME ).getGoogleSupportURL( in );
    expect( supportURL ).toEqual( out );
} );
```

assets/js/googlesitekit/datastore/site/info.js
```diff
@@ -469,6 +470,34 @@ export const selectors = {
...
+	 * @return {(string|null)} The URL containing the user's locale or `null` if path is not set.
+	 */
+	getGoogleSupportURL: ( state, args ) => {
+		if ( ! args ) {
+			return null;
+		}
+
+		const { path, query, hash } = args;
+
+		if ( ! path ) {
```
eugene-manuilov, line 492:
```suggestion
		const { path, query, hash } = args || {};
		if ( ! path ) {
```

assets/js/modules/adsense/components/setup/SetupAccountCreate.js
```diff
@@ -31,19 +31,26 @@ import Link from '../../../../components/Link';
...
+	const userEmail = useSelect( ( select ) => select( CORE_USER ).getEmail() );
 	const existingTag = useSelect( ( select ) => select( STORE_NAME ).getExistingTag() );
 	const signUpURL = useSelect( ( select ) => select( STORE_NAME ).getServiceCreateAccountURL() );
+	const supportURL = useSelect(
+		( select ) => select( CORE_SITE ).getGoogleSupportURL(
+			{
+				path: 'https://support.google.com/adsense/answer/2659101',
+			}
+		)
+	);
```
eugene-manuilov, line 53: @asvinb let's use the compact version for this hook. The same applies to other occurrences as well.

```suggestion
	const supportURL = useSelect( ( select ) => select( CORE_SITE ).getGoogleSupportURL( {
		path: 'https://support.google.com/adsense/answer/2659101',
	} ) );
```

eugene-manuilov (changes requested):
Thanks, @asvinb! Let's make a few changes and we should be good to go.

assets/js/googlesitekit/datastore/site/info.test.js
asvinb: Thanks for the tip @eugene-manuilov . Updated!

assets/js/googlesitekit/datastore/site/info.js
asvinb: Updated @eugene-manuilov

assets/js/modules/adsense/components/setup/SetupAccountCreate.js
asvinb: Updated @eugene-manuilov

eugene-manuilov (dismissed):
LGTM! Thanks, @asvinb!

assets/js/googlesitekit/datastore/site/info.js
```diff
@@ -469,6 +470,30 @@ export const selectors = {
...
+	 * @return {(string|null)} The URL containing the user's locale or `null` if path is not set.
+	 */
+	getGoogleSupportURL: ( state, args ) => {
+		const { path, query, hash } = args || {};
+
+		if ( ! path ) {
+			return null;
+		}
+
+		const url = new URL( addQueryArgs( path, { ...query, hl: getLocale() } ) );
```
felixarntz, line 492: I didn't see the exact IB for this one before (which this matches), but I was thinking the `path` should be an actual path, not a full URL. The `getGoogleSupportURL` selector should always return URLs starting in `https://support.google.com`, and only the other parts (`path`, `query`, and `hash`) would be passed to the selector.

It's confusing to provide a full URL as an argument called `path` IMO.

assets/js/googlesitekit/datastore/site/info.test.js
```diff
@@ -140,6 +140,79 @@ describe( 'core/site site info', () => {
...
+					'returns null if no path is supplied or is empty',
+					{
+						path: '',
+					},
+					null,
+				],
+				[
+					'returns the path, hash and the user locale',
+					{
+						path: 'https://support.google.com/analytics/answer/1032415',
```
felixarntz, line 160: See above, this for example should pass only `/analytics/answer/1032415` instead.

felixarntz (changes requested):
Against the IB this is good, but I think it's odd to pass a full URL for an argument called `path`. All support URLs start with `https://support.google.com`, and that should be part of the selector - this might have been not specific enough in the ACs, but was the intention here.

assets/js/googlesitekit/datastore/site/info.js
asvinb: @felixarntz Code updated. Can you check?

assets/js/googlesitekit/datastore/site/info.test.js
asvinb: Updated @felixarntz

felixarntz (approved):
Great!

Issue #475, opened 2019-08-29
Refactor tag implementations with polymorphism


aaemnnosttv:
## Feature Description

Currently there is a fair amount of code that deals with different forms of tags and their implementations (ie. standard web / amp). This leads to many conditionals and cognitive overhead when working with modules that leverage tags, as well as difficulty testing.

This could be simplified by introducing a Tag interface with dedicated implementations for each case, which would encapsulate their implementation details. Then the consuming module would only need to perform the logic to choose the necessary implementation once, and the rest of the code need not be concerned about which specific implementation is being used.

The selection itself would likely be the same for all modules (amp vs non-amp) so this could likely live in a trait, such as `Module_With_Tags`

This trait might have methods like `get_amp_tag()`, `get_web_tag()`, and `get_tag()`

<details>
    <summary>A basic outline might look something like this</summary>
    
```php
trait Module_With_Tags {
    public function get_tag() : Tag_Interface {
        if ( $this->get_tag_context()->is_amp() ) {
            return $this->get_amp_tag();
        }

        return $this->get_web_tag();
    }
    abstract protected function get_tag_context() : Context;
    abstract protected function get_amp_tag() : Tag_Interface;
    abstract protected function get_web_tag() : Tag_Interface;
}

interface Tag_Interface {
    public function register();
    public function render();
}
```

> Return types used for brevity.

Then each module can decide which implementation to use in its respective methods. Any case where there is no implementation for a given tag could simply return a `Null_Tag` as a no-op.

</details>

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be an interface `Tag_Interface` with public methods `can_register()` and `register()`.
* There should be an interface `Blockable_Tag_Interface` with public methods `is_tag_blocked()` and `get_tag_blocked_on_consent_attribute()`.
* There should be a `Module_With_Tags` interface with a single public `get_tag()` method.
* There should be a `Module_With_Tags_Trait`:
    * protected methods `get_web_tag()` and `get_amp_tag()`
    * public method `get_tag()`: It should rely on `$this->context` (which because this is a trait for modules we can assume is there) and based on that call `get_web_tag` or `get_amp_tag`.
* There should be the following new classes implementing `Tag_Interface` and `Blockable_Tag_Interface`, as well as using one of the above two traits, each of which include the registration logic that is currently within the respective class's `template_redirect` hook, for the respective AMP/non-AMP case:
[...]

PR #2517, opened 2020-12-10
Enhancement/475 refactor tags

eugene-manuilov:
## Summary

Addresses issue #475

## Relevant technical choices

* The implementation is slightly different from what is requested in the IB but it has been discussed and agreed upon with Felix/Evan.
* Tests for Analytics, AdSense, and Tag Manager AMP/Web tags remain in the module classes.

## Checklist

## Simplified class diagram

tags


includes/Core/Tags/AMP_Tag.php
```diff
@@ -0,0 +1,118 @@
...
+namespace Google\Site_Kit\Core\Tags;
+
+/**
+ * Base class for AMP tag.
+ *
+ * @since n.e.x.t
+ * @access private
+ * @ignore
+ */
+abstract class AMP_Tag extends Tag implements Blockable_Tag_Interface {
```
felixarntz, line 20: Let's call these `Module_AMP_Tag` and `Module_Web_Tag` like in the IB and have in them within `Core\Modules` (or `Core\Modules\Tags`), as they are module-specific classes. In theory we could have module-agnostic `Tag` interfaces and classes, but it doesn't satisfy our use-cases and isn't necessary (for example we need to access `$module_slug` like you're doing here), so let's reflect that in their names.

includes/Core/Tags/Tag.php
```diff
@@ -0,0 +1,97 @@
...
+	 * Constructor.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $module_slug Module slug.
+	 * @param string $tag_id Tag ID.
+	 */
+	public function __construct( $module_slug, $tag_id ) {
+		$this->module_slug = $module_slug;
+		$this->tag_id      = $tag_id;
```
felixarntz, line 58: I'm wary about this abstraction here. Since these classes are module-specific either way, I think it makes more sense to require a `Module_Settings` instance to be passed and handle the logic to get the ID from the settings in the `*_Tag` classes. There are other settings like `useSnippet` for example (which right now you're handling via the "guard" approach), but it could also be that a snippet requires two tags (e.g. we will need that for Google Analytics 4 support).

Handling the logic to get the respective settings and determine whether it should be rendered based on settings _inside_ of the `*_Tag` classes also removes more code from the parent class which I think makes sense. These module tag classes should encompass all logic related to rendering tags, which includes figuring out where the tag ID comes from etc. I think setting those things from the outside in the already bloated main module class makes this less intuitive to follow and maintain.

includes/Core/Tags/Tag_Interface.php
```diff
@@ -0,0 +1,29 @@
...
+ */
+interface Tag_Interface {
+
+	/**
+	 * Registers tag hooks.
+	 *
+	 * @since n.e.x.t
+	 */
+	public function register();
+
```
felixarntz, line 28: +1 on omitting `render` as public method here. We _should_ still call the main method for rendering each tag `render` for consistency, but it's true it doesn't need to be part of the interface, since the only public piece is `register` to add hooks.

includes/Core/Tags/Tag.php
```diff
@@ -0,0 +1,97 @@
...
+	 */
+	final public function register() {
+		foreach ( $this->guards as $guard ) {
+			if ( $guard instanceof Guard_Interface ) {
+				$can_activate = $guard->can_activate();
+				if ( is_wp_error( $can_activate ) || ! $can_activate ) {
+					return;
+				}
+			}
+		}
```
felixarntz, line 85: I'm not convinced of the "guard" approach in this PR. I think it would make sense if these were truly standalone/agnostic `Tag` classes, but all of these are more `Module_Tag` classes. We could potentially introduce an even lower level in the future, but I think that would be better as a follow-up PR.

I suggest to do the following:
* Stick with the IB for relying on `Module_Tag` classes that include their own logic based on `Module_Settings` instances for whether they should render or not etc.
* Add a follow-up issue for low-level `Tag` classes that are not associated with modules at all and could use more abstract logic like the `Guard` approach. The `Module_Tag` classes could then be migrated to use `Tag` classes internally.

felixarntz (commented):
@eugene-manuilov I didn't do a proper code review (since it's still a draft), but I left some feedback on the altering patterns here from the IB. Overall, while I'm in favor of SOLID principles, with Site Kit we're not really embracing these, and while some of it makes sense within this infrastructure too, I wanna be careful of moving too far into this direction. I'm all for abstracting reusable pieces of code at low levels, but I think we should continue to have higher-level layers which are easier to follow without going through multiple levels of abstraction.

Mainly, my feedback is: Let's try to go for `Module_Tag` interfaces and classes in this PR, and we can then go from there to create lower-level `Tag` interfaces and classes where IMO it would make more sense to use some of the ideas you're introducing here.

includes/Core/Tags/AMP_Tag.php
eugene-manuilov: Yep, updated.

tofumatt (approved):
I'll admit that I'm not as familiar with this area of the code, but the code here all makes sense to me and unifies the handling of the different tags.

It's still complex, but part of that is the nature of what we're dealing with. I think this implementation makes it easier to understand all tags by understanding one though, so that's great 👍🏻 

I'll leave it to Felix and Evan to review as well—I suspect they'll have stronger opinions about PHP approach—but I see this as an improvement.

includes/Modules/AdSense.php
```diff
@@ -93,88 +93,26 @@ function () {
...
 
-					add_filter( // Load amp-auto-ads component for AMP Reader.
-						'amp_post_template_data',
-						function( $data ) {
-							return $this->amp_data_load_auto_ads_component( $data );
-						}
-					);
+				if ( ! $tag->is_tag_blocked() ) {
+					$tag->use_guard( new TagVerifyGuard( $this->context->input() ) );
+					$tag->use_guard( new Tag_Guard( $module_settings ) );
```
felixarntz, line 111: Instead of instantiating the `*_Tag` instances here throughout the three module classes, please follow the ACs here, implementing a `Module_With_Tags` interface that would return the complete instance and allows us to remove all the `template_redirect` hooks entirely, replaced with one central piece of invocation. This is also specified in more details in the IB.

While we diverged from the original ACs/IB a bit, I don't think we discussed changing that part, and I think that should be kept as requirement as it simplifies the module-specific implementation. A module would only need to declare itself as `Module_With_Tags` and could then implement the necessary methods.

felixarntz (changes requested):
Mostly LGTM, however there's one thing that wasn't discussed to change where we should stick to the original ACs instead.

eugene-manuilov:
@felixarntz I have updated Analytics, AdSense, and Tag Manager modules to have dedicated methods for tags setup instead of using anonymous functions. Please, let me know if it makes sense to you or if you want to have a short call with me and Evan to discuss it.

felixarntz (approved):
@eugene-manuilov The current approach looks solid to me - let's get this merged!

Issue #2405, opened 2020-11-19
Update Analytics report deep links to use `getServiceReportURL` selector


felixarntz:
After #2228, all Analytics frontend report deep links should be updated to rely on the new `getServiceReportURL` selector of the `modules/analytics` store. This will fix various instances in which arguments for those URLs are incorrectly encoded.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All deep links to Analytics (i.e. links that rely on the `getServiceURL` selector of `modules/analytics` with a URL starting with `/report`) should be updated to rely on the `getServiceReportURL` selector.
    * Parameters that are currently manually "encoded" should be passed to `getServiceReportURL` as `reportArgs`.
    * Those links no longer need to pass the Analytics account, property and view IDs since those are handled internally by that selector, simplifying the surrounding code.

## Implementation Brief
- Using `assets/js/modules/analytics/components/common/AcquisitionPieChart.js`
	- Update the `sourceURI` constant to use the `getServiceReportURL` selector with `url` being the current entity URL.
		```js
		const sourceURI = useSelect(
			( select ) => select( STORE_NAME ).getServiceReportURL( 'trafficsources-overview', { '_r.drilldown': `analytics.pagePath:${ url }` } )
		);

		```
- Using `assets/js/modules/analytics/components/dashboard/AnalyticsDashboardWidget.js`
	- Update the `topContentServiceURL`, `topAcquisitionServiceURL`, `visitorsOverview` variables to use the `getServiceReportURL` selector:
		```js
		const topContentServiceURL = useSelect( ( select ) => select( STORE_NAME ).getServiceReportURL( 'content-pages' ) );
		const topAcquisitionServiceURL = useSelect( ( select ) => select( STORE_NAME ).getServiceReportURL( 'trafficsources-overview' ) );
		const visitorsOverview = useSelect( ( select ) => select( STORE_NAME ).getServiceReportURL( 'visitors-overview' ) );
		```
- Using `assets/js/modules/analytics/components/dashboard/AnalyticsDashboardWidgetTopPagesTable.js`
	- Remove the `baseServiceURL` variable
	- In the `dataMapped` array, set the `link` for the item being looped to `url` only.
		```js
		links[ i ] = url;
		```
[...]

PR #2521, opened 2020-12-10
Update Analytics report deep links.

tofumatt:
## Summary

Addresses issue #2405.

## Checklist


assets/js/modules/analytics/components/dashboard/LegacyAnalyticsDashboardWidgetPopularPagesTable.js
```diff
@@ -38,14 +38,9 @@ import TableOverflowContainer from '../../../../components/TableOverflowContaine
...
-		const accountID = select( STORE_NAME ).getAccountID();
-		const profileID = select( STORE_NAME ).getProfileID();
-		const internalWebPropertyID = select( STORE_NAME ).getInternalWebPropertyID();
-		return select( STORE_NAME ).getServiceURL(
-			{ path: `/report/content-pages/a${ accountID }w${ internalWebPropertyID }p${ profileID }` }
-		);
-	} );
+	const serviceURL = useSelect(
+		( select ) => select( STORE_NAME ).getServiceReportURL( 'content-pages' )
+	);
```
eugene-manuilov, line 43:
```suggestion
	const serviceURL = useSelect( ( select ) => select( STORE_NAME ).getServiceReportURL( 'content-pages' ) );
```

assets/js/modules/analytics/components/dashboard/LegacyAnalyticsDashboardWidgetTopLevel.js
```diff
@@ -101,23 +100,18 @@ function LegacyAnalyticsDashboardWidgetTopLevel( { data, requestDataToState } )
...
+	const uniqueVisitorsServiceURL = useSelect(
+		( select ) => select( STORE_NAME ).getServiceReportURL( `visitors-overview`, {
+			'_r.drilldown': `analytics.pagePath:${ url }`,
+		} )
+	);
+	const goalsServiceURL = useSelect(
+		( select ) => select( STORE_NAME ).getServiceReportURL( `conversions-goals-overview`, {
+			'_r.drilldown': `analytics.pagePath:${ url }`,
+		} )
+	);
```
eugene-manuilov, line 114:
```suggestion
	const uniqueVisitorsServiceURL = useSelect( ( select ) => select( STORE_NAME ).getServiceReportURL( `visitors-overview`, {
		'_r.drilldown': `analytics.pagePath:${ url }`,
	} ) );
	const goalsServiceURL = useSelect( ( select ) => select( STORE_NAME ).getServiceReportURL( `conversions-goals-overview`, {
		'_r.drilldown': `analytics.pagePath:${ url }`,
	} ) );
```

assets/js/modules/analytics/components/dashboard/LegacyDashboardAcquisitionPieChart.js
```diff
@@ -40,19 +40,17 @@ import { TYPE_MODULES } from '../../../../components/data';
...
 	const url = useSelect( ( select ) => select( CORE_SITE ).getCurrentEntityURL() );
 
-	const sourceURI = useSelect( ( select ) => select( STORE_NAME ).getServiceURL(
-		{ path: applyEntityToReportPath( url, `/report/trafficsources-overview/a${ accountID }w${ internalWebPropertyID }p${ profileID }/` ) }
-	) );
+	const sourceURI = useSelect(
+		( select ) => select( STORE_NAME ).getServiceReportURL( 'trafficsources-overview', {
+			'_r.drilldown': `analytics.pagePath:${ url }`,
+		} )
+	);
```
eugene-manuilov, line 53:
```suggestion
	const sourceURI = useSelect( ( select ) => select( STORE_NAME ).getServiceReportURL( 'trafficsources-overview', {
		'_r.drilldown': `analytics.pagePath:${ url }`,
	} ) );
```

assets/js/modules/analytics/components/dashboard/AnalyticsAdSenseDashboardWidgetTopPagesTable.js
```diff
@@ -97,7 +97,7 @@ const AnalyticsAdSenseDashboardWidgetTopPagesTable = ( { data } ) => {
 		chartsEnabled: false,
 		links: rows.map( ( row ) => row.dimensions[ 1 ] || '/' ),
 		PrimaryLink: withSelect( ( select, { href = '/' } ) => {
-			const serviceURL = select( STORE_NAME ).getServiceReportURL( 'content-pages', {
+			const serviceURL = select( STORE_NAME ).getServiceReportURL( 'content-drilldown', {
```
eugene-manuilov, line 100: @tofumatt, has it been changed intentionally? It doesn't seem to be required in IB, plus we haven't had `content-drilldown` before... If it has been changed intentionally, then should we update the `AnalyticsDashboardWidget` component to be use `content-drilldown` instead of `content-pages` as well?

eugene-manuilov (changes requested):
@tofumatt everything looks good to me. Just added some suggestions to make hooks formatting be consistent with how we do it for other hooks, plus added one question. Could you please take a look at it?

assets/js/modules/analytics/components/dashboard/AnalyticsAdSenseDashboardWidgetTopPagesTable.js
tofumatt: Nope, that was totally my bad. I meant to copy this key/value and move it to another file, but I must've lost my place and placed it here instead. 😅

Fixed!

eugene-manuilov (approved):
Awesome! Thanks, @tofumatt!

assets/js/modules/analytics/components/common/AcquisitionPieChart.js
```diff
@@ -67,14 +66,10 @@ const GOOGLE_CHART_PIE_SETTINGS = {
...
-	const profileID = useSelect( ( select ) => select( STORE_NAME ).getProfileID() );
-	const internalWebPropertyID = useSelect( ( select ) => select( STORE_NAME ).getInternalWebPropertyID() );
 	const url = useSelect( ( select ) => select( CORE_SITE ).getCurrentEntityURL() );
-	const sourceURI = useSelect( ( select ) => select( STORE_NAME ).getServiceURL(
-		{
-			path: applyEntityToReportPath( url, `/report/trafficsources-overview/a${ accountID }w${ internalWebPropertyID }p${ profileID }/` ),
-		} ) );
+	const sourceURI = useSelect( ( select ) => select( STORE_NAME ).getServiceReportURL( 'trafficsources-overview', {
+		'_r.drilldown': `analytics.pagePath:${ url }`,
+	} );
```
eugene-manuilov, line 72:
```suggestion
	const sourceURI = useSelect( ( select ) => select( STORE_NAME ).getServiceReportURL( 'trafficsources-overview', {
		'_r.drilldown': `analytics.pagePath:${ url }`,
	} ) );
```

eugene-manuilov (changes requested):
Found one more typo.

eugene-manuilov (changes requested):
@tofumatt, VR tests fail, could you please check it and figure out why it happens?

tofumatt:
I'm trying to do so but getting hit with remarkably slow VRT runtimes (over 20 minutes). I'll investigate further…

tofumatt:
Turns out this is because the storybook is loading legacy data for that story, and now that the component has been changed to use data from the datastore it's not getting a `href` prop to the `Link` component, causing it to render as a `button`, which is `display: inline-block` over `display: inline`.

I've investigated the effort into getting all the test data into that story but it seems way out of scope of this issue, and arguably the link should render the same visually regardless of `href` property. So I'm tweaking its style and submitting that in the PR to maintain VRT.

tofumatt:
Sadly, that won't work because of how buttons work 😭 

I'm going to revisit this one next week, as the legacy data that needs replacing is a lot.

tofumatt:
Heh, turns out it was much easier than that. Dinner and some time away made that clear 😄 

Set for review now 👍🏻

eugene-manuilov (approved):
Awesome! Thanks a lot, @tofumatt!

stories/data-table.stories.js
```diff
@@ -32,13 +32,26 @@ import { __, _x } from '@wordpress/i18n';
...
 storiesOf( 'Global', module )
 	.add( 'Data Table', () => {
 		global._googlesitekitLegacyData = analyticsDashboardData;
 
+		const setupRegistry = ( { dispatch } ) => {
+			dispatch( MODULES_ANALYTICS ).receiveGetSettings( {
+				accountID: '123456789',
+				propertyID: '123456789',
+				internalWebPropertyID: '123456789',
+				profileID: '123456789',
```
aaemnnosttv, line 49: Are these values necessary at all? If not the Analytics datastore fixtures has a `settings.default` which would be good to provide valid (but empty) defaults. Otherwise, the `propertyID` is not a valid format and should be `UA-#######-#`

aaemnnosttv (commented):
Looks great @tofumatt – just one question for you about some of the filler data in storybook.

stories/data-table.stories.js
tofumatt: Ah, good shout. I actually copied these from another storybook test. I'll take a quick look to see if these—and the other test—can use `settings.default`. They probably can.

tofumatt: Ah, they can't be empty, just tried. But I'll use valid values, cheers 👍🏻

aaemnnosttv (approved):
👍

assets/js/modules/analytics/components/common/AcquisitionPieChart.js
```diff
@@ -67,14 +66,10 @@ const GOOGLE_CHART_PIE_SETTINGS = {
...
-	const profileID = useSelect( ( select ) => select( STORE_NAME ).getProfileID() );
-	const internalWebPropertyID = useSelect( ( select ) => select( STORE_NAME ).getInternalWebPropertyID() );
 	const url = useSelect( ( select ) => select( CORE_SITE ).getCurrentEntityURL() );
-	const sourceURI = useSelect( ( select ) => select( STORE_NAME ).getServiceURL(
-		{
-			path: applyEntityToReportPath( url, `/report/trafficsources-overview/a${ accountID }w${ internalWebPropertyID }p${ profileID }/` ),
-		} ) );
+	const sourceURI = useSelect( ( select ) => select( STORE_NAME ).getServiceReportURL( 'trafficsources-overview', {
+		'_r.drilldown': `analytics.pagePath:${ url }`,
+	} ) );
```
felixarntz, line 72: You unconditionally add this `url`-based report argument here, which doesn't cater for the fact that `url` can be `null`. `url` is only not null on the single URL details view. `applyEntityToReportPath` had a conditional for this, and we need to maintain this here. Right now you'll see a bunch of links including `null` for that reason.

assets/js/modules/analytics/components/dashboard/DashboardBounceRateWidget.js
```diff
@@ -75,11 +70,9 @@ function DashboardBounceRateWidget() {
 			data: store.getReport( args ),
 			error: store.getErrorForSelector( 'getReport', [ args ] ),
 			loading: ! store.hasFinishedResolution( 'getReport', [ args ] ),
-			serviceURL: store.getServiceURL(
-				{
-					path: applyEntityToReportPath( url, `/report/visitors-overview/a${ accountID }w${ internalWebPropertyID }p${ profileID }/` ),
-				}
-			),
+			serviceURL: store.getServiceReportURL( 'visitors-overview', {
+				'_r.drilldown': `analytics.pagePath:${ url }`,
```
felixarntz, line 74: See above, `url` may be `null` here.

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.js
```diff
@@ -91,11 +87,9 @@ function DashboardUniqueVisitorsWidget() {
 			error: store.getErrorForSelector( 'getReport', [ sparklineArgs ] ) || store.getErrorForSelector( 'getReport', [ args ] ),
 			// Due to the nature of these queries, we need to run them separately.
 			sparkData: store.getReport( sparklineArgs ),
-			serviceURL: store.getServiceURL(
-				{
-					path: applyEntityToReportPath( url, `/report/visitors-overview/a${ accountID }w${ internalWebPropertyID }p${ profileID }/` ),
-				}
-			),
+			serviceURL: store.getServiceReportURL( 'visitors-overview', {
+				'_r.drilldown': `analytics.pagePath:${ url }`,
```
felixarntz, line 91: See above.

assets/js/modules/analytics/components/dashboard/LegacyAnalyticsDashboardWidgetTopLevel.js
```diff
@@ -101,23 +100,14 @@ function LegacyAnalyticsDashboardWidgetTopLevel( { data, requestDataToState } )
...
-		{
-			path: applyEntityToReportPath( url, `/report/conversions-goals-overview/${ pathIDSegment }` ),
-		}
-	) );
+	const uniqueVisitorsServiceURL = useSelect( ( select ) => select( STORE_NAME ).getServiceReportURL( `visitors-overview`, {
+		'_r.drilldown': `analytics.pagePath:${ url }`,
+	} ) );
+	const goalsServiceURL = useSelect( ( select ) => select( STORE_NAME ).getServiceReportURL( `conversions-goals-overview`, {
+		'_r.drilldown': `analytics.pagePath:${ url }`,
+	} ) );
```
felixarntz, line 110: See above.

assets/js/modules/analytics/components/dashboard/LegacyDashboardAcquisitionPieChart.js
```diff
@@ -40,19 +40,15 @@ import { TYPE_MODULES } from '../../../../components/data';
...
-	const accountID = useSelect( ( select ) => select( STORE_NAME ).getAccountID() );
-	const profileID = useSelect( ( select ) => select( STORE_NAME ).getProfileID() );
-	const internalWebPropertyID = useSelect( ( select ) => select( STORE_NAME ).getInternalWebPropertyID() );
 	const url = useSelect( ( select ) => select( CORE_SITE ).getCurrentEntityURL() );
 
-	const sourceURI = useSelect( ( select ) => select( STORE_NAME ).getServiceURL(
-		{ path: applyEntityToReportPath( url, `/report/trafficsources-overview/a${ accountID }w${ internalWebPropertyID }p${ profileID }/` ) }
-	) );
+	const sourceURI = useSelect( ( select ) => select( STORE_NAME ).getServiceReportURL( 'trafficsources-overview', {
+		'_r.drilldown': `analytics.pagePath:${ url }`,
```
felixarntz, line 50: See above.

felixarntz (changes requested):
Mostly LGTM, but there is one crucial problem here that was missed before: We must not drop the conditional check for whether there's a current entity URL or not.

tofumatt:
Ah, good catch on the `null` URLs… I've used `undefined` as the argument there when `url` isn't truthy, I think that should work fine.

felixarntz (changes requested):
One follow-up issue here, but otherwise this should be good to go.

felixarntz (changes requested):
(somehow the comment I thought to have posted in my last review didn't get published)

There is one follow-up problem here, which is that with the changes here the full `url` is added after `analytics.pagePath:`. It must be only the path segment of the URL though, starting with the initial slash. The `applyEntityToReportPath` function was previously taking care of that - maybe we should take the relevant parts from there and move them into a `getURLPath` utility function.

Please test the PR by running it - you'll see that the Analytics deep links result in errors in the Analytics FE because passing a full URL as `analytics.pagePath` is not valid.

tofumatt:
Ah, yes, the full URL was wrong.

But even after fixing it there's an issue.
 
Turns out part of the issue was the `getServiceReportURL()` selector escaping content that the Analytics site expects to be unescaped, see: https://github.com/google/site-kit-wp/pull/2521/files#diff-b76c9acca4f528cdea41f30e3983b26cd19bfa3a57d051bc385951948820e72cL133-R133

I'm concerned about that because I recall us escaping that URL so something else wouldn't break, but I can't see what that would be now, and all of the tests pass. But Analytics wants that URL to be like `_r.drilldown=analytics.pagePath:~2Fblog/` not `analytics.pagePath%25253A~2Ftesting/`—the escaped version (second) causes an error.

I think this is ready now, but note that I'm still not 100% sure about the escaping behaviour change—QAing all of the service URLs is a good idea and I've noted that in the QA brief for this issue now.

assets/js/modules/analytics/datastore/service.js
```diff
@@ -90,7 +90,7 @@ export const selectors = {
 		let path = escapeURI`/report/${ type }/a${ accountID }w${ internalWebPropertyID }p${ profileID }/`;
 
 		if ( argsSegment ) {
-			path += escapeURI`${ argsSegment }/`;
+			path += `${ argsSegment }/`;
```
felixarntz, line 93: @aaemnnosttv Does any of this actually need to be URI-escaped? Looks to me like it doesn't - maybe we should also remove that from the `path` assignment above in line 90. See @tofumatt's comment above.

felixarntz (approved):
Awesome! I think this is good to go, but wanted to double-check with @aaemnnosttv on the escaping. I _do_ think it's unnecessary and actually breaks URLs for how they're required in Analytics.

assets/js/modules/analytics/datastore/service.js
aaemnnosttv: Report arg _values_ need to be URL encoded but nothing else should actually need it, assuming valid values are returned by selectors for `accountID` etc. If they weren't valid the URL wouldn't work anyways though. The `argsSegment` is already URL encoded as needed so this shouldn't have had `escapeURI` applied again because that would encode other characters in the string that shouldn't be, but it's correct now.

As far as the use in the `path` assignment, the `escapeURI` template function only encodes the variables in the string (e.g. `${ type }`), so as long as these are normal valid values it would essentially have no effect because these are all simple characters and numbers. I'm not sure it's really necessary there, but it doesn't hurt.

aaemnnosttv (commented):
Looks good to me overall. There are a few file headers to fix and I would prefer an alternate logic syntax for the conditional pagePath parameter but nothing blocking.

assets/js/util/getURLPath.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * Utility functions for Tags
```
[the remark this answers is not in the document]
aaemnnosttv:
```suggestion
 * URL pathname getter utility function.
```

assets/js/util/getURLPath.test.js
```diff
@@ -0,0 +1,72 @@
+/**
+ * Escape URI components utility tests.
```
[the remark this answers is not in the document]
aaemnnosttv:
```suggestion
 * URL pathname getter utility function tests.
```

aaemnnosttv (approved):
I quickly addressed the remaining feedback here:
* Updated file headers
* Updated `reportArgsToURLSegment` to ignore entries with `undefined` values, similar to `addQueryArgs` + updated tests
* Updated `getServiceReportURL` selectors to conditionally pass a value for the `_r.drilldown` key if `url` similar to before, but the ternary is specific to the key now rather than the object as a whole

Everything else LGTM 👍

aaemnnosttv:
@googlebot I consent.

felixarntz (approved):

PR #2523, opened 2020-12-11
Update webpack config

asvinb:
## Summary

Addresses issue #2344 

## Relevant technical choices
- Add hash only when in production mode

## Checklist


tofumatt (approved):

felixarntz (changes requested):
I think this PR misses one specific instance, which is the `googlesitekit-vendor` file which currently still receives the hash in a development build (see the `googlesitekit-vendor.[contenthash].js` in the Webpack config).

We'll need to add a similar change there so that the hash is only added for production.

asvinb:
> I think this PR misses one specific instance, which is the `googlesitekit-vendor` file which currently still receives the hash in a development build (see the `googlesitekit-vendor.[contenthash].js` in the Webpack config).
> 
> We'll need to add a similar change there so that the hash is only added for production.

Updated @felixarntz

felixarntz (approved):
Great!

Issue #1941, opened 2020-08-21
Update invariant use within datastore generator functions


aaemnnosttv:
## Bug Description

In #1801 we updated the way `createFetchStore` works internally so that invalid use raises errors as expected.

As discovered in that issue, the main problem is that errors raised in a generator function do not throw but instead result in a rejected promise which is not handled well.

We still have many instances of `invariant` used within generators in the datastore outside of the `createFetchStore` use.

E.g. https://github.com/google/site-kit-wp/blob/62c82436e2a69f7d7e4c544f07c73d7e838269a8/assets/js/googlesitekit/modules/datastore/modules.js#L181-L183

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new `createValidatedAction` utility should be created for simplifying the definition of an async action that has validation
	- It should take two arguments: `validate` (a non-generator function), `actionCreator` (a function which may or may not be a generator that returns and/or yields actions)
* All instances of action creators in the data store which are generators that perform validation on parameters (i.e. `invariant`) should be updated to use the new utility

## Implementation Brief

* Finish and merge https://github.com/google/site-kit-wp/pull/3158
	- Add tests for `createValidatedAction`
	- Update remaining actions

## QA Brief

* Review the codebase to make sure there are no longer any actions defined as generator functions which call `invariant` internally

## Changelog entry

* N/A

aaemnnosttv:
@felixarntz moving this forward to AC as there are still quite a few of these in the codebase.

felixarntz:
@aaemnnosttv I like the approach you're suggesting, great! IB ✅

wpdarren:
@aaemnnosttv just checking this should be a `QA:Eng` ticket since it mentions looking at the codebase?

aaemnnosttv:
Yep, thanks @wpdarren – I've added it now 👍

tofumatt:
QA ✅

PR #2525, opened 2020-12-11
Enhancement/2271 allow registering module store name

benbowler:
## Summary

Implement `storeName` argument during module registration and apply its uses across the application.
Addresses issue #2271

## Checklist


benbowler:
Hey @aaemnnosttv,

The core of the changes to add the storeName argument worked and are built as defined. 

The IB task to "`[update] actions.submitChanges in assets/js/googlesitekit/modules/datastore/settings.js
change to be a normal action that returns an object`" was not possible - I wasn't able to call the control without keeping it as a generator function. Refactoring to a control also reintroduced an issue [as mentioned in #1941](https://github.com/google/site-kit-wp/issues/1941) for the invariant violation when missing slugs and introduced some testing issues across the whole of the `settings.js` file because each of the actions and selectors that use the new `getModuleStoreName` selector now call `getModules` (both in `assets/js/googlesitekit/modules/datastore/modules.js`) which returns undefined in the testing context.

Because of the amount of time I've already spent on this over the estimate I wanted to start a discussion/get your thoughts on how to complete the refactor of `settings.js` as well as to reliably test it to complete the task.

benbowler:
Hey @aaemnnosttv, this is now ready for CR, I've updated the `submitChanges` function to pass the invariant error correctly and I've updated the tests to use `provideModules` to get them passing. 

I combined your and Eugene's approaches to test the `isDoingSubmitChanges` selector by adding in a callback into our test `submitChanges` overide that let's us call the selector at that point giving us an explicit test case rather than relying on `untilResolved` to fail like so:

```
				// Check isDoingSubmitChanges is true whiule submitChanges is being preformed
				testDuringSubmissionCallback = () => {
					expect( registry.select( STORE_NAME ).isDoingSubmitChanges( slug ) ).toBe( true );
				};
				await registry.dispatch( STORE_NAME ).submitChanges( slug );
```

assets/js/components/settings/SettingsRenderer.js
```diff
@@ -30,8 +30,8 @@ const { useSelect, useDispatch } = Data;
 
 export default function SettingsRenderer( { slug, isOpen, isEditing } ) {
 	const storeName = `modules/${ slug }`;
```
aaemnnosttv, line 32: This still needs to change to use the registered store name.

assets/js/components/settings/SettingsRenderer.js
```diff
@@ -30,8 +30,8 @@ const { useSelect, useDispatch } = Data;
...
-	const isDoingSubmitChanges = useSelect( ( select ) => select( storeName )?.isDoingSubmitChanges?.() );
-	const haveSettingsChanged = useSelect( ( select ) => select( storeName )?.haveSettingsChanged?.() );
+	const isDoingSubmitChanges = useSelect( ( select ) => select( 'core/modules' )?.isDoingSubmitChanges?.( slug ) );
```
aaemnnosttv, line 33: This should use a constant for the `STORE_NAME` (aliased to `CORE_MODULES`).
Also, the conditional `?` chaining isn't necessary here since we can rely on the selector existing now.

assets/js/googlesitekit/modules/create-info-store.js
```diff
@@ -36,18 +41,20 @@ const { createRegistrySelector } = Data;
...
  * @return {Object} The info store object.
  */
 export const createInfoStore = ( slug, {
 	storeName = undefined,
 	adminPage = 'googlesitekit-dashboard',
 	requiresSetup = true,
 } = {} ) => {
+	invariant( storeName, 'storeName is required.' );
+
 	const STORE_NAME = storeName || `modules/${ slug }`;
```
aaemnnosttv, line 58: * Since `storeName` is required now, this fallback should be removed.
* Also, let's update references below to use `storeName` rather than `STORE_NAME` to show that this is from the argument rather than an imported constant.

assets/js/googlesitekit/modules/create-info-store.test.js
```diff
@@ -30,6 +30,7 @@ import { STORE_NAME as CORE_SITE } from '../datastore/site/constants';
 import { STORE_NAME as CORE_USER } from '../datastore/user/constants';
 
 const MODULE_SLUG = 'test-slug';
+const TEST_STORE_NAME = 'test/' + MODULE_SLUG;
```
aaemnnosttv, line 33: Minor detail, but let's use the interpolated template string syntax instead.

assets/js/googlesitekit/modules/create-info-store.test.js
```diff
@@ -44,9 +45,9 @@ describe( 'createInfoStore store', () => {
 
 	describe( 'storeName', () => {
 		it( 'returns the correct default store name', () => {
```
aaemnnosttv, line 47: This test should be updated to test that it throws an error if no `storeName` is provided since it is a required argument now (there is no default now).

assets/js/googlesitekit/modules/create-module-store.js
```diff
@@ -47,15 +47,15 @@ import { createSubmitChangesStore } from './create-submit-changes-store';
  *
  * @since 1.6.0
  *
- * @param {string}   slug                             Slug of the module that the store is for.
- * @param {Object}   options                          Optional. Options to consider for the store.
- * @param {number}   options.storeName                Store name to use. Default is 'modules/{slug}'.
- * @param {Array}    options.settingSlugs             If the module store should support settings, this needs to be a list of the slugs that are part of the module and handled by the module's 'modules/{slug}/data/settings' API endpoint.
+ * @param {string}   slug                            Slug of the module that the store is for.
+ * @param {Object}   args                            Arguments to consider for the store.
+ * @param {number}   args.storeName                  Store name to use. Default is 'modules/{slug}'.
```
aaemnnosttv, line 52: Again, no more default 😄

assets/js/googlesitekit/modules/create-module-store.js
```diff
@@ -68,8 +68,9 @@ export const createModuleStore = ( slug, {
 	validateCanSubmitChanges = undefined,
 } = {} ) => {
 	invariant( slug, 'slug is required.' );
+	invariant( storeName, 'storeName is required.' );
 
-	storeName = storeName || `modules/${ slug }`;
+	storeName = storeName || `modules/${ slug }`; // TODO: remove this now that the storeName is required
```
aaemnnosttv, line 73: This line should be removed now as the TODO states 😄

assets/js/googlesitekit/modules/create-module-store.test.js
```diff
@@ -34,6 +34,7 @@ import { createSubmitChangesStore } from './create-submit-changes-store';
 
 const SETTING_SLUG = 'testSetting';
 const MODULE_SLUG = 'base';
+const TEST_STORE_NAME = 'test/' + MODULE_SLUG;
```
aaemnnosttv, line 37:
```suggestion
const TEST_STORE_NAME = `test/${ MODULE_SLUG }`;
```

assets/js/googlesitekit/modules/create-module-store.test.js
```diff
@@ -48,7 +49,7 @@ describe( 'createModuleStore store', () => {
 	beforeEach( () => {
 		registry = createRegistry();
 
-		storeDefinition = createModuleStore( MODULE_SLUG, { settingSlugs, registry } );
+		storeDefinition = createModuleStore( MODULE_SLUG, { storeName: TEST_STORE_NAME, settingSlugs, registry } );
```
aaemnnosttv, line 52: There is no `registry` argument to `createModuleStore` anymore so this can be removed 👍

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -222,9 +223,11 @@ const baseActions = {
 	 * @since 1.20.0 Introduced the ability to register settings and setup components.
 	 * @since 1.22.0 Introduced the ability to add a checkRequirements function.
 	 * @since n.e.x.t Introduced the ability to register an Icon component.
+	 * @since n.e.x.t Introduced the ability to explictly define a module store name.
 	 *
 	 * @param {string}      slug                             Module slug.
 	 * @param {Object}      [settings]                       Optional. Module settings.
+	 * @param {Object}      [settings.storeName]             Optional. Module storeName. If non is provided we assume no store is required for this module.
```
aaemnnosttv, line 230:
```suggestion
	 * @param {string}      [settings.storeName]             Optional. Module storeName. If none is provided we assume no store exists for this module.
```

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -321,7 +326,14 @@ export const baseControls = {
 	} ),
 	[ SELECT_MODULE_REAUTH_URL ]: createRegistryControl( ( { select } ) => ( { payload } ) => {
 		const { slug } = payload;
-		const getAdminReauthURL = select( `modules/${ slug }` )?.getAdminReauthURL;
+		const storeName = select( 'core/modules' )?.getModuleStoreName( slug );
```
aaemnnosttv, line 329: Please use the core store name constant instead. Also, no conditional chaining needed here since we can rely on the presence of the core selector.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -601,6 +613,29 @@ const baseSelectors = {
...
+	/**
+	 * Gets module store name by slug.
+	 *
+	 * Returns the store name if preset or null if there is no store name for this module.
+	 * Returns `undefined` if state is still loading or if said module doesn't exist.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {string} slug Module slug.
+	 * @return {(string|null|undefined)} An string of the store name if a name has been set for this module; null if no store name was set; `undefined` if state is still loading.
```
aaemnnosttv, line 625: When there are multiple types like this, it helps to break each onto its own line.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -601,6 +613,29 @@ const baseSelectors = {
...
+	getModuleStoreName: createRegistrySelector( ( select ) => ( state, slug ) => {
+		const module = select( STORE_NAME ).getModule( slug );
+
+		// Return `undefined` if module with this slug isn't loaded yet.
+		if ( module === undefined ) {
+			return undefined;
+		}
+
+		// Return the store name or null if no store name was set.
+		return module?.storeName || null;
```
aaemnnosttv, line 636: This can be simplified like so due to the defaults applied in `getModules`

```suggestion
		return module.storeName;
```
```

File: assets/js/googlesitekit/modules/datastore/settings.js [hunk 14]
```diff
@@ -7,7 +7,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     https://www.apache.org/licenses/LICENSE-2.0
+ *     https://www.apache.org/licenses/LICENSE2.0
```
aaemnnosttv, line 10: Oops 😄 
```suggestion
 *     https://www.apache.org/licenses/LICENSE-2.0
```

assets/js/googlesitekit/modules/datastore/settings.js
```diff
@@ -25,7 +25,12 @@ import invariant from 'invariant';
  * Internal dependencies
  */
 import Data from 'googlesitekit-data';
-const { createRegistrySelector } = Data;
+// import { dispatch } from '@wordpress/data';
+const { createRegistrySelector, createRegistryControl } = Data;
+// import { STORE_NAME } from './constants';
+import { STORE_NAME as CORE_MODULES } from './constants';
```
aaemnnosttv, line 31: Since this file/module is in the context of the `core/modules` store, `STORE_NAME` should be used to reference its own store rather than the alias.

assets/js/googlesitekit/modules/datastore/settings.js
```diff
@@ -25,7 +25,12 @@ import invariant from 'invariant';
...
+
+const SETTINGS_SUBMIT_CHANGES = 'SETTINGS_SUBMIT_CHANGES';
```
aaemnnosttv, line 33: Let's update the action type here to be a little more specific to this store:

```suggestion
const SUBMIT_MODULE_CHANGES = 'SUBMIT_MODULE_CHANGES';
```

assets/js/googlesitekit/modules/datastore/settings.js
```diff
@@ -38,19 +43,34 @@ export const actions = {
...
+				type: SETTINGS_SUBMIT_CHANGES,
+			};
 		}() );
 	},
 };
 
+export const controls = {
+	[ SETTINGS_SUBMIT_CHANGES ]: createRegistryControl( ( registry ) => async ( { payload } ) => {
+		const { slug } = payload;
+		const storeName = registry.select( CORE_MODULES )?.getModuleStoreName( slug );
```
aaemnnosttv, line 59: As mentioned before, the conditional chaining is unnecessary when selecting from a known store.

assets/js/googlesitekit/modules/datastore/settings.js
```diff
@@ -38,19 +43,34 @@ export const actions = {
...
+
+		if ( ! storeName ) {
+			return { error: `The module ${ slug } does not have a store.` };
+		}
+
+		const { submitChanges } = await registry.dispatch( storeName );
```
aaemnnosttv, line 65: The `dispatch` here is not asynchronous, it simply returns the object of bound action creators for the given store name, so the `await` can be removed.

```suggestion
		const { submitChanges } = registry.dispatch( storeName );
```

assets/js/googlesitekit/modules/datastore/settings.js
```diff
@@ -62,7 +82,8 @@ export const selectors = {
 	 */
 	isDoingSubmitChanges: createRegistrySelector( ( select ) => ( state, slug ) => {
 		invariant( slug, 'slug is required.' );
-		return !! select( `modules/${ slug }` )?.isDoingSubmitChanges?.();
+		const storeName = select( CORE_MODULES )?.getModuleStoreName( slug );
```
aaemnnosttv, line 85:
```suggestion
		const storeName = select( CORE_MODULES ).getModuleStoreName( slug );
```

assets/js/googlesitekit/modules/datastore/settings.js
```diff
@@ -75,12 +96,13 @@ export const selectors = {
 	 */
 	canSubmitChanges: createRegistrySelector( ( select ) => ( state, slug ) => {
 		invariant( slug, 'slug is required.' );
-		return !! select( `modules/${ slug }` )?.canSubmitChanges?.();
+		const storeName = select( CORE_MODULES )?.getModuleStoreName( slug );
```
aaemnnosttv, line 99:
```suggestion
		const storeName = select( CORE_MODULES ).getModuleStoreName( slug );
```

assets/js/googlesitekit/modules/datastore/settings.test.js
```diff
@@ -38,25 +39,32 @@ describe( 'core/modules settings', () => {
 			moduleStoreName,
 			Modules.createModuleStore( slug, {
 				storeName: moduleStoreName,
-				submitChanges: () => controlReturn,
+				submitChanges: async () => {
+					await testDuringSubmissionCallback();
+					return testReturnValue;
+				},
```
aaemnnosttv, line 45: Let's update this to use a mock function instead. That way we can both make assertions that the function was called, with what arguments, but also we can change the implementation after the store has been registered.
E.g. above
```js
let submitChanges;
// ...
submitChanges = jest.fn()
```

And here just pass the function
```suggestion
				submitChanges,
```

Then in tests where we care about the return value we only need to set the internal function to return whatever we want:
```js
submitChanges.mockImplementation( () => 'foo' );
```
Or in the case of the test for `isDoingSubmitChanges`, we can simply have it return the value we care about:
```js
submitChanges.mockImplementation( ( { select } ) => select( STORE_NAME ).isDoingSubmitChanges( slug ) );
```
Then we can keep the assertion out of the callback (which might not have even failed the test anyways).

We can also add `expect( submitChanges ).toHaveBeenCalled()` as well, where this was more implicit before.

assets/js/googlesitekit/modules/datastore/settings.test.js
```diff
@@ -74,22 +82,14 @@ describe( 'core/modules settings', () => {
...
-					Modules.createModuleStore( slug, {
-						storeName: moduleStoreName,
-						validateCanSubmitChanges: () => true,
-						submitChanges: async ( { select } ) => {
-							expect( select( STORE_NAME ).isDoingSubmitChanges( slug ) ).toBe( true );
-							return {};
-						},
-					} ),
-				);
+				// Check isDoingSubmitChanges is true whiule submitChanges is being preformed
```
aaemnnosttv, line 85: Typos
```suggestion
				// Check isDoingSubmitChanges is true while submitChanges is being performed
```

assets/js/googlesitekit/modules/datastore/settings.js
```diff
@@ -38,19 +43,34 @@ export const actions = {
...
 	},
 };
 
+export const controls = {
+	[ SETTINGS_SUBMIT_CHANGES ]: createRegistryControl( ( registry ) => async ( { payload } ) => {
+		const { slug } = payload;
+		const storeName = registry.select( CORE_MODULES )?.getModuleStoreName( slug );
+
+		if ( ! storeName ) {
+			return { error: `The module ${ slug } does not have a store.` };
```
aaemnnosttv, line 62: Similar to how it was before with the store name, let's quote the module slug in the string.
```suggestion
			return { error: `The module '${ slug }' does not have a store.` };
```

assets/js/googlesitekit/modules/datastore/settings.js
```diff
@@ -38,19 +43,34 @@ export const actions = {
...
+		}
+
+		const { submitChanges } = await registry.dispatch( storeName );
+		if ( ! submitChanges ) {
+			return { error: `The module ${ slug } does not have a submitChanges() action.` };
```
aaemnnosttv, line 67: Same here
```suggestion
			return { error: `The module '${ slug }' does not have a submitChanges() action.` };
```

aaemnnosttv (changes requested):
Thanks @benbowler – this one is on the right track but still has a few things to clean up and address. See my comments below and let me know if you have any questions about anything.

assets/js/googlesitekit/modules/datastore/settings.test.js
benbowler: Nice, this is a great use of jest.fn()

assets/js/components/settings/SettingsRenderer.js
aaemnnosttv: @benbowler you marked this as resolved but the comment is not yet addressed. As an aside, please don't mark feedback as resolved as it makes it harder to see (possibly how you missed it in this case as well).

assets/js/components/settings/SettingsRenderer.js
aaemnnosttv: This is still unaddressed.

assets/js/googlesitekit/modules/create-info-store.test.js
aaemnnosttv: Marked as resolved but still unaddressed.

assets/js/googlesitekit/modules/create-info-store.test.js
aaemnnosttv: Marked as resolved but still unaddressed.

aaemnnosttv (commented):
@benbowler – not sure what happened here but I see that you've marked the previous feedback as resolved yet the comments were not addressed in the code. Did you perhaps forget to push your changes?

In the future, please don't mark CR conversations as resolved as this is really for the reviewer to do. (I saw you added 👍  reactions, which is great – please continue to do this where you otherwise don't have a response to add)

Please review all of the feedback from the previous round to make sure everything is addressed or responded to before sending back into review.

benbowler:
@aaemnnosttv I didn't push my branch. Changes up to date now for you to properly review.

Noted around resolving conversations.

aaemnnosttv:
Thanks @benbowler – would you please resolve the merge conflict? 🙂

assets/js/googlesitekit/modules/create-info-store.js
```diff
@@ -106,7 +111,7 @@ export const createInfoStore = ( slug, {
 	};
 
 	return {
-		STORE_NAME,
+		storeName,
```
aaemnnosttv, line 114: Let's preserve the case of the key in the return value here for consistency with other `create*Store` utilities.

assets/js/googlesitekit/modules/create-info-store.test.js
aaemnnosttv: I see you've added a new test to address the initial request here. In that case, this test should be removed since there is no default store name now.

assets/js/googlesitekit/modules/create-info-store.test.js
```diff
@@ -76,7 +83,7 @@ describe( 'createInfoStore store', () => {
 			// It uses `adminPage` when provided.
 			it( 'returns adminPage url when `adminPage` is provided', () => {
 				registry.dispatch( CORE_SITE ).receiveSiteInfo( { adminURL: 'http://example.com/wp-admin/' } );
-				const { STORE_NAME, ...store } = createInfoStore( MODULE_SLUG, { adminPage: 'test-admin-page' } );
+				const { STORE_NAME, ...store } = createInfoStore( MODULE_SLUG, { storeName: TEST_STORE_NAME, adminPage: 'test-admin-page' } );
```
aaemnnosttv, line 86: This line is fine going forward (in the context of other feedback above), I just wanted to highlight that the returned store name case was updated in some cases and not in others, so please make sure that `STORE_NAME` is destrucutred or otherwise referenced from the return in all instances.

assets/js/googlesitekit/modules/create-module-store.js
aaemnnosttv: This is still yet to be addressed.

assets/js/googlesitekit/modules/create-module-store.js
```diff
@@ -68,8 +68,9 @@ export const createModuleStore = ( slug, {
 	validateCanSubmitChanges = undefined,
 } = {} ) => {
 	invariant( slug, 'slug is required.' );
+	invariant( storeName, 'storeName is required.' );
 
-	storeName = storeName || `modules/${ slug }`;
+	storeName = storeName || `test/${ slug }`; // TODO: remove this now that the storeName is required
```
aaemnnosttv, line 73: Again, no more default so the `||` logic and fallback should be removed, as per the comment you added.

assets/js/googlesitekit/modules/create-module-store.test.js
aaemnnosttv: This comment is yet to be addressed.

assets/js/googlesitekit/modules/create-module-store.test.js
aaemnnosttv: This comment is yet to be addressed.

assets/js/googlesitekit/modules/datastore/modules.js
aaemnnosttv: This comment is yet to be addressed.

assets/js/googlesitekit/modules/datastore/modules.js
aaemnnosttv: This comment is yet to be addressed.

assets/js/googlesitekit/modules/datastore/modules.js
aaemnnosttv: This comment is yet to be addressed.

assets/js/googlesitekit/modules/datastore/modules.js
aaemnnosttv: This comment is yet to be addressed.

assets/js/googlesitekit/modules/datastore/settings.js
aaemnnosttv: This comment is yet to be addressed.

assets/js/googlesitekit/modules/datastore/settings.js
aaemnnosttv: This comment is yet to be addressed.

assets/js/googlesitekit/modules/datastore/settings.js
aaemnnosttv: This comment is yet to be addressed.

assets/js/googlesitekit/modules/datastore/settings.js
aaemnnosttv: This comment is yet to be addressed.

assets/js/googlesitekit/modules/datastore/settings.js
aaemnnosttv: This comment is yet to be addressed.

assets/js/googlesitekit/modules/datastore/settings.js
aaemnnosttv: This comment is yet to be addressed.

assets/js/googlesitekit/modules/datastore/settings.test.js
aaemnnosttv: This comment is yet to be addressed.

assets/js/googlesitekit/modules/create-module-store.js
aaemnnosttv: This comment is yet to be addressed.

aaemnnosttv (changes requested):
Thanks @benbowler – There's still quite a few comments and requested changes that were missed from the last round.

Also, there's still a merge conflict with `assets/js/googlesitekit/modules/datastore/modules.js` which is preventing the tests from running with the latest on the branch here.

Please take another look through the proposed changes on the **Files changed** tab here to make sure everything is addressed before sending back.

assets/js/components/settings/SettingsRenderer.js
benbowler: [Addressed here](https://github.com/google/site-kit-wp/pull/2525/files#diff-c7a8015e4e1004ea9e43de2e0a06e5b64345dedb0dac21dc94ef5a449969ac7eR32-R34)

assets/js/components/settings/SettingsRenderer.js
benbowler: [Addressed here](https://github.com/google/site-kit-wp/pull/2525/files#diff-c7a8015e4e1004ea9e43de2e0a06e5b64345dedb0dac21dc94ef5a449969ac7eR32-R34)

assets/js/googlesitekit/modules/create-info-store.test.js
benbowler: [Addressed here](https://github.com/google/site-kit-wp/pull/2525/files#diff-909552a7f570a1ec29e503680ef0539b36b2a060ea2c81f324b530caf8989352R32)

assets/js/googlesitekit/modules/create-info-store.test.js
benbowler: I've now [removed the default store name test](https://github.com/google/site-kit-wp/pull/2525/files#diff-734cdd487d9f3b1bb2aea8d73f6c8317c89712f1df2fec7a8141915bee9edf4fL46) in favour of testing passing and failing to pass the storeName.

assets/js/googlesitekit/modules/create-info-store.test.js
benbowler: Based on the above comment, `createInfoStore` now returns `STORE_NAME` not `storeName` so I've reverted these to match.

assets/js/googlesitekit/modules/datastore/settings.js
benbowler: Updated.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -601,6 +613,29 @@ const baseSelectors = {
...
+	getModuleStoreName: createRegistrySelector( ( select ) => ( state, slug ) => {
+		const module = select( STORE_NAME ).getModule( slug );
+
+		// Return `undefined` if module with this slug isn't loaded yet.
+		if ( module === undefined ) {
+			return undefined;
+		}
+
+		// Return the store name or null if no store name was set.
+		return module?.storeName;
```
benbowler, line 636: Removing optional chaining here causes a sea of tests to fail with `Cannot read property 'storeName' of null` as failure cases aren't handled here.

assets/js/googlesitekit/modules/datastore/modules.js
benbowler: I've now separated these with tabs.

assets/js/googlesitekit/modules/datastore/settings.js
benbowler: [Removed](https://github.com/google/site-kit-wp/pull/2525/files#diff-80a2c5f54e6197f4fad0b4cf508c8337f5af754bf8ddd8b6286f65db5b020442R97).

assets/js/googlesitekit/modules/datastore/settings.js
benbowler: Updated.

assets/js/googlesitekit/modules/datastore/settings.js
benbowler: Updated.

assets/js/googlesitekit/modules/create-info-store.js
benbowler: I've changed this and reverted the tests to use this constant form instead of camel case.

benbowler:
Morning @aaemnnosttv, I've been through and double checked all of the points and made some additional updates to tests in response to the changes. Let me know how this version looks.

assets/js/googlesitekit/modules/datastore/settings.js
```diff
@@ -25,7 +25,10 @@ import invariant from 'invariant';
  * Internal dependencies
  */
 import Data from 'googlesitekit-data';
-const { createRegistrySelector } = Data;
+const { createRegistrySelector, createRegistryControl } = Data;
+import { STORE_NAME as CORE_MODULES } from './constants';
```
aaemnnosttv, line 29: No alias should be used here since it is using its own store name; we only alias store names when imported from other stores when in the context of a store (or store partial in this case).

```suggestion
import { STORE_NAME } from './constants';
```

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -601,6 +613,36 @@ const baseSelectors = {
...
+		if ( module === undefined ) {
+			return undefined;
+		}
+
+		// Return null if no store name was set
+		if ( module === null ) {
+			return null;
+		}
+
+		// Return the store name.
```
aaemnnosttv, line 642: This is a bit of a nitpick but let's not add comments that don't add additional context. We don't need to document every line 😄 

```suggestion
```

assets/js/googlesitekit/modules/datastore/settings.js
```diff
@@ -38,19 +41,34 @@ export const actions = {
...
+				type: SUBMIT_MODULE_CHANGES,
+			};
 		}() );
 	},
 };
 
+export const controls = {
+	[ SUBMIT_MODULE_CHANGES ]: createRegistryControl( ( registry ) => ( { payload } ) => {
+		const { slug } = payload;
+		const storeName = registry.select( CORE_MODULES ).getModuleStoreName( slug );
```
aaemnnosttv, line 57:
```suggestion
		const storeName = registry.select( STORE_NAME ).getModuleStoreName( slug );
```

assets/js/googlesitekit/modules/datastore/settings.js
```diff
@@ -62,7 +80,8 @@ export const selectors = {
 	 */
 	isDoingSubmitChanges: createRegistrySelector( ( select ) => ( state, slug ) => {
 		invariant( slug, 'slug is required.' );
-		return !! select( `modules/${ slug }` )?.isDoingSubmitChanges?.();
+		const storeName = select( CORE_MODULES ).getModuleStoreName( slug );
```
aaemnnosttv, line 83:
```suggestion
		const storeName = select( STORE_NAME ).getModuleStoreName( slug );
```

assets/js/googlesitekit/modules/datastore/settings.js
```diff
@@ -75,12 +94,13 @@ export const selectors = {
 	 */
 	canSubmitChanges: createRegistrySelector( ( select ) => ( state, slug ) => {
 		invariant( slug, 'slug is required.' );
-		return !! select( `modules/${ slug }` )?.canSubmitChanges?.();
+		const storeName = select( CORE_MODULES ).getModuleStoreName( slug );
```
aaemnnosttv, line 97:
```suggestion
		const storeName = select( STORE_NAME ).getModuleStoreName( slug );
```

assets/js/googlesitekit/modules/datastore/modules.js
aaemnnosttv: Looks like this may have been from something else since I see this was implemented with passing tests now 👍

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -612,7 +654,9 @@ const baseSelectors = {
 	 *
 	 * @param {Object} state Data store's state.
 	 * @param {string} slug  Module slug.
-	 * @return {(boolean|null|undefined)} TRUE when the module exists and is active; `undefined` if state is still loading or `null` if said module doesn't exist.
+	 * @return {(boolean|null|undefined)} `true` when the module exists and is active.
+	 * 									  `undefined` if state is still loading.
+	 * 									   `null` if said module doesn't exist.
```
aaemnnosttv, line 659: Extra space here 😉 
```suggestion
	 * 									  `null` if said module doesn't exist.
```

aaemnnosttv (commented):
Thanks @benbowler – this looks just about good to go, just a few tiny things left to clean up. I'll go ahead and apply them since they're all simple suggestions on the diff then this should be 🚀

aaemnnosttv (approved):
🎉

Issue #2426, opened 2020-11-24
Add interactivity to `UserDimensionsPieChart` component for new All Traffic widget


felixarntz:
The `UserDimensionsPieChart` component implemented via #2425 should be enhanced so that it is possible to click on a certain pie slice to make it the "current" slice (`dimensionValue`), which will result in the adjacent `TotalUserCount` and `UserCountGraph` components rendering information for only that `dimensionValue`.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `UserDimensionsPieChart` component should be enhanced to support an "active" state for the different pie slices:
    * It should select `getValue( FORM_ALL_TRAFFIC_WIDGET, 'dimensionValue' )` from `core/forms` to have awareness of what is the active `dimensionValue` (default none).
    * If there is an active `dimensionValue`, that slice should be highlighted like it is in the [Figma file](https://www.figma.com/file/NzvJTPFdqUVJC9Q1mmih4e/Google-SK-UX-Enhancements10up?node-id=1198%3A400). It should also be highlighted in the legend below.
    * Clicking on another slice should update the active slice to that one, via dispatching `setValues( FORM_ALL_TRAFFIC_WIDGET, { dimensionName, dimensionValue: newlySelectedValue } )` on `core/forms` - **except** when clicking on an aggregated "Others" entry, it should not be possible to select those.
    * Clicking the active slice should unselect it, via dispatching `setValues( FORM_ALL_TRAFFIC_WIDGET, { dimensionName, dimensionValue: '' } )` on `core/forms`.
* This functionality can be tested by looking at the `DashboardAllTrafficWidgetV2` as a whole: Clicking a pie slice should result in the total user count and the graph on the left to show data specific to that selected slice.
* If not already present, some slice-specific test data should be added to the Storybook story for `DashboardAllTrafficWidgetV2` so that the behavior is testable there.

## Implementation Brief

The main part of this ticket is going to be implemented in the PR for #2423. To complete this ticket, we need to implement the following task:

* Update the select event listener in the `UserDimensionsPieChart` component to reset the `dimensionValue` property when a slice is unselected (the `getSelected` method returns an empty array) or the `Others` slice is selected;

### Test Coverage

* N/A

### Visual Regression Changes

* N/A

## QA Brief

* Set up the plugin on your site;
* Activate and connect the Analytics module;
* Go to the dashboard page;
[...]

Issue #2425, opened 2020-11-24
Implement `UserDimensionsPieChart` component for new All Traffic widget


felixarntz:
After scaffolding the `UserDimensionsPieChart` component in #2392, it should now actually be implemented.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The pie chart for the new All Traffic widget should be fully implemented and should encompass the UI around the pie chart in the bottom right corner of the widget (see [Figma file](https://www.figma.com/file/NzvJTPFdqUVJC9Q1mmih4e/Google-SK-UX-Enhancements10up?node-id=1198%3A42)), including the legend below.
* The chart should use real report data and display a tooltip when one hovers over a slice.
* The chart should be redrawn with new data when another dimension tab is selected.
* If there are more than 5 different dimension values in the resulting report, the chart should include in the piechart only the 4 accounting for the largest amount, and all other values should be aggregated in a single "Others" slice.

## Implementation Brief

* Implement the `UserDimensionsPieChart` component for the new All Traffic widget following these instructions:
    * It should use `<GoogleChart/>` component to display the chart and should encompass the UI around the pie chart in the bottom right corner of the widget (see [Figma file](https://www.figma.com/file/NzvJTPFdqUVJC9Q1mmih4e/Google-SK-UX-Enhancements10up?node-id=1198%3A42)), including the legend below.
    * The component should select the current dimension from the `core/forms` store via `getValue( FORM_ALL_TRAFFIC_WIDGET, 'dimensionName' )`. If not set, the default should be `ga:channelGrouping`.
    * The piechart should display data from an Analytics report for the metric `ga:users` and the dimension `dimensionName` (see above), dates based on the current date range, and optionally the current entity URL, if there is one.
    * If there are more than 5 different dimension values in the resulting report, only the 4 accounting for the largest amount should be included in the piechart, and all other values should be aggregated in a single "Others" slice.
    * Hovering over a pie slice should show a tooltip exposing the name, percentage value, absolute value, and percentage change over previous period, like it is in the [Figma file](https://www.figma.com/file/NzvJTPFdqUVJC9Q1mmih4e/Google-SK-UX-Enhancements10up?node-id=1198%3A271). The exception to this is the "Others" slice:
[...]

PR #2532, opened 2020-12-15
Feature/2425 user dimensions pie chart

eugene-manuilov:
## Summary

Addresses issue #2425

## Relevant technical choices

* Had to rework the way how we load Google Charts to fix some race condition issues;
* Added event listener handler for the `select` event which will be needed in the #2426.

## Checklist


eugene-manuilov:
@googlebot I consent.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/UserDimensionsPieChart.js
```diff
@@ -20,12 +20,228 @@
...
+
+	const args = {
+		...dateRangeDates,
+		metrics: [ { expression: 'ga:users' } ],
+		dimensions: [ dimensionName ],
+		orderby: {
+			fieldName: 'ga:users',
+			sortOrder: 'DESCENDING',
+		},
+		limit: 4,
```
felixarntz, line 56: I think there's some room for improvement here: For the case where there are exactly five rows, all five rows should be shown. See the ACs:
> If there are more than 5 different dimension values in the resulting report, the chart should include in the piechart only the 4 accounting for the largest amount

To satisfy that requirement, I believe we'll need to use a limit of `6` here and then, if then handle limiting the number of pie slices to display via another option for `extractAnalyticsDataForPieChart`. We'd pass `5` to that (the default should be unlimited).
* If more than 5 entries and `withOthers` is `true`, show the first 4 entries plus the "Others" slice.
* Otherwise, show only the first 5 entries.

includes/Core/Assets/Assets.php
```diff
@@ -383,6 +384,12 @@ private function get_assets() {
 					},
 				)
 			),
+			new Google_Charts(
+				'googlesitekit-google-charts',
+				array(
+					'in_footer' => false,
+				)
+			),
```
felixarntz, line 392: I don't think it's worth introducing a class for this. We can define the details and pass them to a `Script` just like we do elsewhere. The `Script` class should also work for external assets.
```suggestion
			new Script(
				'googlesitekit-google-charts',
				array(
					'src'          => 'https://www.gstatic.com/charts/loader.js',
					'in_footer'    => false,
					'before_print' => function() {
						wp_add_inline_script(
							'googlesitekit-google-charts',
							'google.charts.load( "current", { packages: [ "corechart" ] } );'
						);
					},
				)
			),
```

includes/Core/Assets/External/Google_Charts.php
```diff
@@ -0,0 +1,47 @@
+<?php
```
felixarntz, line 1: See above, let's remove this class.

felixarntz (changes requested):
Awesome work here so far! Just a few points of feedback.

assets/js/modules/analytics/util/index.js
```diff
@@ -45,32 +45,67 @@ export * from './validation';
...
-	} );
+		const rowData = [ row.dimensions[ 0 ], percent ];
+		if ( withTooltips ) {
+			rowData.push( tooltipCallback( row, rowData ) );
+		}
+
+		dataMap.push( rowData );
+	}
+
+	if ( withOthers && others > 0.01 ) {
```
felixarntz, line 101: Based on my previous feedback, here we could rely on whether there are remaining rows instead of checking `others` being greater than 0.01.

felixarntz (changes requested):
@eugene-manuilov You'll also need to update the Storybook story for the widget here to pass some dummy data. Currently the story is broken.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/UserDimensionsPieChart.js
eugene-manuilov: Nice catch! Updated.

includes/Core/Assets/Assets.php
eugene-manuilov: Oh.. Nice! Updated :smile:

includes/Core/Assets/External/Google_Charts.php
eugene-manuilov: Updated.

assets/js/modules/analytics/util/index.js
eugene-manuilov: Updated.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/UserDimensionsPieChart.js
```diff
@@ -20,12 +20,229 @@
...
+import { numberFormat, sanitizeHTML } from '../../../../../util';
+import { extractAnalyticsDataForPieChart } from '../../../util';
+import GoogleChart from '../../../../../components/GoogleChart';
+import PreviewBlock from '../../../../../components/PreviewBlock';
+import ReportError from '../../../../../components/ReportError';
+const { useSelect } = Data;
+
+export default function UserDimensionsPieChart( { dimensionName, entityURL, sourceLink } ) {
+	const [ chartLoaded, setChartLoaded ] = useState( false );
+	const dateRangeDates = useSelect( ( select ) => select( CORE_USER ).getDateRangeDates( { compare: true } ) );
```
felixarntz, line 46: There is kind of a dependency on #2236 here. We'll also need to pass the Analytics-specific offset of 1 here, otherwise it will request data up to the current day, which is not available. Let's do it hard-coded for now and update it to use the constant that #2236 will introduce as part of that.

assets/js/modules/analytics/datastore/__fixtures__/dashboard-user-dimensions-args.json
```diff
@@ -0,0 +1,59 @@
+{
+	"ga:channelGrouping": {
+		"startDate": "2020-12-09",
+		"endDate": "2021-01-05",
+		"compareStartDate": "2020-11-11",
+		"compareEndDate": "2020-12-08",
```
felixarntz, line 6: These dates will only work for a specific `referenceDate`, so that needs to be passed in the Storybook story. In combination with my comment above (the offset of `1`), the dates here will only work with a reference date of `2021-01-06`. Either way, we'll need to pass `referenceDate`, otherwise the story will break tomorrow :)

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/UserDimensionsPieChart.js
```diff
@@ -20,12 +20,229 @@
...
+				}
+			} );
+		}
+	}, [] );
+
+	if ( error ) {
+		return <ReportError moduleSlug="analytics" error={ error } />;
+	}
+
+	if ( ! loaded || ! report?.length ) {
```
felixarntz, line 91: This condition should only check for `! loaded` and be placed above the `error` condition.

Then, we should after the `error` condition have another `isZeroReport( report )` condition and in that case return a `ReportZero` component instance, like in the other widgets.

felixarntz (changes requested):
There are only a few remaining problems and one behavioral feedback here, but they're fairly minor, so I'll push an update here to get this merged.

felixarntz (approved):
Awesome, good to go!

Issue #2080, opened 2020-09-18
Allow modules to control their incomplete setup settings view UI through registered `settingsComponent`


felixarntz:
The capabilities for a module's registered `settingsComponent` should be expanded so that that component can also optionally specify a sub-component for the module's "setup incomplete" state (i.e. if the module is `active` but not `connected`). If none is provided, a `DefaultSetupIncomplete` component should be rendered. Site Kit's `SettingsModule` component should then be updated to render the registered `settingsComponent` even if the module is not connected so that full control is given to that component. The existing `googlesitekit.ModuleSetupIncomplete` filter should be removed.

All modules that have a custom "setup incomplete" component (currently only AdSense) should be updated to register that component via their registered `settingsComponent` instead of relying on the removed `googlesitekit.ModuleSetupIncomplete` filter.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `ModuleSetupIncomplete` component should be changed into a `DefaultSettingsSetupIncomplete` component. It should be changed to a functional component using the datastore (i.e. no more `getReAuthURL` usage), and it should no longer be filtered.
* The `registerModule` action of `core/modules` should be expanded to support an optional `SettingsSetupIncompleteComponent`.
* The `SettingsRenderer` component should be updated to also cover rendering `SettingsSetupIncompleteComponent` of the module as needed: If the settings panel `isOpen` and the module is not connected, that component should be rendered instead of `SettingsEdit` or `SettingsView`. If the module doesn't have a `SettingsSetupIncomplete`, the `DefaultSettingsSetupIncomplete` component should be rendered passing the module slug as `slug`.
* The `SettingsModule` part where it renders the legacy `ModuleSetupIncomplete` component can be removed since that will now be encompassed by `SettingsRenderer`. The `setupComplete` check around that can be removed accordingly.
* The AdSense module should be updated to no longer rely on the `googlesitekit.ModuleSetupIncomplete` filter, but instead specify that component in the module registration via `SettingsSetupIncompleteComponent`.
* Storybook should be updated so that the AdSense story for incomplete setup still works as before with the new API.

## Implementation Brief

* Remove the code from `SettingsModule` which renders `ModuleSetupIncomplete` component.
* Refactor and rename `ModuleSetupIncomplete` to `DefaultSettingsSetupIncomplete`.
	* Rename file name to `DefaultSettingsSetupIncomplete.js`
[...]

PR #2534, opened 2020-12-16
Introduce SettingsSetupIncompleteComponent to module registrations.

kostyalmm:
## Summary

Addresses issue #2080

## Relevant technical choices

* Refactor `ModuleSetupIncomplete` to `DefaultSettingsSetupIncomplete`
* Remove filter for incomplete setup
* Update `SettingsRenderer` accordingly

## Checklist


assets/js/components/settings/DefaultSettingsSetupIncomplete.js
```diff
@@ -0,0 +1,63 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * WordPress dependencies
+ */
+import { createInterpolateElement } from '@wordpress/element';
+import { __, sprintf } from '@wordpress/i18n';
+import PropTypes from 'prop-types';
```
eugene-manuilov, line 24: This is not a WordPress dependency. Please, move it to the external dependencies section.

assets/js/components/settings/DefaultSettingsSetupIncomplete.js
```diff
@@ -0,0 +1,63 @@
...
+import Link from '../Link';
+import ModuleSettingsWarning from '../legacy-notifications/module-settings-warning';
+const { useSelect } = Data;
+
+export default function DefaultSettingsSetupIncomplete( { slug } ) {
+	const storeName = `modules/${ slug }`;
+	const adminReauthURL = useSelect( ( select ) => select( storeName )?.getAdminReauthURL?.() );
+
+	return (
+		<div className="mdc-layout-grid__cell mdc-layout-grid__cell--span-12">
```
eugene-manuilov, line 39: Use the `<Cell>` component:

```suggestion
		<Cell size={ 12 }>
```

assets/js/components/settings/DefaultSettingsSetupIncomplete.js
```diff
@@ -0,0 +1,63 @@
...
+				{
+					a: <Link
+						className="googlesitekit-settings-module__edit-button"
+						onClick={ () => {
+							global.location = adminReauthURL;
+						} }
+						inherit
+					/>,
+				}
+			) }
```
eugene-manuilov, line 56: Let's use another approach because the current one will be hard to translate correctly. See how we do it in the `ErrorText` component and do similar things here:

https://github.com/google/site-kit-wp/blob/6f6ea93b7971159bc506b8d6decaf0af1c4c59f2/assets/js/components/ErrorText.js#L46-L63

Use empty <A> tag for translation and replace it with a tag that has the correct `href` attribute afterward:

```javascript
const message = __( 'Setup incomplete: <a>continue module setup</a>', 'google-site-kit' );
const html = message.replace( '<a>', `<a href=${ adminReauthURL }>` );
```

assets/js/components/settings/settings-module.js
```diff
@@ -323,26 +322,16 @@ class SettingsModule extends Component {
...
-									{
-										hasSettings && ! setupComplete &&
-											<ModuleSetupIncomplete
-												slug={ slug }
-											/>
-									}
+									<div className="
+										mdc-layout-grid__cell
+										mdc-layout-grid__cell--span-12
+									">
```
eugene-manuilov, line 328: Use `<Cell>` component instead. Also, please, update `<div className="mdc-layout-grid">` to be `<Grid>` and `<div className="mdc-layout-grid__inner">` to be `<Row>`

tests/js/utils.js
```diff
@@ -28,7 +28,7 @@ import modulesAdSenseStore from '../../assets/js/modules/adsense/datastore';
 import { STORE_NAME as MODULES_ADSENSE } from '../../assets/js/modules/adsense/datastore/constants';
 import { SetupMain as AdSenseSetupMain } from '../../assets/js/modules/adsense/components/setup';
 import {
-	SettingsEdit as AdSenseSettingsEdit,
+	SettingsEdit as AdSenseSettingsEdit, SettingsSetupIncomplete,
```
eugene-manuilov, line 31:
```suggestion
	SettingsEdit as AdSenseSettingsEdit,
	SettingsSetupIncomplete,
```

eugene-manuilov (changes requested):
Thanks, @kostyalmm! Found a few things to address. Could you please check it?

assets/js/components/settings/DefaultSettingsSetupIncomplete.js
kostyalmm: @eugene-manuilov, I used `createInterpolateElement` as it seemed to be a better approach than `.replace` and seemed to be a standard across the project.
Let me know if it is alright.

eugene-manuilov: It works! Thanks, @kostyalmm!

assets/js/components/settings/DefaultSettingsSetupIncomplete.js
```diff
@@ -0,0 +1,64 @@
...
+			{ createInterpolateElement(
+				__( 'Setup incomplete: <a>continue module setup</a>', 'google-site-kit' ),
+				{
+					a: <Link
+						className="googlesitekit-settings-module__edit-button"
+						onClick={ () => {
+							global.location = adminReauthURL;
+						} }
+						inherit
+					/>,
```
eugene-manuilov, line 55: I think it can be shortened to the following:

```suggestion
					a: <Link href={ adminReauthURL } inherit />,
```

eugene-manuilov (changes requested):
LGTM, @kostyalmm! However, I think there is still one thing that we can improve before I can approve this PR. @kostyalmm could you please address it?

eugene-manuilov (approved):
Awesome! Thanks, @kostyalmm!

assets/js/components/settings/SettingsRenderer.js
```diff
@@ -26,13 +26,26 @@ import { useEffect } from '@wordpress/element';
...
+		SettingsEditComponent,
+		SettingsViewComponent,
+		SettingsSetupIncompleteComponent,
+		doesModuleExist,
+		connected,
+	} = useSelect( ( select ) => {
+		const module = select( CORE_MODULES ).getModule( slug );
+		return {
+			...module,
+			doesModuleExist: !! module,
```
aaemnnosttv, line 46: I like the idea to return this as a named value but the name is not quite accurate.
A module that doesn't exist here could be either a module that hasn't been registered/loaded yet, or a module that really doesn't exist (will not be registered).

In this case I think it would be more accurate for our use to name this something like `moduleLoaded` instead.
```suggestion
			moduleLoaded: !! module,
```

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -225,18 +225,19 @@ const baseActions = {
...
+	 * @param {string}      [settings.name]                             Optional. Module name. Default is the slug.
+	 * @param {string}      [settings.description]                      Optional. Module description. Default empty string.
+	 * @param {WPComponent} [settings.Icon]                             Optional. React component to render module icon. Default none.
+	 * @param {number}      [settings.order]                            Optional. Numeric indicator for module order. Default 10.
+	 * @param {string}      [settings.homepage]                         Optional. Module homepage URL. Default empty string.
+	 * @param {WPComponent} [settings.SettingsEditComponent]            Optional. React component to render the settings edit panel. Default none.
+	 * @param {WPComponent} [settings.SettingsViewComponent]            Optional. React component to render the settings view panel. Default none.
+	 * @param {WPComponent} [settings.SetupComponent]                   Optional. React component to render the setup panel. Default none.
+	 * @param {WPComponent} [settings.SettingsSetupIncompleteComponent] Optional. React component to render the incomplete settings panel. Default none.
```
aaemnnosttv, line 239: Let's move this param up one line so that it's grouped with the other `settings.Settings*` params.

Per other comments, the default should be `DefaultSettingsSetupIncomplete`.

assets/js/components/settings/SettingsRenderer.js
```diff
@@ -42,8 +55,13 @@ export default function SettingsRenderer( { slug, isOpen, isEditing } ) {
...
 	}, [ rollbackSettings, haveSettingsChanged, isDoingSubmitChanges, isEditing ] );
 
-	if ( ! isOpen ) {
+	if ( ! isOpen || ! doesModuleExist ) {
 		return null;
+	} else if ( isOpen && ! connected ) {
+		if ( !! SettingsSetupIncompleteComponent ) {
+			return <SettingsSetupIncompleteComponent />;
+		}
+		return <DefaultSettingsSetupIncomplete slug={ slug } />;
```
aaemnnosttv, line 64: We can simplify this to 
```suggestion
		return <SettingsSetupIncompleteComponent slug={ slug } />;
```

We only need to define `moduleDefaults.SettingsSetupIncompleteComponent = DefaultSettingsSetupIncomplete` in `modules.js` (see comment below).

This is slightly different than the IB but I think we should apply defaults consistently where possible and there's no reason why we can't pass `slug` to both components.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -262,6 +264,7 @@ const baseActions = {
 			SettingsEditComponent,
 			SettingsViewComponent,
 			SetupComponent,
+			SettingsSetupIncompleteComponent,
```
aaemnnosttv, line 267: This is missing a definition in `moduleDefaults` above:

https://github.com/google/site-kit-wp/blob/8c37f05d1f1a951f91292de6ec52209a3a82a956/assets/js/googlesitekit/modules/datastore/modules.js#L67-L68

This can be set to the `DefaultSettingsSetupIncomplete` component type as long as we pass the `slug` unconditionally in `SettingsRenderer`.

aaemnnosttv (changes requested):
Thanks @kostyalmm – this is looking good, just a few comments to address.
Overall looking good though 👍

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -248,6 +250,7 @@ const baseActions = {
 		SettingsEditComponent,
 		SettingsViewComponent,
 		SetupComponent,
+		SettingsSetupIncompleteComponent = DefaultSettingsSetupIncomplete,
```
aaemnnosttv, line 253: This assignment belongs in the [`moduleDefaults`](https://github.com/google/site-kit-wp/blob/8c37f05d1f1a951f91292de6ec52209a3a82a956/assets/js/googlesitekit/modules/datastore/modules.js#L54) variable above, not as a default to the action.

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -225,18 +226,19 @@ const baseActions = {
...
+	 * @param {string}      [settings.name]                             Optional. Module name. Default is the slug.
+	 * @param {string}      [settings.description]                      Optional. Module description. Default empty string.
+	 * @param {WPComponent} [settings.Icon]                             Optional. React component to render module icon. Default none.
+	 * @param {number}      [settings.order]                            Optional. Numeric indicator for module order. Default 10.
+	 * @param {string}      [settings.homepage]                         Optional. Module homepage URL. Default empty string.
+	 * @param {WPComponent} [settings.SettingsEditComponent]            Optional. React component to render the settings edit panel. Default none.
+	 * @param {WPComponent} [settings.SettingsViewComponent]            Optional. React component to render the settings view panel. Default none.
+	 * @param {WPComponent} [settings.SettingsSetupIncompleteComponent] Optional. React component to render the incomplete settings panel. Default DefaultSettingsSetupIncomplete.
```
aaemnnosttv, line 239: Sorry, I think I was incorrect here which may have been confusing as to where this goes. The default here should remain "none" as it was, as there is no default _parameter value_ – the default is only applied when the module is retrieved, not on registration.

aaemnnosttv (commented):
Thanks @kostyalmm – this is very close just a few remaining tweaks.

aaemnnosttv (approved):
Great, thanks!

Issue #2252, opened 2020-10-22
Combine similar CTA components returned from widgets


felixarntz:
Widgets may return certain predefined components to indicate certain states:
* `ReportError`
* `ReportZero`
* `ActivateModuleCTA`
* `CompleteModuleActivationCTA`

The last three of those are module-specific and will result in the same UI for each instance they are displayed in (for the same module). If multiple widgets next to each other return the same of those components (i.e. same component with same props), such components should be aggregated and only displayed once. For example, if the Analytics module is not active, only a single `ActivateModuleCTA` for Analytics should be displayed in the "Search Funnel" widget area instead of two, even though it results from two different widgets.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Multiple instances of `ReportZero`, `ActivateModuleCTA`, and `CompleteModuleActivationCTA` returned by adjacent widgets in the same row should be combined per the module that they apply to.
    * The widget widths of widgets combined should be combined. For example, three quarter widgets returning the same component for the same module should result in a single three-quarter width version of that component being rendered.
    * For example, if two quarter adjacent widgets return `ReportZero` for Search Console and the next two quarter widgets return `ActivateModuleCTA` for Analytics, they should be combined to be one half `ReportZero` for Search Console and one half `ActivateModuleCTA` for Analytics.
    * Components should not be combined if they would span across multiple rows. For example, if in the above scenario there was another two quarter widgets returning `ActivateModuleCTA` for Analytics, they should be combined with each other, but not with the previous combination of two `ActivateModuleCTA` for Analytics because this would mean they would be combined across rows, which would mess up the UI.
* Widgets should not have to set state for this to work. They should only have to return the specific comments, and those components should internally support handling the necessary details (e.g. via state and `useEffect` or similar).
* There should be Storybook coverage for widgets where some of them return certain special state components, some in ways that would be combined etc., to test the behavior. The following stories in particular should be included:
    1. 4 quarter widgets: Normal, `ReportZeroData` Search Console, `ReportZeroData` Analytics, `ActivateModuleCTA` AdSense
    2. 4 quarter widgets: `ReportZeroData` Search Console, `ReportZeroData` Search Console, `ReportZeroData` Analytics, `ReportZeroData` Analytics
    3. 4 half widgets: `ReportZeroData` Search Console, `ActivateModuleCTA` Analytics, `ActivateModuleCTA` Analytics, `ActivateModuleCTA` AdSense
[...]

PR #2538, opened 2020-12-17
Combine similar widget CTAs

felixarntz:
## Summary

Addresses issue #2252

## Checklist


felixarntz:
@aaemnnosttv Ready for review, except for missing test coverage.

felixarntz:
@aaemnnosttv This is now ready for full code review.

assets/js/googlesitekit/widgets/util/constants.js
```diff
@@ -0,0 +1,46 @@
+/**
+ * Widgets layout constants.
+ *
+ * Site Kit by Google, Copyright 2020 Google LLC
```
aaemnnosttv, line 4:
```suggestion
 * Site Kit by Google, Copyright 2021 Google LLC
```

assets/js/googlesitekit/widgets/util/get-widget-component-props.js
```diff
@@ -0,0 +1,48 @@
...
+	return {
+		WidgetReportZero: withWidgetSlug( widgetSlug )( WidgetReportZero ),
+		WidgetActivateModuleCTA: withWidgetSlug( widgetSlug )( WidgetActivateModuleCTA ),
+		WidgetCompleteModuleActivationCTA: withWidgetSlug( widgetSlug )( WidgetCompleteModuleActivationCTA ),
+	};
+}
+
+function withWidgetSlug( widgetSlug ) {
+	return ( WrappedComponent ) => {
+		return ( props ) => <WrappedComponent widgetSlug={ widgetSlug } { ...props } />;
```
aaemnnosttv, line 46: Should we not pass `widgetSlug` after the props spread here? My original thought was that the scoped components would not be able to pass a different slug (or at least it wouldn't make a difference if it did). I can't think of a reason why a widget would want/need to use a different slug for these components. Since it's intended to only be used within that widget, I think it should be fixed to only work for that widget.
```suggestion
		return ( props ) => <WrappedComponent { ...props } widgetSlug={ widgetSlug } />;
```

assets/js/googlesitekit/widgets/util/get-widget-component-props.js
```diff
@@ -0,0 +1,48 @@
...
+	// component does not need to (re-)specify the widget slug.
+	return {
+		WidgetReportZero: withWidgetSlug( widgetSlug )( WidgetReportZero ),
+		WidgetActivateModuleCTA: withWidgetSlug( widgetSlug )( WidgetActivateModuleCTA ),
+		WidgetCompleteModuleActivationCTA: withWidgetSlug( widgetSlug )( WidgetCompleteModuleActivationCTA ),
+	};
+}
+
+function withWidgetSlug( widgetSlug ) {
+	return ( WrappedComponent ) => {
```
aaemnnosttv, line 45: Would be nice to apply a proper [`displayName`](https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging) to these.

assets/js/googlesitekit/widgets/util/get-widget-layout.js
```diff
@@ -19,49 +19,29 @@
...
+ * @param {Array.<number>} columnWidths Current column widths.
+ * @param {number}         counter      Current counter.
+ * @return {Array} Array where the first element is the updated list of class
+ *                 names, the second element is the updated list of column
+ *                 widths, and the third element is the resulting counter
+ *                 after the update.
  */
-function resizeClasses( classNames, counter ) {
+function resizeClasses( classNames, columnWidths, counter ) {
 	// Safeguard: counter must always be 9 for this to work.
```
aaemnnosttv, line 38: I've seen this before and it still seems like an odd condition. Would be great if you could elaborate on this a little bit to explain 😄

assets/js/googlesitekit/widgets/util/get-widget-layout.js
```diff
@@ -19,49 +19,29 @@
...
 	if ( counter !== 9 ) {
-		return [ classNames, counter ];
+		return [ classNames, columnWidths, counter ];
```
aaemnnosttv, line 40: Maybe we should change this to return an object now that we're adding another value to the return?

assets/js/googlesitekit/widgets/util/get-widget-layout.js
```diff
@@ -19,49 +19,29 @@
...
+ * @param {Array.<string>} classNames   Current class names.
+ * @param {Array.<number>} columnWidths Current column widths.
+ * @param {number}         counter      Current counter.
+ * @return {Array} Array where the first element is the updated list of class
+ *                 names, the second element is the updated list of column
+ *                 widths, and the third element is the resulting counter
+ *                 after the update.
  */
-function resizeClasses( classNames, counter ) {
+function resizeClasses( classNames, columnWidths, counter ) {
```
aaemnnosttv, line 37: Similar to a previous comment, it's a bit odd that this function handles grid classNames rather than grid cell sizes/props.

Maybe not worth changing right now but it seems like this would be a little less complicated if it were only dealing with widths rather than widths and class names.

assets/js/googlesitekit/widgets/util/get-widget-layout.js
```diff
@@ -156,27 +141,36 @@ export function getWidgetClassNames( activeWidgets ) {
...
 			// the widths of these widgets slightly to fill the entire 12 columns.
 			if ( counter === 9 ) {
-				[ classNames, counter ] = resizeClasses( classNames, counter );
+				[ classNames, columnWidths, counter ] = resizeClasses( classNames, columnWidths, counter );
 			}
 
 			// See above, initial counter for the next row of widgets.
 			counter = WIDTH_GRID_COUNTER_MAP[ width ];
+			rowIndex++;
+		} else if ( counter % 12 === 0 ) {
```
aaemnnosttv, line 170: I think this can be simplified since the condition for greater than 12 is right before this.
```suggestion
		} else if ( counter === 12 ) {
```

assets/js/googlesitekit/widgets/util/get-widget-layout.test.js
```diff
@@ -0,0 +1,129 @@
+/**
+ * Widgets layout utilities tests.
+ *
+ * Site Kit by Google, Copyright 2020 Google LLC
```
aaemnnosttv, line 4:
```suggestion
 * Site Kit by Google, Copyright 2021 Google LLC
```

assets/js/googlesitekit/widgets/util/get-widget-layout.test.js
```diff
@@ -0,0 +1,129 @@
...
+				'mdc-layout-grid__cell',
+				'mdc-layout-grid__cell--span-2-phone',
+				'mdc-layout-grid__cell--span-3-desktop',
+				'mdc-layout-grid__cell--span-4-tablet',
+			],
+			[
+				'mdc-layout-grid__cell',
+				'mdc-layout-grid__cell--span-2-phone',
+				'mdc-layout-grid__cell--span-3-desktop',
+				'mdc-layout-grid__cell--span-4-tablet',
```
aaemnnosttv, line 53: This looks odd – these two quarter widgets have grid classes for a span of 2 in mobile, which would be 1/6 and a span of 4 in tablet (1/3). Is that correct? I would expect the grid classes to increase in width as the viewport gets smaller but tablet is growing while phone is shrinking.

assets/js/googlesitekit/widgets/util/get-widget-layout.test.js
```diff
@@ -0,0 +1,129 @@
...
+			[
+				'mdc-layout-grid__cell',
+				'mdc-layout-grid__cell--span-6-desktop',
+				'mdc-layout-grid__cell--span-8-tablet',
+			],
+			[
+				'mdc-layout-grid__cell',
+				'mdc-layout-grid__cell--span-8-desktop',
+				'mdc-layout-grid__cell--span-8-tablet',
+			],
```
aaemnnosttv, line 64: I'm starting to see what's going on but it is rather complicated. Would you please add comments as to why the various classes are expected for each?

assets/sass/widgets/_widget-area.scss
```diff
@@ -41,6 +41,6 @@
 	}
 }
 
-.googlesitekit-widget-area--hidden {
+.googlesitekit-widget-grid-hidden {
```
aaemnnosttv, line 44: If we do use a class for this, why not use a generic display utility instead? We shouldn't really need to add a new class for hiding in each context where it's needed.

stories/activate-module.stories.js
```diff
@@ -57,4 +57,22 @@ storiesOf( 'Global', module )
...
+		const setupRegistry = ( { dispatch } ) => {
+			dispatch( CORE_USER ).receiveCapabilities( {
+				[ PERMISSION_MANAGE_OPTIONS ]: true,
+			} );
+			dispatch( CORE_MODULES ).receiveGetModules( [
+				{
+					slug: 'analytics',
+					name: 'Analytics',
+				},
+			] );
```
aaemnnosttv, line 71: How about using these instead?
```js
provideUserCapabilities( registry, { ... } );
provideModules( registry );
```

aaemnnosttv (changes requested):
Great job @felixarntz – this is a huge change and it looks great to see all of the components updated!

I left a number of comments through out and a few questions. Nothing major to call out, overall a very solid PR. The one thing that comes to mind is more related to tests and not wanting to couple the tests for utilities to specific components or other implementation details.

My only other thought (probably beyond the scope of this PR) is more around decoupling the logic from classnames and leaving those to the responsibility of the `Cell` component, so that this stuff can focus solely on the numbers.

Overall, very nice and big props for tackling this beast!

assets/js/googlesitekit/widgets/components/WidgetActivateModuleCTA.test.js
```diff
@@ -0,0 +1,77 @@
...
+
+	beforeAll( () => {
+		registry = createTestRegistry();
+		provideModules( registry );
+		provideUserCapabilities( registry );
+	} );
+
+	it( 'sets widget state when rendered and unsets when unmounted', () => {
+		const widgetSlug = 'testWidget';
+		const moduleSlug = 'analytics';
```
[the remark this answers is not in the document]
felixarntz: The `ActivateModuleCTA` component relies on the module identified by the `moduleSlug` being available, so I think we should ensure that. We could also use `testModule` and manually register a module with that slug, but I think it's more straightforward to rely on `provideModules` and use real slugs.

assets/js/googlesitekit/widgets/components/WidgetActivateModuleCTA.test.js
```diff
@@ -0,0 +1,77 @@
...
+		render( <WidgetActivateModuleCTA widgetSlug={ widgetSlug } moduleSlug={ moduleSlug } extraProp="propValue" />, { registry } );
+
+		expect( registry.select( CORE_WIDGETS ).getWidgetState( widgetSlug ) ).toMatchObject( {
+			Component: ActivateModuleCTA,
+			metadata: { moduleSlug },
+		} );
+	} );
+
+	it( 'renders the same output as ActivateModuleCTA with the same props (except widgetSlug)', () => {
+		const props = { moduleSlug: 'analytics', description: 'Activate the module!' };
```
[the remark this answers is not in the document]
felixarntz: See above.

assets/js/googlesitekit/widgets/components/WidgetActivateModuleCTA.js
```diff
@@ -0,0 +1,56 @@
+/**
+ * WidgetActivateModuleCTA component.
+ *
+ * Site Kit by Google, Copyright 2020 Google LLC
```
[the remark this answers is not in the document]
felixarntz: Replaced throughout.

assets/js/googlesitekit/widgets/components/WidgetActivateModuleCTA.js
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Internal dependencies
+ */
+import Data from 'googlesitekit-data';
+import { STORE_NAME } from '../datastore/constants';
+import ActivateModuleCTA from '../../../components/ActivateModuleCTA';
+
+const { useDispatch } = Data;
+
+// The supported props must match `ActivateModuleCTA` (except `widgetSlug`).
```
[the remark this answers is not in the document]
felixarntz: Updated, also for the two other `Widget*` components.

assets/js/googlesitekit/widgets/components/WidgetRenderer.js
```diff
@@ -21,30 +21,57 @@
...
+		// It always needs to be wrapped as it is expected to be a
+		// widget-agnostic component.
+		// The real widget component will still be rendered, but it will be
+		// hidden via CSS.
+		widgetComponent = (
+			<Fragment>
+				<Widget slug="overridden">
+					<OverrideComponent />
+				</Widget>
+				<div className="googlesitekit-widget-grid-hidden">
```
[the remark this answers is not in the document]
felixarntz: Not sure about this. Is there a benefit of using the `template` tag instead? While using CSS is a bit hacky, to me `template` feels just as hacky. If we could avoid it, nothing of this should be rendered at all, but that wouldn't work with our approach.

assets/js/googlesitekit/widgets/util/combine-widgets.js
```diff
@@ -0,0 +1,119 @@
...
+function stateAndRowMatch( stateA, stateB, rowA, rowB ) {
+	return isEqual( stateA, stateB ) && rowA === rowB;
+}
+
+/**
+ * Combines consecutive widgets with similar states within the same row.
+ *
+ * @since n.e.x.t
+ *
+ * @param {Array.<Object>} activeWidgets       List of active widgets.
```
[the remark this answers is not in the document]
felixarntz: How would we deal with the array of arrays in that syntax? Might feel a bit odd.

assets/js/googlesitekit/widgets/util/combine-widgets.js
```diff
@@ -0,0 +1,119 @@
...
+ *                                             `Component` and `metadata`, or
+ *                                             `null`).
+ * @param {Object}         layout              Layout arguments from
+ *                                             	`getWidgetLayout()`.
+ * @param {Array.<Array>}  layout.classNames   List of class name arrays for
+ *                                             each widget.
+ * @param {Array.<number>} layout.columnWidths List of column widths for each
+ *                                             widget.
+ * @param {Array.<number>} layout.rowIndexes   List of row indexes for each
+ *                                             widget.
```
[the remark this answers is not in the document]
felixarntz: I use 80 typically, but since we haven't defined that anywhere we can also revisit that.

assets/js/googlesitekit/widgets/util/combine-widgets.js
```diff
@@ -0,0 +1,119 @@
...
+				// Get total (desktop) column width and use corresponding grid
+				// classes. For tablet and phone, the component should span the
+				// full width as by definition it is at least a "half" widget
+				// wide (which has that behavior).
+				const combinedColumnWidth = columnWidthsBuffer.reduce( ( sum, columnWidth ) => sum + columnWidth, 0 );
+
+				gridClassNames[ i ] = [
+					'mdc-layout-grid__cell',
+					`mdc-layout-grid__cell--span-${ combinedColumnWidth }`,
+				];
```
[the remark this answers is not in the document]
felixarntz: That's a good point. However, the concept of passing around class names like this has already been there before this PR to compose the widget area layout. Changing it in this PR would make it even more complex and potentially result in a delay. I'll create a follow-up issue to refactor that out, I agree it's a simplification worth adding. Since these functions and components are only used within Site Kit, changing it will be okay even when the Widgets API is production-ready.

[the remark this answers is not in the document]
felixarntz: #2671

assets/js/googlesitekit/widgets/util/combine-widgets.test.js
```diff
@@ -0,0 +1,183 @@
...
+import CompleteModuleActivationCTA from '../../../components/CompleteModuleActivationCTA';
+
+describe( 'combineWidgets', () => {
+	const getQuarterWidget = ( slug ) => ( { slug, width: WIDGET_WIDTHS.QUARTER } );
+	const getHalfWidget = ( slug ) => ( { slug, width: WIDGET_WIDTHS.HALF } );
+
+	const getRegularState = () => null;
+	const getReportZeroState = ( moduleSlug ) => ( { Component: ReportZero, metadata: { moduleSlug } } );
+	const getActivateModuleCTAState = ( moduleSlug ) => ( { Component: ActivateModuleCTA, metadata: { moduleSlug } } );
+	const getCompleteModuleActivationCTAState = ( moduleSlug ) => ( { Component: CompleteModuleActivationCTA, metadata: { moduleSlug } } );
```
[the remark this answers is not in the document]
felixarntz: This is somewhat related to my response in https://github.com/google/site-kit-wp/pull/2538#discussion_r559516608. I'm not sure we should change it because these tests match the same ones in Storybook and thus are using "real" UI components, which I think better clarifies the intent. Maybe we could add another test case at the top that uses "random" components, but that also wouldn't really improve coverage.

assets/js/googlesitekit/widgets/util/combine-widgets.test.js
```diff
@@ -0,0 +1,183 @@
...
+
+	// Every test case below corresponds to a matching story in `stories/widgets.stories.js` under
+	// "Global/Widgets/Widget Area/Special combination states".
```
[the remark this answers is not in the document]
felixarntz: I was more thinking that coupling them will ensure we keep the coverage consistent. We don't need to test more scenarios of similar order etc., so I don't think using random other combinations would help. The comments here IMO are a reminder that we should keep the scenarios aligned. If there is a new relevant scenario to be tested, it should be added to both the Jest test and the Storybook story.

assets/js/googlesitekit/widgets/util/get-widget-layout.js
felixarntz: Normally I would agree, but I think here it makes sense to keep it as array:
* The return value is basically a "filtered" version of each function parameter, so having them in an array is intuitive from that perspective.
* Destructuring the values into existing variables like we currently do from an array when calling `resizeClasses` work, but with an object it wouldn't be valid (I wasn't aware of this JS detail before). We would need to first assign it to an object and then write each property's value into the respective variable.

assets/js/googlesitekit/widgets/util/get-widget-layout.js
felixarntz: See above, that can be handled via #2671.

assets/js/googlesitekit/widgets/util/get-widget-layout.test.js
felixarntz: Yes that's correct as far as I can see. Only the desktop classes are dynamic at all. Tablet and phone classes are static per widget width declaration, and `QUARTER` widget should take half the viewport space in both tablet and phone view. The tablet viewport has 8 columns in total while the phone viewport has 4 columns in total, so using 4 and 2 respectively here is correct.

assets/js/googlesitekit/widgets/components/WidgetRenderer.js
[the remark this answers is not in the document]
aaemnnosttv: I think this is only a temporary work around until #2612 is merged, correct?

Regarding the use of `<template>` – it is automatically hidden by the browser (no CSS required) but also the browser does minimal parsing of the HTML within, so there's a bit of a performance optimization at the same time. It's probably not a huge difference though.

I think I mentioned this elsewhere here too, but I don't really think we need to add extra CSS classes just to `display: none`. For elements like this which will and should always remain hidden, what about adding a `Hidden` component?
```js
const Hidden = ( { children } ) => <div style={ { display: 'none' } }>{ children }</div>
```

I think using an inline style in this way would be okay I think.

assets/js/googlesitekit/widgets/util/combine-widgets.js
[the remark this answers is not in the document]
aaemnnosttv: I think that would be `Array[]` – I don't think we have any instances of that though? 🤔

assets/js/googlesitekit/widgets/util/combine-widgets.test.js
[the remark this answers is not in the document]
aaemnnosttv: I suppose that's okay in this case since the logic is very much visually-oriented. I don't think it's something that we want to maintain as a general convention however.

assets/js/googlesitekit/widgets/util/get-widget-layout.js
aaemnnosttv: I think you mean this:

image

If so, it is possible (and we do it in a few places in the codebase), it just requires the added parentheses 👍

assets/js/googlesitekit/widgets/util/get-widget-layout.test.js
aaemnnosttv: Yeah I didn't realize before that the total columns is different per viewport; I just learned this when looking at a different issue. Thanks!

aaemnnosttv (approved):
Awesome! 

I left a few replies to some of the feedback from the previous round, but nothing blocking.

Great work!

Issue #2076, opened 2020-09-18
Update admin bar menu to functional component using datastore


felixarntz:
The admin bar menu for Site Kit should be refactored to become a functional component relying on the datastore. It should not rely on the Widget API at this point and not be extensible in any way. All sub-components of the admin bar menu should be part of "Site Kit core" (i.e. in `assets/js/components`), even those relying on certain modules (`assets/js/modules` should only contain files that are relevant for the respective module JS entrypoints, i.e. register things in Site Kit core).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A feature flag `widgets.adminBar` should be introduced. (While this won't for now use the actual Widgets API, it still underlies several similar requirements.)
* Individual components (using hooks and datastore exclusively) should be added in `assets/js/components/adminbar` for all components that are added via the `googlesitekit.AdminbarModules` filter:
    * `AdminBarUniqueVisitors` and `AdminBarSessions` (from legacy component `AnalyticsAdminbarWidgetOverview`)
    * `AdminBarClicks` and `AdminBarImpressions` (from legacy component `SearchConsoleAdminbarWidgetOverview`)
    * See the existing `Dashboard*Widget` components which will have most of the code for these already included, so a lot is available for copy-paste.
* If the above feature flag is enabled, `AdminBarApp` should render the above new components right above the legacy `AdminBarModules` component, in the same order in which they're displayed now.
    * If the Analytics module is not active, instead of the `AdminBarUniqueVisitors` and `AdminBarSessions` components, the `AnalyticsInactiveCTA` should be displayed, similarly to today.
* `AdminBarModules` should be renamed to `LegacyAdminBarModules`.
* All files in `assets/js/components/adminbar` should have names that match the component name that they export.

## Implementation Brief

* Add a feature flag in `feature-flags.json` `widgets: { adminBar: { enabled: 'development' }`

* Create four new components:
* `AdminBarUniqueVisitors` `assets/js/components/adminbar/AdminBarUniqueVisitors.js` code should be a functional component based on https://github.com/google/site-kit-wp/blob/39a79205909d539f2cf2b7bcbfc07d2cce90f10f/assets/js/modules/analytics/components/adminbar/AnalyticsAdminbarWidgetOverview.js#L95-L107

[...]

PR #2543, opened 2020-12-17
Refactor AdminBar into functional components.

asvinb:
## Summary

Addresses issue #2076 

## Checklist


feature-flags.json
```diff
@@ -5,6 +5,9 @@
     },
     "pageDashboard": {
       "enabled": "development"
+    },
+    "adminBar": {
+      "enabled": "development"
```
eugene-manuilov, line 10: @asvinb is it widget related feature? If not, it should be declared outside of the widgets scope.

asvinb: @eugene-manuilov It's not a widget related feature, however as per the AC:
_While this won't for now use the actual Widgets API, it still underlies several similar requirements._

aaemnnosttv:
@asvinb is this ready for review?

assets/js/components/adminbar/AdminBarClicks.js
```diff
@@ -0,0 +1,83 @@
...
+	const searchConsoleData = useSelect( ( select ) => select( MODULES_SEARCH_CONSOLE ).getReport( reportArgs ) );
+
+	if ( undefined === searchConsoleData ) {
+		return <AdminBarPreview />;
+	}
+
+	const {
+		totalClicks,
+		totalClicksChange,
+	} = extractSearchConsoleDashboardData( searchConsoleData );
```
aaemnnosttv, line 60: We can't use this utility anymore since it is coupled to the format of the old return value which combines both date ranges into a single range in the returned report data.
See here for an example of how these are handled in a similar new component:
https://github.com/google/site-kit-wp/blob/6f6ea93b7971159bc506b8d6decaf0af1c4c59f2/assets/js/modules/search-console/components/dashboard/DashboardClicksWidget.js#L97-L99
Note that this (linked) component doesn't use true multi-date a range report either so the latest and older data need to be sourced a little differently from the returned report data.

assets/js/components/adminbar/AdminBarImpressions.js
```diff
@@ -0,0 +1,83 @@
...
+	const searchConsoleData = useSelect( ( select ) => select( MODULES_SEARCH_CONSOLE ).getReport( reportArgs ) );
+
+	if ( undefined === searchConsoleData ) {
+		return <AdminBarPreview />;
+	}
+
+	const {
+		totalImpressions,
+		totalImpressionsChange,
+	} = extractSearchConsoleDashboardData( searchConsoleData );
```
aaemnnosttv, line 60: Same here.

assets/js/components/adminbar/AdminBarSessions.js
```diff
@@ -0,0 +1,108 @@
...
+		],
+		url,
+	};
+	const analyticsData = useSelect( ( select ) => select( MODULES_ANALYTICS ).getReport( reportArgs ) );
+
+	if ( undefined === analyticsData ) {
+		return <AdminBarPreview />;
+	}
+
+	const { totalSessions, totalSessionsChange } = calculateOverviewData( analyticsData );
```
aaemnnosttv, line 85: I'm not sure if this utility expects separate date ranges in the response (it looks like it may) but we probably shouldn't use it anymore and parse this out in the component instead (only what we need though of course).

aaemnnosttv (commented):
@asvinb not sure if this is ready for review yet or not, but left a few comments here that may also explain why you're seeing different data with these changes.

asvinb:
> @asvinb is this ready for review?

Ready for review now @aaemnnosttv :)

assets/js/components/adminbar/AdminBarApp.js
```diff
@@ -75,7 +80,36 @@ export default function AdminBarApp() {
...
+												)
+										}
+
+										<AdminBarImpressions />
+										<AdminBarClicks />
+									</Fragment>
+								)
+								: (
+									<AdminbarModules />
+								)
```
aaemnnosttv, line 111: According to the ACs, the `AdminbarModules` would still render regardless of the feature flag value. Similar to the main dashboard – if the flag is enabled, the new widget version would render above _in addition_ to the legacy version below it.

assets/js/components/adminbar/AdminBarApp.js
```diff
@@ -25,20 +25,25 @@ import { __ } from '@wordpress/i18n';
...
 import { STORE_NAME as CORE_SITE } from '../../googlesitekit/datastore/site/constants';
+import { STORE_NAME as CORE_MODULES } from '../../googlesitekit/modules/datastore/constants';
+import { decodeHTMLEntity, trackEvent } from '../../util';
 const { useSelect } = Data;
 
 export default function AdminBarApp() {
 	const currentEntityURL = useSelect( ( select ) => select( CORE_SITE ).getCurrentEntityURL() );
 	const currentEntityTitle = useSelect( ( select ) => select( CORE_SITE ).getCurrentEntityTitle() );
 	const detailsURL = useSelect( ( select ) => select( CORE_SITE ).getAdminURL( 'googlesitekit-dashboard', { permaLink: currentEntityURL } ) );
+	const analyticsModuleActive = useSelect( ( select ) => select( CORE_MODULES ).isModuleActive( 'analytics' ) );
```
aaemnnosttv, line 45: We shouldn't need both of these selected since a module cannot be connected when inactive (but the inverse is possible).

```suggestion
```

assets/js/components/adminbar/AdminBarApp.js
```diff
@@ -75,7 +80,36 @@ export default function AdminBarApp() {
...
+												: (
+													<div className="
+														mdc-layout-grid__cell
+														mdc-layout-grid__cell--span-6-desktop
+														mdc-layout-grid__cell--span-4-tablet
+													">
+														<AnalyticsInactiveCTA />
+													</div>
+												)
+										}
```
aaemnnosttv, line 103: Let's simplify this to be less nested using basic conditions rather than ternary.
```js
{ analyticsModuleConnected && (
	<Fragment>
		<AdminBarUniqueVisitors />
		<AdminBarSessions />
	</Fragment>
) }
{ ! analyticsModuleConnected && (
	<div className="
		mdc-layout-grid__cell
		mdc-layout-grid__cell--span-6-desktop
		mdc-layout-grid__cell--span-4-tablet
	">
		<AnalyticsInactiveCTA />
	</div>
) }
```

assets/js/components/adminbar/AdminBarSessions.js
```diff
@@ -0,0 +1,116 @@
...
+				expression: 'ga:sessions',
+				alias: 'Sessions',
+			},
+			{
+				expression: 'ga:bounceRate',
+				alias: 'Bounce Rate',
+			},
+			{
+				expression: 'ga:avgSessionDuration',
+				alias: 'Average Session Duration',
```
aaemnnosttv, line 66: @felixarntz – is there any reason to pass these aliases in the request? I don't think we use them at all – certainly not for display since they're not translated, but just curious if there is any practical reason for this other than maybe inspectability in the response?

If only for us – that's a perfectly valid reason but it's slightly misleading in that it's formatted in a way you might think it's intended to be a user-facing string.

aaemnnosttv (changes requested):
Thanks @asvinb – this is looking pretty good, just a few things to address here and there. Some of the points apply to the same places in other components with the same structure, prop name, etc so just keep that in mind as you go through as I didn't point out every instance of a comment where it applies in multiple components. I did however try to point out where the comment has multiple applications 😄  

Let me know if you have any questions!

aaemnnosttv:
@asvinb also wanted to note that VRT is failing – looks like a timeout from a not-found selector at first glance.

assets/js/components/adminbar/AdminBarClicks.js
asvinb: Updated @aaemnnosttv

assets/js/components/adminbar/AdminBarSessions.js
asvinb: Updated @aaemnnosttv

assets/js/components/adminbar/AdminBarImpressions.js
asvinb: Updated @aaemnnosttv

assets/js/components/adminbar/AdminBarClicks.js
```diff
@@ -0,0 +1,90 @@
...
+				datapoint={ readableLargeNumber( totalClicks ) }
+				change={ totalClicksChange }
+				changeDataUnit="%"
+			/>
+		</div>
+	);
+};
+
+AdminBarClicks.propTypes = {
+	classNames: PropTypes.string,
```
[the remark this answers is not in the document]
asvinb: Updated @aaemnnosttv

assets/js/components/adminbar/AdminBarClicks.js
```diff
@@ -0,0 +1,90 @@
...
+};
+
+AdminBarClicks.defaultProps = {
+	classNames: 'mdc-layout-grid__cell mdc-layout-grid__cell--span-2-tablet mdc-layout-grid__cell--span-3-desktop',
```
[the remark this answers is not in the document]
asvinb: Updated @aaemnnosttv

assets/js/modules/analytics/datastore/report.js
```diff
@@ -122,6 +122,8 @@ const baseSelectors = {
 	 * @param {Object}         options                     Options for generating the report.
 	 * @param {string}         options.startDate           Required, unless dateRange is provided. Start date to query report data for as YYYY-mm-dd.
 	 * @param {string}         options.endDate             Required, unless dateRange is provided. End date to query report data for as YYYY-mm-dd.
+	 * @param {string}         options.compareStartDate    Optional. Start date to compare report data for as YYYY-mm-dd.
+	 * @param {string}         options.compareEndDate      Optional. End date to compare report data for as YYYY-mm-dd.
```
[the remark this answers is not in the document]
asvinb: Updated @aaemnnosttv

assets/js/components/adminbar/AdminBarPreview.js
```diff
@@ -0,0 +1,48 @@
...
+		</div>
+		<div className="
+			mdc-layout-grid__cell
+			mdc-layout-grid__cell--span-2-tablet
+			mdc-layout-grid__cell--span-3-desktop
+		">
+			<PreviewBlock width="auto" height="59px" />
+		</div>
+	</Fragment>
+);
```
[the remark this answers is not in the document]
asvinb: @aaemnnosttv I've actually removed that component and just included the `PreviewBlock` component in the other components. Like you said, it'll be overkill to just have the `PreviewBlock` component in the `AdminBarPreview`  component.

assets/js/components/adminbar/AdminBarSessions.js
```diff
@@ -0,0 +1,116 @@
...
+				alias: 'Pageviews',
+			},
+		],
+		url,
+	};
+	const analyticsData = useSelect( ( select ) => select( MODULES_ANALYTICS ).getReport( reportArgs ) );
+
+	if ( undefined === analyticsData ) {
+		return <AdminBarPreview />;
+	}
```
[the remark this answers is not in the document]
asvinb: Updated @aaemnnosttv . See my previous comment.

assets/js/components/adminbar/AdminBarSessions.js
```diff
@@ -0,0 +1,116 @@
...
+
+	if ( ! analyticsData || ! analyticsData.length ) {
+		return false;
```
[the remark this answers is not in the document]
asvinb: Updated @aaemnnosttv

assets/js/components/adminbar/AdminBarSessions.js
```diff
@@ -0,0 +1,116 @@
...
+			{
+				expression: 'ga:pageviews',
+				alias: 'Pageviews',
+			},
+		],
+		url,
+	};
+	const analyticsData = useSelect( ( select ) => select( MODULES_ANALYTICS ).getReport( reportArgs ) );
+
+	if ( undefined === analyticsData ) {
```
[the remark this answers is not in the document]
asvinb: Updated @aaemnnosttv

assets/js/components/adminbar/AdminBarSessions.js
```diff
@@ -0,0 +1,116 @@
...
+		</div>
+	);
+};
+
+AdminBarSessions.propTypes = {
+	classNames: PropTypes.string,
+};
+
+AdminBarSessions.defaultProps = {
+	classNames: 'mdc-layout-grid__cell mdc-layout-grid__cell--span-2-tablet mdc-layout-grid__cell--span-3-desktop',
```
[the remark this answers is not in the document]
asvinb: Updated @aaemnnosttv

assets/js/components/adminbar/AdminBarUniqueVisitors.js
```diff
@@ -0,0 +1,94 @@
...
+		],
+		url,
+	};
+	const analyticsData = useSelect( ( select ) => select( MODULES_ANALYTICS ).getReport( reportArgs ) );
+
+	if ( undefined === analyticsData ) {
+		return <AdminBarPreview />;
+	}
+
+	if ( ! analyticsData || ! analyticsData.length ) {
```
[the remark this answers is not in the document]
asvinb: Updated @aaemnnosttv

assets/js/components/adminbar/AdminBarApp.js
asvinb: Updated @aaemnnosttv

assets/js/components/adminbar/AdminBarApp.js
asvinb: Updated @aaemnnosttv

assets/js/components/adminbar/AdminBarApp.js
asvinb: Updated @aaemnnosttv . Any reason why we don't use ternary? Is it because of performance reasons?

asvinb:
> @asvinb also wanted to note that VRT is failing – looks like a timeout from a not-found selector at first glance.

@aaemnnosttv VRT ok now :)

assets/js/components/adminbar/AdminBarApp.js
aaemnnosttv: I think it's mostly our convention and a preference that makes the code a bit easier to follow, especially with larger blocks. In this case it results in substantially less indentation/nesting. It also makes it a bit easier to see the logic for each rendered chunk whereas with a ternary you have to kind of backtrack to the original conditional and mentally flip the logic. We don't avoid it in all cases, but we generally prefer not to use it in JSX.

assets/js/components/adminbar/AdminBarSessions.js
```diff
@@ -0,0 +1,128 @@
...
+				alias: 'Users',
+			},
+			{
+				expression: 'ga:sessions',
+				alias: 'Sessions',
+			},
+			{
+				expression: 'ga:bounceRate',
+				alias: 'Bounce Rate',
+			},
```
aaemnnosttv, line 64: Sorry I missed this before @asvinb but we shouldn't be requesting all of these metrics when we only care about total sessions. We should only request `ga:sessions` here, then references to this should be updated below (`values[ 1 ]` would become `values[ 0 ]`).

assets/js/modules/analytics/datastore/report.js
```diff
@@ -122,6 +122,8 @@ const baseSelectors = {
 	 * @param {Object}         options                     Options for generating the report.
 	 * @param {string}         options.startDate           Required, unless dateRange is provided. Start date to query report data for as YYYY-mm-dd.
 	 * @param {string}         options.endDate             Required, unless dateRange is provided. End date to query report data for as YYYY-mm-dd.
+	 * @param {string}         [options.compareStartDate]  Optional. Start date to compare report data for as YYYY-mm-dd.
+	 * @param {string}         [options.compareEndDate]    Optional. End date to compare report data for as YYYY-mm-dd.
```
aaemnnosttv, line 126: Let's move these down to be after all required arguments rather than in between.

stories/adminbar.stories.js
```diff
@@ -42,6 +43,7 @@ import CollectModuleData from '../assets/js/components/data/collect-module-data'
 storiesOf( 'Global', module )
 	.add( 'Admin Bar', () => {
 		global._googlesitekitLegacyData = wpAdminBarData;
+		disableFeature( 'widgets.adminBar' );
```
aaemnnosttv, line 46: This should be added to `resetFeatures` in `.storybook/config` – stories should only enable features where needed. Features are then opt-in per-story, never opt-out.

aaemnnosttv (commented):
This is looking really good @asvinb – just a few minor details left to address then this should be good to go 👍

assets/js/components/adminbar/AdminBarSessions.js
asvinb: Updated @aaemnnosttv

assets/js/modules/analytics/datastore/report.js
asvinb: Updated @aaemnnosttv

stories/adminbar.stories.js
asvinb: Updated @aaemnnosttv

assets/js/components/adminbar/AdminBarSessions.js
```diff
@@ -0,0 +1,108 @@
...
+const AdminBarSessions = ( { className } ) => {
+	const url = useSelect( ( select ) => select( CORE_SITE ).getCurrentEntityURL() );
+	const dateRangeDates = useSelect( ( select ) => select( CORE_USER ).getDateRangeDates( {
+		compare: true,
+		offsetDays: DATE_RANGE_OFFSET,
+	} ) );
+	const reportArgs = {
+		...dateRangeDates,
+		dimensions: 'ga:date',
+		limit: 10,
```
aaemnnosttv, line 51: I'm not sure if the `limit` arg is needed anymore 🤔  
@felixarntz ?

assets/js/components/adminbar/AdminBarApp.js
```diff
@@ -75,6 +79,30 @@ export default function AdminBarApp() {
...
+											mdc-layout-grid__cell
+											mdc-layout-grid__cell--span-6-desktop
+											mdc-layout-grid__cell--span-4-tablet
+										">
+											<AnalyticsInactiveCTA />
+										</div>
+									) }
+
+									<AdminBarImpressions />
+									<AdminBarClicks />
```
aaemnnosttv, line 102: It looks like these need to go before the Analytics components rather than after.

From the current version:
image

aaemnnosttv (changes requested):
Looking good @asvinb ! This is really close but I just noticed that we're missing zero-data and error components to use when there is no data yet or a report error respectively.

image

Also note that the order of the components is off (see comment below)

These were missed as needed in the IB but are needed for consistency with the legacy components as these are provided by `withData` which we no longer use here.

See https://github.com/google/site-kit-wp/blob/8dc8de99f9225f62972c2bb82f67f6e3d39651ad/assets/js/modules/adsense/components/dashboard/DashboardTopEarningPagesWidget.js#L92-L98

Otherwise, this is looking solid. I have one question for @felixarntz below as well but nothing blocking.

assets/js/components/adminbar/AdminBarApp.js
asvinb: Updated @aaemnnosttv

assets/js/components/adminbar/AdminBarSessions.js
aaemnnosttv: Looks like it returns more rows without it, and we use `10` elsewhere so we should probably address this in a separate issue if needed.

assets/js/components/adminbar/AdminBarSessions.js
```diff
@@ -0,0 +1,120 @@
...
+		return <ReportError moduleSlug="analytics" error={ error } />;
+	}
+
+	if ( isZeroReport( analyticsData ) ) {
+		return <ReportZero moduleSlug="analytics" />;
+	}
+
+	if ( ! analyticsData?.length ) {
+		return null;
+	}
```
aaemnnosttv, line 88: This condition is no longer necessary as it is covered by the `isZeroReport` logic above.
```suggestion
```

assets/js/components/adminbar/AdminBarUniqueVisitors.js
```diff
@@ -0,0 +1,118 @@
...
+		return <ReportError moduleSlug="analytics" error={ error } />;
+	}
+
+	if ( isZeroReport( analyticsData ) ) {
+		return <ReportZero moduleSlug="analytics" />;
+	}
+
+	if ( ! analyticsData?.length ) {
+		return null;
+	}
```
aaemnnosttv, line 86: Same here 😄 
```suggestion
```

aaemnnosttv (commented):
Thanks @asvinb – just a few last details to clean up now that the missing components are in place. It should all be covered by the inline suggestions so I'll go ahead and apply them.

aaemnnosttv (changes requested):
@asvinb one last thing I just noticed – we're now showing the `AnalyticsInactiveCTA` when the module is activated but not connected yet. This should instead show the `CompleteModuleActivationCTA` as it does now
image

I.e. the `AnalyticsInactiveCTA` should only show when the module is not active, and `CompleteModuleActivationCTA` should show if it is active but not connected. Then this should be good to go 😅

asvinb:
> @asvinb one last thing I just noticed – we're now showing the `AnalyticsInactiveCTA` when the module is activated but not connected yet. This should instead show the `CompleteModuleActivationCTA` as it does now
> image
> 
> I.e. the `AnalyticsInactiveCTA` should only show when the module is not active, and `CompleteModuleActivationCTA` should show if it is active but not connected. Then this should be good to go 😅

@aaemnnosttv PR has been updated 😅

assets/js/components/adminbar/AdminBarClicks.js
```diff
@@ -0,0 +1,114 @@
...
+
+	return (
+		<div className={ classnames(
+			'mdc-layout-grid__cell',
+			className,
+		) }>
+			<DataBlock
+				className="overview-total-clicks"
+				title={ __( 'Total Clicks', 'google-site-kit' ) }
+				datapoint={ readableLargeNumber( totalClicks ) }
```
aaemnnosttv, line 98: With #2200 just merged, this should be the raw value now as `DataBlock` will format it internally.
```suggestion
				datapoint={ totalClicks }
```

assets/js/components/adminbar/AdminBarClicks.js
```diff
@@ -0,0 +1,114 @@
...
+ */
+import Data from 'googlesitekit-data';
+import DataBlock from '../DataBlock';
+import PreviewBlock from '../PreviewBlock';
+import ReportError from '../ReportError';
+import ReportZero from '../ReportZero';
+import { STORE_NAME as CORE_USER } from '../../googlesitekit/datastore/user/constants';
+import { STORE_NAME as CORE_SITE } from '../../googlesitekit/datastore/site/constants';
+import { STORE_NAME as MODULES_SEARCH_CONSOLE, DATE_RANGE_OFFSET } from '../../modules/search-console/datastore/constants';
+import { calculateChange, readableLargeNumber } from '../../util';
```
aaemnnosttv, line 41:
```suggestion
import { calculateChange } from '../../util';
```

assets/js/components/adminbar/AdminBarImpressions.js
```diff
@@ -0,0 +1,114 @@
...
+
+	return (
+		<div className={ classnames(
+			'mdc-layout-grid__cell',
+			className,
+		) }>
+			<DataBlock
+				className="overview-total-impressions"
+				title={ __( 'Total Impressions', 'google-site-kit' ) }
+				datapoint={ readableLargeNumber( totalImpressions ) }
```
aaemnnosttv, line 98:
```suggestion
				datapoint={ totalImpressions }
```

assets/js/components/adminbar/AdminBarImpressions.js
```diff
@@ -0,0 +1,114 @@
...
+ */
+import Data from 'googlesitekit-data';
+import DataBlock from '../DataBlock';
+import PreviewBlock from '../PreviewBlock';
+import ReportError from '../ReportError';
+import ReportZero from '../ReportZero';
+import { STORE_NAME as CORE_USER } from '../../googlesitekit/datastore/user/constants';
+import { STORE_NAME as CORE_SITE } from '../../googlesitekit/datastore/site/constants';
+import { STORE_NAME as MODULES_SEARCH_CONSOLE, DATE_RANGE_OFFSET } from '../../modules/search-console/datastore/constants';
+import { calculateChange, readableLargeNumber } from '../../util';
```
aaemnnosttv, line 41:
```suggestion
import { calculateChange } from '../../util';
```

assets/js/components/adminbar/AdminBarSessions.js
```diff
@@ -0,0 +1,116 @@
...
+
+	return (
+		<div className={ classnames(
+			'mdc-layout-grid__cell',
+			className,
+		) }>
+			<DataBlock
+				className="overview-total-sessions"
+				title={ __( 'Total Sessions', 'google-site-kit' ) }
+				datapoint={ readableLargeNumber( totalSessions ) }
```
aaemnnosttv, line 100:
```suggestion
				datapoint={ totalSessions }
```

assets/js/components/adminbar/AdminBarSessions.js
```diff
@@ -0,0 +1,116 @@
...
+ */
+import Data from 'googlesitekit-data';
+import DataBlock from '../DataBlock';
+import PreviewBlock from '../PreviewBlock';
+import ReportError from '../ReportError';
+import ReportZero from '../ReportZero';
+import { STORE_NAME as CORE_USER } from '../../googlesitekit/datastore/user/constants';
+import { STORE_NAME as CORE_SITE } from '../../googlesitekit/datastore/site/constants';
+import { STORE_NAME as MODULES_ANALYTICS, DATE_RANGE_OFFSET } from '../../modules/analytics/datastore/constants';
+import { calculateChange, readableLargeNumber } from '../../util';
```
aaemnnosttv, line 41:
```suggestion
import { calculateChange } from '../../util';
```

assets/js/components/adminbar/AdminBarUniqueVisitors.js
```diff
@@ -0,0 +1,114 @@
...
+
+	return (
+		<div className={ classnames(
+			'mdc-layout-grid__cell',
+			className,
+		) }>
+			<DataBlock
+				className="overview-total-users"
+				title={ __( 'Total Users', 'google-site-kit' ) }
+				datapoint={ readableLargeNumber( totalUsers ) }
```
aaemnnosttv, line 98:
```suggestion
				datapoint={ totalUsers }
```

assets/js/components/adminbar/AdminBarUniqueVisitors.js
```diff
@@ -0,0 +1,114 @@
...
+ */
+import DataBlock from '../DataBlock';
+import Data from 'googlesitekit-data';
+import PreviewBlock from '../PreviewBlock';
+import ReportError from '../ReportError';
+import ReportZero from '../ReportZero';
+import { STORE_NAME as CORE_USER } from '../../googlesitekit/datastore/user/constants';
+import { STORE_NAME as CORE_SITE } from '../../googlesitekit/datastore/site/constants';
+import { STORE_NAME as MODULES_ANALYTICS, DATE_RANGE_OFFSET } from '../../modules/analytics/datastore/constants';
+import { calculateChange, readableLargeNumber } from '../../util';
```
aaemnnosttv, line 41:
```suggestion
import { calculateChange } from '../../util';
```

aaemnnosttv (approved):
Awesome, thanks!

Issue #2368, opened 2020-11-16
Move PHP Tests to GitHub Actions


aaemnnosttv:
## Feature Description

As part of our initiative to move our CI jobs to GitHub Actions, this issue is for moving our PHP tests. This corresponds to the job that runs on Travis by the same name.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new GitHub Actions workflow definition should be created for each of the "PHP Tests" jobs we have currently defined in Travis
* The workflows should contain a job for running the same PHP Test jobs that we run on Travis
* The workflows and jobs should fail under the same conditions as they do today when running on Travis
* All PHP Tests, PHP-only variables, and related script steps should be removed from `.travis.yml`
* Previous runs of the workflow should be cancelled automatically by newer runs using our `cancel.yml` workflow by [adding the workflow filename of the workflow to the list provided to `workflow_id`](https://github.com/google/site-kit-wp/blob/07d95eddab386c3c5b9f8d1dad1f123d7cb33dae/.github/workflows/cancel.yml#L23) 
* The workflows should be a required check (set in repo settings)
	- Exception: The PHP Tests for WordPress nightly should not be required

## Implementation Brief

* Add a `php-tests.yml` file to our `.github/workflows/` folder.
* Following the format of the storybook tests, add our existing "PHP tests": `composer test` using the setup in `.travis.yml`
* Use the default `mysql` service provided by GitHub Actions using https://github.com/google/site-kit-wp/pull/2545/files#diff-8b304272438502fbcb449090052d2e707da4b71c84982948503c8f5f6532b0adR20-R24
* Use a "matrix" approach to run the tests on multiple versions of PHP.
* Merge #2545.
* Set the Repo Settings to require the new workflow named "PHP Tests" except those allowed to fail (@felixarntz will need to handle this):
  * All tests using `WordPress: Nightly` are allowed to fail

### Test Coverage

* None needed.

### Visual Regression Changes

* No changes.

## QA Brief

* The PR (#2545) should pass and use GitHub Actions to run the new check named "PHP Tests":

<img width="1061" alt="Screenshot 2020-12-17 at 18 38 08" src="https://user-images.githubusercontent.com/90871/102528852-0438c200-4097-11eb-9491-892e769de5fa.png">

## Changelog entry

* N/A

felixarntz:
IB ✅  😆  Moving over to Execution for the last pieces @tofumatt

[...]

PR #2545, opened 2020-12-17
Add GitHub actions for PHPUnit tests.

tofumatt:
## Summary

Addresses issue #2368.

## Checklist


.github/workflows/php-tests.yml
```diff
@@ -0,0 +1,48 @@
...
+on:
+  pull_request:
+    branches:
+      - master
+      - develop
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
```
felixarntz, line 12: @tofumatt Let's also run this on push, see other comments on #2537:
```suggestion
on:
  push:
    branches:
      - develop
      - master
  pull_request:
    types:
      - opened
      - reopened
      - synchronize
      - ready_for_review
```

.github/workflows/php-tests.yml
```diff
@@ -0,0 +1,70 @@
...
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
+
+jobs:
+  php-tests:
+    name: Run PHP Tests
```
aaemnnosttv: We can use values from the matrix to make the job name dynamic.

See here how I have this [on another project](https://github.com/aaemnnosttv/version-assets/blob/9480d7e7666341840e04d831a38223ddf71dc19b/.github/workflows/tests.yml).

.github/workflows/php-tests.yml
```diff
@@ -0,0 +1,70 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        php: ['5.6', '7.4']
```
aaemnnosttv: 5.6 is a special case, so this should go in the matrices `include` key since we don't want to generate a permutation of each WP version for it.

This would only be `[ '7.4' ]`.

.github/workflows/php-tests.yml
```diff
@@ -0,0 +1,70 @@
...
+        wordpress: ['4.7', 'latest-stable', 'latest-stable (multisite)', 'nightly']
```
aaemnnosttv: We should be able to use versions here that are passed to the test install script.

This should only be `latest` and `nightly`. 4.7 and latest with multisite should be added via `include`.

`multisite` itself would be part of the matrix and be set like `env: WP_MULTISITE: {{ matrix.multisite }}` (compressed for brevity). See a similar example on the other project.

.github/workflows/php-tests.yml
```diff
@@ -0,0 +1,70 @@
...
+    services:
+      mysql:
+        image: mysql:5.7
+        env:
+          MYSQL_DATABASE: wordpress
```
aaemnnosttv, line 23: This part should get the added health check command and may need the port exposed as well otherwise the test may be using the mysql server on the runner.
See https://github.com/aaemnnosttv/version-assets/blob/9480d7e7666341840e04d831a38223ddf71dc19b/.github/workflows/tests.yml#L32

.github/workflows/php-tests.yml
```diff
@@ -0,0 +1,70 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: shivammathur/setup-php@v2
+        with:
+          extensions: mbstring, intl, redis, pdo_mysql
```
aaemnnosttv, line 28: I don't _think_ we need `intl` but we for sure don't need `redis`, and `pdo_mysql` should probably be `mysqli` instead.

.github/workflows/php-tests.yml
```diff
@@ -0,0 +1,70 @@
...
+          php-version: ${{ matrix.php }}
+      - name: Start MySQL
+        run: sudo service mysql start
```
aaemnnosttv, line 31: This is the mysql instance on the runner (which we probably shouldn't use because we have no control over its version), not the one in the container added above 🙂

.github/workflows/php-tests.yml
```diff
@@ -0,0 +1,70 @@
...
+      - name: Enable multisite
+        if: ${{ matrix.wordpress == 'latest-stable-multisite' }}
+        run: |
+          echo "WP_MULTISITE=1" >> $GITHUB_ENV
```
aaemnnosttv: As mentioned before, this can be set unconditionally in the job's `env` (or in the test step) based on the value in the matrix. By default we can set it to `0` if needed, but probably unset/blank is fine too since this needs to be truthy to enable multisite.

.github/workflows/php-tests.yml
```diff
@@ -0,0 +1,70 @@
...
+      - name: Extract version from stable WordPress
+        if: ${{ matrix.wordpress == 'latest-stable' || matrix.wordpress == 'latest-stable-multisite' }}
+        run: |
+          curl -s http://api.wordpress.org/core/version-check/1.7/ > /tmp/wp-latest.json
+          echo "WP_VERSION=$(grep -o '"version":"[^"]*' /tmp/wp-latest.json | sed 's/"version":"//')" >> $GITHUB_ENV
```
aaemnnosttv: We shouldn't need to do this since the test setup/install script will install the latest version if we use the version `latest`.

.github/workflows/php-tests.yml
```diff
@@ -0,0 +1,70 @@
...
+        env:
+          WP_VERSION: ${{ matrix.wordpress }}
+      - name: Set WordPress version
+        if: ${{ !(matrix.wordpress == 'latest-stable' || matrix.wordpress == 'latest-stable (multisite)') }}
+        run: echo "WP_VERSION=$WP_VERSION" >> $GITHUB_ENV
+        env:
+          WP_VERSION: ${{ matrix.wordpress }}
```
aaemnnosttv: I don't think we need it via env either - it's only needed for passing to the `tests/bin/install-wp-tests.sh` script.

.github/workflows/php-tests.yml
```diff
@@ -0,0 +1,70 @@
...
+          echo "::set-output name=dir::$(composer config cache-files-dir)"
+      - uses: actions/cache@v2
+        with:
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-composer-
+      - name: Composer Install
+        if: ${{ (matrix.wordpress == '4.7' && matrix.php == '5.6') || (matrix.wordpress == 'latest-stable' && matrix.php == '7.4') || (matrix.wordpress == 'latest-stable (multisite)' && matrix.php == '7.4') || (matrix.wordpress == 'nightly' && matrix.php == '7.4') }}
+        run: docker run --rm -v "$PWD:/app" -v "${{ steps.composer-cache.outputs.dir }}:/tmp/cache" composer install --no-interaction --no-progress
```
aaemnnosttv: Not sure why this is needed – we don't conditionally run composer any differently between the current PHP test jobs on Travis.

.github/workflows/php-tests.yml
```diff
@@ -0,0 +1,70 @@
...
+      - name: Set up PHP test data
+        if: ${{ (matrix.wordpress == '4.7' && matrix.php == '5.6') || (matrix.wordpress == 'latest-stable' && matrix.php == '7.4') || (matrix.wordpress == 'latest-stable (multisite)' && matrix.php == '7.4') || (matrix.wordpress == 'nightly' && matrix.php == '7.4') }}
+        run: tests/bin/install-wp-tests.sh wordpress_test root root 127.0.0.1 $WP_VERSION
```
aaemnnosttv, line 153: We should be able to do this instead:

```suggestion
        run: tests/bin/install-wp-tests.sh wordpress_test root root 127.0.0.1 {{ matrix.wordpress }}
```

.github/workflows/php-tests.yml
```diff
@@ -0,0 +1,70 @@
...
+      - name: Run Unit Tests
+        if: ${{ (matrix.wordpress == '4.7' && matrix.php == '5.6') || (matrix.wordpress == 'latest-stable' && matrix.php == '7.4') || (matrix.wordpress == 'latest-stable (multisite)' && matrix.php == '7.4') || (matrix.wordpress == 'nightly' && matrix.php == '7.4') }}
```
aaemnnosttv: This line should always run. Why would the main step run conditionally?

```suggestion
```

tests/bin/install-wp-tests.sh
```diff
@@ -88,7 +88,7 @@ install_wp() {
 		else
 			local ARCHIVE_NAME="wordpress-$WP_VERSION"
 		fi
-		download https://wordpress.org/${ARCHIVE_NAME}.tar.gz  $TMPDIR/wordpress.tar.gz
+		download https://wordpress.org/${ARCHIVE_NAME}.tar.gz $TMPDIR/wordpress.tar.gz
```
aaemnnosttv, line 91: Why was this line changed? This should be a mirror of the "main" script which lives in wp-cli's scaffolding.

aaemnnosttv (commented):
Thanks @tofumatt ! Leaving a few comments here to provide some context for our discussion 😄

.github/workflows/php-tests.yml
tofumatt: Argh, sure, copypasta from another test getting 5.6 working 😅

.github/workflows/php-tests.yml
tofumatt: This is to prevent these from running at all for combos of the matrix we don't run (on Travis). But I'll tidy it up...

.github/workflows/php-tests.yml
tofumatt: See above: to skip it for combos of matrix values we want to skip.

tests/bin/install-wp-tests.sh
tofumatt: Oh, I had an error from something unrelated, noticed the double space, and might've just been tidying it. But if it's a mirror I'll restore 😉

.github/workflows/php-tests.yml
tofumatt: Urgh, right, sorry. 😓

.github/workflows/php-tests.yml
aaemnnosttv: Better to simplify the matrix and add includes for additional values that you don't want to generate permutations for.

felixarntz:
@tofumatt I've removed `.travis.yml` and all Travis-CI references as part of this PR, since it's the last piece to migrate. Can you split out the PHP tests into separate jobs and then put each one into its own `.yml` file per scenario as discussed? This way it's aligned with the other tests.

Super excited to have everything running on GH Actions!

.github/workflows/php-tests-wp-4-7-php-5-6.yml
```diff
@@ -0,0 +1,69 @@
...
+        env:
+          MYSQL_ALLOW_EMPTY_PASSWORD: yes
+        ports:
+          - 3306:3306
+        options: --health-cmd="mysqladmin ping" --health-interval=5s --health-timeout=2s --health-retries=3
+    env:
+      DB_HOST: 127.0.0.1
+      DB_PORT: 3306
+      MYSQL_USER: user
+      MYSQL_PASSWORD: password
```
eugene-manuilov, line 30:
```suggestion
      MYSQL_USER: root
      MYSQL_PASSWORD: root
```

.github/workflows/php-tests-wp-4-7-php-5-6.yml
```diff
@@ -0,0 +1,69 @@
...
+    name: WP 4.7, PHP 5.6
+    runs-on: ubuntu-latest
+    services:
+      mysql:
+        image: mysql:5.7
+        env:
+          MYSQL_ALLOW_EMPTY_PASSWORD: yes
+        ports:
+          - 3306:3306
+        options: --health-cmd="mysqladmin ping" --health-interval=5s --health-timeout=2s --health-retries=3
```
eugene-manuilov, line 25: @tofumatt MySQL service isn't needed because there is a MySQL server on the virtual machine itself. Just run sudo systemctl start mysql.service to start it and you can use it with root user and root password.

.github/workflows/php-tests-wp-4-7-php-5-6.yml
```diff
@@ -0,0 +1,69 @@
...
+          while ! mysqladmin ping -h"$DB_HOST" -P"$DB_PORT" --silent; do
+            sleep 1
+          done
+      - name: Grant database access
+        env:
+          MYSQL_TCP_PORT: ${{ env.DB_PORT }}
+        run: |
+          mysql -u root -h ${DB_HOST} -e "CREATE USER ${MYSQL_USER} IDENTIFIED BY '${MYSQL_PASSWORD}';"
+          mysql -u root -h ${DB_HOST} -e "GRANT ALL PRIVILEGES ON *.* TO ${MYSQL_USER};"
+          mysql -u root -h ${DB_HOST} -e "FLUSH PRIVILEGES;"
```
eugene-manuilov, line 47: This is not needed.

eugene-manuilov (changes requested):
Thanks, @tofumatt! This is a good start, but we need to adjust it. I have added a few comments to one of the new workflows, the same applies to others. Could you please take a look at it?

.github/workflows/php-tests.yml
eugene-manuilov: @aaemnnosttv the version of MySQL doesn't matter if it is 5.6 or higher. Why do you want to control its version?

.github/workflows/php-tests-wp-4-7-php-5-6.yml
```diff
@@ -0,0 +1,69 @@
...
+          restore-keys: |
+            ${{ runner.os }}-composer-
+      - name: Composer Install
+        run: docker run --rm -v "$PWD:/app" -v "${{ steps.composer-cache.outputs.dir }}:/tmp/cache" composer install --no-interaction --no-progress # Run inside docker so composer install doesn't error...
+      - name: Set up PHP test data
+        run: tests/bin/install-wp-tests.sh ${MYSQL_DATABASE} ${MYSQL_USER} ${MYSQL_PASSWORD} ${DB_HOST}:${DB_PORT} ${WP_VERSION}
+      - name: Run Unit Tests
+        run: composer test
+        env:
+          WP_DB_HOST: ${{ env.DB_HOST }}:${{ env.DB_PORT }}
```
aaemnnosttv, line 69: This variable isn't necessary unless you wanted to substitute it above in the place of `${DB_HOST}:${DB_PORT}` but I think it's better to keep that part as-is for now.

aaemnnosttv (commented):
@tofumatt @eugene-manuilov – I left a few comments/replies to add onto the initial review here.

.github/workflows/php-tests-wp-4-7-php-5-6.yml
eugene-manuilov: @aaemnnosttv I don't have such doubts and think that it should work without any problem. Anyway, we can try it with the available version of MySQL server on the host and if it doesn't work for WordPress 4.7, then switch back to using a service, ok?

aaemnnosttv: @eugene-manuilov My concern there is that while it may work today, unexpected changes to the MySQL version on the runner could break things in the future. The [requirements page](https://make.wordpress.org/hosting/handbook/handbook/server-environment/#database) shows it supports MySQL `8.0.x` – does that mean older versions might not work with MySQL 8.1?

There are already a few tickets around MySQL 8:
- https://core.trac.wordpress.org/ticket/51740
- https://core.trac.wordpress.org/ticket/49364

It seems like it would be a good idea to stick to the latest MySQL v5 for now.

eugene-manuilov: I wish we could try MySQL 8 for our tests anyway and use it if it doesn't cause problems.

composer.lock
```diff
@@ -1578,6 +1578,7 @@
             "keywords": [
                 "tokenizer"
             ],
+            "abandoned": true,
```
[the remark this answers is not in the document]
tofumatt: I don't think so—I was getting composer errors but they were unrelated to the lockfile, so reverted 👍🏻

.github/workflows/php-tests-wp-4-7-php-5-6.yml
```diff
@@ -0,0 +1,69 @@
...
+        run: |
+          echo "::set-output name=dir::$(composer config cache-files-dir)"
+      - uses: actions/cache@v2
+        with:
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-composer-
+      - name: Composer Install
+        run: docker run --rm -v "$PWD:/app" -v "${{ steps.composer-cache.outputs.dir }}:/tmp/cache" composer install --no-interaction --no-progress # Run inside docker so composer install doesn't error...
```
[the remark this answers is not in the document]
tofumatt: Yeah I think in this case we just want the install to work, and the speed improvements here are enough to use the docker command instead. Agreed it's less clean, but the comment at least clarifies the purpose 🙂

.github/workflows/php-tests-wp-4-7-php-5-6.yml
tofumatt: The existing ones, I thought, assumed v5, so let's stick with that for now but add a v8 test (or move to it) in another issue? I'm just wary of introducing changes to the tests here during the migration.

.github/workflows/php-tests-wp-4-7-php-5-6.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Grant database access
+        env:
+          MYSQL_TCP_PORT: ${{ env.DB_PORT }}
+        run: |
+          mysql -u root -h ${DB_HOST} -e "CREATE USER ${MYSQL_USER} IDENTIFIED BY '${MYSQL_PASSWORD}';"
+          mysql -u root -h ${DB_HOST} -e "GRANT ALL PRIVILEGES ON *.* TO ${MYSQL_USER};"
+          mysql -u root -h ${DB_HOST} -e "FLUSH PRIVILEGES;"
+      - uses: shivammathur/setup-php@v2
+        with:
+          extensions: composer, mysqli
```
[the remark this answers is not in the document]
tofumatt: Oh, shoot, you're right, I thought I needed it for set up but the docker path worked. Either this isn't needed, or I could move this to `tools` and save the `docker run` command…

.github/workflows/php-tests-wp-4-7-php-5-6.yml
tofumatt: Trying to use the root user always results in an error for me, so I think we need this:

Screenshot 2020-12-21 at 19 20 36

felixarntz (approved):
Awesome!

PR #2556, opened 2020-12-21
Fix potential error with CompleteModuleActivationCTA with updates for consistency

aaemnnosttv:
## Summary

Addresses issue #2299

This PR fixes a potential fatal client side error when rendering `CompleteModuleActivationCTA` before the module has been fully loaded
image

## Relevant technical choices

* Targets `master` for inclusion in release
* Fixes potential "Cannot read property 'name' of undefined error" in `CompleteModuleActivationCTA` due to no guard for undefined `module`
* Inlined prop values for consistency with most other components
* Added missing dependencies to `useCallback` hooks
* Confirmed both components render without raising an error now in dev and production builds
image
image

## Checklist

### DIFF
--- a/assets/js/components/ActivateModuleCTA.js
+++ b/assets/js/components/ActivateModuleCTA.js
@@ -44,6 +44,7 @@ const ActivateModuleCTA = ( { slug, title, description } ) => {
 	const module = useSelect( ( select ) => select( CORE_MODULES ).getModule( slug ) );
 	const canManageOptions = useSelect( ( select ) => select( CORE_USER ).hasCapability( PERMISSION_MANAGE_OPTIONS ) );
 	const { activateModule } = useDispatch( CORE_MODULES );
+
 	const onCTAClick = useCallback( async () => {
 		const { error, response } = await activateModule( slug );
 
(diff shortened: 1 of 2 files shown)


assets/js/components/ActivateModuleCTA.js
```diff
@@ -58,37 +59,36 @@ const ActivateModuleCTA = ( { slug, title, description } ) => {
 				type: 'win-error',
 			} );
 		}
-	} );
+	}, [ activateModule ] );
```
aaemnnosttv, line 62: `activateModule` comes from `useDispatch` here which is generally stable, but could potentially return different values due to async store registration, so good to be a little extra safe here.

assets/js/components/ActivateModuleCTA.js
```diff
@@ -58,37 +59,36 @@ const ActivateModuleCTA = ( { slug, title, description } ) => {
...
 
-	if ( ! module || ! canManageOptions ) {
+	if ( ! module?.name || ! canManageOptions ) {
```
aaemnnosttv, line 64: This component didn't have the same issue but updated the guard here for consistency with the other since `name` is the only property we use from `module` – we shouldn't render without a name which should never be empty anyways.

assets/js/components/ActivateModuleCTA.js
```diff
@@ -58,37 +59,36 @@ const ActivateModuleCTA = ( { slug, title, description } ) => {
...
-		sprintf(
-			/* translators: %s: Module name */
-			__( '%s module needs to be configured', 'google-site-kit' ),
-			name,
-		);
-	const moduleCTALabel = sprintf(
-		/* translators: %s: Module name */
-		__( 'Set up %s', 'google-site-kit' ),
-		name,
-	);
```
aaemnnosttv, line 84: Separate variable names are redundant when used as props here, since the prop serves the same purpose.

assets/js/components/CompleteModuleActivationCTA.js
```diff
@@ -40,43 +40,42 @@ const CompleteModuleActivationCTA = ( { slug, title, description } ) => {
 	const module = useSelect( ( select ) => select( MODULES_STORE ).getModule( slug ) );
 	const adminReauthURL = useSelect( ( select ) => select( `modules/${ slug }` )?.getAdminReauthURL() );
 	const canManageOptions = useSelect( ( select ) => select( CORE_USER ).hasCapability( PERMISSION_MANAGE_OPTIONS ) );
-	const { name } = module;
 
 	const onCTAClick = useCallback( async () => {
-		global.location = adminReauthURL;
-	} );
+		global.location.assign( adminReauthURL );
+	}, [ adminReauthURL ] );
```
aaemnnosttv, line 46: * Used `location.assign` instead
* Added the selected `adminReauthURL` to the callback's dependencies – no array was present before which would cause the callback to be updated each time the component was rendered. Probably not an expensive omission but if we're using `useCallback` we should always provide a dependency list, otherwise it defeats the purpose.

assets/js/components/CompleteModuleActivationCTA.js
```diff
@@ -40,43 +40,42 @@ const CompleteModuleActivationCTA = ( { slug, title, description } ) => {
...
 
-	if ( ! canManageOptions ) {
+	if ( ! module?.name || ! canManageOptions ) {
```
aaemnnosttv, line 48: Again, guarding with the module name, if present since that's the only thing we need from it.

assets/js/components/CompleteModuleActivationCTA.js
```diff
@@ -40,43 +40,42 @@ const CompleteModuleActivationCTA = ( { slug, title, description } ) => {
...
-			__( '%s module setup needs to be completed', 'google-site-kit' ),
-			name,
-		);
-	const moduleCTA = __( 'Complete setup', 'google-site-kit' );
-
-	const moduleAriaLabel = sprintf(
-		/* translators: %s: Module name */
-		__( 'Complete %s setup', 'google-site-kit' ),
-		name,
-	);
```
aaemnnosttv, line 71: Inlining as props instead – reads much cleaner IMO.

felixarntz (approved):
Awesome!

Issue #2062, opened 2020-09-17
Register widget contexts and widget areas for module pages


felixarntz:
Widget contexts and the available widget areas should be registered via the Widget API for the Site Kit modules pages (Search Console, AdSense, Analytics). Furthermore the internal widget rendering components should be put in place on those pages, to render widget areas for the respective context (above the original "widgets" not using the Widget API).

Individual feature flags should be used, which will only be active for development builds. The existing UI should remain untouched (similar to how it was with the dashboard before) - the new widgets implementation will for now show up as a "copy" of the legacy version of the UI.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The following additional default context slugs should receive constants in `assets/js/googlesitekit/widgets`:
    * `CONTEXT_MODULE_SEARCH_CONSOLE`
    * `CONTEXT_MODULE_ANALYTICS`
    * `CONTEXT_MODULE_ADSENSE`
* The following additional widget area slugs should receive constants in `assets/js/googlesitekit/widgets`:
    * `AREA_MODULE_SEARCH_CONSOLE_MAIN`
    * `AREA_MODULE_ANALYTICS_MAIN`
    * `AREA_MODULE_ADSENSE_MAIN`
* The `registerDefaults` function in `assets/js/googlesitekit/widgets` should be expanded to register the above widget areas in their respective above context:
    * `title` and `subtitle` should always be empty.
    * `priority` should always be 1.
    * `style` should always be `WIDGET_AREA_STYLES.BOXES`.

## Implementation Brief

* Add three context slugs to `/assets/js/googlesitekit/widgets/default-contexts.js` `CONTEXT_MODULE_SEARCH_CONSOLE`, `CONTEXT_MODULE_ANALYTICS`, and `CONTEXT_MODULE_ADSENSE`.
* Add three additional widget areas in `/assets/js/googlesitekit/widgets/default-areas.js`: `AREA_MODULE_SEARCH_CONSOLE_MAIN`, `AREA_MODULE_ANALYTICS_MAIN`, and `AREA_MODULE_ADSENSE_MAIN`.
* In `/assets/js/googlesitekit/widgets/register-defaults.js` registerDefaults() add the above three widget areas to their context areas ( `AREA_MODULE_SEARCH_CONSOLE_MAIN` / `CONTEXT_MODULE_SEARCH_CONSOLE` and so on ) using `widgetsAPI.registerWidgetArea()`:
```js
{
[...]

PR #2564, opened 2020-12-23
Register widget contexts and widget areas for module pages.

asvinb:
## Summary

Addresses issue #2062 

## Checklist


assets/js/googlesitekit/widgets/register-defaults.js
```diff
@@ -159,4 +168,37 @@ export function registerDefaults( widgetsAPI ) {
 			AREA_DASHBOARD_POPULARITY,
 		],
 	);
+
+	widgetsAPI.registerWidgetArea(
+		AREA_MODULE_SEARCH_CONSOLE_MAIN,
+		{
+			priority: 1,
+			subtitle: ' ',
```
tofumatt, line 176: The IB mentions using blank/empty strings for the title/subtitle here, why are `' '` strings used throughout instead?

tofumatt (changes requested):
Looks like this is slightly off compared to the IB, just needs a minor adjustment.

assets/js/googlesitekit/widgets/register-defaults.js
asvinb: Hey @tofumatt 
it's because if we use empty strings, we get the following error message:
`Uncaught Invariant Violation: settings.title is required.`
I don't think we'll need to review the condition which requires the title in case we are using it in some other ways. Also, not sure if the title and subtitle will remain blank later because it looks to me they are placeholders for now.

tofumatt: @felixarntz What's the reason behind these titles/subtitles being blank in the issue? Since we explicitly have a check that forbids empty titles/subtitles (and it doesn't seem like a good idea), should we deal with this another way?

I think either we should add:

1. titles/subtitles for these areas
2. a way to hide the title/subtitle but still register them (if the intent here was not to visually show said title/subtitle)

But the current "fake blank" approach feels like a hack and something we're doing to get around rules we explicitly set up, so probably isn't the right approach. 🙂 

Thoughts @felixarntz?

tofumatt (commented):
I think we need to tweak the approach here, I'm a bit curious why the blank titles were proposed anyhow. Let's see what @felixarntz's intent was there before we go ahead with this 🙂

assets/js/googlesitekit/widgets/register-defaults.js
felixarntz: @tofumatt @asvinb The titles and subtitles need to be empty because this is to replicate the current module pages ("Search Console" etc.) via the Widgets API. In those pages, there is only the overall page headline (the module title), and then a single widget area for the entire page basically. So this requirement is valid and should be kept as specified in the issue.

However, I agree the way we hack around our own API here is not great. Maybe we can do the following:
* Specify all widget area `title`s as `__( 'Overview', 'google-site-kit' )`.
* Do not specify any `subtitle` for them.
* Make `subtitle` optional in `registerWidgetArea` - that is really something that shouldn't be required.
* Update `WidgetAreaRenderer` so that it does not render the area's header entirely if this widget area is the **only** widget area in its context.

felixarntz (changes requested):
Responded to the feedback from @tofumatt. We need to not display any title or subtitle for these areas, but the implementation _does_ need to be improved. I left a suggestion.

assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js
```diff
@@ -65,25 +65,27 @@ export default function WidgetAreaRenderer( { slug } ) {
...
-				<Cell className="googlesitekit-widget-area-header" size={ 12 }>
-					{ Icon && (
-						<Icon width={ 33 } height={ 33 } />
-					) }
+			{ totalAreas > 1 && (
+				<Row>
+					<Cell className="googlesitekit-widget-area-header" size={ 12 }>
+						{ Icon && (
+							<Icon width={ 33 } height={ 33 } />
+						) }
```
tofumatt, line 73: This would be good to add a unit test for. There's already test infrastructure for the `WidgetAreaRenderer`—could you add a test in https://github.com/google/site-kit-wp/blob/develop/assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.test.js#L68 to ensure the title only appears with multiple areas?

assets/js/googlesitekit/widgets/register-defaults.js
```diff
@@ -159,4 +168,37 @@ export function registerDefaults( widgetsAPI ) {
 			AREA_DASHBOARD_POPULARITY,
 		],
 	);
+
+	widgetsAPI.registerWidgetArea(
+		AREA_MODULE_SEARCH_CONSOLE_MAIN,
+		{
+			priority: 1,
+			subtitle: '',
```
tofumatt, line 176: The subtitles can be removed entirely now, since they're optional. I know `''` isn't truthy in JS, but omitting the param seems more explicit than including the empty one.

```suggestion
```

assets/js/googlesitekit/widgets/register-defaults.js
```diff
@@ -159,4 +168,37 @@ export function registerDefaults( widgetsAPI ) {
...
+			title: __( 'Overview', 'google-site-kit' ),
+		},
+		CONTEXT_MODULE_SEARCH_CONSOLE,
+	);
+
+	widgetsAPI.registerWidgetArea(
+		AREA_MODULE_ANALYTICS_MAIN,
+		{
+			priority: 1,
+			subtitle: '',
```
tofumatt, line 187: Same here:

```suggestion
```

assets/js/googlesitekit/widgets/register-defaults.js
```diff
@@ -159,4 +168,37 @@ export function registerDefaults( widgetsAPI ) {
...
+			title: __( 'Overview', 'google-site-kit' ),
+		},
+		CONTEXT_MODULE_ANALYTICS,
+	);
+
+	widgetsAPI.registerWidgetArea(
+		AREA_MODULE_ADSENSE_MAIN,
+		{
+			priority: 1,
+			subtitle: '',
```
tofumatt, line 198: Same here:

```suggestion
```

tofumatt (changes requested):
I like this much better, let's add a test for the title appearing only when there are more than one areas, but then we can ship this 👍🏻

tofumatt (approved):
Awesome, thanks so much!

aaemnnosttv:
Assigning to @felixarntz for final review as he previously requested changes as well.

felixarntz (approved):
Awesome!

Issue #2495, opened 2020-12-07
Do not make proxy notifications request when not using proxy


felixarntz:
## Bug Description

Currently, for sites that aren't using the proxy (because of using custom OAuth client credentials), Site Kit still regularly requests the notifications endpoint of the proxy, which naturally will fail, as the client ID used in that case is not a valid site ID against the proxy.

Those requests should be avoided, either by the client not making the request to the server for notifications at all, or by the server not requesting remote notifications when not using the proxy. The latter probably makes more sense because the `/core/site/data/notifications` endpoint is a general notifications endpoint and, while it is currently only used to return remote notifications, could also return notifications coming directly from the plugin server-side.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `core/site/data/notifications` endpoint should be modified to only issue its API request to the proxy endpoint if the site is actually using the proxy. If not, it should simply return an empty array of notifications. (If we need plugin-side notifications to be returned from here in the future, we could do that in this endpoint too.)

## Implementation Brief

* Move the `if` check for `using_proxy()` in `Notifications.php` (https://github.com/google/site-kit-wp/blob/dc2cb956993b8d6e47592884264c37a70bcd9895/includes/Core/Notifications/Notifications.php#L143) above the `$response = wp_remote_get( $endpoint );` line, so that a request is not made to the proxy if it is not in use.

### Test Coverage

* A test to ensure `wp_remote_get` is not called when not using the proxy would be useful to add in PHPUnit, if straightforward.

### Visual Regression Changes

* There should be no changes to VRT.

## QA Brief

* Configure your site without the proxy, confirm that the `core/site/data/notifications` REST endpoint **does not** attempt to call the Google proxy.
* Configure your site with the proxy,  confirm that the `core/site/data/notifications` REST endpoint **does** call the Google proxy.

## Changelog entry

*

felixarntz:
@tofumatt Nitpick, but I think it would be a bit better to move the check to the very top of that function because we can avoid any other logic in that case. That's a detail though that can be covered during code review. IB ✅

aaemnnosttv:
@benbowler would you please add a QA Brief? Once added this is good to move over to QA as the PR is merged 👍

benbowler:
QA Brief added.

asvinb:
- Configured site without the proxy, no calls to Google proxy 👍 
- Configured site with the proxy, calls to Google proxy 👍 

QA: ✅

PR #2569, opened 2020-12-29
Do not make proxy notifications request when not using proxy.

benbowler:
## Summary

Addresses issue #2495

## Relevant technical choices

Implemented as in the IB.

For the test, I created and authenticate a user with permissions to access the notification REST API endpoint, then I created a new `WP_REST_Request` dispatching it with `$server->dispatch` and asserted it returns the empty array of notifications.

## Checklist


tests/phpunit/integration/Core/Notifications/NotificationsTest.php
```diff
@@ -42,4 +47,35 @@ function ( REST_Route $route ) {
...
+		// Fake setup and authentication for access to dashboard.
+		$this->fake_site_connection();
+		remove_all_filters( 'googlesitekit_setup_complete' );
+		$authentication = new Authentication( new Context( GOOGLESITEKIT_PLUGIN_MAIN_FILE ) );
+		$authentication->verification()->set( true );
+		$authentication->get_oauth_client()->set_access_token( 'test-access-token', HOUR_IN_SECONDS );
+
+		// Set up the REST Routes including the notification route to test.
+		$rest_routes = new REST_Routes( new Context( GOOGLESITEKIT_PLUGIN_MAIN_FILE ) );
+		$rest_routes->register();
```
aaemnnosttv, line 66: This part isn't actually needed because the routes are set up during the plugin bootstrap. I like the idea to use dedicated instances in the tests but I'm not sure if we can. This is also still using the `Route` instance from the main/bootstrapped `Notifications` class.

tests/phpunit/integration/Core/Notifications/NotificationsTest.php
```diff
@@ -42,4 +47,35 @@ function ( REST_Route $route ) {
...
+
+		// Make the test request.
+		$request  = new WP_REST_Request( 'GET', '/' . REST_Routes::REST_ROOT . '/core/site/data/notifications' );
+		$response = $server->dispatch( $request );
+		$data     = $response->get_data();
+
+		$this->assertEquals(
+			array(),
+			$data
+		);
```
aaemnnosttv, line 79: While this does somewhat test the request, it doesn't actually test that the request wasn't made to the proxy which is the most important thing here. Also, it's worth noting that an empty array doesn't specifically mean that a request wasn't made as the proxy could return an empty array as well as a valid response (i.e. no notifications).

In this case, we should simply hook into the `pre_http_request` filter before dispatching the rest request, and if a request is made with a URL containing the proxy's notifications endpoint (`$this->google_proxy->url( '/notifications/' )`), then the test should fail (regardless of the response). We have a few other tests where this filter is used that you could use as a reference, but essentially we want to do something like this:

```php
// in test
add_filter( 'pre_http_request', function ( $override, $args, $url ) {
	if ( /* $url starts with notifications endpoint URL */ ) {
		$this->fail( 'Expected no request to be made to Google proxy notifications endpoint' );
	}
	return $override;
}, 10, 3 );
///
```

The rest of the test can be simplified to meet the minimum requirements needed to make the request and if the response code is a 200 then we know the permissions were met.

aaemnnosttv (changes requested):
Thanks @benbowler – this looks good just one change to make in the test. Let me know if you have any questions there 👍

tests/phpunit/integration/Core/Notifications/NotificationsTest.php
```diff
@@ -42,4 +49,44 @@ function ( REST_Route $route ) {
 			$routes
 		);
 	}
+
+	public function test_non_proxy_request_does_not_request_proxy() {
+
+		$context                        = new Context( GOOGLESITEKIT_PLUGIN_MAIN_FILE, new MutableInput() );
```
aaemnnosttv, line 55: No need for `MutableInput` here as we aren't mutating any super global input arrays.
```suggestion
		$context                        = new Context( GOOGLESITEKIT_PLUGIN_MAIN_FILE );
```

tests/phpunit/integration/Core/Notifications/NotificationsTest.php
```diff
@@ -42,4 +49,44 @@ function ( REST_Route $route ) {
...
+		$google_proxy                   = new Google_Proxy( $context );
+		$google_proxy_notifications_url = $google_proxy->url( '/notifications/' );
+
+		// Create a user with access to the WP REST API and log in.
+		$user = $this->factory()->user->create_and_get( array( 'role' => 'administrator' ) );
+		wp_set_current_user( $user->ID );
+		do_action( 'wp_login', $user->user_login, $user );
+
+		// Fake setup and authentication for access to dashboard.
```
aaemnnosttv, line 64: We're not accessing the dashboard so this comment isn't accurate.
```suggestion
```

tests/phpunit/integration/Core/Notifications/NotificationsTest.php
```diff
@@ -42,4 +49,44 @@ function ( REST_Route $route ) {
...
+			10,
+			3
+		);
+
+		// Get a REST server instance.
+		$server = rest_get_server();
+
+		// Make the test request.
+		$request  = new WP_REST_Request( 'GET', '/' . REST_Routes::REST_ROOT . '/core/site/data/notifications' );
+		$response = $server->dispatch( $request );
```
aaemnnosttv, line 84: The server instance can be inlined, the extra comment there is a bit redundant as well.
```suggestion
		// Make the test request.
		$request  = new WP_REST_Request( 'GET', '/' . REST_Routes::REST_ROOT . '/core/site/data/notifications' );
		$response = rest_get_server()->dispatch( $request );
```

aaemnnosttv (commented):
LGTM, just a few minor details to clean up which I'll take care of in just a second 🙂

tests/phpunit/integration/Core/Notifications/NotificationsTest.php
```diff
@@ -42,4 +49,44 @@ function ( REST_Route $route ) {
...
+
+		// Confirm the request returns 200 status code.
+		$this->assertEquals(
+			$response->get_status(),
+			200
+		);
```
aaemnnosttv, line 90: Expected and actual values are backwards here 😄 
```suggestion
		$this->assertEquals( 200, $response->get_status() );
```

aaemnnosttv (approved):
👍  Thanks!

Issue #2477, opened 2020-12-03
Update AdSense widgets to no longer pass special dateRange parameters


felixarntz:
#2235 (and follow-up #2459) already modified the "typical" `dateRange` parameters in refactored AdSense widgets to be replaced with `startDate` and `endDate` parameters. There are some special cases in the `DashboardSummaryWidget` though which use `dateRange` values that are only used and supported by the AdSense module specifically, so those cannot use the centralized `getDateRangeDates` selector. Let's migrate those too though via another appropriate way 

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The AdSense `DashboardSummaryWidget` should no longer pass `dateRange` parameters to the API.
    * For `today`, this can be done by simply getting the reference date from the `core/user` store and using it for both `startDate` and `endDate`.
    * For `this-month`, we need to introduce some other selector (maybe best in the `modules/adsense` store because it's specific to that for now) that returns the current reference date as `endDate` and the first date of the same month as `startDate`.
* The Storybook setup for this widget needs to be adjusted so that it still works as before these changes.

## Implementation Brief

* Change `todayArgs` in `DashboardSummaryWidget` to use:
```js
const referenceDate = select( CORE_USER ).getReferenceDate();

const todayArgs = {
	startDate: referenceDate,
	endDate: referenceDate,
	metrics,
};
```
* Get the start of the month using an approach like:
```js
const referenceDateFormat = new Intl.DateTimeFormat( 'en', { year: 'numeric', month: '2-digit', day: '2-digit' } );
const startOfMonth = referenceDate ? referenceDateFormat.format(
	new Date(
		new Date( referenceDate ).getFullYear(),
		new Date( referenceDate ).getMonth(),
		1
	)
) : undefined;
```
* Change `dailyArgs` to use:
```js
const dailyArgs = {
	startDate: startOfMonth,
	endDate: referenceDate,
	metrics,
	dimensions: [ 'DATE' ],
};
```

I don't feel a selector is relevant right now because this format is only used in this single component (along with the legacy component). If required the logic could be split out into a registry selector that used `getReferenceDate`, but right now doing the calculation inside the only component that needs it seems fine to me. Let me know what you think, but if you wanted we can easily amend the IB to add a `getStartOfMonthRelativeToReferenceDate()` selector 😅 

### Test Coverage

[...]

PR #2570, opened 2020-12-30
Update AdSense widgets to no longer pass special dateRange parameters.

benbowler:
## Summary

Update AdSense widgets to no longer pass special dateRange parameters.
Addresses issue #2477

## Relevant technical choices

I changed the date formatting code slightly because Site Kit uses the ISO 8601 date format.

```
const startOfMonth = new Date(
	new Date( referenceDate ).getFullYear(),
	new Date( referenceDate ).getMonth(),
	1
).toLocaleDateString( 'sv-SE' ); // Sweden uses ISO 8601 as their official date format.
```

## Checklist


assets/js/modules/adsense/components/dashboard/DashboardSummaryWidget.js
```diff
@@ -65,8 +68,16 @@ function DashboardSummaryWidget() {
 			metrics,
 		};
 
+		// Get the first day of the month an ISO 8601 date string.
+		const startOfMonth = new Date(
+			new Date( referenceDate ).getFullYear(),
+			new Date( referenceDate ).getMonth(),
+			1
+		).toLocaleDateString( 'sv-SE' ); // Sweden uses ISO 8601 as their official date format.
```
tofumatt, line 76: So, that's neat, but my concern about this is… what happens if Sweden changes their date format? 😆 

I know it's unlikely, but this feels like using the fact that Swedish in Sweden uses that format as a hack rather than outputting exactly what we meant to output clearly.

Is the main reason for doing this to avoid using `Intl.DateTimeFormat`? What's the reason behind it?

This is a case of "shorter code" but less clear code to me, so I'd rather we didn't rely on a specific locale's implementation for our internal date formats 😄

tofumatt (changes requested):
I left a comment about it, but basically I think we should be more explicit in how we construct the date string than relying on a locale. Unless there's a reason to avoid `Intl.DateTimeFormat`, as mentioned in my comment.

assets/js/modules/adsense/components/dashboard/DashboardSummaryWidget.js
benbowler: Hey @tofumatt, the issue with the `Intl.DateTimeFormat` implementation is that, if set to `'en'`, it will return dates as: `1/7/2021`, whereas the API is looking for an ISO 8601 data which is delineated by dashes not slashes and is in reverse order: `2021-01-07`. 

I get your point about this solution though. Another more reliable solution might be to get a substring of:

```
new Date().toISOString()
```

Which outputs:

```
2021-01-07T12:13:46.567Z
```

tofumatt: Ah, yeah, I think that's a bit better and more clear regarding intent. Happy to go with that 👍🏻

tofumatt (changes requested):
Let's use the `toISOString()` method you mentioned instead of relying on Sweden's time format. After that this looks good to go. 👍🏻

assets/js/modules/adsense/components/dashboard/DashboardSummaryWidget.js
benbowler: Update and back over to you.

tofumatt (changes requested):
The diff here is pretty massive, looks like something else snuck in (an unintentional merge/rebase issue or something?)



Screenshot 2021-01-13 at 21 09 35

Can you have a look and tidy up the branch? After that just ping me and I'll take a look ASAP!

tofumatt (dismissed):

assets/js/modules/adsense/components/dashboard/DashboardSummaryWidget.js
```diff
@@ -65,8 +68,16 @@ function DashboardSummaryWidget() {
 			metrics,
 		};
 
+		// Get the first day of the month an ISO 8601 date string without the time.
+		const startOfMonth = new Date(
+			new Date( referenceDate ).getFullYear(),
+			new Date( referenceDate ).getMonth(),
+			1
+		).toISOString().substr( 0, 10 );
```
aaemnnosttv, line 76: We already have a utility for formatting a `Date` into our `YYYY-MM-DD` format, which we should use here instead of the ISO+substring approach. See:
https://github.com/google/site-kit-wp/blob/37c5b0456cd9aa8545e139a0adbc8b6aa7102039/assets/js/googlesitekit/datastore/user/utils/get-date-string.js#L38

This is the same utility used by our datastore date range selectors, so as long as we use this utility, we know that it will be formatted correctly.

While this is currently a utility of the `core/user` datastore, it should really be a generic date utility, so I'll open a new issue for moving these.

assets/js/modules/adsense/components/dashboard/DashboardSummaryWidget.js
```diff
@@ -51,8 +51,11 @@ function DashboardSummaryWidget() {
 	} = useSelect( ( select ) => {
 		const metrics = [ 'EARNINGS', 'PAGE_VIEWS_RPM', 'IMPRESSIONS' ];
 
+		const referenceDate = select( CORE_USER ).getReferenceDate(); // The issues is the reference date needs to be offset to match the data...
```
aaemnnosttv, line 54: A bit of an oddly worded comment here which doesn't really provide useful context, so let's remove it.
```suggestion
		const referenceDate = select( CORE_USER ).getReferenceDate();
```

assets/js/modules/adsense/components/dashboard/DashboardSummaryWidget.js
```diff
@@ -65,8 +68,16 @@ function DashboardSummaryWidget() {
 			metrics,
 		};
 
+		// Get the first day of the month an ISO 8601 date string without the time.
```
aaemnnosttv, line 71:
```suggestion
		// Get the first day of the month as an ISO 8601 date string without the time.
```

aaemnnosttv (changes requested):
Looks good @benbowler although this requires one minor change for consistency with our current date handling/formatting with a few other minor corrections.

assets/js/modules/adsense/components/dashboard/DashboardSummaryWidget.js
aaemnnosttv: Aforementioned new issue: https://github.com/google/site-kit-wp/issues/2641

benbowler: Thanks @aaemnnosttv, this makes sense to me. I've added the blocking association with this ticket.

benbowler: I resolved this using the existing location of the helper, it can be updated later in #2641

aaemnnosttv:
@googlebot I consent.

aaemnnosttv (approved):
Awesome, thanks!

PR #2571, opened 2020-12-30
Pass actual dates in API calls in Analytics widgets

benbowler:
## Summary

Pass actual dates in API calls in Analytics widgets
Addresses issue #2236

## Relevant technical choices

Completed to the IB. I tried to keep the use of selectors/destructuring/assignment consistent across all the other widgets.
I also had to update the data for the Goals Widget story as it wasn't working even on `develop`.

## Checklist


stories/module-analytics-components.stories.js
```diff
@@ -29,26 +29,6 @@ import { STORE_NAME } from '../assets/js/modules/analytics/datastore/constants';
...
-	pageDashboardUniqueVisitorsVisitorArgs,
-	pageDashboardUniqueVisitorsVisitorData,
-	dashboardBounceRateWidgetArgs,
-	dashboardBounceRateWidgetData,
-	dashboardGoalsWidgetArgs,
-	dashboardGoalsWidgetData,
-	dashboardUniqueVisitorsSparkArgs,
-	dashboardUniqueVisitorsVisitorArgs,
-	dashboardUniqueVisitorsSparkData,
-	dashboardUniqueVisitorsVisitorData,
```
aaemnnosttv, line 51: The data these stories use shouldn't need to change. If for some reason the data does need to be updated (curious why this would be), it should be updated in the fixture rather than inlined.

assets/js/modules/analytics/components/dashboard/DashboardBounceRateWidget.js
```diff
@@ -55,9 +55,13 @@ function DashboardBounceRateWidget() {
 		const profileID = store.getProfileID();
 		const internalWebPropertyID = store.getInternalWebPropertyID();
 
+		const { compareStartDate, compareEndDate, startDate, endDate } = select( CORE_USER ).getDateRangeDates( { offsetDays: DATE_RANGE_OFFSET, compare: true, weekdayAlign: true } );
```
aaemnnosttv, line 58: This line is a bit long and would be better spread across a few. I thought we were spreading the date range dates into the args but it looks like we aren't doing that anymore.
Let's do something like this https://github.com/google/site-kit-wp/blob/8dc8de99f9225f62972c2bb82f67f6e3d39651ad/assets/js/modules/adsense/components/dashboard/AdSensePerformanceWidget.js#L42-L47

The args to `getDateRangeDates` should also be on separate lines though.

A good rule of thumb here is 1-2 keys destructured or shorthand key/value is okay. More than that destructuring, or when using key/values without the shorthand should probably be on separate lines.

```suggestion
		const {
			compareStartDate,
			compareEndDate,
			startDate,
			endDate
		} = select( CORE_USER ).getDateRangeDates( {
			offsetDays: DATE_RANGE_OFFSET,
			compare: true,
			weekdayAlign: true,
		} );
```

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget.js
```diff
@@ -104,7 +107,7 @@ function DashboardAllTrafficWidget() {
 				mdc-layout-grid__cell
 				mdc-layout-grid__cell--span-12
 			">
-				<ReportZero moduleSlug="analytics" />;
```
aaemnnosttv, line 107: Nice catch 👍

aaemnnosttv (changes requested):
Thanks @benbowler ! This is looking good for the most part. The main thing to address is the change to the stories. Let's figure out what's going on there and fix that if we can. The rest is looking good though 👍

stories/module-analytics-components.stories.js
benbowler: Hey @aaemnnosttv, I've inlined the widget data based on [this comment from Felix](https://github.com/google/site-kit-wp/issues/2459#issuecomment-736892298) and to keep it consistent with the other widget tickets I've worked on.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget.js
benbowler: ;)

stories/module-analytics-components.stories.js
felixarntz: @aaemnnosttv The problem here is that fixtures aren't a great approach because the data is dependent on another value passed here, the `referenceDate`. I think that would justify its own issue though. For now, although I agree it's terrible it's most appropriate to inline the data since it's dependent and it's not a static "fixture".

assets/js/modules/analytics/components/dashboard/DashboardGoalsWidget.js
```diff
@@ -55,9 +55,13 @@ function DashboardGoalsWidget() {
 		const profileID = store.getProfileID();
 		const internalWebPropertyID = store.getInternalWebPropertyID();
 
+		const { compareStartDate, compareEndDate, startDate, endDate } = select( CORE_USER ).getDateRangeDates( { offsetDays: DATE_RANGE_OFFSET, compare: true, weekdayAlign: true } );
```
felixarntz, line 58: Similar to the previous comment by @aaemnnosttv, this would be better readable if it was split between multiple lines. We should only put objects in a single line if they have one or two short properties and the line length overall wouldn't suffer.

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.js
```diff
@@ -55,35 +55,38 @@ function DashboardUniqueVisitorsWidget() {
...
-			commonArgs.url = url;
-		}
+		const { compareStartDate, compareEndDate, startDate, endDate } = select( CORE_USER ).getDateRangeDates( { offsetDays: DATE_RANGE_OFFSET, compare: true } );
+
 		const sparklineArgs = {
+			compareStartDate,
+			compareEndDate,
+			startDate,
+			endDate,
+			url,
```
felixarntz, line 67: Adding this in here unconditionally is an unrelated change against how it was before. The args should only include `url` if it is a truthy value.

felixarntz (changes requested):
Mostly LGTM now, just a few things, which I can address here quickly to get it merged (including updating to resolve conflicts with #2425 which touched some related areas). I've opened #2584 for introducing a factory instead of inlining massive mock responses.

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.js
```diff
@@ -55,35 +55,38 @@ function DashboardUniqueVisitorsWidget() {
...
 
 		const url = select( CORE_SITE ).getCurrentEntityURL();
-		if ( url ) {
-			commonArgs.url = url;
-		}
+		const { compareStartDate, compareEndDate, startDate, endDate } = select( CORE_USER ).getDateRangeDates( { offsetDays: DATE_RANGE_OFFSET, compare: true } );
+
 		const sparklineArgs = {
+			compareStartDate,
+			compareEndDate,
```
felixarntz, line 64: These two shouldn't be here, the `sparklineArgs` object previously didn't have `multiDateRange: 1`, so it should only have `startDate` and `endDate` to keep that behavior.

felixarntz (changes requested):

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.js
felixarntz, line 64: These two shouldn't be here, the `sparklineArgs` object previously didn't have `multiDateRange: 1`, so it should only have `startDate` and `endDate` to keep that behavior. Only `args` should have the two extra properties because that had `multiDateRange: 1`.

felixarntz (changes requested):

felixarntz (approved):
Good for merge! Let's focus on #2584 soon to improve the inlined mock responses.

Issue #2499, opened 2020-12-07
Only point to specific question and back when editing user input question


felixarntz:
Currently, when previewing user input questions at the end of the flow or when viewing the responses from within settings, when clicking "Edit", you land on the `googlesitekit-user-input` screen with the regular behavior that you'll have to go through all (following) questions.

This should be updated so that you only have to re-answer the question you're intending to edit, rather than having to click through the following ones again too.

@tofumatt already highlighted that UX flaw in a comment for #2041 around the settings UI. That has now been revisited and should be enhanced for _both_ the settings UI and the regular preview at the end of the flow.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* When clicking "Edit" in the Preview UI of the `googlesitekit-user-input` screen, the question displayed should not surface "Previous" and "Next" buttons, but rather a single "Back" button (is that the best wording here?). That button should internally set the new responses in the datastore and then return the user to the preview UI (rather than requiring to go through all following questions again).
* When clicking "Edit" in the settings UI around user input questions, the question displayed should not surface "Previous" and "Next" buttons, but rather a single "Submit" button. That button should internally set the new responses in the datastore _and_ immediately submit it to the REST API / authentication service (**different from the first bullet point**) and then return the user to the settings UI (rather than requiring to go through all following questions again).
* The copy and behavior of the buttons on a user input question should be contextually modifiable (e.g. via props) to cover the above use-cases.

**Note**: The difference in behavior between the two cases is due to the fact that in the user input screen, responses are submitted via the "Submit" button in the preview UI (which is also why the question button there should not say "Submit"). In the settings UI there is no "Submit" button so changing the responses for a question should immediately submit it.

## Implementation Brief

We should use hash-based URLs to track and set the current question/position/behaviour of the `UserInputQuestionnaire` component, rather than the internal state variable (`activeSlug`) currently used.

While in the future this could (and likely should) be moved to use a real router like React Router, for now we'll implement the URL handling with the `react-use` `useHash` hook mentioned in #2041. What would make sense is supporting URLs like:

* `/wp-admin/admin.php?page=googlesitekit-user-input#/` -> Show the User Input preview UI with all questions answered (if already answered). Redirect to `#/question/1` if not answered.
* `/wp-admin/admin.php?page=googlesitekit-user-input#/question/1` -> Show the first question with the ability to advance to the next one.
[...]

PR #2573, opened 2020-12-31
Only point to specific question and back when editing user input question

benbowler:
You can now edit a single question from both the user-input and settings pages and return instantly to the previous page without having to go through the entire question flow.

## Summary

Addresses issue #2499

## Relevant technical choices

I had to implement this quite differently to the IB because #2041 used the query string, not URL hashes for its implementation. I achieved the same outcome. One larger change I had to make was moving the `submitChanges` logic from the `UserInputPreview` component to the parent `UserInputQuestionnaire` component so that a question can now trigger a submit. 

## Checklist


benbowler:
As this is blocked by #2041 I've created this PR against that branch until it's complete.

assets/js/components/user-input/UserInputQuestionnaire.js
```diff
@@ -42,13 +43,16 @@ import {
...
+import { STORE_NAME as CORE_SITE } from '../../googlesitekit/datastore/site/constants';
+const { useSelect, useDispatch } = Data;
 
 export default function UserInputQuestionnaire() {
 	const steps = [ ...USER_INPUT_QUESTIONS_LIST, 'preview' ];
 
 	const [ activeSlug, setActiveSlug ] = useQueryArg( 'question', steps[ 0 ] );
 	const [ redirectURL ] = useQueryArg( 'redirect_url' );
+	// The single variable is used to edit *just one question* from the summary screens on the user input page and the settings page.
+	const [ single, setSingle ] = useQueryArg( 'single', false );
```
eugene-manuilov, line 55: @benbowler let's move this hook to the `UserInputQuestionWrapper` component to escape a need to pass it via props. Just destruct the first value only in the wrapper component: `const [ single ] = useQueryArg( ... )`

assets/js/components/user-input/UserInputQuestionnaire.js
```diff
@@ -81,7 +85,11 @@ export default function UserInputQuestionnaire() {
 		setActiveSlug( steps[ activeSlugIndex + 1 ] );
 	}, [ activeSlugIndex ] );
 
-	const goTo = useCallback( ( num = 1 ) => {
+	const goTo = useCallback( ( num = 1, shouldSetSingle = false ) => {
+		// Add the single edit context to the query string when navigating to a specific question on the user-input page.
+		if ( shouldSetSingle ) {
+			setSingle( 'user-input' );
+		}
```
eugene-manuilov, line 92: @benbowler this functionality doesn't seem to be needed in the questionary component. Let's add a new callback in the preview component that will set this and call the `goTo` function passed to the component via props.

Also, please, use the `useQueryArg` hook in the preview component to get the setter function instead of passing it via props.

assets/js/components/user-input/UserInputQuestionnaire.js
```diff
@@ -189,9 +245,13 @@ export default function UserInputQuestionnaire() {
 
 			{ activeSlug === 'preview' && (
 				<UserInputPreview
+					isNavigating={ isNavigating }
+					isSavingSettings={ isSavingSettings }
+					error={ error }
+					submitChanges={ submitChanges }
+					goToPreview={ goToPreview }
```
eugene-manuilov, line 252: @benbowler the `goToPreview` property isn't needed for the `UserInputPreview` component. Let's get rid of it.

assets/js/components/user-input/UserInputQuestionnaire.js
```diff
@@ -107,6 +148,9 @@ export default function UserInputQuestionnaire() {
 					questionNumber={ 1 }
 					title={ __( 'Which best describes your team/role in relation to this site?', 'google-site-kit' ) }
 					description={ __( 'This will help Site Kit show tips that help you specifically in your role.', 'google-site-kit' ) }
+					single={ single }
+					submitChanges={ submitChanges }
+					goToPreview={ goToPreview }
```
eugene-manuilov, line 153: @benbowler instead of passing it to every wrapper, let's leverage `next` and `nextLabel` props. It can be done like this:

```javascript
let nextCallback = next;
let nextLabel = undefined;

if ( single === 'user-input' ) {
    nextCallback = goToPreview;
    nextLabel = __( 'Update', 'google-site-kit' );
} else if ( single === 'settings' ) {
    nextCallback = submitChanges;
    nextLabel = __( 'Submit', 'google-site-kit' );
}

...

return (
    ...
    <UserInputQuestionWrapper
        next={ nextCallback }
        nextLabel={ nextLabel }
    />
    ...
);
```

> For the last question you can use `nextLabel={ nextLabel || __( 'Preview', 'google-site-kit' ) }` to make sure it works correctly.

In this case, we won't need additional logic in the wrapper component.

eugene-manuilov (changes requested):
Thanks, @benbowler! It's a good start, but we need to adjust your logic a little bit. Please, check the comments below and let me know if you need additional details.

assets/js/components/user-input/UserInputPreview.js
```diff
@@ -25,52 +25,33 @@ import PropTypes from 'prop-types';
...
+export default function UserInputPreview(
+	{
+		noFooter,
+		back,
+		goTo,
+		isNavigating,
+		isSavingSettings,
+		error,
+		submitChanges,
+	} ) {
```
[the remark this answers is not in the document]
benbowler: Is this format of prop destructuring a project-wide standard?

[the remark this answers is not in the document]
eugene-manuilov: @benbowler I don't think we have a project-wide standard for it but in general, it's much easier to read if props with many elements are destructed in the function body.

benbowler:
Hey @eugene-manuilov, thanks for the comments, I've been through and updated the PR.

assets/js/components/user-input/UserInputPreview.js
```diff
@@ -79,80 +58,90 @@ export default function UserInputPreview( { noFooter, back, goTo, redirectURL }
 		USER_INPUT_ANSWERS_ROLE,
 	} = getUserInputAnwsers();
 
+	const { error } = useSelect( ( select ) => ( {
+		error: select( CORE_USER ).getErrorForAction( 'saveUserInputSettings', [] ),
+	} ) );
```
eugene-manuilov, line 63: @benbowler :man_facepalming: excuse me for the incorrect suggestion at the beginning, but we need to move error handling back to the `UserInputQuestionnaire` component and render it there instead of keeping it in the preview component. The idea here is that we need to display error no matter which screens we are on (a single question or the preview).

assets/js/components/user-input/UserInputPreview.js
```diff
@@ -79,80 +58,90 @@ export default function UserInputPreview( { noFooter, back, goTo, redirectURL }
...
+
+	const [ single, setSingle ] = useQueryArg( 'single', false );
+
+	const editSingle = useCallback(
+		( questionNumber ) => {
+			setSingle( 'user-input' );
+			goTo( questionNumber );
+		},
+		[ single ],
+	);
```
eugene-manuilov, line 73: The `single` value isn't needed here. We can simplify it to be:

```suggestion
	const [ , setSingle ] = useQueryArg( 'single', false );

	const editSingle = useCallback( ( questionNumber ) => {
		setSingle( 'user-input' );
		goTo( questionNumber );
	}, [] );
```

assets/js/components/user-input/UserInputQuestionWrapper.js
```diff
@@ -94,6 +97,19 @@ export default function UserInputQuestionWrapper( props ) {
...
+								</Button>
+							) }
+							{ single && single === 'settings' && (
+								<Button
+									onClick={ submitChanges }
+									disabled={ values.filter( ( value ) => value.trim().length > 0 ).length === 0 }
+								>
+									{ __( 'Submit', 'google-site-kit' ) }
+								</Button>
+							) } */ }
```
eugene-manuilov, line 112: Let's remove comments:

```suggestion
```

assets/js/components/user-input/UserInputQuestionWrapper.js
```diff
@@ -81,7 +84,7 @@ export default function UserInputQuestionWrapper( props ) {
 
 					{ isActive && (
 						<div className="googlesitekit-user-input__buttons">
-							{ back && (
+							{ ! single && back && (
```
eugene-manuilov, line 87: Let's use the same approach as we use for the next callback in the `UserInputQuestionnaire` component:

```javascript
	// Update the callbacks and labels for the questions if the user is editing a *single question*.
	let backCallback = back;
	let nextCallback = next;
	let nextLabel;

	if ( single === 'user-input' ) {
		// When the user is editing a single question in the user-input screen send them back to the preview when they click Update.
		nextCallback = goToPreview;
		nextLabel = __( 'Update', 'google-site-kit' );
		backCallback = undefined;
	} else if ( single === 'settings' ) {
		// When the user is editing a single question from the settings screen, submit changes and send them back to the settings pages when they click Submit.
		nextCallback = submitChanges;
		nextLabel = __( 'Submit', 'google-site-kit' );
		backCallback = undefined;
	}
```

In this case we can escape changing the wrapper component at all.

assets/js/components/user-input/UserInputQuestionnaire.js
```diff
@@ -187,11 +242,14 @@ export default function UserInputQuestionnaire() {
 				</UserInputQuestionWrapper>
 			) }
 
-			{ activeSlug === 'preview' && (
+			{ activeSlug === 'preview' && ( isSavingSettings || isNavigating ) && (
+				<ProgressBar />
+			) }
```
eugene-manuilov, line 247: We need to display this progress bar not only when we are on the preview step. We need to display it when we are saving a single question as well. Let's move it above, something like this:

```javascript
if ( isSavingSettings || isNavigating ) {
	return (
		<ProgressBar />
	);
}

return (
	<Fragment>
		<ProgressBar
			height={ 0 }
			indeterminate={ false }
			progress={ ( activeSlugIndex + 1 ) / USER_INPUT_QUESTIONS_LIST.length }
		/>

		...
	</Fragment>
);
```

Also, please, make sure the progress bar has the same styles as it had when it was in the preview component.

assets/js/components/user-input/UserInputQuestionnaire.js
```diff
@@ -92,6 +94,55 @@ export default function UserInputQuestionnaire() {
...
+			setIsNavigating( false );
+		}
+	}, [ dashboardURL ] );
+
+	const goToPreview = useCallback( () => {
+		setActiveSlug( steps[ steps.length - 1 ] );
+	}, [ activeSlugIndex ] );
+
+	// The single variable is used to edit *just one question* from the summary screens on the user input page and the settings page.
+	const [ single ] = useQueryArg( 'single', false );
```
eugene-manuilov, line 130: Let's move this `useQueryArg` hook to other query arg hooks defined above:

```javascript
	const [ activeSlug, setActiveSlug ] = useQueryArg( 'question', steps[ 0 ] );
	const [ redirectURL ] = useQueryArg( 'redirect_url' );
	const [ single ] = useQueryArg( 'single', false );
```

eugene-manuilov (changes requested):
Thanks, @benbowler! It's better now but there are a few more things to address.

assets/js/components/user-input/UserInputPreview.js
benbowler: Are you happy with the error message displaying here outside the questions box? If we want it inside the box we will need to pass it as a prop to `UserInputQuestionWrapper` instead.

Screenshot 2021-01-05 at 16 59 23

benbowler:
Hey @eugene-manuilov, I've updated everything else and the `ErrorNotice` component looks like so:

Screenshot 2021-01-05 at 17 14 56

assets/js/components/user-input/UserInputPreview.js
benbowler: I've inlined this now so that it looks like the following:

Screenshot 2021-01-06 at 10 38 32
Screenshot 2021-01-06 at 10 39 21

assets/js/components/user-input/UserInputPreview.js
benbowler: I wasn't able to do this because the `useQueryArgs` hook doesn't update in the parent component when set in the child component. We'd need to work on the hook some more to get this to work as expected.

benbowler:
Hey @eugene-manuilov, back with you. I inlined the errors into each component here's how it looks:

Screenshot 2021-01-06 at 10 38 32
Screenshot 2021-01-06 at 10 39 21

I wasn't able to complete the change to [set the single URL argument in the child component because the hook doesn't currently work as expected there](https://github.com/google/site-kit-wp/pull/2573#discussion_r552508347).

benbowler:
All the rest of the changes have been made.

benbowler:
I can also confirm that the design of the `ProgressBar` component still matches `develop`:

2499-edit-specific-questions-user-input

eugene-manuilov:
> I can also confirm that the design of the `ProgressBar` component still matches `develop`:
> 
> 2499-edit-specific-questions-user-input

I mean the progress bar which appears when you save settings, not the progress at the top which hasn't been affected by your changes.

In the develop branch:
develop

In your branch:
after

benbowler:
@eugene-manuilov I've also updated the ProgressBar to show as it does in develop on save.

benbowler:
JS test failing is from [base branch](https://github.com/google/site-kit-wp/pull/2476).

assets/js/components/user-input/UserInputQuestionnaire.js
```diff
@@ -92,6 +99,75 @@ export default function UserInputQuestionnaire() {
...
+	}
+
+	if ( isSavingSettings || isNavigating ) {
+		return (
+			<Fragment>
+				<ProgressBar
+					height={ 0 }
+					indeterminate={ false }
+					progress={ ( activeSlugIndex + 1 ) / USER_INPUT_QUESTIONS_LIST.length }
+				/>
```
eugene-manuilov, line 159: Let's assign it to a variable and use it to escape duplicates:

```javascript
const settingsProgress = (
    <ProgressBar
        height={ 0 }
        indeterminate={ false }
        progress={ ( activeSlugIndex + 1 ) / USER_INPUT_QUESTIONS_LIST.length }
    />
);

if ( isSavingSettings || isNavigating ) {
    return (
        <Fragment>
            { settingsProgress }
            <div> ... </div>
        </Fragment>
    );
}

return (
    <Fragment>
        { settingsProgress }
        ...
    </Fragment>
);
```

assets/js/components/user-input/UserInputQuestionnaire.js
```diff
@@ -187,11 +272,12 @@ export default function UserInputQuestionnaire() {
 				</UserInputQuestionWrapper>
 			) }
 
-			{ activeSlug === 'preview' && (
+			{ activeSlug === 'preview' && ! isSavingSettings && ! isNavigating && (
```
eugene-manuilov, line 275: This is not needed because if either `isSavingSettings` or `isNavigating` is true, we will return earlier.

```suggestion
			{ activeSlug === 'preview' && (
```

eugene-manuilov (changes requested):
Thanks, @benbowler! The last two comments from me.

benbowler:
@eugene-manuilov all requested changes resolved, thanks for working through this with me.

eugene-manuilov (approved):
Awesome! Thanks, @benbowler!

eugene-manuilov:
@benbowler i am going to leave this ticket in the Code Review column until the PR for #2041 is merged.

felixarntz (approved):
Awesome work!

Issue #2522, opened 2020-12-10
Implement Search Console updates to rely on latest API version and display fresh data


felixarntz:
[Search Console recently released a big API update](https://developers.google.com/search/blog/2020/12/search-console-api-updates) relevant for Site Kit: Related to #2284 (where we actually incorrectly assumed this was already in place), we can now display "fresh data" in Search Console reports via the API.

We should migrate from using the legacy API `https://www.googleapis.com/webmasters/v3/` to `https://searchconsole.googleapis.com/` as the new functionality is only available on the latter. Other than that though, the new API is fully backward-compatible, in fact it uses the same URLs (including "legacy prefix" `/webmasters/v3/` within the new API) to achieve that.

We can easily take care of this migration by updating the Google API client services library to the latest version and relying on the [`SearchConsole` class](https://github.com/googleapis/google-api-php-client-services/blob/master/src/Google/Service/SearchConsole.php) instead of the [`Webmasters` class](https://github.com/googleapis/google-api-php-client-services/blob/master/src/Google/Service/Webmasters.php). We can then set the new `dataState` argument via the new [request class](https://github.com/googleapis/google-api-php-client-services/blob/master/src/Google/Service/SearchConsole/SearchAnalyticsQueryRequest.php) to include fresh data in the API responses. In combination with that change, we can now finally change the day offset to 1 day before, from currently 2 days before.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The Search Console module should rely on the [new `SearchConsole` API](https://github.com/googleapis/google-api-php-client-services/blob/master/src/Google/Service/SearchConsole.php) instead of the [legacy `Webmasters` API](https://github.com/googleapis/google-api-php-client-services/blob/master/src/Google/Service/Webmasters.php).
    * The `webmasters` service of the module should accordingly be renamed to `searchconsole`.
    * Since we're fully using the PHP client services library, this shouldn't require any actual migration code other than exchanging the PHP classes relied upon (and included in the bundle).
* The `GET:searchanalytics` REST datapoint should be updated to always pass the `dataState` argument with value "all" to the Google API request.
* The Search Console day offset for parsing a date range to actual dates should be updated from 2 to just 1 day delay, in both PHP and JavaScript.

## Implementation Brief

* Use composer to update google/apiclient-services to version 0.156 with `composer update google/apiclient-services `
[...]

PR #2574, opened 2020-12-31
Update to the new Search Console API version

benbowler:
## Summary

Addresses issue #2522

## Relevant technical choices

As mentioned in [a comment on the original issue](https://github.com/google/site-kit-wp/issues/2522#issuecomment-752991239),  the `dataState` property is [only available on `master` branch](https://github.com/googleapis/google-api-php-client-services/blob/master/src/Google/Service/SearchConsole/SearchAnalyticsQueryRequest.php#L22) not [on the v0.156 release](https://github.com/googleapis/google-api-php-client-services/blob/v0.156/src/Google/Service/SearchConsole/SearchAnalyticsQueryRequest.php) as of yet so I've implemented this with the `dev-master` branch of the repository in `composer.json`.

## Checklist


composer.json
```diff
@@ -18,7 +18,7 @@
     "ext-json": "*",
     "ext-openssl": "*",
     "google/apiclient": "^2.0",
-    "google/apiclient-services": "~0.1",
+    "google/apiclient-services": "dev-master",
```
aaemnnosttv, line 21: This is probably okay to change since this `composer.json` is more of a dev-dependency (you can't install a fully-working version of the plugin by requiring `google/google-site-kit`).

It's a bit odd that the new version hasn't been tagged yet but 🤷‍♂️ 

I'm okay with this change if @felixarntz is (we control when its updated anyways) but I would prefer we used a version constraint instead; I understand the latest tag doesn't include `dataState` yet.

aaemnnosttv (approved):
LGTM but asking @felixarntz to review the version constraint change to `google/apiclient-services`

composer.json
felixarntz: Good catch - I hadn't noticed that the change hasn't been released yet. Let's roll this change here back and wait for the next release instead. I've just emailed the maintainer of the PHP library and will keep an eye out for the update. We can have this issue wait in MR until then.

composer.lock
```diff
@@ -72,7 +72,7 @@
             },
             "require": {
                 "firebase/php-jwt": "~2.0||~3.0||~4.0||~5.0",
-                "google/apiclient-services": "~0.13",
+                "google/apiclient-services": "dev-master",
```
felixarntz, line 75: See above, let's revert the `composer.lock` changes.

includes/Modules/Search_Console.php
```diff
@@ -538,12 +540,12 @@ protected function setup_info() {
 	}
 
 	/**
-	 * Get the configured Webmasters service instance.
+	 * Get the configured SearchConsole service instance.
```
felixarntz, line 543: While we're at it, let's add a `since` annotation here, which was previously missing. Since it's a new method though, we can use the `n.e.x.t` tag.
```suggestion
	 * Get the configured SearchConsole service instance.
	 *
	 * @since n.e.x.t
```

felixarntz (changes requested):
Awesome work! LGTM, just some minor things below, TLDR we should wait for the library update.

composer.lock
benbowler: Reverted.

composer.json
```diff
@@ -18,7 +18,7 @@
     "ext-json": "*",
     "ext-openssl": "*",
     "google/apiclient": "^2.0",
-    "google/apiclient-services": "~0.1",
+    "google/apiclient-services": ">=0.156",
```
felixarntz, line 21: @benbowler @aaemnnosttv Shouldn't this be `>0.156` instead? Not a big deal for now though - alternatively we can wait for the new version to have the actual version number and then use `>=` with that one instead.

felixarntz (changes requested):
@benbowler I've updated the `composer.json` here to use the new 0.157 or higher, which was released just 2 days ago and includes the necessary updates.

The production code here is good to go, however this PR breaks the Storybook stories for all Search Console widgets (they display the zero data state instead of their regular loaded state) because the fixtures / reference date are still relying on the 2 day offset. Can you update these accordingly? Simplest would probably be to move the reference dates by one day.

benbowler:
The storybooks are now fixed and this is ready for CR and MR.

felixarntz (approved):
Awesome!

Issue #2587, opened 2021-01-06
Decouple JS registrations


aaemnnosttv:
## Feature Description

In #2415 we introduced a new `provideModuleRegistrations` utility for providing default module registrations to the `core/modules` data store. While this works as-is, it requires us to duplicate and maintain the arguments to `registerModule` in two places: module indexes and in the utility.

Also, we currently have various references to global instances of core API objects, such as when registering a datastore.

We should decouple our registration code from the specific instances that are used and make this sharable in a consistent way across the codebase to avoid such duplication.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All core and module main index modules (e.g. `assets/js/googlesitekit/modules`, `assets/js/modules/analytics`) should export the following named functions, **as needed**
	- `registerStore( registry: WPDataRegistry )`
	- `registerModule( modules: Modules )`
	- `registerWidgets( widgets: Widgets )`
* Existing registration calls on these APIs should be updated to happen within these functions for each module and relevant core API (widgets, modules, + all core datastores)
* The `registerStore` function should be defined and exported from the datastore index
	- This functions should then be re-exported from the main index for external use
* Registration on the global instances of these objects should only happen in a module's respective entry point (e.g. `assets/js/googlesitekit-datastore-forms.js`)
	- Registration should happen at module load time (no need to wrap with `domReady`)
* Test infrastructure should be updated to use these exports for provisioning test instances in the same way by providing the test API object instance to the module's registration functions
* `assets/js/googlesitekit/modules/index.js` should be updated to export a new `createModules( registry: WPDataRegistry ) : Modules` function
	- Requires a `registry` instance and returns a new `Modules` API instance for the given `registry`
	- The `default` export (`Modules` instance) should be removed as the factory function is required to build it
* `assets/js/googlesitekit/widgets/index.js` should be updated to export a new `createWidgets( registry: WPDataRegistry ) : Widgets` function
	- Requires a `registry` instance and returns a new `Widgets` API instance for the given `registry`
	- The `default` export (`Modules` instance) should be removed as the factory function is required to build it

## Implementation Brief

* Update all datastores to export a `registerStore` function from their datastore's index module in place of the current `Data.registerStore` call
	- Re-export this from the module's main index
* Update service modules to export a `registerModule` function from their index for registering their modules on the given `Modules` instance
* Update all modules that register widgets to export a `registerWidgets` function from their index for registering their widgets on the given `Widgets` instance
[...]

PR #2593, opened 2021-01-07
Refactor with dedicated registration functions

aaemnnosttv:
## Summary

Addresses issue #2587 

## Relevant technical choices

* Reworked `withData`'s wrapping component a bit to apply the default date range from the current registry
    * Should be a bit more efficient now with less memory leaks
* See inline comments below  

## Checklist


assets/js/components/higherorder/withData.js
```diff
@@ -141,6 +147,23 @@ const withData = (
...
+			castArray( dataRequest.context ).forEach( ( context ) => {
+				acc[ context ] = acc[ context ] || [];
+				acc[ context ].push( dataRequest );
+			} );
+			return acc;
+		},
+		{}
+	);
 	// ...and returns another component...
 	class NewComponent extends Component {
```
aaemnnosttv, line 168: The changes here will likely be easier to review with whitespace changes hidden
image

assets/js/components/higherorder/withData.js
```diff
@@ -153,100 +176,141 @@ const withData = (
 				moduleRequiringSetup: '',
 			};
 
+			this.handleModuleDataReset = this.handleModuleDataReset.bind( this );
+			this.handleReturnedData = this.handleReturnedData.bind( this );
+			this.addDataRequests = this.addDataRequests.bind( this );
+			this.removeDataRequests = this.removeDataRequests.bind( this );
```
aaemnnosttv, line 182: The main reason for these changes is that initially I had simply updated the filter callback to apply the global date range as a default. This worked for the initial render, but because the callbacks were hooked in once from the component's constructor, they didn't respond to date range changes even when the prop was being passed through correctly.

The component also does a better job of cleaning up after itself now; where it previously would have caused potential memory leaks due to its callbacks never being removed.

assets/js/components/higherorder/withData.js
```diff
@@ -153,100 +176,141 @@ const withData = (
...
+
+		addDataRequests() {
+			const { dateRange } = this.props;
+
+			Object.entries( dataRequestsByContext ).forEach(
+				( [ context, dataRequests ] ) => {
+					addFilter(
+						`googlesitekit.module${ context }DataRequest`,
+						`googlesitekit.withData.${ hashRequests( dataRequests ) }`,
+						( contextRequests ) => {
```
aaemnnosttv, line 226: This is also a bit more efficient than before. Previously we were adding a filter callback for each request, for each of its contexts. Now we only add a filter once for each _context_ and add all relevant requests at once, for each wrapped component.

assets/js/components/higherorder/withData.js
```diff
@@ -153,100 +176,141 @@ const withData = (
...
+								const { type, identifier } = dataRequest || {};
+								// If a dataRequest's module requires setup, set it in the state.
+								// This will cause the setup incomplete component to be rendered in all cases.
+								if ( TYPE_MODULES === type && ! getModulesData()[ identifier ]?.setupComplete ) {
+									this.setState( { moduleRequiringSetup: identifier } );
+									continue;
+								}
+
+								// Apply default date range if not set.
+								const request = requestWithDateRange( dataRequest, dateRange );
```
aaemnnosttv, line 238: Here's where the default date range is applied now, where before it was done in `dataAPI.combinedGet`.

assets/js/modules/adsense/index.js
```diff
@@ -47,17 +41,24 @@ import {
...
-	// IMPORTANT: When updating arguments here, also update the same call in
-	// `provideModuleRegistrations`.
-	Modules.registerModule(
+let isAdBlockerActive = () => {};
+
+export const registerStore = ( registry ) => {
+	registerDataStore( registry );
+	// TODO: fix hack
+	isAdBlockerActive = () => registry.select( STORE_NAME ).isAdBlockerActive();
+};
```
aaemnnosttv, line 58: I think this is okay as a _temporary workaround_, but it is rather ugly; it's the only thing left that isn't a clean change though. `isAdBlockerActive` here is used within the `checkRequirements` function in the `registerModule` call below. I think the better approach here would be to add support for something like "registry checks" which would be created like a registry control, using a function that works like `createRegistryControl`. That way it could still use the registry if needed (which I think is a reasonable dependency for such a function), or not if not and still be very WP-data like as well as consistent with the overall theme of dependency injection in this PR.

I found a need for a similar pattern with the refactored `CompatibilityChecks` since one of the checks needed access to the registry as well so I used a similar pattern, but just handled the registry part in the same way just without the added API/function/abstraction.

I'm happy to open a new issue for this as a follow-up related issue if we're in agreement about the way forward here?

.storybook/config.js
```diff
@@ -61,8 +61,8 @@ const resetGlobals = () => {
 		isFirstAdmin: true,
```
aaemnnosttv, line 61: First thing's first – the changes here will likely be easier to review with whitespace changes hidden

image

assets/js/modules/adsense/index.js
tofumatt: > I think the better approach here would be to add support for something like "registry checks" which would be created like a registry control, using a function that works like createRegistryControl. That way it could still use the registry if needed (which I think is a reasonable dependency for such a function), or not if not and still be very WP-data like as well as consistent with the overall theme of dependency injection in this PR.

That sounds like a good path forward to me, I like it.

tofumatt (approved):
Looks good to me. I like the idea you mentioned for the registry checks, but overall this looks good to me. 👍🏻

assets/js/googlesitekit-data.js
```diff
@@ -25,9 +25,7 @@ if ( typeof global.googlesitekit === 'undefined' ) {
 	global.googlesitekit = {};
 }
 
-if ( typeof global.googlesitekit.data === 'undefined' ) {
-	global.googlesitekit.data = Data;
-}
```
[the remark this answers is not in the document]
aaemnnosttv: Just seemed like an odd precaution to take here. It shouldn't ever be set by anything other than these entry points.

assets/js/modules/adsense/index.js
felixarntz: Why not pass the `registry` (i.e. `Data`) as parameter to a registered `checkRequirements`? I think that's a simpler way which is reasonable. Not sure why we would want to introduce a separate utility for that. If the function wants to use the `registry`, it can - if not, it doesn't need to.

stories/modal-dialog.stories.js
```diff
@@ -40,6 +40,7 @@ storiesOf( 'Global', module )
 				title={ __( 'Modal Dialog Title', 'google-site-kit' ) }
 				subtitle={ __( 'Modal Dialog Subtitle', 'google-site-kit' ) }
 				provides={ provides }
+				handleConfirm={ global.console.log.bind( null, 'Dialog::handleConfirm' ) }
```
felixarntz, line 43: Why was this change needed?

tests/js/utils.js
```diff
@@ -369,18 +306,10 @@ export const freezeFetch = ( matcher ) => {
...
-	registry.registerStore( MODULES_ANALYTICS, modulesAnalyticsStore );
-	registry.registerStore( MODULES_PAGESPEED_INSIGHTS, modulesPageSpeedInsightsStore );
-	registry.registerStore( MODULES_SEARCH_CONSOLE, modulesSearchConsoleStore );
-	registry.registerStore( MODULES_TAGMANAGER, modulesTagManagerStore );
-	registry.registerStore( MODULES_OPTIMIZE, modulesOptimizeStore );
+	[
+		...allCoreStores,
+		...allModules,
+	].forEach( ( { registerStore } ) => registerStore?.( registry ) );
 };
```
felixarntz, line 353: Awesome to see all the simplifications in these functions!

felixarntz (changes requested):
There appears to be a problem with this approach which affects the new widget components.

On the dashboard, using a development build, I get this:
Screen Shot 2021-02-02 at 11 20 07 AM

It works fine using a production build, where the Widget API-powered components are not used.

The code overall looks solid, I left a few specific questions.

stories/modal-dialog.stories.js
aaemnnosttv: `handleConfirm` is a required prop.

assets/js/googlesitekit-widgets.js
```diff
@@ -16,28 +16,22 @@
...
 if ( typeof global.googlesitekit === 'undefined' ) {
 	global.googlesitekit = {};
 }
 
-if ( global.googlesitekit.widgets === undefined ) {
-	global.googlesitekit.widgets = Widgets;
-}
-
-// register default widget areas
-domReady( () => {
```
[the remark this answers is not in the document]
aaemnnosttv: I didn't test translations yet but they should definitely be loaded. Now that translations are loaded per-script using an inline script that runs before the dependent script, we can be sure that translations are immediately available at execution time. It's essentially no different than relying on something in a `Script_Data` global that is a dependency of the script, except the translation loading is automatic.

https://github.com/google/site-kit-wp/blob/1680f18750ea7fc778d39a5a5feab64d0d1b24d0/includes/Core/Assets/Script.php#L119-L137

assets/js/components/data/utils/request-with-date-range.js
```diff
@@ -0,0 +1,47 @@
...
+ * Gets a copy of the given data request object with the data.dateRange populated via filter, if not set.
+ * Respects the current dateRange value, if set.
+ *
+ * @since 1.0.0
+ *
+ * @param {Object} originalRequest Data request object.
+ * @param {string} dateRange       Default date range slug to use if not specified in the request.
+ * @return {Object} New data request object.
+ */
+export const requestWithDateRange = ( originalRequest, dateRange ) => {
```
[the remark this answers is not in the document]
aaemnnosttv: I took a look at how often it's actually being called and with all modules active on the main dashboard, it's only 11 times (and again when the date range changes). The legacy data stuff doesn't get retriggered the same way selectors do so it's not an issue 👍

assets/js/modules/adsense/index.js
aaemnnosttv: @felixarntz that's certainly one solution I considered. I opened #2749 to discuss and address this 👍

aaemnnosttv:
> It works fine using a production build, where the Widget API-powered components are not used.

Thanks @felixarntz – after taking a closer look, the error there was related to recent change where `args.dateRange` was referenced within a `useSelect` where it is no longer set in the `args` object (refactored with actual dates). This resulted in an error in the select function and caused it to return early which had the effect of returning `undefined` which raised the error when the return value was attempted to be destructured. This no longer needed though with the incoming changes from `develop` as the date range was only needed for use with the legacy `getCurrentDateRangeDayCount` utility that was used for passing `num_of_days` in the service URL. This was refactored in #2634.

That said, everything seems to be working fine now, so this should be ready for another pass 👍

felixarntz (approved):
Awesome!

Issue #2478, opened 2020-12-03
Invalid URL error on confirming site kit reset after disconnecting


eugene-manuilov:
## Bug Description

I have encountered an error during smoke testing the latest release (`1.22.0`) which happened after I reset the disconnected site kit. The error itself doesn't seem to block redirect that happens after we confirm resetting which is good. However, it may confuse users, so we still need to address it.

## Steps to reproduce

1. Enable the `Preserve log` option in your browser console;
2. Set up and connect the site kit on your site;
3. Disconnect it by clicking on `Disconnect` button in the user menu;
4. Once the splash screen appears and the `Reset Site Kit` button shows up, click on it to reset the plugin.
5. See the error for a few seconds before you will be redirected and check the console log for more details.

## Screenshots

![Screenshot from 2020-12-03 20-32-59](https://user-images.githubusercontent.com/873049/101074015-ece2da80-35a8-11eb-88f8-868f52a2ee4b.png)

There's also a screencast that I have recorded, sorry unable to upload it here due to file size https://d.pr/v/6KsggE 

## Additional Context

 - PHP Version: 7.4.12
 - OS: [e.g. iOS] Ubuntu
 - Browser [e.g. chrome, safari] Google Chrome
 - Plugin Version [e.g. 22] 1.22.0
 - Device: [e.g. iPhone6] Desktop computer

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* No JS error should occur / flash up in the background after having clicked the Reset button.

## Implementation Brief

* Using `/assets/js/components/ResetButton.js`
	- Edit the `handleUnlinkConfirm` function to remove the line where the state is updated to hide the dialog.
	```
	setDialogActive( false )
	```
	- This is causing a console error where the component has already been unmounted but we are trying to update the state. Since we are redirecting the user to another page, we can omit updating the state to hide the dialog.
* Using `assets/js/googlesitekit/datastore/site/reset.js`
	- Remove `reducerCallback` from `fetchResetStore`
### Test Coverage

* Replace relevant tests in `assets/js/googlesitekit/datastore/site/reset.test.js` which expect state to be changed as a result of the reset to expect that state _does not change_ from the reset action
    - The one exception here is the state related to the fetch request progress, but we should be able to assert that the state is the same before the reset request and once the request is finished.
[...]

PR #2595, opened 2021-01-08
Cleanup reset site kit code, change test accordingly.

kostyalmm:
## Summary

Addresses issue #2478

## Relevant technical choices

* Updated state management while resetting site kit
* Updated tests accordingly for `reset`

## Checklist


assets/js/googlesitekit/datastore/site/reset.test.js
```diff
@@ -85,7 +85,7 @@ describe( 'core/site reset', () => {
 
 				// After a successful reset, `connection` should be `undefined` again.
```
aaemnnosttv, line 86: This comment is no longer accurate.

```suggestion
				// After a successful reset, `connection` state will be updated on the next page load.
```

assets/js/googlesitekit/datastore/site/reset.test.js
```diff
@@ -85,7 +85,7 @@ describe( 'core/site reset', () => {
...
 				const connection = await registry.select( STORE_NAME ).getConnection();
-				expect( connection ).toEqual( undefined );
+				expect( connection ).toEqual( { connected: true, resettable: true } );
```
aaemnnosttv, line 88: The name of this test should also be updated to be more accurate as well.
How about: it "resets connection on server only" ? I think this would be a bit more clear as to what is going on in combination with the updated comment.
Unfortunately GitHub won't let me suggest this change because it's out of range of the current changes.

aaemnnosttv (changes requested):
Looks great, just two small things to address for this to make sense when we come back to look at this in the future 😄

assets/js/googlesitekit/datastore/site/reset.test.js
kostyalmm: Sure. Sounds good.

aaemnnosttv (approved):
Super, thanks!

Issue #2403, opened 2020-11-19
Search Console dashboard infinite loading state when lacking permissions to property


aaemnnosttv:
## Bug Description

Currently the Search Console dashboard will remain in an infinite loading state if the user does not have sufficient permissions to the property. This is visible in the batch data responses in the console.

<details>
<summary>batch response</summary>
	
```json
{
	"modules::search-console::searchanalytics::14bf456def2282c15006e28a71ed9b45": {
		"code": 403,
		"message": "User does not have sufficient permission for site 'https://example.com/'. See also: https://support.google.com/webmasters/answer/2451999.",
		"data": {
			"status": 403,
			"reason": "forbidden"
		}
	},
	"modules::search-console::searchanalytics::8d4c4f45a7123a18487d0b9fb7d2587c": {
		"code": 403,
		"message": "User does not have sufficient permission for site 'https://example.com/'. See also: https://support.google.com/webmasters/answer/2451999.",
		"data": {
			"status": 403,
			"reason": "forbidden"
		}
	}
}
```

</details>

![image](https://user-images.githubusercontent.com/1621608/99720805-f3585900-2ab6-11eb-8b20-aef54ac22ad4.png)

## Steps to reproduce

1. Set up Site Kit
1. Using the Developer Settings plugin, change your Custom Site URL to `https://example.com`
1. Navigate to the Search Console module dashboard page
1. See perpetual loading state

## Additional Context

Tested with 1.21 but confirmed the issue is present in 1.20 as well.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The Search Console dashboard should not display an infinite loading state under any conditions, but specifically when the user does not have sufficient permissions to the property

## Implementation Brief

After digging into this a bit I found that the problem had to do with a bug in the way the legacy dataAPI handled multiple requests for the same request (cache-key). It's related to how when a batch request is made, the individual requests are grouped by request key (cache key), so that each unique data request is only made once to the server. When the response comes back, the handler needs to iterate over all request objects for the cache key for both error handling and resolving the response (sending data back to the caller for each request object).

I believe the problem was due to the line below where an error was raised due to attempted destructuring from `undefined` which stopped iteration and resulted in a rejected promise before data could be resolved (resulting in infinite loading).
https://github.com/google/site-kit-wp/blob/24e4cf02daffd797106b9566a6185229ae2e4983/assets/js/components/data/index.js#L207

[...]

PR #2600, opened 2021-01-08
Fix request resolution bug in batch response handling.

aaemnnosttv:
## Summary

Addresses issue #2403

## Checklist


felixarntz (changes requested):
LGTM regarding the fix - can you update the relevant JS tests so that they pass again?

assets/js/components/data/index.js
```diff
@@ -196,33 +196,35 @@ const dataAPI = {
...
-					console.debug( 'data_error', 'unknown response key ' + key ); // eslint-disable-line no-console
+			Object.entries( results ).forEach( ( [ requestKey, response ] ) => {
+				if ( ! keyIndexesMap[ requestKey ] ) {
+					console.debug( 'data_error', 'unknown response key ' + requestKey ); // eslint-disable-line no-console
 					return;
 				}
 
-				const isError = isWPError( result );
-				if ( isError ) {
-					const { datapoint, type, identifier } = dataRequest[ keyIndexesMap[ key ] ];
```
aaemnnosttv, line 207: This seems to have been possible due to this odd behavior when using an array as a key:

image

This allowed it to work for single requests, but it would fail as soon as there were multiple requests for data with the same cache key, in the same `combinedRequest`. This only affected requests that resulted in an error, but still not good of course.

felixarntz (approved):
Great!

Issue #2457, opened 2020-11-30
Reduce excessive amount of Analytics report requests for AdSense linked check


felixarntz:
When Analytics and AdSense are active, some widgets use the Analytics Reporting API to query for AdSense-related metrics. Those requests result in an error if the AdSense account and the Analytics property are not linked.

This has been a long-standing problem, because there is unfortunately no endpoint in the Analytics Management API that exposes whether such a link exists or not. There is an [endpoint for Google Ads links](https://developers.google.com/analytics/devguides/config/mgmt/v3/mgmtReference/management/webPropertyAdWordsLinks), but none for AdSense links.

The Analytics frontend uses an internal `linkableproperties` endpoint for this information as seen here:
<img width="1669" alt="Screenshot 2020-11-30 at 11 29 45" src="https://user-images.githubusercontent.com/3531426/100654941-6e174480-32ff-11eb-8a51-65522822e6ff.png">

We cannot use it, so it looks like we're limited to relying on the approach we're already using. However, we should explore ways to issue fewer of these erroring requests: Generally we don't cache API error responses, which is why this request is fired basically on every pageload, resulting in an excessive amount of error requests, which could be avoided. We should consider caching certain errors on the client-side too.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `googlesitekit-api` layer's cache approach should be modified so that cache TTL is determined based on the cache entry and is stored with each entry rather than specified manually when reading an entry:
    * `setItem` should receive:
        * A new `ttl` parameter that specifies the entry's time-to-live in seconds and defaults to `3600`. `ttl` should be included in the cache field alongside the existing `timestamp` and `value`.
        * An optional `isError` parameter (default `false`) that indicates whether the cached value is an error. That should also be stored in the cache field as `isError`.
    * `getItem` should have its `cacheTimeToLive` parameter removed. Instead, the found item's `ttl` property value should be used to determine whether it's expired or not. If an item doesn't have a `ttl` property (needed for BC), it should be considered expired.
* The `cacheTTL` argument of the `siteKitRequest` function should be used in the `setItem` call instead of the `getItem` call, based on the above changes.
* The logic for potentially dispatching some store actions, which is currently part of the `catch` clause in `siteKitRequest`, should be moved to a reusable `dispatchAPIError` function. The `trackAPIError` and console error logging parts should not be included in that.
[...]

PR #2605, opened 2021-01-11
Enhancement/2457 excessive analytics requests

asvinb:
## Summary

Addresses issue #2457 

## Checklist


assets/js/googlesitekit/api/cache.js
```diff
@@ -202,19 +203,28 @@ export const getItem = async ( key, cacheTimeToLive = null ) => {
...
+ * @param {*}       value          Value to store in the cache.
+ * @param {Object}  args           Optional object containing ttl, timestamp and isError keys.
+ * @param {number}  args.ttl       Validity of the cached item in seconds.
+ * @param {number}  args.timestamp Timestamp when the cached item was created.
+ * @param {boolean} args.isError   Whether the cached item is an error.
  * @return {Promise} A promise: resolves to `true` if the value was saved; `false` if not (usually because no storage method was available).
  */
-export const setItem = async ( key, value, _timestamp = undefined ) => {
+export const setItem = async ( key, value, {
+	ttl = 60 * 60,
```
eugene-manuilov, line 215:
```suggestion
	ttl = 3600, // One hour.
```

assets/js/googlesitekit/api/cache.js
```diff
@@ -202,19 +203,28 @@ export const getItem = async ( key, cacheTimeToLive = null ) => {
...
  *
- * @param {string} key        Name of cache key.
- * @param {*}      value      Value to store in the cache.
- * @param {number} _timestamp Timestamp to set as the cache data save time.
+ * @param {string}  key            Name of cache key.
+ * @param {*}       value          Value to store in the cache.
+ * @param {Object}  args           Optional object containing ttl, timestamp and isError keys.
+ * @param {number}  args.ttl       Validity of the cached item in seconds.
+ * @param {number}  args.timestamp Timestamp when the cached item was created.
+ * @param {boolean} args.isError   Whether the cached item is an error.
```
eugene-manuilov, line 211:
```suggestion
 * @param {number}  [args.ttl]       Optional. Validity of the cached item in seconds.
 * @param {number}  [args.timestamp] Optional. Timestamp when the cached item was created.
 * @param {boolean} [args.isError]   Optional. Whether the cached item is an error.
```

assets/js/googlesitekit/api/cache.js
```diff
@@ -161,29 +161,30 @@ export const getStorage = async () => {
...
+			const { timestamp, ttl, value, isError } = parsedData;
 
 			// Ensure a timestamp is found, otherwise this isn't a valid cache hit.
 			// (We don't check for a truthy `value`, because it could be legitimately
 			// false-y if `0`, `null`, etc.)
-			if ( parsedData.timestamp && (
-				cacheTimeToLive === null || // Ensure the cached data isn't too old.
-				Math.round( Date.now() / 1000 ) - parsedData.timestamp < cacheTimeToLive
+			if ( timestamp && (
+				ttl === null || ttl === undefined || // Ensure the cached data isn't too old.
```
eugene-manuilov, line 181:
```suggestion
				! ttl || // Ensure the cached data isn't too old.
```

eugene-manuilov (changes requested):
Nice work, @asvinb! Just found a few minor things to improve and it should be good to go. Could you please check it?

includes/Core/Modules/Module.php
```diff
@@ -831,15 +831,21 @@ protected function exception_to_error( Exception $e, $datapoint ) {
...
 		$reconnect_url = '';
 
 		if ( $e instanceof Google_Service_Exception ) {
+
 			$errors = $e->getErrors();
 			if ( isset( $errors[0]['message'] ) ) {
 				$message = $errors[0]['message'];
+
+				if ( preg_match( '#^Restricted metric\(s\):(.*)$#im', $message ) ) {
+					$cache_ttl = ( 10 * 60 );
```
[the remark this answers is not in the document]
asvinb: Updated @eugene-manuilov

includes/Core/Modules/Module.php
```diff
@@ -831,15 +831,21 @@ protected function exception_to_error( Exception $e, $datapoint ) {
...
 		$reason        = '';
 		$reconnect_url = '';
 
 		if ( $e instanceof Google_Service_Exception ) {
+
 			$errors = $e->getErrors();
 			if ( isset( $errors[0]['message'] ) ) {
 				$message = $errors[0]['message'];
+
+				if ( preg_match( '#^Restricted metric\(s\):(.*)$#im', $message ) ) {
```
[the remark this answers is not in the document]
asvinb: Nice. Thanks!

assets/js/googlesitekit/api/cache.js
asvinb: 👍

assets/js/googlesitekit/api/cache.js
asvinb: 👍

assets/js/googlesitekit/api/cache.js
asvinb: 👍

eugene-manuilov (approved):
LGTM!

includes/Core/Modules/Module.php
```diff
@@ -861,6 +867,10 @@ protected function exception_to_error( Exception $e, $datapoint ) {
 			'reason' => $reason,
 		);
 
+		if ( $cache_ttl ) {
+			$data['cacheTTL'] = $cache_ttl;
+		}
+
```
felixarntz, line 873: While this works, the entire PHP logic needed here is specific to the Analytics API, so I don't think it is suitable to be added to the general `Module` class (also see "the `Analytics` exception handling function" reference in the ACs). Let's instead put these changes into the `Analytics` class:
* Modify the overridden `exception_to_error` method, which already includes the necessary check.
* Instead of immediately returning the `WP_Error` returned by the parent method, modify it to add the `cacheTTL` to the `$data` property of the `WP_Error` object based on the checks made.
    * Basically, the existing clause can be moved below the parent method call so that we can then within it update the `WP_Error` as needed before returning it.

felixarntz (changes requested):
Mostly LGTM, however the PHP logic should not be in the general `Module` class since it is specific to Analytics (see ACs and IB).

includes/Core/Modules/Module.php
asvinb: Ah! Makes sense @felixarntz I thought we were going to do that for all "Restricted metrics..." error. The PR has been updated. Can you take a look @eugene-manuilov ?

eugene-manuilov (dismissed):
Looks good!

includes/Modules/Analytics.php
```diff
@@ -1299,15 +1299,33 @@ protected function has_access_to_property( $property_id ) {
...
 
-		return parent::exception_to_error( $e, $datapoint );
+		$error = parent::exception_to_error( $e, $datapoint );
+
+		if ( $cache_ttl && is_wp_error( $error ) ) {
+			$error->add_data(
+				array(
+					'cacheTTL' => $cache_ttl,
+				)
+			);
```
felixarntz, line 1325: This unconditionally overrides any existing `data` that is already present in the `$error` (e.g. the HTTP response code in `status` or the `errorReason`). We'll need to change it to manually alter the `$error->error_data` property if it already has content (modifying properties directly isn't great, but that's all that WordPress core allows here).

felixarntz (changes requested):
Almost good to go, just one final comment to address.

felixarntz (approved):
Fixed the last bit, so this is good to go now!

Issue #2497, opened 2020-12-07
Make data store navigation-aware


aaemnnosttv:
## Feature Description

Currently there are a few instances in the plugin where we need to account for whether or not a navigation is in progress, so that we can show a progress bar after a form submission for example. Until now, we have managed to work around this on a case-by-case basis and in a local component scope.

As a reoccurring need, it makes sense to address this in a more holistic way by exposing navigation state to consumers via the data store. There are also cases where it makes sense to be managed at a bit of a higher level as a navigation may not "belong" to any specific component, or may not even be triggered by UI within Site Kit.

See related thread here: https://github.com/google/site-kit-wp/issues/2261#issuecomment-737322569

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a new `core/location` data store, defined in `assets/js/googlesitekit/datastore/location/`. It should have an entrypoint file `assets/js/googlesitekit-datastore-location.js` which also has a corresponding PHP asset. The asset should become part of the core dependencies.
* The store should include:
    * A `navigateTo( url )` action, which sets a flag that a navigation is happening and then redirects the user to that URL (via `global.location.assign`).
    * A `isNavigating()` selector, which returns a boolean based on whether or not a navigation is currently in progress.
    * A `getNavigateURL()` selector, which returns the URL that is currently being navigated to (if a navigation is currently in progress).
* Existing workarounds for handling in-progress navigation (e.g. post module setup submit) should be refactored to use the new `isNavigating` selector.
* The `UnsatisfiedScopesAlert` should be updated to only show up if there is currently no navigation happening using the new `isNavigating` selector (potentially this should be more specific to only navigations to OAuth connect URLs).

## Implementation Brief

- Create new folder `/assets/js/googlesitekit/location`
- Using newly created folder:
	- Create `constants.js` which exports the `STORE_NAME` constant which is set to `core/location`
	- Create `navigation.js` which will contain the different selectors.
- Create `/assets/js/googlesitekit/location/index.js` taking another datastore as example.
- Create `/assets/js/googlesitekit-datastore-location.js` which imports `assets/js/googlesitekit/datastore/location`
- Edit `webpack.config.js` to include a new entry point file `assets/js/googlesitekit-datastore-location.js`
	```js
	'googlesitekit-datastore-location': './assets/js/googlesitekit-datastore-location.js',
	```
- Using `/assets/js/googlesitekit/location/navigation`
	- Add the following action:
[...]

PR #2609, opened 2021-01-11
Enhancement/2497 navigation aware datastores

eugene-manuilov:
## Summary

Addresses issue #2497

## Relevant technical choices

* Changes in the user input settings datastore are required because the original `isFetchingSaveUserInputSettings` has never worked as expected.

## Checklist


assets/js/googlesitekit/datastore/user/user-input-settings.js
```diff
@@ -159,6 +177,18 @@ const baseResolvers = {
...
+	 * Determines whether the user input settings are being saved or not.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {boolean} TRUE if the user input settings are being saved, otherwise FALSE.
+	 */
+	isSavingUserInputSettings( state ) {
+		return !! state?.isSavingInputSettings;
+	},
```
felixarntz, line 190: I saw your comment on the PR, but can you clarify what the problem was with `isFetchingSaveUserInputSettings`? Since the above calls `fetchSaveUserInputSettings`, why can't we rely on that?

assets/js/components/setup/ModuleSetup.js
```diff
@@ -65,9 +67,7 @@ export default function ModuleSetup( { moduleSlug } ) {
 			redirectURL = getSiteKitAdminURL( 'googlesitekit-dashboard', args );
 		}
 
-		delay( function() {
-			global.location.replace( redirectURL );
-		}, 500, 'later' );
+		navigateTo( redirectURL );
```
felixarntz, line 70: We need to rely on this one in a couple more components. The original specific goal of the issue was to fix the scopes notification from showing up when activating a module before getting redirected to OAuth. In order to satisfy that criteria, we'll need to update the `assets/js/components/settings/SetupModule.js` file accordingly.

There are various other files where we redirect to the OAuth flow, all of which would be worth adjusting, to fix the same problem:
* `ActivateModuleCTA`
* `CompleteModuleActivationCTA`
* `PermissionsModal`

Or, even better, we could replace all usages of `global.location.assign` with the `navigateTo` action to set the right precedent also for the future.

assets/js/googlesitekit/datastore/location/navigation.js
```diff
@@ -0,0 +1,122 @@
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {string|null} The current navigating URL if doing navigation, otherwise NULL.
+	 */
+	getNavigateURL( state ) {
+		return state.navigatingTo || null;
+	},
+
```
felixarntz, line 112: As an additional suggestion, maybe we could introduce another selector `isNavigatingTo( url )` that supports receiving either a URL string or a regular expression? That would allow us to perform more accurate checks - to be strict, the OAuth scopes warning should _only_ not show up if the user is being redirected to the OAuth URL. If they were redirected to some other random URL, that wouldn't really affect the state of OAuth scopes.

Just an idea - if we decide it's worth it, could also be a follow-up issue. But maybe it's overkill.

felixarntz (changes requested):
Awesome! Mostly LGTM, but we need to use `navigateTo` at least in one other component to satisfy the ACs - and it would probably be best to use it for all redirects.

assets/js/googlesitekit/datastore/user/user-input-settings.js
eugene-manuilov: @felixarntz the `isFetchingSaveUserInputSettings` selector works only when we pass the appropriate argument (an array of user input settings without scopes) into it which we can't get in the user input components. So, it has never worked as expected.

assets/js/components/setup/ModuleSetup.js
eugene-manuilov: Yep, updated all usages of `global.location.assign` with the `navigateTo`.

assets/js/googlesitekit/datastore/location/navigation.js
eugene-manuilov: Ok, added the `isNavigateTo` selector and updated the `UnsatisfiedScopesAlert` component to use it.

assets/js/components/notifications/UnsatisfiedScopesAlert.js
```diff
@@ -58,12 +59,13 @@ function mapScopesToModuleNames( scopes ) {
 }
 
 export default function UnsatisfiedScopesAlert() {
+	const isNavigating = useSelect( ( select ) => select( CORE_LOCATION ).isNavigatingTo( /\/o\/oauth2/i ) );
```
felixarntz, line 62: This is not fully accurate and doesn't fix the underlying issue, since Site Kit redirects to a WP Admin URL containing a `googlesitekit_connect=1` query parameter. Only from there the user is redirected to the authentication service's OAuth proxy endpoint.

It's a good idea to keep the `/o/oauth2` in here to be compatible, should we switch to using OAuth in a more typical way, but we need to also support the real URL Site Kit uses for this purpose.

felixarntz (approved):
Updated the last bit here to unblock, so good to go!

Issue #2422, opened 2020-11-24
Implement `UserCountGraph` component for new All Traffic widget


felixarntz:
After scaffolding the `UserCountGraph` component in #2392, it should now actually be implemented.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `UserCountGraph` component for the Analytics `DashboardAllTrafficWidgetV2` should be fully implemented:
    * It should encompass the full graph UI in the bottom left corner of the widget (see [Figma file](https://www.figma.com/file/NzvJTPFdqUVJC9Q1mmih4e/Google-SK-UX-Enhancements10up?node-id=1198%3A271)), including the filled area below and the curved UI (see e.g. https://developers.google.com/chart/interactive/docs/gallery/linechart#curving-the-lines).
    * It should display data from an Analytics report for the metric `ga:users`, the dimension `ga:date`, dates based on the current date range, and optionally the current entity URL, if there is one.
    * If both the `dimensionName` and `dimensionValue` props are present, it should also consider those for querying the report, as `dimensionFilters`.
    * Background: If both props are present, the graph shows user counts over time for all users that satisfy the filter. Otherwise, the graph shows total user counts over time, not filtered by any dimension.
* The existing Storybook story for  `DashboardAllTrafficWidgetV2` should have its setup updated, to populate the registry with corresponding test report data, so that the UI rendered by `UserCountGraph` is visible there.

## Implementation Brief

* Update the `GoogleChart` component:
  * Update chart instantiation functionality to create the `AreaChart` chart when the `chartType` property equals to `area`.
* Edit the `UserCountGraph` component:
  * Add new `useSelect` hooks which do the following:
    * Pulls `startDate`/`endDate` dates from the `core/user` datastore using the `getDateRangeDates` selector;
    * Pulls `url` from the `core/site` datastore using the `getCurrentEntityURL` selector;
    * Pulls report data from the `module/analytics` datastore using the `getReport` selector with `ga:users` metrics, selected dimension name and value (if it's not empty), `startDate`/`endDate` dates and `url` received from the aforementioned selectors;
  * Implement the required markup and styles to match the UI defined in the provided Figma file.
    * Use the `GoogleChart` component to display the chart. Check `AnalyticsDashboardWidgetSiteStats` as an example to see how we use this component and the AreaChart [documentation](https://developers.google.com/chart/interactive/docs/gallery/areachart#overview) for more details on how to configure it.
[...]

PR #2589, opened 2021-01-06
Feature/2423 total user count

eugene-manuilov:
## Summary

Addresses issue #2423

## Checklist

- [ ] My code is tested and passes existing unit tests.
- [ ] My code has an appropriate set of unit tests which all pass.
- [x] My code is backward-compatible with WordPress 4.7 and PHP 5.6.
- [x] My code follows the [WordPress](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) coding standards.
- [x] My code has proper inline documentation.
- [ ] I have added a QA Brief on the issue linked above.
- [x] I have signed the Contributor License Agreement (see <https://cla.developers.google.com/>).

PR #2615, opened 2021-01-13
Add UserCountGraph component.

tofumatt:
## Summary

Addresses issue #2422

## Relevant technical choices

It wasn't possible to fully recreate the mockups in #2422/Figma because of limitations of the Google Charts API.

Here's the mockup:

Screenshot 2021-01-13 at 19 53 19

Here's the actual component using a `LineChart`, which allows for a smoothly curved line:

Screenshot 2021-01-13 at 20 00 06

**Above is the version that is checked in.**

Here is an alternate with the proposed `AreaChart`:

Screenshot 2021-01-13 at 19 54 53

There are several issues with the above that we aren't able to address with the limitations in the Google Charts API:

* The background colour under the line can't be changed to a lighter blue.
* The line cannot be "smooth", only harsh lines are allowed in AreaCharts.
But we can't use `LineChart` to overcome the above as it doesn't support backgrounds.

---

Additionally, I've changed the dates used from the mockup to add more context (month/day, as the previous one wouldn't work well when spanned across multiple months) but have fewer markings to reduce visual/cognitive noise.

I also added/updated the loading states for the left components; previously no loading animations appeared, now several do:

Screenshot 2021-01-13 at 19 58 30

## Checklist


assets/js/components/GoogleChart.js
```diff
@@ -179,7 +193,7 @@ export default function GoogleChart( props ) {
 
 GoogleChart.propTypes = {
 	chartID: PropTypes.string,
-	chartType: PropTypes.oneOf( [ 'pie', 'line', '' ] ),
+	chartType: PropTypes.oneOf( [ 'area', 'pie', 'line' ] ).isRequired,
```
felixarntz, line 196: Why make this prop required? We could easily keep the default of `line`, this way the PR also doesn't affect too many other areas this late in the release cycle.

assets/js/components/Sparkline.js
```diff
@@ -79,6 +79,7 @@ function Sparkline( {
 	return (
 		<div className="googlesitekit-analytics-sparkline-chart-wrap">
 			<GoogleChart
+				chartType="line"
```
felixarntz, line 82: See above, we could avoid adding that (same applies below).

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/TotalUserCount.js
```diff
@@ -73,8 +74,19 @@ export default function TotalUserCount( { dimensionName, dimensionValue } ) {
...
+	if ( ! loaded ) {
+		return (
+			<PreviewBlock
+				className="googlesitekit-widget--analyticsAllTrafficV2__totalcount--loading"
+				width="220px"
+				height="130px"
+				shape="square"
+			/>
+		);
+	}
```
felixarntz, line 86: I don't think we need to modify this component here. The original idea was to include a single loader / `ReportZero` / `ReportError` because `TotalUserCount` and `UserCountGraph` should be displayed as one - basically `TotalUserCount` always returns `null` for those scenarios while `UserCountGraph` would return the actual component for it. See https://github.com/google/site-kit-wp/pull/2589#discussion_r553093002. Is there a specific reason why you added the `PreviewBlock` here?

assets/js/modules/analytics/util/index.js
```diff
@@ -286,6 +286,39 @@ export const extractAnalyticsDashboardData = ( reports, selectedStats, days ) =>
...
+			const date = parseDimensionStringToDate( dateString );
+
+			return [
+				date,
+				null,
+				values[ 0 ],
+			];
+		} ),
+	];
+};
```
felixarntz, line 320: Let's keep this inline (or in a privately used function) within `UserCountGraph` for now. The name here is very broad for that the function does something fairly specific. We only require it in one place for now, so let's move this as an internal function into `UserCountGraph.js`. We can reconsider reusing/improving/abstracting it when this becomes necessary elsewhere.

Normally we might be able to do that as part of this PR and invest a bit more time, but since it's late, let's go with the above for now.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/UserCountGraph.js
```diff
@@ -21,9 +22,153 @@
...
+import ReportError from '../../../../../components/ReportError';
+import ReportZero from '../../../../../components/ReportZero';
+const { useSelect } = Data;
+
 export default function UserCountGraph( { dimensionName, dimensionValue } ) {
-	return null;
+	const currentEntityURL = useSelect( ( select ) => select( CORE_SITE ).getCurrentEntityURL() );
+	const dateRange = useSelect( ( select ) => select( CORE_USER ).getDateRange() );
+	const dateRangeDates = useSelect( ( select ) => select( CORE_USER ).getDateRangeDates( {
+		compare: true,
```
felixarntz, line 44: We don't need to compare two date ranges here, so this should be removed.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/UserCountGraph.js
```diff
@@ -21,9 +22,153 @@
...
+		offsetDays: DATE_RANGE_OFFSET,
+	} ) );
+
+	const args = {
+		startDate: dateRangeDates.startDate,
+		endDate: dateRangeDates.endDate,
```
felixarntz, line 50: Nit-pick, but you could simplify this then to `...dateRangeDates` (which without the `compare: true` above only include `startDate` and `endDate`).

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/UserCountGraph.js
```diff
@@ -21,9 +22,153 @@
...
+		metrics: [
+			{
+				expression: 'ga:users',
+				alias: 'Users Over Time',
+			},
+		],
+		orderby: {
+			fieldName: 'ga:date',
+			sortOrder: 'DESCENDING',
+		},
```
felixarntz, line 61: Why sort by date in reverse order here? That doesn't make sense to me. The results in API responses are normally sorted correctly: If `ga:date` is the sole dimension, they should be by date ascending.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/UserCountGraph.js
```diff
@@ -21,9 +22,153 @@
...
+	};
+
+	if ( currentEntityURL ) {
+		args.url = currentEntityURL;
+	}
+
+	if ( dimensionName && dimensionValue ) {
+		args.dimensions.push( dimensionName );
```
felixarntz, line 69: I don't think we need to include an extra dimension here. We only filter the entire query by a single dimension value anyway, so grouping by the same dimension is kind of pointless. This actually also applies to `TotalUserCount`, where this is incorrectly present. We only need `dimensionFilters` here.

felixarntz (changes requested):
This looks solid for the most part! I left a bit of feedback.

felixarntz:
@tofumatt I've pushed two commits to address the request simplifications I mentioned above, also some similar ones from `TotalUserCount` that only now became apparent to me. I've also moved the fixtures for Storybook into their own files.

assets/js/components/GoogleChart.js
tofumatt: Mostly that I found the default of "line" quite confusing and it'd add to the switch-case… we could do this later but I tested every spot this is in and it didn't fail any VRTs either, so is safe to make I think.

The default of `line` didn't seem very intuitive, especially with multiple options for this prop now. We can revert this change, but I think the original prop setup wasn't very intentional.

If you'd still prefer we can revert this though, let me know 😄

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/TotalUserCount.js
tofumatt: These reports are slightly different: one requests the total user count and the other requests users for each day. This means they resolve at different times, and putting the loading state for another component inside this one seems odd to me. If they should be anywhere shared it should be the parent component I think.

But in my usage these components can exit "loading" state at different times, so they shouldn't share a loading state. This results in one disappearing and the layout "jumping" around while one is loaded and another isn't.

I suppose we could consolidate the API requests but the total user API request does not include the same data as the new one for the daily users report. Again: combining the numbers from the daily report could give us the user count and reduce API requests which would be good, but I wasn't sure if we wanted to do that now or later 🙂 

Thoughts?

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/UserCountGraph.js
tofumatt: Ah, true, whoops! 👍🏻

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/UserCountGraph.js
tofumatt: I like the more explicit nature of specifying the args, so you don't need to trace back to the `getDateRangeDates` function to know what's being returned.

But I'll use `startDate` and `endDate` instead here, to simplify it a bit. 😄

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/UserCountGraph.js
tofumatt: Argh, that was testing something with the charts and it stayed in, my bad! 😓

assets/js/components/GoogleChart.js
felixarntz: Fair point, I was mostly wanting to avoid unrelated changes this late in the cycle. But this one should be straightforward enough.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/TotalUserCount.js
felixarntz: I originally thought we would use the same API request for both, but it's okay not to do that for now. We definitely _can_ do it:
* Request with compare dates (needed for `TotalUserCount`), then in `UserCountGraph` only consider the second half of rows (the "main" date range).
* Request with `ga:date` dimension (needed for `UserCountGraph`), which is irrelevant for `TotalUserCount` since it only looks at the totals anyway.

Maybe worth a follow-up issue.

felixarntz (approved):
Awesome work! I'll merge this, we can open follow-up issues if the review results in UI changes being required.

Issue #2429, opened 2020-11-24
Expose new Analytics All Traffic widget in production


felixarntz:
Once ready the new `DashboardAllTrafficWidgetV2` widget should replace the `DashboardAllTrafficWidget` widget and be exposed in production.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `DashboardAllTrafficWidget` component and any sub-components that are only used by that one should be removed. All references to that widget should be removed too, including tests, Storybook etc.
* The `DashboardAllTrafficWidgetV2` should be renamed to `DashboardAllTrafficWidget`, all references should be updated accordingly. Its widget registration should be updated to drop the v2 indicator from the widget slug as well.
* The heading of both `*AllTraffic` widget areas should be updated to: _Your Traffic at a Glance_
* Pending on whether the widget API is not live yet, we'll need to also replace the legacy All Traffic widget with the new component (outside of the widget API) and updated the area heading and description.

## Implementation Brief

**Widgets**

* Remove `DashboardAllTrafficWidget` component
* Remove `AcquisitionSources` component
* Remove `AcquisitionPieChart` component
* Rename `DashboardAllTrafficWidgetV2` to `DashboardAllTrafficWidget`, change `<Widget>`'s `slug` to `analyticsAllTraffic`  and update all the references accordingly.
* Remove `analyticsAllTrafficV2` widget registration in `assets/js/modules/analytics/index.js`. Update `analyticsAllTraffic` registration to use the new `DashboardAllTrafficWidget`

**Stories**

* Remove `Analytics Module/Components/Page Dashboard/All Traffic Widget`
* Rename `Analytics Module/Components/Dashboard/All Traffic Widget V2` to `Analytics Module/Components/Dashboard/All Traffic Widget` and update the `Component` to use the new `DashboardAllTrafficWidget`
* Update `Analytics Module/Components/Page Dashboard/All Traffic Widget` to use the newer v2 of `DashboardAllTrafficWidget`

**Language Strings**

* Update `title` for `AREA_DASHBOARD_ALL_TRAFFIC` and `AREA_PAGE_DASHBOARD_ALL_TRAFFIC` to `Your Traffic at a Glance` in `assets/js/googlesitekit/widgets/register-defaults.js`

For legacy support:
* Add a `LegacyAllTraffic` component in `assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget`.
[...]

PR #2617, opened 2021-01-13
Expose new Analytics All Traffic widget in production.

benbowler:
## Summary

Addresses issue #2429

## Relevant technical choices

The IB wasn't too clear with the Stories updates, instead of "Remove[ing] Analytics Module/Components/Page Dashboard/All Traffic Widget" I updated it to render with the new component and updated the test data also, inlining it as Felix has in other widget stories. Otherwise, I followed the IB end to end.

## Checklist


benbowler:
While still blocked by #2422 this is ready for initial code review. Once #2422 is merged this can be passed for QA/MR.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
```diff
@@ -204,7 +204,7 @@ export default function UserDimensionsPieChart( { dimensionName, entityURL, sour
 	}
```
felixarntz, line 205: I noticed a problem while testing, related to this component (see #2425): When I run a development build, the piechart does not show up, in neither the new nor the old widget. I assume that is happening because the chart uses a specific ID but is now used in multiple places?

@eugene-manuilov Do you have an idea how we could fix this? I wonder whether we could somehow increase a static count to prevent duplicate IDs.

Alternatively, if that doesn't work, we could also go with a more pragmatic approach specific to the situation, e.g. `DashboardAllTrafficWidget` could have an optional `isLegacy` prop and based on that the ID used for the inner component's chart would be different - not a great approach, but this situation is temporary anyway, so might be okay. If we can solve it in `UserDimensionsPieChart` internally though, that'd be perfect.

felixarntz (changes requested):
Awesome work! There's only one thing that needs to be addressed here, and two minor comments. It also looks like this broke the e2e tests, so it would be great if you could check those. Potentially it's just that we were relying on some DOM element from the old All Traffic widget somewhere.

assets/js/modules/analytics/index.legacy.js
```diff
@@ -85,7 +85,7 @@ if ( modulesData.analytics.active ) {
 	 */
 	addFilter( 'googlesitekit.DashboardModule',
 		'googlesitekit.Analytics',
-		addAnalyticsAllTraffic, 9 );
+		legacyDashboardAllTraffic, 8 );
```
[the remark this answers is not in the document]
benbowler: Yes, I'm not sure, that was right from the IB. I've reverted it.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/LegacyDashboardAllTraffic.js
```diff
@@ -0,0 +1,56 @@
+/**
+ * DashboardSpeed component.
```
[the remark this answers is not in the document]
benbowler: Fixed.

benbowler:
I've fixed the small comments and I'll take a look at the E2E tests now.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
eugene-manuilov: > Do you have an idea how we could fix this? I wonder whether we could somehow increase a static count to prevent duplicate IDs.

@felixarntz I think we can leverage the [`useInstanceId`](https://github.com/WordPress/gutenberg/tree/master/packages/compose#useInstanceId) hook to get a unique id for the `UserDimensionsPieChart` component and then add it as a suffix to the `chartID` prop. What do you think?

felixarntz: @eugene-manuilov SGTM! cc @benbowler

benbowler: @felixarntz this has been [implemented here](https://github.com/google/site-kit-wp/pull/2617/commits/3add375d269f207bea6b3da62d9933e5a23b1923).

felixarntz:
@benbowler Merged `develop` back in which now includes the `UserCountGraph` implemented.

felixarntz (changes requested):
@benbowler @eugene-manuilov Not sure what's going on there. I saw you added `useInstanceId`, but when I run the latest code in a development build, I see only one of the two charts showing up at a time.

They still appear to be connected in some way they shouldn't be. For example, when I switched tabs in the new widget, it scrolled down to the legacy widget and showed the graph there.

aaemnnosttv:
@googlebot I consent.

aaemnnosttv:
@felixarntz 

> when I switched tabs in the new widget, it scrolled down to the legacy widget and showed the graph there.

We had the same problem with the PageSpeed widget; it has to do with the default behavior of the `TabBar`. We just need to configure the `Tab` components to not `focusOnActivate`. I've pushed a commit that fixes this.

> when I run the latest code in a development build, I see only one of the two charts showing up at a time.

This seems to be limited to the initial render/draw. When I click a dimension tab, it appears. Taking a closer look at this now.

benbowler:
Looping back here, this is the only comment outstanding to look at I believe:

> @felixarntz  when I run the latest code in a development build, I see only one of the two charts showing up at a time.

> @aaemnnosttv  This seems to be limited to the initial render/draw. When I click a dimension tab, it appears. Taking a closer look at this now.

aaemnnosttv:
The problem I was seeing before is that the `onReady` callback for the chart wasn't being called for both. Not sure if it consistently works when there's only one either, but I can look into this a bit more.

eugene-manuilov:
@benbowler i think the issue is caused by the `fontName` options. Try to comment out the following lines, those charts should start working correctly:

https://github.com/google/site-kit-wp/blob/df38a87cef996a23f35434400d5eac632b604df8/assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/UserDimensionsPieChart.js#L245

https://github.com/google/site-kit-wp/blob/df38a87cef996a23f35434400d5eac632b604df8/assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/UserDimensionsPieChart.js#L259

benbowler:
@eugene-manuilov's fix has worked for me when testing. I've committed it, removed the now unused `useInstanceID` hook and pushed ready for review.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
```diff
@@ -207,17 +207,16 @@ export default function UserDimensionsPieChart( { dimensionName, entityURL, sour
...
 
 	return (
-		<div className="googlesitekit-widget--analyticsAllTrafficV2__dimensions-chart">
+		<div className="googlesitekit-widget--analyticsAllTraffic__dimensions-chart">
 			<GoogleChart
-				chartID="user-dimensions-pie-chart"
```
eugene-manuilov, line 212: @benbowler the `chartID` is important for this chart and it should be unique. Please, use the `useInstanceId` hook again to make the `chartID` property unique on per instance basis. Plus, make sure that `GoogleChart.charts.get( 'user-dimensions-pie-chart' )` calls in the `onReady` callback use the same chart id as we pass to the `GoogleChart` component.

eugene-manuilov (changes requested):

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
benbowler: Resolved.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/LegacyDashboardAllTraffic.js
```diff
@@ -0,0 +1,56 @@
+/**
+ * Legacy Dashboard All Traffic component.
+ *
+ * Site Kit by Google, Copyright 2019 Google LLC
```
eugene-manuilov, line 4:
```suggestion
 * Site Kit by Google, Copyright 2021 Google LLC
```

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/LegacyDashboardAllTraffic.js
```diff
@@ -0,0 +1,56 @@
...
+import DashboardModuleHeader from '../../../../../components/dashboard/DashboardModuleHeader';
+import DashboardAllTrafficWidget from '.';
+
+function LegacyDashboardAllTraffic() {
+	return (
+		<Fragment>
+			<div id="googlesitekit-pagespeed-header" className="
+					mdc-layout-grid__cell
+					mdc-layout-grid__cell--span-12
+				">
```
eugene-manuilov, line 38:
```suggestion
			<Cell id="googlesitekit-pagespeed-header" size={ 12 }>
```

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/LegacyDashboardAllTraffic.js
```diff
@@ -0,0 +1,56 @@
...
+				<DashboardModuleHeader
+					title={ __( 'Your Traffic at a Glance', 'google-site-kit' ) }
+					description={ __( 'How people found your site.', 'google-site-kit' ) }
+				/>
+			</div>
+			<div className="
+				mdc-layout-grid__cell
+				mdc-layout-grid__cell--span-12
+			">
```
eugene-manuilov, line 47:
```suggestion
			<Cell size={ 12 }>
```

eugene-manuilov (changes requested):
Thanks, @benbowler! Found a few minor things to change and I think it should be good to go. Could you please take a look at it?

benbowler:
Updated and back to you @eugene-manuilov

eugene-manuilov (dismissed):
Awesome!

eugene-manuilov (approved):
Looks good now!

felixarntz:
@benbowler @eugene-manuilov FYI I pushed one extra commit here just for some wording changes that came up during review with @marrrmarrr.

felixarntz (approved):
Awesome work - let's 🚢 !

Issue #2627, opened 2021-01-15
Settings not preserved after granting additional scope


aaemnnosttv:
## Bug Description

When performing an action that requires more than the default read-only scopes, we request the additional scope needed on-the-fly; e.g. creating a new property or profile in Analytics setup/settings. While the underlying functionality (store snapshot/restore) is working as expected, a recent change is causing the settings rollback to be triggered after the state is restored which essentially "forgets" the previous settings state which leaves the form in a different state since the "create" options are no longer selected.

## Steps to reproduce

1. Set up Analytics
2. Edit Analytics settings to create a new property (without previously granting scope)
3. Submit (pop-up should prompt to grant additional scope)
4. Grant scope
5. Upon returning to settings UI, notice that property select is no longer on the "create" option

## Additional Context

This is likely due to changes/refactoring to `SettingsApp` where the initial render may not have `isEditing` as `true` which would cause the settings rollback to be triggered since settings would have changed. See #2187.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* 

## Implementation Brief

* 

### Test Coverage

* 

### Visual Regression Changes

* 

## QA Brief

* 

## Changelog entry

*

felixarntz:
Closing this in favor of tracking via #2187 where we'll need to fix this for the same release.

Issue #2187, opened 2020-10-15
Refactor `SettingsApp` component


felixarntz:
The `SettingsApp` component should be refactored to become a functional component.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `SettingsApp` component should be refactored to become a functional component using the datastore.
    * The component should rely on internal `useState` to manage the current tab.
    * The component should rely on a `useEffect` to potentially set the current tab (via above internal state) and currently active module settings panel (via `setModuleSettingsPanelState` action in `core/modules` store) based on the current URL hash.
* Its file should be renamed to `SettingsApp.js`, and all import references should be updated accordingly.

## Implementation Brief

While we're refactoring this component, we should stop our direct interaction with `location.hash`. There are better, more React-y libraries available to handle this for us. There are two options here for a refactor:

Use the `useHash` hook from [react-use](https://github.com/streamich/react-use). This will allow us to treat that particular hash param as a piece of state that is also stored in `location.hash`. This will lead to much easier-to-follow code, so it's my recommendation and the one I'll reference in this IB. 🙂 

* Rename `assets/js/components/settings/settings-app.js` to `assets/js/components/settings/SettingsApp.js`, updating all imports accordingly (VSCode will do this automatically with the right settings enabled).
* Add `react-use` to the project: `npm install --save react-use`.
* Use [`useHash()`](https://github.com/streamich/react-use/blob/master/docs/useHash.md) to get and store the `activeTabID`, `moduleSlug`, and `moduleState`:

```js
const [ hash, setHash ] = useHash();

// Route, e.g. #settings/analytics/(view|edit)
let [ activeTabID = 'settings', moduleSlug, moduleState ] = hash.replace( '#', '' ).split( /\// );
```
* Use `select( 'core/modules' ).getModuleSettingsPanelState( moduleSlug )` to get the `moduleSlug`'s panel state.
* Use `useDispatch( 'core/modules' ).setModuleSettingsPanelState( moduleSlug, viewState )` to set a panel's state.
* See: https://gist.github.com/tofumatt/cd40b93462e5356c277d9f3c06d1409f for a rough idea of how this would work once #2181 is complete. That is a proof-of-concept that largely works and doesn't rely on internal component state for values—instead using the URL as the source of truth to make things easier.

### Test Coverage

* Add tests to ensure panel click behaviour is respected; activating/deactivating tabs and panels should work as-expected.
[...]

PR #2631, opened 2021-01-15
Fix settings app

aaemnnosttv:
## Summary

Addresses issue #2187 (follow-up), #2627

Targets `master` branch

## Relevant technical choices

* Fixes unwanted side-effects (settings rollback) due to settings panel state not being initialized on initial render

## Checklist

### DIFF
--- a/assets/js/components/settings/SettingsApp.js
+++ b/assets/js/components/settings/SettingsApp.js
@@ -21,12 +21,12 @@
  */
 import Tab from '@material/react-tab';
 import TabBar from '@material/react-tab-bar';
-import { useHash } from 'react-use';
+import { useFirstMountState, useHash } from 'react-use';
 
 /**
  * WordPress dependencies
  */
-import { Fragment, useEffect } from '@wordpress/element';
+import { Fragment, useState } from '@wordpress/element';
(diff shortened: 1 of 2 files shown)


assets/js/components/settings/SettingsApp.js
```diff
@@ -44,62 +44,57 @@ import { Cell, Grid, Row } from '../../material-components';
...
-		const fragments = [ tabID ];
-
-		if ( tabID === 'settings' && slug && state !== 'closed' ) {
-			fragments.push( slug );
-			if ( state ) {
-				fragments.push( state );
-			}
+	if ( isFirstMount ) {
+		if ( initialModuleSlug ) {
+			setModuleSettingsPanelState( initialModuleSlug, initialModuleState );
```
aaemnnosttv, line 73: This is the important change here. `useEffect` runs too late and `moduleState` would otherwise be `closed` on the initial render, even when properly parsed as `view` or `edit` from the hash before this. This will essentially run synchronously but only on the very first render. Since `SettingsApp` is never unmounted, this only happens once. It's all somewhat temporary though as most of this will change again once we add the router.

felixarntz (approved):
LGTM!

Issue #2286, opened 2020-10-29
Pass specific dates in Search Console deep links


felixarntz:
After #2237, all dynamic JS Search Console deep links in the new widgets (any JS links pointing to `performance/search-analytics...`) should be updated to pass the correct dates in the URLs.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All widgets (using the Widget API) that include Search Console deep links should be updated to pass the exact dates from the currently selected date range in the URL:
    * The `num_of_days` query parameter should no longer be used.
    * Instead, it should be query parameters `start_date` and `end_date`, both in the format `YYYYMMDD` (e.g. `start_date=20201001&end_date=20201028`).

## Implementation Brief
* Create new file `/assets/js/modules/search-console/util/report-date-range-args` and using newly created file,
	* Create a new function `generateDateRangeArgs` which accepts the returned object from the getDateRangeDates selector. The object contains the following keys:
		* `startDate`
		* `endDate`
		* `compareStartDate`: if compare is true when calling `getDateRangeDates`.
		* `compareEndDate`: if compare is true when calling `getDateRangeDates`.
	* The `generateDateRangeArgs` should return an object with the following keys:
		* `'start_date': startDate.replace( /-/g, '' ),`
		* `'end_date': endDate.replace( /-/g, '' ),`
	* `start_date` and `end_date` must be in the format: `YYYMMDD`
* Using the files listed below:
	- Import the `generateDateRangeArgs` function from `/assets/js/modules/search-console/util/report-date-range-args` and call the function by passing in the result of the `getDateRangeDates` selector of the user store (if not already present in the component).For e.g
		```js
		const { startDate, endDate } = select( CORE_USER ).getDateRangeDates( { offsetDays: DATE_RANGE_OFFSET } );
		```
	- Replace occurrences of `num_of_days` parameter by spreading the returned object of the `generateDateRangeArgs` function.
* List of files:
	- `assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js`
	- `assets/js/modules/search-console/components/dashboard/LegacySearchConsoleDashboardWidgetKeywordTable.js`
	- `assets/js/modules/search-console/components/dashboard/LegacyDashboardWidgetPopularKeywordsTable.js`
	- `assets/js/modules/search-console/components/dashboard/DashboardClicksWidget.js`
[...]

Issue #2563, opened 2020-12-23
Refactor store aliases with named exports


aaemnnosttv:
## Feature Description

We currently export a `STORE_NAME` constant from the `constants.js` module for each module and core data store. When referenced within the context of itself (the same store) we use it with the same name. However, when this constant is imported for use in another store as is often needed, we alias it to be named in a way.

E.g. the `export const STORE_NAME = 'core/site'` would be imported as `import { STORE_NAME as CORE_SITE }`

https://github.com/google/site-kit-wp/blob/8dc8de99f9225f62972c2bb82f67f6e3d39651ad/assets/js/modules/optimize/components/common/UseSnippetInstructions.js#L30-L34

We have _many_ instances of this throughout the codebase, but there are a few common challenges that we face when importing the store name in this way:

* Inconsistency: new engineers often get tripped up by this convention which is somewhat arbitrary as an alias
* Inefficiency: importing `STORE_NAME` is a bit cumbersome since we have so many and especially when the constant is already imported, we lose the ability to leverage auto-import functionality (e.g. using VS Code)

These can both be solved by providing the "external" store name as a named export rather than using an alias.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All registered data store `STORE_NAME` constants should be available to import as named exports
* All `STORE_NAME` **aliases** should be replaced with imports of the named export
    * All un-aliased uses should remain unchanged

## Implementation Brief
* Review occurrences where `STORE_NAME` is aliased and add a new export name where the aliased `STORE_NAME` is defined. For e.g
	```js
	export const STORE_NAME = 'core/site';
	export { STORE_NAME as CORE_SITE };
	```
* Review occurrences where `STORE_NAME` is aliased throughout the app and update them to use the named export instead.
* All un-aliased uses should remain unchanged

### Test Coverage
* To tests needed or to be updated.

### Visual Regression Changes
* There should be no visual changes.

## QA Brief
* Search through the codebase for instances of `STORE_NAME` imported with an alias, e.g. `import { STORE_NAME as `
* There should be none remaining
* Instances where `STORE_NAME` is imported, but *not* as an alias are fine, e.g. `import { STORE_NAME, ... } from`

## Changelog entry

* N/A

aaemnnosttv:
@asvinb 

> For e.g
> ```js
> export const STORE_NAME = 'core/site';
> ```
> becomes
> ```js
> export const CORE_SITE = 'core/site';
[...]

PR #2634, opened 2021-01-18
Pass specific dates in Search Console deep links

johnPhillips:
## Summary

Addresses issue #2286

## Checklist

### DIFF
--- a/assets/js/components/PostSearcher.js
+++ b/assets/js/components/PostSearcher.js
@@ -42,7 +42,7 @@ const { useSelect, useDispatch } = Data;
 function PostSearcher() {
 	const [ canSubmit, setCanSubmit ] = useState( false );
 	const [ match, setMatch ] = useState( {} );
-	const analyticsModuleActive = useSelect( ( select ) => select( CORE_MODULES ).isModuleActive( 'analytics' ) );
+	const analyticsModuleConnected = useSelect( ( select ) => select( CORE_MODULES ).isModuleConnected( 'analytics' ) );
 
 	const detailsURL = useSelect( ( select ) => {
 		return select( CORE_SITE ).getAdminURL( 'googlesitekit-dashboard', {
@@ -56,9 +56,9 @@ function PostSearcher() {
 	return (
 		<div
 			className={ classnames( 'mdc-layout-grid__cell', {
-				'mdc-layout-grid__cell--span-12': analyticsModuleActive,
-				'mdc-layout-grid__cell--span-4-tablet': ! analyticsModuleActive,
-				'mdc-layout-grid__cell--span-6-desktop': ! analyticsModuleActive,
+				'mdc-layout-grid__cell--span-12': analyticsModuleConnected,
+				'mdc-layout-grid__cell--span-4-tablet': ! analyticsModuleConnected,
+				'mdc-layout-grid__cell--span-6-desktop': ! analyticsModuleConnected,
 			} ) }
 		>
 			<Layout
--- a/assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
+++ b/assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
@@ -29,19 +29,21 @@ import Data from 'googlesitekit-data';
 import LegacySearchConsoleDashboardWidgetKeywordTable from '../dashboard/LegacySearchConsoleDashboardWidgetKeywordTable';
 import DashboardModuleHeader from '../../../../components/dashboard/DashboardModuleHeader';
 import Layout from '../../../../components/layout/Layout';
-import { STORE_NAME } from '../../datastore/constants';
+import { STORE_NAME, DATE_RANGE_OFFSET } from '../../datastore/constants';
 import { CORE_SITE } from '../../../../googlesitekit/datastore/site/constants';
-import { getCurrentDateRangeDayCount } from '../../../../util/date-range';
+import { CORE_USER } from '../../../../googlesitekit/datastore/user/constants';
 import { untrailingslashit } from '../../../../util';
+import { generateDateRangeArgs } from '../../util/report-date-range-args';
 
 const { useSelect } = Data;
 
 const DashboardDetailsWidgetKeywordsTable = () => {
 	const propertyID = useSelect( ( select ) => select( STORE_NAME ).getPropertyID() );
 	const url = useSelect( ( select ) => select( CORE_SITE ).getCurrentEntityURL() );
+	const { startDate, endDate } = useSelect( ( select ) => select( CORE_USER ).getDateRangeDates( { offsetDays: DATE_RANGE_OFFSET } ) );
 	const footerCTALinkArgs = {
 		resource_id: propertyID,
-		num_of_days: getCurrentDateRangeDayCount(),
+		...generateDateRangeArgs( { startDate, endDate } ),
 	};
 	const isDomainProperty = useSelect( ( select ) => select( STORE_NAME ).isDomainProperty() );
 	const referenceSiteURL = useSelect( ( select ) => {
--- a/assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
+++ b/assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
@@ -29,7 +29,7 @@ import Data from 'googlesitekit-data';
 import LegacySearchConsoleDashboardWidgetKeywordTable from '../dashboard/LegacySearchConsoleDashboardWidgetKeywordTable';
 import DashboardModuleHeader from '../../../../components/dashboard/DashboardModuleHeader';
 import Layout from '../../../../components/layout/Layout';
-import { STORE_NAME, DATE_RANGE_OFFSET } from '../../datastore/constants';
+import { MODULES_SEARCH_CONSOLE, DATE_RANGE_OFFSET } from '../../datastore/constants';
 import { CORE_SITE } from '../../../../googlesitekit/datastore/site/constants';
 import { CORE_USER } from '../../../../googlesitekit/datastore/user/constants';
 import { untrailingslashit } from '../../../../util';
@@ -38,14 +38,14 @@ import { generateDateRangeArgs } from '../../util/report-date-range-args';
 const { useSelect } = Data;
 
 const DashboardDetailsWidgetKeywordsTable = () => {
-	const propertyID = useSelect( ( select ) => select( STORE_NAME ).getPropertyID() );
+	const propertyID = useSelect( ( select ) => select( MODULES_SEARCH_CONSOLE ).getPropertyID() );
 	const url = useSelect( ( select ) => select( CORE_SITE ).getCurrentEntityURL() );
 	const { startDate, endDate } = useSelect( ( select ) => select( CORE_USER ).getDateRangeDates( { offsetDays: DATE_RANGE_OFFSET } ) );
 	const footerCTALinkArgs = {
 		resource_id: propertyID,
 		...generateDateRangeArgs( { startDate, endDate } ),
 	};
-	const isDomainProperty = useSelect( ( select ) => select( STORE_NAME ).isDomainProperty() );
+	const isDomainProperty = useSelect( ( select ) => select( MODULES_SEARCH_CONSOLE ).isDomainProperty() );
 	const referenceSiteURL = useSelect( ( select ) => {
 		return untrailingslashit( select( CORE_SITE ).getReferenceSiteURL() );
 	} );
@@ -54,7 +54,7 @@ const DashboardDetailsWidgetKeywordsTable = () => {
 	} else if ( isDomainProperty && referenceSiteURL ) {
 		footerCTALinkArgs.page = `*${ referenceSiteURL }`;
 	}
-	const footerCTALink = useSelect( ( select ) => select( STORE_NAME ).getServiceURL( {
+	const footerCTALink = useSelect( ( select ) => select( MODULES_SEARCH_CONSOLE ).getServiceURL( {
 		path: '/performance/search-analytics',
 		query: footerCTALinkArgs,
 	} ) );
--- a/assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
+++ b/assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
@@ -29,7 +29,7 @@ import Data from 'googlesitekit-data';
 import LegacySearchConsoleDashboardWidgetKeywordTable from '../dashboard/LegacySearchConsoleDashboardWidgetKeywordTable';
 import DashboardModuleHeader from '../../../../components/dashboard/DashboardModuleHeader';
 import Layout from '../../../../components/layout/Layout';
-import { MODULES_SEARCH_CONSOLE, DATE_RANGE_OFFSET } from '../../datastore/constants';
+import { STORE_NAME, DATE_RANGE_OFFSET } from '../../datastore/constants';
 import { CORE_SITE } from '../../../../googlesitekit/datastore/site/constants';
 import { CORE_USER } from '../../../../googlesitekit/datastore/user/constants';
 import { untrailingslashit } from '../../../../util';
@@ -38,14 +38,14 @@ import { generateDateRangeArgs } from '../../util/report-date-range-args';
 const { useSelect } = Data;
 
 const DashboardDetailsWidgetKeywordsTable = () => {
-	const propertyID = useSelect( ( select ) => select( MODULES_SEARCH_CONSOLE ).getPropertyID() );
+	const propertyID = useSelect( ( select ) => select( STORE_NAME ).getPropertyID() );
 	const url = useSelect( ( select ) => select( CORE_SITE ).getCurrentEntityURL() );
 	const { startDate, endDate } = useSelect( ( select ) => select( CORE_USER ).getDateRangeDates( { offsetDays: DATE_RANGE_OFFSET } ) );
 	const footerCTALinkArgs = {
 		resource_id: propertyID,
 		...generateDateRangeArgs( { startDate, endDate } ),
 	};
-	const isDomainProperty = useSelect( ( select ) => select( MODULES_SEARCH_CONSOLE ).isDomainProperty() );
+	const isDomainProperty = useSelect( ( select ) => select( STORE_NAME ).isDomainProperty() );
 	const referenceSiteURL = useSelect( ( select ) => {
 		return untrailingslashit( select( CORE_SITE ).getReferenceSiteURL() );
 	} );
@@ -54,7 +54,7 @@ const DashboardDetailsWidgetKeywordsTable = () => {
 	} else if ( isDomainProperty && referenceSiteURL ) {
 		footerCTALinkArgs.page = `*${ referenceSiteURL }`;
 	}
-	const footerCTALink = useSelect( ( select ) => select( MODULES_SEARCH_CONSOLE ).getServiceURL( {
+	const footerCTALink = useSelect( ( select ) => select( STORE_NAME ).getServiceURL( {
 		path: '/performance/search-analytics',
 		query: footerCTALinkArgs,
 	} ) );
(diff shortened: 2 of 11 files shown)


assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
```diff
@@ -29,19 +29,21 @@ import Data from 'googlesitekit-data';
 import LegacySearchConsoleDashboardWidgetKeywordTable from '../dashboard/LegacySearchConsoleDashboardWidgetKeywordTable';
 import DashboardModuleHeader from '../../../../components/dashboard/DashboardModuleHeader';
 import Layout from '../../../../components/layout/Layout';
-import { STORE_NAME } from '../../datastore/constants';
+import { STORE_NAME, DATE_RANGE_OFFSET } from '../../datastore/constants';
 import { STORE_NAME as CORE_SITE } from '../../../../googlesitekit/datastore/site/constants';
-import { getCurrentDateRangeDayCount } from '../../../../util/date-range';
+import { STORE_NAME as CORE_USER } from '../../../../googlesitekit/datastore/user/constants';
```
tofumatt, line 34: Now that #2563 is merged these imports can be cleaned up and are the likely source of merge conflicts in the PR—this can now be imported directly as `CORE_USER` instead of `STORE_NAME as CORE_USER`.

assets/js/modules/search-console/util/report-date-range-args.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * Report date range args utils.
+ *
+ * Site Kit by Google, Copyright 2020 Google LLC
```
tofumatt, line 4: 2020 felt like it would never end, but it did 😉 

```suggestion
 * Site Kit by Google, Copyright 2021 Google LLC
```

assets/js/modules/search-console/util/report-date-range-args.js
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Generates an object with the appropriate structure to pass to report selectors.
```
tofumatt, line 20:
```suggestion
 * Generates a date range object with the appropriate structure to pass to report selectors.
```

Let's be specific about the kind of objects and/or reports we're talking about—this docstring will lack a bit of context without that.

assets/js/modules/search-console/util/report-date-range-args.js
```diff
@@ -0,0 +1,49 @@
...
+ */
+export const generateDateRangeArgs = ( dates ) => {
+	if ( ! dates ) {
+		return undefined;
+	}
+
+	const { startDate, endDate } = dates;
+
+	if ( ! startDate || ! endDate ) {
+		return dates;
```
tofumatt, line 39: Why return this if either of the expected arguments are false-y? I'd expect this either to error or to return undefined, but not to return the original arguments.

assets/js/modules/search-console/util/report-date-range-args.test.js
```diff
@@ -0,0 +1,57 @@
+/**
+ * Tests for reporting date range utilities.
+ *
+ * Site Kit by Google, Copyright 2020 Google LLC
```
tofumatt, line 4:
```suggestion
 * Site Kit by Google, Copyright 2021 Google LLC
```

assets/js/modules/search-console/util/report-date-range-args.js
```diff
@@ -0,0 +1,49 @@
...
+ * Generates an object with the appropriate structure to pass to report selectors.
+ *
+ * @since n.e.x.t
+ *
+ * @param {Object} dates                    Object containing the dates.
+ * @param {string} dates.startDate          Start date in YYYY-MM-DD format.
+ * @param {string} dates.endDate            End date in YYYY-MM-DD format.
+ * @param {string} [dates.compareStartDate] Optional. Start date in YYYY-MM-DD format.
+ * @param {string} [dates.compareEndDate]   Optional. End date in YYYY-MM-DD format.
+ * @return {Object} Date range object.
```
tofumatt, line 29: This can also return `undefined`, so this should be updated to be:

```suggestion
 * @return {Object|undefined} Date range object.
```

assets/js/modules/search-console/util/report-date-range-args.test.js
```diff
@@ -0,0 +1,57 @@
...
+
+describe( 'Search Console reporting date range arguments', () => {
+	describe( 'generateDateRangeArgs', () => {
+		const allArgs = {
+			startDate: '2020-12-18',
+			endDate: '2021-01-14',
+		};
+
+		describe( 'if `startDate` or `endDate` arguments are not provided', () => {
+			it( 'should return args untouched', () => {
```
tofumatt, line 28: This is nowhere in the ACs/IBs and seems unexpected to me. Let's have it return `undefined` if the args aren't present. Or use `invariant` to warn against improper usage—either works for me.

Since it's a new function that isn't used anywhere else it might be nicer to use `invariant` when the args aren't a valid format. 🤔

tofumatt (changes requested):
Looks good for the most part, but there are a few merge conflicts, likely related to the `STORE_NAME` change in #2563.

Also: let's not return the arguments as mentioned when they don't conform to what we expect—instead we should either error or return `undefined`.

assets/js/modules/search-console/util/report-date-range-args.test.js
johnPhillips: Bit of a misunderstanding I think: I picked this up from https://github.com/google/site-kit-wp/commit/58101f479d9cec82fe81b86a4234edc3161e3a6d , and understood / thought that the function had been written to spec. When I came to write the tests I included this behaviour. I did think it was unusual, but I'll just ask next time 🙂

assets/js/modules/search-console/util/report-date-range-args.js
johnPhillips: It was a handover mixup, see my comment below 🙈

tofumatt:
@googlebot I consent.

tofumatt (commented):
Just caught a few `STORE_NAME` references that needed updating, but otherwise the code looks good. I updated them to reduce the back-and-forth on the issue 🙂 

I saw E2E here failing but I'm not sure if that's related to the changes as I've seen them fail unexpected elsewhere on Actions too today. I'll wait until they pass to move to merge review, but it looks good to me.

johnPhillips:
Thanks @tofumatt - I've renamed the imports as you suggest 👍

johnPhillips:
E2E tests are passing locally...🤔

tofumatt:
Odd, I see this test failure locally:

Screenshot 2021-01-21 at 14 53 59

tofumatt (commented):
At this point I'm thinking merging this might be blocked by #2674, because E2E tests that are unrelated and should be passing are failing. I'll leave this in Code Review now but mark the issue as blocked by #2674.

tofumatt:
@googlebot I consent.

aaemnnosttv:
@johnPhillips the main problem here is that your linting is failing which is causing many other jobs to fail because it errors during the build.

johnPhillips:
I rebased and fixed the linting issue (no idea how that crept in there, as I do check before I push). Looks like there is still an E2E test failing here. I'll try to get to this today and investigate further 👍

aaemnnosttv:
> Looks like there is still an E2E test failing here

That doesn't look related to your changes here and seems to be one of the remaining instabilities. I've triggered it to run again after which it will _probably_ pass 😄  Usually legitimate E2E failures cause them all to fail, otherwise it's almost always a stability thing.

tofumatt (approved):

aaemnnosttv:
@googlebot I consent.

assets/js/modules/search-console/util/report-date-range-args.js
```diff
@@ -0,0 +1,49 @@
...
+ * Generates a date range object with the appropriate structure to pass to report selectors.
+ *
+ * @since n.e.x.t
+ *
+ * @param {Object} dates                    Object containing the dates.
+ * @param {string} dates.startDate          Start date in YYYY-MM-DD format.
+ * @param {string} dates.endDate            End date in YYYY-MM-DD format.
+ * @param {string} [dates.compareStartDate] Optional. Start date in YYYY-MM-DD format.
+ * @param {string} [dates.compareEndDate]   Optional. End date in YYYY-MM-DD format.
+ * @return {Object|undefined} Date range object.
```
aaemnnosttv, line 29: As a function which expects the return value from the `getDateRangeDates` selector, we should require this to be an object since that selector will always return one (it never returns `undefined`).

assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
```diff
@@ -29,21 +29,23 @@ import Data from 'googlesitekit-data';
 import LegacySearchConsoleDashboardWidgetKeywordTable from '../dashboard/LegacySearchConsoleDashboardWidgetKeywordTable';
 import DashboardModuleHeader from '../../../../components/dashboard/DashboardModuleHeader';
 import Layout from '../../../../components/layout/Layout';
-import { STORE_NAME } from '../../datastore/constants';
+import { MODULES_SEARCH_CONSOLE, DATE_RANGE_OFFSET } from '../../datastore/constants';
```
aaemnnosttv, line 32: I realize @tofumatt raised this in his review https://github.com/google/site-kit-wp/pull/2634#discussion_r561120025, but our convention has always been to leave the `STORE_NAME` unaliased within the scope of the module's datastore and components. If this is confusing, we can revisit it in a dedicated issue to iterate on this, but we need to keep this for consistency.

TL;DR – modules or core components importing _their own_ `STORE_NAME` should do so using the generic `STORE_NAME` constant. We only used the named variant for referencing stores external to the current context.

This will reduce the size of the diff here as well.

assets/js/modules/search-console/components/dashboard-details/DashboardDetailsWidgetKeywordsTable.js
```diff
@@ -29,21 +29,23 @@ import Data from 'googlesitekit-data';
...
 import { CORE_SITE } from '../../../../googlesitekit/datastore/site/constants';
-import { getCurrentDateRangeDayCount } from '../../../../util/date-range';
+import { CORE_USER } from '../../../../googlesitekit/datastore/user/constants';
```
aaemnnosttv, line 34: Just to reinforce the previous comment, this is correct.

assets/js/modules/search-console/util/report-date-range-args.js
```diff
@@ -0,0 +1,49 @@
...
+export const generateDateRangeArgs = ( dates ) => {
+	if ( ! dates ) {
+		return undefined;
+	}
+
+	const { startDate, endDate } = dates;
+
+	if ( ! startDate || ! endDate ) {
+		return undefined;
+	}
```
aaemnnosttv, line 40: We should avoid using `undefined` in utility classes like this as a general rule regarding invalid input.

We primarily use `undefined` intentionally when we want to use a default argument (usually in another function), or to mean that some data is not resolved yet (i.e. `! hasFinishedResolution`).

In this case, I think raising an error is fine since it should never happen anyways and `getDateRangeDates` _always_ returns an object with these keys, so an error would indicate improper use.

```js
invariant( isValidDateString( startDate ), 'a valid startDate is required.' );
invariant( isValidDateString( endDate ), 'a valid endDate is required.' );
```

cc: @tofumatt

assets/js/modules/search-console/util/report-date-range-args.js
```diff
@@ -0,0 +1,49 @@
...
+
+	const range = {
+		start_date: startDate.replace( /-/g, '' ),
+		end_date: endDate.replace( /-/g, '' ),
+	};
+
+	return range;
```
aaemnnosttv, line 47: Minor nitpick, but let's avoid creating unnecessary temporary variables before returning.
```suggestion
	return {
		start_date: startDate.replace( /-/g, '' ),
		end_date: endDate.replace( /-/g, '' ),
	};
```

assets/js/modules/search-console/util/report-date-range-args.test.js
```diff
@@ -0,0 +1,57 @@
...
+				expect( endDatePassed ).toBe( undefined );
+			} );
+		} );
+
+		it( 'should return an object containing a `start_date` key, the value of which is the `startDate` argument with "-" stripped', () => {
+			const { startDate, endDate } = allArgs;
+			const result = generateDateRangeArgs( { startDate, endDate } );
+			const expected = startDate.replace( /-/g, '' );
+
+			expect( result.start_date ).toBe( expected );
```
aaemnnosttv, line 46: While this works fine, using a variable for all inputs somewhat obscures the readability of the test.
That is, variables are great for describing what it represents and using for default valid values throughout the tests, but for cases where we want to bring attention to the expectation, it helps to use literal values.
E.g.
```suggestion
			const { startDate, endDate } = allArgs;
			const result = generateDateRangeArgs( { startDate: '2020-12-31', endDate } );

			expect( result.start_date ).toBe( '20201231' );
```

assets/js/modules/search-console/util/report-date-range-args.test.js
```diff
@@ -0,0 +1,57 @@
...
+ * limitations under the License.
+ */
+import { generateDateRangeArgs } from './report-date-range-args';
+
+describe( 'Search Console reporting date range arguments', () => {
+	describe( 'generateDateRangeArgs', () => {
+		const allArgs = {
+			startDate: '2020-12-18',
+			endDate: '2021-01-14',
+		};
```
aaemnnosttv, line 25: Since this variable is only destructured throughout, it would be cleaner to simply define `startDate` and `endDate` individually. It would also prevent these values from potentially being mutated and leaking between tests (not really a risk here but you get the idea).

This isn't a strict rule but a pattern you will likely see throughout the tests which has worked well for us.

aaemnnosttv (changes requested):
Thanks @johnPhillips – I see there was a lack of clarity around how the utility should work regarding invalid input, as well as confusion around the store name import.

I've left a few comments below to address, but it's all mostly small things. I left a few suggestions around the tests wells but nothing major.

All in all, this looks good just a handful of things to clean up then this should be good to go!

assets/js/modules/search-console/components/dashboard/GoogleSitekitSearchConsoleDashboardWidget.js
```diff
@@ -45,10 +45,11 @@ import getNoDataComponent from '../../../../components/legacy-notifications/noda
 import getDataErrorComponent from '../../../../components/legacy-notifications/data-error';
 import HelpLink from '../../../../components/HelpLink';
 import { getCurrentDateRangeDayCount } from '../../../../util/date-range';
-import { STORE_NAME } from '../../datastore/constants';
+import { MODULES_SEARCH_CONSOLE, DATE_RANGE_OFFSET } from '../../datastore/constants';
```
aaemnnosttv, line 48:
```suggestion
import { STORE_NAME, DATE_RANGE_OFFSET } from '../../datastore/constants';
```

assets/js/modules/search-console/components/dashboard/GoogleSitekitSearchConsoleDashboardWidget.js
```diff
@@ -59,19 +60,20 @@ const GoogleSitekitSearchConsoleDashboardWidget = () => {
 	const [ errorObj, setErrorObject ] = useState();
 	const [ loading, setLoading ] = useState( true );
 	const dateRange = useSelect( ( select ) => select( CORE_USER ).getDateRange() );
-	const propertyID = useSelect( ( select ) => select( STORE_NAME ).getPropertyID() );
-	const isDomainProperty = useSelect( ( select ) => select( STORE_NAME ).isDomainProperty() );
+	const propertyID = useSelect( ( select ) => select( MODULES_SEARCH_CONSOLE ).getPropertyID() );
+	const isDomainProperty = useSelect( ( select ) => select( MODULES_SEARCH_CONSOLE ).isDomainProperty() );
```
aaemnnosttv, line 64:
```suggestion
	const propertyID = useSelect( ( select ) => select( STORE_NAME ).getPropertyID() );
	const isDomainProperty = useSelect( ( select ) => select( STORE_NAME ).isDomainProperty() );
```

assets/js/modules/search-console/components/dashboard/GoogleSitekitSearchConsoleDashboardWidget.js
```diff
@@ -59,19 +60,20 @@ const GoogleSitekitSearchConsoleDashboardWidget = () => {
...
 	const searchConsoleDeepArgs = {
 		resource_id: propertyID,
-		num_of_days: getCurrentDateRangeDayCount(),
+		...generateDateRangeArgs( { startDate, endDate } ),
 	};
 	if ( isDomainProperty && referenceSiteURL ) {
 		searchConsoleDeepArgs.page = `*${ referenceSiteURL }`;
 	}
-	const searchConsoleDeepLink = useSelect( ( select ) => select( STORE_NAME ).getServiceURL(
+	const searchConsoleDeepLink = useSelect( ( select ) => select( MODULES_SEARCH_CONSOLE ).getServiceURL(
```
aaemnnosttv, line 76:
```suggestion
	const searchConsoleDeepLink = useSelect( ( select ) => select( STORE_NAME ).getServiceURL(
```

assets/js/modules/search-console/util/report-date-range-args.test.js
```diff
@@ -0,0 +1,52 @@
...
+import { generateDateRangeArgs } from './report-date-range-args';
+
+describe( 'Search Console reporting date range arguments', () => {
+	describe( 'generateDateRangeArgs', () => {
+		it( 'should throw if `startDate` or `endDate` arguments are not provided', () => {
+			try {
+				generateDateRangeArgs( { startDate: '2020-12-18' } );
+			} catch ( error ) {
+				expect( error.message ).toEqual( 'A valid endDate is required.' );
+			}
```
aaemnnosttv, line 27: This works but Jest provides a [`toThrow`](https://jestjs.io/docs/en/expect#tothrowerror) assertion which makes this a bit cleaner:

```suggestion
			expect(
				() => generateDateRangeArgs( { startDate: '2020-12-18' } )
			).toThrow( 'A valid endDate is required' )
```

Note the value for matching the exception message is optional and supports various methods.
So, not that we can't/shouldn't use try/catch for this, but we should use `toMatch` where possible (this is where we use elsewhere as well).

assets/js/modules/search-console/util/report-date-range-args.test.js
```diff
@@ -0,0 +1,52 @@
...
+
+			try {
+				generateDateRangeArgs( { endDate: '2021-01-14' } );
+			} catch ( error ) {
+				expect( error.message ).toEqual( 'A valid startDate is required.' );
+			}
```
aaemnnosttv, line 33:
```suggestion
			expect(
				() => generateDateRangeArgs( { endDate: '2021-01-14' } )
			).toThrow( 'A valid startDate is required' );
```

aaemnnosttv (commented):
This one is super close, just a few minor details; they all have suggestions so I'll apply them now.

aaemnnosttv (dismissed):
This worked great in my testing – thanks @johnPhillips !

assets/js/components/PostSearcher.js
```diff
@@ -42,7 +42,7 @@ const { useSelect, useDispatch } = Data;
 function PostSearcher() {
 	const [ canSubmit, setCanSubmit ] = useState( false );
 	const [ match, setMatch ] = useState( {} );
-	const analyticsModuleActive = useSelect( ( select ) => select( CORE_MODULES ).isModuleActive( 'analytics' ) );
+	const analyticsModuleConnected = useSelect( ( select ) => select( CORE_MODULES ).isModuleConnected( 'analytics' ) );
```
aaemnnosttv, line 45: This is weird because it's already in `develop` but shows as a change... 🤔 
https://github.com/google/site-kit-wp/blob/01e0edf785ba0ab381c212359d8da885be1e6620/assets/js/components/PostSearcher.js#L45-L63

aaemnnosttv (approved):

Issue #2287, opened 2020-10-29
Pass specific dates in Analytics deep links


felixarntz:
After #2236, all dynamic JS Analytics deep links in the new widgets should be updated to pass the correct dates in the URLs.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All widgets (using the Widget API) that include Analytics deep links should be updated to pass the exact dates from the currently selected date range in the URL:
    * The path of the URL should include "query parameter-like" values like `_u.date00=YYYYMMDD&_u.date01=YYYYMMDD` (they need to be in the path though, not actual query parameters). Go to  the Analytics FE and select a custom date range to see an example. Note that there may already be other "query-parameter-like" values in the URL in which case they need to be combined accordingly (concatenated with `&`).
* For widgets that also _compare_ with a second date range (i.e. `compareStartDate` and `compareEndDate`), those dates _also_ need to be passed in the URL in addition to the above, e.g. `_u.date10=YYYYMMDD&_u.date11=YYYYMMDD`.
* A full example (with all four dates) could be `_u.date00=20201023&_u.date01=20201029&_u.date10=20201016&_u.date11=20201022`.
* The additional arguments should be provided via `getServiceReportURL` (as probably already in place for all of these widgets anyway).

## Implementation Brief
* **Once #2405 has been merged**,
* Using `/assets/js/modules/analytics/components/dashboard/AnalyticsDashboardWidgetTopPagesTable.js`
	- Query the user store to get the date range:
	```js
		const dateRangeDates = useSelect( ( select ) => select( CORE_USER ).getDateRangeDates( {
			offsetDays: DATE_RANGE_OFFSET,
		} ) );
	```
	- The `DATE_RANGE_OFFSET` should be imported from `/assets/js/modules/analytics/datastore/constants.js`
* Create new file `/assets/js/modules/analytics/util/report-date-range-args` and using newly created file,
	- Create a new function `generateDateRangeArgs` which accepts the returned object from the `getDateRangeDates` selector. The object contains the following keys:
		- `startDate`
		- `endDate`
		- `compareStartDate` : if `compare` is true when calling `getDateRangeDates`. See example below.
		- `compareEndDate`: if `compare` is true when calling `getDateRangeDates`. See example below.
	- The `generateDateRangeArgs` should return an object with the following keys:
		- `'_u.date00': startDate.replace( /-/g, '' ),`
[...]

PR #2636, opened 2021-01-18
Pass specific dates in Analytics deep links

johnPhillips:
## Summary

Addresses issue #2287

## Checklist


assets/js/modules/analytics/util/report-date-range-args.test.js
```diff
@@ -0,0 +1,74 @@
...
+				const { startDate, endDate } = allArgs;
+
+				const noArgsPassed = generateDateRangeArgs();
+				const startDatePassed = generateDateRangeArgs( { startDate } );
+				const endDatePassed = generateDateRangeArgs( { endDate } );
+
+				expect( noArgsPassed ).toBe( undefined );
+				expect( startDatePassed ).toMatchObject( { startDate } );
+				expect( endDatePassed ).toMatchObject( { endDate } );
+			} );
```
tofumatt, line 40: As in the Search Console PR, let's not return these arguments when they aren't valid. Whichever method you use in #2634 let's do here as well. 🙂

assets/js/modules/analytics/util/report-date-range-args.js
```diff
@@ -0,0 +1,53 @@
...
+ * Generates an object with the appropriate structure to pass to report selectors.
+ *
+ * @since n.e.x.t
+ *
+ * @param {Object} dates                    Object containing the dates.
+ * @param {string} dates.startDate          Start date in YYYY-MM-DD format.
+ * @param {string} dates.endDate            End date in YYYY-MM-DD format.
+ * @param {string} [dates.compareStartDate] Optional. Start date in YYYY-MM-DD format.
+ * @param {string} [dates.compareEndDate]   Optional. End date in YYYY-MM-DD format.
+ * @return {Object} Date range object.
```
tofumatt, line 29:
```suggestion
 * @return {Object|undefined} Date range object (or `undefined` if no dates are supplied).
```

tofumatt (changes requested):
Approach looks good, but the same updates required in #2634 are required here:

- [ ] Use 2021 as the copyright year not 2020
- [ ] Update `STORE_NAME` imports
- [ ] Don't return `dates` when they are invalid

assets/js/modules/analytics/components/dashboard/AnalyticsDashboardWidgetTopPagesTable.js
```diff
@@ -118,6 +124,7 @@ const AnalyticsDashboardWidgetTopPagesTable = ( props ) => {
 			'mdc-layout-grid__cell',
 			`mdc-layout-grid__cell--span-${ colspan }`
 		) }>
+			zoe mamaa
```
tofumatt, line 127: 😆 

```suggestion
```

assets/js/modules/analytics/util/report-date-range-args.test.js
```diff
@@ -0,0 +1,74 @@
...
+ */
+import { generateDateRangeArgs } from './report-date-range-args';
+
+describe( 'Analytics reporting date range arguments', () => {
+	describe( 'generateDateRangeArgs', () => {
+		const allArgs = {
+			startDate: 'start-date',
+			endDate: 'end-date',
+			compareStartDate: 'compare-start-date',
+			compareEndDate: 'compare-end-date',
```
tofumatt, line 26: Are these values reflective of the actual ones used? Wouldn't they be `YYYY-MM-DD` or something like that?

Let's use actual values to test this so it's easier for the next developer to read and understand the motivation of the test. Technically speaking it's the same if it's letters or numbers, but it better conveys the intent and improves readability of the test for it to look like real data.

assets/js/modules/analytics/util/report-date-range-args.test.js
```diff
@@ -0,0 +1,74 @@
...
+				const result = generateDateRangeArgs( allArgs );
+				const expected = allArgs.compareStartDate.replace( /-/g, '' );
+
+				expect( result[ '_u.date10' ] ).toBe( expected );
+			} );
+			it( 'should return an object containing a `_u.date11` key, the value of which is the `compareEndDate` argument with "-" stripped', () => {
+				const result = generateDateRangeArgs( allArgs );
+				const expected = allArgs.compareEndDate.replace( /-/g, '' );
+
+				expect( result[ '_u.date11' ] ).toBe( expected );
```
tofumatt, line 70: Having to jump around the code a bit less to compare the expectation here might be nice, and I think improves the readability of the test. The spec outlines the intent well, but leaving the hardcoded argument here seems fine to me, and it makes the test code a bit more readable/immediately understandable.

What do you think?

```suggestion
				const expected = allArgs.compareEndDate.replace( /-/g, '' );

				expect( result[ '_u.date11' ] ).toBe( 'compareenddate' );
```

tofumatt (changes requested):
I think there are some readability improvements we can make to the tests, and there's one string that looks like a debugging string in the changes, but otherwise looks good. 👍🏻

johnPhillips:
Thanks for the review @tofumatt 
I agree with your suggestions and I'll implement those.
The debugging string was there when I took over the ticket, and even though I removed it, it seems to have come back like all good debugging strings do 😂

tofumatt:
@googlebot I consent.

tofumatt:
(Saw this had a bit of a merge issue thanks to some updates/file name changes so I fixed the conflict.)

tofumatt (dismissed):

assets/js/modules/analytics/util/report-date-range-args.test.js
```diff
@@ -0,0 +1,68 @@
...
+
+			expect( result[ '_u.date01' ] ).toBe( '20210112' );
+		} );
+
+		describe( 'if `compareStartDate` and `compareEndDate` keys are passed', () => {
+			it( 'should return an object containing a `_u.date10` key, the value of which is the `compareStartDate` argument with "-" stripped', () => {
+				const result = generateDateRangeArgs( allArgs );
+				expect( result[ '_u.date10' ] ).toBe( '20200112' );
+			} );
+			it( 'should return an object containing a `_u.date11` key, the value of which is the `compareEndDate` argument with "-" stripped', () => {
```
aaemnnosttv, line 62: Nitpicking this but please provide a blank link between `it` cases 👍 
```suggestion

			it( 'should return an object containing a `_u.date11` key, the value of which is the `compareEndDate` argument with "-" stripped', () => {
```

aaemnnosttv (dismissed):
Looks good @johnPhillips – just a few things to address; most of it applies from [my review of the Search Console version of this same issue](https://github.com/google/site-kit-wp/pull/2634#pullrequestreview-581225020) so I won't duplicate it here. Mostly small changes though.

assets/js/modules/analytics/util/report-date-range-args.js
```diff
@@ -0,0 +1,57 @@
...
+
+	invariant( isValidDateString( startDate ), 'A valid startDate is required.' );
+	invariant( isValidDateString( endDate ), 'A valid endDate is required.' );
+
+	const range = {
+		'_u.date00': startDate.replace( /-/g, '' ),
+		'_u.date01': endDate.replace( /-/g, '' ),
+	};
+
+	if ( isValidDateString( compareStartDate ) && isValidDateString( compareEndDate ) ) {
```
aaemnnosttv, line 51: Let's use `invariant` with these dates too, but it should only raise an error if provided of course.
That is, if either one is provided, we should require and validate both.

assets/js/modules/analytics/util/report-date-range-args.test.js
```diff
@@ -0,0 +1,76 @@
...
+				generateDateRangeArgs( { startDate: '2020-12-18' } );
+			} catch ( error ) {
+				expect( error.message ).toEqual( 'A valid endDate is required.' );
+			}
+
+			try {
+				generateDateRangeArgs( { endDate: '2021-01-14' } );
+			} catch ( error ) {
+				expect( error.message ).toEqual( 'A valid startDate is required.' );
+			}
```
aaemnnosttv, line 33: Please use [`.toThrow`](https://jestjs.io/docs/en/expect#tothrowerror) for these. I mentioned this on the SC PR too but you probably didn't see [my comment](https://github.com/google/site-kit-wp/pull/2634#discussion_r568844042) since it was resolved shortly after 😄

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/index.js
```diff
@@ -138,7 +138,7 @@ function DashboardAllTrafficWidget() {
 		};
```
aaemnnosttv, line 138: Looks like we're missing the date range args in the report args here ^

assets/js/modules/analytics/components/dashboard/DashboardPopularPagesWidget.js
```diff
@@ -72,7 +74,10 @@ function DashboardPopularPagesWidget( { WidgetReportZero } ) {
 		};
 
 		return {
-			analyticsMainURL: store.getServiceReportURL( 'content-pages' ),
+			analyticsMainURL: store.getServiceReportURL(
+				'content-pages',
+				{ ...generateDateRangeArgs( { startDate, endDate, compareStartDate, compareEndDate } ) }
```
aaemnnosttv, line 79: No need to spread here since there are no other args provided for this URL.

assets/js/modules/analytics/util/report-date-range-args.test.js
```diff
@@ -0,0 +1,76 @@
...
+			} );
+
+			it( 'should return an object containing a `_u.date11` key, the value of which is the `compareEndDate` argument with "-" stripped', () => {
+				const result = generateDateRangeArgs( {
+					startDate: '2020-01-12',
+					endDate: '2021-01-12',
+					compareStartDate: '2020-01-12',
+					compareEndDate: '2021-01-12',
+				} );
+				expect( result[ '_u.date11' ] ).toBe( '20210112' );
```
aaemnnosttv, line 72: This applies to the other tests here but please use different dates for each key so that the assertions ensure the proper date is assigned to the correct key in the resulting object.

Also, since this is a simple utility, let's assert the full result matches our expectation. 
E.g.
```js
expect( result ).toEqual( {
    '_u.date00': // date1,
    '_u.date01': // date2,
    '_u.date10': // date3,
    '_u.date11': // date4,
} )
```

aaemnnosttv (commented):
Thanks @johnPhillips – just a few things left to address here 👍

aaemnnosttv:
@googlebot I consent.

assets/js/modules/analytics/util/report-date-range-args.test.js
aaemnnosttv: Sorry, I missed that you had a separate test for each key which also works, but is a little overkill 😄 
No need for further changes here though 👍

aaemnnosttv (commented):
@johnPhillips everything seems to be working as expected here, although I'm noticing a few places where it may be missing but I'm not sure based on the way the ACs are worded.

Passing over to @felixarntz for a final look over. Otherwise this looks good to me!

aaemnnosttv:
@johnPhillips would you please fix the merge conflict?

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/index.js
```diff
@@ -132,10 +134,11 @@ function DashboardAllTrafficWidget( { Widget, WidgetReportZero } ) {
 		reportArgs = {
 			'explorer-table.plotKeys': '[]',
 			'_r.drilldown': `analytics.pagePath:${ getURLPath( entityURL ) }`,
+			...generateDateRangeArgs( { startDate, endDate, compareStartDate, compareEndDate } ),
```
felixarntz, line 137: You currently add the dates only in this `entityURL` clause, which only applies on the single URL dashboard view, where data is displayed not for the entire site but only a single URL. The date arguments need to be added unconditionally though, so you could put them into the `reportArgs` assignment above, and then in this clause, merge the `entityURL`-specific arguments into it.

felixarntz (changes requested):
This is almost good to go, just one thing is slightly incorrect which results in the All Traffic widget on the Site Kit dashboard not including date arguments.

felixarntz (approved):
Awesome!

felixarntz (approved):

Issue #2347, opened 2020-11-10
Improve Reset button UX (don't close modal on reset)


tofumatt:
## Feature Description

Currently the Reset flow is slightly awkward and doesn't provide the user much feedback of what's happening after they reset their site. See the following screencast:

![2020-11-10 23 31 26](https://user-images.githubusercontent.com/90871/98746382-4c383b00-23ad-11eb-8372-82952df33973.gif)

I outlined some of the issues with this UX and some suggestions here: https://github.com/google/site-kit-wp/pull/2268#discussion_r512715549

-----

I don't think showing a generic loading animation is a good indication of what's _currently happening_, because it would remove "Reset Site Kit" with "…" (some loading animation) so you might not _know_ what's loading/resetting.

I think it'd be nicer if the dialog didn't _close_ on `handleUnlinkConfirm`. Instead, the modal's content should update after the "Reset" button is clicked—changing the button to be disabled with the text "Site Kit is resetting" would be nice.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Clicking the "Reset" button should not close the modal because this action relies on a page reload, and closing it without any loading indicator results in confusing UX.
* A small loading indicator should appear next to the "Reset" button once clicked. It should persist until the page is reloaded (e.g. relying on `isNavigating` selector from #2497).

## Implementation Brief
* **Once #2497 has been merged**
* Using `/assets/js/components/Dialog.js`
	* Import the `Spinner` component. `/assets/js/components/Spinner.js`
	* Add a new boolean prop called `showSpinner` which is false by default.
	* if `showSpinner` is true, then show the `<Spinner/>` component. The `Spinner` component should be outputted next to the "Confirm" button.
* Using `/assets/js/components/ResetButton.js`
	* Query the navigation store via the `isNavigating` and `getNavigateURL` selectors.
	* If `isNavigating` is true, and `postResetURL` is equal to the URL returned by `getNavigateURL`, set `showSpinner` to true for the `Dialog` component.

### Test Coverage

* No changes needed.

### Visual Regression Changes

* No changes expected.

## QA Brief

* Reset Site Kit and you should see a spinning animation next to the "Confirm" button.

## Changelog entry

* Improve UX when clicking the Reset dialog button, keeping the dialog open in a loading state while completing the process.

felixarntz:
[...]

PR #2637, opened 2021-01-18
Improve Reset button UX (don't close modal on reset) 

johnPhillips:
## Summary

Addresses issue #2347

## Checklist

### DIFF
--- a/assets/js/components/Dialog.js
+++ b/assets/js/components/Dialog.js
@@ -115,7 +115,10 @@ const Dialog = ( {
 								>
 									{ confirmButton ? confirmButton : __( 'Disconnect', 'google-site-kit' ) }
 								</Button>
-								<Spinner isSaving={ showSpinner } />
+								<Spinner
+									className="googlesitekit-margin-right-auto"
+									isSaving={ showSpinner }
+								/>
 								<Link
 									className="mdc-dialog__cancel-button"
 									onClick={ handleDialog }
--- a/assets/js/components/Dialog.js
+++ b/assets/js/components/Dialog.js
@@ -36,6 +36,7 @@ import { __ } from '@wordpress/i18n';
 import Button from './Button';
 import Link from './Link';
 import { MDCDialog } from '../material-components';
+import Spinner from './Spinner.js';
 
 const Dialog = ( {
 	dialogActive,
@@ -47,6 +48,7 @@ const Dialog = ( {
 	confirmButton,
 	dependentModules,
 	danger,
+	showSpinner = false,
 } ) => {
 	const dialogRef = useCallback( ( el ) => {
 		if ( el !== null ) {
@@ -109,9 +111,11 @@ const Dialog = ( {
 								<Button
 									onClick={ handleConfirm }
 									danger={ danger }
+									disabled={ showSpinner }
 								>
 									{ confirmButton ? confirmButton : __( 'Disconnect', 'google-site-kit' ) }
 								</Button>
+								<Spinner isSaving={ showSpinner } />
 								<Link
 									className="mdc-dialog__cancel-button"
 									onClick={ handleDialog }
--- a/assets/js/components/Dialog.js
+++ b/assets/js/components/Dialog.js
@@ -48,7 +48,7 @@ const Dialog = ( {
 	confirmButton,
 	dependentModules,
 	danger,
-	showSpinner = false,
+	inProgress = false,
 } ) => {
 	const dialogRef = useCallback( ( el ) => {
 		if ( el !== null ) {
@@ -111,18 +111,18 @@ const Dialog = ( {
 								<Button
 									onClick={ handleConfirm }
 									danger={ danger }
-									disabled={ showSpinner }
+									disabled={ inProgress }
 								>
 									{ confirmButton ? confirmButton : __( 'Disconnect', 'google-site-kit' ) }
 								</Button>
 								<Spinner
-									className="googlesitekit-margin-right-auto"
-									isSaving={ showSpinner }
+									isSaving={ inProgress }
 								/>
 								<Link
-									className="mdc-dialog__cancel-button"
+									className="googlesitekit-margin-left-auto mdc-dialog__cancel-button"
 									onClick={ handleDialog }
 									inherit
+									disabled={ inProgress }
 								>
 									{ __( 'Cancel', 'google-site-kit' ) }
 								</Link>
--- a/assets/js/components/Spinner.js
+++ b/assets/js/components/Spinner.js
@@ -20,10 +20,10 @@
  * A single module. Keeps track of its own active state and settings.
  */
 
-function Spinner( { isSaving } ) {
+function Spinner( { isSaving, className } ) {
 	return (
 		<span
-			className="spinner"
+			className={ [ 'spinner', className ].join( ' ' ) }
 			style={ {
 				display: ( isSaving ? 'inline-block' : 'none' ),
 				float: 'none',
--- a/assets/js/components/Spinner.js
+++ b/assets/js/components/Spinner.js
@@ -20,10 +20,10 @@
  * A single module. Keeps track of its own active state and settings.
  */
 
-function Spinner( { isSaving, className } ) {
+function Spinner( { isSaving } ) {
 	return (
 		<span
-			className={ [ 'spinner', className ].join( ' ' ) }
+			className="spinner"
 			style={ {
 				display: ( isSaving ? 'inline-block' : 'none' ),
 				float: 'none',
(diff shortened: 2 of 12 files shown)


assets/js/components/Dialog.js
```diff
@@ -47,6 +48,7 @@ const Dialog = ( {
 	confirmButton,
 	dependentModules,
 	danger,
+	showSpinner = false,
```
aaemnnosttv, line 51: Let's name this a bit more abstractly since the spinner is more of an implementation detail, but also because this prop controls more than just the visibility of the spinner.

How about `inProgress` ?

assets/js/components/Spinner.js
```diff
@@ -20,10 +20,10 @@
  * A single module. Keeps track of its own active state and settings.
  */
 
-function Spinner( { isSaving } ) {
+function Spinner( { isSaving, className } ) {
 	return (
 		<span
-			className="spinner"
+			className={ [ 'spinner', className ].join( ' ' ) }
```
aaemnnosttv, line 26: Please use the `classnames` package for dynamic class names. We already use it in many places and is our go-to utility for this kind of thing 👍

assets/js/components/Dialog.js
```diff
@@ -109,9 +111,14 @@ const Dialog = ( {
...
+									disabled={ showSpinner }
 								>
 									{ confirmButton ? confirmButton : __( 'Disconnect', 'google-site-kit' ) }
 								</Button>
+								<Spinner
+									className="googlesitekit-margin-right-auto"
+									isSaving={ showSpinner }
+								/>
 								<Link
 									className="mdc-dialog__cancel-button"
```
aaemnnosttv, line 123: How about we put the margin on the cancel button (as a `left-auto`) instead since it will always be the last thing. Then in this case we shouldn't need to update `Spinner` to support an additional `className` either.

assets/js/components/Dialog.js
```diff
@@ -109,9 +111,14 @@ const Dialog = ( {
...
 									onClick={ handleDialog }
```
aaemnnosttv, line 124: Let's also disable the `Link` here using the same logic as the button – our link component supports the `disabled` prop even though normal `<a>` tags don't 😉

aaemnnosttv (changes requested):
Thanks @johnPhillips – this is looking good! Just a few points to address and then this should be good to go 👍

assets/js/components/Dialog.js
johnPhillips: I agree and am very happy to change it. However, `showSpinner` was specified in the IB:
> Add a new boolean prop called showSpinner which is false by default.

Just wanted to check there wasn't some specific reason for this requirement before I change it?

assets/js/components/Spinner.js
johnPhillips: Ah, perfect. Thank you

assets/js/components/Dialog.js
johnPhillips: Great idea, I hadn't thought of that 👍

johnPhillips:
@aaemnnosttv One question I did have was: why rely on the navigation selectors and not just show the spinner once the 'reset' button is clicked?
The navigation call happens last, after the async call to perform all the resetting:
```js
// assets/js/components/ResetButton.js: 70

const handleUnlinkConfirm = useCallback( async () => {
	await reset();
	clearWebStorage();
	navigateTo( postResetURL ); // This is the point at which we will show the spinner
}, [ reset, postResetURL ] );
```
If I understand all this correctly, we are waiting until everything is done before re-rendering the modal just to show the spinner? Couldn't we do something like:
```js
const handleUnlinkConfirm = useCallback( async () => {
	// Update the 'inProgress' prop to show the spinner first
	await reset();
	clearWebStorage();
	navigateTo( postResetURL ); 
}, [ reset, postResetURL ] );
```

aaemnnosttv:
> why rely on the navigation selectors and not just show the spinner once the 'reset' button is clicked?

Good call @johnPhillips – in most places we have two pieces of state that control the `inProgress` state for this kind of thing: one piece of state for the submit action (which in some cases might fail for some reason) and then the navigation which persists afterwards until the page reloads which is only set once _we know_ we are going to navigate. Then the progress would be passed along as something like `isSubmitting || isNavigating` to remain true for either.

We're obviously missing the first part of it here which is a detail that was missed in definition (good catch!). Let's add that here since it should be very quick to do. We don't actually need to set some state on click though since we already have selectors built in for all of our fetch actions (part of our async fetch infrastructure – see `createFetchStore`). In this case however, we just need to select `isDoingReset` from `core/site` which will return `true` as long as the reset fetch request is in progress; that should be good enough for our purposes here. Can you give this a try and see if this works without introducing any gaps in the `inProgress` state?

Also, it looks like there is a JS unit test failure related to the use of the new selector. Can you have a look?

johnPhillips:
@aaemnnosttv I have updated the PR with the addition of the `isDoingReset` selector, but it seems there might be a gap. 
When I log the values of `isDoingReset` and `isNavigatingToPostResetURL`, I see this in the console after I hit the reset button:
```js
{ isDoingReset: true, isNavigatingToPostResetURL: false }
{ isDoingReset: false, isNavigatingToPostResetURL: false } // render cycle where both are false
{ isDoingReset: false, isNavigatingToPostResetURL: true }
```
There is no visible gap (the spinner remains), but I am seeing a little stutter / jank:

reset

What do you think?

aaemnnosttv:
Hey John, thanks for the screencast. It's better than what we have now but I agree that it is a bit odd.

If the DOM is updating in that render cycle I would expect to see the button and link become un-disabled at the same time. Can you take a look in devtools to see if you can see if the spinner is actually being removed from the DOM temporarily there?

The PR overall looks good to me so this is the only thing left to address I think.

Worst case, I think we just need to add a minimum amount of time to the spinner displaying when the button is clicked which would expire after a few seconds after which it would only spin if it was actually navigating. Not that great, but not too bad. @felixarntz and I talked about this before as well with the definition of the current actions and selectors so if this is a problem, then we might need to adjust how the datastore part of this needs to work to account for these kinds of things.

johnPhillips:
Hey @aaemnnosttv, thanks for the comment. 

> If the DOM is updating in that render cycle I would expect to see the button and link become un-disabled at the same time. Can you take a look in devtools to see if you can see if the spinner is actually being removed from the DOM temporarily there?

What actually happens is that the spinner has its `display` property toggled - it's always in the DOM, just with a `display: none` when it isn't `inProgress`. 
Apart from that, what I see if I use breakpoints is exactly what you'd expect though:

1. Click reset button
2. The reset action fires and the API request is sent
3. The selectors update and force a re-render with `isDoingReset: true` and `isNavigatingToPostResetURL: false`
4. This results in `inProgress: true`, and the greyed out button and spinner are shown
5. The reset completes and we get `isDoingReset: false` and `isNavigatingToPostResetURL: false` so `inProgress: false`
6. The original red button is shown, spinner is hidden
7. Navigation starts and we get `isNavigatingToPostResetURL: true` therefore `inProgress: true`
8. Render greyed out button and spinner again
9. Navigates to next page

The render at step 6 happens too fast for me to see without breakpoints though.

> Worst case, I think we just need to add a minimum amount of time to the spinner displaying when the button is clicked which would expire after a few seconds after which it would only spin if it was actually navigating. 

I am not _entirely_ sure what you mean here, sorry. Are you suggesting a `setTimeout()` or similar? Let me know what you think and I'll happily implement, or we can jump on a quick call if that's easier 👍

aaemnnosttv:
@johnPhillips 

> I am not _entirely_ sure what you mean here, sorry. Are you suggesting a `setTimeout()` or similar? Let me know what you think and I'll happily implement, or we can jump on a quick call if that's easier 👍

Yeah, something like this...

```js
// import { useDebounce } from '@wordpress/compose'
const isDoingReset = select() //...
const isNavigatingToPostResetURL = select() //...
const [ inProgress, setInProgress ] = useState( false );
const debouncedSetInProgress = useDebounce( setInProgress, 2000 );
// 
const mediatedSetInProgress = ( bool ) => bool ? setInProgress( true ) : debouncedSetInProgress( false );

useEffect( () => {
	mediatedSetInProgress( isDoingReset || isNavigatingToPostResetURL );
}, [ isDoingReset, isNavigatingToPostResetURL ] )
```

aaemnnosttv (commented):
Looking good @johnPhillips – this is pretty much good to go, we just need to re-work the polyfilling of the hook a little bit and tidy up a few things. I tried it out though and it works well!

Should be fairly quick to address; let me know if you have any questions 👍

aaemnnosttv:
@googlebot I consent.

aaemnnosttv (approved):
LGTM, thanks @johnPhillips !

Issue #2288, opened 2020-10-29
Add e2e coverage for user input flow


felixarntz:
e2e test coverage should be added for the new user input flow.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* e2e test coverage should be added for the new user input flow. The following scenarios should be covered:

### "New" user
1. Click "Sign in with Google" on setup screen (which will internally mark user input as required before redirecting to authentication service).
2. Come back from setup to the Site Kit dashboard URL technically, but get redirected to user input screen.
3. Go through all five question, choose some answers.
4. Arrive at preview screen with all questions and the given answers, click "Submit".
5. Land on dashboard with the "Congrats" notification for the user input flow present.

### "Existing" user (already connected)
1. Visit the Site Kit dashboard, see notification prompting to fill user input.
2. Click the CTA link in the notification, arriving on the user input screen.
3. The rest matches steps 3.-5. for the **"New" user** scenario above.

## Implementation Brief

* Create an E2E plugin that creates a REST endpoint, similar to other E2E REST Routes:
```php
<?php
namespace Google\Site_Kit\Core\Authentication;

use Google\Site_Kit\Core\Authentication\User_Input_State;
use Google\Site_Kit\Core\REST_API\REST_Routes;
use Google\Site_Kit\Core\Storage\User_Setting;

add_action(
	'rest_api_init',
	function () {
		if ( ! defined( 'GOOGLESITEKIT_PLUGIN_MAIN_FILE' ) ) {
			return;
		}

		register_rest_route(
			REST_Routes::REST_ROOT,
			'e2e/auth/user-input',
			array(
				'methods'             => WP_REST_Server::EDITABLE,
				'callback'            => function ( WP_REST_Request $request ) {
					update_option(
						User_Setting::OPTION,
						sanitize_text_field( $request['user_input'] )
					);

					return array( 'success' => true );
				},
				'permission_callback' => '__return_true',
			)
		);
	},
	0
);
```
* Create an E2E JS Util (eg: `tests/e2e/utils/set-user-input.js`) that allows this REST endpoint to be used to set a User Input settings option value:
```js
/**
 * Internal dependencies
 */
import { wpApiFetch } from './wp-api-fetch';

/**
 * Sets the User Input option.
 *
 * @since n.e.x.t
 *
 * @param {string} value The new user input state to set.
 * @return {*} Resolved value from `apiFetch` promise.
 */
[...]

PR #2638, opened 2021-01-18
Enhancement/2288 user input e2e tests

eugene-manuilov:
## Summary

Addresses issue #2288

## Checklist


tests/e2e/specs/user-input-questions.test.js
```diff
@@ -0,0 +1,152 @@
...
+
+		await Promise.all( [
+			expect( page ).toClick( '.googlesitekit-user-input__buttons--next' ),
+			page.waitForNavigation(),
+		] );
+	}
+
+	async function expectSuccessNotification() {
+		await page.waitForSelector( '#user-input-success' );
+		await expect( page ).toMatchElement( '#user-input-success' );
```
aaemnnosttv, line 66: This is redundant here as `waitForSelector` will fail the test if it times out (default 30s).

We shouldn't need to wait that long so I think it would be better to use just the `toMatchElement` here, but using the `timeout` option to use a more reasonable limit in the event of a legitimate failure (maybe 5s?).

tests/e2e/plugins/user-input.php
```diff
@@ -0,0 +1,139 @@
...
+
+function e2e_user_input_settings_proxy_handler( $pre, $args, $url ) {
+	$google_proxy            = new Google_Proxy( new Context( GOOGLESITEKIT_PLUGIN_MAIN_FILE ) );
+	$user_input_settings_url = $google_proxy->url( Google_Proxy::USER_INPUT_SETTINGS_URI );
+	if ( $url !== $user_input_settings_url ) {
+		return $pre;
+	}
+
+	$user_id  = get_current_user_id();
+	$defaults = array(
```
aaemnnosttv, line 43: Is all of this necessary/used by the tests? It looks like most of this might have been copied from before so I realize it probably wasn't all written for this, but it would be good to keep this to only what's needed if we can reduce it. What do you think?

tests/e2e/plugins/user-input.php
```diff
@@ -0,0 +1,139 @@
...
+use Google\Site_Kit\Core\Storage\User_Options;
+
+register_activation_hook( __FILE__, 'e2e_user_input_settings_reset' );
+register_deactivation_hook( __FILE__, 'e2e_user_input_settings_reset' );
+
+add_filter( 'pre_http_request', 'e2e_user_input_settings_proxy_handler', 10, 3 );
+
+function e2e_user_input_settings_reset() {
+	$user_options     = new User_Options( new Context( GOOGLESITEKIT_PLUGIN_MAIN_FILE ) );
+	$user_input_state = new User_Input_State( $user_options );
```
aaemnnosttv, line 27: While this works fine for the tests, we should account for this potentially being activated (or deactivated) with Site Kit inactive. Other E2E plugins check for a constant or use Site Kit provided actions to prevent such errors.

Maybe we should `wp_die` on `plugins_loaded` if Site Kit isn't active? I'm open to ideas here.

assets/js/googlesitekit/datastore/user/user-input-settings.js
```diff
@@ -109,7 +109,7 @@ const baseActions = {
 		const registry = yield Data.commonActions.getRegistry();
 		yield clearError( 'saveUserInputSettings', [] );
 
-		const settings = registry.select( STORE_NAME ).getUserInputSettings();
+		const settings = registry.select( STORE_NAME ).getUserInputSettings() || {};
```
aaemnnosttv, line 112: It seems like maybe a test should be added/updated for the changes to this action.

aaemnnosttv (changes requested):
Thanks @eugene-manuilov – this worked well in my testing, and just a few things below to address, but nothing major; nice work!

One thing that I noticed (which I had to change to get the tests to run) was a PHP notice that seems to indicate a problem in the `User_Input_Settings` class.

```
Notice: Array to string conversion in /var/www/html/wp-content/plugins/google-site-kit/includes/Core/Util/User_Input_Settings.php on line 134 Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/wp-content/plugins/google-site-kit/includes/Core/Util/User_Input_Settings.php:134) in /var/www/html/wp-admin/includes/misc.php on line 1310
Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/wp-content/plugins/google-site-kit/includes/Core/Util/User_Input_Settings.php:134) in /var/www/html/wp-includes/pluggable.php on line 1296
Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/wp-content/plugins/google-site-kit/includes/Core/Util/User_Input_Settings.php:134) in /var/www/html/wp-includes/pluggable.php on line 1299
```
Can you take a look?

Also, I noticed that when the success notice is shown, the CTA to provide user input is also showing:

image

tests/e2e/plugins/user-input.php
eugene-manuilov: It's needed because that's what we expect to receive even if there is nothing saved yet. We can't remove it.

tests/e2e/specs/user-input-questions.test.js
eugene-manuilov: This is needed because we should wait till the page is redirected. If I remove it, tests fail.

tests/e2e/plugins/user-input.php
eugene-manuilov: Updated to check whether the `GOOGLESITEKIT_PLUGIN_MAIN_FILE` constant is defined.

tests/e2e/specs/user-input-questions.test.js
eugene-manuilov: What we can remove is the expect clause: `await expect( page ).toMatchElement( '#user-input-success' );`. Updated.

assets/js/googlesitekit/datastore/user/user-input-settings.js
eugene-manuilov: This seems to be not needed anymore. Reverted it.

eugene-manuilov:
> One thing that I noticed (which I had to change to get the tests to run) was a PHP notice that seems to indicate a problem in the User_Input_Settings class.

This should be fixed.

> Also, I noticed that when the success notice is shown, the CTA to provide user input is also showing:

This happens because we don't use a real proxy server and the authentication class can't set the appropriate user input flag.

eugene-manuilov:
@aaemnnosttv this is ready for the next round of review. I have reworked it slightly, plus added a few utility functions and updated e2e actions to help with troubleshooting. Hope it works for you.

includes/Core/Util/User_Input_Settings.php
```diff
@@ -128,18 +128,22 @@ function( $setting ) {
...
 			'body'    => array(
 				'site_id'     => $creds['oauth2_client_id'],
 				'site_secret' => $creds['oauth2_client_secret'],
 			),
 		);
 
+		if ( ! empty( $access_token ) && is_string( $access_token ) ) {
+			$user_input_settings_args['headers']['Authorization'] = 'Bearer ' . $access_token;
+		}
+
```
aaemnnosttv, line 146: Why change this? The access token is required for making this request, is it not?

tests/e2e/utils/step-and-screenshot.js
```diff
@@ -0,0 +1,78 @@
...
+ * Makes a step and takes a screenshot.
+ *
+ * @since n.e.x.t
+ *
+ * @param {string}           name Step name.
+ * @param {Function|Promise} cb   Step callback function or a promise object.
+ * @return {Promise} Promise object.
+ */
+export function step( name, cb ) {
+	return new Promise( async ( success, error ) => {
```
aaemnnosttv, line 68: Minor difference, but let's use the same names we do elsewhere for consistency.
```suggestion
	return new Promise( async ( resolve, reject ) => {
```

tests/e2e/utils/step-and-screenshot.js
```diff
@@ -0,0 +1,78 @@
...
+		try {
+			const results = await ( typeof cb === 'function' ? cb() : cb );
+			await screenshot( name );
+			success( results );
```
aaemnnosttv, line 72:
```suggestion
			resolve( results );
```

tests/e2e/utils/step-and-screenshot.js
```diff
@@ -0,0 +1,78 @@
...
+		} catch ( err ) {
+			await screenshot( name );
+			error( err );
```
aaemnnosttv, line 75:
```suggestion
			reject( err );
```

tests/e2e/utils/step-and-screenshot.js
```diff
@@ -0,0 +1,78 @@
...
+
+	await mkdir( screenshotsDir, {
+		mode: 0o755,
+		recursive: true,
+	} );
+
+	await page.screenshot( {
+		path: `${ screenshotsDir }/${ screenshotIndex.toString().padStart( 2, '0' ) }-${ name.replace( /\W+/, '-' ) }.png`,
+		type: 'png',
+		fullPage: true,
```
aaemnnosttv, line 54: I would prefer we don't default to `fullPage` here as this loses what the browser "sees" at that moment in the viewport which can be relevant as to why the test failed. I see how this could be useful in some cases though so perhaps we should accept an optional second argument for screenshot options that we can spread into the call here. We could then add this as the 3rd parameter to `step` below.

aaemnnosttv (commented):
Thanks @eugene-manuilov – overall this is looking good and close to ready. Just a few questions and suggestions from me.

tests/e2e/utils/step-and-screenshot.js
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * @since n.e.x.t
+ *
+ * @param {string}           name Step name.
+ * @param {Function|Promise} cb   Step callback function or a promise object.
+ * @return {Promise} Promise object.
+ */
+export function step( name, cb ) {
+	return new Promise( async ( resolve, reject ) => {
+		try {
```
aaemnnosttv, line 69: Regarding the lint errors, let's import `path from path` rather than using named imports for those functions to avoid this conflict.

eugene-manuilov: Yep, already done it on my end.

tests/e2e/utils/step-and-screenshot.js
eugene-manuilov: Updated.

includes/Core/Util/User_Input_Settings.php
eugene-manuilov: Had to change it because we don't use real OAuth client during e2e tests, thus it doesn't return an access token and causes unexpected issues.

> The access token is required for making this request, is it not?

Yes, it is required by the proxy server itself. The user input settings class just passes what it has.

aaemnnosttv: Hmm, okay. I suppose this is fine then since we weren't guarding against making the request if there was no token before.

aaemnnosttv (approved):
LGTM 👍  Nice touch with the added screenshot artifacts!

Issue #2623, opened 2021-01-14
Change font styles in All Traffic V2 `TotalUserCount` to match those of our regular `DataBlock`s


felixarntz:
#2423 introduced the `TotalUserCount` component. Unfortunately, the Figma file design linked to there was outdated, so we added some custom styles there that should not be applied.

[This is the correct Figma design](https://www.figma.com/file/NzvJTPFdqUVJC9Q1mmih4e/Google-SK-UX-Enhancements10up?node-id=1284%3A175) - to be precise, the font style/size/color etc. for every part of the `TotalUserCount` UI should entirely match those that are used in our regular `DataBlock` components.

This will also result in the new widget integrating more nicely with the existing widgets where otherwise it would have felt a bit too foreign.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The "Users" label, the total count, and the percentage change should all match in text style/size/color the respective similar text that is used in `DataBlock` components (e.g. in the "Search Funnel" widget area below).
* Custom styles added earlier to make the current version of the text appear like it does should be removed.

## Implementation Brief

* Update styles for the `TotalUserCount` (found in [assets/sass/widgets/_googlesitekit-widget-analyticsAllTrafficV2.scss](https://github.com/google/site-kit-wp/blob/7ee258d11ee6c429be9e2e931fedfed5d92dd2f4/assets/sass/widgets/_googlesitekit-widget-analyticsAllTrafficV2.scss)) component to match the design in the Figma file provided in the description.
* Remove the previous styles that are no longer needed for the total user count component.

### Test Coverage

* N/A

### Visual Regression Changes

* The total user count section of the new all traffic widget should look like on the new Figma file.

## QA Brief

* Set up the plugin;
* Activate and connect the Analytics module;
* Go to the dashboard page and make sure that the new all traffic widget has appropriate styles for the total user count block and that the fonts match the fonts in e.g. the Search Funnel widget below.

## Changelog entry

* Modify font styles in the new All Traffic widget to match font styles used in Search Funnel widget and elsewhere throughout the plugin.

tofumatt:
IB ✅

wpdarren:
### QA Update: Pass ✅ 

- Verified that the font style matches that of the values in the Search Funnel widget.  Compared CSS and font family. Also checked on mobile devices to make sure that the value was readable on all screen sizes. 

[...]

PR #2645, opened 2021-01-19
Change font styles in All Traffic V2 'TotalUserCount'.

johnPhillips:
## Summary

Addresses issue #2623

## Checklist

### DIFF
--- a/assets/sass/widgets/_googlesitekit-widget-analyticsAllTrafficV2.scss
+++ b/assets/sass/widgets/_googlesitekit-widget-analyticsAllTrafficV2.scss
@@ -88,53 +88,15 @@
 	.googlesitekit-widget--analyticsAllTrafficV2__totalcount {
 
 		h3 {
-			color: $c-boulder;
-			font-size: 16px;
-			font-weight: 400;
-			margin: 0;
 
 			span::before {
 				content: ">";
 				margin: 0 0.5rem;
 			}
 		}
-
-		h2 {
-			font-size: 58px;
-			font-weight: 400;
-			margin: 0;
-		}
 	}
 
 	.googlesitekit-widget--analyticsAllTrafficV2__totalcount--loading {
 		margin: 0 0 10px;
-	}
-
-	.googlesitekit-widget--analyticsAllTrafficV2__totalcount--change {
-		display: inline-block;
-		font-size: 12px;
-		font-weight: 400;
-		margin: 0;
-		margin-right: 4px;
-
-		svg {
-			margin: 0;
-			margin-right: 5px;
-		}
-	}
-
-	.googlesitekit-widget--analyticsAllTrafficV2__totalcount--up-change {
-		color: $c-chateau-green;
-	}
-
-	.googlesitekit-widget--analyticsAllTrafficV2__totalcount--down-change {
-		color: $c-thunderbird;
-	}
-
-	.googlesitekit-widget--analyticsAllTrafficV2__totalcount--daterange {
-		display: inline-block;
-		font-size: 12px;
-		font-weight: 400;
-		margin: 0;
 	}
 }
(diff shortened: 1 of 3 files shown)


eugene-manuilov (approved):
LGTM!

assets/sass/widgets/_googlesitekit-widget-analyticsAllTrafficV2.scss
```diff
@@ -88,8 +88,9 @@
 	.googlesitekit-widget--analyticsAllTrafficV2__totalcount {
```
felixarntz, line 88: Now that this issue has clarified the requirement of matching the style for these numbers and labels to be similar to our `DataBlock` styles, I think it's rather counter-productive to redeclare all of those styles here, since this results in duplicate code, and furthermore the styles used here don't _exactly_ match because e.g. responsive behavior is slightly different.

I think it would be better to reuse `DataBlock` CSS classes in the `TotalUserCount` component, since pretty much all pieces in there are similar. Only a single design tweak here requires extra CSS, which is the percentage value should also be in red/green according to the change (contrary to using the regular gray).

Reusing the classes clarifies the relation here, and it ensures even in the future these numbers and labels will be formatted consistently.

assets/sass/widgets/_googlesitekit-widget-analyticsAllTrafficV2.scss
```diff
@@ -99,8 +100,8 @@
 		}
 
 		h2 {
```
felixarntz, line 102: This was already a problem before this PR, but I just noticed it: We should not use a `h2` in the component here as it results in incorrectly nested heading levels. Let's use a `div` like we do in the regular `DataBlock`s too.

assets/sass/widgets/_googlesitekit-widget-analyticsAllTrafficV2.scss
```diff
@@ -111,10 +112,10 @@
 
 	.googlesitekit-widget--analyticsAllTrafficV2__totalcount--change {
```
felixarntz, line 113: In addition to CSS changes, the component also needs to be modified for the change arrow, since it looks way bigger than the font now, which is different from the Figma design. The SVG needs to use a size of 9x9 pixels, the same dimensions that the `DataBlock`s use.

felixarntz (changes requested):
There's one piece that visually doesn't match the Figma design (the change arrow), however I think we can address the design disparities in a better way: The goal here is to have them more consistent with what the `DataBlock`s look like, so relying on the same classes makes more sense than redeclaring similar styles.

felixarntz:
@johnPhillips @eugene-manuilov I've made changes according to the feedback above (relying on `DataBlock` styles instead of redeclaring them) in a separate branch built on top of this one: https://github.com/google/site-kit-wp/compare/enhancement/2623-rely-on-datablock-updates?compare=1

Comparison screenshots:

**Current PR branch:** (style looks different from other numbers in Search Funnel below)
Screen Shot 2021-01-20 at 2 39 33 PM

**Extra branch:** (style looks similar to other numbers in Search Funnel below)
Screen Shot 2021-01-20 at 2 39 45 PM

The visual differences are minor, especially in the screenshots they're barely visible, but when seeing them next to the Search Funnel widgets, it becomes obvious.

johnPhillips:
@felixarntz Thanks for the review and feedback! I agree with what you are suggesting, makes total sense 👍
Just to be clear what it is you would like me to do now though: should I merge this branch of yours into my branch?
Let me know and I'll get it done.

felixarntz:
@johnPhillips If you agree, of course! I was opening a second branch just to leave this open for discussion, should there have been concerns about the approach.

johnPhillips:
@felixarntz Yes, I think it makes total sense. I have integrated your suggestions - hopefully I got all of them - and it looks good locally 👌
Would you mind having another look through when you have a moment? I'll assign the ticket back to you.

felixarntz (approved):
Great!

Issue #2624, opened 2021-01-14
Fix `UserCountGraph` to always show axis labels


felixarntz:
On smaller viewports, the graph from the `UserCountGraph` component in the new All Traffic widget has its labels replaced with dots, which should not happen. For higher user counts like e.g. 1000+ this trimming already starts to happen even on small desktop sizes.

We should ensure these numbers are always fully present. The graph itself should be made more narrow as needed, but the axis labels should not go away.

Screenshot:
![Screen Shot 2021-01-14 at 11 09 18 AM](https://user-images.githubusercontent.com/3531426/104636953-fc9a1800-5658-11eb-8b5d-b7f013f16316.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The labels on the `UserCountGraph` (particularly the y-axis) should not be trimmed on smaller viewports.

## Implementation Brief

* Update the `chartArea.width` option to be `80%` for the `UserCountGraph` component.

### Test Coverage

* Not required

### Visual Regression Changes

* Axis labels should be displayed correctly on smaller screens.

## QA Brief

* Set up the plugin;
* Activate and connect the Analytics module;
* Go to the dashboard page and make sure that the user count graph shows axis labels correctly;
* Resize your browser window to much smaller sizes and make sure that axis labels are still visible and you can read it.

## Changelog entry

* Fix graph in All Traffic widget to not cut off axis labels on viewports smaller than desktop.

aaemnnosttv:
@eugene-manuilov While I think your suggested approach would work, are we sure that we can't solve it with `vAxes` configuration? Looking at the [configuration options](https://developers.google.com/chart/interactive/docs/gallery/areachart#configuration-options) for the chart, I think the solution might instead involve changing `vAxes.viewWindow` and/or `vAxes.viewWindowMode`.

I'm curious if removing both of these, or using alternate values would solve the problem?

eugene-manuilov:
> I'm curious if removing both of these, or using alternate values would solve the problem?

@aaemnnosttv I have checked it on my local but it doesn't help.

aaemnnosttv:
Thanks for checking @eugene-manuilov 👍 

IB ✅

wpdarren:
### QA Update: Pass ✅ 

[...]

PR #2661, opened 2021-01-20
Fix bug with UserCountGraph where axis labels would become ellipses.

benbowler:
## Summary

Addresses issue #2624

## Checklist


eugene-manuilov (approved):
LGTM!

felixarntz (approved):
That works great mostly!

On small screens there's still a chance for cutting off numbers (see below), but I think on those viewports the widget content looks a bit squished anyway - we shouldn't try to fit it into two columns on such narrow viewports. I'll open a follow-up issue to adjust it so that on phones both grid columns wrap into their own rows.

Issue #2539, opened 2020-12-17
Extract Bundlesize to its own job/check


aaemnnosttv:
## Feature Description

In #2366 we migrated our JS tests to GitHub Actions, which created a new workflow for running our Jest tests as well as our Bundlesize tests.

These currently run in the same job, making both slower to run, but also Bundlesize has some integration with GitHub checks which makes the results of the run more visible, rather than simply pass or fail.

[![](https://camo.githubusercontent.com/172d76ab2e6658114b0eecc25aeaf3c2b38c5ae3660c456a1b8147bcb5191aa6/68747470733a2f2f63646e2e7261776769742e636f6d2f7369646468617274686b702f62756e646c6573697a652f6d61737465722f6172742f7374617475732e706e67)](https://github.com/siddharthkp/bundlesize#build-status-for-github)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Bundlesize should be reported in each Pull Request via GitHub Actions checks.

## Implementation Brief

Because this process needs to be done by someone with admin access to the repo and the ability to authorise Bundlesize to access the repo, it's best done by @felixarntz.

### Work completed by @felixarntz, required to allow the check to appear

* Visit https://github.com/login/oauth/authorize?scope=repo%3Astatus&client_id=6756cb03a8d6528aca5a to get a Bundlesize GitHub token
* Allow Bundlesize access to Google repos. This is needed to add the check to each PR we run `bundlesize` on.
* Add the token obtained in the above step to the `google/site-kit-wp` repo as a GitHub secret named `BUNDLESIZE_GITHUB_TOKEN`.

### Work to be done
* Add the following environment variables to the `npm run test:bundlesize` command in the ["JS Tests" GitHub Action workflow](https://github.com/google/site-kit-wp/blob/develop/.github/workflows/js-tests.yml#L42):
  * `CI_REPO_OWNER='google'`
  * `CI_REPO_NAME='site-kit-wp'`
  * `CI_COMMIT_MESSAGE={The git commit message for this commit, should be available as a GitHub Actions ENV var}`
  * `CI_COMMIT_SHA={The git commit SHA for this commit, should be available as a GitHub Actions ENV var}`
  * `CI=true` (if not already set by GitHub Actions)
* See:  https://github.com/siddharthkp/bundlesize#using-a-different-ci

[...]

PR #2667, opened 2021-01-21
Extract Bundlesize to its own job/check

benbowler:
Extract Bundlesize to its own job/check.

## Summary

Addresses issue #2539

## Relevant technical choices

I followed [these documents](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context) to get the head commit message and SHA.

## Checklist


benbowler:
If the context is incorrect we can check is using this action step:

```
      - name : GITHUB CONTEXT
        env:
          GITHUB_CONTEXT: ${{ toJson(github) }}
        run: echo "$GITHUB_CONTEXT"
```

aaemnnosttv:
The request to report the check is failing it seems:
https://github.com/google/site-kit-wp/runs/1761296111?check_suite_focus=true#step:9:16

Not sure what's going on exactly but it's resulting in a 500 Internal server error.

benbowler:
Hey @aaemnnosttv, I have done some rearch, including following [this thread](https://github.com/siddharthkp/bundlesize/issues/308). As we don't have control over the Vercel (now.sh) set up I wasn't able to dig into why this is returning a 500 error. 

There is a more actively maintained alternative called [bundlewatch](https://github.com/bundlewatch/bundlewatch) with more active commits and a similar web service to https://bundlesize-store.now.sh.

Would you consider switching?

aaemnnosttv:
Thanks @benbowler – it looks like this is working which is great! However one thing I noticed as a feature on the `compressed-size-action` which is something we'll definitely want to use is the `strip-hash` option. This is really important for our use since the hash will change any time the resulting file changes which would undermine the whole process of comparing the size from one version to another since without this it would always see them as separate files.

This should be fairly straight forward to add, again because the action offers support for this, we just need to configure it to match our hashes so they can be stripped.

See https://github.com/preactjs/compressed-size-action#dealing-with-hashed-filenames

felixarntz (approved):
Great!

Issue #2613, opened 2021-01-13
Pass scoped `Widget` component to registered widget component function


felixarntz:
Widgets that pass `wrapWidget: false` need to render the wrapping `Widget` component themselves. For this purpose, we right now export the `Widget` component publicly on `googlesitekit.widgets.components`.

Somewhat annoying here, components that render `Widget` on its own currently are required to pass the widget's slug to the component - duplicating that from where the widget is registered, making it error-prone.

We should change this so that we pass a scoped version of `Widget` to every registered component so that they don't need to import it and don't need to pass the `slug` prop. This follows the pattern established in #2252 where components already get access to special widget state components via props, such as `WidgetReportZero`.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `slug` prop of `Widget` should be renamed to a more explicit `widgetSlug`, matching what other publicly usable `Widget*` components are doing.
* The `getWidgetComponentProps` function should be expanded to include a scoped version of `Widget` in the props object. (This will have `WidgetRenderer` pass it through to every registered widget component.)
* The `googlesitekit.widgets.components` property should be removed (i.e. `Widgets.components` internally).
* Every Site Kit widget that currently renders `<Widget />` manually in its component should rely on the `Widget` it receives from props instead of importing it. The `slug` / `widgetSlug` must no longer be passed since it's already taken care of.

## Implementation Brief

* Rename the `slug` property to be `widgetSlug` in the `Widget` component. Update all occurrences of the `Widget` component to use the new property.
* Add a new property to the returning object from the `getWidgetComponentProps` function. It should have the `Widget` name and scoped the `Widget` component as a value (similarly to other props in the returned object).
* Edit the following components to rely on the `Widget` property passed to the component instead of `Widgets.components.Widget` one. Plus remove slug properties from the rendered `<Widget>` components since they are not needed anymore.
  * The `DashboardSummaryWidget` component in the AdSense module;
  * The `DashboardTopEarningPagesWidget` component in the AdSense module;
  * The `DashboardAllTrafficWidget` component in the Analytics module;
  * The `DashboardPopularPagesWidget` component in the Analytics module;
  * The `DashboardPageSpeedWidget` component in the Page Speed module;
  * The `DashboardPopularKeywordsWidget` component in the Search Console module;
* Remove the `components` property of the `Widgets` object in the `assets/js/googlesitekit/widgets/index.js` file.
[...]

PR #2676, opened 2021-01-22
Pass scoped widget component

benbowler:
## Summary

Addresses issue #2613

## Checklist


benbowler:
There is an issue with this code:

https://github.com/google/site-kit-wp/blob/717946a1eebcf756bdef672bcd755f1fd3c90479/assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js#L58-L61

Causing a missing prop type error:

```
checkPropTypes.js:20 Warning: Failed prop type: The prop `widgetSlug` is marked as required in `Widget`, but its value is `undefined`.
```

I spent a little time seeing if there was a quick fix but, as this is marked with a TODO: to "Solve this in a better way." I left it for now. Is there an existing ticket to update this function or should I create one?

eugene-manuilov (changes requested):
Thanks, @benbowler! It's a good start, however, there are still a few issues that we need to address before approving this PR. Here are else things we need to do:

1. Fix `slug` properties in the `stories/widgets.stories.js` file;
2. Fix this storybook: https://google.github.io/site-kit-wp/storybook/pull/2676/?path=/story/pagespeed-insights-module-components--dashboard-widget

assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js
```diff
@@ -55,6 +55,7 @@ export default function WidgetAreaRenderer( { slug, totalAreas } ) {
 			return false;
 		}
 		const widgetComponentProps = getWidgetComponentProps( widget.slug );
+		// NOTE: this causes a "Failed prop type: The prop `widgetSlug` is marked as required in `Widget`" error.
```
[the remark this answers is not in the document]
benbowler: Updated, thanks.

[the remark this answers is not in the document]
benbowler: I actually tracked this down to the `URLSearchWidget` still passing `slug` rather than `widgetSlug`.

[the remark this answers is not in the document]
eugene-manuilov: Ok, cool. Let's get rid of this note then:

```suggestion
```

stories/utils/generate-widget-stories.js
```diff
@@ -24,16 +24,18 @@ import { storiesOf, Story } from '@storybook/react';
 /**
  * Internal dependencies
  */
-import Widgets from 'googlesitekit-widgets';
+// import Widgets from 'googlesitekit-widgets'; TODO:
```
eugene-manuilov, line 27: Is it still needed or we can get rid of it?

stories/utils/generate-widget-stories.js
```diff
@@ -24,16 +24,18 @@ import { storiesOf, Story } from '@storybook/react';
...
 	createTestRegistry,
 	WithTestRegistry,
 	provideModules,
 	provideSiteInfo,
 } from '../../tests/js/utils';
+import Widget from '../../assets/js/googlesitekit/widgets/components/Widget';
+import { getWidgetComponentProps } from '../../assets/js/googlesitekit/widgets/util';
 
-const { components: { Widget } } = Widgets;
+// const { components: { Widget } } = Widgets; TODO:
```
eugene-manuilov, line 38: Same here, do we still need this or it can be removed?

stories/utils/generate-widget-stories.js
```diff
@@ -210,24 +212,27 @@ export function generateReportBasedWidgetStories( {
 		...customVariants,
 	};
 
-	let widget;
+	const slugs = moduleSlugs.map( ( slug ) => {
+		return `${ slug }-widget`;
+	} );
+	const slug = slugs.join( ' ' );
```
eugene-manuilov, line 218: Let's use shorter version:

```suggestion
	const slug = moduleSlugs.map( ( slug ) => `${ slug }-widget` ).join( ' ' );
```

stories/utils/generate-widget-stories.js
```diff
@@ -210,24 +212,27 @@ export function generateReportBasedWidgetStories( {
...
+
+	const widgetComponentProps = getWidgetComponentProps( slug );
+
+	let widgetElement = <Component { ...widgetComponentProps } />;
```
eugene-manuilov, line 222: Let's leave it in the `else` statement where it was before:

```javascript
let widgetElement;

if ( wrapWidget ) {
    widgetElement = ( ... );
} else {
    const slug = moduleSlugs.map( ( slug ) => `${ slug }-widget` ).join( ' ' );
    const widgetComponentProps = getWidgetComponentProps( slug );
    widgetElement = <Component { ...widgetComponentProps } />;
}
```

eugene-manuilov (changes requested):
Thanks, @benbowler! I have added a few minor comments and it should be ready to go. Please, take a look at it.

stories/utils/generate-widget-stories.js
benbowler: We don't need it, it was a note from a merge conflict I've now fixed.

stories/utils/generate-widget-stories.js
benbowler: Switched back in.

benbowler:
I've addressed your feedback and the merge conflicts.

eugene-manuilov (changes requested):
@benbowler everything looks good except the LegacyDashboardAllTraffic component. It renders DashboardAllTrafficWidget without props which causes errors and broken e2e tests. We need to manually add props generated using `getWidgetComponentProps` function:

```javascript
const props = getWidgetComponentProps( 'legacy-all-traffic-widget' );
return (
    ...
    <DashboardAllTrafficWidget { ...props } />
    ...
);
```

https://github.com/google/site-kit-wp/blob/9da5402e860b9333022b2082dfd74cbec0709f22/assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/LegacyDashboardAllTraffic.js#L44

benbowler:
Thanks Eugene, I've addressed the E2E testing issue.

eugene-manuilov (approved):
Awesome! Thanks, @benbowler!

stories/utils/generate-widget-stories.js
```diff
@@ -210,24 +208,20 @@ export function generateReportBasedWidgetStories( {
...
 	if ( wrapWidget ) {
-		const slugs = moduleSlugs.map( ( slug ) => {
-			return `${ slug }-widget`;
-		} );
-		widget = (
-			<Widget slug={ slugs.join( ' ' ) }>
-				<Component />
-			</Widget>
-		);
+		widgetElement = <Component />;
```
eugene-manuilov, line 214: @benbowler storybook is broken again. See data unavailable states of widgets: 
* https://google.github.io/site-kit-wp/storybook/pull/2676/?path=/story/analytics-module-components-dashboard-bounce-rate-widget--data-unavailable
* https://google.github.io/site-kit-wp/storybook/pull/2676/?path=/story/analytics-module-components-dashboard-goals-widget--data-unavailable

I think it happens because this `<Component />` should be wrapped with a `Widget` element and receive additional props too:

```javascript
let widgetElement;

const slug = moduleSlugs.map( ( mapSlug ) => `${ mapSlug }-widget` ).join( ' ' );
const widgetComponentProps = getWidgetComponentProps( slug );

if ( wrapWidget ) {
    const { Widget } = widgetComponentProps;
    widgetElement = (
        <Widget>
            <Component />
        </Widget>
    );
} else {
    widgetElement = <Component { ...widgetComponentProps } />;
}
```

eugene-manuilov (changes requested):

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/LegacyDashboardAllTraffic.js
```diff
@@ -29,8 +29,10 @@ import Layout from '../../../../../components/layout/Layout';
 import DashboardModuleHeader from '../../../../../components/dashboard/DashboardModuleHeader';
 import DashboardAllTrafficWidget from '.';
 import { Cell } from '../../../../../material-components';
+import { getWidgetComponentProps } from '../../../../../googlesitekit/widgets/util';
 
 function LegacyDashboardAllTraffic() {
+	const widgetComponentProps = getWidgetComponentProps( 'legacy-all-traffic-widget' );
```
felixarntz, line 35: Nice!

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/index.js
```diff
@@ -42,10 +41,9 @@ import UserCountGraph from './UserCountGraph';
 import DimensionTabs from './DimensionTabs';
 import UserDimensionsPieChart from './UserDimensionsPieChart';
 import { isZeroReport } from '../../../util';
-const { Widget } = Widgets.components;
 const { useSelect } = Data;
 
-function DashboardAllTrafficWidget() {
+function DashboardAllTrafficWidget( { Widget } ) {
```
felixarntz, line 46: Now that this component also receives the widget component props (in both the new and legacy version, let's also address the TODO comment in line 161/162 and rely on the passed `WidgetReportZero` instead of the imported `ReportZero`.

assets/js/googlesitekit/widgets/components/URLSearchWidget.js
```diff
@@ -49,7 +49,7 @@ function URLSearchWidget() {
 	return (
 		<div className="mdc-layout-grid__cell">
 			<Widget
-				slug="urlSearchWidget"
+				widgetSlug="urlSearchWidget"
```
felixarntz, line 52: Here we should also be able to use `props.Widget` instead of the imported component (this way we also don't need to pass the slug anymore).

stories/utils/generate-widget-stories.js
felixarntz: Agreed - we must not remove the `Widget` wrapper here: The widget is registered with `wrapWidget: true`, so the `Widget` needs to be present _around_ the registered `Component`.

One thing that would need to be fixed in @eugene-manuilov's code snippet above: We would still need to pass `...widgetComponentProps` to the `Component` either way, since that's expected. Even if the component doesn't use `props.Widget`, it might use other components on there, e.g. `props.WidgetReportZero`.

eugene-manuilov: Ah... yes, thanks! I mentioned it in the comment, but forgot to add it to the snipped... :smile:

benbowler: Ok, thanks for updating the comment, I've applied that and will push shortly.

felixarntz (changes requested):
Just one thing to fix here, otherwise this is good to go.

felixarntz (approved):
Updated this, should be good to go now - great work!

Issue #2668, opened 2021-01-21
Tag Manager Snippet is output twice in AMP URLs


jamesozzie:
As reported in the [WordPress support forums](https://wordpress.org/support/topic/tag-manager-code-snippet-inserted-twice/#post-13938356) the Tag Manager snippet is inserted twice in AMP URLs. 

Example duplicate snippet:

```html
<!-- Google Tag Manager added by Site Kit -->
<amp-analytics config="https://www.googletagmanager.com/amp.json?id=GTM-MH4PNSJ" data-credentials="include" class="i-amphtml-layout-fixed i-amphtml-layout-size-defined" style="width:1px;height:1px;" i-amphtml-layout="fixed"><script type="application/json">{"optoutElementId":"__gaOptOutExtension"}</script></amp-analytics>
<!-- End Google Tag Manager -->
```

This only started occurring since 1.24.0. Reproducible in support. 

## Steps to reproduce

1. Install the AMP plugin and set the mode to "Transitional" or "Standard"
2. Install Site Kit version 1.24.0 and setup the Google Tag Manager module
3. Access your site and view the snippet place via Site Kit (snippet is duplicated)

## Screenshots

![image](https://user-images.githubusercontent.com/41326532/105378016-4a87c080-5c03-11eb-8d59-f324f59753fb.png)

## Additional Context

 - Only occurs with 1.24.0
 - Reproducible in support, two sites tested
 - Snippets evident by viewing https://www.pribrat.cz/
 - GIF created showing example (happy to share if required)
 - Occurs with only the SC & Tag Manager modules active, no need for additional modules to be connected

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The Tag Manager snippet should only be inserted once in AMP pages.
* It should be verified that the snippets for Analytics and AdSense are only inserted once (since potentially there could be a similar regression there).

## Implementation Brief

* Merge #2678

### Test Coverage

* Add phpunit tests to test the `Method_Proxy_Trait` trait.

### Visual Regression Changes

* Tags should be inserted only once after the fix for this ticket is implemented.

## QA Brief

* Install and activate the AMP plugin;
* Set up the plugin on your local;
* Activate and connect Analytics, AdSense, and Tag Manager modules;
* Make sure these modules insert their tags only once both on web and amp pages.

## Changelog entry

* Fix bug where Tag Manager and Analytics snippet could be inserted twice on AMP pages.

[...]

PR #2678, opened 2021-01-22
Bug/2668 duplicate tags

eugene-manuilov:
## Summary

Addresses issue #2668

## Checklist

### DIFF
--- a/includes/Core/Util/Method_Proxy_Trait.php
+++ b/includes/Core/Util/Method_Proxy_Trait.php
@@ -35,15 +35,16 @@ trait Method_Proxy_Trait {
 	 * @return callable A proxy function.
 	 */
 	private function get_method_proxy_once( $method ) {
-		static $calls = array();
+		return function ( ...$args ) use ( $method ) {
+			static $called;
+			static $return_value;
 
-		return function ( ...$args ) use ( $method, &$calls ) {
-			$key = get_class( $this ) . '::' . spl_object_hash( $this ) . '::' . $method;
-			if ( ! array_key_exists( $key, $calls ) ) {
-				$calls[ $key ] = $this->{ $method }( ...$args );
+			if ( ! $called ) {
+				$called       = true;
+				$return_value = $this->{ $method }( ...$args );
 			}
 
-			return $calls[ $key ];
+			return $return_value;
 		};
 	}
 
--- a/includes/Core/Util/Method_Proxy_Trait.php
+++ b/includes/Core/Util/Method_Proxy_Trait.php
@@ -38,7 +38,7 @@ trait Method_Proxy_Trait {
 		static $calls = array();
 
 		return function ( ...$args ) use ( $method, &$calls ) {
-			$key = get_class( $this ) . '::' . $method;
+			$key = get_class( $this ) . '::' . spl_object_hash( $this ) . '::' . $method;
 			if ( ! array_key_exists( $key, $calls ) ) {
 				$calls[ $key ] = $this->{ $method }( ...$args );
 			}
--- a/includes/Core/Util/Method_Proxy_Trait.php
+++ b/includes/Core/Util/Method_Proxy_Trait.php
@@ -37,7 +37,7 @@ trait Method_Proxy_Trait {
 	private function get_method_proxy_once( $method ) {
 		static $calls = array();
 
-		return function ( ...$args ) use ( $method, $calls ) {
+		return function ( ...$args ) use ( $method, &$calls ) {
 			$key = get_class( $this ) . '::' . $method;
 			if ( ! array_key_exists( $key, $calls ) ) {
 				$calls[ $key ] = $this->{ $method }( ...$args );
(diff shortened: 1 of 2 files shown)


tests/phpunit/integration/Core/Util/Method_Proxy_TraitTest.php
```diff
@@ -0,0 +1,51 @@
...
+
+		call_user_func( $callback );
+		call_user_func( $callback );
+
+		$this->assertEquals( $this->calls, 1 );
+	}
+
+	private function hidden_function() {
+		$this->calls++;
+	}
```
aaemnnosttv, line 49: You could use `MethodSpy` or use that as the base for a mock instance to use here instead of reimplementing that.

aaemnnosttv (commented):
This is still in IB but looks to me that there is a simpler way to solve the problem.

includes/Core/Util/Method_Proxy_Trait.php
```diff
@@ -37,8 +37,8 @@ private function get_method_proxy( $method ) {
 	private function get_method_proxy_once( $method ) {
 		static $calls = array();
 
-		return function ( ...$args ) use ( $method, $calls ) {
-			$key = get_class( $this ) . '::' . $method;
+		return function ( ...$args ) use ( $method, &$calls ) {
+			$key = get_class( $this ) . '::' . spl_object_hash( $this ) . '::' . $method;
```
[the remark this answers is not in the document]
eugene-manuilov: Yep, nice idea. Updated.

tests/phpunit/integration/Core/Util/Method_Proxy_TraitTest.php
eugene-manuilov: I haven't reimplemented it, this case is different and I think don't think it is going to work with MethodSpy or a mock instance due to the nature of traits.

tests/phpunit/integration/Core/Util/Method_Proxy_TraitTest.php
```diff
@@ -0,0 +1,51 @@
...
+	private $calls;
+
+	public function test_method_proxy() {
+		$this->calls = 0;
+		$callback    = $this->get_method_proxy( 'hidden_function' );
+
+		call_user_func( $callback );
+		call_user_func( $callback );
+
+		$this->assertEquals( $this->calls, 2 );
```
aaemnnosttv, line 34: The order of arguments is backwards here. It should be `expected, actual`.
```suggestion
		$this->assertEquals( 2, $this->calls );
```

tests/phpunit/integration/Core/Util/Method_Proxy_TraitTest.php
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	public function test_method_proxy_once() {
+		$this->calls = 0;
+		$callback    = $this->get_method_proxy_once( 'hidden_function' );
+
+		call_user_func( $callback );
+		call_user_func( $callback );
+
+		$this->assertEquals( $this->calls, 1 );
```
aaemnnosttv, line 44:
```suggestion
		$this->assertEquals( 1, $this->calls );
```

tests/phpunit/integration/Core/Util/Method_Proxy_TraitTest.php
aaemnnosttv: I think it's possible but it looks more complicated than it's worth here for sure 👍

tests/phpunit/integration/Core/Util/Method_Proxy_TraitTest.php
```diff
@@ -0,0 +1,51 @@
...
+	use Method_Proxy_Trait;
+
+	private $calls;
+
+	public function test_method_proxy() {
+		$this->calls = 0;
+		$callback    = $this->get_method_proxy( 'hidden_function' );
+
+		call_user_func( $callback );
+		call_user_func( $callback );
```
aaemnnosttv, line 32: While we're at it, let's simplify the calls, just like we do in the plugin.
```suggestion
		$callback();
		$callback();
```

tests/phpunit/integration/Core/Util/Method_Proxy_TraitTest.php
```diff
@@ -0,0 +1,51 @@
...
+
+		$this->assertEquals( $this->calls, 2 );
+	}
+
+	public function test_method_proxy_once() {
+		$this->calls = 0;
+		$callback    = $this->get_method_proxy_once( 'hidden_function' );
+
+		call_user_func( $callback );
+		call_user_func( $callback );
```
aaemnnosttv, line 42:
```suggestion
		$callback();
		$callback();
```

aaemnnosttv (commented):
Looks good to me, just a few tweaks (mostly the argument swapping needed for assertions). I've tested that this does fix the problem though 👍

aaemnnosttv (approved):
Great, thanks!

felixarntz (approved):
Great!

Issue #2610, opened 2021-01-11
Migrate WP dashboard widget's "Top content over the last x days" table to functional component using datastore


felixarntz:
#2075 focused on replicating the WP dashboard widget's UI via new functional components using the datastore. One piece that was missed during ACs and IB review was the "Top content over the last x days" table that appears at the bottom when Analytics is active.

Since the new UI is currently only shown in a development build, that is okay for now, but we need to also migrate that table before we can launch the new WP dashboard widget in production. Let's do that.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a new functional hook-based `WPDashboardPopularPages` component that relies on the datastore and renders the exact same UI as its legacy version.
    * It should not use any legacy functions, i.e. it should create its request object individually and extract data from its response manually, similar to closely related refactored components, such as the Analytics Top Pages table in the dashboard (a lot of the code we'll be able to copy from there).
* The component should be integrated into the new WP dashboard widget component infrastructure (behind `widgets.wpDashboard` feature flag) with similar placement as it has in the legacy WP dashboard widget.

## Implementation Brief

* Create a new file to export a functional component `WPDashboardPopularPages` in `/assets/js/components/wp-dashboard/WPDashboardPopularPages.js`
* For rendering of the widget, its content will be based on: `/assets/js/modules/analytics/components/wp-dashboard/WPAnalyticsDashboardWidgetTopPagesTable.js`
https://github.com/google/site-kit-wp/blob/7ee258d11ee6c429be9e2e931fedfed5d92dd2f4/assets/js/modules/analytics/components/wp-dashboard/WPAnalyticsDashboardWidgetTopPagesTable.js
* For functionality of the widget, see these components:
* https://github.com/google/site-kit-wp/blob/7ee258d11ee6c429be9e2e931fedfed5d92dd2f4/assets/js/modules/analytics/components/dashboard/AnalyticsDashboardWidget.js#L1
* https://github.com/google/site-kit-wp/blob/7ee258d11ee6c429be9e2e931fedfed5d92dd2f4/assets/js/modules/analytics/components/dashboard/AnalyticsDashboardWidgetTopPagesTable.js#L1
[...]

PR #2682, opened 2021-01-25
Migrate dashboard 'Top content over last x days' widget.

johnPhillips:
## Summary

Addresses issue #2610

## Relevant technical choices
Tried to follow the IB as closely as possible.

## Checklist


assets/js/components/wp-dashboard/WPDashboardPopularPages.js
```diff
@@ -0,0 +1,119 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+/**
```
eugene-manuilov, line 23:
```suggestion
import { __ } from '@wordpress/i18n';

/**
```

eugene-manuilov (changes requested):
Thanks, @johnPhillips! Everything looks good to me, however, let's return a `ReportZero` element if the resulting report is empty:

```js
if ( isZeroReport( report ) ) {
    return <ReportZero moduleSlug="analytics" />;
}
```

johnPhillips:
Thanks @eugene-manuilov! I've incorporated those changes, let me know what you think.

eugene-manuilov (dismissed):

eugene-manuilov:
Thanks, @johnPhillips! All looks good now.

assets/js/components/wp-dashboard/WPDashboardPopularPages.js
```diff
@@ -0,0 +1,126 @@
...
+			numFmt( row.metrics[ 0 ].values[ 0 ], { style: 'decimal' } ),
+		];
+	} );
+	const options = {
+		hideHeader: false,
+		chartsEnabled: false,
+		links,
+		hideColumns: {
+			mobile: [ 2, 3 ],
+		},
```
felixarntz, line 109: The `options` here are not correct, which causes the table to have a different appearance between the old and new (the new one is missing the URLs visually, and it's linking to the wrong location). See here:
Screen Shot 2021-02-02 at 11 44 47 AM

I think you took them from the wrong component - they need to come from `WPAnalyticsDashboardWidgetTopPagesTable`, which is the legacy version of this.

assets/js/components/wp-dashboard/WPDashboardApp.js
```diff
@@ -84,6 +85,7 @@ const WPDashboardApp = () => {
 							) }
 						</div>
 					) }
+					<WPDashboardPopularPages />
```
felixarntz, line 88: This must only be rendered if the Analytics Module is active and connected, similar to the other Analytics components above - otherwise it will cause breakage when the module is not active/not connected.

felixarntz (changes requested):
There are two problems that need to be addressed - other than that, the code looks solid.

assets/js/components/wp-dashboard/WPDashboardApp.js
johnPhillips: Ah, thank you. I did wonder, but I couldn't see anything specific in the issue about this. I'll add it in 👍

assets/js/components/wp-dashboard/WPDashboardPopularPages.js
johnPhillips: Good catch, thank you

assets/js/components/wp-dashboard/WPDashboardPopularPages.js
```diff
@@ -0,0 +1,127 @@
...
+		],
+		limit: 5,
+	};
+
+	const data = useSelect( ( select ) => select( MODULES_ANALYTICS ).getReport( reportArgs ) );
+	const error = useSelect( ( select ) => select( MODULES_ANALYTICS ).getErrorForSelector( 'getReport', [ reportArgs ] ) );
+	const loading = useSelect( ( select ) => ! select( MODULES_ANALYTICS ).hasFinishedResolution( 'getReport', [ reportArgs ] ) );
+
+	if ( loading ) {
+		return <PreviewBlock width="48%" height="92px" />;
```
felixarntz, line 71: This preview component is incorrect here, which makes the loading state of the widget look odd.

Right here you're using the same like the other dashboard widget components, which however overall have different distribution and only take half the horizontal widget width. The component used here should match what is used in the legacy component `WPAnalyticsDashboardWidgetTopPagesTabl`, where it's a `<PreviewTable rows={ 6 } />`.

felixarntz (changes requested):
Just one piece of feedback here that I missed before, which affects the loading state of the component. Otherwise this is good to go!

assets/js/components/wp-dashboard/WPDashboardPopularPages.js
johnPhillips: Ah, good catch, sorry about that. I didn't see the loading component passed into the HOC right at the bottom of the file, and I must've used an example from elsewhere 🤦  

I've addressed this. I had to add a line of CSS to get the `PreviewTable` to display correctly because its parent was `display: flex`, but now it's looking the same as the legacy version from what I can see.

felixarntz (approved):
Awesome!

Issue #2644, opened 2021-01-19
UX/UI Improvement for Pie Chart Feature on the New All Traffic Module


wpdarren:
## Bug Description

This is a UI improvement suggestion. For the New All Traffic module on Site Kit Dashboard.  On the pie chart, select a slice, and then change the date selection to another period, i.e. 90 days. The module reloads and the correct number of users is displayed for that slice, but it is no longer selected. It would be good to see the slice continued to be highlighted so the user knows the user stats are related to that slice on first look. 

## Steps to reproduce

1. Go to The New Analytics module on the Site Kit Dashboard
2. Click on a slice within the pie chart for any dimension.
3. The number of users changes.  
4. Change the date selection to `90 days`
5. The page reloads with the new stats for Users but slice is not selected.

## Screenshots

![image](https://user-images.githubusercontent.com/73545194/105056964-b78a4180-5aaf-11eb-9029-d53e191b91f1.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* When changing the date range for the current page (e.g. on the Site Kit dashboard), if a specific pie slice in the All Traffic widget is currently selected, the visual cue for the slice being selected (i.e. the thin border around that slice) should still be visible.
    * Currently, while the graph displayed for the pie slice remains the same as expected, the slice itself is no longer visually selected. That's what needs to be fixed here.

## Implementation Brief

* When a section of the pie chart is selected (https://github.com/google/site-kit-wp/blob/562423f7d36b9a61009d1425e0a51b42c139931e/assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/UserDimensionsPieChart.js#L80), store the selected value in the `FORM_ALL_TRAFFIC_WIDGET` as a value (eg. `selectedRow` or something similar). Equally when the selection is cleared, set the `selectedRow` to `undefined`)
* Select that value from the data store at the top of `UserDimensionsPieChart` using `const selectedRow = useSelect( ( select ) => select( CORE_FORMS ).getValue( FORM_ALL_TRAFFIC_WIDGET, 'selectedRow' ) );`
[...]

PR #2684, opened 2021-01-26
Keep row selected when changing dateRange.

tofumatt:
## Summary

Addresses issue #2644.

## Checklist

### DIFF
--- a/assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/UserDimensionsPieChart.js
+++ b/assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/UserDimensionsPieChart.js
@@ -25,7 +25,7 @@ import classnames from 'classnames';
 /**
  * WordPress dependencies
  */
-import { useCallback, useState } from '@wordpress/element';
+import { useCallback, useEffect, useState } from '@wordpress/element';
 import { __, _x, sprintf } from '@wordpress/i18n';
 
 /**
@@ -120,6 +120,12 @@ export default function UserDimensionsPieChart( { dimensionName, entityURL, sour
 			} );
 		}
 	}, [ dimensionName, selectedRow, setValues ] );
+
+	useEffect( () => {
+		if ( selectedRow ) {
+			setValues( FORM_ALL_TRAFFIC_WIDGET, { selectedRow: undefined } );
+		}
+	}, [ dimensionName ] );
 
 	if ( ! loaded ) {
 		return <PreviewBlock width="282px" height="282px" shape="circular" />;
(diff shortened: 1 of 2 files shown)


felixarntz (changes requested):
The existing code looks good, but there's one problem this introduces: We need to make sure that `selectedRow` is no longer set when changing the tab (i.e. the `dimensionName`). Right now, when I select a pie slice in e.g. "Channels" and then go over to "Locations", there will still be a slice initially selected (I guess the one with the same row index or something?). That should not be the case, the selection should only persist within the same dimension's pie chart.

We should reset `selectedRow` when changing tabs.

tofumatt:
Ah right, I missed that! I've added a reset on `dimensionName` change, should be good now 👍🏻

felixarntz (changes requested):
There still seems to be something off here, although I thought your code would fix it. When I run it, it only works sometimes.

Instead of relying on `useEffect`, wouldn't it be more reliable to unset it when switching the tab, together with the dimension? That would IMO also better clarify the intent here.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/UserDimensionsPieChart.js
```diff
@@ -84,6 +85,10 @@ export default function UserDimensionsPieChart( { dimensionName, entityURL, sour
 		const { chart, onSelect } = chartData || {};
 		const { slices } = UserDimensionsPieChart.chartOptions;
 
+		if ( selectedRow !== undefined ) {
+			chart.setSelection( [ { row: selectedRow } ] );
```
felixarntz, line 89: I think the underlying problem here might be that we're selecting the `row` which is not specific enough, and it also shouldn't be necessary. We already store when a pie slice is selected, by storing the `dimensionValue` which _is_ the pie slice.

Basically, _if_ there is a (non-empty) `dimensionValue`, we need to ensure that the `row` which has that value as label (i.e. `row[ 0 ]` if I'm not mistaken) is selected.

felixarntz (changes requested):
This is still not working as expected, it's unreliable and only appears to work on initial pageload. As soon as I do a couple more interactions (change date range, change tabs, back and forth etc) it doesn't keep the pie slice for the same dimension value selected like it should.

I left a comment below for what I think is the underlying problem here. I don't think we need to store a new piece of information actually, since `dimensionValue` already indicates which pie slice is/should be selected. Let's try to use that one.

felixarntz:
Changed this to be against `master` since a fix for this issue should preferably go into the release.

felixarntz:
Closing since this will be covered by #2702.

Issue #2507, opened 2020-12-09
Remove Module datapoints for individual settings


aaemnnosttv:
## Feature Description

Currently we have a number of datapoints defined for modules that simply return a single value from the module's settings.

These are legacy datapoints which were used to both request individual settings from the client but also were used internally to retrieve these values in many places. Another purpose that these datapoints provided in the past was to abstract the key of the setting it was stored in. This is because the convention used for setting keys had changed or been renamed several times in the past.

We addressed much of this with the addition of `Module_Settings` classes quite a while ago, which handles things like default values, as well as legacy key migration automatically.

In addition to the abstraction provided by the data store that we use for interacting with settings on the client now, these individual datapoints are no longer needed and unnecessarily complicate things where used.

A few examples of these datapoints are (not a comprehensive list):
* Analytics: `(GET|POST):(account-id|profile-id|use-snippet`
* AdSense: `(GET|POST):(account-id|client-id)`
* Tag Manager: `(GET|POST):(account-id|container-id)`

<details>
	<summary>All current settings by module</summary>
	
```json
{
  "site-verification": [],
  "search-console": [
    "propertyID"
  ],
  "adsense": [
    "ownerID",
    "accountID",
    "clientID",
    "accountStatus",
    "siteStatus",
    "accountSetupComplete",
    "siteSetupComplete",
    "useSnippet"
  ],
  "analytics": [
    "ownerID",
    "accountID",
    "adsenseLinked",
    "anonymizeIP",
    "internalWebPropertyID",
    "profileID",
    "propertyID",
    "trackingDisabled",
    "useSnippet"
  ],
  "pagespeed-insights": [],
  "optimize": [
    "ownerID",
    "ampExperimentJSON",
    "optimizeID"
  ],
  "tagmanager": [
    "ownerID",
    "accountID",
    "ampContainerID",
    "containerID",
    "internalContainerID",
    "internalAMPContainerID",
    "useSnippet"
  ]
}
```

</details>

Note: datapoint names are snake-case while setting keys are camelCase (with the exception of our acronym casing rules)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All module datapoint definitions and create/parse handlers for individual settings should be removed
* All internal (server-side) `get_data` and `set_data` calls for these datapoints should be updated to use the module's `Settings` instance directly instead
[...]

Issue #2468, opened 2020-12-02
Require SVGs for all images and remove unnecessary image support


caseymorrisus:
## Feature Description

All existing images should be removed and replaced with SVGs as we will only support SVGs going forward. Existing logic for handling images can be removed to simplify logic.

~~We currently need to explicitly append a global variable (`global._googlesitekitLegacyData.admin.assetsRoot`) to every image source that we use. This is a bit tedious and could be handled better via a Webpack loader that appends this variable to every image source automatically.~~

See [this conversation](https://github.com/google/site-kit-wp/pull/2313#discussion_r517701856) for the origins of this issue.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All plugin usage of non-vector images (e.g. `.png`, `.jpg`, ...) should be replaced with using SVG.
    * Note most non-vector images will be removed via #2285 or #2290 anyway.
    * Please ping @felixarntz / @marrrmarrr with a list of images where we need SVG replacements that we don't have.
* The `assets/images` folder should be removed and it should be removed from the build process.
* The `smallImage` and `winImage` props of the `Notification` component should be removed to encourage usage of vector images.
* Storybook stories using images should be updated accordingly and checked that they display as expected as well.

## Implementation Brief

- Please ping @felixarntz / @marrrmarrr with a list of images where we need SVG replacements that we don't have.
- Remove `allImagesLoaded` logic from `./storybook/preview-head.html` which can be found [here](https://github.com/google/site-kit-wp/pull/2173/files).
- Remove build infra and dependencies
  - Within `package.json` remove all `imagemin` related scripts
  - Remove all `imagemin-*` (dev)dependencies
- Remove images within `/assets/images/`
- Update usage of removed images to use vectors
  - **`logo-g_white_small.png`**
    - Create a new `Core\Util\Google_Icon` class which holds the source SVG string in an `::XML` class constant  
      - Add a public static `to_base64` method which returns a base64-encoded version of the html string
        ```html
[...]

PR #2685, opened 2021-01-26
Fix PHP notice in AdSense constructor by removing additional unused datapoints

aaemnnosttv:
## Summary

Addresses issue #2507

## Relevant technical choices

* Fixes initial follow-up issue related to PHP notice raised during `setup_info` call by inlining the fallback case that we were already using for the module's `homeurl`
* `GET:reports-url` datapoint is no longer used and removed
* Initially extracted `GET:account-url` datapoint to a method for it to be used in `GET:notifications` so the now unused `GET:account-url` could be removed
* ~~Found that AdSense's `GET:notifications` datapoint is also unused and removed this as well, which allowed for the added method for `get_account_url` to also be removed as it was now unused as well~~
    - ~~I was a bit surprised about this but after looking through the code, the AdSense module uses `getAlerts`, never `getNotifications` and all references to the notifications datapoint are all for site notifications, no module notifications are ever requested. Even the datapoint for AdSense's `GET:notifications` used `get_data( 'alerts' )` internally so it makes sense that this would no longer be used. This should simplify the changes needed in #2468 as well.~~

## Checklist

### DIFF
--- a/includes/Modules/AdSense.php
+++ b/includes/Modules/AdSense.php
@@ -359,26 +359,6 @@ final class AdSense extends Module
 	}
 
 	/**
-	 * Gets the service URL for the current account or signup if none.
-	 *
-	 * @since n.e.x.t
-	 *
-	 * @return string
-	 */
-	protected function get_account_url() {
-		$option = $this->get_settings()->get();
-
-		if ( ! empty( $option['accountID'] ) && $this->authentication->profile()->has() ) {
-			return add_query_arg(
-				array( 'authuser' => $this->authentication->profile()->get()['email'] ),
-				sprintf( 'https://www.google.com/adsense/new/%s/home', $option['accountID'] )
-			);
-		}
-
-		return 'https://www.google.com/adsense/signup/new';
-	}
-
-	/**
 	 * Parses a response for the given datapoint.
 	 *
 	 * @since 1.0.0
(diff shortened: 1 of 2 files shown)


includes/Modules/AdSense.php
```diff
@@ -123,21 +123,6 @@ public function prepare_info_for_js() {
...
-		);
-
-		$signup_args = array(
-			'utm_source' => 'site-kit',
-			'utm_medium' => 'wordpress_signup',
-		);
-
-		$info['accountURL'] = add_query_arg( $idenfifier_args, $this->get_data( 'account-url' ) );
-		$info['signupURL']  = add_query_arg( $signup_args, $info['accountURL'] );
-		$info['rootURL']    = add_query_arg( $idenfifier_args, 'https://www.google.com/adsense/' );
```
aaemnnosttv, line 139: These are all unused now. We likely missed this cleanup when we added the `getServiceURL` selectors for AdSense.

includes/Modules/AdSense.php
```diff
@@ -310,44 +279,6 @@ protected function create_data_request( Data_Request $data ) {
 				}
 
 				return $this->create_adsense_earning_data_request( array_filter( $args ) );
-			case 'GET:notifications':
-				return function() {
-					$alerts = $this->get_data( 'alerts' );
```
aaemnnosttv, line 315: AdSense uses `getAlerts` directly which uses `GET:alerts` so this is no longer used.

eugene-manuilov (dismissed):

includes/Modules/AdSense.php
felixarntz: @aaemnnosttv This must not be removed. `getAlerts` is used during setup, but we still have legacy notifications functionality relying on this. The overall `GET:modules/adsense/notifications` endpoint will look if this particular datapoint here exists and use it, so this block needs to be kept in place.

felixarntz (changes requested):
One thing here shouldn't be removed, otherwise LGTM.

includes/Modules/AdSense.php
aaemnnosttv: @felixarntz I know it's connected to that endpoint but I couldn't find where it's actually used. Can you point me to it?

felixarntz: @aaemnnosttv It's used in the `Alert` component which is included on every module page for the respective module. Basically if some alert comes up in your AdSense account, you'll see it on the AdSense module page.

aaemnnosttv: Ah, I see it now. Apparently also used in `DashboardModulesAlerts`. Thanks!

aaemnnosttv: Restored 👍

felixarntz (approved):
Great!

Issue #2674, opened 2021-01-22
Insignificant fetch request errors causing E2E failures


aaemnnosttv:
## Bug Description

In #1770 we upgraded `@wordpress/scripts` to the latest which included the addition of `@wordpress/jest-console` which is enabled automatically for our JS unit tests.

In #842, we added `@wordpress/jest-console` to our Jest config for E2E, which achieved the cleaner output we were looking for but this seems to have been at the cost of stability.

This Jest preset mocks all of the main console methods (`log`, `warn`, `error`) and by default will fail the test if none of these are explicitly expected. This is a problem for some tests particularly where an asynchronous request fails, simply because the test completed successfully and the page navigates before the response is received. In most cases where this response is significant, we explicitly wait for it so this wouldn't be a problem here, but in others we don't and this is where some tests can become unstable and fail simply due to this unexpected console error.

A few possible solutions:
1. Introduce a new E2E utility to record subsequent fetch requests and build an array of `page.waitForResponse` promises.  
Then we'd need to identify tests affected by this problem and update them to use the utility before these requests are made, and `await Promise.all( responsePromises )` before a subsequent navigation is triggered.  
_Ideally this kind of thing could be done for all tests in a global `beforeEach`/`afterEach` but that won't really work_
1. Remove `@wordpress/jest-console` from the E2E Jest config, and only add it selectively for specific tests where we want to be more strict about console logging. This would essentially undo the cleanliness to the output we added in #842 but would result in more stable (although less strict) tests. The console output is actually more verbose now as several lines of context are added around each console log so this is not a trivial amount of noise to re-introduce.

<details>
<summary>Example console error</summary>

`JSHandle` data destructured into a string here to make it readable. The `You are probably offline` error is from `@wordpress/api-fetch` when fetch fails.

```
  console.error
    Google Site Kit API Error method:GET datapoint:health-checks type:core identifier:site error:You are probably offline. }

      214 | 
      215 | 		// eslint-disable-next-line no-console
    > 216 | 		console[ logFunction ]( text );
          | 		^
      217 | 	} );
      218 | }
      219 | 

      at Page.<anonymous> (config/bootstrap.js:216:3)
      at Page._addConsoleMessage (../../node_modules/puppeteer/lib/Page.js:628:10)
      at Page._onConsoleAPI (../../node_modules/puppeteer/lib/Page.js:550:10)
      at CDPSession.<anonymous> (../../node_modules/puppeteer/lib/Page.js:119:57)
[...]

PR #2686, opened 2021-01-26
Fix/2674 e2e fetch stability

aaemnnosttv:
## Summary

Addresses issue #2674

## Relevant technical choices

The changes to `plugin-activation.test.js` look like more than there really is due to some restructuring:
* Removed an unneeded `describe` block about JS being enabled
* Updated before/after functions to optimize plugin activation handling
* Updated GCP tests to test the same cases as the proxy auth case by extracting a `matrix` variable to share test functions

I tried testing `specs/modules/tagmanager/setup.test.js` several times but I was not able to reproduce any failure locally. I have seen it fail in GHA (including in some of these runs) although most of the time it is usually an odd log statement that _does not_ cause the test to fail and some other test is the one that fails the job, or in one case I saw a timeout error which seems to be related to click instability that we've seen before, so somewhat beyond the scope here.

## Checklist

### DIFF
--- a/tests/e2e/specs/dashboard/dashboard-noscript.test.js
+++ b/tests/e2e/specs/dashboard/dashboard-noscript.test.js
@@ -1,4 +1,22 @@
 /**
+ * Dashboard noscript tests.
+ *
+ * Site Kit by Google, Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


felixarntz (approved):
@aaemnnosttv Awesome, that should help a lot!

There's still another e2e instability I've been seeing fairly often, around Tag Manager and AMP:
Screen Shot 2021-01-26 at 8 28 55 AM

Maybe we should open another issue to review that?

Issue #2675, opened 2021-01-22
All Traffic Widget: Don't show Dimensions tabs when no data is available for Analytics


tofumatt:
When the All Traffic Widget doesn't have any data to show, the tabs that allow the user to select their dimension should not appear, because there's no data to filter. Currently, the tabs appear even when `isZeroData( report )` is `truth`-y, but they should only be shown when `isZeroData( report )` is `false`-y.

## Screenshots

![image](https://user-images.githubusercontent.com/90871/105494162-4bc3f680-5cb2-11eb-9bef-9ae1c9610791.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* When a "zero data" or "error" state is present for any selector used by the `DashboardAllTrafficWidget`, the entire component should show a "Zero data" or "error" state, and not render the individual children components/charts.

## Implementation Brief

After a call with @felixarntz, @aaemnnosttv, and @eugene-manuilov, we decided the best solution is to move all selectors for the `DashboardAllTrafficWidget` children components into the `DashboardAllTrafficWidget` component itself, then pass the `loading` and `data`/`report` variables to the child components as props. Data in `error` (from the `getErrorForSelector` selector) should not be passed to children—instead, if the component has `zeroData` from the `UserDimensionsPieChart` component or an **error** state, one large "zero data" or "error" state should be rendered.

* Move all `loading`/`error`/`getReport` selectors from `TotalUserCount`, `UserCountGraph`, and `UserDimensionsPieChart` to `DashboardAllTrafficWidget`.
  * Pass the `loading` and `data/report` variables as props to their respective components
  * _Optional:_ All components should show their "loading" state until _all_ data has loaded on first component load/render. After the first load, updates for things like clicking on a dimension in the pie chart to filter results should not enable a loading state in all components, only the ones that are updating (this can be done using `useEffect`/`useState`). If this is done, `dateRange` updates should trigger a full `loading` state.
* The widget's main component should only return an overall `error` or "zero data" state as a result for the `UserDimensionsPieChart` request. That one is the only request which on failure/zero would indicate the entire widget is unusable. If that component's `getReport` requests has no `error` and is not "zero data" then each child component with its own error/loading/report variables should be rendered.

### Test Coverage

* No test coverage needed.

### Visual Regression Changes

* If the "zero data" state has a VRT, it will need adjustment.

## QA Brief

[...]

PR #2687, opened 2021-01-26
Enhancement/2675 all traffic updates

eugene-manuilov:
## Summary

Addresses issue #2675

## Checklist

### DIFF
--- a/assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/TotalUserCount.js
+++ b/assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/TotalUserCount.js
@@ -39,7 +39,7 @@ import ChangeArrow from '../../../../../components/ChangeArrow';
 import PreviewBlock from '../../../../../components/PreviewBlock';
 const { useSelect } = Data;
 
-export default function TotalUserCount( { loaded, error, report, dimensionValue } ) {
+export default function TotalUserCount( { loaded, report, dimensionValue } ) {
 	const dateRange = useSelect( ( select ) => select( CORE_USER ).getDateRange() );
 
 	if ( ! loaded ) {
@@ -53,7 +53,7 @@ export default function TotalUserCount( { loaded, error, report, dimensionValue 
 		);
 	}
 
-	if ( error || isZeroReport( report ) ) {
+	if ( isZeroReport( report ) ) {
 		// The UserCountGraph component will return appropriate ReportError/ReportZero component
 		// based on the report fetching status, so we can return just NULL here to make sure it doesn't take extra space.
 		return null;
@@ -108,7 +108,6 @@ export default function TotalUserCount( { loaded, error, report, dimensionValue 
 
 TotalUserCount.propTypes = {
 	loaded: PropTypes.bool,
-	error: PropTypes.shape( {} ),
 	report: PropTypes.arrayOf( PropTypes.object ),
 	dimensionValue: PropTypes.string,
 };
--- a/assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/TotalUserCount.js
+++ b/assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/TotalUserCount.js
@@ -36,9 +36,10 @@ import { numFmt, calculateChange } from '../../../../../util';
 import { getAvailableDateRanges } from '../../../../../util/date-range';
 import ChangeArrow from '../../../../../components/ChangeArrow';
 import PreviewBlock from '../../../../../components/PreviewBlock';
+import ReportError from '../../../../../components/ReportError';
 const { useSelect } = Data;
 
-export default function TotalUserCount( { loaded, report, dimensionValue } ) {
+export default function TotalUserCount( { loaded, error, report, dimensionValue } ) {
 	const dateRange = useSelect( ( select ) => select( CORE_USER ).getDateRange() );
 
 	if ( ! loaded ) {
@@ -50,6 +51,10 @@ export default function TotalUserCount( { loaded, report, dimensionValue } ) {
 				shape="square"
 			/>
 		);
+	}
+
+	if ( error ) {
+		return <ReportError moduleSlug="analytics" error={ error } />;
 	}
 
 	const { totals } = report?.[ 0 ]?.data || {};
--- a/assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/TotalUserCount.js
+++ b/assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/TotalUserCount.js
@@ -34,7 +34,6 @@ import Data from 'googlesitekit-data';
 import { CORE_USER } from '../../../../../googlesitekit/datastore/user/constants';
 import { numFmt, calculateChange } from '../../../../../util';
 import { getAvailableDateRanges } from '../../../../../util/date-range';
-import { isZeroReport } from '../../../util';
 import ChangeArrow from '../../../../../components/ChangeArrow';
 import PreviewBlock from '../../../../../components/PreviewBlock';
 const { useSelect } = Data;
@@ -51,12 +50,6 @@ export default function TotalUserCount( { loaded, report, dimensionValue } ) {
 				shape="square"
 			/>
 		);
-	}
-
-	if ( isZeroReport( report ) ) {
-		// The UserCountGraph component will return appropriate ReportError/ReportZero component
-		// based on the report fetching status, so we can return just NULL here to make sure it doesn't take extra space.
-		return null;
 	}
 
 	const { totals } = report?.[ 0 ]?.data || {};
@@ -94,7 +87,10 @@ export default function TotalUserCount( { loaded, report, dimensionValue } ) {
 				</span>
 				<span className={ classnames(
 					'googlesitekit-data-block__value',
-					`googlesitekit-data-block__value--${ 0 <= change ? 'up' : 'down' }`
+					{
+						'googlesitekit-data-block__value--up': 0 <= change,
+						'googlesitekit-data-block__value--down': 0 > change,
+					},
 				) }>
 					{ numFmt( Math.abs( change ), { style: 'percent', maximumFractionDigits: 1 } ) }
 				</span>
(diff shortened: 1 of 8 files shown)


assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/TotalUserCount.js
```diff
@@ -125,7 +87,10 @@ export default function TotalUserCount( { dimensionName, dimensionValue } ) {
 				</span>
 				<span className={ classnames(
 					'googlesitekit-data-block__value',
-					`googlesitekit-data-block__value--${ 0 <= change ? 'up' : 'down' }`
+					{
+						'googlesitekit-data-block__value--up': 0 <= change,
+						'googlesitekit-data-block__value--down': 0 > change,
+					},
```
tofumatt, line 93: 👍🏻

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/index.js
```diff
@@ -20,31 +20,101 @@
...
 import { _x } from '@wordpress/i18n';
+import { useState, useEffect } from '@wordpress/element';
 
 /**
  * Internal dependencies
  */
 import Data from 'googlesitekit-data';
 import Widgets from 'googlesitekit-widgets';
-import { MODULES_ANALYTICS, FORM_ALL_TRAFFIC_WIDGET } from '../../../datastore/constants';
+import { FORM_ALL_TRAFFIC_WIDGET, DATE_RANGE_OFFSET, STORE_NAME } from '../../../datastore/constants';
```
tofumatt, line 30: Let's use the new constants available and import this as `MODULES_ANALYTICS` instead, eg:

```suggestion
import { FORM_ALL_TRAFFIC_WIDGET, DATE_RANGE_OFFSET, MODULES_ANALYTICS } from '../../../datastore/constants';
```

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/index.js
```diff
@@ -20,31 +20,101 @@
...
+	const pieError = useSelect( ( select ) => select( STORE_NAME ).getErrorForSelector( 'getReport', [ pieArgs ] ) );
+	const pieReport = useSelect( ( select ) => select( STORE_NAME ).getReport( pieArgs ) );
+
+	const graphLoaded = useSelect( ( select ) => select( STORE_NAME ).hasFinishedResolution( 'getReport', [ graphArgs ] ) );
+	const graphError = useSelect( ( select ) => select( STORE_NAME ).getErrorForSelector( 'getReport', [ graphArgs ] ) );
+	const graphReport = useSelect( ( select ) => select( STORE_NAME ).getReport( graphArgs ) );
+
+	const totalsLoaded = useSelect( ( select ) => select( STORE_NAME ).hasFinishedResolution( 'getReport', [ totalsArgs ] ) );
+	const totalsError = useSelect( ( select ) => select( STORE_NAME ).getErrorForSelector( 'getReport', [ totalsArgs ] ) );
+	const totalsReport = useSelect( ( select ) => select( STORE_NAME ).getReport( totalsArgs ) );
```
tofumatt, line 116:
```suggestion
	const pieLoaded = useSelect( ( select ) => select( MODULES_ANALYTICS ).hasFinishedResolution( 'getReport', [ pieArgs ] ) );
	const pieError = useSelect( ( select ) => select( MODULES_ANALYTICS ).getErrorForSelector( 'getReport', [ pieArgs ] ) );
	const pieReport = useSelect( ( select ) => select( MODULES_ANALYTICS ).getReport( pieArgs ) );

	const graphLoaded = useSelect( ( select ) => select( MODULES_ANALYTICS ).hasFinishedResolution( 'getReport', [ graphArgs ] ) );
	const graphError = useSelect( ( select ) => select( MODULES_ANALYTICS ).getErrorForSelector( 'getReport', [ graphArgs ] ) );
	const graphReport = useSelect( ( select ) => select( MODULES_ANALYTICS ).getReport( graphArgs ) );

	const totalsLoaded = useSelect( ( select ) => select( MODULES_ANALYTICS ).hasFinishedResolution( 'getReport', [ totalsArgs ] ) );
	const totalsError = useSelect( ( select ) => select( MODULES_ANALYTICS ).getErrorForSelector( 'getReport', [ totalsArgs ] ) );
	const totalsReport = useSelect( ( select ) => select( MODULES_ANALYTICS ).getReport( totalsArgs ) );
```

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/index.js
```diff
@@ -67,7 +137,31 @@ function DashboardAllTrafficWidget() {
 		};
 	}
 
-	const serviceReportURL = useSelect( ( select ) => select( MODULES_ANALYTICS ).getServiceReportURL( reportType, reportArgs ) );
+	const serviceReportURL = useSelect( ( select ) => select( STORE_NAME ).getServiceReportURL( reportType, reportArgs ) );
```
tofumatt, line 140:
```suggestion
	const serviceReportURL = useSelect( ( select ) => select( MODULES_ANALYTICS ).getServiceReportURL( reportType, reportArgs ) );
```

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/index.js
```diff
@@ -20,31 +20,101 @@
...
+	}
+
+	if ( dimensionName && dimensionValue ) {
+		graphArgs.dimensionFilters = { [ dimensionName ]: dimensionValue };
+		totalsArgs.dimensionFilters = { [ dimensionName ]: dimensionValue };
+	}
+
+	const pieLoaded = useSelect( ( select ) => select( STORE_NAME ).hasFinishedResolution( 'getReport', [ pieArgs ] ) );
+	const pieError = useSelect( ( select ) => select( STORE_NAME ).getErrorForSelector( 'getReport', [ pieArgs ] ) );
+	const pieReport = useSelect( ( select ) => select( STORE_NAME ).getReport( pieArgs ) );
```
tofumatt, line 108: Bit of a nitpick, but can we use slightly-more verbose variable names for these? Maybe `pieChartLoaded` and `userCountGraph`, etc.? Just to make things a bit clearer.

tofumatt (changes requested):
I tested this locally and it looks great, but I think it's missing one piece from the IB:

> The widget's main component should only return an overall error or "zero data" state as a result for the UserDimensionsPieChart request. That one is the only request which on failure/zero would indicate the entire widget is unusable. If that component's getReport requests has no error and is not "zero data" then each child component with its own error/loading/report variables should be rendered.

[@felixarntz mentioned that we should only show an error for the entire component if the "main data source" (eg the PieChart report request).](https://github.com/google/site-kit-wp/issues/2675#issuecomment-767075125). So I think we should still be passing the `error` prop to the children component and rendering smaller errors when they happen.

Otherwise if an error occurs when selecting, for instance, a slice of the pie chart after everything has loaded, the entire component will show an error rather than just the graph. The component would still be usable if the pie slice was de-selected, so it's worth passing the errors around.

That and a few tweaks to the code in the PR, but overall looks great 👍🏻

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/index.js
eugene-manuilov: Updated.

eugene-manuilov:
Thanks, @tofumatt! I have addressed all issues. Could you please take another look?

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/UserCountGraph.js
```diff
@@ -114,98 +52,118 @@ export default function UserCountGraph( { dimensionName, dimensionValue } ) {
 		return <ReportError moduleSlug="analytics" error={ error } />;
 	}
 
-	if ( isZeroReport( report ) ) {
-		return <ReportZero moduleSlug="analytics" />;
-	}
```
felixarntz, line 119: This is different from before, but good point on removing the `isZeroReport` check from `UserCountGraph` and `TotalUserCount`. These might be 0 because of a certain filtered dimension having 0 users. In that case, it still makes sense to show the 0s instead of the `ReportZero` box since the latter only exists to indicate that likely your site doesn't have enough data for meaningful results. For these two components though, if they are rendered at all, it means there _is_ data overall, so 👍 

Nice UX improvement here (or we could even argue the behavior from before was a bug).

tofumatt (dismissed):
Looks great! And second what @felixarntz mentioned about the `isZeroReport` on the filtered data. 👍🏻

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/index.js
```diff
@@ -69,6 +139,30 @@ function DashboardAllTrafficWidget() {
...
+		dateRange,
+		currentRange,
+	] );
+
+	if ( pieChartError ) {
+		return <ReportError moduleSlug="analytics" error={ pieChartError } />;
+	}
+
+	const totalUsers = parseInt( pieChartReport?.[ 0 ]?.data?.totals?.[ 0 ]?.values?.[ 0 ] );
+	if ( pieChartLoaded && ( Number.isNaN( totalUsers ) || totalUsers < 1 ) ) {
```
felixarntz, line 162: Why not use `isZeroReport( pieChartReport )`? That's the intended function for this purpose - we shouldn't manually inspect the response to check for a "zero data" state.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidgetV2/index.js
```diff
@@ -69,6 +139,30 @@ function DashboardAllTrafficWidget() {
...
+		return <ReportZero moduleSlug="analytics" />;
```
felixarntz, line 163: This should in principle use `props.WidgetReportZero` now instead since #2252 has been merged. However, we'll still need to use this component in the legacy UI where it doesn't receive these props, so it can remain as is for now.

However, let's add a comment:
```
// TODO: Replace with `props.WidgetReportZero` once legacy UI has been removed.
```

felixarntz (commented):
Awesome work!

Two remaining points of feedback here, but nothing critical.

felixarntz (approved):
Pushed the last changes here, good to go!

Issue #2693, opened 2021-01-26
Trigger authorization action hooks also for sites that do not use the proxy


felixarntz:
The `googlesitekit_authorize_user` action (receiving a new set of access token and refresh token) and `googlesitekit_reauthorize_user` action (receiving a new access token using a refresh token) are currently only fired when using the proxy. `googlesitekit_authorize_user` was originally introduced like that because the proxy includes some extra arguments in the response, but that's not really a good reason to not fire that hook for other sites that use manual OAuth credentials and directly access Google's OAuth infrastructure. `googlesitekit_reauthorize_user` was simply put behind the same flag for parity with the existing action.

We should change it so that these actions run regardless of whether or not a site is using the Site Kit authentication proxy, since the action names are not specific to that and imply a certain action that is also relevant without the proxy.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `googlesitekit_authorize_user` and `googlesitekit_reauthorize_user` actions should be fired regardless of whether the proxy is being used. The `TODO` comments around the clauses should be removed accordingly.
* Any callbacks hooked into one of these actions should be individually reviewed, and it should be determined whether they need to be modified:
    * If the callback relies on data in the `$token_response` action parameter that is only present when using the proxy, it should be ensured this callback does not cause an error when not using the proxy (e.g. because of an undefined array index).
    * If the callback performs functionality that is only relevant for the proxy (e.g. anything around user input, remote feature detection, checking proxy URLs, ...), the callback should receive an early return for when the proxy is not being used.

## Implementation Brief

* Update `includes/Core/Authentication/Clients/OAuth_Client.php`, removing the if statement around the `googlesitekit_authorize_user` action trigger *and removing the TODO: comment*:

https://github.com/google/site-kit-wp/blob/fb2bbdf94d435753a3c84f2a1064108144e390df/includes/Core/Authentication/Clients/OAuth_Client.php#L729-L745

---

* Update hooks into the action `googlesitekit_authorize_user`:
* Update `includes/Core/Authentication/Authentication.php` use of the `googlesitekit_authorize_user` hook to return early if the proxy is not being used, to prevent	`$this->set_connected_proxy_url();` and `$this->require_user_input();` being called in this case:

https://github.com/google/site-kit-wp/blob/baec79d391492896a1ac4d0b41a6c93ac67ccf5c/includes/Core/Authentication/Authentication.php#L295-L301

[...]

Issue #2692, opened 2021-01-26
Store initially used Site Kit version per user


felixarntz:
Originally part of #2648, we decided to take the part of storing a user's initially used Site Kit version out of that issue so that we can squeeze it into the current 1.25.0 release. This is necessary because we want to ship the first set of tooltips in the following release, and it requires us to already start collecting active version information before.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new user setting `googlesitekitpersistent_initial_version` (string) should be introduced and registered.
    * The `googlesitekitpersistent_` prefix should be used for any settings that require more persistent storage, **which should only be in specific circumstances**. By using the special prefix, this data will not be wiped on disconnect or reset.
* When a user receives a new access token (action `googlesitekit_authorize_user`) or receives a refreshed access token (new action to introduce: `googlesitekit_reauthorize_user`), the `googlesitekitpersistent_initial_version` option should be set (to the current plugin version) if it wasn't set before.
    * Of course this doesn't work retroactively, but that's fine. This mechanism will allow us in the future to know for every user what version of Site Kit they first used, and we'll be able to provide tooltip contextually.

## Implementation Brief

* Add `Core\Authentication\Initial_Version` class which extends `User_Setting` for the new `googlesitekitpersistent_initial_version`.
* Add protected property for it in `Core\Authentication\Authentication` and instantiate it in the constructor.
* Add `do_action` call for new `googlesitekit_reauthorize_user` to `Core\Authentication\Clients\Google_Site_Kit_Client::fetchAccessTokenWithRefreshToken` in the clause where the access token is set, at the end. It should pass through the response data (in `$creds` variable) similar to the existing related `googlesitekit_authorize_user` action.
* In `Core\Authentication\Authentication`, if the `Core\Authentication\Initial_Version` option value is empty, add a callback to both `googlesitekit_authorize_user` and `googlesitekit_reauthorize_user` which will set the `Core\Authentication\Initial_Version` option value to the current Site Kit version.

### Test Coverage

* Add tests to `AuthenticationTest` class that ensures the `Authentication::register` method adds the new hooks as expected, and also ensure that triggering the hook in that case results in the option being set afterwards.

### Visual Regression Changes

* N/A

## QA Brief

**This requires direct database access to verify.**

* After going through the initial Site Kit setup/OAuth flow, a `googlesitekitpersistent_initial_version` user meta should be set to the current Site Kit version for the user you connected.
[...]

PR #2694, opened 2021-01-26
Store initial Site Kit version per user.

felixarntz:
## Summary

Addresses issue #2692

## Relevant technical choices

* The new hook is, similar to the old one, only run when using the proxy. This is a bit odd, but I didn't wanna diverge from that behavior in here, and I also didn't wanna introduce a broader change where we would run these hooks for any sites. Mostly irrelevant anyway since probably almost nobody does not use the proxy, but I think it's worth addressing, so I opened a follow-up issue #2693 to run these hooks in any scenario.

## Checklist

### DIFF
--- a/includes/Core/Authentication/Authentication.php
+++ b/includes/Core/Authentication/Authentication.php
@@ -480,17 +480,6 @@ final class Authentication {
 	}
 
 	/**
-	 * Gets the Initial_Version instance.
-	 *
-	 * @since n.e.x.t
-	 *
-	 * @return Initial_Version An instance of the Initial_Version class.
-	 */
-	public function get_initial_version() {
-		return $this->initial_version;
-	}
-
-	/**
 	 * Revokes authentication along with user options settings.
 	 *
 	 * @since 1.0.0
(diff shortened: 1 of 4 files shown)


includes/Core/Authentication/Authentication.php
```diff
@@ -182,6 +182,15 @@ final class Authentication {
...
 	protected $google_proxy;
 
+	/**
+	 * Initial_Version instance.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @var User_Input_State
+	 */
+	private $initial_version = null;
```
aaemnnosttv, line 192: The `@var` definition needs updating here. Also, the IB (as well as other instance properties here) are `protected` but this is `private`. Also, let's omit the `null` assignment for these since that is the value it will have simply by declaring it anyways.
```suggestion
	 * @var Initial_Version
	 */
	protected $initial_version;
```
The empty line between `@since` and `@var` is removed in the suggestion as well for consistency with the others.

aaemnnosttv (commented):
Looks good to me, just the one change to the property definition; I'll apply that now.

aaemnnosttv (approved):
Awesome!

Issue #2195, opened 2020-10-16
Enhance acronym ESLint rule to enforce lowercase when applicable


aaemnnosttv:
## Feature Description

Our custom `sitekit/camelcase-acronyms` ESLint rule enforces our conventions around when acronyms should be capitalized.

While this works great for acronyms used within a variable name, it misses the inverse of this logic which applies to variable names that start with an acronym (e.g. `ampMode`), or variables which _are_ an acronym (e.g. `url`).

In these cases, the acronyms should be required to be _lowercase_ instead.

Perhaps the rule itself should be renamed to `sitekit/acronym-case` since the logic applies to a bit more than just camel-case ?

Note: one exception to this case are React components, so if the name is a class or function that starts with a capital letter and an acronym, then it must be all capitalized 😄 

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Acronyms defined in the `sitekit/camelcase-acronyms` rule should be required to be lowercase whenever a qualifying name to be validated starts with the acronym
* React component names (or just all functions) that begin with an acronym must require the acronym to be all uppercase (due to React component naming restrictions)

## Implementation Brief

* Rename `packages/eslint-plugin/rules/camelcase-acronyms.js` to `packages/eslint-plugin/rules/acronym-case.js`
* Add checks for lowercase acronyms to https://github.com/google/site-kit-wp/blob/4708e025503a007f27f5b45e9d15295f7d8f8952/packages/eslint-plugin/rules/camelcase-acronyms.js#L106
  * This should entail checking to see if the match was found at the start of the variable name, and if so, that it is _entirely_ lowercase instead of uppercase.
* Add a check to see if the variable is a function, in which case it _can_ begin with either:
  * an all-lowercase acronym (normal function)
  * an all-uppercase acronym (a React component)

There's no way to determine if a function is a React component, so allowing either is the best option here. Any React component will start with uppercase and thus would require the all-caps uppercase anyway, so `AmpError` should fail the linter but `ampError` and `AMPError` would work.

### Test Coverage

* Add ESlint tests for this component to cover the above cases.

### Visual Regression Changes

* No changes to VRT.

## QA Brief

* `npm run test:js` should pass as our rules cover the changes made to the linter here.

## Changelog entry

* N/A

aaemnnosttv:
IB ✅

johnPhillips:
[...]

PR #2696, opened 2021-01-27
Enhance acronym ESLint rule to enforce lowercase when applicable

johnPhillips:
## Summary

Addresses issue #2195

## Checklist


tofumatt:
@googlebot I consent.

tofumatt (dismissed):

packages/eslint-plugin/utils.js
```diff
@@ -100,14 +101,29 @@ function isImported( node ) {
...
+	if (
+		node?.type === 'Identifier' && (
+			hasFunctionType( node?.parent?.type ) ||
+			hasFunctionType( node?.parent?.init?.type )
+		)
+	) {
+		return true;
+	}
+
+	const isFunctionDeclaration = hasFunctionType( node?.type );
```
tofumatt, line 126: I think we weren't using optional chaining in these files before because our version of node was too old, but now that we require Node 14 via NVM this should be fine. 👍🏻

tofumatt (approved):

aaemnnosttv (commented):
LGTM!

@johnPhillips it looks like all the changes were from renaming the rule, is that correct? Were there any actual violations that needed fixing?

Otherwise, the only thing needed here is a file conflict to be resolved.

johnPhillips:
@aaemnnosttv the bulk of the changes were just the renaming, yes. I also added a couple of extra checks to the rule as per the AC. There weren't any violations that I can remember, so it was pretty clean.

aaemnnosttv (approved):
Great!

aaemnnosttv (changes requested):
Oops! I just tried this out and found that it did not detect a few cases where it should.

image

I pushed an update to the test case for this rule so it should be as simple as getting the tests to pass again from here 😄

aaemnnosttv:
@googlebot I consent.

johnPhillips:
@aaemnnosttv CC @tofumatt 
I fixed the broken test there, but it seems to have been down to this line:
```js
// If the name of this variable is the same length as the acronym,
// it should be lowercase or uppercase.
- if ( name.length === acronym.length && ( acronymMatch === acronym.toLowerCase() || acronymMatch !== acronym ) ) {
+ if ( name.length === acronym.length && ( acronymMatch === acronym.toLowerCase() || acronymMatch === acronym.toUpperCase() ) ) {
```
I couldn't figure out what the original intent was with the ` || acronymMatch !== acronym` check here. It didn't _seem_ necessary, so I removed it. I hope I didn't break something - more than happy to add it back in if I did 👍

aaemnnosttv (approved):
Super, thanks!

Issue #2452, opened 2020-11-30
Make feature flags dynamic


aaemnnosttv:
## Feature Description

Currently our feature flags are defined in `webpack.feature-flags.config.js` (soon to be `feature-flags.json`) which is provided as `featureFlags` to all modules in JS via Webpack. Feature flags allow us to using avoid long-running feature branches and merge code before it is ready for production use. Originally, the thought was that non-production feature code would be tree-shaken out for releases but that isn't really the case and is actually more of a feature than a bug as it allows translations to be prepared before the features are enabled in a subsequent release.

One major drawback of the current approach however is that is problematic to test these features, particularly in E2E. The build we use for E2E uses a normal production build, with added utilities for use in the tests. This means that only production features will be enabled (i.e. no flag-specific features).

This is also a bit of a problem in Storybook where `featureFlags` is treated as a global that persists across stories. This means that enabling a feature in one story will cause it to be enabled in another which may be specific to the current production case. This can lead to unexpected results when viewing Storybook and switching between stories. VRT is more/less unaffected by this since Backstop loads the story iframe directly for each case rather than switching via the UI.

Given the changes coming in #2316, it makes sense to build on this to make the server the source of truth when it comes to available/enabled features.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* `feature-flags.json` should be restructured into a single-level key-value structure, removing the `enabled` key entirely
_E.g. `userInput.enabled` would be `userInput` – `widgets.dashboard` may remain, but as a single string key_
* The `FeatureFlagsPlugin` should be removed from the `webpack.config.js`
* `webpack-feature-flags-plugin` dependency should be removed
* `featureFlags` should be removed as an allowed global from the ESLint config and all usage should be updated to consume the value via the `useFeature` hook

### PHP

* Feature activation should be provided from the server as part `_googlesitekitBaseData` as `enabledFeatures` as a list of feature keys/paths that are enabled
* The `FeatureFlags::enabled` method should be made filterable by passing the result through a new `googlesitekit_is_feature_enabled` filter
	- the filterable value is the boolean value to be returned by the method
	- the filter should pass the `FeatureFlags::$mode` as the second parameter to the filter

### E2E

* There should be a new `enableFeature( feature )` E2E utility function to enable a feature by its feature "path" (e.g. `userInput` or `widgets.dashboard`)
* There should be a new `disableFeature( feature )` E2E utility function to disable a feature by its feature "path"
[...]

PR #2698, opened 2021-01-27
Enhancement/2452 e2e feature flags

eugene-manuilov:
## Summary

Addresses issue #2452

## Relevant technical choices

- Renamed the option name to escape it from being deleted when the site kit is reset.

## Checklist

### DIFF
--- a/tests/e2e/mu-plugins/e2e-rest-feature-flag.php
+++ b/tests/e2e/mu-plugins/e2e-rest-feature-flag.php
@@ -26,7 +26,7 @@ add_action(
 			array(
 				'methods'             => WP_REST_Server::EDITABLE,
 				'callback'            => function ( WP_REST_Request $request ) {
-					$feature_flag_overrides = get_option( '_e2e_feature_flags', array() );
+					$feature_flag_overrides = get_option( 'googlesitekit_e2e_feature_flags', array() );
 
 					if ( $request['feature_value'] ) {
 						$feature_flag_overrides[ $request['feature_name'] ] = (bool) $request['feature_value'];
@@ -34,7 +34,7 @@ add_action(
 						unset( $feature_flag_overrides[ $request['feature_name'] ] );
 					}
 
-					update_option( '_e2e_feature_flags', $feature_flag_overrides );
+					update_option( 'googlesitekit_e2e_feature_flags', $feature_flag_overrides );
 
 					return array(
 						'success' => true,
@@ -51,7 +51,7 @@ add_action(
 add_filter(
 	'googlesitekit_is_feature_enabled',
 	function ( $feature_enabled, $feature_name ) {
-		$features = get_option( '_e2e_feature_flags', array() );
+		$features = get_option( 'googlesitekit_e2e_feature_flags', array() );
 
 		return ! empty( $features[ $feature_name ] );
 	},
--- a/tests/e2e/mu-plugins/e2e-rest-feature-flag.php
+++ b/tests/e2e/mu-plugins/e2e-rest-feature-flag.php
@@ -26,7 +26,7 @@ add_action(
 			array(
 				'methods'             => WP_REST_Server::EDITABLE,
 				'callback'            => function ( WP_REST_Request $request ) {
-					$feature_flag_overrides = get_option( 'googlesitekit_e2e_feature_flags', array() );
+					$feature_flag_overrides = get_option( '_e2e_feature_flags', array() );
 
 					if ( $request['feature_value'] ) {
 						$feature_flag_overrides[ $request['feature_name'] ] = (bool) $request['feature_value'];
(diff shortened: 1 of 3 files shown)


tests/e2e/mu-plugins/e2e-rest-feature-flag.php
```diff
@@ -26,15 +26,15 @@ function () {
...
+					$feature_flag_overrides = get_option( '_e2e_feature_flags', array() );
 
 					if ( $request['feature_value'] ) {
 						$feature_flag_overrides[ $request['feature_name'] ] = (bool) $request['feature_value'];
 					} else {
 						unset( $feature_flag_overrides[ $request['feature_name'] ] );
 					}
 
-					update_option( 'googlesitekit_e2e_feature_flags', $feature_flag_overrides );
+					update_option( '_e2e_feature_flags', $feature_flag_overrides );
```
aaemnnosttv, line 37: I'm assuming this change is to allow the flags to persist across resets (which may be needed), it was done this way intentionally (at least for now). If we need the flag to persist across resets, this will require additional changes to satisfy the ACs:

> * All features should be disabled by default in E2E until explicitly enabled
> * Enabled features should be reset after each test case (perhaps any time Site Kit is reset) so that no features are enabled again

In this case, we should probably add a `e2e/feature/reset` endpoint to set an empty array and hook that in `afterAll` in the bootstrap so that one test file doesn't influence another. Each test would still need to reset flags as necessary within its own file though, similar to how we handle the plugin reset.

I would leave it as-is for now though unless we can foresee needing it right away?

tests/e2e/mu-plugins/e2e-rest-feature-flag.php
```diff
@@ -50,10 +50,11 @@ function () {
 // Enforce feature activation as defined by the E2E feature flags option.
 add_filter(
 	'googlesitekit_is_feature_enabled',
-	function ( $feature_name ) {
-		$features = get_option( 'googlesitekit_e2e_feature_flags', array() );
+	function ( $feature_enabled, $feature_name ) {
```
aaemnnosttv, line 53: Yes. Thank you 🤦

tests/e2e/utils/features.js
```diff
@@ -33,7 +33,10 @@ export async function disableFeature( feature ) {
 	return await wpApiFetch( {
 		path: 'google-site-kit/v1/e2e/feature/set-flag',
 		method: 'post',
-		data: { [ feature ]: false },
+		data: {
+			feature_name: feature,
+			feature_value: false,
+		},
```
aaemnnosttv, line 39: Nice catch. We should have probably required this to be used in the PR somehow to test 👍

aaemnnosttv (changes requested):
Thanks @eugene-manuilov – this all around looks great, the only change that we need to address is regarding the option name. See my comments below.

tests/e2e/mu-plugins/e2e-rest-feature-flag.php
eugene-manuilov: Ah... okay, I thought it was done by accident. I'll revert it then.

eugene-manuilov: Updated.

aaemnnosttv (approved):
Great!

I wrote up a quick E2E test to verify this all works as expected but didn't want to include it here since it just takes some screenshots:

**Widgets API Dashboard**
screenshot--googlesitekit-dashboard--widgets

**Legacy Dashboard**
screenshot--googlesitekit-dashboard--legacy

<details>
<summary>Here's the test</summary>

```js
import { activatePlugin, visitAdminPage } from '@wordpress/e2e-test-utils';
import { createWaitForFetchRequests, enableFeature, resetSiteKit, setSearchConsoleProperty, setSiteVerification } from '../../utils';

describe( 'widgets.dashboard feature', () => {
	beforeAll( async () => {
		await activatePlugin( 'e2e-tests-proxy-auth-plugin' );
	} );

	beforeEach( async () => {
		await setSiteVerification();
		await setSearchConsoleProperty();
	} );

	afterEach( async () => {
		await resetSiteKit();
	} );

	it( 'shows the widget-based dashboard', async () => {
		await enableFeature( 'widgets.dashboard' );
		const waitForFetchRequests = createWaitForFetchRequests();
		await visitAdminPage( 'admin.php', 'page=googlesitekit-dashboard' );
		await waitForFetchRequests();
		await page.screenshot( { path: 'screenshot--googlesitekit-dashboard--widgets.png' } );
		expect( console ).toHaveWarned();
		expect( console ).toHaveErrored();
	} );

	it( 'shows the legacy dashboard', async () => {
		const waitForFetchRequests = createWaitForFetchRequests();
		await visitAdminPage( 'admin.php', 'page=googlesitekit-dashboard' );
		await waitForFetchRequests();
		await page.screenshot( { path: 'screenshot--googlesitekit-dashboard--legacy.png' } );
		expect( console ).toHaveWarned();
	} );
} )
;

```

</details>

PR #2702, opened 2021-01-28
Select or unselect initial All Traffic pie slice based on dimension value and whether it is included in the current data

felixarntz:
## Summary

Addresses issue #2644

(alternative approach to #2684)

## Relevant technical choices

* Rely on `dimensionValue` to find out which slice should be selected (separate state field unnecessary).
* In `onReady` chart callback:
    * If there is a `dimensionValue` but no selection, make sure the row with that `dimensionValue` is selected.
    * If there is no `dimensionValue` but (still) a selection, make sure the selection is emptied.
* Also ensure that, when such a "fix" is necessary, the color stored is updated accordingly. Since the dimension values are not always in the same order, the color for the same dimension value may be different when using another date range.
* This PR certainly does not fix all underlying problems with how we use the charts library, there are still some quirks with it overall. IMO this PR clearly improves on the current situation though and addresses the main bug. We can try to improve further and potentially open a follow-up issue to fix our usage of the charts library, but I think it would be worth shipping this in the current release.
* @tofumatt Thoughts? Feel free to iterate on this approach if you like it and have further ideas for improvement.

## Checklist


felixarntz:
Changed this to be against `master` since a fix for this issue should preferably go into the release.

tofumatt (dismissed):
This is similar to an approach I took using just the `dimensionValue` and works to preserve it between date range loads sometimes, but issues with moving between tabs then selecting a slice are still present.

Given the intermittent nature of the chart redraws it's possible the date range change could still select the pie slice then unselect it, but I think this at least doesn't cause regressions and improves things so we should go with it for now. 👍🏻

felixarntz:
https://github.com/google/site-kit-wp/pull/2702/commits/3db918735ff279a31c75dc19229edae664f12ad4 addresses another piece of feedback raised by @marrrmarrr - we should be displaying the pointer cursor when users hover over a pie slice or legend entry in the new widget.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
```diff
@@ -63,19 +63,38 @@ export default function UserDimensionsPieChart( { dimensionName, sourceLink, loa
...
+			}
+
+			// If there is no dimension value set but the initialized chart does have a selection,
+			// ensure it is no longer selected in the chart.
+			if ( ! dimensionValue && chart.getSelection().length ) {
+				chart.setSelection( [] );
+				setValues( FORM_ALL_TRAFFIC_WIDGET, { dimensionColor: '' } );
+			}
+		}
+
```
eugene-manuilov, line 84: @felixarntz this should be moved to its own `useEffect` hook because the `onReady` callback is executed only once per the chart component lifetime.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
```diff
@@ -85,7 +104,7 @@ export default function UserDimensionsPieChart( { dimensionName, sourceLink, loa
 				}
 			} );
 		}
-	}, [ chartID, dimensionName, setValues ] );
+	}, [ chartID, dimensionName, dimensionValue, report, setValues ] );
```
eugene-manuilov, line 107: We probably shouldn't use an object as a dependency item for a hook because it can break memoization. Let's better convert it to a string using the `JSON.stringify` function.

eugene-manuilov (changes requested):

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
felixarntz: @eugene-manuilov I tried that originally, but it didn't result in the expected behavior, while having it here does. I agree with you that in principle it should be in a `useEffect`, but there are some broader problems with our charts component - the `onReady` event fires several times. Since `useEffect` is decoupled from that, the chart might not be ready, so that didn't work.

Overall, the approach here is not the one we should go with for the long term, but it addresses the user-facing issue for this release. I've opened #2714 to address this in a better way.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
felixarntz: @eugene-manuilov I thought about that, but it shouldn't be an issue here? The report object returned by the `getReport` selector is always the same for the same set of request args.

felixarntz:
https://github.com/google/site-kit-wp/pull/2702/commits/2bad5a5d788096ac13190ace40890dfd87b50bfb addresses some feedback from Aaron. The breadcrumb UI should consistently use the same label for the top-level view.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
eugene-manuilov: @felixarntz see #2717 - it works correctly now (at least when you switch date ranges), however, that PR still requires some work. Let me know how it works on your end.

felixarntz:
@eugene-manuilov I've merged your branch in here - let me know if this is good to go now.

felixarntz:
@eugene-manuilov Any idea what's going on here? https://github.com/google/site-kit-wp/pull/2702/checks?check_run_id=1788039052 Maybe something changed in the Docker container configuration for our PHP 5.6 action?

felixarntz:
@eugene-manuilov I've pushed an extra commit https://github.com/google/site-kit-wp/pull/2702/commits/b6eb4048024a51374397ce7f787f81aa3ef9da29 to fix the PHP 5.6 test action, which was failing because the docker container was using PHP 8, apparently it recently changed. I fixed it by enforcing PHP 7.4 for the PHP-Scoper setup, which ensures a consistent set of versions there.

I'll merge this now, since it's pretty much the same as #2717, and this way I can build a new ZIP for testing. We can follow up more if needed.

Issue #2708, opened 2021-01-28
Fix All Traffic line chart alignment


felixarntz:
In the new All Traffic widget, the line chart (`UserCountGraph`) currently shows weirdly "indented", compared to the big number above (`TotalUserCount`). This is different from the original design where the graph on its left side aligns with the outer padding of the entire widget.

When I looked at the DOM, it seems the extra spacing is added by the Google charts library itself. Let's see if we can either get rid of it (preferred) or fix it via CSS in some other way (negative margin?).

Furthermore, the line chart's bottom end should be aligned with the bottom end of the `UserDimensionsPieChart`, i.e. the legends for both should pretty much appear on the same row.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The left end of the line chart in the new All Traffic widget should align with the left end of the big number and text above. It should no longer be indented further than the regular widget padding.
* The bottom end of the line chart in the new All Traffic widget should align with the bottom end of the pie chart, i.e. the legends should appear as if they were on the same line.

## Implementation Brief

* To fix the positioning of the chart horizontally, change the `UserCountGraph` `chartArea` param to:
```js
chartArea: {
	height: '90%',
	left: '-5%',
	top: '5%',
	width: '90%',
}
```
* To line them up vertically add the following CSS:
```scss
.googlesitekit-widget--analyticsAllTraffic__user-count-chart {

	@media (min-width: $bp-desktop) {
		margin-top: 1.4vw;
	}

	@media (min-width: $bp-xlarge) {
		margin-top: 0;
	}
}
```
* To maintain the dates appearing in the wider widget, we (sadly) cannot show the farthest-out dates; if we do, Google Charts will hide them. So let's move the dates we show on the horizontal axis below the chart to be one day ahead of the `startDate` and one day before the `endDate`.
* Merge #2712.

### Test Coverage

* No tests for this issue.

### Visual Regression Changes

* No VRT changes should result from these changes.

## QA Brief

* The charts should appear in-line with each other and the line chart of total users should not be "pushed in" to the center of the left side of the All Traffic Widget, eg:

<img width="1166" alt="Screenshot 2021-01-28 at 21 42 50" src="https://user-images.githubusercontent.com/90871/106202461-e55b3e80-61b1-11eb-9e57-0d0608b8cf6a.png">

## Changelog entry

* Ensure line chart in All Traffic is correctly left-aligned with the total user count.

felixarntz:
[...]

Issue #2699, opened 2021-01-27
Show more dates on All Traffic graph


felixarntz:
The new All Traffic widget graph only shows the start and end date on its x-axis. It should include more dates in between, similar to how our other graphs in the plugin behave.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The new All Traffic widget graph on the left should have more dates on its x-axis, not just start and end date. Regarding how many dates to show it should match the configuration of our other graphs in the plugin (e.g. the one on the Analytics module page).
    * Despite showing more dates like in existing graphs, the labels should remain in the date format that they currently have though (e.g. Dec 29), even though that is also different from our existing graphs (where it is e.g. 12/29/20).

## Implementation Brief

* Remove the `chartOptions.hAxis.ticks` property:
  https://github.com/google/site-kit-wp/blob/8b8340008cf7b687e1c1da118278322a0fe36ef4/assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserCountGraph.js#L77

### Test Coverage

* N/A

### Visual Regression Changes

* The user count graph should start showing more dates on the horizontal axis.

## QA Brief

* Set up the site kit plugin on your local;
* Activate and connect the Analytics module;
* Go to the dashboard page;
* Make sure that you see more than 2 dates on the horizontal axis of the user count graph.

## Changelog entry

* Show more than just two dates on the All Traffic widget's line chart.

tofumatt:
IB ✅

wpdarren:
### QA Update:  Fail ❌ 

There is some inconsistency with the displaying of the horizontal axis dates depending on the date period selected.

**For 7 days on desktop:** there are three dots (...) on the first date and then 6 dates are displayed (the 6th is quite close to the vertical axis)  

![Screenshot](https://user-images.githubusercontent.com/73545194/106763087-71b5a780-6671-11eb-9fb3-49d8acae3f0f.png)

**For 7 days on small screens (iOS and Android phones):** there are no dates on the horizontal axis. 

[Screenshot](https://user-images.githubusercontent.com/73545194/106763774-24860580-6672-11eb-891d-2d6e648a06c5.png)

**For 14 days on desktop:** this looks fine. 

**For 14 days on small screens (iOS and Android phones):**  there are no dates on the horizontal axis. 

[...]

PR #2712, opened 2021-01-28
Align UserCountGraph properly.

tofumatt:
## Summary

Addresses issue #2708.

## Checklist


assets/sass/widgets/_googlesitekit-widget-analyticsAllTraffic.scss
```diff
@@ -46,6 +46,17 @@
 		}
 	}
 
+	.googlesitekit-widget--analyticsAllTraffic__user-count-chart {
+
+		@media (min-width: $bp-desktop) {
+			margin-top: 1.4vw;
```
felixarntz, line 52: I don't think we should use `vw` here since it is dynamic on the viewport width. The margin shouldn't need to be different though based on how many pixels exactly your viewport is wide. I tried using `13px` here, which brings makes the two more aligned.

assets/sass/widgets/_googlesitekit-widget-analyticsAllTraffic.scss
```diff
@@ -46,6 +46,17 @@
...
+		}
+
+		@media (min-width: $bp-xlarge) {
+			margin-top: 0;
+		}
```
felixarntz, line 57: This is okay, but for this viewport (at least on my end), the pie chart is still slightly higher up, so they're not really aligned. I think you need to add `margin-top` to the `.googlesitekit-widget--analyticsAllTraffic__dimensions-chart` element for this viewport - `2px` did it for me. Let me know if that works on your end as well.

felixarntz (changes requested):
@tofumatt Nit-pick, but let's try to be pixel-perfect here - that might be the only way to get the alignment correctly. Unless there are some implications here that differ between our browsers/viewports, they are still not exactly aligned on my end with this.

Please also review the feedback I left above.

tofumatt:
Oddly, seems the "absolute" values passed to charts still account for viewport size, etc.—it's actually what the adjacent pie chart was using. I switched this component over to using absolute values and it seems to work much better:

tofumatt:
Regarding the date cutoff: from some searching on the web this seems to be a common issue with Google Charts as a whole, and difficult to work around. #2699 could solve it by allowing for "automatic" tick marks, but there's no obvious setting to control date cutoff.

Strange that it's cutting off for you but not me too. I don't think we can easily fix that issue though.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserCountGraph.js
```diff
@@ -100,12 +107,15 @@ UserCountGraph.chartOptions = {
 		startup: true,
 	},
 	curveType: 'function',
-	height: 340,
+	height: 410,
```
felixarntz, line 110: There's now a lot of extra whitespace at the bottom of the widget above the "Source" link, which looks odd. Likely due to increasing the height here? Can we achieve the vertical alignment without changing the overall chart height - why was this necessary?

See screenshot of extra whitespace:
Screen Shot 2021-01-28 at 3 36 11 PM

felixarntz (changes requested):
@tofumatt This looks great regarding the issue, the dates now show in their full version too. However, the solution now introduces another problem, unnecessary whitespace between the main widget UI and the source link. Could you check that?

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserCountGraph.js
tofumatt: This is as tight as I can get the margins without changing the grid/widget margins on things:

Screenshot 2021-01-29 at 13 32 24

The margins are accounting for most of the whitespace:

Screenshot 2021-01-29 at 13 45 02
Screenshot 2021-01-29 at 13 44 53

But I think my reductions in the margins is an improvement 🙂

tofumatt:
Sure thing! Should be all sorted now 🙂

felixarntz (approved):
@tofumatt Awesome!

I've pushed one more commit here (https://github.com/google/site-kit-wp/pull/2712/commits/32e2be50d6258fed36aedbae8ca85a9234b16889), which adjusts the height of the `PreviewBlock`s in `TotalUserCount` and `UserCountGraph`. Before, they were not matching the height of the fully loaded UI, resulting in ugly content jumps when selecting another pie slice during loading state.

With the values I specified now, it should result in the widget not changing its height in these loading states.

cc @aaemnnosttv

Issue #2714, opened 2021-01-28
Fix buggy All Traffic pie chart slice selection by introducing new React-ive GoogleChartV2 component


felixarntz:
#2644 originally had the primary goal of ensuring a pie slice in the new All Traffic pie chart remains selected also when changing the date range. When working on that fix (PRs #2684 and #2702), it became apparent that there are broader problems with how we use the charts library, especially in that new widget.

Potentially it's problematic that we use a single component for the three different tabs which should render entirely independent pie charts. Maybe we're having some incorrect hook logic in place that for example could make a certain `onReady` or `onSelect` callback for one pie chart apply to another.

Overall, while #2644 improved persistency of a selected pie slice in the All Traffic widget, selecting a pie slice is still failing sometimes, and sometimes, when switching tabs a random slice might be selected, although on tab switch no slice should ever be selected, since the `dimensionValue` must initially always be empty on tab switch. Another problem that sometimes occurs is that a pie slice might not show its selected slightly opaque border, even though it was just selected. All of these problems do not seem to appear consistently, but they happen.

After exploring more what's going on there, it looks like we should holistically revisit our inclusion of the Google Charts library and potentially replace it with a React wrapper, rather than wrapping it on our own.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a new, truly React-ive version of the `GoogleChart` component. Let's call it `GoogleChartV2` for now.
    * The component should wrap/rely on https://react-google-charts.com/, an already established wrapper library around Google charts.
    * The available props of `GoogleChartV2` should match the ones for the existing `GoogleChart` for now, except for the `selectedStats` and `singleStat` props.
    * In addition to an `onReady` prop (`ready` event), the new component should also support `onSelect` (`select` event), `onMouseOver` (`onmouseover` event), and `onMouseOut` (`onmouseout` event) props to specify individual callbacks that are then passed to the wrapped component (see https://react-google-charts.com/user-selection). We could also support their approach to pass any events as an array, but having dedicated props is a more intuitive API.
        * We currently (with the existing `GoogleChart` component) use `onReady` as a `ready` listener callback that then adds the other listener callbacks. With `GoogleChartV2`, this should no longer be the case; other events should be added via their dedicated props (see above).
    * The component should automatically include the correct loader (based on props) the same way the existing `GoogleChart` component does.
[...]

PR #2717, opened 2021-01-28
Update the user dimensions pie chart to use useEffect hook.

eugene-manuilov:
## Summary

Addresses issue #2644 

## Checklist

### DIFF
--- a/assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/index.js
+++ b/assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/index.js
@@ -20,7 +20,7 @@
  * WordPress dependencies
  */
 import { _x } from '@wordpress/i18n';
-import { useState, useEffect } from '@wordpress/element';
+import { useState, useEffect, Fragment } from '@wordpress/element';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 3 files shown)


felixarntz:
@eugene-manuilov When I test this, maintaining the selection works.

However, it seems to introduce another critical visual issue, where the pie chart labels are off, and legend labels aren't appearing sometimes. I see you removed the loading state here, why is that?

Screen Shot 2021-01-28 at 1 09 12 PM

Can you maybe reduce this PR to only moving the relevant code into a `useEffect`? Let's not change/refactor more than necessary at this point, since we're working against the release. We should invest more time then to resolve remaining problems via #2714.

eugene-manuilov:
@felixarntz all those changes are needed to make sure that the google chart component doesn't get unmounted and recreated when we select another date range or dimension. Let me see why styles get messy.

eugene-manuilov:
@felixarntz it should be good now. Can you check it on your end again?

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/index.js
```diff
@@ -155,71 +155,75 @@ function DashboardAllTrafficWidget() {
...
-							report={ pieChartReport }
+		<Fragment>
+			<div style={ { display: pieChartError ? 'block' : 'none' } }>
+				<ReportError moduleSlug="analytics" error={ pieChartError || {} } />
+			</div>
+			<div style={ { display: isZero && ! pieChartError ? 'block' : 'none' } }>
+				{ /* TODO: Replace with `props.WidgetReportZero` once legacy dashboard UI has been removed. */ }
+				<ReportZero moduleSlug="analytics" />
+			</div>
+			<div style={ { display: pieChartError || isZero ? 'none' : 'block' } }>
```
felixarntz, line 169: Instead of using the `style` attribute here, let's use the `googlesitekit-hidden` class.

felixarntz (commented):
@eugene-manuilov This still feels a bit hacky, relying on CSS so that we can still render out the components, but it works great for now - awesome!

Let's dive deeper into how we can fix the underlying JS rendering without using CSS to keep the component mounted at all times, via #2714.

Left just one comment worth addressing IMO to make it _a little_ less hacky.

felixarntz:
@eugene-manuilov e2e tests seem broken right now.

eugene-manuilov:
Yep, let me check it.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/index.js
eugene-manuilov: Ok, looks like these changes in the index.js file aren't needed. So, I have reverted it.

eugene-manuilov:
@felixarntz it's all green now.

felixarntz (approved):
Great work @eugene-manuilov!

This looks better than my original PR, so I'll merge it in there. However, it still doesn't work fully reliably. When testing this and switching tabs, active slices, and date ranges around a bunch, sometimes the pie slice that should be marked as selected still did not show up as selected - it happens inconsistently though.

It happened 2 or 3 times in 2 minutes of changing the UI around, that when I changed the date range, the pie slice for the current `dimensionValue` no longer looked as if it was selected. (The graph on the left is always correct, but that was already the case before this.)

Issue #2579, opened 2021-01-05
Ensure the toggle to allow Site Kit to place Analytics tracking code on a users site remains once they've disconnected Tag Manager (with an existing Analytics snippet tag)


jamesozzie:
## Bug Description

At present one typical Site Kit setup scenario may be where a user first connects Analytics, allowing Site Kit to insert the tracking snippet. They may then connect the Tag Manager module, with may contain an existing Analytics tag, which Site Kit then switches to. 

If a user then disconnects Tag Manager the plugin reverts to the Analytics module UI property alone. While the toggle to "[Let Site Kit place code on your site](https://i.imgur.com/5S5sPME.png) may have been initially on before connecting Tag Manager the toggle is then off, resulting in no Analytics data being captured. 

https://i.imgur.com/Fw9IkkT.png

Some possible solutions to ensure users have Analytics tracking code remaining on their site (if chosen) include:

1. **Adding a notice when disconnecting Tag Manager**: This must also account for users who disable the Tag Manager configuration to allow Site Kit to insert Tag Manager code. 
2. **Ensure the toggle to allow Site Kit insert Analytics code is on by default**: One consideration with this solution is the user may have intentionally disabled the toggle before activating Tag Manager. This would be then reverting to a configuration the user chose not to select. 

## Steps to reproduce

1. Connect the Analytics module and allow SK to insert the Analytics tracking snippet
2. Connect the Tag Manager module to a container with an existing Analytics Universal Tracking snippet 
3. Site Kit switches to the Tag Manager placed UA snippet
4. Disconnect the Tag Manager module
5. The Analytics snippet once again becomes active although the default configuration is that Site Kit does not insert the Analytics tracking snippet, resulting in a user potentially not have any Analytics data

## Additional Context

 - [Video demonstrating issue](https://recordit.co/Cj5QXfLYY3)
 - While there have been no support topics open in relation to this there is a possibility of users not being aware of the status of their Analytics snippet. 

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Site Kit should store a flag option for the Analytics module in case the Analytics `useSnippet` option gets changed from `true` to `false` due to the user having enabled Tag Manager with a matching Analytics tag.
* Then, if the Tag Manager module later gets disconnected, in case that new Analytics flag is enabled, `useSnippet` should be "restored" to `true`.

#### Analytics module:

[...]

PR #2730, opened 2021-01-30
Update the Analytics module to support a new canUseSnippet setting.

benbowler:
## Summary

Addresses issue #2579

## Relevant technical choices

I used `getLiveContainerAnalyticsPropertyID` instead of `getLiveContainerAnalyticsTag` as it returns the propertyID from the tag for us.

## Checklist

### DIFF
--- a/includes/Modules/Analytics/Settings.php
+++ b/includes/Modules/Analytics/Settings.php
@@ -121,9 +121,22 @@ class Settings extends Module_Settings implements Setting_With_Owned_Keys_Interf
 				 * @param bool $adsense_linked Null by default, will fallback to the option value if not set.
 				 */
 				$adsense_linked = apply_filters( 'googlesitekit_analytics_adsense_linked', null );
-
 				if ( is_bool( $adsense_linked ) ) {
 					$option['adsenseLinked'] = $adsense_linked;
+				}
+
+				/**
+				 * Filters the state of the can use snipped setting.
+				 *
+				 * This filter exists so that useSnippet can be restored to true when the Tag Manager module
+				 * is disconnected, ensuring the analytics snippet is always included.
+				 *
+				 * @since n.e.x.t
+				 * @param bool $can_use_snippet Null by default, will fallback to the option value if not set.
+				 */
+				$can_use_snippet = apply_filters( 'googlesitekit_analytics_can_use_snippet', null );
+				if ( is_bool( $can_use_snippet ) ) {
+					$option['canUseSnippet'] = $can_use_snippet;
 				}
 
 				return $option;
@@ -165,6 +178,7 @@ class Settings extends Module_Settings implements Setting_With_Owned_Keys_Interf
 			'propertyID'            => '',
 			'trackingDisabled'      => array( 'loggedinUsers' ),
 			'useSnippet'            => true,
+			'canUseSnippet'         => true,
 		);
 	}
 
--- a/assets/js/modules/analytics/components/common/UseSnippetSwitch.js
+++ b/assets/js/modules/analytics/components/common/UseSnippetSwitch.js
@@ -33,6 +33,7 @@ const { useSelect, useDispatch } = Data;
 
 export default function UseSnippetSwitch() {
 	const useSnippet = useSelect( ( select ) => select( STORE_NAME ).getUseSnippet() );
+	const canUseSnippet = useSelect( ( select ) => select( STORE_NAME ).getCanUseSnippet() );
 
 	const { setUseSnippet } = useDispatch( STORE_NAME );
 	const onChange = useCallback( () => {
@@ -51,10 +52,12 @@ export default function UseSnippetSwitch() {
 				checked={ useSnippet }
 				onClick={ onChange }
 				hideLabel={ false }
+				disabled={ ! canUseSnippet }
 			/>
 			<p>
-				{ useSnippet && __( 'Site Kit will add the code automatically.', 'google-site-kit' ) }
-				{ ! useSnippet && __( 'Site Kit will not add the code to your site.', 'google-site-kit' ) }
+				{ ! canUseSnippet && __( 'The code is controlled by the Tag Manager module.', 'google-site-kit' ) }
+				{ canUseSnippet && useSnippet && __( 'Site Kit will add the code automatically.', 'google-site-kit' ) }
+				{ canUseSnippet && ! useSnippet && __( 'Site Kit will not add the code to your site.', 'google-site-kit' ) }
 			</p>
 		</div>
 	);
--- a/assets/js/modules/analytics/components/settings/SettingsView.js
+++ b/assets/js/modules/analytics/components/settings/SettingsView.js
@@ -40,6 +40,7 @@ export default function SettingsView() {
 	const internalWebPropertyID = useSelect( ( select ) => select( STORE_NAME ).getInternalWebPropertyID() );
 	const profileID = useSelect( ( select ) => select( STORE_NAME ).getProfileID() );
 	const useSnippet = useSelect( ( select ) => select( STORE_NAME ).getUseSnippet() );
+	const canUseSnippet = useSelect( ( select ) => select( STORE_NAME ).getCanUseSnippet() );
 	const anonymizeIP = useSelect( ( select ) => select( STORE_NAME ).getAnonymizeIP() );
 	const trackingDisabled = useSelect( ( select ) => select( STORE_NAME ).getTrackingDisabled() ) || [];
 	const hasExistingTag = useSelect( ( select ) => select( STORE_NAME ).hasExistingTag() );
@@ -103,7 +104,8 @@ export default function SettingsView() {
 						{ __( 'Analytics Code Snippet', 'google-site-kit' ) }
 					</h5>
 					<p className="googlesitekit-settings-module__meta-item-data">
-						{ useSnippet && __( 'Snippet is inserted', 'google-site-kit' ) }
+						{ ! canUseSnippet && __( 'The code is controlled by the Tag Manager module.', 'google-site-kit' ) }
+						{ canUseSnippet && useSnippet && __( 'Snippet is inserted', 'google-site-kit' ) }
 						{ ( ! useSnippet && ! hasExistingTag ) && __( 'Snippet is not inserted', 'google-site-kit' ) }
 						{ ( ! useSnippet && hasExistingTag ) && __( 'Inserted by another plugin or theme', 'google-site-kit' ) }
 					</p>
--- a/assets/js/modules/analytics/datastore/base.js
+++ b/assets/js/modules/analytics/datastore/base.js
@@ -32,6 +32,7 @@ const baseModuleStore = Modules.createModuleStore( 'analytics', {
 		'propertyID',
 		'internalWebPropertyID',
 		'useSnippet',
+		'canUseSnippet',
 		'trackingDisabled',
 		'ownerID',
 	],
--- a/assets/js/modules/analytics/hooks/useExistingTagEffect.js
+++ b/assets/js/modules/analytics/hooks/useExistingTagEffect.js
@@ -33,7 +33,7 @@ import { MODULES_TAGMANAGER } from '../../tagmanager/datastore/constants';
 const { useSelect, useDispatch } = Data;
 
 export default function useExistingTagEffect() {
-	const { setAccountID, selectProperty, setUseSnippet } = useDispatch( STORE_NAME );
+	const { setAccountID, selectProperty } = useDispatch( STORE_NAME );
 	const gtmModuleActive = useSelect( ( select ) => select( CORE_MODULES ).isModuleActive( 'tagmanager' ) );
 
 	const {
@@ -86,7 +86,6 @@ export default function useExistingTagEffect() {
 			// GTM container has GA tag and user has access to it, force select it.
 			setAccountID( gtmAnalyticsAccountID );
 			selectProperty( gtmAnalyticsPropertyID );
-			setUseSnippet( false );
 		}
 	}, [
 		existingTag,
--- a/includes/Modules/Analytics/Tag_Guard.php
+++ b/includes/Modules/Analytics/Tag_Guard.php
@@ -30,7 +30,7 @@ class Tag_Guard extends Module_Tag_Guard {
 	 */
 	public function can_activate() {
 		$settings = $this->settings->get();
-		return ! empty( $settings['useSnippet'] ) && ! empty( $settings['propertyID'] );
+		return apply_filters( 'googlesitekit_analytics_can_use_snippet', ! empty( $settings['useSnippet'] ) && ! empty( $settings['propertyID'] ) );
 	}
 
 }


eugene-manuilov (dismissed):
@benbowler the PR looks good to me, but it still requires a little bit of work. We need to make sure that the Analytics useSnippet element is unblocked when I change Tag Manager settings and select a different account or container without GA integration.

To achieve this, we need to update Tag Manager's `gaPropertyID` and `gaAMPPropertyId` settings when we select a web or an amp container. See a snippet in my comment: https://github.com/google/site-kit-wp/issues/2579#issuecomment-776046145 We need to do it in `WebContainerSelect` and `AMPContainerSelect` containers.

benbowler:
@aaemnnosttv I ended up with an update to the IB implementation between what we talked about and @eugene-manuilov's suggestions. Let me know how it works for you.

aaemnnosttv (changes requested):
Thanks @benbowler – this is off to a good start I think but needs a little refinement. See my comments below.

One thing I'm wondering about is whether or not we should ever save the value of `canUseSnippet` since it is essentially a computed setting; the saved value seems irrelevant.  
cc: @felixarntz

benbowler:
Hey Evan, I've addressed those points and I feel like it is in a good position now.

assets/js/modules/tagmanager/hooks/useExistingTagEffect.js
```diff
@@ -33,14 +34,29 @@ export default function useExistingTagEffect() {
...
+			// prevent duplicate measurement.
+			// The stored property ID is then used to re-enable the Analytics snippet if
+			// Tag Manager is disconnected in future.
+			if ( singleAnalyticsPropertyID !== undefined && analyticsModuleActive ) {
+				// Set the GA property ID in the Tag Manager store.
+				await setGAPropertyID( singleAnalyticsPropertyID === null ? '' : singleAnalyticsPropertyID );
+			}
 		} )();
-	}, [ hasExistingTag, existingTag, hasExistingTagPermission, existingTagPermission ] );
+	}, [ hasExistingTag, existingTag, hasExistingTagPermission, existingTagPermission, singleAnalyticsPropertyID, analyticsModuleActive ] );
```
aaemnnosttv, line 61: I don't think the effect needs to change here since it isn't related to existing tags. In GA, we do however treat a GA tag in a GTM container as an existing tag, but this is a bit different.

We essentially need a `useGAPropertyIDEffect` that synchronizes the current value based on the selected containers.
This is fairly straightforward since the `singleAnalyticsPropertyID` is already based on container selections.

```js
function useGAPropertyIDEffect() {
	const singleAnalyticsPropertyID = useSelect( ( select ) => select( STORE_NAME ).getSingleAnalyticsPropertyID() );
	const { setGAPropertyID } = useDispatch( STORE_NAME );

	useEffect( () => {
		if ( singleAnalyticsPropertyID !== undefined ) {
			setGAPropertyID( singleAnalyticsPropertyID || '' );
		}
	}, [ singleAnalyticsPropertyID ] );
}
```
This would then be called in the same places where Tag Manager's `useExistingTagEffect` is called (setup + settings). That way, the setting is kept in sync with the current `singleAnalyticsPropertyID` and will be saved automatically when settings are saved.

includes/Modules/Analytics/Settings.php
```diff
@@ -121,11 +121,24 @@ function ( $option ) {
...
+				/**
+				 * Filters the state of the can use snipped setting.
+				 *
+				 * This filter exists so that useSnippet can be restored to true when the Tag Manager module
+				 * is disconnected, ensuring the Analytics snippet is always included.
+				 *
+				 * @since n.e.x.t
+				 * @param bool $can_use_snippet Null by default, will fallback to the option value if not set.
+				 */
+				$can_use_snippet = apply_filters( 'googlesitekit_analytics_can_use_snippet', null );
```
aaemnnosttv, line 137: As mentioned in my last review, I don't think it makes sense for this to use the saved value (that's more of a side effect of including it here). Let's default the value to `true` here so that it always takes precedent, unless explicitly filtered to `false`.

includes/Modules/Tag_Manager.php
```diff
@@ -638,4 +640,28 @@ private function register_tag() {
...
+	private function can_analytics_use_snippet( $original_value ) {
+		$settings = $this->get_settings()->get();
+
+		// The GAPropertyID is set by Tag Manager when it's own snippet shares the GA ID
+		// of an existing, active, Analytics tag.
+		// Returning false here disables the useSnippet setting in the Analytics setting
+		// and shows a message to the user explaining that the snippet is now controlled
+		// by Tag Manager.
+		if ( ! empty( $settings['GAPropertyID'] ) && $settings['useSnippet'] ) {
+			return false;
```
aaemnnosttv, line 661: The comment here is a little off. The `gaPropertyID` will be set whenever it is configured with a GA property in its container(s). We require that this property ID is the same for both containers (if they use AMP) and that it matches the configured property ID of Analytics, if it is already set up.
So the important thing to mention here is that this disables the Analytics snippet if there is a GA tag in the configured containers, and the GTM snippet is enabled.

aaemnnosttv (commented):
Thanks @benbowler – I think we're getting much closer here. Just a few things to address remain.

assets/js/modules/tagmanager/datastore/settings.test.js
```diff
@@ -99,6 +96,12 @@ describe( 'modules/tagmanager settings', () => {
 	beforeEach( () => {
 		registry = createTestRegistry();
 		registry.dispatch( CORE_SITE ).receiveSiteInfo( {} );
+
+		// Provide an inactive Analytics module so that we aren't testing the singleAnalyticsPropertyID in every test.
+		provideModules( registry, [ {
+			slug: 'analytics',
+			active: false,
```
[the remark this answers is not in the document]
benbowler: Changing the default status of the analytics module in`assets/js/googlesitekit/modules/datastore/fixtures.json` causes a couple of other suites to fail so it's worth adding that to another ticket, especially due to the time already logged against this one.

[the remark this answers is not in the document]
benbowler: I can't remove these lines without the tests failing.

includes/Modules/Tag_Manager/Settings.php
```diff
@@ -80,8 +80,7 @@ protected function get_default() {
 			'internalContainerID'    => '',
 			'internalAMPContainerID' => '',
 			'useSnippet'             => true,
-			'gaAMPPropertyID'        => '',
-			'gaPropertyID'           => '',
+			'GAPropertyID'           => '',
```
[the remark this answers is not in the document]
benbowler: @aaemnnosttv If I change this I get `getGAPropertyID` and `setGAPropertyID` as `undefined` throughout the code.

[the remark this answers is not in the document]
benbowler: Is there another acronym list that the store uses to know what should be uppercased?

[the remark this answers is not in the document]
benbowler: I've been looking through the helpers here: `assets/js/googlesitekit/data/transform-case.js` and I think this acronym gets miss translated somewhere.

benbowler:
Hey @aaemnnosttv, I've updated based on your comments. There is an issue with the GA acronym in the auto-generated actions and selectors, I've left comments inline. Let me know your thoughts.

assets/js/modules/tagmanager/datastore/settings.test.js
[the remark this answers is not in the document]
aaemnnosttv: Thanks for trying that. I'll open an issue for updating the fixture 👍

includes/Modules/Tag_Manager/Settings.php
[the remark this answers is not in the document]
aaemnnosttv: The case is transformed to camel case when generating selectors and actions in `createSettingsStore`. It currently doesn't consider our known acronyms though so we need to do some manual renaming.

aaemnnosttv:
@benbowler – this is looks really close but it seems E2E and JS tests are currently failing. I'm guessing this is probably about the naming issue:

> There is an issue with the GA acronym in the auto-generated actions and selectors, I've left comments inline.

Yes, the settings store that generates the actions and selectors for all of the settings only applies a simple transformation to camel case the setting slug when prefixed with `get`/`set`. We need to manually rename the parts that need it to conform to our naming conventions. We already have this for the `ampContainerID` setting for GTM so you can simply extend that, which is what I was referring to in https://github.com/google/site-kit-wp/pull/2730#discussion_r577502768 https://github.com/google/site-kit-wp/pull/2730#discussion_r579073892 🙂 

Let me know if you have any other questions 👍

benbowler:
Hey @aaemnnosttv, I tracked down the mapping between ga -> GA in `assets/js/modules/tagmanager/datastore/base.js`. 

Tests are now passing, should be ready to go.

assets/js/modules/tagmanager/hooks/useGAPropertyIDEffect.test.js
```diff
@@ -0,0 +1,64 @@
...
+		registry.dispatch( STORE_NAME ).receiveGetExistingTag( null );
+
+		// Provide an activated Analytics module.
+		provideModules( registry, [ {
+			slug: 'analytics',
+			active: true,
+		} ] );
+	} );
+
+	it( 'sets the gaPropertyID when property ID exists and Analytics is active', async () => {
```
aaemnnosttv, line 46:
```suggestion
	it( 'sets the gaPropertyID with the current detected singular property ID in selected containers', async () => {
```

includes/Modules/Analytics/Settings.php
```diff
@@ -121,11 +121,24 @@ function ( $option ) {
...
 				}
 
+				/**
+				 * Filters the state of the can use snipped setting.
+				 *
+				 * This filter exists so that useSnippet can be restored to true when the Tag Manager module
+				 * is disconnected, ensuring the Analytics snippet is always included.
+				 *
+				 * @since n.e.x.t
+				 * @param bool $can_use_snippet Null by default, will fallback to the option value if not set.
```
aaemnnosttv, line 135: This description is slightly inaccurate now.
```suggestion
				 * @param bool $can_use_snippet Whether or not `useSnippet` can control snippet output. Default: `true`.
```

includes/Modules/Tag_Manager.php
```diff
@@ -93,6 +93,8 @@ public function register() {
 
 		// Tag Manager tag placement logic.
 		add_action( 'template_redirect', $this->get_method_proxy( 'register_tag' ) );
+		// Filter the Analytics useSnippet value.
```
aaemnnosttv, line 96:
```suggestion
		// Filter the Analytics `canUseSnippet` value.
```

includes/Modules/Tag_Manager.php
```diff
@@ -638,4 +640,25 @@ private function register_tag() {
 		}
 	}
 
+	/**
+	 * Can the Analytics module insert it's own snippet.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param boolean $original_value Original value of useSnippet setting.
+	 *
```
aaemnnosttv, line 649: This is a bit inconsistent in some places, but there shouldn't be a line between `@param` and `@return` annotations. (PHP linting is not as thorough as ESLint about this _yet_ 😄 )
```suggestion
```

includes/Modules/Tag_Manager.php
```diff
@@ -638,4 +640,25 @@ private function register_tag() {
 		}
 	}
 
+	/**
+	 * Can the Analytics module insert it's own snippet.
```
aaemnnosttv, line 644: Similar to JS, the doc block should start with a 3p verb.
```suggestion
	 * Filters whether or not the Analytics module's snippet should be controlled by its `useSnippet` setting.
```

includes/Modules/Tag_Manager.php
```diff
@@ -93,6 +93,8 @@ public function register() {
 
 		// Tag Manager tag placement logic.
 		add_action( 'template_redirect', $this->get_method_proxy( 'register_tag' ) );
+		// Filter the Analytics useSnippet value.
+		add_action( 'googlesitekit_analytics_can_use_snippet', $this->get_method_proxy( 'can_analytics_use_snippet' ) );
```
aaemnnosttv, line 97: It would be great if we could add a quick test for this in `Tag_ManagerTest` to cover all the combinations of `gaPropertyID` and `$settings['useSnippet']`.

aaemnnosttv (commented):
Thanks @benbowler – this is very close, just a few (mostly doc) tweaks, and small things to clean up. Most of it already has suggestions ready to go as well.

assets/js/modules/tagmanager/hooks/useExistingTagEffect.js
```diff
@@ -33,8 +33,9 @@ export default function useExistingTagEffect() {
 	const existingTag = useSelect( ( select ) => select( STORE_NAME ).getExistingTag() );
 	const existingTagPermission = useSelect( ( select ) => select( STORE_NAME ).getTagPermission( existingTag ) );
 	const hasExistingTagPermission = useSelect( ( select ) => select( STORE_NAME ).hasExistingTagPermission() );
-	// Set the accountID and containerID if there is an existing tag.
```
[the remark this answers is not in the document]
benbowler: I've reverted this file and the test as they are no longer edited from develop.

assets/js/modules/tagmanager/hooks/useExistingTagEffect.test.js
```diff
@@ -52,6 +58,14 @@ describe( 'useExistingTagEffect', () => {
...
 		registry.dispatch( STORE_NAME ).setInternalContainerID( firstContainer.containerId );
 
+		// Loop through each container and set up the relevant tag.
+		containers.forEach( ( container ) => {
+			// eslint-disable-next-line sitekit/acronym-case
+			const liveContainerVersion = factories.buildLiveContainerVersionWeb( { accountID: container.accountId, propertyID: container.publicId } );
+			// eslint-disable-next-line sitekit/acronym-case
+			registry.dispatch( STORE_NAME ).receiveGetLiveContainerVersion( liveContainerVersion, { accountID: container.accountId, internalContainerID: container.containerId } );
+		} );
+
```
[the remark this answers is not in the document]
benbowler: I've reverted this also.

assets/js/modules/tagmanager/hooks/useGAPropertyIDEffect.test.js
```diff
@@ -0,0 +1,64 @@
...
+		// Set settings to prevent fetch in resolver.
+		registry.dispatch( STORE_NAME ).receiveGetSettings( {} );
+		// Set set no existing tag.
+		registry.dispatch( STORE_NAME ).receiveGetExistingTag( null );
+
+		// Provide an activated Analytics module.
+		provideModules( registry, [ {
+			slug: 'analytics',
+			active: true,
+		} ] );
```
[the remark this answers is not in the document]
benbowler: Removed it and it's looking good!

includes/Modules/Tag_Manager.php
benbowler: @aaemnnosttv if you're able to add this that would be great as my time remaining is limited.

aaemnnosttv: Will do! 👍

aaemnnosttv: It ended up being a little less quick than I thought but should be good now 😄

aaemnnosttv:
@googlebot I consent.

aaemnnosttv (approved):
Pending the clarification about the requirement for the note mentioned in the ACs, but perhaps this is no longer needed. Otherwise, I helped get this across the finish line so it's 👍  from me 😄  Nice work @benbowler !

Assigning to @felixarntz for a (hopefully) final review.

felixarntz (approved):
@aaemnnosttv LGTM - this should be good without the notice indeed.

Issue #2612, opened 2021-01-13
Fix hacky approach for detecting whether a widget is inactive


felixarntz:
As part of the dynamic widgets API, we want to make it simple for developers to implement their widgets without calling specific Site Kit datastore actions to indicate a certain state of the widgets. #2252 focused on introducing special `Widget*` wrapper components for various special states so that the registered widget component simply needs to return that component. This is further simplified since these components are passed to the widget component as props (in a scoped version), so developers don't need to import it.

For a couple months now, we've had some _terribly_ hacky and ugly logic in `WidgetAreaRenderer` which inspects the return value of the registered widget component by calling its render function directly - while this works, it goes against React best practices, where overall inspecting child components in such a way is not intended. This hack furthermore causes some issues which (among others) prevent us from updating to a newer version of React.

We should replace this hack with a proper mechanism to detect whether a widget is inactive. One possible approach would be to follow along the pattern that #2252 introduced:
* We could introduce a `WidgetNull` component and pass it to every registered widget component.
* The component would dispatch a new `setWidgetActive( false )` action on mount and `setWidgetActive( true )` on unmount, similar to how e.g. `WidgetReportZero` sets widget state. The component would technically just return `null`, i.e. render nothing.
* `WidgetAreaRenderer` could then recognize from the datastore for each widget whether it is active or inactive.
* This would require registered components to return `<WidgetNull />` instead of `null`, which maybe is a bit unintuitive. But it could still be a good solution, we can always document it.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a new `Null` component in `assets/js/components` which simply returns `null`.
* There should be a new `WidgetNull` component in `assets/js/googlesitekit/widgets/components` that requires a `widgetSlug` prop.
    * It should include a `useEffect` hook that calls `setWidgetState( widgetSlug, Null, {} )`, and `unsetWidgetState` with the same arguments as cleanup function. (This is similar to e.g. `WidgetActivateModuleCTA` or `WidgetReportZero`.)
    * It should render `<Null />`.
* The `getWidgetComponentProps` function should be expanded to also include a scoped version of `WidgetNull` in the returned object (which is then passed to every widget's component).
* There should be a new function `separateNullWidgets( widgets, widgetStates )` in `assets/js/googlesitekit/widgets/util`:
[...]

PR #2731, opened 2021-01-31
Enhancement/2612 detecting widget inactivity

ivankruchkoff:
## Summary

Addresses issue #2612 

## Checklist


assets/js/googlesitekit/widgets/components/URLSearchWidget.js
```diff
@@ -30,9 +30,10 @@ import { CORE_SITE } from '../../datastore/site/constants';
 import { CORE_LOCATION } from '../../datastore/location/constants';
 import Button from '../../../components/Button';
 import PostSearcherAutoSuggest from '../../../components/PostSearcherAutoSuggest';
+import WidgetNull from './WidgetNull';
 const { useSelect, useDispatch } = Data;
 
-function URLSearchWidget( { Widget } ) {
+function URLSearchWidget( { Widget = WidgetNull } = {} ) {
```
aaemnnosttv, line 36: Looks you've got a working typo here. Also, no need to default to empty object since props will always be passed.
```suggestion
function URLSearchWidget( { Widget, WidgetNull } ) {
```

assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js
```diff
@@ -141,6 +130,14 @@ export default function WidgetAreaRenderer( { slug, totalAreas } ) {
...
 				</Row>
 			</div>
+
+			<div className="googlesitekit-widget-area-widgets">
+				{ inactiveWidgets.map( ( widget ) => (
+					<div className={ HIDDEN_CLASS } key={ widget.slug }>
+						<widget.Component />
+					</div>
+				) ) }
+			</div>
```
aaemnnosttv, line 140: The whole `div` here should be hidden, it doesn't need any other class on it. We don't need to render an extra div around each widget.

assets/js/googlesitekit/widgets/components/WidgetRenderer.js
```diff
@@ -43,7 +44,7 @@ const WidgetRenderer = ( { slug, gridClassName, OverrideComponent } ) => {
 	const widgetComponentProps = useMemo( () => getWidgetComponentProps( slug ), [ slug ] );
 
 	if ( ! widget ) {
-		return null;
+		return <WidgetNull widgetSlug={ slug } />;
```
aaemnnosttv, line 47: This widget is already provided in `widgetComponentProps`, so no need to re-import it here and pass the slug, since this is already passed as well.

assets/js/googlesitekit/widgets/util/separate-null-widgets.js
```diff
@@ -0,0 +1,31 @@
...
+export function separateNullWidgets( widgets, widgetStates ) {
+	return widgets.reduce( ( separated, widget ) => {
+		if ( 'WidgetNull' !== widget.Component().type.name ) {
+			// @TODO figure out why we have a null state in widgetStates when the component is not WidgetNull
+			separated.activeWidgets.push( widget );
+		} else if ( undefined === widgetStates[ widget.slug ] || null === widgetStates[ widget.slug ] ) {
+			separated.inactiveWidgets.push( widget );
+		} else {
+			separated.activeWidgets.push( widget );
+		}
```
aaemnnosttv, line 28: These comparisons should only use the `widgetState` for the widget rather than touching the `widget.Component`. Keep in mind that the component used by the widget's state for `WidgetNull` will be `Null`. We need to import that here and simply check `widgetState[ slug ]?.Component === Null` where `Null` is the actual component.

For inactive, I think we can simply check `! widgetStates[ widget.slug ]` since any state would be set as an object here.

assets/js/util/when-active.js
```diff
@@ -54,14 +55,14 @@ export default function whenActive( { moduleName, FallbackComponent = null, Inco
 			// eslint-disable-next-line react-hooks/rules-of-hooks
 			const module = useSelect( ( select ) => select( CORE_MODULES ).getModule( moduleName ) );
 
-			// Return null if the module is not loaded yet or doesn't exist.
+			// Return <WidgetNull /> if the module is not loaded yet or doesn't exist.
 			if ( ! module ) {
-				return null;
+				return <WidgetNull widgetSlug={ moduleName } />;
```
aaemnnosttv, line 60: `widgetSlug` is never needed by the component which is passed via props since we pre-provide and "lock" this in the given component already. We have no way of knowing what the widget slug is anyways.
```suggestion
				return <WidgetNull />;
```

assets/js/util/when-active.js
```diff
@@ -45,7 +46,7 @@ const { useSelect } = Data;
  * @param {WPComponent|null} [options.IncompleteComponent] Optional. Fallback component to render when the module is active but not connected.
  * @return {Function} Enhancing function.
  */
-export default function whenActive( { moduleName, FallbackComponent = null, IncompleteComponent = null } ) {
+export default function whenActive( { moduleName, FallbackComponent = WidgetNull, IncompleteComponent = null } ) {
```
aaemnnosttv, line 49: We should rely on `WidgetNull` from props everywhere except for within `getWidgetComponentProps`. Since this is an HOC for use with widgets, we can assume it will be provided here too.
```suggestion
export default function whenActive( {
	moduleName,
	WidgetNull,
	FallbackComponent = WidgetNull,
	IncompleteComponent = null,
} ) {
```

aaemnnosttv (commented):
Looking good @ivankruchkoff ! I left a few comments that should help clarify what's left here.

assets/js/googlesitekit/widgets/components/URLSearchWidget.js
ivankruchkoff: It wasn't so much a typo, the props weren't being passed from /assets/js/googlesitekit/widgets/register-defaults.js, but I think that was more a symptom of the issue we were debugging. No need to destructure WidgetNull since we don't render it.

assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js
ivankruchkoff: Updated.

assets/js/googlesitekit/widgets/components/WidgetRenderer.js
ivankruchkoff: This is already resolved.

assets/js/googlesitekit/widgets/util/separate-null-widgets.js
ivankruchkoff: Removed.

ivankruchkoff:
@googlebot I consent.

aaemnnosttv:
@googlebot I consent.

aaemnnosttv (commented):
Thanks @ivankruchkoff – this was quite close but I think some of my suggestions from before may not have been clear.

Following our discussion earlier, I've pushed a few commits on top of where you left off and I think this is a good place to stop until we sync with @felixarntz as to where this is at.

assets/js/googlesitekit/widgets/components/WidgetNull.test.js
```diff
@@ -0,0 +1,30 @@
...
+ * Internal dependencies
+ */
+import WidgetNull from './WidgetNull';
+import { render } from '../../../../../tests/js/test-utils';
+
+describe( 'WidgetNull', () => {
+	it( 'should return an empty element', async () => {
+		const { container } = render( <WidgetNull widgetSlug="TestWidget" /> );
+		expect( container ).toBeEmptyDOMElement();
+	} );
```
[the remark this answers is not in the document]
ivankruchkoff: Good call, led to this: https://github.com/google/site-kit-wp/pull/2731/commits/3d283ba50b964b06e0591114483b2f59e9012d5f

assets/js/googlesitekit/widgets/datastore/widgets.js
```diff
@@ -316,6 +317,20 @@ export const selectors = {
...
+	},
+
+	getActiveWidgets( state, widgetAreaSlug ) {
+		return selectors.getWidgets( state, widgetAreaSlug )
+			.filter( ( widget ) => selectors.isWidgetActive( state, widget.slug ) );
+	},
+
+	getWidgetStates( state ) {
+		return state.widgetStates;
+	},
```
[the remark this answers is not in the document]
ivankruchkoff: Updated in https://github.com/google/site-kit-wp/pull/2731/commits/63e4502d5a750a72ce9ab7b8b66cede126f7998b

assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js
```diff
@@ -89,7 +61,7 @@ export default function WidgetAreaRenderer( { slug, totalAreas } ) {
 	} );
 
 	// Render all widgets.
-	const widgetsOutput = activeWidgets.map( ( widget, i ) => (
+	const widgetsOutput = widgets.map( ( widget, i ) => (
```
felixarntz, line 64: Passing `widgets` here makes sense given the new approach, but we'll also need to pass `widgets` in the two utility function calls above, instead of passing `activeWidgets`. Otherwise the layout will be messed up, since the utility functions work based on indexes, and those are of course different between `widgets` and `activeWidgets`.

We'll need to consistently act on all widgets now (i.e. the `widgets` variable), and to do that we'll need to slightly expand the `getWidgetLayout` and `combineWidgets` utility functions to also support inactive widgets:
* The function parameters need to be renamed from `activeWidgets` to `widgets` to clarify intent.
* `getWidgetLayout` also needs to receive `widgetStates` to be able to check which widgets are inactive. For those, it should simply set the `classNames` entry to `null`, the `columnWidths` entry to `0`, and the `rowIndexes` to the current `rowIndex`. The other logic should be entirely bypassed for them, since it's only relevant for active widgets.
* `combineWidgets` should be pretty much already compatible, since it already has a clause that will set the `HIDDEN_CLASS` for any widgets that have their `classNames` entry set to `null`. I think here we only need to update the parameter name and its documentation, as mentioned above.
* We should add Jest test coverage for these new pieces of logic.

assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js
```diff
@@ -101,6 +73,15 @@ export default function WidgetAreaRenderer( { slug, totalAreas } ) {
...
 	) );
 
+	// Here we render the bare output as it is guaranteed to render empty.
+	// This is important compared to returning `null` so that the area
+	// can maybe render later if conditions change for widgets to become active.
+	// Returning `null` here however would have the side-effect of making
+	// all widgets active again, which is why we must return the "null" output.
+	if ( ! activeWidgets.length ) {
+		return widgetsOutput;
+	}
```
felixarntz, line 83: Related to my above comment: This is the only place where we should be using `activeWidgets` in this component; since we don't want to render any surrounding UI in that case.

assets/js/googlesitekit/widgets/components/WidgetRenderer.js
```diff
@@ -24,26 +24,26 @@ import PropTypes from 'prop-types';
...
 import { getWidgetComponentProps } from '../util';
 import { HIDDEN_CLASS } from '../util/constants';
 
 const { useSelect } = Data;
 
 const WidgetRenderer = ( { slug, gridClassName, OverrideComponent } ) => {
 	const widget = useSelect( ( select ) => select( STORE_NAME ).getWidget( slug ) );
-
-	const widgetComponentProps = useMemo( () => getWidgetComponentProps( slug ), [ slug ] );
+	const widgetComponentProps = getWidgetComponentProps( slug );
```
felixarntz, line 42: Not worth keeping the `useMemo` here?

assets/js/googlesitekit/widgets/datastore/widgets.js
```diff
@@ -316,6 +317,49 @@ export const selectors = {
...
+	 *
+	 * @since n.e.x.t
+	 * @private
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {Object} Widget states, keyed by slug, with a value containing the component and metadata.
+	 */
+	getWidgetStates( state ) {
+		return state.widgetStates;
+	},
```
felixarntz, line 362: Nit-pick, but let's move this up, right below `getWidgetState` which is closely related.

assets/js/googlesitekit/widgets/datastore/widgets.js
```diff
@@ -316,6 +317,49 @@ export const selectors = {
...
+	 * @private
+	 *
+	 * @param {Object} state Data store's state.
+	 * @param {string} slug  Widget slug.
+	 * @return {boolean} True for active.
+	 */
+	isWidgetActive( state, slug ) {
+		return selectors.getWidgetState( state, slug ) === null ||
+			selectors.getWidgetState( state, slug ).Component !== Null;
+	},
```
felixarntz, line 334: While this selector might be useful, I think for our usage in the component where we have to compare `widgets` with `widgetStates` inside utility functions, it would be more beneficial to outsource the low-level logic to another utility function, e.g. `isInactiveWidgetState( widgetState ) => widgetState && widgetState.Component === Null`. This we could then use in e.g. `getWidgetLayout` (see above comment).

assets/js/googlesitekit/widgets/components/WidgetNull.js
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Internal dependencies
+ */
+import useWidgetStateEffect from '../hooks/useWidgetStateEffect';
+import Null from '../../../components/Null';
+
+// The supported props must match `Null` (except `widgetSlug`).
+export default function WidgetNull( { widgetSlug } ) {
+	useWidgetStateEffect( widgetSlug, Null, null );
```
felixarntz, line 32: Per the ACs, this should set an empty object as `metadata` instead of `null`: Essentially, `metadata` should be directly providable as props (see usage in `WidgetAreaRenderer`), so `null` is not good for use here. Let's just make sure we rely on `useMemo` (like in the other similar components relying on `useWidgetStateEffect`) or a hard-coded empty object, to make sure it doesn't always set a _new_ (different) object.

felixarntz (changes requested):
The overall approach makes sense to me; there are a couple more things though that as a result of the change will need to be addressed here. See comments below for details.

assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js
ivankruchkoff: Updated.

assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js
ivankruchkoff: Updated: This is the only place for activeWidgets.

assets/js/googlesitekit/widgets/components/WidgetNull.js
ivankruchkoff: Updated, https://github.com/google/site-kit-wp/pull/2731/commits/a5596274fdf6b19c25a32e3f99b55cdab1eff061 hard-coded empty object makes more sense.

assets/js/googlesitekit/widgets/components/WidgetRenderer.js
ivankruchkoff: @aaemnnosttv any reason for this?

assets/js/googlesitekit/widgets/datastore/widgets.js
ivankruchkoff: Updated.

assets/js/googlesitekit/widgets/datastore/widgets.js
ivankruchkoff: Updated and removed the selector in favour of the util function in https://github.com/google/site-kit-wp/pull/2731/commits/f8c6a5f4ef31bfb58d58071b4f1c0d6f911be951

assets/js/googlesitekit/widgets/components/WidgetRenderer.js
felixarntz: Ah looks like `getWidgetComponentProps` itself now takes care of that using `memize`, so that makes sense. Cleaner approach, so okay to remove here.

felixarntz (changes requested):
@ivankruchkoff This is almost good to go now, excellent work so far.

There is one more problem that was missed before, one set of checks in `getWidgetLayout` needs to be reworked, now that the function also receives inactive widgets.

assets/js/googlesitekit/widgets/util/get-widget-layout.js
```diff
@@ -101,32 +102,41 @@ function getWidgetSizes( counter, widget ) {
...
 		// Get available sizes for the current widget to select the most appropriate width for the current row.
 		let sizes = getWidgetSizes( counter, widget );
 
 		if (
 			// If it is the last widget in the entire widget area.
-			i + 1 === activeWidgets.length ||
+			i + 1 === widgets.length ||
 			// Or the next widget can't fit into the current row anyway, then we can try to use alternative sizes.
-			getWidgetSizes( sizes.sort( ascending )[ 0 ].counter, activeWidgets[ i + 1 ] ).filter( fitIntoRow ).length === 0
+			getWidgetSizes( sizes.sort( ascending )[ 0 ].counter, widgets[ i + 1 ] ).filter( fitIntoRow ).length === 0
```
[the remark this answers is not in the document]
ivankruchkoff: Addressed in https://github.com/google/site-kit-wp/pull/2731/commits/0774342320a02647f885df161538d3acc1aa162b

felixarntz (approved):
Awesome work @ivankruchkoff, good for merge! And with this, we're concluding the JS API refactoring and widgets API epics (full launch pending, but it's a milestone)!

Issue #2652, opened 2021-01-19
Update `Widget` component `header` and `footer` props to be capitalized


felixarntz:
Per our architecture requirements, any props/arguments/parameters that expect a `WPComponent` / `React.Component` / `PropTypes.elementType` to be passed should be in Pascal-case. That is to differentiate them from `WPElement` / `React.Element` / `PropTypes.element`, where we should be using camel-case.

Using Pascal-case also allows to reuse those names in JSX directly, where otherwise they would have to be renamed.

The `Widget` component is currently still violating that principle, so its `header` and `footer` props should be capitalized.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `header` and `footer` props of the `Widget` component should be renamed to `Header` and `Footer` respectively.
* All usages of these props should be adjusted accordingly.

## Implementation Brief
* Using `/assets/js/googlesitekit/widgets/components/Widget.js`
	* rename `header` and `footer` props to `Header` and `Footer` respectively.
* Review all occurrences where `header` and `footer` are used as props for the `Widget` component across the app and rename props accordingly.

### Test Coverage
* No changes.

### Visual Regression Changes
* No changes

## QA Brief

* Search the codebase for usage of the `Widget` component
* Ensure no components are using the old (lowercase) `header` or `footer` props
* Ensure all widgets using the new `Header` and `Footer` props are passing a [component type](https://github.com/facebook/react/tree/master/packages/react-is#determining-if-a-component-is-valid) rather than [an element](https://github.com/facebook/react/tree/master/packages/react-is#element)

## Changelog entry

* Rename the `header` and `footer` props of the `Widget` component to `Header` and `Footer` to indicate that they require a React component instead of an element.

felixarntz:
IB ✅

tofumatt:
QA ✅

PR #2732, opened 2021-01-31
Update component header and footer props to be capitalized.

johnPhillips:
## Summary

Addresses issue #2652

## Checklist

### DIFF
--- a/assets/js/googlesitekit/widgets/components/Widget.js
+++ b/assets/js/googlesitekit/widgets/components/Widget.js
@@ -27,8 +27,8 @@ const Widget = ( {
 	className,
 	slug,
 	noPadding,
-	header: Header,
-	footer: Footer,
+	Header,
+	Footer,
 } ) => {
 	return (
 		<div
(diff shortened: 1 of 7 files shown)


assets/js/googlesitekit/widgets/components/Widget.js
```diff
@@ -27,8 +27,8 @@ const Widget = ( {
 	className,
 	slug,
 	noPadding,
-	header: Header,
-	footer: Footer,
+	Header,
+	Footer,
```
aaemnnosttv, line 31: PropTypes need to be updated accordingly here. Note that the type also needs updating, not just the case.

That is, these should both expect a `PropTypes.elementType` now instead of an `element`.

aaemnnosttv (changes requested):
Thanks @johnPhillips, this looks good to me, just the one change left to make.

aaemnnosttv (approved):
Great, thanks!

Issue #2514, opened 2020-12-09
Add (React) Router 


tofumatt:
## Feature Description

In order to improve navigation behaviour in components, expose state in the URL when appropriate, give users more control of the plugin with browser history, and lessen our implementation of routers with things like `useQueryParams` and `useHash`, I think it's time we adopted an actual Router and started implementing routes in our components.

There are instances in the plugin already (like the [SettingsApp](https://github.com/google/site-kit-wp/blob/develop/assets/js/components/settings/settings-app.js#L54)) where we effectively implement a `HashRouter`—we could simplify our code and use more idiomatic React Router patterns using React Router.

There are also things like the User Input questions (see #2499) where using a Router would allow us more declarative control over the behaviour of a component—especially when linked to from other parts of the plugin.

Due to its widespread usage, knowledge/guides for it, and its reliability, [React Router](https://reactrouter.com/) is the recommended library from me.

Using React Router could even open us up to inter-linking the entire plugin with React, but this would require more effort in terms of asset management by JS and prefetching of data. It's worth exploring, but much later.

For now, the focus of this issue is to add React Router to the plugin, and to migrate a single section of the plugin that relies on a "homemade" router to React Router. I would nominate the `SettingsApp` to migrate to a `HashRouter`.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A React Router `Router` component should be added as a common wrapper to all React apps
* `SettingsApp` should be updated to use using `<Route />` components and `<Link>` components
    * It should continue to respond to the same top-level (`#settings`/`#connect`/`#admin`) hash URLs and the sub-URLs for `#settings/`
        * These should be updated to "redirect" to the new hashes below
    * Additionally, the hashes used for each tab should be updated as follows:
        * `#/connected-services` - Connected services tab
        * `#/connected-services/{moduleSlug}/{view/edit}` - Connected services tab with specific module panel opened (in view or edit state)
        * `#/connect-more-services` - Connect more services tab
        * `#/admin-settings` - Admin settings tab
* Active Module state for `SettingsApp` is no longer stored in `state`, but in the route.

## Implementation Brief

* `npm install --save react-router-dom`
* If permitted by the components, change the `<Tab>` components to use a `<Link>` directly so the user can see the URL they are about to navigate to in their browser. If not possible, update the `handleActiveIndexUpdate` prop to use `history.push()` instead of `setState()`.
[...]

PR #2733, opened 2021-02-01
Enhancement/2514 add react router and replace homegrown hash router within settings app

caseymorrisus:
## Summary

Addresses issue #2514 

Add React Router and replace existing (from scratch) implementation that holds component state within URL hash. The majority of the logic within `SettingsApp`, `SettingsModules` and `SettingsModule` was tied to our specific hash router implementation. Adding React Router allowed many workarounds to be removed as component props and state are now being updated in a way that React expects. State no longer needs to be managed and passed between components, individual components are now controlled via React Router and can access its state directly (less parent/child coupling).

## Relevant technical choices
<details>
  <summary>Click to expand detailed list of technical changes & decisions</summary>
 
- `Button.js` - Modified `children` prop type to allow non-strings (any valid React child type).
- `Link.js` - Modified render logic to pass through `to` prop if it exists (via `isRouterLink`) as well as conditionally rendering `RouterLink` via `SemanticLink`
- `Root/index.js` - Wrapped `children` in render with `HashRouter` to enable usage of React Router.
- `settings-module.js` - Removed & refactored into `SettingsModule.js`
- `SettingsApp.js` - Logic isolated to tab switching (decoupled from children).
    - `useHash` no longer necessary with React Router implementation, hash state is held via `HashRouter`.
    - `useFirstMountState` no longer necessary to set module or hash state on initial render, is set within URL via React Router.
    - `setModuleSettingsPanelState` no longer necessary, panel state is handled via URL.
    - `parseHash` no longer necessary, React Router handles parsing into params object.
    - `activeTabID` and `moduleSlug` state no longer necessary, tab and module state are handled via URL.
    - `getModuleSettingsPanelState` no longer necessary, `edit` and `view` states are handled via URL.
    - `setModuleState` no longer necessary, module state is handled via URL.
    - `handleTabUpdate` no longer necessary, active tab state is handled via URL.
    - `activeTab` modified to use URL in combination with `SettingsApp.basePathToTabIndex` to map base path values to tab indices.
    - `Tab` components are passed `NavLink` components (to allow active styling vs `Link` component) as well as a `to` prop to handle tab switching via React Router.
    - Removed props being passed to `SettingsModules` component, all module & tab state is handled via URL (decoupled from `SettingsApp` component).
    - Removed `SettingsAdmin` component from render method to add to `SettingsModules` component, which contains the relevant `Switch` & `Route` components that now handle rendering based on state.
    - Removed `tabToIndex` and replaced with `basePathToTabIndex` to map base path values with tab indices.
    - Use `SettingsApp.basePathToTabIndex` to map current path to the `activeIndex` value for `TabBar` , necessary to pass active tab styling to `NavLink` component.
    - Added `getTag` method to conditionally select `button` element instead of `NavLink` to avoid trying to push the same URL twice in a row (throws a warning).
- `SettingsApp.test.js` - Tests modified to use new URL structure and isolate functionality unique to `SettingsApp` component ( `Tab` & `TabBar` functionality).
- `SettingsConnectedServices.js` - Route component (`/connected-services/*`) that contains logic (conditionally returning `null` if no modules) & rendering for active settings modules.
- `SettingsConnectMoreServices.js` - Route component (`/connect-more-services`) for inactive settings modules with conditional rendering of congratulatory `Notification` and `SettingsOverlay` components.
- `SettingsModule.js`
    - Modified filename from `settings-module.js` to `SettingsModule.js` to match current naming standards.
    - Added `componentDidUpdate` lifecycle method to reset module to `initial` state when changing or closing module.
    - Removed `handleConfirmRemoveModule`, it directly calls `deactivate` .
    - Removed `getDependentModules` method to remove the need to call `getModulesData` within each module (called once within parent component).
    - Add `saveSettings` and `handleConfirmEditModule` methods to handle module edit logic directly within the component being edited.
        - This removes previous complexity where module state was being handled within the parent component (replaces `handleButtonAction`)
        - Logic simplified by using `status` vs `isSaving` and `error` booleans (`error.code` and `error.message` are never used)
    - Render method variables & logic have been updated to use state from React Router.
- `SettingsModule.test.js`
    - Includes tests for module display (edit vs view) as well as view switching via Edit & Cancel links.
    - Includes tests for accordion functionality.
- `SettingsModules.js` - Logic isolated to routing (including redirects) and transforming data to pass into routes (`/connected-services` & `/connect-more-services`).
    - Refactored to use implement React Router `Switch` component to handle conditional rendering based on URL, replacing `activeTab` logic.
    - Abstracted tab panels (handled within `Route` components) into `SettingsConnectedServices` and `SettingsConnectMoreServices` components.
    - Moved `SettingsAdmin` component (from `SettingsApp`) to render method within `Route` to align with other tab panels (without moving `Switch` up the component tree).
    - Added redirects to map existing routes to new routes via `Redirect` components. All previous routes should correctly route to the corresponding updated route.
    - Removed `updateModulesList` method, no longer necessary with updated logic (auto-updates on prop changes)
    - Removed `handleAccordion` method, focus and module state are handle via React Router via `HashRouter`.
    - Moved `handleButtonAction` to `SettingsModule` component now that state is decouped between parent/child.
    - Refactored `settingsModuleComponent` and `mapToModule` methods to filter into `activeModules` and `inactiveModules` arrays via `getModulesByActive` to decouple parent/child rendering logic.
    - Convert from class to functional component due to simplified logic.
- `SettingsModule.test.js` - Includes tests for redirects, route display itself is handled while testing Tab/TabBar within `SettingsApp` and testing output within `SettingsModule` tests.
- `_googlesitekit-settings-module.scss` - Must explicitly set `.googlesitekit-settings-module__header` to be block level element now that accordions use React Router `Link` internally (anchors default to inline elements).
- `test-utils.js` - Add `HashRouter` to Jest render `Wrapper` (conditionally via `useRouter` option) to allow React Router usage within component tests (in isolation).
</details>

## Checklist


assets/js/components/Button.js
```diff
@@ -83,7 +83,10 @@ Button.displayName = 'Button';
 
 Button.propTypes = {
 	onClick: PropTypes.func,
-	children: PropTypes.string.isRequired,
+	children: PropTypes.oneOfType( [
+		PropTypes.arrayOf( PropTypes.node ),
+		PropTypes.node,
+	] ),
```
tofumatt, line 89: Isn't `PropTypes.node` sufficient for `children`? I thought `node` described anything that could be rendered.

assets/js/components/Link.js
```diff
@@ -40,13 +41,17 @@ function Link( {
...
 	...extraProps
 } ) {
 	// Note: the disabled attribute does not alter behavior of anchor tags,
 	// so if disabled we force it to be a button.
 	const isAnchor = href && ! disabled;
-	const SemanticLink = isAnchor ? 'a' : 'button';
+	const element = isAnchor ? 'a' : 'button';
+	// If `to` prop is set, a `react-router` link is expected
+	const isRouterLink = typeof to !== 'undefined';
+	const SemanticLink = isRouterLink ? RouterLink : element;
```
tofumatt, line 54: This all makes sense but it's a bit tough, to me, to follow the nested ternary logic here. Some if/else statements might make the conditions clearer:

```suggestion
	let LinkElement;
	if ( typeof to !== 'undefined' ) {
		// If `to` prop is set, a `react-router` link is expected.
		LinkElement = RouterLink;
	} else if ( href && ! disabled ) {
		// If href` prop is set this is a "regular" link, but only render an
		// `<a>` tag if the link isn't disabled. Disabled links should be
		// rendered as `<button>`s instead.
		LinkElement = 'a';
	} else {
		LinkElement = 'button';
	}
```

assets/js/components/Link.js
```diff
@@ -82,6 +87,7 @@ function Link( {
 				},
 			) }
 			href={ isAnchor ? href : undefined }
+			to={ isRouterLink ? to : undefined }
```
tofumatt, line 90: Technically you can pass this one right through; if it's `undefined` it shouldn't be an issue; if it _isn't_ undefined we set the element as a React Router link that expects the "to" prop.

assets/js/components/Link.js
```diff
@@ -82,6 +87,7 @@ function Link( {
 				},
 			) }
 			href={ isAnchor ? href : undefined }
```
tofumatt, line 89: Might be better to pass this through only if `href && ! disabled` is true:
```suggestion
			href={ href !== undefined && ! disabled ? href : undefined }
```

assets/js/components/settings/SettingsApp.js
```diff
@@ -19,81 +19,32 @@
...
-	};
-
-	const handleTabUpdate = ( tabIndex ) => {
-		const newActiveTabID = SettingsApp.tabIDsByIndex[ tabIndex ];
-		setActiveTabID( newActiveTabID );
-		setHash( hashFrom( newActiveTabID ) );
-	};
+function SettingsApp( { location: { pathname } } ) {
+	// Don't pass NavLink component if it would send you to same URL, prevents a warning.
+	const getTag = ( path ) => basePath === path ? 'button' : NavLink;
```
tofumatt, line 45: Interesting, didn't know that would cause a warning. Sort of a shame!

assets/js/components/settings/SettingsConnectedServices.js
```diff
@@ -0,0 +1,68 @@
...
+	return (
+		<div className="
+			mdc-layout-grid__cell
+			mdc-layout-grid__cell--span-12
+		">
+			<Layout>
+				{ modules.map( ( module ) => (
+					<SettingsModule
+						module={ module }
+						key={ module.slug + '-module-wrapper' }
```
tofumatt, line 44: Does this need the extra string in there for the key? The keys are unique to this component, right? So using the `module.slug` should be unique enough in the context of this component.

```suggestion
						key={ module.slug }
```

assets/js/components/settings/SettingsModule.js
```diff
@@ -89,31 +92,39 @@ class SettingsModule extends Component {
 		global.removeEventListener( 'keyup', this.handleCloseModal );
 	}
 
+	componentDidUpdate( prevProps ) {
+		const { slug } = this.props.module;
+		const { moduleSlug: previousSlug } = prevProps.match.params;
+		const { moduleSlug: currentSlug } = this.props.match.params;
+		const changingModule = ( previousSlug === slug ) || ( currentSlug === slug );
+		const openModuleChanged = currentSlug !== previousSlug;
```
tofumatt, line 100: I follow what's happening here but it might be a bit opaque to someone new to the project or the component. Could you provide a bit more documentation in a comment here for why this flow occurs and what we're resetting when this happens? 🙂

assets/js/components/settings/SettingsModule.test.js
```diff
@@ -0,0 +1,128 @@
...
+			</Route>
+		</Switch>
+	);
+
+	// Create hash history to interact with HashRouter using `history.push`
+	const history = createHashHistory( { initialEntries: [ '/' ] } );
+	let registry;
+
+	beforeEach( () => {
+		global.location.hash = '';
```
tofumatt, line 57: I wonder if we should be resetting this after a test or even restoring it? But resetting it at least might be a good idea.

tests/js/test-utils.js
```diff
@@ -30,24 +31,31 @@ import { createTestRegistry } from './utils';
  * @param {Object}   options               Render options.
  * @param {Function} options.setupRegistry A function which accepts the registry instance to configure it.
  * @param {Function} options.registry      A specific registry instance to use. Defaults to a fresh test registry with all stores.
+ * @param {boolean}  options.useRouter     Decides if wrapped with HashRouter component. Defaults to false, no router.
```
tofumatt, line 34: I'm not a huge fan of this approach, as _eventually_ I'd think we'll want to use React Router for more and more of the app, eventually even allowing for switching between module pages without a fullpage reload. I'd prefer if having a Router enabled was assumed/default behaviour for tests, just like having a datastore is.

What's the drawback to always enabling a router?

I suppose I can see where you might want to conditionally enable a hash router as part of the render test, but in that case maybe this should be `hashRouter` or `useHashRouter` or even `wrapInHashRouter`.

tests/js/test-utils.js
```diff
@@ -30,24 +31,31 @@ import { createTestRegistry } from './utils';
...
 		registry = createTestRegistry(),
+		useRouter = false,
 		...renderOptions
 	} = options;
 
 	invariant( typeof setupRegistry === 'function', 'options.setupRegistry must be a function.' );
 	setupRegistry( registry );
 
+	// Conditionally wrap with `HashRouter` component.
+	const MaybeRouter = ( { children } ) => useRouter ? <HashRouter>{ children }</HashRouter> : children;
```
tofumatt, line 50: What about:

```suggestion
	const RouterComponent = useRouter ? HashRouter : Fragment;
```
?

assets/js/components/settings/SettingsModules.test.js
```diff
@@ -0,0 +1,85 @@
...
+ */
+import SettingsModules from './SettingsModules';
+import { render } from '../../../../tests/js/test-utils';
+
+describe( 'SettingsModules', () => {
+	// Create hash history to interact with HashRouter using `history.push`
+	const history = createHashHistory( { initialEntries: [ '/' ] } );
+
+	beforeEach( () => {
+		global.location.hash = '';
```
tofumatt, line 35: Again: would be good to reset this after all the tests have run.

assets/js/components/settings/SettingsModules.js
```diff
@@ -19,283 +19,84 @@
...
+		! module.internal && active === module.active;
+
+	const getModulesByActive = ( active ) =>
+		Object.values( modulesData )
+			.filter( byActiveNoInternals( active ) )
+			.sort( ( a, b ) => a.sort - b.sort )
+			.map( withDependantModulesText );
+
+	const activeModules = getModulesByActive( true );
+	const inactiveModules = getModulesByActive( false );
```
tofumatt, line 74: Nitpick, but this API is a bit tough-to-parse. How about:

```suggestion
	const getModulesByStatus = ( { active } = {} ) =>
		Object.values( modulesData )
			.filter( byActiveNoInternals( active ) )
			.sort( ( a, b ) => a.sort - b.sort )
			.map( withDependantModulesText );

	const activeModules = getModulesByStatus( { active: true } );
	const inactiveModules = getModulesByStatus( { active: false } );
```

assets/js/components/settings/SettingsModules.js
```diff
@@ -19,283 +19,84 @@
...
+
+	return (
+		<Switch>
+			{ /* Settings Module Routes */ }
+			<Route path={ [ '/connected-services/:moduleSlug/:action', '/connected-services/:moduleSlug', '/connected-services' ] }>
+				<SettingsConnectedServices modules={ activeModules } />
+			</Route>
```
tofumatt, line 81: Hmm, I wonder what others thoughts are on this? I'd usually make multiple routes with different `path` values but the same `children` for things like this, eg:

```suggestion
			{ /* Settings Module Routes */ }
			<Route path="/connected-services/:moduleSlug/:action">
				<SettingsConnectedServices modules={ activeModules } />
			</Route>
			<Route path="/connected-services/:moduleSlug">
				<SettingsConnectedServices modules={ activeModules } />
			</Route>
			<Route path="/connected-services">
				<SettingsConnectedServices modules={ activeModules } />
			</Route>
```

I suppose the downside is that it might re-render the entire child/mount a new component with similar props when we change actions, etc.? Is that what this is trying to avoid?

No strong opinion here, but I do like the readability of a path-per-Route approach.

cc @aaemnnosttv @felixarntz

tofumatt (changes requested):
Wow, this looks great! Amazing work! 👍🏻 

I have a few questions, a few ideas where we can improve things, and probably a few nitpicks. This is a bit change and I'll try it out locally a bit more, but for now I wanted to submit my first review of the approach while I continue local testing. Overall looks great though. Ping me if you have any further questions on my review.

(Also, looks like a lot of tests are failing.)

assets/js/components/settings/SettingsModules.js
aaemnnosttv: I tend to like the flatter path-per-route way as well, although it does look a bit odd that the result is the same for each. Would it be non-idiomatic to pass the props through directly here using one of the render prop variants of `Route`?

aaemnnosttv (commented):
Looking good @caseymorrisus ! Just adding a few comments on to the initial review but I like what I see so far.

One question I have is what happens to the datastore parts for module settings panel state – is that all unused now?

assets/js/components/Button.js
caseymorrisus: This is correct, silly me.

assets/js/components/Link.js
caseymorrisus: Good point — passing straight through as it'll be `undefined` when expected.

assets/js/components/Link.js
caseymorrisus: `isAnchor` was checking for falsy `href` values as well as `! disabled`, but I made the modification to typecheck for undef instead to be more clear. Empty strings aren't a possible value due to the ESLint rule (`jsx-a11y/anchor-is-valid`) which warns against it.

assets/js/components/settings/SettingsApp.js
caseymorrisus: The specific warning being: `Warning: Hash history cannot PUSH the same path; a new entry will not be added to the history stack`

After looking at this again with a clearer head, I'm not a big fan of the `getTag` approach I went with, seems hacky.

New approach uses a `shouldReplaceHistory` method that is passed into the `replace` prop to `Tab` components, which is supplied to `NavLink` components. This seems a lot less hacky, we're replacing the history entry with an identical one rather than depending on the `button` element not responding to the `to` prop.

assets/js/components/settings/SettingsModules.js
caseymorrisus: Life saver! I was trying to think of better naming & personally dislike passing booleans straight into functions (no context) but ended up with `getModulesByActive( { active: true } )` which seemed worse. `getModuleByStatus` naming is exactly what I was looking for, naming is hard sometimes. 😅

assets/js/components/settings/SettingsModules.js
caseymorrisus: @aaemnnosttv @tofumatt Modified this section to use flat paths. It will re-mount on change, but it doesn't appear to be an issue. I attempted to use render props but it still requires an array value for the path prop or multiple routes (flat paths).

Issue #2456, opened 2020-11-30
Introduce a new datastore for UI state


eugene-manuilov:
## Feature Description

After discussing #2392 on our internal call, we agreed that it would be more convenient to create a completely new datastore to let different logical components to store UI related state there. This new datastore should become our go-to place when it comes to knowing the selected date range, which tab is open in a widget, or which metric is select for a graph.

Having UI information stored in the new datastore eliminates the weirdness that we have right now when that information is stored across different stores that are not intended for such information. Plus, it gives us a good basis for the next UI/UX enhancements that we can implement having this datastore. For example, we can save information stored in this datastore in persistent storage (for example, in the IndexedDB) and load it on the next visit to restore UI from the last session.

The datastore itself should be pretty simple at the beginning: it should be key-value storage with appropriate setValue action and getValue selector. All additional features can be implemented later on.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a new `core/ui` JS datastore with selector `getValue( key )` and action `setValue( key, value )`. It should also have a `setValues( values )` that allows to set multiple key-value pairs at a time.
    * Essentially, it should be a simple key-value storage, almost similar to the `core/forms` store but without the extra nesting of having key-value pairs as part of a specific form.
* The new Analytics All Traffic widget and the PageSpeed Insights widget should have their implementation modified to rely on the `core/ui` store instead of the `core/forms` store, since their interactions are not really based on form content. The keys used should be:
    * `dashboardAllTrafficWidgetDimensionName`
    * `dashboardAllTrafficWidgetDimensionValue`
    * `dashboardPageSpeedWidgetStrategy`
    * `dashboardPageSpeedWidgetDataSource`

## Implementation Brief

* Create `assets/js/googlesitekit/datastore/ui/constants.js` file:
  * Export a new constant `STORE_NAME` with `core/ui` value;
  * Export the same `STORE_NAME` constant as `CORE_UI` alias.
* Create `assets/js/googlesitekit/datastore/ui/ui.js` file:
  * Define the standard structure of the regular datastore using [forms](https://github.com/google/site-kit-wp/blob/develop/assets/js/googlesitekit/datastore/forms/forms.js) datastore as an example;
  * Add `setValue( key, value )` and `setValues( values )` actions with appropriate reducers that add provided data into the state object;
  * Add `getValue( key )` selector that returns data from the state using provided key.
[...]

PR #2741, opened 2021-02-02
Enhancement/2456 UI data store

asvinb:
## Summary

Addresses issue #2456 

## Checklist


assets/js/googlesitekit/datastore/ui/ui.js
```diff
@@ -0,0 +1,123 @@
...
+	},
+
+	/**
+	 * Sets a particular value for a key.
+	 *
+	 * @since n.e.x.t
+	 * @private
+	 *
+	 * @param {string} key   Ui key to set the value for.
+	 * @param {string} value The value for the key.
```
tofumatt, line 56: _Strictly speaking_ does this have to be a string? I'd think storing `boolean` values here would be common, for UI flags signally whether something is hidden or not.

```suggestion
	 * @param {*} value The value for the key.
```

assets/js/googlesitekit/datastore/ui/ui.js
```diff
@@ -0,0 +1,123 @@
...
+	 * @return {Object} Redux-style action.
+	 */
+	setValue( key, value ) {
+		invariant( key, 'key is required.' );
+		invariant( value, 'value is required.' );
```
tofumatt, line 61: I don't think this is accurate either: `''`, `0`, `false`, etc. would all trigger this to fail which shouldn't happen.

I might not bother to check the `value` at all—even setting `undefined` here doesn't strike me as that terrible. But at most I'd check 
```suggestion
		invariant( value !== undefined, 'value is required.' );
```

assets/js/googlesitekit/datastore/ui/ui.test.js
```diff
@@ -0,0 +1,154 @@
...
+			} );
+
+			it( 'preserves data from state when new data is assigned', () => {
+				registry.dispatch( STORE_NAME ).setValues( { key1: 'value1', key2: 'value2' } );
+
+				registry.dispatch( STORE_NAME ).setValues( { key3: 'value3', key4: 'value4' } );
+
+				const formValue = registry.select( STORE_NAME ).getValue( 'key2' );
+				expect( formValue ).toEqual( 'value2' );
+			} );
```
tofumatt, line 101: I'd add some tests in here with non-string _values_ as I think we want to allow them and it'd be good if the tests explicitly supported that 🙂

assets/js/googlesitekit/datastore/ui/ui.test.js
```diff
@@ -0,0 +1,154 @@
...
+				expect( () => {
+					registry.dispatch( STORE_NAME ).setValue( 'key1' );
+				} ).toThrow( 'value is required.' );
+			} );
+
+			it( 'does not throw with a key and value', () => {
+				expect( () => {
+					registry.dispatch( STORE_NAME ).setValue( 'key1', 'value1' );
+				} ).not.toThrow();
+			} );
```
tofumatt, line 121: I'd expand these tests to account for `null`, non-`string` values, etc. as discussed above.

assets/js/googlesitekit/datastore/ui/ui.js
asvinb: 👍

assets/js/googlesitekit/datastore/ui/ui.js
asvinb: @tofumatt I think you're right. `value` can be anything which might not worth checking for. I've removed the test for it as well.

assets/js/googlesitekit/datastore/ui/ui.test.js
asvinb: 👍

assets/js/googlesitekit/datastore/ui/ui.test.js
asvinb: Updated @tofumatt

tofumatt (approved):
Looks great! 👍🏻

assets/js/googlesitekit/datastore/ui/ui.js
```diff
@@ -0,0 +1,122 @@
...
+ */
+
+/**
+ * External dependencies
+ */
+import invariant from 'invariant';
+import isPlainObject from 'lodash/isPlainObject';
+
+const SET_UI_VALUES = 'SET_UI_VALUES';
+const SET_UI_VALUE = 'SET_UI_VALUE';
```
felixarntz, line 26: These could be simplified to just `SET_VALUES` / `SET_VALUE`.

assets/js/modules/pagespeed-insights/datastore/constants.js
```diff
@@ -19,11 +19,13 @@
...
 // Report strategies.
 export const STRATEGY_MOBILE = 'mobile';
 export const STRATEGY_DESKTOP = 'desktop';
 // Report data sources.
 export const DATA_SRC_LAB = 'data_lab';
 export const DATA_SRC_FIELD = 'data_field';
+
+// Dashboard widgets using the UI store.
+export const WIDGET_STRATEGY = 'dashboardPageSpeedWidgetStrategy';
+export const WIDGET_DATA_SOURCE = 'dashboardPageSpeedWidgetDataSource';
```
felixarntz, line 31: Let's use a consistent naming format for these key variables across the plugin. I like to keep the constant names short, but maybe we should prefix them to indicate that they are fields for the `core/ui` store. The alternative would be to go with super-verbose names (like what is used for the actual values of these constants), but since these constants are usually not mixed between different components, it should be okay keeping their names short.

How about renaming these to `UI_STRATEGY` and `UI_DATA_SOURCE`?

assets/js/modules/analytics/datastore/constants.js
```diff
@@ -37,4 +37,6 @@ export const EDIT_SCOPE = 'https://www.googleapis.com/auth/analytics.edit';
 // Date range offset days for Analytics report requests.
 export const DATE_RANGE_OFFSET = 1;
 
-export const FORM_ALL_TRAFFIC_WIDGET = 'analyticsAllTrafficWidget';
+export const DIMENSION_NAME_ALL_TRAFFIC_WIDGET = 'dashboardAllTrafficWidgetDimensionName';
+export const DIMENSION_COLOR_ALL_TRAFFIC_WIDGET = 'dashboardAllTrafficWidgetDimensionColor';
+export const DIMENSION_VALUE_ALL_TRAFFIC_WIDGET = 'dashboardAllTrafficWidgetDimensionValue';
```
felixarntz, line 42: See my other comment, how about renaming the constants to `UI_DIMENSION_NAME`, `UI_DIMENSION_COLOR`, `UI_DIMENSION_VALUE`?

felixarntz (commented):
Overall LGTM, but some naming suggestions for more consistency and clarity for the constants that are only for the `core/ui` store.

cc @aaemnnosttv @tofumatt

assets/js/googlesitekit/datastore/ui/ui.js
asvinb: Updated @felixarntz

assets/js/modules/pagespeed-insights/datastore/constants.js
asvinb: Sounds good. Updated @felixarntz

assets/js/modules/analytics/datastore/constants.js
asvinb: Sounds good. Updated @felixarntz

felixarntz (approved):
Awesome!

Issue #2689, opened 2021-01-26
Pass specific dates in AdSense deep links


felixarntz:
Similar to #2286 and #2287, we should include specific dates in AdSense deep links for all new datastore-driven components (i.e. those components using `select( 'core/user' ).getDateRangeDates()`).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All AdSense frontend deep links (specifically "source links", not links in AdSense setup or settings) in JS components that rely on the datastore should be updated to include arguments indicating the dates from the current date range, so that the AdSense FE will show metrics for that specific date range.
* Such links should include the date in a `d` query parameter, such as `&d=2020%2F12%2F19-2021%2F01%2F17`.
    * The format of how the dates need to be specified here is different from Analytics and Search Console, but please review the IBs for #2286 and #2287 for inspiration and consistency where possible.

## Implementation Brief
* Using `/assets/js/modules/analytics/components/dashboard/AnalyticsAdSenseDashboardWidgetLayout.js`
	* Query the user store to get the date range:
		```js
		const dateRangeDates = useSelect( ( select ) => select( CORE_USER ).getDateRangeDates( {
			offsetDays: DATE_RANGE_OFFSET,
		} ) );
		```
	* The `DATE_RANGE_OFFSET` should be imported from `/assets/js/modules/adsense/datastore/constants.js`
* Create new file `/assets/js/modules/adsense/util/report-date-range-args` and using newly created file,
	* Create a new function `generateDateRangeArgs` which accepts the returned object from the `getDateRangeDates` selector. The object contains the following keys:
		* `startDate`
		* `endDate`
	* The `generateDateRangeArgs` should return an object with the following key, d:
		```js
		d: `${ startDate.replace( /-/g, '/' ) }-${ endDate.replace( /-/g, '/' ) }`
		```
	* For e.g `{ d: '2021/01/01-2021/01/27' }`
* Using `/assets/js/modules/adsense/datastore/service.js`, update the `getServiceAccountSiteURL` selector to include a parameter `urlParams` (object) which will be appended to the URL returned by the selector. In the `query` constant, spread the `urlParams` parameter.
* Using `/assets/js/modules/analytics/components/dashboard/AnalyticsAdSenseDashboardWidgetLayout.js`
	* Call the `generateDateRangeArgs` function and pass the returned object as args to the `getServiceAccountSiteURL` selector.
[...]

PR #2742, opened 2021-02-02
Pass specific dates in AdSense deep links (2689)

johnPhillips:
## Summary

Addresses issue #2689

## Checklist


assets/js/modules/adsense/util/report-date-range-args.test.js
```diff
@@ -0,0 +1,53 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { generateDateRangeArgs } from './report-date-range-args';
+
+describe( 'AdSense reporting date range arguments', () => {
+	describe( 'generateDateRangeArgs', () => {
```
eugene-manuilov, line 21: No need to use two `describe` blocks here. Let's combine them into one.

assets/js/modules/analytics/components/dashboard/AnalyticsAdSenseDashboardWidgetLayout.js
```diff
@@ -25,14 +25,20 @@ import { __, _n, sprintf } from '@wordpress/i18n';
...
 const { useSelect } = Data;
 
 const AnalyticsAdSenseDashboardWidgetLayout = ( { children } ) => {
-	const accountSiteURL = useSelect( ( select ) => select( MODULES_ADSENSE ).getServiceAccountSiteURL() );
+	const dateRangeDates = useSelect( ( select ) => select( CORE_USER ).getDateRangeDates( {
+		offsetDays: DATE_RANGE_OFFSET,
+	} ) );
+	const accountSiteURL = useSelect( ( select ) => select( MODULES_ADSENSE ).getServiceAccountSiteURL(
+		{ ...generateDateRangeArgs( dateRangeDates ) }
+	) );
```
eugene-manuilov, line 41: I think we can simplify it a bit:

```suggestion
	const accountSiteURL = useSelect( ( select ) => select( MODULES_ADSENSE ).getServiceAccountSiteURL( generateDateRangeArgs( dateRangeDates ) ) );
```

assets/js/modules/adsense/util/report-date-range-args.js
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Generates an object with the appropriate structure to pass to report selectors.
+ *
+ * @since n.e.x.t
+ *
+ * @param {Object} dates           Object containing the start and end dates.
+ * @param {string} dates.startDate Start date in YYYY-MM-DD format.
+ * @param {string} dates.endDate   End date in YYYY-MM-DD format.
+ * @return {Object|undefined} Date range object.
```
eugene-manuilov, line 32:
```suggestion
 * @return {Object} Date range object.
```

assets/js/modules/adsense/components/dashboard/AdSenseDashboardWidget.js
```diff
@@ -35,39 +35,44 @@ import getNoDataComponent from '../../../../components/legacy-notifications/noda
...
+	} ) );
+	const accountSiteURL = useSelect( ( select ) => select( MODULES_ADSENSE ).getServiceAccountSiteURL(
+		{ ...generateDateRangeArgs( dateRangeDates ) }
+	) );
+	const dateRange = useSelect( ( select ) => select( CORE_USER ).getDateRange() );
+	const currentDayCount = getCurrentDateRangeDayCount( dateRange );
 
 	/**
-	 * Handle data errors from the contained AdSense component(s).
+	 * Handles data errors from the contained Analytics component(s).
```
eugene-manuilov, line 73:
```suggestion
	 * Handles data errors from the contained AdSense component(s).
```

assets/js/modules/adsense/components/dashboard/AdSenseDashboardWidget.js
```diff
@@ -35,39 +35,44 @@ import getNoDataComponent from '../../../../components/legacy-notifications/noda
...
 	 *
-	 * Currently handled in the AdSenseEstimateEarningsWidget component.
+	 * Currently handled in the AnalyticsDashboardWidget component.
```
eugene-manuilov, line 75:
```suggestion
	 * Currently handled in the AdSenseEstimateEarningsWidget component.
```

assets/js/modules/adsense/components/dashboard/AdSenseDashboardWidget.js
```diff
@@ -35,39 +35,44 @@ import getNoDataComponent from '../../../../components/legacy-notifications/noda
...
+	const [ loading, setLoading ] = useState( true );
+	const [ zeroData, setZeroData ] = useState( false );
+
+	const adSenseModuleConnected = useSelect( ( select ) => select( CORE_MODULES ).isModuleConnected( 'adsense' ) );
+	const dateRangeDates = useSelect( ( select ) => select( CORE_USER ).getDateRangeDates( {
+		offsetDays: DATE_RANGE_OFFSET,
+	} ) );
+	const accountSiteURL = useSelect( ( select ) => select( MODULES_ADSENSE ).getServiceAccountSiteURL(
+		{ ...generateDateRangeArgs( dateRangeDates ) }
+	) );
```
eugene-manuilov, line 68:
```suggestion
	const accountSiteURL = useSelect( ( select ) => select( MODULES_ADSENSE ).getServiceAccountSiteURL( generateDateRangeArgs( dateRangeDates ) ) );
```

eugene-manuilov (changes requested):
Thanks, @johnPhillips! It mostly looks good to me, just added some comments to address minor issues.

assets/js/modules/adsense/components/dashboard/AdSenseDashboardWidget.js
johnPhillips: Ah yes, of course 🤦

assets/js/modules/adsense/components/dashboard/AdSenseDashboardWidget.js
johnPhillips: Great spot. I was refactoring this based on a similar file and these comment changes must have crept in from there.

eugene-manuilov (approved):
LGTM! Thanks, @johnPhillips!

assets/js/modules/adsense/datastore/service.test.js
```diff
@@ -76,5 +94,51 @@ describe( 'module/adsense service store', () => {
...
+					source: 'site-kit',
+					url: referenceSiteURL,
+				} );
+			} );
+
+			it( 'should append `urlParams` arguments to the `query` if received', () => {
+				const urlParams = { foo: 'bar' };
+				const url = registry.select( STORE_NAME ).getServiceAccountSiteURL( urlParams );
+				expect( url ).toMatchQueryParameters( urlParams );
+			} );
```
aaemnnosttv, line 141: This test doesn't show that the given `urlParams` are _appended_ to the default query params; only that they are present. Let's add another assertion to check that the default query params are also present.

It might also be good to add one more test here to show that `urlParams` take precedence over the default query params for the same key.

assets/js/modules/adsense/util/report-date-range-args.js
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * External dependencies
+ */
+import invariant from 'invariant';
+
+/**
+ * Internal dependencies
+ */
+import { isValidDateString } from '../../../googlesitekit/datastore/user/utils/is-valid-date-string';
+/**
```
aaemnnosttv, line 28: Nitpicking this a bit but please include a blank line between import groups and other comments.
```suggestion

/**
```

assets/js/modules/adsense/util/report-date-range-args.js
```diff
@@ -0,0 +1,47 @@
...
+ * @param {Object} dates           Object containing the start and end dates.
+ * @param {string} dates.startDate Start date in YYYY-MM-DD format.
+ * @param {string} dates.endDate   End date in YYYY-MM-DD format.
+ * @return {Object} Date range object.
+ */
+export const generateDateRangeArgs = ( dates ) => {
+	invariant( dates, 'dates object is required' );
+	const { startDate, endDate } = dates;
+	invariant( isValidDateString( startDate ), 'startDate is required' );
+	invariant( isValidDateString( endDate ), 'endDate is required' );
```
aaemnnosttv, line 42: The validation message should be consistent with the others in that "A valid startDate is required." or "A valid endDate is required."

This is generally the format we use for invariant messages, particularly when the logic uses an `isValid` utility.

assets/js/modules/adsense/util/report-date-range-args.test.js
```diff
@@ -0,0 +1,38 @@
...
+
+	it( 'should throw if `startDate` argument is not provided', () => {
+		expect( () => generateDateRangeArgs( { endDate: '2021-01-27' } ) ).toThrow( 'startDate is required' );
+	} );
+
+	it( 'should throw if `endDate` argument is not provided', () => {
+		expect( () => generateDateRangeArgs( { startDate: '2021-01-01' } ) ).toThrow( 'endDate is required' );
+	} );
+
+	it( 'should return an object containing a `d` key, the value of which is the `startDate` argument with "-" replaced with "/", concatenated to the `endDate` argument also with "-" replaced with "/", using `-` to join the two strings', () => {
```
aaemnnosttv, line 33: This reads a bit overly verbose and is a bit too much detail regarding the implementation. While very explicit, it likely isn't all necessary in the output, even if this test fails.

It's okay for some of the logic to be conveyed by the test itself, as long as it is clear. In this case, I think it's quite clear what it's doing so a more concise name would be easier to read without hurting the clarity of the intent.

aaemnnosttv (changes requested):
Thanks @johnPhillips – this looks good, just a few small things to address.

assets/js/modules/adsense/components/dashboard/AdSenseDashboardWidget.js
```diff
@@ -35,37 +35,40 @@ import getNoDataComponent from '../../../../components/legacy-notifications/noda
...
 import ModuleSettingsWarning from '../../../../components/legacy-notifications/module-settings-warning';
-import { getModulesData } from '../../../../util';
 import DateRangeSelector from '../../../../components/DateRangeSelector';
 import HelpLink from '../../../../components/HelpLink';
 import Header from '../../../../components/Header';
 import PageHeader from '../../../../components/PageHeader';
 import Layout from '../../../../components/layout/Layout';
 import { CORE_USER } from '../../../../googlesitekit/datastore/user/constants';
+import { CORE_MODULES } from '../../../../googlesitekit/modules/datastore/constants';
+import { MODULES_ADSENSE, DATE_RANGE_OFFSET } from '../../../adsense/datastore/constants';
```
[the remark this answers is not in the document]
johnPhillips: Sorry @aaemnnosttv! I know you've pointed this out to me a few times now - I'll get there in the end 😂

assets/js/modules/adsense/util/report-date-range-args.test.js
johnPhillips: LOL yes, now I read it again it does seem a bit long

assets/js/modules/adsense/datastore/service.js
```diff
@@ -106,9 +106,10 @@ export const selectors = {
 	 *
 	 * @since 1.14.0
 	 *
+	 * @param {Object} urlParams URL parameters to be passed to the query.
 	 * @return {(string|undefined)} AdSense account site overview URL (or `undefined` if not loaded).
 	 */
-	getServiceAccountSiteURL: createRegistrySelector( ( select ) => () => {
+	getServiceAccountSiteURL: createRegistrySelector( ( select ) => ( state, urlParams = {} ) => {
```
felixarntz, line 112: @aaemnnosttv noticed during testing that the URL returned by this selector is actually not the correct one to pass these date parameters too, which I just verified.

All these links instead need to go to a `https://www.google.com/adsense/new/u/0/{accountID}/reporting/` base URL. Let's introduce a simple AdSense `getServiceReportURL( reportArgs )` registry selector, similar to the one we have for Analytics. It should do the following:
* Get the `accountID`.
* Call the `getServiceURL( { path, query } )` selector with `path` set to `${ accountID }/reporting` and `query` set to the `reportArgs` provided, simply passing through.

Let's then revert the changes made to `getServiceAccountSiteURL` here and rely on the new selector throughout this PR. Its extra arguments like `source` and `url` are irrelevant for report URLs and thus shouldn't be provided with the new selector.

assets/js/modules/adsense/components/dashboard/AdSenseDashboardWidget.js
```diff
@@ -35,37 +35,40 @@ import getNoDataComponent from '../../../../components/legacy-notifications/noda
...
+	const [ error, setError ] = useState( false );
+	const [ errorObj, setErrorObj ] = useState();
+	const [ loading, setLoading ] = useState( true );
+	const [ zeroData, setZeroData ] = useState( false );
+
+	const adSenseModuleConnected = useSelect( ( select ) => select( CORE_MODULES ).isModuleConnected( 'adsense' ) );
+	const dateRangeDates = useSelect( ( select ) => select( CORE_USER ).getDateRangeDates( {
+		offsetDays: DATE_RANGE_OFFSET,
+	} ) );
+	const accountSiteURL = useSelect( ( select ) => select( STORE_NAME ).getServiceAccountSiteURL( generateDateRangeArgs( dateRangeDates ) ) );
```
felixarntz, line 66: See my other comment, let's use a new `getServiceReportURL` selector here instead.

assets/js/modules/analytics/components/dashboard/AnalyticsAdSenseDashboardWidgetLayout.js
```diff
@@ -25,14 +25,18 @@ import { __, _n, sprintf } from '@wordpress/i18n';
...
 import { getCurrentDateRangeDayCount } from '../../../../util/date-range';
+import { generateDateRangeArgs } from '../../../adsense/util/report-date-range-args';
 const { useSelect } = Data;
 
 const AnalyticsAdSenseDashboardWidgetLayout = ( { children } ) => {
-	const accountSiteURL = useSelect( ( select ) => select( MODULES_ADSENSE ).getServiceAccountSiteURL() );
+	const dateRangeDates = useSelect( ( select ) => select( CORE_USER ).getDateRangeDates( {
+		offsetDays: DATE_RANGE_OFFSET,
+	} ) );
+	const accountSiteURL = useSelect( ( select ) => select( MODULES_ADSENSE ).getServiceAccountSiteURL( generateDateRangeArgs( dateRangeDates ) ) );
```
felixarntz, line 39: See above, let's use the new selector here.

Technically, linking to AdSense here is a bit odd since the data in this widget actually comes from Analytics. Fixing that is out of scope though, so let's just keep linking to AdSense here for now. We can address this in a follow-up issue.

felixarntz (changes requested):
This is almost good to go. There is one thing we need to change though, which wasn't clear from the ACs and IB, see below. It requires a new selector, but it should be fairly straightforward.

assets/js/modules/adsense/datastore/service.js
```diff
@@ -106,6 +106,7 @@ export const selectors = {
 	 *
 	 * @since 1.14.0
 	 *
+	 * @param {Object} urlParams URL parameters to be passed to the query.
```
felixarntz, line 109: This also needs to be reverted.

assets/js/modules/adsense/datastore/service.js
```diff
@@ -125,6 +126,13 @@ export const selectors = {
 		return select( STORE_NAME ).getServiceURL( { path, query } );
 	} ),
 
+	getServiceReportURL: createRegistrySelector( ( select ) => ( state, reportArgs ) => {
+		const accountID = select( STORE_NAME ).getAccountID();
```
felixarntz, line 130: When the data hasn't been loaded yet, `accountID` can be `undefined`. Can you add a condition like in the other selectors above that in such a case returns `undefined`?

assets/js/modules/adsense/datastore/service.js
```diff
@@ -125,6 +126,13 @@ export const selectors = {
 		return select( STORE_NAME ).getServiceURL( { path, query } );
 	} ),
 
+	getServiceReportURL: createRegistrySelector( ( select ) => ( state, reportArgs ) => {
```
felixarntz, line 129: Please add a doc block to this function, since it's a regular selector (which should be documented).

felixarntz (changes requested):
Functionality-wise LGTM now, just 3 minor points of feedback, then this should be good.

assets/js/modules/adsense/datastore/service.js
```diff
@@ -126,8 +125,21 @@ export const selectors = {
 		return select( STORE_NAME ).getServiceURL( { path, query } );
 	} ),
 
+	/**
+	 * Returns the service URL to an AdSense account's site overview page.
```
felixarntz, line 129: This should describe a bit more what this URL is for. Right now it is an exact copy of the other doc-block description. Should be more like `Returns the service URL to an AdSense report.

felixarntz (commented):
Technically good to go, just the doc-block content isn't really accurate. I'll quickly update that.

felixarntz (approved):
Awesome!

Issue #2715, opened 2021-01-28
"All Users" in All Traffic widget breadcrumb should link to overall view


felixarntz:
When the left side of the All Traffic widget shows data for a filtered subset of all users, the "All Users" segment of the "All Users > ..." breadcrumb should look like a link. When clicked, it should result in the overall graph being displayed again and the respective pie slice being unselected.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The "All Users" segment of the "All Users > ..." breadcrumb should be a link that, when clicked, goes back to the regular "All Users" view, displaying the overall graph.
* When that happens, the respective pie slice should also be no longer selected.

## Implementation Brief
* Using `assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/TotalUserCount.js`
	* Add a link around the `All Users` text if there is a `dimensionValue`.
	* Upon clicking the link, it should clear the `dimensionValue` and `dimensionColor`.
		```js
		const { setValues } = useDispatch( CORE_FORMS );
		const showAllUsers = ( e ) => {
			e.preventDefault();

			setValues( FORM_ALL_TRAFFIC_WIDGET, { dimensionValue: '', dimensionColor: '' } );
		};
		```
### Test Coverage
* No changes expected.

### Visual Regression Changes
* No changes expected.

## QA Brief

* Set up your Site Kit and connect the Analytics module.
* Visit the dashboard.
* Click on a segment in the pie chart of the "Your Traffic at a Glance" widget.
* The "All Users" text will become a link that resets the select pie segment.

## Changelog entry

* When viewing a specific segment of users in the All Traffic widget, make the "All Users" breadcrumb a link to go back to the overall users view.

felixarntz:
IB ✅

benbowler:
Added an IB.

wpdarren:
### QA Update:

**Verified:**

- When you click on a segment in the pie chart on the "Your Traffic at a Glance" widget, the "All Users" text becomes a link that resets the select pie segment. 

- Tested this on all 3 dimensions and with different date periods, i.e. 7, 14,28 and 90 days. 
- Tested this on mobile devices with different screen sizes. 

![image](https://user-images.githubusercontent.com/73545194/107221882-2f181480-6a4f-11eb-870d-9a9b09e77f3e.png)

PR #2745, opened 2021-02-02
"All Users" in All Traffic widget breadcrumb should link to overall view.

benbowler:
## Summary

Addresses issue #2715

## Checklist


benbowler:
One thing, I've noticed is that because it is a button with an onClick event it doesn't display the `pointer` cursor that it would if it were a link. We could fix this by adding:

```
&:not([disabled]) {
	cursor: pointer;
}
```

to the default button styles in `assets/sass/base/_defaults.scss`.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/TotalUserCount.js
```diff
@@ -83,7 +92,8 @@ export default function TotalUserCount( { loaded, error, report, dimensionValue
 				{ ! dimensionValue && __( 'All Users', 'google-site-kit' ) }
 				{ dimensionValue && (
 					<Fragment>
-						{ __( 'All Users', 'google-site-kit' ) }
+						{ dimensionValue && <button onClick={ showAllUsers }>{ __( 'All Users', 'google-site-kit' ) }</button> }
```
felixarntz, line 95: This should use the `Link` component instead - in general, we should always rely on `Link` or `Button`, never on the plain HTML elements. Both of our components work correctly regardless of whether semantically we want a link (`a`) or a button (`a`) - the component name is based on the appearance (e.g. a `Link` always looks like a link should look, no matter if it technically only has no `href` and would therefore use the `button` HTML element under the hood).

By using this component, we should also get the correct cursor (referring to your comment).

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/TotalUserCount.js
```diff
@@ -39,11 +39,20 @@ import { getAvailableDateRanges } from '../../../../../util/date-range';
...
+import { CORE_FORMS } from '../../../../../googlesitekit/datastore/forms/constants';
+import { FORM_ALL_TRAFFIC_WIDGET } from '../../../datastore/constants';
+const { useSelect, useDispatch } = Data;
 
 export default function TotalUserCount( { loaded, error, report, dimensionValue } ) {
 	const dateRange = useSelect( ( select ) => select( CORE_USER ).getDateRange() );
 
+	const { setValues } = useDispatch( CORE_FORMS );
+	const showAllUsers = ( e ) => {
+		e.preventDefault();
```
felixarntz, line 51: I don't think we need to call `preventDefault`, we don't typically do this in other button listeners either. It's only necessary to actually prevent default behavior that would otherwise run, e.g. a form submission or opening a link.

felixarntz (changes requested):
Mostly LGTM, two things that should be addressed though.

felixarntz (approved):
Awesome!

felixarntz (changes requested):
@benbowler Actually the font size is slightly off here because it's set in the default styles for `Link` (see below):
Screen Shot 2021-02-03 at 9 04 07 AM

Can you add a CSS rule specific to the All Traffic widget (in its `.scss` file) that this link doesn't result in a smaller font size? `font-size: 1` should do it I think.

benbowler:
> @benbowler Actually the font size is slightly off here because it's set in the default styles for `Link` (see below):
> Screen Shot 2021-02-03 at 9 04 07 AM
> 
> Can you add a CSS rule specific to the All Traffic widget (in its `.scss` file) that this link doesn't result in a smaller font size? `font-size: 1` should do it I think.

I went with:

```
		.googlesitekit-cta-link {
			font-size: 1em;
			letter-spacing: 0.04em;
		}
```

To match the `h3` tag.

felixarntz (approved):
Great!

Issue #2700, opened 2021-01-27
All Traffic widget pie chart legend should not be cut off


felixarntz:
The legend entries on the new All Traffic widget pie chart currently get shortened if there is not enough space available, and sometimes they even receive an arrow to see the other ones. See screenshot below:
![Screen Shot 2021-01-27 at 11 07 42 AM](https://user-images.githubusercontent.com/3531426/106040979-10e60800-6090-11eb-9899-fba03010d18e.png)

Instead of this, they should always show completely and never be shortened/cut off. If they take too much horizontal space, they should instead float into a new line below. We intentionally limit the results to 5, so this should never result in a crazy amount of vertical space.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The legend entries on the new All Traffic widget pie chart should always be shown completely and never be shortened with an ellipsis (unless a single one would be longer than the available horizontal space for the entire pie chart, which is highly unlikely). They should float into multiple rows as needed. In accordance with that, there should also never be an arrow to see the other entries that didn't fit.

#### DashboardAllTrafficWidget/index.js:
* Pass the `dimensionValue` variable to the `UserDimensionsPieChart` element as we do it for the `TotalUserCount` element;

#### DashboardAllTrafficWidget/UserDimensionsPieChart.js:
* Add a new property `dimensionValue` to the `UserDimensionsPieChart.propTypes` map;
* Disable the built-in legend in the user dimensions pie chart by setting the `position: 'none'` setting for the `legend` options;
* Create a new `onLegendClick` callback in the `UserDimensionsPieChart` component:
  * The callback should accept one argument: an index of clicked legend item;
  * Grab the chart instance using the `GoogleChart.charts.get( chartID )` function (see how we do it in the onSelect handler);
  * Get the current selection using `chart.getSelection()` function;
  * Call `chart.setSelection( null )` if there is already a selected row with an index that equals to the clicked legend item index, otherwise call `chart.setSelection( [ { row: selectedItem, column: null } ] )`;
  * Dispatch the new value for the `dimensionValue` property as we do it in the onSelect handler.
* Render the new legend beneath the google chart element using the new `onLegendClick` callback for legend items. It can be done something like this:
  ```javascript
  <div>
      { dataMap.slice( 1 ).map( ( [ label ], i ) => (
          <button key={ label } onClick={ () => onLegendClick( i ) }>{ label }</button>
      ) ) }
  </div>
  ```
[...]

Issue #2710, opened 2021-01-28
Prevent jumps in All Traffic widget for loading states and when switching tabs


felixarntz:
When the new All Traffic widget is in any loading state, you'll see brief jumps where the widget briefly is taller than usually. Similar inconsistencies can be noticed when switching tabs when new data needs to be loaded.

It seems that this is partly caused by the charts library, where on the initial render of any chart it uses a taller height for a split-second sometimes. If the problem here is the charts library and we cannot change that behavior, we should try to wrap it with some container that accounts for that jump, to ensure that at least not the overall widget height jumps.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The new All Traffic widget should always show up with a consistent height. Entering or leaving a loading state in any part should not trigger the height of the overall widget to change (causing content jumps).

## Implementation Brief

* Adjust styles of the preview element in the user dimensions pie chart component to make sure it takes the same height as the loaded component.

### Test Coverage

* N/A

### Visual Regression Changes

* The All Traffic widget should have the same height no matter in which state it is (loading or rendering data).

## QA Brief

* Set up the plugin on your local;
* Activate and connect the Analytics module;
* Go to the dashboard page;
* Check the All Traffic widget, it should remain in the same size when you switch tabs or select pie slices.

## Changelog entry

* Avoid content shifts in All Traffic widget due to incorrect loading UI layout.

felixarntz:
@eugene-manuilov I've by now already updated the `PreviewBlock`s to match their heights in https://github.com/google/site-kit-wp/commit/32e2be50d6258fed36aedbae8ca85a9234b16889. The only weird jump that remains on my end is that the `PreviewBlock` instance for the pie chart specifically jumps at some point _during_ the loading phase. See here:

https://user-images.githubusercontent.com/3531426/106540471-dd9ee100-64b4-11eb-834c-7cbcaef6451b.mov

So the problem here is not even between loading and non-loading state, but _within_ loading state a jump happens. Let me know if you can replicate that, or if you see anything else that doesn't look smooth on your end.

Maybe the loading circle jump is the only thing we'll need to fix here.

eugene-manuilov:
@felixarntz yeah, seems like the last bit is to adjust the size of the preview block for the user dimensions pie chart. IB is updated.

felixarntz:
IB ✅

wpdarren:
QA Update: Pass ✅ 

[...]

PR #2752, opened 2021-02-03
Introduce our own legend for Pie Charts in All Traffic widget.

asvinb:
## Summary

Addresses issue #2700 

## Checklist


asvinb:
The PR needs to be updated now that #2710 has been merged where we need to remove the negative margins introduced here: https://github.com/google/site-kit-wp/pull/2754/files#diff-9c02e442264b614408ff5c987b2652e3470828705cdd0aab2c36f4dc854bfd81R88
The negative margins were added because of the legend which was part of the Chart. Now that the legend is living outside of the chart, we no longer need the negative margins to reposition the placeholder.

aaemnnosttv (changes requested):
Thanks @asvinb – this looks great!

A few things to address here:
1. I'm not sure if we want the global button focus style applied here (@felixarntz?)   
image
2. There is a slight shift when "activating" one of the items due to the change in font-weight  
Kapture 2021-02-04 at 19 37 48
3. I'm not sure exactly what the hover style should be here. The clip above shows an underline but this isn't present in the design. Also it isn't clear if the background color is supposed to be applied on hover as well or only on selection
4. You mentioned we shouldn't need the negative margins that were added in #2710 anymore to account for the legend from the charts – can we remove those styles that are no longer relevant?

I realize some of these things require answers that I don't have yet, but we should still address them here if we can.

asvinb:
> Thanks @asvinb – this looks great!
> 
> A few things to address here:
> 
> 1. I'm not sure if we want the global button focus style applied here (@felixarntz?)
>    image
> 2. There is a slight shift when "activating" one of the items due to the change in font-weight
>    Kapture 2021-02-04 at 19 37 48
> 3. I'm not sure exactly what the hover style should be here. The clip above shows an underline but this isn't present in the design. Also it isn't clear if the background color is supposed to be applied on hover as well or only on selection
> 4. You mentioned we shouldn't need the negative margins that were added in #2710 anymore to account for the legend from the charts – can we remove those styles that are no longer relevant?
> 
> I realize some of these things require answers that I don't have yet, but we should still address them here if we can.

@aaemnnosttv My notes below
1. I would leave them mainly for accessibility reasons.
2. This is as per the designs where the active item has a font weight of 500.
3. I would leave the hover style as is else we'll need to find a way on how to style the hover state efficiently. Because right now, the colors are all set via inline CSS, which uses the same colors defined in the component. https://github.com/google/site-kit-wp/pull/2752/files#diff-3f30def21f83e2a53e6b645675570d834e6aa59efd4e4b2598d50b15543feaceR299 We can either add predefined classes in the CSS, thus duplicating the colors, or dynamically inject the hover styles via JS.
4. Yes, I'll remove the styles.

asvinb:
@aaemnnosttv Negative margins have been removed.

aaemnnosttv:
>> I'm not sure if we want the global button focus style applied here

> I would leave them mainly for accessibility reasons.

@asvinb  Sure, but this should be a deliberate decision though as it's a user-facing change, so we'll need input from @felixarntz about it.

>> There is a slight shift when "activating" one of the items due to the change in font-weight

> This is as per the designs where the active item has a font weight of 500.

Yes, I understand. The font weight is fine, but we should be able to [tweak the styles to prevent the shifting](https://css-tricks.com/bold-on-hover-without-the-layout-shift/) that this causes.

> I would leave the hover style as is else we'll need to find a way on how to style the hover state efficiently. Because right now, the colors are all set via inline CSS, which uses the same colors defined in the component. We can either add predefined classes in the CSS, thus duplicating the colors, or dynamically inject the hover styles via JS.

Looking at the markup there, we should be able to do this with CSS alone. We simply need to render the `span.googlesitekit-widget--analyticsAllTraffic__active-underlay` unconditionally (rather than only when `isActive`). Then in CSS, we just need to do something like this:
```scss
.googlesitekit-widget--analyticsAllTraffic__active,
.googlesitekit-widget--analyticsAllTraffic__legend-slice:hover {
    > .googlesitekit-widget--analyticsAllTraffic__active-underlay { display: block; }
}
```

Or, since the underlay is a separate element, could hide it with `opacity: 0` by default instead and vary this to create a better hover experience.

```scss
.googlesitekit-widget--analyticsAllTraffic__active-underlay {
    opacity: 0;
}
.googlesitekit-widget--analyticsAllTraffic__legend-slice:hover  .googlesitekit-widget--analyticsAllTraffic__active-underlay {
    opacity: .7;
}
.googlesitekit-widget--analyticsAllTraffic__active .googlesitekit-widget--analyticsAllTraffic__active-underlay {
    opacity: 1;
}
```
What do you think? There seems to be some room for improvement with the class names as well.
* `googlesitekit-widget--analyticsAllTraffic__active-underlay` should probably be:  
`googlesitekit-widget--analyticsAllTraffic__underlay` since it shouldn't really be active-only I think.
* `googlesitekit-widget--analyticsAllTraffic__active` should probably be:  
`googlesitekit-widget--analyticsAllTraffic__legend-slice--active` to match its main class.
Aside – shouldn't the `--` after `googlesitekit-widget` be a single dash, or underscore(s) since double dash usually indicates state? I guess we're not doing strict BEM here but would be good to maintain consistency where we can.

> Negative margins have been removed.

Awesome, thanks!

asvinb:
@aaemnnosttv Thanks for the layout shift tip. I didn't know about that. :) Hover state has been added. About the class names, ideally they should have been in the following format: `googlesitekit-widget__xxx--yy` but to keep things consistent in the file, I used the same "convention"

felixarntz:
@asvinb @aaemnnosttv We should fully stick to the Figma design for the hover, focus and active styles. Specifically:
* The legend entries should never be underlined.
* The legend entries should never have that outline.

I agree though that having hover, focus, and active all look the same is not a great idea. Maybe we could use the bolder font-weight only when active?

Regarding the slight shift, we should try to fix that while maintaining the bold font-weight as per the design.

aaemnnosttv:
> I agree though that having hover, focus, and active all look the same is not a great idea. Maybe we could use the bolder font-weight only when active?

If by "active" you mean selected, that is how it is now:
image
- Direct `:hover`
- Organic Search: **selected** + `:focus`

> Regarding the slight shift, we should try to fix that while maintaining the bold font-weight as per the design.

That has already been fixed 😄

felixarntz (changes requested):
@asvinb Can you remove the underline on hover and the outline on focus? Afterwards, this should be good to go.

asvinb:
@aaemnnosttv @felixarntz Hover and focus styles removed. I also tweaked the code so that the "Others" legend is not clickable.

aaemnnosttv (approved):
Great, thanks @asvinb !

Issue #2716, opened 2021-01-28
"Others" slice in All Traffic widget should not be selectable


felixarntz:
The "Others" slice in the new All Traffic widget should not be selectable, since it doesn't alter the UI on the left at all. Indicating it as being selectable results in confusing UX.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Any "Others" slice displayed in the All Traffic widget pie chart should not be selectable, and it visually shouldn't be indicated as selectable.
* Its tooltip should still show when hovering.

## Implementation Brief
* Using `assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js`
	* Clear the chart's selection if the 'Others' slice is selected by calling:
		```js
		chart.setSelection( [] );
		```
	* Add a new state variable called `selectable`. This will act as a flag whether we want to have the chart slices to be selectable by adding `cursor: pointer` to the graph.
		```js
		const [ selectable, setSelectable ] = useState( false );
		```
	* Listen to the `onmouseover` event of the chart and set the chart to be not selectable if the label of the slice hovered is `Others`
		```js
		chartData.onMouseOver = global.google.visualization.events.addListener( chart, 'onmouseover', ( e ) => {
			const { row } = e;

			if ( ! row ) {
				setSelectable( false );
			}
			const { dataTable } = GoogleChart.charts.get( chartID ) || {};
			setSelectable( dataTable.getValue( row, 0 ) !== __( 'Others', 'google-site-kit' ) );
		} );
		```
	* Listen to the `onmouseout` event as well, to make the chart not selectable.
		```js
		chartData.onMouseOut = global.google.visualization.events.addListener( chart, 'onmouseout', () => {
			setSelectable( false );
		} );
		```
	* Conditionally add the `googlesitekit-widget--analyticsAllTraffic__selectable` class name to the container containing the chart if `selectable` is `true`.
		```js
			<div className={ classnames(
				'googlesitekit-widget--analyticsAllTraffic__dimensions-chart',
				{
					'googlesitekit-widget--analyticsAllTraffic__dimensions--loading': ! loaded,
					'googlesitekit-widget--analyticsAllTraffic__selectable': selectable,
				}
			) }>
		```
	* Update the code for the `select` event so that `setValues` is not called when `isOthers` is `true`.
* Using `assets/js/components/GoogleChart.js`
[...]

PR #2758, opened 2021-02-04
'Others' slice in All Traffic widget should not be selectable.

johnPhillips:
## Summary

Addresses issue #2716

## Checklist


tofumatt (dismissed):
Looks good. Just one suggestion about the `event` variable naming but I added it to reduce turnaround time 🙂

tofumatt:
@googlebot I consent.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
```diff
@@ -85,6 +90,20 @@ export default function UserDimensionsPieChart( { dimensionName, dimensionValue,
 				}
 			} );
 		}
+
+		chartData.onMouseOver = global.google.visualization.events.addListener( chart, 'onmouseover', ( e ) => {
+			const { row } = e;
```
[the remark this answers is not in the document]
johnPhillips: Good shout, thank you @tofumatt!

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
```diff
@@ -72,6 +73,10 @@ export default function UserDimensionsPieChart( { dimensionName, dimensionValue,
 						const newDimensionValue = dataTable.getValue( row, 0 );
 						const isOthers = __( 'Others', 'google-site-kit' ) === newDimensionValue;
 
+						if ( isOthers ) {
+							chart.setSelection( [] );
+						}
```
felixarntz, line 78: I think instead of adjusting the selection differently if `isOthers`, we rather need to not make _any_ change when the "Others" slice is clicked.

felixarntz (changes requested):
This is almost there, but one problem to address: When you select a slice and then click the "Others" slice, it still goes (back) to the "All Users" line chart view. Clicking the "Others" slice should not have _any_ effect on the UI.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
felixarntz: Hmm, maybe my previous advice was a bit too literally: Right now, the behavior is still incorrect, as you can visually select the "Others" slice right now. It needs to work the following way:
* If you click "Others" to select it when nothing else is selected, it should not be visually selected. For this, it looks like it's necessary to empty the current selection.
* However, if you click "Others" when another slice was previously selected, that other selection should remain (_that_ part was broken in my previous review).

So, instead of always emptying the selection when "Others" is clicked, we always need to maintain the previous selection basically. We can accomplish this as follows if `isOthers`:
* When a `dimensionValue` is set, set the selection to the respective row indicated by the current `dimensionValue`.
* Only when no `dimensionValue` is set, set the selection to an empty array.

felixarntz (changes requested):
There's now something else off here, I think we need to handle the "Others" selection more granularly.

johnPhillips:
Hi @felixarntz,

> - When a dimensionValue is set, set the selection to the respective row indicated by the current dimensionValue.
> - Only when no dimensionValue is set, set the selection to an empty array.

@tofumatt and I spent a bit of time trying to find a way to do this, but the issue is that `dimensionValue` gets passed into the `onready` callback at the start and then it doesn't have access to its updated values - it's always `undefined`. This means we can't determine what the current value is. 

Matt is currently working on a related issue that involves changes to the `useEffect` in this file, and he thinks he might have an idea he can try here. We will keep you updated.

felixarntz (approved):
@johnPhillips I went ahead and brought this PR back into its previous state where it always empties the selection when the "Others" slice is selected or unselected. While this is not all the way where we want it, it's already an improvement in UX, especially for showing the correct cursor.

#2714 requires some more work, and hopefully we'll eventually be able to reliably use Google charts.

Issue #2533, opened 2020-12-15
Check authentication service to determine remote-enabled features


felixarntz:
The `Feature_Flags` class currently only determines feature availability based on the current build mode. As part of the "Remote-controlled features" epic, this should be expanded to connect to the authentication service which will expose a new endpoint that for a given site returns the remotely controlled features and whether they're enabled.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a new utility class `Remote_Controlled_Features`. It should include a public method `get_features()` which issues a (cached) request to the authentication service (rely on the `Google_Proxy` class for the URL) as follows:
    * POST request to the proxy URL with path `/site-management/features/`
    * send URL-encoded body parameters `site_id`, `site_secret`, `platform` ("wordpress/google-site-kit"), and `version` (plugin version)
    * the response will be an object map of feature identifiers and their enabled/disabled state as boolean
    * The request should not be issued if the site is not using the proxy or if there are no site ID and site secret.
    * The response should be cached in a transient for 1 day.
* The new `Remote_Controlled_Features` class should include a `register` method that adds a filter for `googlesitekit_is_feature_enabled`.
    * It should return early if the site does not have `site_id` and `site_secret`.
    * For every feature that is included as a key/property in the API response object from the `get_features()` method, it should override the value and set it to `true` or `false` based on that response (every key/property contains an object with an `enabled` property that has that boolean). Any value that is not within the response is not a remotely controlled feature and should remain untouched, i.e. rely on its original value (coming from the `feature-flags.json` file).

Example response:
```json
{
    "userInput": { "enabled": true },
    "widgets.dashboard": { "enabled": false }
}
```

#### includes/Core/Authentication/Google_Proxy.php:

* Add a new constant `FEATURES_URI` with `'/site-management/features/'` value;
* Add a new public method `request( $uri, Credentials $credentials, array $args = array() )`
  * Return a new WP_Error with the `oauth_credentials_not_exist` code if the `$credentials->has()` call returns falsy result;
  * Create a new array `$request_args` that has empty arrays under `headers` and `body` keys;
  * Copy headers and body from the incoming `$args` array, if they are not empty, into appropriate `headers` and `body` properties of the `$request_args`;
  * If the incoming `$args['access_token']` variable is not empty, use it to compose the `Authorization` header (`'Bearer ' . $access_token`) and add it to the `$request_args['headers']` array;
[...]

PR #2759, opened 2021-02-04
Feature/2533 remote enabled features

eugene-manuilov:
## Summary

Addresses issue #2533

## Checklist


includes/Core/Authentication/Authentication.php
```diff
@@ -1277,4 +1277,28 @@ private function verify_user_input_settings() {
...
+	 * @param boolean $feature_enabled Original value of the feature.
+	 * @param string  $feature_name    Feature name.
+	 * @return boolean State flag from the proxy server if it is available, otherwise the original value.
+	 */
+	private function filter_features_via_proxy( $feature_enabled, $feature_name ) {
+		$transient_name = 'googlesitekit_remote_features';
+		$features       = $this->transients->get( $transient_name );
+		if ( empty( $features ) ) {
+			$features = $this->google_proxy->get_features( $this->credentials );
+			$this->transients->set( $transient_name, $features, DAY_IN_SECONDS );
```
adamsilverstein, line 1294: noticed that `$features` may be an error here if `google_proxy->get_features( $this->credentials );` returns an error. In that case, we may want to try again more quickly, maybe in an hour?

includes/Core/Authentication/Google_Proxy.php
```diff
@@ -78,6 +79,67 @@ public function url( $path = '' ) {
...
+
+		$code = wp_remote_retrieve_response_code( $response );
+		$body = wp_remote_retrieve_body( $response );
+		$body = json_decode( $body, true );
+		if ( $code < 200 || 299 < $code ) {
+			$message = is_array( $body ) && ! empty( $body['error'] ) ? $body['error'] : '';
+			return new WP_Error( 'request_failed', $message, array( 'status' => $code ) );
+		}
+
+		if ( ! $body ) {
```
adamsilverstein, line 136: This is throwing an error when the response is empty, eg:
`"body": "{}\n"` which is currently the case for the features endpoint (since no features are active). Wondering if we should make this check more explicitly look for a null? (`$body === null`). cc: @felixarntz

felixarntz: @eugene-manuilov See above. Let's maybe change this to a `is_array` check, since the service will always return an JSON array or object of data.

eugene-manuilov: @felixarntz @adamsilverstein yep, already updated it to check if it is `is_null( $body )`.

includes/Core/Authentication/Google_Proxy.php
```diff
@@ -78,6 +79,67 @@ public function url( $path = '' ) {
...
+		$code = wp_remote_retrieve_response_code( $response );
+		$body = wp_remote_retrieve_body( $response );
+		$body = json_decode( $body, true );
+		if ( $code < 200 || 299 < $code ) {
+			$message = is_array( $body ) && ! empty( $body['error'] ) ? $body['error'] : '';
+			return new WP_Error( 'request_failed', $message, array( 'status' => $code ) );
+		}
+
+		if ( is_null( $body ) ) {
+			return new WP_Error( 'failed_to_parse_response' );
```
felixarntz, line 137: @eugene-manuilov Let's always also add a message for `WP_Error`s, and also a status code via `$error_data`. Could use a simple `__( 'Failed to parse response.', 'google-site-kit' ), array( 'status' => 500 )` here.

includes/Core/Authentication/Authentication.php
eugene-manuilov: Yep, sounds good to me. Updated to be an hour for errors.

includes/Core/Authentication/Authentication.php
```diff
@@ -1279,4 +1279,29 @@ private function verify_user_input_settings() {
...
+	 * @since n.e.x.t
+	 *
+	 * @param boolean $feature_enabled Original value of the feature.
+	 * @param string  $feature_name    Feature name.
+	 * @return boolean State flag from the proxy server if it is available, otherwise the original value.
+	 */
+	private function filter_features_via_proxy( $feature_enabled, $feature_name ) {
+		$transient_name = 'googlesitekit_remote_features';
+		$features       = $this->transients->get( $transient_name );
+		if ( empty( $features ) ) {
```
adamsilverstein, line 1294: When features are empty (as is currently the case), this will store an empty array in the transient, which will pass the empty check, causing the plugin to check the endpoint repeatedly.

When the transient is expired, get_transient will return false, let's check for that explicitly here.

```suggestion
		if ( false === $features ) {
```

adamsilverstein:
@eugene-manuilov once final point came up in my final round of testing, otherwise looks good!

includes/Core/Authentication/Google_Proxy.php
```diff
@@ -282,23 +298,36 @@ public function exchange_site_code( $site_code, $undelegated_code ) {
...
-			throw new Exception(
-				isset( $response_data['error'] ) ? $response_data['error'] : 'failed_to_parse_response'
-			);
+		if ( is_wp_error( $response_data ) ) {
+			return $response_data;
 		}
 
 		if ( ! isset( $response_data['site_id'], $response_data['site_secret'] ) ) {
-			throw new Exception( 'oauth_credentials_not_exist' );
+			return new WP_Error( 'oauth_credentials_not_exist' );
```
felixarntz, line 306: I know this was missing before too, but let's add a corresponding message here now.

felixarntz (changes requested):
This is looking solid, just a bit of feedback.

includes/Core/Authentication/Authentication.php
eugene-manuilov: Updated.

includes/Core/Authentication/Authentication.php
```diff
@@ -1279,4 +1279,29 @@ private function verify_user_input_settings() {
...
+			$features = $this->google_proxy->get_features( $this->credentials );
+			$ttl      = is_wp_error( $features ) ? HOUR_IN_SECONDS : DAY_IN_SECONDS;
+			$this->transients->set( $transient_name, $features, $ttl );
```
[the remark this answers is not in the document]
eugene-manuilov: Updated to cache an empty array if the response is an error and to do nothing if we don't have credentials.

includes/Core/Authentication/Google_Proxy.php
eugene-manuilov: Updated.

includes/Core/Authentication/Google_Proxy.php
```diff
@@ -78,6 +79,67 @@ public function url( $path = '' ) {
...
+	 */
+	private function request( $uri, $credentials, array $args = array() ) {
+		$request_args = array(
+			'headers' => ! empty( $args['headers'] ) && is_array( $args['headers'] ) ? $args['headers'] : array(),
+			'body'    => ! empty( $args['body'] ) && is_array( $args['body'] ) ? $args['body'] : array(),
+		);
+
+		if ( $credentials && $credentials instanceof Credentials ) {
+			if ( ! $credentials->has() ) {
+				return new WP_Error( 'oauth_credentials_not_exist', __( 'OAuth credentials haven\'t been found.', 'google-site-kit' ), array( 'status' => 401 ) );
```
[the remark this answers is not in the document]
eugene-manuilov: The `! $credentials->has()` does check that there are actual credentials, doesn't it? Or I miss something?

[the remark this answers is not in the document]
felixarntz: @eugene-manuilov Yes that's correct; I was just referring to that we should check whether credentials are there before calling any of the `Google_Proxy` methods that attempt an API request, since otherwise we would cache data in a transient that would be incorrect, but not because an API request error but simply because we would have made a request that shouldn't have been attempted in the first place.

Essentially, we should always only make proxy API requests that require credentials when we know that there are already credentials.

With your latest changes, this looks good to go here.

felixarntz (approved):
Awesome!

I'll hold off merging until we have pushed the proxy-side part to production.

Issue #2648, opened 2021-01-19
Implement logic for triggering feature tour tooltips


felixarntz:
While #2626 introduces the essential JS functionality for rendering tooltip tours, we'll also need to implement a reliable way of detecting whether a specific tooltip tour should be triggered. For example:

* A tour specific to introducing a newly introduced feature to Site Kit should only be displayed to users that have already used the plugin before. A new user should not be confronted with various tooltip tours for features which for them have "always" been there.
* When a user completes or dismisses a tooltip tour, it should never be displayed to that user again, so it needs to be stored reliably in a persistent way. Using `localStorage` is infeasible, and even for the server-side storage it needs to be ensured that this storage does not get wiped on disconnect/reset like typical `googlesitekit_*` options are.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

### Exposing the user's initial Site Kit version in JS
* A new JS selector `getInitialSiteKitVersion()` should be added to the `core/user` store. It should receive the value of the `Core\Authentication\Initial_Version` option from #2692, which should be initially provided on pageload via PHP-to-JS data on `_googlesitekitUserData`.

### Storing a user's completed/dismissed tours
* A new user setting `googlesitekitpersistent_dismissed_tours` (array of strings; the `googlesitekitpersistent_` prefix must be used to prevent the option from being deleted on reset/disconnect) should be introduced and registered.
* A new REST datapoint `GET:core/user/data/dismissed-tours` should be introduced. It should return an array of dismissed tour slugs (basically the current value of the current user's `googlesitekitpersistent_dismissed_tours` option). This endpoint should be added to the REST endpoints to be preloaded on pageload.
* A new REST datapoint `POST:core/user/data/dismiss-tour` should be introduced which requires a `slug` parameter. It should add the tour identified by `slug` to the current user's `googlesitekitpersistent_dismissed_tours` option. It should return the updated list of dismissed tours.
* A new JS selector `getDismissedTours()` should be added to the `core/user` store, relying on the response from `GET:core/user/data/dismissed-tours`.
* A new JS selector `isTourDismissed( slug )` should be added to the `core/user` store, relying on the `getDismissedTours()` selector to check whether the tour identified by `slug` is included in that list.
* A new JS action `dismissTour( slug )` should be added to the `core/user` store, relying on calling the `POST:core/user/data/dismiss-tour` datapoint. It should also update the dismissed tours stored on the client, via the response from the API datapoint.

### PHP 

[...]

PR #2763, opened 2021-02-05
Add foundation for dismissing feature tours

aaemnnosttv:
## Summary

Addresses issue #2648

## Checklist


includes/Core/Feature_Tours/REST_Feature_Tours_Controller.php
```diff
@@ -0,0 +1,149 @@
...
+	 * Gets REST route instances.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return REST_Route[] List of REST_Route objects.
+	 */
+	protected function get_rest_routes() {
+		$can_authenticate = function () {
+			return current_user_can( Permissions::AUTHENTICATE );
+		};
```
aaemnnosttv, line 105: This is different than I had in the IB which was `Permissions:VIEW_DASHBOARD`. This permission requires being authenticated which I don't think is necessary for this. Any user who can authenticate (and thus use Site Kit) should be able to see and dismiss tours. We can always change this of course, but this was my thought in changing it.

includes/Core/Feature_Tours/Feature_Tours.php
```diff
@@ -0,0 +1,81 @@
...
+use Google\Site_Kit\Core\Storage\User_Options;
+
+/**
+ * Class for handling feature tours.
+ *
+ * @since n.e.x.t
+ * @access private
+ * @ignore
+ */
+class Feature_Tours {
```
aaemnnosttv, line 23: This class seems a bit thin as it is, but I don't think that's necessarily a bad thing. I definitely prefer separating out the controller so this is just a small feature as it is 😄

includes/Core/Feature_Tours/REST_Feature_Tours_Controller.php
```diff
@@ -0,0 +1,149 @@
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Context      $context Plugin context.
+	 * @param User_Options $user_options Optional. User option API. Default is a new instance.
+	 */
+	public function __construct( Context $context, User_Options $user_options = null ) {
+		$this->context         = $context;
+		$this->user_options    = $user_options ?: new User_Options( $context );
+		$this->dismissed_tours = new Dismissed_Tours( $this->user_options );
```
aaemnnosttv, line 67: The IB says the `Dismissed_Tours` instance would be an optional constructor dependency but I followed the example we have in `Authentication` and other places where only the API instances are passed via the constructor, and these "lesser entities" are constructed internally. It would be great to make this more flexible with a DI container in the future but this is fine for now I think 👍

assets/js/googlesitekit/datastore/user/feature-tours.js
```diff
@@ -0,0 +1,185 @@
...
+				payload: { slug },
+				type: DISMISS_TOUR,
+			};
+			// Dispatch a request to persist the dismissal.
+			const { response, error } = yield fetchDismissTour( slug );
+			if ( ! error ) {
+				yield receiveGetDismissedTours( response );
+			}
+			return { response, error };
+		}() );
```
felixarntz, line 94: Why not make this a generator action like we have elsewhere for these? What is the benefit of this approach? From a code readability perspective this seems unnecessarily complicated.

assets/js/googlesitekit/datastore/user/feature-tours.js
```diff
@@ -0,0 +1,185 @@
...
+	 * Checks whether or not the given tour is dismissed.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @param {string} slug  Tour slug to check.
+	 * @return {boolean} `true` if dismissed, otherwise `false`.
+	 */
+	isTourDismissed: createRegistrySelector( ( select ) => ( state, slug ) => {
+		const dismissedTours = select( STORE_NAME ).getDismissedTours() || [];
```
felixarntz, line 154: Should we maybe return `true` if `getDismissedTours` hasn't resolved yet? When deciding whether a tour should be displayed, the current behavior could result the tour UI to briefly pop up before "recognizing" that it actually was already dismissed.

includes/Core/Feature_Tours/Dismissed_Tours.php
```diff
@@ -0,0 +1,95 @@
...
+	const OPTION = 'googlesitekitpersistent_dismissed_tours';
+
+	/**
+	 * Adds one or more tours to the list of dismissed tours.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string ...$tour_slug The tour identifier to dismiss.
+	 */
+	public function add( ...$tour_slug ) {
```
felixarntz, line 40: This doesn't seem necessary at this point. I'm not opposed to keeping it, but in general I prefer separate explicit methods for one vs multiple.

includes/Core/Feature_Tours/Feature_Tours.php
felixarntz: Yeah I like it, this class is basically the PHP entry point to the entire feature, so that `Plugin` has to only initialize that rather than every "sub-class" of the feature.

felixarntz (commented):
This looks great! I left a couple of feedback and some questions - maybe nothing blocking, but especially the `dismissTour` action implementation I wanted to discuss first.

assets/js/googlesitekit/datastore/user/feature-tours.js
aaemnnosttv: @felixarntz It's needed for validation to work as expected. If you recall, we do the same thing in `createFetchStore` this is just the inline version of it:
https://github.com/google/site-kit-wp/blob/fb2bbdf94d435753a3c84f2a1064108144e390df/assets/js/googlesitekit/data/create-fetch-store.js#L176-L185

assets/js/googlesitekit/datastore/user/feature-tours.js
aaemnnosttv: Good point. Similar to other selectors, I don't think we should return `true` or `false` if the dependent selector hasn't been resolved yet. Let's return `undefined` if `getDismissedTours` is `undefined` so that this only returns a boolean if there is state to compute it, in which case it would always be accurate.

includes/Core/Feature_Tours/Dismissed_Tours.php
aaemnnosttv: Are you referring to the method as a whole or just the ability to add more than one at a time?

I see this as similar to `Settings::merge` in that you can set multiple "inner values" in one go, rather than requiring a separate DB query for each addition. The method body is basically the same to add a single value as it is for multiple as well.

assets/js/googlesitekit/datastore/user/feature-tours.js
felixarntz: Ah right, makes sense.

includes/Core/Feature_Tours/Dismissed_Tours.php
felixarntz: I'm not opposed to having a method to dismiss multiple tours in one go (one DB update), but I meant that I'd prefer to have a separate method to be more explicit in usage. `Module_Settings::merge` only works with an array of values, while this here supports any number of values. I guess that's just a personal preference, but I'm not a large fan of using variable parameters for this purpose, rather would go with `add( $tour_slug )` and something like `add_multiple( $tour_slugs )`; or alternatively not have anything for single and only have `add( $tour_slugs )`.

The way it is now is okay, my other point was also that we just don't need it at this point since the API doesn't support dismissing multiple tours at once anyway, so we could basically avoid this discussion and only add `add( $tour_slug )` for now. But let's go with this as is, not heavily opposed.

felixarntz (dismissed):
Awesome!

felixarntz (approved):

Issue #2688, opened 2021-01-26
Remove custom retry map for exponential retries.


adamsilverstein:
## Bug Description

In https://github.com/google/site-kit-wp/pull/2442 / https://github.com/google/site-kit-wp/commit/b1b73c93215f0cd1eac88df8a88dfc7ae561fb7c we added a custom retry map for exponential retries. The main reason we added this was to avoid retrying for lightouseErrors which can return a 500 code even when not retryable.

Since that time, I opened a PR to add this config to the upstream library which has now been merged: https://github.com/googleapis/google-api-php-client/pull/2010. The client config now matches ours, so we can remove our custom config:

https://github.com/adamsilverstein/google-api-php-client/blob/b4aadd3a71213a19a9076af90a1cbef85eec8a83/src/Task/Runner.php#L75-L86

## Additional Context

 - PHP Version: 
 - Plugin Version [e.g. 22]

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The exponential retry feature should use the default configuration.
* The PHP API client should be running at least version 2.9.0 (when the upstream change was released)

## Implementation Brief

* In `/includes/Core/Authentication/Clients/OAuth_Client.php`:
  * Remove the retry map array.
  * Remove the line setting the retry config: `$client->setConfig( 'retry_map', $this->retry_map );`
* In `composer.json` update the version for `google/apiclient` to the [latest released version](https://github.com/googleapis/google-api-php-client/releases). (currently 2.9.1). 

### Test Coverage

* No change.

### Visual Regression Changes

* No change.

## QA Brief

* Engineer to confirm the retry map has been removed.

## Changelog entry

* Update Google API client library and remove custom configuration to retry failed API requests as it is now covered in the library itself.

felixarntz:
Noting here that we'll need to require at least version 2.9.0 of the `google-api-php-client` to make this change.

felixarntz:
@adamsilverstein Can you update the ACs and IB based on the above? Afterwards should be good.

adamsilverstein:
👍🏼 Updated.

felixarntz:
IB ✅

benbowler:
Added a PR and QA:Eng line.

tofumatt:
QA ✅

PR #2769, opened 2021-02-08
Remove custom retry map for exponential retries.

benbowler:
## Summary

Addresses issue #2688

## Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
     "php": ">=5.6",
     "ext-json": "*",
     "ext-openssl": "*",
-    "google/apiclient": "^2.0",
+    "google/apiclient": "^2.9.1",
     "google/apiclient-services": ">=0.157",
     "guzzlehttp/guzzle": "^5.3.3"
   },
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -17,7 +17,7 @@
     "php": ">=5.6",
     "ext-json": "*",
     "ext-openssl": "*",
-    "google/apiclient": "^2.0",
+    "google/apiclient": "^2.9.1",
```
eugene-manuilov, line 20: @benbowler changing the version manually in the composer.json file is not enough because in this case the composer.lock file remains the same. We need to update the dependency version correctly which will mean that the lock file should be updated as well. 

Just run `composer remove google/apiclient` to remove the existing version and run `composer require google/apiclient` to install the latest one. It will correctly update composer.json and composer.lock files.

eugene-manuilov (changes requested):

composer.json
benbowler: Good point, I failed to commit the `composer.lock`. Pushed up now.

eugene-manuilov (approved):
Awesome! LGTM.

felixarntz (approved):
Great!

Issue #2738, opened 2021-02-01
Fix All Traffic pie chart legend and line chart x-axis labels to vertically align


felixarntz:
This was previously _attempted_ in #2708, but did not resolve the issue. This is a separate follow-up issue because we didn't have time to make this work in the 1.25.0 release.

Please have a look at #2712: At some point, the legend and the x-axis labels properly aligned. However, that fix introduced a super-large extra whitespace below which made the UI look worse than before. When that margin was removed again, we missed noticing that this once again made the two pieces unaligned 🤦‍♂️ 

Let's find a _proper_ solution here that does not introduce any other problems.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The x-axis labels of the All Traffic line chart should vertically align with the legend of the All Traffic pie chart.
* This must not introduce any other visual regressions. :)

## Implementation Brief

After trying to make the sizes work with set values for way too much time, it seems a set value or a percentage value won't work properly to line up with the pie chart for the Analytics All Traffic widget. I think the best bet is to change the height of the `UserCountGraph` component's `GoogleChart` component based on the window size.

* Use [`useWindowSize`](https://github.com/streamich/react-use/blob/master/docs/useWindowSize.md) to get the size of the window and set the chart height according to the current viewport size.

From my testing something around:

```js
	height: 350,
	width: '100%',
	colors: [ '#1a73e8' ],
	chartArea: {
		left: '1%',
		height: 300,
		top: 18,
		width: '90%',
	},
```

works well for both a desktop and a large desktop viewport, but adjustments need to be made to line up with the legends for both viewports.

* Use the [existing CSS breakpoints](https://github.com/google/site-kit-wp/blob/c61cc0847f47f5c330024a1f4db00803dfa47831/assets/sass/config/_variables.scss#L157-L162) as reference points for when to adjust the `height` vaules of the chart.

### Test Coverage

* No test coverage needed.

### Visual Regression Changes

* Adding VRT tests for the new widget that ensures the legends line up would be good.

## QA Brief

* The legends should vertically align on both smaller and larger desktop screens.

## Changelog entry

* Ensure x-axis labels of the All Traffic line chart are correctly aligned with the pie chart legend.

eugene-manuilov:
> The x-axis labels of the All Traffic line chart should vertically align with the legend of the All Traffic pie chart.

[...]

PR #2770, opened 2021-02-08
Align All Traffic pie chart legend and line chart x-axis labels. (#2738)

johnPhillips:
## Summary

Addresses issue #2738

## Relevant technical choices
I felt that it was worth briefly investigating a CSS solution, intending to follow the IB if I couldn't find one. 

Messing around in-browser, I found that on the current release a couple of changes on the parent `.googlesitekit-widget--analyticsAllTraffic__dimensions` div lined the legends up correctly, but the issue was that in `develop` there have been [a few changes which made it less straightforward](https://github.com/google/site-kit-wp/commit/13348bdbee4a44e12b20b0760a8e570e0d98385e). 

I therefore removed them and then attempted to add them back in one at a time to get them to coexist with the changes I made. I've done a fair bit of cross browser testing (FF, Safari, Chrome, Brave) - hopefully I haven't introduced a regression, and both issues are still solved 🤞.

I'm assuming a CSS solution was preferable - I'm very happy to attempt the method in the IB if needed.

Feb-08-2021 15-11-02

## Checklist

### DIFF
--- a/assets/sass/widgets/_googlesitekit-widget-analyticsAllTraffic.scss
+++ b/assets/sass/widgets/_googlesitekit-widget-analyticsAllTraffic.scss
@@ -50,7 +50,6 @@
 
 		@media (min-width: $bp-desktop) {
 			position: relative;
-			top: 16px;
 		}
 
 		@media (min-width: $bp-xlarge) {
@@ -73,6 +72,9 @@
 	}
 
 	.googlesitekit-widget--analyticsAllTraffic__dimensions {
+		display: flex;
+		flex-direction: column;
+		justify-content: space-between;
 		min-height: 450px;
 		padding-top: 30px;
 		position: relative;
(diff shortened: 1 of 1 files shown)


eugene-manuilov:
I am not sure about this approach because it pushes the pie chart down which makes it look less natural to me. @tofumatt can you review it on your end and tell if it is okay or not?

tofumatt (approved):
When I tested this locally and compared it to `develop` the Pie Chart wasn't pushed down for me in Chrome or Firefox on macOS, and it led to everything being lined up. I much prefer the CSS-only solution, and it worked at many viewpoints.

This works for me, I think it looks good. 👍🏻

aaemnnosttv (changes requested):
Thanks @johnPhillips – I think this can work but needs a little tweaking.

It looks good in larger viewports, but the alignment is lost at **1279px** and below, until it breaks onto two rows. This seems to be the same breakpoint where the font size of the main total users metric decreases as well.
image

felixarntz:
Superseded by #2783.

PR #2775, opened 2021-02-09
Fix/e2e logging

aaemnnosttv:
## Summary

Addresses issue #2777

## Relevant technical choices

* This fixes a few E2E failures as well which were easier to fix now that we can see what the failure was

## Checklist

### DIFF
--- a/bin/local-env/install-wordpress.sh
+++ b/bin/local-env/install-wordpress.sh
@@ -127,6 +127,8 @@ container chmod 767 /var/www/html/favicon.ico
 # Activate Google Site Kit plugin.
 status_message "Activating Google Site Kit plugin..."
 wp plugin activate google-site-kit --quiet
+# Reset post-activate state.
+wp google-site-kit reset
 
 # Set pretty permalinks.
 status_message "Setting permalink structure..."
--- a/tests/e2e/specs/modules/search-console/admin-bar.test.js
+++ b/tests/e2e/specs/modules/search-console/admin-bar.test.js
@@ -42,6 +42,11 @@ describe( 'Site Kit admin bar component display', () => {
 					status: 200,
 					body: JSON.stringify( mockBatchResponse[ 'modules::search-console::searchanalytics::e74216dd17533dcb67fa2d433c23467c' ] ),
 				} );
+			} else if ( request.url().match( 'google-site-kit/v1/data/' ) ) {
+				request.respond( {
+					status: 200,
+					body: JSON.stringify( mockBatchResponse ),
+				} );
 			} else {
 				request.continue();
 			}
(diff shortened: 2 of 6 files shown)


bin/local-env/install-wordpress.sh
```diff
@@ -127,6 +127,8 @@ container chmod 767 /var/www/html/favicon.ico
 # Activate Google Site Kit plugin.
 status_message "Activating Google Site Kit plugin..."
 wp plugin activate google-site-kit --quiet
+# Reset post-activate state.
+wp google-site-kit reset
```
aaemnnosttv, line 131: This fixes a potential error from showing up in the log which may or may not have caused a failure before but has been seen many times in the E2E log as part of the tagmanager setup test, but it's unrelated to that. It was only happening here because it was the first test that always runs.

This was due to compatibility checks that ran the very first time the `plugins.php` screen loaded due to our activation flag that controls the presence of the activation notification (which runs compatibility checks). To fix this, we simply reset the plugin after the initial activation. We already do a plugin reset in the bootstrap.js in the main `beforeAll` but it's too late as it's after the call to deactivate utility plugins. I thought about moving or changing this, but the above change is just simpler, and actually gives us slightly better coverage in that we actually use our reset command.

tests/e2e/specs/modules/search-console/admin-bar.test.js
```diff
@@ -42,6 +42,11 @@ describe( 'Site Kit admin bar component display', () => {
 					status: 200,
 					body: JSON.stringify( mockBatchResponse[ 'modules::search-console::searchanalytics::e74216dd17533dcb67fa2d433c23467c' ] ),
 				} );
+			} else if ( request.url().match( 'google-site-kit/v1/data/' ) ) {
+				request.respond( {
+					status: 200,
+					body: JSON.stringify( mockBatchResponse ),
+				} );
```
aaemnnosttv, line 49: This was removed in a recent PR for the admin bar, but should have been kept. It's been restored in the other admin bar PR but I've added it here to for completeness.

tests/e2e/specs/modules/tagmanager/setup.test.js
```diff
@@ -59,6 +59,11 @@ describe( 'Tag Manager module setup', () => {
 				request.respond( {
 					status: 200,
 				} );
+			} else if ( request.url().match( 'google-site-kit/v1/core/site/data/notifications' ) ) {
+				request.respond( {
+					status: 200,
+					body: JSON.stringify( [] ),
+				} );
```
aaemnnosttv, line 66: This was the request that was often causing the test to fail as an unexpected API failure (or at least one of them 😄 ).

felixarntz (approved):
Awesome!

PR #2779, opened 2021-02-09
Update selection properly with GoogleChart component.

tofumatt:
## Summary

Addresses issue #2714.

## Relevant technical choices

After what started as a straightforward fix I ran into issues with the GoogleChart component and ultimately with it not reliably setting the selected value after being drawn.

The only reliable solution I've found, given our current implementation and our lack of a function to run after Google Charts has drawn, was to use `useEffect` + `setTimeout(() => {}, 0)`. I'm not in love with this solution, but on the other hand it seems to work _very_ well in my testing.

I think going forward we need to rethink how we're interacting with the non-React-aware `GoogleChart` library and our own components. Right now it seems shakey.

But this fixes all selection issues I've been seeing.

## Checklist

### DIFF
--- a/assets/js/components/GoogleChart.js
+++ b/assets/js/components/GoogleChart.js
@@ -25,7 +25,7 @@ import debounce from 'lodash/debounce';
  * WordPress dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { useEffect, useState, useRef, useLayoutEffect } from '@wordpress/element';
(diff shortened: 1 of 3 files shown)


felixarntz:
Closing this per https://github.com/google/site-kit-wp/issues/2714#issuecomment-776326854. I think we should invest our time into a better solution instead of adding another hack to all the existing hacks :)

Issue #2103, opened 2020-09-24
Tracking opt-in request fails on multisite for super admins that don't belong to the site


aaemnnosttv:
## Bug Description

On a regular multisite network, Site Kit can be configured per-site just like on a normal single site install.

When Site Kit is active on a sub-site on the network, the REST request used when toggling anonymous usage tracking fails with a 404 when the user is not a member of the current site. This is only possible for super admins who can access all sites across the network without being added.

## Steps to reproduce

1. Activate Site Kit on a sub-site of a multisite network
1. Log in as a super admin who is not added to the sub-site (`wp-admin/network/site-users.php`)
1. Go to the Site Kit splash page
1. Toggle tracking opt-in
1. See "Invalid user ID" error

## Screenshots

![image](https://user-images.githubusercontent.com/1621608/94190599-fc60eb80-feb4-11ea-8c0e-9fac58721553.png)

![image](https://user-images.githubusercontent.com/1621608/94191834-ce7ca680-feb6-11ea-82bf-eaed19ecd544.png)

## Additional Context

This only seems to affect the `wp/v2/users/me` route we use for this. Other Site Kit routes seem to work fine.

The actual response from the server:

```json
{
  "code": "rest_user_invalid_id",
  "message": "Invalid user ID.",
  "data": { "status": 404 }
}
```

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a new `GET:core/user/data/tracking` datapoint in PHP which returns an object like `{ enabled: true|false }`, based on the option's value.
    * The response for this REST endpoint should be preloaded on pageload.
    * Note: `trackingEnabled` should still be included in `_googlesitekitBaseData` since it's needed for the general admin tracking logic. Since we don't load all our JS code (e.g. the datastores) on all WP admin pages (which should be kept that way), for the basic tracking we need to continue to rely on the PHP-to-JS object data.
* There should be a new `POST:core/user/data/tracking` datapoint in PHP which expects an `enabled` argument, updates the option based on it, and then returns an object like `{ enabled: true|false }`, based on the updated value.
* There should be a new JS datastore selector `isTrackingEnabled()` in the `core/user` store which relies on the `GET:core/user/data/tracking` datapoint.
* There should be a new JS datastore action `setTrackingEnabled( enabled )` in the `core/user` store which sends the data to `POST:core/user/data/tracking`.
[...]

PR #2795, opened 2021-02-11
Bug/2103 opt in request fails

eugene-manuilov:
## Summary

Addresses issue #2103

## Checklist


includes/Core/Util/Tracking.php
```diff
@@ -106,4 +122,67 @@ private function inline_js_base_data( $data ) {
...
+	 * Gets tracking routes.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $routes Array of routes.
+	 * @return array Modified array of routes that contains tracking related routes.
+	 */
+	private function get_rest_routes( $routes ) {
+		$valid_user = function() {
+			return is_user_logged_in();
```
felixarntz, line 136: This should rather use a Site Kit base capability, e.g. whether the user can authenticate. Not anybody should be able to use this route (e.g. non-administrators right now).

assets/js/googlesitekit/datastore/user/tracking.js
```diff
@@ -0,0 +1,154 @@
...
+const baseActions = {
+	/**
+	 * Saves user tracking settings.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {boolean} enabled Tracking status.
+	 * @return {Object} Object with `response` and `error`.
+	 */
+	*saveUserTracking( enabled ) {
```
felixarntz, line 63: This name conflicts with the ACs where this is outlined as `setTrackingEnabled`. Let's go with that as described there. Alternatively, we could define our actions (and selectors) more closely to the REST endpoint, e.g.:
* `setTracking( data )` as low-level (private) action
* utility action `enableTracking()` which calls `setTracking( { enabled: true } )`
* utility action `disableTracking()` which calls `setTracking( { enabled: false } )`
* same for the selectors, we could have `getTracking()` which returns the full object, and `isTrackingEnabled()` which only returns the `enabled` property value

Anyway, let's use "set" throughout here instead of "save", as there is no need for differentiation, and "set" is a more intuitive action name that we use in most places.

assets/js/components/OptIn.js
```diff
@@ -21,119 +21,75 @@
...
+		<div className={ classnames( 'googlesitekit-opt-in', className ) }>
+			<Checkbox
+				id={ id }
+				name={ name }
+				value="1"
+				checked={ enabled }
+				disabled={ saving }
+				onChange={ handleOptIn }
+			>
+				<span dangerouslySetInnerHTML={ sanitizeHTML( labelHTML, allowedDOM ) } />
```
felixarntz, line 83: While we're refactoring this, we should be able to clean this up by using `createInterpolateElement` instead.

assets/js/components/OptIn.js
```diff
@@ -21,119 +21,75 @@
...
-			optIn,
-			error,
-		} = this.state;
 
-		const {
-			id,
-			name,
-			className,
-		} = this.props;
+		saveUserTracking( checked );
```
felixarntz, line 55: We need to look at the response value here. If there is an error and the tracking value wasn't saved, we need to re-set `toggleTracking` to the previous checked value, since that part is decoupled from the datastore, since it should be able to run agnostically without all of React etc.

Maybe it would even be cleaner to do that as a `useEffect` based on `enabled`, i.e. handle the `toggleTracking` and `trackEvent` side effects based on value changes of that (since it will only change when the request is successful). If we do that, we just need to make sure it doesn't fire on the _initial_ call when data was just loaded.

felixarntz (changes requested):
This looks solid overall, but there are a couple things to iterate on.

includes/Core/Util/Tracking.php
eugene-manuilov: Updated.

assets/js/googlesitekit/datastore/user/tracking.js
eugene-manuilov: Updated to use names from the ACs.

assets/js/components/OptIn.js
eugene-manuilov: Updated.

assets/js/components/OptIn.js
eugene-manuilov: Updated to use the response of the saving tracking enabled response.

felixarntz (approved):
Awesome!

Issue #2095, opened 2020-09-23
Vertical Axis displays format inconsistent with tooltip


ryanwelcher:
The vertical axis for in `AnalyticsDashboardWidgetSiteStats` displays standard numbers which is inconsistent with what it being displayed in the tooltip.

The most obvious example is the Session Duration. The tool tip displays `mm:ss` while the vertical axis displays seconds:
![sitekit_session_duraction.png](https://images.zenhubusercontent.com/5e74b5396e1fdc722577f6d3/bf8244df-34c0-4807-a9a4-a47bdd67c382)
It may also be an improvement to add a % sign after the label for Bounce Rate as well.
![sitekit_bounce_rate.png](https://images.zenhubusercontent.com/5e74b5396e1fdc722577f6d3/59cc3c23-0bd9-4119-9852-76759f9218d0)
This may be a limitation of the [API](https://developers.google.com/chart/interactive/docs/customizing_axes#number-formats), but further discovery is required to determine if we can format the labels appropriately.

Discovered during QA for #2008.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The Analytics graph on the Analytics module page should have its y-axis labels number formatting based on the values that the graph displays (relying on `numFmt` if possible):
    * For the "Bounce Rate" the values should be formatted as percentages.
    * For the "Session Duration" the values should be formatted as durations. It would be great if we could display these labels in multitudes of minutes instead of seconds when applicable (e.g. "1m", "2m", instead of "50", "100", the latter of which do not correspond to a full minute).

## Implementation Brief

We currently show Session Duration in the graph as seconds, however charts can take a 'timeofday', https://developers.google.com/chart/interactive/docs/datesandtimes#working-with-timeofday which takes an array of integers of the following format:
`[HOURS, MINUTES, SECONDS, MILLIS]`

This function is used for formatting the data:
https://github.com/google/site-kit-wp/blob/fa3df25c8c6cb11ab600262174c7fbfbcbaf7692/assets/js/modules/analytics/util/index.js#L164

and it is called from here to render the chart https://github.com/google/site-kit-wp/blob/fa3df25c8c6cb11ab600262174c7fbfbcbaf7692/assets/js/modules/analytics/components/dashboard/AnalyticsDashboardWidgetSiteStats.js#L121

For Sesson Duration data, we need to change it from 
`{"type":"number","label":"Session Duration"}`
to
[...]

PR #2800, opened 2021-02-12
Enhancement/2095 vertical axis

asvinb:
## Summary

Addresses issue #2095 

## Checklist


assets/js/util/i18n.js
```diff
@@ -57,6 +57,36 @@ export const prepareSecondsForDisplay = ( seconds ) => {
...
+	}
+
+	return [
+		Math.floor( seconds / 60 / 60 ),
+		Math.floor( ( seconds / 60 ) % 60 ),
+		Math.floor( seconds % 60 ),
+		parseInt( parseFloat( seconds ).toFixed( 3 ).slice( -3 ) ),
+	];
+};
+
```
tofumatt, line 89: The logic here seems fine, but two notes:

1. I don't think this belongs in the `i18n` utils file. It's not i18n-related—it can either go in a date/time-related file or in its own file, whichever makes more sense.
2. It'd be good to add a few simple unit tests for this to ensure it's behaving as expected

tofumatt (changes requested):
Approach seems good, but the placement of the helper function seems wrong to me. After it's moved (and tests are added as mentioned) we're good-to-go 👍🏻

assets/js/util/i18n.js
asvinb: @tofumatt I've created a new file, `convert-time.js` and moved the function over there. I also added the tests.

tofumatt (approved):
Lovely! Thanks for adding the tests 👍🏻

assets/js/util/convert-time.js
```diff
@@ -0,0 +1,47 @@
...
+
+	if ( isNaN( seconds ) || 0 === seconds ) {
+		return [ 0, 0, 0, 0 ];
+	}
+
+	return [
+		Math.floor( seconds / 60 / 60 ),
+		Math.floor( ( seconds / 60 ) % 60 ),
+		Math.floor( seconds % 60 ),
+		parseInt( parseFloat( seconds ).toFixed( 3 ).slice( -3 ) ),
```
aaemnnosttv, line 45: Rather than parsing this back and fourth and using string operations, we can get the decimal value as an integer up to the thousandths precision with some basic math like so:
```suggestion
		Math.floor( seconds * 1000 ) - Math.floor( seconds ) * 1000,
```

For example: if `seconds` were `123.4567`, this would result in the following:
```
123456 - 123000 = 456
```

assets/js/util/convert-time.js
```diff
@@ -0,0 +1,47 @@
...
+ * SS: seconds.
+ * MS: milliseconds.
+ *
+ * For example, passing 196.385 returns [ 0, 3, 16, 385 ].
+ *
+ * @since n.e.x.t
+ *
+ * @param {number} seconds The number of seconds.
+ * @return {Array} Array containing the hours, minutes, seconds and milliseconds.
+ *
```
aaemnnosttv, line 32: Extra line 😄 
```suggestion
```

aaemnnosttv (commented):
LGTM, just one suggestion around the added utility.

assets/js/util/convert-time.js
asvinb: Updated @aaemnnosttv

assets/js/util/convert-time.js
asvinb: nice. updated!

aaemnnosttv (approved):
Super, thanks!

Issue #2740, opened 2021-02-01
Add Analytics events for interacting with All Traffic widget


felixarntz:
We should add Analytics events for certain interactions with the new All Traffic widget that was introduced in 1.25.0.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The following Analytics events should be sent when interacting with the new All Traffic widget:
    * `tab_select`: User selects a specific tab (event label should be the dimension name, e.g. `ga:device`)
    * `slice_select`: User selects a specific pie slice (event label should be the dimension name and dimension value, e.g. `ga:device:tablet`)
    * `help_click`: User clicks "(other)" or "(not set)" help link in tooltip (event label should be `(other)` or `(not set)`)
    * `others_source_click`: User clicks on the Analytics deep link in the "Others" tooltip (no event label)
* For all of these events, the event category should be `all_traffic_widget`.

## Implementation Brief
* Using `/assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/DimensionTabs.js`
	* Import `trackEvent` from `../../../../../util`
	* Update the `handleTabUpdate` function to call the `trackEvent` function with the following values:
		* Category: **all_traffic_widget**
		* Action: **tab_select**
		* Label: The selected dimension name.
* Using `/assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js`
	* Import `trackEvent` from `../../../../../util`
	* Update the `onSelect` callback of the chart to track when a slice is selected using the values below. `trackEvent` function should be called only then a slice is selected, i.e when dimension name is set, and not when deselecting a slice.
		* Category: **all_traffic_widget**
		* Action: **slice_select**
		* Label: The selected dimension name and dimension value, separated by **:**. For e.g `ga:device:tablet`.
	* Update the `getTooltipHelp` function to:
		* Add a new param to the function, `rowLabel` which will populate a new `data-row-label` attribute on the anchor tag.
		* Add a new class name to the anchor tag, `googlesitekit-cta-link__tooltip`.
	* Update calls where `getTooltipHelp` is called, to add the `rowLabel` as the third parameter.
[...]

PR #2805, opened 2021-02-12
Create a GoogleChartV2 component.

tofumatt:
## Summary

Addresses issue #2714.

## Relevant technical choices

This creates a new GoogleChartV2 component that integrates the more modern "PreviewBlock"-style loading animations for the chart. It also allows the developer to create `ref`s to the `chart` value from the parent function, so we no longer need to rely on `GoogleChart.charts.get( chartID )` in our code.

I explored unit testing the component, but the reliance on DOM and on loading the charts library and rendering meant it was a nightmare, see: https://github.com/rakannimer/react-google-charts/issues/214

Best to rely on VRT and E2E tests for this one 🙂 

## Checklist


assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
```diff
@@ -297,71 +286,97 @@ export default function UserDimensionsPieChart( { dimensionName, dimensionValue,
...
+							<span className="googlesitekit-widget--analyticsAllTraffic__legend-slice">
+								<span className="googlesitekit-widget--analyticsAllTraffic__dot" style={ { backgroundColor: '#ccc' } } />
+								<PreviewBlock height="18px" width="40px" shape="square" />
+							</span>
+
+							<span className="googlesitekit-widget--analyticsAllTraffic__legend-slice">
+								<span className="googlesitekit-widget--analyticsAllTraffic__dot" style={ { backgroundColor: '#ccc' } } />
+								<PreviewBlock height="18px" width="52px" shape="square" />
+							</span>
+						</Fragment>
```
eugene-manuilov, line 377: How about this?

```javascript
new Array( 4 ).fill().map( ( _, i ) => (
    <span key={ i } className="googlesitekit-widget--analyticsAllTraffic__legend-slice">
        <span className="googlesitekit-widget--analyticsAllTraffic__dot" style={ { backgroundColor: '#ccc' } } />
        <PreviewBlock height="18px" width="68px" shape="square" />
    </span>
) )
```

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
```diff
@@ -297,71 +286,97 @@ export default function UserDimensionsPieChart( { dimensionName, dimensionValue,
...
+							</Link>
+						);
+					} ) }
+
+					{ ! loaded && (
+						<Fragment>
+							<span className="googlesitekit-widget--analyticsAllTraffic__legend-slice">
+								<span className="googlesitekit-widget--analyticsAllTraffic__dot" style={ { backgroundColor: '#ccc' } } />
+								<PreviewBlock height="18px" width="68px" shape="square" />
+							</span>
```
eugene-manuilov, line 361: PreviewBlock renders a div which means that we will have divs nested in spans. Let's use `div` tags for legend preview items instead.

eugene-manuilov (changes requested):
Awesome work, @tofumatt! Added some comments to address some issues. Could you please take a look at it?

assets/js/components/GoogleChartV2.js
```diff
@@ -0,0 +1,164 @@
...
+
+		if ( onMouseOut ) {
+			// eslint-disable-next-line no-unused-expressions
+			googleRef.current?.visualization.events.addListener( chartWrapperRef.current.getChart(), 'onmouseout', ( event ) => {
+				onMouseOut( event, {
+					chartWrapper: chartWrapperRef.current,
+					google: googleRef.current,
+				} );
+			} );
+		}
```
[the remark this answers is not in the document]
tofumatt: Unfortunately the `react-google-charts` component only supports a limited number of events with that API, so I had to implement these events this way: https://github.com/rakannimer/react-google-charts/blob/master/src/types.ts#L217-L225

But I've added a comment to clarify why we handle them this way. 👍🏻

assets/js/components/GoogleChartV2.js
```diff
@@ -0,0 +1,164 @@
...
+				className="googlesitekit-chart-v2-loading__wrapper"
+				height={ loadingHeightToUse }
+				shape={ loadingShape }
+				width={ loadingWidthToUse }
+			/>
+		</div>
+	);
+
+	const chartWrapperRef = useRef();
+	const googleRef = useRef();
```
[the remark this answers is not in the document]
tofumatt: See above: we still need these for now because we can't use the `chartEvents` API for the `mouse*` events.

assets/js/components/GoogleChartV2.js
```diff
@@ -0,0 +1,164 @@
...
+		onMouseOut,
+		onReady,
+		onSelect,
+		width,
+		...otherProps
+	} = props;
+
+	const loadingHeightToUse = loadingHeight || height;
+	const loadingWidthToUse = loadingWidth || width;
+	const loadingShape = getLoadingShape( chartType );
```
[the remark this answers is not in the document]
tofumatt: In the future other charts might require their own shapes… but actually I had a look through all the exist charts and I think every chart aside from `PieChart` would make sense to use `square` as a default shape.

I was thinking we keep it flexible/extensible, but this is easy to expand in the future, so fair enough! I've simplified it to use your suggestion—honestly, that's probably fine forever anyway 🙂

assets/js/components/GoogleChartV2.js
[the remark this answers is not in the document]
eugene-manuilov: @tofumatt those types are valid only for when you use ts, once it's transpiled into regular js those types do nothing. Better check the GoogleChartEvents component here: https://github.com/rakannimer/react-google-charts/blob/master/src/components/GoogleChartEvents.tsx - it doesn't filter passed events, so looks like it should work with onmouse events too.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
tofumatt: I was going to, but the varied sizes create a slightly more "natural" loading size (usually the four dimension names have different widths, so making them all the same width during loading feels less natural.)

A bonus is that for some locals these will line up perfectly with the actual loaded data, but even when they don't: the variance looks a bit nicer 🙂

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
tofumatt: Good catch, thanks!

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
```diff
@@ -56,125 +60,110 @@ export default function UserDimensionsPieChart( { dimensionName, dimensionValue,
...
+			const newDimensionValue = chartWrapperRef.current.getDataTable().getValue( index, 0 );
 			const isOthers = __( 'Others', 'google-site-kit' ) === newDimensionValue;
 
 			if ( isOthers ) {
 				return;
 			}
 
-			const { row } = chart.getSelection()?.[ 0 ] || {};
+			const { row } = chartWrapperRef.current.getChart().getSelection()?.[ 0 ] || {};
 			if ( row === index ) {
```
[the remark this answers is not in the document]
tofumatt: Oh, nice catch. Fewer ternary operators is always nice 👍🏻

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
```diff
@@ -56,125 +60,110 @@ export default function UserDimensionsPieChart( { dimensionName, dimensionValue,
...
-			}
-			const { dataTable } = GoogleChart.charts.get( chartID ) || {};
-			setSelectable( dataTable.getValue( row, 0 ) !== __( 'Others', 'google-site-kit' ) );
-		} );
 
-		chartData.onMouseOut = global.google.visualization.events.addListener( chart, 'onmouseout', () => {
-			setSelectable( false );
-		} );
-	}, [ chartID, dimensionName, setValues ] );
+	const chartWrapperRef = useRef();
```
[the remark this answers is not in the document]
tofumatt: `chartWrapper` is actually a `ref` that `react-google-charts` makes—it's a bit of its own custom API to get the chart variable, the data table, etc. It's not a `<div>`, and I don't think makes sense to store in `state` as we just want to keep a reference to it to use in the `onLegendClick` callback. Storing it in state will force another component render after finishing up as the ref is set/updated after the chart renders, which `useRef` should avoid.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
```diff
@@ -297,71 +286,97 @@ export default function UserDimensionsPieChart( { dimensionName, dimensionValue,
...
+				<GoogleChartV2
+					chartType="PieChart"
+					data={ dataMap || [] }
+					getChartWrapper={ ( chartWrapper ) => {
+						chartWrapperRef.current = chartWrapper;
+						// Forces a re-render of the component to re-run useEffect hooks.
+						// If this is not called and the chart is updated while the mouse cursor
+						// is already over the chart, it won't properly run the `onmousenter` event
+						// and the "selectable" slices of the pie chart won't be enabled consistently.
+						setSelectable( null );
```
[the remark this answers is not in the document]
tofumatt: Indeed! I don't love it, and I'm also not even 100% it always fixes the bug. If it doesn't, I'll remove it.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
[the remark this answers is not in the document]
eugene-manuilov: `chartWrapper` is an instance of the [google.visualization.ChartWrapper](https://developers.google.com/chart/interactive/docs/reference#chartwrapperobject) class. Here is how `react-google-charts` set it:

https://github.com/rakannimer/react-google-charts/blob/288ce66a2d8e89012305ee3bb5c200fed0a1c9ef/src/components/GoogleChart.tsx#L161-L165

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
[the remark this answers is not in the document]
tofumatt: Turns out it makes the "Others" section selectable instead so there's really no winning.

This is a big of a bug with loaded charts and mouseenter, so for now we'll leave it be.

tofumatt:
Awesome, thanks for the review. I've made all the suggested fixes or commented/explained where I thought a different approach was needed. Let me know if that looks good 🙂

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
```diff
@@ -56,125 +60,119 @@ export default function UserDimensionsPieChart( { dimensionName, dimensionValue,
...
+	const onLegendClick = ( index ) => {
+		if ( ! chartWrapperRef.current ) {
+			return;
+		}
+
+		const newDimensionValue = chartWrapperRef.current.getDataTable().getValue( index, 0 );
+		const isOthers = __( 'Others', 'google-site-kit' ) === newDimensionValue;
+
+		if ( isOthers ) {
+			return;
```
eugene-manuilov, line 80: @tofumatt we probably need to remove this `return` statement because it prevents dimension value reset when "others" is selected in the legend. Currently, if I select a slice, then click on the "others" slice - my selection is reset. But it doesn't happen if I click on the legend item.

assets/js/components/GoogleChartV2.js
```diff
@@ -0,0 +1,174 @@
...
+	);
+
+	const chartWrapperRef = useRef();
+	const googleRef = useRef();
+
+	useEffect( () => {
+		// Remove all event listeners after the component has unmounted.
+		return () => {
+			// eslint-disable-next-line no-unused-expressions
+			googleRef.current?.visualization.events.removeAllListeners( chartWrapperRef.current?.getChart() );
```
eugene-manuilov, line 76: @tofumatt we also need to remove all events assigned to the `chartWrapper` itself because this is how the library assigns `select` and `ready` events.
```suggestion
			if ( googleRef.current && chartWrapperRef.current ) {
				const { events } = googleRef.current.visualization;
				events.removeAllListeners( chartWrapperRef.current.getChart() );
				events.removeAllListeners( chartWrapperRef.current );
			}
```

Also, will we have the correct references to the google and chartWrapper instances here? Will useEffect memoize it and work with the original values? :thinking:

assets/js/components/GoogleChartV2.js
```diff
@@ -0,0 +1,174 @@
...
+				loader={ loader }
+				height={ height }
+				getChartWrapper={ ( chartWrapper, google ) => {
+					// Remove all the event listeners on the old chart before we draw
+					// a new one. Only run this if the old chart and the new chart aren't
+					// the same reference though, otherwise we'll remove existing `onReady`
+					// events and other event listeners, which will cause bugs.
+					if ( chartWrapper !== chartWrapperRef.current ) {
+						// eslint-disable-next-line no-unused-expressions
+						googleRef.current?.visualization.events.removeAllListeners( chartWrapperRef.current?.getChart() );
```
eugene-manuilov, line 153: Same here, we also need to remove events assigned to the `chartWrapper` instance.

eugene-manuilov (changes requested):
Thanks, @tofumatt! This is very close now. I have 3 more comments and I think it should be good to go.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
eugene-manuilov: @tofumatt ok, then how about this?

```javascript
[ 68, 52, 40, 52  ].map( ( width, i ) => (
    <span key={ i } className="googlesitekit-widget--analyticsAllTraffic__legend-slice">
        <span className="googlesitekit-widget--analyticsAllTraffic__dot" style={ { backgroundColor: '#ccc' } } />
        <PreviewBlock height="18px" width={ `${ width }px` } shape="square" />
    </span>
) )
```

assets/js/components/GoogleChartV2.js
tofumatt: These are the "previous" instances since we haven't yet re-assigned the `ref`. These refs should still be intact at this point, yeah.

eugene-manuilov (dismissed):
Thanks, @tofumatt!

felixarntz (changes requested):
This looks like an awesome enhancement overall! While testing, I noticed a few problems still:
* The "Others" pie slice still receives a visual outline when being selected/unselected, and it also modifies the dimension value currently used. This was particularly mentioned as requirement in the ACs. When you click the "Others" slice, basically nothing should change at all in the widget's UI.
* After changing the date range when no slice is selected currently, the cursor remains the regular arrow when hovering over a pie slice, instead of changing into the pointer as expected. As soon as you click a slice, it goes back to the expected behavior; just after changing the date range, it does not show as a pointer on hover.

@tofumatt Can you have a look at those things? I know particularly dealing with the "Others" slice was already tricky before, but I was hoping we might be able to finally address these problems here. If this is still not possible due to limited support by the charts library, I guess we might need to accept that, but I'd like to confirm first.

felixarntz:
@tofumatt With #2740 now merged, can you resolve the merge conflict / integrate the new tracking code into your new approach here?

tofumatt:
> After changing the date range when no slice is selected currently, the cursor remains the regular arrow when hovering over a pie slice, instead of changing into the pointer as expected. As soon as you click a slice, it goes back to the expected behavior; just after changing the date range, it does not show as a pointer on hover.

Unfortunately this is because of how the `mouseEnter` events are fired that enable the `--selectable` class. We need to know the position of the mouse pointer compared to a chart slice to determine if the slice is selectable, but there's no (straightforward) way to know which row/pie slice the user is hovering over when the date range changes. It does get fixed after you re-enter the chart.

(I think this already happens in the existing implementation as well. If not, I'd say it'd be best to file a new issue for it as fixing it might require storing the last know.)

We _could_ keep a record of:

1. The last know row/slice the user hovered over.
2. If their mouse has left the chart area.

And then check that when the date range updates. I can give that a go, but equally we could file a new follow-up issue for that if you like.

felixarntz (changes requested):
@tofumatt The behavior is still off here:
* When clicking the "Others" legend item, I get a JavaScript error (see screenshot below).
* When selecting a pie slice, and then clicking "Others" twice, it still results in the dimension value being changed to empty ("All Users").

Screen Shot 2021-02-24 at 1 12 59 PM

Regarding the minor bug where the cursor remains the regular arrow, sounds good dealing with that separately, since it seems more tricky and also isn't critical. But we should figure out the above, especially the JS error of course.

felixarntz:
@tofumatt I've addressed the problems here as discussed:
* https://github.com/google/site-kit-wp/pull/2805/commits/7e573f71ff816718efead54598a453e3692276fc ensures that the "Others" slice can no longer be selected and does not have any effect when clicked. It does that by overriding the selection to what was previously selected (either nothing or the previously selected slice). In order to do so, the active row index is saved in the `core/ui` store.
* https://github.com/google/site-kit-wp/pull/2805/commits/a8925ed4f3bbd63a84ab26b422067a6b295a78f3 fixes the legend click handler logic, which was inversed (only if "Others" is selected, it shouldn't do anything). It also adds the necessary calls to update the active row index (per the first bullet point).
* https://github.com/google/site-kit-wp/pull/2805/commits/66f36ac26c788002e4d465deb7bdcf79e056ee6a moves a bunch of code further up in the component, without making _any_ changes. I kept this in a separate commit to keep review easy. Moving this code was needed to allow for the following bullet point.
* https://github.com/google/site-kit-wp/pull/2805/commits/5b301f5be46e724815efb43c720a985826bc6e3b fixes the JS error that would happen when the dimension value for the selected slice was no longer available in the new data, e.g. after modifying the date range. The logic there was flawed in a few ways (not caused by this PR, but has been a problem for a while):
    * It was relying on the report to look up the matching slice, but the report includes some extra entries (1-2 depending on whether "Others" is used) which are not actually displayed as a slice. It could therefore happen that the lookup would be successful (e.g. finding the `ga:country` dimension value in the data as the sixth entry) when it shouldn't be (the real slice data to consider here is only five entries, or even only four, if "Others" is present). This is why the `onReady` callback needs access to the slice data, hence the logic to assemble that was moved up (see above).
    * When this lookup fails, there was previously no logic to react to that, which caused the line chart on the left to still show data for that dimension value, even though that dimension value does not have any pie slice and therefore shouldn't be displayed. Now, whenever there is no match in the new data, the dimension value (and related data) will be correctly reset.
* https://github.com/google/site-kit-wp/pull/2805/commits/f96f96c262847364fca6020fbd631a002604226e fixes an incorrect event name, please re-add the other events from #2740 here, using the same code introduced there (manual `useEffect` hook to listen to tooltip link clicks).

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
```diff
@@ -96,12 +123,6 @@ export default function UserDimensionsPieChart( {
 				[ UI_DIMENSION_VALUE ]: newDimensionValue,
 				[ UI_ACTIVE_ROW_INDEX ]: index,
 			} );
-
-			trackEvent(
-				'all_traffic_widget',
-				'slice_select',
-				`${ dimensionName }:${ newDimensionValue }`,
-			);
```
felixarntz: This shouldn't be removed, as it also selects the respective slice.

felixarntz (approved):
Awesome! Let's 🚢

Issue #2653, opened 2021-01-20
Modify `ModuleApp` to support rendering Widgets API widgets


felixarntz:
In order to migrate all module pages to use the Widgets API, we first need to modify the `ModuleApp` component so that it can render widgets (i.e. render a separate version of the page in addition to the current/legacy UI).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Modules should be able to register a widget context in JS that is used to render their admin screen via a new `screenWidgetContext` argument for the `core/modules` `registerModule` action.
* All modules that are configured to have their own admin screen in PHP (see `Module_With_Screen`) should also register a `screenWidgetContext`:
    * `moduleSearchConsole` for `search-console`
    * `moduleAnalytics` for `analytics`
    * `moduleAdsense` for `adsense`
    * These context names were already added via #2062, but they should no longer be defined as constants in the core `assets/js/googlesitekit/widgets` module (since they are not really parts of the core infrastructure). They should instead be defined in a `assets/js/modules/*/constants.js` file per module.
* All these modules should also register a single "main" widget area for their own context. This has already been implemented in #2062, however, similar as the contexts, they should no longer be defined in the core `assets/js/googlesitekit/widgets` module. Instead, the widget area slugs should be defined in `assets/js/modules/*/constants.js` per module, and the calls to register every module's widget area should also be moved from `assets/js/googlesitekit/widgets/register-defaults.js` into `assets/js/modules/*/index.js` per module.
* A new feature flag `widgets.moduleScreens` should be added, similar to the existing `widgets.*` flags.
* The module's screen component in JS should, if the above feature flag is enabled, render the widgets for their registered context (_if_ a `screenWidgetContext` for the module is registered) via `WidgetContextRenderer`. The widgets will be implemented later, but the screen should already include a duplicate version of the page's header and footer (see #2646 for related changes to the dashboard screen).

### Screen widget context registration

* Add `screenWidgetContext` argument to the `registerModule` action of `core/modules`.
* Add a `getScreenWidgetContext( moduleSlug )` selector to the `core/modules` store which returns the registered context string, or `null`.
* For Search Console, Analytics, and AdSense:
[...]

PR #2808, opened 2021-02-15
Modify `ModuleApp` to support rendering Widgets API widgets (#2653)

johnPhillips:
## Summary

Addresses issue #2653

## Checklist


assets/js/components/module/ModuleApp.js
```diff
@@ -0,0 +1,100 @@
...
+
+import DateRangeSelector from '../DateRangeSelector';
+import { CORE_MODULES } from '../../googlesitekit/modules/datastore/constants';
+
+const { useSelect } = Data;
+
+function ModuleApp( { moduleSlug } ) {
+	const screenWidgetContext = useSelect( ( select ) => select( CORE_MODULES ).getScreenWidgetContext( moduleSlug ) );
+	const { slug, name } = moduleSlug;
+	const serviceName = _x( name, 'Service name', 'google-site-kit' ); // eslint-disable-line @wordpress/i18n-no-variables
```
tofumatt, line 50: Why is this rule disabled? It seems the violation here is legitimate and this won't be able to be extracted by the i18n tools: https://github.com/WordPress/gutenberg/blob/master/packages/eslint-plugin/docs/rules/i18n-no-variables.md

assets/js/components/module/ModuleApp.js
```diff
@@ -0,0 +1,100 @@
...
+import ModuleFooter from './ModuleFooter';
+import LegacyModuleApp from './LegacyModuleApp';
+import WidgetContextRenderer from '../../googlesitekit/widgets/components/WidgetContextRenderer';
+
+import DateRangeSelector from '../DateRangeSelector';
+import { CORE_MODULES } from '../../googlesitekit/modules/datastore/constants';
+
+const { useSelect } = Data;
+
+function ModuleApp( { moduleSlug } ) {
```
tofumatt, line 47: This is a confusing prop name, because it isn't a module slug: it's an object with the module's `slug` and `name` attributes as properties.

Can we rename this to something more representative? Or we could have `slug` and `name` be props.

assets/js/components/module/ModuleApp.js
```diff
@@ -0,0 +1,100 @@
...
+		<Fragment>
+			<Header>
+				{ moduleConnected && <DateRangeSelector /> }
+			</Header>
+			<Alert module={ slug } />
+			{ shouldRenderWidget &&
+				<Fragment>
+					<WidgetContextRenderer
+						slug={ screenWidgetContext }
+						className={ `googlesitekit-module-page googlesitekit-module-page--${ slug }` }
```
tofumatt, line 72: This is minor, but anytime we have multiple classnames, especially when one is dynamic, I think it's a lot easier to read when we use `classnames`:

```suggestion
						className={ classnames(
							'googlesitekit-module-page',
							`googlesitekit-module-page--${ slug }`
						) }
```

assets/js/components/module/ModuleApp.js
```diff
@@ -0,0 +1,100 @@
...
+						Header={ () => (
```
tofumatt, line 73: This is technically fine but it's a bit unexpected to see this sort of render props pattern in the render section in our codebase. Can we define this as a constant above to improve readability? Especially as `ModuleFooter` is already defined that way. 🤷🏻‍♂️

assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -790,6 +793,31 @@ const baseSelectors = {
...
+	/**
+	 * Gets the module's screenWidgetContext.
+	 *
+	 * Returns `null` if there is no registered context string for the given module.
+	 * Returns `string` the registered context string, screenWidgetContext for the given module.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state      Data store's state.
+	 * @param {string} moduleSlug Module slug.
```
tofumatt, line 806: This isn't accurate as the current `moduleSlug` variable here is an object.

But I think this should remain and the module slug ONLY should be supplied to this function, not the object. The `name` attribute isn't used anywhere and makes this a bit more confusing to read.

Let's update this to take only the actual `slug` string value as an argument.

tofumatt (changes requested):
This looks good but I think we can simplify/clarify the `moduleSlug` object variable throughout the new code.

I'm also not clear on why we're ignoring the ESLint rule for `_x()`, as from what the rule implies using a variable as the first argument will cause the string not to be extracted—since all i18n extraction is done using static analysis.

assets/js/components/module/ModuleApp.js
johnPhillips: I agree, this could be clearer. The IB was particular about it:

> - Introduce a new ModuleApp functional hook-based component in assets/js/components/module:
>    - Require moduleSlug as prop.

Looks like that might be to keep parity with the other component that could be rendered in [assets/js/googlesitekit-module.js](https://github.com/google/site-kit-wp/blob/develop/assets/js/googlesitekit-module.js#L43)?

I like the idea of specific props, so let's go with `moduleName` and `moduleSlug`?

assets/js/components/module/ModuleApp.js
johnPhillips: Removed as discussed this morning - thanks 👍

assets/js/components/module/ModuleApp.js
tofumatt: We're already changing the API anyway so I don't think we need to keep the `moduleName` prop around, it's fine to update this and accept the slug only.

I might just use `slug` as the prop name given the component itself is called `ModuleSetup`, so `slug` implies a module `slug`.

But I think think we should remove the `moduleName`/`name` prop entirely, because the component can select the module name based on the supplied slug:

```js
const module = useSelect( ( select ) => select( CORE_MODULES ).getModule( slug ) );
const { name } = module;
```

I'm a fan of not requiring more data as a prop than needed for a component to work, so since we can already get a module's name from inside the component when the `slug` is known and we are already selecting other parts of the module's info from state ([like its `icon`](https://github.com/google/site-kit-wp/pull/2808/files#diff-7214c466b8f8c31951bc809134e9dcaad0ee7a61a96c860a838721f8ca891141R51)), I think it'd be better to get the `name` from a `useSelect` instead of as a prop.

tofumatt (changes requested):
Just one more thing to fix on this; I think we should avoid requiring the module's `name` as a prop when we can access it from `useSelect` instead.

assets/js/components/module/ModuleApp.js
johnPhillips: But the selector will return `undefined` at first, no?

tofumatt (approved):
Looks good, nice! 👍🏻

assets/js/components/module/ModuleApp.js
```diff
@@ -0,0 +1,105 @@
...
+import { CORE_MODULES } from '../../googlesitekit/modules/datastore/constants';
+
+const { useSelect } = Data;
+
+function ModuleApp( { slug } ) {
+	const { name } = useSelect( ( select ) => select( CORE_MODULES ).getModule( slug ) ) || {};
+	const screenWidgetContext = useSelect( ( select ) => select( CORE_MODULES ).getScreenWidgetContext( slug ) );
+	const moduleConnected = useSelect( ( select ) => select( CORE_MODULES ).isModuleConnected( slug ) );
+	const ModuleIcon = useSelect( ( select ) => select( CORE_MODULES ).getModuleIcon( slug ) );
+	const shouldRenderWidget = isFeatureEnabled( 'widgets.moduleScreens' ) && screenWidgetContext;
```
felixarntz, line 53: This would be better to use the `useFeature` hook instead.

assets/js/components/module/ModuleApp.js
```diff
@@ -0,0 +1,105 @@
...
+		/>
+	);
+
+	return (
+		<Fragment>
+			<Header>
+				{ moduleConnected && <DateRangeSelector /> }
+			</Header>
+			<Alert module={ slug } />
+			{ shouldRenderWidget &&
```
felixarntz, line 83: We should also include `screenWidgetContext` in the conditional here, since that is technically optional, even though all our modules that have their own dedicated screen relying on `ModuleApp` do so. Otherwise this can result in errors due to not passing a valid `slug` to `WidgetContextRenderer`.

assets/js/components/module/ModuleApp.js
```diff
@@ -0,0 +1,105 @@
...
+				<ModuleIcon
+					className="googlesitekit-page-header__icon"
+					height="21"
+					width="23"
+				/>
+			}
+			status={ moduleStatus }
+			statusText={ moduleStatusText }
+		/>
+	);
```
felixarntz, line 75: I think it would be a bit cleaner to also move this one into its own dedicated component file, similar to e.g. `DashboardHeader` and `DashboardDetailsHeader`. That component could also require the `moduleSlug` prop, and then retrieve the other details via hooks (basically moving some of those here to there). That would make the component reusable too.

assets/js/components/module/ModuleApp.js
```diff
@@ -0,0 +1,105 @@
...
+import ModuleFooter from './ModuleFooter';
+import LegacyModuleApp from './LegacyModuleApp';
+import WidgetContextRenderer from '../../googlesitekit/widgets/components/WidgetContextRenderer';
+
+import DateRangeSelector from '../DateRangeSelector';
+import { CORE_MODULES } from '../../googlesitekit/modules/datastore/constants';
+
+const { useSelect } = Data;
+
+function ModuleApp( { slug } ) {
```
felixarntz, line 48: This prop should be called `moduleSlug` to be a bit more verbose. We previously ran into issues where we needed multiple slugs in one component (e.g. `widgetSlug` and `moduleSlug`), and while it should be obvious here that it's a module's `slug`, being more verbose is a bit more future-proof and clear I think.

assets/js/components/module/ModuleApp.js
```diff
@@ -0,0 +1,105 @@
...
+					/>
+				</Fragment>
+			}
+			<LegacyModuleApp />
+		</Fragment>
+	);
+}
+
+ModuleApp.propTypes = {
+	slug: PropTypes.string,
```
felixarntz, line 102: See above (also in other references).

assets/js/modules/adsense/index.js
```diff
@@ -78,6 +78,7 @@ export const registerModule = ( modules ) => {
 					data: null,
 				};
 			},
+			screenWidgetContext: CONTEXT_MODULE_ADSENSE,
 		}
 	);
 };
```
felixarntz, line 84: This file is missing the logic to register the AdSense main widget area (it's only missing here; you've added it as needed in Analytics and Search Console).

felixarntz (changes requested):
Stellar work here, this looks very solid overall! There are a few points of feedback, some things to correct, but this is really close.

assets/js/components/module/ModuleApp.js
johnPhillips: I think I already am on line 53?
```js
const shouldRenderWidget = isFeatureEnabled( 'widgets.moduleScreens' ) && screenWidgetContext;
```
Or perhaps I have misunderstood.

assets/js/components/module/ModuleApp.js
johnPhillips: Haha, there was some back and forth on this already (see discussion above with @tofumatt). I've changed it to `moduleSlug`.

assets/js/modules/adsense/index.js
johnPhillips: Good catch!

johnPhillips:
Thanks for the review @felixarntz - I think I addressed everything you've raised 👍

assets/js/components/module/ModuleApp.js
felixarntz: @johnPhillips Oh my bad, I missed that in there. All good then!

felixarntz (approved):
Awesome work - all good now!

Issue #2509, opened 2020-12-09
Replace deprecated eslint-loader with eslint-webpack-plugin


aaemnnosttv:
## Bug Description

[The `eslint-loader` package is now deprecated](https://github.com/webpack-contrib/eslint-loader#warning-deprecated) and [`eslint-webpack-plugin`](https://github.com/webpack-contrib/eslint-webpack-plugin) is recommended in its place.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The functionality provided by `eslint-loader` should be replaced with `eslint-webpack-plugin`  
See https://github.com/webpack-contrib/eslint-webpack-plugin#migrate-from-eslint-loader

## Implementation Brief

* Install `eslint-webpack-plugin` as a development dependency;
* Delete `eslint-loader` dependency;
* Remove eslint-loader from the common rules array:
  https://github.com/google/site-kit-wp/blob/026f7d3e022356335c420d276fdd31366ea6342c/webpack.config.js#L117-L123
* Create a new `plugins` array similar to the `rules` one and add the `ESLintPlugin` plugin to it;
  * Pass options from the `eslint-loader` to the eslint plugin constructor.
* Destruct plugins array in the plugin section of each configuration as we do it with the rules array.

### Test Coverage

* Not needed

### Visual Regression Changes

* No visual changes are expected

## QA Brief

* Make sure the eslint plugin catches formatting errors when you build assets.

## Changelog entry

* N/A

aaemnnosttv:
IB ✅

Hazlitte:
QA ✅
I tested npm builds and watch and edited a few different files with linting errors.  All errors were caught.

PR #2813, opened 2021-02-16
Migrate to eslint-webpack-plugin and quiet webpack output.

tofumatt:
## Summary

Addresses issue #2509.

## Relevant technical choices

This wasn't part of the original IB, but the webpack noise is currently _extremely_ high, and it makes it difficult to see things like ESLint errors, etc. I added:

```js
config.stats = 'errors-warnings';
```

to all of the webpack configs so they're quieter. We can back that out of the change if you like, but I thought it was useful.

## Checklist


eugene-manuilov (changes requested):
@tofumatt don't we need to add the eslint plugin to the storybook webpack config?

tofumatt:
We weren't using `eslint-loader` in the old storybook config, and I don't think it's very useful since the tests/build would fail during a normal `npm run build` (or warn during `npm run watch`) anyway. I don't think we _also_ need to do it in Storybook as our `lint` task covers storybook files, so it mostly just adds extra time to storybook compiles.

But I removed the devserver bit as suggested 👍🏻

eugene-manuilov (dismissed):
Awesome!

aaemnnosttv (changes requested):
Thanks @tofumatt – this looks good but didn't work quite as expected in my testing.

Running a production build fails as expected with ESLint errors when present, however a development build (`dev` or `watch`) did not emit any error or warnings at all during the build or when rebuilding (during watch). I didn't test any ESLint warnings (I think we error on everything.. do we have anything that warns?). Currently ESLint errors do show in webpack output. A `dev` build will fail with ESLint errors (although I realize this PR changes that) but we should still see the errors. Similarly, a `watch` build won't fail but the errors are shown in the build output.

The less verbose output with webpack is nice though if we can keep that 👍

tofumatt:
> Running a production build fails as expected with ESLint errors when present, however a development build (dev or watch) did not emit any error or warnings at all during the build or when rebuilding (during watch). I didn't test any ESLint warnings (I think we error on everything.. do we have anything that warns?).

This is the issue I mentioned during a previous call that I wasn't sure was related to this change or not. I guess maybe it is. It seems the errors are only omitted during the first webpack build/compliation—subsequent builds don't warn/error 😞

I'll look into it further, but there aren't any obvious config settings to enable the desired behaviour…

tofumatt:
So I had a look and adding `failOnError` improves things, but it seems that still the errors sometimes don't show after saving. It's hard to pin down but seems to be when a file with errors is saved, then corrected, then saved again: the errors won't be output. I've no idea why that is, but adding `failOnError` seemed to improve output consistency.

That's all I've got though 😞

aaemnnosttv (approved):
Thanks @tofumatt – I found this worked well, showing the errors on rebuild when there was and it didn't break the build either. The more concise output is nice as well 👍

Issue #2588, opened 2021-01-06
Enhance localization of prepareSecondsForDisplay


aaemnnosttv:
## Feature Description

The `prepareSecondsForDisplay` utility takes a number in seconds and returns a formatted human-readable duration as the respective number of hours, minutes, and seconds. As it is, it currently uses hardcoded abbreviations for `s`, `h`, and `m` as well as building the resulting string by concatenating the pieces from left to right. This is not ideal for localization and should be improved to provide a more consistent experience for users across supported locales.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* `prepareSecondsForDisplay` should return properly localized numeric values, handled by the `numberFormat` utility  
_Note: we normally use `numFmt` for all formatting, but we're using `numberFormat` directly here since this will be called from within `numFmt` – there's no reason to support extra shorthand, etc_
* hours, minutes, and seconds should be formatted as `style: unit` with `hour`, `minute`, or `second` as the `unit`, respectively
* The formatted units of time should be concatenated with a translatable format string as there is no standard for formatting a number with multiple units
* The output should remain consistent with the current implementation for the default locale
* `numFmt` should respect any unit-related (e.g. `unitDisplay`) options, if provided when `style: duration` is explicitly passed
* `prepareSecondsForDisplay` should be renamed to better reflect its function of formatting the given `seconds` into a duration
* This function should no longer be used outside of `numFmt` and tests

## Implementation Brief

1. Refactor `prepareSecondsForDisplay` so that:
  * it is no longer exported
  * it has a new name: `durationFormat`
  * it accepts a number `seconds` as a parameter
  * it accepts a second `options` parameter, which is an object of type `Intl.NumberFormatOptions`
  * it parses the seconds into hours, minutes and seconds
  * it passes each into `numberFormat`, using appropriate `style` and `unit` options (e.g. hour would be `{ style: 'unit', unit: 'hour' }`) 
  * if the `options` parameter has been passed, then it should be passed through to `numberFormat` as well so that options like `unitDisplay` are reflected in the output.
  * it concatenates the formatted units using a translatable format string (e.g. `_x( '%$1s %$2s %$3s', 'duration of time', 'google-site-kit' )`). Remember to include a `translators: comment` above that to indicate that the first is hours, the second placeholder is for minutes, and the third is for seconds
[...]

PR #2818, opened 2021-02-17
Enhance localization of prepareSecondsForDisplay (#2588)

johnPhillips:
## Summary

Addresses issue #2588

## Checklist


assets/js/util/i18n.js
```diff
@@ -33,28 +33,51 @@ import { __, sprintf } from '@wordpress/i18n';
...
+ * @param {(Intl.NumberFormatOptions)} [options] Optional formatting options.
+ * @return {string}   		Human readable string indicating time elapsed.
  */
-export const prepareSecondsForDisplay = ( seconds ) => {
+const durationFormat = ( seconds, options = { unitDisplay: 'short' } ) => {
 	seconds = parseInt( seconds, 10 );
 
-	if ( isNaN( seconds ) || 0 === seconds ) {
-		return '0.0s';
+	if ( isNaN( seconds ) ) {
```
eugene-manuilov, line 46: We should use `Number.isNaN` isntead of `isNaN`:

```suggestion
	if ( Number.isNaN( seconds ) ) {
```

assets/js/util/i18n.js
```diff
@@ -33,28 +33,51 @@ import { __, sprintf } from '@wordpress/i18n';
...
+	}
+
+	seconds = numberFormat( seconds, {
+		style: 'unit',
+		unit: 'second',
+		...options,
+	} );
+
+	/* translators: 1: hours, 2: minutes, 3: seconds */
+	const formattedString = sprintf( _x( '%1$s %2$s %3$s', 'google-site-kit', 'google-site-kit' ), hours, minutes, seconds );
```
eugene-manuilov, line 78:
```suggestion
	const formattedString = sprintf( _x( '%1$s %2$s %3$s', 'duration of time: hh mm ss', 'google-site-kit' ), hours, minutes, seconds );
```

eugene-manuilov (changes requested):
Overall looks good to me. Just added two minor comments to address. Once fixed it should be good to go.

eugene-manuilov (approved):

assets/js/util/i18n.js
```diff
@@ -33,28 +33,51 @@ import { __, sprintf } from '@wordpress/i18n';
...
+	let hours = Math.floor( seconds / 60 / 60 ) || '';
+	let minutes = Math.floor( ( seconds / 60 ) % 60 ) || '';
+
+	seconds = Math.floor( seconds % 60 );
+
+	if ( hours ) {
+		hours = numberFormat( hours, {
+			style: 'unit',
+			unit: 'hour',
+			...options,
```
aaemnnosttv, line 59: Let's move where the spread happens since the `style` and `unit` shouldn't be overridable.

assets/js/util/i18n.js
```diff
@@ -33,28 +33,51 @@ import { __, sprintf } from '@wordpress/i18n';
...
+	}
+
+	seconds = numberFormat( seconds, {
+		style: 'unit',
+		unit: 'second',
+		...options,
+	} );
+
+	/* translators: 1: hours, 2: minutes, 3: seconds */
+	const formattedString = sprintf( _x( '%1$s %2$s %3$s', 'duration of time: hh mm ss', 'google-site-kit' ), hours, minutes, seconds );
```
aaemnnosttv, line 78: Let's split this onto a few lines as `sprintf` gets a bit hard to follow with the translated string and multiple parameters all on one line. This should do it (although it gets confused about the indentation of the translators comment sometimes 😄 )
```suggestion
	const formattedString = sprintf(
		/* translators: 1: hours, 2: minutes, 3: seconds */
		_x( '%1$s %2$s %3$s', 'duration of time: hh mm ss', 'google-site-kit' ),
		hours,
		minutes,
		seconds
	);
```

One additional thought here – perhaps we should document the placeholders a bit more descriptively since each one will also have its localized units included (if non-zero), so we should make it clear to avoid translators adding unit prefix/suffixes.

assets/js/util/i18n.js
```diff
@@ -33,28 +33,51 @@ import { __, sprintf } from '@wordpress/i18n';
...
  *
- * @param {number} seconds The number of seconds.
- * @return {string} Human readable string indicating time elapsed.
- *
+ * @param {number}                     seconds   The number of seconds.
+ * @param {(Intl.NumberFormatOptions)} [options] Optional formatting options.
+ * @return {string}   		Human readable string indicating time elapsed.
  */
-export const prepareSecondsForDisplay = ( seconds ) => {
+const durationFormat = ( seconds, options = { unitDisplay: 'short' } ) => {
```
aaemnnosttv, line 43: Generally we always apply defaults in the destructuring, not in the default value itself. That way the individual defaults aren't lost if `options` are passed.

```suggestion
const durationFormat = ( seconds, { unitDisplay = 'short' } = {} ) => {
```

In this case, since we want all of the options in the one object it makes more sense to handle the defaults in the function, then we don't need to worry about passing `options.style` as well.

```js
const durationFormat = ( seconds, options = {} ) => {
	const options = {
		unitDisplay: 'short',
		...options,
		style: 'unit',
	};

```

assets/js/util/i18n.js
```diff
@@ -180,10 +203,9 @@ export const numFmt = ( number, options = {} ) => {
 
 	if ( 'metric' === style ) {
 		return readableLargeNumber( number );
-	}
-
-	if ( 'duration' === style ) {
-		return prepareSecondsForDisplay( number );
+	} else if ( 'duration' === style ) {
+		formatOptions = omit( options, 'style' );
```
aaemnnosttv, line 207: Let's remove the `omit` here – `durationFormat` should simply ignore/override it if passed.

assets/js/util/i18n.js
aaemnnosttv, line 78: Looking at this again and thinking longer term, it seems reasonable that we might extend this to include larger units of time in the future (e.g. days, weeks...) at which point adding these placeholders would be a bit awkward.

Let's reverse the order so that `%1$s` refers to seconds, `%2$s` to minutes, and `%3$s` to hours so that the number of the placeholder increases with the unit of time.

aaemnnosttv (changes requested):
Thanks @johnPhillips – this looks great, just a few small things to address and then I think this should be good to go.

assets/js/util/i18n.js
johnPhillips: @aaemnnosttv Something like `duration of time (numeric): hh mm ss` perhaps, or does it need to be more verbose than that?

assets/js/util/i18n.js
johnPhillips: Good catch

assets/js/util/i18n.js
johnPhillips: Good idea

assets/js/util/i18n.js
```diff
@@ -33,28 +33,62 @@ import { __, sprintf } from '@wordpress/i18n';
...
+
+	let hours = Math.floor( seconds / 60 / 60 ) || '';
+	let minutes = Math.floor( ( seconds / 60 ) % 60 ) || '';
+
+	seconds = Math.floor( seconds % 60 );
+
+	if ( hours ) {
+		hours = numberFormat( hours, {
+			...options,
+			style: 'unit',
```
aaemnnosttv, line 64: We can remove the `style` in these calls since it's applied to all options now above.

assets/js/util/i18n.js
```diff
@@ -33,28 +33,62 @@ import { __, sprintf } from '@wordpress/i18n';
...
+		style: 'unit',
+		unit: 'second',
+	} );
+
+	const formattedString = sprintf(
+		/* translators: 1: seconds, 2: minutes, 3: hours */
+		_x( '%1$s %2$s %3$s', 'duration of time (numeric): hh mm ss', 'google-site-kit' ),
+		hours,
+		minutes,
+		seconds
```
aaemnnosttv, line 88: The resulting `formattedString` will be correct for the default locale here but `hours` should refer to `%3$s` where it is currently filling the `%1$s` placeholder. We just need to reverse the order of the placeholders in the string and the order they're passed to `sprintf` which should result in the same output but fix the relationship between placeholders and inputs. The translators comment is correct although it might be more clear to emphasize that each unit is formatted (rather than a bare number).

assets/js/util/test/prepareSecondsForDisplay.js
```diff
@@ -1,41 +0,0 @@
...
-		'1m 5s',
-	],
-	[
-		( 60 * 60 * 3 ) + ( 60 * 5 ) + 12,
-		'3h 5m 12s',
-	],
-	[
-		( 60 * 60 * 7 ) + ( 60 * 2 ) + 42,
-		'7h 2m 42s',
-	],
```
aaemnnosttv, line 34: Let's also include/merge these cases with the new tests. There are a few cases here which are important to keep such as `35`, `60` and `65` (although we don't need to include `65` twice like it is here 😄 ). It would also be good to ensure we're not breaking anything in the process.

assets/js/util/i18n.js
```diff
@@ -33,28 +33,62 @@ import { __, sprintf } from '@wordpress/i18n';
...
  * @since 1.0.0
+ * @since n.e.x.t Refactored and renamed to improve localization.
+ * @private
  *
- * @param {number} seconds The number of seconds.
- * @return {string} Human readable string indicating time elapsed.
- *
+ * @param {number}                     seconds   The number of seconds.
+ * @param {(Intl.NumberFormatOptions)} [options] Optional formatting options.
+ * @return {string}   		Human readable string indicating time elapsed.
```
aaemnnosttv, line 41: The return description doesn't require any special indentation, only parameters need to be aligned.

aaemnnosttv (commented):
Almost there!

assets/js/util/i18n.js
johnPhillips: Sorry, I think I misunderstood this last time round. Hopefully fixed now.

aaemnnosttv:
@googlebot I consent.

aaemnnosttv (approved):
Super, thanks!

Issue #924, opened 2019-11-27
Extend PHPCS rules to enforce all style rules


aaemnnosttv:
## Feature Description

In addition to following [WordPress core coding styles for PHP](https://github.com/WordPress/WordPress-Coding-Standards), there are some additional coding style rules which are applied that are not covered by the PHPCS configuration. 

This may be partially covered by #547 but some rules will likely still not be enforced.

E.g.
* The order of annotations in a docblock
* The required presence of specific annotations in a docblock
* The presence of blank lines before/after annotations in a docblock

We should extend the PHPCS configuration to include these additions as much as possible. 

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

There should be PHPCS rules to enforce the following:

### Semantic
* Every description of everything starts with a capital letter (descriptions, parameter docs, return docs, ...).
* Every description of everything ends in a dot (descriptions, parameter docs, return docs, ...).
* For functions and methods, the first word of the description must end in an "s" (this is probably 99% reliable, but most importantly should never yield a false positive - I don't think there are any exceptions to the grammatical rule of 3P verbs ending in an "s").

### Organizational
* Every property/function/method/class/interface/trait doc-block should have a `since` tag.
* The order of tags (if present, for each) should be `since` (multiple entries allowed), `deprecated`, `access`, `static`, `global`, `var`, `param` (multiple entries allowed), `return`.
* For methods and functions, tags should be grouped: `since`/`deprecated`/`access`/`static` should be grouped together, separated by a blank line from `param`/`return` which should be grouped together. If there is `global`, it should be separated from both others by a blank line (essentially be its own group).
* For properties, all tags should be in a single group (no blank lines between them).
* All `param` tag types and descriptions in a single doc-block should be indented in a way that they are aligned with each other.

Bonus points for opening this also against WordPress core, which probably could use all of it :-)

## Implementation Brief

* Create a new directory `includes/CodeStandards/GoogleSiteKit/` where our new coding sniffs will live.
* Create a new ruleset file: `includes/CodeStandards/GoogleSiteKit/ruleset.xml` to define our new ruleset:
```
<?xml version="1.0"?>
<ruleset name="GoogleSiteKit">
 <description>Google Site Kit's additional coding standards.</description>
</ruleset>
```
* To get composer to include our new coding standard rules whenever the vendor version of phpcs in run, add the following to the `phpcs.xml` file inside the `<ruleset>` tags:
```
[...]

PR #2826, opened 2021-02-18
Extend PHPCS rules

benbowler:
## Summary

Addresses issue #924

## Relevant technical choices

I went with `/CodingStandards/` not `/CodeStandards/` to keep it consistent with other similar standards.

Some elements of these rules were covered by other rules so I've disabled any duplicates in `includes/CodingStandards/GoogleSiteKit/ruleset.xml`.

## Checklist


benbowler:
I've completed the new sniffs. Can you review the implementation? Once that review is complete I'll run the new rules across and fix files across the whole project.

As one of the sniffs is to add a `@since` tag to every comment, how would you suggest I find the correct version for existing files? Look through the git commits for each?

includes/CodingStandards/GoogleSiteKit/Sniffs/Organizational/CommentHasSinceTagSniff.php
```diff
@@ -0,0 +1,71 @@
...
+				continue;
+			}
+		}
+
+		if ( ! $has_since_tag ) {
+			$error = 'Doc comment must include a @since tag';
+
+			$phpcs_file->addError( $error, $stack_ptr, 'MissingSinceTag' );
+		}
+	}
```
felixarntz, line 70: This should not check for `@since` tags on file header comments which it currently does; `@since` tags are only required for code entities, like classes, functions, interfaces, actions, etc.

includes/CodingStandards/GoogleSiteKit/Sniffs/Organizational/CommentTagsCorrectlyGroupedSniff.php
```diff
@@ -0,0 +1,216 @@
...
+			),
+			array(
+				'@global',
+			),
+			array(
+				'@var',
+				'@param',
+				'@return',
+			),
+		);
```
felixarntz, line 80: This does not fully implement the ACs; grouping should be slightly different between property docs and function/method docs; in particular: `@var` should only be relevant for property doc comments, and for those type of comments there should be just a single group with everything (most tags, e.g. `@param` and `@return`, are irrelevant for that kind of comment anyway).

includes/CodingStandards/GoogleSiteKit/Sniffs/Organizational/CommentTagsDescriptionsAlignedSniff.php
```diff
@@ -0,0 +1,247 @@
...
+	 * Processes this sniff, when one of its tokens is encountered.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param \PHP_CodeSniffer\Files\File $phpcs_file The current file being checked.
+	 * @param int                         $stack_ptr  The position of the current token in the
+	 *                                                stack passed in $tokens.
+	 * @return void
+	 */
+	public function process( File $phpcs_file, $stack_ptr ) {
```
felixarntz, line 47: This sniff should look for the minimum indentation needed, not just that they're aligned. For example, right now this test would (incorrectly) pass for the following:
```php
 * @param array $avar      A description.
 * @param array $bvar      Another description.
```

It should only pass for:
```php
 * @param array $avar A description.
 * @param array $bvar Another description.
```

includes/CodingStandards/GoogleSiteKit/Sniffs/Semantic/DescriptionEndsWithFullStopSniff.php
```diff
@@ -0,0 +1,204 @@
...
+	 * Processes this sniff, when one of its tokens is encountered.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param \PHP_CodeSniffer\Files\File $phpcs_file The current file being checked.
+	 * @param int                         $stack_ptr  The position of the current token in the
+	 *                                                stack passed in $tokens.
+	 * @return void
+	 */
+	public function process( File $phpcs_file, $stack_ptr ) {
```
felixarntz, line 47: This should also not apply to file comments - the descriptions on top of files are (I think) the only thing that should not require a full stop.

includes/CodingStandards/GoogleSiteKit/Sniffs/Semantic/DescriptionStartsWithThirdPersonVerbSniff.php
```diff
@@ -0,0 +1,91 @@
...
+	 * Processes this sniff, when one of its tokens is encountered.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param \PHP_CodeSniffer\Files\File $phpcs_file The current file being checked.
+	 * @param int                         $stack_ptr  The position of the current token in the
+	 *                                                stack passed in $tokens.
+	 * @return void
+	 */
+	public function process( File $phpcs_file, $stack_ptr ) {
```
felixarntz, line 47: There's one particular instance where this should not apply, which is for constructor descriptions (`__construct`). Those (should) usually have the shape of simply `Constructor.` or `MyClass constructor.`.

felixarntz (changes requested):
This is off to a great start! I tested this and found a few parts where the sniffs are not working as expected, so we'll need some iterations.

Great idea to first have the sniffs reviewed before (unnecessarily) fixing some of the false positives. Let's iterate on the sniffs further until they're good to go, and then fix the failures.

aaemnnosttv:
Closing this for now as we've decided to keep these outside of Site Kit.

Issue #2772, opened 2021-02-08
Link to Analytics FE for AdSense widgets based on Analytics data


felixarntz:
## Bug Description

Site Kit currently has deep links to the AdSense frontend on all AdSense widgets. While at first glance this seems to be expected, there are some AdSense widgets which are actually powered by Analytics. These widgets only show data when Analytics is also active and when the AdSense account is linked to the Analytics property.

These links should be updated to go to the relevant Analytics frontend area instead. The base URL for the relevant report is `https://analytics.google.com/analytics/web/#/report/content-publisher-overview`

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The AdSense top earning pages widget on the Site Kit dashboard as well as the top earning pages widget on the AdSense module page should deep link to the relevant Analytics report, using the `getServiceReportURL` selector from the `modules/analytics` store. For `type`, they should pass `content-publisher-overview`.
* The labels for all of these source links should be modified to refer to "Analytics" instead of "AdSense", where not already the case.
* For those components where the datastore is being used (at least the refactored version of the AdSense dashboard widget), the dates based on the current date range should be passed to the selector (see other usages of the Analytics `getServiceReportURL` selector for reference).
    * If this is trivial to achieve for the legacy components, it's worth doing - but we shouldn't make any significant changes to the code in these components.

## Implementation Brief

* Modify the `DashboardTopEarningPagesWidget` component in the AdSense module to use the `getServiceReportURL` selector instead of the `getServiceURL`;
  * Pass `content-publisher-overview` as the first argument into the `getServiceReportURL` selector;
  * Pass `{ '_u.date00': startDate, '_u.date01': endDate }` as the second argument into the selector (`startDate` and `endDate` should be formatted to `YYYYMMDD`).
* Modify the `AnalyticsAdSenseDashboardWidgetLayout` component in the Analytics module:
  * Grab the current range start and end dates using the following selector:
    ```javascript
    const dateRangeDates = useSelect( ( select ) => select( CORE_USER ).getDateRangeDates( {
        offsetDays: DATE_RANGE_OFFSET,
    } ) );
    ```
  * Use the `getServiceReportURL` selector from the `modules/analytics` datastore instead of the `getServiceAccountSiteURL` from the `modules/adsense` datastore:
    * Pass `content-publisher-overview` as the first argument into the `getServiceReportURL` selector;
[...]

PR #2831, opened 2021-02-19
Link to Analytics FE for AdSense widgets based on Analytics data (#2772)

johnPhillips:
## Summary

Addresses issue #2772

## Checklist


eugene-manuilov (dismissed):
LGTM

assets/js/modules/adsense/components/dashboard/DashboardTopEarningPagesWidget.js
```diff
@@ -66,7 +66,9 @@ function DashboardTopEarningPagesWidget( { Widget, WidgetReportZero, WidgetRepor
 
 		return {
 			isAdSenseLinked: select( MODULES_ANALYTICS ).getAdsenseLinked(),
-			analyticsMainURL: select( MODULES_ANALYTICS ).getServiceURL(),
+			analyticsMainURL: select( MODULES_ANALYTICS ).getServiceReportURL( 'content-publisher-overview', {
+				'_u.date00': startDate, '_u.date01': endDate,
```
aaemnnosttv, line 70: We can't pass raw dates like this because they need to be formatted as `YYYYMMDD` as noted in the IB.
We added the `generateDateRangeArgs` utilities to handle this for us, as well as abstracting the relationship between arg name and specific dates which is not obvious just by looking at it.

Let's update it to use `generateDateRangeArgs( { startDate, endDate } )` instead 👍 
Use care to make sure we're sourcing the utility from Analytics utils though since AdSense has its own version by the same name, which is different.

assets/js/modules/analytics/components/dashboard/AnalyticsAdSenseDashboardWidgetLayout.js
```diff
@@ -25,18 +25,21 @@ import { __, _n, sprintf } from '@wordpress/i18n';
...
-import { generateDateRangeArgs } from '../../../adsense/util/report-date-range-args';
 const { useSelect } = Data;
 
 const AnalyticsAdSenseDashboardWidgetLayout = ( { children } ) => {
-	const dateRangeDates = useSelect( ( select ) => select( CORE_USER ).getDateRangeDates( {
+	const { startDate, endDate } = useSelect( ( select ) => select( CORE_USER ).getDateRangeDates( {
 		offsetDays: DATE_RANGE_OFFSET,
 	} ) );
-	const accountSiteURL = useSelect( ( select ) => select( MODULES_ADSENSE ).getServiceReportURL( generateDateRangeArgs( dateRangeDates ) ) );
+	const accountSiteURL = useSelect( ( select ) => select( MODULES_ANALYTICS ).getServiceReportURL(
```
aaemnnosttv, line 39: "accountSiteURL" is more of an AdSense terminology (the store also has a `getServiceAccountSiteURL` selector). Let's call this `analyticsReportURL` instead since the selector is using Analytics rather than AdSense now.

assets/js/modules/analytics/components/dashboard/AnalyticsAdSenseDashboardWidgetLayout.js
```diff
@@ -25,18 +25,21 @@ import { __, _n, sprintf } from '@wordpress/i18n';
...
+		'content-publisher-overview',
+		{ '_u.date00': startDate, '_u.date01': endDate },
```
aaemnnosttv, line 41: Same here; let's continue using `generateDateRangeArgs` as we did before. Again, this should use the Analytics version of the utility.

aaemnnosttv (changes requested):
Thanks @johnPhillips – this is looking good, just a few quick tweaks to make for consistency with other code that uses date range args.

I realize the IB doesn't explicitly call for using `generateDateRangeArgs`, but that doesn't mean it shouldn't be used. When in doubt, always opt for maintaining consistency 👍

assets/js/modules/analytics/components/dashboard/AnalyticsAdSenseDashboardWidgetLayout.js
```diff
@@ -25,18 +25,22 @@ import { __, _n, sprintf } from '@wordpress/i18n';
  * Internal dependencies
  */
 import Data from 'googlesitekit-data';
-import { MODULES_ADSENSE, DATE_RANGE_OFFSET } from '../../../adsense/datastore/constants';
+import { DATE_RANGE_OFFSET } from '../../../adsense/datastore/constants';
+import { MODULES_ANALYTICS } from '../../../analytics/datastore/constants';
```
aaemnnosttv, line 29: The date range offset values are currently all the same, but given that this links to Analytics I'm thinking we should probably be using the date range offset from Analytics as well. Let's update this and then we should be good to go here.
```suggestion
import { MODULES_ANALYTICS, DATE_RANGE_OFFSET } from '../../../analytics/datastore/constants';
```

aaemnnosttv (approved):
Super!

assets/js/modules/analytics/components/dashboard/AnalyticsAdSenseDashboardWidgetLayout.js
johnPhillips: Ah yes! Good catch 👍

johnPhillips:
Thanks for that @aaemnnosttv 👍

felixarntz (approved):
Awesome!

Issue #2660, opened 2021-01-20
Sitekit Dashboard Locations Label Placement


ivankruchkoff:
## Bug Description

Site Kit Dashboard label for Locations overlaps with the text 'BY' when 100% of traffic comes from one location.

## Steps to reproduce

1. Using latest develop branch as I'm pretty sure the feature is behind a feature flag.
2. Open Site Kit dashboard with Analytics module connected
2. Scroll down to Users widget in Site Overview.
3. Open Locations Tab.
4. Ensure 100% of Locations are attributed to one location, e.g. United States
5. See error

## Screenshots

<img width="1368" alt="Screen Shot 2021-01-19 at 10 05 23 PM" src="https://user-images.githubusercontent.com/373931/105134139-c9191c80-5aa2-11eb-8b25-6896a0acdec1.png">

## Additional Context

 - PHP Version: NA
 - OS: OSX
 - Browser chrome Version 87.0.4280.141 (Official Build) (x86_64)
 - Plugin Version develop branch (7ee258d11 Head commit)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* When the All Traffic widget pie chart only includes a single dimension (i.e. 100% of users come from that dimension), the percentage should still be displayed within the pie slice, it shouldn't overlap with the overall pie chart label.
    * This might be hard to reproduce consistently, please check in different browsers.

## Implementation Brief

* Inside [UserDimensionsPieChart.js](https://github.com/google/site-kit-wp/blob/develop/assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js):
  * Perform a check to determine if the `report` prop for `UserDimensionsPieChart` only has 1 row:
  ```js
    report?.[0]?.data?.rows?.length === 1 
  ```
* If there is only one row, we need to add `pieSliceText = 'none'` to the `UserDimensionsPieChart.chartOptions`. This will hide the '100%' text that appears in the centre of the pie. 
* if there is more than one row, we do nothing, i.e. current behaviour
* _Note: You could do this around line [288](https://github.com/google/site-kit-wp/blob/develop/assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js#L288) where a similar check is also performed_

### Test Coverage

* All existing tests should pass; no new tests required

### Visual Regression Changes

[...]

PR #2833, opened 2021-02-19
Fix all traffic widget label when there is only a single row of data (#2660)

johnPhillips:
## Summary

Addresses issue #2660

## Checklist


eugene-manuilov (approved):
:+1:

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
```diff
@@ -292,11 +292,17 @@ export default function UserDimensionsPieChart( { dimensionName, dimensionValue,
...
+
 	if ( report?.[ 0 ]?.data?.rows?.length < 2 ) {
 		// Hide pie slice text when there is just one slice because it will overlap with the chart title.
 		options.pieSliceTextStyle.color = 'transparent';
 	}
 
+	if ( report?.[0]?.data?.rows?.length === 1 ) {
+		// When there is only one row, the chart will add a label which we need to hide - see issue #2660
+		options.pieSliceText = 'none';
+	}
```
aaemnnosttv, line 304: @johnPhillips does this condition need to be separate than the one above? They both seem to cater for the single slice case and do different things. Is there any reason to change the `pieSliceTextStyle.color` in this case if we're changing the `pieSliceText` to `none` ?

johnPhillips: I spoke to @eugene-manuilov who wrote the other condition here and it seems they are both attacking the same problem. I've combined them.

aaemnnosttv (commented):
Code looks good! Sending back to add a story for this state per our convo.

aaemnnosttv (commented):
Thanks @johnPhillips – the added story is really helpful.

The only thing I see now is the alignment of the text looks different than the image in the QA Brief. The component in Storybook shows the text somewhat below center. Did this change with the new V2 chart component maybe? (cc: @tofumatt )

**QA Brief**
original

**Storybook**
image

johnPhillips:
@aaemnnosttv  Looks like as part of the GoogleChartV2 work we have added the `googlesitekit-cta-link googlesitekit-widget--analyticsAllTraffic__legend-slice` at the bottom. It has a negative top margin, which changes the height of the parent, meaning the absolutely positioned title appears off-center. I added a fix 👍

assets/sass/widgets/_googlesitekit-widget-analyticsAllTraffic.scss
```diff
@@ -167,7 +167,10 @@
 		position: absolute;
 		text-align: center;
 		text-transform: uppercase;
-		top: 50%;
+		// The -6px margin-top on the legend affects the parent height
+		// so we have to factor it in here to center the title
+		top: calc(50% - 6px);
```
aaemnnosttv, line 172: I'm not sure this is the right solution here. Looking at the component, the title seems to be intended to be a child of the chart component and the legend below:
https://github.com/google/site-kit-wp/blob/0075f6de4035b67d5830ada399bbc8cdca9c5a37/assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js#L387-L410
However, due to an implementation detail of the chart being output as a fragment with the chart followed by the children, this results in some unexpected output where the title and legend are siblings of each other which causes the legend to influence the height of the pie chart when it really shouldn't.

I would suggest we update `GoogleChartV2` to provide a wrapping `div` instead so that the title is always relative to the area of the chart.

<details>
<summary>This worked well in my testing:</summary>

Kapture 2021-03-03 at 13 00 25

</details>
@tofumatt any reason we can't do something like this?

<details>
<summary>git diff</summary>

```diff
diff --git a/assets/js/components/GoogleChartV2.js b/assets/js/components/GoogleChartV2.js
index cfef6c26d..4e2fcc4fc 100644
--- a/assets/js/components/GoogleChartV2.js
+++ b/assets/js/components/GoogleChartV2.js
@@ -136,13 +136,19 @@ export default function GoogleChartV2( props ) {
 	}
 
 	return (
-		<Fragment>
+		<div
+			className={ classnames(
+				'googlesitekit-chart-v2',
+				`googlesitekit-chart-v2--${ chartType }`,
+				className
+			) }
+		>
 			<Chart
 				chartEvents={ combinedChartEvents }
 				chartType={ chartType }
 				className={ classnames(
-					'googlesitekit-chart-v2',
-					`googlesitekit-chart-v2--${ chartType }`,
+					'googlesitekit-chart-v2__chart',
+					`googlesitekit-chart-v2__chart--${ chartType }`,
 					className
 				) }
 				loader={ loader }
@@ -170,7 +176,7 @@ export default function GoogleChartV2( props ) {
 				{ ...otherProps }
 			/>
 			{ children }
-		</Fragment>
+		</div>
 	);
 }
 
diff --git a/assets/sass/components/global/_googlesitekit-chart.scss b/assets/sass/components/global/_googlesitekit-chart.scss
index 03294e911..7d083cf0b 100644
--- a/assets/sass/components/global/_googlesitekit-chart.scss
+++ b/assets/sass/components/global/_googlesitekit-chart.scss
@@ -24,6 +24,10 @@
 	}
 }
 
+.googlesitekit-chart-v2 {
+	position: relative;
+}
+
 .googlesitekit-chart-v2--PieChart svg > g > g {
 	pointer-events: none;
 }
```
</details>

aaemnnosttv (changes requested):
Thanks @johnPhillips – the CSS change works but I don't think it's actually centered due to the height of the legend also affecting the total height. The good news is I don't think we need to change the CSS to fix it. See my comments below.

cc: @tofumatt

assets/sass/widgets/_googlesitekit-widget-analyticsAllTraffic.scss
tofumatt: No reason at all that shouldn't work, though we'll want to adjust a single CSS rule we have that has Hyper-Selector-Specificity™: https://github.com/google/site-kit-wp/blob/0075f6de4035b67d5830ada399bbc8cdca9c5a37/assets/sass/widgets/_googlesitekit-widget-analyticsAllTraffic.scss#L144-L150

I think it's a matter of adding another `> div` to make up for what `Chart` renders. We have to be really specific there to override and target specific SVG styles, if I recall correctly. (Alternatively, add a `googlesitekit-chart-v2__inner` to `Chart`'s `className` or something and use that.)

(Also: I'd strip out the `className` from the `Chart` component entirely there rather than duplicate it in the `div` and `Chart` component, but I think that might've been done by mistake 😅)

TL;DR is yes replacing the `Fragment` with a `<div>` should be fine 👍🏻

johnPhillips: Thanks @tofumatt 
Right, let me give this a whirl...

.storybook/storybook-data.js
```diff
@@ -65,7 +65,7 @@ module.exports = [
 			options: {
 				hierarchyRootSeparator: '|',
 				hierarchySeparator: {},
-				readySelector: '.googlesitekit-chart-v2 > div[style="position: relative;"]',
+				readySelector: '.googlesitekit-chart-v2',
```
aaemnnosttv, line 68: I think the intention here is to wait for the _inner_ selector – the one generated by a rendered chart, whereas `googlesitekit-chart-v2` will be present on all renderings of the chart, (loading, etc) so this may cause problems with the reliability of the test.

aaemnnosttv (commented):
Thanks @johnPhillips – the title looks great now and is no longer affected by the legend 👍 

I left one comment about the `readySelector` change but also VRT is failing (maybe related?)

.storybook/storybook-data.js
johnPhillips: Ah yes you are right, my bad. I've addressed this (hopefully!)

aaemnnosttv (approved):
Super, thanks @johnPhillips !

Issue #2529, opened 2020-12-14
Redirect authenticated users from splash to dashboard


felixarntz:
## Bug Description

The `googlesitekit-splash` screen currently has an incorrect `<title>` tag where the main segment is missing. At the moment it looks like the following:

```
 < example.com — WordPress
```

However, there's currently no way for a user to reach the splash screen after they're already authenticated – even the activation notice shown after activating the plugin will link to the dashboard. This makes this situation only possible if the user navigated directly to the splash screen (from browser history/back, etc). The splash screen is only a stepping stone to send the user through Google OAuth; if they're already authenticated, there's no reason to go through it again so they should be sent straight to the dashboard.

If required scopes have changed, the user will still need to go through oAuth of course, but we already prompt the user to "Redo Setup" in this case.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `googlesitekit-splash` screen should redirect to `googlesitekit-dashboard` if the user is already authenticated (`Authentication::is_authenticated`)
* The connect and disconnect URLs should be updated to no longer use the base splash screen URL, but instead use their own admin `action` URLs, like we do for Google Proxy setup/settings URLs

## Implementation Brief

* Update screen definition of `googlesitekit-splash` to redirect to the dashboard if authenticated (preserving existing logic to ensure permissions first)
* Move connect/disconnect handlers from `googlesitekit-splash` URL to dedicated `admin_action_` URLs:
	- `index.php?action=googlesitekit_connect&nonce=...`
	- `index.php?action=googlesitekit_disconnect&nonce=...`
* Finish and merge https://github.com/google/site-kit-wp/pull/3037
	- Update references to these URLs in JS/fixtures
	- Updating tests

### Test Coverage

* Add/update tests for decoupled action handlers
* Update JS tests as necessary

### Visual Regression Changes

* None.

## QA Brief

This issue primarily updates the splash screen to redirect the user to the SK dashboard if they're already authenticated and have sufficient permissions.

The following cases should be tested to work (and continue to work) properly

**Unauthenticated / Site Kit not set up yet**
- Splash screen loads without redirecting
- Navigating directly to SK dashboard redirects to splash screen

**Authenticated / Site Kit set up**
- Navigating directly to splash screen redirects to SK dashboard
- SK dashboard does not redirect to splash

**Disconnect via user menu**
- User lands on splash screen with same disconnect messaging as today
- Session storage should be cleared

[...]

PR #2834, opened 2021-02-21
Enhancement/2468 svg replacing images

ivankruchkoff:
## Summary

Addresses issue #2468 

## Checklist


assets/js/components/legacy-notifications/dashboard-modules-alerts.js
```diff
@@ -74,7 +74,7 @@ class DashboardModulesAlerts extends Component {
 						title={ notification.title || '' }
 						description={ notification.description || '' }
 						blockData={ notification.blockData || [] }
-						winImage={ notification.winImage ? `${ global._googlesitekitLegacyData.admin.assetsRoot }images/${ notification.winImage }` : '' }
+						WinImageSVG={ notification.WinImageSVG || null }
```
tofumatt, line 77: If `WinImageSVG` is false-y then it won't render anything, so it's safe to omit the `|| null` here as the `Notification` component will handle false-y `WinImageSVG` prop values for us 🙂 
```suggestion
						WinImageSVG={ notification.WinImageSVG }
```

includes/Core/Admin/Screen.php
```diff
@@ -129,9 +130,53 @@ public function add( Context $context ) {
...
+					 * A user can pick a custom Admin Color Scheme, which is only available in admin_init
+					 * or later actions. add_menu_page runs in admin_menu actino, which precedes admin_init
+					 * per https://codex.wordpress.org/Plugin_API/Action_Reference
+					 *
+					 * WordPress provides some color schemes out of the box, but they can also be added via
+					 * wp_admin_css_color()
+					 *
+					 * Our workaround is to set the icon and subsequently replace it in admin_init, which is
+					 * what we do in the following action.
+					 */
```
tofumatt, line 155: Thanks for the detailed comment, very helpful! 👍🏻

tofumatt (approved):
This looks good, just one little tweak from me but it's minor enough I'll just commit it before merge review rather than do another round of code review 🙂

tofumatt:
@googlebot I consent.

assets/svg/rocket.svg
```diff
@@ -0,0 +1,219 @@
+<svg width="148" height="123" viewBox="0 0 148 123" fill="none" xmlns="http://www.w3.org/2000/svg">
```
aaemnnosttv, line 1: Looking at the change in VRT, I think we need to update the dimensions to match the image to make this a transparent replacement. If that doesn't work, then we'll need to specify the dimensions in the `RocketSVG` attributes so that there is no visual change.

includes/Core/Util/Google_Icon.php
```diff
@@ -0,0 +1,49 @@
...
+	/**
+	 * Returns SVG XML with fill color replaced.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $color Any valid color for css, either word or hex code.
+	 *
+	 * @return string SVG XML with the fill color replaced
+	 */
+	public static function replace_fill( $color ) {
```
aaemnnosttv, line 45: The replacement is a bit of an implementation detail. Let's name this a little more generically.
```suggestion
	public static function with_fill( $color ) {
```

includes/Core/Util/Google_Icon.php
```diff
@@ -0,0 +1,49 @@
...
+ * @license   https://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
+ * @link      https://sitekit.withgoogle.com
+ */
+
+namespace Google\Site_Kit\Core\Util;
+
+/**
+ * Class for the Google SVG Icon
+ *
+ * @since n.e.x.t
```
aaemnnosttv, line 16: There are a few more annotations that we add to all classes which we should include here as well:
```suggestion
 * @since n.e.x.t
 * @access private
 * @ignore
```

assets/js/components/legacy-notifications/dashboard-modules-alerts.js
```diff
@@ -74,7 +74,7 @@ class DashboardModulesAlerts extends Component {
 						title={ notification.title || '' }
 						description={ notification.description || '' }
 						blockData={ notification.blockData || [] }
-						winImage={ notification.winImage ? `${ global._googlesitekitLegacyData.admin.assetsRoot }images/${ notification.winImage }` : '' }
+						WinImageSVG={ notification.WinImageSVG }
```
aaemnnosttv, line 77: This won't work because `notification` comes from an API response here. I think this would be the same as the change to `Alert` where it would get `SmallSunSVG`.

includes/Core/Admin/Screen.php
```diff
@@ -129,9 +130,53 @@ public function add( Context $context ) {
...
+					add_action(
+						'admin_init',
+						function() {
+
```
aaemnnosttv, line 159:
```suggestion
```

aaemnnosttv (changes requested):
Thanks @ivankruchkoff, this is looking good, just a few things to address. Most importantly is that there shouldn't be any substantial visual changes. Our VRT threshold for failure is zero so it's okay if there are microscopic changes in the image due to differences between img/svg but they should be visually equivalent.

includes/Core/Util/Google_Icon.php
```diff
@@ -0,0 +1,49 @@
...
+
+	const XML = '<svg width="147" height="150" viewBox="0 0 147 150" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="M30.5611 74.7909C30.5611 70.0315 31.2403 65.272 32.9381 60.8526L7.81006 41.4749C2.71654 51.6736 0 62.8923 0 74.7909C0 86.6895 2.71654 97.9082 7.81006 107.767L32.9381 88.3893C31.2403 84.3098 30.5611 79.5503 30.5611 74.7909Z" fill="white"/><path d="M74.7029 30.5964C85.2295 30.5964 94.7374 34.3359 102.208 40.4552L123.94 18.6978C110.697 7.13915 93.7187 0 74.7029 0C45.1605 0 19.6928 16.998 7.80796 41.815L32.936 61.1928C38.7087 43.1749 55.0079 30.5964 74.7029 30.5964Z" fill="white"/><path d="M74.7029 118.986C55.0079 118.986 38.7087 106.408 32.936 88.7297L7.80796 108.107C19.6928 132.585 45.1605 149.583 74.7029 149.583C93.0395 149.583 110.358 143.123 123.261 130.885L99.4913 112.527C92.7 116.606 84.2108 118.986 74.7029 118.986Z" fill="white"/><path d="M32.936 88.7297L7.80796 108.107L32.936 88.7297Z" fill="#34A853"/><path d="M74.7036 61.1929V90.0894H114.773C112.735 99.9483 107.302 107.427 99.492 112.527L123.262 130.885C136.844 117.966 146.013 99.2684 146.013 74.7913C146.013 70.3718 145.334 65.6123 144.315 61.1929H74.7036Z" fill="white"/></svg>';
```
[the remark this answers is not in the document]
ivankruchkoff: This isn't the svg from the ticket, all of these images are from Pratheep.

assets/svg/rocket.svg
ivankruchkoff: I played around with image size, and it's not the size that's an issue, it's positioning relative to image boundary, looking to get that cleared up which should resolve the discrepancies.
Screen Shot 2021-02-25 at 5 22 48 PM

ivankruchkoff: Updated both the image to have the same positioning and size relative to image boundaries, and also added a css rule to control the svg size.

assets/js/components/legacy-notifications/dashboard-modules-alerts.js
ivankruchkoff: Updated in https://github.com/google/site-kit-wp/pull/2834/commits/2502ef0c930661df6ff9359f7342328d4124e6be

includes/Core/Util/Google_Icon.php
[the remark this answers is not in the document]
ivankruchkoff: Took the svg from the ticket, added the dimensions to it and updated it here. Also added a comment about the fill attribute.

includes/Core/Admin/Screen.php
```diff
@@ -129,9 +130,57 @@ public function add( Context $context ) {
...
+						'admin_init',
+						function() {
+							global $menu, $_wp_admin_css_colors, $pagenow;
+							$color_scheme = get_user_option( 'admin_color' ) ?: 'fresh';
+
+							$prefix = 'googlesitekit-';
+
+							// If we're on one of the sitekit pages, use the 'current' color, otherwise use the 'base' color.
+							// @see wp_admin_css_color().
+							$color_key = 'admin.php' === $pagenow && substr( filter_input( INPUT_GET, 'page', FILTER_SANITIZE_STRING ), 0, strlen( $prefix ) ) === $prefix ? 'current' : 'base';
```
ivankruchkoff, line 166: @aaemnnosttv a new addition here as the following two URLs will have different fill colours:

`/wp-admin/admin.php?page=googlesitekit-dashboard` and `/wp-admin/index.php`

aaemnnosttv: Thanks, I noticed that in my testing too. Can we base this on `get_current_screen` instead of a query param? Assuming that the timing isn't an issue, we should simply be able to compare the Screen's `parent_file` or `parent_base` 
image

assets/sass/components/global/_googlesitekit-publisher-wins.scss
```diff
@@ -84,6 +84,10 @@
 			display: inline-block;
 			max-width: 100%;
 		}
+
+		svg {
+			width: 200px;
+		}
```
aaemnnosttv, line 90: Looks like we might need to add some responsive styles here or make it relative so that it naturally fills the grid cell width? I'm guessing the previous images would have scaled as well.

aaemnnosttv (commented):
@ivankruchkoff this is looking really close, just a few tweaks left I think.

One comment about the added dynamic screen-based fill and added css rule. I still have [a question here](https://github.com/google/site-kit-wp/pull/2834/files#r582775160) about the big change in lock file size as well.

I noticed the VRT images look nearly identical now for small and large sizes, however there was a substantial difference still with the medium size:

**Onion skin**
image
**Side-by-side**
image

assets/sass/components/global/_googlesitekit-publisher-wins.scss
ivankruchkoff: This is purely for the rocket image which was 200px x 167px, and never scaled, see the comparison images. Further confirmation is that the addition of the rule brings no additional VRT test failures, so if this rule has an unintended side effect, it also means we are missing some storybook tests.

SVG1
https://github.com/google/site-kit-wp/blob/7dc1eb3947860fb77a9e98b2fcccdebca1248c91/tests/backstop/reference/google-site-kit_Global_Notifications_Module_Setup_Complete_0_document_0_small.png

PNG1
https://github.com/google/site-kit-wp/blob/develop/tests/backstop/reference/google-site-kit_Global_Notifications_Module_Setup_Complete_0_document_0_small.png

SVG2
https://github.com/google/site-kit-wp/blob/7dc1eb3947860fb77a9e98b2fcccdebca1248c91/tests/backstop/reference/google-site-kit_Global_Notifications_Module_Setup_Complete_0_document_1_medium.png

PNG2
https://github.com/google/site-kit-wp/blob/develop/tests/backstop/reference/google-site-kit_Global_Notifications_Module_Setup_Complete_0_document_1_medium.png

SVG3
https://github.com/google/site-kit-wp/blob/7dc1eb3947860fb77a9e98b2fcccdebca1248c91/tests/backstop/reference/google-site-kit_Global_Notifications_Module_Setup_Complete_0_document_2_large.png

PNG3
https://github.com/google/site-kit-wp/blob/develop/tests/backstop/reference/google-site-kit_Global_Notifications_Module_Setup_Complete_0_document_2_large.png

package.json
```diff
@@ -197,10 +193,6 @@
     "fsevents": "^2.2.1",
     "git-repo-info": "^2.1.1",
     "husky": "^3.1.0",
-    "imagemin-cli": "^6.0.0",
-    "imagemin-jpegtran": "^7.0.0",
-    "imagemin-optipng": "^8.0.0",
-    "imagemin-webpack": "^5.1.1",
```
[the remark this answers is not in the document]
ivankruchkoff: We're on different npm versions:

https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json#lockfileversion

```
Note that the file format changed significantly in npm v7 to track information that would have otherwise required looking in node_modules or the npm registry. Lockfiles generated by npm v7 will contain lockfileVersion: 2.

2: The lockfile version used by npm v7, which is backwards compatible to v1 lockfiles.
```

```
$ node -v
v14.13.1
[<_html/wp-content/plugins/google-site-kit]$ npm -v
7.5.6
[<_html/wp-content/plugins/google-site-kit]$ cat .nvmrc
14
```

I went and pulled the old package-lock.json and removed the imagemin packages again and pushed it up, that gets us to `2,716 changes: 133 additions & 2,583 deletions`

includes/Core/Admin/Screen.php
ivankruchkoff: Tried that, however:

/wp-admin/admin.php?page=googlesitekit-settings
```
["base"]=>
string(36) "site-kit_page_googlesitekit-settings"
["parent_base"]=>
NULL
["parent_file"]=>
NULL
```

/wp-admin/admin.php?page=googlesitekit-dashboard
```
["base"]=>
string(37) "toplevel_page_googlesitekit-dashboard"
["parent_base"]=>
NULL
["parent_file"]=>
NULL
```

/wp-admin/admin.php?page=googlesitekit-module-analytics
```
["base"]=>
string(44) "site-kit_page_googlesitekit-module-analytics"
["parent_base"]=>
NULL
["parent_file"]=>
NULL
```

If there's an aversion to using `$_GET['page']`, we can use get_current_screen and see if the `base` or `id` params contain `googlesitekit`. It does mean we hook `current_screen` instead of `admin_init` but that doesn't break any functionality.

assets/sass/components/global/_googlesitekit-publisher-wins.scss
ivankruchkoff: @aaemnnosttv updated the rule in https://github.com/google/site-kit-wp/pull/2834/commits/6f76bed4afb4e3870f9b2ad0fd54e74335c231b4

Updated the image in https://github.com/google/site-kit-wp/pull/2834/commits/cbeb8a16b4f4ebc10f2b88ebd09eef07fa072bef

includes/Core/Admin/Screen.php
aaemnnosttv: @ivankruchkoff Ah yes, I see the current screen isn't set yet when `admin_init` fires.

> If there's an aversion to using `$_GET['page']`, we can use get_current_screen and see if the `base` or `id` params contain `googlesitekit`. It does mean we hook `current_screen` instead of `admin_init` but that doesn't break any functionality.

We should prefer to use the screen where possible over query parameters directly. From what I can see, `get_current_screen()->parent_base === 'googlesitekit-dashboard'` will be true on all of our screens with one exception (which is a bit of an edge case). When visiting the splash screen when already authenticated (we don't actually do this) the splash screen is registered as an "orphaned" admin page (not a child of our dashboard); see #2529. We talked about this yesterday but I think we may fix this by introducing a redirect to the dashboard in this case so it may be a non-issue.

So to answer your question, I think it's fine to hook onto `current_screen` to do the same thing since it's before the menu is generated/output, but we should be able to be more precise than simply checking a substring match. We should check that the parent base matches.

package.json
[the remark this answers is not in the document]
aaemnnosttv: Thanks! I'm not against using the new lock file format, but we should make this change as a team to avoid back and forth changes to the lock file like this 👍  
I'll open an issue for it.

assets/sass/components/global/_googlesitekit-publisher-wins.scss
```diff
@@ -84,6 +84,20 @@
...
 		}
+
+		@media (max-width: $bp-mobileOnly ) {
+
+			svg {
+				width: 200px;
+			}
+		}
+
+		@media (min-width: $bp-desktop) {
```
aaemnnosttv, line 95: We can combine these into a single media query
```suggestion
		@media (max-width: $bp-mobileOnly ), (min-width: $bp-desktop) {
```

includes/Core/Admin/Screen.php
ivankruchkoff: If you take a look at my previous comment, `parent_base` is null in most scenarios.

aaemnnosttv: @ivankruchkoff It's not that I didn't look, but it seemed to be set in my testing but I think it appeared that way due to my test method and those being populated later on the instance. Logging it out differently now I see what you mean.

This works however, although I'm not sure what the ramifications of calling this earlier would be:
```php
add_action( 'current_screen', function ( WP_Screen $screen ) {
	global $parent_file;
	$screen->set_parentage( $parent_file );
	// $screen->parent_base is now set
} );
```

Normally, this isn't called until [after the menu is generated](https://github.com/WordPress/WordPress/blob/c504140c321776860fbb48703c36415e933c7e70/wp-admin/admin-header.php#L259) but I'm not sure why 🤔 

This is getting a bit too far down the rabbit hole though so if it's not possible without further hackiness then I'm happy to fallback to the query param-based logic if that's cleanest.

**One other screen-based alternative we could try** would be using the `$screen->id` which _is_ set on `current_screen` and should match the return value of `add_submenu_page` or `add_menu_page` in `Screen::add()` which is where we're already making changes. What do you think?

package.json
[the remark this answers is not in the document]
ivankruchkoff: Already opened: https://github.com/google/site-kit-wp/issues/2911

includes/Core/Admin/Screen.php
ivankruchkoff: Updated to use `get_current_screen()->id` per our sync and it also covers the dev/tester plugin pages too.

includes/Core/Admin/Screen.php
```diff
@@ -129,9 +130,55 @@ public function add( Context $context ) {
...
+					 * WordPress provides some color schemes out of the box, but they can also be added via
+					 * wp_admin_css_color()
+					 *
+					 * Our workaround is to set the icon and subsequently replace it in current_screen, which is
+					 * what we do in the following action.
+					 */
+					add_action(
+						'current_screen',
+						function() {
+							global $menu, $_wp_admin_css_colors, $pagenow;
```
aaemnnosttv, line 159: `$pagenow` is no longer used and can be removed.
```suggestion
							global $menu, $_wp_admin_css_colors;
```

includes/Core/Admin/Screen.php
```diff
@@ -129,9 +130,55 @@ public function add( Context $context ) {
...
+							// @see wp_admin_css_color().
+							$color_key = false === strpos( get_current_screen()->id, 'googlesitekit' ) ? 'base' : 'current';
+
+							if ( empty( $_wp_admin_css_colors[ $color_scheme ]->icon_colors[ $color_key ] ) ) {
+								return;
+							}
+
+							$color = $_wp_admin_css_colors[ $color_scheme ]->icon_colors[ $color_key ];
+
+							foreach ( $menu as &$item ) {
```
aaemnnosttv, line 172: This should be fine, but since `current_screen` will run any time `set_current_screen` is called, we shouldn't assume `$menu` will always be an array. Let's add a guard at the top to simply bail if `$menu` isn't an array to be extra safe.

aaemnnosttv (commented):
This is super close. The main thing left to address is a discrepancy in the size of the icon which is now slightly larger due to being an svg background image rather than an image.

image

We need to preserve the 16x16 dimensions of the icon. I found this was possible by applying a `background-size: 16px 16px` rule to it with CSS but maybe there's a better way?

includes/Core/Admin/Screen.php
ivankruchkoff: Resolved in https://github.com/google/site-kit-wp/pull/2834/commits/373c8c0fab5be7835786b64c91c6d21694140a6b

ivankruchkoff:
> This is super close. The main thing left to address is a discrepancy in the size of the icon which is now slightly larger due to being an svg background image rather than an image.
> 
> image
> 
> We need to preserve the 16x16 dimensions of the icon. I found this was possible by applying a `background-size: 16px 16px` rule to it with CSS but maybe there's a better way?

So I think WP icons are made to trigger people with OCD. I added some padding to the svg itself, updated it in https://github.com/google/site-kit-wp/pull/2834/commits/574bb2debe8147516e19ecbb2307c0ae60f54755 and then took a screenshot and cropped it to see if the dimensions fit. 

- It aligns with the widest icon, the media library.
- It's about 2px larger than the dashboard image above it.
- It's about 1px larger than the posts pin.
- Pages is smaller and off centre.
- Comments icon is smaller and off centre.
- 
Screen Shot 2021-03-04 at 12 32 54 PM


For comparison, the old svg icon approach:
Screen Shot 2021-03-04 at 12 43 44 PM
- The G icon is touching the border on left, and 1px shy on right.
- The pin is larger than the dashboard icon
- The media icon is 2px larger

aaemnnosttv (commented):
Thanks @ivankruchkoff. I'm less worried about the icon matching other WP icons and that the size matches the image we had before.

It looks like it's just a _tiny bit_ larger that it was before:
image
image
image

I tried fiddling with it a little and found that it seems to line up exactly when the svg dimensions match the `viewbox` using the original svg on the issue.
image

I'll push this up so we can put this one to rest 😄

aaemnnosttv:
@googlebot I consent.

aaemnnosttv (approved):
Awesome! Thanks @ivankruchkoff 👍

Issue #2626, opened 2021-01-14
Implement framework for "feature tour tooltips"


felixarntz:
We need to integrate a set of JS components to Site Kit to enable tooltip-based guided tours through new product features. These should be possible to be applied across any Site Kit-owned screen and encompass one or more tooltips that can be navigated through sequentially.

<img width="606" alt="Screen Shot 2021-01-14 at 1 52 28 PM" src="https://user-images.githubusercontent.com/3531426/104653644-c3b96d80-566f-11eb-9722-f35d9b21458b.png">

We should try to find a suitable JS library that helps us with this issue, rather than implementing everything ourselves.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Site Kit should include a set of reusable JavaScript components (e.g. `TourTooltips` with `TourTooltip` children) that can be used to display a tooltip-based guided tour through certain new product features.
* The `TourTooltips` component should be configurable with one or more `TourTooltip` instances that have a sequential order.
    * Each tooltip should follow the [design of the blue tooltips in this Figma file](https://www.figma.com/file/RL66yqI7W2Gdlt910KFVYK/Tooltips-%2F-coachmarks?node-id=105%3A65). They're using Material design, so we should use a corresponding dependency for Material tooltips if available.
    * The headline and body content should be custom-specified per `TourTooltip`. It should be possible to use basic HTML elements like links, emphasized content etc in the copy.
    * Every tooltip should have an "x" in the upper right corner which dismisses the tooltips sequence overall.
    * There first tooltip in the sequence must always have a "Next" button which goes to the next tooltip (unless it is also the last tooltip, see below).
    * The last tooltip in the sequence must always have a "Got it" button which dismisses the tooltips sequence overall.
    * All tooltips in between should have both a "Back" and "Next" button for navigating between tooltips.
* Every tooltip should be attached to a certain DOM element on the page, and the tiny triangle on its edge should point to that element (the triangle can appear on any edge of the tooltip). Tooltips should appear in a way that they're always within the viewport, and if the next tooltip is outside of the viewport, it should be scrolled there.
* While a tooltips tour is currently shown, no interaction with other elements on the screen should be possible (i.e. there should be an overlay).
* When a sequence of tooltips is dismissed, it should be stored in `localStorage` (for now) so that the tooltips do not show up again for that user.
[...]

Issue #2650, opened 2021-01-19
Update tooltip tour components to persistently dismiss tours


felixarntz:
#2626 introduced tooltip tour JS components, relying on a temporary mechanism to dismiss a specific tour by storing its slug in `localStorage`. #2648 introduced the server-side and client-side infrastructure for persistent storage of such a dismissal, so we now need to switch the JS components to rely on that mechanism.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Instead of storing dismissed tooltip tours in `localStorage`, they should be dismissed via the `dismissTour( slug )` action of the `core/user` store.

## Implementation Brief

* Update the `TourTooltips` component:
  * Use the `isTourDismissed( slug )` selector of the `core/user` datastore instead of checking a variable in the localStorage to determine whether a tour is dismissed;
  * Call the `dismissTour( slug )` action of the `core/user` datastore instead of setting a variable in the localStorage when a tour ends.

### Test Coverage

* N/A

### Visual Regression Changes

* N/A

## QA Brief

* Connect Site Kit and set up Analytics.
* Ensure your `googlesitekit_initial_version` user option is set to an old version of Site Kit, below 1.25.0.
* Go to the dashboard, the "All Traffic" widget feature tour should show up.
* Once you've completed it, it shouldn't show again when you reload the dashboard.
* Delete your local storage. When visiting the dashboard now, the tour should _still_ not show again, since it's stored in the WordPress database now as part of this issue.

## Changelog entry

* Implement persistent dismissal of feature tours, stored in the database.

felixarntz:
IB ✅

felixarntz:
@eugene-manuilov #2626 is now merged.

felixarntz:
@eugene-manuilov Looks like @aaemnnosttv included this in #2934 (PR for #2739). We somewhat forgot to tag this issue here with the release. It's probably fine to keep this in #2934 now, since it's only a small chunk of code there anyway and since #2866 is out of date.

I'll add the release now (keep this issue for tracking) and close the outdated PR.

aaemnnosttv:
Unfortunately the PR where this was introduced broke the story for the `TourTooltips` component.

![image](https://user-images.githubusercontent.com/1621608/112619778-604a7980-8e30-11eb-9a1d-395d353a7b9e.png)

Sending back for a quick followup.

felixarntz:
[...]

PR #2835, opened 2021-02-22
Feature/2626 add TourTooltips component to enable tooltip-based guided tours through product features.

caseymorrisus:
## Summary

Addresses issue [#2626](https://github.com/google/site-kit-wp/issues/2626)

**Note:** In the interest of providing clarity for those working on related tasks, this PR is being created as a draft while unit test coverage is being finalized. Final Jest commit should follow shortly after the creation of this PR and shouldn't affect any clarity or expectations set within this PR. See component story within `tour-tooltips.stories.js` for example usage.

### No Overlay Blur

Overlay (darkening layer) within Figma design has `backdrop-filter: blur(20px)` applied — difficult or impossible to work into `react-joyride`.

### How React Joyride Applies Masked Overlay & How That Affected Implementation:

Library uses a mask filter (black background, white rectangle w/ hard light blending) to darken everything except the area within the spotlight. Applying a blur within this setup causes the spotlight content to blur as well — this is not wanted. I experimented to verify if it was possible to hack around it with pseudo-elements (`::before` or `::after`) — no luck. I can theoretically think of a way to do this outside of `react-joyride` but this would require building a touring library ourselves from scratch which seems unnecessary unless the blur effect is a hard requirement.

### Spotlight Blue Border

Due to limitations with the masking technique explained above, the blue border is not completely opaque as it is within the Figma design — I was able to get close with a slight modification to the masking technique.

### Action Items

Verify if style-related technical choices are blockers. If not blockers & still desired after considering the amount of work involved — create an issue(s) to fix.

## Checklist

- [ ]  My code is tested and passes existing unit tests.
- [ ]  My code has an appropriate set of unit tests which all pass.
- [x]  My code is backward-compatible with WordPress 4.7 and PHP 5.6.
- [x]  My code follows the [WordPress](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) coding standards.
- [x]  My code has proper inline documentation.
- [ ]  I have added a QA Brief on the issue linked above.
- [x]  I have signed the Contributor License Agreement (see [https://cla.developers.google.com/](https://cla.developers.google.com/)).


assets/js/components/TourTooltip.js
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * Creates an array of length equal to size with incremental values starting at 0.
+ *
+ * @since n.e.x.t
+ *
+ * @param {number} size Array length.
+ * @return {number[]} Array of incremental numbers.
+ * @example createIncrementalArrayBySize(3); // [0, 1, 2]
+ */
+const createIncrementalArrayBySize = ( size ) => new Array( size ).fill().map( ( _, i ) => i );
```
eugene-manuilov, line 42: Looks like this function is used only once in the TourTooltip component. Why not use it directly there?

```javascript
const indicatorArray = new Array( size ).fill().map( ( _, i ) => i );
const getIndicatorClassName = ( indicatorIndex ) => ...
```

assets/js/components/TourTooltip.js
```diff
@@ -0,0 +1,140 @@
...
+								{ ...closeProps }
+								aria-label="Got it"
+								title="Got it"
+							>
+								Got it
+							</Button>
+						) }
+					</div>
+				</CardActions>
+				<button
```
eugene-manuilov, line 111: This also should be the `<Button...> ... </Button>` element, shouldn't it?

eugene-manuilov (changes requested):
This looks great! Added some comments. @caseymorrisus could you please take a look at it?

stories/tour-tooltips.stories.js
```diff
@@ -0,0 +1,369 @@
...
+						This is the fourth step with an external link that should go to Google, <Link href="http://google.com" external>link.</Link>
+					</div>
+				),
+			},
+		];
+
+		return (
+			<WithTestRegistry>
+				<MockWPDashboard />
+				<TourTooltips steps={ steps } tourID="feature" />
```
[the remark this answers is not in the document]
caseymorrisus: @eugene-manuilov While making this change, I realized it would remove our ability to verify conditional rendering within the QA Brief. I'm on the fence — using a static `tourID` allows us to test the component using expected functionality within Storybook, but could potentially confuse an engineer if they had already cleared the tour, forgot, then tested within Storybook again at a later date.

What do you think?

assets/js/components/TourTooltip.js
caseymorrisus: @eugene-manuilov Forgive the wordy response! I tried my best to refine it down to be as short as possible and still convey the intended info. 😅

I created this function to prevent technical debt by adding clarity around an otherwise confusing or scary-looking piece of logic via explicit function naming. I know the `Array.fill.map` trick now but cannot guarantee I (or someone else) will know it in the future. This is useful while debugging or otherwise trying to understand component logic — when working memory is a significant factor in how long it takes to fix an issue. I prefer this method over comments alone as comments are easy to ignore when skimming through logic and have a greater chance to contain stale information.

This has the added benefit of providing example usage (and data return) on-hover, in-editor, which can help if there's any confusion with function naming (JSDoc description can go even further). While not a big deal in this case as there's nothing to be mentally parsed above `indicatorArray` — this abstraction keeps the render function clean (easy to parse quickly) by keeping related logic colocated. There are no comments or unnecessary logic visually breaking the top-down flow of logic.

This also decouples the logic within `createIncrementalArrayBySize` away from the render method of the component. As a result, we can assume the function does not contain any side effects that are component related (props, state, anything component-specific) — meaning we can ignore it when mentally parsing through component logic as any unit complex enough to warrant such a worry should be guarded by unit tests in the first place. Logic & implementation details of the function can be ignored altogether — only the data returned is of concern. 

Significantly less of a concern to me — and somewhat counterintuitively, this can result in micro-perf benefits under certain circumstances due to JIT compiler optimizations (otherwise, within the margin of error). This is usually only relevant when being called within a loop, but it doesn't hurt to mention.

Because I'm a caveman and probably explained things poorly, I'll end this with a Martin quote: *“Indeed, the ratio of time spent reading versus writing is well over 10 to 1. We are constantly reading old code as part of the effort to write new code. ...[Therefore,] making it easy to read makes it easier to write.”*

Let me know how you feel about my thoughts above! In no way am I particularly attached to them. I'm happy to make the change if it seems more confusing than helpful. Just wanted to shed light on the thought process behind the decision. 🙏

eugene-manuilov: Well, I have nothing against it, but in this case, let's move it to util functions then. So, we can use it in different components if need be.

assets/js/components/TourTooltip.js
```diff
@@ -0,0 +1,147 @@
...
+							</Button>
+						) }
+					</div>
+				</CardActions>
+				<Button
+					className="googlesitekit-tooltip-close"
+					text
+					{ ...closeProps }
+				>
+					<CloseIcon width="14" height="14" />
```
eugene-manuilov, line 122: This icon should be passed via `icon` prop:

```javascript
<Button
    className="googlesitekit-tooltip-close"
    text
    icon={ <CloseIcon width="14" height="14" /> }
    { ...closeProps }
>
```

Also currently the Button component requires a string to be passed as its children. We probably need to update the Button component to make the children prop be optional.

assets/js/components/TourTooltip.js
```diff
@@ -0,0 +1,147 @@
...
+				</div>
+				<CardActions className="googlesitekit-tooltip-actions">
+					<ul className="googlesitekit-tooltip-indicators">
+						{ indicatorArray.map( ( indicatorIndex ) => (
+							<li
+								key={ `indicator-${ indicatorIndex }` }
+								className={ getIndicatorClassName( indicatorIndex ) }
+							/>
+						) ) }
+					</ul>
```
eugene-manuilov, line 84: Currently, if there is only one step in the tour, it shows a single bullet which looks a bit confusing. What if we hide these indicators if we have less than 2 steps? cc @felixarntz

stories/tour-tooltips.stories.js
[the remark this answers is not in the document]
eugene-manuilov: Ok, let's leave it as is for now. It will be updated in the #2650.

eugene-manuilov (changes requested):
Thanks, @caseymorrisus. This is almost good to go. Just added a few more comments. Please, take a look.

eugene-manuilov:
@caseymorrisus could you please js tests? Something is broken there.

assets/js/components/TourTooltip.js
caseymorrisus: Hiding indicators when tour only contains a single step seems like the most-clear option. I added this, but can quickly remove it if necessary.

caseymorrisus:
> @caseymorrisus could you please js tests? Something is broken there.

@eugene-manuilov JS tests are now passing. Based on logs, E2E test failures should be unrelated.

eugene-manuilov (dismissed):
Awesome! Thanks, @caseymorrisus. Looks good to me.

assets/js/components/TourTooltip.js
```diff
@@ -0,0 +1,136 @@
...
+								{ __( 'Next', 'google-site-kit' ) }
+							</Button>
+						) }
+						{ isLastStep && (
+							<Button
+								className="googlesitekit-tooltip-button"
+								text
+								{ ...closeProps }
+								aria-label={ gotItText }
+								title={ gotItText }
```
felixarntz, line 100: @caseymorrisus What's the reason for making this an `aria-label` and `title` here, but not do the same for the other buttons? Or, why is that even necessary here? Since the button has a human-readable text as content, why the extra props/attributes here?

caseymorrisus: @felixarntz I agree that it seems like overkill (and may even be bad a11y) but wanted to be consistent with the `react-joyride` library and its other buttons. Props provided by `react-joyride` that are being spread (`closeProps`) include `aria-label` and `title` values. The reason this button is the only one to have overrides is because `closeProps` prop is trying to set the `aria-label` to `Close` (as well as `title`) for the button that says `Got it` as I'm re-using the rest of the functionality provided.

felixarntz: Leaving this here after our Slack message exchange: We'll need to ensure no user-facing strings come from Joyride defaults, since otherwise they wouldn't be translated via WordPress.

assets/js/components/TourTooltips.js
```diff
@@ -0,0 +1,147 @@
...
+			callback={ handleJoyrideCallback }
+			continuous
+			disableOverlayClose
+			floaterProps={ floaterProps }
+			run={ run }
+			showProgress
+			stepIndex={ stepIndex }
+			steps={ parsedSteps }
+			styles={ joyrideStyles }
+			tooltipComponent={ TourTooltip }
```
felixarntz, line 138: We should provide all standard strings here via the `locale` prop, see https://docs.react-joyride.com/props. The individual `TourTooltip` component shouldn't need to override anything then and could use the provided props, making for a cleaner solution.

felixarntz (changes requested):
Looks almost good, just the issue about localization.

assets/js/components/TourTooltip.js
```diff
@@ -0,0 +1,136 @@
...
+								{ ...primaryProps }
+							>
+								{ __( 'Next', 'google-site-kit' ) }
+							</Button>
+						) }
+						{ isLastStep && (
+							<Button
+								className="googlesitekit-tooltip-button"
+								text
+								{ ...closeProps }
```
felixarntz, line 98: This button should also use `primaryProps` instead of `closeProps`; the latter is meant only for the Close icon, not for the last primary button (although that also closes the tooltip). See https://docs.react-joyride.com/custom-components#tooltipcomponent

caseymorrisus:
@felixarntz I just pushed up changes to handle translating string content and attribute values using the Joyride `locale` prop. I had the choice between `title` and `aria-label`  to use for button content and went with `title` — both pull from the same `locale` property.

aaemnnosttv:
@googlebot I consent.

felixarntz (approved):
Excellent!

Issue #2737, opened 2021-02-01
All Traffic pie chart tooltip is too far away to click link


wpdarren:
## Bug Description

For the new All Traffic pie chart, the tooltip is displayed slightly away from the slice. So, when you move your mouse to go over the tooltip to click on the link, it closes.  In the example below the blue slice doesn't have a link but if it did I would not be able to click it. I suspect it might also depend on the screen size. 

## Steps to reproduce

1. Go to 'Dashboard' and scroll to the new pie chart
2. Click on a Slice and observe position of tooltips

## Screenshots

![pie](https://user-images.githubusercontent.com/73545194/106247983-4daa2e80-624b-11eb-9daa-1c379dbaca60.gif)

--------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The All Traffic widget pie chart slice tooltips should always overlap their respective pie slice so that the user can move their cursor over there without leaving the pie slice, which would have the tooltip disappear, i.e. make it unreachable.
    * This might be hard to reproduce consistently, please check in different browsers.
* The "Others" tooltip should no longer include a link. It should instead have a text (different per tab / dimension name):
    * _See the full list of channels in Analytics_
    * _See the full list of locations in Analytics_
    * _See the full list of devices in Analytics_

## Implementation Brief

* Add the following code before the render in `GoogleChartV2`:
```js
if ( dimensionValue?.length ) {
	options.tooltip.trigger = 'selection';
} else {
	options.tooltip.trigger = 'focus';
}
```

### Test Coverage

* N/A

### Visual Regression Changes

* 

## QA Brief

* Click a pie slice.
* After the slice is selected, you should be able to move the mouse outside the slice and the tooltip should stay visible. This means you can select it.
* This should work on a browser window with a viewport larger than `1400px` wide.

## Changelog entry

* Show All Traffic pie chart tooltips persistently when a slice is selected to allow interaction with the tooltip, which would not be possible when hovering.

tofumatt:
This component will use the new `GoogleChartV2` component, so marking it as blocked by that issue.

tofumatt:
I a fair bit of time today going through Google Chart issues in GitHub/Google Code archives, Stack Overflow, etc.

Unfortunately this seems to be an issue with Google Charts that we can't resolve ourselves, if we want to maintain the hover/focus state only.

One thing to note is that this problem is especially bad on wider screens, because the chart is set to a width of 100%:

[...]

PR #2845, opened 2021-02-22
Keep slice selected after clicking it in All Traffic Widget

tofumatt:
## Summary

Addresses issue #2737.

## Relevant technical choices

The IB had a typo and mentioned putting this bit of code in `GoogleChartV2`, but it should have been in the `UserDimensionsPieChart` component before it _renders_ the `GoogleChartV2`.

**Note:** this PR includes code from the PR for #2714 and can't be reasonably reviewed until #2805 is merged.

## Checklist


felixarntz (changes requested):
Given the limitations the charts library imposes upon us, this looks good. Of course the perfect scenario would be if tooltips were always overlapping their respective slice and then, that hovering over the tooltip would keep the tooltip open. But that seems infeasible with the library while at the same time not re-introducing the weird flickering bug on hover.

There is just one thing we should also change here: Right now it is still impossible to click a link in the tooltip or interact with it in any other way, due to `pointer-events: none` in CSS. We should be able to override that behavior for when a slice is selected.

felixarntz (approved):
I've made the quick CSS change here.

This now results in okay behavior; the main problems is still that the "Others" slice tooltip link is not reachable because the only way to get to such a link is when the slice is selected, but "Others" cannot be selected.

@tofumatt If you have an idea on how to address that, potentially via altering some other behavior, let's discuss that. I'll also bring it up with Mariya to explore alternative options for that link.

Issue #2856, opened 2021-02-23
Single URL dashboard still displays old All Traffic widget


felixarntz:
## Bug Description

In a production build (i.e. without widgets feature flags enabled), the single URL dashboard still displays the old All Traffic widget. This must have been an oversight, and I assume we have been testing too much using the widgets API-based version, thus missing that critical bug.

Originally noticed in https://github.com/google/site-kit-wp/issues/2646#issuecomment-779773565

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The legacy version (which is still used in production for now) of the single URL dashboard should display the new All Traffic widget instead of the old one.
* The old All Traffic widget component used there should be removed, unless it is also used somewhere else.

## Implementation Brief

* Review and merge #2859

### Test Coverage

* N/A

### Visual Regression Changes

* The legacy version of the single URL dashboard should display the new All Traffic widget now.

## QA Brief

* Make sure you don't have the `widgets.pageDashboard` enabled on your local site;
* Go to the **single URL** dashboard page and make sure you see the new version of the All Traffic widget.

## Changelog entry

* Fix bug where the old All Traffic widget was still displaying in the dashboard for a single URL.

felixarntz:
IB ✅  😆

wpdarren:
### QA Update: ⚠️ Engineer to confirm 

- On single URL dashboard page the new version of the All Traffic widget appears. 

![image](https://user-images.githubusercontent.com/73545194/108999345-1d846d00-76dd-11eb-9756-86b65b7b7329.png)

@eugene-manuilov @felixarntz  I've run through additional checks to make sure that it functions as expected.  I found two issues, that either need new tickets, or, they need to be fixed as part of this ticket? 

1) For channel dimension. For a period of 28 days. There's a legend for `social` and `(other)` that have zero values, so I am assuming these should not appear.  They do not appear in the pie chart, just the legends. 

![image](https://user-images.githubusercontent.com/73545194/109000464-b667b800-76de-11eb-866d-51be8aa1da4a.png)

2) For channel dimension. When I extended the period to 90 days, the `(other)` option had 1 user, but it does not appear in the pie chart. I am assuming it should. 

![image](https://user-images.githubusercontent.com/73545194/109000934-43127600-76df-11eb-9db8-1aee199b78d6.png)

I couldn't see any issues on other periods and the other dimensions appear to be fine on my test site.

PR #2859, opened 2021-02-23
Bug/2856 old all traffic

eugene-manuilov:
## Summary

Addresses issue #2856 

## Checklist


assets/js/modules/analytics/components/dashboard-details/AnalyticsDashboardDetailsWidgetTopAcquisitionSources.js
```diff
@@ -20,71 +20,33 @@
...
-									mdc-layout-grid__cell--span-4-phone
-								">
-								<LegacyAnalyticsDashboardWidgetTopAcquisitionSources />
-							</div>
-						</div>
-					</div>
+			</Cell>
+			<Cell size={ 12 }>
+				<Layout>
+					<DashboardAllTrafficWidget { ...widgetComponentProps } />
```
aaemnnosttv, line 47: @eugene-manuilov Should we maybe use [`LegacyDashboardAllTraffic`](https://github.com/google/site-kit-wp/blob/404d20abe2685203fa71e9ef3af7b7ec1c4e8c91/assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/LegacyDashboardAllTraffic.js) here instead?

In which case, I think we may even be able to use that in place of `AnalyticsDashboardDetailsWidgetTopAcquisitionSources` altogether?

assets/js/modules/analytics/components/dashboard-details/AnalyticsDashboardDetailsWidgetTopAcquisitionSources.js
```diff
@@ -20,71 +20,33 @@
...
 
 	return (
 		<Fragment>
-			<div className="
-					mdc-layout-grid__cell
-					mdc-layout-grid__cell--span-12
-				">
+			<Cell size={ 12 }>
 				<DashboardModuleHeader
 					title={ __( 'All Traffic', 'google-site-kit' ) }
```
felixarntz, line 41: This should also say "Your Traffic at a Glance" instead.

felixarntz (changes requested):
LGTM, just one more thing to change here.

I'll quickly do that, and I'll also take care of removing some of these components, which we shouldn't need anymore.

assets/js/modules/analytics/components/dashboard-details/AnalyticsDashboardDetailsWidgetTopAcquisitionSources.js
felixarntz: @aaemnnosttv @eugene-manuilov I went ahead and deleted this and other now unused components now, I think keeping it in the component where it used to be is fine, even though naming-wise this is a bit worse. But it's soon-to-be-removed legacy code anyway.

felixarntz (approved):
Great!

Issue #2724, opened 2021-01-29
Allow PHP8


TorbenLundsgaard:
## Feature Description

Start testing all pull requests on PHP8

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a new GitHub action that replicates the "PHP (PHP 7.4, WordPress Latest)" one, except that it uses PHP 8.0.
* Any failures, errors or warnings that this job results in should be addressed.
    * For the IB, it's probably best to set up PHP 8 locally and run the tests to see where failures, errors and warnings come up.

## Implementation Brief

* Create a new PHP test Github Action: '.github/workflows/php-tests-wp-latest-php-8-0.yml' from the existing: `.github/workflows/php-tests-wp-latest-php-7-4.yml` actions, updating the PHP version to `8.0`.
* Update the 7 tests that fail because of [`Method ReflectionParameter::isArray() is deprecated`](https://php.watch/versions/8.0/deprecated-reflectionparameter-methods#isArray)
   1. `Google\Site_Kit\Tests\Core\Authentication\AuthenticationTest::test_require_user_input` - `tests/phpunit/integration/Core/Authentication/AuthenticationTest.php`
   1. `Google\Site_Kit\Tests\Core\Authentication\AuthenticationTest::test_require_user_input__without_feature` - `tests/phpunit/integration/Core/Authentication/AuthenticationTest.php`
   1. `Google\Site_Kit\Tests\Core\Authentication\AuthenticationTest::test_disconnect` - `tests/phpunit/integration/Core/Authentication/AuthenticationTest.php`
   1. `Google\Site_Kit\Tests\Core\Authentication\Clients\OAuth_ClientTest::test_authorize_user` - `tests/phpunit/integration/Core/Authentication/Clients/OAuth_ClientTest.php`
   1. `Google\Site_Kit\Tests\Core\Authentication\Clients\OAuth_ClientTest::test_refresh_profile_data` - `tests/phpunit/integration/Core/Authentication/Clients/OAuth_ClientTest.php`
   1. `Google\Site_Kit\Tests\Modules\Analytics\Advanced_Tracking\Script_InjectorTest::test_inject_event_script` - `tests/phpunit/integration/Modules/Analytics/Advanced_Tracking/Script_InjectorTest.php`
[...]

PR #2867, opened 2021-02-24
Add a new workflow for PHP Tests with PHP 8.0 and update PHPUnit setup for compatibility

benbowler:
## Summary

Addresses issue #2724

Started by @benbowler, finished by @aaemnnosttv 

## Relevant technical choices

* WP core compatibility with PHP 8 is for PHPUnit 7 only so we needed to allow for this to be installed by widening the version constraint for `phpunit/phpunit` 
    * `config.platform.php` will still limit packages to those compatible with PHP 5.6 during install/update so we explicitly update it in the workflow with an added step, which ignores platform requirements
* PHPUnit 7 updated the `TestCase::checkRequirements` method to be private so it's no longer extendable. This was only used for polyfiling the `ms-required|ms-excluded` group test skip-by-annotation that we had added to fix for older versions of WordPress but this will no longer work at all because `checkRequirements` is no longer extendable.
    * To work around this, I split out a separate `phpunit.multisite.xml` config file and updated the default config file to exclude the groups that don't apply (this is what core does). I looked for a ticket on trac for this but found nothing so they may not even be aware that this skip-by-annotation no longer works.
    * The result is a bit better experience anyways IMO since PHPUnit would previously finish with warnings due to skipped tests. Now these tests are filtered out by group so they aren't even included in the list of tests to be run, so they're no longer counted as skipped, resulting in happier looking finish when all passes 😄 
* I added missing assertions to one test (my own from recently) that cause the test to fail in the newer version of PHPUnit
* Added a new `test:multisite` script to Composer

## Checklist

### DIFF
--- a/phpunit.multisite.xml
+++ b/phpunit.multisite.xml
@@ -10,7 +10,7 @@
     </testsuites>
     <groups>
         <exclude>
-            <group>ms-required</group>
+            <group>ms-excluded</group>
         </exclude>
     </groups>
     <php>
@@ -19,5 +19,6 @@
         <ini name="display_startup_errors" value="1" />
         <const name="GOOGLESITEKIT_TESTS" value="1" />
         <const name="WP_HTTP_BLOCK_EXTERNAL" value="1" />
+        <const name="WP_TESTS_MULTISITE" value="1" />
     </php>
 </phpunit>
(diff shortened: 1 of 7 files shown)


aaemnnosttv:
@googlebot I consent.

phpunit.multisite.xml
```diff
@@ -0,0 +1,24 @@
+<phpunit
```
aaemnnosttv, line 1: Core puts this file in `tests/phpunit/multisite.xml` but I thought it would be better to keep them side-by-side to ensure they're kept up-to-date. AFAIK there is no support for inheritance for these files so we need to duplicate them 😕

felixarntz (approved):
Awesome! +1 for improved multisite test configuration

PR #2879, opened 2021-02-25
Fix Sparkline redraw bug by using GoogleChartV2 for Sparkline.

tofumatt:
## Summary

Addresses issue #2714. Applied against `master` to fix the release issue.

## Relevant technical choices

I'll admit I'm not _exactly_ sure why this happens. While I think it’s related to our own legacy `GoogleChart` component’s usage of the `google` globals for interacting with charts in a debounced manner… I’m honestly not sure. What I do know is that it only happens for `SparkLine` charts and migrating them to use `GoogleChartV2` was easy and didn't seem to cause any issues.

It seems largely that our old Google Charts component dealt with a lot of globals and assumed it existed in isolation. Now that we have another component handling charts code it gets messed up, as this bug didn’t seem to exist in the previous version before I introduced `GoogleChartV2` when I tested locally.

## Checklist

### DIFF
--- a/.storybook/storybook-data.js
+++ b/.storybook/storybook-data.js
@@ -65,7 +65,7 @@ module.exports = [
 			options: {
 				hierarchyRootSeparator: '|',
 				hierarchySeparator: {},
-				readySelector: '.googlesitekit-line-chart > div[style="position: relative;"]',
+				readySelector: '.googlesitekit-chart-v2 > div[style="position: relative;"]',
 			},
 		},
 	},
@@ -79,7 +79,7 @@ module.exports = [
 			options: {
 				hierarchyRootSeparator: '|',
 				hierarchySeparator: {},
-				readySelector: '.googlesitekit-line-chart > div[style="position: relative;"]',
+				readySelector: '.googlesitekit-chart-v2 > div[style="position: relative;"]',
 			},
 		},
 	},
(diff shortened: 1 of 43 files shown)


felixarntz (changes requested):
@tofumatt The PR seems to be based off `develop`, since it includes additional changes that shouldn't be in `master` yet. Can you check?

I'm somewhat hesitant to go with this approach. In how many other places do we use `GoogleChart`? If we now already change it in so many places, maybe there aren't so many more anyway. Still, it would be best if we could find a solution for this problem specifically. The error message at least seems fairly precise in where it's happening (`chart.draw` call in our `GoogleChart` component).

felixarntz:
@tofumatt If #2880 works as expected, we should be able to close this one.

tofumatt:
> @tofumatt The PR seems to be based off `develop`, since it includes additional changes that shouldn't be in master yet. Can you check?

Ah, indeed, whoops! That's fixed now and this only includes the relevant commit.

It looks like this is a more stable solution than #2880 given https://github.com/google/site-kit-wp/pull/2880#pullrequestreview-598971548.

It _seems_ that the problem with our GoogleChart component is made worse when there are many on-screen, as is especially the case with Sparkline pages.

I tested the old component on other charts pages and didn't see the resize error like there was here.

I know it's a more drastic change than we wanted but I think this is the most stable approach as the new charts library seems to have zero issues causing call stack or `draw()` errors, especially on resize. I'd say we'll get more stability by moving to it entirely. 🤷🏻‍♂️

felixarntz:
Closing in favor of #2881 which is based off `master`.

PR #2880, opened 2021-02-25
Use current Google charts library verrsion again to prevent conflict with react-google-charts

felixarntz:
## Summary

Addresses issue #2714

## Relevant technical choices

* `react-google-charts` (used for new `GoogleChartV2` component) doesn't specify any Google charts library version, i.e. it relies on the latest one.
* In PHP however (used for old `GoogleChart` component), we load the Google charts library in a specific version 42. This results in a conflict between versions, which causes the error.
* We only used the specific 42 version for about the last month (see https://github.com/google/site-kit-wp/commit/8097bbdd5f9d617348546a536ba021e64b7f924c#diff-3ff56820a1aad4fa15985200da4db645e6e443ca74ce9495f1faa159fa43e2d9), to fix the tooltip flickering issue (it was basically introduced after version 42). However, in the meantime we've already resolved that issue in a different way anyway, so we can safely go back to "current".

## Checklist

### DIFF
--- a/includes/Core/Assets/Assets.php
+++ b/includes/Core/Assets/Assets.php
@@ -393,7 +393,7 @@ final class Assets {
 					'in_footer'    => false,
 					'before_print' => function( $handle ) {
 						// The "42" version is important because it contains a fix for the tooltip flickering issue.
-						wp_add_inline_script( $handle, 'google.charts.load( "42", { packages: [ "corechart" ] } );' );
+						wp_add_inline_script( $handle, 'google.charts.load( "current", { packages: [ "corechart" ] } );' );
 					},
 				)
 			),


tofumatt (approved):
Awesome! That's safer than my fix in #2879, let's 🚢

tofumatt (commented):
Actually, I just noticed this can still cause errors. It stops the [error reported here](https://github.com/google/site-kit-wp/issues/2714#issuecomment-785909760), but can cause another error with enough resizes:

tofumatt (changes requested):
After testing out both locally, I think #2879 contains the more robust solution here, so I think that's what we should go with.

tofumatt (approved):
Argh, that was supposed to be a comment, user-error.

What I was going to say is that merging this as well is a good idea, as there's no reason to force the older version of Google Charts now.

felixarntz:
Included this fix in #2879 - closing.

Issue #2794, opened 2021-02-11
AdGuard adblocker no longer detected


aaemnnosttv:
## Bug Description

The popular [AdGuard ad blocker](https://chrome.google.com/webstore/detail/adguard-adblocker/bgnkhhnnamicmpeenaelnjfhikgbkllg) is no longer detected by our detection mechanism.

This allows users to proceed with AdSense setup where it will fail when AdSense module API requests are blocked.

![image](https://user-images.githubusercontent.com/1621608/107667509-4dbd1c00-6c98-11eb-8e7a-e953e784e82f.png)

![image](https://user-images.githubusercontent.com/1621608/107667515-4eee4900-6c98-11eb-8ce7-0d02c501875d.png)

![image](https://user-images.githubusercontent.com/1621608/107667527-5150a300-6c98-11eb-9ff6-9809de6ac41d.png)

Another popular ad blocker, Ad Block Plus is still detected so the breakage seems to be a result of changes to the blocking rules no longer preventing our test script from loading.

![image](https://user-images.githubusercontent.com/1621608/107667782-9aa0f280-6c98-11eb-8e17-4938999869af.png)

The chrome extension alone reports 10M+ users and their website advertises 30M+ cross-platform users so this will likely impact some of our users.

## Steps to reproduce

1. Start with the AdSense module inactive
1. Install [AdGuard ad blocker for your browser](https://github.com/AdguardTeam/AdguardBrowserExtension#installation)
2. Go to Site Kit settings > Connect tab
3. Notice there is no ad blocker detected warning
4. Activate AdSense and start setup
5. Note errors preventing setup from completing

## Additional Context

This is likely due to a change in the AdGuard pattern matching rather than a regression in Site Kit itself since the detection logic is purely based on the URL of the [detection script](https://github.com/google/site-kit-wp/blob/5817746473cc66880feb5c0b621ce2b3cddf0c44/assets/js/pagead2.ads.js).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The ad blocker detection script should be updated so that it is blocked by the current [AdGuard](https://adguard.com/) ad blocker
* Other currently supported ad blockers should continue to be detected

## Implementation Brief

* Remove dedicated external ad-detection script entirely
	- Entry in `Assets`
	- References to `googlesitekit-pagead2.ads` handle
	- Entry point in webpack.config
[...]

Issue #2749, opened 2021-02-03
Expose registry select and dispatch to module `checkRequirements` functions


aaemnnosttv:
## Feature Description

In #2587 we did away with accessing the global registry outside of our JS entry points. A complications of this change exposed areas where we were previously reaching for this instance where we shouldn't, such as the `checkRequirements` function for AdSense, which selects `isAdBlockerActive` from its datastore.

Given that datastore access is a reasonable dependency for module `checkRequirements` functions (see #2130) to have access to, we should provide it to those functions rather than encouraging global access.

See the original discussion here: https://github.com/google/site-kit-wp/pull/2593#discussion_r562093340

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `checkRequirements` function that can be registered for each module (via `registerModule` action of `core/modules`) should be passed the current `registry`, so that it can e.g. access `registry.select` as needed.
* The AdSense module `checkRequirements` function should be modified to use the provided `registry`, and the hack where the necessary selector is temporarily saved into a variable which was needed before should be removed.

## Implementation Brief
In `assets/js/googlesitekit/modules/datastore/modules.js`:
* Pass `registry` as a parameter to the call to `module.checkRequirements()`:
https://github.com/google/site-kit-wp/blob/b460234894bcbe28dd31a9422b57293e78021977/assets/js/googlesitekit/modules/datastore/modules.js#L453
* Add an `@since` comment to the doc block for `registerModule` in `assets/js/googlesitekit/modules/index.js` to reflect that checkRequirements now accepts a `registry` parameter.

In `assets/js/modules/adsense/index.js`:
* Remove the `let isAdBlockerActive = () => {};` on line 55.
* Remove the line in the `registerStore` function that sets `isAdBlockerActive`, and the `TODO` comment.
* Modify the signature of `checkRequirements` to accept the new `registry` parameter:
https://github.com/google/site-kit-wp/blob/b460234894bcbe28dd31a9422b57293e78021977/assets/js/modules/adsense/index.js#L77
* Refactor the call to `isAdBlockerActive` to use `registry.select( STORE_NAME ).isAdBlockerActive()` instead:
https://github.com/google/site-kit-wp/blob/b460234894bcbe28dd31a9422b57293e78021977/assets/js/modules/adsense/index.js#L78-L80

### Test Coverage

* No new tests required; existing tests should pass.

### Visual Regression Changes

* N/A

## QA Brief

[...]

PR #2882, opened 2021-02-26
Enhance adblocker detection with more robust implementation

aaemnnosttv:
## Summary

Addresses issue #2794 

## Relevant technical choices

* Replaces URL pattern-based adblocker detection with more robust library-based solution
* Fixes duplicate module `checkRequirements` function invocations as a result of duplicate resolver functions

image

## Checklist


assets/js/googlesitekit/modules/datastore/modules.js
```diff
@@ -785,11 +783,16 @@ const baseSelectors = {
...
 	 */
-	getCheckRequirementsError( state, slug ) {
+	getCheckRequirementsError: createRegistrySelector( ( select ) => ( state, slug ) => {
 		invariant( slug, 'slug is required.' );
-		const requirementsStatus = state.checkRequirementsResults[ slug ];
-		return requirementsStatus === true ? null : requirementsStatus;
-	},
+
+		// Need to use registry selector here to ensure resolver is invoked.
+		if ( select( STORE_NAME ).canActivateModule( slug ) ) {
```
aaemnnosttv, line 790: This change isn't explicitly needed by this PR but a bug I noticed during implementation.

`checkRequirements` functions were being invoked twice due to separate resolvers being used for `canActivateModule` and `getCheckRequirementsError`. Even though they use the same underlying function, the datastore will run them if present.

assets/js/modules/adsense/datastore/adblocker.js
```diff
@@ -20,6 +20,7 @@
  * External dependencies
  */
 import invariant from 'invariant';
+import { detectAnyAdblocker } from 'just-detect-adblock';
```
aaemnnosttv, line 23: Even though it uses this library here, this could easily be replaced with something else as there are a few that work very similarly. This one seems to be one of the newer ones and seems to work well. If needed we could always fork one of these as an internal package to tweak for our needs.

Part of the issue here is a proof of concept that since our use is encapsulated within the datastore and connected components, we can rely on on-demand asynchronous logic rather than running this on every page load proactively.

assets/js/modules/adsense/datastore/adblocker.test.js
```diff
@@ -55,35 +68,23 @@ describe( 'modules/adsense adblocker', () => {
 
 	describe( 'selectors', () => {
 		describe( 'isAdBlockerActive', () => {
-			it( 'uses a resolver to load status from a global variable by default', async () => {
-				if ( ! global._googlesitekitLegacyData ) {
-					global._googlesitekitLegacyData = {};
-				}
-				global._googlesitekitLegacyData.canAdsRun = true;
```
aaemnnosttv, line 62: Much nicer than messing with a global 👌

assets/js/modules/adsense/index.js
```diff
@@ -54,7 +54,7 @@ let isAdBlockerActive = () => {};
 export const registerStore = ( registry ) => {
 	registerDataStore( registry );
 	// TODO: fix hack
-	isAdBlockerActive = () => registry.select( STORE_NAME ).isAdBlockerActive();
+	isAdBlockerActive = () => registry.__experimentalResolveSelect( STORE_NAME ).isAdBlockerActive();
```
aaemnnosttv, line 57: Issue to fix this hack is here: https://github.com/google/site-kit-wp/issues/2749

felixarntz (changes requested):
Excellent - just one thing more to clean up, which I'll quickly do here.

felixarntz (approved):
Excellent!

Issue #2865, opened 2021-02-24
Issue with Legends/Pie Chart on All Traffic Widget for a Single URL Dashboard Page


wpdarren:
## Bug Description

As part of the QA for the All Traffic Widget on [Single URL dashboard page](https://github.com/google/site-kit-wp/issues/2856)  I've found two issues that are linked. 

1) For channel dimension. For a period of 28 days. There's a legend for `social` and `(other)` that have zero values, so I am assuming these should not appear.  They do not appear in the pie chart, just the legends. 

![image](https://user-images.githubusercontent.com/73545194/109000464-b667b800-76de-11eb-866d-51be8aa1da4a.png)

<details><summary>Non-issue hidden. (Click to expand.)</summary>
(Not an issue, <a href="https://github.com/google/site-kit-wp/issues/2865#issuecomment-785221206">see comment below.</a>)
<strike>
2) For channel dimension. When I extended the period to 90 days, the `(other)` option had 1 user, but it does not appear in the pie chart. I am assuming it should.
</strike>

![image](https://user-images.githubusercontent.com/73545194/109002458-33942c80-76e1-11eb-96e8-87d5626895eb.png)
</details>

I couldn't see any issues on other periods and the other dimensions appear to be fine on my test site. 

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The All Traffic pie chart legend should only include dimension values which are also present in the pie chart (i.e. have data).

## Implementation Brief

* Edit the `extractAnalyticsDataForPieChart` function:
  * Add a new option `hideEmpty` with `false` as its default value;
  * Don't add new row data to the dataMap array if the `hideEmpty` option is truthy and `users` equals to 0.
* Add `hideEmpty: true` option to the extractAnalyticsDataForPieChart function options in the UserDimensionsPieChart component.

### Test Coverage

* N/A

### Visual Regression Changes

* Empty slices should become hidden after implementing the aforementioned changes.

## QA Brief

To test this a channel dimension users data value needs to be set to zero in assets/js/modules/analytics/util/index.js extractAnalyticsDataForPieChart.  Then the legend for that users value should not appear beneath the pie chart.

## Changelog entry

*

felixarntz:
[...]

PR #2890, opened 2021-03-01
2865 Hide legends with no users on all traffic widget.

Hazlitte:
## Summary
Hide legends with no users on all traffic widget.

Addresses issue #2865

## Checklist

### DIFF
--- a/assets/js/modules/analytics/util/index.js
+++ b/assets/js/modules/analytics/util/index.js
@@ -68,6 +68,7 @@ export function extractAnalyticsDataForPieChart( reports, options = {} ) {
 		maxSlices,
 		withOthers = false,
 		tooltipCallback,
+		hideEmpty = false,
 	} = options;
 
 	const data = reports[ 0 ].data;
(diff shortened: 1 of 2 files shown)


assets/js/modules/analytics/util/index.js
```diff
@@ -111,7 +112,9 @@ export function extractAnalyticsDataForPieChart( reports, options = {} ) {
 			rowData.push( tooltipCallback( row, rowData ) );
 		}
 
-		dataMap.push( rowData );
+		if ( ! ( hideEmpty && users < 1 ) ) {
+			dataMap.push( rowData );
+		}
```
felixarntz, line 117: We should be able to wrap the whole block around the `rowData` variable in this clause, right now the `rowData` variable logic starting in line 110 may be unnecessarily run.

felixarntz (changes requested):
This looks solid in relation to the IB, just one point of feedback there.

However, this now introduces a greater underlying problem around whether or not we need to include the "Others" slice or not. The `hasOthers` variable relies on checking the number of `rows`, however now that some of them may be filtered out because they are "empty" which would mess up that logic as is.

@Hazlitte Please hold off on continuing work on this PR for now though, since there is another point in the issue definition that we should clarify first, see https://github.com/google/site-kit-wp/issues/2865#issuecomment-789226645. If we decide this is still valid as is, we'll need to think about a more appropriate solution though, which will also ensure the need for an "Others" slice is still determined correctly.

felixarntz:
Closing this as the issue has been determined as invalid. Aplogies @Hazlitte that this wasn't noticed earlier in the pipeline.

Issue #2249, opened 2020-10-22
Introduce `ReportTable` component


felixarntz:
The `getDataTableFromData` function is overly complex and should be superseded by a `ReportTable` component.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a new functional component `ReportTable` in `assets/js/components/ReportTable.js`:
    * It should require `props.rows` as an array of objects of row data (in no specific structure as long as it is consistent across all rows)
    * It should require `props.columns` as an array of objects with the following properties:
        * `{string} title`: Column header title.
        * `{string} description`: Optional. Column header description, displayed as tooltip. Replaces today's `tooltip` property.
        * `{bool} primary`: Optional. Whether the column is primary, like today.
        * `{string} className`: Optional. Class name(s) to add to each `th` and `td` of the column
        * `{string} field`: Optional. The path to access the value to render for the column in each row.  
        If provided, the value rendered should be essentially `get( row, field )` via `lodash/get`  
        **Any further customization needed should use `Component` instead**
        * `{WPComponent} Component`: Optional. Component type to use for rendering cells for this column in each row.  
        If provided, `Component` will be rendered with the full `row` object passed as a prop
        **Takes precedence over `field`**
        * Each column must provide either a `field` or a `Component`
    * It should produce similar output like `getDataTableFromData`
    * It should render the header row using the title and description from each column in `columns`
    * It should render each row by iterating over each column definition for each row
        * If `column.Component`, it should render `<Component row={ row } />`
        * If `column.field` is provided and no `Component`, the value referred to by `column.field` should be rendered for that row
* The following components should be updated to rely on `ReportTable` instead of `getDataTableFromData`:
    * `DashboardPopularKeywordsWidget`
    * `DashboardPopularPagesWidget`
    * `DashboardTopEarningPagesWidget`
    * `AcquisitionSources`
    * `WPDashboardPopularPages` (Introduced in https://github.com/google/site-kit-wp/issues/2610)
* Visually, nothing should change about these four components using the new `ReportTable`.

## Implementation Brief
- Using `assets/js/components/`, create a new file `ReportTable.js` which will contain the `ReportTable` functional component with the following props:
	- `rows`: array of objects of row data (in no specific structure as long as it is consistent across all rows)
	- `columns`: array of objects with the following properties
		- `title`: {string} Column header title.
[...]

PR #2892, opened 2021-03-01
Introduce <ReportTable /> and refactor new components to use

aaemnnosttv:
## Summary

Addresses issue #2249

## Relevant technical choices

* Added support for `column.field` in conjunction with `column.Component` as it makes for cleaner use in some cases. 
This was originally defined as only `field` or `Component` could be used. We still need one or the other, but it's also to pass both in which case the value referred to by `field` is passed to the component as `fieldValue`.
* Defined `ReportTable` `columns` outside of the consuming component to avoid unnecessary renders caused by defining the columns object and inner components inline on every render
* Added a `limit` prop which functions similar to the previous `options.cap` – this was mostly needed for VRT since one of the stories provides extra data to the component than it actually requests (report limit now matches the row limit), but I figured it was probably a good idea to include this since it's better to slice the array within the table than in the parent to avoid extra renders or the need to memoize the rows after slicing
* Added support for `column.hideOnMobile` – in the IB, we were going to use `className="hidden-on-mobile"` for this but we have specific styles that require the number of columns shown in mobile to be known. As long as we have the need for hiding columns on mobile, it's probably better for the table component to control this rather than passing the class from consumers.
* Extracted a new `DetailsPermaLinks` component as a fully generic component for rendering a Link to the dashboard-details view for a given title and `path` (URL pathname) which has another smaller link below it to the path itself that opens in a new tab
* Extracted a new `Num` component as a wrapper for `numFmt`
* Extracted a new `Decimal` component as a child of `Num` which has more natural props for dealing with the number of digits it displays (precision)
* Added a `getServiceReportURL` selector to the Search Console datastore for consistency with others and to simplify consuming code for deep links to search analytics reports
    * Updated existing tests in `service.test.js` for consistency

## Checklist


assets/sass/utilities/_overflow-wrap.scss
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+.googlesitekit-plugin {
+	// See https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap
+
+	.googlesitekit-overflow-wrap-break-word {
+		overflow-wrap: break-word;
+	}
```
tofumatt, line 24: Maybe this should be closer-linked to the style in-question?
```suggestion

	// See https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap
	.googlesitekit-overflow-wrap-break-word {
		overflow-wrap: break-word;
	}
```

But also an explanation of why we're using this is probably better if it warrants a comment at all? Otherwise it can probably be removed; if anyone's confused about the attribute itself they can look it up straightforwardly 🙂

tofumatt (changes requested):
I noticed in Storybook the first item in the report is lacking an icon:

Screenshot 2021-03-02 at 16 28 10

I'm guessing that's okay but just checking 🙂 

----

Overall this looks good but I have some thoughts about a few naming things, etc. Let me know what you think 🙂

assets/js/components/DetailsPermaLinks.js
```diff
@@ -0,0 +1,68 @@
...
+				{ title }
+			</Link>
+
+			<Link
+				className={ classnames(
+					'googlesitekit-display-block',
+					'googlesitekit-overflow-wrap-break-word'
+				) }
+				href={ permaLink }
+				target="_blank" // No external styling.
```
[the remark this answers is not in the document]
aaemnnosttv: > Why are we forcing a link to open in a new tab/window here? That limits user choice/interaction capabilities and is generally user-hostile unless there's a real chance of them losing valuable info here by changing the current/in-focus tab.

That's just how the old version of the component that this is replacing behaves.

> Also: what's the comment here mean?

Normally for these kinds of links we use the `external` prop, but this also applies additional styling which these links shouldn't have. So the target is added to avoid the additional side-effect of using `external`.

assets/js/components/Num/Decimal.js
```diff
@@ -0,0 +1,48 @@
+/**
```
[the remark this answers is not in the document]
aaemnnosttv: @tofumatt Can we? I was thinking we couldn't due to `Number` being a core JS object 🤔

assets/js/components/Num/index.js
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Internal dependencies
+ */
+import { numFmt } from '../../util';
+
+export default function Num( {
+	value,
+	shorthand,
+	children, // eslint-disable-line no-unused-vars
```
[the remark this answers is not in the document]
aaemnnosttv: It's destructured to not be included in the following `...options` object from all other props.

assets/js/components/ReportTable.js
```diff
@@ -0,0 +1,128 @@
...
+import PropTypes from 'prop-types';
+
+export default function ReportTable( { rows, columns, className, limit } ) {
+	invariant( Array.isArray( rows ), 'rows must be an array.' );
+	invariant( Array.isArray( columns ), 'columns must be an array.' );
+	columns.forEach( ( { Component, field = null } ) => {
+		invariant( Component || field !== null, 'each column must define a Component and/or a field.' );
+	} );
+	invariant( Number.isInteger( limit ) || limit === undefined, 'limit must be an integer, if provided.' );
+	const mobileColumns = columns.filter( ( col ) => col.hideOnMobile );
```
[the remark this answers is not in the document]
aaemnnosttv: Oh good catch, you're right! I think in this case it ended up working the same because the same number of columns were hidden/visible in mobile.

assets/js/modules/adsense/components/dashboard/DashboardTopEarningPagesWidget.js
```diff
@@ -137,12 +107,45 @@ function DashboardTopEarningPagesWidget( { Widget, WidgetReportZero, WidgetRepor
...
+				/>
 			</TableOverflowContainer>
 		</Widget>
 	);
 }
 
+const tableColumns = [
+	{
+		title: __( 'Top Earning Pages', 'google-site-kit' ),
+		tooltip: __( 'Top Earning Pages', 'google-site-kit' ),
```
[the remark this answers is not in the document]
aaemnnosttv: Hmm, yeah not sure. We should probably update it. @felixarntz ?

aaemnnosttv:
> I noticed in Storybook the first item in the report is lacking an icon

@tofumatt – that's expected because we don't even register Site Verification on the client side at all since it's purely an internal module. It's the only one that doesn't have an icon so far though.

assets/js/modules/adsense/components/dashboard/DashboardTopEarningPagesWidget.js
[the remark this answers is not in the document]
felixarntz: Let's keep this as is for now, I'd say that's out of scope here.

assets/js/components/Num/index.js
```diff
@@ -0,0 +1,50 @@
...
+ * External dependencies
+ */
+import PropTypes from 'prop-types';
+
+/**
+ * Internal dependencies
+ */
+import { numFmt } from '../../util';
+
+export default function Num( {
```
felixarntz, line 29: I think this component as well as the related `Decimal` component are a bit overkill and too low-level for components. It makes sense to create reusable components to create a specific UI, but these two components are purely for formatting a single value, that could easily be achieved by just calling `numFmt`.

Rather than introducing yet another layer around formatting numbers, let's use the function directly.

assets/js/components/wp-dashboard/WPDashboardPopularPages.js
```diff
@@ -77,49 +78,39 @@ const WPDashboardPopularPages = ( { WidgetReportZero, WidgetReportError } ) => {
...
+		Component: ( { row } ) => {
+			const [ title, path ] = row.dimensions;
+			return <DetailsPermaLinks title={ title } path={ path } />;
+		},
+	},
+	{
+		title: __( 'Pageviews', 'google-site-kit' ),
+		description: __( 'Pageviews', 'google-site-kit' ),
+		field: 'metrics.0.values.0',
+		Component: ( { fieldValue } ) => <Decimal value={ fieldValue } />,
```
felixarntz, line 111: See above. We could simply use `Fragment` and `numFmt` here, that would simplify the implementation by reducing that extra abstraction.

assets/js/modules/adsense/components/dashboard/DashboardTopEarningPagesWidget.js
```diff
@@ -137,12 +107,45 @@ function DashboardTopEarningPagesWidget( { Widget, WidgetReportZero, WidgetRepor
...
+	},
+	{
+		title: __( 'Revenue', 'google-site-kit' ),
+		tooltip: __( 'Revenue', 'google-site-kit' ),
+		Component: ( { row } ) => (
+			<Decimal
+				value={ row.metrics[ 0 ].values[ 0 ] }
+				precision={ 2 }
+				fixed
+			/>
```
felixarntz, line 144: See above, let's use `Fragment` and `numFmt` instead.

assets/js/modules/analytics/components/dashboard/AnalyticsAdSenseDashboardWidgetTopPagesTable.js
```diff
@@ -51,75 +52,60 @@ const AnalyticsAdSenseDashboardWidgetTopPagesTable = ( { data } ) => {
...
+		title: __( 'Page RPM', 'google-site-kit' ),
+		description: __( 'Page RPM', 'google-site-kit' ),
+		field: 'metrics.0.values.1',
+		Component: ( { fieldValue } ) => <Decimal value={ fieldValue } precision={ 2 } fixed />,
+	},
+	{
+		title: __( 'Impressions', 'google-site-kit' ),
+		description: __( 'Impressions', 'google-site-kit' ),
+		field: 'metrics.0.values.2',
+		Component: ( { fieldValue } ) => <Decimal value={ fieldValue } />,
```
felixarntz, line 105: See above.

assets/js/modules/analytics/components/dashboard/AnalyticsAdSenseDashboardWidgetTopPagesTable.js
```diff
@@ -51,75 +52,60 @@ const AnalyticsAdSenseDashboardWidgetTopPagesTable = ( { data } ) => {
...
+					inherit
+				>
+					{ title }
+				</Link>
+			);
+		},
+	},
+	{
+		title: __( 'Earnings', 'google-site-kit' ),
+		description: __( 'Earnings', 'google-site-kit' ),
```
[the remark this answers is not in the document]
felixarntz: See above, let's keep this as is for now. @tofumatt Can you open a separate issue for us to iterate/revisit these duplicate tooltips and descriptions?

assets/js/modules/analytics/components/dashboard/AnalyticsDashboardWidgetTopPagesTable.js
```diff
@@ -59,80 +58,72 @@ const AnalyticsDashboardWidgetTopPagesTable = ( props ) => {
...
+		hideOnMobile: true,
+		field: 'metrics.0.values.1',
+		Component: ( { fieldValue } ) => <Decimal value={ fieldValue } />,
+	},
+	{
+		title: __( 'Bounce Rate', 'google-site-kit' ),
+		description: __( 'Bounce Rate', 'google-site-kit' ),
+		hideOnMobile: true,
+		field: 'metrics.0.values.2',
+		Component: ( { fieldValue } ) => <Num value={ Number( fieldValue ) / 100 } shorthand="%" />,
```
felixarntz, line 123: See above.

assets/js/modules/analytics/components/dashboard/DashboardPopularPagesWidget.js
```diff
@@ -137,10 +107,29 @@ function DashboardPopularPagesWidget( { Widget, WidgetReportZero, WidgetReportEr
...
+		primary: true,
+		Component: ( { row } ) => {
+			const [ title, path ] = row.dimensions;
+			return <DetailsPermaLinks title={ title } path={ path } />;
+		},
+	},
+	{
+		title: __( 'Views', 'google-site-kit' ),
+		field: 'metrics.0.values.0',
+		Component: ( { fieldValue } ) => <Decimal value={ fieldValue } />,
```
felixarntz, line 131: See above.

assets/js/modules/search-console/components/dashboard/DashboardPopularKeywordsWidget.js
```diff
@@ -142,10 +83,53 @@ function DashboardPopularKeywordsWidget( { Widget, WidgetReportZero, WidgetRepor
...
+	},
+	{
+		title: __( 'Clicks', 'google-site-kit' ),
+		description: __( 'Number of times users clicked on your content in search results', 'google-site-kit' ),
+		Component: ( { row } ) => <Decimal value={ row.clicks } />,
+	},
+	{
+		title: __( 'Impressions', 'google-site-kit' ),
+		description: __( 'Counted each time your content appears in search results', 'google-site-kit' ),
+		Component: ( { row } ) => <Decimal value={ row.impressions } />,
```
felixarntz, line 131: See above.

assets/js/components/DetailsPermaLinks.js
```diff
@@ -0,0 +1,68 @@
...
+		<Fragment>
+			<Link
+				className="googlesitekit-display-block"
+				href={ detailsURL }
+				inherit
+			>
+				{ title }
+			</Link>
+
+			<Link
```
felixarntz, line 54: There appears to be a problem here; these two links can currently appear in the same row (e.g. in the WP dashboard widget) which would be a regression and looks broken.

assets/js/modules/search-console/components/dashboard/DashboardPopularKeywordsWidget.js
```diff
@@ -20,72 +20,48 @@
...
+		limit: 10,
+		url: url || undefined,
+	};
+	const data = useSelect( ( select ) => select( STORE_NAME ).getReport( reportArgs ) );
+	const error = useSelect( ( select ) => select( STORE_NAME ).getErrorForSelector( 'getReport', [ reportArgs ] ) );
+	const loading = useSelect( ( select ) => ! select( STORE_NAME ).hasFinishedResolution( 'getReport', [ reportArgs ] ) );
+	const baseServiceURL = useSelect( ( select ) => select( STORE_NAME ).getServiceReportURL( {
+		...generateDateRangeArgs( dateRangeDates ),
+		page: url ? `!${ url }` : undefined,
+	} ) );
```
felixarntz, line 59: There's a regression here, see above: For a domain property this should, if no entity URL exists, still ensure that the site URL is included as `page` query parameter, to filter it down as needed; in the new code the `page` query parameter is instead omitted.

Was there a need to implement a `getServiceReportURL` selector here? It's probably a good idea to have it, but feels a bit out of scope in this already large PR. I guess since this is the only usage here, we should be okay keeping it in.

felixarntz (changes requested):
The overall implementation looks solid, however I think the abstraction with having components just to format numbers goes a bit too far. Also I've found two regressions with the usage of the new component.

ryihan (changes requested):
Problem ...

assets/js/components/Num/index.js
aaemnnosttv: To me the component/JSX form is easier to read and a bit more natural semantically:

image

I'll update it to use `numFmt` though, not a problem.

assets/js/components/wp-dashboard/WPDashboardPopularPages.js
aaemnnosttv: `Fragment` isn't necessary here since we're returning a single root value/node.

assets/js/modules/search-console/components/dashboard/DashboardPopularKeywordsWidget.js
aaemnnosttv: > Was there a need to implement a `getServiceReportURL` selector here? It's probably a good idea to have it, but feels a bit out of scope in this already large PR.

It wasn't strictly required, no, but it helped reduce a substantial amount of duplication as well as handle the domain property logic which is something that individual components shouldn't need to be concerned with. The effort was fairly minimal so I think it was a worthwhile enhancement to include.

assets/js/components/DetailsPermaLinks.js
aaemnnosttv: This was due to missing styles since wpdashboard has a separate stylesheet which I've updated.

Also, related to this was that I noticed the widget app's render target did not have the `.googlesitekit-plugin` class which was preventing some styles from being applied as well, so I added this too. This resulted in a slight difference in the font rendering in the widget (since core uses a different font by default) but it seems more likely that this was missed rather than an intentional omission (let me know if that's not a correct assumption).

felixarntz (approved):
Excellent!

ryihan (commented):
Done

Issue #2747, opened 2021-02-02
Search Console x-axis labels truncated on 90 day charts


aaemnnosttv:
## Bug Description

Currently our charts used on most overview pages show dates below the chart along the x-axis. This works fine in most cases, and the chart will dynamically show less labels as the width of the viewport gets narrower. However, even on a large display, the 90-day range results in some of the labels being individually truncated with what seems to be a `text-overflow: ellipsis;` style.

![image](https://user-images.githubusercontent.com/1621608/106648510-0b5d5600-6599-11eb-9a32-8ed2980ce81f.png)

## Steps to reproduce

1. Go to the Search Console dashboard page with a site that has at least 90 days of data
2. Select the 90 day date range
3. See squished labels

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* On the Search Console module page, when selecting 90 days, the dates on the x-axis of the graph should not be truncated. Less dates should be displayed as applicable to make it appear in a better way.

## Implementation Brief

* In `modules/search-console/components/dashboard/SearchConsoleDashboardWidgetSiteStats.js`, we need to ensure that  the following rule is added to the chart's `options`:
```javascript
// This small reduction in height gives the x-axis labels the room they need
chartArea.height = '77%'; 
```

### Test Coverage

* No new tests need to be added; existing ones should all pass.

### Visual Regression Changes

* This should introduce no new regressions.

## QA Brief

* Go to the Search Console module page
* Select the 'last 90 days' date range
* Investigate the chart at various screen widths: none of the horizontal axis labels should truncate 

## Changelog entry

* Fix Search Console x-axis labels on 90 day charts

tofumatt:
The ACs here mention reducing the number of labels that appear rather than making them all fit, but I think I prefer your solution as it's less-complex and lets the chart library manage things for us.

But I wonder why we'd bother to use anything but this solution for all sizes? I tried it and it seems/looks fine, so rather than change the style based on the number of days shown (which seems a bit odd to me and might strike someone as unexpected), let's just use those options as the default options for all date range views for this component, if others are cool with that.

johnPhillips:
@aaemnnosttv Do you have any thoughts on this? 

[...]

PR #2893, opened 2021-03-01
Search Console x-axis labels truncated on 90 day charts #2747

johnPhillips:
## Summary

Addresses issue #2747

## Checklist


aaemnnosttv (approved):
LGTM 😄  Thanks @johnPhillips !

aaemnnosttv (changes requested):
Oops, looks like this requires VRT reference images to be updated. 
image

Would you please review the failures and approve the changes if they're all expected?

johnPhillips:
@aaemnnosttv mostly the VRT changes are just down to the new height on the chart, which is barely perceptible.

However, the one you shared is a change from slanted labels to stacked labels, which is happening when the 28-day range is selected on medium-large screens up to about 1400px. I'm guessing this is due to the library using the extra space to fit in some more labels. I did check for regressions like this before, so I am slightly confused as to why it's now happening. Maybe the charts library is a bit of a moving target in this regard.

| Stacked  | Slanted |
| ------------- | ------------- |
| Screenshot 2021-03-02 at 14 46 43 | Screenshot 2021-03-02 at 14 43 21 |

I've approved them because I think this is a natural side-effect of applying this change to all date ranges, rather than just the 90-day range, which is what we agreed in the IB. We also agreed that it might be best to let the library manage the labels its own way as long as the truncation was solved.

However, very happy to go back to the drawing board if you don't think this is an acceptable regression. Let me know what you think?

aaemnnosttv (approved):

Issue #2848, opened 2021-02-22
Changing question doesn't scroll new question into viewport


tofumatt:
## Bug Description

On the User Input screen, when the Back/Next buttons are clicked the viewport doesn't change. This is a problem on smaller screens (eg mobile) where advancing to the next question causes it to appear "above the fold". We should scroll the viewport to the new question whenever we change the "active" question.

It's worth noting this happens—at least in part—because we don't use a dedicated router (see #2514). I think #2514 would fix this issue as React Router is good about scrolling new content from a route change when it isn't visible. And there are more predictable ways to get the content to scroll into view with a router.

So it might be best to wait on this until #2514 is fixed, to see if it covers the issue.

## Steps to reproduce

1. Go to the User Input questionnaire on a new site where the answers have not been submitted.
2. Using a small/mobile viewport, scroll to the bottom of the first question, select an answer, then click "Next".
3. Notice the question is outside the viewport and the screen needs to be scrolled up to read it.

## Screenshots

![2021-02-22 22 48 33](https://user-images.githubusercontent.com/90871/108780429-54814400-7560-11eb-98d7-047bb0dfd1e5.gif)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The next question should clearly appear after the "Next" button is clicked in the user input screen.
* The viewport should be scrolled so the _beginning_ of the new question component is in view, not the end of the component.

## Implementation Brief

* Add <code>id={ \`googlesitekit-user-input-question-${ questionNumber }\` }</code> to the top-most `div` in `assets/js/components/user-input/UserInputQuestionWrapper.js`
* Add a `useEffect()` hook in `UserInputQuestionnaire` that scrolls to the above `googlesitekit-user-input-question-${ questionNumber }` (eg `googlesitekit-user-input-question-${ activeSlugIndex + 1 }`) element after the first render when `activeSlugIndex` changes.
* Merge #2912

### Test Coverage

* No change

### Visual Regression Changes

* N/A

## QA Brief

* Follow the steps to reproduce above, but note that the question is fully visible after being scrolled into view when clicking "Next" or "Back".

## Changelog entry

* Fix bug where switching between questions in the user input flow did not scroll the question into view.

aaemnnosttv:
[...]

PR #2912, opened 2021-03-03
Scroll to the previous/next question in UserInputApp.

tofumatt:
## Summary

Addresses issue #2848.

## Checklist


assets/js/components/user-input/UserInputQuestionnaire.js
```diff
@@ -125,6 +126,15 @@ export default function UserInputQuestionnaire() {
...
 	}, [ activeSlugIndex ] );
 
+	useEffect( () => {
+		if ( firstRender ) {
+			setFirstRender( false );
+			return;
+		}
+
+		global.document.getElementById( `googlesitekit-user-input-question-${ activeSlugIndex + 1 }` )?.scrollIntoView( { behavior: 'smooth' } );
+	}, [ activeSlugIndex ] );
```
aaemnnosttv, line 136: `firstRender|shouldScrollToActiveQuestion` should probably be in the dependency list here as well.

assets/js/components/user-input/UserInputQuestionnaire.js
```diff
@@ -125,6 +126,15 @@ export default function UserInputQuestionnaire() {
 		setActiveSlug( steps[ steps.length - 1 ] );
 	}, [ activeSlugIndex ] );
 
+	useEffect( () => {
+		if ( firstRender ) {
+			setFirstRender( false );
+			return;
+		}
+
+		global.document.getElementById( `googlesitekit-user-input-question-${ activeSlugIndex + 1 }` )?.scrollIntoView( { behavior: 'smooth' } );
```
aaemnnosttv, line 135: Shouldn't this be simply `activeSlugIndex` rather than +1 since the effect is already triggered by this state changing?

assets/js/components/user-input/UserInputQuestionnaire.js
```diff
@@ -125,6 +126,15 @@ export default function UserInputQuestionnaire() {
 		setActiveSlug( steps[ steps.length - 1 ] );
 	}, [ activeSlugIndex ] );
 
+	useEffect( () => {
+		if ( firstRender ) {
+			setFirstRender( false );
+			return;
+		}
```
aaemnnosttv, line 133: Alternatively, we could maybe just move the effect here to its cleanup function so that it's only invoked on the "backside" which would also skip the initial run like we want. As long as that didn't trigger an unwanted scroll at the very end, I would think this might be the better way to go. What do you think?

aaemnnosttv (commented):
Thanks @tofumatt – assuming this is ready for review since this is still marked as a draft but left a few comments/suggestions below.

assets/js/components/user-input/UserInputQuestionnaire.js
tofumatt: Ah, the indexes are mismatched in their base; `activeSlugIndex` references a zero-based index:

Screenshot 2021-03-18 at 17 15 09

but the question IDs start at "1":
Screenshot 2021-03-18 at 17 12 25

So this is expected but you're right: it's super-unclear. I've added a comment to clarify it 🙂

assets/js/components/user-input/UserInputQuestionnaire.js
tofumatt: I think that's a bit more confusing—it's clever, but it's not straightforward.

As a rule I'd expect to look in a `useEffect`'s main function for its intended effects, and its returned function for the cleanup. In that case, we're be running the intended effect in the cleanup. It would work, but I think it's a bit too clever for its own good 😅

assets/js/components/user-input/UserInputQuestionnaire.js
tofumatt: Adding `shouldScrollToActiveQuestion` will mean that when we call `setShouldScrollToActiveQuestion( true )` in the effect this will run again and then scroll to the first question on page load.

The existing code works as intended and doesn't warn about dependency issues… I think it's fine 🤷🏻‍♂️

aaemnnosttv (approved):
Super!

Issue #2837, opened 2021-02-22
User Input preview display quirks


aaemnnosttv:
## Bug Description

The preview for user input has a few quirks which were recently discovered which should be addressed before launch.

1. Long freeform answers can overflow horizontally  
![image](https://user-images.githubusercontent.com/1621608/108726706-677c2000-7530-11eb-8a03-2beab2b10909.png)
2. On mobile, the preview pill/bubble radius looks a odd/inconsistent with multi-line answers
![image](https://user-images.githubusercontent.com/1621608/108726907-9eeacc80-7530-11eb-9504-f242d926f5b5.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The answer bubbles in the User Input flow preview should have a consistent border radius, no matter the number of lines, width, height etc.
* Long content in a custom answer should not be able to mess up the layout; in other words, the bubbles should should always be limited to the maximum available width, and the inner content should wrap as possible and in no case overflow.
* Custom answers should be trimmed (i.e. spaces at the beginning and end should be cut off).
* Any answers provided under "Other" should be prefixed with "Other:" in their bubble, and if nothing is entered they should just say "Other". In other words, the bubble content should be like "Other: something custom I entered" or "Other".

## Implementation Brief
* Using `/assets/sass/components/user-input/googlesitekit-user-input-preview.scss`,
	* Update the `.googlesitekit-user-input__preview-answer` class to:
		* Add `word-break: break-word;` to prevent characters from overflowing.
		* Set `border-radius` to `30px` on small screens only. The border radius should be `100px` for tablet and above.
* Using `/assets/js/components/user-input/UserInputPreviewGroup.js`,
	* Create a new constant, `isOther`, which checks whether `values` can be found in `options` (example can be found in `/assets/js/components/user-input/UserInputSelectOptions.js`) and the question number is not `5`.
	* Before rendering the `value`, check if `isOther` is `true`. If this is the case, display the `Other:` text before the `value`
* Using `/assets/js/googlesitekit/datastore/user/user-input-settings.js`,
	* Update the `saveUserInputSettings` action to trim values within the `settingssettings[ key ]?.values` variable.

### Test Coverage
* Add an additional test for the `saveUserInputSettings` action which checks whether user input values are saved without spaces at the beginning and at the end.

### Visual Regression Changes
* N/A

## QA Brief

* Go to the user input settings flow and select answers providing long custom answers in `other` fields;
[...]

PR #2914, opened 2021-03-03
Bug/2837 user input preview quirks

eugene-manuilov:
## Summary

Addresses issue #2837 

## Checklist


aaemnnosttv (approved):
Looks good to me!

Theres one detail in the ACs that may not be consistent with what we want now (allowing empty value for Other) so I'll leave this for Felix to confirm.

assets/sass/components/user-input/googlesitekit-user-input-preview.scss
```diff
@@ -54,15 +54,20 @@
...
 		margin-bottom: 12px;
 		margin-right: 12px;
 		padding: 8px 13px;
+		word-break: break-word;
 
 		@include responsive-text("body1", "body3");
+
+		@media (min-width: $bp-tablet) {
+			border-radius: 100px;
+		}
```
felixarntz, line 70: With a multiline answer for example, this still looks weird. How about always using `30px`, even on wider viewports? IMO this always looks better than `100px` - for one-line content, there's no visual difference anyway (since the height of the bubble then isn't even close to 100 pixels), for multi-line content, using 100 pixels makes the border way too round IMO and leaves little visual spacing around the words.

cc @asvinb Please double-check, since you wrote the IB.

felixarntz (commented):
Technically LGTM, but the border radius still looks odd to me and IMO could be improved (cc @asvinb).

**100px border radius on large screens** (looks odd IMO)
Screen Shot 2021-03-08 at 12 10 33 PM

**30px border radius on large screens (i.e. no override compared to small screens)** (better)
Screen Shot 2021-03-08 at 12 10 45 PM

assets/sass/components/user-input/googlesitekit-user-input-preview.scss
eugene-manuilov: Ok, updated. Now it should always be 30px.

felixarntz (approved):
Great!

Issue #2887, opened 2021-02-26
Migrate all components to use GoogleChartV2 component instead of GoogleChart


felixarntz:
#2714 introduced a new `GoogleChartV2` component and started using it, mostly in the All Traffic widget. However, the old `GoogleChart` component is still in use, and it has previously caused conflicts. While we've resolved those, we should sooner than later completely phase out usage of the old component (and its way of loading the charts library via PHP).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All JS code that displays charts should use `GoogleChartV2` instead of `GoogleChart`.
* The `GoogleChart` component, as well as any related side effects logic (e.g. enqueuing the Google charts library script in PHP) or CSS code only used for `GoogleChart` should be removed.
* After the removal, `GoogleChartV2` should be renamed to `GoogleChart` so that it is the new de-facto standard for displaying charts. Any similar references (e.g. CSS classes including the `v2`) should also be renamed accordingly.

## Implementation Brief

* Rename `GoogleChartV2` to `GoogleChart`, along with imports and SCSS/class names.
* Remove the old `GoogleChart` component, replacing it with `GoogleChartV2`.
* Add a new `selectedColumns` prop that filters the data supplied to the underlying `Chart` component from `react-google-charts` to be only the relevant/selected columns. See implementation here: https://github.com/google/site-kit-wp/blob/671161504d66c3cc514770daec462829ef445ad8/assets/js/components/GoogleChart.js#L56-L65
* Find all instances of `GoogleChart` and update the arguments to use `chartType="${type}Chart"` instead of `chartType="$type"`, and update `selectedStats` to the new `selectedColumns` prop.
* Merge #2916.

### Test Coverage

* N/A.

### Visual Regression Changes

* Loading states for charts will change as the new `GoogleChart` uses `PreviewBlock` instead of `ProgressBar` for loading.

## QA Brief

* Test all pages that have charts (essentially every page aside from settings pages) and ensure the charts still render when data is present. Test using all the widget feature flags and with no feature flags enabled.
* All charts should continue to function as normal.

## Changelog entry

* Migrate all components to use the new version of the GoogleChart component

felixarntz:
IB ✅

PR #2916, opened 2021-03-04
Migrate all components to use new GoogleChartV2 component.

tofumatt:
## Summary

Addresses issue #2887.

## Relevant technical choices

Renamed the component `GoogleChart` as the `V2` moniker is no longer needed.

## Checklist


assets/js/modules/analytics/components/dashboard/LegacyDashboardAcquisitionPieChart.js
```diff
@@ -90,11 +90,11 @@ const LegacyDashboardAcquisitionPieChart = ( { data, source } ) => {
 	return (
 		<div className="googlesitekit-chart googlesitekit-chart--pie">
 			<GoogleChart
+				chartType="PieChart"
 				data={ processedData }
+				loadingHeight={ 205 }
+				loadingWidth={ 205 }
```
eugene-manuilov, line 96: These should be strings:

```suggestion
				loadingHeight="205px"
				loadingWidth="205px"
```

Otherwise it throws errors:
Screenshot from 2021-04-23 16-58-01

assets/js/components/MiniChart.js
```diff
@@ -72,14 +72,10 @@ function MiniChart( { index, change } ) {
 	return (
 		<div className="googlesitekit-mini-chart">
 			<GoogleChart
+				chartType="PieChart"
 				data={ data }
+				loadingHeight={ 24 }
```
eugene-manuilov, line 77: Same here, it should be a string:

```suggestion
				loadingHeight="24px"
```

assets/js/components/GoogleChart.js
```diff
@@ -19,205 +19,186 @@
...
 /**
  * Internal dependencies
  */
-import ProgressBar from './ProgressBar';
+import PreviewBlock from './PreviewBlock';
 
 export default function GoogleChart( props ) {
 	const {
-		chartID,
+		chartEvents,
```
eugene-manuilov, line 37: @tofumatt we also need to add prop types for this component as well.

assets/js/components/GoogleChart.js
```diff
@@ -19,205 +19,186 @@
...
-					GoogleChart.charts.set( chartID, chartData );
-				}
-			} else {
-				googleChart = GoogleChart.charts.get( chartID ).chart;
-			}
+	// Ensure we don't filter out columns that aren't data, but are things like
+	// tooltips or other content.
+	const nonDataColumns = data?.length ? data[ 0 ].reduce( ( acc, row, rowIndex ) => {
+		return row?.role ? [ ...acc, rowIndex ] : acc;
+	}, [] ) : [];
```
eugene-manuilov, line 60: Too many question marks and colons here. Could you, please, re-format it to make it possible to ready? :smile: Maybe something like this will do it better:

```suggestion
	const nonDataColumns = data?.length 
		? data[ 0 ].reduce( ( acc, row, rowIndex ) => row?.role ? [ ...acc, rowIndex ] : acc }, [] )
		: [];
```

assets/js/modules/analytics/components/dashboard/LegacyDashboardAcquisitionPieChart.js
```diff
@@ -90,11 +90,11 @@ const LegacyDashboardAcquisitionPieChart = ( { data, source } ) => {
 	return (
 		<div className="googlesitekit-chart googlesitekit-chart--pie">
 			<GoogleChart
+				chartType="PieChart"
```
eugene-manuilov, line 93: Another observation for this chart is that the new version cuts highlight curves:

Screenshot from 2021-04-23 17-21-22

assets/js/modules/analytics/components/module/ModuleOverviewWidget/SiteStats.js
```diff
@@ -65,7 +65,7 @@ export default function SiteStats( { selectedStat, report } ) {
 			<Row>
 				<Cell size={ 12 }>
 					<GoogleChart
-						chartType="line"
+						chartType="LineChart"
```
eugene-manuilov, line 68: Seems like we need to add more props to this chart because its loading state doesn't look right:
Screenshot from 2021-04-23 17-35-11

Plus, it looks like this chart stopped showing previous data:
Screenshot from 2021-04-23 17-35-22

assets/js/modules/adsense/components/dashboard/LegacyAdSenseDashboardWidgetSiteStats.js
```diff
@@ -161,8 +161,7 @@ export default function LegacyAdSenseDashboardWidgetSiteStats( props ) {
 			<Row>
 				<Cell size={ 12 }>
 					<GoogleChart
-						chartType="line"
-						selectedStats={ [ selectedStats ] }
+						chartType="LineChart"
```
eugene-manuilov, line 164: Same here, this chart stopped showing the previous period. See it here: https://google.github.io/site-kit-wp/storybook/pull/2916/?path=/story/adsense-module-components-module-overview-widget--loaded

Here is how it looks in the develop branch: https://google.github.io/site-kit-wp/storybook/develop/?path=/story/adsense-module-components-module-overview-widget--loaded

assets/js/modules/search-console/components/module/ModuleOverviewWidget/Stats.js
```diff
@@ -122,7 +122,7 @@ const Stats = ( { data, metrics, selectedStats } ) => {
 			<Row>
 				<Cell size={ 12 }>
 					<GoogleChart
-						chartType="line"
+						chartType="LineChart"
```
eugene-manuilov, line 125: Same here, before: https://google.github.io/site-kit-wp/storybook/develop/?path=/story/search-console-module-components-module-page-overview-widget--loaded

after: https://google.github.io/site-kit-wp/storybook/pull/2916/?path=/story/search-console-module-components-module-page-overview-widget--loaded

eugene-manuilov (changes requested):
Thanks, @tofumatt! This looks like a good start, but updated charts work slightly differently. We need to fix it and to make sure that the existing behavior remains the same.

assets/js/modules/analytics/components/dashboard/LegacyDashboardAcquisitionPieChart.js
tofumatt: Whoops, thanks! 👍🏻

assets/js/components/GoogleChart.js
tofumatt: The `chartEvents` should be part of the propTypes from the existing GoogleChart that we pass through. If not I'll add them though 👍🏻

eugene-manuilov: @tofumatt the new GoogleChart component does not have propTypes at all. We need to add it.

tofumatt: Oh no, I see. My editor had been doing weird removals of prop types sometimes and I don't know why. It seems they were removed here even though I thought I added them. I'll add them back! 😓

danielgent: @tofumatt I had a similar experience (on a PR that @eugene-manuilov reviewed and luckily he spotted the problem!)

I notice that our linting rules are very aggressive with removing unreachable code.  As opposed to just flagging it, it also autofixes it by deleting (which throws me off while I'm developing! :smiley: ).  I'm not sure if it's related to this :thinking:

tofumatt: @danielgent Yup, I think that's it. I usually catch it, but it's easy to miss the proptypes being removed at the bottom of the page. 🤷🏻

assets/js/modules/analytics/components/dashboard/LegacyDashboardAcquisitionPieChart.js
tofumatt: Odd, I'm not able to recreate that:

tofumatt: I can't reproduce that bug locally; which browser are you using to cause that bug?

Screenshot 2021-04-29 at 15 41 25

tofumatt:
Thanks for all of the widget components you pointed out had differences—I added VRTs for those components because I didn't want them to fail unnoticed again 🙂

eugene-manuilov (changes requested):
Thanks, @tofumatt. It looks much better and thank you for adding those charts to VR tests. Found a few more issues to address. Please, see my comments below.

assets/js/components/Sparkline.js
```diff
@@ -75,7 +75,7 @@ function Sparkline( {
 
 	return (
 		<div className="googlesitekit-analytics-sparkline-chart-wrap">
-			<GoogleChartV2
+			<GoogleChart
```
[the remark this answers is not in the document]
tofumatt: Maybe that should be done by the chart component itself—if only one is set, the other one gets set to `100%`? I'll try that first so we don't have to add boilerplate args to all the components, but if not I'll go through each one 👍🏻

assets/js/modules/analytics/components/dashboard/LegacyDashboardAcquisitionPieChart.js
tofumatt: Ah, I do see it now. Weird that it only happens at large viewports. I'll see what's causing it 👍🏻

tofumatt: Ah, I caught it, we had a CSS fix for it in the old charts code that I used for the new one. Nice catch 👍🏻

tofumatt:
All feedback addressed and fixed, thanks so much for catching all of that stuff! 👍🏻

I think this is good-to-go now 😅

tofumatt:
Sorry, just taking a look at these VRT failures; I didn't see any locally so not sure why that happened…

tofumatt:
Ah, seems it was an Actions bug, ready for review 👍🏻

eugene-manuilov (changes requested):
Thanks, @tofumatt. This is looking really great now. There are a few tiny issues that we need to address and this should be good to go. Could you please take a look at it?

assets/js/components/GoogleChart.js
```diff
@@ -20,205 +20,251 @@
...
+		'Sankey',
+		'ScatterChart',
+		'SteppedAreaChart',
+		'Table',
+		'Timeline',
+		'TreeMap',
+		'WaterfallChart',
+		'WordTree',
+	] ).isRequired,
+	data: PropTypes.array,
```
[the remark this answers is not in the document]
tofumatt: Actually, I tried that and got errors, because sometimes `data` is actually an array of arrays of objects. But:

```js
PropTypes.arrayOf(
	PropTypes.oneOf(
		PropTypes.arrayOf( PropTypes.object ),
		PropTypes.object,
	),
)
```

works, so I've added that.

assets/js/components/GoogleChart.js
```diff
@@ -20,205 +20,251 @@
...
+		'PieChart',
+		'Sankey',
+		'ScatterChart',
+		'SteppedAreaChart',
+		'Table',
+		'Timeline',
+		'TreeMap',
+		'WaterfallChart',
+		'WordTree',
+	] ).isRequired,
```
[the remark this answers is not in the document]
tofumatt: My only worry there is not allowing valid types might be confusing for future developers who will think our charts can't/shouldn't use those types…

But I'll update it and add a comment that lets them know more can be added, that seems fine… in case we want to add any special handling for other charts before allowing their types, like we do with loaders here.

assets/js/components/GoogleChart.js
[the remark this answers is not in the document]
tofumatt: Actually the contents are really varied and the size of the arrays varies too. I don't think there's much benefit to doing anything beyond `array` here:

Screenshot 2021-05-05 at 18 10 44

tofumatt:
Ah, so sorry! I meant to get those proptypes and just blanked on it. Should be good now 🙂

assets/js/components/GoogleChart.js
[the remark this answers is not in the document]
eugene-manuilov: :facepalm: you are right. For some reason, I thought that we pass an array of objects received from the `getReport` selector. Looks good to me.

eugene-manuilov (approved):
:tada:

Issue #2067, opened 2020-09-18
Migrate Analytics page "Audience overview" widget to use Widget API


felixarntz:
The "Audience overview for the last x days" widget on the Analytics module page should be refactored as a new component that is registered and rendered using the Widget API.

The existing component should be untouched and only be renamed, giving it a `Legacy` prefix.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new component `ModuleOverviewWidget` should be added in `assets/js/modules/analytics/components/module/ModuleOverviewWidget.js`.
* It should be implemented as a functional, hook-based, datastore-driven widget component that renders the existing "Audience overview for the last x days" block on the Analytics module page, in a refactored way.
    * In other words, it should replicate what the legacy `AnalyticsDashboardWidgetOverview` and `AnalyticsDashboardWidgetSiteStats` components display.
    * Sub-components can be implemented as necessary. If reusable, they should go into `assets/js/modules/analytics/components/common`. Otherwise (if specific to just this widget), they should go into a new `assets/js/modules/analytics/components/module/ModuleOverviewWidget` directory; in this case, the main widget component should go into the `index.js` file within that directory.
* The new component should be used to register a new widget in `assets/js/modules/analytics/index.js` (for testing, requires to enable the `widgets.moduleScreens` feature flag):
    * slug: `analyticsModuleOverview`
    * width: `WIDGET_WIDTHS.FULL`
    * priority: `1`
    * wrapWidget: `false` (since it needs to render `Widget` manually to add the `Header` prop
    * widget area: `AREA_MODULE_ANALYTICS_MAIN`
* `AnalyticsDashboardWidgetOverview` and `AnalyticsDashboardWidgetSiteStats` should be renamed to have a `Legacy*` prefix added.

Hint: For the implementation it might be useful to look at `AdSenseDashboardWidgetOverview` and `AdSenseDashboardWidgetSiteStats` which have implemented a somewhat similar functionality. These components are not widget components, but they are already using the datastore, so for that parts they can be relevant inspiration.

## Implementation Brief

* Implement requirements defined in the AC

### Test Coverage

* N/A

### Visual Regression Changes

* N/A

## QA Brief

* Set up sitekit and tester plugins on your local;
* Activate and connect the Analytics module;
* Enable the `widgets.moduleScreens` feature in the tester plugin settings;
[...]

PR #2918, opened 2021-03-04
Enhancement/2067 audience overview

eugene-manuilov:
## Summary

Addresses issue #2067

## Checklist


assets/js/modules/analytics/components/module/ModuleOverviewWidget/Header.js
```diff
@@ -0,0 +1,58 @@
...
+import { CORE_USER } from '../../../../../googlesitekit/datastore/user/constants';
+import { getCurrentDateRangeDayCount } from '../../../../../util/date-range';
+import { MODULES_ANALYTICS } from '../../../datastore/constants';
+import LayoutHeader from '../../../../../components/layout/LayoutHeader';
+const { useSelect } = Data;
+
+export default function Header() {
+	const visitorsOverview = useSelect( ( select ) => select( MODULES_ANALYTICS ).getServiceReportURL( 'visitors-overview' ) );
+	const dateRange = useSelect( ( select ) => select( CORE_USER ).getDateRange() );
+	const currentDayCount = getCurrentDateRangeDayCount( dateRange );
```
felixarntz, line 37: This can use the `getDateRangeNumberOfDays` selector instead.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/Header.js
```diff
@@ -0,0 +1,58 @@
...
+		__( 'See full stats in %s', 'google-site-kit' ),
+		_x( 'Analytics', 'Service name', 'google-site-kit' )
+	);
+
+	return (
+		<LayoutHeader
+			title={ title }
+			ctaLabel={ headerCTALabel }
+			ctaLink={ visitorsOverview }
+		/>
```
felixarntz, line 56: This shouldn't use `LayoutHeader` which is a legacy component and mostly intended for use within `Layout`. The `Header` prop of the `Widget` component already adds necessary padding around this, and we shouldn't need to add another similar layer within it. Let's use a custom class or update the CSS rules for `.googlesitekit-widget__header` if it's something that generally makes sense.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/Overview.js
```diff
@@ -0,0 +1,128 @@
...
+		);
+	}
+
+	const getDatapointAndChange = ( [ report ], divider = 1 ) => ( {
+		datapoint: report?.data?.totals?.[0]?.values?.[0] / divider,
+		change: calculateChange(
+			report?.data?.totals?.[1]?.values?.[0],
+			report?.data?.totals?.[0]?.values?.[0],
+		),
+	} );
```
felixarntz, line 60: Maybe move out of the component since it doesn't have to be inside of it regarding dependencies?

assets/js/modules/analytics/components/module/ModuleOverviewWidget/SiteStats.js
```diff
@@ -0,0 +1,160 @@
...
+export default function SiteStats( { loaded, selectedStat, report } ) {
+	const dateRange = useSelect( ( select ) => select( CORE_USER ).getDateRange() );
+
+	if ( ! loaded ) {
+		return (
+			<PreviewBlock width="100%" height="270px" />
+		);
+	}
+
+	const currentDayCount = getCurrentDateRangeDayCount( dateRange );
```
felixarntz, line 45: See above, there is a dedicated selector for this.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/SiteStats.js
```diff
@@ -0,0 +1,160 @@
...
+	const dataMap = extractAnalyticsDashboardData( report, {
+		selectedStats: selectedStat,
+		selectedDataIndex: 0,
+		currentMonthMetricIndex: 0,
+		previousMonthMetricIndex: 1,
+		days: currentDayCount,
+	} );
```
felixarntz, line 52: Instead of using and modifying this legacy function, let's come up with a more flexible (or at least more understandable) way to do that. Like in the other widget components implemented for the Site Kit dashboard, we shouldn't use old functions that serve a super-specific purpose. Let's introduce some less specific utility functions instead, and the parts that need to be specific to the `report` used here should be kept within this component.

Essentially, `extractAnalyticsDashboardData` can be broken down into multiple, parameter-controlled functions where possible. The new component should then use a mix of those flexible functions while hard-coding e.g. `dataLabels` in the component since it's specific to that.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/index.js
```diff
@@ -0,0 +1,111 @@
...
+				return {
+					...acc,
+					[ metric ]: select( MODULES_ANALYTICS ).getReport( args ),
+					loaded: acc.loaded && select( MODULES_ANALYTICS ).hasFinishedResolution( 'getReport', [ args ] ),
+					error: acc.error || select( MODULES_ANALYTICS ).getErrorForSelector( 'getReport', [ args ] ),
+				};
+			},
+			{ loaded: true },
+		);
+	} );
```
felixarntz, line 81: This should be queried in a single report. The dimensions and dates etc that we need are the same for every metric, so we can just query it all at once.

The overall goal is to keep reports specific to the widget they are used in. We shouldn't combine reports between widgets for less bandwidth, but a single widget should ensure its report is queried in an optimal way. Instead of having 4 API queries, this can become a single one.

assets/js/modules/analytics/util/index.js
```diff
@@ -156,23 +157,29 @@ function reduceAnalyticsRowsData( rows, selectedStats ) {
...
  *
- * @param {Object} reports       The data returned from the Analytics API call.
- * @param {Array}  selectedStats The currently selected stat we need to return data for.
- * @param {number} days          The number of days to extract data for. Pads empty data days.
+ * @param {Object} reports The data returned from the Analytics API call.
+ * @param {Object} options Extraction options.
  * @return {Array} The dataMap ready for charting.
  */
-export const extractAnalyticsDashboardData = ( reports, selectedStats, days ) => {
+export function extractAnalyticsDashboardData( reports, options ) {
```
felixarntz, line 164: See above, let's not change this function's signature - the new code shouldn't use it.

assets/js/modules/analytics/util/index.js
```diff
@@ -131,15 +131,16 @@ export function extractAnalyticsDataForPieChart( reports, options = {} ) {
...
  *
- * @param {Array} rows          An array of rows to reduce.
- * @param {Array} selectedStats The currently selected stat we need to return data for.
+ * @param {Array}  rows           An array of rows to reduce.
+ * @param {number} selectedMetric The currently selected metric index.
+ * @param {number} selectedStats  The currently selected stat we need to return data for.
  * @return {Array} Array of selected stats from analytics row data.
  */
-function reduceAnalyticsRowsData( rows, selectedStats ) {
+function reduceAnalyticsRowsData( rows, selectedMetric, selectedStats ) {
```
felixarntz, line 139: Potentially this is worth refining as a more reusable function - but if we touch this legacy function to improve it, we should think it through, i.e. not just patch what's needed but also try to make it more understandable. The naming here is completely unclear to me.

I suggest we either change this function further, or we introduce an updated (separate) version of it that new code should use. Either is fine with me, but we should make sure the new code only uses functions we can consider well reusable and understandable.

assets/sass/widgets/_googlesitekit-widget-analytics-module.scss
```diff
@@ -0,0 +1,29 @@
...
+
+	.googlesitekit-widget--analyticsModuleOverview {
+
+		.mdc-layout-grid,
+		.googlesitekit-layout__header-title {
+			margin: 0;
+			padding: 0;
+		}
+	}
+}
```
felixarntz, line 29: See my above comment. This should be reverted - the only reason it's needed right now is to hack around using `LayoutHeader` within `WidgetHeader`, which shouldn't be done.

felixarntz (changes requested):
While this is off to a great start regarding its user-facing behavior, there are a couple of technical iterations needed. Most importantly, we need to optimize the amount of queries for this; since they're all within the same component anyway, that shouldn't be a stretch. We also should stop using legacy functions in the new components and rather abstract them into more manageable reusable utility functions, while keeping the very specific parts directly within the component.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/Header.js
eugene-manuilov: Updated.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/Overview.js
eugene-manuilov: Updated.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/SiteStats.js
eugene-manuilov: Updated.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/index.js
eugene-manuilov: Updated.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/Header.js
eugene-manuilov: Ok, updated it.

assets/sass/widgets/_googlesitekit-widget-analytics-module.scss
eugene-manuilov: Updated.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/SiteStats.js
eugene-manuilov: @felixarntz this `extractAnalyticsDashboardData` is only intended for the legacy and new site stats components. We don't use it anywhere else. So I don't think we need to over-engineering here and introduce a more complicated approach.

I have updated it to minimize the number of changes in that function to make it less complicated. Let me know if it works for you.

assets/js/modules/analytics/util/index.js
eugene-manuilov: This is a private function that is used by the `extractAnalyticsDashboardData` function only. I don't think it is worth reworking this function. I have just adopted it to work with a slightly different dataset than is used in the legacy component.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/Header.js
```diff
@@ -0,0 +1,67 @@
...
+
+				<Cell className="mdc-layout-grid__cell--align-right-tablet" lgSize={ 6 } mdSize={ 4 } smSize={ 4 } alignMiddle>
+					<Link href={ visitorsOverview } external inherit>
+						{ headerCTALabel }
+					</Link>
+				</Cell>
+			</Row>
+		</Grid>
+	);
+}
```
felixarntz, line 67: #3015 introduces reusable `WidgetHeaderTitle` and `WidgetHeaderCTA` components; these should be used here as well, instead of introducing an extra grid wrap here which isn't necessary for this simple UI.

Since this PR here is pretty much ready for being merged though, it shouldn't need to wait for the other one. Let's open a follow-up PR to update this once the components and CSS changes from #3015 have landed.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/SiteStats.js
```diff
@@ -0,0 +1,161 @@
...
+export default function SiteStats( { loaded, selectedStat, report } ) {
+	const currentDayCount = useSelect( ( select ) => select( CORE_USER ).getDateRangeNumberOfDays() );
+
+	if ( ! loaded ) {
+		return (
+			<PreviewBlock width="100%" height="270px" />
+		);
+	}
+
+	const dataMap = extractAnalyticsDashboardData( report, selectedStat, currentDayCount, 0, 1 );
```
felixarntz, line 44: @eugene-manuilov Following up on our previous conversation: My main issue with this function is that it is for a super-specific purpose; I'd prefer if it wasn't a general utility function but rather lived close to the one or two components that it is specific to.

We can keep this for now, but I think we should revisit it once we remove the legacy site stats component. Could be as simple as moving right into this file and not exporting it.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/Header.js
felixarntz: I've left a follow-up comment in my new review. There is some overlap with #3015 here.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/SiteStats.js
felixarntz: Left a follow-up comment on my new review.

assets/sass/widgets/_widgets.scss
```diff
@@ -45,6 +45,12 @@
 	&__body,
 	&__footer {
 
+		& .googlesitekit-table,
+		& > .mdc-layout-grid {
+			margin: 0;
+			padding: 0;
+		}
+
```
felixarntz, line 53: This change should also be reverted, as it visually breaks the All Traffic widget in the dashboard. It shouldn't be needed in the first place if we use the approach from #3015 for the widget header. Either you can copy the components from there to here and go with the same approach right away (shouldn't cause a merge conflict if you use the exact same code), or you can add a temporary CSS rule that only affects the specific widget here.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/index.js
```diff
@@ -0,0 +1,88 @@
...
+	const args = {
+		...dates,
+		dimensions: 'ga:date',
+		metrics: [
+			'ga:users',
+			'ga:sessions',
+			'ga:bounceRate',
+			'ga:avgSessionDuration',
+		],
+	};
```
felixarntz, line 57: While I suggested to combine the metrics into one request, now while testing I noticed a regression over what the current/legacy overview widget displays, which is due to a really weird issue with the Analytics API itself 😞 

We'll need to request the `ga:users` metric total separately, because when adding the `ga:date` dimension, it is different (for whatever inexplicable reason, this has come up before, and removing the dimension is the way to fix/work around it). I just re-tested this again now, it really is only the `ga:users` metric where the totals differ based on whether the dimension is there or not. The other three metrics have the same totals between both requests.

I suggest we make two requests here, the one you have, and another one that is exactly the same, just without the `dimensions` argument. Then, `Overview` can use the one without the dimensions to have the correct number.

felixarntz (changes requested):
This is almost good to go now, though there are two problems that need to be addressed. There is also some overlap with #3015 which I explain in the comments. We should eventually render the widget's header in the same way it is established in that PR.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/SiteStats.js
eugene-manuilov: Yes, agreed. Once we remove the legacy component, we won't need it to be in the utils folder and can simply move it to the component file.

assets/sass/widgets/_widgets.scss
eugene-manuilov: Yep, copied those components and reverted my css chagnes.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/index.js
eugene-manuilov: Sounds good. Updated.

felixarntz (approved):
Awesome work!

Issue #2919, opened 2021-03-05
Use alternate Composer setup in E2E jobs for efficiency


aaemnnosttv:
## Feature Description

We recently updated our E2E actions to run Composer using the `php-actions/composer` action. Using this action adds ~2 minutes to the job run because it requires building the container to use which provides both Composer v1 and v2.

This is much less efficient than how we run Composer for PHP Tests which installs Composer using `shivammathur/setup-php@v2` – which takes ~45s.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* E2E test jobs for GitHub actions should be updated to install and/or run Composer in a way that is more efficient than `php-actions/composer`

## Implementation Brief

* Merge #2921

### Test Coverage

* No change

### Visual Regression Changes

* N/A

## QA Brief

* Review run time for `setup-php` and `composer install` steps in E2E jobs on GitHub actions
* See total time for these steps is now ~15s compared to ~2.5m before  
![image](https://user-images.githubusercontent.com/1621608/110109452-24367280-7db6-11eb-9f02-a66533bcdf1a.png)

## Changelog entry

* N/A

eugene-manuilov:
IB :heavy_check_mark:

wpdarren:
@eugene-manuilov @aaemnnosttv please could you provide steps for running E2E jobs on GitHub actions, this isn't something I have done before, but happy to learn!

eugene-manuilov:
@wpdarren E2E jobs are run automatically, we can't run them manually. You can compare existing e2e tests for a PR before merging changes for this ticket with a PR that has been created after these changes have been merged.

wpdarren:
@eugene-manuilov @aaemnnosttv wooosh, that went right over my head! :) More direction here would be welcome.

cole10up:
### QA ✅

Before:
PR 20 days ago
![image](https://user-images.githubusercontent.com/57115907/110593601-a2857280-8130-11eb-9264-24260428995a.png)

After:
PR today
![image](https://user-images.githubusercontent.com/57115907/110593497-7d90ff80-8130-11eb-931e-a43b0b3f947d.png)

Sending to approved

PR #2921, opened 2021-03-05
Update E2E jobs to use setup-php vs php-actions/composer

aaemnnosttv:
## Summary

Addresses issue #2919 

On the first run with this PR:
* `setup-php@v2` took a maximum of 2s
* `composer install` took a maximum of 14s

It seems as though the cache was already primed (perhaps from other action runs?) but again, this was on the first run for this PR so 🤷‍♂️  😄 

Looking at other E2E runs recently on `develop` it looks like some are just as fast as these but those still took 2 min extra to build the `php-actions/composer`. There seems to a substantial variation in the length of the tests themselves which is a bit odd.

## Checklist


.github/workflows/e2e-tests-wp-4-7.yml
```diff
@@ -65,10 +71,7 @@ jobs:
           restore-keys: ${{ runner.os }}-composer-
 
       - name: Composer Install
-        uses: php-actions/composer@v4
-        with:
-          dev: no
-          php_version: 7.3
+        run: composer install --no-interaction --no-progress
```
eugene-manuilov, line 74: @aaemnnosttv I think we can also use the `--no-dev` flag since we don't use development dependencies in the e2e tests, right? Let's add it then.

.github/workflows/e2e-tests-wp-4-7.yml
```diff
@@ -54,6 +54,12 @@ jobs:
       - name: Build
         run: npm run build:test
 
+      # E2E tests use a dedicated container, this is just for Composer.
+      - uses: shivammathur/setup-php@v2
+        with:
+          php-version: '7.x'
+          tools: composer
+
```
eugene-manuilov, line 62: I don't think we need this anymore. [Ubuntu 20.04](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#php) (aka `ubuntu-latest`) comes with PHP 7.4.x and Composer 2.0.x preinstalled. We can simply run `php7.4 /usr/bin/composer install --no-interaction --no-progress --no-dev` in the `Composer Install` step and it will work just fine.

eugene-manuilov (changes requested):

.github/workflows/e2e-tests-wp-4-7.yml
aaemnnosttv: Oops yeah, you're right 👍  Will fix.

.github/workflows/e2e-tests-wp-4-7.yml
aaemnnosttv: I didn't realize that the runner came with both PHP 7.4 and 8 (I thought it was only 8 now). I think this step is okay to keep since it runs in literally 1-2s and won't rely as much on the runner installed versions. Maybe this action even uses the same preinstalled binary when run like this?

Let's keep it like this for now to avoid surprises due to changes on what is installed on the runner.

eugene-manuilov (approved):
LGTM

Issue #2727, opened 2021-01-29
Add support for breakpoint-specific sizes to `PreviewBlock`


felixarntz:
The `PreviewBlock` component only supports a single `width` and `height`. This is not always sufficient, as the "loaded" versions of where the `PreviewBlock` is used may differ in height e.g. between tablet and desktop, or even desktop and larger desktops.

See https://github.com/google/site-kit-wp/pull/2712#pullrequestreview-579513552, where a workaround around this limitation was necessary.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `PreviewBlock` component should be expanded to support additional props `smWidth`, `smHeight`, `mdWidth`, `mdHeight`, `lgWidth`, `lgHeight`.
* The values passed to these props should result in the respective width/height only being applied in the specific viewport.
* `width` and `height` should continue to be supported, e.g. if no further specification per breakpoint is necessary, or as fallback.

## Implementation Brief
* Install the `@react-hook/window-size` [package](https://www.npmjs.com/package/@react-hook/window-size).
* Create new file `assets/js/hooks/useBreakpoint.js` which will export the `useBreakpoint` function. The function should:
	* use the `useWindowWidth` hook, from `@react-hook/window-size/throttled`
	* based on the width returned from `useWindowWidth`, return the following string if:
		* width > 1280: `xlarge`
		* width > 960: `desktop`
		* width > 600: `tablet`
		* default: `small`
	* The values above are the same as defined in the SASS variables file, to maintain consistency between the JS and CSS breakpoints. https://github.com/google/site-kit-wp/blob/develop/assets/sass/config/_variables.scss#L157-L162
* Edit the `PreviewBlock` component:
	* To accept the following additional props, all string:
		* `smallWidth`
		* `smallHeight`
		* `mediumWidth`
		* `mediumHeight`
		* `largeWidth`
		* `largeHeight`
	* Use the `useBreakpoint` hook to get the current breakpoint.
	* Add 2 new variables, `blockWidth` and `blockHeight` which should be initialized to the `width` and `height` props values respectively.
	* Update the output of the component to use the `blockWidth` and `blockHeight` variables when setting the width and height of the `googlesitekit-preview-block` container.
	* If the breakpoint is `small` and the props `smallWidth`, `smallHeight` are defined, set the `blockWidth` and `blockHeight` to the `smallWidth`, `smallHeight` respectively.
[...]

PR #2922, opened 2021-03-05
Enhancement/2727 preview block

asvinb:
## Summary

Addresses issue #2727 

## Checklist


eugene-manuilov (approved):
From the technical point of view, everything looks okay. However, the selected approach seems to be excessive and goes beyond what has been requested in the AC. 

I don't think we need to introduce new hooks and a new dependency to implement responsive preview blocks. We can use `window.matchMedia()` to check breakpoints and it will do the trick. The only benefit of the new hook is that it listens to the `onresize` event on our behalf, but I am not convinced that it's really needed. Even though if we really needed it, it will be better to implement the `onresize` event listener in the component itself instead of bringing in a new dependency. Fewer dependencies we use are better for us.

I am going to approve it and leave the final decision to @aaemnnosttv or @felixarntz.

aaemnnosttv:
Thoughts @tofumatt? You suggested the added package that @asvinb added per the IB, but @eugene-manuilov feels it may not be necessary. What do you think?

tofumatt (approved):
My instinct is that especially for code like this, where the dependency is not a large API surface area or likely to change much—it does one thing and does it well—we're better off using an external dependency over maintaining the code ourselves. There's less code we need to write and test, less logic for more mundane things like this, etc.

I can see both arguments, but in this case the additional dependency doesn't create upgrade difficulties or anything. Changes like Styled Components, a UI component, or even some sort of Redux middleware introduce issues. But in this case—it's a small amount of code we don't have to maintain, and were we to write this code ourselves I think we'd do it exactly the same as the hook.

It's a useful feature to have in the code so I'd rather it be reusable than not.

So my take is that we include the dependency. 🙂

aaemnnosttv:
@googlebot I consent.

aaemnnosttv (approved):

Issue #2671, opened 2021-01-21
Refactor `getWidgetLayout`, `combineWidgets` and widget components to rely on grid components


felixarntz:
Although we have dedicated `Grid`, `Row` and `Cell` components, the Widgets API internally relies on `div`s with the respective classes being manually applied. We should simplify `getWidgetLayout`, `combineWidgets`, `WidgetAreaRenderer` and `WidgetRenderer` to rely on these components and only pass around grid column widths instead of full class names.

This change is okay to add even with the Widgets API production-ready, since all of it is only relevant internally.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* 

## Implementation Brief

* 

### Test Coverage

* 

### Visual Regression Changes

* 

## QA Brief

* 

## Changelog entry

*

benbowler:
IB added.

I've given this one a 19 point estimate because I think there could be some edge cases that could come with updating such a fundamental component, for example, I know that there have been issues using the HIDDEN_CLASS being used to render null components while keeping them in the DOM and because with the IB this will no longer add this class we may need to investigate side effects there.

aaemnnosttv:
@felixarntz assigning this to you to review since you're the most familiar with these parts. Also wanted to note the changes in #2612 may require adjustments here so it would be good to sync on that first.

felixarntz:
IB ✅ 

@benbowler @aaemnnosttv The IB is still valid, even against the changes that #2612 will make. However, the actual implementation will surely run into merge conflicts, so I've marked this one as blocked by #2612 for further execution.

johnPhillips:
> If gridColumnWidths[ i ] >= 6, props should be <Cell size={ gridColumnWidths[ i ] }>.

@aaemnnosttv or @felixarntz - Just noticing that in these situations with widgets of width 6, I'm seeing a lot of snapshot diffs that look like this:

```js
// Snapshot output for several tests in WidgetAreaRenderer.test.js
-  class="mdc-layout-grid__cell mdc-layout-grid__cell--span-6-desktop mdc-layout-grid__cell--span-8-tablet"
+  class="mdc-layout-grid__cell mdc-layout-grid__cell--span-6"
```

I haven't seen any visual regressions yet, but I'm still very much learning this part of the codebase. Is losing that tablet / desktop distinction going to be a problem?

aaemnnosttv:
[...]

PR #2927, opened 2021-03-08
Refactor `getWidgetLayout`, `combineWidgets` and widget components to rely on grid components (#2671)

johnPhillips:
## Summary

Addresses issue #2671

## Relevant technical choices

Tried to follow the AC and IB as closely as I could, as this is my first foray into this part of the codebase.

## Checklist


assets/js/googlesitekit/widgets/components/WidgetWrapper.js
```diff
@@ -0,0 +1,58 @@
...
+	if ( gridColumnWidth === 0 ) {
+		// Widget with columnWidth of 0 should be hidden
+		return (
+			<div className={ HIDDEN_CLASS }>
+				{ children }
+			</div>
+		);
+	} else if ( gridColumnWidth >= 6 ) {
+		return (
+			<Cell size={ gridColumnWidth }>
```
johnPhillips, line 40: I left a [comment on the issue](https://github.com/google/site-kit-wp/issues/2671#issuecomment-792789473) regarding this. I have implemented this according to the IB / AC, but I think we may introduce a visual regression here. 
Happy to do as the reviewer recommends.

assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js
```diff
@@ -53,26 +52,31 @@ export default function WidgetAreaRenderer( { slug, totalAreas } ) {
...
-			key={ widget.slug }
-			slug={ widget.slug }
-		/>
-	) );
+	const widgetsOutput = widgets.map( ( widget, i ) => {
+		return (
+			<WidgetWrapper
+				key={ `${ widget.slug }-wrapper` }
+				gridColumnWidth={ gridColumnWidths[ i ] }
+			>
```
eugene-manuilov, line 68: Let's use a shorter notation:

```javascript
const widgetsOutput = widgets.map( ( widget, i ) => (
    <WidgetWrapper
        key={ `${ widget.slug }-wrapper` }
        gridColumnWidth={ gridColumnWidths[ i ] }
    >
        ...
    </WidgetWrapper>
) );
```

assets/js/googlesitekit/widgets/components/WidgetRenderer.js
```diff
@@ -34,10 +34,11 @@ import { STORE_NAME } from '../datastore/constants';
 import BaseWidget from './Widget';
 import { getWidgetComponentProps } from '../util';
 import { HIDDEN_CLASS } from '../util/constants';
+// import { Cell } from '../../../material-components';
```
eugene-manuilov, line 37:
```suggestion
```

assets/js/googlesitekit/widgets/components/WidgetWrapper.js
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Internal dependencies
+ */
+import { HIDDEN_CLASS } from '../util/constants';
+import { Cell } from '../../../material-components';
+
+const WidgetWrapper = ( { gridColumnWidth, children } ) => {
+	switch ( true ) {
+		case 0 === gridColumnWidth:
```
eugene-manuilov, line 32: @johnPhillips let's better use `if` conditions instead of this fancy switch:

```javascript
if ( 0 === gridColumnWidth ) {
    return ...;
}

if ( gridColumnWidth === 3 || gridColumnWidth === 4 ) {
    return ...;
}

if ( gridColumnWidth === 6 || gridColumnWidth === 8 ) {
    return ...;
}

...
```

assets/js/googlesitekit/widgets/components/WidgetWrapper.js
```diff
@@ -0,0 +1,66 @@
...
+ */
+import PropTypes from 'prop-types';
+
+/**
+ * Internal dependencies
+ */
+import { HIDDEN_CLASS } from '../util/constants';
+import { Cell } from '../../../material-components';
+
+const WidgetWrapper = ( { gridColumnWidth, children } ) => {
```
eugene-manuilov, line 30: @johnPhillips how about other sizes? What if the `gridColumnWidth` equals to 5 or 7? It will return undefined and the widget won't be displayed at all, right?

eugene-manuilov (changes requested):
Thanks, @johnPhillips! It's a good start, however, not everything has been implemented yet. See the following statement from the IB that still needs to be addressed:

> Remove the WIDTH_GRID_CLASS_MAP constant and all of it's uses in:
> 
> ...
> assets/js/googlesitekit/widgets/util/constants.js:
> ...

assets/js/googlesitekit/widgets/components/WidgetWrapper.js
johnPhillips: Yeah, true. Looking at the code and tests I'm not sure 5 or 7 could happen, but I will handle this just in case.

johnPhillips: Note that I tried to keep the output (in terms of CSS classes in the snapshots and unit tests etc.) the same after this refactor as it was before, and there doesn't seem to be any pre-existing examples of 5 or 7 that I can see. I added these to be handled like `12` as I'm unsure how the mobile / tablet sizing would go in this scenario.

eugene-manuilov (dismissed):
LGTM

assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js
```diff
@@ -53,25 +52,28 @@ export default function WidgetAreaRenderer( { slug, totalAreas } ) {
...
-		<WidgetRenderer
-			gridClassName={ classnames( gridClassNames[ i ] ) }
-			OverrideComponent={ overrideComponents[ i ] ? () => {
-				const { Component, metadata } = overrideComponents[ i ];
-				return <Component { ...metadata } />;
-			} : undefined }
-			key={ widget.slug }
-			slug={ widget.slug }
-		/>
+		<WidgetWrapper
```
felixarntz, line 64: Maybe let's use a more specific name here? How about `WidgetCellWrapper`? While technically it can also wrap the widget in a div that hides it, the primary purpose is to wrap it in a grid cell.

assets/js/googlesitekit/widgets/components/WidgetAreaRenderer.js
```diff
@@ -53,25 +52,28 @@ export default function WidgetAreaRenderer( { slug, totalAreas } ) {
...
+			key={ `${ widget.slug }-wrapper` }
+			gridColumnWidth={ gridColumnWidths[ i ] }
+		>
+			<WidgetRenderer
+				OverrideComponent={ overrideComponents[ i ] ? () => {
+					const { Component, metadata } = overrideComponents[ i ];
+					return <Component { ...metadata } />;
+				} : undefined }
+				key={ widget.slug }
```
felixarntz, line 73: This prop shouldn't be necessary anymore now that it's not the top component in the loop.

assets/js/googlesitekit/widgets/components/WidgetWrapper.js
```diff
@@ -0,0 +1,69 @@
...
+		);
+	}
+
+	if ( gridColumnWidth === 6 || gridColumnWidth === 8 ) {
+		return (
+			<Cell lgSize={ gridColumnWidth } mdSize={ 8 }>
+				{ children }
+			</Cell>
+		);
+	}
```
felixarntz, line 54: This doesn't fully match the AC requirements:
1. The extra rule for `6` and `8` shouldn't be needed; in that case it should simply be a single `size`.
2. The rule for `3` and `4` shouldn't be that specific. While the ACs mention that these are realistically the only values that would be used, it states that this should apply to any width lower than 6 (except 0 of course).

So basically, we can remove one condition here completely while making the other apply to anything below 6. Unless there was a specific reason you diverged from the ACs here?

assets/js/googlesitekit/widgets/util/get-widget-layout.js
```diff
@@ -196,7 +202,18 @@ export function getWidgetLayout( widgets, widgetStates ) {
...
+				// Extract the columns on the current row to be resized (we
+				// don't want to include the overflowing widget or previously
+				// calculated rows)
+				let resizeRow = rowColMap[ rowIndex ];
+				// Resize them
+				[ resizeRow, counter ] = resizeColumns( resizeRow, counter );
+				// Insert the resized row back in
+				rowColMap[ rowIndex ] = resizeRow;
+				// Update the columnWidths with the resized row
+				columnWidths = Object.values( rowColMap ).flat();
```
felixarntz, line 216: Why was the logic here altered and why a new function `getRowColumnMap` introduced? This should work as before, so why not just get rid of `classNames` and leave the other logic unaltered?

felixarntz (changes requested):
This looks great for the most part, but there are some minor suggestions for improvement and a few questions to clarify.

assets/js/googlesitekit/widgets/util/get-widget-layout.js
johnPhillips: My source of truth while working on this was the snapshots, the rationale being that if they all passed then I had a reasonable hope that the refactor didn't cause a regression. So I only deviated from the IB / ACs when the snapshot tests failed.

In this case, the resize operation seemed to malfunction and some snapshot tests started failing. For example this test:
```
WidgetAreaRenderer should resize widgets in a row that spans 9 columns to fill the full 12 columns 
(3, 12, 3-6, 6 -> 3, 12, 4-8, 6)
```
The resize operation calculated the 9-column row correctly (`3-6 -> 4-8`), but then continued into the previous rows. So here it adjusted the `3` on the first row to be a `4`, meaning the final output was `4, 12, 4-8, 6`.

I found that if I only passed the offending row, it was resized correctly and the tests passed.

I think it's to do with this change, because we lose the `includes` from the original, which might have been why we didn't see this issue with the previous code:
```js
-	if ( singleWidgetClassNames.includes( 'mdc-layout-grid__cell--span-3-desktop' ) ) {
-		singleWidgetClassNames[ singleWidgetClassNames.indexOf( 'mdc-layout-grid__cell--span-3-desktop' ) ] = 'mdc-layout-grid__cell--span-4-desktop';
+	if ( columnWidths[ i ] === 3 ) {
```

That's my best guess though - I'll be honest, I found this quite a tricky ticket! 😂
I'm very happy to remove this if the regression is not an issue. Or maybe I've misunderstood something along the way here.

assets/js/googlesitekit/widgets/components/WidgetWrapper.js
johnPhillips: As I explained below, I tried to use the snapshots as the source of truth throughout. 

> The extra rule for 6 and 8 shouldn't be needed; in that case it should simply be a single size.

In this case I deviated from the ACs because I was seeing diffs like this:
### 
```diff
-  class="mdc-layout-grid__cell mdc-layout-grid__cell--span-6-desktop mdc-layout-grid__cell--span-8-tablet"
+  class="mdc-layout-grid__cell mdc-layout-grid__cell--span-6"
```
I was concerned that this might cause a regression. It seems that for a `6` and `8` on desktop, we currently do `8` on tablet. Can you advise?

> The rule for 3 and 4 shouldn't be that specific. While the ACs mention that these are realistically the only values that would be used, it states that this should apply to any width lower than 6 (except 0 of course).

I can certainly change this 👍

felixarntz: @johnPhillips Ah right, thanks, that's a good catch. In that case, it should be adjusted as follows:
* If smaller than 6, use the first condition (4 on medium, 2 on small)
* If smaller than 8, use the second condition (8 on medium)
* Otherwise use the default (no fallback, always full amount of columns)

assets/js/googlesitekit/widgets/util/get-widget-layout.js
felixarntz: @johnPhillips I found the problem here. It was indeed broken before, but since previously we weren't using the column widths (nor really testing them) it never surfaced.

* The problem is https://github.com/google/site-kit-wp/pull/2927/files#diff-1accd10359d13467fb6f9186730cc4ed6a8b0aac9139453bcd5cb3a6f165e807R187
* In that line we add the column width for the current widget to the list of column widths before we have even fitted it into the current row. This is different from before where the class names were only added to the actual area at the very end (see https://github.com/google/site-kit-wp/pull/2927/files#diff-1accd10359d13467fb6f9186730cc4ed6a8b0aac9139453bcd5cb3a6f165e807L214, there's also a comment on why that was done).
* The problem is basically that due to adding the current column width to the list of widths too early, one more entry is in there than it should. Before this PR, this simply meant the column widths were out of sync with the class names passed to `resizeClasses`. Now that we actually rely on the column widths though, this surfaces as a problem, breaking the expected behavior.
* When reverting to the simpler previous code where it was only calling `resizeClasses` here and adding a `console.log` to the beginning of `resizeClasses`, you will see that it's always one too many numbers that is passed to the function. That becomes more apparent when you add the same `console.log` to the current `develop` codebase: In there, you'll be able to see that it's always one more entry in `columnWidths` than in `classNames`, and the last entry is the one that shouldn't be there.

In other words, TLDR:
* Please remove the new extra logic and function you've added here and rely on calling `resizeClasses` as before.
* Instead, to fix the problem remove the early addition to the `columnWidths` array in https://github.com/google/site-kit-wp/pull/2927/files#diff-1accd10359d13467fb6f9186730cc4ed6a8b0aac9139453bcd5cb3a6f165e807R187 and instead do that in https://github.com/google/site-kit-wp/pull/2927/files#diff-1accd10359d13467fb6f9186730cc4ed6a8b0aac9139453bcd5cb3a6f165e807R231.

assets/js/googlesitekit/widgets/components/WidgetWrapper.js
johnPhillips: No problem, thanks!

assets/js/googlesitekit/widgets/util/get-widget-layout.js
johnPhillips: Ah, well spotted! I think that bug required an extra brain in the end 😂

I can't seem to get the link in your comment to work, but I think you meant [this line](https://github.com/google/site-kit-wp/blob/develop/assets/js/googlesitekit/widgets/util/get-widget-layout.js#L181)?
```js
// Populate initial column width and row index for the widget.
// These may be corrected below in certain situations.
columnWidths.push( WIDTH_GRID_COUNTER_MAP[ width ] ); //  ← 💥
rowIndexes.push( rowIndex );
```

When I remove that, the tests pass, so hopefully that's it. Let me know if I misunderstood.

I've removed the extra logic too 👍

felixarntz (commented):
Awesome, this is pretty much good for merge!

Just a few nit-picks, I'll quickly address them.

felixarntz (approved):

Issue #2739, opened 2021-02-01
Add tooltip tour for Analytics All Traffic widget


felixarntz:
There should be a tooltip tour explaining the new All Traffic widget. While it was already shipped in 1.25.0, it's still worth shipping this a bit later. The copy will be worded accordingly to reflect that.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a new tooltip tour for the All Traffic widget. The following tooltips should be displayed:
     1. Element to attach the tooltip to: the new chronological trend chart, see [Figma](https://www.figma.com/file/lcOhMAcYcc4KMOZ3XsgNFe/Tooltips?node-id=132%3A149). Tooltip copy: 
          * Heading: "It's now easier to see your traffic at a glance!"  
          * Text: "Check the trend graph to see how your traffic changed over time"
     2. Element to attach the tooltip to: the donut chart, see [Figma](https://www.figma.com/file/lcOhMAcYcc4KMOZ3XsgNFe/Tooltips?node-id=132%3A294). Tooltip copy: 
          * Heading: "See where your visitors come from"  
          * Text: "Click on the chart slices to see how each segment changed over time"
     3. Element to attach the tooltip to: the date picker, see [Figma](https://www.figma.com/file/lcOhMAcYcc4KMOZ3XsgNFe/Tooltips?node-id=132%3A438). Tooltip copy: 
          * Heading: "Check how your traffic changed since you last looked"  
          * Text: "Pick a time frame to see the comparison with the previous period"
* This tooltip tour should be shown to everybody who initial Site Kit version is `1.25.0` or lower.

## Implementation Brief

**1. Updating API contexts**

Revert this commit (remove the comment)
https://github.com/google/site-kit-wp/commit/3936d3dd0

Create a new file, `/assets/js/googlesitekit/default-screen-contexts.js`, with the following content:

```js
const SCREEN_CONTEXT_ACTIVATION = 'Activation';
const SCREEN_CONTEXT_ADMINBAR = 'Adminbar';
const SCREEN_CONTEXT_DASHBOARD = 'Dashboard';
const SCREEN_CONTEXT_SETTINGS = 'Settings';
const SCREEN_CONTEXT_SINGLE = 'Single';
const SCREEN_CONTEXT_SPLASH = 'Splash';
const SCREEN_CONTEXT_USER_INPUT = 'UserInput';
[...]

PR #2934, opened 2021-03-09
Connect feature tour infra with React and add first tour

aaemnnosttv:
## Summary

Addresses issue #2739

## Relevant technical choices

* A few places may appear to have unnecessary changes, which is technically true, but in the process of reaching the end result here there was an added React context for the `viewContext` passed in the root. This was later removed as unnecessary, but I kept some of the things touched in the process as cleanup/touchups.

## Checklist


assets/js/components/TourTooltips.js
```diff
@@ -73,27 +74,26 @@ export default function TourTooltips( { steps, tourID } ) {
 	const stepKey = `${ tourID }-step`;
 	const runKey = `${ tourID }-run`;
 	const { setValue } = useDispatch( CORE_UI );
-	const [ shouldSkipTour, setShouldSkipTour ] = useLocalStorage( `sitekit-${ tourID }__has-encountered-tour` );
+	const { dismissTour } = useDispatch( CORE_USER );
```
felixarntz, line 77: I'm noticing now that #2650 has kind of been forgotten for this release - and it looks like you seem to be implementing that here. Since the PR on #2650 is out of date, maybe we can just keep it in here. cc @eugene-manuilov Can you give this here a pass as well?

assets/js/googlesitekit/datastore/user/feature-tours.js
```diff
@@ -141,6 +234,32 @@ const baseSelectors = {
...
+	 * Gets a list of tour objects that qualify for the given view context.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state       Data store's state.
+	 * @param {string} viewContext View context.
+	 * @return {(Object[]|undefined)} Array of qualifying tour objects
+	 *                                `undefined` while readiness is being resolved.
+	 */
+	getReadyTours( state, viewContext ) {
```
felixarntz, line 247: Let's name this `getCurrentTours` or, I think better, `getCurrentFeatureTours`, similar to how it was discussed during IB review. If we rename this to `getCurrentFeatureTours`, I'd also suggest to rename the above selector to `getDismissedFeatureTours`.

assets/js/googlesitekit/datastore/user/feature-tours.js
```diff
@@ -141,6 +234,32 @@ const baseSelectors = {
...
+	 * Gets a list of all tour objects.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {Object[]} Array of tour objects.
+	 */
+	getTours( state ) {
+		return state.tours;
+	},
```
felixarntz, line 261: Let's not add this selector, as it shouldn't be necessary. We only need to be able to get the relevant tours for a context, not all tours that are generally available. If we need it at some point, we can still add it.

assets/js/googlesitekit/datastore/user/feature-tours.js
```diff
@@ -80,19 +90,69 @@ const baseActions = {
...
+		};
+	},
+
+	receiveTours( tours ) {
+		invariant( Array.isArray( tours ), 'tours must be an array.' );
+		return {
+			payload: { tours },
+			type: RECEIVE_TOURS,
+		};
+	},
```
felixarntz, line 123: This appears to be unnecessary, and it isn't used since we can import the available feature tours directly. Let's remove it.

assets/js/googlesitekit/datastore/user/feature-tours.js
```diff
@@ -117,13 +195,28 @@ const baseReducer = ( state, { type, payload } ) => {
...
+	*getReadyTours( viewContext ) {
+		const registry = yield getRegistry();
+		const tours = registry.select( STORE_NAME ).getTours();
+		const readyTours = [];
+
+		for ( const tour of tours ) {
+			const tourQualifies = yield {
+				payload: { tour, viewContext },
+				type: CHECK_TOUR_REQUIREMENTS,
+			};
```
felixarntz, line 213: Nit-pick, but should we maybe make this available as a (private) action too, rather than only having the action creator in here?

assets/js/googlesitekit/datastore/user/feature-tours.js
```diff
@@ -80,19 +90,69 @@ const baseActions = {
...
-			if ( ! error ) {
-				yield receiveGetDismissedTours( response );
-			}
-			return { response, error };
+			// Dispatch a request to persist and receive updated dismissed tours.
+			return yield fetchDismissTour( slug );
 		}() );
 	},
+
+	receiveReadyTours( readyTours, { viewContext } = {} ) {
```
felixarntz, line 108: Let's rename this alongside the related selector, to `receiveGetCurrentFeatureTours`.

assets/js/feature-tours/index.js
```diff
@@ -0,0 +1,64 @@
...
+		// Here we need to wait for the underlying selector to be resolved before selecting `isModuleConnected`.
+		await registry.__experimentalResolveSelect( CORE_MODULES ).getModules();
+		if ( ! registry.select( CORE_MODULES ).isModuleConnected( 'analytics' ) ) {
+			return false;
+		}
+		return true;
+	},
+	steps: [
+		{
+			target: '.googlesitekit-widget--analyticsAllTraffic__user-count-chart',
```
felixarntz, line 44: Looking at the feature while testing, I think it would be actually more appropriate to target the entire left column of the All Traffic widget (i.e. user count number and chart); currently it looks weird what the focus is on.

assets/js/feature-tours/index.js
```diff
@@ -0,0 +1,64 @@
...
+			title: __( "It's now easier to see your site's traffic at a glance", 'google-site-kit' ),
+			content: __( 'Check the trend graph to see how your traffic changed over time', 'google-site-kit' ),
+		},
+		{
+			target: '.googlesitekit-widget--analyticsAllTraffic__dimensions-chart',
```
felixarntz, line 49: Same here, this should probably target the entire right column of the All Traffic widget (tabs and pie chart).

felixarntz (changes requested):
@aaemnnosttv This looks great for a start, I've mostly left some comments about naming and some cleanup.

As you've mentioned earlier today, the scroll behavior is a bit odd. Is there a way to update how `react-joyride` behaves in that regard? Preferably, the focused component and tooltip should be as much in the vertical center of the viewport as possible. The tooltip being visible is highest priority of course, but if the entire focused component _and_ tooltip can fit in the viewport, that should be the case too.

assets/js/components/TourTooltips.js
aaemnnosttv: I didn't even know about that issue and this was a simple change to make that seemed necessary 😄

assets/js/googlesitekit/datastore/user/feature-tours.js
aaemnnosttv: Not a problem to rename, but I feel like "current" doesn't quite have the same meaning that the returned tours are "qualified" and thus ready to show. I think an added `ForView` would also be a bit more clear.

assets/js/googlesitekit/datastore/user/feature-tours.js
aaemnnosttv: It's used in tests right now which is necessary for being able to test with different tours.

assets/js/googlesitekit/datastore/user/feature-tours.js
aaemnnosttv: I think you mean making this available as a private action creator rather than as a private action (no bound action creator exposed)? That would be a simple change to make but why would we do that unless we wanted to be able to call it elsewhere?

assets/js/googlesitekit/datastore/user/feature-tours.js
aaemnnosttv: How about `receiveCurrentFeatureTours`? `receiveGetCurrentFeatureTours` sounds like it is a fetch store action.

assets/js/googlesitekit/datastore/user/feature-tours.js
aaemnnosttv: It's needed by the resolver for `get(Ready|Current)FeatureTours`. I'll mark it as `@private` for now, but unfortunately the only alternative is reaching into `state` directly which we don't do anywhere else.

assets/sass/widgets/_googlesitekit-widget-analyticsAllTraffic.scss
```diff
@@ -196,7 +196,7 @@
 	.googlesitekit-widget--analyticsAllTraffic__totals {
 		// HACK: Fixes jumping of widget when a dimension is selected.
 		@media (min-width: $bp-desktop) {
-			height: 1px;
+			margin-bottom: $grid-gap-desktop * -1;
```
aaemnnosttv, line 199: Still feels like kind of a hack, but at least a better one 😄 
cc: @tofumatt

assets/js/googlesitekit/datastore/user/feature-tours.js
```diff
@@ -80,19 +90,69 @@ const baseActions = {
...
-			if ( ! error ) {
-				yield receiveGetDismissedTours( response );
-			}
-			return { response, error };
+			// Dispatch a request to persist and receive updated dismissed tours.
+			return yield fetchDismissTour( slug );
 		}() );
 	},
+
+	receiveReadyFeatureTours( readyTours, { viewContext } = {} ) {
```
felixarntz, line 108: Let's name this accordingly to `receiveFeatureToursForView`.

assets/js/googlesitekit/datastore/user/feature-tours.js
```diff
@@ -80,19 +90,69 @@ const baseActions = {
...
+		invariant( Array.isArray( readyTours ), 'readyTours must be an array.' );
+		invariant( viewContext, 'viewContext is required.' );
+		return {
+			payload: { readyTours, viewContext },
+			type: RECEIVE_READY_TOURS,
+		};
+	},
+
+	receiveFeatureTours( tours ) {
```
felixarntz, line 117: Let's name this accordingly to `receiveAllFeatureTours`.

felixarntz (approved):
LGTM functionality-wise, just left some more naming feedback. @aaemnnosttv I can quickly make these updates if you're okay with that.

felixarntz (approved):

Issue #2844, opened 2021-02-22
Avoid loss of user input responses due to navigating away


felixarntz:
## Bug Description

When the user navigates away from the user input screen without submitting their responses, they currently all get lost. This is most problematic because there is a link to "Settings" on that page that seems like it would be worth clicking it; but overall it is a problem that the responses aren't saved anywhere because the user could navigate away from the page at any point.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The user input responses should be temporarily stored in the browser (e.g. local storage) as the user goes through the user input flow, so that they are still in place, should they leave the page before submitting them.
* The storage should explicitly still be temporary, i.e. nothing should be sent to WordPress or the authentication service other than on submission, as it is now.
* Once the user submits the responses to the API, the temporary browser storage should be cleared, as the source of truth is now the API.
* The "Settings" link that appears on each user input question should no longer be a link. The copy should be kept, it just shouldn't link to "Settings" because otherwise it makes it too "easy" to accidentally navigate away from the screen.

## Implementation Brief

* Edit `user-input-settings` datastore:
  * Update the `setUserInputSetting` action:
    * Read the `user-input` property from the local storage with an empty object fallback if the property doesn't exist;
    * Add the current timestamp plus one hour to the `user-input` object under the `ttl` property;
    * Add the incoming values to the `values` property of the `user-input` object;
    * Save the `user-input` object in the local storage. Here is an example of the `user-input` object structure:
      ```javascript
      {
          "ttl": 1234567890,
          "values": {
              "setting1": [...],
              "setting2": [...],
              ...
          }
      }
      ```
  * Update the `saveUserInputSettings` action to remote the `user-input` item from the local storage.
  * Update the `getUserInputSettings` selector:
    * Read the `user-input` property from the local storage;
    * Do nothing if the `user-input` object is empty or its `ttl` value is less than the current timestamp;
    * Otherwise, merge values from the `values` property of the `user-input` object with user input settings received from the server.
* Remove the link to "Settings" page in the `UserInputQuestionNotice` component.

### Test Coverage

* Update the existing tests if any is failing due to the aforementioned changes.

### Visual Regression Changes

* N/A

## QA Brief

_Note: it's easiest to test this in Storybook and it will still work there, because the User Input questionnaire is only shown once._

* Load the User Input Questionnaire on a new site. By default, no questions should show as answered. (This requires enabling the "`userInput`" feature flag on the site.
[...]

PR #2943, opened 2021-03-10
Cache user input questions if the question state isn't "complete" and remove the "Settings" link.

tofumatt:
## Summary

Addresses issue #2844.

## Relevant technical choices

Rather than implement a custom TTL setting/logic and also safe storage checks, this re-uses our caching module—which already has flexible/configurable backend support and cache TTL logic.

## Checklist


assets/js/googlesitekit/datastore/user/user-input-settings.js
```diff
@@ -65,30 +75,50 @@ const baseInitialState = {
...
+	 * @since n.e.x.t
 	 *
-	 * @param {Object} values User input settings.
-	 * @return {Object} Redux-style action.
+	 * @return {Object} Cached user input answer values.
 	 */
-	setUserInputSettings( values ) {
-		return {
-			type: SET_USER_INPUT_SETTINGS,
-			payload: { values },
```
tofumatt, line 78: Note: this was _not_ working as-expected in my usage… using it to set all values resulted in bugs. It's possible the API wasn't clear/documented incorrectly, but more-to-the-point: we don't use this anywhere in the code and it wasn't even useful for this PR really, so I've removed it as:

1. we weren't using it
2. it wasn't working as-expected

The second point matters more, as it led me down a bit of a rabbit hole wondering why the code didn't work. Upon switching to `setUserInputSetting` it worked immediately, so I figured it was better to remove unused, unexpected code.

eugene-manuilov (approved):
Overall, everything looks good. The only thing that I am not sure about is whether we need to rely on the user input state flag. It has been requested neither in AC nor IB, so I leave it to @felixarntz to judge here.

eugene-manuilov (changes requested):
@tofumatt could you please check the broken js test?

tofumatt:
> The only thing that I am not sure about is whether we need to rely on the user input state flag.

Yeah, it wasn't in the ACs or IB, but I think it's the best way to handle the logic for storing temporary values. It'll still work if we store them all the time, but it's wasteful so I figured I'd only store when the state isn't "complete".

I've fixed the tests and this is good-to-go now, so I've moved it to merge review given the previous approval. 😄

assets/js/googlesitekit/datastore/user/user-input-settings.js
```diff
@@ -27,25 +27,33 @@ import isPlainObject from 'lodash/isPlainObject';
...
 	reducerCallback: fetchStoreReducerCallback,
 } );
 
 const fetchSaveUserInputSettingsStore = createFetchStore( {
 	baseName: 'saveUserInputSettings',
-	controlCallback: ( settings ) => API.set( 'core', 'user', 'user-input-settings', { settings } ),
+	controlCallback: ( settings ) => {
+		return API.set( 'core', 'user', 'user-input-settings', { settings } )
+		;
+	},
```
felixarntz, line 56: These changes seem unrelated/unnecessary?

assets/js/googlesitekit/datastore/user/user-input-settings.js
```diff
@@ -65,30 +75,50 @@ const baseInitialState = {
 
 const baseActions = {
 	/**
-	 * Sets user input settings.
+	 * Gets cached user input settings and save them to the data store.
 	 *
-	 * @since 1.19.0
+	 * @since n.e.x.t
```
felixarntz, line 80: Since this action is documented, it should be marked as private, since it's a low-level action that should only be used via the `getUserInputSettings` resolver. The alternative is to not document it at all.

assets/js/googlesitekit/datastore/user/user-input-settings.js
```diff
@@ -65,30 +75,50 @@ const baseInitialState = {
...
 	 *
-	 * @param {Object} values User input settings.
-	 * @return {Object} Redux-style action.
+	 * @return {Object} Cached user input answer values.
 	 */
-	setUserInputSettings( values ) {
-		return {
-			type: SET_USER_INPUT_SETTINGS,
-			payload: { values },
+	*fetchCachedUserInputSettings() {
```
felixarntz, line 84: This name is a bit off for two reasons:
* The `fetch*` prefix should only be used for the internal fetch actions related to a public API-based selector or action (typically using `createFetchStore`). The action here isn't that though.
* The function sets user input settings based on what is cached, which is not clear from the name (fetch implies it's getting/fetching something, but it also sets something).

I suggest something like `setUserInputSettingsFromCache`.

assets/js/googlesitekit/datastore/user/user-input-settings.js
```diff
@@ -65,30 +75,50 @@ const baseInitialState = {
...
-	setUserInputSetting( settingID, values ) {
+	*setUserInputSetting( settingID, values ) {
+		const registry = yield Data.commonActions.getRegistry();
+		if ( registry.select( STORE_NAME ).getUserInputState() !== 'completed' ) {
+			// Save this setting in the cache.
+			yield {
+				type: SET_CACHED_USER_INPUT_SETTING,
+				payload: {
+					settingID,
+					values,
```
felixarntz, line 117: This is missing the `trim` logic that is needed, see right below here.

assets/js/googlesitekit/datastore/user/user-input-settings.js
```diff
@@ -27,25 +27,33 @@ import isPlainObject from 'lodash/isPlainObject';
...
 import Data from 'googlesitekit-data';
+import { deleteItem, getItem, setItem } from '../../../googlesitekit/api/cache';
 import { STORE_NAME } from './constants';
 import { createFetchStore } from '../../data/create-fetch-store';
 import { actions as errorStoreActions } from '../../data/create-error-store';
-const { commonActions, createRegistrySelector } = Data;
+const { commonActions, createRegistryControl, createRegistrySelector } = Data;
 const { receiveError, clearError } = errorStoreActions;
 
+const CACHE_KEY_NAME = 'fetchGetUserInputSettingsStore.fetchGetUserInputSettingsStore';
```
felixarntz, line 37: Why this overly cryptic/specific cache key name with a duplicated segment? It makes me think that name is somehow required due to how it integrates with `fetchGetUserInputSettingsStore`. We can just give it an arbitrary name like `userInputSettings`.

assets/js/googlesitekit/datastore/user/user-input-settings.js
```diff
@@ -173,9 +218,14 @@ export const baseReducer = ( state, { type, payload } ) => {
 const baseResolvers = {
 	*getUserInputSettings() {
 		const { select } = yield commonActions.getRegistry();
-		if ( ! select( STORE_NAME ).getUserInputSettings() ) {
+
+		if ( ! select( STORE_NAME ).getUserInputSettings() && select( STORE_NAME ).getUserInputState() === 'completed' ) {
```
felixarntz, line 222: This seems wrong. There might be saved user input settings also when the state is not `completed`. This part shouldn't be changed, we need to only make sure that the cached user input settings take precedence over what comes from the server.

felixarntz (changes requested):
Overall looks solid; I left a bit of minor feedback, a few things need iteration though.

assets/js/googlesitekit/datastore/user/user-input-settings.js
tofumatt: Ah, I think I was doing something in there with console logs previously, removed them, and that's why the change is there. I've reverted it.

assets/js/googlesitekit/datastore/user/user-input-settings.js
tofumatt: Fair point—there's no reason for it other than specificity, but `userInputSettings` should be fine 👍🏻

assets/js/googlesitekit/datastore/user/user-input-settings.js
tofumatt: Ah, fair point. I guess since we never store it _after_ the user input status is `complete` that's safe to do. 👍🏻

felixarntz (approved):
Great!

PR #2956, opened 2021-03-12
Add fallback to durationFormat for browsers that do not support style: 'unit'.

johnPhillips:
## Summary

Some browsers do not fully support `Intl.NumberFormatOptions`, so I've added a fallback as discussed here: https://github.com/google/site-kit-wp/issues/2588#issuecomment-797318556

Addresses issue #2588

## Checklist

### DIFF
--- a/assets/js/util/i18n.js
+++ b/assets/js/util/i18n.js
@@ -47,6 +47,28 @@ const durationFormat = ( seconds, options = {} ) => {
 		style: 'unit',
 	};
 
+	// Some browsers, e.g. Safari, throw a RangeError when options.style is
+	// not one of decimal, percent, or currency.
+	function testNumberFormat() {
+		try {
+			// test call to numberFormat
+			numberFormat( 10, {
+				...options,
+				unit: 'hour',
+			} );
+			return true;
+		} catch ( e ) {
+			return false;
+		}
+	}
+
+	const browserSupportsStyleUnit = testNumberFormat();
+	let formattedString = '';
+
+	if ( ! browserSupportsStyleUnit ) {
+		options.style = 'decimal';
+	}
+
 	seconds = parseInt( seconds, 10 );
 
 	if ( Number.isNaN( seconds ) ) {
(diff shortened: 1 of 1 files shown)


assets/js/util/i18n.js
```diff
@@ -47,6 +47,28 @@ const durationFormat = ( seconds, options = {} ) => {
...
+				...options,
+				unit: 'hour',
+			} );
+			return true;
+		} catch ( e ) {
+			return false;
+		}
+	}
+
+	const browserSupportsStyleUnit = testNumberFormat();
```
ivankruchkoff, line 65: Since we're declaring and using the function in one place, there's no need to create a function and then call it once.

Why not something like:

```
let browserSupportsStyleUnit = true;

try {
    // test call to numberFormat
    numberFormat( 10, {
        ...options,
        unit: 'hour',
    } );
} catch (error) {
    options.style = 'decimal';
    browserSupportsStyleUnit = false;
}
```

assets/js/util/i18n.js
```diff
@@ -83,13 +105,21 @@ const durationFormat = ( seconds, options = {} ) => {
...
-		minutes,
-		hours,
-	);
+			_x( '%3$s %2$s %1$s', 'duration of time: hh mm ss', 'google-site-kit' ),
+			seconds,
+			minutes,
+			hours,
+		);
+	} else {
+		formattedString =
```
ivankruchkoff, line 118: Let's go with the consistent approach to localisation / translation rather than two different ways to do it, the only difference here is that we also add the units into the translatable string.:

```
formattedString = sprintf(
    /* translators: 1: formatted seconds, 2: formatted minutes, 3: formatted hours */
    _x( '%3$sh %2$sm %1$ss', 'duration of time with time units: XXh YYm ZZs', 'google-site-kit' ),
    seconds,
    minutes,
    hours,
);
```

ivankruchkoff (changes requested):

felixarntz:
@aaemnnosttv @johnPhillips I assume this is superseded by #2958?

Issue #2846, opened 2021-02-22
Introduce (contextual) help icon in Site Kit header bar


felixarntz:
The Site Kit header bar should include a help icon that, when clicked, exposes a dropdown with various help links. This should globally replace the need for the "Need help?" link in the Site Kit footer.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new feature flag `helpVisibility` should be introduced.
* If the feature flag is enabled, the Site Kit header bar should include a Help icon which exposes a dropdown with three help links per the [Figma file](https://www.figma.com/file/pvsqJE5yZe9cFeNlTKUEiU/Google-SK-UX-Part-3-Enhancements10up?node-id=1437%3A11).
    * On the AdSense screen, an additional "Get help with AdSense" link should be displayed.
* If the feature flag is enabled, the "Need help?" link in the footer should no longer appear.
* A new `HelpMenu` component should be introduced which renders the help menu icon (use a translatable `Help` label for screen readers) and a list of links in the dropdown, per the [Figma file](https://www.figma.com/file/pvsqJE5yZe9cFeNlTKUEiU/Google-SK-UX-Part-3-Enhancements10up?node-id=1437%3A11).
* The individual links should be using a new `HelpMenuLink` component each. The default implementation of `HelpMenu` should automatically render three `HelpMenuLink`s (also per Figma):
    * "Fix common issues" (https://sitekit.withgoogle.com/documentation/fix-common-issues/)
    * "Read help docs" (https://sitekit.withgoogle.com/documentation/)
    * "Get support" (https://wordpress.org/support/plugin/google-site-kit/)
* The `HelpMenu` component should support rendering additional `HelpMenuLink`s passed via `children`. These should be rendered _before_ the default links.
* The `Header` component should be updated to render its `children` unconditionally, not only when the user is authenticated. Any usage where currently `children` are passed require an authenticated user anyway and otherwise are unreachable.
* All screens that render the `Header` component should be updated to render the `HelpMenu` as its first (or only) child, if the `helpVisibility` feature flag is enabled.
[...]

PR #2957, opened 2021-03-12
Add help menu behind feature flag

tofumatt:
## Summary

Addresses issue #2846.

## Relevant technical choices

The help menu is not display in one section of the site: the GCP, non-proxy setup header. This is because it is not a functional component with access to the feature flags hook. This wasn't mentioned in the ACs/IBs, and my thinking was it was both not very necessary and it's a bit out-of-scope of the feature to add a `withFeatureFlags` HOC to make it work. But let me know if you'd like to see that and we could probably add it without too much fuss.

The IBs mentioned copying the entire `Menu` render output and reusing it, but that seemed a bit much. It was more straightforward to allow `Menu` to accept `children` and output them directly. Ultimately it seems like it'd be good if we had a `Dropdown` component, as the dropdown logic is already replicated in a few places, but I didn't want to make this PR too big as each dropdown is a bit difference so it'd be out-of-scope.

A minor note: the `HelpMenuLink` requested a `label` prop but `children` seemed a more natural/obvious choice for what it does, so that's what I used 😄 

## Checklist


tofumatt:
This is ready for review, but is marked as a draft PR until #2955 is merged, because until then all tests user the `wordpress` docker image will fail.

Hazlitte:
@googlebot I consent.

aaemnnosttv:
@googlebot I consent.

assets/js/components/help/HelpMenu.js
```diff
@@ -0,0 +1,104 @@
...
+					aria-controls="help-menu"
+					aria-expanded={ menuOpen }
+					aria-haspopup="menu"
+					className="googlesitekit-header__dropdown googlesitekit-help-menu__button mdc-button--dropdown"
+					icon={ <HelpIcon fill="#476282" width="20" height="20" /> }
+					onClick={ handleMenu }
+					ref={ menuButtonRef }
+					text
+				>
+					<VisuallyHidden>{ __( 'Help', 'google-site-kit' ) }</VisuallyHidden>
```
aaemnnosttv, line 81: Maybe an aria-label on the button would be simpler here?

assets/js/components/help/HelpMenu.js
```diff
@@ -0,0 +1,104 @@
...
+		};
+	}, [] );
+
+	const handleMenu = useCallback( () => {
+		toggleMenu( ! menuOpen );
+	}, [ menuOpen ] );
+
+	return (
+		<Fragment>
+			<div className="googlesitekit-user-selector googlesitekit-dropdown-menu googlesitekit-help-menu mdc-menu-surface--anchor">
```
aaemnnosttv, line 70: I noticed the `googlesitekit-user-selector` class here which doesn't really belong except that it removes the dropdown arrow. Let's extract this to a different class to not mix user-selector styles with this element.

assets/js/components/help/HelpMenuLink.js
```diff
@@ -0,0 +1,49 @@
...
+import PropTypes from 'prop-types';
+
+/**
+ * Internal dependencies
+ */
+import Link from '../Link';
+
+function HelpMenuLink( { children, href } ) {
+	return (
+		<li className="googlesitekit-help-menu-link mdc-list-item" role="menuitem">
```
aaemnnosttv, line 31: It looks like the list item here should actually have a `role` of "none" and the _link_ should have the `role="menuitem"`
Source: https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html

Looking at the menu button itself (within the `HelpMenu`) I think what we have is correct 🤔

assets/svg/help.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="#1d1d1d" xmlns="http://www.w3.org/2000/svg">
```
aaemnnosttv, line 1: Let's avoid using hardcoded fills on svgs for simple single color icons. In this case we should default to `currentColor` (I think this has to be on the respective `path` anyways, no?)
```suggestion
<svg width="24" height="24" viewBox="0 0 24 24" fill="currentColor" xmlns="http://www.w3.org/2000/svg">
```

package-lock.json
```diff
@@ -10089,11 +10089,6 @@
           "resolved": "https://registry.npmjs.org/prelude-ls/-/prelude-ls-1.2.1.tgz",
           "integrity": "sha512-vkcDPrRZo1QZLbn5RLGPpg/WmIQ65qoWWhcGKf/b5eplkkarX0m9z8ppCat4mlOqUsWpyNuYgO3VRyrYHSzX5g=="
         },
-        "prettier": {
```
aaemnnosttv, line 10092: I see this gets bounced around a lot between PRs. While we generally prefer to avoid committing changes to the lock file when not required, it's not a great experience. It would be great to implement a change that would prevent this from happening in the first place. I'm not sure there's much we can do about it (except maybe switch to using Yarn or some other npm client instead?). Thoughts?

stories/header.stories.js
```diff
@@ -51,4 +52,13 @@ storiesOf( 'Global', module )
 				</Header>
 			</WithTestRegistry>
 		);
+	} )
+	.add( 'Plugin Header with Help Menu ', ( registry ) => {
```
aaemnnosttv, line 56: Trailing space
```suggestion
	.add( 'Plugin Header with Help Menu', ( registry ) => {
```

stories/header.stories.js
```diff
@@ -51,4 +52,13 @@ storiesOf( 'Global', module )
...
+		return (
+			<WithTestRegistry features={ [ 'helpVisibility', 'storeErrorNotifications' ] } registry={ registry }>
+				<Header>
+					<HelpMenu />
```
aaemnnosttv, line 60: Can we also add a story that includes additional arbitrary menu items? Maybe we should add a story for the `HelpMenu` itself?

assets/sass/components/global/_googlesitekit-help-menu.scss
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+.googlesitekit-header .googlesitekit-user-selector .googlesitekit-help-menu__button {
+	margin-right: 0;
+}
+
+.googlesitekit-plugin .googlesitekit-help-menu .mdc-menu-surface {
+	width: auto;
+}
```
aaemnnosttv, line 25: These two might be better as utility classes rather than hidden away here.
E.g.
- `.googlesitekit-margin-right-0`
- `.googlesitekit-width-auto`

What do you think?

assets/sass/components/global/_googlesitekit-help-menu-link.scss
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+.googlesitekit-plugin .googlesitekit-help-menu-link .googlesitekit-cta-link {
+	color: $c-nearblack;
+	font-size: 16px;
```
aaemnnosttv, line 21: It would be great to use a variable for the size instead of the fixed value. It looks like our definitions in `$typography` are the only place it is defined but I'm not sure if the `responsive-text` mixin is appropriate here as I don't see it used on the other menus/items. I suppose it's okay to keep as is; what do you think?

aaemnnosttv (changes requested):
Thanks @tofumatt – this looks great overall. Just a few suggestions and comments, otherwise I think this is pretty close. Nice work! 👍

assets/js/components/help/HelpMenu.js
tofumatt: Sure thing—I think it's like this in another menu this was based off, but good point 👍🏻

stories/header.stories.js
tofumatt: I think an extra story with it in the plugin header is fine for now, as that's (mainly) where we use it, and it isn't more complicated than a few stories 🤷🏻‍♂️

assets/sass/components/global/_googlesitekit-help-menu-link.scss
tofumatt: I thought about using a variable too, but as this is from newer designs I don't know what to call this. I think the newer design mockups are using slightly different sizes, so we are stuck in the "in-between" state.

I don't see this size being used elsewhere or clearly defined yet, so I can't think of a useful variable to assign it to. I'd prefer to, but I think for now this is fine. We can tweak it later though 👍🏻

assets/sass/components/global/_googlesitekit-help-menu.scss
tofumatt: Works for me! I had to tweak another, older selector to be less specific so these'd work, but it seems fine. 👍🏻

tofumatt:
Thanks for the feedback, especially around the styles/CSS classes. Should be ready for another review 🙂

aaemnnosttv (commented):
This is really close, just a few final things to clean up. I pushed a few commits to fix some of these quick and easy details already so there shouldn't be too much left to do.

assets/sass/config/_variables.scss
```diff
@@ -58,6 +58,7 @@ $c-mango-tango: #e67700;
 $c-thunderbird: #c7221f;
 $c-solitude: #e8f0fe;
 $c-heather: #bdc1c6;
+$c-helplink: #476282;
```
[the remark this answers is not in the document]
tofumatt: Argh, it totally changed recently 😓 . It used to be a blue, but I see it's now using the colours you referenced. I'll update it 👍🏻

assets/sass/base/_defaults.scss
```diff
@@ -37,6 +37,8 @@
 	}
 
 	svg:not(:root) {
+		/* stylelint-disable-next-line sh-waqar/declaration-use-variable */
+		color: inherit;
```
tofumatt, line 41: Pretty sure this caused a bunch of VRT failures:

Screenshot 2021-03-18 at 19 17 08

assets/svg/help.svg
```diff
@@ -1,4 +1,4 @@
-<svg width="24" height="24" viewBox="0 0 24 24" fill="#1d1d1d" xmlns="http://www.w3.org/2000/svg">
+<svg width="24" height="24" viewBox="0 0 24 24" fill="currentColor" xmlns="http://www.w3.org/2000/svg">
```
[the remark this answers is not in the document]
tofumatt: Pretty sure this caused a bunch of VRT failures 😅 

Screenshot 2021-03-18 at 19 17 08

I think we need to target just this SVG for now...

aaemnnosttv:
@tofumatt there's was a merge conflict with the header stories now that the v6 issue has been merged which I resolved.

I also fixed the styles so that our svgs inherit `color` by default (there was a problem with the specificity of the change arrow styles that was breaking VRT).

aaemnnosttv (approved):
Awesome!

Issue #2900, opened 2021-03-01
User Input: Can't Delete Search Terms with Keyboard


ivankruchkoff:
## Bug Description

When I fill out search terms, I can't then delete them with the backspace/delete key.

## Steps to reproduce

1. Go to user input flow
2. Get to 5/5 search terms
3. Fill out some search terms
4. Try and delete with backspace
5. See that search terms don't get deleted

## Additional Context

 - PHP Version: 
 - Plugin Version [e.g. 22]

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* When pressing the backspace key within a search term (i.e. cursor is within the search term bubble), it should delete characters from that search term.
* When pressing the backspace key when right after a search term (i.e. the cursor is outside a search term bubble, right after one), it should delete that entire search term.

## Implementation Brief
* Once #2842 has been merged, using `/assets/js/components/user-input/UserInputKeywords.js`,
	* import `BACKSPACE` from `@wordpress/keycodes`.
	* create new function `deleteKeyword` which does the same thing as `onKeywordDelete`.
	* update `onKeywordDelete` to use the `deleteKeyword` function.
	* update the `onKeyDown` function to:
		* get the length of `nonEmptyValues`
		* if the length is less than `3` and the `BACKSPACE` key is pressed, call the `deleteKeyword` function passing `nonEmptyValuesLength - 1` as the index to delete.
		* set the focus to the input text again using the same logic when the `ENTER` or `COMMA` keys are pressed, but this time, having the following selector: `#${ slug }-keyword-${ nonEmptyValuesLength - 1 }`

### Test Coverage
N/A

### Visual Regression Changes
N/A

## QA Brief

* Fill out the user input flow until you reach step 5 (search terms)
* You should be able to:
  * Enter up to 3 words, pressing comma or enter/ return to enter a word
  * Delete within a word you are typing (i.e. hitting backspace while typing a word should delete the character you have just typed)
  * Delete a term after entering it (e.g. type `test`, hit comma to enter it, then `testing`, hitting comma again. Now hit the the backspace and `testing` should be removed, but not `test`).

## Changelog entry

* Fix accessibility problem where it wasn't possible to delete user input search terms using the keyboard.

tofumatt:
IB largely looks good but one thing I'm not clear on:

[...]

PR #2964, opened 2021-03-15
User Input: Can't Delete Search Terms with Keyboard (#2900)

johnPhillips:
## Summary

Addresses issue #2900

## Relevant technical choices

My understanding of the AC was that the backspace behaviour essentially toggles between deleting the previous character and deleting the previous keyword:
- if in the middle of typing, it should delete the last character
- if in the middle of keywords (i.e. one just got added via comma or enter/return, or one just got deleted) then it should delete the last keyword.

The strategy in the IB got most of the way there but missed a few edge cases, so I added a `useState` to keep track of the backspace behaviour. 

## Checklist


assets/js/components/user-input/UserInputKeywords.js
```diff
@@ -85,30 +86,59 @@ export default function UserInputKeywords( { slug, max } ) {
...
-			setTimeout( () => {
-				const input = keywordsContainer.current.querySelector( `#${ slug }-keyword-${ index + 1 }` );
-				if ( input ) {
-					input.focus();
-				}
-			}, 50 );
+			setCanDeleteKeyword( true ); // New keyword has been added, so hitting backspace now will remove it
+			setTimeout( () =>
+				focusInput( `#${ slug }-keyword-${ index + 1 }` )
+			, 50 );
```
tofumatt, line 103: This code style is a bit awkward; do we _need_ to be returning `focusInput()` here? I don't think the return value here is needed, so seems like this is using this syntax to save on curly braces but not because it needs to return.

Let's be clearer with:
```suggestion
			setTimeout( () => {
				focusInput( `#${ slug }-keyword-${ index + 1 }` );
			}, 50 );
```

Sure, it's more characters but I think expresses the intent better. Whenever I see an implicit return I assume the return value matters—because it doesn't here let's avoid it.

Same goes for other bits of this PR 🙂

assets/js/components/user-input/UserInputKeywords.js
```diff
@@ -85,30 +86,59 @@ export default function UserInputKeywords( { slug, max } ) {
...
+		}
+
+		if ( keyCode === BACKSPACE && canDeleteKeyword ) {
+			deleteKeyword( nonEmptyValuesLength - 1 );
+			setTimeout( () =>
+				focusInput( `#${ slug }-keyword-${ nonEmptyValuesLength - 1 }` )
+			, 50 );
```
tofumatt, line 110:
```suggestion
			setTimeout( () => {
				focusInput( `#${ slug }-keyword-${ nonEmptyValuesLength - 1 }` );
			}, 50 );
```

assets/js/components/user-input/UserInputKeywords.js
```diff
@@ -85,30 +86,59 @@ export default function UserInputKeywords( { slug, max } ) {
...
+		}
+
+		if ( target.value.length <= 1 && keyCode === BACKSPACE ) {
+			// user deleted the only character, or just landed in this input after
+			// deleting a keyword so hitting backspace again should delete the previous keyword
```
tofumatt, line 115: Nitpick, but these comments are sentences, so please start them with a capital letter. 🙂 
```suggestion
			// User deleted the only character, or just landed in this input after
			// deleting a keyword so hitting backspace should delete the previous keyword.
```

assets/js/components/user-input/UserInputKeywords.js
```diff
@@ -85,30 +86,59 @@ export default function UserInputKeywords( { slug, max } ) {
...
+			setCanDeleteKeyword( true );
+		} else {
+			// user is typing, so backspace should delete the last character
```
tofumatt, line 118: Similarly, this should end with a period.
```suggestion
			// User is typing a keyword, so backspace should delete the last character,
			// not the entire keyword.
```

assets/js/components/user-input/UserInputKeywords.js
```diff
@@ -85,30 +86,59 @@ export default function UserInputKeywords( { slug, max } ) {
...
 				...values.slice( index + 1 ),
 			] );
 
-			setTimeout( () => {
-				const input = keywordsContainer.current.querySelector( `#${ slug }-keyword-${ index + 1 }` );
-				if ( input ) {
-					input.focus();
-				}
-			}, 50 );
+			setCanDeleteKeyword( true ); // New keyword has been added, so hitting backspace now will remove it
```
tofumatt, line 100:
```suggestion
			// A new keyword has been added. Pressing backspace will remove the entire keyword.
			setCanDeleteKeyword( true );
```

assets/js/components/user-input/UserInputKeywords.js
```diff
@@ -85,30 +86,59 @@ export default function UserInputKeywords( { slug, max } ) {
...
 		}
-	}, [ keywordsContainer.current, ...dependencies ] );
+	}, [ keywordsContainer.current, ...dependencies, canDeleteKeyword ] );
+
+	const focusInput = ( querySelector ) => {
+		const input = keywordsContainer.current.querySelector( querySelector );
+		if ( input ) {
+			input.focus();
+		}
+	};
```
tofumatt, line 128: It's a bit confusing that this is defined after it is referenced above. I know it's technically okay and such—it isn't used until later.

But if we can move this up it might make it a bit easier to mentally parse is all. 🤷🏻‍♂️

assets/js/components/user-input/UserInputKeywords.js
```diff
@@ -85,30 +86,59 @@ export default function UserInputKeywords( { slug, max } ) {
...
+		deleteKeyword( index );
+	}, dependencies );
+
+	const deleteKeyword = useCallback( ( index ) => {
 		updateKeywords( [
 			...values.slice( 0, index ),
 			...values.slice( index + 1 ),
 		] );
-	}, dependencies );
+		// After deleting a keyword, hitting backspace will delete the next keyword
```
tofumatt, line 139:
```suggestion
		// After deleting a keyword, hitting backspace will continue to delete keywords.
		// Keywords are removed from the end of the list when Backspace is used.
```

tofumatt (changes requested):
Mostly this is good, but this doesn't appear to behave _exactly_ as-intended and as spelled out in the ACs:



Seems like if you focus the input (either via keyboard input as done in the above GIF, or with the mouse), the first Backspace key press doesn't trigger the keyword delete action. Everything after that works as-intended, so seems like there's a bug there in the state management for when `canDeleteKeyword` is set.

assets/js/components/user-input/UserInputKeywords.js
johnPhillips: I agree - I've moved all these declarations higher up.

assets/js/components/user-input/UserInputKeywords.js
```diff
@@ -55,6 +56,26 @@ export default function UserInputKeywords( { slug, max } ) {
...
+			input.focus();
+		}
+	};
+
+	const deleteKeyword = useCallback( ( index ) => {
+		updateKeywords( [
+			...values.slice( 0, index ),
+			...values.slice( index + 1 ),
+		] );
+		// After deleting a keyword, hitting backspace will delete the next keyword
```
tofumatt, line 71:
```suggestion
		// After deleting a keyword, hitting backspace will delete the next keyword.
```

assets/js/components/user-input/UserInputKeywords.js
```diff
@@ -86,30 +107,36 @@ export default function UserInputKeywords( { slug, max } ) {
...
+		const nonEmptyValuesLength = nonEmptyValues.length;
+
 		if ( ( keyCode === ENTER || keyCode === COMMA ) && nonEmptyValues.length < max ) {
 			updateKeywords( [
 				...values.slice( 0, index + 1 ),
 				'',
 				...values.slice( index + 1 ),
 			] );
 
+			setCanDeleteKeyword( true ); // A new keyword has been added. Pressing backspace now will remove the entire keyword.
```
tofumatt, line 121:
```suggestion
			// A new keyword has been added. Pressing backspace now will remove the entire keyword.
			setCanDeleteKeyword( true );
```

tofumatt (dismissed):
Looks great—tested locally and this behaves as-expected. Thanks! 👍🏻

tofumatt:
@googlebot I consent.

tofumatt (approved):

Issue #2907, opened 2021-03-02
User Input: "Other" choice not required when multiple


aaemnnosttv:
## Bug Description

Steps 3 and 4 of the user input flow allow multiple choices with an option to provide an "other" value. The  single-option steps that offer an option to provide "Other" all require a non-empty, non-whitespace value to be provided.

## Steps to reproduce

1. Enable `userInput` feature
1. Start user input flow
2. On step 3 or 4 select one less than the required number of choices (excluding "Other")
3. Select "Other", without providing a value for it
4. See "Next" button is enabled

## Screenshots

![image](https://user-images.githubusercontent.com/1621608/109695129-3c2eac00-7b94-11eb-8be1-0d8481815cbd.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Entering a text alongside the "Other" choice in a User Input question should _always_ be required when the radio button/checkbox is enabled in order to proceed, for every question.

## Implementation Brief

* edit `assets/js/components/user-input/UserInputQuestionWrapper.js`
* modify the `disabled` prop of the next `Button` to check if any of the values in the `values` array are empty and disable the button if there is. Specifically, https://github.com/google/site-kit-wp/blob/f3eea273ba67e3319457f8b22601461e3813954b/assets/js/components/user-input/UserInputQuestionWrapper.js#L101 can be changed to `disabled={ values.some( ( value ) => value.trim().length === 0 ) }`

### Test Coverage

* 

### Visual Regression Changes

* 

## QA Brief

* Check the following on the multiple choice questions 3 and 4 of the user input flow:
  * The next button is disabled when other is selected but no text has been entered in the other text field.
* Check that the next button is not disabled on the 5th question when only one or two search terms are entered.

## Changelog entry

* Ensure entering a text is required when choosing the "Other" option for a question in the user input flow.

felixarntz:
@aaemnnosttv Did we decide here that entering something should _not_ be required? Not sure I recall; there's definitely the issue of visibility of the input, but I'm not sure about this one here.

aaemnnosttv:
@felixarntz you'll have to check with @marrrmarrr. I'm also not sure but it's similar/related to #2897 in that it allows an empty value where others do not. So one way or the other it's inconsistent.

felixarntz:
@aaemnnosttv Talked through this with @marrrmarrr: Entering something here should always be required.

tofumatt:
IB ✅

[...]

PR #2965, opened 2021-03-15
Update next button logic so that it is disabled if other is selected but the other text input is empty.

Hazlitte:
## Summary

Update next button logic so that it is disabled if other is selected but the other text input is empty.

Addresses issue #2907 

## Checklist


felixarntz:
@Hazlitte Is this ready for review? (Please mark as such)

assets/js/components/user-input/UserInputQuestionWrapper.js
```diff
@@ -98,7 +98,7 @@ export default function UserInputQuestionWrapper( props ) {
 								<Button
 									className="googlesitekit-user-input__buttons--next"
 									onClick={ next }
-									disabled={ values.filter( ( value ) => value.trim().length > 0 ).length === 0 }
+									disabled={ values.some( ( value ) => value.trim().length === 0 ) || values.length === 0 }
```
aaemnnosttv, line 101: Generally speaking, we should always check the `length` (or other property) first before using a collection check to avoid unnecessary iterations. In this case, it doesn't really matter since there are no iterations to perform on an empty list but something to keep in mind as a general order of operations 👍

aaemnnosttv (approved):
LGTM 👍

aaemnnosttv:
@googlebot I consent.

aaemnnosttv (changes requested):
Looks like this change has broken E2E tests for user input.

See `tests/e2e/specs/user-input-questions.test.js`

aaemnnosttv (changes requested):
Looks like this change has broken E2E tests for user input.

See `tests/e2e/specs/user-input-questions.test.js`

eugene-manuilov:
@googlebot I consent.

assets/js/components/user-input/UserInputQuestionWrapper.js
```diff
@@ -51,12 +51,17 @@ export default function UserInputQuestionWrapper( props ) {
...
 	} = props;
 
 	const values = useSelect( ( select ) => select( CORE_USER ).getUserInputSetting( slug ) || [] );
 	const scope = useSelect( ( select ) => select( CORE_USER ).getUserInputSettingScope( slug ) );
 	const author = useSelect( ( select ) => select( CORE_USER ).getUserInputSettingAuthor( slug ) );
 
+	const nextDisabled = allowEmptyValues
+		? values.filter( ( value ) => value.trim().length > 0 ).length === 0
+		: values.length === 0 || values.some( ( value ) => value.trim().length === 0 );
+
```
aaemnnosttv, line 64: The changes in this component seem unnecessary. The tests still pass with `allowEmptyValues` set to `true` or `false`. Both branches of the condition seem like they would result in the same result.

tests/e2e/specs/user-input-questions.test.js
```diff
@@ -76,7 +76,9 @@ describe( 'User Input Settings', () => {
 			'enter keywords',
 			async () => {
 				await expect( page ).toClick( '.googlesitekit-user-input__buttons--next' );
-				await expect( page ).toFill( '#searchTerms-keyword-0', 'One,Two,Three,' );
+				await expect( page ).toFill( '#searchTerms-keyword-0', 'One' );
+				await expect( page ).toFill( '#searchTerms-keyword-1', 'Two' );
+				await expect( page ).toFill( '#searchTerms-keyword-2', 'Three' );
```
aaemnnosttv, line 81: I think this is likely the only additional change needed on the original PR.

aaemnnosttv (changes requested):
Thanks for the helping hand @eugene-manuilov – it looks like there may be some unnecessary additions here.
See my comments below.

assets/js/components/user-input/UserInputQuestionWrapper.js
eugene-manuilov: Tests work because we enter all three non-empty keywords. If you comment out one of the keywords, tests will fail.

The problem with the initial approach was that it disabled the Next button on the keywords screen because there was an empty string in the values (which is a possible case for that screen). This "allowEmptyValues" property fixes this problem.

tests/e2e/specs/user-input-questions.test.js
eugene-manuilov: Not the only one, see my comment above. Try to revert `allowEmptyValues` changes and comment out one of these lines. Tests will start failing.

assets/js/components/user-input/UserInputQuestionWrapper.js
aaemnnosttv: Okay, got it; thanks for clarifying 👍

aaemnnosttv: @eugene-manuilov please add a comment to explain here as we discussed, then this should be good to go 👍

eugene-manuilov: @aaemnnosttv a comment is added.

assets/js/components/user-input/UserInputQuestionWrapper.js
```diff
@@ -51,12 +51,29 @@ export default function UserInputQuestionWrapper( props ) {
...
+	//   2. When a user is on the last 5th screen (search terms) - we SHOULD NOT disable the next
+	//      button if at least one search term is entered. In other words we should disable the next
+	//      button only when all values are empty strings.
+	//
+	const hasInvalidValues = allowEmptyValues
+		// Consider the values array invalid if it contains all values empty.
+		? values.filter( ( value ) => value.trim().length > 0 ).length === 0
+		// Consider the values array invalid if it contains at least one value that is empty.
+		: values.some( ( value ) => value.trim().length === 0 );
+
```
aaemnnosttv, line 76: This is great, thanks for the elaborate detail 👍👍👍👍

aaemnnosttv (approved):
Awesome!

PR #2970, opened 2021-03-16
Compressed Size Action configuration

aaemnnosttv:
## Summary

Addresses issue #2539 (follow-up)

## Relevant technical choices

* Moves the compressed size check to its own workflow as it currently only works for PRs
    * Included built CSS files in report
    * Scoped workflow to only run when source assets or config files change
    * Added condition to only run on non-drafts
    * This makes sense due to the runtime of this job which runs NPM install and the build two times: once for the current branch and once for the base branch to generate a diff
* Scopes the analyzed files to `dist/assets` only (this excludes some insignificant files in the dist root such as JS files generated for scss inputs)
* Adds `strip-hash` configuration to allow comparing of built files since hashes will always differ
* Configures hash to be replaced by asterisks rather than be completely stripped to differentiate between files that have a hash and those that don't

## Checklist

### DIFF
--- a/.github/workflows/compressed-size.yml
+++ b/.github/workflows/compressed-size.yml
@@ -1,0 +1,32 @@
+name: Compressed Size
+
+on:
+  pull_request:
+    branches:
+      - develop
+      - master
+    types:
+      - opened
+      - reopened
+      - synchronize
+    paths:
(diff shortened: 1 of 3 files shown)


felixarntz (approved):
Awesome, thanks!

Issue #2901, opened 2021-03-01
User Input: Add label for keyword input


ivankruchkoff:
## Feature Description

The keywords `input` has no label because it renders the `label` attribute on the `TextField` as a `placeholder` HTML attribute. This isn't accessible, so let's add a label.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Because the `label` prop supplied to the `TextField` component here is rendered as a `placeholder`, there should also be a manually-created `<label>` added for the `TextField` for assistive technology. Ensure there is always a `<label>` tag for the `TextField` with the text `"Keywords (minimum one, maximum three)"`.

## Implementation Brief

*  edit `assets/js/components/user-input/UserInputKeywords.js`
*  add a `<label>` to the `TextField` with the text `"Keyword"` and hide it with the `VisuallyHidden` component.

### Test Coverage

* No tests needed for this change.

### Visual Regression Changes

* VRT changes to the placeholder being empty after a value is entered might occur; otherwise none should occur.

## QA Brief

* Check that the term text fields have hidden labels

## Changelog entry

* Introduce hidden labels for search term fields of the User Input Settings flow.

tofumatt:
While the label for this text input isn't _great_, it is instructive for screenreaders and shouldn't be removed as it then lacks context, especially because we shift focus to it after creating a new keyword.

Instead we should move the label into the top of the text field as many component do elsewhere for `TextField`/`Input` components, like "Text Field" in this screenshot:

<img width="380" alt="Screenshot 2021-03-11 at 20 36 30" src="https://user-images.githubusercontent.com/90871/110851821-78a38b80-82a9-11eb-90ee-68a4ecb268c4.png">

----

Failing that, though, we should use a `className` to _visually_ hide the label rather than delete its text, so it's still usable for assistive technologies.

Hazlitte:
@tofumatt On further investigation at present the input fields don't have a label at all.  When the prop `noLabel` is passed to the `TextField` component the `label` prop text is rendered as a `placeholder`.  We could remove the `noLabel` prop, remove the conditional setting of the label text and set the `floatingLabelClassName` prop of the `TextField` to `screen-reader-text` if no text has been entered.

I also noted that `display: inline-flex` is missing from the input wrapper which is causing it to render below entered keywords.  I'm not sure if this is intended behaviour.
[...]

PR #2984, opened 2021-03-18
Add hidden label to user input search terms.

Hazlitte:
## Summary

Add hidden label to user input search terms.

Addresses issue #2901 

## Checklist

### DIFF
--- a/assets/js/components/user-input/UserInputKeywords.js
+++ b/assets/js/components/user-input/UserInputKeywords.js
@@ -26,7 +26,7 @@ import classnames from 'classnames';
  * WordPress dependencies
  */
 import { useCallback, useRef } from '@wordpress/element';
-import { __ } from '@wordpress/i18n';
+import { __, sprintf } from '@wordpress/i18n';
 import { ENTER } from '@wordpress/keycodes';
 
 /**
@@ -122,16 +122,16 @@ export default function UserInputKeywords( { slug, max } ) {
 						} ) }
 					>
 						<VisuallyHiden>
-							<label
-								htmlFor={ `${ slug }-keyword-${ i }` }
-							>
-								{ __( 'Enter minimum one (1), maximum three (3) terms', 'google-site-kit' ) }
+							<label htmlFor={ `${ slug }-keyword-${ i }` } >
+								{ sprintf(
+									/* translators: %s is the keyword number; 1, 2, or 3 */
+									__( 'Keyword %s', 'google-site-kit' ),
+									i + 1, // Keys are zero-indexed; this starts keyword at "1".
+								) }
 							</label>
 						</VisuallyHiden>
 						<TextField
-							label={ values.length === 1 && values[ 0 ] === ''
-								? __( 'Enter minimum one (1), maximum three (3) terms', 'google-site-kit' )
-								: '' }
+							label={ __( 'Enter minimum one (1), maximum three (3) terms', 'google-site-kit' ) }
 							noLabel
 						>
 							<Input
--- a/assets/js/components/user-input/UserInputKeywords.js
+++ b/assets/js/components/user-input/UserInputKeywords.js
@@ -120,8 +120,16 @@ export default function UserInputKeywords( { slug, max } ) {
 							'googlesitekit-user-input__text-option': values.length > i + 1 || value.length > 0,
 						} ) }
 					>
+						<label
+							htmlFor={ `${ slug }-keyword-${ i }` }
+							className="screen-reader-text"
+						>
+							{ __( 'Enter minimum one (1), maximum three (3) terms', 'google-site-kit' ) }
+						</label>
 						<TextField
-							label={ i + 1 === values.length ? __( 'Enter minimum one (1), maximum three (3) terms', 'google-site-kit' ) : '' }
+							label={ values.length === 1 && values[ 0 ] === ''
+								? __( 'Enter minimum one (1), maximum three (3) terms', 'google-site-kit' )
+								: '' }
 							noLabel
 						>
 							<Input
--- a/assets/js/components/user-input/UserInputKeywords.js
+++ b/assets/js/components/user-input/UserInputKeywords.js
@@ -113,6 +113,12 @@ export default function UserInputKeywords( { slug, max } ) {
 	return (
 		<Cell lgStart={ 6 } lgSize={ 6 } mdSize={ 8 } smSize={ 4 }>
 			<div ref={ keywordsContainer } className="googlesitekit-user-input__text-options">
+				<label
+					htmlFor={ `${ slug }-keyword-0` }
+					className="screen-reader-text"
+				>
+					{ __( 'Enter minimum one (1), maximum three (3) terms', 'google-site-kit' ) }
+				</label>
 				{ values.map( ( value, i ) => (
 					<div
 						key={ i }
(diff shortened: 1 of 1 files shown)


assets/js/components/user-input/UserInputKeywords.js
```diff
@@ -120,8 +120,16 @@ export default function UserInputKeywords( { slug, max } ) {
 							'googlesitekit-user-input__text-option': values.length > i + 1 || value.length > 0,
 						} ) }
 					>
+						<label
+							htmlFor={ `${ slug }-keyword-${ i }` }
+							className="screen-reader-text"
+						>
+							{ __( 'Enter minimum one (1), maximum three (3) terms', 'google-site-kit' ) }
+						</label>
```
eugene-manuilov, line 128: @Hazlitte let's move this label out of the `values.map` callback because currently, it generates three identical labels with the same text which might confuse users that use screen readers. Just move it above the `values.map` section and I believe it should be good to go.

eugene-manuilov (changes requested):

assets/js/components/user-input/UserInputKeywords.js
Hazlitte: @eugene-manuilov Ok.

assets/js/components/user-input/UserInputKeywords.js
```diff
@@ -121,7 +127,9 @@ export default function UserInputKeywords( { slug, max } ) {
 						} ) }
 					>
 						<TextField
-							label={ i + 1 === values.length ? __( 'Enter minimum one (1), maximum three (3) terms', 'google-site-kit' ) : '' }
+							label={ values.length === 1 && values[ 0 ] === ''
+								? __( 'Enter minimum one (1), maximum three (3) terms', 'google-site-kit' )
+								: '' }
```
eugene-manuilov, line 132: @tofumatt do we really need to hide this placeholder? I think it's quite helpful as is especially after all modifications in the keywords component.

eugene-manuilov: The text field becomes invisible, so I think we need to revert this. 
google github io_site-kit-wp_storybook_pull_2984__path=_story_user-input--userinputapp


https://google.github.io/site-kit-wp/storybook/pull/2984/?path=/story/user-input--userinputapp

tofumatt: Oh, agreed actually. Especially now that the input often spans multiple lines the removal of both the placeholder _and_ the blue underline makes it look both spartan and unclear.

I mean, the original issue was when everything was on one line, eg:

Screen Shot 2021-02-17 at 8 44 00 PM

Maybe now it's best if we leave this alone and mark the bug as won't fix.

Thoughts @felixarntz and @ivankruchkoff (as the original issue creator)?

tofumatt (commented):
The original issue was when everything was on one line, eg:

Screen Shot 2021-02-17 at 8 44 00 PM

But as mentioned in my review, this might not be appropriate to do now that the input spans multiple lines. @eugene-manuilov pointed out it's pretty awkward: https://github.com/google/site-kit-wp/pull/2984/files#r598668340

I think we should close out this issue as a won't fix.

assets/js/components/user-input/UserInputKeywords.js
```diff
@@ -113,6 +113,12 @@ export default function UserInputKeywords( { slug, max } ) {
 	return (
 		<Cell lgStart={ 6 } lgSize={ 6 } mdSize={ 8 } smSize={ 4 }>
 			<div ref={ keywordsContainer } className="googlesitekit-user-input__text-options">
+				<label
+					htmlFor={ `${ slug }-keyword-0` }
```
tofumatt, line 117: This will always generate a label for the first keyword, which means this will be a label linked to no element after the first keyword is entered. This label should be generated inside the `values.map` next to the `<TextField><Input /></TextField>` it is for, because the ID changes based on the index.

Also, we have a `VisuallyHidden` component I'd prefer we use over the CSS class. Building as many things out of components is more React-y. Admittedly it's sort of a preference, but it's why we have the `VisuallyHidden` component. So let's use that instead 🙂

tofumatt (changes requested):
Let's never hide this placeholder as mentioned in a comment, but still implement the visually hidden label for accessibility.

I amended the ACs/IBs for the issue as well based on this comment: https://github.com/google/site-kit-wp/issues/2901#issuecomment-806010206

assets/js/components/user-input/UserInputKeywords.js
eugene-manuilov: @tofumatt if we want to generate that label for every keyword, we need to update label text because otherwise screen readers will read the same label text triple times which won't help users with disabilities too much.

tofumatt: True, let's do that. Having only one and linking it with a specific index would mean the other inputs don't have useful labels still though.

That or we change the `htmlFor` based on the input, but having a label for each strikes me as better…

assets/js/components/user-input/UserInputKeywords.js
```diff
@@ -120,8 +121,17 @@ export default function UserInputKeywords( { slug, max } ) {
...
+								htmlFor={ `${ slug }-keyword-${ i }` }
+							>
+								{ __( 'Enter minimum one (1), maximum three (3) terms', 'google-site-kit' ) }
+							</label>
+						</VisuallyHiden>
 						<TextField
-							label={ i + 1 === values.length ? __( 'Enter minimum one (1), maximum three (3) terms', 'google-site-kit' ) : '' }
+							label={ values.length === 1 && values[ 0 ] === ''
+								? __( 'Enter minimum one (1), maximum three (3) terms', 'google-site-kit' )
+								: '' }
```
tofumatt, line 134: This change should be removed; it's no longer part of the ACs/IB as we decided it isn't needed anymore. See: https://github.com/google/site-kit-wp/pull/2984/files#r598867190

assets/js/components/user-input/UserInputKeywords.js
```diff
@@ -120,8 +121,17 @@ export default function UserInputKeywords( { slug, max } ) {
 							'googlesitekit-user-input__text-option': values.length > i + 1 || value.length > 0,
 						} ) }
 					>
+						<VisuallyHiden>
+							<label
+								htmlFor={ `${ slug }-keyword-${ i }` }
+							>
+								{ __( 'Enter minimum one (1), maximum three (3) terms', 'google-site-kit' ) }
+							</label>
```
tofumatt, line 129: @eugene-manuilov pointed out that this means each keyword input gets the same label, which isn't super-helpful. See: https://github.com/google/site-kit-wp/pull/2984#discussion_r601467779

Let's go with something like:

```js
<label htmlFor={ `${ slug }-keyword-${ i }` }>
	{ sprintf( __( 'Keyword %s', 'google-site-kit' ), i ) }
</label>
```

tofumatt (changes requested):
This still needs the empty label change reverted, and @eugene-manuilov made a good point about this label text being the same for three different inputs. I made a suggestion to clear it up.

tofumatt:
@googlebot I consent.

eugene-manuilov (approved):
Awesome! Looks good to me.

Issue #2935, opened 2021-03-09
Fix OAuth callback logic to not run before WordPress login redirect


felixarntz:
## Bug Description

It is currently possible to access the OAuth callback URL on any site to trigger Site Kit's logic to handle it. While this is not a real security flaw since the necessary permission checks are in place, it is still concerning, and more importantly, it can result in a user-facing problem in some certain circumstances, e.g. when for some reason being logged out while going through the Site Kit setup flow.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* When accessing the `{adminURL}/index.php?oauth2callback=1` URL on a site where you're not logged in, you should get redirected to `wp-login.php`, like it is expected to happen for admin URLs.
* _Any_ Site Kit handler logic for specific query parameters (i.e. also `googlesitekit_connect`, `googlesitekit_disconnect`) should be run at a hook late enough so that it is only reached if the user is already logged in.

## Implementation Brief

**Allow authentication urls for logged in users only**
Change `handle_oauth` hook to run on `admin_init` in
https://github.com/google/site-kit-wp/blob/2bac88b/includes/Core/Authentication/Authentication.php#L268

Remove the is_admin() call in `handle_oauth`
https://github.com/google/site-kit-wp/blob/2bac88b/includes/Core/Authentication/Authentication.php#L633-L635

**Allow logout after starting authentication**
If we logout during authentication, when we first go to the google oauth screen, all of our WordPress nonce verification will fail due to https://core.trac.wordpress.org/browser/tags/5.7/src/wp-includes/pluggable.php#L2187. To get around this we can create a custom nonce verification approach:

```php
function get_nonce() {
    // Checks a user option: googlesitekit_proxy_nonce
    // If an option exists it is verified with wp_verify_nonce(), if it returns false, or no nonce existed, one is generated and saved using `wp_create_nonce( self::ACTION_SETUP )` 
}
```

```php
function verify_nonce( $nonce ) {
    // Verify the supplied option against the option value, returning true if they are identical, false otherwise.
}
```

```php
function delete_nonce() {} // delete the option if it exists
```

Update nonce verification function
https://github.com/google/site-kit-wp/blob/2bac88b/includes/Core/Authentication/Authentication.php#L1040-L1051

Update the nonce from url:
https://github.com/google/site-kit-wp/blob/2bac88b/includes/Core/Authentication/Google_Proxy.php#L113

Delete the nonce option after we no longer need it:
[...]

PR #2996, opened 2021-03-19
Feature/2935 oauth callback logic

ivankruchkoff:
## Summary

Addresses issue #2935

## Checklist


includes/Core/Authentication/Authentication.php
```diff
@@ -1219,7 +1216,7 @@ private function get_proxy_setup_url() {
 		return add_query_arg(
 			array(
 				'action' => Google_Proxy::ACTION_SETUP,
-				'nonce'  => wp_create_nonce( Google_Proxy::ACTION_SETUP ),
+				'nonce'  => $this->google_proxy->get_nonce(),
```
felixarntz, line 1219: See @aaemnnosttv's comment on the issue: The nonce here should be a regular WordPress nonce (i.e. it shouldn't be using the new persistent approach). Only the actual proxy URL (`https://sitekit.withgoogle.com/site-management/setup/...`) should receive that "more persistent" proxy nonce.

includes/Core/Authentication/Authentication.php
```diff
@@ -1045,7 +1042,7 @@ private function get_authentication_oauth_error_notice() {
 	private function verify_proxy_setup_nonce() {
 		$nonce = $this->context->input()->filter( INPUT_GET, 'nonce', FILTER_SANITIZE_STRING );
 
-		if ( ! wp_verify_nonce( $nonce, Google_Proxy::ACTION_SETUP ) ) {
+		if ( ! $this->google_proxy->verify_nonce( $nonce ) ) {
```
felixarntz, line 1045: We might need to update this logic here:
* If the request comes from Site Kit, it is to redirect to the proxy. In that case, the nonce should be a regular WordPress nonce.
* Otherwise, the request comes from the proxy in which case it should be a "proxy nonce".

Handling these different requests in the same callback is arguably a bit confusing, we may wanna decouple them via a separate request. In that case, the proxy one should be kept intact. However, for now it might be okay to just try both? If `wp_verify_nonce` fails, try proxy nonce verification. With this we should be set up for both scenarios.

felixarntz (changes requested):
This looks good for the most part, but as @aaemnnosttv mentioned there is a bit of (understandable) confusion with the nonces - only the proxy itself should use these "proxy nonces".

aaemnnosttv (commented):
LGTM – we just need to update the E2E version in `tests/e2e/plugins/oauth-callback.php`

I'll push this up since it's a super simple change.

aaemnnosttv:
@googlebot I consent.

aaemnnosttv (approved):
👍

felixarntz (approved):
Great!

Issue #2944, opened 2021-03-10
Implement GA tracking events for All Traffic feature tooltip tour (and tooltip tours in general)


felixarntz:
Feature tour should send GA tracking events for interaction with them (for users that have opted in).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Every feature tour should include its event category for GA tracking in its definition. For now, the only tour that exists is `allTrafficWidget`; it should use `all_traffic_widget` for its GA event category (for parity with other tracking events for the feature).
* The tooltip tour components should trigger the following GA tracking events to be sent:
    * `feature_tooltip_view` (for every tooltip displayed; event label should be the current tooltip step, starting at index 1)
    * `feature_tooltip_dismiss` (when dismissing via the X button; event label should be the current tooltip step, starting at index 1)
    * `feature_tooltip_advance` (when clicking the "Next" button; event label should be the current tooltip step, starting at index 1)
    * `feature_tooltip_return` (when clicking the "Back" button; event label should be the current tooltip step, starting at index 1)
    * `feature_tooltip_complete` (when clicking the "Got it" button; event label should be the current tooltip step, starting at index 1)

## Implementation Brief

* Using `TourTooltips` component, import the `trackEvent` function from the `assets/js/util/tracking/index.js` file;
* Add a new property `gaEventCategory` to the TourTooltips component and mark it required;
* Update `handleJoyrideCallback` function in the TourTooltips component to track the following events using `gaEventCategory` property as actual event category value and `data.index + 1` value as label:
    * `feature_tooltip_view` event when `data.type = EVENTS.TOOLTIP && data.lifecycle = LIFECYCLE.TOOLTIP`;
    * `feature_tooltip_dismiss` event when `data.action = ACTIONS.CLOSE && data.lifecycle = LIFECYCLE.COMPLETE`;
    * `feature_tooltip_advance` event when `data.action = ACTIONS.NEXT && data.lifecycle = LIFECYCLE.COMPLETE`;
    * `feature_tooltip_return` event when `data.action = ACTIONS.PREV && data.lifecycle = LIFECYCLE.COMPLETE`;
    * `feature_tooltip_complete` event when `data.status = STATUS.FINISHED && data.lifecycle = LIFECYCLE.COMPLETE`;
* Add `gaEventCategory` property with `all_traffic_widget` value to the `allTrafficWidget` object in the `assets/js/feature-tours/index.js` file;
[...]

PR #2999, opened 2021-03-22
Add event tracking for feature tours

aaemnnosttv:
## Summary

Addresses issue #2944

## Relevant technical choices

* Added tests for event tracking to the existing component tests as the logic proved to be a bit tricky

## Console logs (logging not included)

* Full tour, with back navigation, completing at end  
image
* Tour forward to step 2, back to 1, and then dismissing
image

## Checklist

### DIFF
--- a/assets/js/components/TourTooltips.js
+++ b/assets/js/components/TourTooltips.js
@@ -20,7 +20,7 @@
  * External dependencies
  */
 import { useMount } from 'react-use';
-import Joyride, { ACTIONS, EVENTS, STATUS } from 'react-joyride';
+import Joyride, { ACTIONS, EVENTS, LIFECYCLE, STATUS } from 'react-joyride';
 import PropTypes from 'prop-types';
 
 /**
@@ -33,8 +33,9 @@ import { __ } from '@wordpress/i18n';
  */
 import Data from 'googlesitekit-data';
 import { CORE_UI } from '../googlesitekit/datastore/ui/constants';
-import TourTooltip from './TourTooltip';
 import { CORE_USER } from '../googlesitekit/datastore/user/constants';
+import { trackEvent } from '../util/tracking';
+import TourTooltip from './TourTooltip';
 const { useSelect, useDispatch } = Data;
 
 /** For available options, see: {@link https://github.com/gilbarbara/react-joyride/blob/3e08384415a831b20ce21c8423b6c271ad419fbf/src/styles.js}. */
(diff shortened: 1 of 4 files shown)


assets/js/components/TourTooltips.js
```diff
@@ -96,6 +106,33 @@ export default function TourTooltips( { steps, tourID } ) {
...
+		// The index is 0-based, but step numbers are 1-based.
+		const stepNumber = index + 1;
+
+		if ( type === EVENTS.TOOLTIP && lifecycle === LIFECYCLE.TOOLTIP ) {
+			trackEvent( gaEventCategory, GA_ACTIONS.VIEW, stepNumber );
+		} else if ( status === STATUS.FINISHED && type === EVENTS.TOUR_END && size === stepNumber ) {
+			// Here we need to additionally check the size === stepNumber because
+			// it is the only way to differentiate the status/event combination
+			// from an identical combination that happens immediately after completion
+			// on index `0` to avoid duplicate measurement.
```
tofumatt, line 119: Thanks, that's helpful! 👍🏻

assets/js/components/TourTooltips.js
```diff
@@ -96,6 +106,33 @@ export default function TourTooltips( { steps, tourID } ) {
...
+			return;
+		}
+
+		if ( action === ACTIONS.CLOSE ) {
+			trackEvent( gaEventCategory, GA_ACTIONS.DISMISS, stepNumber );
+		} else if ( action === ACTIONS.PREV ) {
+			trackEvent( gaEventCategory, GA_ACTIONS.PREV, stepNumber );
+		} else if ( action === ACTIONS.NEXT ) {
+			trackEvent( gaEventCategory, GA_ACTIONS.NEXT, stepNumber );
+		}
```
tofumatt, line 133: This is pretty minor, but do these need to be `else if`s really? It can only ever be one.

I tend to prefer `if` statements over `switch/case`s, and this is fine, but I think technically a bunch of `if`s would work too 🤷🏻‍♂️

tofumatt (approved):
Awesome, the tests here are great and the comments super-helpful. One note about if/elses but it's fine. 👍🏻

assets/js/components/TourTooltips.js
aaemnnosttv: Good point; they obviously don't need to be, but I think it somewhat ended up like that when the logic for each was more complex and it was simplified later. I generally prefer a single dictionary look-up with 1 or no condition in cases like these 😄  That way the mapping doesn't even need to live in the component itself.

PR #2987, opened 2021-03-18
Migrate AdSense page "Performance by page" widget to use Widget API (2066)

johnPhillips:
## Summary

Addresses issue #2066

## Relevant technical choices
1. The IB for this one missed a few details on how to handle AdSense not linked and restricted metric error scenarios. I've had a stab at implementing it correctly here - basing the logic on the existing component and on `DashboardTopEarningPagesWidget.js` (which has already been refactored to use the Widgets API), but I am still getting my head around how all this works.
2. I also created a new `ModuleHeader` component which is the refactored analogue of `LayoutHeader`, but works with the `googlesitekit-widget...` CSS structure. I've put it in `modules/adsense/components/common` for now as suggested in the AC but as more widgets are refactored we might consider moving it to the `components` directory so they can all use it.

## Checklist

- [x] My code is tested and passes existing unit tests.
- [ ] My code has an appropriate set of unit tests which all pass.
- [x] My code is backward-compatible with WordPress 4.7 and PHP 5.6.
- [x] My code follows the [WordPress](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) coding standards.
- [x] My code has proper inline documentation.
- [ ] I have added a QA Brief on the issue linked above.
- [x] I have signed the Contributor License Agreement (see <https://cla.developers.google.com/>).

Issue #2064, opened 2020-09-18
Migrate Search Console page "Top search queries" widget to use Widget API


felixarntz:
The "Top search queries over the last x days" widget on the Search Console module page should be refactored as a new component that is registered and rendered using the Widget API.

The existing component should be untouched and only be renamed, giving it a `Legacy` prefix.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new component `ModulePopularKeywordsWidget` should be added in `assets/js/modules/search-console/components/module/ModulePopularPagesWidget.js`.
* It should be implemented as a functional, hook-based, datastore-driven widget component that renders the existing "Top search queries over the last x days" block on the Search Console module page, in a refactored way.
    * In other words, it should replicate what the legacy `LegacySearchConsoleDashboardWidgetKeywordTable` component displays.
    * Sub-components can be implemented as necessary. If reusable, they should go into `assets/js/modules/search-console/components/common`. Otherwise (if specific to just this widget), they should go into a new `assets/js/modules/search-console/components/module/ModulePopularKeywordsWidget` directory; in this case, the main widget component should go into the `index.js` file within that directory.
* The new component should be used to register a new widget in `assets/js/modules/search-console/index.js` (for testing, requires to enable the `widgets.moduleScreens` feature flag):
    * slug: `searchConsoleModulePopularKeywords`
    * width: `WIDGET_WIDTHS.FULL`
    * priority: `2`
    * wrapWidget: `false` (since it needs to render `Widget` manually to add the `Header` prop
    * widget area: `AREA_MODULE_SEARCH_CONSOLE_MAIN`

Hint: For the implementation it might be useful to look at `DashboardPopularKeywordsWidget` which renders something very similar, just with less columns and different header/footer in the dashboard.

## Implementation Brief
* Create a new file `assets/js/modules/search-console/components/module/ModulePopularPagesWidget.js` which will export the `ModulePopularPagesWidget`  functional component.
* `ModulePopularPagesWidget` should have the following props: `Widget`, `WidgetReportZero` and `WidgetReportError`, similar to what is done in `assets/js/modules/search-console/components/dashboard/DashboardPopularKeywordsWidget.js`.
* `ModulePopularPagesWidget` should take most of the code from `assets/js/modules/search-console/components/dashboard/LegacySearchConsoleDashboardWidgetKeywordTable.js` with the following changes:
[...]

PR #3002, opened 2021-03-22
Migrate Search Console page "Top search queries" widget to use Widget API (2064)

johnPhillips:
## Summary

Addresses issue #2064

## Relevant technical choices

These new widgets seem to all require a shared header component, so I've added it in `assets/js/modules/search-console/components/common/ModuleHeader.js`. 

Note that I have another PR open (https://github.com/google/site-kit-wp/pull/2987) where I've done the same thing, so if we agree that this component should be added, I'll need to briefly refactor one PR when the other is merged to avoid conflicts.

## Checklist


assets/js/modules/search-console/components/ModulePopularPagesWidget.js
```diff
@@ -0,0 +1,167 @@
...
+						/* translators: %s: number of days */
+						_n( 'Top search queries over the last %s day', 'Top search queries over last %s days', currentDayCount, 'google-site-kit', ),
+						currentDayCount,
+					) }
+					ctaLink={ baseServiceURL }
+					ctaLabel={ sprintf(
+						/* translators: %s: module name. */
+						__( 'See full stats in %s', 'google-site-kit' ),
+						_x( 'Search Console', 'Service name', 'google-site-kit' )
+					) }
```
felixarntz, line 110: There is some overlap with #3015 here which introduces the necessary CSS styles and two new `WidgetHeaderTitle` and `WidgetHeaderCTA` components that should be used instead. I already left a similar comment in https://github.com/google/site-kit-wp/pull/2918#discussion_r600648816, so maybe it would be best to take these two components and the related CSS changes out of #3015 and copy them into here, to unblock this.

assets/js/modules/search-console/components/common/ModuleHeader.js
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * External dependencies
+ */
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+import Link from '../../../../components/Link';
+
+function ModuleHeader( { title, ctaLabel, ctaLink } ) {
```
felixarntz, line 27: See above, this is a parallel effort to what's established in #3015 - apologies for not having flagged/noticed this as overlap. We can remove this component again and rely on the approach in #3015 which is quite simpler.

felixarntz (changes requested):

johnPhillips:
@felixarntz No problem - I guessed that there might be some duplication of effort going on, so was happy to make the changes. I've hopefully brought this in line with #3015 now.  👍

assets/js/modules/search-console/components/module/ModulePopularPagesWidget/index.js
```diff
@@ -0,0 +1,144 @@
...
+
+		return {
+			data: store.getReport( reportArgs ),
+			isLoading: ! store.hasFinishedResolution( 'getReport', [ reportArgs ] ),
+			error: store.getErrorForSelector( 'getReport', [ reportArgs ] ),
+		};
+	} );
+
+	if ( isLoading ) {
+		return <PreviewTable padding />;
```
felixarntz, line 75: It would be nice for the loading state here to still return it wrapped in the regular `Widget` including the `Header` (and `Footer`), so that the loading UI is displayed in the body of the regular widget and the header and CTA are already there. This is similar to how the new Analytics `ModuleOverviewWidget` does it.

assets/js/modules/search-console/components/module/ModulePopularPagesWidget/index.js
```diff
@@ -0,0 +1,144 @@
...
+
+	if ( isZeroReport( data ) ) {
+		return <WidgetReportZero moduleSlug="search-console" />;
+	}
+
+	return (
+		<Widget
+			noPadding
+			Header={ Header }
+		>
```
felixarntz, line 90: This is currently missing the footer that the current version of the widget has. It could be added via the `Footer` prop (see the Analytics `ModulePopularPagesWidget` for example).

stories/module-search-console-components.stories.js
```diff
@@ -2336,3 +2337,112 @@ generateReportBasedWidgetStories( {
 	wrapWidget: false,
 	setup,
 } );
+
+generateReportBasedWidgetStories( {
+	moduleSlugs: [ 'search-console' ],
+	datastore: STORE_NAME,
+	group: 'Search Console Module/Components/Module Page/ Popular Pages Widget',
```
felixarntz, line 2344: +1 for also adding a story!

@eugene-manuilov Could you maybe do this as a follow-up to #2067 and #2068 as well?

felixarntz (changes requested):
Excellent! Just two small things, then this is good to go!

assets/js/modules/search-console/components/module/ModulePopularPagesWidget/Header.js
```diff
@@ -0,0 +1,75 @@
...
+	} ) );
+	const dateRange = useSelect( ( select ) => select( CORE_USER ).getDateRange() );
+	const currentDayCount = getCurrentDateRangeDayCount( dateRange );
+
+	return (
+		<Fragment>
+			<WidgetHeaderTitle
+				title={ sprintf(
+					/* translators: %s: number of days */
+					_n( 'Overview for the last %s day', 'Overview for the last %s days', currentDayCount, 'google-site-kit', ),
```
felixarntz, line 59: Actually the label here is not correct. Please use the same one as the old/current widget.

assets/js/modules/search-console/components/module/ModulePopularPagesWidget/index.js
```diff
@@ -0,0 +1,144 @@
...
+import { CORE_USER } from '../../../../../googlesitekit/datastore/user/constants';
+import TableOverflowContainer from '../../../../../components/TableOverflowContainer';
+import ReportTable from '../../../../../components/ReportTable';
+import { isZeroReport } from '../../../util/is-zero-report';
+import { generateDateRangeArgs } from '../../../util/report-date-range-args';
+import Header from './Header';
+
+const { useSelect } = Data;
+
+function ModulePopularPagesWidget( { Widget, WidgetReportZero, WidgetReportError } ) {
```
felixarntz, line 47: The component needs to be renamed to `ModulePopularKeywordsWidget`, as defined in the ACs (also the directory similarly).

felixarntz (changes requested):
Just noticed two other naming / wording issues where it diverges from the legacy widget / ACs.

assets/js/modules/search-console/components/module/ModulePopularKeywordsWidget/index.js
```diff
@@ -0,0 +1,160 @@
...
+			noPadding
+			Header={ Header }
+			Footer={ () => (
+				<SourceLink
+					className="googlesitekit-data-block__source"
+					name={ _x( 'Search Console', 'Service name', 'google-site-kit' ) }
+					href={ serviceURL }
+					external
+				/>
+			) }
```
felixarntz, line 99: Let's move this into its own `Footer` component the same way that you've done with `Header`.

assets/js/modules/search-console/components/module/ModulePopularKeywordsWidget/index.js
```diff
@@ -0,0 +1,160 @@
...
+		};
+	} );
+
+	if ( error ) {
+		return <WidgetReportError moduleSlug="search-console" error={ error } />;
+	}
+
+	if ( isZeroReport( data ) ) {
+		return <WidgetReportZero moduleSlug="search-console" />;
+	}
```
felixarntz, line 86: It would be better if there were wrapped in the regular `Widget` container (including `Header` and `Footer`, but **without** `noPadding`, since the padding removal is only relevant for when rendering the data table itself).

Right now, this doesn't return the results within the widget container which looks odd. We have this problem in a couple widgets, but it's worth improving going forward. See for example the Analytics `ModulePopularPagesWidget` which does this correctly now.

assets/js/modules/search-console/components/module/ModulePopularKeywordsWidget/index.js
```diff
@@ -0,0 +1,160 @@
...
+					className="googlesitekit-data-block__source"
+					name={ _x( 'Search Console', 'Service name', 'google-site-kit' ) }
+					href={ serviceURL }
+					external
+				/>
+			) }
+		>
+			{
+
+				isLoading ? <PreviewTable padding />
```
felixarntz, line 103: Instead of having this conditional here, it would be better to handle the loading state in an early return, the same way you're doing it for the "error" and "zero data" states above. When doing so, please consider my other comment too (the `PreviewTable` should be wrapped in a `Widget` with `Header`, `Footer` and `noPadding`).

assets/js/modules/search-console/components/module/ModulePopularKeywordsWidget/index.js
```diff
@@ -0,0 +1,160 @@
...
+		title: __( 'Clicks', 'google-site-kit' ),
+		description: __( 'Number of times users clicked on your content in search results', 'google-site-kit' ),
+		Component: ( { row } ) => numFmt( row.clicks, { style: 'decimal' } ),
+	},
+	{
+		title: __( 'Impressions', 'google-site-kit' ),
+		description: __( 'Counted each time your content appears in search results', 'google-site-kit' ),
+		Component: ( { row } ) => numFmt( row.impressions, { style: 'decimal' } ),
+	},
+];
```
felixarntz, line 152: Although this was done in a similar way initially in related widgets, this entire array needs to be moved into the component; otherwise the strings might not get translated.

assets/js/modules/search-console/index.js
```diff
@@ -102,4 +103,16 @@ export const registerWidgets = ( widgets ) => {
 		},
 		CONTEXT_MODULE_SEARCH_CONSOLE,
 	);
+	widgets.registerWidget(
+		'searchConsoleModulePopularKeywords',
+		{
+			Component: ModulePopularKeywordsWidget,
+			width: [ widgets.WIDGET_WIDTHS.FULL ],
+			priority: 1,
```
felixarntz, line 111: This should be `2` per the ACs.

stories/module-search-console-components.stories.js
felixarntz, line 2344: Can you please remove the space in front of "Popular Pages Widget" here?

package-lock.json
```diff
@@ -39784,7 +39784,7 @@
       "dependencies": {
         "strip-ansi": {
           "version": "3.0.1",
-          "resolved": "https://registry.npmjs.org/strip-ansi/-/strip-ansi-3.0.1.tgz",
+          "resolved": "http://registry.npmjs.org/strip-ansi/-/strip-ansi-3.0.1.tgz",
```
felixarntz, line 39787: Can you revert the changes in this file? They seem to be unrelated.

felixarntz (changes requested):
This is almost good to go. A few last recommendations to simplify the code and fix some minor visual issues.

assets/js/modules/search-console/components/module/ModulePopularKeywordsWidget/Footer.js
```diff
@@ -0,0 +1,58 @@
...
+import { CORE_SITE } from '../../../../../googlesitekit/datastore/site/constants';
+import { CORE_USER } from '../../../../../googlesitekit/datastore/user/constants';
+import SourceLink from '../../../../../components/SourceLink';
+import { generateDateRangeArgs } from '../../../util/report-date-range-args';
+
+const { useSelect } = Data;
+
+export default function Footer() {
+	const { serviceURL } = useSelect( ( select ) => {
+		const url = select( CORE_SITE ).getCurrentEntityURL();
```
felixarntz, line 38: This entire component doesn't need awareness of any current entity URL, since there will never be one (the current entity is only relevant for widgets that are also used in the single URL view).

felixarntz (changes requested):
This is close, but there's a few more follow-up things I've noticed. Since those are quick to address though, in the interest of time I'll take care of it.

assets/js/modules/search-console/components/module/ModuleOverviewWidget/Stats.js
```diff
@@ -24,31 +24,19 @@ import PropTypes from 'prop-types';
...
-	let title = __( 'Search Traffic Summary', 'google-site-kit' );
-	if ( currentEntityTitle ) {
-		/* translators: %s: page title */
-		title = sprintf( __( 'Search Traffic Summary for %s', 'google-site-kit' ), decodeHTMLEntity( currentEntityTitle ) );
-	}
-
 	const options = {
 		chart: {
-			title,
+			title: __( 'Search Traffic Summary', 'google-site-kit' ),
```
felixarntz, line 39: The changes in this file are kind of unrelated, but I spotted this in merged code while noticing that this PR here originally handled the current entity which is generally unnecessary for module page widgets.

felixarntz (approved):
Great work! Made the last few changes, so this is ready to go.

Issue #2847, opened 2021-02-22
Implement tooltip for new contextual help feature


felixarntz:
The contextual help feature introduced via #2846 should receive a tooltip that highlights the addition to the screen.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new tooltip "tour" (with just a single step) should be added: It should appear on the dashboard if the user's initial Site Kit version is below 1.29.0 (TBD, depending on when we'll ship this feature), target the new help icon, and the copy for the only step should be as follows:
    * Heading: _Need help?_
    * Content: _Access the Site Kit docs or ask a question in our support forum from here._
    * As tooltip tours work by default, this should have a "Got it" button to dismiss the tooltip "tour".
    * The tooltip tour should use the slug `helpVisibility` internally.
    * The tooltip tour should only be shown if the `helpVisibility` feature flag is enabled.
    * See also [Figma](https://www.figma.com/file/pvsqJE5yZe9cFeNlTKUEiU/Google-SK-UX-Part-3-Enhancements10up?node-id=1408%3A475).
* Only one tooltip tour should be displayed at a time. If both the existing `allTrafficWidget` and the new `helpVisibility` tooltip tour are relevant, the `allTrafficWidget` one should be displayed first. Once that is completed/dismissed, the `helpVisibility` one should show up, **after a minimum period of 2 hours (so that the user isn't overwhelmed by many tooltips right after each other)**.

## Implementation Brief

* Create a new file `assets/js/feature-tours/help-visibility.js` that exports an object with the following properties:
  * `slug` should be `helpVisibility`;
  * `contexts` should be an array with `VIEW_CONTEXT_DASHBOARD` and `VIEW_CONTEXT_PAGE_DASHBOARD` values;
  * `version` should be `1.29.0`;
  * `checkRequirements` should ensure the tour only shows up when the `helpVisibility` feature flag is enabled
  * `steps` should be an array with a single item: 
    * `target` should contain a selector which can be used to find the help menu icon in the DOM (coordinate/get that information based on #2846, which implements that icon)
    * `title` should be `Need help?`;
    * `content` should be `Access the Site Kit docs or ask a question in our support forum from here.`;
    * `placement` should be `bottom-start`.
* In the `assets/js/feature-tours/index.js` file, import an object from the `help-visibility.js` file and add it to the end of the array that is exported by default in this file.

**Preventing successive tours**

[...]

PR #3010, opened 2021-03-23
Implement tooltip for new contextual help feature (2847)

johnPhillips:
## Summary

Addresses issue #2847

## Checklist


assets/js/googlesitekit/datastore/user/feature-tours.js
```diff
@@ -283,6 +340,40 @@ const baseSelectors = {
...
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {undefined|boolean} Whether feature tours are on cooldown or undefined.
+	 */
+	areFeatureToursOnCooldown: ( createRegistrySelector( ( select ) => () => {
+		const lastDismissedAt = select( STORE_NAME ).getLastDismissedAt();
+
+		if ( undefined === lastDismissedAt ) {
+			return undefined;
+		}
```
tofumatt, line 372: `getLastDismissedAt()` has a resolver that sets `lastDismissedAt` only if it has a truthy value. But if nothing is obtained, `lastDismissedAt` isn't set as `false`, it simply isn't set resulting in it being `undefined`.

This means, from what I can see, it'll never be set as `false`.

assets/js/googlesitekit/datastore/user/feature-tours.js
```diff
@@ -218,6 +268,13 @@ const baseResolvers = {
 		}
 		yield actions.receiveFeatureToursForView( viewTours, { viewContext } );
 	},
+
+	*getLastDismissedAt() {
+		const { value: lastDismissedAt } = yield Data.commonActions.await( getItem( 'feature_tour_last_dismissed_at' ) );
```
tofumatt, line 273: The `await` action is neat 🙂

assets/js/googlesitekit/datastore/user/feature-tours.js
```diff
@@ -121,11 +132,36 @@ const baseActions = {
...
+			yield {
+				type: CACHE_LAST_DISMISSED_AT,
+				payload: { timestamp },
+			};
+			yield {
+				type: RECEIVE_LAST_DISMISSED_AT,
+				payload: { timestamp },
+			};
+		}() );
+	},
```
tofumatt, line 158: This is a bit awkward to read—what's wrong with:

```suggestion
	*setLastDismissedAt( timestamp ) {
		invariant( timestamp, 'A timestamp is required.' );

		yield {
			type: CACHE_LAST_DISMISSED_AT,
			payload: { timestamp },
		}
		
		yield {
			type: RECEIVE_LAST_DISMISSED_AT,
			payload: { timestamp },
		};
	},
```

tofumatt (changes requested):
Looks like this approach won't ever set a `false` value on `areFeatureToursOnCooldown`, from what I can see.

If the components check for a false-y value that's fine, but checking to see if it's loading and if it's false-y should be two separate things. We don't tend to treat `undefined` as `false`-y when it's a result from our data store, because it usually means we're waiting for something to resolve and thus we can't render anything accurate yet.

assets/js/googlesitekit/datastore/user/feature-tours.js
aaemnnosttv: @tofumatt the `invariant` won't throw as expected in the caller if the function is a generator. Instead it will result in a rejected promise.

assets/js/googlesitekit/datastore/user/feature-tours.js
johnPhillips: Yeah, thanks to @aaemnnosttv for telling me about it - this was giving me grey hair 😂

assets/js/googlesitekit/datastore/user/feature-tours.js
johnPhillips: But `lastDismissedAt` might be set and have a timestamp from more than 2 hours ago, which would mean this does return `false`? 🤔 

| `lastDismissedAt`  timestamp    | Function returns |
| ----------- | ----------- |
| Over 2 hours ago      | `false`       |
| Under 2 hours ago   | `true`        |
| `undefined`   | `undefined`        |

Maybe I have misunderstood?

assets/js/googlesitekit/datastore/user/feature-tours.js
```diff
@@ -153,6 +189,13 @@ const baseControls = {
 
 		return true;
 	} ),
+	[ CACHE_LAST_DISMISSED_AT ]: createRegistryControl( ( ) => async ( { payload } ) => {
```
aaemnnosttv, line 192: This doesn't need to be a registry control since it doesn't use the registry, it can simply be the inner `async` function.

assets/js/googlesitekit/datastore/user/feature-tours.js
```diff
@@ -153,6 +189,13 @@ const baseControls = {
...
+		const { timestamp } = payload;
+
+		setItem( 'feature_tour_last_dismissed_at', timestamp, {
```
aaemnnosttv, line 195: `setItem` is the async part, so we should await it, or return it (the control will await the value automatically).

The return value isn't really used/needed here so we can simply await it.
```suggestion
		await setItem( 'feature_tour_last_dismissed_at', timestamp, {
```

assets/js/googlesitekit/datastore/user/feature-tours.js
aaemnnosttv, line 273: Let's use a constant for this key so we can reference it in tests as well.

assets/js/googlesitekit/datastore/user/feature-tours.js
```diff
@@ -283,6 +340,40 @@ const baseSelectors = {
...
 	} ),
+
+	/**
+	 * Gets the timestamp for the last dismissed feature tour.
+	 *
+	 * @since n.e.x.t
+	 * @private
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {number} Timestamp of the last dismissal.
```
aaemnnosttv, line 351: This may also return `undefined` if the value has not been resolved yet, or if there is no entry in the cache yet.

assets/js/googlesitekit/datastore/user/feature-tours.js
```diff
@@ -283,6 +340,40 @@ const baseSelectors = {
...
+	 * Determines whether feature tours are on cooldown (i.e. the last
+	 * dismissal was within the cooldown time span).
+	 *
+	 * @since n.e.x.t
+	 * @private
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {undefined|boolean} Whether feature tours are on cooldown or undefined.
+	 */
+	areFeatureToursOnCooldown: ( createRegistrySelector( ( select ) => () => {
```
aaemnnosttv, line 367: There's an extra wrapping `()` around the `createRegistrySelector` here which isn't needed.

aaemnnosttv (commented):
Looking good @johnPhillips – just a few minor details to address then this should be ready for tests 👍

assets/js/googlesitekit/datastore/user/feature-tours.test.js
```diff
@@ -286,4 +350,51 @@ describe( 'core/user feature-tours', () => {
...
+			expect( lastDismissedAt ).toEqual( undefined );
+		} );
+
+		it( 'returns the lastDismissedAt timestamp if there is one', () => {
+			const timestamp = Date.now();
+			registry.dispatch( STORE_NAME ).receiveLastDismissedAt( timestamp );
+			expect( registry.select( STORE_NAME ).getLastDismissedAt() ).toEqual( timestamp );
+		} );
+
+		it( 'sets lastDismissedAt in the store if there is a value in the cache', async () => {
```
aaemnnosttv, line 366: Let's clarify that this is a side-effect performed by its resolver rather than the selector itself.

assets/js/googlesitekit/datastore/user/feature-tours.test.js
```diff
@@ -286,4 +350,51 @@ describe( 'core/user feature-tours', () => {
...
+			const timestamp = Date.now();
+			const { setItem, deleteItem } = CacheModule;
+
+			setItem( FEATURE_TOUR_CACHE_KEY, timestamp, { ttl: FEATURE_TOUR_COOLDOWN_PERIOD } );
+
+			registry.select( STORE_NAME ).getLastDismissedAt();
+			await untilResolved( registry, STORE_NAME ).getLastDismissedAt();
+
+			expect( store.getState().lastDismissedAt ).toEqual( timestamp );
+			await	deleteItem( FEATURE_TOUR_CACHE_KEY );
```
aaemnnosttv, line 376: Please move any clean up like this to an `afterEach` callback 👍

assets/js/googlesitekit/datastore/user/feature-tours.test.js
```diff
@@ -286,4 +350,51 @@ describe( 'core/user feature-tours', () => {
...
+	} );
+
+	describe( 'areFeatureToursOnCooldown', () => {
+		it( 'returns undefined if there is no lastDismissedAt timestamp', () => {
+			expect( registry.select( STORE_NAME ).areFeatureToursOnCooldown() ).toBeUndefined();
+		} );
+
+		it( 'returns true if the lastDismissedAt timestamp is within the feature tour cooldown period', () => {
+			const timestamp = Date.now();
+			registry.dispatch( STORE_NAME ).receiveLastDismissedAt( timestamp );
```
aaemnnosttv, line 387: Let's use the `FEATURE_TOUR_COOLDOWN_PERIOD` here as well but tweak the resulting timestamp (+1 sec or so) so that it falls just inside the cooldown range rather than right after it's been dismissed. This way we're more likely to catch a bug if the TTL were to be updated somewhere without using the constant for example.

assets/js/googlesitekit/datastore/user/feature-tours.js
```diff
@@ -153,6 +191,13 @@ const baseControls = {
 
 		return true;
 	} ),
+	[ CACHE_LAST_DISMISSED_AT ]: async ( { payload } ) => {
+		const { timestamp } = payload;
+
+		await setItem( FEATURE_TOUR_CACHE_KEY, timestamp, {
+			ttl: FEATURE_TOUR_COOLDOWN_PERIOD,
```
aaemnnosttv, line 198: The `ttl` here is expected to be a number in seconds, however `FEATURE_TOUR_COOLDOWN_PERIOD` is defined in ms (`1000 * ...`), so this would result in a cooldown 1000x longer than we want 😄 

It would be good to rename `FEATURE_TOUR_COOLDOWN_PERIOD` to `FEATURE_TOUR_COOLDOWN_SECONDS` to clarify the unit.

assets/js/googlesitekit/datastore/user/feature-tours.js
```diff
@@ -30,14 +30,23 @@ import Data from 'googlesitekit-data';
...
 import { STORE_NAME } from './constants';
 import featureTours from '../../../feature-tours';
+import { setItem, getItem } from '../../../googlesitekit/api/cache';
+
 const { createRegistrySelector, createRegistryControl } = Data;
 const { getRegistry } = Data.commonActions;
 
+// Feature tour cooldown period is 2 hours
+export const FEATURE_TOUR_COOLDOWN_PERIOD = 1000 * 60 * 60 * 2;
+export const FEATURE_TOUR_CACHE_KEY = 'feature_tour_last_dismissed_at';
```
aaemnnosttv, line 40: Let's rename this to be more specific to the cache value for the last dismissed at timestamp. The actual cache key is currently more descriptive than it's constant 😄

aaemnnosttv (changes requested):
Thanks @johnPhillips – this is looking good, just a few things left to address. Let me know if you have any questions.

assets/js/googlesitekit/datastore/user/feature-tours.js
johnPhillips: Oops 😂
Yes, good idea!

assets/js/googlesitekit/datastore/user/feature-tours.js
aaemnnosttv: Looks right to me 👍

aaemnnosttv: Actually I think what @tofumatt was getting at here is that the selector doesn't differentiate between a resolving state (`undefined`) and a cache miss which would result in nothing being received so the state would be the same as resolving. I've updated this to receive `null` for the cache miss so this is always called, and then this selector would always return `false` in that case. This also fixed a bug where it was calling `setLastDismissedAt` instead of `receive*` from the resolver which was resulting in unnecessary calls to set the cache.

aaemnnosttv (approved):
Thanks @johnPhillips – I've pushed a few finishing touches on to where you left off, so this looks good to go now.

@tofumatt – would you please give this a final look over since you previously requested changes?

aaemnnosttv:
@googlebot I consent.

assets/js/googlesitekit/datastore/user/feature-tours.js
tofumatt: Oh right, sorry, I forgot about that 👍🏻

tofumatt (approved):
Great, looks good to me too. 👍🏻

PR #3017, opened 2021-03-24
Hide all instances of HelpLink and create a generic withFeatureFlag HOC.

tofumatt:
## Summary

Addresses issue #2846.

This hides the HelpLink everywhere when `helpVisibility` is enabled.

## Relevant technical choices

Added a more generic `withFeatureFlag` HOC as it was needed by several components.

## Checklist

### DIFF
--- a/assets/js/components/module/ModuleFooter.js
+++ b/assets/js/components/module/ModuleFooter.js
@@ -30,6 +30,10 @@ import { useFeature } from '../../hooks/useFeature';
 const ModuleFooter = () => {
 	const helpVisibilityEnabled = useFeature( 'helpVisibility' );
 
+	if ( helpVisibilityEnabled ) {
+		return null;
+	}
+
 	return (
 		<div className="mdc-layout-grid__inner">
 			<div className={ classNames( [
@@ -37,7 +41,7 @@ const ModuleFooter = () => {
 				'mdc-layout-grid__cell--span-12',
 				'mdc-layout-grid__cell--align-right',
 			] ) }>
-				{ ! helpVisibilityEnabled && <HelpLink /> }
+				<HelpLink />
 			</div>
 		</div>
 	);
--- a/assets/js/components/module/ModuleFooter.js
+++ b/assets/js/components/module/ModuleFooter.js
@@ -17,33 +17,13 @@
  */
 
 /**
- * External dependencies
- */
-import classNames from 'classnames';
-
-/**
  * Internal dependencies
  */
-import HelpLink from '../HelpLink';
-import { useFeature } from '../../hooks/useFeature';
+import PageFooter from '../PageFooter';
 
 const ModuleFooter = () => {
-	const helpVisibilityEnabled = useFeature( 'helpVisibility' );
-
-	if ( helpVisibilityEnabled ) {
-		return null;
-	}
-
 	return (
-		<div className="mdc-layout-grid__inner">
-			<div className={ classNames( [
-				'mdc-layout-grid__cell',
-				'mdc-layout-grid__cell--span-12',
-				'mdc-layout-grid__cell--align-right',
-			] ) }>
-				<HelpLink />
-			</div>
-		</div>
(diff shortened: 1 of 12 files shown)


assets/js/components/module/ModuleFooter.js
```diff
@@ -25,17 +25,22 @@ import classNames from 'classnames';
...
+				'mdc-layout-grid__cell--span-12',
+				'mdc-layout-grid__cell--align-right',
+			] ) }>
+				{ ! helpVisibilityEnabled && <HelpLink /> }
+			</div>
 		</div>
-	</div>
-);
+	);
+};
```
felixarntz, line 44: For cases like here where `HelpLink` is the only element within the container / grid, the flag should be around the entire container / grid, to avoid ending up with unnecessary empty containers.

There might in theory be cases where some of the containers are needed to not break the layout, and if there is one like that, it's okay to keep there (we shouldn't change any CSS as part of this PR). However I'm pretty sure all of these containers that have _only_ a `HelpLink` are full-width (span 12) containers, so we should just return `null` from the entire component if the flag is disabled, rather than rendering an empty container.

**Update:** Due to another observation I made here (see other comment on this file), the change right here isn't needed. Still keeping this comment for context.

assets/js/components/settings/SettingsApp.js
```diff
@@ -140,7 +140,7 @@ export default function SettingsApp() {
 							<SettingsAdmin />
 						) }
 						<Cell size={ 12 } alignRight>
-							<HelpLink />
+							{ ! helpVisibilityEnabled && <HelpLink /> }
 						</Cell>
```
felixarntz, line 144: Same here.

assets/js/modules/adsense/components/dashboard/AdSenseDashboardWidget.js
```diff
@@ -193,7 +196,7 @@ export default function AdSenseDashboardWidget() {
 						</Cell>
 
 						<Cell alignRight size={ 12 }>
-							<HelpLink />
+							{ ! helpVisibilityEnabled && <HelpLink /> }
 						</Cell>
```
felixarntz, line 200: Same here - also check other components throughout this PR, I won't annotate the other ones.

This problem also is present in components that you already changed in the previous PR, **please update the `PageFooter` component as well**.

assets/js/components/module/ModuleFooter.js
```diff
@@ -25,17 +25,22 @@ import classNames from 'classnames';
...
 
-const ModuleFooter = () => (
-	<div className="mdc-layout-grid__inner">
-		<div className={ classNames( [
-			'mdc-layout-grid__cell',
-			'mdc-layout-grid__cell--span-12',
-			'mdc-layout-grid__cell--align-right',
-		] ) }>
-			<HelpLink />
+const ModuleFooter = () => {
```
felixarntz, line 30: Just noticed this component is not correctly implemented (not a result of this PR): Can you please reduce it to simply returning `PageFooter`? It should just wrap that component, the same way `DashboardFooter` and `DashboardDetailsFooter` do. The current markup this creates below does not fit correctly with the widget-based module pages.

felixarntz (changes requested):

assets/js/components/module/ModuleFooter.js
tofumatt: That's true, I just figured it was just easier to grep for the conditional rendering by keeping it all the same (eg. `{ ! helpVisibilityEnabled && <HelpLink /> }`). Either way it renders more/less the same visually, but I've removed the containers now 🙂

felixarntz (changes requested):
https://github.com/google/site-kit-wp/pull/3017#discussion_r600614949 and https://github.com/google/site-kit-wp/pull/3017#discussion_r600617820 still haven't been addressed.

felixarntz (approved):
Great!

Issue #2065, opened 2020-09-18
Migrate AdSense page "Performance" widget to use Widget API


felixarntz:
The "Performance over the last x days" widget on the AdSense module page should be refactored as a new component that is registered and rendered using the Widget API.

The existing component should be untouched and only be renamed, giving it a `Legacy` prefix.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new component `ModuleOverviewWidget` should be added in `assets/js/modules/adsense/components/module/ModuleOverviewWidget.js`.
* It should be implemented as a functional, hook-based, datastore-driven widget component that renders the existing "Performance over the last x days" block on the AdSense module page, in a refactored way.
    * In other words, it should replicate what the legacy `AdSenseDashboardWidgetOverview` and `AdSenseDashboardWidgetSiteStats` components display. These components were already refactored at a previous point so the new component should be able to reuse most of the code. The main things to change is to create the API requests themselves instead of receiving data via props, as well as to use `Widget*` components via props.
    * Sub-components can be implemented as necessary. If reusable, they should go into `assets/js/modules/adsense/components/common`. Otherwise (if specific to just this widget), they should go into a new `assets/js/modules/adsense/components/module/ModuleOverviewWidget` directory; in this case, the main widget component should go into the `index.js` file within that directory.
* The `AdSenseDashboardWidgetOverview` and `AdSenseDashboardWidgetSiteStats` components should be renamed, receiving a `Legacy*` prefix.
* The new component should be used to register a new widget in `assets/js/modules/adsense/index.js` (for testing, requires to enable the `widgets.moduleScreens` feature flag):
    * slug: `adsenseModuleOverview`
    * width: `WIDGET_WIDTHS.FULL`
    * priority: `1`
    * wrapWidget: `false` (since it needs to render `Widget` manually to add the `Header` prop
    * widget area: `AREA_MODULE_ADSENSE_MAIN`

## Implementation Brief
* Create a new file `assets/js/modules/adsense/components/module/ModuleOverviewWidget.js` which will export the `ModuleOverviewWidget` functional component.
* `ModuleOverviewWidget` should have the following props: `Widget`, `WidgetReportZero` and `WidgetReportError`, similar to what is done in `assets/js/modules/search-console/components/dashboard/DashboardPopularKeywordsWidget.js`.
* `ModuleOverviewWidget` should take most of the code from the following files:
[...]

Issue #2940, opened 2021-03-10
Migrate `DashboardUniqueVisitorsWidget` Storybook coverage to follow new v6 best practices


felixarntz:
With Storybook now in v6, we should migrate a few particular stories to follow the new best practices established in [this doc](https://docs.google.com/document/d/1Ny2Tlo1_FxZY6s6dC2-E1TWJYi1rhp0OMEtEfXxwXjY/edit). Migrating all existing stories is not a high priority, we can keep most old stories in their existing state; they should only be migrated on demand.

This issue is about migrating a specific datastore-driven (connected) component to follow the new Storybook v6 best practices. As part of this, some general infrastructure for the new best practices (e.g. global decorators, reusable components) should be implemented.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A global Storybook decorator should be implemented that wraps every story in a `WithTestRegistry`.
    * It needs to be verified that this does not break existing stories which might already have their own `WithTestRegistry` (i.e. the closest test registry wins?).
* A new component `WithRegistrySetup` should be added within `tests/js`, that should be used in any decorators that need to populate the test registry:
    * It should take `func` and `children` props.
    * It should only render `children` if an internal state flag `ready` (with initial value `false`) is set to `true`; otherwise should return `null`.
    * It should run the `func` on first render, passing the registry to it via `useRegistry()`, and once it has completed set the internal state flag `ready` to `true`.
* The Analytics `DashboardUniqueVisitorsWidget` component should receive an accompanying `DashboardUniqueVisitorsWidget.stories.js` file, with new Storybook coverage that follows the [new Storybook best practices](https://docs.google.com/document/d/1Ny2Tlo1_FxZY6s6dC2-E1TWJYi1rhp0OMEtEfXxwXjY/edit). Particularly:
    * It should use `Modules/Analytics/Widgets/DashboardUniqueVisitorsWidget` for its title.
    * It should have 6 different stories: one for its regularly loaded state, one for zero data state, one for error state; then, the exact same three states for the case where an entity URL is set (i.e. irl that would be on the single URL page dashboard).
    * If there is some data needed in the datastore which is common for all 6 stories, a component decorator should be implemented that takes care of it via `WithRegistrySetup`.
    * For any other data needed in the datastore, story-specific decorators should be added, which use `WithRegistrySetup`.
* The existing Storybook coverage for `DashboardUniqueVisitorsWidget` from before should be removed.

[...]

PR #3023, opened 2021-03-25
Migrate AdSense page "Performance" widget to use Widget API

asvinb:
## Summary

Addresses issue #2065 

## Checklist


asvinb:
@felixarntz @aaemnnosttv In the IB, there is mention about creating stories for the "Performance" widget. My question is, since we are now using Storybook v6, I'am not sure whether we should still be using the `generateReportBasedWidgetStories` to create the stories for the widget. Let me know what you think.

aaemnnosttv:
Good question @asvinb – I believe the intention is not to use that utility for any stories for new components but if the component already has stories, it should continue to build on what's already there for now but I'll leave that to @felixarntz to confirm.

assets/js/modules/adsense/components/module/ModuleOverviewWidget/Overview.js
```diff
@@ -0,0 +1,124 @@
...
+		<Grid>
+			<Row>
+				{ dataBlocks.map( ( block, i ) => (
+					<Cell
+						key={ i }
+						smSize={ 2 }
+						mdSize={ 2 }
+						lgSize={ 3 }
+					>
+						<DataBlock
```
aaemnnosttv, line 97: We only have 4 stats and this is fixed so let's keep them them inline (no iteration necessary).

The wrapping `Cell` props can be extracted to a variable instead to avoid duplicating those for each one since they are always the same (`key` was the one different and that's no longer needed if we remove the map) 🙂

assets/js/modules/adsense/components/module/ModuleOverviewWidget/Stats.js
```diff
@@ -0,0 +1,151 @@
...
+			},
+		},
+	};
+
+	const dataMap = getSiteStatsDataForGoogleChart(
+		currentRangeData,
+		previousRangeData,
+		Object.values( metrics )[ selectedStats ],
+		selectedStats + 1,
+		currentRangeData.headers[ selectedStats + 1 ],
```
aaemnnosttv, line 125: Why the `+ 1` ?

assets/js/modules/adsense/components/module/ModuleOverviewWidget/index.js
```diff
@@ -0,0 +1,142 @@
...
+
+	const currentRangeLoading = useSelect( ( select ) => ! select( STORE_NAME ).hasFinishedResolution( 'getReport', [ currentRangeArgs ] ) );
+	const previousRangeLoading = useSelect( ( select ) => ! select( STORE_NAME ).hasFinishedResolution( 'getReport', [ previousRangeArgs ] ) );
+	const currentRangeChartLoading = useSelect( ( select ) => ! select( STORE_NAME ).hasFinishedResolution( 'getReport', [ currentRangeChartArgs ] ) );
+	const previousRangeChartLoading = useSelect( ( select ) => ! select( STORE_NAME ).hasFinishedResolution( 'getReport', [ previousRangeChartArgs ] ) );
+
+	const currentRangeError = useSelect( ( select ) => select( STORE_NAME ).getErrorForSelector( 'getReport', [ currentRangeArgs ] ) );
+	const previousRangeError = useSelect( ( select ) => select( STORE_NAME ).getErrorForSelector( 'getReport', [ previousRangeArgs ] ) );
+	const currentRangeChartError = useSelect( ( select ) => select( STORE_NAME ).getErrorForSelector( 'getReport', [ currentRangeChartArgs ] ) );
+	const previousRangeChartError = useSelect( ( select ) => select( STORE_NAME ).getErrorForSelector( 'getReport', [ previousRangeChartArgs ] ) );
```
aaemnnosttv, line 91: These could be combined into 2 selects. We generally do a single `useSelect` with a single select within, but the convention is to use a single `useSelect` for each value that we need in the component scope. In this case, we only care about `loading` and `error` – each individual value isn't important in this case, we just want the first one or none. Furthermore, this is a little more efficient because within the selector function we can return as soon as we've found something rather than selecting them all unconditionally.

assets/js/modules/adsense/components/module/ModuleOverviewWidget/index.js
```diff
@@ -0,0 +1,142 @@
...
+import Data from 'googlesitekit-data';
+const { useSelect } = Data;
+
+const ModuleOverviewWidget = ( { Widget, WidgetReportZero, WidgetReportError } ) => {
+	const metrics = {
+		EARNINGS: __( 'Earnings', 'google-site-kit' ),
+		PAGE_VIEWS_RPM: __( 'Page RPM', 'google-site-kit' ),
+		IMPRESSIONS: __( 'Impressions', 'google-site-kit' ),
+		PAGE_VIEWS_CTR: __( 'Page CTR', 'google-site-kit' ),
+	};
```
aaemnnosttv, line 49: Static things like this no longer need to be defined within the component because translations will always be loaded before the JS entry point is initialized. Not a big deal but keeps the component body a little cleaner 🙂

assets/js/modules/adsense/components/module/ModuleOverviewWidget/index.js
```diff
@@ -0,0 +1,142 @@
...
+		<Widget
+			noPadding
+			Header={ Header }
+		>
+			<Overview
+				metrics={ metrics }
+				currentRangeData={ currentRangeData }
+				previousRangeData={ previousRangeData }
+				selectedStats={ selectedStats }
+				handleStatsSelection={ handleStatsSelection }
```
aaemnnosttv, line 124: We can pass the setter here instead which is stable from `useState` – no need for an added `useCallback` around it 👍 

```suggestion
				handleStatsSelection={ setSelectedStats }
```

aaemnnosttv (changes requested):
Thanks @asvinb – this is looking good with a few changes to address.

Regarding the stories, let's write these as we would for any other new component: co-located using CSF, without `generateReportBasedWidgetStories`.

assets/js/modules/adsense/components/module/ModuleOverviewWidget/Overview.js
asvinb: @aaemnnosttv I had to remove the `Cell` component to instead use custom class names. I think there was a valid reason to use the custom class names before especially in cases we have very long numbers, which the custom classes take care of. So the new widget is similar to the legacy one. About using `map`, I prefer if we keep it like that, because it reduces the markup and everything is cleaner IMHO. Also, I think it'll be consistent about how we display the blocks on other modules `Overview` components as well.

assets/js/modules/adsense/components/module/ModuleOverviewWidget/Stats.js
asvinb: @aaemnnosttv Not sure about that since am taking what we already have in the legacy component. https://github.com/google/site-kit-wp/blob/develop/assets/js/modules/adsense/components/dashboard/AdSenseDashboardWidgetSiteStats.js#L155 
If i remove the +1, the data is incorrectly displayed in the graph. Guess, the index of some data is off by 1.

assets/js/modules/adsense/components/module/ModuleOverviewWidget/index.js
asvinb: @aaemnnosttv I made the updates, but not sure if it's what you meant. Can you check?

assets/js/modules/adsense/components/module/ModuleOverviewWidget/index.js
asvinb: Sounds good. Updated!

assets/js/modules/adsense/components/module/ModuleOverviewWidget/index.js
asvinb: Nice. Updated!

asvinb:
@aaemnnosttv For the story in CSF, I wrapped the story within the `WithTestRegistry` component. Ideally it `WithTestRegistry` should be in the decorator of `.storybook/preview.js` where all stories have access to the registry. However, I was running into an issue where storybook was failing. Spent some time trying to debug but I saw there's a similar ticket #2940 where a global decorator is added which gives access to the registry. I guess we can just update this story when #2940 is done. What do you think?

felixarntz:
@aaemnnosttv @asvinb Per our conversation this morning: I'm not opposed to having this use CSF if already done - but I'd imagine this is not trivial until we've established that our proposed approach actually works, so maybe it's better to keep using the old widget Storybook approach for now (which btw is further enhanced via #3034, where a "Loading" state story is also added).

assets/js/modules/adsense/components/module/ModuleOverviewWidget/Overview.js
```diff
@@ -0,0 +1,115 @@
...
+					title={ block.title }
+					datapoint={ block.datapoint }
+					datapointUnit={ block.datapointUnit }
+					change={ block.change }
+					changeDataUnit={ block.changeDataUnit }
+					context={ block.context }
+					selected={ block.selected }
+					handleStatSelection={ block.handleStatsSelection }
+				/>
+			) ) }
```
aaemnnosttv, line 102: > About using `map`, I prefer if we keep it like that, because it reduces the markup and everything is cleaner IMHO. Also, I think it'll be consistent about how we display the blocks on other modules `Overview` components as well.

I think there are pros/cons with both approaches, but I don't think we should iterate over a fixed list of props objects for a few reasons:
* It makes it easier to define one of the props wrong because the editor/IDE has no awareness of the component it is being used with. Since it's just a plain object, you can define anything and won't see it until it runs in the browser. Said another way, you lose the benefits of static analysis for prop names/types.
* There's actually a bit of a (potential) bug here in that each object above is intended to be a specific index (i.e. `selectedStats === 2`), but `stat` is being passed as the index from the `dataBlocks` array where everywhere else it is explicitly defined. If the order of the objects in the array changed, the stat each block refers to would no longer be accurate.

As for consistency, we have the inline approach in the [`AdSenseDashboardWidgetOverview`](https://github.com/google/site-kit-wp/blob/55c3280c1e61cbce8aa9360592c72a3fe452460b/assets/js/modules/adsense/components/dashboard/AdSenseDashboardWidgetOverview.js) component already. 

I think there's room to clean this up a bit more, but we can revisit that in a dedicated issue, perhaps after the legacy components are removed 👍

assets/js/modules/adsense/components/module/ModuleOverviewWidget/Stats.js
aaemnnosttv: Yes, I found that the first index is from the DATE dimension. `currentRangeData.headers`:
```js
[
    {
        "name": "DATE",
        "type": "DIMENSION"
    },
    {
        "name": "EARNINGS",
        "type": "METRIC_CURRENCY",
        "currency": "USD"
    },
    {
        "name": "PAGE_VIEWS_RPM",
        "type": "METRIC_CURRENCY",
        "currency": "USD"
    },
    {
        "name": "IMPRESSIONS",
        "type": "METRIC_TALLY"
    },
    {
        "name": "PAGE_VIEWS_CTR",
        "type": "METRIC_RATIO"
    }
]
```

So `selectedStats` is a 0-based index, but `currentRangeData.headers` is an array with metrics after the dimensions.

Let's document this so that this relationship is easier to see 👍

assets/js/modules/adsense/components/module/ModuleOverviewWidget/ModuleOverviewWidget.stories.js
```diff
@@ -0,0 +1,98 @@
...
+ */
+import { WithTestRegistry } from '../../../../../../../tests/js/utils';
+import * as fixtures from '../../../datastore/__fixtures__';
+import { STORE_NAME } from '../../../datastore/constants';
+import { CORE_USER } from '../../../../../googlesitekit/datastore/user/constants';
+import { CORE_WIDGETS, WIDGET_WIDTHS, WIDGET_AREA_STYLES } from '../../../../../googlesitekit/widgets/datastore/constants';
+import WidgetAreaRenderer from '../../../../../googlesitekit/widgets/components/WidgetAreaRenderer';
+import ModuleOverviewWidget from './index';
+
+export const Performance = () => {
```
aaemnnosttv, line 30: This would only export the single story; we should still have stories for the other states like we do now with the generate-widget-based stories utility.

Per @felixarntz 's https://github.com/google/site-kit-wp/pull/3023#issuecomment-809621278, it looks like I was wrong about going full SB6 here just yet, so feel free to use the old approach for now until we've fully fleshed out how the new approach will work going forward, including these multi-variant cases 👍

aaemnnosttv (commented):
Thanks @asvinb! This looks good and works well for me, just a few things left to address. Please ping me directly if you have any questions.

assets/js/modules/adsense/components/module/ModuleOverviewWidget/Overview.js
asvinb: 👍  Reverted the changes to not use `map`

assets/js/modules/adsense/components/module/ModuleOverviewWidget/index.js
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	if ( currentRangeError || previousRangeError || currentRangeChartError || previousRangeChartError ) {
+		return (
+			<WidgetReportError
+				moduleSlug="adsense"
+				error={ currentRangeError || previousRangeError || currentRangeChartError || previousRangeChartError }
+			/>
+		);
+	}
```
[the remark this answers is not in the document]
asvinb: Updated @aaemnnosttv

assets/js/modules/adsense/components/module/ModuleOverviewWidget/Overview.js
```diff
@@ -37,69 +37,54 @@ const Overview = ( {
...
-					changeDataUnit={ block.changeDataUnit }
-					context={ block.context }
-					selected={ block.selected }
-					handleStatSelection={ block.handleStatsSelection }
-				/>
-			) ) }
+			<DataBlock
+				className="googlesitekit-data-block--page-rpm googlesitekit-data-block--button-1"
+				title={ metrics[ headers[ 0 ].name ] }
+				datapoint={ totals[ 0 ] }
```
aaemnnosttv, line 45: These `DataBlock` components are missing the `stat` prop now, which is still needed.

assets/js/modules/adsense/components/module/ModuleOverviewWidget/Overview.js
```diff
@@ -37,69 +37,54 @@ const Overview = ( {
...
+				datapointUnit={ headers[ 0 ]?.currency }
+				change={ calculateChange( previousTotals[ 0 ], totals[ 0 ] ) }
+				changeDataUnit="%"
+				context="button"
+				selected={ selectedStats === 0 }
+				handleStatSelection={ () => handleStatsSelection( 0 ) }
```
aaemnnosttv, line 51: We should be able to pass `handleStatsSelection` directly like we did before. I'm guessing you did it this way due to the missing `stat` prop since this is what determines the value passed to the provided function.

aaemnnosttv (commented):
Thanks @asvinb – this is super close! See my comments below.

assets/js/modules/adsense/components/module/ModuleOverviewWidget/Overview.js
asvinb: Ah good catch @aaemnnosttv . Updated.

assets/js/modules/adsense/components/module/ModuleOverviewWidget/Overview.js
asvinb: 👍

aaemnnosttv (approved):
Great, thanks @asvinb !

PR #3030, opened 2021-03-26
Fix broken TourTooltip stories, add reset

aaemnnosttv:
## Summary

Addresses issue #2650

## Relevant technical choices

* This PR fixes the stories for TourTooltip which were broken due to the default fetch mock in Storybook which was causing an empty object to be received for `dismissedTourSlugs` which failed when calling an array method on it.
    * We _could_ be more defensive about making sure we only receive an array but not sure we should be paranoid about what our own API returns for a successful response. 
    * If we decide to make changes to the underlying component, we should probably update this to target `master` instead in which case it might need to be rebased

## Checklist

### DIFF
--- a/stories/tour-tooltips.stories.js
+++ b/stories/tour-tooltips.stories.js
@@ -1,5 +1,5 @@
 /**
- * VisuallyHidden stories.
+ * TourTooltip stories.
  *
  * Site Kit by Google, Copyright 2021 Google LLC
  *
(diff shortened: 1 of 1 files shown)


felixarntz (changes requested):
@aaemnnosttv Can you open this against `master` instead? :)

stories/tour-tooltips.stories.js
```diff
@@ -329,6 +335,23 @@ const MockWPDashboard = () => (
...
+
+	return (
+		<div style={ { textAlign: 'right' } }>
+			<Button onClick={ reset }>
+				Reset Dismissed Tours
+			</Button>
+		</div>
+	);
+};
+
```
felixarntz, line 354: Nice!

felixarntz (approved):
Great catch and fix!

Issue #3031, opened 2021-03-26
Improve language throughout codebase


felixarntz:
The Site Kit codebase should always use inclusive language; this is not fully the case currently as there are still some terms being used which should be removed.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The default branch should be renamed to `main`.
* The following terms should no longer be used in the Site Kit codebase (roughly sorted by priority for removal):
    * `master`
    * `whitelist`
    * `ninja`
    * `dummy`
    * `kebab`
    * `sane`, `sanity`, etc.
    * `agnostic`
    * `primitive`
    * `hang`
    * ~~`hit`~~ (we only use "cache hit" which has a different meaning)
* The only place where these terms are okay to be kept is when required due to third-party code.

## Implementation Brief

* Merge #3032.

### Test Coverage

* N/A

### Visual Regression Changes

* N/A

## QA Brief

* Ensure the terms in the ACs are no longer to be found in the codebase, unless it is in places where it is required due to third-party code usage.

## Changelog entry

* N/A

aaemnnosttv:
IB ✅

asvinb:
@felixarntz I just wanted to confirm something about the strikethrough word: `hit`. Is it no longer relevant because I can see occurrences of this word in the following files:
- `assets/js/googlesitekit/api/middleware/preloading.js`
- `assets/js/modules/analytics/components/dashboard/LegacyAdSenseDashboardWidgetTopPagesTableSmall.js`
- `assets/js/modules/analytics/components/dashboard/AnalyticsAdSenseDashboardWidgetTopPagesTable.js`

If that's fine, then QA approved.

felixarntz:
@asvinb Ah, good catch! Yes, it should be replaced everywhere where it's not "cache hit". **QA** ❌

asvinb:
Thanks for confirming @felixarntz . Assigning task back to you.

asvinb:
Checked the codebase for the terms in the AC and none found.
### QA: ✅

PR #3032, opened 2021-03-26
Remove problematic terms and facilitate default branch renaming

felixarntz:
## Summary

Addresses issue #3031

## Relevant technical choices

* The only changes here that go beyond changing wording were added in distinct commits - please review those separately.
* This PR leaves us with some problematic terms only in `composer.lock` and `package-lock.json`, as well as in a few other areas related to third-party code (e.g. stylelint and PHP-Scoper config).
    * `stylelint` already has improved language throughout its codebase too, so can update it. That comes with a bunch of side effects though, so should be done separately.

## Checklist


tests/phpunit/integration/Core/Util/ScopesTest.php
```diff
@@ -96,7 +96,7 @@ public function is_satisfied_by_provider() {
 				),
 				true,
 			),
-			// webmaster is not satisfied by webmasters.readonly
+			// webmasters is not satisfied by webmasters.readonly
```
felixarntz, line 99: Unrelated change - just fixes a typo.

tofumatt:
@googlebot I consent.

tofumatt:
😆  Oops, I noticed the `zips` workflow still had the `master` branch listed and changed it, only to see that was intentional and from this branch. I'll change it back, but add a comment to note why it isn't changed, as that wasn't clear to me.

I venture they'll eventually change that too, hopefully, but we'll see.

tofumatt (approved):
Awesome 👍🏻

tests/bin/install-wp-tests.sh
```diff
@@ -121,6 +119,9 @@ install_test_suite() {
 		sed $ioption "s/yourusernamehere/$DB_USER/" "$WP_TESTS_DIR"/wp-tests-config.php
 		sed $ioption "s/yourpasswordhere/$DB_PASS/" "$WP_TESTS_DIR"/wp-tests-config.php
 		sed $ioption "s|localhost|${DB_HOST}|" "$WP_TESTS_DIR"/wp-tests-config.php
+
+		# Enforce mysqli.
+		echo "define( 'WP_USE_EXT_MYSQL', false );" >> "$WP_TESTS_DIR"/wp-tests-config.php
```
tofumatt, line 124: Just noting, for any future reviewers/on-lookers, that this change removes the link to a `master` repo URL and that the file being downloaded does what this new code does. If anything, this is probably more stable in case that repo is removed 🙂

aaemnnosttv: @tofumatt @felixarntz I don't think we should bother making customizations to this file. This script is essentially an [external dependency](https://github.com/wp-cli/scaffold-command/blob/479cd4f10f638d7a46a5b4a60c80bd4b4a6bd5e2/templates/install-wp-tests.sh) that we mirror in our source because it's a shell script.

If anything, this patch should be an upstream contribution 🙂 

In the future, we should be able to update this by replacing this file with the newest version from upstream without requiring manual updates afterwards.

PR #3034, opened 2021-03-26
Enhancement/2068 storybook stories

eugene-manuilov:
## Summary

Addresses issue #2068

## Relevant technical choices

* Updated mock factory to generate reports for multiple metrics.
* Reworked loading state of Overview and Top Pages widgets to be more consistent with the legacy components.
* Reworked zero and error states of Overview and Top Pages widgets to be more clear and understandable by users.

## Checklist


assets/js/modules/analytics/components/module/ModuleOverviewWidget/index.js
```diff
@@ -74,29 +76,46 @@ export default function ModuleOverviewWidget( { Widget, WidgetReportError } ) {
 	const statsReport = useSelect( ( select ) => select( MODULES_ANALYTICS ).getReport( statsArgs ) );
 	const statsError = useSelect( ( select ) => select( MODULES_ANALYTICS ).getErrorForSelector( 'getReport', [ statsArgs ] ) );
 
-	if ( overviewError || statsError ) {
-		return <WidgetReportError statsError={ overviewError || statsError } />;
+	const isZero = isZeroReport( overviewReport );
+	const isError = overviewError || statsError;
+
+	if ( ! overviewLoaded || ! statsLoaded ) {
+		return <ProgressBar />;
```
felixarntz, line 83: Let's use `PreviewBlock` here. While the legacy component here relies on `ProgressBar` and this is indeed a change, we should build the new widget implementation consistent with how the widgets in the dashboard work. It'd be better to then also wrap that `PreviewBlock` in the regular `Widget` UI below.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/index.js
```diff
@@ -74,29 +76,46 @@ export default function ModuleOverviewWidget( { Widget, WidgetReportError } ) {
...
-				handleStatSelection={ setSelectedState }
-			/>
+		<Widget
+			Header={ Header }
+			noPadding={ reportsAreValid }
+		>
+			{ isError && (
+				<WidgetReportError moduleSlug="analytics" error={ overviewError || statsError } />
+			) }
+			{ ! isError && isZero && (
```
felixarntz, line 96: Don't we need to wrap this to prevent some unexpected output from `isZero`?
```suggestion
			{ ( ! isError && isZero ) && (
```

assets/js/modules/analytics/components/module/ModuleOverviewWidget/index.js
```diff
@@ -74,29 +76,46 @@ export default function ModuleOverviewWidget( { Widget, WidgetReportError } ) {
...
+				<WidgetReportZero moduleSlug="analytics" />
+			) }
+			{ reportsAreValid && (
```
felixarntz, line 99: Nitpick, but using a variable here IMO that makes code readability harder. You have the related conditions right above, but here they are "hidden" behind a variable that is further above in the code.
```suggestion
			{ ( ! isError && ! isZero ) && (
```

assets/js/modules/analytics/components/module/ModuleOverviewWidget/index.js
```diff
@@ -74,29 +76,46 @@ export default function ModuleOverviewWidget( { Widget, WidgetReportError } ) {
...
-		<Widget Header={ Header } noPadding>
-			<Overview
-				loaded={ overviewLoaded }
-				report={ overviewReport }
-				selectedStat={ selectedStat }
-				handleStatSelection={ setSelectedState }
-			/>
+		<Widget
+			Header={ Header }
+			noPadding={ reportsAreValid }
```
felixarntz, line 91: Maybe you can just keep this as
```suggestion
			noPadding={ ! isError && ! isZero }
```
? I can see why it makes sense to put it in a variable, but also this condition is super simple, and per my other comment IMO it's easier to understand when writing out these `isError` and `isZero` conditions always.

assets/js/modules/analytics/components/module/ModulePopularPagesWidget/index.js
```diff
@@ -72,20 +72,27 @@ export default function ModulePopularPagesWidget( { Widget, WidgetReportError }
...
-			noPadding
+			Footer={ reportIsValid ? Footer : undefined }
+			noPadding={ reportIsValid }
 		>
-			{ ! loaded && (
-				<PreviewTable padding />
+			{ error && (
+				<WidgetReportError moduleSlug="analytics" error={ error } />
+			) }
+			{ ! error && isZero && (
```
felixarntz, line 92: Same here, shouldn't this be
```suggestion
			{ ( ! error && isZero ) && (
```

assets/js/modules/analytics/components/module/ModulePopularPagesWidget/index.js
```diff
@@ -72,20 +72,27 @@ export default function ModulePopularPagesWidget( { Widget, WidgetReportError }
...
+				<WidgetReportZero moduleSlug="analytics" />
 			) }
-			{ loaded && (
+			{ reportIsValid && (
```
felixarntz, line 95: See similar comment on `ModuleOverviewWidget`.

assets/js/modules/analytics/components/module/ModulePopularPagesWidget/index.js
```diff
@@ -72,20 +72,27 @@ export default function ModulePopularPagesWidget( { Widget, WidgetReportError }
...
+	const isZero = isZeroReport( report );
+	const reportIsValid = ! error && ! isZero;
+
 	return (
 		<Widget
 			Header={ Header }
-			Footer={ Footer }
-			noPadding
+			Footer={ reportIsValid ? Footer : undefined }
+			noPadding={ reportIsValid }
```
felixarntz, line 87: See above.

assets/js/modules/analytics/components/module/ModulePopularPagesWidget/index.js
```diff
@@ -72,20 +72,27 @@ export default function ModulePopularPagesWidget( { Widget, WidgetReportError }
...
 
+	const isZero = isZeroReport( report );
+	const reportIsValid = ! error && ! isZero;
+
 	return (
 		<Widget
 			Header={ Header }
-			Footer={ Footer }
-			noPadding
+			Footer={ reportIsValid ? Footer : undefined }
```
felixarntz, line 86: Why not always render the `Footer` here, same way as the `Header`? It has the source link only and that shouldn't depend on whether or not the API request is successful IMO.

assets/js/modules/analytics/components/module/ModulePopularPagesWidget/index.js
```diff
@@ -72,20 +72,27 @@ export default function ModulePopularPagesWidget( { Widget, WidgetReportError }
 	const loaded = useSelect( ( select ) => select( MODULES_ANALYTICS ).hasFinishedResolution( 'getReport', [ args ] ) );
 	const error = useSelect( ( select ) => select( MODULES_ANALYTICS ).getErrorForSelector( 'getReport', [ args ] ) );
 
-	if ( error ) {
-		return <WidgetReportError error={ error } />;
+	if ( ! loaded ) {
+		// Return NULL while loading to replicate the behavior of the legacy components.
+		return null;
```
felixarntz, line 77: While this in principle makes sense to exactly replicate the old behavior, let's not do this here. Let's build out each widget component individually, unaware of its surroundings, the same way the refactored version of the dashboard is built. In other words, this shouldn't return `null` just because we assume the `ModuleOverviewWidget` already takes care of the loading state.

We can later combine loading states etc as an enhancement if needed. I'd say it's more about consistency among all our widgets here - i.e. this should use `PreviewBlock`, wrapped within the `Widget` UI below.

felixarntz (changes requested):
This looks solid mostly, but I have a few points of feedback. One general thing I'd suggest:

* Instead of now having all conditions for `isError`, `isZero`, `! loaded` etc within the single `Widget` container below (which requires extra conditionals to make them mutually exclusive) which makes readability harder and diverges from the easy-to-read "early return" approach we have in most components, maybe we could keep the early return approach?
* Using the early return _does_ require us to always wrap the right `Widget` component around the respective content, so that adds _some_ duplicate code but it improves readability and reduces conditional complexity.
* If we want to reduce the duplicate code from writing out `Widget` with the specific props multiple times, I think we can avoid extra duplicate code by defining the specific version of `Widget` that we need for both components in the same file, e.g. as `function WidgetContainer( { Widget, children, noPadding } )`.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/index.js
eugene-manuilov: Updated.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/index.js
eugene-manuilov: Updated.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/index.js
eugene-manuilov: Updated.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/index.js
eugene-manuilov: Updated.

assets/js/modules/analytics/components/module/ModulePopularPagesWidget/index.js
eugene-manuilov: Updated.

assets/js/modules/analytics/components/module/ModulePopularPagesWidget/index.js
eugene-manuilov: Updated.

assets/js/modules/analytics/components/module/ModulePopularPagesWidget/index.js
eugene-manuilov: Updated.

assets/js/modules/analytics/components/module/ModulePopularPagesWidget/index.js
eugene-manuilov: Updated.

assets/js/modules/analytics/components/module/ModulePopularPagesWidget/index.js
eugene-manuilov: Updated.

assets/js/modules/adsense/components/dashboard/DashboardTopEarningPagesWidget.js
```diff
@@ -94,6 +94,37 @@ function DashboardTopEarningPagesWidget( { Widget, WidgetReportZero, WidgetRepor
...
+			Component: ( { row } ) => numFmt(
+				row.metrics[ 0 ].values[ 0 ],
+				{
+					style: 'decimal',
+					minimumFractionDigits: 2,
+					maximumFractionDigits: 2,
+				}
+			),
+		},
+	];
```
felixarntz, line 126: Why did you move them into the components? It reads a bit more intuitive, but it also means that the arrays will be regenerated on every render although they're static - or am I missing something here?

assets/js/modules/analytics/components/module/ModuleOverviewWidget/index.js
```diff
@@ -74,29 +75,42 @@ export default function ModuleOverviewWidget( { Widget, WidgetReportError } ) {
...
+			Header={ Header }
+			noPadding={ ! isError && ! isZero }
+		>
+			{ isError && (
+				<WidgetReportError moduleSlug="analytics" error={ overviewError || statsError } />
+			) }
+			{ ( ! isError && isZero ) && (
+				<WidgetReportZero moduleSlug="analytics" />
+			) }
+			{ ( ! isError && ! isZero ) && (
```
felixarntz, line 92: Nitpick: Instead of having all these conditions, I think the code would be easier to follow if it was using early returns, the same way that you've adjusted it in `ModulePopularPagesWidget`:
* If `isError`, return `WidgetReportError` wrapped in `Widget` with the `Header`.
* If `isZero`, return `WidgetReportZero` wrapped in `Widget` with the `Header`.
* Otherwise render the default content with `noPadding` simply set to `true`.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/index.js
```diff
@@ -74,29 +75,42 @@ export default function ModuleOverviewWidget( { Widget, WidgetReportError } ) {
...
-			<SiteStats
-				loaded={ statsLoaded }
-				selectedStat={ selectedStat }
-				report={ statsReport }
-			/>
+					<SiteStats
+						loaded={ statsLoaded }
+						selectedStat={ selectedStat }
+						report={ statsReport }
+					/>
```
felixarntz, line 105: I'm not sure how valuable it is to have individual loading states for these two. They just render as a big block anyway, and the only point of having two loaders is that one of them might finish faster - which is not necessarily an advantage: In the All Traffic widget for example we intentionally made it so that all pieces of the widget remain in loading state until everything is done.

Maybe instead of passing a `loaded` prop to `Overview` and `SiteStats`, we could have a single loading state returned by `ModuleOverviewWidget` itself? This would also simplify it and fall nicely in line with my above comment. Basically, there could be another early return for `! overviewLoaded || ! statsLoaded`, which should be wrapped in a `Widget` with the `Header`. This also shouldn't have `noPadding`, which is the case right now for the loading state, making it look different than most our widgets which don't have the `PreviewBlock`s extend to the edges of the widget box.

assets/js/modules/search-console/components/dashboard/DashboardPopularKeywordsWidget.js
```diff
@@ -70,6 +73,46 @@ function DashboardPopularKeywordsWidget( { Widget, WidgetReportZero, WidgetRepor
...
+			title: __( 'Clicks', 'google-site-kit' ),
+			description: __( 'Number of times users clicked on your content in search results', 'google-site-kit' ),
+			Component: ( { row } ) => numFmt( row.clicks, { style: 'decimal' } ),
+		},
+		{
+			title: __( 'Impressions', 'google-site-kit' ),
+			description: __( 'Counted each time your content appears in search results', 'google-site-kit' ),
+			Component: ( { row } ) => numFmt( row.impressions, { style: 'decimal' } ),
+		},
+	];
```
felixarntz, line 114: Same here: This is probably not an issue, but was just curious why you moved this block of code.

stories/utils/generate-widget-stories.js
```diff
@@ -131,12 +132,28 @@ export function generateReportBasedWidgetStories( {
 				dispatch( datastore ).receiveGetReport( data, { options } );
 			}
 
+			// Run additional callback if it exists.
+			if ( additionalLoadedCallback ) {
+				additionalLoadedCallback( dispatch, data, options );
+			}
+		},
+		Loading( { dispatch } ) {
```
felixarntz, line 140: Awesome! 👍

felixarntz (changes requested):
@eugene-manuilov Awesome work, +1 to the additional `Loading` stories!

I left a few more comments, should be the final round. This is almost good to go, I just have some recommendations for further simplifying the `ModuleOverviewWidget` implementation.

assets/js/modules/adsense/components/dashboard/DashboardTopEarningPagesWidget.js
eugene-manuilov: @felixarntz translations might not be ready/loaded by the time we define that array. By moving that array into the component we make sure that static copy will be translated correctly.

assets/js/modules/analytics/components/module/ModuleOverviewWidget/index.js
eugene-manuilov: Yep, updated!

assets/js/modules/analytics/components/module/ModuleOverviewWidget/index.js
eugene-manuilov: You are right, update!

assets/js/modules/search-console/components/dashboard/DashboardPopularKeywordsWidget.js
eugene-manuilov: Same here: to make sure translations are loaded by the time we render this component.

assets/js/modules/adsense/components/dashboard/DashboardTopEarningPagesWidget.js
felixarntz: @eugene-manuilov Thanks, makes sense!

felixarntz (approved):
Excellent!

PR #3037, opened 2021-03-29
Redirect splash to dashboard if authenticated

aaemnnosttv:
## Summary

Addresses issue #2529 

## Relevant technical choices

* Extracts `connect` and `disconnect` actions to dedicated action URLs and handlers
* I wanted to add tests around the behavior of the splash screen itself but the changes needed to support this are rather out of scope, so I will create a new issue for this

## Checklist


includes/Core/Authentication/Authentication.php
```diff
@@ -618,71 +625,80 @@ private function cron_refresh_profile_data( $user_id ) {
...
-					'googlesitekit_reset_session' => 1,
-				)
-			);
+		// User is trying to authenticate, but access token hasn't been set.
+		$additional_scopes = $input->filter( INPUT_GET, 'additional_scopes', FILTER_DEFAULT, FILTER_REQUIRE_ARRAY );
+		wp_safe_redirect(
+			esc_url_raw(
+				$this->get_oauth_client()->get_authentication_url( $redirect_url, $additional_scopes )
+			)
+		);
```
eugene-manuilov, line 672: The `esc_url_raw` call here is redundant because it is intended for database usage.

```suggestion
		wp_safe_redirect(
			$this->get_oauth_client()->get_authentication_url( $redirect_url, $additional_scopes )
		);
```

eugene-manuilov (changes requested):
Overall looks good to me. Just left one suggestion plus we need to fix broken e2e tests.

aaemnnosttv:
Thanks for the review @eugene-manuilov. I believe the E2E is failing now due to the added redirection combined with our fake authentication which is usually set at the beginning. Will take a closer look at this soon.

includes/Core/Authentication/Authentication.php
aaemnnosttv: It's not only for DB usage but rather anything other than for display. WP is rather inconsistent about its use with redirects, but there are instances of it used in core and it is mentioned in the docs:  https://developer.wordpress.org/reference/functions/esc_url_raw/#more-information

In this case, it may be unnecessary as the returned URL shouldn't have any encoded entities. I forget now but I believe the reason these were added in a number of places was from one of them biting us where it was needed. I'll take a closer look.

aaemnnosttv: Let's leave this as-is for now. We should probably re-evaluate this and other uses in a separate issue.

eugene-manuilov (changes requested):
In general looks good to me. Just found one tiny thing that we need to address.

includes/Core/Authentication/Authentication.php
```diff
@@ -618,71 +625,80 @@ private function cron_refresh_profile_data( $user_id ) {
...
-			// User is trying to authenticate, but access token hasn't been set.
-			$additional_scopes = $input->filter( INPUT_GET, 'additional_scopes', FILTER_DEFAULT, FILTER_REQUIRE_ARRAY );
-			wp_safe_redirect(
-				esc_url_raw(
-					$auth_client->get_authentication_url( $redirect_url, $additional_scopes )
-				)
-			);
-			exit();
-		}
+		wp_safe_redirect( $redirect_url );
```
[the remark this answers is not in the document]
aaemnnosttv: I removed it from `handle_connect` as it doesn't seem to be necessary here.

eugene-manuilov (approved):

felixarntz (approved):
Great!

Issue #2969, opened 2021-03-16
Improve GitHub Actions Hygiene


felixarntz:
GitHub Actions have recently been stagnating for days, causing significant delays in being able to review and merge PRs. While there have been a few issues on the administrative side, it looks like our actions are also consuming quota quite aggressively, and we should try to reduce that footprint, consolidating workflows and running them more selectively where possible.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All jobs that do not depend on other jobs should be updated to run the `styfle/cancel-workflow-action` action as its first step
    * The dedicated `cancel.yml` workflow should be removed
* All workflows should define [`defaults`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#defaults) that sets [`timeout-minutes`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes) to `20` as a safeguard against run-away steps/jobs
* All workflows should define [`paths` and/or `paths-ignore`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths) to scope the run as a whole to only run when specified paths change
    *  Note: this is specific to the `push` or `pull_request` event configuration

## Implementation Brief

* Add `timeout-minutes: 20` as a default for all actions.
* Add `styfle/cancel-workflow-action` action to all actions as the first step aside from the `zips.yml` `deploy-to-wiki` job.
* Remove the `cancel.yml` action/file.
* Add `paths` for Jest/CSS/JS/PHP files accordingly, so JS tests are only run when JS files are changed, etc. (For now, always run E2E tests if any file other than a test file has changed, so that adding JS tests doesn't result in E2E tests being run…)
See #3046 for a PR that encompasses all changes sans the last one about combining jobs.

### Test Coverage

* N/A.

### Visual Regression Changes

* N/A.

## QA Brief

* GitHub Actions should only be run when code is pushed to a PR that affects said action—PHP changes should trigger PHPCS and PHP Unit runs, but not JS tests/linting, etc.
* Push a test branch to GitHub with only PHP changes. Observe that only PHP and E2E tests run, not JS/CSS-related tests.
* Push another commit with only JS/CSS changes. Observe only JS/CSS/E2E tests are run.

## Changelog entry

* N/A

felixarntz:
[...]

PR #3046, opened 2021-03-30
Improve GitHub Actions by running them selectively

tofumatt:
## Summary

Addresses issue #2969.

## Relevant technical choices

Adds path-ignore rules for E2E tests that prevent them running when test/storybook-only code is modified.

## Checklist


.github/workflows/e2e-tests-wp-nightly.yml
```diff
@@ -1,11 +1,48 @@
...
+      - '.storybook/**'
+      - 'stories/**'
+      - 'tests/backstop/**'
+      - 'tests/phpunit/**'
+      - 'assets/**/__tests__/**/*.js'
+      - 'assets/**/test/*.js'
+      - 'assets/**/?(*.)test.js'
+      - 'packages/**/__tests__/**/*.js'
+      - 'packages/**/test/*.js'
+      - 'packages/**/?(*.)test.js'
```
aaemnnosttv, line 27: I don't think this will work as expected because filter patterns aren't regexs.

See the docs here: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

.github/workflows/e2e-tests-wp-nightly.yml
```diff
@@ -1,11 +1,48 @@
...
 
+defaults:
+  timeout-minutes: 20
+
 on:
   push:
     branches:
       - develop
       - main
+    paths-ignore:
```
aaemnnosttv, line 11: I think we should probably always include a `paths` list of patterns rather than ignore only.

Maybe this would be:
* assets/**
* includes/**
* /google-site-kit.php
* lock files (any dependency changes)
* config files

While continuing to ignore tests

That way, if we add some new file outside of the ignore patterns it wouldn't trigger E2E.

.github/workflows/js-css-lint.yml
```diff
@@ -1,11 +1,42 @@
...
       - main
+    # Only run if CSS/JS/MD-related files changed.
+    paths:
+      - '**.js'
+      - '**.json'
+      - '**.cjs'
+      - '**.css'
+      - 'packages/**/*.md'
+      - '.eslint*'
+      - '.markdownlint*'
```
aaemnnosttv, line 19: I don't think we currently have any linting for markdown files.

.github/workflows/js-css-lint.yml
```diff
@@ -1,11 +1,42 @@
...
 on:
   push:
     branches:
       - develop
       - main
+    # Only run if CSS/JS/MD-related files changed.
+    paths:
+      - '**.js'
+      - '**.json'
+      - '**.cjs'
```
aaemnnosttv, line 15: I don't think this is used either.
```suggestion
```

.github/workflows/compressed-size.yml
```diff
@@ -24,25 +27,27 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event.pull_request.draft == false
     steps:
+      - uses: styfle/cancel-workflow-action
```
aaemnnosttv, line 30: We should include the version number in the reference here. [GitHub strongly recommends against using actions without a version constraint](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsuses).
```suggestion
      - uses: styfle/cancel-workflow-action@0.8.0
```

.github/workflows/php-lint.yml
```diff
@@ -1,11 +1,30 @@
...
   push:
     branches:
       - develop
       - main
+    # Only run if PHP-related files changed.
+    paths:
+      - '**.php'
+      - 'phpcs.xml.dist'
+      - 'phpmd.xml'
+      - 'phpstan.neon.dist'
```
aaemnnosttv, line 16: We don't have this 😄 
```suggestion
```

.github/workflows/php-tests-wp-4-7-php-5-6.yml
```diff
@@ -1,11 +1,30 @@
...
       - develop
       - main
+    # Only run if PHP-related files changed.
+    paths:
+      - '**.php'
+      - 'phpcs.xml.dist'
+      - 'phpmd.xml'
+      - 'phpstan.neon.dist'
+      - 'composer.json'
+      - 'composer.lock'
```
aaemnnosttv, line 18: We don't have a `phpmd.xml` or `phpstan` config file so let's exclude those.

Also, we should probably add `phpunit*.xml*` to cover dist + multisite variants.

.github/workflows/storybook.yml
```diff
@@ -1,11 +1,42 @@
...
 on:
   push:
     branches:
       - develop
       - main
+    # Only run if CSS/JS/MD-related files changed.
+    paths:
+      - '**.js'
+      - '**.json'
+      - '**.cjs'
```
aaemnnosttv, line 15:
```suggestion
```

aaemnnosttv (changes requested):
Thanks @tofumatt ! This looks like a great start, just a few things to tweak.

A few things to keep in mind:

- Much of this is duplicated throughout so please update all instances where feedback applies
- Let's make sure we're only referencing files/paths that apply to our codebase
- We can't merge this until required checks are updated (read: all of these jobs can no longer be required) since this PR will mean they no longer run in all cases which would block us until this is updated in the repo settings
    - This is more for clarity since this PR can't be merged anyways for the same reason 😄

.github/workflows/e2e-tests-wp-nightly.yml
tofumatt: I saw that but thought it would work, but looking again I think you're right. Okay, I'll tweak! 👍🏻

.github/workflows/compressed-size.yml
tofumatt: Good point 👍🏻

.github/workflows/compressed-size.yml
```diff
@@ -10,39 +13,41 @@ on:
...
-      - 'assets/svg/**/*.svg'
-      - './*.config.js'
-      - './package-lock.json'
+      - "assets/js/**.js"
+      - "!assets/js/**.test.js"
+      - "!assets/js/**/test/**"
+      - "assets/sass/**/*.scss"
+      - "assets/svg/**/*.svg"
+      - "./*.config.js"
+      - "./package-lock.json"
```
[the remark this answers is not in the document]
tofumatt: Argh, I didn't notice my editor was auto-saving YAML in weird ways; I didn't write these like this, thanks 👍🏻

.github/workflows/e2e-tests-wp-nightly.yml
```diff
@@ -1,11 +1,68 @@
...
+      # Ignore Storybook-only and unit-test-only changes when checking to run
+      # E2E tests. This is useful when adding/modifying/refactoring/removing
+      # test code ONLY in a commit, but saves running the expensive E2E test
+      # suite on GitHub Actions when all that's changing are tests and not code
+      # that would affect E2E behavior.
+      - ".babelrc"
+      - "assets/**"
+      - "includes/**"
+      - "google-site-kit.php"
+      - "feature-flags.json"
```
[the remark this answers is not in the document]
tofumatt: But changing which ones are available might warrant a re-run, no?

.github/workflows/visual-regression.yml
```diff
@@ -1,10 +1,24 @@
...
       - develop
+    # Only run if CSS/JS/MD-related files changed.
+    paths:
+      - "**.js"
+      - "**.json"
+      - "**.css"
+      - ".eslint*"
+      - ".nvmrc"
+      - ".prettier*"
+      - ".stylelint*"
```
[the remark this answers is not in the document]
tofumatt: Fair point 👍🏻

.github/workflows/visual-regression.yml
```diff
@@ -1,10 +1,24 @@
...
+
 on:
   pull_request:
     branches:
       - main
       - develop
+    # Only run if CSS/JS/MD-related files changed.
+    paths:
+      - "**.js"
+      - "**.json"
```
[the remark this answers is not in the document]
tofumatt: My thinking was fixture files actually, but I guess if they're all in `assets/**` anyway that'll work. Good call 👍🏻

tofumatt:
I spent a few minutes wondering why the checks weren't running, then realised they wouldn't for these changes.

I wonder if we should run the checks anytime the workflow files change to ensure they are working correctly? I'm going to do that for now to ensure we don't accidentally land bad workflows and force most to run when the workflows themselves change.

.github/workflows/visual-regression.yml
```diff
@@ -5,6 +5,19 @@ on:
...
       - develop
+    # Only run if CSS/JS/MD-related files changed.
+    paths:
+      - '.github/workflows/visual-regression.yml'
+      - 'assets/**'
+      - 'packages/**'
+      - 'tests/backstop/**'
+      - '.storybook/**'
+      - '**.scss'
+      - '.eslint*'
```
aaemnnosttv, line 16: Let's remove eslint here since it doesn't affect VRT.
```suggestion
```

.github/workflows/visual-regression.yml
```diff
@@ -5,6 +5,19 @@ on:
...
+      - '.nvmrc'
+      - '.prettier*'
```
aaemnnosttv, line 18: We don't have any prettier config files so let's remove this.
```suggestion
```

.github/workflows/visual-regression.yml
```diff
@@ -5,6 +5,19 @@ on:
     branches:
       - main
       - develop
+    # Only run if CSS/JS/MD-related files changed.
+    paths:
+      - '.github/workflows/visual-regression.yml'
+      - 'assets/**'
+      - 'packages/**'
```
aaemnnosttv, line 12: Let's exclude tests here as well.

.github/workflows/storybook.yml
```diff
@@ -5,7 +5,35 @@ on:
...
+      - '.nvmrc'
+      - '.prettier*'
+      - '**/package.json'
+      - 'package-lock.json'
   pull_request:
+    # Only run if CSS/JS/MD-related files changed.
+    paths:
+      - '.github/workflows/storybook.yml'
+      - 'assets/**'
+      - 'packages/**'
```
aaemnnosttv, line 27: Let's exclude tests here as these aren't relevant to Storybook.

.github/workflows/storybook.yml
```diff
@@ -5,7 +5,35 @@ on:
...
+      - 'tests/backstop/**'
```
aaemnnosttv, line 28: Storybook changes are relevant to VRT, but backstop changes aren't relevant for Storybook 😄 
```suggestion
```

.github/workflows/storybook.yml
```diff
@@ -5,7 +5,35 @@ on:
...
+      - '.storybook/**'
+      - '!.storybook/storybook-data.js'
+      - '**.scss'
+      - '.eslint*'
```
aaemnnosttv, line 32:
```suggestion
```

.github/workflows/storybook.yml
```diff
@@ -5,7 +5,35 @@ on:
...
+      - '.nvmrc'
+      - '.prettier*'
```
aaemnnosttv, line 34:
```suggestion
```

.github/workflows/storybook.yml
```diff
@@ -5,7 +5,35 @@ on:
...
+      - '**/package.json'
+      - 'package-lock.json'
```
aaemnnosttv, line 21: Same here 🙂

.github/workflows/php-tests-wp-nightly-php-7-4.yml
```diff
@@ -32,6 +51,7 @@ jobs:
       MYSQL_ROOT_PASSWORD: root
       WP_VERSION: nightly
     steps:
+      - uses: styfle/cancel-workflow-action@0.8.0
       - uses: actions/checkout@v2
       - name: Grant database access
```
aaemnnosttv, line 56: Looks like this PR needs to be synced with develop since this step is no longer here.

.github/workflows/php-lint.yml
```diff
@@ -5,7 +5,23 @@ on:
     branches:
       - develop
       - main
+    # Only run if PHP-related files changed.
+    paths:
+      - '.github/workflows/php-lint.yml'
+      - '**.php'
+      - 'phpcs.xml.dist'
+      - 'phpmd.xml'
```
aaemnnosttv, line 13: We don't have this file.
```suggestion
```

.github/workflows/php-lint.yml
```diff
@@ -5,7 +5,23 @@ on:
...
+      - 'composer.json'
+      - 'composer.lock'
   pull_request:
+    # Only run if PHP-related files changed.
+    paths:
+      - '.github/workflows/php-lint.yml'
+      - '**.php'
+      - 'phpcs.xml.dist'
+      - 'phpmd.xml'
```
aaemnnosttv, line 22:
```suggestion
```

.github/workflows/js-tests.yml
```diff
@@ -5,7 +5,29 @@ on:
     branches:
       - develop
       - main
+    # Only run if CSS/JS/MD-related files changed.
+    paths:
+      - '.github/workflows/js-tests.yml'
+      - '**.js'
+      - '**.json'
+      - '**.scss'
```
aaemnnosttv, line 13: Why would JS tests be concerned with `scss` file changes? 🤔

.github/workflows/js-tests.yml
```diff
@@ -5,7 +5,29 @@ on:
...
+      - '.eslint*'
```
aaemnnosttv, line 14: This workflow doesn't do JS linting yet, so let's leave this out for now.
```suggestion
```

.github/workflows/js-tests.yml
```diff
@@ -5,7 +5,29 @@ on:
...
+      - '.nvmrc'
+      - '.prettier*'
```
aaemnnosttv, line 16:
```suggestion
```

.github/workflows/js-tests.yml
```diff
@@ -5,7 +5,29 @@ on:
     branches:
       - develop
       - main
+    # Only run if CSS/JS/MD-related files changed.
+    paths:
+      - '.github/workflows/js-tests.yml'
+      - '**.js'
+      - '**.json'
```
aaemnnosttv, line 12: Any json file would also include `composer.json` so we should be a tiny bit more specific, or at least exclude a few specific ones like this `.eslintrc.json`, and `feature-flags.json`.

.github/workflows/js-tests.yml
```diff
@@ -5,7 +5,29 @@ on:
...
+      - '**/package.json'
+      - 'package-lock.json'
   pull_request:
+    # Only run if CSS/JS/MD-related files changed.
+    paths:
+      - '.github/workflows/js-tests.yml'
+      - '**.js'
+      - '**.json'
+      - '**.scss'
+      - '.eslint*'
```
aaemnnosttv, line 26:
```suggestion
```

.github/workflows/js-tests.yml
```diff
@@ -5,7 +5,29 @@ on:
...
+      - '.nvmrc'
+      - '.prettier*'
```
aaemnnosttv, line 28:
```suggestion
```

.github/workflows/js-css-lint.yml
```diff
@@ -5,7 +5,31 @@ on:
     branches:
       - develop
       - main
+    # Only run if CSS/JS/MD-related files changed.
+    paths:
+      - '.github/workflows/js-css-lint.yml'
+      - '**.js'
+      - '**.json'
```
aaemnnosttv, line 12: Similar to my comment on the JS tests workflow, we should exclude irrelevant json files such as `composer.json`. See other comment.

.github/workflows/js-css-lint.yml
```diff
@@ -5,7 +5,31 @@ on:
...
+      - '**.scss'
+      - '.eslint*'
+      - '.nvmrc'
+      - '.prettier*'
```
aaemnnosttv, line 16:
```suggestion
```

.github/workflows/js-css-lint.yml
```diff
@@ -5,7 +5,31 @@ on:
...
   pull_request:
+    # Only run if CSS/JS/MD-related files changed.
+    paths:
+      - '.github/workflows/js-css-lint.yml'
+      - '**.js'
+      - '**.json'
+      - '**.scss'
+      - '.eslint*'
+      - '.nvmrc'
+      - '.prettier*'
```
aaemnnosttv, line 29:
```suggestion
```

.github/workflows/e2e-tests-wp-nightly.yml
```diff
@@ -5,7 +5,61 @@ on:
...
+      - 'includes/**'
+      - 'google-site-kit.php'
+      - 'jest-puppeteer.config.js'
+      - 'composer.json'
+      - 'composer.lock'
+      - 'package.json'
+      - 'package-lock.json'
+      - 'postcss.config.js'
+      - 'scoper.inc.php'
+      - 'uninstall.php'
```
aaemnnosttv, line 54: We can simplify a few of these root php ones to a single `/*.php` like we have in a few of the other workflows.

.github/workflows/e2e-tests-wp-nightly.yml
```diff
@@ -5,7 +5,61 @@ on:
...
+      - 'assets/**'
+      - 'includes/**'
+      - 'google-site-kit.php'
+      - 'jest-puppeteer.config.js'
+      - 'composer.json'
+      - 'composer.lock'
+      - 'package.json'
+      - 'package-lock.json'
+      - 'postcss.config.js'
+      - 'scoper.inc.php'
```
aaemnnosttv, line 53: We should also include `/php-scoper/composer.lock` since we manage the installation of php-scoper separately here.

aaemnnosttv (commented):
Thanks @tofumatt – a few things left to update but I think this should be the last pass needed. I left a number of comments (many with inline suggestions) to help but didn't comment on every instance where it was already mentioned.

.github/workflows/e2e-tests-wp-nightly.yml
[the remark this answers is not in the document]
aaemnnosttv: That makes sense but this file is really only used by the tester plugin right now. The plugin itself has no awareness of available features, only enabled features. Feature activation is controlled entirely by the filter when `enabled()` is called.

.github/workflows/visual-regression.yml
tofumatt: My concern was that _eventually_ we might have our own packages that would warrant running the regression tests, but I guess we can cross that bridge when we come to it 😅

.github/workflows/storybook.yml
tofumatt: Hah, right, my brain got turned around 😆

.github/workflows/js-tests.yml
tofumatt: Ah, for us I guess they wouldn't. Probably in general they won't, good point 👍🏻

tofumatt:
Awesome, thanks for the review. Should be good-to-go now 👍🏻

.github/workflows/visual-regression.yml
aaemnnosttv: I don't mind including `packages/` here if you want, I was just saying we should exclude JS tests the paths 😄

.github/workflows/storybook.yml
aaemnnosttv: Please exclude tests from triggering this workflow.

aaemnnosttv (commented):
_So close_ 😅   The only thing still missing are the exclusions for test paths for Storybook and VRT, then this is 👌

tofumatt:
Ah, right! Okay, added the test paths to the exclude list now. 👍🏻

aaemnnosttv (approved):
Awesome! Thanks @tofumatt 🎉

Issue #3018, opened 2021-03-24
Ad Blocker incorrectly detected when using GoDaddy CDN


jamesozzie:
## Bug Description

As reported in #3007 when using the GoDaddy CDN Site Kit is incorrectly detecting an Ad Blocker, resulting in non reporting of AdSense dashboards and the inability to connect the module. 

A temporarily workaround for impacted users is to deactivate this CDN configuration from their GoDaddy control panel. 

## Screenshots
 
![image](https://user-images.githubusercontent.com/41326532/112326956-e9d13e80-8cac-11eb-8a07-7313b237e123.png)

## Impacted Users

In addition to the helpful insights @rafapontederino provided there have been 3 additional reports in the support forms, details below, all occurring in the past week:

- [Report 1](https://github.com/google/site-kit-wp/issues/3007) _**unresolved**_ ([SH info provided](https://docs.google.com/forms/d/1lhjns2K7c0Ny7ME4gxTcsnYtNUtcMr8BvFpPQi23Y1E/edit#response=ACYDBNh0jEAF3gSYjbYC-KvfLitAqa3rI5m-STxEhLDndzyRh1SLB1MFHyEgOtE8r9nvsXg))
- [Report 2](https://wordpress.org/support/topic/ad-blocker-detected-you-need-to-disable-it-in-order-to-set-up-adsense-2) _**unresolved**_ ([SH info provided](https://docs.google.com/forms/d/1lhjns2K7c0Ny7ME4gxTcsnYtNUtcMr8BvFpPQi23Y1E/edit#response=ACYDBNik_JWvvqQPOfYIGqQPkDJ_chOIDKDXyImYwWERLGZKWeAXvlmD4K0LQj5ZPwE98Lk))
- [Report 3](https://wordpress.org/support/topic/ad-blocker-detected-you-need-to-disable-it-in-order-to-set-up-adsense-3) **_unresolved_** ([SH info provided](https://docs.google.com/forms/d/1lhjns2K7c0Ny7ME4gxTcsnYtNUtcMr8BvFpPQi23Y1E/edit#response=ACYDBNi7daldAUBE5p1kdG0XMwmH_gs9bp_betJdB_nRNjtKG7chm4J8iD92WV2NJxj9IuM))
- [Report 4](https://wordpress.org/support/topic/ad-blocker-detected-you-need-to-disable-it-in-order-to-set-up-adsense/) _**unresolved**_ 
[...]

PR #3055, opened 2021-03-31
Fix/3018 adblocker fallback

aaemnnosttv:
## Summary

Addresses issue #3018

## Relevant technical choices

* Updates fallback mechanism to no longer follow redirects which is the default behavior for `fetch`
* Adds `timestamp` for cache busting to ensure we always make the request rather than serve from browser cache which would bypass detection
* Unfortunately `fetchMock` doesn't seem to handle redirects the same as the browser does so we can't add a test to cover this case specifically

**Before**
image

**After**
image

## Checklist


assets/js/modules/adsense/datastore/adblocker.js
```diff
@@ -63,8 +68,21 @@ export const controls = {
...
+				addQueryArgs(
+					'/favicon.ico',
+					{
+						// The name of the parameter here doesn't really matter
+						// since adblockers look at the URL as a whole.
+						'google-site-kit': '/adsense/pagead2.googlesyndication.com/pagead/js/adsbygoogle.js',
+						// Add a timestamp for cache-busting.
+						timestamp: Date.now(),
+					}
+				),
```
eugene-manuilov, line 80: The only thing that concerns me is that `addQueryArgs` escapes arguments which means that `/adsense/pagead2.googlesyndication.com/pagead/js/adsbygoogle.js` will be escaped to `%2Fadsense%2Fpagead2.googlesyndication.com%2Fpagead%2Fjs%2Fadsbygoogle.js`. However, I don't think it should bother adblocker and it should still detect it. :thinking:

eugene-manuilov (approved):
Going to approve it with one comment. Assigning to you, @aaemnnosttv, to take a look at my comment and merge this PR if you have tested it and it works for you.

assets/js/modules/adsense/datastore/adblocker.js
aaemnnosttv: Good catch! I found that certain adblockers such as uBlock Origin did not work as before when the value was encoded. I've removed the use of `addQueryArgs` which was mostly added to make the construction of the URL a bit easier to read. It now uses a simpler method and explicitly mentions the importance of not being encoded.

aaemnnosttv:
Thanks for catching that @eugene-manuilov – back to you for another pass 👍

eugene-manuilov (approved):

Issue #3026, opened 2021-03-25
Verify that Help/header icons should be highlighted in inactive state


marrrmarrr:
## Bug Description

The Help icon is highlighted even when the menu is deactivated. 

## Steps to reproduce

1. Click on the Help icon, menu dropdown appears
2. Click on the Help icon again, the menu disappears, but the icon stays highlighted. 
This happens both on the Site Kit Service and in product, with different visual treatments: 

Let's verify that (1) this is expected behaviour and if yes (2) that this is the expected visual treatment. 
cc @aarondruck for an expert consult.

## Screenshots

Highlighted icon on Site Kit Service:
![SKS-help](https://user-images.githubusercontent.com/13304791/112533620-6c8bf380-8daa-11eb-93ae-81455c5dd29c.png)

Highlighted icon in product:
![plugin-help](https://user-images.githubusercontent.com/13304791/112533657-77468880-8daa-11eb-8814-ae25788b6206.png)

## Additional Context

 - PHP Version: 
 - Plugin Version [e.g. 22]

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The following UX bugs with the help visibility icon in the Site Kit header bar (behind `helpVisibility` feature flag) should be fixed:
    * The blue outline that is displayed e.g. when clicking the menu icon to close the menu should be completely removed (applies to both active and focus state). This should be removed from _all_ header buttons, in fact—see https://github.com/google/site-kit-wp/issues/3026#issuecomment-811074432.
    * When clicking a link from the menu (opens in a new tab) and then going back...
        * ...the menu icon should be visually active but the menu should be closed. It should still be the currently focused item.
        * ...clicking on the menu icon should open the menu again. Currently this is clearly broken, as the first click in this scenario doesn't do anything, and only on the second click it will reopen.
* As a guidance, in terms of the above three bullet points it should behave similar to the question mark icon you can see e.g. in [Google Drive](https://drive.google.com/drive/my-drive). But the help icon should still be visually highlighted upon return, as it is the focused element and that visual indicator is useful to a keyboard user, especially after the menu was closed on clicking a link.

## Implementation Brief

* edit the `HelpMenu` component and remove the `<Fragment>`
* edit the file `assets/sass/components/global/_googlesitekit-dropdown-menu.scss`
* add the following to the class `.googlesitekit-header_dropdown` to the file with:
	`&:focus {
		outline: 0;
	}`
[...]

PR #3063, opened 2021-03-31
Update to fix help menu button bug.

Hazlitte:
## Summary

Update to fix help menu button bug.

Addresses issue #3026 

## Checklist


Hazlitte:
Hi @tofumatt, this PR includes the code for passing an `onSelected` function from the parent to the `Menu` component.  I looked at adding a default prop function to the `Menu` component as an alternative and didn't find a quick solution.  So I thought it would be best to solve the bug and then pass it over to you to have a look at the `Menu` component.

tofumatt (changes requested):
Ohhh, I see, `menuOpen` is a prop that `Menu` takes, so internally closing the menu isn't great as the [`useEffect`](https://github.com/google/site-kit-wp/blob/111c6653e27da89ed30568ee425fcbc0d27036e7/assets/js/components/Menu.js#L67-L73) in the `Menu` component will override it.

We could change that by adjusting `Menu`'s props API to handle the `menuOpen` internally when the `menuOpen` prop isn't a boolean (so `undefined` would cause internal handling but `true`/`false` would cause prop-based handling), which is a _nicer_ API but out-of-scope of this issue.

And since the other menus already handle it, let's leave that alone for now. It could be nice as a follow-up, code quality issue but let's see if we need that before we do it 🙂 

----

In that case, this largely looks good, but it looks like one of the menu buttons doesn't have the `className` we're targeting (`.googlesitekit-header__dropdown `), so the outlines in the header look like this:



I think that `className` should be added to that date menu button as well, assuming no other adverse side effects, so they all get the new outline-less styling. After that this is good-to-go 👍🏻

tofumatt:
@googlebot I consent.

tofumatt (approved):
Thanks, looks good 👍🏻

tofumatt (approved):

Issue #2616, opened 2021-01-13
Add new Site Kit illustrations


felixarntz:
Tracking issue for exchanging the existing illustrations in the plugin for new branding illustrations, all SVG. Somewhat related is #2468. This should either be behind the `userInput` flag (since those will be rolled out together), or behind a separate illustration-specific flag.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The following images should be removed and replaced by corresponding new "side kick" illustrations:
    * `rocket.svg`: With green success illustration
    * `person-sitting.svg`: With user input prompt illustration (used only in user input)
    * `congrats.svg`: With blue milestone illustration (used only in user input for now)
    * `sun-small.svg`: With notification alert illustration (used only for module alerts, i.e. AdSense, for now)
    * `thumbs-up.svg`: With award illustration (used when all modules have been activated)
    * `person-walking.svg`: With welcome illustrations (used for setup behind `serviceSetupV2` feature flag)

## Implementation Brief

* Exchange all SVG graphics per the ACs.
* Merge #3077.

### Test Coverage

* N/A

### Visual Regression Changes

* VRT reference images have to be refreshed to account for expected difference in illustrations.

## QA Brief

* Ensure the SVG files outlined in the ACs are no longer in the codebase.
* Ensure the following areas have the above graphics:
    * In the success notification for connecting Site Kit or activating a module, the green jumping sidekick should appear.
    * When having activated all modules and going to "Connect more services", the yellow sidekick holding an award should appear.
    * When there is a remote AdSense alert notification (shows in the AdSense module page if you have one), it should show the blue sidekick's arm holding a red warning sign.
    * With the `serviceSetupV2` feature flag active:
        * The Site Kit setup screen should show the four sidekicks as a welcome.
    * With the `userInput` feature flag active:
        * When visiting the dashboard as an existing user that has not filled the user input, the user input prompt notification should show the green sidekick with a pencil filling the user input.
        * When completing the user input questions, the success notification should show the blue sidekick holding up a star.

## Changelog entry

* Update assets across the plugin with new side kick-based branding illustrations.

aaemnnosttv:
IB ✅ 

Going to bump the estimate a tiny bit though to allow a little more time for CR + QA (particularly the latter).

wpdarren:
Adding a note to confirm I have asked on Slack for the illustrations in Figma, so we can compare what is on the  test site.

[...]

PR #3077, opened 2021-04-01
Replace old graphics with new side kick illustrations

felixarntz:
## Summary

Addresses issue #2616

## Relevant technical choices

* Also updates VRT reference images for the 3 stories where any of those updated illustrations are used.

## Checklist


assets/svg/award.svg
```diff
@@ -0,0 +1 @@
...0,0,1,71.91,138a28,28,0,0,1-7.23,4.93"/><line class="cls-8" x1="269.28" y1="98.8" x2="244.62" y2="190.46"/><path class="cls-9" d="M197.37,113.58a11.4,11.4,0,1,0,1.41-22.31,38.5,38.5,0,0,1-8.49-.63,22.81,22.81,0,1,1,28-16.1"/><path class="cls-9" d="M324.07,147.67a11.4,11.4,0,1,1,10-20s4.61,2.89,7.65,3.71a22.8,22.8,0,1,0-16.09-27.95"/><path class="cls-10" d="M253.5,157.45c37.27,10,80.69-30.92,97-91.47l-135-36.32C199.23,90.21,216.23,147.42,253.5,157.45Z"/><path class="cls-11" d="M224,109.65c-5.87-19.74-6.24-43.45-1.16-67.66"/><path class="cls-11" d="M241.18,139.46a54.17,54.17,0,0,1-11.61-15.67"/><path class="cls-10" d="M248,177.83c-13.07-3.52-26,2.23-28.83,12.84L266.5,203.4C269.36,192.79,261.08,181.34,248,177.83Z"/><path class="cls-9" d="M266,110.86l-17.44,3.92a2.18,2.18,0,0,1-2.46-3l7.5-16.23-9.12-15.37a2.19,2.19,0,0,1,2.14-3.28L264.4,79l11.8-13.42a2.18,2.18,0,0,1,3.78,1l3.48,17.53,16.41,7.08a2.18,2.18,0,0,1,.2,3.91l-15.6,8.72-1.66,17.79a2.18,2.18,0,0,1-3.66,1.4Z"/><path class="cls-12" d="M74.48,209.27c74.32,45.28,152.4-1.07,170.14-18.81"/><path class="cls-12" d="M87.61,216.33a129.71,129.71,0,0,0,13.13,5.3"/><path class="cls-3" d="M74.48,209.27a136.76,136.76,0,0,0,13.13,7.06"/></svg>
```
aaemnnosttv, line 1: Let's update the formatting to be consistent with other source files. The compression and optimization should happen at build time, but the source file should be optimized for readability. That is, I would expect this to be formatted more like a normal xml document like `assets/svg/logo-g.svg`.

Also, we should check if attributes like `id="_1_Step_Darker" data-name="1 Step Darker"` are preserved when importing, and if so remove them here since they shouldn't be relevant.

assets/svg/success-green.svg
```diff
@@ -0,0 +1 @@
....46" x2="279.29" y2="62.66"/><path class="cls-2" d="M251.2,51.21c-.67-1,10.33-9.56,11.16-10.16,5.81-4.22,12.64-8,19.94-8.46,7.91-.53,15.45,2.58,23,4.58,10.31,2.77,21.44,3.8,32.09,1.69,9.59-1.9,20.83-5.9,27.87-12.93-.24.24.3,2.83.33,3.18a65.31,65.31,0,0,1,0,8.51,59.1,59.1,0,0,1-7,25.06,43.41,43.41,0,0,1-18.94,18.13C329.14,86,317.78,88.93,307,93.41c-6.7,2.79-18.11,7.23-18.11,16.07Z"/><line class="cls-3" x1="248.83" y1="47.55" x2="342.46" y2="192.24"/><path class="cls-4" d="M279.89,172.88c18.52,6.68,53.35,9.87,64.53-10.87,14.77-27.39-23.47-38.52-43.87-34.53"/><path class="cls-4" d="M88.94,145C13,233.15,70,296.93,119,267.62"/><path class="cls-5" d="M234.54,260.65c59.23,111,7.37,98-23.53,93.23l2,20.19"/><path class="cls-5" d="M186.58,274.6c31.93,144.76-28.4,65.93-35.58,53.33l-12.25,14.36"/><path class="cls-6" d="M234.54,260.65q3.92,7.32,7.19,14"/><path class="cls-6" d="M186.58,274.6q2.19,9.88,3.81,18.42"/><polygon class="cls-7" points="170.71 172.85 72.71 212.2 112.06 310.2 210.06 270.85 308.06 231.5 268.71 133.5 170.71 172.85"/><path class="cls-8" d="M158.45,268.63C142.2,271,125.93,263.43,114.7,250l83.92-22.8"/><line class="cls-3" x1="316.62" y1="152.31" x2="342.46" y2="192.24"/></svg>
```
aaemnnosttv, line 1: It's a bit odd that these all have the same `id` and `data-name` attributes. I think these should be removed as mentioned before but seems strange that all of these would have these in common.

aaemnnosttv (changes requested):
Thanks @felixarntz – this looks good, just a few things to review, mostly regarding the new source files.

felixarntz:
@aaemnnosttv Good catch! I've replaced the SVGs with more appropriate versions which have these fixed, as well as using inline attributes instead of style tags which would have global implications.

aaemnnosttv (approved):
Super!

Issue #3068, opened 2021-04-01
PHP Tests jobs sometimes fail to due to DB connection


aaemnnosttv:
## Bug Description

It seems that there may be a bit of a race condition when the PHP test jobs run in that the MySQL server may not be ready by the time we attempt to run mysql commands to set up the DB:

![image](https://user-images.githubusercontent.com/1621608/113267318-e2c6b380-92de-11eb-8de6-893c38eb7ef3.png)

The weird part is that it already seems to be doing something like this:

![image](https://user-images.githubusercontent.com/1621608/113272804-b150e680-92e4-11eb-8bdd-baa6e6439cea.png)

Relevant post: [Using the MySQL Service with Github Actions](https://firefart.at/post/using-mysql-service-with-github-actions/)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* PHP Tests job definitions should be updated to ensure the mysql service is properly waited for to ensure these jobs don't fail due to the above mentioned connection issue

## Implementation Brief

* Merge #3082

### Test Coverage

* N/A

### Visual Regression Changes

* N/A

## QA Brief

* PHP Tests check should stop failing due to MySQL connection issues.

## Changelog entry

* N/A

aaemnnosttv:
IB ✅

tofumatt:
QA ✅

PR #3082, opened 2021-04-02
Fix mysql issues during GH Action runs.

eugene-manuilov:
## Summary

Addresses issue #3068

## Checklist

### DIFF
--- a/.github/workflows/php-tests-wp-4-7-php-5-6.yml
+++ b/.github/workflows/php-tests-wp-4-7-php-5-6.yml
@@ -19,7 +19,6 @@ jobs:
       mysql:
         image: mysql:5.7
         env:
-          MYSQL_ALLOW_EMPTY_PASSWORD: yes
           MYSQL_ROOT_PASSWORD: wordpress
         ports:
           - 3306:3306
--- a/.github/workflows/php-tests-wp-4-7-php-5-6.yml
+++ b/.github/workflows/php-tests-wp-4-7-php-5-6.yml
@@ -20,25 +20,19 @@ jobs:
         image: mysql:5.7
         env:
           MYSQL_ALLOW_EMPTY_PASSWORD: yes
+          MYSQL_ROOT_PASSWORD: wordpress
         ports:
           - 3306:3306
         options: --health-cmd="mysqladmin ping" --health-interval=5s --health-timeout=2s --health-retries=3
     env:
       DB_HOST: 127.0.0.1
       DB_PORT: 3306
-      MYSQL_USER: wordpress
+      MYSQL_USER: root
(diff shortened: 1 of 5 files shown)


aaemnnosttv (commented):
@eugene-manuilov – it looks like the changes here may only solve the problems we were seeing due to the delayed execution of mysql commands; we're not truly waiting for the service to be fully available/ready.

See the note under ["No connections until MySQL init completes"](https://hub.docker.com/_/mysql#:~:text=No%20connections%20until%20MySQL%20init%20completes) on Docker hub. Ideally, we'd wait for this (maybe we could extract a common helper for use with E2E as well?). As it is, we're still assuming it's ready, just checking later.

If this is too much effort right now, we can go with this and revisit if it comes up again.

eugene-manuilov:
@aaemnnosttv I don't think we need to re-invent the wheel. GitHub Actions already do it for us:
github com_google_site-kit-wp_runs_2255462654_check_suite_focus=true

The runner waits till MySQL container is ready and only then starts executing our steps. Of course, we can use the following command instead of `mysqladmin ping`, but I think it is overkill and not really needed:

```sh-session
mysql -e "select 1" -u root -pwordpress > /dev/null 2>&1
```

aaemnnosttv:
> The runner waits till MySQL container is ready and only then starts executing our steps.

Right, that's partly what I mean. GitHub Actions was already doing this before and we were still seeing connection errors.

Is there a difference between the service being "healthy" and ready to accept commands?

eugene-manuilov:
> > The runner waits till MySQL container is ready and only then starts executing our steps.
> 
> Right, that's partly what I mean. GitHub Actions was already doing this before and we were still seeing connection errors.

Yes, I think you are right, GHA was doing it before and we were still seeing connection errors. :thinking:  

> Is there a difference between the service being "healthy" and ready to accept commands?

In our case, GHA uses what we put in `--health-cmd` argument: https://github.com/google/site-kit-wp/blob/05e1c2f9c856609f713bdf58e0d920a49a42199b/.github/workflows/php-tests-wp-latest-php-7-4.yml#L25

Looks like `phpadmin ping` doesn't guaranty that MySQL is ready to accept requests. If we want to be more precise, we can use MySQL command to execute `select 1` query as I posted above, but I still think that it is overkill and we don't need to change it.

aaemnnosttv:
> Looks like `phpadmin ping` doesn't guaranty that MySQL is ready to accept requests. If we want to be more precise, we can use MySQL command to execute `select 1` query as I posted above, but I still think that it is overkill and we don't need to change it.

Okay, let's go with how it is for now and reassess if needed 👍

aaemnnosttv (approved):
👍

Issue #2975, opened 2021-03-17
Add ga4setup feature flag


felixarntz:
There should be a new `ga4setup` feature flag that toggles whether the new GA4 support during the Analytics setup flow (plus related infrastructure such as the internal module, see #2974) is enabled. [See more information on the overall approach](https://docs.google.com/document/d/11DMYyAGhtFHO8bMH107pacC8iMpD9PJ9BTaR6IF_B5E/edit?resourcekey=0-0CgoW31PILS2t6JGZ6w_yA#heading=h.ft5rt9liuag2).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a new feature flag `ga4setup`.
* If the `ga4setup` feature flag is enabled, the new `Google\Site_Kit\Modules\Analytics_4` class (module slug `analytics-4`) should be registered alongside the existing core modules (see `Google\Site_Kit\Core\Modules\Modules`).
    * While this class is being implemented in #2974, the implementation of this issue can already get started regardless, since the class name is defined. #2974 is required to be merged first though.
* The `Google\Site_Kit\Core\Modules\Modules` PHP class should be further amended so that the `analytics-4` module's activation status always depends on the `analytics` module; in other words: Instead of storing `analytics-4` in the list of active modules in the database, rely on the value for `analytics` to determine activation status.

## Implementation Brief

* Add `ga4setup` to the feature flags array in the `feature-flags.json` file.
* Update the `Modules::__construct` method to add the new `Google\Site_Kit\Modules\Analytics_4` module to the `$core_modules` array if the `ga4setup` feature flag is enabled. Use the `Feature_Flags::enabled` method to determine whether the `ga4setup` flag is enabled.
* Update the `Modules::get_active_modules_option` method to add `analytics-4` slug to the `$option` variable if it contains the `analytics` slug.

### Test Coverage

* Update the existing phpunit tests if any is failing due to the aforementioned changes.

### Visual Regression Changes

* N/A

## QA Brief

* 

## Changelog entry

* N/A

aaemnnosttv:
@eugene-manuilov IB looks good, with one suggestion:
- Instead of modifying `Modules::get_active_modules_option`, how about we return `'force_active' => true,` from the GA4 module's `setup_info` method when Analytics is in the list of active module slugs. We'd probably have to check the option directly (module classes don't have access to `Modules` anyways) but this seems like it might be a better alternative. What do you think?

[...]

PR #3087, opened 2021-04-05
Enhancement/2975 ga4setup

eugene-manuilov:
## Summary

Addresses issue #2975

## Checklist


includes/Core/Modules/Modules.php
```diff
@@ -963,6 +969,10 @@ private function get_active_modules_option() {
 		$option = $this->options->get( self::OPTION_ACTIVE_MODULES );
 
 		if ( is_array( $option ) ) {
+			if ( in_array( 'analytics', $option, true ) ) {
+				$option[] = 'analytics-4';
+			}
+
```
aaemnnosttv, line 975: We should update this to work regardless of which option the value comes from (see legacy fallback below).

aaemnnosttv (changes requested):
Thanks @eugene-manuilov - just one small change needed then this should be good to go 👍

includes/Core/Modules/Modules.php
eugene-manuilov: Updated.

aaemnnosttv (approved):
Super, thanks!

includes/Core/Modules/Modules.php
```diff
@@ -962,17 +968,19 @@ private function get_module_schema() {
...
 
-		$legacy_option = $this->options->get( 'googlesitekit-active-modules' );
+		if ( ! is_array( $option ) ) {
+			$option = array();
+		}
 
-		if ( is_array( $legacy_option ) ) {
-			return $legacy_option;
+		if ( in_array( 'analytics', $option, true ) ) {
+			$option[] = 'analytics-4';
```
aaemnnosttv, line 980: One more thing – we should probably do one or both of these:
- ensure `analytics-4` is not returned by `get_active_modules_option` if  `analytics` is not in the list
- always remove `analytics-4` in `set_active_modules_option` so that `analytics-4` is never set in the option.

aaemnnosttv (commented):
One more thought here @eugene-manuilov – let me know what you think.

aaemnnosttv (approved):

Issue #2184, opened 2020-10-15
Refactor `SettingsModules` component


felixarntz:
The `SettingsModules` component should be refactored to be a functional component, relying fully on the datastore. It should be split into sub-components where it makes sense.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `SettingsModules` component should be refactored into two new functional, hook-based, datastore-driven components `SettingsActiveModules` and `SettingsInactiveModules`.
* Both of these components should render a list of `SettingsModule` components. The `SettingsModule` component should be refactored at a later point though, to not make this too much work in one issue/PR - all props that `SettingsModule` currently requires should be passed. This can be simplified later once the `SettingsModule` component gets simplified.
* The only thing that should be changed which also affects `SettingsModule` a bit is that the `handleButtonAction` function should be split into more targeted separate action handlers which should be passed as separate props from `SettingsModules` to `SettingsModule`:
    * `handleEdit` should be invoked when the user clicks on the "Edit" button (current action is `edit`)
    * `handleConfirm` should be invoked when the user clicks on the "Confirm" button when editing (current action is `confirm`)
    * `handleCancel` should be invoked when the user clicks on the "Cancel" button when editing (current action is `cancel`)
    * In `SettingsModule`, the three functions should then be invoked on the relevant buttons, instead of the single `handleEdit` with an action passed.
* Both new components should rely on the `getModules` selector of `core/modules` instead of the legacy `getModulesData()` function, then filter out active/inactive modules respectively, and then they should provide props based on the registered module data.
    * There are some pieces of information which are only present in `getModulesData` right now, but not in the data from the `getModules()` selector. When writing the IB, please list these pieces of data. We can then discuss the best way to make them available in the datastore-driven approach.

#### Step 1

* Add a new property `forceActive` to module information data that is returned from the `Modules::prepare_module_data_for_response` method. Use `$module->force_active` to get a value for this property.
* In the `assets/js/googlesitekit/modules/datastore/modules.js` file:
  * Add the new `features` property to the `moduleDefaults` object with an empty array value;
  * Add the new optional `features` argument to the `registerModule` action;
  * Add the new `getModuleFeatures` selector that returns `features` for a module.
* Update modules to set their `features` during module registration in javascript instead of defining the `provides` array in their PHP classes:
[...]

PR #3088, opened 2021-04-05
Update SettingsInactiveModules to preserve list based on initial active state

aaemnnosttv:
## Summary

Addresses issue #2184 (follow-up)

## Relevant technical choices

* Fixes a regression where the activated module would be removed from the list of modules to activate while the page was reloading
* Active state is determined by the server. As soon as the `getModules` selector returns a value, it is guaranteed to include the active state as it waits for server definitions before it returns. Here we store the inactive slugs as soon as they're available to capture only the modules which were initially inactive for rendering.

## Checklist

### DIFF
--- a/assets/js/components/settings/SettingsInactiveModules.js
+++ b/assets/js/components/settings/SettingsInactiveModules.js
@@ -19,6 +19,7 @@
 /**
  * WordPress dependencies
  */
+import { useEffect, useState } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 
 /**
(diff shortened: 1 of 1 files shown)


assets/js/components/settings/SettingsInactiveModules.js
```diff
@@ -34,15 +35,32 @@ import ThumbsUpSVG from '../../../svg/thumbs-up.svg';
...
 
 const SettingsInactiveModules = () => {
+	// We store `initialInactiveSlugs` separately to avoid
+	// layout shifts when activating a module as it would otherwise
+	// cause the activated module to be removed upon activation.
+	const [ initialInactiveSlugs, setInitialInactiveSlugs ] = useState();
 	const modules = useSelect( ( select ) => select( CORE_MODULES ).getModules() );
+	useEffect( () => {
+		// Only set initialInactiveSlugs once, as soon as modules are available.
+		if ( ! modules || initialInactiveSlugs !== undefined ) {
```
tofumatt, line 45: Pretty minor, but if `modules` are `null` or something, wouldn't that be valid but cause this to still be `! modules`? Not a big deal, but wouldn't it be safer to check for `undefined` only here?

tofumatt (approved):
Looks good! Tested locally and it works as expected 👍🏻

assets/js/components/settings/SettingsInactiveModules.js
aaemnnosttv: `modules` here should only ever be an object or `undefined` (which I suppose is the same for `initialInactiveSlugs`, except it uses an array it's just more obvious that it could be `undefined` here because the initial state is empty).

Either should be fine, and I usually go with whatever seems easiest to read.

Issue #2384, opened 2020-11-18
Make Reset button available in a consistent location (for "emergency")


felixarntz:
At the moment, Site Kit provides two places where the plugin can be reset:
* The setup screen, via a secondary action button
* The settings screen (Admin Settings tab)

It has proven multiple times to be a problem that both of these buttons can only be accessed conditionally. The settings button can only be accessed when connected (for obvious reasons), while the setup button can only be accessed when there is no other user that currently has credentials. While the latter should mostly show up when relevant, there is still potential that the plugin is stuck and needs to be reset while some users already (or rather still) have an access token, preventing the button from showing up in the setup screen.

We could make the button in the setup screen show up unconditionally (which is how it used to be), however that is not a good idea because then it becomes too "easy" to reset which may unexpectedly mess up the setup for other already connected users. Instead, we should add another Reset button in a "somewhat hidden/special" area which can always be accessed by an administrator.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Add a section in the Tools > Available Tools menu
* Heading: Reset Site Kit 
* Content: Resetting will disconnect all users and remove all Site Kit settings and data within WordPress. You and any other users who wish to use Site Kit will need to reconnect to restore access. 
* Link or button: Reset Site Kit 
* When clicked, Site Kit should be reset (similar to the existing Reset button in Site Kit).
* Only administrators (more specifically, users that can setup Site Kit) should be able to see this and reset.
* The implementation should be basic WordPress & PHP, not load/use React like we do usually.

## Implementation Brief

* Create a new `includes/Core/Admin/Available_Tools.php` file with a new class:
  * Use `Method_Proxy_Trait` trait to bind action/filter hooks with private methods;
  * Add a new private method `render_tool_box` that should do the following:
    * Do nothing if the current user can't set up Site Kit (`! current_user_can( Permissions::SETUP )`);
    * Output the following block using heading, content and button from the AC:
      ```html
      <div class="card">
          <h2 class="title">/* heading */</h2>
          <p>
              /* content */
              /* link styled as a button */
          </p>
      </div>
      ```
    * Use `admin_url( 'admin.php?action=googlesitekit_reset' )` to get reset URL and `wp_nonce_url` function to add a nonce argument to it.
  * Create a new public method `register` that adds a new hook for the `tool_box` action and calls the `render_tool_box` method (via using `get_method_proxy` from the aforementioned trait).
[...]

PR #3095, opened 2021-04-06
Add dedicated reset tool

aaemnnosttv:
## Summary

Addresses issue #2384

## Relevant technical choices

* Redirects to splash page rather than dashboard to avoid an immediate redirect from dashboard to splash due to missing authentication
* Reset also triggers client side session reset and adds success notification as are done/shown when doing a normal client-side reset

## Checklist

### DIFF
--- a/includes/Core/Admin/Available_Tools.php
+++ b/includes/Core/Admin/Available_Tools.php
@@ -42,13 +42,6 @@ class Available_Tools {
 		if ( ! current_user_can( Permissions::SETUP ) ) {
 			return;
 		}
-		$reset_url = add_query_arg(
-			array(
-				'action' => Reset::ACTION,
-				'nonce'  => wp_create_nonce( Reset::ACTION ),
-			),
-			admin_url( 'index.php' )
-		);
 		?>
 		<div class="card">
 			<h2 class="title"><?php esc_html_e( 'Reset Site Kit', 'google-site-kit' ); ?></h2>
@@ -63,7 +56,7 @@ class Available_Tools {
 			<p>
 				<a
 					class="button button-primary"
-					href="<?php echo esc_url( $reset_url ); ?>"
+					href="<?php echo esc_url( Reset::url() ); ?>"
 				>
 					<?php esc_html_e( 'Reset Site Kit', 'google-site-kit' ); ?>
 				</a>
--- a/includes/Core/Admin/Available_Tools.php
+++ b/includes/Core/Admin/Available_Tools.php
@@ -1,0 +1,74 @@
+<?php
+/**
+ * Class Google\Site_Kit\Core\Admin\Available_Tools
+ *
+ * @package   Google\Site_Kit\Core\Admin
+ * @copyright 2021 Google LLC
+ * @license   https://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
+ * @link      https://sitekit.withgoogle.com
+ */
+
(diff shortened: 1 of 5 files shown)


includes/Core/Util/Reset.php
```diff
@@ -178,4 +192,35 @@ private function get_rest_routes() {
...
+
+	/**
+	 * Handles the reset admin action.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $nonce WP nonce for action.
+	 */
+	private function handle_reset_action( $nonce ) {
+		if ( empty( $nonce ) || ! wp_verify_nonce( $nonce, static::ACTION ) ) {
```
eugene-manuilov, line 204: Checking whether `$nonce` is empty is redundant, `wp_verify_nonce` does the same.
```suggestion
		if ( ! wp_verify_nonce( $nonce, static::ACTION ) ) {
```

includes/Core/Util/Reset.php
```diff
@@ -178,4 +192,35 @@ private function get_rest_routes() {
...
+			wp_die( esc_html__( 'Invalid nonce.', 'google-site-kit' ), 400 );
```
eugene-manuilov, line 205: The `Invalid nonce.` error message is very confusing and doesn't tell users how they can solve this problem. Let's update it to something like `Your request to reset Site Kit is expired. Please, try again.` - where `try again` is the newly generated reset link.

includes/Core/Admin/Available_Tools.php
```diff
@@ -0,0 +1,74 @@
...
+		if ( ! current_user_can( Permissions::SETUP ) ) {
+			return;
+		}
+		$reset_url = add_query_arg(
+			array(
+				'action' => Reset::ACTION,
+				'nonce'  => wp_create_nonce( Reset::ACTION ),
+			),
+			admin_url( 'index.php' )
+		);
```
eugene-manuilov, line 51: Let's probably move it to a new public static method `Reset::get_reset_link`, so we can reuse it to generate reset links.

eugene-manuilov (changes requested):
In general, LGTM. Found a few things to update before we can merge it.

includes/Core/Util/Reset.php
aaemnnosttv: Sounds good to me. We do the same thing many other places though 😄  Will update.

includes/Core/Util/Reset.php
aaemnnosttv: I agree the messaging could be better here, but this is the same message we use in many other places for the same thing. Let's address this in a separate issue so we can improve it everywhere rather than just here.

includes/Core/Admin/Available_Tools.php
aaemnnosttv: I thought about this too and I agree. I was thinking maybe `Reset_Persistent` shouldn't have this method but we can always overload it there. Otherwise, it's fine if they both point to the same URL for now.

includes/Core/Util/Reset.php
aaemnnosttv: Created https://github.com/google/site-kit-wp/issues/3098 as this relates to a follow-up issue we were planning to open as well 👍

eugene-manuilov (approved):

felixarntz (approved):
Great!

Issue #3005, opened 2021-03-22
Ensure adding a www prefix doesn't result  in reporting of Analytics and AdSense from that date only (or vice versa)


jamesozzie:
## Bug Description

As raised in the [WordPress support forums](https://wordpress.org/support/topic/past-performance-not-showing-after-url-change-from-www-to-non-www-subdomain/page/2/#post-14219206) at present when a user adds www to their site URL they're then prompted to setup Site Kit once more, with the plugin identifying this new URL.

With any such www prefix modifications Site Kit should maintain the same Analytics or AdSense data, and not populate dashboards or display data based only on this new URL.  

## Steps to reproduce

1. Login to a site with Analytics and AdSense data
2. Add www in the defined URLs under "_Settings > General Settings_" in a standard WordPress setup
3. Login to your site once more and setup Site Kit
4. Wait for data to appear - the services will already be connected, but the dashboards and data that are populated will be based on the www version of your site only, with no legacy information from before the www change. 

## Screenshots

![image](https://user-images.githubusercontent.com/41326532/112059363-a9aa7880-8b53-11eb-8dce-6db36a9a08de.png)

## Additional Context

 - Tested and reproduced in support, Site Kit version 1.28.0
 - A filter was introduced recently for reporting on a WordPress specific install only, with some sites using WordPress within a subdirectory of a primary non WordPress domain

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Analytics reports should request data for the current domain **regardless of with or without `www.` prefix**.
    * The existing [Analytics report endpoint](https://developers.google.com/analytics/devguides/reporting/core/v4/rest/v4/reports/batchGet) dimension filter that filters by `ga:hostname` can be updated relying on the [`IN LIST` operator](https://developers.google.com/analytics/devguides/reporting/core/v4/rest/v4/reports/batchGet#operator).
* AdSense reports should request data for the current domain **regardless of with or without `www.` prefix**.
    * The existing [AdSense report endpoint](https://developers.google.com/adsense/management/v1.4/reference/accounts/reports/generate) `filter` parameter can be updated with the necessary [`OR` logic](https://developers.google.com/adsense/management/reporting/filtering#OR).

## Implementation Brief

* Update the `Analytics::create_analytics_site_data_request` method:
  * Use the `IN_LIST` operator instead of `EXACT` for the `ga:hostname` dimension;
[...]

PR #3107, opened 2021-04-07
Update to ensure that Adsense and Analytics data is correct, www prefix.

Hazlitte:
## Summary

Update to ensure that Adsense and Analytics data is correct.

Addresses issue #3005

## Checklist

### DIFF
--- a/includes/Core/Modules/Module.php
+++ b/includes/Core/Modules/Module.php
@@ -663,6 +663,31 @@ abstract class Module {
 	}
 
 	/**
+	 * Permutes hostname to cover all different variants of it (not considering the path).
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $site_url Site URL to get hostname permutations for.
+	 * @return array List of permutations.
+	 */
+	final protected function permute_hostname( $site_url ) {
+		$hostnames = array();
+
+		// Get hostname.
+		$host = wp_parse_url( $site_url, PHP_URL_HOST );
+
+		if ( strncmp( $host, 'www.', 4 ) === 0 ) {
+			$hostnames['with_www']    = $host;
+			$hostnames['without_www'] = substr( $host, 4 );
+		} else {
+			$hostnames['with_www']    = 'www.' . $host;
+			$hostnames['without_www'] = $host;
+		}
+
+		return $hostnames;
+	}
+
+	/**
 	 * Gets the Google client the module uses.
 	 *
 	 * This method should be used to access the client.
--- a/includes/Core/Modules/Module.php
+++ b/includes/Core/Modules/Module.php
@@ -663,31 +663,6 @@ abstract class Module {
 	}
 
 	/**
-	 * Permutes hostname to cover all different variants of it (not considering the path).
-	 *
-	 * @since n.e.x.t
-	 *
-	 * @param string $site_url Site URL to get hostname permutations for.
-	 * @return array List of permutations.
-	 */
-	final protected function permute_hostname( $site_url ) {
-		$hostnames = array();
-
-		// Get hostname.
-		$host = wp_parse_url( $site_url, PHP_URL_HOST );
-
-		if ( strncmp( $host, 'www.', 4 ) === 0 ) {
-			$hostnames['with_www']    = $host;
-			$hostnames['without_www'] = substr( $host, 4 );
-		} else {
-			$hostnames['with_www']    = 'www.' . $host;
-			$hostnames['without_www'] = $host;
-		}
-
-		return $hostnames;
-	}
-
-	/**
 	 * Gets the Google client the module uses.
 	 *
 	 * This method should be used to access the client.
--- a/includes/Modules/AdSense.php
+++ b/includes/Modules/AdSense.php
@@ -537,17 +537,10 @@ final class AdSense extends Module
 			$opt_params['maxResults'] = (int) $args['limit'];
 		}
 
-		$host = wp_parse_url( $this->context->get_reference_site_url(), PHP_URL_HOST );
+		$hostnames = $this->permute_hostname( $this->context->get_reference_site_url() );
 
-		if ( ! empty( $host ) ) {
-			if ( strpos( $host, 'www.' ) !== false ) {
-				$host_with_www    = $host;
-				$host_without_www = str_replace( 'www.', '', $host );
-			} else {
-				$host_with_www    = 'www.' . $host;
-				$host_without_www = $host;
-			}
-			$opt_params['filter'] = 'DOMAIN_NAME==' . $host_with_www . ',DOMAIN_NAME==' . $host_without_www;
+		if ( ! empty( $hostnames['with_www'] ) && ! empty( $hostnames['without_www'] ) ) {
+			$opt_params['filter'] = 'DOMAIN_NAME==' . $hostnames['with_www'] . ',DOMAIN_NAME==' . $hostnames['without_www'];
 		}
 
 		$service = $this->get_service( 'adsense' );
--- a/includes/Modules/AdSense.php
+++ b/includes/Modules/AdSense.php
@@ -537,13 +537,11 @@ final class AdSense extends Module
 			$opt_params['maxResults'] = (int) $args['limit'];
 		}
 
-		$hostnames = array_unique(
-			array_map(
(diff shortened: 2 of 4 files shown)


Hazlitte:
Hi @aaemnnosttv, The code changes in this PR are similar in two of the files and it may be preferable to create a separate function instead.  I wasn't sure where that function should be added and as it is urgent to get the issue into this release I focused on meeting the IB.  The creation of a separate function could be addressed in this PR or as a separate issue.

includes/Modules/AdSense.php
```diff
@@ -538,8 +538,16 @@ protected function create_adsense_earning_data_request( array $args = array() )
 		}
 
 		$host = wp_parse_url( $this->context->get_reference_site_url(), PHP_URL_HOST );
+
 		if ( ! empty( $host ) ) {
-			$opt_params['filter'] = 'DOMAIN_NAME==' . $host;
+			if ( strpos( $host, 'www.' ) !== false ) {
+				$host_with_www    = $host;
+				$host_without_www = str_replace( 'www.', '', $host );
```
aaemnnosttv, line 545: This is too loose in the search and replacement. It must only consider a hostname that starts with `www.`. If the hostname happens to include `www.` elsewhere, it doesn't matter.

Similarly, we only want to remove a leading `www.` if it is present. `str_replace` will replace all occurrences, so we can't use that.

Turns out we already have a method on the base `Module` class for doing this: `Module::permute_site_url` – the difference is that it returns an array of URLs, so we just need to map the return value into host names first, otherwise it does exactly what we need 👍

includes/Modules/Analytics.php
```diff
@@ -939,11 +939,20 @@ protected function create_analytics_site_data_request( array $args = array() ) {
...
-		$hostname         = wp_parse_url( $this->context->get_reference_site_url(), PHP_URL_HOST );
+		$hostname = wp_parse_url( $this->context->get_reference_site_url(), PHP_URL_HOST );
+
+		if ( strpos( $hostname, 'www.' ) !== false ) {
+			$hostname_with_www    = $hostname;
+			$hostname_without_www = str_replace( 'www.', '', $hostname );
+		} else {
+			$hostname_with_www    = 'www.' . $hostname;
+			$hostname_without_www = $hostname;
+		}
```
aaemnnosttv, line 950: See previous comment.

tests/phpunit/integration/Modules/AnalyticsTest.php
```diff
@@ -696,13 +696,14 @@ public function test_create_analytics_site_data_request() {
...
+		$this->assertEquals( 'IN_LIST', $filters[0]->getOperator() );
 
 		$hostname    = wp_parse_url( $context->get_reference_site_url(), PHP_URL_HOST );
 		$expressions = $filters[0]->getExpressions();
+
 		$this->assertTrue( is_array( $expressions ) );
-		$this->assertEquals( 1, count( $expressions ) );
-		$this->assertEquals( $hostname, $expressions[0] );
+		$this->assertEquals( 2, count( $expressions ) );
+		$this->assertContains( $hostname, $expressions );
```
aaemnnosttv, line 706: Would be nice to add another assertion here that it also contains `"www.$hostname"`

aaemnnosttv (changes requested):
Thanks @Hazlitte – this looks good, just needs a little refactoring and then this should be good to go.

includes/Modules/AdSense.php
Hazlitte: 👍

includes/Modules/Analytics.php
Hazlitte: 👍

includes/Core/Modules/Module.php
```diff
@@ -662,6 +662,31 @@ final protected function permute_site_url( $site_url ) {
...
 
+	/**
+	 * Permutes hostname to cover all different variants of it (not considering the path).
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $site_url Site URL to get hostname permutations for.
+	 * @return array List of permutations.
+	 */
+	final protected function permute_hostname( $site_url ) {
```
aaemnnosttv, line 673: Sorry if my previous comment was unclear, but we don't need to add a new method for permuting hostnames, we can simply use the `permute_site_url` method we have now. Like I said, we just need to map the result into the hostname from the array of URLs.

It also doesn't matter which element is which, so we don't need to key them differently to identify www vs non-www. We just need to map them into the array of values, (formatted slightly different for AdSense) and provide them to the respective filters.

Happy to show you what I mean if you have any questions.

aaemnnosttv (commented):
Almost there @Hazlitte – see my last comment and let me know if you have any questions.

includes/Core/Modules/Module.php
Hazlitte: I'm sorry, I don't understand what the first part of your comment means.

I understand the second part.

includes/Modules/AdSense.php
```diff
@@ -537,9 +537,17 @@ protected function create_adsense_earning_data_request( array $args = array() )
...
+		$hostnames = array_unique(
+			array_map(
+				function ( $site_url ) {
+					return 'DOMAIN_NAME==' . wp_parse_url( $site_url, PHP_URL_HOST );
+				},
+				$this->permute_site_url( $this->context->get_reference_site_url() )
+			)
+		);
+
+		if ( ! in_array( '', $hostnames, true ) ) {
```
aaemnnosttv, line 549: This condition is unnecessary because it would never contain an empty string.
https://github.com/google/site-kit-wp/blob/1680f18750ea7fc778d39a5a5feab64d0d1b24d0/includes/Core/Modules/Module.php#L645-L652

aaemnnosttv (commented):
Looks great, just one tiny thing left to tweak. ~~I'll push this real quick since we're looking to merge it for the release.~~

Hazlitte:
@googlebot

includes/Modules/Analytics.php
```diff
@@ -939,11 +939,21 @@ protected function create_analytics_site_data_request( array $args = array() ) {
...
+		$hostnames = array_values(
+			array_unique(
+				array_map(
+					function ( $site_url ) {
+						return wp_parse_url( $site_url, PHP_URL_HOST );
+					},
+					$this->permute_site_url( $this->context->get_reference_site_url() )
+				)
+			)
+		);
```
aaemnnosttv, line 951: One last thing here that I missed before – the added values + unique calls aren't necessary since the values will always be unique (one will always be the current hostname, and the other a www-added variant).
```suggestion
		$hostnames = array_map(
			function ( $site_url ) {
				return wp_parse_url( $site_url, PHP_URL_HOST );
			},
			$this->permute_site_url( $this->context->get_reference_site_url() )
		);
```

includes/Modules/AdSense.php
```diff
@@ -537,10 +537,16 @@ protected function create_adsense_earning_data_request( array $args = array() )
...
-			$opt_params['filter'] = 'DOMAIN_NAME==' . $host;
-		}
+		$hostnames = array_unique(
+			array_map(
+				function ( $site_url ) {
+					return 'DOMAIN_NAME==' . wp_parse_url( $site_url, PHP_URL_HOST );
+				},
+				$this->permute_site_url( $this->context->get_reference_site_url() )
+			)
+		);
```
aaemnnosttv, line 547: Just as previous comment – the permuted site URLs will always be unique so no need for this extra call 👍 
```suggestion
		$hostnames = array_map(
			function ( $site_url ) {
				return 'DOMAIN_NAME==' . wp_parse_url( $site_url, PHP_URL_HOST );
			},
			$this->permute_site_url( $this->context->get_reference_site_url() )
		);
```

aaemnnosttv:
@googlebot I consent.

includes/Modules/AdSense.php
aaemnnosttv: Oops! Sorry about that, you were right 👍  🤦

aaemnnosttv (approved):
Thanks @Hazlitte !

includes/Modules/AdSense.php
Hazlitte: 🙂👍

Issue #2981, opened 2021-03-17
Implement modules/analytics-4 datastore functionality for matching web datastream with property


felixarntz:
The `modules/analytics-4` datastore should be able to find a matching web data stream for a given property and the current site, as well as selecting it based on the property selected.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `modules/analytics-4` datastore should have a new selector `getMatchingWebDataStream( propertyID )` which, for the given property, checks whether there is a web data stream for that property that matches the current site’s URL in its `defaultUri` field (similar comparison to how we’re typically doing, i.e. ignore HTTP/HTTPS, www/non-www) and if so returns that web data stream object, or otherwise `null`.
* The `modules/analytics-4` datastore should have a new action `selectProperty( propertyID )` that sets the GA4 `propertyID` setting based on the given property and then handles side effect selections (similar to the `selectProperty` action in `modules/analytics`, however with slightly different behavior, focusing on the nested web data stream instead of profile/view):
    * Start by emptying any `webDataStreamID` and `measurementID` if set from before.
    * If set to create a new property, do nothing else.
    * Otherwise, check if there is a matching web data stream via `getMatchingWebDataStream` and if so set `webDataStreamID` and `measurementID` accordingly (the latter via the web data stream’s `measurementId` field).
    * Note that GA4 property objects have a [`name` attribute in the format `properties/{propertyID}`](https://developers.google.com/analytics/devguides/config/admin/v1/rest/v1alpha/properties#Property).  
The actual `propertyID` needs to be extracted from that string.
    * Note that GA4 web data stream objects have a [`name` attribute in the format `properties/{propertyID}/webDataStreams/{webDataStreamID}`](https://developers.google.com/analytics/devguides/config/admin/v1/rest/v1alpha/properties.webDataStreams#WebDataStream).  
The actual `webDataStreamID` needs to be extracted from that string.

#### Step 1

* Add `export const PROPERTY_CREATE = 'property_create';` to the `assets/js/modules/analytics-4/datastore/constants.js` file.
* Create a utility function `isValidPropertyID( propertyID )` in the `assets/js/modules/analytics-4/util/validation.js` file. It should return `TRUE` if the `propertyID` is a string and consists of digits only.
[...]

PR #3108, opened 2021-04-07
Enhancement/2981 ga4 matching datastream with property

eugene-manuilov:
## Summary

Addresses issue #2981

## Checklist


assets/js/modules/analytics-4/datastore/webdatastreams.js
```diff
@@ -134,6 +173,37 @@ const baseSelectors = {
...
+	 * @param {string} propertyID The GA4 property ID to find matched web data stream.
+	 * @return {(Object|null|undefined)} A web data stream object if found, otherwise null; `undefined` if web data streams are not loaded.
+	 */
+	getMatchingWebDataStream: createRegistrySelector( ( select ) => ( state, propertyID ) => {
+		const datastreams = select( STORE_NAME ).getWebDataStreams( propertyID );
+		if ( datastreams === undefined ) {
+			return undefined;
+		}
+
+		if ( Array.isArray( datastreams ) ) {
```
aaemnnosttv, line 192: `getWebDataStreams` only returns `Array|undefined` so this would be an unnecessary check. I would rather we enforce the type when receiving the data into state than not trusting our own APIs.

assets/js/modules/analytics-4/datastore/webdatastreams.test.js
```diff
@@ -159,5 +159,44 @@ describe( 'modules/analytics-4 webdatastreams', () => {
...
+				const datastream = registry.select( STORE_NAME ).getMatchingWebDataStream( propertyID );
+				expect( datastream ).toBeNull();
+			} );
+
+			it( 'should return the correct datastream when reference site URL matches exactly', () => {
+				provideSiteInfo( registry, { referenceSiteURL: 'http://example.com' } );
+				registry.dispatch( STORE_NAME ).receiveGetWebDataStreams( fixtures.webDataStreams, { propertyID } );
+
+				const datastream = registry.select( STORE_NAME ).getMatchingWebDataStream( propertyID );
+				expect( datastream ).toEqual( fixtures.webDataStreams[ 0 ] );
```
aaemnnosttv, line 186: I would prefer not to reference individual items from fixtures like this (same for the `it.each` below). These fixtures are mostly for use in tests which need a fetch mock with a decent sized object. It would be a bit more useful for the test to be able to see the definition of the webdatastream object - particularly since we're not using a factory to create it and the specific object you're referencing is significant. This was why builders were introduced for GTM (which was done after Analytics – it would be great to eventually do the same there though); not sure if it makes sense for GA4 or not; what do you think?

The simplest thing to do here would be to add definitions for these in the describe above (or at the top) for each of the two webdatastream objects and then receive them more explicitly in the test:

E.g.
```js
// above
const webDataStreamDotCom = { ... };
const webDataStreamDotOrg = { ... };
const webDataStreams = [ webDataStreamDotCom, webDataStreamDotOrg ];
// in test:
registry.dispatch( STORE_NAME ).receiveGetWebDataStreams( webDataStreams, { propertyID } );
// ...
expect( datastream ).toEqual( webDataStreamDotOrg );
```

This would be easier to reason about.

aaemnnosttv (changes requested):
Thanks @eugene-manuilov – this looks great, just a few things to address 👍

assets/js/modules/analytics-4/datastore/webdatastreams.test.js
eugene-manuilov: Updated.

assets/js/modules/analytics-4/datastore/webdatastreams.js
eugene-manuilov: Updated.

assets/js/modules/analytics-4/datastore/webdatastreams.js
```diff
@@ -97,9 +102,43 @@ const baseActions = {
...
+
+			return new Promise( ( resolve ) => {
+				const unsubscribe = subscribe( () => {
+					if ( areLoaded() ) {
+						unsubscribe();
+						resolve();
+					}
+				} );
+			} );
+		};
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/analytics-4/datastore/webdatastreams.js
```diff
@@ -97,9 +102,43 @@ const baseActions = {
...
+	/**
+	 * Waits for web data streams to be loaded for a property.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {string} propertyID GA4 property ID.
+	 * @return {Object} Redux action.
+	 */
+	waitForWebDataStreams( propertyID ) {
+		return {
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/analytics-4/datastore/properties.js
```diff
@@ -99,6 +100,42 @@ const baseActions = {
...
+				propertyID,
+				webDataStreamID: '',
+				measurementID: '',
+			} );
+
+			if ( PROPERTY_CREATE === propertyID ) {
+				return;
+			}
+
+			yield registry.dispatch( STORE_NAME ).waitForWebDataStreams( propertyID );
```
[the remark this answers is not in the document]
eugene-manuilov: Not sure why but these approaches don't work :man_shrugging:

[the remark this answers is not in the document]
aaemnnosttv: I'm not exactly sure what the problem was, but I got it to work. Pushing an update now.

aaemnnosttv:
@googlebot I consent.

assets/js/modules/analytics-4/datastore/properties.js
```diff
@@ -99,6 +101,38 @@ const baseActions = {
...
+	 */
+	selectProperty( propertyID ) {
+		invariant( isValidPropertySelection( propertyID ), 'A valid propertyID selection is required.' );
+
+		return ( function* () {
+			const registry = yield Data.commonActions.getRegistry();
+
+			registry.dispatch( STORE_NAME ).setPropertyID( propertyID );
+			registry.dispatch( STORE_NAME ).setWebDataStreamID( '' );
+			registry.dispatch( STORE_NAME ).setMeasurementID( '' );
```
aaemnnosttv, line 121: It's important that we always set these individually rather than using `setSettings` since this updates the object with _all settings_, it doesn't merge them. Using `setSettings` here would unset others like `useSnippet` or any other new setting in the future, so it's important that we always update them individually whenever we're not working with the full settings object.

assets/js/modules/analytics-4/datastore/properties.js
```diff
@@ -99,6 +101,38 @@ const baseActions = {
...
+
+			if ( PROPERTY_CREATE === propertyID ) {
+				return;
+			}
+
+			yield webDataStreamActions.waitForWebDataStreams( propertyID );
+
+			const webdatastream = registry.select( STORE_NAME ).getMatchingWebDataStream( propertyID );
+			if ( webdatastream ) {
+				registry.dispatch( STORE_NAME ).setWebDataStreamID( webdatastream.name.split( '/' ).pop() );
```
aaemnnosttv, line 131: We should probably extract a utility for this in the future, but this is okay for now.

aaemnnosttv (approved):
Looks good now, thanks @eugene-manuilov !

I pushed a few small changes and added a test for the behavior around `selectProperty` and waiting for web data streams so we can be sure that it is indeed working as expected now.

assets/js/modules/analytics-4/datastore/properties.js
eugene-manuilov: @aaemnnosttv it will be done in another ticket.

PR #3111, opened 2021-04-07
Fix cursor jump in user input by also storing state locally.

tofumatt:
## Summary

Addresses issue #2900, specifically the issue mentioned in QA feedback here: https://github.com/google/site-kit-wp/issues/2900#issuecomment-814843972

## Relevant technical choices

This addresses an issue with controlled components and—I think—React's scheduler and handling of controlled components like `<input />`. It's something I've seen before but it's easy to get bit by.

@johnPhillips provided some helpful context in a Slack convo so I'm copy-pasting what he wrote 🙂 

> So we’ve effectively lifted the state out of the component, and I think during the round trip React loses its grip on where the cursor should be. It seems that this is [something people run into a lot](https://github.com/facebook/react/issues/14904), there is even a [sandbox here which reproduces something extremely similar](https://codesandbox.io/s/distracted-rhodes-gkric).

I am vastly over-simplifying here, but my understanding is that, essentially, our `<Input />` component "loses track" of cursor position because of the "complex" route from the data store to the component's `value` prop. An easy fix is to copy what is in the data store locally in the component—all validation and management of the data remains the same, but the component has a local copy of the value data and thus doesn't get "confused" about cursor positioning.

## Checklist

### DIFF
--- a/assets/js/components/user-input/UserInputKeywords.js
+++ b/assets/js/components/user-input/UserInputKeywords.js
@@ -53,6 +53,10 @@ export default function UserInputKeywords( { slug, max } ) {
 		values.push( '' );
 	}
 
+	// Store values in local state to prevent
+	// https://github.com/google/site-kit-wp/issues/2900#issuecomment-814843972.
+	const [ localValues, setLocalValues ] = useState( [ '' ] );
+
 	// Need to make sure that dependencies list always has the same number of elements.
 	const dependencies = values.concat( Array( max ) ).slice( 0, max );
 
@@ -92,6 +96,7 @@ export default function UserInputKeywords( { slug, max } ) {
 			newKeywords = newKeywords.split( EOT ).slice( 0, max );
 		}
 
+		setLocalValues( newKeywords );
 		setUserInputSetting( slug, newKeywords );
 	}, [ slug ] );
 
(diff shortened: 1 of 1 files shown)


assets/js/components/user-input/UserInputKeywords.js
```diff
@@ -53,6 +53,10 @@ export default function UserInputKeywords( { slug, max } ) {
 		values.push( '' );
 	}
 
+	// Store values in local state to prevent
+	// https://github.com/google/site-kit-wp/issues/2900#issuecomment-814843972.
+	const [ localValues, setLocalValues ] = useState( [ '' ] );
```
aaemnnosttv, line 58: Shouldn't the initial state be `values`?

aaemnnosttv (changes requested):
@tofumatt this works as expected for the most part but introduces another bug 😄 

https://user-images.githubusercontent.com/1621608/113987653-53c61800-9857-11eb-8154-3c68ded3b525.mp4

assets/js/components/user-input/UserInputKeywords.js
tofumatt: Oops, yeah, that'd be fine 👍🏻

tofumatt:
@aaemnnosttv Argh, shoot! What are you typing to make that bug happen? I can't see the keyboard input so it isn't clear.

aaemnnosttv:
To reiterate from our other conversation, this happened in [Storybook for this PR](https://google.github.io/site-kit-wp/storybook/pull/3111/?path=/story/user-input--userinputapp) after filling in all the information, clicking _Preview_ and then clicking on _Edit_ for the keywords.

assets/js/components/user-input/UserInputKeywords.js
tofumatt: And as I suspected, that fixes the bug too. Thanks! 👍🏻

aaemnnosttv (approved):
Great, thanks!

Issue #2877, opened 2021-02-25
Add tracking events for PageSpeed widget


felixarntz:
There are a few new tracking events that should be implemented for the PageSpeed widget.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* GA tracking events for interacting with the PageSpeed Insights widget should be added:
    * `widget_view`: should be triggered **once** during page lifecycle when the widget becomes visible in the viewport, e.g. when the user scrolled to it
    * `default_tab_view`: should be triggered without user interaction, based on which tab is rendered by default (event label should be either "field" or "lab")
    * `tab_select`: when clicking the "Field" or "Lab" tab (event label should be either "field" or "lab")
    * `stack_pack_expand`: when opening a stack pack recommendation in the accordion (event label should be the audit ID)
* The event category should be `pagespeed_widget` for all of them

## Implementation Brief
* Install the `react-intersection-observer` package as a dependency which will allow us to know when an element is visible on the page.
* Using `assets/js/modules/pagespeed-insights/components/dashboard/DashboardPageSpeed.js`,
	* Use the `useEffect` and `useInView` hooks as described here [https://github.com/thebuilder/react-intersection-observer/blob/master/docs/Recipes.md#track-impressions] to track when the widget is visible within the viewport. If the widget is visible, i.e `inView` is `true`, call the `trackEvent` function with the following parameter:
		* action: `widget_view`
	* Use the same `useEffect` hook as in the previous bullet point to also track which tab is rendered by default. The `trackEvent` function must be called with the following parameters:
		* action: `default_tab_view`
		* label: `field` or `lab` depending on the selected tab. The latter can be obtained via the `dataSrc` variable which needs to be stripped of the `data_` prefix because the values can be `data_lab` or `data_field`.
	* Update the `updateActiveTab` function to call the `trackEvent` function with the following parameters:
		* action: `tab_select`
		* label: `lab` if the "Lab" tab has been selected or `field` if the "Field" tab has been selected.
* Using `assets/js/components/Accordion.js`,
	* add 2 new props:
		* `onOpen`: callback when the accordion item is opened.
		* `onClose`: callback when the accordion item is closed. Though it will not be used in this ticket, it can be helpful for other use cases later down the road.
	* Update the `toggleAccordion` function to call the `onOpen` and `onClose` props based on the `isActive` variable.
* Using `assets/js/modules/pagespeed-insights/components/common/Recommendation.js`,
[...]

PR #3116, opened 2021-04-08
Add tracking for PageSpeed Widget.

asvinb:
## Summary

Addresses issue #2877 

## Relevant technical choices
* The threshold for the widget to be in view is set to 0.25.

## Checklist


assets/js/components/Accordion.js
```diff
@@ -25,10 +25,19 @@ import classnames from 'classnames';
...
+export default function Accordion( { title, children, initialOpen, onOpen, onClose } ) {
 	const [ isActive, setActive ] = useState( !! initialOpen );
+
+	useEffect( () => {
+		if ( isActive && onOpen && typeof onOpen === 'function' ) {
+			onOpen();
+		} else if ( ! isActive && onClose && typeof onClose === 'function' ) {
+			onClose();
+		}
+	}, [ isActive ] );
```
eugene-manuilov, line 39: `onOpen` and `onClose` are properties, thus we need to add it to the dependencies list for this effect hook.

assets/js/modules/pagespeed-insights/components/common/Recommendation.js
```diff
@@ -38,6 +39,10 @@ export default function Recommendation( props ) {
 		strategy,
 	} = props;
 
+	const onOpen = useCallback( () => {
+		trackEvent( 'pagespeed_widget', 'stack_pack_expand', auditID );
```
eugene-manuilov, line 43: @felixarntz maybe we should limit this event tracking to only one per stack pack per page view? In other words, if a user opens and closes the `Remove unused CSS` accordion item three times during a single page view, we will track the appropriate event only once instead of three. What do you think?

eugene-manuilov (changes requested):
@asvinb, this PR looks solid to me. Just found one minor issue in the dependencies list and have one question for Felix.

assets/js/modules/pagespeed-insights/components/common/Recommendation.js
felixarntz: I think we should keep this as is for now and track every user interaction.

assets/js/components/Accordion.js
asvinb: Updated @eugene-manuilov

eugene-manuilov (approved):

Issue #2992, opened 2021-03-19
Mis-alignment of Goals title and data in Site Kit Dashboard 


wpdarren:
## Bug Description

While testing #2858  I noticed that the `Goals Completed` title and data does not align with the `Unique Visitors from Search ` due to the missing change arrow because a change of zero currently results in nothing output there.

![image](https://user-images.githubusercontent.com/73545194/111764032-bcb41300-88dd-11eb-846e-24cc72853df4.png)

Would be good that it is aligned in the widget. 

## Steps to reproduce

Find a test site with `Google Analytics` connected. 
The GA account has to have goals set up, but with zero completions. 
Go to Site Kit Dashboard and look under `Search Funnel`

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The Goals widget content should be vertically aligned with the other widgets in the Search Funnel, even if there is no change percentage displayed below the sparkline.
* This should be fixed in a holistic way, so that other widgets with that appearance would get that fix as well.

## Implementation Brief

* Refactor `assets/js/components/DataBlock.js`:
  * where the percentage change is formatted, change the control flow to format 0 if there is no change (i.e. `changeFormatted` should be `0%` in this scenario) instead of ignoring it.
  *  Output this `0%` inside the `span.googlesitekit-data-block__value` when there is no change, instead of outputting nothing.
  * Add a CSS class, `googlesitekit-data-block__change--no-change` to `.googlesitekit-data-block__change` when the change is 0.
* In `assets/sass/components/global/_googlesitekit-data-block.scss`, add a `visibility: hidden` style rule for `googlesitekit-data-block__change--no-change`. 

### Test Coverage

* No new tests; existing tests should pass.

### Visual Regression Changes

* The goals widget should have the same visual spacing as the other widgets when there is no change in Goals, even though the percentage change is not visible.

## QA Brief

* Follow 'steps to reproduce' above
* The goals widget should have the same visual spacing and alignment as the other Search Funnel widgets, even though it is not displaying the percentage change.

## Changelog entry

* Fix broken alignment of data block elements when there is one with zero data.

tofumatt:
IB ✅

cole10up:
### **QA ✅**

Setup goals on my personal site, confirmed the alignment below after the merge in develop:
![image](https://user-images.githubusercontent.com/57115907/114996616-dc464900-9e53-11eb-940f-fcb0998de1e8.png)

Checked window sizes:
[...]

PR #3121, opened 2021-04-09
Fix misalignment of zero change in datablock with visibility:hidden. (2992)

johnPhillips:
## Summary

Addresses issue #2992

## Relevant technical choices

Previously the `googlesitekit-data-block__change` div in the `DataBlock` component was not rendered if there was zero change. This caused a misalignment with neighbouring blocks.
I've made the component render the div in all cases, so a zero percent change will now output `0%`, however, in this case the div is hidden with `visibility: hidden`. This hides the div without losing the space it occupies, fixing the misalignment.

## Checklist


eugene-manuilov (changes requested):
@johnPhillips, overall this PR looks good to me, but there are issues with VRT tests. Those visual changes are expected, so we will need to approve broken VR tests. Could you, please, check whether you can make the sparkline with zero changes to be on the same baseline with other charts?

google-site-kit_Dashboard_Search_Funnel_0_document_1_medium

johnPhillips:
Hey @eugene-manuilov, thanks for the review. 

It seems like the default behaviour of the graph is to plot the line halfway up the container. Presumably because that represents zero change and change can be negative or positive. So a negative change would be plotted below the line and a positive change above. At least, that's my interpretation of what I am seeing, I might be incorrect.

Screenshot 2021-04-12 at 11 54 05

I think if that's the case then I'd have to do something funky with negative margins or `position: absolute` to move the container down by 50% to make it appear that the line is at the base of the container. 

I'm happy to give that a whirl, but thought I'd double check first. What do you think?

eugene-manuilov:
> It seems like the default behaviour of the graph is to plot the line halfway up the container. Presumably because that represents zero change and change can be negative or positive. So a negative change would be plotted below the line and a positive change above.

We don't have negative values in the data arrays that we pass to the `Sparkline` component. That graph displays actual values, not changes from day-to-day.

> I think if that's the case then I'd have to do something funky with negative margins or position: absolute to move the container down by 50% to make it appear that the line is at the base of the container.

I think we can add the following to the Sparkline component to make it display horizontal axis at the bottom when there are all values equal to zero:

```diff
diff --git a/assets/js/components/Sparkline.js b/assets/js/components/Sparkline.js
index b1d2bde52..43f74182d 100644
--- a/assets/js/components/Sparkline.js
+++ b/assets/js/components/Sparkline.js
@@ -66,6 +66,11 @@ function Sparkline( {
                ],
        };
 
+       if ( ! data.slice( 1 ).some( ( row ) => row[ 1 ] > 0 ) ) {
+               chartOptions.vAxis.minValue = 0;
+               chartOptions.vAxis.maxValue = 1;
+       }
+
        return (
                <div className="googlesitekit-analytics-sparkline-chart-wrap">
                        <GoogleChartV2
```

assets/js/components/Sparkline.js
```diff
@@ -66,6 +66,11 @@ function Sparkline( {
 		],
 	};
 
+	if ( ! data.slice( 1 ).some( ( row ) => row[ 1 ] > 0 ) ) {
+		chartOptions.vAxis.minValue = 0;
+		chartOptions.vAxis.maxValue = 1;
+	}
```
eugene-manuilov, line 72: @johnPhillips could you please add a comment to explain why we need this?

johnPhillips: Of course, sorry 👍

eugene-manuilov (approved):
Awesome! Thanks, @johnPhillips.

Issue #2941, opened 2021-03-10
Migrate `Button` Storybook coverage to follow new v6 best practices


felixarntz:
With Storybook now in v6, we should migrate a few particular stories to follow the new best practices established in [this doc](https://docs.google.com/document/d/1Ny2Tlo1_FxZY6s6dC2-E1TWJYi1rhp0OMEtEfXxwXjY/edit). Migrating all existing stories is not a high priority, we can keep most old stories in their existing state; they should only be migrated on demand.

This issue is about migrating a specific presentational (non-connected) component to follow the new Storybook v6 best practices.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The general `Button` component should receive an accompanying `Button.stories.js` file, with new Storybook coverage that follows the [new Storybook best practices](https://docs.google.com/document/d/1Ny2Tlo1_FxZY6s6dC2-E1TWJYi1rhp0OMEtEfXxwXjY/edit). Particularly:
    * It should use `Components/Button` for its title.
    * It should have individual stories for each of its 5 different states (see existing stories), relying on [story args](https://storybook.js.org/docs/react/writing-stories/args#story-args). Similar to the examples in the documentation, a single `Template` should be used with arguments different for each of the 5 scenarios.
* The existing Storybook coverage for `Button` from before should be removed.

## Implementation Brief

* Familiarise yourself with the `Site Kit: Storybook v6 usage` doc.
* Make a copy of `stories/buttons.stories.js` to `assets/js/components/Button.stories.js`
* Break up the combined stories into a set of individual stories for each button variant.
* For the backstop params in https://github.com/google/site-kit-wp/blob/ef1a7e5/stories/buttons.stories.js#L68-L72 use `component parameters` for them: https://storybook.js.org/docs/react/writing-stories/parameters#component-parameters
* Rebuild the stories via `npm run build:storybook` or by keeping `npm run storybook` running as you make story changes. 
* Compare visually each individual story in storybook v6 format to the old format of story (v5), you can do this by comparing your updated stories with https://google.github.io/site-kit-wp/storybook/master/ or https://google.github.io/site-kit-wp/storybook/develop/ and comparing with `/dist/index.html`
* When the stories match visually, remove the old format of story `stories/buttons.stories.js`
* Run backstop to ensure visual regression testing work correctly `npm run test:visualtest`
[...]

PR #3140, opened 2021-04-13
Migrate `Button` Storybook coverage to follow new v6 best practices.

Hazlitte:
## Summary

Migrate `Button` Storybook coverage to follow new v6 best practices.

Addresses issue #2941 

## Checklist


Hazlitte:
@googlebot

tofumatt:
@googlebot I consent.

.storybook/storybook-data.js
```diff
@@ -14,12 +14,12 @@ module.exports = [
...
-		id: 'global--buttons',
-		kind: 'Global',
+		id: 'components-button--hover-button',
+		kind: 'Components',
 		name: 'Buttons',
-		story: 'Buttons',
+		story: 'Default Button Hover',
 		parameters: {
-			fileName: './stories/buttons.stories.js',
+			fileName: './assets/js/components/Button.stories.js',
```
aaemnnosttv, line 22: I'm exactly sure what should happen with the scenario definition here, but we can't simply replace it with one of the single stories since the previous story used for VRT had multiple cases in the one story and this is just one.

If anything, maybe this should just be removed here and re-added later when we update the new stories to support defining scenarios.

assets/js/components/Button.stories.js
```diff
@@ -0,0 +1,65 @@
...
+const Template = ( args ) => <Button { ...args } />;
+
+export const DefaultButton = () => <Button>Default Button</Button>;
+DefaultButton.storyName = 'Default Button';
+
+export const LinkButton = Template.bind( {} );
+LinkButton.storyName = 'Default Button Link';
+LinkButton.args = {
+	children: 'Default Button Link',
+	href: 'http://google.com',
```
aaemnnosttv, line 33: Let's use a `#` here when a `href` is needed but the value isn't significant.
```suggestion
	href: '#',
```

assets/js/components/Button.stories.js
```diff
@@ -0,0 +1,65 @@
...
+DisabledButton.storyName = 'Disabled Button';
+DisabledButton.args = {
+	children: 'Disabled Button',
+	disabled: true,
+};
+
+export const HoverButton = Template.bind( {} );
+HoverButton.storyName = 'Default Button Hover';
+HoverButton.args = {
+	children: 'VRT: Default Button Hover',
```
aaemnnosttv, line 53: I realize this is how it was before, but it seems oddly inconsistent for this one to have such a prefix and it's already clear that this is for VRT due to the associated scenario definition.
```suggestion
	children: 'Default Button Hover',
```

aaemnnosttv (changes requested):
Thanks @Hazlitte – a few minor changes to make but should be quick. The bigger question is what we should do with VRT as this wasn't mentioned in the ACs explicitly.

@felixarntz what would you like us to do with the backstop scenario definition for this? My understanding is that we aren't ready to start implementing the new backstop scenario definitions just yet (there's some other infra needed for that too).

felixarntz:
@aaemnnosttv Regarding the Backstop implementation, is it possible to keep the old approach (everything in `storybook-data.js`) and update it to point to the new story? Yes, we can't do the new Backstop implementation for this yet, which is fine, but if possible we should keep the Storybook coverage available.

Btw do we already have an issue to establish the new Backstop convention? We should probably ensure then that it is applied to our three "demo issues" for Storybook v6 approach.

aaemnnosttv:
> is it possible to keep the old approach (everything in `storybook-data.js`) and update it to point to the new story?

@felixarntz Yes of course, except the new story is 5 stories 😄  So we no longer have a single story using the new approach we can use for the same VRT scenario since the old story contained all 5 cases.

> if possible we should keep the Storybook coverage available

I think you mean Backstop/VRT coverage? The ACs state "The existing Storybook coverage for `Button` from before should be removed." which I think makes sense, but we'd need to add an additional "aggregate" story like we had before in order to preserve the single VRT test. I think this is probably okay to do as a temporary step until the new approach is ready.

> Btw do we already have an issue to establish the new Backstop convention? We should probably ensure then that it is applied to our three "demo issues" for Storybook v6 approach.

I don't think we do yet, I'll create it.

felixarntz:
@aaemnnosttv 
> I think you mean Backstop/VRT coverage? The ACs state "The existing Storybook coverage for `Button` from before should be removed." which I think makes sense, but we'd need to add an additional "aggregate" story like we had before in order to preserve the single VRT test. I think this is probably okay to do as a temporary step until the new approach is ready.

Ah yes, I meant "Backstop/VRT". Good point on the overload of stories, yeah we surely wouldn't want 5 VRT scenarios for just the buttons. How straightforward would it be to have one aggregate story? It's probably worth doing that, since this need will be relevant in more cases than this.

aaemnnosttv:
> How straightforward would it be to have one aggregate story?

@felixarntz In this case, it would be pretty straightforward since these are simple UI components. We have to add a story for backstop to use regardless so we could have a story like the old one that renders each variant. I'm not sure if we could simply render each story component in a larger component though since the props are passed via args 🤔 

Looking in the docs, it seems like this is possible, it just requires passing through the args as props manually:
https://storybook.js.org/docs/react/workflows/stories-for-multiple-components#reusing-subcomponent-stories

For sub-stories that have their own datastore set up, this could be problematic though as different sub-stories may compete for the same state in the shared registry so we'll need to see how to best handle this when it comes up.

assets/js/components/Button.stories.js
```diff
@@ -0,0 +1,65 @@
...
+	className: 'googlesitekit-button--hover',
+	options: {
```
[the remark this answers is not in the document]
aaemnnosttv: Let's be sure to move these to the new `VRTStory` as well since it will be the only one with a respective scenario.

assets/js/components/Button.stories.js
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Internal dependencies
+ */
+import Button from './Button';
+
+const Template = ( args ) => <Button { ...args } />;
+
+export const DefaultButton = () => <Button>Default Button</Button>;
+DefaultButton.storyName = 'Default Button';
```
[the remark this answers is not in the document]
Hazlitte: 👍

assets/js/components/Button.stories.js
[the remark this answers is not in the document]
Hazlitte: 👍

.storybook/storybook-data.js
Hazlitte: 👍

assets/js/components/Button.stories.js
Hazlitte: 👍

assets/js/components/Button.stories.js
Hazlitte: 👍

Hazlitte:
@aaemnnosttv I have followed all the comments above however I haven't as yet succeeded in getting the VRT tests to pass.  I have committed the changes in case anyone else wishes to take a look and will pick this up again when I am back online on Monday.

aaemnnosttv (approved):
Thanks @Hazlitte – I pushed a few finishing touches but this should be good to go now.

Changes I made:
- Updating `fileName` referenced in storybook data
- Removing added reference images that no longer matched a scenario (only the existing images changed)

Also, don't forget to keep the branch up to date with the base when iterating on changes 😉

aaemnnosttv:
@googlebot I consent.

Issue #3058, opened 2021-03-31
Help icon needs circular mouse hover 


Pratheep-lab:
The help icon that is yet to be launched has a rectangular mouse hover. This needs to be changed to a circular mouse hover. 

**Current:** 
![image](https://user-images.githubusercontent.com/81642523/113162656-6331d980-9237-11eb-9612-dc9d240d6a74.png)

**The shape of the grey mouse hover has to look like this:**
![image](https://user-images.githubusercontent.com/81642523/113162870-8e1c2d80-9237-11eb-857f-06c02d2a5925.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The background for the global help icon (behind `helpVisibility` feature flag) that is shown e.g. when hovering, clicking etc. should be circular instead of rectangular (see screenshots above).

## Implementation Brief
* Using `assets/sass/components/global/_googlesitekit-help-menu-link.scss`,
	* Add new styles for the `.googlesitekit-help-menu` class name, targeting the following class names:
		* `.mdc-button--dropdown`
			* Set the `border-radius` to `50%`
			* Set the `min-width` to `36px`
			* Set the `padding` to `0`
* Using `assets/sass/components/global/_googlesitekit-header.scss`,
	* Add the `.googlesitekit-dropdown-menu + .googlesitekit-dropdown-menu` selector and set the `margin-left` to `24px`. Instead of adding `margin-right` to the `.googlesitekit-dropdown-menu` element, we are adding `margin-left` to the adjacent element so that the dropdown menu for the help icon is aligned with the button.
	* Update the `.googlesitekit-date-range-selector` class to remove the `margin-right`.
* The help icon looks slightly smaller than the calendar icon next to it because the dimension of the path within the help icon is `16px` whereas it should  have been `20px`. To fix that, update `assets/svg/help.svg` with the following SVG:
	```js
	<svg width="20" height="20" viewBox="0 0 20 20" fill="none" xmlns="http://www.w3.org/2000/svg">
[...]

PR #3143, opened 2021-04-14
Enhancement/3058 help icon needs circular mouse hover

danielgent:
## Summary

Addresses issue #3058

## Checklist

**Before**
header-before

**After**
header-after


danielgent:
Do check out the visual diff change! In storybook the flag `helpVisibility` is not set

Screenshot from 2021-04-14 15-23-01

tofumatt:
@googlebot I consent.

tofumatt (changes requested):
First off: I added a new story to this PR because we didn't have a header like one on most dashboard views—with both the Help Menu and the Date Range selector: 687315bf462c02f5a90ce0181509bd07b0cd59ac

You can view the story here: https://google.github.io/site-kit-wp/storybook/pull/3143/?path=/story/global--plugin-header-with-help-menu-and-date-range-selector

The spacing looks nice on larger viewports, but unfortunately on mobile this design gets very cramped. On a very large mobile it works _okay_, save for an odd hover effect, but note what happens when a smaller viewport is used:



This is what it looks like on an iPhone X:

Screenshot 2021-04-15 at 01 26 03

-----

This might be an issue with the IB itself—I'm not sure if mobile viewports were taken into account. But further design work might need to be done to allow for the proposed spacing in the header.

asvinb:
@googlebot I consent.

tofumatt (changes requested):
This still has issues with not only iPhone SE sizes (`320px` wide) but also iPhone X sizes:

#### iPhone SE

Screenshot 2021-04-15 at 21 00 52

#### iPhone X

Screenshot 2021-04-15 at 21 00 45

You can see this at https://google.github.io/site-kit-wp/storybook/pull/3143/iframe.html?id=global--plugin-header-with-help-menu-and-date-range-selector&viewMode=story using a mobile viewport.

So I think the spacing still needs adjusting.

asvinb:
@tofumatt In Storybook, there is 50px padding on each side, meaning that the content area is further reduced. If you're testing on iPhone 5, the content area in Storybook is not 320px but only 220px. Can you test within WordPress itself instead of Storybook? This is how it should look:
image

I think to avoid any confusion in the further, we need to revisit the "small mobile" viewport option in Storybook so that the content area is actually 320px, not including the extra padding from Storybook.

tofumatt:
@asvinb Oh, shoot, I didn't realise that! I've filed #3174 to address that issue, thanks!

I'll test in the WP dashboard instead. 👍🏻

tofumatt (approved):
Looks great now:

Screenshot 2021-04-16 at 15 02 33

Thanks 👍🏻

tofumatt (approved):
(Pushed the latest `develop` to fix the E2E Nightly test failure.)

Issue #2951, opened 2021-03-11
AdSense "Performance by page over the last X days" deep links do not include date range


aaemnnosttv:
## Bug Description

The links in the top pages table on the AdSense module dashboard which is shown when there is a linked Analytics account does not include the current date range in its deep link.

## Steps to reproduce

1. Set up Site Kit with AdSense and a linked Analytics account
2. Navigate to AdSense dashboard
3. Note the current date range
3. Click on a link in the "Performance by page over the last X days" table
4. See the date range in Analytics does not match the date range selected in Site Kit

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Deep links in the `LegacyAnalyticsAdSenseDashboardWidgetTopPagesTable` component should include date range args in the report args used for constructing the report service URL

## Implementation Brief
* Using `/assets/js/modules/analytics/components/dashboard/LegacyAnalyticsAdSenseDashboardWidgetTopPagesTable.js`,
	* Import the `generateDateRangeArgs` from `assets/js/modules/analytics/util/report-date-range-args.js`.
	* Query the user's date range selection by querying the user data store with the `getDateRangeDates` selector, passing the `DATE_RANGE_OFFSET` constant from `assets/js/modules/analytics/datastore/constants.js` as value to the `offsetDays` key.
	* Update usage of the `getServiceReportURL` selector to spread the result of the `generateDateRangeArgs` function, passing the user's date range as parameter to the function.
	* Changes above should also be applied to the new widget based component, most specifically the following file: `assets/js/modules/adsense/components/module/ModuleTopEarningPagesWidget/Table.js`

### Test Coverage
* N/A

### Visual Regression Changes
* N/A

## QA Brief
You will need to enable the `widgets.moduleScreens` feature using the tester plugin.
* Follow the 'steps to reproduce' above
* The links in the new (top) widget and the legacy (bottom) widget should all contain a date range 

## Changelog entry

* Update AdSense "Performance by page over the last X days" deep links to include date range

tofumatt:
The IB had adjustments made now that #2987 is merged (the original IB mentioned this, but I made them concrete). Essentially just that the component `AnalyticsAdSenseDashboardWidgetTopPagesTable` is now `LegacyAnalyticsAdSenseDashboardWidgetTopPagesTable` Otherwise looks good!

IB ✅

johnPhillips:
[...]

Issue #656, opened 2019-10-14
AdSense stats/reporting fixes on the main dashboard


OisinOConnor:
#### Top Earning Pages panel (First image below)
Revenue -> change "Revenue" to “Earnings” for consistency. Should be in currency format. Currently in plain number format.

#### Earning panel (Second image below)
RPM -> change "RPM" to “Page RPM” for consistency. There is no trend (red/green arrows & change) showing up for RPM. Should be in same currency format as RPM.
Total Earnings -> Trend number format does not match the currency format of the main number.
Ad Impressions -> Trend number is missing.

![Image2](https://user-images.githubusercontent.com/6488453/65443536-34f36280-de26-11e9-87c4-c6752729ed66.png)
![Image1](https://user-images.githubusercontent.com/6488453/65443535-34f36280-de26-11e9-86a4-0ac2c5a84707.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* In the AdSense summary widget on the Site Kit dashboard:
    * The "RPM" label should be changed to "Page RPM".
* In the AdSense "Top Earning Pages" widget on the Site Kit dashboard:
    * The "Revenue" column should have its label changed to "Earnings".
    * Each value in that column should be formatted with the currency.
* In the AdSense "Performance by page over the last 28 days" widget on the AdSense module page:
    * The "Earnings" and "Page RPM" columns should be formatted as a currency instead of plain numerics (similar to above in the "Top Earning Pages" widget on the Site Kit dashboard).
* These changes should be made to both the legacy version of the widgets as well as to the refactored versions.

## Implementation Brief
* Using `assets/js/modules/adsense/components/dashboard/DashboardSummaryWidget.js`, update the title of the first `DataBlock` from `RPM` to `Page RPM`.
* Using `assets/js/modules/adsense/components/dashboard/DashboardTopEarningPagesWidget.js` and `assets/js/modules/adsense/components/module/ModuleTopEarningPagesWidget/Table.js`, update the `title` and `tooltip` of the second column of `tableColumns` from `Revenue` to `Earnings`.
* Using `assets/js/modules/analytics/components/dashboard/LegacyAnalyticsAdSenseDashboardWidgetTopPagesTable.js`,
	* Query the AdSense data store to get the currency by using the `getReport` selector and passing `EARNINGS` as `metrics`. The currency value can be found in the headers array from the response.
[...]

Issue #576, opened 2019-09-23
AdSense dashboard: fix number formatting 


OisinOConnor:
## Fixing small things with the AdSense reporting screen

Summary: Some of the numbers are wrongly formatted and a few other tweaks. Details below.

>AdSense reporting screen (First image below):
>>Performance over previous 28 days highlights panel
Page RPM: should be in currency format. Currently plain number.

>>Performance over 28 days panel (page title list when linked with GA)
Earnings: should be in currency format. Currently plain number.
Page RPM: should be in currency format. Currently plain number.

![Image3](https://user-images.githubusercontent.com/6488453/65443537-34f36280-de26-11e9-8e13-da471eb19078.png)
---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The values in the "Earnings" and "Page RPM" columns of the "Performance by page over the last 28 days" widget on the AdSense module page should be formatted as a currency instead of plain numerics.
* This change should be made to both the legacy version of this widget as well as to the refactored version (see #2066).

## Implementation Brief

* 

## Changelog entry

*

felixarntz:
In parts this has already been addressed through the new version of the "Performance" widget, but it's still relevant for the "Performance by page" widget.

johnPhillips:
@felixarntz I picked up https://github.com/google/site-kit-wp/issues/656 and have been working through the IB. It covers adding currencies to the values in `LegacyAnalyticsAdSenseDashboardWidgetTopPagesTable`,  `DashboardTopEarningPagesWidget` and `ModuleTopEarningPagesWidget`.

Just wondering if this issue is effectively a duplicate?

felixarntz:
@johnPhillips From an ACs perspective, these two are not duplicates since this one only touches the AdSense module page while #656 only touches the overall Site Kit dashboard. As far as I can tell:
* The `LegacyAnalyticsAdSenseDashboardWidgetTopPagesTable` component is related to this issue, not #656. Instead, the `LegacyAdSenseDashboardWidgetTopPagesTableSmall` component is the one that relates to #656, since that component is rendered on the Site Kit dashboard.
* The `ModuleTopEarningPagesWidget` also is only related to this issue since it's for the AdSense module page, not the Site Kit dashboard.

So the work to format numbers across these components is almost the same, but they are technically part of the different issues based on where the components are used.

[...]

PR #3147, opened 2021-04-14
AdSense stats/reporting fixes on the main dashboard (656).

johnPhillips:
## Summary

Addresses issue #656 
Note also the discussion on #576 - we ended up pulling that work into this PR.

## Relevant technical choices
I pulled out the currency extraction from the AdSense report into a utility. It falls back to the original decimal format if the report hasn't yet loaded.

Also, regarding `LegacyAdSenseDashboardWidgetTopPagesTableSmall`:
- it needed to be refactored into a functional component to use the `useSelect` hook and `ReportTable` component.
- like `LegacyAnalyticsAdSenseDashboardWidgetTopPagesTable` it also didn't include deep links  (see https://github.com/google/site-kit-wp/issues/2951), so I added these.

## Checklist


assets/js/modules/adsense/components/module/ModuleTopEarningPagesWidget/Table.js
```diff
@@ -33,12 +33,30 @@ import ReportTable from '../../../../../components/ReportTable';
...
 import { CORE_USER } from '../../../../../googlesitekit/datastore/user/constants';
+import { getCurrencyFormat } from '../../../util/currency';
 import { generateDateRangeArgs } from '../../../../analytics/util/report-date-range-args';
 import { numFmt } from '../../../../../util';
 const { useSelect } = Data;
 
 export default function Table( { report } ) {
+	const {
+		currencyFormat,
+	} = useSelect( ( select ) => {
```
felixarntz, line 46: Nit-pick, but no need to wrap this in an object here since it's the only returned value.

assets/js/modules/adsense/components/module/ModuleTopEarningPagesWidget/Table.js
```diff
@@ -33,12 +33,30 @@ import ReportTable from '../../../../../components/ReportTable';
 import TableOverflowContainer from '../../../../../components/TableOverflowContainer';
 import Link from '../../../../../components/Link';
 import { MODULES_ANALYTICS, DATE_RANGE_OFFSET } from '../../../../analytics/datastore/constants';
+import { MODULES_ADSENSE } from '../../../datastore/constants';
```
felixarntz, line 36: Per our "best practices" we should use `STORE_NAME` for code (e.g. components) that is part of the same module as the relevant store. In other words, you're editing an AdSense component here, so when referring to the AdSense store, it should use `STORE_NAME`. Only when referring to other stores, the more expressive constant should be used (as you correctly do here for `MODULES_ANALYTICS`.

assets/js/modules/analytics/components/dashboard/LegacyAdSenseDashboardWidgetTopPagesTableSmall.js
```diff
@@ -20,95 +20,126 @@
...
+const { useSelect } = Data;
 
-class LegacyAdSenseDashboardWidgetTopPagesTableSmall extends Component {
-	static renderLayout( component ) {
-		return (
-			<div className="
+let currencyFormat;
+const setCurrencyFormat = ( report ) => {
+	currencyFormat = getCurrencyFormat( report );
+};
```
felixarntz, line 51: This seems really hacky. Why not handle this value relying on hooks in the component as usual?

assets/js/modules/analytics/components/dashboard/LegacyAdSenseDashboardWidgetTopPagesTableSmall.js
```diff
@@ -20,95 +20,126 @@
...
+		metrics: 'EARNINGS',
+	} ) );
 
-		const options = {
-			hideHeader: false,
-			chartsEnabled: false,
-			cap: 5,
-			links,
-		};
+	setCurrencyFormat( adsenseData );
```
felixarntz, line 84: See above, we should be able to get the currency format within this component.

assets/js/modules/analytics/components/dashboard/LegacyAdSenseDashboardWidgetTopPagesTableSmall.js
```diff
@@ -20,95 +20,126 @@
...
+	{
+		title: __( 'Earnings', 'google-site-kit' ),
+		description: __( 'Earnings', 'google-site-kit' ),
+		field: 'metrics.0.values.0',
+		Component: ( { fieldValue } ) => numFmt(
+			fieldValue,
+			currencyFormat,
+		),
+	},
+];
```
felixarntz, line 142: `tableColumns` needs to be defined within the component as we do elsewhere - this way the localization works as expected and also we don't have to use the `setCurrencyFormat` approach.

assets/js/modules/analytics/components/dashboard/LegacyAnalyticsAdSenseDashboardWidgetTopPagesTable.js
```diff
@@ -33,15 +33,34 @@ import ctaWrapper from '../../../../components/legacy-notifications/cta-wrapper'
...
 import Link from '../../../../components/Link';
 import ReportTable from '../../../../components/ReportTable';
+import { getCurrencyFormat } from '../../../adsense/util/currency';
 import { generateDateRangeArgs } from '../../util/report-date-range-args';
 const { useSelect } = Data;
 
+let currencyFormat;
+const setCurrencyFormat = ( report ) => {
+	currencyFormat = getCurrencyFormat( report );
+};
```
felixarntz, line 49: See above.

stories/module-adsense-components.stories.js
```diff
@@ -189,6 +189,32 @@ const topEarningPagesArgs = {
...
+		[ '2103.23', '1.89', '5693093' ],
+	],
+	totals: [
+		'2103.23', '1.89', '5693093' ],
+	averages: [
+		'2103.23', null, '5693093',
+	],
+	startDate: '2020-08-15',
+	endDate: '2020-09-11',
+};
```
felixarntz, line 210: Nitpick, but this is not a "correct" mock report for the data requested below:
* You're only requesting the `EARNINGS` metric below, as expected.
* This means the response must only include `EARNINGS` data. However the response here also includes data for two other metrics.

stories/module-adsense-components.stories.js
```diff
@@ -189,6 +189,32 @@ const topEarningPagesArgs = {
 	limit: 5,
 };
 
+const topEarningPagesAdSenseReport = {
```
felixarntz, line 192: Maybe add a comment above this and potentially even rename the variable to something that more clearly indicates that this is only used to get the currency to display, not any "actual" numbers.

felixarntz (changes requested):
This looks solid mostly and works as expected. A few points of feedback where we can clean some things up.

assets/js/modules/analytics/components/dashboard/LegacyAdSenseDashboardWidgetTopPagesTableSmall.js
johnPhillips: Ah, I hadn't grasped that the `Table` should be inside the component - I was just copying the approach in `LegacyAnalyticsAdSenseDashboardWidgetTopPagesTable`. Good to know about the localisation.
I'll refactor both - no problem 👍

stories/module-adsense-components.stories.js
johnPhillips: Good idea, I'll do both.

felixarntz (approved):
Excellent!

PR #3158, opened 2021-04-15
Fix/1941 validated actions

aaemnnosttv:
## Summary

Addresses issue #1941

## Relevant technical choices

**Add tests for createValidatedAction**
We don't have code coverage set up on this project so I just sensibly thought about what needed testing!  Happy to add any more edge cases anyone can think of

**Update remaining actions**
I went through the code looking for all use of `invariant(` inside a generator function `function*`

I went through twice, but any reviewer is more than welcome to look again!  Really weird bugs can be caused by redux sagas swallowing exceptions! (I speak from experience :smile: )

## Checklist


danielgent:
@googlebot I consent

assets/js/googlesitekit/datastore/location/navigation.test.js
```diff
@@ -52,8 +52,7 @@ describe( 'core/location', () => {
 	describe( 'actions', () => {
 		describe( 'navigateTo', () => {
 			it( 'should require a valid URL', () => {
-				return expect( registry.dispatch( STORE_NAME ).navigateTo( 'testurl' ) )
-					.rejects
+				return expect( () => registry.dispatch( STORE_NAME ).navigateTo( 'testurl' ) )
```
danielgent, line 55: Note how I've had to update this test though

If I don't wrap the call inside an anonymous function the Jest fails the test because the exception isn't caught by this syntax  (which makes sense as previous the exceptions were getting swallowed thus this ticket)

Instead, I can wrap it in a try-catch block but then that's ugly and prone to false positives!

That does mean that all my other changes are untested though! I had a look through the other tests and they're not testing that `invariant` is catching bad input (which does make sense as previously the code was very basic)

I'm happy to add more tests though

eugene-manuilov (changes requested):
Nice work. There are a few things to address before we can merge it.

assets/js/googlesitekit/data/utils.test.js
```diff
@@ -787,4 +793,68 @@ describe( 'data utils', () => {
...
+			expect( validator ).toHaveBeenCalledTimes( 1 );
+			expect( validator ).toHaveBeenCalledWith( args );
+		} );
+
+		it( 'should call action creator', () => {
+			const args = { foo: 'bar' };
+			const validator = noop;
+			const actionCreator = jest.fn();
+
+			expect( actionCreator ).toHaveBeenCalledTimes( 0 );
```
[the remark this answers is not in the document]
danielgent: Good point.  It's a force of habit after being burned a fair few times with odd tests

But I'm literally declaring the mock one line above. :joy:

assets/js/googlesitekit/data/utils.js
```diff
@@ -377,3 +379,24 @@ export function createValidationSelector( validate ) {
...
+ * @since n.e.x.t
+ *
+ * @param {Function} validate      A function for validating action arguments.
+ * @param {Function} actionCreator A function for returning or yielding redux-style actions.
+ * @return {Function} An enhanced action creator.
+ */
+export function createValidatedAction( validate, actionCreator ) {
+	invariant( typeof validate === 'function', 'a validator function is required.' );
+	invariant( typeof actionCreator === 'function', 'an action creator function is required.' );
+	invariant( ! isGenerator( validate ), 'an action’s validator function must not be a generator.' );
```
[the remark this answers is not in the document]
danielgent: @eugene-manuilov Good point on it being a private API! I didn't spot that. I had to install this package in order for Jest to use it so I can revert that change

This was already on @aaemnnosttv's PR (and the IB was only to add to the PR :upside_down_face: )

I'm making your requested change anyway @eugene-manuilov  but just tagged Evan so he can see :eyes: 

My only concern was that maybe in the future there's a better way to check for generator functions but I doubt it will change :smile:

assets/js/googlesitekit/data/utils.test.js
```diff
@@ -787,4 +793,68 @@ describe( 'data utils', () => {
 			} );
 		} );
 	} );
+
+	describe( 'createValidatedAction', () => {
```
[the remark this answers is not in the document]
danielgent: Done below in test `should not call action creator if validator throws an exception` @eugene-manuilov

assets/js/googlesitekit/data/utils.test.js
```diff
@@ -787,4 +793,81 @@ describe( 'data utils', () => {
...
+			const validator = noop;
+			function* actionCreator() { }
+
+			return expect( () => createValidatedAction(
+				validator,
+				// this has to be called to return generator object from generator function
+				actionCreator()
+			) )
+				.toThrow( 'an action creator function is required.' );
+		} );
```
eugene-manuilov, line 822: This test doesn't really make sense because it tests the logic that has already been tested in the previous test where we called the `createValidatedAction` function without the second parameter (in other words passed `undefined`). I suggest removing this test because it is redundant.

assets/js/googlesitekit/data/utils.test.js
```diff
@@ -787,4 +793,81 @@ describe( 'data utils', () => {
...
+				throw new Error( 'foo' );
+			};
+			const actionCreator = jest.fn();
+
+			const validatedAction = createValidatedAction(
+				validator,
+				actionCreator
+			);
+
+			expect( () =>	validatedAction( args ) ).toThrow( 'foo' );
```
eugene-manuilov, line 868:
```suggestion
			expect( () => validatedAction( args ) ).toThrow( 'foo' );
```

eugene-manuilov (changes requested):
Thanks, @danielgent! In general looks good to me. Just found a few more formatting issues and one suggestion to remove one of the tests. Please, take a look at it.

assets/js/googlesitekit/data/utils.test.js
danielgent: Apologies @eugene-manuilov ! I managed to confuse myself here with some previous changes :facepalm: 

I've fixed this test now to correctly check if the validator is a generator (which is the whole point of the `invariant` call in the previous file!)

eugene-manuilov (approved):

Issue #3105, opened 2021-04-07
Add general selector to match the site URL and use it to get the matching GA4 web data stream


felixarntz:
In order to know whether Site Kit needs to create a new web data stream for the GA4 property that the user has selected or whether it can use an existing one, the `modules/analytics-4` store should offer a selector.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `core/site` datastore should get a new selector `isSiteURLMatch( url )`.
    * It should rely on the `getReferenceSiteURL` selector and then create the permutations of it, in a similar way that the `Module::permute_site_url` method does in PHP (creating variants with HTTP/HTTPS and www/non-www).
    * It should then check whether the passed `url` is among those variants (ensuring the comparison is made in lower-case, for the edge-case that one of the URLs is capitalized).
    * If the passed `url` is a match, it should return `true`, otherwise `false`.
* The `modules/analytics-4` datastore's `getMatchingWebDataStream( propertyID )` selector should be updated to rely on the new `isSiteURLMatch` selector instead of using its own matching logic.

## Implementation Brief

* Add a new selector `isSiteURLMatch( url )` to the `googlesitekit/datastore/site/info.js` datastore. It should pull the current reference site URL and compare it with the incoming `url` argument ignoring protocol, `www.` subdomain prefix, and trailing slash in both URLs. In other words, it should return `TRUE` for `http://example.com/` and `https://www.example.com` URLs, but `FALSE` for `http://example.com/` and `http://example.com/subsite/`.
* Update the `getMatchingWebDataStream( propertyID )` selector from the `analytics-4` module to use the new `isSiteURLMatch( url )` selector instead of relying on custom logic to compare URLs.

### Test Coverage

* Add js tests to check the new selector.

### Visual Regression Changes

* N/A

## QA Brief

* Verify that the new `isSiteURLMatch` selector is added to the `core/site` datastore.
* Check that the `getMatchingWebDataStream` selector from the `modules/analytics-4` datastore uses it.

## Changelog entry

* Add a selector to the `core/site` datastore for checking whether or not a given URL matches the site's reference URL.

felixarntz:
This is almost a duplicate of #2981, but it comes with one nuance worth including, which is the reusable `isSiteURLMatch` selector. I'll rephrase the ACs accordingly.

felixarntz:
IB ✅

ivankruchkoff:
[...]

PR #3176, opened 2021-04-16
Enhancement/3105 match site url selector

eugene-manuilov:
## Summary

Addresses issue #3105

## Checklist


assets/js/googlesitekit/datastore/site/info.test.js
```diff
@@ -359,5 +359,27 @@ describe( 'core/site site info', () => {
...
 			} );
 		} );
+
+		describe( 'isSiteURLMatch', () => {
+			beforeEach( async () => {
+				await registry.dispatch( STORE_NAME ).receiveSiteInfo( baseInfo );
+			} );
+
+			it( 'should return TRUE when URL matches the reference site URL even if the protocol is different', () => {
+				expect( registry.select( STORE_NAME ).isSiteURLMatch( 'https://example.com' ) ).toBe( true );
```
aaemnnosttv, line 369: It would be good to add both here for clarity.
```suggestion
				expect( registry.select( STORE_NAME ).isSiteURLMatch( 'http://example.com' ) ).toBe( true );
				expect( registry.select( STORE_NAME ).isSiteURLMatch( 'https://example.com' ) ).toBe( true );
```

assets/js/googlesitekit/datastore/site/info.test.js
```diff
@@ -359,5 +359,27 @@ describe( 'core/site site info', () => {
...
+			} );
+
+			it( 'should return TRUE when URL matches the reference site URL even with www. subdomain', () => {
+				expect( registry.select( STORE_NAME ).isSiteURLMatch( 'http://www.example.com' ) ).toBe( true );
```
aaemnnosttv, line 373: Just like previous suggestion, with slightly different title.
```suggestion
			it( 'should return TRUE when URL matches the reference site URL with or without a www. subdomain', () => {
				expect( registry.select( STORE_NAME ).isSiteURLMatch( 'http://example.com' ) ).toBe( true );
				expect( registry.select( STORE_NAME ).isSiteURLMatch( 'http://www.example.com' ) ).toBe( true );
```

assets/js/googlesitekit/datastore/site/info.test.js
```diff
@@ -359,5 +359,27 @@ describe( 'core/site site info', () => {
...
+			} );
+
+			it( 'should return TRUE when URL matches the reference site URL even with trailing slash', () => {
+				expect( registry.select( STORE_NAME ).isSiteURLMatch( 'http://example.com/' ) ).toBe( true );
```
aaemnnosttv, line 377:
```suggestion
			it( 'should return TRUE when URL matches the reference site URL with or without a trailing slash', () => {
				expect( registry.select( STORE_NAME ).isSiteURLMatch( 'http://example.com' ) ).toBe( true );
				expect( registry.select( STORE_NAME ).isSiteURLMatch( 'http://example.com/' ) ).toBe( true );
```

assets/js/googlesitekit/datastore/site/info.js
```diff
@@ -508,6 +508,23 @@ export const selectors = {
...
+	 * @since n.e.x.t
+	 *
+	 * @param {string} url The URL to compare with the reference site URL.
+	 * @return {boolean} TRUE if the URL matches reference site URL, otherwise FALSE.
+	 */
+	isSiteURLMatch: createRegistrySelector( ( select ) => ( state, url ) => {
+		const referenceURL = select( STORE_NAME ).getReferenceSiteURL();
+		const normalizeURL = ( incomingURL ) => incomingURL
+			.replace( /^https?:\/\/(www\.)?/i, '' ) // Remove protocol and optional "www." prefix from the URL.
+			.replace( /\/$/, '' ); // Remove trailing slash.
```
aaemnnosttv, line 524: The ACs reference `Module::permute_site_url` for the behavior here (although "similar" to it). That method does not consider the path of the site URL although this function will as-is since only a trailing slash is stripped. I'm not sure if that's intended.

@felixarntz should this selector consider the site URL's path at all in the matching or only hostname and scheme variants?

aaemnnosttv (commented):
Thanks @eugene-manuilov – this looks good, just a few suggestions for the tests and one question about the logic around the matching for @felixarntz to clarify.

aaemnnosttv (commented):
_Duplicate from review submit quirk_

assets/js/googlesitekit/datastore/site/info.js
felixarntz: @aaemnnosttv Good catch! This is actually wrong in `Module::permute_site_url`, but correct here. We _do_ need to consider a potential path in the site URL.

For the current behavior of the plugin, this only causes a problem for the site verification logic in the non-proxy setup flow, so that is not a high priority. Elsewhere that we use `Module::permute_site_url` we are only using it to get to hostnames anyway, so there it doesn't make a difference either way.

But I'll open an issue to fix this method, since we should have a unified approach to this. Long story short, the implementation _here_ is good.

felixarntz: Opened https://github.com/google/site-kit-wp/issues/3233

aaemnnosttv (approved):
Great, thanks!

Issue #2888, opened 2021-02-26
Analytics view selection not preserved by useExistingTagEffect for properties with multiple views


aaemnnosttv:
## Bug Description

The `useExistingTagEffect` in Analytics will set/select the account and property for an existing detected Analytics property, either from a tag on the page or GTM container tag. Due to logic in the `selectProperty` action which clears the current profile (view), this has the effect of selecting the first profile whenever Analytics settings are edited.

https://github.com/google/site-kit-wp/blob/5268ae9f21e589a8c3104a4d98f63b372a88a916/assets/js/modules/analytics/hooks/useExistingTagEffect.js#L78-L88

The presence of an existing tag forces the detected account and property to be selected, but the view can always be changed since this is specific to reporting only.

For properties with multiple views, this will cause the chosen view to be changed when Analytics settings are edited if the previously selected view is any one other than the first.

### Pre-Setup ###
1. Find a UA tag that your account has access to, you can do this by setting up with analytics module and then looking at the page source for the following `gtag("config", "UA-` and noting that `UA-XXXXXXXX-X` tag.
2. Setup a new view for your tag that is not `All Web Site Data` for the analytics tag that you had just noted.
<img width="227" alt="Screen Shot 2021-04-19 at 10 18 31 AM" src="https://user-images.githubusercontent.com/373931/115277112-b9888e80-a0f8-11eb-8f10-fc33b66a9f0c.png">

### Repro ###
See the video attached:

1. Enable sitekit tester plugin, and add the UA tag from above to `Place Analytics existing tag` in the settings page `/wp-admin/admin.php?page=googlesitekit-tester`
2. Disconnect Analytics if it's already connected, and open the `Set Up Analytics` link
3. Pick your new view, save it.
3. Edit Analytics settings again.
4. Notice the first view is now selected `All Web Site Data` in the dropdown and "Configure Analytics" is blue indicating settings have changed and saving is possible, if you pick the view you created, the `Configure Analytics` button will go grey as you are no longer changing the setting.

https://user-images.githubusercontent.com/373931/115278119-ee491580-a0f9-11eb-900d-e7833d70acf8.mp4

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The presence of an Analytics tag should not change the selected view when the detected account and property matches the currently configured values

## Implementation Brief

[...]

PR #3180, opened 2021-04-17
Preserve selected profile ID when selecting its respective property

ivankruchkoff:
## Summary

Addresses issue #2888 

## Checklist


aaemnnosttv:
@ivankruchkoff we should only update the base to main after its cut in preparation for the release for issues that are tagged with that version 👍  2888 wasn't tagged yet so it should be assumed to target develop as usual.

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -173,21 +173,28 @@ const baseActions = {
...
 
 			const profiles = registry.select( STORE_NAME ).getProfiles( accountID, propertyID );
-			if ( property.defaultProfileId && profiles?.some( ( profile ) => profile.id === property.defaultProfileId ) ) { // eslint-disable-line sitekit/acronym-case
-				registry.dispatch( STORE_NAME ).setProfileID( property.defaultProfileId ); // eslint-disable-line sitekit/acronym-case
-				return;
-			}
 
 			if ( profiles === undefined ) {
 				return; // Selection will happen in in getProfiles resolver.
 			}
```
aaemnnosttv, line 185: If `profiles` is `undefined` and the selected profile was cleared, then this will still cause the same problem as before (selecting the first even when the selected profile belongs to the selected property).

In this case, I think we shouldn't rely on the `getProfiles` resolver to make any selection for us and it should be handled in this action instead.

Here's what I'm thinking:

- Remove the clearing of the profile ID – the `getProfiles` resolver will only set it if it isn't set
- Select `getProfiles` using `__experimentalResolveSelect` and use `commonActions.await` for the result – after which its resolver will be completed so we won't need to worry about it
- Check if the selected profile ID exists in the returned profiles
    - If it does, nothing more should be needed
    - If not, fallback to the default profile ID from the selected property as we do now
    - Finally, fall back to the first profile returned, or if none: profile_create

assets/js/modules/analytics/components/common/ProfileSelect.test.js
```diff
@@ -105,9 +105,10 @@ describe( 'ProfileSelect', () => {
 		const { getAllByRole } = render( <ProfileSelect />, { setupRegistry } );
 
 		const listItems = getAllByRole( 'menuitem', { hidden: true } );
-		// Note: we do length + 1 here because there should also be an item for
+
+		// Note: we do a length = 3 corresponding to the 2 items in the properties profiles and one more for
 		// "Set up a new property".
-		expect( listItems ).toHaveLength( fixtures.accountsPropertiesProfiles.profiles.length + 1 );
+		expect( listItems ).toHaveLength( 3 );
```
aaemnnosttv, line 111: I prefer the previous version here as it will be more resilient to change (i.e. someone may update `toHaveLength( 3 )` in the future but not the comment).

Also, the comment is a little confusing as written: "items in the properties profiles"

aaemnnosttv (changes requested):
Thanks @ivankruchkoff ! This is looking good but needs a slightly different approach I think. See my comment below.

assets/js/modules/analytics/datastore/properties.js
ivankruchkoff: Updated in https://github.com/google/site-kit-wp/pull/3180/commits/3c3c7835690e44b8f0c3e4d81fe5d3014d7a50f7#diff-5da51f519c7f1141354d71a4f94d41b1de0ca5589b8f731be7edaedb1079864dR177

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -173,21 +173,24 @@ const baseActions = {
...
-			if ( profiles === undefined ) {
-				return; // Selection will happen in in getProfiles resolver.
+			let returnProfile = profiles.find( ( { id: ID } ) => ID === existingProfileID );
+
+			if ( ! returnProfile ) {
+				if ( property.defaultProfileId && profiles?.some( ( profile ) => profile.id === property.defaultProfileId ) ) { // eslint-disable-line sitekit/acronym-case
+					registry.dispatch( STORE_NAME ).setProfileID( property.defaultProfileId ); // eslint-disable-line sitekit/acronym-case
+					return;
+				}
+				returnProfile = profiles.find( ( { webPropertyId } ) => webPropertyId === propertyID ) || { id: PROFILE_CREATE }; // eslint-disable-line sitekit/acronym-case
```
aaemnnosttv, line 190: I realize this `find` was here from before it looks like it's my own left over from when another action was consolidated here that relied the `propertyID` alone. See https://github.com/google/site-kit-wp/commit/28fe5230a8cfd3c8973dd3f3bcb1b6ad6e61be6e

This can be simplified to get the first profile in the list instead (or fall back to create as it does) since all profiles in `profiles` should be for the given `propertyID`.

aaemnnosttv (commented):
Thanks @ivankruchkoff – this is pretty close 😄 
See my comments below and let me know if you have any questions.

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -173,21 +173,24 @@ const baseActions = {
...
-			registry.dispatch( STORE_NAME ).setProfileID( '' );
+			const existingProfileID = registry.select( STORE_NAME ).getProfileID(); // eslint-disable-line @wordpress/no-unused-vars-before-return
+			const profiles = yield Data.commonActions.await( registry.__experimentalResolveSelect( STORE_NAME ).getProfiles( accountID, propertyID ) );
 
-			const profiles = registry.select( STORE_NAME ).getProfiles( accountID, propertyID );
-			if ( property.defaultProfileId && profiles?.some( ( profile ) => profile.id === property.defaultProfileId ) ) { // eslint-disable-line sitekit/acronym-case
-				registry.dispatch( STORE_NAME ).setProfileID( property.defaultProfileId ); // eslint-disable-line sitekit/acronym-case
-				return;
+			if ( profiles === undefined ) {
+				return; // Selection will happen in getProfiles resolver.
```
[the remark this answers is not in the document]
ivankruchkoff: Updated in https://github.com/google/site-kit-wp/pull/3180/commits/3d0cb75b81486afec4975bdce2ed6e13b3b2c920

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -173,21 +173,24 @@ const baseActions = {
...
 			}
 
-			if ( profiles === undefined ) {
-				return; // Selection will happen in in getProfiles resolver.
+			let returnProfile = profiles.find( ( { id: ID } ) => ID === existingProfileID );
```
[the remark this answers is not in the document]
ivankruchkoff: Updated in https://github.com/google/site-kit-wp/pull/3180/commits/27342ee6405bbea9a567439024d735214f4ca1c2

assets/js/modules/analytics/datastore/properties.js
ivankruchkoff: Updated in https://github.com/google/site-kit-wp/pull/3180/commits/27342ee6405bbea9a567439024d735214f4ca1c2

assets/js/modules/analytics/datastore/properties.js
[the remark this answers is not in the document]
ivankruchkoff: updated in https://github.com/google/site-kit-wp/pull/3180/commits/27342ee6405bbea9a567439024d735214f4ca1c2

assets/js/modules/analytics/datastore/properties.js
ivankruchkoff: updated in https://github.com/google/site-kit-wp/pull/3180/commits/27342ee6405bbea9a567439024d735214f4ca1c2

assets/js/modules/analytics/datastore/properties.js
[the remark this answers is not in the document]
ivankruchkoff: Updated in https://github.com/google/site-kit-wp/pull/3180/commits/3d0cb75b81486afec4975bdce2ed6e13b3b2c920

assets/js/modules/analytics/components/common/ProfileSelect.test.js
ivankruchkoff: Resolved

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -179,21 +179,26 @@ const baseActions = {
 
 			registry.dispatch( STORE_NAME ).setInternalWebPropertyID( internalPropertyID || '' );
 
-			// Clear any profile ID selection in the case that selection falls to the getProfiles resolver.
-			registry.dispatch( STORE_NAME ).setProfileID( '' );
+			const existingProfileID = registry.select( STORE_NAME ).getProfileID(); // eslint-disable-line @wordpress/no-unused-vars-before-return
+			const profiles = yield Data.commonActions.await( registry.__experimentalResolveSelect( STORE_NAME ).getProfiles( accountID, propertyID ) );
```
aaemnnosttv, line 183: This is a bit long for one line 🙂 
```suggestion
			const profiles = yield Data.commonActions.await(
				registry.__experimentalResolveSelect( STORE_NAME ).getProfiles( accountID, propertyID )
			);
```

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -179,21 +179,26 @@ const baseActions = {
...
-			if ( property.defaultProfileId && profiles?.some( ( profile ) => profile.id === property.defaultProfileId ) ) { // eslint-disable-line sitekit/acronym-case
-				registry.dispatch( STORE_NAME ).setProfileID( property.defaultProfileId ); // eslint-disable-line sitekit/acronym-case
-				return;
+			if ( ! Array.isArray( profiles ) ) {
+				return; // Something unexpected occurred and we want to avoid type errors.
 			}
 
-			if ( profiles === undefined ) {
-				return; // Selection will happen in in getProfiles resolver.
+			const matchedProfileID = profiles.find( ( { id: ID } ) => ID === existingProfileID )?.id;
```
aaemnnosttv, line 189: This is only relevant if a profile ID was previously set. `existingProfileID` may be an empty string here.

aaemnnosttv (commented):
Thanks @ivankruchkoff – this is looks close to ready. I left a few comments below; nothing major. Most importantly though, all E2E is currently broken.

assets/js/modules/analytics/datastore/properties.js
ivankruchkoff: Addressed in https://github.com/google/site-kit-wp/pull/3180/commits/03c3db0aa55a9d3e503d8aab50d6318fe85f8b28

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -179,21 +179,26 @@ const baseActions = {
...
+
+			if ( matchedProfileID ) {
+				registry.dispatch( STORE_NAME ).setProfileID( matchedProfileID );
```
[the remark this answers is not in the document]
ivankruchkoff: Addressed in https://github.com/google/site-kit-wp/pull/3180/commits/03c3db0aa55a9d3e503d8aab50d6318fe85f8b28

aaemnnosttv:
@googlebot I consent.

assets/js/modules/analytics/datastore/__fixtures__/properties-profiles.json
```diff
@@ -620,6 +620,47 @@
...
+      "selfLink": "https://www.googleapis.com/analytics/v3/management/accounts/151753095/webproperties/UA-151753095-1/profiles/205382011",
+      "siteSearchCategoryParameters": null,
+      "siteSearchQueryParameters": null,
+      "starred": null,
+      "stripSiteSearchCategoryParameters": null,
+      "stripSiteSearchQueryParameters": null,
+      "timezone": "America/Los_Angeles",
+      "type": "WEB",
+      "updated": "2019-11-07T11:19:21.809Z",
+      "webPropertyId": "UA-151753095-1",
```
aaemnnosttv, line 646: @ivankruchkoff you added an additional profile to the list here but it was referencing a different property ID than the other profile in the list which would be incorrect here. The `properties-profiles` endpoint returns all properties, but the `profiles` returned are always for a single property (either the matched property if there is one, or just the first) – it doesn't return all profiles for each of the properties (that would be way too many requests on the server side). I updated this to be mostly a copy of the other one in the list here with a few differences as this is still needed for the added test 👍

aaemnnosttv (approved):
Thanks @ivankruchkoff – As you know, I've put a few finishing touches on this in working through some of the problems we were seeing so this should be good to go now.

eugene-manuilov (approved):
Looks good to me.

tofumatt (approved):
Looks good to me as well 👍🏻

Issue #98, opened 2019-07-02
Add full URL to Analytics dashboard


jamesozzie:
## Feature Description

It would be great to see the full URL on hover in the Analytics dashboard rather than just the page titles. In some cases websites feature the same page titles for some of their URLs, in particular websites with a large inventory of content with auto generated sub-pages. 

## Proposed solution
Add a full URL on hover of the page title. Or alternatively add the full URL below the page title. 

![uidrBdWHhaE](https://user-images.githubusercontent.com/41326532/60498595-d13a1d00-9cae-11e9-99ec-0db86905b4e0.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The list of the most popular pages on the Analytics module page should also include the URL path for each page instead of only the page title (the same way this is in the popular pages widget in the Site Kit dashboard or WordPress dashboard widget).
* This should be updated in the legacy version of the widget as well as the refactored version (see #2068).

## Implementation Brief
* Using `/assets/js/components/DetailsPermaLinks.js`
	* Add a new prop, `serviceURL`: String and having `null` as default value.
	* The `googlesitekit-display-block` class name should only be applied to the first link if `serviceURL` is null, else the external icon will be aligned to the far right of the link.
	* The `href` of the first link should display `serviceURL` if it's not null, else display `detailsURL`.
	* The `external` prop of the first link should be true if `serviceURL` is not null, so that the external icon is displayed.
* Using `/assets/js/modules/analytics/components/module/ModulePopularPagesWidget/index.js`,
	* Update the first item of the `tableColumns` array to use the `DetailsPermaLinks` instead of the `Link` component, passing the `title`, `url` as the `path` prop and `serviceURL`.
* The changes from the previous bullet point should be applied to the legacy component as well, i.e `/assets/js/modules/analytics/components/dashboard/LegacyAnalyticsDashboardWidgetTopPagesTable.js`.

### Test Coverage
* N/A

### Visual Regression Changes
* Visual changes should be noted for the above components.

## QA Brief

* on the top pages table of the Analytics dashboard check that:
  * the title links are being displayed by the 'DetailsPermaLinks' component
  * a full URL is displayed on hover of the page title or below the page title.

## Changelog entry

* Update popular pages list tables on the Analytics module dashboard to be consistent with the same components on the main Site Kit and WordPress dashboards.

tofumatt:
IB ✅

johnPhillips:
[...]

PR #3184, opened 2021-04-19
Enhancement/98 analytics url

Hazlitte:
## Summary

Add full URL to Analytics dashboard

Addresses issue #98 

## Checklist


assets/js/components/DetailsPermaLinks.js
```diff
@@ -44,9 +45,10 @@ export default function DetailsPermaLinks( { title, path } ) {
 	return (
 		<Fragment>
 			<Link
-				className="googlesitekit-display-block"
-				href={ detailsURL }
+				className={ serviceURL === null ? 'googlesitekit-display-block' : '' }
```
aaemnnosttv, line 48: Even though we only have one class here, the fact that it's conditional makes it a bit cleaner and more consistent to use `classnames` instead. Also, we don't really need to check if `serviceURL === null` specifically, we can simply use `serviceURL` in a looser truthy comparison. In this case, the class should be added if `! serviceURL`.

assets/js/components/DetailsPermaLinks.js
```diff
@@ -44,9 +45,10 @@ export default function DetailsPermaLinks( { title, path } ) {
...
+				href={ serviceURL !== null ? serviceURL : detailsURL }
```
aaemnnosttv, line 49:
```suggestion
				href={ serviceURL || detailsURL }
```

assets/js/components/DetailsPermaLinks.js
```diff
@@ -44,9 +45,10 @@ export default function DetailsPermaLinks( { title, path } ) {
...
 				inherit
+				external={ serviceURL !== null ? true : false }
```
aaemnnosttv, line 51:
```suggestion
				external={ !! serviceURL }
```

assets/js/modules/analytics/components/dashboard/LegacyAnalyticsDashboardWidgetTopPagesTable.js
```diff
@@ -74,13 +74,7 @@ const LegacyAnalyticsDashboardWidgetTopPagesTable = ( props ) => {
...
 				} );
 				return (
-					<Link
-						href={ serviceURL }
-						external
-						inherit
-					>
-						{ title }
-					</Link>
+					<DetailsPermaLinks title={ title } path={ url } serviceURL={ serviceURL } />
```
aaemnnosttv, line 77: This looks good but please spread props onto their own line when more than 1 (or 2 very short).

assets/js/modules/analytics/components/module/ModulePopularPagesWidget/index.js
```diff
@@ -117,9 +117,7 @@ export default function ModulePopularPagesWidget( { Widget, WidgetReportError, W
 				} ) );
 
 				return (
-					<Link href={ serviceURL } external inherit>
-						{ title }
-					</Link>
+					<DetailsPermaLinks title={ title } path={ url } serviceURL={ serviceURL } />
```
aaemnnosttv, line 120: See previous comment.

assets/js/components/DetailsPermaLinks.js
```diff
@@ -66,3 +68,13 @@ export default function DetailsPermaLinks( { title, path } ) {
...
+
+DetailsPermaLinks.propTypes = {
+	title: PropTypes.string,
+	path: PropTypes.string,
+	serviceURL: PropTypes.string,
+};
+
+DetailsPermaLinks.defaultProps = {
+	serviceURL: null,
+};
```
aaemnnosttv, line 80: Unless there is a reason why we need a specific value to handle this case, I think this can be removed. We generally use `null` as our intentional value for none. In this case the default `undefined` should fine if it isn't passed.

aaemnnosttv (changes requested):
Thanks @Hazlitte – just a few things to address below, but also VRT is failing. This will probably require a reference image update but let's review it first to make sure we're getting the expected results first 👍

assets/js/components/DetailsPermaLinks.js
Hazlitte: 👍

assets/js/components/DetailsPermaLinks.js
Hazlitte: 👍

assets/js/components/DetailsPermaLinks.js
Hazlitte: 👍

assets/js/modules/analytics/components/dashboard/LegacyAnalyticsDashboardWidgetTopPagesTable.js
Hazlitte: 👍

assets/js/modules/analytics/components/module/ModulePopularPagesWidget/index.js
Hazlitte: 👍

assets/js/components/DetailsPermaLinks.js
Hazlitte: 👍

assets/js/components/DetailsPermaLinks.js
aaemnnosttv: Sorry @Hazlitte, what I meant was that `classnames` is designed to handle conditional classes and you'll see it used this way throughout much of the codebase. [See here for more information](https://github.com/JedWatson/classnames#usage).

assets/js/components/DetailsPermaLinks.js
```diff
@@ -44,9 +45,10 @@ export default function DetailsPermaLinks( { title, path } ) {
 	return (
 		<Fragment>
 			<Link
-				className="googlesitekit-display-block"
-				href={ detailsURL }
+				className={ ! serviceURL ? classnames( 'googlesitekit-display-block' ) : '' }
```
aaemnnosttv, line 48: Maybe this doesn't need the class at all since the link below has it as well, they'll always be on separate lines visually. I took a quick look and it seems to look the same in both cases. Please confirm and if so, let's just remove it since it otherwise doesn't make much sense to render it conditionally.

aaemnnosttv (commented):
Thanks @Hazlitte – this looks good, just one comment left about the conditional class.

assets/js/components/DetailsPermaLinks.js
Hazlitte: You are right, I have removed it 👍

aaemnnosttv (approved):
Great, thanks @Hazlitte !

Issue #2183, opened 2020-10-15
Refactor `SettingsModule` component


felixarntz:
The `SettingsModule` component should be refactored to be a functional component, relying fully on the datastore. It should be split into sub-components where it makes sense.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `SettingsModule` component should be refactored into a new functional, hook-based, datastore-driven `SettingsActiveModule` component (the file should also be renamed to match the component name).
* The new component should be only for the use-case that the module is already active, so it can assume that. For the scenario where previously `SettingsModule` was also rendered for the inactive modules (see now `SettingsInactiveModules` as of #2184), that component should now use the existing `SetupModule` component directly, which the current `SettingsModule` component simply wraps. Further refactoring of the _inactive_ module settings component should not be necessary.
* The `SettingsActiveModule` component should now entirely rely on module data from the datastore (`getModule` selector based on the `slug` prop) instead of using the `getModulesData()` legacy function.
* In general, it should no longer use _any_ legacy functions, e.g. module deactivation should also happen via the datastore.
* The `SettingsActiveModules` component should be updated accordingly, if needed (those changes should be minimal though since its refactoring should already be completed when defining this issue, see #2184).
* The following now unused legacy functions should be removed:
    * `getReAuthURL`
    * `activateOrDeactivateModule`
    * if anything else comes up as unused during IB / execution, it should be removed too

## Implementation Brief
* Rename `SettingsModule` component from `/assets/js/components/settings/settings-module.js` to `SettingsActiveModule` component and rename file accordingly.
* Refactor `SettingsActiveModule` to be a functional component with the following changes:
	* Legacy functions such as `getReAuthURL` and `activateOrDeactivateModule` should removed in favour of querying the module data store.
	* To get the auth URL, query the data store like this:
		```js
		const moduleStoreName = useSelect( ( select ) => select( CORE_MODULES ).getModuleStoreName( slug ) );
		const adminReauthURL = useSelect( ( select ) => select( moduleStoreName )?.getAdminReauthURL() );
		```
	* The `getDependentModules` should be removed in favour of getting the module dependants like this:
		```js
		const dependents = useSelect( ( select ) => select( CORE_MODULES )?.getModuleDependantNames( slug ) );
		```
	* Since all modules are assumed to be active, the `activateOrDeactivate` function can be simplified and possibly renamed to use `deactivateModule` function which can be imported as such:
		```js
[...]

PR #3196, opened 2021-04-20
Navigate to dashboard after module deactivation.

aaemnnosttv:
## Summary

Addresses issue #2183 

## Checklist

### DIFF
--- a/assets/js/components/settings/SettingsActiveModule/ConfirmDisconnect.js
+++ b/assets/js/components/settings/SettingsActiveModule/ConfirmDisconnect.js
@@ -33,6 +33,7 @@ import { ESCAPE } from '@wordpress/keycodes';
  */
 import Data from 'googlesitekit-data';
 import { CORE_MODULES } from '../../../googlesitekit/modules/datastore/constants';
+import { CORE_SITE } from '../../../googlesitekit/datastore/site/constants';
 import { clearWebStorage } from '../../../util';
 import Dialog from '../../Dialog';
 const { useSelect, useDispatch } = Data;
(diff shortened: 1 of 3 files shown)


assets/js/components/settings/SettingsActiveModule/ConfirmDisconnect.js
```diff
@@ -71,9 +71,9 @@ export default function ConfirmDisconnect( { slug, handleDialog } ) {
 
 		if ( ! error ) {
 			clearWebStorage();
-			global.location.assign( adminReauthURL );
+			global.location.assign( dashboardURL );
```
felixarntz, line 74: This should use the `core/location` data store instead.

Then, related to this, we also need to make sure the dialog remains open until the navigation/redirect is complete (basically it just shouldn't close). See https://github.com/google/site-kit-wp/issues/2183#issuecomment-823985934

felixarntz (changes requested):
There's one more thing that needs to be fixed here.

aaemnnosttv:
@felixarntz @wpdarren I think this should be good now. I've added a few changes that result in a cleaner experience although I think this should probably solved a little differently in the future given that this is a pattern that will come up in other places/situations, even now.

The latest changes here:

- Keeps dialog on screen until page refreshes – previously this was lost because of the disconnection, the module went into an unconnected state and so the settings component switched to `SettingsSetupIncompleteComponent` which had the side-effect of unmounting the dialog as well
- Updates dialog to preserve "in progress" state until page refreshes
- Uses datastore for navigation (as we should), although that state isn't currently used elsewhere in these changes

https://user-images.githubusercontent.com/1621608/115608898-852af300-a2ef-11eb-8c9e-1cc9981ef1a5.mp4

You'll notice there are still a few details which change as a result of the disconnect (button text is perhaps most obvious) but also the connected state indicator of the module in the top right, although this is probably to be expected 😄  These are pretty minor and not worth further changes here IMO.

assets/js/components/settings/SettingsActiveModules.js
```diff
@@ -39,9 +39,23 @@ const { useDispatch, useSelect } = Data;
...
+		// Only set initialActiveSlugs once, as soon as modules are available.
+		if ( ! modules || initialActiveSlugs !== undefined ) {
+			return;
+		}
 
+		const activeSlugs = Object.keys( modules )
+			.filter( ( slug ) => modules[ slug ].active );
+
+		setInitialActiveSlugs( activeSlugs );
+	}, [ modules ] );
```
felixarntz, line 57: Nice! I wish we wouldn't need such kind of changes - at _some_ point we can hopefully take some time to explore making all the screens a real single page application so that we don't need to have those "hacks" in order to have the UI wait work navigations between screens.

felixarntz (approved):
Awesome!

assets/js/components/settings/SettingsActiveModules.js
felixarntz: @aaemnnosttv Not for this PR, but in general: Maybe we could introduce a utility to simplify getting the "initial loaded" value of a selector? Maybe something like `useInitialSelect` that uses a mix of `useSelect`, `useState`, and `useEffect` to achieve behavior close to this (just simpler), so that the function eventually always returns the initial loaded value (i.e. the first value that `useSelect` returns which is not `undefined`).

Issue #3072, opened 2021-04-01
Add use-exhaustive-deps to our ESLint rules.


tofumatt:
## Feature Description

Further to [our decision to use `useCallback` when defining callback functions in functional components](https://github.com/google/site-kit-wp/issues/2351#issuecomment-812013536), we should add `use-exhaustive-deps` to our list of ESLint rules so we're alerted to any `useCallback` hooks that don't include all of the relevant dependencies in them.

This will affect `useEffect` code as well, and might highlight existing places where our dependency lists weren't complete.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All `useCallback` and `useEffect` hooks should be linted to ensure their dependencies are accounted for, so `useCallback` and `useEffect` hooks do not contain stale references to variables.

## Implementation Brief

* Add https://www.npmjs.com/package/eslint-plugin-react-hooks to our `package.json` and enable it in our `eslintrc` by adding `"react-hooks"` to our ESLint `plugins` array and adding `"react-hooks/exhaustive-deps": ["error"}]` to our `rules` configuration.
* Fix any existing ESLint errors or document ones that are not legitimate errors (there should be few/none, but it's possible a few `useEffect`s might have legitimate dependency mismatches.

### Test Coverage

* N/A.

### Visual Regression Changes

* N/A.

## QA Brief

* `npm run lint:js` should pass.

## Changelog entry

* N/A

eugene-manuilov:
> Add https://www.npmjs.com/package/eslint-plugin-react-hooks to our package.json and enable it in our eslintrc by adding "plugin:react-hooks/recommended" to our "extends" array.

@tofumatt it means that we will add all rules configured by `eslint-plugin-react-hooks` by default with their default settings. I think it can easily cause side effects and throw unexpected errors that are out of the scope for this ticket. Let's better use custom configuration as it's described in the [readme](https://www.npmjs.com/package/eslint-plugin-react-hooks#custom-configuration) to enable only `react-hooks/exhaustive-deps` rule from that package.

tofumatt:
Fair enough. The other hooks linting is useful as well and I think worth using, but you're right that it might inflate the scope of the issue. I've adjusted the IB to just worry about exhaustive-deps for now.

eugene-manuilov:
Thanks, @tofumatt! IB :heavy_check_mark:

danielgent:
[...]

PR #3199, opened 2021-04-20
Infrastructure/3072 add use exhaustive deps to es lint

danielgent:
## Summary

Addresses issue #3072

## Checklist


.eslintrc.json
```diff
@@ -2,7 +2,8 @@
   "extends": [
     "plugin:@wordpress/eslint-plugin/recommended-with-formatting",
     "plugin:jest/recommended",
-    "plugin:sitekit/main"
+    "plugin:sitekit/main",
+    "plugin:react-hooks/recommended"
```
danielgent, line 6: The IB says to only enable one of `react-hook`'s two rules, *but* enabling `react-hooks/rules-of-hooks` didn't generate a single warning (Eugene was worried here => https://github.com/google/site-kit-wp/issues/3072#issuecomment-817684043 )

`react-hooks/rules-of-hooks` gives great warnings for people who are learning hooks

assets/js/components/ActivateModuleCTA.js
```diff
@@ -62,7 +62,7 @@ const ActivateModuleCTA = ( { moduleSlug, title, description } ) => {
 				type: 'win-error',
 			} );
 		}
-	}, [ activateModule ] );
+	}, [ activateModule, navigateTo, moduleSlug ] );
```
danielgent, line 65: most of this PR are `useCallback` dependency arrays being populated with all the dependencies that eslint tells us

assets/js/components/GoogleChart.js
```diff
@@ -169,7 +171,7 @@ export default function GoogleChart( props ) {
 
 			clearInterval( interval );
 		};
-	}, [] );
+	} );
```
danielgent, line 174: I've assumed that any `useEffect` with an empty dependency array is meant to only run on mount (as opposed to being an error!)

assets/js/components/GoogleChart.js
```diff
@@ -88,7 +89,7 @@ export default function GoogleChart( props ) {
 
 			setChart( googleChart );
 		}
-	}, [ loading, !! chartRef.current, visualizationLoaded, !! chart ] );
+	}, [ loading, visualizationLoaded, chart, chartID, chartType, onReady ] );
```
danielgent, line 92: There are some dependency arrays in the code which have values that don't look like React best practices, and eslint gives all sorts of warnings about these now (not having complex values in the dependencies array, not having unused values), so I'm following eslint's recommendations

assets/js/components/GoogleChart.js
```diff
@@ -126,13 +127,14 @@ export default function GoogleChart( props ) {
 		};
 	}, [
 		chart,
-		JSON.stringify( data ),
-		JSON.stringify( options ),
 		selectedStats,
 		singleStat,
+		chartID,
+		data,
+		options,
```
danielgent, line 134: same with this. `JSON.stringify` creates strings which are immutable and equal in JS if they have the same value, but that is not true with the object references I'm not passing in :thinking: 

The tests still pass but I'm worried in this file especially that I may have changed something (I'm happy to revert this and just disable eslint here)

assets/js/components/ResetButton.js
```diff
@@ -49,11 +49,14 @@ function ResetButton( { children } ) {
 	 * the navigate call starting, we will just set a debounce to keep the spinner for 3 seconds.
 	 */
 	const debouncedSetInProgress = useDebounce( setInProgress, 3000 );
-	const mediatedSetInProgress = ( bool ) => bool ? setInProgress( true ) : debouncedSetInProgress( false );
+	const mediatedSetInProgress = useCallback(
```
danielgent, line 52: this dependency to a callback also has to be a callback (eslint now warns us)

assets/js/components/help/HelpMenu.js
```diff
@@ -66,7 +66,7 @@ function HelpMenu( { children } ) {
 
 	const handleMenuSelected = useCallback( () => {
 		toggleMenu( false );
-	} );
+	}, [] );
```
danielgent, line 69: There's no reason for this to be a callback now technically but as per our tech decision, we want all handlers to use `useCallback` now

danielgent: `eslint-plugin-react-hooks` knows that `toggleMenu` comes from `useState` and thus does not need to be a dependency. Which is pretty cool

assets/js/components/settings/SettingsActiveModule/ConfirmDisconnect.js
```diff
@@ -73,7 +73,7 @@ export default function ConfirmDisconnect( { slug, handleDialog } ) {
 			clearWebStorage();
 			global.location.assign( adminReauthURL );
 		}
-	}, [ module?.slug ] );
+	}, [ adminReauthURL, deactivateModule, module.forceActive, slug ] );
```
danielgent, line 76: as this is a `useCallback` and not `useEffect` then IMHO we can just follow `eslint-plugin-react-hooks` recommendations (it's not re-running code on any change here, it's just assuring there are no stale closures)

assets/js/components/user-input/UserInputQuestionnaire.js
```diff
@@ -48,9 +48,9 @@ import { Cell, Row } from '../../material-components';
 import { trackEvent } from '../../util';
 const { useSelect, useDispatch } = Data;
 
-export default function UserInputQuestionnaire() {
-	const steps = [ ...USER_INPUT_QUESTIONS_LIST, 'preview' ];
+const steps = [ ...USER_INPUT_QUESTIONS_LIST, 'preview' ];
```
danielgent, line 51: move out of component so does not become a dynamic dependency

assets/js/modules/analytics/components/common/AccountCreate/CountrySelect.js
```diff
@@ -48,7 +48,7 @@ export default function CountrySelect() {
 				timezone: countriesByCode[ newCountryCode ].defaultTimeZoneId,
 			} );
 		}
-	}, [ setValues ] );
+	}, [ setValues, value ] );
```
danielgent, line 51: I don't know how we haven't got bugs from some of these! As here this would have had a stale `value`

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/DimensionTabs.js
```diff
@@ -45,24 +45,24 @@ import { Select, Option } from '../../../../../material-components';
 import { trackEvent } from '../../../../../util';
 const { useDispatch } = Data;
 
+const tabs = [
```
danielgent, line 48: extracted out as needs to be static as is a dependency

assets/js/modules/adsense/components/dashboard/LegacyAdSenseDashboardWidgetOverview.js
```diff
@@ -85,9 +85,13 @@ export default function LegacyAdSenseDashboardWidgetOverview( props ) {
...
-		currentDataLoaded && previousDataLoaded, // All reports are fetched.
-		!! currentError || !! previousError, // Whether there is an error or not.
-		JSON.stringify( currentRangeData ),
+		currentDataLoaded,
+		currentError,
+		currentRangeData,
+		handleDataError,
+		handleDataSuccess,
+		previousDataLoaded,
+		previousError,
```
danielgent, line 94: As this is a `useEffect` this could be a cause for concern!

Note that `useEffect` re-runs when a dependency changes **not** when a dependency is truthy (the previous code does look a bit like it's expecting that)

assets/js/modules/tagmanager/components/setup/SetupForm.js
```diff
@@ -61,7 +60,6 @@ export default function SetupForm( { finishSetup } ) {
 	const { setValues } = useDispatch( CORE_FORMS );
 	const { activateModule } = useDispatch( CORE_MODULES );
 	const { submitChanges } = useDispatch( STORE_NAME );
-	const dispatchAnalytics = useDispatch( MODULES_ANALYTICS );
```
danielgent, line 64: :thinking:   I'm assuming this was left around from a previous change. But it's fine to remove

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
```diff
@@ -81,16 +81,18 @@ export default function UserDimensionsPieChart( {
...
 
 		return () => {
-			if ( containerRef.current ) {
-				containerRef.current.removeEventListener( 'click', onTooltipClick );
+			if ( currentContainerRef ) {
+				currentContainerRef.removeEventListener( 'click', onTooltipClick );
 			}
 		};
-	}, [ containerRef.current ] );
+	}, [ ] );
```
danielgent, line 95: I'm not sure what was intended before here.  

Changing a ref doesn't re-render so this was just attaching the handler when mounting.  The original div will probably persist because of React's reconciliation algorithm preserving the DOM *but* it could wipe it at any point!

I'm not sure why the handler isn't just assigned straight to the `div` below in the JSX 
e.g. instead of `<div ref={ containerRef }` do `<div onClick={onTooltipClick}`

assets/js/components/user-input/UserInputKeywords.js
```diff
@@ -57,9 +56,6 @@ export default function UserInputKeywords( { slug, max } ) {
 	// https://github.com/google/site-kit-wp/issues/2900#issuecomment-814843972.
 	const [ localValues, setLocalValues ] = useState( values );
 
-	// Need to make sure that dependencies list always has the same number of elements.
-	const dependencies = values.concat( Array( max ) ).slice( 0, max );
```
danielgent, line 61: I'm not following this. The dependencies array needs to have the dependencies so that the callback doesn't have stale values.

assets/js/components/GoogleChart.js
danielgent: Note that this is a `useEffect`, the `useEffect` changes are what concern me!

danielgent: As we are re-running code (where as `useCallback` is just creating a new function that could be ran. so there are no side-effects)

assets/js/modules/tagmanager/hooks/useExistingTagEffect.js
```diff
@@ -42,5 +42,13 @@ export default function useExistingTagEffect() {
...
-	}, [ hasExistingTag, existingTag, hasExistingTagPermission, existingTagPermission ] );
+	},
+	[
+		hasExistingTag,
+		existingTag,
+		hasExistingTagPermission,
+		existingTagPermission,
+		selectAccount,
+		selectContainerByID,
+	]	);
```
tofumatt, line 53: Surprised ESLint doesn't care about this, but how about:

```suggestion
	}, [
		hasExistingTag,
		existingTag,
		hasExistingTagPermission,
		existingTagPermission,
		selectAccount,
		selectContainerByID,
	] );
```

tofumatt (changes requested):
I'm a bit curious why these rules request usage of `useMount`, because there's nothing special about it: [it's just `useEffect(fn, [])`](https://github.com/streamich/react-use/blob/master/src/useEffectOnce.ts#L4).

`useEffect(fn, [])` isn't an obscure React paradigm or anything, it's a very common hooks approach, so I don't see the need for a new function that obscures what's happening here.

The ACs/IBs don't mention including `useMount` at all, and given it's a wrapper around exactly what we're doing I think we should ignore that rule/revert the change. Arguably it's more boilerplate for many React developers who would think to use `useEffect`. Instead of the empty array argument "boilerplate" it's the different import.

I'd rather we stick with `useEffect` as it's immediately recognisable for any React developer so requires less developer onboarding.

----

The rest of the PR largely looks good—I've made some comments about a few things I think need fixing/tweaking, but overall looks good 👍🏻

assets/js/components/GoogleChart.js
danielgent: Is that a request @tofumatt ?  :smile:   I'm happy to disable eslint for any existing code

tofumatt:
Looks like this is failing the linters too:

Screenshot 2021-04-23 at 13 54 50

assets/js/components/settings/SettingsActiveModule/ConfirmDisconnect.js
```diff
@@ -16,11 +16,6 @@
  * limitations under the License.
  */
 
-/**
- * External dependencies
- */
-import PropTypes from 'prop-types';
```
eugene-manuilov, line 22: @danielgent why did you delete prop types for this component? Could you, please, add it back?

assets/js/components/user-input/UserInputKeywords.js
```diff
@@ -43,7 +43,6 @@ const { useSelect, useDispatch } = Data;
 
 export default function UserInputKeywords( { slug, max } ) {
 	const keywordsContainer = useRef();
-	const [ canDeleteKeyword, setCanDeleteKeyword ] = useState( false );
```
eugene-manuilov, line 46: @danielgent why did you delete this?

assets/js/components/ResetButton.js
```diff
@@ -49,11 +49,14 @@ function ResetButton( { children } ) {
...
+		( bool ) => bool ? setInProgress( true ) : debouncedSetInProgress( false ),
+		[ debouncedSetInProgress ]
+	);
```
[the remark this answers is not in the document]
danielgent: @tofumatt If we're allowed to make code changes then I inlined `mediatedSetInProgress` right into the `useEffect` (it's not used elsewhere) **AND** I used proper if-else blocks (as per your suggestion)

package.json
```diff
@@ -122,6 +122,7 @@
     "compare-versions": "^3.6.0",
     "dompurify": "^2.0.12",
     "element-closest": "^3.0.2",
+    "eslint-plugin-react-hooks": "^4.2.0",
```
eugene-manuilov, line 125: @danielgent please, add it as a dev dependency.

eugene-manuilov (dismissed):
Thanks, @danielgent! It looks good in general. Just found a few more issues to address. Could you please check it?

assets/js/components/settings/SettingsActiveModule/ConfirmDisconnect.js
danielgent: Fixed @tofumatt

package.json
danielgent: Ooops @eugene-manuilov ! nice spot! :facepalm: . Fixed :heavy_check_mark:

assets/js/components/settings/SettingsActiveModule/ConfirmDisconnect.js
danielgent: Apolgies @eugene-manuilov .  I managed to remove them in a merge :thinking:   No idea what happened there! They're added back now :heavy_check_mark:

assets/js/components/user-input/UserInputKeywords.js
danielgent: @eugene-manuilov This file really confused me. I'm happy to revert it all and just disable eslint for the file :smile: 

This state didn't appear to do anything. `canDeleteKeyword` was only used for unnecessary dependencies.

the react hooks eslint plugin forced me to remove it as a dependency, and then linting forced me to delete the variable due to it being unused!

assets/js/components/GoogleChart.js
danielgent: @tofumatt  I reverted this file and just added an eslint ignore for this rule :+1:

tofumatt (commented):
I had a bit of a conversion with @danielgent about this and the reasoning for not allowing empty `useEffect` dependency arrays is pretty fair. It adds complexity, but reduces potential for errors in an acceptable way.

So I retract my request for `useMount` to be removed; let's add it back.

assets/js/modules/tagmanager/hooks/useExistingTagEffect.js
```diff
@@ -42,5 +42,12 @@ export default function useExistingTagEffect() {
...
 		} )();
-	}, [ hasExistingTag, existingTag, hasExistingTagPermission, existingTagPermission ] );
+	},	[
+		hasExistingTag,
+		existingTag,
+		hasExistingTagPermission,
+		existingTagPermission,
+		selectAccount,
+		selectContainerByID,
+	]	);
```
tofumatt, line 52: What's with the indentation of the `[` and `]` here?

```suggestion
	}, [
		hasExistingTag,
		existingTag,
		hasExistingTagPermission,
		existingTagPermission,
		selectAccount,
		selectContainerByID,
	] );
```

tofumatt (changes requested):
Overall looks great! One little styling tweak and a few merge conflicts need resolving, but then this should be good-to-merge. 👍🏻

tofumatt (approved):
Looks good to me, thanks for the work on this! 👍🏻

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/UserDimensionsPieChart.js
tofumatt: Actually, upon further inspection this is a very specific handler for the tooltips that this chart renders, so keeping this here is appropriate.

Issue #3192, opened 2021-04-19
Only run GH actions for PRs against specific branches


felixarntz:
While we restrict our GitHub actions for pushes to only run against `main` and `develop`, we don't do that for pull-requests - they currently trigger the workflows no matter which branch a PR is opened against.

This currently causes quite some overload in our Actions queue, which is largely caused by a separate branch `feature/swg-module` (in which an early exploration for a potential future feature is happening), since the early work is also being reviewed in a PR (against another separate branch `feature/swg-module-reviewed`).

While it can be beneficial to run workflows also for PRs against other branches than `main` and `develop`, we should choose an opt-in approach on a case-by-case basis, rather than running the workflows on any PR, due to the problems this can cause (see above).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All GitHub workflows should only be triggered for pull-requests which are opened against either `main` or `develop`.

## Implementation Brief

* Merge #3193.

### Test Coverage

* N/A

### Visual Regression Changes

* N/A

## QA Brief

* When opening a PR against a random branch other than `develop` and `main`, the GitHub workflows should not run.

## Changelog entry

* N/A

felixarntz:
@aaemnnosttv I've already pushed the same changes that are in #3193 to the feature branch in question, to prevent more workflows from spinning up. This effectively serves as QA I'd argue :)

aaemnnosttv:
IB+CR ✅

ivankruchkoff:
QA:Eng ✅

If your branches `feature/swg-module` and `feature/swg-module-reviewed` are constantly getting `develop`/`main` merged into them, then that will solve the issue on those branches too, otherwise make sure to cherrypick this commit into each of those branches @felixarntz

Issue #3130, opened 2021-04-12
Allow 3P capability management plugins to get list of Site Kit capabilities


felixarntz:
There are various 3P plugins which allow managing WordPress capabilities. Site Kit's capabilities are currently not available in an easy-to-access way, so introducing a utility method that returns a list with all of its base capabilities would be helpful.

Originally requested by @JoryHogeveen via #987.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a `get_capabilities()` method in the `Permissions` class which returns an array with all of Site Kit's base capabilities.

## Implementation Brief

* Complete and merge #987. Or, since the PR #987 is quite old and has gotten stale, we may need to open a new one for this.

### Test Coverage

* N/A

### Visual Regression Changes

* N/A

## QA Brief

* Ensure the new method returns all Site Kit base capabilities (expect the publish one, which is unused and will soon be completely removed).

## Changelog entry

* Add method to get list of all base capabilities used by Site Kit.

Hazlitte:
Hi @felixarntz, I'm not sure what to do about test coverage here.  I checked the PHPUnit files and there isn't a test file for the Permissions class.

Hazlitte:
@felixarntz Ok, thank you for confirming.

ivankruchkoff:
QA ✅

```
$ wp eval 'var_dump( \Google\Site_Kit\Core\Permissions\Permissions::get_capabilities() );'
array(6) {
  [0]=>
  string(26) "googlesitekit_authenticate"
  [1]=>
  string(19) "googlesitekit_setup"
  [2]=>
  string(33) "googlesitekit_view_posts_insights"
  [3]=>
  string(28) "googlesitekit_view_dashboard"
  [4]=>
  string(33) "googlesitekit_view_module_details"
  [5]=>
  string(28) "googlesitekit_manage_options"
}
```

These correspond to:
```php
	/*
	 * Custom base capabilities.
	 */
	const AUTHENTICATE        = 'googlesitekit_authenticate';
	const SETUP               = 'googlesitekit_setup';
	const VIEW_POSTS_INSIGHTS = 'googlesitekit_view_posts_insights';
	const VIEW_DASHBOARD      = 'googlesitekit_view_dashboard';
	const VIEW_MODULE_DETAILS = 'googlesitekit_view_module_details';
	const MANAGE_OPTIONS      = 'googlesitekit_manage_options';
```

Issue #3102, opened 2021-04-07
Implement the ability to query a single GA4 property by ID


felixarntz:
Following up on #2979, it should also be possible to look up a single GA4 property by ID, in the `analytics-4` module class in PHP and the `modules/analytics-4` store in JS.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `Analytics_4` module class should have a new REST datapoint `GET:property`.
    * It should accept a `propertyID` request parameter, which is a plain property ID (without `properties/` prefix).
    * It should rely on the [properties.get](https://developers.google.com/analytics/devguides/config/admin/v1/rest/v1alpha/properties/get) API endpoint.
    * It should return the property object.
* The `modules/analytics-4` datastore should get a new selector `getProperty( propertyID )`.
    * It should rely on the `GET:modules/analytics-4/data/property` REST datapoint to get the property if it's not already in store state.
    * In other words, the existing `getProperties` resolver should also be updated to reduce the amount of necessary API calls: Any properties returned by `getProperties` should **also** be keyed by ID in state (instead of just under the account they belong to).

## Implementation Brief

* In the `includes/Modules/Analytics_4.php` file:
  * Add `'GET:property' => array( 'service' => 'analyticsadmin' )` to the returning array of the `get_datapoint_definitions` method.
  * Add a new case `GET:property` to the `create_data_request` method. It should return an error (like [here](https://github.com/google/site-kit-wp/blob/develop/includes/Modules/Analytics_4.php#L233-L240)) if `propertyID` is not set in the `$data` array. Otherwise it should return the result of the `$this->get_service( 'analyticsadmin' )->properties->get( $data['propertyID'] )` call.
* In the `modules/analytics-4/datastore/properties.js` file:
  * Create a new fetch store for the `property` endpoint:
    * Use `getProperty` base name.
    * Issue a GET request to the `modules/analytics-4/data/property` endpoint with `propertyID` query argument.
    * Set received property under appropriate `propertyID` key in the `propertiesByIDs`  property of the state object in the `reducerCallback` callback.
    * The `argsToParams` callback should convert the incoming `propertyID` argument to an object with `propertyID` property.
    * Make sure that `propertyID` is valid using the `isValidPropertyID` function in the `validateParams` callback.
  * Update the `reducerCallback` function of the `fetchGetPropertiesStore` datastore. In addition to the existing functionality, it should also populate the `propertiesByID` section of the state object using received properties.
[...]

PR #3200, opened 2021-04-20
Fix pseudo attributes issue in the Analytics 4 module.

eugene-manuilov:
## Summary

Addresses issue #3102 (also merges #3192 and #3130 into `main`, which are fine to be included in the release)

## Checklist


felixarntz:
Updated this against `main`. It now has a little extra code from other issues, which normally would require changing this PR, but each of them is so minor / irrelevant for the release that we can just merge as is. I'll update the other issues touched by this to also be against 1.31.0 accordingly.

Issue #3164, opened 2021-04-15
Display UA property IDs next to property names in dropdowns


felixarntz:
While this issue is not strictly related to GA4 and will immediately result in a user-facing change, it is a step towards GA4 where this will be more crucial; as it is now, it is still an enhancement, since sometimes the name of a property may be duplicated and therefore not clearly indicate which property it is.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The existing Analytics (UA) `PropertySelect` dropdown should be updated to always display the property ID after the name in parentheses, e.g. "My Property (UA-123456-7)".

## Implementation Brief

* Update dropdown option labels in the `PropertySelect` component (`assets/js/modules/analytics/components/common/PropertySelect.js`) to display the property `id` attribute in parentheses after `name`: https://github.com/google/site-kit-wp/blob/a24a1f3c588c73eb1b3a72d0221f7a66ca461882/assets/js/modules/analytics/components/common/PropertySelect.js#L90-L98
* Use `sprintf( __( '%1$s (%2$s)' ), name, id )` to ensure the string is translated/localized properly.

### Test Coverage

* Update existing PropertySelect tests if any of them fails due to the aforementioned changes.

### Visual Regression Changes

* Property IDs will be displayed in the dropdown field after property names.

## QA Brief

* Connect Analytics in Site Kite
* In `Settings` edit the Analytics settings
* Check that the `Property` field displays the name and property ID

## Changelog entry

* Update Analytics property select options to display property IDs in addition to names.

tofumatt:
Looks good, but I added one note to use `sprintf( __( '%1$s (%2$s)' ), name, id )`, so the string is translated/localized properly. I figured better to leave it here than in the code review in case someone missed it, as it's an easy oversight not to translate things like these.

IB ✅

aaemnnosttv:
@Hazlitte would you please add a QA brief?

Hazlitte:
@aaemnnosttv Yes of course.  Apologies that I forgot to do that.

wpdarren:
### QA Update: Pass ✅ 

**Verified:**

- Confirmed that the Property field displays the name and property ID when you edit Analytics in the settings. 

![image](https://user-images.githubusercontent.com/73545194/116280610-c5232780-a780-11eb-9449-1c2d611e735d.png)

PR #3209, opened 2021-04-21
Update to display property id and name.

Hazlitte:
## Summary

Update to display property id and name.

Addresses issue #3164

## Checklist

### DIFF
--- a/assets/js/modules/analytics/components/common/PropertySelect.js
+++ b/assets/js/modules/analytics/components/common/PropertySelect.js
@@ -91,7 +91,7 @@ export default function PropertySelect() {
 					<Option
 						key={ index }
 						value={ id }
-						data-internal-id={ internalWebPropertyId } /* eslint-disable-line sitekit/acronym-case */
+						data-internal-id={ internalWebPropertyId } // eslint-disable-line sitekit/acronym-case
 					>
 						{ internalWebPropertyId // eslint-disable-line sitekit/acronym-case
 							? sprintf(
--- a/assets/js/modules/analytics/components/common/PropertySelect.js
+++ b/assets/js/modules/analytics/components/common/PropertySelect.js
@@ -91,14 +91,16 @@ export default function PropertySelect() {
 					<Option
 						key={ index }
 						value={ id }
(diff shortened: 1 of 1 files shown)


assets/js/modules/analytics/components/common/PropertySelect.js
```diff
@@ -93,7 +93,12 @@ export default function PropertySelect() {
 						value={ id }
 						data-internal-id={ internalWebPropertyId } // eslint-disable-line sitekit/acronym-case
 					>
-						{ name }
+						{ sprintf(
+							/* translators: %1$s: property name, %2$s: property ID */
+							__( '%1$s (%2$s)', 'google-site-kit' ),
+							name,
+							id
+						) }
```
aaemnnosttv, line 101: This looks great for the most part, except we should only do this for actual properties. We don't want to format the value in this way for `property_create` since this is an internal detail.

image

The easiest solution here is to use `internalWebPropertyId` to conditionally render with:
* If truthy, format the value with the name and ID, 
* if there is no internal web property ID, only output the name

aaemnnosttv (changes requested):
Thanks @Hazlitte – just one thing to fix which wasn't foreseen in the original definition.

assets/js/modules/analytics/components/common/PropertySelect.js
```diff
@@ -91,9 +91,16 @@ export default function PropertySelect() {
 					<Option
 						key={ index }
 						value={ id }
-						data-internal-id={ internalWebPropertyId } // eslint-disable-line sitekit/acronym-case
+						data-internal-id={ internalWebPropertyId } /* eslint-disable-line sitekit/acronym-case */
```
aaemnnosttv, line 94: Please use `//` comments for single lines as it was before (was this maybe changed on accident?
```suggestion
						data-internal-id={ internalWebPropertyId } // eslint-disable-line sitekit/acronym-case
```

aaemnnosttv (commented):
Looks good to me, just the one comment 👍

aaemnnosttv (approved):
Thanks @Hazlitte !

Issue #3091, opened 2021-04-05
Show change values in AdSense summary widget


felixarntz:
Originally reported as part of #656: In the AdSense summary widget, only the "Total Earnings" value currently shows up with a change percentage. The "RPM" (or "Page RPM" as of #656) and "Ad Impressions" values should also have a change value.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The AdSense summary widget on the dashboard should be updated as follows (**only for the refactored version**, i.e. `DashboardSummaryWidget`):
    * The change value displayed for the monetary earnings value should be the change value compared to the previous period, rather than the value of "today's" earnings.
    * There should be a change value displayed for the "RPM" / "Page RPM" metric: It should be the absolute change from the previous period value and should be formatted with the currency (similar to how it is for the "Total Earnings" change value, i.e. do not use percentage like we mostly do otherwise).
    * There should be a change value displayed for the "Ad Impressions" metric: Similar to the above, it should be the absolute change from the previous period value and it should be formatted without any unit (i.e. just the absolute difference, not a percentage).

## Implementation Brief

* Using `assets/js/modules/adsense/components/dashboard/DashboardSummaryWidget.js`,
	* Get the `compareStartDate` and `comparedEndDates` to be used for querying the data for the previous period by passing `compare: true` to the `getDateRangeDates` selector.
	* Get a report for the previous period with the same metrics as `periodArgs` and set `startDate` and `endDate` to `compareStartDate` and `compareEndDate` respectively.
	* ~~Use the `calculateChange` function from `assets/js/util/index.js`, to calculate the changed values between the current period and previous period reports data and pass the value as the `change` prop of each `DataBlock` component.~~
	* The `changeDataUnit` prop should be the currency for the first 2 `DataBlock` components and metric for the last one.

### Test Coverage

* N/A

### Visual Regression Changes

* Visual changes to the `DashboardSummaryWidget` component.

## QA Brief

* Make sure you have the ` widgets.dashboard` feature enabled, and have Analytics and AdSense set up.
* Go to the Site Kit dashboard and scroll down to the `DashboardSummaryWidget`. This is the one under the `Earnings` heading that sits on the left hand side, with the spark lines.
* All three values should now display a 'change' (i.e. with up/ down arrow) if the values have changed.
* Page RPM and Earnings should show 'change' with currency units (e.g. Dollar sign).

## Changelog entry

[...]

PR #3214, opened 2021-04-22
Show change values in AdSense summary widget. (3091)

johnPhillips:
## Summary

Addresses issue #3091

## Relevant technical choices

I added a small utility at the top of the file to calculate the absolute change, as the `calculateChange` utility deals in percentages.

## Checklist


assets/js/modules/adsense/components/dashboard/DashboardSummaryWidget.js
```diff
@@ -77,12 +99,15 @@ function DashboardSummaryWidget( { Widget, WidgetReportZero, WidgetReportError }
 		return {
 			today: select( STORE_NAME ).getReport( todayArgs ),
```
aaemnnosttv, line 100: With these changes, we should have no more need for the today-specific report and related logic so I think this (and related) should be removed.

assets/js/modules/adsense/components/dashboard/DashboardSummaryWidget.js
```diff
@@ -38,23 +38,45 @@ import AdBlockerWarning from '../common/AdBlockerWarning';
...
 
+/**
+ * Calculates the absolute change between two values.
+ *
+ * @since n.e.x.t
+ *
+ * @param {number} previous The previous value.
+ * @param {number} current  The current value.
+ * @return {number} The absolute change between values.
+ */
```
aaemnnosttv, line 49: As a local helper, I think the doc block probably isn't necessary. It's basically a 1 line function anyways.

assets/js/modules/adsense/components/dashboard/DashboardSummaryWidget.js
```diff
@@ -38,23 +38,45 @@ import AdBlockerWarning from '../common/AdBlockerWarning';
...
+const calculateAbsoluteChange = ( previous, current ) => {
+	const change = Number( current ) - Number( previous );
+	return change.toFixed( 2 );
```
aaemnnosttv, line 52: We shouldn't need to use `toFixed` in most cases since this is essentially number formatting which should be done by `numFmt` instead.

Without this, the function here essentially becomes `return current - previous` which seems unnecessary to make into a function 😄

assets/js/modules/adsense/components/dashboard/DashboardSummaryWidget.js
```diff
@@ -137,6 +162,12 @@ function DashboardSummaryWidget( { Widget, WidgetReportZero, WidgetReportError }
 	const currencyHeader = period.headers.find( ( header ) => null !== header.currency && 0 < header.currency.length );
 	const currencyCode = currencyHeader ? currencyHeader.currency : false;
 
+	const change = {
+		earnings: calculateAbsoluteChange( previousPeriod.totals[ 0 ], period.totals[ 0 ] ),
+		rpm: calculateAbsoluteChange( previousPeriod.totals[ 1 ], period.totals[ 1 ] ),
+		impressions: calculateAbsoluteChange( previousPeriod.totals[ 2 ], period.totals[ 2 ] ),
```
aaemnnosttv, line 168: I'm not sure the utility adds a benefit as-is. I like the idea but I think it ends up being clearer to do the subtraction here instead as it isn't obvious by the signature which parameter is current and which is previous. If it would be doing more then it would make sense to keep but as it is, it's basically just subtraction and casting numbers (which will happen automatically when subtracting if needed as well).

assets/js/modules/adsense/components/dashboard/DashboardSummaryWidget.js
```diff
@@ -146,6 +177,8 @@ function DashboardSummaryWidget( { Widget, WidgetReportZero, WidgetReportError }
 						title={ __( 'RPM', 'google-site-kit' ) }
 						datapoint={ period.totals[ 1 ] }
 						datapointUnit={ currencyCode }
+						change={ Number( change.rpm ) }
```
aaemnnosttv, line 180: The `Number` cast shouldn't be necessary anymore as the value should already be a number.

aaemnnosttv (changes requested):
Thanks @johnPhillips – this is looking good but has a few things to address yet. I left a few comments below, but also the ["Loaded" story](https://google.github.io/site-kit-wp/storybook/pull/3214/?path=/story/adsense-module-components-dashboard-summary-widget--loaded) for this component is broken at the moment as well.

assets/js/modules/adsense/components/dashboard/DashboardSummaryWidget.js
```diff
@@ -146,6 +146,8 @@ function DashboardSummaryWidget( { Widget, WidgetReportZero, WidgetReportError }
 						title={ __( 'RPM', 'google-site-kit' ) }
 						datapoint={ period.totals[ 1 ] }
 						datapointUnit={ currencyCode }
+						change={ previousPeriod.totals[ 1 ] - period.totals[ 1 ] }
```
aaemnnosttv, line 149: Should it not be `current - previous` ? That is [how you had it originally with the utility function](https://github.com/google/site-kit-wp/pull/3214/files/faaebdf7181d8f931af21c0140b52f854a793a46..9f2a3316950ec4da7faba99ca79dc579e71ed822#diff-5cc841e545f161a83da5721c030f752769179ee5ffec24803b63cbae17e947c0L50-L51).

assets/js/modules/adsense/components/dashboard/DashboardSummaryWidget.js
```diff
@@ -124,7 +124,7 @@ function DashboardSummaryWidget( { Widget, WidgetReportZero, WidgetReportError }
 		);
 	}
 
-	if ( isZeroReport( today ) && isZeroReport( period ) && isZeroReport( daily ) ) {
+	if ( isZeroReport( period ) && isZeroReport( daily ) ) {
```
aaemnnosttv, line 127: We need to include the previous period in this check as well since it should only show the report zero component when both current and previous periods are zero data, not just the current. It looks like this was missing before but would be good to fix now.

aaemnnosttv (commented):
Thanks @johnPhillips – a few comments left for you.

Also, the story is still broken 😄  
image
https://google.github.io/site-kit-wp/storybook/pull/3214/?path=/story/adsense-module-components-dashboard-summary-widget--loaded

johnPhillips:
@aaemnnosttv ah, sorry about the story, I missed that. Hopefully fixed now 👍
Also addressed the bad arithmetic and added the `isZeroReport`.

assets/js/modules/adsense/components/dashboard/DashboardSummaryWidget.js
```diff
@@ -189,6 +191,7 @@ function DashboardSummaryWidget( { Widget, WidgetReportZero, WidgetReportError }
 						className="overview-adsense-impressions"
 						title={ __( 'Ad Impressions', 'google-site-kit' ) }
 						datapoint={ period.totals[ 2 ] }
+						change={ period.totals[ 2 ] - previousPeriod.totals[ 2 ] }
```
aaemnnosttv, line 194: I think we also need a `changeDataUnit` here, otherwise this looks inconsistent with other numbers that are formatted: 
image

Unfortunately, we don't have a shorthand option for `DataBlock` to use the default "metric" style (which is what we'd want here I think). Passing an empty object would work but that would also be a bit too cryptic for my taste.

I think the easiest thing we can do is pass `changeDataUnit` as a boolean but leave it empty to use the default (`true`) which will also result in the default format, which I think makes more sense than adding something else to mean "use the default".
```suggestion
						change={ period.totals[ 2 ] - previousPeriod.totals[ 2 ] }
						changeDataUnit
```
How does that sound @felixarntz ?

aaemnnosttv (commented):
Thanks @johnPhillips – just one last thing I think, then this should be good to go.

assets/js/modules/adsense/components/dashboard/DashboardSummaryWidget.js
felixarntz: @aaemnnosttv Sounds good to me!

aaemnnosttv (approved):

Issue #3171, opened 2021-04-15
Implement Analytics datastore infrastructure to control primary property type in GA4-transitional flow


felixarntz:
Related to #3170 (learn more about the different setup flows between UA and GA4 there), if the "ga4-transitional" setup flow mode is active, we will eventually have a single dropdown which lists _all_ Analytics properties (i.e. including both UA and GA4, see also #3167). In addition to that, there will also be a dropdown to choose the respective other one (e.g. if you choose a GA4 property in the overall dropdown, a separate UA property dropdown would appear, and vice-versa).

In order to know which property should be selected in the overall dropdown, we need to know whether to "give precedence" to the current GA4 or UA property. This is only relevant on the client-side, to not alter whether we show the UA property or GA4 property in the dropdown other than what the user originally selected there. From a data perspective, it doesn't matter at all from which point we look at it, and this selection therefore should not be persisted in any way.

In other words, this infrastructure is only relevant when the "ga4-transitional" setup flow mode is used. Otherwise, the user would only have UA properties _or_ GA4 properties anyway, but not both types.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `modules/analytics` datastore should have a new selector `getPrimaryPropertyType()`:
    * It should either return “ua” or “ga4” (eventually based on what type of property is selected in the primary dropdown for the "ga4-transitional" flow), relying on a value stored in state.
    * The initial/default value should be “ua”.
* To modify the result for the above selector, the `modules/analytics` datastore should also have an action `setPrimaryPropertyType( type )`:
    * It should only allow “ua” or “ga4” as possible values for `type`.
    * It should modify the value stored in state.

## Implementation Brief
* Using `assets/js/modules/analytics/datastore/properties.js`,
	* Add a new state variable `primaryPropertyType` having the default value: `ua`
	* Add a new action `setPrimaryPropertyType` which has `type` as parameter. The latter can either be `ua` or `ga4`.
	* The `setPrimaryPropertyType` action should update the state variable to the `type` value.

### Test Coverage
* Add test for the `setPrimaryPropertyType` action.

### Visual Regression Changes
* N/A.

## QA Brief

Check that the code has been implemented as the AC describes

## Changelog entry

* Implement Analytics datastore infrastructure to control primary property type in GA4-transitional flow

tofumatt:
IB ✅

danielgent:
[...]

PR #3219, opened 2021-04-23
Enhancement/3171 analytics datastore infrastructure ga4

danielgent:
## Summary

Addresses issue #3171

## Checklist

### DIFF
--- a/assets/js/modules/analytics/datastore/properties.js
+++ b/assets/js/modules/analytics/datastore/properties.js
@@ -366,6 +366,18 @@ const baseSelectors = {
 	},
 
 	/**
+	 * Gets the primary property type.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {string} "ua" or "ga4".
+	 */
+	getPrimaryPropertyType( state ) {
+		return state.primaryPropertyType;
+	},
+
+	/**
 	 * Gets the matched property, if any.
 	 *
 	 * @since 1.8.0
--- a/assets/js/modules/analytics/datastore/properties.js
+++ b/assets/js/modules/analytics/datastore/properties.js
@@ -93,11 +93,13 @@ const RECEIVE_MATCHED_PROPERTY = 'RECEIVE_MATCHED_PROPERTY';
 const RECEIVE_GET_PROPERTIES = 'RECEIVE_GET_PROPERTIES';
 const RECEIVE_PROPERTIES_PROFILES_COMPLETION = 'RECEIVE_PROPERTIES_PROFILES_COMPLETION';
 const WAIT_FOR_PROPERTIES = 'WAIT_FOR_PROPERTIES';
+const SET_PRIMARY_PROPERTY_TYPE = 'SET_PRIMARY_PROPERTY_TYPE';
 
 const baseInitialState = {
 	properties: {},
 	isAwaitingPropertiesProfilesCompletion: {},
 	matchedProperty: undefined,
+	primaryPropertyType: 'ua',
 };
 
 const baseActions = {
@@ -216,6 +218,15 @@ const baseActions = {
 			type: WAIT_FOR_PROPERTIES,
 		};
 	},
+
+	setPrimaryPropertyType( primaryPropertyType ) {
+		invariant( [ 'ua', 'ga4' ].includes( primaryPropertyType ), 'type must be "ua" or "ga4"' );
+
+		return {
+			payload: { primaryPropertyType },
+			type: SET_PRIMARY_PROPERTY_TYPE,
+		};
+	},
 };
 
 const baseControls = {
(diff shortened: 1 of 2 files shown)


assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -281,5 +304,10 @@ describe( 'modules/analytics properties', () => {
 				expect( foundProperty ).toEqual( undefined );
 			} );
 		} );
+		describe( 'getPrimaryPropertyType', () => {
```
eugene-manuilov, line 307: We also need to add a test (one is enough) that checks that this selector returns the correct type from the state object after it has been changed.

assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -281,5 +304,10 @@ describe( 'modules/analytics properties', () => {
...
+			it( 'returns the primary property type from the store', () => {
+				expect( registry.select( STORE_NAME ).getPrimaryPropertyType( ) ).toEqual( 'ua' );
```
eugene-manuilov, line 309: @danielgent, please, use `.toBe( ... )` for primitive types instead of `.toEqual( ... )` which is intended for complex types like objects.

eugene-manuilov (changes requested):
Nice work, @danielgent! There are a few comments regarding your tests. Please, take a look at it and make the appropriate changes.

assets/js/modules/analytics/datastore/properties.test.js
eugene-manuilov: The same applies to other occurrences of `.toEqual()` in your tests.

assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -165,6 +165,29 @@ describe( 'modules/analytics properties', () => {
...
+
+		describe( 'setPrimaryPropertyType', () => {
+			it( 'can only be "ga" or "ua4"', () => {
+				registry.dispatch( STORE_NAME ).setPrimaryPropertyType( 'ua' );
+				registry.dispatch( STORE_NAME ).setPrimaryPropertyType( 'ga4' );
+
+				expect( () => {
+					registry.dispatch( STORE_NAME ).setPrimaryPropertyType( 'foo-bar' );
+				} ).toThrow( 'type must be "ua" or "ga4"' );
+			} );
```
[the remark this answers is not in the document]
danielgent: @eugene-manuilov *much better!*.  I always forget the `.each` syntax :facepalm: .  So much more readable

assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -165,6 +165,29 @@ describe( 'modules/analytics properties', () => {
...
+			it( 'can be set and read', () => {
+				expect( registry.select( STORE_NAME ).getPrimaryPropertyType( ) ).toEqual( 'ua' );
+
+				registry.dispatch( STORE_NAME ).setPrimaryPropertyType( 'ga4' );
+
+				expect( registry.select( STORE_NAME ).getPrimaryPropertyType( ) ).toEqual( 'ga4' );
+
+				registry.dispatch( STORE_NAME ).setPrimaryPropertyType( 'ua' );
+
+				expect( registry.select( STORE_NAME ).getPrimaryPropertyType( ) ).toEqual( 'ua' );
```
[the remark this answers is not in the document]
danielgent: Yes @eugene-manuilov !  Very good point.  Better to check the state directly and then there's a different test for the selector below :+1:

assets/js/modules/analytics/datastore/properties.test.js
danielgent: @eugene-manuilov  Sounds great. I wasn't sure what to put here because `setPrimaryPropertyType` was testing everything above (the test felt redundant), but (as per your suggestion) I'll change the above test to check the state directly

And then this test can check the selector reads the state after it has changed :+1:

assets/js/modules/analytics/datastore/properties.test.js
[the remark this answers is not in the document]
danielgent: @eugene-manuilov  I'm changing to use `.each` as requested *but* in this case I did actually want a test that tests that the value can be toggled back and forth between the two values

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -216,6 +218,15 @@ const baseActions = {
 			type: WAIT_FOR_PROPERTIES,
 		};
 	},
+
+	setPrimaryPropertyType( primaryPropertyType ) {
```
eugene-manuilov, line 222: @danielgent, please, add jsdoc to this action.

eugene-manuilov (changes requested):
Amazing! Just add one jsdoc for the action function and we are good to go.

eugene-manuilov (approved):

Issue #3166, opened 2021-04-15
Implement GA4 PropertySelect component in JS


felixarntz:
There should be a GA4 version of the existing UA `PropertySelect` component, which covers the same functionality for GA4 properties.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a new GA4 `PropertySelect` component in `assets/js/modules/analytics-4/components/common/PropertySelect.js` which at a high level does the same that the existing Analytics (UA) `PropertySelect` component does, just for GA4 properties instead.
    * It should rely on the `modules/analytics-4` datastore:
        * to list all GA4 properties for the currently selected account
        * to select a GA4 property in the datastore when the user changes the selection
    * It should visually look similar, and it should display the property ID after the property name in parentheses (e.g. "My Property (12345678)" (see related #3164 which does this in the UA `PropertySelect` component).
    * It should show a loading state in the similar conditions the existing component does.
    * It should also send the same tracking event on property change.
    * Other than the existing component, it should not deal with an existing tag for now, i.e. ignore that part.
* There should be Storybook coverage for the new component.

## Implementation Brief
* Create file `assets/js/modules/analytics-4/components/common/PropertySelect.js` which exports the `PropertySelect` functional component.
* The component should take most of the code from `assets/js/modules/analytics/components/common/PropertySelect.js` with the following changes:
	* It should query the `modules/analytics-4` data store instead of `modules/analytics`.
	* Disregard all references to `hasExistingTag` which is not in scope for now.
* Add stories for the newly added `PropertySelect` component in `assets/js/modules/analytics-4/components/common/PropertySelect.stories.js`.

### Test Coverage
* Add tests for the `PropertySelect` component.

### Visual Regression Changes
* N/A

## QA Brief
Note: this component isn't currently rendered anywhere except in storybook, so the ACs will need to be verified from looking at the code, tests and story.
* Verify that the ACs above have been met.
* There should be a story file at <del>`assets/js/modules/analytics-4/components/common/PropertySelect.stories.js`</del> (see: https://github.com/google/site-kit-wp/pull/3221#discussion_r629584884) `stories/module-analytics-4.stories.js`.
* There should be test coverage for the new component.

## Changelog entry

* N/A

tofumatt:
IB ✅

johnPhillips:
Hi @asvinb / @tofumatt 
I haven’t dealt with GA4 yet, so I just had a couple of questions around this:

[...]

PR #3221, opened 2021-04-26
 Implement GA4 PropertySelect component in JS (3166).

johnPhillips:
## Summary

Addresses issue #3166

## Checklist


assets/js/modules/analytics-4/components/common/PropertySelect.js
```diff
@@ -0,0 +1,99 @@
...
+	} = useSelect( ( select ) => {
+		const data = {
+			accountID: select( MODULES_ANALYTICS ).getAccountID(),
+			properties: [],
+			isResolvingProperties: false,
+		};
+
+		if ( data.accountID ) {
+			data.properties = select( STORE_NAME ).getProperties( data.accountID );
+			data.isResolvingProperties = select( STORE_NAME ).isResolving( 'getProperties', [ data.accountID ] );
```
eugene-manuilov, line 51: I think we need to use `hasFinishedResolution` instead of `isResolving`. At least this is what we use across the code base to determine whether something has been resolved or not.

assets/js/modules/analytics-4/components/common/PropertySelect.js
```diff
@@ -0,0 +1,99 @@
...
+
+	const propertyID = useSelect( ( select ) => select( STORE_NAME ).getPropertyID() );
+	const hasResolvedAccounts = useSelect( ( select ) => select( MODULES_ANALYTICS ).hasFinishedResolution( 'getAccounts' ) );
+
+	const { selectProperty } = useDispatch( STORE_NAME );
+	const onChange = useCallback( ( index, item ) => {
+		const newPropertyID = item.dataset.value;
+		if ( propertyID !== newPropertyID ) {
+			selectProperty( newPropertyID, item.dataset.internalId ); // eslint-disable-line sitekit/acronym-case
+			trackEvent( 'analytics_setup', 'property_change', newPropertyID );
```
eugene-manuilov, line 65: Not sure that we need it for ga4. I believe we will have a separate ticket for ga4 tracking.

assets/js/modules/analytics-4/components/common/PropertySelect.js
johnPhillips: Yeah I've not seen this before. It did come from the analytics module's `PropertySelect` component though...

https://github.com/google/site-kit-wp/blob/11b7cff0245e1eae5e3d2ffb085cdb0fceabeef5/assets/js/modules/analytics/components/common/PropertySelect.js#L51

assets/js/modules/analytics-4/components/common/PropertySelect.js
johnPhillips: Good spot, but I double checked the ACs and I think we are good to keep this.

assets/js/modules/analytics-4/components/common/PropertySelect.js
```diff
@@ -0,0 +1,102 @@
...
+import { __, sprintf } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import Data from 'googlesitekit-data';
+import { Select, Option } from '../../../../material-components';
+import ProgressBar from '../../../../components/ProgressBar';
+import { STORE_NAME } from '../../datastore/constants';
+import { MODULES_ANALYTICS, PROPERTY_CREATE } from '../../../analytics/datastore/constants';
```
eugene-manuilov, line 32: We need to import the `PROPERTY_CREATE` constant from the `analytics-4` module constants.

assets/js/modules/analytics-4/components/common/PropertySelect.js
eugene-manuilov: Let's use `hasFinishedResolution` since we primarily use it right now.

assets/js/modules/analytics-4/components/common/PropertySelect.js
```diff
@@ -0,0 +1,102 @@
...
+	const hasResolvedAccounts = useSelect( ( select ) => select( MODULES_ANALYTICS ).hasFinishedResolution( 'getAccounts' ) );
+
+	const { selectProperty } = useDispatch( STORE_NAME );
+	const onChange = useCallback( ( index, item ) => {
+		const newPropertyID = item.dataset.value;
+		if ( propertyID !== newPropertyID ) {
+			selectProperty( newPropertyID );
+			trackEvent( 'analytics_setup', 'property_change', newPropertyID );
+		}
+	}, [ propertyID ] );
```
eugene-manuilov, line 67: We also need to add `selectProperty` to the dependencies list here.

assets/js/modules/analytics-4/components/common/PropertySelect.js
```diff
@@ -0,0 +1,102 @@
...
+				} )
+				.map( ( { _id, displayName }, index ) => (
+					<Option
+						key={ index }
+						value={ _id }
+						data-internal-id={ _id }
+					>
+						{
+							/* translators: 1: Property name. 2: Property ID. */
+							sprintf( __( '%1$s (%2$s)', 'google-site-kit' ), displayName, _id )
```
eugene-manuilov, line 96: Let's make sure we don't display `_id` if it equals to the `PROPERTY_CREATE` constant.

eugene-manuilov (changes requested):
Thanks, @johnPhillips. It looks good to me in general. Found a few issues that we need to address before we can merge this PR. Please, take a look at it.

assets/js/modules/analytics-4/components/common/PropertySelect.js
johnPhillips: Good catch 👍

assets/js/modules/analytics-4/components/common/PropertySelect.js
johnPhillips: Things seem to break when I do that 🤔

assets/js/modules/analytics-4/components/common/PropertySelect.js
```diff
@@ -0,0 +1,102 @@
...
+const { useSelect, useDispatch } = Data;
+
+export default function PropertySelect() {
+	const {
+		accountID,
+		properties,
+		isResolvingProperties,
+	} = useSelect( ( select ) => {
+		const data = {
+			accountID: select( MODULES_ANALYTICS ).getAccountID(),
```
felixarntz, line 44: Maybe would be good to add a comment here to indicate that this should eventually be stored in the GA4 module itself (i.e. in `modules/analytics-4`), to decouple from the old Analytics/GA module.

eugene-manuilov (approved):
LGTM

felixarntz (dismissed):
Mostly this is looking good, I have two follow-up concerns.

assets/js/modules/analytics-4/components/common/PropertySelect.stories.js
```diff
@@ -0,0 +1,80 @@
...
+				<div className="googlesitekit-setup-module">
+					{ children }
+				</div>
+			</section>
+		</div>
+	);
+}
+
+storiesOf( 'GA4', module )
+	.add( 'PropertySelect', () => {
```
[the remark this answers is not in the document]
tofumatt: I agree… I think adding these to existing stories is the best approach. Assuming there are any where it'd make sense to add them.

I'd also settle for a comment that marks them as exactly what you said above:

```js
// @TODO: Use the new Component Story Format for these stories.
// These stories are not written using our new Storybook best
// practices, so do not reference them for new stories.
```

If anything, a comment is even more explicit, and the aim is to make sure no one gets confused 🙂

aaemnnosttv (commented):
Left a few comments in addition to others' but nothing major.

assets/js/modules/analytics-4/components/common/PropertySelect.js
johnPhillips: > I'm guessing you probably tried replacing isResolving for hasFinishedResolution

I'm unable to recreate the issue now so I think that must have been what was I was doing before, sorry. 🤦‍♂️ 
I've gone with your suggestion below as well Evan, regarding using a single `isLoading` variable. Thanks!

assets/js/modules/analytics-4/components/common/PropertySelect.stories.js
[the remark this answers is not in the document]
johnPhillips: I've moved this back into a `stories/module-analytics-4.stories.js` file for now. I was unaware of our plans to use the new CSF convention and it wasn't mentioned in the IB (although I should have guessed changes were afoot from the new story location). Sorry about that.
Hope this is ok for now, and I am more than happy to refactor once the new approach is established.

assets/js/modules/analytics-4/components/common/PropertySelect.js
```diff
@@ -0,0 +1,103 @@
...
+				} )
+				.map( ( { _id, displayName }, index ) => (
+					<Option
+						key={ index }
+						value={ _id }
+						data-internal-id={ _id }
+					>
+						{
+							/* translators: 1: Property name. 2: Property ID. */
+							_id === PROPERTY_CREATE ? displayName : sprintf( __( '%1$s (%2$s)', 'google-site-kit' ), displayName, _id )
```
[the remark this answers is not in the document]
johnPhillips: I was getting pushback from eslint when I tried to do this. It doesn't seem to cope with the translators comment being put in a ternary, hence the long line. I've now pulled it out into a simple function - hope that's ok. Maybe I missed an easier way to do this?

assets/js/modules/analytics-4/components/common/PropertySelect.js
```diff
@@ -0,0 +1,103 @@
...
+import { isValidAccountID } from '../../../analytics/util';
+import { trackEvent } from '../../../../util';
+const { useSelect, useDispatch } = Data;
+
+export default function PropertySelect() {
+	const {
+		accountID,
+		properties,
+		isResolvingProperties,
+	} = useSelect( ( select ) => {
```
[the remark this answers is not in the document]
johnPhillips: This actually helped a lot. The IB suggested using most of the code from the existing analytics component so I went with that, but I think the older approaches in that file were a bit confusing for me.

assets/js/modules/analytics-4/components/common/PropertySelect.test.js
```diff
@@ -0,0 +1,122 @@
...
+		// Note: we do length + 1 here because there should also be an item for
+		// "Set up a new property".
+		expect( listItems ).toHaveLength( properties.length + 1 );
+	} );
+
+	it( 'should be disabled when in the absence of an valid account ID.', async () => {
+		muteFetch( /^\/google-site-kit\/v1\/modules\/analytics-4\/data\/settings/, [] );
+		const { container, registry } = render( <PropertySelect />, {
+			setupRegistry( { dispatch } ) {
+				setupRegistry( { dispatch } );
```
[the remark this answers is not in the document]
johnPhillips: This was shamelessly borrowed from the existing Analytics component's tests 🙊 
I agree, it needs a couple of reads. I've had a go at a refactor to make it easier to understand, let me know what you think.

assets/js/modules/analytics-4/components/common/PropertySelect.js
```diff
@@ -0,0 +1,97 @@
...
+				} )
+				.map( ( { _id, displayName }, index ) => (
+					<Option
+						key={ index }
+						value={ _id }
+						data-internal-id={ _id }
+					>
+						{
+							getPropertyName( displayName, _id )
+						}
```
aaemnnosttv, line 92: Looks like this would work instead
```suggestion
						{ _id === PROPERTY_CREATE
							? displayName
							: sprintf(
								/* translators: 1: Property name. 2: Property ID. */
								__( '%1$s (%2$s)', 'google-site-kit' ),
								displayName,
								_id
							)
						}
```

assets/js/modules/analytics-4/components/common/PropertySelect.js
```diff
@@ -0,0 +1,97 @@
...
+			{ ( properties || [] )
+				.concat( {
+					_id: PROPERTY_CREATE,
+					displayName: __( 'Set up a new property', 'google-site-kit' ),
+				} )
+				.map( ( { _id, displayName }, index ) => (
+					<Option
+						key={ index }
+						value={ _id }
+						data-internal-id={ _id }
```
aaemnnosttv, line 88: This prop is only relevant for UA properties; GA4 properties don't have a different internal ID.
```suggestion
```

assets/js/modules/analytics-4/components/common/PropertySelect.js
```diff
@@ -0,0 +1,97 @@
...
+
+export default function PropertySelect() {
+	// TODO: Update this select hook to pull accountID from the modules/analytics-4 datastore when GA4 module becomes separated from the Analytics one
+	const accountID = useSelect( ( select ) => select( MODULES_ANALYTICS ).getAccountID() );
+	const properties = useSelect( ( select ) => select( STORE_NAME ).getProperties( accountID ) );
+	const propertyID = useSelect( ( select ) => select( STORE_NAME ).getPropertyID() );
+
+	const isLoading = useSelect( ( select ) => {
+		return ! select( MODULES_ANALYTICS ).hasFinishedResolution( 'getAccounts' ) ||
+			select( STORE_NAME ).isResolving( 'getProperties', [ accountID ] );
```
aaemnnosttv, line 45: I realize the previous is what is used in the `ProfileSelect` but I think we should probably use `hasFinishedResolution` for consistency, but also this is usually what we want.
```suggestion
			! select( STORE_NAME ).hasFinishedResolution( 'getProperties', [ accountID ] );
```

stories/module-analytics-4.stories.js
```diff
@@ -0,0 +1,80 @@
...
+				<div className="googlesitekit-setup-module">
+					{ children }
+				</div>
+			</section>
+		</div>
+	);
+}
+
+storiesOf( 'Analytics-4 Module', module )
+	.add( 'Account Property Select', () => {
```
aaemnnosttv, line 46:
```suggestion
	.add( 'PropertySelect', () => {
```

assets/js/modules/analytics-4/components/common/PropertySelect.test.js
```diff
@@ -0,0 +1,124 @@
...
+	dispatch( MODULES_ANALYTICS ).finishResolution( 'getAccounts', [] );
+
+	dispatch( STORE_NAME ).receiveGetProperties( [], { accountID } );
+	dispatch( STORE_NAME ).finishResolution( 'getProperties', [ accountID ] );
+};
+
+const setupRegistryWithNoAccountID = ( registry ) => {
+	setupRegistry( registry );
+	registry.dispatch( STORE_NAME ).finishResolution( 'getProperties', [ ACCOUNT_CREATE ] );
+};
```
aaemnnosttv, line 69: This is a bit confusing since `setupRegistry` will set an `accountID` 🤔

aaemnnosttv (commented):
Thanks @johnPhillips – this is looking very close, just a few things left to tweak then I think we'll be good to go.

assets/js/modules/analytics-4/components/common/PropertySelect.js
johnPhillips: Ah nice 💡

assets/js/modules/analytics-4/components/common/PropertySelect.test.js
johnPhillips: Looking at the original test, I think the intent was: 
- check it renders enabled when there is a valid account ID
- dispatch an action to set account ID to `ACCOUNT_CREATE`, which is a valid selection but an invalid ID
- check it now renders disabled.

https://github.com/google/site-kit-wp/blob/11b7cff0245e1eae5e3d2ffb085cdb0fceabeef5/assets/js/modules/analytics/components/common/PropertySelect.test.js#L106-L116

Perhaps there is a simpler way to go about it, I agree it doesn't read well. Happy to do as you advise.

On a related note: looking into this again, I now realise that the slightly unconventional use of `useSelect` [in the original component ](https://github.com/google/site-kit-wp/blob/11b7cff0245e1eae5e3d2ffb085cdb0fceabeef5/assets/js/modules/analytics/components/common/PropertySelect.js#L42)must've have been to do with this requirement because we only want to get properties etc if there is actually a valid accountID.
I've gone back and refactored this, but tried to keep it more in line with how we currently use `useSelect`. I've also added a story for this scenario as it wasn't immediately obvious to me.

It might be an idea to have a call and discuss if I'm still wide of the mark with all of this. In hindsight this was probably the wrong issue to pick as my first adventure with GA4 😂

johnPhillips: Just a note here about our slack conversation. I'm going to try to add this guard against no accountID in the resolver like in Analytics - maybe we can clear this up a bit.

johnPhillips: I've added a guard in the `getProperties` resolver as per our conversation on Slack, so was able to tidy up the use of `useSelect`, thank you for that.

aaemnnosttv (commented):
Looking good! Left a few last comments but I think we're almost there 😄

assets/js/modules/analytics-4/components/common/PropertySelect.test.js
```diff
@@ -0,0 +1,124 @@
...
+
+describe( 'PropertySelect', () => {
+	it( 'should render an option for each analytics property of the currently selected account.', async () => {
+		muteFetch( /^\/google-site-kit\/v1\/modules\/analytics-4\/data\/settings/, [] );
```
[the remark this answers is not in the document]
johnPhillips: If I remove the `muteFetch` the tests start failing with e.g.
```console
["Google Site Kit API Error", "method:GET", "datapoint:settings", 
"type:modules", "identifier:analytics-4", "error:\"window.fetch is not a function\""]
```
So things are indeed calling `fetch` under the hood somewhere to get at the `settings` endpoint. I'm not sure where this is happening. Is there something I'm doing wrong in `setupRegistry` perhaps?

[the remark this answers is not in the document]
aaemnnosttv: Ah that helps. It's because you're only receiving settings for `modules/analytics` not GA4 but it uses both so it's trying to fetch the settings for the GA4 module.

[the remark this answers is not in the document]
johnPhillips: Bingo. Thanks - I've removed `muteFetch` now 👍

stories/module-analytics-4.stories.js
```diff
@@ -0,0 +1,98 @@
...
+			dispatch( MODULES_ANALYTICS ).receiveGetSettings( {
+				accountID,
+				propertyID,
+			} );
+			dispatch( MODULES_ANALYTICS ).receiveGetAccounts( [ {
+				id: properties[ 0 ]._accountID,
+				kind: 'analytics#account',
+				name: 'GA4 Account',
+			} ] );
+			dispatch( MODULES_ANALYTICS ).finishResolution( 'getAccounts', [] );
```
aaemnnosttv, line 84: We shouldn't need to use `finishResolution` any time we're pre-receiving the related data (as above).
```suggestion
```

stories/module-analytics-4.stories.js
```diff
@@ -0,0 +1,98 @@
...
+			webDataStreams,
+		} = fixtures;
+		const accountID = createProperty._accountID;
+		const propertyID = createWebDataStream._propertyID;
+
+		const setupRegistry = ( { dispatch } ) => {
+			dispatch( MODULES_ANALYTICS ).receiveGetSettings( {
+				accountID,
+				propertyID,
+			} );
```
aaemnnosttv, line 78: It seems like the story here is intended to show a selected property, but it does not. It may be because the settings set here are for classic Analytics but I'm not entirely sure. We should also make sure that the `propertyID` which we're setting is available in the `properties` we're receiving otherwise it will also not be able to select it.

assets/js/modules/index.test.js
```diff
@@ -35,7 +35,7 @@ function directories( relativePath ) {
 
 function getComponentNames( componentPath ) {
 	return fs.readdirSync( componentPath )
-		.filter( ( name ) => ! /^index|\.test\.js$/.test( name ) )
+		.filter( ( name ) => ! /^index|\.test|\.stories\.js$/.test( name ) )
```
aaemnnosttv, line 38: Maybe a little easier to read?
```suggestion
		.filter( ( name ) => ! /^index|\.(test|stories)\.js$/.test( name ) )
```

aaemnnosttv (commented):
Thanks @johnPhillips – just a few comments related to the added stories but otherwise very close.

stories/module-analytics-4.stories.js
johnPhillips: I made the property pre-selection work but it does seem to require this line, otherwise `select( MODULES_ANALYTICS ).hasFinishedResolution( 'getAccounts' )` returns false and it is stuck in a loading state. 🤔

aaemnnosttv (approved):
Great, thanks!

PR #3230, opened 2021-04-27
Enhancement/2940 migrate storybook

asvinb:
## Summary

Addresses issue #2940 

## Relevant technical choices
* Renamed the file `DashboardSearchVisitorsWidget.js` to `DashboardUniqueVisitorsWidget.js` since the latter exported the `DashboardUniqueVisitorsWidget` component.
* `WidgetAreaRenderer` is being used in individual stories decorators as opposed to component decorators. No data flows in if it's the latter.
* Updated the tests for modules to exclude `*.stories.js` files

## Checklist


tests/js/WithRegistrySetup.js
```diff
@@ -0,0 +1,49 @@
...
+ * Internal dependencies
+ */
+import Data from 'googlesitekit-data';
+const { useRegistry } = Data;
+
+const WithRegistrySetup = ( { func, children } ) => {
+	const registry = useRegistry();
+	const [ ready, setReady ] = useState( false );
+
+	useEffect( () => {
```
tofumatt, line 34: Looks like there's a linting error here, probably good to clear up 🙂

tofumatt (changes requested):
Overall this looks awesome, thanks 👍🏻 

A few notes:

1. Looks like there's a linting error for the `useEffect` hook in `WithRegistrySetup`, but should be easy-to-fix.
2. I know this isn't mentioned in the doc or the ACs/IBs, but think we can take this opportunity to differentiate the naming of story states better, starting here. The "Loaded" and "Loading" states are very different but their names are only a few letters apart; it's really hard to visually distinguish them in the list:

Screenshot 2021-04-28 at 22 47 44

Why don't we try something like `"Ready"` or `"Available"` instead of `"Loaded"` as the prefix for those stories? I think that'd make it way easier to read.

Thoughts @felixarntz, @aaemnnosttv, @eugene-manuilov?

tests/js/WithRegistrySetup.js
```diff
@@ -0,0 +1,49 @@
...
+		if ( func && typeof func === 'function' ) {
+			func( registry );
+
+			setReady( true );
```
ivankruchkoff, line 38: func is optional and we're calling it as such `<WithRegistrySetup func={ args?.setupRegistry }>` but without a supplied func, we will still need to `setReady(true)`.

tests/js/WithRegistrySetup.js
asvinb: @tofumatt Originally there was a linting error which I fixed. Do you still see the error? All the tests are passing for me as well as the Github checks.

tests/js/WithRegistrySetup.js
asvinb: Good catch @ivankruchkoff . Updated.

eugene-manuilov:
> Why don't we try something like "Ready" or "Available" instead of "Loaded" as the prefix for those stories? I think that'd make it way easier to read.

@tofumatt sounds good to me. I think `Ready` is the best option for it.

.storybook/polyfill-globals.js
```diff
@@ -0,0 +1,3 @@
+import { resetGlobals } from './utils/resetGlobals';
```
aaemnnosttv, line 1: Please add a file header ☝️

tests/js/WithRegistrySetup.js
aaemnnosttv: This should probably use `useMount` instead of `useEffect` like we do in other places now when an effect should only run once. I think the linting error here is from the missing dependency array though below.

tests/js/WithRegistrySetup.js
```diff
@@ -0,0 +1,49 @@
...
+const { useRegistry } = Data;
+
+const WithRegistrySetup = ( { func, children } ) => {
+	const registry = useRegistry();
+	const [ ready, setReady ] = useState( false );
+
+	useEffect( () => {
+		if ( func && typeof func === 'function' ) {
+			func( registry );
+		}
```
aaemnnosttv, line 37: Let's move the current condition here to `invariant` so that it fails if `func` isn't passed or is a non-function.

The only condition really needed for invoking `func` is `ready === false`.

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.stories.js
```diff
@@ -0,0 +1,215 @@
...
+		},
+	],
+};
+const optionsCompareEntityURL = {
+	...optionsCompare,
+	url: currentEntityURL,
+};
+const WidgetAreaTemplate = ( args ) => {
+	return (
+		<WithRegistrySetup func={ args?.setupRegistry }>
```
aaemnnosttv, line 71: Since `func` should always be required, if there is a story that doesn't need it, we could provide an empty function here as a fallback. As it is though, every story has the setup function defined.

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.stories.js
```diff
@@ -0,0 +1,215 @@
...
+		{
+			expression: 'ga:users',
+			alias: 'Total Users',
+		},
+	],
+};
+const optionsCompareEntityURL = {
+	...optionsCompare,
+	url: currentEntityURL,
+};
```
aaemnnosttv, line 68: This looks like it would be too easy for the story to get out of sync with the component like this. Maybe we can expose the base/default options from the component as a static property or export? The dates should be handled automatically though by setting the reference date. Same with the entity URL.

We'd done something like this before in a similar situation by exposing select functions on the component but it ended up being replaced with widget state. In this case, maybe we can expose functions for selecting the report args which would be used in the component as well. 🤔 

I will see about trying this on another branch.

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.stories.js
```diff
@@ -0,0 +1,215 @@
...
+
+				dispatch( CORE_WIDGETS ).registerWidgetArea( areaName, {
+					title: 'Overview',
+					style: WIDGET_AREA_STYLES.BOXES,
+				} );
+				dispatch( CORE_WIDGETS ).registerWidget( widgetSlug, {
+					Component: DashboardUniqueVisitorsWidget,
+					width: WIDGET_WIDTHS.FULL,
+				} );
+				dispatch( CORE_WIDGETS ).assignWidget( widgetSlug, areaName );
```
aaemnnosttv, line 198: This part looks like it will be a common need for most widget stories. I'm wondering why we use a widget area here at all rather than just rendering the widget component directly 🤔 

If we really need it, this part (along with maybe the Template component & registration) would be great to extract into something more reusable (maybe from the widget area stories themselves?).

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.stories.js
```diff
@@ -0,0 +1,215 @@
...
+				dispatch( CORE_USER ).setReferenceDate( '2020-09-08' );
+
+				provideModules( registry, [ {
+					active: true,
+					connected: true,
+					slug: 'analytics',
+				} ] );
```
aaemnnosttv, line 205: If we weren't rendering a widget area (+widget renderer internally), this part probably wouldn't be needed either.

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.stories.js
```diff
@@ -0,0 +1,215 @@
...
+ErrorEntityURL.args = {
+	setupRegistry: ( registry ) => {
+		const { dispatch } = registry;
+		const error = {
+			code: 'missing_required_param',
+			message: 'Request parameter is empty: metrics.',
+			data: {},
+		};
+
+		provideSiteInfo( registry, { currentEntityURL } );
```
aaemnnosttv, line 167: This line looks like it belongs in the global decorator, unless there's something about it which is specific to the error case here.

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.stories.js
```diff
@@ -0,0 +1,215 @@
...
+		dispatch( STORE_NAME ).receiveError( error, 'getReport', [ options ] );
+		dispatch( STORE_NAME ).finishResolution( 'getReport', [ options ] );
+	},
+};
+
+export const LoadedEntityURL = WidgetAreaTemplate.bind();
+LoadedEntityURL.storyName = 'Loaded with entity URL set';
+LoadedEntityURL.args = {
+	setupRegistry: ( registry ) => {
+		const { dispatch } = registry;
```
aaemnnosttv, line 123: The destructuring makes it a bit cleaner when the full `registry` isn't needed, but when it is it's probably better to call `registry.dispatch` below and modify the formatting of those lines to be spread across multiple as they're already a bit too long as-is.

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.stories.js
```diff
@@ -0,0 +1,215 @@
...
+
+export default {
+	title: 'Modules/Analytics/Widgets/DashboardUniqueVisitorsWidget',
+	decorators: [
+		( Story ) => {
+			const setupRegistry = ( registry ) => {
+				const { dispatch } = registry;
+
+				const [ property ] = accountsPropertiesProfiles.properties;
+				registry.dispatch( STORE_NAME ).receiveGetSettings( {
```
aaemnnosttv, line 181: Here `dispatch` is destructured, but `registry.dispatch` is also called. Since `registry` is used below, we should just use `registry.` methods on all as mentioned above.

aaemnnosttv (commented):
Thanks @asvinb ! I left a few comments regarding the current implementation. It looks like there is room to make these a bit more composable but not sure if there is a common decorator we can extract for all widgets in the Analytics module. If composition is easy without being overly verbose, we don't need to add many additional layers of abstraction.

I'm going to take a look at a potential solution for helping with the composition here, but this is off to a good start 👍

aaemnnosttv:
@asvinb @tofumatt I pushed a branch to test my ideas about exposing functions to select args so that the underlying component maintains ownership of that so that we don't need to duplicate them in story definitions. Some bits are a bit rough/unfinished (missing docs/etc) but it seems like an improvement. I also added a little utility for generating a mock response and receiving the report for Analytics with the given options to make that a little cleaner as well.

Let me know what you think: https://github.com/google/site-kit-wp/compare/enhancement/2940-migrate-storybook...poc/2940-alternate-widget-storybook

tofumatt:
@aaemnnosttv I like the approach, that seems like a good refactoring 👍🏻 

Let's run with that idea for this—I can see how it'd be useful when we have multiple stories here.

asvinb:
@googlebot I consent.

asvinb:
@googlebot I fixed it.

.storybook/polyfill-globals.js
asvinb: Added @aaemnnosttv

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.stories.js
asvinb: I think we can leave it as is because we know for sure that every stories using the Template will definitely have a function.

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.stories.js
asvinb: The error message/layout will be the same, but it's just making sure it displays correctly when an entity URL is set.

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.stories.js
asvinb: @aaemnnosttv You're right. I've updated the stories to render the widget component directly.

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.stories.js
asvinb: 👍

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.stories.js
```diff
@@ -0,0 +1,167 @@
...
+const Template = ( args ) => {
+	const widgetSlug = 'dashboardUniqueVisitorsWidget';
+
+	return (
+		<WithRegistrySetup func={ args.setupRegistry }>
+			<div className="googlesitekit-widget">
+				<div className="googlesitekit-widget__body">
+					<DashboardUniqueVisitorsWidget
+						WidgetReportError={ withWidgetSlug( widgetSlug )( WidgetReportError ) }
+						WidgetReportZero={ withWidgetSlug( widgetSlug )( WidgetReportZero ) }
```
aaemnnosttv, line 43: Let's wrap `DashboardUniqueVisitorsWidget` with `withWidgetComponentProps` instead which provides all of the expected component props 👍

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.stories.js
```diff
@@ -0,0 +1,167 @@
...
+					/>
+				</div>
+			</div>
+		</WithRegistrySetup>
+	);
+};
+
+export const Ready = Template.bind();
```
aaemnnosttv, line 51: Storybook docs show `.bind` always called with `{}` although it doesn't explain why (the best I could find was [this](https://stackoverflow.com/questions/65054041/how-does-technique-for-making-a-copy-of-a-function-works-in-javascript-storyboo)). This corresponds to the `thisArg` parameter in the function so my guess is that it matters for class-based components. While we don't really use these anymore, there are cases where it is unavoidable (e.g. error boundaries). We also use this in the stories for `Buttons` now as well so we should be consistent and use it everywhere.

.storybook/utils/resetGlobals.js
```diff
@@ -0,0 +1,77 @@
+/**
+ * Storybook utilities.
```
aaemnnosttv, line 2: Let's update this to be specific to the utility since the name implies that it only provides a single utility.

assets/js/googlesitekit/widgets/util/get-widget-component-props.js
```diff
@@ -54,7 +54,7 @@ export const getWidgetComponentProps = memize(
 	}
 );
 
-function withWidgetSlug( widgetSlug ) {
+export const withWidgetSlug = ( widgetSlug ) => {
```
aaemnnosttv, line 57: This change shouldn't be necessary per my other comment.

aaemnnosttv (changes requested):
Thanks @asvinb – this is looking really good I think, just a few things to touch up.

.storybook/polyfill-globals.js
asvinb: Updated @aaemnnosttv  I also updated the year in `main.js`

tests/js/WithRegistrySetup.js
```diff
@@ -0,0 +1,54 @@
...
+const { useRegistry } = Data;
+
+const WithRegistrySetup = ( { func, children } ) => {
+	const registry = useRegistry();
+	const [ ready, setReady ] = useState( false );
+
+	invariant( typeof func === 'function', 'func must be a function.' );
+
+	useMount( () => {
+		func( registry );
```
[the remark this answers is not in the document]
asvinb: 👍

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.stories.js
asvinb: 👍

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.stories.js
asvinb: 👍

.storybook/utils/resetGlobals.js
asvinb: 👍

assets/js/googlesitekit/widgets/util/get-widget-component-props.js
asvinb: Reverted the change @aaemnnosttv

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.stories.js
asvinb: @aaemnnosttv I do see we are using the `withActive` HOC in the `DashboardUniqueVisitorsWidget` component. Thus if the module is not loaded, nothing will be displayed: https://github.com/google/site-kit-wp/blob/d6ea05554e20353df9e770abd193c98b006cdf79/assets/js/util/when-active.js#L57-L60

aaemnnosttv: Ah yes, I forgot about that HOC we use with most widgets 👍

aaemnnosttv:
@googlebot I consent.

aaemnnosttv:
@asvinb I pushed a few commits here to make a few corrections but also to add some important functionality which was missing in the global decorators related to enabled features (which we recently changed) so I think this is just about good to go, but I'd like to get @tofumatt, @felixarntz, and @eugene-manuilov to have a look over as well since this is setting some important precedents.

Basically, `features` can now be passed via Story `parameters` which should be set before the registry is created so that things like idea-hub's conditional registration will be easily controllable per-story.

.storybook/preview.js
```diff
@@ -29,78 +28,47 @@ import '../assets/sass/wpdashboard.scss';
 import '../assets/sass/adminbar.scss';
 import '../assets/sass/admin.scss';
 import './assets/sass/wp-admin.scss';
+// Ensure all globals are setup before any other imports are run.
```
tofumatt, line 31:
```suggestion
// Ensure all globals are set up before any other imports are run.
```

.storybook/preview.js
```diff
@@ -29,78 +28,47 @@ import '../assets/sass/wpdashboard.scss';
...
-			googlesitekit_view_posts_insights: true,
-			googlesitekit_view_dashboard: true,
-			googlesitekit_view_module_details: true,
-			googlesitekit_manage_options: true,
-			googlesitekit_publish_posts: true,
-		},
-	};
-};
-resetGlobals();
+// Decorators run from last added to first.
```
tofumatt, line 40: This one just took my brain an extra second to parse; let's clarify so it makes sense immediately to as many folks as possible—some will prefer the original explanation so no harm in leaving both:

```suggestion
// Decorators run from last added to first. (Eg. In reverse order as listed.)
```

tofumatt (commented):
The new features parameters approach looks good to me. As discussed with @aaemnnosttv I think it'd be good to have an example of a story that relies on the features parameters here, but that's all!

eugene-manuilov (approved):
Going to approve this, however, I can't say that I like the new implementation of the `DashboardUniqueVisitorsWidget` component :thinking:. I understand why it has been changed that way, but IMO we sacrificed readability and maintainability for getting new storybooks working.

From my point of view, we should have separated the component into two: one for fetching data and passing it to another one, and one for rendering data passed from the first component. Then we would easily create stories for the second component to visualize it and would avoid a need to replicate the datastore environment for it.

assets/js/modules/analytics/components/dashboard/DashboardUniqueVisitorsWidget.js
```diff
@@ -172,6 +130,57 @@ function DashboardUniqueVisitorsWidget( { WidgetReportZero, WidgetReportError }
...
+			},
+		],
+		dimensions: [ 'ga:channelGrouping' ],
+		...select( CORE_USER ).getDateRangeDates( {
+			offsetDays: DATE_RANGE_OFFSET,
+			compare: true,
+		} ),
+	};
+};
+
```
felixarntz, line 183: I agree with @eugene-manuilov, we shouldn't change this production code just to get the Storybook story setup simpler. Having exported functions in the same file as the component isn't good practice, so we shouldn't do that.

Let's provide the report options in Storybook manually the same way we do in the legacy versions of the story. This issue is not about changing _that_, it's only about migrating to the new proper way of CSF.

felixarntz (changes requested):
I agree with @eugene-manuilov's assessment, and wanna emphasize here that we shouldn't need to change the component to get the new CSF stories set up. The approach of manually passing the request data for the selector and generated response data based on it should be maintained like we have been doing it in the old stories.

felixarntz:
@aaemnnosttv I just saw your comment in https://github.com/google/site-kit-wp/pull/3230#issuecomment-832484422, I don't think I agree though. We can certainly look into making some of this data more usable, but throwing those functions into the same file as the component and making the component code harder to understand is not worth the benefit of not duplicating the object with the data.

I suggest we keep it simple and focus on only the actual migration of this to CSF, rather than also changing the way we provide data for the selectors which wasn't really in scope here.

package.json
```diff
@@ -153,8 +153,8 @@
     "@babel/plugin-transform-runtime": "^7.10.4",
     "@babel/preset-react": "^7.10.4",
     "@jackfranklin/test-data-bot": "^1.3.0",
-    "@storybook/addon-viewport": "^6.1.21",
-    "@storybook/react": "^6.1.21",
+    "@storybook/addon-viewport": "^6.2.9",
+    "@storybook/react": "^6.2.9",
```
aaemnnosttv, line 157: Not sure if this is necessary or not, but Storybook would get stuck in a HMR loop when starting on the previous version. Technically out of scope here but given the minor change (and problem that it fixes for me) I think it's okay.

aaemnnosttv (commented):
Thanks all – this should be ready for another pass now 😄

assets/js/modules/analytics/components/dashboard/DashboardSearchVisitorsWidget.stories.js
```diff
@@ -0,0 +1,193 @@
...
+];
+const currentEntityURL = 'https://www.example.com/example-page/';
+const reportOptionsWithEntity = reportOptions.map( ( options ) => {
+	return {
+		...options,
+		url: currentEntityURL,
+	};
+} );
+
+const WidgetWithComponentProps = withWidgetComponentProps( 'widget-slug' )( DashboardSearchVisitorsWidget );
```
felixarntz, line 66: Maybe it would be more appropriate for this to be the real widget slug instead? Not really an issue though.

felixarntz (approved):
Excellent! One comment, but nothing blocking.

assets/js/modules/analytics/components/dashboard/DashboardSearchVisitorsWidget.stories.js
aaemnnosttv: My preference is to use dummy values when it isn't necessary/significant. It's also one less place to change if the slug were to change 😄

Issue #3167, opened 2021-04-15
Implement selector to get all UA + GA4 properties


felixarntz:
There should be a selector to get all Analytics properties, i.e. including both UA and GA4. The transitional UI for users that have both UA and GA4 should be implemented as part of the `analytics` module, as such this selector should become part of the `modules/analytics` datastore.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `modules/analytics` store should have a new selector `getPropertiesIncludingGA4( accountID )`.
    * It should rely on the `modules/analytics` selector `getProperties` to get all UA properties for the account.
    * _If_ the `modules/analytics-4` store is available, it should rely on its `getProperties` selector to get all GA4 properties for the account.
    * If both are being combined, it should sort the two lists of property objects (**which have a different format**) together, first by property name, then by property ID (e.g. "Same Name (12345)" should come before "Same Name (UA-12345-6)"). The sorting should happen in a memoized way to not cause performance issues.

## Implementation Brief
* Using `assets/js/modules/analytics/datastore/properties.js`,
	* Add new selector `getPropertiesIncludingGA4`:
		* which relies on the `getProperties` selector to get all UA properties for the account.
		* If the `modules/analytics-4` data store is available, use the `getProperties` selector to get all GA4 properties for the account.
		* Combine the results of the `getProperties` selector for both data stores first by property name, then by property ID (e.g. "Same Name (12345)" should come before "Same Name (UA-12345-6)") and then return the combined list.

### Test Coverage
* Add tests for the new `getPropertiesIncludingGA4` selector.

### Visual Regression Changes
* N/A

## QA Brief

Ensure tests pass and `npm run build` runs without errors.

## Changelog entry

* Implement selector to get all UA + GA4 properties.

tofumatt:
@felixarntz @asvinb I'm not sure about the ACs here, namely:

> The `modules/analytics` store should have a new selector `getAllProperties( accountID )`.

This will result in two very similar selectors, `getProperties( accountID )` and `getAllProperties( accountID )` in the store, which is confusing to read. The distinction between the two isn't evident from their names, and I could even see the potential for a developer to read them as the same thing. I think it's better we go one of two routes:

[...]

PR #3231, opened 2021-04-27
Enhancement/3167 implement selector all properties

danielgent:
## Summary

Addresses issue #3167

## Checklist


assets/js/modules/analytics/datastore/__fixtures__/properties-ga4.json
```diff
@@ -0,0 +1,80 @@
+[
```
danielgent, line 1: Copied from https://github.com/google/site-kit-wp/blob/develop/assets/js/modules/analytics-4/datastore/__fixtures__/properties.json

assets/js/modules/analytics/datastore/__snapshots__/properties.test.js.snap
```diff
@@ -0,0 +1,668 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
danielgent, line 1: I'm not a huge fan of snapshots but this seemed the easiest way of checking the sort order!

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -407,6 +408,33 @@ const baseSelectors = {
 
 		return select( STORE_NAME ).isFetchingGetPropertiesProfiles( accountID );
 	} ),
+
+	/**
+	 * Gets all "classic" Universal Analytic and Google Analytics 4 (GA4) properties this account can access.
```
danielgent, line 413: :eyes: my terminology here. I took this from the GA4 Google docs we have

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -407,6 +408,33 @@ const baseSelectors = {
...
+	getPropertiesIncludingGA4: createRegistrySelector( ( select ) => ( state, accountID ) => {
+		const properties = select( STORE_NAME ).getProperties( accountID );
+
+		let propertiesGA4 = [];
+
+		if ( select( 'modules/analytics-4' ) ) {
+			propertiesGA4 = select( 'modules/analytics-4' ).getProperties( accountID );
+		}
+
+		const propertiesGA4Mapped = propertiesGA4.map( ( property ) => ( { ...property, id: property._id } ) );
```
danielgent, line 434: according to the fixtures (assuming these to be correct) `ua` properties have an `id` key, but `ga4` have an `_id` key

assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -255,6 +255,22 @@ describe( 'modules/analytics properties', () => {
...
+
+				// Load data into stores
+				registry.dispatch( STORE_NAME ).receiveGetProperties( fixtures.propertiesProfiles.properties, { accountID } );
+				registry.dispatch( 'modules/analytics-4' ).receiveGetProperties( fixtures.propertiesGA4, { accountID } );
+
+				const properties = registry.select( STORE_NAME ).getPropertiesIncludingGA4( testAccountID );
+
+				expect( properties ).toMatchSnapshot();
+				expect( properties ).toHaveLength( 23 );
+			} );
```
danielgent, line 271: I would also like to test for when the `modules/analytics-4` datastore is not available, but I don't any way to conditionally load data stores when testing

the test util registerAllStoresOn registers all the stores!
https://github.com/google/site-kit-wp/blob/4cac32bb3a3a5fc977d07546ef3fe5d7f3bc94c3/tests/js/utils.js#L359

assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -255,6 +255,22 @@ describe( 'modules/analytics properties', () => {
...
 			} );
 		} );
+		describe( 'getPropertiesIncludingGA4', () => {
+			it( 'returns a sorted list of ua and ga4 properties ', async () => {
+				const testAccountID = fixtures.profiles[ 0 ].accountId; // eslint-disable-line sitekit/acronym-case
+				const accountID = testAccountID;
+
+				// Load data into stores
+				registry.dispatch( STORE_NAME ).receiveGetProperties( fixtures.propertiesProfiles.properties, { accountID } );
+				registry.dispatch( 'modules/analytics-4' ).receiveGetProperties( fixtures.propertiesGA4, { accountID } );
```
danielgent, line 265: Copied store hydration code from previous tests

I could also test for when both of these selectors have not yet resolved *but* that feels out of scope for this as we are already testing that in the respective selector's tests

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -407,6 +409,31 @@ const baseSelectors = {
...
+	 * @param {Object} state     Data store's state.
+	 * @param {string} accountID The Analytics Account ID to fetch properties for.
+	 * @return {(Array.<Object>|undefined)} An array of Analytics properties; `undefined` if not loaded.
+	 */
+	getPropertiesIncludingGA4: createRegistrySelector( ( select ) => ( state, accountID ) => {
+		let properties = select( STORE_NAME ).getProperties( accountID );
+
+		if ( select( MODULES_ANALYTICS_4 ) ) {
+			const	propertiesGA4 = select( MODULES_ANALYTICS_4 ).getProperties( accountID );
+			const propertiesGA4Mapped = propertiesGA4.map( ( property ) => ( { ...property, id: property._id } ) );
```
danielgent, line 431: according to the fixtures (assuming these to be correct) `ua` properties have an `id` key, but `ga4` have an `_id` key

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -407,6 +409,31 @@ const baseSelectors = {
...
+			properties = properties.concat( propertiesGA4Mapped );
+		}
+
+		return sortBy( properties, [ 'name', 'id' ] );
```
eugene-manuilov, line 435: @danielgent i think it won't work as expected because GA4 properties store property names in the `displayName` attribute. Let's rewrite this to use custom compare function `properties.sort( ( a, b ) => { ... } )`.

eugene-manuilov (changes requested):
Thanks, @danielgent. It's a good start. Found a few issues that need to be addressed before we can merge this PR. Could you please take a look at it?

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -407,6 +409,31 @@ const baseSelectors = {
...
+	 *
+	 * @param {Object} state     Data store's state.
+	 * @param {string} accountID The Analytics Account ID to fetch properties for.
+	 * @return {(Array.<Object>|undefined)} An array of Analytics properties; `undefined` if not loaded.
+	 */
+	getPropertiesIncludingGA4: createRegistrySelector( ( select ) => ( state, accountID ) => {
+		let properties = select( STORE_NAME ).getProperties( accountID );
+
+		if ( select( MODULES_ANALYTICS_4 ) ) {
+			const	propertiesGA4 = select( MODULES_ANALYTICS_4 ).getProperties( accountID );
```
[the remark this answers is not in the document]
danielgent: @eugene-manuilov I increased the size of tabs in vs-code so I can spot these tabs
tabs-spaces

I also found a setting where vscode was rewriting tabs as spaces :facepalm: 
`"editor.insertSpaces": false,`

So hopefully I'll make less of these formatting issues now

assets/js/modules/analytics/datastore/properties.js
danielgent: @eugene-manuilov so I can differentiate between the `ga` and `ua4` properties based on if they have the key `_id` or `id`, right? (from looking at the fixtures)

So I can do that.  :+1:   (and drop the `propertiesGA4Mapped` above)

assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -255,6 +255,22 @@ describe( 'modules/analytics properties', () => {
...
+
+				const properties = registry.select( STORE_NAME ).getPropertiesIncludingGA4( testAccountID );
+
+				expect( properties ).toMatchSnapshot();
```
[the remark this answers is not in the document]
danielgent: @eugene-manuilov  Makes total sense!  I can assert that each element is in the correct place, and then anyone else can quickly check that I've done the ticket correctly

danielgent:
@eugene-manuilov  I think I've addressed all your feedback now! Thanks for that.

There are quite a few changes so I imagine this will require another round of feedback.  All suggestions welcome :smile:

assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -255,6 +259,98 @@ describe( 'modules/analytics properties', () => {
...
+				expect( properties[ 2 ].id ).toBe( 'UA-151753095-1' );
+				expect( properties[ 3 ]._id ).toBe( '151753095-3' );
+
+				expect( properties[ 0 ].name ).toBe( 'rwh' );
+				expect( properties[ 1 ].displayName ).toBe( 'troubled-tipped.example.com' );
+				expect( properties[ 2 ].name ).toBe( 'troubled-tipped.example.com' );
+				expect( properties[ 3 ].displayName ).toBe( 'www.elasticpress.io' );
+			} );
+
+			it( 'returns only ua properties when ga4 datastore not available ', async () => {
```
eugene-manuilov, line 314: @danielgent please take my apologies, but I think we don't need this test because it does nothing but repeats the test above. Let's remove this.

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -325,6 +326,18 @@ const baseResolvers = {
...
+
+function compare( a, b ) {
+	if ( a < b ) {
+		return -1;
+	}
+	if ( a > b ) {
+		return 1;
+	}
+	return 0;
+}
```
eugene-manuilov, line 339: These functions are very specific to the `getPropertiesIncludingGA4` selector and used only there. Let's move it into the selector body to encapsulate it there.

eugene-manuilov (changes requested):
Good work, @danielgent. There are two minor things to address and I'll approve this PR. Could you please update it?

eugene-manuilov (changes requested):
Good work, @danielgent. There are two minor things to address and I'll approve this PR. Could you please update it?

assets/js/modules/analytics/datastore/properties.js
danielgent: Sure thing @eugene-manuilov !  I was half expecting you to point out that they already existed in native JS or in Lodash :see_no_evil:

assets/js/modules/analytics/datastore/properties.test.js
danielgent: No problem @eugene-manuilov 

It just tests that the code I've written doesn't explode if the ga4 datastore isn't available.  It's not where I'd expect it to fail though :+1:

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -407,6 +408,54 @@ const baseSelectors = {
...
+
+		function compare( a, b ) {
+			if ( a < b ) {
+				return -1;
+			}
+			if ( a > b ) {
+				return 1;
+			}
+			return 0;
+		}
```
eugene-manuilov, line 443:
```suggestion
		const isGA4 = ( property ) => !! property._id;
		const compare = ( a, b ) => {
			if ( a < b ) {
				return -1;
			}
			if ( a > b ) {
				return 1;
			}
			return 0;
		}
```

assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -255,6 +256,59 @@ describe( 'modules/analytics properties', () => {
...
+						_id: '151753095-3',
+						_accountID: '151753095',
+						displayName: 'www.elasticpress.io',
+					},
+					{
+						_id: '151753095-4',
+						_accountID: '151753095',
+						displayName: 'troubled-tipped.example.com',
+					},
+				], { accountID } );
```
eugene-manuilov, line 294:
```suggestion

				registry.dispatch( MODULES_ANALYTICS_4 ).receiveGetProperties(
					[
						{
							_id: '151753095-3',
							_accountID: '151753095',
							displayName: 'www.elasticpress.io',
						},
						{
							_id: '151753095-4',
							_accountID: '151753095',
							displayName: 'troubled-tipped.example.com',
						},
					],
					{ accountID }
				);
```

eugene-manuilov (changes requested):
Awesome! Thanks, @danielgent! Looks good to me. There are two formatting issues, I'll take care of it.

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -407,6 +408,53 @@ const baseSelectors = {
...
+		const isGA4 = ( property ) => !! property._id;
+		const compare = ( a, b ) => {
+			if ( a < b ) {
+				return -1;
+			}
+			if ( a > b ) {
+				return 1;
+			}
+			return 0;
+		}
```
eugene-manuilov, line 442: Forgot to update here as well:
```suggestion
		};
```

eugene-manuilov (changes requested):

eugene-manuilov (approved):

Issue #3169, opened 2021-04-15
Implement GA4 selector to check if Admin API is generally working correctly


felixarntz:
The GA4 implementation will initially rely on an alpha version of the Analytics Admin API, which is clearly problematic in a way, but is currently needed. Furthermore, given that the GA4 functionality at this point is not essential to using Site Kit, it will entirely be implemented in a way that, should one of the key API requests fail, the functionality is ignored, assuming that there was a breaking change in the API.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a new `modules/analytics-4` selector `isAdminAPIWorking()` which returns a boolean for whether the Admin API is "essentially working correctly" (or `undefined` if not loaded yet).
    * Returns `true` if both `modules/analytics-4` selectors `getProperties` and `getWebDataStreams` have resolved successfully (for _any_ selector parameters).
    * Returns `false` if one of them resulted in an error.
    * Returns `undefined` otherwise.

## Implementation Brief

* Create a new partial datastore `assets/js/modules/analytics-4/datastore/api.js`. It should follow the general approach of defining different parts of a datastore (like base actions, controllers, resolvers, etc) and export combined store along with base bits via named export. See `properties` partial datastore for reference.
* Add a new `isAdminAPIWorking` selector:
  * It should return `TRUE` if:
    * The `state['properties']` object has at least one account with a non-empty array of properties;
    * The `state['webdatastreams']` object has at least one property with a non-empty array of web datastreams;
    * There are no entries in the `state['errors']` object which key start with `getProperties::` prefix;
    * There are no entries in the `state['errors']` object which key start with `getWebDataStreams::` prefix.
  * It should return `FALSE` if there is at least one error in the `state['errors']` object with these prefixes;
  * It should return `undefined` if `state['properties']` and `state['webdatastreams']` objects are empty and there is no errors in the `state['errors']` object with previously defined prefixes.

### Test Coverage

* Add js tests to cover possible cases for the new selector.

### Visual Regression Changes

* N/A

## QA Brief

* Assert that the new selector has been implemented as per AC
* Check that my tests are testing against the AC correctly
* All other tests and checks should pass

## Changelog entry

* N/A

tofumatt:
Livin' on the edge! 😎 

IB ✅

johnPhillips:
QA ✅

- Everything seems implemented correctly as per the AC
- The tests pass

PR #3235, opened 2021-04-28
Enhancement/3169 ga4 selector admin api working

danielgent:
## Summary

Addresses issue #3169

## Checklist


assets/js/modules/analytics-4/datastore/api.test.js
```diff
@@ -0,0 +1,119 @@
...
+			${ true }  | ${ true }  |  ${ false }   | ${ false }   | ${ true }
+			${ true }  | ${ false } |  ${ false }   | ${ false }   | ${ undefined }
+			${ false } | ${ true }  |  ${ false }   | ${ false }   | ${ undefined }
+			${ false } | ${ false } |  ${ false }   | ${ false }   | ${ undefined }
+			${ true }  | ${ true }  |  ${ true }    | ${ false }   | ${ false }
+			${ true }  | ${ true }  |  ${ false }   | ${ true }    | ${ false }
+			${ true }  | ${ true }  |  ${ true }    | ${ true }    | ${ false }
+			${ false } | ${ true }  |  ${ true }    | ${ true }    | ${ false }
+			${ true }  | ${ false } |  ${ true }    | ${ true }    | ${ false }
+			${ false } | ${ false } |  ${ true }    | ${ true }    | ${ false }
```
danielgent, line 60: Previously I had more descriptive individual tests (see commit 
https://github.com/google/site-kit-wp/pull/3235/commits/33d24a36b90faa04365ea0bfb82a9ffdb4ee0873 )

But the AC implies a lot of combinations so this test is more comprehensive, but less readable!

assets/js/modules/analytics-4/datastore/api.js
```diff
@@ -0,0 +1,92 @@
...
+
+		for ( const propertyList of Object.values( state.properties ) ) {
+			if ( propertyList && propertyList.length && propertyList.length > 0 ) {
+				for ( const webdatastreamList of Object.values( state.webdatastreams ) ) {
+					if ( webdatastreamList && webdatastreamList.length && webdatastreamList.length > 0 ) {
+						return true;
+					}
+				}
+			}
+		}
```
tofumatt, line 68: Because this is a temporary, admittedly kinda hacky selector I think this approach is okay (I approved it in the IB 😅), but this kind of direct state access is generally not something we use or endorse—usually we build selectors, even ones marked as `@private`, for everything, so if the internal shape of our state data changes we don't get bugs.

It might be good to explicitly highlight the approach here as against normal conventions so it's acknowledged and doesn't confuse anyone in the future. Just a comment mentioning that this only exists because we're working with an Alpha, unstable API at the time of development.

assets/js/modules/analytics-4/datastore/api.js
```diff
@@ -0,0 +1,92 @@
...
+	/**
+	 * Checks if the Admin API is working.
+	 *
+	 * Returns false if there are any errors for getProperties or getWebDataStreams.
+	 *
+	 * Returns true if getProperties or getWebDataStreams has fetched successfully.
+	 *
+	 * Else returns undefined.
+	 *
+	 * @since n.e.x.t
```
tofumatt, line 45: Minor thing but let's mark this as an internal-only selector.

```suggestion
	 * @since n.e.x.t
	 * @private
```

assets/js/modules/analytics-4/datastore/api.js
```diff
@@ -0,0 +1,92 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Internal dependencies
+ */
+import Data from 'googlesitekit-data';
+
+const baseInitialState = { };
```
tofumatt, line 24:
```suggestion
const baseInitialState = {};
```

assets/js/modules/analytics-4/datastore/api.js
```diff
@@ -0,0 +1,92 @@
...
+};
+
+const store = Data.combineStores(
+	{
+		initialState: baseInitialState,
+		actions: baseActions,
+		controls: baseControls,
+		reducer: baseReducer,
+		resolvers: baseResolvers,
+		selectors: baseSelectors,
```
tofumatt, line 81: Any of these values that are empty aren't technically needed; the `combineStores` function will accept `undefined` values just fine: https://github.com/google/site-kit-wp/blob/3fb4b73421597f329da0ce0022ba315a6cf18630/assets/js/googlesitekit/data/utils.js#L183-L201

I'd remove any of these that you aren't using, especially as it reduces boilerplate and saves creating the passthrough reducer in this file.

assets/js/modules/analytics-4/datastore/api.js
```diff
@@ -0,0 +1,92 @@
...
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {(boolean | undefined)} True, false, or undefined.
+	 */
+	isAdminAPIWorking( state ) {
+		for ( const errorKey in state.errors ) {
+			if (
+				errorKey.startsWith( 'getProperties' ) ||
+				errorKey.startsWith( 'getWebDataStreams' )
```
tofumatt, line 54: The IB mentions that the keys should start with `getProperties::` not `getProperties`, so I think this is more appropriate:

```suggestion
				errorKey.startsWith( 'getProperties::' ) ||
				errorKey.startsWith( 'getWebDataStreams::' )
```

assets/js/modules/analytics-4/datastore/api.test.js
```diff
@@ -0,0 +1,119 @@
...
+							new Error( 'foo' ),	'getProperties', [ 'foo', 'bar' ]
+						);
+					}
+					if ( errorWebData ) {
+						registry.dispatch( STORE_NAME ).receiveError(
+							new Error( 'foo' ),	'getWebDataStreams', [ 'foo', 'bar' ]
+						);
+					}
+
+					const isAdminAPIWorking = registry.select( STORE_NAME ).isAdminAPIWorking( );
```
tofumatt, line 113: Weird ESLint doesn't complain 😆 
```suggestion
					const isAdminAPIWorking = registry.select( STORE_NAME ).isAdminAPIWorking();
```

assets/js/modules/analytics-4/datastore/api.test.js
```diff
@@ -0,0 +1,119 @@
...
+									defaultUri: 'http://example.com',
+									displayName: 'Test GA4 WebDataStream',
+								},
+							],
+							{ propertyID: 'foobar' }
+						);
+					}
+					if ( errorProperty ) {
+						registry.dispatch( STORE_NAME ).receiveError(
+							new Error( 'foo' ),	'getProperties', [ 'foo', 'bar' ]
```
tofumatt, line 104: The spacing here with the... tabs? is sort of throwing me off. I guess the idea is to make it readable but tab widths vary per-editor and I don't think we do this elsewhere. Mind just setting each argument apart with a space instead? Eg
```suggestion
							new Error( 'foo' ), 'getProperties', [ 'foo', 'bar' ]
```

assets/js/modules/analytics-4/datastore/api.test.js
tofumatt: I'm usually a fan of tests that are more verbose in exchange for readability—but this formatting is actually pretty handy and readable for what this is. Nice 👍🏻

tofumatt (changes requested):
A few minor tweaks and documentation adjustments here but largely looks good 👍🏻

assets/js/modules/analytics-4/datastore/api.js
danielgent: makes sense @tofumatt 

I saw an empty `baseControls`  in the ga-4 properties datastores so I just copied that :smile: 

https://github.com/google/site-kit-wp/blob/666153eda0986002b2533c10f2821e0d8a3cc0f8/assets/js/modules/analytics-4/datastore/properties.js#L167

Removing now

assets/js/modules/analytics-4/datastore/api.js
danielgent: The thought did cross my mind @tofumatt but I was just following the IB! :angel: 
(Although the IB says to access the properties like `state['properties']` but I thought this was not needed :wink: 

I'll add a comment now

tofumatt:
@googlebot I consent.

tofumatt:
@googlebot I consent.

tofumatt:
@googlebot I consent.

tofumatt (approved):

assets/js/modules/analytics-4/datastore/api.js
```diff
@@ -38,8 +38,11 @@ const baseSelectors = {
 	 * @return {(boolean|undefined)} True, false, or undefined.
 	 */
 	isAdminAPIWorking( state ) {
-		// NOTE - This is against normal conventions. Usually we build selectors for everything
-		// we are working with an Alpha, unstable API at the time of development.
+		// This is against normal state access conventions.
+		// Usually we build selectors for any state access and avoid
+		// looking around state internals.
+		// But we are working with an Alpha, unstable Google API at the time
+		// of development.
```
danielgent, line 45: Nice @tofumatt !  Mine looks like a rant on Twitter in comparison

It makes me want to re-read The Elements of Style now I'm working on OSS :smile:

Issue #3154, opened 2021-04-14
Implement wrapping infrastructure for dealing with WordPress post metadata


felixarntz:
As part of the Idea Hub module functionality, it will be necessary for the first time in Site Kit to deal with post metadata. Therefore, we should add basic infrastructure wrapping the relevant WordPress API (similar to e.g. `Options` and `User_Options` classes). [See more information on the overall approach](https://docs.google.com/document/d/1mIBLvv-v2Y3OMiHsOn5kMu3Fnt-wAqbWDnAiuraTy0k/edit).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a new interface `Google\Site_Kit\Core\Storage\Post_Meta_Interface`, with the following methods:
    * `get( int $post_id, string $key, bool $single = false ) : mixed`: Gets the value of the given post meta key.
    * `update( int $post_id, string $key, mixed $value, mixed $prev_value = '' ): bool`: Updates the value for a post meta key.
    * `add( int $post_id, string $key, mixed $value, bool $unique = false ) : bool`: Adds a new value for a post meta key.
    * `delete( int $post_id, string $key, mixed $value = '' ): bool`: Deletes the given post meta key.
* There should be a new class `Google\Site_Kit\Core\Storage\Post_Meta` which implements the above interface:
    * `get` should rely on `get_post_meta`.
    * `update` should rely on `update_post_meta`.
    * `add` should rely on `add_post_meta`.
    * `delete` should rely on `delete_post_meta`.
* There should be a new abstract class `Google\Site_Kit\Core\Storage\Post_Meta_Setting`:
    * Relies on a `META_KEY` class constant, to be defined in each sub-class.
    * Relies on a `Post_Meta_Interface` being passed into the constructor.
    * Has a public `register()` method, which calls `register_meta( 'post', ... )`. It should rely on protected methods `get_type()` and `get_sanitize_callback()` which can be overridden by each sub-class (similar to e.g. in the existing `User_Setting` class. The `single` argument should be enforced to `true` for now.
    * Should have methods `get( $post_id )`, `set( $post_id )` and `delete( $post_id )` which wrap the respective `Post_Meta` methods.
        * `get` should always use `$single` as `true`, and it should return the value of a protected `get_default()` method if no actual meta value is set.
[...]

PR #3241, opened 2021-04-28
Enhancement/3154 post metadata

eugene-manuilov:
## Summary

Addresses issue #3154 

## Checklist


includes/Core/Storage/Post_Meta_Setting.php
```diff
@@ -0,0 +1,151 @@
...
+	 * Gets the value of the setting.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param int $post_id Post ID.
+	 * @return mixed Value set for the setting, or default if not set.
+	 */
+	public function get( $post_id ) {
+		$value = $this->post_meta->get( $post_id, static::META_KEY, true );
+		if ( false === $value ) {
```
aaemnnosttv, line 119: This will return an empty string as the default/non-existent value, not `false`.

How about we add a `has( $post_id )` method here (similar to options) which calls [`metadata_exists( ... )`](https://developer.wordpress.org/reference/functions/metadata_exists/)? That way we'd know for sure if the value truly does not exist (and should return the default) or if the saved value is just empty (which should not return the default).

I suppose this change would require an addition to the interface as well which is okay I think (any objections @felixarntz ?)  If he'd rather not add that, we can at least use it inline here but I think it would be worth adding.

includes/Core/Storage/Post_Meta_Setting.php
```diff
@@ -0,0 +1,151 @@
...
+			$value = $this->get_default();
```
aaemnnosttv, line 120: Mostly a stylistic preference, but IMO early returns are preferable over changing variables.
```suggestion
			return $this->get_default();
```

tests/phpunit/integration/Core/Storage/Post_MetaTest.php
```diff
@@ -0,0 +1,110 @@
...
+	public function test_add() {
+		global $wpdb;
+
+		$post_meta  = new Post_Meta();
+		$post_id    = $this->factory()->post->create();
+		$meta_key   = 'test_meta';
+		$meta_value = 'test_value';
+
+		$meta_id   = $post_meta->add( $post_id, $meta_key, $meta_value );
+		$new_value = $wpdb->get_var( $wpdb->prepare( "SELECT meta_value FROM {$wpdb->postmeta} WHERE post_id = %d AND meta_key = %s", $post_id, $meta_key ) );
```
aaemnnosttv, line 53: How about we extract the query to the test case like we have for `TestCase::queryOption` ?

tests/phpunit/integration/Core/Storage/Post_MetaTest.php
```diff
@@ -0,0 +1,110 @@
...
+				'meta_value' => $meta_value,
+			)
+		);
+
+		$meta_id = $wpdb->insert_id;
+		$deleted = $post_meta->delete( $post_id, $meta_key );
+		$meta    = $wpdb->get_row( $wpdb->prepare( "SELECT * FROM {$wpdb->postmeta} WHERE meta_id = %d", $meta_id ) );
+
+		$this->assertTrue( $deleted );
+		$this->assertNull( $meta );
```
aaemnnosttv, line 107: Similarly, for options we have `assertOptionNotExists` and `assertOptionExists` which would be useful for meta too.

tests/phpunit/integration/Core/Util/ResetTest.php
```diff
@@ -50,11 +50,21 @@ public function test_all() {
...
+		add_post_meta( $post_id, 'googlesitekit-keep', 'keep' );
+		add_post_meta( $post_id, 'googlesitekit_keep', 'delete' );
+
 		$this->run_reset( $context );
 
 		// Ensure options that don't start with googlesitekit_ are not deleted.
 		$this->assertEquals( 'keep', get_option( 'googlesitekitkeep' ) );
 		$this->assertEquals( 'keep', get_option( 'googlesitekit-keep' ) );
+
+		// Ensure post meta that don't start with googlesitekit_ isn't deleted.
```
aaemnnosttv, line 64:
```suggestion
		// Ensure post meta that do not start with googlesitekit_ are not deleted.
```

aaemnnosttv (changes requested):
Thanks @eugene-manuilov – this looks great overall, just one thing which needs fixing and a few minor suggestions.

includes/Core/Storage/Post_Meta_Setting.php
eugene-manuilov: Updated.

includes/Core/Storage/Post_Meta_Setting.php
eugene-manuilov: Updated.

tests/phpunit/integration/Core/Storage/Post_MetaTest.php
eugene-manuilov: Updated.

tests/phpunit/integration/Core/Storage/Post_MetaTest.php
eugene-manuilov: Updated.

tests/phpunit/integration/Core/Util/ResetTest.php
eugene-manuilov: Updated.

includes/Core/Storage/Post_Meta_Setting.php
```diff
@@ -115,12 +128,11 @@ protected function get_sanitize_callback() {
 	 * @return mixed Value set for the setting, or default if not set.
 	 */
 	public function get( $post_id ) {
-		$value = $this->post_meta->get( $post_id, static::META_KEY, true );
-		if ( false === $value ) {
-			$value = $this->get_default();
+		if ( $this->has( $post_id, static::META_KEY ) ) {
```
aaemnnosttv, line 131: This should be `! has` for returning the default 😄 

Also, this should only need to pass through the `$post_id`.

includes/Core/Storage/Post_Meta_Setting.php
```diff
@@ -106,6 +106,19 @@ protected function get_sanitize_callback() {
...
+	 * Checks whether a post meta exists or not.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param int    $post_id  Post ID.
+	 * @param string $meta_key Meta key.
+	 * @return bool True if the meta key exists, otherwise false.
+	 */
+	public function has( $post_id, $meta_key ) {
+		return metadata_exists( 'post', $post_id, $meta_key );
```
aaemnnosttv, line 119: Let's update this to use the meta key via the class constant here like we do in the rest of the class – only the `$post_id` is needed here.

aaemnnosttv (commented):
Thanks @eugene-manuilov – this is very close, just a few small details left to fix.

includes/Core/Storage/Post_Meta_Setting.php
eugene-manuilov: :facepalm: yes, updated.

includes/Core/Storage/Post_Meta_Setting.php
eugene-manuilov: :+1: updated.

aaemnnosttv (approved):
Thanks @eugene-manuilov !

Issue #3161, opened 2021-04-15
Enable easy code placement for Ads ID


marrrmarrr:
## Feature Description

For sites with gtag.js (meaning Site Kit has placed the GA tag, can be either existing or newly provisioned accounts), enable the user to copy their Ads ID and incorporate that into the gtag.js snippet. 

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* For sites with gtag.js (Analytics), display an "Ads Conversion ID" field in the GA settings tab
* As context, display text "If you're using Google Ads, insert your Ads conversion ID if you'd like Site Kit to place the snippet on your site"
* The field should only be displayed if the Analytics snippet is inserted from Site Kit (i.e. if `useSnippet` is enabled).
* The field should be the last in the Analytics settings panel, in its own row.
* Once the user copies their ID into the field and clicks "Confirm changes", place the extra line in the gtag.js snippet. The line will have the format `gtag( 'config', 'AW-CONVERSION_ID');`
    * This should be implemented for both the UA Analytics snippet as well as the GA4 snippet (see #3106).
    * The field input should be validated to start with `AW-`.
* After the user has clicked "confirm changes", display an entry with heading "Ads Conversion ID" and the inserted conversion ID (`AW-...`) in the GA settings _if_ the Analytics snippet is inserted from Site Kit (i.e. if `useSnippet` is enabled). 

### JS/Frontend

* Create a new setting for Analytics in [`settingsSlugs`](https://github.com/google/site-kit-wp/blob/df83274d9db0d795e2d02dda236ff24e5ce59b20/assets/js/modules/analytics/datastore/base.js#L28-L38): `adsID`
* Create a new component for the Analytics `SetupForm` to use in `assets/js/modules/analytics/components/common` named `AdsIDInput` containing a `TextField` that the user can use to submit the `adsID` setting to Analytics module, similar to how other settings are loaded and saved in the Analytics module. The label for the Ads ID value should read *"Insert your Ads Conversion ID here if you want Site Kit to place the Ads snippet on your site"*. `ProfileNameTextField` can serve as a reference for how to build this setting, as it is another text entry setting in Analytics.
  * This input should only appear if the settings `useSnippet` is enabled—if Site Kit isn't outputting the Analytics tag for this site, don't show this setting.
[...]

Issue #3165, opened 2021-04-15
Remove GA4 account ID setting since it is the same as UA account ID


felixarntz:
While eventually both the `analytics` and the `analytics-4` module need to support a setting for the account ID, the accounts are the same within both ecosystems. Therefore, the `accountID` setting in the `analytics-4` module should be removed for now, and if it is needed anywhere should be taken from the `analytics` module which is a requirement to have active anyway.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `accountID` setting should be entirely removed from the `analytics-4` module.
    * In PHP (e.g. the settings class and Site Health integration etc.).
    * In JS (e.g. from the `modules/analytics-4` store).

## Implementation Brief

* Comment out (soft delete) `accountID` usage in the `Analytics_4::is_connected` method.
* Comment out `analytics_4_account_id` section in the `Analytics_4::get_debug_fields` method.
* Comment out all occurrences of the `accountID` in the `Analytics_4/Settings` class.
* Comment out `accountID` setting slug in the `assets/js/modules/analytics-4/datastore/base.js` file.
* Add a new TODO note to all commented out sections which says to uncomment it when `Analytics` and `Analytics 4` modules become officially separated.

### Test Coverage

* Update existing phpunite tests that are affected by the aforementioned changes.

### Visual Regression Changes

* N/A

## QA Brief

* Check that the code is correctly commented as per the IB
* Check that PHPUnit tests pass

## Changelog entry

* N/A

eugene-manuilov:
Sending this ticket back to @felixarntz to re-define the AC section.

felixarntz:
@eugene-manuilov Updated the ACs (and description and title).

eugene-manuilov:
IB is updated.

felixarntz:
@eugene-manuilov Good idea keeping the code, as commented-out. IB ✅

johnPhillips:
QA ✅

Verified that:
- there are no more uncommented instances of `accountID` in `includes/Modules/Analytics_4` by using ripgrep for `account` and manually checking the instances.
- the relevant `accountID` instances in `includes/Modules/Analytics_4.php` (`::get_debug_fields` and `::is_connected` sections) have been commented out
- the `accountID` settings slug in `assets/js/modules/analytics-4/datastore/base.js` has been commented out
- In all cases, the `TODO` has been added as per the ACs

PR #3257, opened 2021-04-29
Add support for Ads conversion tracking in Analytics

aaemnnosttv:
## Summary

Addresses issue #3161

## Relevant technical choices

- No AMP implementation in place for the conversion ID yet (will be added in a follow-up pending guidance) 

## Checklist


assets/js/modules/analytics-4/datastore/base.js
```diff
@@ -27,6 +27,7 @@ const baseModuleStore = Modules.createModuleStore( 'analytics-4', {
 	settingSlugs: [
 		// TODO: This can be uncommented when Analytics and Analytics 4 modules are officially separated.
 		// 'accountID',
+		'adsConversionID',
```
felixarntz, line 30: I think it would be better to also add this here only as a commented-out setting, the same way as `accountID`. As decided while discussing #3165, we shouldn't double-store or sync values between both modules. By adding this here as commented out we indicate that in principle it needs to be stored in the module as well, but let's not worry about it for now, since `analytics-4` is only available with `analytics` active anyway.

assets/js/modules/analytics/components/common/AdsConversionIDTextField.js
```diff
@@ -0,0 +1,87 @@
...
+ */
+import Data from 'googlesitekit-data';
+import { TextField, HelperText, Input } from '../../../../material-components';
+import { STORE_NAME } from '../../datastore/constants';
+import VisuallyHidden from '../../../../components/VisuallyHidden';
+import { isValidAdsConversionID } from '../../util';
+const { useSelect, useDispatch } = Data;
+
+export default function AdsConversionIDTextField() {
+	const adsConversionID = useSelect( ( select ) => select( STORE_NAME ).getAdsConversionID() );
```
felixarntz, line 41: +1 for calling the setting `adsConversionID`.

assets/js/modules/analytics/components/settings/SettingsView.js
```diff
@@ -141,6 +142,18 @@ export default function SettingsView() {
...
+			<div className="googlesitekit-settings-module__meta-items">
+				<div className="googlesitekit-settings-module__meta-item">
+					<h5 className="googlesitekit-settings-module__meta-item-type">
+						{ __( 'Ads Conversion ID', 'google-site-kit' ) }
+					</h5>
+					<p className="googlesitekit-settings-module__meta-item-data">
+						<DisplaySetting value={ adsConversionID } />
+					</p>
+				</div>
+			</div>
```
felixarntz, line 155: Probably better to only show this if `useSnippet`, the same as in `SettingsEdit` and like we do here for e.g. Anonymize IP setting.

includes/Modules/Analytics/Tag_Interface.php
```diff
@@ -0,0 +1,47 @@
...
+namespace Google\Site_Kit\Modules\Analytics;
+
+/**
+ * Interface for an Analytics tag.
+ *
+ * @since n.e.x.t
+ * @access private
+ * @ignore
+ */
+interface Tag_Interface {
```
felixarntz, line 20: +1 for adding this interface. While quite specific, it helps clarifying what an Analytics tag class implementation needs.

includes/Modules/Analytics/Tag_Interface.php
```diff
@@ -0,0 +1,47 @@
...
+	 * @param string $domain Domain name.
+	 */
+	public function set_home_domain( $domain );
+
+	/**
+	 * Sets whether or not to anonymize IP addresses.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param bool $anonymize_ip Anonymize or not.
```
felixarntz, line 35: Nitpick, but maybe let's use a bit more "human" clear language like "Whether to anonymize IP addresses or not.".

includes/Modules/Analytics_4.php
```diff
@@ -445,15 +445,20 @@ private function register_tag() {
...
+
+		if ( $tag->is_tag_blocked() ) {
+			return;
+		}
+
+		$tag->use_guard( new Tag_Verify_Guard( $this->context->input() ) );
+		$tag->use_guard( new Tag_Guard( $this->get_settings() ) );
+
+		if ( $tag->can_register() ) {
+			$tag->set_ads_conversion_id( $settings['adsConversionID'] );
```
felixarntz, line 459: See my other comment, this needs to come from the `Analytics\Settings` class instead. Since it won't be needed forever, let's not try to find a nice way of dependency injection here. What would be good to have though (also temporarily) is a private method like `get_analytics_module_settings()`, because that can also be used if we need to use the account ID from the `analytics` module in this class.

That method implementation could probably just manually instantiate the `Analytics\Settings` class and then call `get` on it.

includes/Modules/Analytics_4/Settings.php
```diff
@@ -66,6 +66,7 @@ protected function get_default() {
 			'ownerID'         => 0,
 			// TODO: This can be uncommented when Analytics and Analytics 4 modules are officially separated.
 			/* 'accountID'       => '', */ // phpcs:ignore Squiz.PHP.CommentedOutCode.Found
+			'adsConversionID' => '',
```
felixarntz, line 69: See above, let's add this here only as commented out.

tests/phpunit/integration/Modules/Analytics_4/SettingsTest.php
```diff
@@ -32,6 +32,7 @@ public function test_get_default() {
 			array(
 				// TODO: This can be uncommented when Analytics and Analytics 4 modules are officially separated.
 				// 'accountID'       => '',
+				'adsConversionID' => '',
```
felixarntz, line 35: See above, let's comment out this.

includes/Modules/Analytics_4/Web_Tag.php
```diff
@@ -45,6 +45,8 @@ protected function enqueue_gtag_script() {
 			// If the gtag script is already registered in the Analytics module, then we need to add G-<MEASUREMENT_ID> configuration only.
 			$config = sprintf( 'gtag("config", "G-%s");', esc_js( $this->tag_id ) );
 			wp_add_inline_script( 'google_gtagjs', $config );
+
+			$this->add_inline_ads_conversion_id_config();
```
felixarntz, line 49: I don't think this part should be here:
* If the `Analytics\Web_Tag` class will output the GA snippet (this conditional), that class already ads the conversion ID as needed.
* Otherwise, the `Analytics_4\Web_Tag` class calls the parent method (below), which already includes this anyway.

So I don't think this class needs to be changed.

felixarntz (changes requested):
This looks solid so far, +1 for the changes from the original IB. A few points of feedback, primarily let's not duplicate the `adsConversionID` setting between both modules for now.

includes/Modules/Analytics/Tag_Interface.php
aaemnnosttv: Agreed! I'll updated it in the other places then as well https://github.com/google/site-kit-wp/blob/d6ea05554e20353df9e770abd193c98b006cdf79/includes/Modules/Analytics/Web_Tag.php#L59

includes/Modules/Analytics_4.php
aaemnnosttv: I think we can handle this in a more centralized way by proxying a few settings values from the classic settings rather than adding inline exceptions across the codebase. This way, the `Settings` class is responsible for this and the usage need not change later when we decouple them. I'll add this to the PR and we can revert it if you'd prefer not to, but I think it should make for an easier transition.

includes/Modules/Analytics_4/Settings.php
aaemnnosttv: Updated to proxy from UA settings which I think is preferable to having dead code in place for later.

tests/phpunit/integration/Modules/Analytics_4/SettingsTest.php
aaemnnosttv: See previous comment.

assets/js/modules/analytics/components/settings/SettingsView.js
aaemnnosttv: Should we not also consider `canUseSnippet`? Wondering since `useSnippet` will remain `true` if GTM takes over adding the tag for Analytics even though Analytics isn't actually outputting the tag.

assets/js/modules/analytics-4/datastore/base.js
aaemnnosttv: Sorry, looks like I missed that issue. You mentioned that we shouldn't double-store or sync (I think you also mean double-storing here?) but what about proxying the value? Not for all the settings but only those that are relevant from the other module. That would allow us to use settings in GA4 from UA settings without requiring any special handling from the consumer. We can encapsulate this handling in its settings class entirely so we only have one place to change later when they are decoupled. I have an example in here in one commit, so we can revert if you'd prefer.

tests/phpunit/includes/Core/Modules/Module_With_Settings_ContractTests.php
```diff
@@ -39,7 +39,10 @@ public function test_setting_registration() {
 		$testcase    = $this->get_testcase();
 		$settings    = $this->get_module_with_settings()->get_settings();
 		$option_name = constant( get_class( $settings ) . '::OPTION' );
-		unregister_setting( $option_name, $option_name );
+
+		if ( array_key_exists( $option_name, get_registered_settings() ) ) {
+			unregister_setting( $option_name, $option_name );
+		}
```
aaemnnosttv, line 45: Not sure why this wasn't causing a problem before but the `Analytics_4Test` was failing from this because its setting isn't registered during the plugin bootstrap due to its conditional registration and all feature flags being off by default at that point.

includes/Modules/Analytics_4/Settings.php
```diff
@@ -90,4 +92,27 @@ protected function get_sanitize_callback() {
...
+	 * @TODO Remove once Analytics and Analytics 4 modules are officially separated.
+	 *
+	 * @since n.e.x.t
+	 */
+	protected function proxy_ua_settings() {
+		$analytics_settings   = new Analytics_Settings( $this->options );
+		$proxy_values_from_ua = function ( $ga4_settings ) use ( $analytics_settings ) {
+			$ua_settings                     = $analytics_settings->get();
+			$ga4_settings['accountID']       = $ua_settings['accountID'];
+			$ga4_settings['adsConversionID'] = $ua_settings['adsConversionID'];
```
aaemnnosttv, line 111: @felixarntz As mentioned before, I think it would be preferable to encapsulate the common settings as much as possible here so we don't need to add so many special cases and todos elsewhere in the codebase.

assets/js/modules/analytics-4/datastore/base.js
```diff
@@ -25,8 +25,9 @@ import { STORE_NAME } from './constants';
 const baseModuleStore = Modules.createModuleStore( 'analytics-4', {
 	storeName: STORE_NAME,
 	settingSlugs: [
-		// TODO: This can be uncommented when Analytics and Analytics 4 modules are officially separated.
+		// TODO: These can be uncommented when Analytics and Analytics 4 modules are officially separated.
 		// 'accountID',
+		// 'adsConversionID',
```
aaemnnosttv, line 30: Per the other changes, I think we can uncomment these here too 🤔

assets/js/modules/analytics/components/common/AdsConversionIDTextField.js
```diff
@@ -0,0 +1,95 @@
...
+
+	const { setAdsConversionID } = useDispatch( STORE_NAME );
+	const onChange = useCallback( ( { currentTarget } ) => {
+		let newValue = currentTarget.value.trim().toUpperCase();
+		// Automatically add the AW- prefix if not provided.
+		if ( 'AW-'.length < newValue.length && ! /^AW-/.test( newValue ) ) {
+			newValue = `AW-${ newValue }`;
+		}
```
aaemnnosttv, line 49: I added this to help with the formatting since I've seen the conversion ID referred to as a simple integer. The prefix only seems to be added when it's referenced in the `gtag` config.

aaemnnosttv:
This is still pending the AMP implementation for UA but otherwise it should be ready for review.

assets/js/modules/analytics/datastore/settings.js
```diff
@@ -126,6 +129,10 @@ export function validateCanSubmitChanges( select ) {
 	invariant( isValidPropertySelection( getPropertyID() ), INVARIANT_INVALID_PROPERTY_SELECTION );
 	invariant( isValidProfileSelection( getProfileID() ), INVARIANT_INVALID_PROFILE_SELECTION );
 
+	if ( getAdsConversionID() ) {
+		invariant( isValidAdsConversionID( getAdsConversionID() ), INVARIANT_INVALID_CONVERSION_ID );
+	}
```
aaemnnosttv, line 134: This could result in the user being blocked if they were to get an invalid value in the setting (either saved or just the current unsaved value) if the field is then hidden. I've updated the conditional display to be handled by the component itself in the edit view now so that it's only hidden if it has a valid value.

assets/js/modules/analytics/components/settings/SettingsView.js
felixarntz: Makes sense to check for `canUseSnippet` too. There are a few other places where we should probably add that (separate issue though).

includes/Modules/Analytics_4/Settings.php
felixarntz: I'm not sure this approach is better for now. From the server-side perspective, absolutely. However, I think this can easily invoke false expectations on the client that the `accountID` from the `analytics-4` store is the current value, when it might have been updated client-side in the `analytics` store, i.e. now being invalid.

If we want to go with this approach, I think we should ensure on the client-side the value also _always_ comes from the `analytics` module. In other words, the `modules/analytics-4` store should include `accountID` and `adsConversionID` via `createSettingsStore`, instead we could manually implement `getAccountID` and `getAdsConversionID` selectors that take it from `modules/analytics`.

Either way, I think this is a too wide-reaching change that could have other implications at this point. I suggest we keep the solution here in a way that does not affect other code (i.e. keep commented-out code and don't "automagically" take settings from the `analytics` module). In other words, the `analytics-4` module should technically not have any `accountID` or `adsConversionID` settings for now.

We can then open a follow-up issue to implement the approach you're suggesting, probably also including the client-side approach I'm suggesting above.

includes/Modules/Analytics_4/Settings.php
felixarntz: See above, let's not do this for now.

tests/phpunit/integration/Modules/Analytics_4/SettingsTest.php
felixarntz: See above.

felixarntz (changes requested):
I think we should go with the simple solution regarding the `adsConversionID` setting for now (see comment for details).

Also, for some reason VRT are failing here, not sure how that's related 🤔

includes/Modules/Analytics_4.php
aaemnnosttv: Updated to come from Analytics settings for now 👍

includes/Modules/Analytics_4.php
```diff
@@ -132,14 +134,21 @@ public function get_debug_fields() {
 		$settings = $this->get_settings()->get();
 
 		return array(
-			// TODO: This can be uncommented when Analytics and Analytics 4 modules are officially separated.
-			/* // phpcs:ignore Squiz.PHP.CommentedOutCode.Found
+			// phpcs:disable
```
aaemnnosttv, line 137: I couldn't not get PHPCS to be happy here due to various conflicting rules so needed to disable it for this block.

felixarntz (approved):
Great!

Issue #3155, opened 2021-04-14
Implement Idea Hub infrastructure (with mocks) for querying new ideas


felixarntz:
The Idea Hub infrastructure to retrieve new ideas should be implemented, with mock data for now since we cannot use the real API yet. [See more information on the overall approach](https://docs.google.com/document/d/1mIBLvv-v2Y3OMiHsOn5kMu3Fnt-wAqbWDnAiuraTy0k/edit).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `GET:new-ideas` data point of the new `Idea_Hub` module class should be implemented to return a mocked response, hard-coded into a variable (see below for the response to return). The `TODO` annotation for this data point should be modified to say `TODO: Implement this with the real API endpoint.`.
* The `modules/idea-hub` datastore should get a new selector `getNewIdeas( options )` which uses the `GET:modules/idea-hub/data/new-ideas` REST datapoint to list ideas.
    * `options` should be optional and support properties `offset` and `length`.
    * `options` **must not** be passed through to the server at this point. The server should always return all new ideas. Instead, it should only be used within the selector itself, to return a slice of the available new ideas. In other words, the resolver should also ignore `options` and only request the data once, if it's not available in the store state yet.

The following mock response should be used for the `GET:new-ideas` data point:
```json
[
    {
        "name": "ideas/17450692223393508734",
        "text": "Why Penguins are guanotelic?",
        "topics": [
            {
                "mid": "/m/05z6w",
                "display_name": "Penguins"
            }
        ]
    },
    {
        "name": "ideas/14025103994557865535",
        "text": "When was sushi Kalam introduced?",
        "topics": [
            {
                "mid": "/m/07030",
                "display_name": "Sushi"
            }
        ]
    },
    {
        "name": "ideas/7612031899179595408",
        "text": "How to speed up your WordPress site",
        "topics": [
            {
                "mid": "/m/09kqc",
                "display_name": "Websites"
            }
        ]
    },
    {
        "name": "ideas/2285812891948871921",
        "text": "Using Site Kit to analyze your success",
        "topics": [
            {
                "mid": "/m/080ag",
                "display_name": "Analytics"
            }
        ]
    },
    {
        "name": "ideas/68182298994557866271",
        "text": "How to make carne asada",
        "topics": [
            {
                "mid": "/m/07fhc",
                "display_name": "Cooking"
            }
        ]
    }
]
```

## Implementation Brief
[...]

PR #3259, opened 2021-04-30
Feature/3155 idea hub get new ideas

ivankruchkoff:
## Summary

Addresses issue #3155 

## Checklist


assets/js/modules/idea-hub/datastore/new-ideas.js
```diff
@@ -0,0 +1,109 @@
...
+	 * @param {Object} options          Options for getting new ideas.
+	 * @param {number} [options.offset] Optional. From which array index to get ideas.
+	 * @param {number} [options.length] Optional. Amount of new ideas to return.
+	 * @return {(Array.<Object>|undefined)} A list of idea hub ideas; `undefined` if not loaded.
+	 */
+	getNewIdeas( state, options = {} ) {
+		const { newIdeas } = state;
+		const offset = options?.offset || 0;
+		const length = options.length ? offset + options.length : newIdeas.length;
+		return 'offset' in options || 'length' in options ? newIdeas.slice( offset, length ) : newIdeas;
```
tofumatt, line 89: <del>
The ternary as a fallback option is a bit too dense for me 😅 

Even moving it to a newline might be nice, what about:

```suggestion
		return (
			'offset' in options ||
			'length' in options && newIdeas.slice( offset, length ) ||
			newIdeas
		);
```
</del>

Ah, scratch all that, I see what this is meant to do, but the order-of-operations isn't super-clear. I think the way this intends to behave is:

```js
return ( 'offset' in options || 'length' in options ) ? newIdeas.slice( offset, length ) : newIdeas;
```

but I read it as returning a boolean at first because the ternary conditions weren't grouped in parens. It made me think the return value here was a boolean, but it's not. Let's add parens around it so that's more immediately clear to the next developer 😅:
```suggestion
		return ( 'offset' in options || 'length' in options ) ? newIdeas.slice( offset, length ) : newIdeas;
```

assets/js/modules/idea-hub/datastore/new-ideas.test.js
```diff
@@ -0,0 +1,199 @@
...
+				expect( pendingNewIdeas ).toEqual( [] );
+				await untilResolved( registry, STORE_NAME ).getNewIdeas( options );
+
+				const newIdeas = registry.select( STORE_NAME ).getNewIdeas( options );
+
+				expect( fetchMock ).toHaveFetchedTimes( 1 );
+				expect( newIdeas ).toEqual( fixtures.newIdeas );
+			} );
+
+			it( 'correctly handles offset and length parameters', async () => {
```
tofumatt, line 74:
```suggestion
			it( 'uses offset and length parameters to adjust/limit the ideas returned by the selector', async () => {
```

assets/js/modules/idea-hub/datastore/new-ideas.test.js
```diff
@@ -0,0 +1,199 @@
...
+				registry.select( STORE_NAME ).getNewIdeas( customOptions );
+				await untilResolved( registry, STORE_NAME ).getNewIdeas( customOptions );
+
+				const newIdeas = registry.select( STORE_NAME ).getNewIdeas( customOptions );
+
+				expect( fetchMock ).toHaveFetchedTimes( 1 );
+				expect( newIdeas ).toEqual( fixtures.newIdeas.slice( 2, 4 ) );
+			} );
+
+			it( 'correctly handles missing options', async () => {
```
tofumatt, line 93:
```suggestion
			it( 'treats all options as optional', async () => {
```

assets/js/modules/idea-hub/datastore/new-ideas.test.js
```diff
@@ -0,0 +1,199 @@
...
+				registry.select( STORE_NAME ).getNewIdeas( {} );
+				await untilResolved( registry, STORE_NAME ).getNewIdeas( {} );
+
+				const newIdeas = registry.select( STORE_NAME ).getNewIdeas( {} );
+
+				expect( fetchMock ).toHaveFetchedTimes( 1 );
+				expect( newIdeas ).toEqual( fixtures.newIdeas );
+			} );
+
+			it( 'correctly handles offset parameter only', async () => {
```
tofumatt, line 108:
```suggestion
			it( 'adjusts idea results when only offset parameter is supplied', async () => {
```

assets/js/modules/idea-hub/datastore/new-ideas.test.js
```diff
@@ -0,0 +1,199 @@
...
+				registry.select( STORE_NAME ).getNewIdeas( customOptions );
+				await untilResolved( registry, STORE_NAME ).getNewIdeas( customOptions );
+
+				const newIdeas = registry.select( STORE_NAME ).getNewIdeas( customOptions );
+
+				expect( fetchMock ).toHaveFetchedTimes( 1 );
+				expect( newIdeas ).toEqual( fixtures.newIdeas.slice( 2 ) );
+			} );
+
+			it( 'correctly handles length parameter only', async () => {
```
tofumatt, line 126:
```suggestion
			it( 'adjusts idea results when only limit parameter is supplied', async () => {
```

assets/js/modules/idea-hub/datastore/new-ideas.test.js
```diff
@@ -0,0 +1,199 @@
...
+				registry.select( STORE_NAME ).getNewIdeas( customOptions );
+				await untilResolved( registry, STORE_NAME ).getNewIdeas( customOptions );
+
+				const newIdeas = registry.select( STORE_NAME ).getNewIdeas( customOptions );
+
+				expect( fetchMock ).toHaveFetchedTimes( 1 );
+				expect( newIdeas ).toEqual( fixtures.newIdeas.slice( 0, 3 ) );
+			} );
+
+			it( 'only fetches once even with different options passed', async () => {
```
tofumatt, line 144:
```suggestion
			it( 'only fetches once even with different options are passed', async () => {
```

includes/Modules/Idea_Hub.php
```diff
@@ -161,9 +161,65 @@ protected function create_data_request( Data_Request $data ) {
 					return null;
 				};
 			case 'GET:new-ideas':
-				// @TODO implementation
+				// @TODO: Implement this with the real API endpoint.
 				return function() {
-					return null;
+					return array(
+						array(
```
tofumatt, line 167: I guess we can't load in the fixture file here instead of duplicating it, eh? (I'm assuming it isn't included with the bundled plugin.)

Not a big deal, just sort of a bummer to need to duplicate it! Though I guess the work is already done and it's temporary, so best to leave well-enough alone 🙂

tofumatt (changes requested):
Looks good to me—I have a few readability adjustments for the code and the test spec names, but that's about it.

Mostly I think phrasing like "correctly handles X" isn't as helpful a test spec as "X causes Y to happen". ("Correctly" is a bit of an unneeded adjective given the point of tests is to outline the correct behaviour 😅 ) I made suggestions for a number of test specs so they outline what the intended effect is rather than only describing the test input.

After those tweaks though this should be good-to-go, the behaviour is all correct 👍🏻

includes/Modules/Idea_Hub.php
ivankruchkoff: We can consider it a motivation to get the real API in place quicker ;)

tofumatt (approved):

Issue #3175, opened 2021-04-16
Consolidate JS linting and JS tests workflows into one


aaemnnosttv:
## Feature Description

In #2969 we overhauled our GitHub actions to be much more selective with when each of them runs to be more efficient with build time. This issue is intended to further enhance that by combining JS linting and tests into a single workflow so that JS tests don't run if there is a linting failure.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The current `js-tests.yml` and `js-css-lint.yml` workflows should be combined into a single workflow
* The JS tests should run only after JS linting has passed successfully
* CSS linting should continue to run first as it runs the fastest
* `paths` should be updated to combine rules from both as needed

## Implementation Brief

* Check PR https://github.com/google/site-kit-wp/pull/3302
* Check that  `js-tests.yml` and `js-css-lint.yml` have been merged together successfully (no paths missing)
* Check that workflow is running

### Test Coverage

* The same number of tests should be run as before.

### Visual Regression Changes

* N/A

## QA Brief

* Check that new job is working correctly on GitHub

## Changelog entry

* N/A

aaemnnosttv:
@tofumatt anything you want to add/change here?

tofumatt:
Nothing to add, that looks exactly right 👍🏻

tofumatt:
(Note: I'm leaving this assigned to me for Code Review as I have some feedback for the PR and having reviewed the IB think it makes sense, but we'll get to this in a later sprint or at least after everything in Sprint 49 is complete 🙂 )

johnPhillips:
QA ✅
New job seems to be running well

PR #3302, opened 2021-05-05
Consolidate JS linting and JS tests workflows into one

danielgent:
## Summary

Addresses issue #3175.

## Checklist


.github/workflows/js-css-lint-test.yml
```diff
@@ -18,6 +18,7 @@ on:
       - 'package-lock.json'
       - '!feature-flags.json'
       - '!composer.json'
+      - '!.eslintrc.json'
```
tofumatt, line 21: Looks like this was copied from the JS Tests YML, but we shouldn't ignore ESLint changes in this file anymore since this file is now responsible for running the ESLint task. 🙂 
```suggestion
```

.github/workflows/js-css-lint-test.yml
```diff
@@ -33,6 +34,7 @@ on:
       - '.stylelint*'
       - '**/package.json'
       - 'package-lock.json'
+      - '!.eslintrc.json'
```
tofumatt, line 37: Likewise here:
```suggestion
```

tofumatt (changes requested):
The `.github/workflows/js-css-lint.yml` path in `js-css-lint-test.yml` needs to be updated to use the new filename (`js-css-lint-test.yml`). That file is included in the paths used to trigger the workflow so that if the workflow is updated—but no other files are—the workflow is still run so we can verify any changes being made to it work.

tofumatt (approved):
Lovely! This should take a few minutes off any JS-related PR, but mainly reduces the number of jobs we're running. JS tests alone in most other PRs seem to take pretty much the same time to run as this combined lint + test job, but saves the minute of install/setup overhead.

And again: mainly we don't need to run them in parallel so other jobs should come up faster 🙂

PR #3309, opened 2021-05-06
Add React router

tofumatt:
## Summary

Addresses issue #2514. See #2733 for original PR.

This is big PR but I tried to keep tangential refactoring low. It comes from an original PR that was already a bit out-of-date when I took it over, but has been brought up-to-date with other changes to Settings. I've gone through it all locally—in the plugin and in Storybook—and I think it works (and it passes all of the E2E tests). I _think_ I've also removed any unused files but if you catch any "dead" files let me know 🙂 

## Relevant technical choices

Originally I didn't think this required too much refactoring, but the Storybook stories needed considerable work to get working, and there was also a lot of needless prop-passing and `setState` function passing in the `SettingsActiveModule` component that really wasn't needed after moving to the router approach and relying on `CORE_UI` for things.

I ended up removing all of the `onConfirm`, `onCancel` type functions and also removed things like `isOpen`/`isSaving` from props, instead using `CORE_UI` state to handle those variables. This simplified the components and made them more flexible, removing the need to pass the props around.

The stories (and our custom `render` function we use in Jest tests) all needed to be updated to use the router, and particularly to hide modules except the "intended" one in Storybook since Search Console is always enabled. See the `_hideExcept` prop for that. 

## Checklist


tofumatt:
(I'll need @felixarntz to set the CLA manually here.)

assets/js/components/Root/index.js
```diff
@@ -50,19 +51,21 @@ export default function Root( {
...
-							TODO: Replace `FeatureToursDesktop` with `FeatureTours`
-							once tour conflicts in smaller viewports are resolved.
-							@see https://github.com/google/site-kit-wp/issues/3003
-						*/ }
-						{ viewContext && <FeatureToursDesktop viewContext={ viewContext } /> }
-						{ dataAPIContext && (
-						// Legacy dataAPI support.
-							<CollectModuleData context={ dataAPIContext } args={ dataAPIModuleArgs } />
-						) }
+						<HashRouter>
```
aaemnnosttv, line 54: This seems to be adding a `#` to all URLs where Site Kit apps are rendered

image

image

Can we avoid this except for where it's actively used? If not, we probably need to move the `HashRouter` into the `SettingsApp` for now.

assets/js/components/settings/SettingsModules.js
```diff
@@ -0,0 +1,85 @@
...
+				<SettingsActiveModules _hideExcept={ _hideExcept } />
+			</Route>
+			<Route path="/connected-services">
+				<SettingsActiveModules _hideExcept={ _hideExcept } />
+			</Route>
+			<Route path="/connect-more-services">
+				<SettingsInactiveModules />
+			</Route>
+			<Route path="/admin-settings">
+				<SettingsAdmin />
```
aaemnnosttv, line 69: Something is different about the way this component is rendered now as the gap is missing between the sections: 
image

stories/module-analytics-settings.stories.js
```diff
@@ -202,7 +201,7 @@ storiesOf( 'Analytics Module/Settings', module )
 			permission: true,
 		}, { propertyID: existingTag.propertyID } );
 
-		return <Settings isOpen={ true } isEditing={ true } registry={ registry } />;
+		return <Settings registry={ registry } route="/connected-services/analytics" />;
```
aaemnnosttv, line 204: This was `isEditing={ true }` before but that seems to have been lost here.

stories/modules-tagmanager-settings.stories.js
```diff
@@ -185,7 +185,7 @@ storiesOf( 'Tag Manager Module/Settings', module )
 		registry.dispatch( STORE_NAME ).setInternalContainerID( container.containerId );
 		registry.dispatch( STORE_NAME ).receiveGetSettings( defaultSettings );
 
-		return <Settings isOpen isEditing registry={ registry } />;
+		return <Settings registry={ registry } route="/connected-services/tagmanager" />;
```
aaemnnosttv, line 188: Missing edit state.

stories/modules-tagmanager-settings.stories.js
```diff
@@ -195,7 +195,7 @@ storiesOf( 'Tag Manager Module/Settings', module )
 		registry.dispatch( STORE_NAME ).receiveGetAccounts( [] );
 		registry.dispatch( STORE_NAME ).receiveGetSettings( defaultSettings );
 
-		return <Settings isOpen isEditing registry={ registry } />;
+		return <Settings registry={ registry } route="/connected-services/tagmanager" />;
```
aaemnnosttv, line 198: Missing edit state.

stories/modules-tagmanager-settings.stories.js
```diff
@@ -206,7 +206,7 @@ storiesOf( 'Tag Manager Module/Settings', module )
 		registry.dispatch( STORE_NAME ).receiveGetAccounts( fixtures.accounts );
 		registry.dispatch( STORE_NAME ).receiveGetSettings( defaultSettings );
 
-		return <Settings isOpen isEditing registry={ registry } />;
+		return <Settings registry={ registry } route="/connected-services/tagmanager" />;
```
aaemnnosttv, line 209: Missing edit state.

stories/modules-tagmanager-settings.stories.js
```diff
@@ -368,7 +368,7 @@ storiesOf( 'Tag Manager Module/Settings/Secondary AMP', module )
 			ampContainerName: fixtures.getContainers.amp[ 0 ].name,
 		} );
 
-		return <Settings isOpen isEditing registry={ registry } />;
+		return <Settings registry={ registry } route="/connected-services/tagmanager" />;
```
aaemnnosttv, line 371: All missing edit state until here 😄

stories/settings.stories.js
```diff
@@ -82,13 +83,13 @@ storiesOf( 'Settings', module )
...
+		const history = createHashHistory();
+
 		return (
 			<WithTestRegistry callback={ setupRegistry } >
-				<div className="mdc-layout-grid__inner">
-					<div className="mdc-layout-grid__cell mdc-layout-grid__cell--span-12">
-						<SettingsActiveModules />
-					</div>
-				</div>
+				<Router history={ history }>
```
aaemnnosttv, line 90: If we're applying the Router in the `Root` we should probably add it with a global decorator in `preview.js`. If needed, an alternate `history` could be provided via `parameters` similar to `features`.

assets/js/components/settings/SettingsRenderer.test.js
```diff
@@ -1,136 +0,0 @@
-/**
```
aaemnnosttv, line 1: Why was this test deleted when `SettingsRenderer` was not?

assets/js/components/settings/SettingsModules.js
```diff
@@ -0,0 +1,85 @@
...
+		<Switch>
+			{ /* Settings Module Routes */ }
+			<Route path="/connected-services/:moduleSlug/:action">
+				<SettingsActiveModules _hideExcept={ _hideExcept } />
+			</Route>
+			<Route path="/connected-services/:moduleSlug">
+				<SettingsActiveModules _hideExcept={ _hideExcept } />
+			</Route>
+			<Route path="/connected-services">
+				<SettingsActiveModules _hideExcept={ _hideExcept } />
```
aaemnnosttv, line 63: The `_hideExcept` prop is a bit confusing and seems to only be used in Storybook. Can you elaborate on this?

assets/js/components/settings/SettingsActiveModule/index.test.js
```diff
@@ -0,0 +1,128 @@
...
+	// 	provides: [ 'foo', 'bar' ],
+	// 	setupComplete: false,
+	// 	slug: 'analytics',
+	// };
+
+	const SettingsModuleWithWrapper = () => (
+		<Switch>
+			<Route path={ [ '/connected-services/:moduleSlug/:action', '/connected-services/:moduleSlug', '/connected-services' ] }>
+				<SettingsActiveModule slug="analytics" />
+			</Route>
```
aaemnnosttv, line 48: Can we avoid duplicating the routes here and use a "catch-all" route instead? It seems like it will always render analytics anyways so I would think the `:moduleSlug` part isn't needed anyways 🤔

assets/js/components/settings/SettingsActiveModule/index.test.js
```diff
@@ -0,0 +1,128 @@
...
+	it( 'should display SettingsViewComponent when on module view route.', async () => {
+		history.push( '/connected-services/analytics' );
+
+		const { queryByTestID } = render( <SettingsModuleWithWrapper />, { history, registry } );
+
+		expect( queryByTestID( 'view-component' ) ).toBeInTheDocument();
+	} );
+
+	it( 'should display SettingsEditComponent when on module view route.', async () => {
+		history.push( '/connected-services/analytics/edit' );
```
aaemnnosttv, line 80:
```suggestion
	it( 'should display SettingsEditComponent when on module edit route.', async () => {
		history.push( '/connected-services/analytics/edit' );
```

assets/js/components/settings/SettingsActiveModule/index.test.js
```diff
@@ -0,0 +1,128 @@
...
+			name: 'Analytics',
+			active: true,
+			connected: true,
+			setupComplete: true,
+			SettingsEditComponent: () => <div data-testid="edit-component">edit</div>,
+			SettingsViewComponent: () => <div data-testid="view-component">view</div>,
+		} ] );
+	} );
+
+	it( 'should display SettingsViewComponent when on module view route.', async () => {
```
aaemnnosttv, line 71: Let's remove the full stops from the test names here since we don't do this anywhere else.

assets/js/components/settings/SettingsModules.test.js
```diff
@@ -0,0 +1,114 @@
...
+		expect( global.location.hash ).toEqual( '#/connect-more-services' );
+	} );
+
+	it( 'should redirect from #admin to #/admin-settings', async () => {
+		const coreUserTrackingSettingsEndpointRegExp = /^\/google-site-kit\/v1\/core\/user\/data\/tracking/;
+		const coreUserTrackingResponse = { status: 200, body: { enabled: false } };
+		fetchMock.getOnce( coreUserTrackingSettingsEndpointRegExp, coreUserTrackingResponse );
+		fetchMock.postOnce( coreUserTrackingSettingsEndpointRegExp, coreUserTrackingResponse );
+
+		await registry.dispatch( CORE_USER ).setTrackingEnabled( false );
```
aaemnnosttv, line 74: This part looks like it could be simplified as it doesn't look particularly relevant to the test.

Can we use `muteFetch` instead?

assets/js/components/settings/SettingsModules.test.js
```diff
@@ -0,0 +1,114 @@
...
+
+	it( 'should redirect from #settings to #/connected-services', async () => {
+		history.push( '/settings' );
+
+		render( <SettingsModules />, { history, registry } );
+
+		expect( global.location.hash ).toEqual( '#/connected-services' );
+	} );
+
+	it( 'should redirect from #settings/analytics/view to #/connected-services/analytics', async () => {
```
aaemnnosttv, line 91: We should probably name these independent of the module used for testing.
```suggestion
	it( 'should redirect from #settings/:moduleSlug/view to #/connected-services/:moduleSlug', async () => {
```

assets/js/components/settings/SettingsModules.test.js
```diff
@@ -0,0 +1,114 @@
...
+		history.push( '/settings/analytics/view' );
+
+		render( <SettingsModules />, { history, registry } );
+
+		expect( global.location.hash ).toEqual( '#/connected-services/analytics' );
+	} );
+
+	it( 'should redirect from #settings/adsense/edit to #/connected-services/adsense/edit', async () => {
```
aaemnnosttv, line 99: See above.

assets/js/components/Link.js
```diff
@@ -21,52 +21,118 @@
...
+		return getNonIdenticalLabel();
+	};
+
+	const type = getType();
+	const LinkComponent = getLinkComponent( type );
+	const ariaLabel = getAriaLabel( type );
+
 	return (
-		<SemanticLink
+		<LinkComponent
```
aaemnnosttv, line 134: There's quite a few changes here and it seems like the component is getting more complex. Are there any tests we should add to `Link.test.js`?

aaemnnosttv (changes requested):
Thanks @tofumatt ! There's a lot here but I've left some initial comments for you.

All in all it seems to work great, although there is a bit of a concern with it adding hashes to any screen where a Site Kit app is rendered.

I also noticed some of the VRT tests which have very slight changes but still odd that those would have changed; do you know why?

assets/js/components/settings/SettingsActiveModules.js
```diff
@@ -24,26 +24,35 @@ import PropTypes from 'prop-types';
...
 	const [ initialActiveSlugs, setInitialActiveSlugs ] = useState();
 	const modules = useSelect( ( select ) => select( CORE_MODULES ).getModules() );
+
+	if ( _hideExcept ) {
+		Object.keys( modules ).forEach( ( moduleSlug ) => {
+			if ( _hideExcept.includes( moduleSlug ) ) {
+				return;
+			}
+			modules[ moduleSlug ].active = false;
+			modules[ moduleSlug ].forceActive = false;
```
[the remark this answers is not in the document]
tofumatt: It won't mutate the state, but this is something I overlooked in code review! It's a bit odd, I'll tidy it up.

assets/js/components/settings/SettingsRenderer.test.js
tofumatt: The tests contained here were updated and moved to `SettingsActiveModule`; the tests here were testing behaviour now more specific to `SettingsActiveModule`, so it made sense to move them there.

assets/js/components/settings/SettingsModules.js
tofumatt: It was kinda a hack, and I figured out a _slightly_ better approach in d778e6b404670e446c394e925c6e7a77e0caac94, let me know what you think 🙂

assets/js/components/settings/SettingsModules.test.js
tofumatt: Using `muteFetch` seems to throw errors because we need to mock a get and post... we need two mocks and only having one fails. But calling it twice also fails because it sets up the same mock API endpoint twice.

assets/js/components/Root/index.js
tofumatt: Thanks for spotting that! It didn't happen in Storybook strangely, so I fixed it here AND tweaked Storybook so it should fail if that were to happen in the future.

tofumatt: Yeah, I moved it for now because realistically we should be keeping it scoped to there for now 👍🏻

assets/js/components/settings/SettingsModules.js
tofumatt: Thanks for spotting that! It didn't happen in Storybook strangely, so I fixed it here AND tweaked Storybook so it should fail if that were to happen in the future.

stories/settings.stories.js
tofumatt: It already exists in https://github.com/google/site-kit-wp/blob/9d9c1c3a57455927c9e5773eb51cb7df2630f4ca/tests/js/utils.js#L105, but I've moved it from `<Root>` for now anyway, so I think this is better for the time being.

assets/js/components/settings/SettingsActiveModule/index.test.js
tofumatt: What do you mean by a "catch-all" route? I don't quite think I follow 😅

assets/js/components/Link.js
tofumatt: Looking at it further, it's mostly a refactoring of what was here before, but I've added tests for the component type and a few for the ARIA label behaviour.

tofumatt:
@googlebot I consent.

aaemnnosttv:
@tofumatt would you please resolve the merge conflicts?

assets/js/components/settings/SettingsModules.js
aaemnnosttv: I much prefer the weirdness to live in Storybook than using the previous prop 👍

assets/js/components/Link.test.js
```diff
@@ -48,10 +48,58 @@ describe( 'Link', () => {
...
+			expect( container.firstChild ).toHaveAttribute( 'aria-label', 'foo' );
+		} );
+
+		it( 'does not set the ARIA label based on children when there is no suffix', () => {
+			const { container } = render( <Link>text content</Link> );
+
+			expect( container.firstChild ).not.toHaveAttribute( 'aria-label' );
+		} );
+
+		it( 'creates an <a> attribute by default', () => {
```
aaemnnosttv, line 81: I think "tag" or "element" may be more appropriate here?
```suggestion
		it( 'creates an <a> tag by default', () => {
```

assets/js/components/Link.test.js
```diff
@@ -48,10 +48,58 @@ describe( 'Link', () => {
...
+			expect( container.firstChild.tagName ).toEqual( 'A' );
+		} );
+
+		it( 'creates an <a> attribute when using React Router', () => {
+			const { container } = render( <Link to="/">text content</Link> );
+
+			expect( container.firstChild.tagName ).toEqual( 'A' );
+		} );
+
+		it( 'creates a <button> attribute when `onClick` is set', () => {
```
aaemnnosttv, line 99:
```suggestion
		it( 'creates a <button> tag when `onClick` is set', () => {
```

assets/js/components/settings/SettingsActiveModule/index.test.js
aaemnnosttv: Sorry, I think I misunderstood. It looked like we were duplicating a route definition here that could render different modules but would always render Analytics regardless. I see that the parameters are significant to the behavior of the component now (via `useParams`), it just wasn't immediately obvious before.

assets/js/components/settings/SettingsAdmin.js
```diff
@@ -68,9 +67,9 @@ const SettingsAdmin = () => {
 	}, [ isUserInputCompleted ] );
 
 	return (
-		<Fragment>
+		<div className="mdc-layout-grid__inner">
```
aaemnnosttv, line 70: This looks like it should be a `Row` instead 🙂

assets/js/components/settings/SettingsApp.js
```diff
@@ -21,83 +21,34 @@
...
-	const handleTabUpdate = ( tabIndex ) => {
-		const newActiveTabID = SettingsApp.tabIDsByIndex[ tabIndex ];
-		setActiveTabID( newActiveTabID );
-		setHash( hashFrom( newActiveTabID ) );
-	};
+	const location = useLocation();
+	// Prevent pushing to hash history if it would send you to the same URL.
+	// (Without this React Router will trigger a warning.)
+	const shouldReplaceHistory = ( path ) => false && basePath === path;
+	const [ , basePath ] = location?.pathname.split( '/' );
```
aaemnnosttv, line 50: There is a conditional chain here (would `location` ever not be set?) but it always destructures a list which would fail if the chain short-circuited.

assets/js/components/settings/SettingsModules.test.js
aaemnnosttv: Ah, that makes sense. I think we could avoid the mock for `get` though by calling `dispatch('core/user').receiveGetTracking(...)`. Then we'd only need to mock the post. Not a big deal though.

aaemnnosttv (commented):
Thanks @tofumatt – this looks good to me, just a few small details left and a merge conflict to resolve and then I think this is good to go!

assets/js/components/settings/SettingsApp.js
tofumatt: Hmm, no, `location` should always be set. Not sure why I did that, but removing the conditional chain here should be safe.

assets/js/components/settings/SettingsAdmin.js
tofumatt: Nice catch, thanks 👍🏻

tofumatt (commented):
Thanks! Merge conflicts resolved; this should be set now 👍🏻

aaemnnosttv (approved):
Thanks @tofumatt – great work! 🎉

Issue #3157, opened 2021-04-15
Implement infrastructure for storing Idea Hub idea information with WordPress posts


felixarntz:
Since the Idea Hub module will allow creating WordPress posts based on Idea Hub ideas, it should be possible to store an idea's full information within a post. [See more information on the overall approach](https://docs.google.com/document/d/1mIBLvv-v2Y3OMiHsOn5kMu3Fnt-wAqbWDnAiuraTy0k/edit).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new `Google\Site_Kit\Modules\Idea_Hub\Post_Idea_Name` class should be implemented which extends `Google\Site_Kit\Core\Storage\Post_Meta_Setting`.
    * Its meta key should be `googlesitekitpersistent_idea_name`.
    * Its type should be a string (it's expecting a slug).
* A new `Google\Site_Kit\Modules\Idea_Hub\Post_Idea_Text` class should be implemented which extends `Google\Site_Kit\Core\Storage\Post_Meta_Setting`.
    * Its meta key should be `googlesitekitpersistent_idea_text`.
    * Its type should be a string (it's expecting a human-readable label).
* A new `Google\Site_Kit\Modules\Idea_Hub\Post_Idea_Topics` class should be implemented which extends `Google\Site_Kit\Core\Storage\Post_Meta_Setting`.
    * Its meta key should be `googlesitekitpersistent_idea_topics`.
    * Its type should be an array (it's expecting a list of objects / associative arrays, each with `mid` and `display_name` fields).
    * Its sanitize callback should ensure every entry within the array is an associative array with `mid` and `display_name` string fields.
* The `Idea_Hub` class should register the new `Post_Meta_Setting`s implemented.
* The `Idea_Hub` class should receive two new protected methods:
    * `set_post_idea( int $post_id, array $idea ) : bool`: Stores metadata in the given post based on the idea passed, relying on the 3 `Post_Meta_Setting`s. `$idea` is expected to be an associative array for a single idea (see #3155 for the exact format of such an idea).
    * `get_post_idea( int $post_id ) : array | null`: Returns an idea array based on the post metadata (relying on the 3 `Post_Meta_Setting`s) stored in the post, or returns `null` if any of the 3 pieces of data is not stored. The returned idea format should match the idea format returned by the API (see #3155 for the exact format of such an idea).

## Implementation Brief

Following #3154,

[...]

PR #3320, opened 2021-05-07
Enhancement/3157 save idea info

eugene-manuilov:
## Summary

Addresses issue #3157

## Checklist


tests/phpunit/includes/TestCase.php
```diff
@@ -248,6 +248,12 @@ protected function assertPostMetaExists( $post_id, $meta_key ) {
 		);
 	}
 
+	protected function assertPostMetaHasValue( $post_id, $meta_key, $meta_value ) {
+		$meta = $this->queryPostMeta( $post_id, $meta_key );
+		$this->assertNotNull( $meta );
+		$this->assertEquals( $meta_value, $meta['meta_value'], "Failted to assert that post $post_id has \"$meta_key\" meta with \"$meta_value\" value." );
```
tofumatt, line 254:
```suggestion
		$this->assertEquals( $meta_value, $meta['meta_value'], "Failed to assert that post $post_id has \"$meta_key\" meta with \"$meta_value\" value." );
```

tofumatt: Seems like a misspelling here, but then it _did_ pass. What's causing this to emit "failted"? I couldn't find it in our code, but is it worth noting why that… typo appears? 🤷🏻‍♂️

tofumatt (changes requested):
Seems good, but I'm curious about a misspelling in the test case.

Also: can you add a QA brief? I know in this case it's essentially "Make sure tests pass", but better to leave a minimal one than an empty one 🙂

tests/phpunit/includes/TestCase.php
eugene-manuilov: @tofumatt the last parameter is just a friendly message that will be displayed if the first two parameters don't equal.

tofumatt (approved):

Issue #3244, opened 2021-04-28
Implement submitChanges and validateCanSubmitChanges logic for GA4 store


felixarntz:
The GA4 store in JS `modules/analytics-4` should receive a `submitChanges` and `validateCanSubmitChanges` implementation to handle GA4 data submissions correctly. Since for now all GA4 behavior will be controlled through existing Analytics UI, it should also be incorporated into the similar action/selector respectively of the `modules/analytics` store.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `modules/analytics-4` store should receive its own implementation of `submitChanges` and `validateCanSubmitChanges` (via the `createModuleStore` call), with these functions themselves being implemented in `assets/js/modules/analytics-4/datastore/settings.js` (similar to how it is for the `modules/analytics` store). Both of the new function implementations should follow the same patterns used in the similar functions in the `modules/analytics` store.
* The `submitChanges` function should do the following:
    1. If the store's `getPropertyID` is `PROPERTY_CREATE`, a new GA4 property should be created using the `createProperty` action. Afterwards, the new property should be set.
    2. If the store's `getWebDataStreamID` is not valid (a new utility function `isValidWebDataStreamID` should be added to `assets/js/modules/analytics-4/utils/validation.js` for this purpose), a new GA4 web data stream should be created using the `createWebDataStream` action. Afterwards, the new web data stream ID and measurement ID from it should be set.
    3. Then, if settings have changed, the changes should be saved using `saveSettings`, and on success the `analytics-4` cache should be invalidated (no need to include the legacy data API cache here, since it isn't used for any `analytics-4` API requests).
    4. If any error occurs during the above, it should not be returned by the function - the function should in that case simply return the success response (empty object) early. This is necessary since the GA4 APIs are only in alpha, and errors shouldn't prevent any user-facing processes to fail/be blocked, also given that GA4 is not essential at this point to use Site Kit.
* The `validateCanSubmitChanges` function should do the following:
    1. Return early if the GA4 API is not working (i.e. `isAdminAPIWorking === false`). This is necessary to include since the GA4 APIs are only in alpha, and errors shouldn't prevent any user-facing processes to fail/be blocked, also given that GA4 is not essential at this point to use Site Kit.
    2. Fail if the store currently `isDoingSubmitChanges`.
    3. Fail if for the store `! haveSettingsChanged`.
[...]

PR #3325, opened 2021-05-10
Enhancement/3244 ga4 submit changes

eugene-manuilov:
## Summary

Addresses issue #3244 

## Checklist


assets/js/modules/analytics-4/datastore/settings.js
```diff
@@ -0,0 +1,97 @@
...
+	if ( propertyID === PROPERTY_CREATE ) {
+		const accountID = select( MODULES_ANALYTICS ).getAccountID();
+		const { response: property, error } = await dispatch( STORE_NAME ).createProperty( accountID );
+		if ( error ) {
+			return { error };
+		}
+
+		propertyID = property._id;
+		await dispatch( STORE_NAME ).setPropertyID( propertyID );
+		await dispatch( STORE_NAME ).setWebDataStreamID( '' );
```
tofumatt, line 51: Why do we need to set the webstream data ID as an empty string here? We select it and then check to see if it's valid, below, so presumably a `null`/`undefined` value or similar won't be valid, right?

Or does it need to be a string, but an invalid string, in order for the `createWebDataStream` path below to work? If so, it's worth adding a comment here to clarify that. At any rate it isn't obvious to me why this line is here.

tofumatt (changes requested):
Looks good to me, but I have one question about the `setWebDataStreamID()` call.

After that's addressed this should be good to merge. 👍🏻

assets/js/modules/analytics-4/datastore/settings.js
eugene-manuilov: We need to set an invalid web data stream ID to make sure a new data stream will be created. I have added a comment to explain it.

tofumatt (approved):
Thanks for the explanation on the empty string for the web stream ID! Looks good 👍🏻

Issue #3243, opened 2021-04-28
Update UA and GA4 dropdowns behavior to only show once parent is specified


felixarntz:
Currently, the Analytics dropdowns for account, property and profile/view always display, with the latter two being disabled until the respective parent entity has been specified. In relation to the upcoming GA4 setup flow, this should be modified so that the user only sees the dropdowns once they actually allow a selection to be made. This should be changed for the existing infrastructure though too, i.e. it will affect Site Kit production behavior already today.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The Analytics (UA) `PropertySelect` component and the Analytics 4 (GA4) `PropertySelect` component should be adjusted so that they return `null` as long as no valid account ID has been specified yet (i.e. based on `! isValidAccountID`).
    * They should otherwise maintain their loading state as is, returning `null` should only happen as long as no account ID is specified yet.
* The Analytics (UA) `ProfileSelect` component should be adjusted so that it returns `null` as long as no valid account ID or no valid (UA) property ID has been specified yet (i.e. based on `! isValidAccountID` and `! isValidPropertyID`).
* As usual, it must be ensured that the Storybook coverage for these components still works as expected, with any potentially now irrelevant stories for them (e.g. disabled state that now is no longer possible) being removed.

## Implementation Brief

For the three dropdowns:
* Analytics (UA) `PropertySelect`
* Analytics 4 (GA4) `PropertySelect` 
* Analytics (UA) `ProfileSelect`

Return null if there is no account ID

And also for the Analytics (UA) `ProfileSelect`

Return null is there is no property ID

#### Storybook changes

The Analytics section has two stories:  `Account Property Profile Select (none selected)` and `Account Property Profile Select (all selected)`
* Remove `"all selected"` story
* Prefill redux with data for every account in the existing accounts fixture: _(By dispatching `properties` and `profiles` for each account id)_
* The story should be interactive

### Test Coverage

No existing tests currently fail because of the above change

Add extra tests for the three dropdowns to assert that null is returned when
* no accountID for all three
* no propertyID for the Profile Select

### Visual Regression Changes

Checked and there should be no visual regression changes with above changes

## QA Brief

The UA `PropertySelect` and the `ProfileSelect` changes can easily be seen on storybook
https://google.github.io/site-kit-wp/storybook/pull/3330/?path=/story/analytics-module--account-property-profile-select

Here you can observe the changes are made as per the AC

[...]

PR #3330, opened 2021-05-10
Update UA and GA4 dropdowns behavior to only show once parent is specified

danielgent:
## Summary

Addresses issue #3243

## Checklist


assets/js/modules/analytics-4/components/common/PropertySelect.test.js
```diff
@@ -75,6 +76,27 @@ describe( 'PropertySelect', () => {
...
+
+				dispatch( STORE_NAME ).receiveGetProperties( [], { accountID } );
+				dispatch( STORE_NAME ).finishResolution( 'getProperties', [ accountID ] );
+			},
+		} );
+
+		expect( container ).toBeEmptyDOMElement();
+
+		// in order to not get act errors need to do something as DOM is empty at the start and end of this test AND wp-data does lots of stuff
+		await act( () => registry.dispatch( MODULES_ANALYTICS ).setAccountID( accountID ) );
```
danielgent, line 97: in order to not get act errors need to do something as DOM is empty at the start and end of this test (so there's nothing to wait for in the DOM as per react-testing-library best practices), **AND** wp-data is doing lots of stuff

stories/module-analytics.stories.js
```diff
@@ -62,54 +62,30 @@ function SetupWrap( { children } ) {
...
-				// eslint-disable-next-line sitekit/acronym-case
-				accountID: profiles[ 0 ].accountId,
-				// eslint-disable-next-line sitekit/acronym-case
-				propertyID: profiles[ 0 ].webPropertyId,
-				// eslint-disable-next-line sitekit/acronym-case
-				internalWebPropertyID: profiles[ 0 ].internalWebPropertyId,
-				profileID: profiles[ 0 ].id,
-			} );
+			// TODO - should instead have new static fixtures? is this too confusing? NOT tests though... could scramble names and generate fixtures dynamically?
+			for ( const account of accounts ) {
```
danielgent, line 73: Instead, should I make new static fixtures?  (they could get pretty big though!)

tests/e2e/specs/modules/analytics/setup-with-account-no-tag.test.js
```diff
@@ -157,6 +158,7 @@ describe( 'setting up the Analytics module with an existing account and no exist
 			// See the selects populate
 			await expect( page ).toMatchElement( '.mdc-select__selected-text', { text: /test account a/i } );
 			await expect( page ).toMatchElement( '.mdc-select__selected-text', { text: /test property x/i } );
+			// NOTE - according to spec this should be hidden. Same in storybook. Displays when account is selected. TO INVESTIGATE
 			await expect( page ).toMatchElement( '.mdc-select__selected-text', { text: /test profile x/i } );
```
danielgent, line 162: Should this profile select be hidden? I assumed so (and then I would change this line as I did with line 150)

Is the account Id and profile Id being set automatically when the account is selected?

danielgent: See here the interactive demo in storybook https://google.github.io/site-kit-wp/storybook/pull/3330/?path=/story/analytics-module--account-property-profile-select

stories/module-analytics.stories.js
eugene-manuilov: I think there is no need to create a new fixture for it. We just need to take and load the first three items from the accounts list. Then load some properties for the first two accounts of selected accounts. And some profiles for the first account to simulate all possible situations.

```javascript
const { accounts: allAccounts, properties, profiles } = fixtures.accountsPropertiesProfiles;
const accounts = allAccounts.slice( 0, 3 );

dispatch( STORE_NAME ).receiveGetAccounts( accounts );

dispatch( STORE_NAME ).receiveGetProperties( properties.map( ( property ) => { ...property, accountId: accounts[0].id } ), { accountID: accounts[ 0 ].id } );
dispatch( STORE_NAME ).receiveGetProfiles( profiles.map( ( profile ) => { ...profile, accountId: accounts[ 0 ].id } ), { accountID: accounts[ 0 ].id, propertyID: properties[ 0 ].id } );

dispatch( STORE_NAME ).receiveGetProperties( properties.map( ( property ) => { ...property, accountId: accounts[1].id } ), { accountID: accounts[ 1 ].id } );
dispatch( STORE_NAME ).receiveGetProfiles( [], { accountID: accounts[ 1 ].id, propertyID: properties[ 0 ].id } );

dispatch( STORE_NAME ).receiveGetProperties( [], { accountID: accounts[ 2 ].id } );
dispatch( STORE_NAME ).receiveGetProfiles( [], { accountID: accounts[ 2 ].id, propertyID: properties[ 0 ].id } );
```

I hope this will be sufficient to see the new functionality in the storybook :thinking:

assets/js/modules/analytics/components/common/PropertySelect.test.js
```diff
@@ -95,6 +95,30 @@ describe( 'PropertySelect', () => {
...
+				dispatch( STORE_NAME ).finishResolution( 'getAccounts', [] );
+
+				dispatch( STORE_NAME ).receiveGetProperties( [], { accountID } );
+				dispatch( STORE_NAME ).finishResolution( 'getProperties', [ accountID ] );
+			},
+		} );
+
+		expect( container ).toBeEmptyDOMElement();
+
+		await act( () => registry.dispatch( STORE_NAME ).setAccountID( ACCOUNT_CREATE ) );
```
eugene-manuilov, line 119: Once we set the accountID, we need to check that the PropertySelect element becomes visible.

assets/js/modules/analytics/components/common/ProfileSelect.test.js
```diff
@@ -130,6 +130,56 @@ describe( 'ProfileSelect', () => {
...
+				dispatch( STORE_NAME ).finishResolution( 'getProperties', [ accountID ] );
+
+				dispatch( STORE_NAME ).receiveGetProfiles( [], { accountID, propertyID } );
+				dispatch( STORE_NAME ).finishResolution( 'getProfiles', [ accountID, propertyID ] );
+			},
+		} );
+
+		expect( container ).toBeEmptyDOMElement();
+
+		await act( () => registry.dispatch( STORE_NAME ).setAccountID( 'pub-12345678' ) );
```
eugene-manuilov, line 154: Same here, but we also need to set propertyID and check that ProfileSelect element becomes visible.

assets/js/modules/analytics/components/common/ProfileSelect.test.js
```diff
@@ -130,6 +130,56 @@ describe( 'ProfileSelect', () => {
...
+	} );
+
+	it( 'should not render if property ID is not set.', async () => {
+		const { container, registry } = render( <ProfileSelect />, {
+			setupRegistry: ( { dispatch } ) => {
+				const accountID = fixtures.accountsPropertiesProfiles.profiles[ 0 ].accountId; // eslint-disable-line sitekit/acronym-case
+				const propertyID = fixtures.accountsPropertiesProfiles.profiles[ 0 ].webPropertyId; // eslint-disable-line sitekit/acronym-case
+
+				dispatch( STORE_NAME ).setSettings( {} );
+				dispatch( STORE_NAME ).setAccountID( accountID );
```
eugene-manuilov, line 164: Just combine it into one action:

```javascript
dispatch( STORE_NAME ).setSettings( { accountID } );
```

assets/js/modules/analytics/components/common/ProfileSelect.test.js
```diff
@@ -130,6 +130,56 @@ describe( 'ProfileSelect', () => {
...
+
+				dispatch( STORE_NAME ).receiveGetProfiles( [], { accountID, propertyID } );
+				dispatch( STORE_NAME ).finishResolution( 'getProfiles', [ accountID, propertyID ] );
+			},
+		} );
+
+		expect( container ).toBeEmptyDOMElement();
+
+		// in order to not get act errors need to do something as DOM is empty at the start and end of this test AND wp-data does lots of stuff
+		await act( () => registry.dispatch( STORE_NAME ).setAccountID( 'pub-12345678' ) );
```
eugene-manuilov, line 180: We should set the `propertyID` instead, because the `accountID` setting is already set in the setup function. Plus, we need to check that the ProfileSelect element becomes visible after doing these changes.

eugene-manuilov (changes requested):
Thanks, @danielgent. It's a good start. Added some comments for you. Please, take a look.

assets/js/modules/analytics-4/components/common/PropertySelect.test.js
```diff
@@ -90,9 +90,9 @@ describe( 'PropertySelect', () => {
 			registry.dispatch( MODULES_ANALYTICS ).setAccountID( ACCOUNT_CREATE );
 			registry.dispatch( STORE_NAME ).finishResolution( 'getProperties', [ ACCOUNT_CREATE ] );
 		} );
-		// ACCOUNT_CREATE is an invalid accountID (but valid selection), so ensure the select IS currently disabled.
-		expect( selectWrapper ).toHaveClass( 'mdc-select--disabled' );
-		expect( selectedText ).toHaveAttribute( 'aria-disabled', 'true' );
+
+		// ACCOUNT_CREATE is an invalid (but valid selection), so ensure the select is not rendered
+		expect( container ).toBeEmptyDOMElement();
```
danielgent, line 95: Just to note @eugene-manuilov that the changed functionality of using ` isValidAccountID` and `isValidPropertyID` results in this test changing

Just checking that's ok with you! :+1:

assets/js/modules/analytics/components/common/PropertySelect.test.js
danielgent: So (as per my last comment) I'm reusing the existing tests but changing them to expect hidden

The current tests go from visible to hidden but I'll also dispatch extra actions to check it toggles back to visible :+1:

stories/module-analytics.stories.js
danielgent: I tried this out and it worked @eugene-manuilov  :exploding_head: 

I don't understand how though.  Surely when I select the third account on the list we should get errors as `wp-data` starts fetching properties for that account :thinking: 

Ah well, I must be missing something deeper in the code but I've already spent some time looking so I'll just trust you :smile:

danielgent:
Due to these changes there is now an e2e test which fails when the assertion is both true *and* false (see last two commits)

eugene-manuilov:
@googlebot I consent.

assets/js/modules/analytics/components/common/ProfileSelect.test.js
```diff
@@ -130,32 +130,55 @@ describe( 'ProfileSelect', () => {
...
 		await act( () => registry.dispatch( STORE_NAME ).setAccountID( '0' ) );
 
-		// An empty accountID is invalid, so ensure the select IS currently disabled.
+		// An empty accountID is invalid, so ensure the select is not rendered.
+		expect( container ).toBeEmptyDOMElement();
+
+		// eslint-disable-next-line sitekit/acronym-case
+		await act( () => registry.dispatch( STORE_NAME ).setAccountID( fixtures.propertiesProfiles.profiles[ 0 ].accountId ) );
+
+		// now select should be visible again
```
tofumatt, line 153:
```suggestion
		// A valid account ID was set, so the select should be visible.
```

assets/js/modules/analytics/components/common/ProfileSelect.test.js
```diff
@@ -130,32 +130,55 @@ describe( 'ProfileSelect', () => {
...
+			setupRegistry( { dispatch } ) {
+				setupRegistry( { dispatch } );
+				dispatch( STORE_NAME ).finishResolution( 'getProperties', [ '0' ] );
+			},
+		} );
+
+		const validAccountID = registry.select( STORE_NAME ).getAccountID();
+		const validPropertyID = registry.select( STORE_NAME ).getPropertyID();
+
+		// A valid accountID is provided, so ensure it is not currently disabled.
```
tofumatt, line 168:
```suggestion
		// A valid accountID is provided, so the select component should not be disabled.
```

assets/js/modules/analytics/components/common/ProfileSelect.test.js
```diff
@@ -130,32 +130,55 @@ describe( 'ProfileSelect', () => {
...
 
-		// The accountID is valid, but an empty propertyID is invalid, so ensure the select IS currently disabled.
-		expect( container.querySelector( '.googlesitekit-analytics__select-profile' ) )
-			.toHaveClass( 'mdc-select--disabled' );
+		// The accountID is valid, but an empty propertyID is invalid, so ensure the select is not rendered.
+		expect( container ).toBeEmptyDOMElement();
+
+		await act( () => registry.dispatch( STORE_NAME ).setPropertyID( validPropertyID ) );
+
+		// now select should be visible again
```
tofumatt, line 180:
```suggestion
		// After setting a valid property ID, the select should be visible.
```

assets/js/modules/analytics/components/common/PropertySelect.test.js
```diff
@@ -95,14 +95,16 @@ describe( 'PropertySelect', () => {
 			.toHaveClass( 'mdc-select--disabled' );
 	} );
 
-	it( 'should be disabled when in the absence of an valid account ID.', async () => {
+	it( 'should not render in the absence of an valid account ID.', async () => {
```
tofumatt, line 98: Again: I know this wasn't your wording but let's take the time to simplify it:

```suggestion
	it( 'should not render if account ID is invalid', async () => {
```

assets/js/modules/analytics/components/common/PropertySelect.test.js
```diff
@@ -111,9 +113,13 @@ describe( 'PropertySelect', () => {
 
 		await act( () => registry.dispatch( STORE_NAME ).setAccountID( ACCOUNT_CREATE ) );
 
-		// ACCOUNT_CREATE is an invalid (but valid selection), so ensure the select IS currently disabled.
-		expect( selectWrapper ).toHaveClass( 'mdc-select--disabled' );
-		expect( selectedText ).toHaveAttribute( 'aria-disabled', 'true' );
+		// ACCOUNT_CREATE is an invalid (but valid selection), so ensure the select is not rendered
```
tofumatt, line 116: Similar to the previous comment, let's clarify this and add at least what seems to be a missing noun 🙂

assets/js/modules/analytics/components/common/PropertySelect.test.js
```diff
@@ -111,9 +113,13 @@ describe( 'PropertySelect', () => {
...
+		expect( container ).toBeEmptyDOMElement();
+
+		await act( () => registry.dispatch( STORE_NAME ).setAccountID( accountID ) );
+
+		// now select should be visible again
```
tofumatt, line 121:
```suggestion
		// A valid account ID was set, so the select should be visible.
```

stories/module-analytics-4.stories.js
```diff
@@ -43,55 +43,39 @@ function SetupWrap( { children } ) {
...
-			createWebDataStream,
-			properties,
-			webDataStreams,
-		} = fixtures;
-		const accountID = createProperty._accountID;
-		const propertyID = createWebDataStream._propertyID;
+			const accountA = {
+				id: '1000',
+				name: 'Account A',
+			};
```
tofumatt, line 55: Why does this need to be called `accountA`? It's the only account for this test, right?

```suggestion
			const account = {
				id: '1000',
				name: 'Account A',
			};
```

Same goes for `propertyA` below.

stories/module-analytics.stories.js
```diff
@@ -66,54 +66,42 @@ function SetupWrap( { children } ) {
...
+
+			const propertyB = {
+				id: 'UA-2001-1',
+				name: 'Property B',
+			};
+
+			const profileA = {
+				id: '3000',
+				name: 'Profile A',
+			};
```
tofumatt, line 89: Here it makes sense to label these, but I find these single-letter differentiators a bit tough to scan.

Let's use more descriptive labels—`propertyOne` and `propertyTwo` works, or `firstProperty` and `secondProperty`. Equally `properties = [ {}, {} ]` would work 🙂 

I know it seems nitpicky, but these end up being equivalent to single-letter-variables like `e` or `a` because that's the only differentiator between the two variables, which isn't as easy-to-scan as my suggestions 😅

tests/e2e/specs/modules/analytics/setup-with-account-no-tag.test.js
```diff
@@ -183,9 +184,10 @@ describe( 'setting up the Analytics module with an existing account and no exist
 
 					// See the selects populate
 					await expect( page ).toMatchElement( '.mdc-select__selected-text', { text: /test account a/i } );
-					// Property and profile dropdowns should select "Set up a new property/view" options because there is no property associated with the current reference URL.
+					// Property dropdown should select "Set up a new property" options because there is no property associated with the current reference URL.
```
tofumatt, line 187:
```suggestion
					// Property dropdown should select "Set up a new property" option because there is no property associated with the current reference URL.
```

tofumatt (changes requested):
Overall looks good and the story works well when I tried it out using the link, thanks! 👍🏻 

I think there's a bit of wording and test tweaks to do here before we merge this, but after that should be good. After you've taken a look through just assign me for another review and it should be good 👍🏻

assets/js/modules/analytics-4/components/common/PropertySelect.test.js
```diff
@@ -90,9 +90,17 @@ describe( 'PropertySelect', () => {
 			registry.dispatch( MODULES_ANALYTICS ).setAccountID( ACCOUNT_CREATE );
 			registry.dispatch( STORE_NAME ).finishResolution( 'getProperties', [ ACCOUNT_CREATE ] );
 		} );
-		// ACCOUNT_CREATE is an invalid accountID (but valid selection), so ensure the select IS currently disabled.
-		expect( selectWrapper ).toHaveClass( 'mdc-select--disabled' );
-		expect( selectedText ).toHaveAttribute( 'aria-disabled', 'true' );
+
+		// ACCOUNT_CREATE is an invalid (but valid selection), so ensure the select is not rendered
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/analytics/components/common/PropertySelect.test.js
eugene-manuilov: Updated.

stories/module-analytics-4.stories.js
eugene-manuilov: Updated.

stories/module-analytics.stories.js
eugene-manuilov: Updated.

tofumatt (approved):
Looks great 👍🏻

tofumatt:
(Merging despite the Nightly E2E tests failing, as they appear unrelated—something to do with AMP + Nightly.)

Issue #1891, opened 2020-08-07
Allow ability to select what user roles are excluded


scottshefler:
## Feature Description

Currently only allowed to enable/disable excluding logged-in users. If you enable it this will also exclude front-end user logins for membership type sites. It's obviously very important to be able to track these types of logged in users. 

The plugin should loop through and return all user roles that are setup on the site with a checkbox option to select which roles to exclude.

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Site Kit should allow to disable tracking for only users that can create content (even if they are logged in with a WordPress account), e.g. to satisfy the tracking requirements for membership sites:
    * The Analytics `trackingDisabled` option needs to support a new possible value `contentCreators`. If that value is enabled but `loggedinUsers` is not, tracking should be disabled if the current user can `edit_posts` (which is a WordPress capability every role except subscriber has).
    * The respective Analytics UI in the module settings panel should be updated for the new possible value:
        * The toggle for `contentCreators` should appear right of the existing toggle for `loggedinUsers`.
        * The label for the new toggle should be _Users that can write posts_. The label for the existing toggle should be rephrased to _All logged-in users_.
        * If the `loggedinUsers` toggle is enabled, the `contentCreators` toggle should not be visible. Only if it is disabled, the `contentCreators` toggle should be visible.
    * Keep in mind the option is an array. While that is not really applicable right now, it should be maintained as such. Even though `loggedinUsers` technically includes `contentCreators`, that is not a big deal since it should simply take precedence.

## Implementation Brief

* Edit `includes/Modules/Analytics.php` and update the `is_tracking_enabled` method to also set `$disabled` to true if 'contentCreators' is in the option array and the user can edit posts ([`current_user_can( 'edit_posts' );`](https://developer.wordpress.org/reference/functions/current_user_can/))
* Edit `assets/js/modules/analytics/components/common/TrackingExclusionSwitches.js`
* Add a constant of `TRACKING_CONTENT_CREATORS` with the value `contentCreators`
* Add to and update the `trackingExclusionLabels` values
  *  `TRACKING_LOGGED_IN_USERS` : `All logged-in users`
  *  `TRACKING_CONTENT_CREATORS` : `Users that can write posts`
* Duplicate the `<Switch>` component and update the code of the new one to use `TRACKING_CONTENT_CREATORS`
[...]

PR #3333, opened 2021-05-11
Allow ability to select what user roles are excluded - UI fix (1891).

johnPhillips:
## Summary

Addresses issue #1891

## Relevant technical choices

I played around with the UI a bit here but settled for the simplest implementation I could think of to show the switches inline on desktop and tablet but stacked on mobile.
I was unsure where to put the new CSS. In the end `assets/sass/components/settings/_googlesitekit-settings-module.scss` seemed like the most logical place.

| Mobile  | Tablet, desktop |
| ------------- | ------------- |
| Screenshot 2021-05-11 at 12 10 28 | Screenshot 2021-05-11 at 12 10 47 |
| Screenshot 2021-05-11 at 12 10 14 | Screenshot 2021-05-11 at 12 11 11 |

## Checklist

### DIFF
--- a/assets/js/modules/analytics/components/common/TrackingExclusionSwitches.js
+++ b/assets/js/modules/analytics/components/common/TrackingExclusionSwitches.js
@@ -78,8 +78,8 @@ export default function TrackingExclusionSwitches() {
 			<legend className="googlesitekit-setup-module__text">
 				{ __( 'Exclude from Analytics', 'google-site-kit' ) }
 			</legend>
-			<div>
-				<div className="googlesitekit-analytics-trackingdisabled__inlineswitch">
+			<div className="googlesitekit-settings-module__inline-items">
+				<div className="googlesitekit-settings-module__inline-item">
 					<Switch
 						label={ trackingExclusionLabels[ TRACKING_LOGGED_IN_USERS ] }
 						checked={ trackingDisabled.includes( TRACKING_LOGGED_IN_USERS ) }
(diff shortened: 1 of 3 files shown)


assets/js/modules/analytics/components/common/TrackingExclusionSwitches.js
```diff
@@ -81,29 +78,26 @@ export default function TrackingExclusionSwitches() {
 			<legend className="googlesitekit-setup-module__text">
 				{ __( 'Exclude from Analytics', 'google-site-kit' ) }
 			</legend>
-			<Grid>
-				<Row>
-					<Cell lgSize={ 6 } mdSize={ 4 } smSize={ 4 }>
+			<div>
```
eugene-manuilov, line 81: @johnPhillips I think it would be better to add a new class to the container div (let's say `googlesitekit-settings-module__inline-items`) and use CSS flex styles to make children elements inline instead of introducing a new ad-hoc class for a very narrow use case. In other words, let's have a reusable CSS class that we can use in the future for other elements as well.

eugene-manuilov (changes requested):

assets/js/modules/analytics/components/common/TrackingExclusionSwitches.js
johnPhillips: Thanks @eugene-manuilov. I have added two more generic classes and an atomic rule for the top margin. Let me know what you think?

eugene-manuilov: Sounds good to me. Thanks @johnPhillips.

Issue #3274, opened 2021-05-03
Scaffold Idea Hub widget with tabbed UI


felixarntz:
The new ideas widget for the Idea Hub module should be scaffolded, including its tabbed UI. For now, each tab should be empty though, as these will be implemented in a follow-up issue. It should be possible to link to a specific tab in the widget.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new component `DashboardIdeasWidget` should be scaffolded in `assets/js/modules/idea-hub/components/dashboard`:
    * The widget should be scaffolded to look like [in this Figma file](https://www.figma.com/file/YB79eKcGOwyUjlWG8mjDZA/Prompt-Engine-%2F-Idea-Hub?node-id=20332%3A2653), **excluding the actual content of each tab**, i.e. the list of ideas and the footer that has the pagination should be left out, to be implemented later. For now each tab should just be empty.
    * The three tabs should for now just have their basic labels, without any number suffix indicating the amount of ideas.
    * The filtering UI from the Figma design should not be implemented, i.e. completely ignored.
    * If the `idea-hub` module is not active or not connected, the widget should not be rendered.
    * The component should manually render the `props.Widget` component to ensure the default padding is not added, since manual padding needs to be added for the inner components per the Figma design.
    * It should be possible to link to one of the tabs directly, i.e. every tab should have its own unique ID attribute. If the current URL has that ID as anchor, the tab should be automatically focused (otherwise, i.e. by default, the "New" tab should be focused).
        * This should explicitly not be implemented using React Router, since it is not really a route, as it is embedded in the dashboard where it is just one of many widgets. Using an anchor link is more appropriate here.
* The new component should be registered via a new `registerWidgets( widgets )` function in `assets/js/modules/idea-hub/index.js` which is then called in `assets/js/googlesitekit-modules-idea-hub.js`, as usual. The widget should be registered as follows:
    * `width` should be `WIDGET_WIDTHS.HALF`
    * `wrapWidget` should be `false` as the widget needs to have special padding manually taken care of (see above)
    * The only widget area should be `AREA_DASHBOARD_ACQUISITION`.
    * `priority` should be `2`, taking over this slot from the existing `analyticsPopularPages` widget, which should instead be moved to `3`.
* A Storybook story for the scaffolded widget should be added.

## Implementation Brief
[...]

PR #3344, opened 2021-05-12
Enhancement/3274 idea hub tab scaffold

asvinb:
## Summary

Addresses issue #3274 

## Relevant technical choices
* Added tests for the `DashboardIdeasWidget` to make sure the hash is set even though it was not in the IB
* Added stories in CSF, however if #2940 gets merged before, then we'll have to update the PR to use the `WithRegistrySetup` decorator.

## Checklist


tofumatt:
@googlebot I consent.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -0,0 +1,141 @@
...
+		}
+
+		if ( desktopBreakpoint.matches ) {
+			siteHeaderHeight = 84;
+			marginBottom = 24;
+			adminBarHeight = 32;
+		}
+		const headerOffset = ( siteHeaderHeight + adminBarHeight + marginBottom ) * -1;
+		return ideaHubContainer.current.getBoundingClientRect().top + global.window.pageYOffset + headerOffset;
+	};
```
eugene-manuilov, line 67: Let's pass `ideaHubContainer.current.getBoundingClientRect().top` as a parameter to this function and move the function outside of this component.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -0,0 +1,141 @@
...
+
+		if ( tabletBreakpoint.matches ) {
+			adminBarHeight = 46;
+		}
+
+		if ( desktopBreakpoint.matches ) {
+			siteHeaderHeight = 84;
+			marginBottom = 24;
+			adminBarHeight = 32;
+		}
```
eugene-manuilov, line 64: Instead of hardcoding all these numbers, we should use actual heights of appropriate elements:

```
siteHeaderHeight = document.querySelector('.googlesitekit-header')?.offsetHeight || 0;
adminBarHeight = document.getElementById( 'wpadminbar' )?.offsetHeight || 0;
```

Not sure about marginBottom, do we really need it? :thinking:

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -0,0 +1,141 @@
...
+		}, 1000 );
+	} );
+
+	const handleTabUpdate = useCallback( ( tabIndex ) => {
+		setActiveTabIndex( tabIndex );
+		setHash( DashboardIdeasWidget.tabIDsByIndex[ tabIndex ] );
+	}, [ setHash, setActiveTabIndex ] );
+
+	return (
+		<div className="googlesitekit-widget">
```
eugene-manuilov, line 85: Please, use `<Widget noPadding>` component passed as a property per AC:

```
The component should manually render the props.Widget component to ensure the default padding is not added...
```

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -0,0 +1,141 @@
...
+import SavedIdeas from './SavedIdeas';
+import DraftIdeas from './DraftIdeas';
+
+const getHash = ( hash ) => hash ? hash.replace( '#', '' ) : false;
+const isValidHash = ( hash ) => getHash( hash ) in DashboardIdeasWidget.tabToIndex;
+
+const DashboardIdeasWidget = () => {
+	const ideaHubContainer = useRef();
+	const [ hash, setHash ] = useHash();
+	const [ activeTabIndex, setActiveTabIndex ] = useState( DashboardIdeasWidget.tabToIndex[ getHash( hash ) ] );
```
eugene-manuilov, line 46: The `getHash` function may return `false` if nothing is selected yet. It means that we will use `undefined` as a default value for the `activeTabIndex` state variable. Let's fall back to the first tab index (which is `0`) in this case.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.test.js
```diff
@@ -0,0 +1,76 @@
...
+			'should change location hash & DOM correctly when the Saved ideas tab is clicked',
+			'Saved',
+			'#saved-ideas',
+		],
+		[
+			'should change location hash & DOM correctly when the Draft ideas tab is clicked',
+			'Drafts',
+			'#draft-ideas',
+		],
+	] )( '%s', async ( _, args, expected ) => {
```
eugene-manuilov, line 64: Let's not repeat text in the test names and use a template for it:

```suggestion
	it.each( [
		[ 'New', '#new-ideas' ],
		[ 'Saved', '#saved-ideas' ],
		[ 'Drafts', '#draft-ideas' ],
	] )( 'should change location hash & DOM correctly when the %s tab is clicked', async ( args, expected ) => {
```

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -0,0 +1,141 @@
...
+import NewIdeas from './NewIdeas';
+import SavedIdeas from './SavedIdeas';
+import DraftIdeas from './DraftIdeas';
+
+const getHash = ( hash ) => hash ? hash.replace( '#', '' ) : false;
+const isValidHash = ( hash ) => getHash( hash ) in DashboardIdeasWidget.tabToIndex;
+
+const DashboardIdeasWidget = () => {
+	const ideaHubContainer = useRef();
+	const [ hash, setHash ] = useHash();
```
eugene-manuilov, line 45: I know that IB says that we need to use hashes, but let's use `core/ui` store to control which tab is open instead of hashes. We use the same approach for other widgets as well, for example, PageSpeed Insights uses it for its tabs: https://github.com/google/site-kit-wp/blob/cf3e8916594c9cb43288f4a639ec938ed53f3a63/assets/js/modules/pagespeed-insights/components/dashboard/DashboardPageSpeed.js#L62-L63

eugene-manuilov (changes requested):
Thanks, @asvinb. It's a good start, however, we need to address a few issues before we can merge this PR. Could you please take a look at it?

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
asvinb: 👍

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
asvinb: I was on the fence about hardcoding those values since they will never change, but better be safe and make them dynamic. Updated. As for the margin, it's just to have some spacing between the header and the widget.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
asvinb: 👍

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
asvinb: Updated @eugene-manuilov

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.test.js
asvinb: 👍

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
asvinb: @eugene-manuilov The thing is that we need to be able to link to a specific tab via the URL as per the ticket description:
> It should be possible to link to a specific tab in the widget.

Unless you want to say to not use the `activeTabIndex` state variable and use the `core/ui` store see what tab is opened. If this is the case, I think it's overkilled in this scenario unless am missing something? 🤔 
I checked the PageSpeed Insights and I can't find a way where you can specify a tab to be open.
Let me know :)

eugene-manuilov: Ok, let it be as is then.

assets/sass/components/idea-hub/_googlesitekit-idea-hub-dashboard-ideas-widget.scss
```diff
@@ -0,0 +1,70 @@
...
+.googlesitekit-plugin {
+
+	.googlesitekit-idea-hub {
+		background-color: $c-idea-hub-background;
+		height: 100%;
+	}
+
+	.googlesitekit-idea-hub__header {
+		padding: 24px 16px 0;
+	}
```
eugene-manuilov, line 28: These styles seem to be off. As far as I see in the figma design it should be `20px 24px 0`, shouldn't it? 

Screenshot from 2021-05-14 15-01-54

eugene-manuilov (changes requested):
Awesome. It is almost good to go. Left the final comment about styles. @asvinb could you please take a look at it?

assets/sass/components/idea-hub/_googlesitekit-idea-hub-dashboard-ideas-widget.scss
asvinb: @eugene-manuilov If we follow the designs, then the title will not be aligned with titles from other widgets. Check the screenshot below:
Padding top set to `24px` 
image
compared to padding top set to `20px` as per the designs:
image

You will notice that there is some misalignment.

eugene-manuilov: Ah... okay. Got it. Thanks!

eugene-manuilov (approved):
Looks good now. Thanks, @asvinb.

Issue #3245, opened 2021-04-29
Integrate the GA4 submit changes store functionality into the UA submit changes store


felixarntz:
#3244 implemented the `canSubmitChanges` (plus `__dangerousCanSubmitChanges`) selector and the `submitChanges` action for the `modules/analytics-4` store.

For the first milestone of adding GA4 support, the GA4 module only exists internally, while the only user-facing touch points are exposed via the existing Analytics module, which primarily is for UA but now allows for some GA4 support. As such, since all GA4 data is controlled via UI in the existing Analytics module, some parts of the `modules/analytics-4` store need to be integrated into the `modules/analytics` store so that calls to the latter from the UI are sufficient to cover both.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `submitChanges` action of the `modules/analytics` store should after successful execution of its own functionality also trigger the `submitChanges` action of the `modules/analytics-4` store **if that store is loaded** (which depends on the feature flag being enabled, via the server side).
    * This should be trivial by calling the `submitChanges` action of `modules/analytics-4` in the `submitChanges` function from `modules/analytics`.
* Similarly, the `canSubmitChanges` selector (plus its "dangerous" companion selector) of the `modules/analytics` store should also require the corresponding functionality of the `modules/analytics-4` store to pass **if that store is loaded** (as above).
    * This should be possible to achieve simply by relying on the `modules/analytics-4` store's `__dangerousCanSubmitChanges` selector within the `validateCanSubmitChanges` code from `modules/analytics`.

## Implementation Brief

1. Inside the `submitChanges` action of the Analytics store, after successfully completing its own functionality:
  * Ascertain that the Analytics-4 feature (`ga4setup`) is enabled via `isFeatureEnabled`
	* If it is enabled, call the `submitChanges` action of the Analytics-4 module (see #3244)
2. Inside the `validateCanSubmitChanges` selector of the Analytics store, 
  * Ascertain that the Analytics-4 feature (`ga4setup`) is enabled via `isFeatureEnabled`
	* If it is enabled, call the `__dangerousCanSubmitChanges` selector in the Analytics-4 store (important to use this version of the `canSubmitChanges` selector because it will throw if it fails).

### Test Coverage

* Add test coverage for the new functionality; existing tests should pass.

### Visual Regression Changes

* None anticipated.

## QA Brief

[...]

PR #3345, opened 2021-05-12
Enhancement/3245 ga4 submit settings

eugene-manuilov:
## Summary

Addresses issue #3245

## Checklist


assets/js/modules/analytics/datastore/settings.test.js
```diff
@@ -333,6 +339,70 @@ describe( 'modules/analytics settings', () => {
...
+			describe( 'analytics-4', () => {
+				beforeEach( () => {
+					registry.dispatch( STORE_NAME ).receiveGetExistingTag( null );
+					registry.dispatch( STORE_NAME ).setSettings( validSettings );
+
+					enabledFeatures.add( 'ga4setup' );
+				} );
+
+				afterEach( () => {
+					enabledFeatures.delete( 'ga4setup' );
```
aaemnnosttv, line 352: This shouldn't be necessary as `enabledFeatures` is already cleared after each test.

assets/js/modules/analytics/datastore/settings.test.js
```diff
@@ -333,6 +339,70 @@ describe( 'modules/analytics settings', () => {
...
+					expect( saveChangesError ).toBeUndefined();
+
+					expect( fetchMock ).toHaveFetched( gaSettingsEndpoint, { body: { data: validSettings } } );
+					expect( fetchMock ).toHaveFetched( ga4SettingsEndpoint, { body: { data: ga4Settings } } );
+
+					expect( registry.select( STORE_NAME ).haveSettingsChanged() ).toBe( false );
+					expect( registry.select( MODULES_ANALYTICS_4 ).haveSettingsChanged() ).toBe( false );
+				} );
+
+				it( 'should ignore analytics-4 errors if it fails', async () => {
```
aaemnnosttv, line 379: I'm not sure if this is the expected behavior or not as it isn't in the ACs, but I would expect an error from GA4 to be returned from UA `submitChanges` if there was one since it's functioning as a combined method.

@felixarntz what should happen in the event of a GA4 error when submitting changes?

assets/js/modules/analytics/datastore/settings.test.js
```diff
@@ -525,6 +595,38 @@ describe( 'modules/analytics settings', () => {
...
+			describe( 'analytics-4', () => {
+				beforeEach( () => {
+					registry.dispatch( STORE_NAME ).receiveGetExistingTag( null );
+					registry.dispatch( STORE_NAME ).setSettings( validSettings );
+
+					enabledFeatures.add( 'ga4setup' );
+				} );
+
+				afterEach( () => {
+					enabledFeatures.delete( 'ga4setup' );
```
aaemnnosttv, line 608: See previous comment

aaemnnosttv (commented):
Looks great @eugene-manuilov ! Just a few suggestions and a question for @felixarntz.

assets/js/modules/analytics/datastore/settings.test.js
felixarntz: @aaemnnosttv Indeed it looks like that's missing in the ACs, but it is correct (and mentioned in the ACs of the related #3244). The reason that errors here should be ignored from a user perspective is that the APIs powering these interactions are unstable, and if something fails with GA4 for that reason, it's not a blocker from proceeding.

This should all change once the GA4 APIs are stable. For now, we only should track such errors to have awareness if we need to make changes (e.g. in a hotfix), which should already be the case through the regular API error tracking for opted-in users (correct @eugene-manuilov?).

eugene-manuilov: Yep, we track it here: https://github.com/google/site-kit-wp/blob/db4d6444f12e4185f1ff36e2602e3b6fe14a1fc6/assets/js/googlesitekit/api/index.js#L171-L173

assets/js/modules/analytics/datastore/settings.test.js
eugene-manuilov: Updated.

assets/js/modules/analytics/datastore/settings.test.js
eugene-manuilov: Updated.

aaemnnosttv (approved):
Great, thanks!

Issue #3348, opened 2021-05-13
`Cannot read property 'clicks' of undefined` error on Search Console widget-based dashboard


aaemnnosttv:
## Bug Description

When viewing the Search Console dashboard, it is possible to encounter the following fatal error:

```
Cannot read property 'clicks' of undefined

    in Stats (created by ModuleOverviewWidget)
    in div (created by Widget)
    in div (created by Widget)
    in Widget (created by WithWidgetSlug(Widget))
    in WithWidgetSlug(Widget) (created by ModuleOverviewWidget)
    in ModuleOverviewWidget (created by WidgetRenderer)
    in WidgetRenderer (created by WidgetAreaRenderer)
    in div (created by Cell)
    in Cell (created by WidgetCellWrapper)
    in WidgetCellWrapper (created by WidgetAreaRenderer)
    in div (created by Row)
    in Row (created by WidgetAreaRenderer)
    in div (created by WidgetAreaRenderer)
    in div (created by Grid)
    in Grid (created by WidgetAreaRenderer)
    in WidgetAreaRenderer (created by WidgetContextRenderer)
    in div (created by WidgetContextRenderer)
    in WidgetContextRenderer (created by ModuleApp)
    in ModuleApp (created by GoogleSitekitModule)
    in GoogleSitekitModule
    in RestoreSnapshots (created by Root)
    in ErrorHandler (created by Root)
    in Root
```

The problem happens when the API returns an odd number of rows for a request that includes the previous range for comparison. This leads to an off-by-one situation when iterating over the previous data due to the way the rows are sliced.

https://github.com/google/site-kit-wp/blob/db4d6444f12e4185f1ff36e2602e3b6fe14a1fc6/assets/js/modules/search-console/components/module/ModuleOverviewWidget/Stats.js#L108-L115

The error happens right here because the `index` in the array of `previous` rows does not exist (because it was not returned from the API:

https://github.com/google/site-kit-wp/blob/db4d6444f12e4185f1ff36e2602e3b6fe14a1fc6/assets/js/modules/search-console/util/site-stats-data.js#L65

This is likely a matter of timing regarding when data is available as it is the very last (most recent) row which is missing
![image](https://user-images.githubusercontent.com/1621608/118090608-71e0e400-b3d2-11eb-82ba-6ac55f705018.png)

This request was for a 28 day range + compare which should have returned 56 rows (28 * 2), but instead it returned 55 (54 + 1 for the 0th row).

[...]

PR #3349, opened 2021-05-13
Normalize report partitioning with new utility

aaemnnosttv:
## Summary

Addresses issue #3348

## Relevant technical choices

* Promoted to general utility as the same partitioning was done in a number of other places, not just in Search Console
* Updated all instances of previous array slicing to use the new utility

## Checklist


aaemnnosttv:
VRT failed with one widget rendering a slightly different date range in the chart although stats remained exactly the same so it looks like this may have been a bug before.

| Before | After |
| --- | --- |
| image | image |

image

assets/js/modules/search-console/components/module/ModuleOverviewWidget/index.js
```diff
@@ -87,6 +89,8 @@ const ModuleOverviewWidget = ( { Widget, WidgetReportZero, WidgetReportError } )
 		);
 	}
 
+	const { compareRange, currentRange } = partitionReport( data, { rangeLength: dateRangeLength } );
```
felixarntz, line 92: Since this is only needed in the `Stats` component, why not do it there?

assets/js/util/partition-report.js
```diff
@@ -0,0 +1,47 @@
...
+	invariant( Number.isInteger( rangeLength ) && rangeLength > 0, 'rangeLength must be a positive integer.' );
+
+	const compareRange = report.slice( 0, rangeLength );
+	const currentRange = report.slice( rangeLength, rangeLength * 2 );
+	// Normalize the length of both ranges to be the same.
+	const minLength = Math.min( compareRange.length, currentRange.length );
+
+	return {
+		compareRange: compareRange.slice( 0, minLength ),
+		currentRange: currentRange.slice( 0, minLength ),
```
felixarntz, line 45: Could this result in a potential problem that the compare date range would be incorrect (e.g. missing some days)? For example, if the `currentRange` is 1 entry longer than `compareRange`, right now this would strip the last result of the `compareRange`, so effectively that date would be skipped which would result in somewhat of a wrong comparison (where furthermore even for a "multitude of a week" period weekdays may not be aligned).

A more appropriate solution may be to slice the `compareRange` from its end, effectively trimming away the oldest dates. This may be more specific to Site Kit's behavior, but I think that makes sense - we don't need this utility to be overly generic.

assets/js/util/partition-report.js
```diff
@@ -0,0 +1,47 @@
...
+ * Splits report data into segments for current and comparison ranges.
+ *
+ * @since n.e.x.t
+ *
+ * @param {Array}  report           Report rows.
+ * @param {Object} args             Additional arguments.
+ * @param {number} args.rangeLength Length of an individual range.
+ * @return {Object} Object with keys for `compareRange` and `currentRange`.
+ */
+export const partitionReport = ( report, { rangeLength } ) => {
```
felixarntz, line 34: While the code for this utility function right now appears to be quite generic, I'm not sure that's really necessary (see also other comment). It's somewhat of a nit-pick, but it may be better to call the argument here for example `dateRangeLength` for more clarity on what it refers to. The return object properties could possibly be something like `dateRangeEntries` and `compareDateRangeEntries`, just throwing some ideas out there.

TLDR I think it's a good idea to name and document this more focused around date ranges since that's what we actually use it for.

felixarntz (commented):
This looks solid mostly, I have a few points of feedback, particularly one thing is worth improving.

assets/js/modules/search-console/components/module/ModuleOverviewWidget/index.js
aaemnnosttv: I originally did but thought it might be cleaner to do this way since it needed the date range as well which it would need another prop for. I guess it really wasn't much of an improvement so I've moved it back into `Stats` with the added prop rather than changing the props it receives data from.

assets/js/util/partition-report.js
aaemnnosttv: > Could this result in a potential problem that the compare date range would be incorrect (e.g. missing some days)?

I don't think so – it shouldn't be possible for the `compareRange` to be shorter than the `currentRange` since it is sliced from the beginning of the `report` array. Due to the way the two ranges are sliced before this, `currentRange` can never be longer because they both have the same length limit.

I suppose this makes the slicing on the `compareRange` here redundant and we could simply slice the `currentRange` to the length of the `compareRange` instead.

assets/js/util/partition-report.js
aaemnnosttv: I've updated the range length argument to be `dateRangeLength` as you suggested – this cleaned up the consuming code as well 👍 

> The return object properties could possibly be something like `dateRangeEntries` and `compareDateRangeEntries`, just throwing some ideas out there.

This seems a bit more verbose than necessary. We should keep the return focused on the particular segment of the report.

We could also simply return an array from the function and leave the naming to the consumer via destructuring:

```js
const [ compare, current ] = partitionReport( data, { dateRangeLength } )
```

I wouldn't be opposed to this, but our use would be more consistent to return an object like it does now.

felixarntz: Using an object sounds good.

felixarntz (approved):
@aaemnnosttv I've added one more test here as mentioned that, while it doesn't really test anything extra, uses more realistic data to prove that the implementation maintains the weekday alignment when data is requested for a multitude of weeks. Let me know if that looks good to you / feel free to merge if so.

aaemnnosttv (commented):
The added test LGTM – thanks for adding that @felixarntz 

I can't actually approve this since it's my own PR but ✅  from me 😄

Issue #3248, opened 2021-04-29
Implement Analytics setup form UI for accounts that only have UA properties


felixarntz:
The UI for the scenario that an Analytics account has only UA properties available (yet with GA4 support) should be implemented.

For details on this UI, see:
* [design doc section](https://docs.google.com/document/d/11DMYyAGhtFHO8bMH107pacC8iMpD9PJ9BTaR6IF_B5E/edit#heading=h.6t1gq9q9bzg3)
* [Figma design](https://www.figma.com/file/ix0pcdTQi1Bf9QFNBlBByB/GA4-%2B-UA-onboarding?node-id=20269%3A9057)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The UI from the [Figma design](https://www.figma.com/file/ix0pcdTQi1Bf9QFNBlBByB/GA4-%2B-UA-onboarding?node-id=20269%3A9057) should be implemented in the `SetupFormUA` component that was scaffolded in #3247.
    * One exception: For the blue notification, rely on our existing `SettingsNotice` component for the look for now, to keep this simple. The notification design will be updated to the one specified in the Figma file in a follow-up issue.
* Technically, the UI implementation should largely follow what is in `SetupFormLegacy`:
    * A `StoreErrorNotices` for the `modules/analytics` store should be on top, followed conditionally by `ExistingTagNotice` and `ExistingGTMPropertyNotice`.
    * The main dropdowns should be rendered in a `.googlesitekit-setup-module__inputs` container, followed by another `.googlesitekit-setup-module__inputs` container for the `ProfileNameTextField`, if creating a new profile/view.
* Another technical requirement to specifically point out is that this component should automatically set the `modules/analytics-4` store's property ID setting to `PROPERTY_CREATE`: Since this scenario is for when the account has no GA4 properties, the user doesn't need to select anything, instead it is clear that a new GA4 property needs to be created.
* Basic Storybook coverage for rendering `SetupForm` with this UI state (i.e. so that `SetupFormUA` is shown) should be implemented.

## Implementation Brief
* Using `SetupFormUA`,
	* Copy the code from `SetupFormLegacy` with changes as per the second bullet point in the AC.
	* Drop downs should be implemented as per the AC.
	* Upon component mount, use the `selectProperty` action of the `analytics-4` data store to set the property ID to `PROPERTY_CREATE` as per the third bullet point in the AC.
[...]

PR #3351, opened 2021-05-13
Enhancement/3248 analytics form UI

asvinb:
## Summary

Addresses issue #3248 

## Checklist


asvinb:
@aaemnnosttv I have a question regarding component stories. In the PR, we should have a story which renders the `SetupFormUA` component. However the later is nested within the `SetupModule` and `SetupForm` component. So my question is, shall we create a story which renders the `SetupFormUA` component directly (thus we need to recreate some of the styles) or a story which renders `SetupFormUA` via the `SetupModule` component?

aaemnnosttv:
@asvinb – good question! In this case, it probably makes sense to do something similar to what we already have in the existing Analytics setup stories which renders the `ModuleSetup` component (which is the one I think you were referring to – `SetupModule` is the little CTA on the Connect Services tab).

assets/js/modules/analytics/components/setup/SetupFormUA.stories.js
```diff
@@ -0,0 +1,91 @@
...
+
+			registry.dispatch( STORE_NAME ).receiveGetSettings( {} );
+			registry.dispatch( STORE_NAME ).receiveGetAccounts( accounts );
+			registry.dispatch( STORE_NAME ).finishResolution( 'getAccounts', [] );
+
+			// eslint-disable-next-line sitekit/acronym-case
+			registry.dispatch( STORE_NAME ).receiveGetProperties( properties, { accountID } );
+			registry.dispatch( STORE_NAME ).receiveGetProfiles( profiles, { accountID, propertyID } );
+
+			registry.dispatch( STORE_NAME ).receiveGetSettings( { accountID } );
```
eugene-manuilov, line 48: We need to select this account instead of loading it as part of the module settings:
```suggestion
			registry.dispatch( STORE_NAME ).selectAccount( { accountID } );
```

eugene-manuilov (changes requested):
Nice work, @asvinb. It almost looks good to me. Left a single comment for the storybook store. Please, take a look at it and delete the empty `assets/js/components/setup/ModuleSetup.stories.js` file.

assets/js/modules/analytics/components/setup/SetupFormUA.stories.js
```diff
@@ -0,0 +1,91 @@
...
+	},
+];
+
+export default {
+	title: 'Modules/Analytics/Setup/SetupFormUA',
+	decorators: [
+		( Story ) => {
+			const setupRegistry = ( registry ) => {
+				enabledFeatures.clear();
+				enabledFeatures.add( 'ga4setup' );
```
aaemnnosttv, line 66: Stories shouldn't need to interact with `enabledFeatures` directly. `features` should be provided as an array to `Story.parameters.features`. The `clear` before/after is handled automatically 👍

assets/js/modules/analytics/components/setup/SetupFormUA.stories.js
asvinb: @eugene-manuilov If i make that change, then the UI will be different from what we have in Figma + there's a console error.
image

assets/js/modules/analytics/components/setup/SetupFormUA.stories.js
asvinb: Nice. thanks @aaemnnosttv ! Updated!

eugene-manuilov (changes requested):

assets/js/modules/analytics/components/setup/SetupFormUA.stories.js
eugene-manuilov: @asvinb see an example here: https://github.com/google/site-kit-wp/pull/3421/files#diff-f9c7297e7f2970b10e8c61bab07940db10fd1b64f72c07b1b690e91d3adbc1f6R34-R54

assets/js/modules/analytics/components/setup/SetupFormUA.js
```diff
@@ -16,24 +16,78 @@
...
 			</div>
-			<div>SetupFormUA</div>
+
+			{ profileID === PROFILE_CREATE && (
+				<div className="googlesitekit-setup-module__inputs googlesitekit-setup-module__inputs--multiline">
+					<ProfileNameTextField />
+				</div>
+			) }
+
+			{ isValidPropertyID( propertyID ) && <GA4PropertyNotice /> }
```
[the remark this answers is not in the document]
asvinb: @eugene-manuilov The doc here: https://docs.google.com/document/d/11DMYyAGhtFHO8bMH107pacC8iMpD9PJ9BTaR6IF_B5E/edit?resourcekey=0-0CgoW31PILS2t6JGZ6w_yA#heading=h.6t1gq9q9bzg3 states that 
> Once a property has been selected, the user sees the view dropdown, as well as a notice below that informs them that an associated GA4 property will also be created.

So the notice is not visible when the component is rendered.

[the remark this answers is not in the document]
eugene-manuilov: It doesn't matter which property you select, a new GA4 property will be created anyway because the currently selected account doesn't have GA4 properties at all.

assets/js/modules/analytics/components/setup/SetupFormUA.stories.js
asvinb: @eugene-manuilov Story updated as per the designs in Figma. Actually I didn't want it to match a property so that we could test when the notice would show, i.e after selecting a property. :-)

assets/js/modules/analytics/components/common/GA4PropertyNotice.js
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Internal dependencies
+ */
+import SettingsNotice, { TYPE_INFO } from '../../../../components/SettingsNotice';
+import Link from '../../../../components/Link';
+
+export default function GA4PropertyNotice() {
+	return (
+		<SettingsNotice type={ TYPE_INFO }>
+			{ __( 'A Google Analytics 4 property will also be created.', 'google-site-kit' ) }
```
eugene-manuilov, line 33: @asvinb let's make this text passed via props to be able to reuse this component in other versions of the setup form.

assets/js/modules/analytics/components/setup/SetupFormUA.js
```diff
@@ -16,24 +16,78 @@
...
+					{ __( 'Please select the account information below. You can change this view later in your settings.', 'google-site-kit' ) }
+				</p>
+			) }
+
 			<div className="googlesitekit-setup-module__inputs">
 				<AccountSelect />
+
+				<PropertySelect />
+
+				{ isValidPropertyID( propertyID ) && <ProfileSelect /> }
```
eugene-manuilov, line 81: The `isValidPropertyID( propertyID )` check isn't needed here because the `ProfileSelect` component does the same internally: https://github.com/google/site-kit-wp/blob/6543b1cd989c03bfc62b22764b48a2e1e83464e5/assets/js/modules/analytics/components/common/ProfileSelect.js#L56-L58

assets/js/modules/analytics/components/setup/SetupFormUA.stories.js
```diff
@@ -0,0 +1,89 @@
...
+
+export const Ready = () => <ModuleSetup moduleSlug="analytics" />;
+Ready.storyName = 'SetupFormUA';
+Ready.decorators = [
+	( Story ) => {
+		const setupRegistry = ( registry ) => {
+			const { accounts, properties, profiles, matchedProperty } = fixtures.accountsPropertiesProfiles;
+			/* eslint-disable sitekit/acronym-case */
+			const accountID = properties[ 0 ].accountId;
+			const propertyID = profiles[ 0 ].webPropertyId;
```
eugene-manuilov, line 37:
```suggestion
			/* eslint-disable sitekit/acronym-case */
			const accountID = properties[ 0 ].accountId;
			const propertyID = profiles[ 0 ].webPropertyId;
			/* eslint-enable */
```

assets/js/modules/analytics/components/setup/SetupFormUA.stories.js
```diff
@@ -0,0 +1,89 @@
...
+
+			registry.dispatch( STORE_NAME ).receiveGetSettings( {} );
```
eugene-manuilov, line 39:
```suggestion
			registry.dispatch( STORE_NAME ).receiveGetSettings( { adsConversionID: '' } );
```

assets/js/modules/analytics/components/setup/SetupFormUA.stories.js
```diff
@@ -0,0 +1,89 @@
...
+			registry.dispatch( STORE_NAME ).receiveGetAccounts( accounts );
+			registry.dispatch( STORE_NAME ).receiveGetProperties( properties, { accountID } );
+			registry.dispatch( STORE_NAME ).receiveGetProfiles( profiles, { accountID, propertyID } );
+			registry.dispatch( STORE_NAME ).receiveGetExistingTag( null );
+			registry.dispatch( STORE_NAME ).receiveMatchedProperty( matchedProperty );
+			registry.dispatch( MODULES_ANALYTICS_4 ).receiveGetProperties( [], { accountID } );
```
eugene-manuilov, line 45: We must use `selectAccount` action:

```suggestion
			registry.dispatch( MODULES_ANALYTICS_4 ).receiveGetProperties( [], { accountID } );
			registry.dispatch( STORE_NAME ).selectAccount( accountID );
```

assets/js/modules/analytics/components/setup/SetupFormUA.stories.js
```diff
@@ -0,0 +1,89 @@
...
+						connected: true,
+					},
+					{
+						slug: 'analytics-4',
+						active: true,
+						connected: true,
+					},
+				] );
+
+				provideModuleRegistrations( registry );
```
eugene-manuilov, line 79: It would be also helpful to call `provideSiteInfo( registry )` to set up site settings.

eugene-manuilov (changes requested):
Thanks, @asvinb. A few more changes are required.

assets/js/modules/analytics/components/setup/SetupFormUA.js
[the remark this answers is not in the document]
asvinb: I see. Updated

assets/js/modules/analytics/components/common/GA4PropertyNotice.js
asvinb: 👍

assets/js/modules/analytics/components/setup/SetupFormUA.js
asvinb: 👍

assets/js/modules/analytics/components/setup/SetupFormUA.stories.js
asvinb: 👍

assets/js/modules/analytics/components/setup/SetupFormUA.stories.js
asvinb: 👍

assets/js/modules/analytics/components/setup/SetupFormUA.stories.js
asvinb: 👍

assets/js/modules/analytics/components/setup/SetupFormUA.stories.js
```diff
@@ -0,0 +1,91 @@
...
+		const setupRegistry = ( registry ) => {
+			const { accounts, properties, profiles } = fixtures.accountsPropertiesProfiles;
+			/* eslint-disable sitekit/acronym-case */
+			const accountID = properties[ 0 ].accountId;
+			const propertyID = profiles[ 0 ].webPropertyId;
+			/* eslint-enable */
+
+			registry.dispatch( STORE_NAME ).receiveGetSettings( { adsConversionID: '' } );
+			registry.dispatch( STORE_NAME ).receiveGetAccounts( accounts );
+			registry.dispatch( STORE_NAME ).receiveGetProperties( properties, { accountID } );
```
eugene-manuilov, line 42: Let's limit the number of accounts and properties to 1 which is enough for the storybook story:
```suggestion
			registry.dispatch( STORE_NAME ).receiveGetAccounts( accounts.slice( 0, 1 ) );
			registry.dispatch( STORE_NAME ).receiveGetProperties(
				// eslint-disable-next-line sitekit/acronym-case
				properties.slice( 0, 1 ).map( ( property ) => ( { ...property, websiteUrl: 'http://example.com' } ) ),
				{ accountID },
			);
```

assets/js/modules/analytics/components/setup/SetupFormUA.stories.js
```diff
@@ -0,0 +1,91 @@
...
+export const Ready = () => <ModuleSetup moduleSlug="analytics" />;
+Ready.storyName = 'SetupFormUA';
+Ready.decorators = [
+	( Story ) => {
+		const setupRegistry = ( registry ) => {
+			const { accounts, properties, profiles } = fixtures.accountsPropertiesProfiles;
+			/* eslint-disable sitekit/acronym-case */
+			const accountID = properties[ 0 ].accountId;
+			const propertyID = profiles[ 0 ].webPropertyId;
+			/* eslint-enable */
```
eugene-manuilov, line 38: I think we can make it even simpler: 
```suggestion
			const accountID = accounts[ 0 ].id;
			const propertyID = properties[ 0 ].id;
```

eugene-manuilov (changes requested):
Ok, two more minor updates for the storybook story and it should be good to go. I'll apply it.

Issue #3247, opened 2021-04-29
Prepare Analytics setup form UI for split into four sub-variants (depending GA4 support and available properties)


felixarntz:
As specified in #3170, the Analytics setup flow with GA4 support will be tiered into three different versions, one of which will be rendered based on the currently selected account and whether it has only UA properties, only GA4 properties, or properties of both types.

Then, in addition there is the existing UI which also needs to be maintained until we launch GA4 support (currently behind `ga4setup` feature flag). The existing UI is basically the UI for no GA4 support (which technically also means it only shows UA properties, but other than the new UI for that scenario the existing one doesn't even consider GA4 such as create matching properties).

This issue is about splitting the Analytics `SetupForm` component to prepare for the 4 different variants.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Four new components should be scaffolded in `assets/js/modules/analytics/components/setup`:
    * `SetupFormLegacy`
    * `SetupFormUA`
    * `SetupFormGA4`
    * `SetupFormGA4Transitional`
* Each of them should be responsible for rendering the form UI within the `SetupForm` component, specifically everything within the `form` element except for the `.googlesitekit-setup-module__action` with the submit button.
    * The currently existing form UI in `SetupForm` should be moved as is into `SetupFormLegacy`. In other words, that component can already be completed here.
    * The other 3 components should only be scaffolded as part of this issue:
        * Each of them should have a version of the `.googlesitekit-setup-module__inputs` container, for now with the `AccountSelect` dropdown only.
        * After the `.googlesitekit-setup-module__inputs` container, each of them should for now have a simple placeholder `div` with text what they are for (e.g. `SetupFormUA` should return `<div>SetupFormUA</div>`).
* The `SetupForm` component should then be updated to conditionally render one of the four new components, based on the `getSetupFlowMode` selector:
    * `legacy` --> `SetupFormLegacy`
    * `ua` --> `SetupFormUA`
    * `ga4` --> `SetupFormGA4`
    * `ga4transitional` --> `SetupFormGA4Transitional`
* One small related change should also be made to the existing `SetupMain` component: If the `getSetupFlowMode` selector returns `undefined` (i.e. not loaded), the `ProgressBar` condition should apply, i.e. the whole setup should show a loading state.
* It needs to be ensured that existing Storybook coverage (i.e. for the `legacy` case) remains intact.

## Implementation Brief
* Create four new components as per the AC.
[...]

Issue #3170, opened 2021-04-15
Implement Analytics datastore infrastructure to determine which UI to display (UA vs GA4 vs both)


felixarntz:
Now that we have both an UA Analytics module (`analytics`) and an (internal) GA4 Analytics module (`analytics-4`), we need to cater for a combined setup flow. At this point, the `analytics` module is still the source of truth and required, and the `analytics-4` module solely provides extra functionality purely related to GA4. Eventually, they will be further decoupled though until at some point the `analytics` module can be removed, keeping only the `analytics-4` module (this is likely more than a year away though, and not part of the initial GA4 MVP at all).

For now, the entire setup flow will remain part of the `analytics` module, however it will now have to cater for whether the `analytics-4` module is also there and whether the user has GA4 properties or not.

The end result should be that every user going through the Analytics setup flow now chooses both a UA property and a GA4 property, effectively double tagging to populate data for both of them.

One of the pieces needed is a separated setup UI based on circumstances:
* Either `ua`, if the user only has UA properties in the currently selected account. This is effectively what we have today.
* Or `ga4`, if the user only has GA4 properties in the currently selected account.
* Or `ga4-transitional`, if the user has both UA properties and GA4 properties in the currently selected account.
* Another possible return value needs to be `legacy`, which is for only the case that the GA4 store is not loaded or that the API does not work.

This issue adds the first piece of infrastructure for this, which is a selector to determine the correct "setup flow mode".

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `modules/analytics` datastore should get a new selector `getSetupFlowMode()`:
    * If the `modules/analytics-4` store is not available, it should return “legacy”.
    * If the `modules/analytics-4` store selector `isAdminAPIWorking()` (see #3169) returns `false`, it should return “legacy”. If it returns `undefined`, it should also return `undefined` (indicating "not loaded yet").
    * If there is no account selected, it should return “ua”.
    * If there is an account selected for which the `modules/analytics-4` store selector `getProperties` returns an empty array (i.e. no GA4 properties), it should return “ua”.
    * If there is an account selected for which the `modules/analytics` store selector `getProperties` returns an empty array (i.e. no UA properties), it should return “ga4”.
    * If there is an account selected for which both the `modules/analytics` and `modules/analytics-4` selectors `getProperties` return a non-empty array, it should return “ga4-transitional”.

## Implementation Brief

[...]

PR #3352, opened 2021-05-13
Enhancement/3247 fix loading issue

eugene-manuilov:
## Summary

Addresses issue #3247 

## Relevant technical choices

* Added the new `GET:account-summaries` endpoint to the Analytics module to get information about available properties for the current user.
* Added new `hasProperties` selectors to Analytics and Analytics-4 modules to check whether the module has available properties or not.
* Reworked the `getSetupFlowMode` selector to rely on `hasProperties` selectors.
* Reworked the `isAdminAPIWorking` selector not to rely on available properties and web data streams because if they were not loaded, it would always return `undefined`.

## Checklist

### DIFF
--- a/assets/js/modules/analytics-4/datastore/api.js
+++ b/assets/js/modules/analytics-4/datastore/api.js
@@ -29,13 +29,11 @@ const baseSelectors = {
 	 *
 	 * Returns true if getProperties or getWebDataStreams has fetched successfully.
 	 *
-	 * This selector returns undefined if either `getProperties` or `getWebDataStreams` selectors have not resolved.
-	 *
 	 * @since 1.32.0
 	 * @private
 	 *
 	 * @param {Object} state Data store's state.
-	 * @return {(boolean|undefined)} True, false, or undefined.
+	 * @return {(boolean)} True if there are no GA4 errors, otherwise false.
 	 */
 	isAdminAPIWorking( state ) {
 		// This is against normal state access conventions.
--- a/assets/js/modules/analytics-4/datastore/api.js
+++ b/assets/js/modules/analytics-4/datastore/api.js
@@ -52,17 +52,7 @@ const baseSelectors = {
 			}
 		}
 
-		for ( const propertyList of Object.values( state.properties ) ) {
-			if ( propertyList && propertyList.length && propertyList.length > 0 ) {
-				for ( const webdatastreamList of Object.values( state.webdatastreams ) ) {
-					if ( webdatastreamList && webdatastreamList.length && webdatastreamList.length > 0 ) {
-						return true;
-					}
-				}
-			}
-		}
-
-		return undefined;
+		return true;
 	},
 };
 
--- a/assets/js/modules/analytics/datastore/accounts.js
+++ b/assets/js/modules/analytics/datastore/accounts.js
@@ -60,6 +60,21 @@ const fetchGetAccountsPropertiesProfilesStore = createFetchStore( {
 	},
 } );
 
+const fetchGetAccountSummaries = createFetchStore( {
+	baseName: 'getAccountSummaries',
+	controlCallback() {
+		return API.get( 'modules', 'analytics', 'account-summaries', {}, {
+			useCache: false,
+		} );
+	},
+	reducerCallback( state, accountSummaries ) {
+		return {
+			...state,
+			accountSummaries,
+		};
+	},
+} );
+
 const fetchCreateAccountStore = createFetchStore( {
 	baseName: 'createAccount',
 	controlCallback: ( { data } ) => {
@@ -87,6 +102,7 @@ const RESET_ACCOUNTS = 'RESET_ACCOUNTS';
 
 const baseInitialState = {
 	accounts: undefined,
+	accountSummaries: undefined,
 	isAwaitingAccountsPropertiesProfilesCompletion: false,
 	accountTicketID: undefined,
 };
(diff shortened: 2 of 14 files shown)


assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -294,6 +320,18 @@ const baseSelectors = {
...
+	 * Gets account summaries.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {(Array.<Object>|undefined)} An array of Analytics account summaries; `undefined` if not loaded.
+	 */
+	getAccountSummaries( state ) {
+		return state?.accountSummaries;
+	},
```
felixarntz, line 333: This was only added so that it can be used in the `modules/analytics` `hasProperties` selector correct?

assets/js/modules/analytics/datastore/setup-flow.js
```diff
@@ -40,57 +40,31 @@ const baseSelectors = {
...
-		// Check to see if the Admin API is working—if it's `undefined` the request
-		// is loading, but if it's `false` we should also use the legacy analytics
-		// because the API isn't working properly.
+		// Check to see if the Admin API is working—if it's `false` we should
+		// use the legacy analytics because the API isn't working properly.
 		const isAdminAPIWorking = select( MODULES_ANALYTICS_4 ).isAdminAPIWorking();
-
-		if ( isAdminAPIWorking === undefined ) {
-			return undefined;
-		}
```
felixarntz, line 50: I think this (and the `isAdminAPIWorking` ability to differentiate between `undefined` and `true`) shouldn't be removed, it needs to be maintained. Otherwise, the selector returns `true` even though it doesn't even know (yet) that the GA4 Admin API _actually_ works.

includes/Modules/Analytics.php
```diff
@@ -348,6 +349,10 @@ protected function get_datapoint_definitions() {
 	 */
 	protected function create_data_request( Data_Request $data ) {
 		switch ( "{$data->method}:{$data->datapoint}" ) {
+			case 'GET:account-summaries':
+				return function() {
+					return $this->get_service( 'analytics' )->management_accountSummaries->listManagementAccountSummaries();
+				};
```
felixarntz, line 355: This must not be wrapped in a closure. Wrapping in a function should only be used if the data point logic relies primarily on custom code. If it is for a Google API request (like here), the API request itself needs to be returned (see e.g. `POST:create-account-ticket` below). This allows the batch mechanism to work correctly and combine these requests, and the post-processing then happens in `parse_data_response`.

assets/js/modules/analytics-4/datastore/api.js
```diff
@@ -52,17 +50,7 @@ const baseSelectors = {
...
-				for ( const webdatastreamList of Object.values( state.webdatastreams ) ) {
-					if ( webdatastreamList && webdatastreamList.length && webdatastreamList.length > 0 ) {
-						return true;
-					}
-				}
-			}
-		}
-
-		return undefined;
+		return true;
```
felixarntz, line 53: See my other comment, this still needs to be able to differentiate between `undefined` and `true`. The logic you removed here indeed looked incorrect though.

What would be more appropriate here would be to:
* If any `getProperties` or `getWebDataStreams` (or `getAccountSummaries`) selector resolution resulted in an error, it should return `false`.
* Otherwise, if any `getProperties` or `getWebDataStreams` (or `getAccountSummaries`) selector has finished resolution, it should return `true`.
* Otherwise, it should return `undefined`.

assets/js/modules/analytics-4/datastore/api.js
```diff
@@ -52,17 +50,7 @@ const baseSelectors = {
 			}
 		}
```
felixarntz, line 51: We should probably also consider the GA4 `getAccountSummaries` selector here - if that randomly failed, we also have breakage likely due to an API change.

Also, why isn't this selector using `createRegistrySelector` and actually checking for something like `hasFinishedResolution`? Regarding the errors, it would be more appropriate to introduce new `getErrorsForSelector( selectorName )` and `getErrorsForAction( actionName )` selectors via `createErrorStore` which we could then use here - but that would increase scope further, so I think the manual `state.errors` checks here are okay.

assets/js/modules/analytics/datastore/setup-flow.js
```diff
@@ -40,57 +40,31 @@ const baseSelectors = {
...
-
-		const accountID = select( MODULES_ANALYTICS ).getAccountID();
-
-		// If no accountID exists then no account is selected. This means we should
-		// use the UA setup flow.
-		if ( ! accountID ) {
-			return SETUP_FLOW_MODE_UA;
-		}
-
-		const ga4Properties = select( MODULES_ANALYTICS_4 ).getProperties( accountID );
```
felixarntz, line 74: Why was this removed? What specifically wasn't working before? While the new approach here partially makes sense (if the user only has properties of one type we know for sure it's that specific type of flow), it's not really appropriate and contradicts the ACs on #3170 (which the original selector was implemented for). We can only really know the flow once the account is selected.

Right now with this PR, the user may select an account with e.g. only UA properties, but if they have GA4 properties in any other account it would still show `ga4transitional`, which would be incorrect.

felixarntz (changes requested):
@eugene-manuilov The approach here is somewhat flawed as it contradicts the requirements on #3170 (which arguably this PR relates more to than to #3247).

What was the specific problem that caused https://github.com/google/site-kit-wp/issues/3247#issuecomment-840575192? Yes, at the beginning no account may be selected, but then it should rely on the `ua` or `legacy` flow (depending on whether `isAdminAPIWorking`). Once the user then selects the account, the flow may need to update - it is crucial that this logic is still based on the currently selected account, not across what the user has between all accounts.

Let's sync on this if needed. This PR is rather big and probably took quite some time to implement - for major changes like this during a QA fix, especially if they change user-facing behavior, we should have some discussion first (potentially even a second IB), to avoid this in the future.

eugene-manuilov:
Closing this PR out in favor of #3386

Issue #3282, opened 2021-05-04
Add support to place Google Ads conversion ID in Analytics AMP snippet


felixarntz:
While #3161 allows specifying the Google Ads conversion ID and having Site Kit integrate it with the Analytics snippet, the initial enhancement didn't include AMP support due to time constraints. Support for also integrating the Ads conversion ID with the `amp-analytics` snippet placed by the Analytics module should be added via this issue.

See [this comment](https://github.com/google/site-kit-wp/issues/3161#issuecomment-827975122) for more context, plus [the official documentation](https://developers.google.com/google-ads/amp/conversion-measurement#option_1_global_site_tag).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `amp-analytics` snippet placed by the `Analytics\AMP_Tag` class should incorporate the Google Ads conversion ID (`adsConversionID` setting) if specified, placing it in the `config` JSON object alongside the GA measurement ID as outlined in https://developers.google.com/gtagjs/devguide/amp#installation.

## Implementation Brief

* Change the `'config'` value in https://github.com/google/site-kit-wp/blob/db4d6444f12e4185f1ff36e2602e3b6fe14a1fc6/includes/Modules/Analytics/AMP_Tag.php#L128 to be built separately and conditionally, adding the Ads Conversion ID if available, eg:

```php
$config = array(
    $this->tag_id => array(
        'groups' => 'default',
        'linker' => array(
            'domains' => array( $this->home_domain ),
        ),
    ),
);

if ( ! empty( $this->ads_conversion_id ) ) {
    $config[ $this->ads_conversion_id ] => array(
        'groups' => 'default',
    );
}
```

### Test Coverage

* Add a PHPUnit test to ensure this value is only output in the config when available—if this is straightforward to do. (I'm not sure if we have tag insertion test infrastructure already, but adding it would be overkill.)

### Visual Regression Changes

* N/A.

## QA Brief

* Enable AMP (Standard mode) on your site with a compatible theme (Twenty-Twenty works), enable the Analytics module, then add an Ads Conversion ID to the Analytics module in Settings.
* View the page in AMP mode (anonymously, so the `<amp-analytics>` tag appears, as it won't for signed-in users.
* Ensure the `{ "$adsConversionID": { "groups": "default" } }` JSON appears in the tag's `config`.

## Changelog entry

* Update `amp-analytics` tag to include configuration for an ads conversion ID when provided.

[...]

PR #3361, opened 2021-05-13
Add Ads Conversion ID to AMP output.

tofumatt:
## Summary

Addresses issue #3282.

For the life of me I can't get the `<amp-analytics />` tag that outputs this to render on my development site. I've tried all AMP modes, viewing anonymously, etc. but it never appears. Oddly the Tag Manager `<amp-analytics />` tag _will_ appear when I enable Tag Manager, so I'm not sure what's up.

I can't see the tag without this commit as well, so it's unrelated to this change.

But if the reviewer could see if they too aren't seeing the AMP Analytics tag that'd be handy—if so seems there's another bug? Or if not something's up with my install/approach.

Either way, the logic here seems right so I think this is best to review and have someone else idiot-check my tag not working 😅

## Relevant technical choices

I attempted to write some PHPUnit tests for this, but given the `render()` method is protected it implied that we shouldn't be testing this behaviour, so I left it alone.

## Checklist

### DIFF
--- a/includes/Modules/Analytics/AMP_Tag.php
+++ b/includes/Modules/Analytics/AMP_Tag.php
@@ -106,18 +106,26 @@ class AMP_Tag extends Module_AMP_Tag implements Tag_Interface {
 	 * @since 1.24.0
 	 */
 	protected function render() {
+		$config = array(
+			$this->tag_id => array(
+				'groups' => 'default',
+				'linker' => array(
+					'domains' => array( $this->home_domain ),
+				),
+			),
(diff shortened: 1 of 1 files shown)


aaemnnosttv (approved):
LGTM, thanks @tofumatt !

aaemnnosttv:
@tofumatt I did check the tag output and it worked as expected for me so not sure what your issue may have been if not the logged-in opt-out or GTM suppressing it if configured with an Analytics tag.

Issue #3286, opened 2021-05-04
Automatically pre-select GA4 property in dropdown based on current site URL


felixarntz:
If the current user already has a GA4 property with a web data stream that refers to the current site URL, such a GA4 property should be pre-selected in the dropdown during Analytics module setup.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The functionality to find and set a "matched property" when selecting a specific account (relevant for `modules/analytics` store, as enhanced in #3291) should be expanded to cover GA4:
    * The `modules/analytics` action `selectAccount` should initially also reset the `analytics-4` property ID (to an empty string) **if the `modules/analytics-4` store is available**. It should also `setPrimaryPropertyType` to "ua" (i.e. reset the "primary property type", since "ua" is the default).
    * If the `modules/analytics-4` store is available, it should then call and wait for the `modules/analytics-4` selector `getProperties` and, once properties have been fetched, rely on the `matchPropertyByURL` action (see #3168) passing the `getSiteURLPermutations` array of URLs to attempt to find a matching property for the current site URL. If one is found, the matched property should be set via `setPropertyID`.
    * At the end of the `selectAccount` action, if a matching GA4 property was found (i.e. in `modules/analytics-4` store) and no matching UA property was found (i.e. in `modules/analytics` store), the "primary property type" should be set to "ga4", to indicate that the user likely already has a matching GA4 property but no matching UA property yet.
        * In any other scenario this is not necessary, since "ua" is the default anyway, and if matching properties for both types exist, then "ua" should still be preferred.
* The functionality to find and set a "matched property" on pageload (see `getAccounts` resolver in `modules/analytics` store) should be expanded to cover GA4 **if the `modules/analytics-4` store exists**:
    * At the end of the `getAccounts` resolver, if there is a matched property (based on which an account and UA property were just set) and if there either is no GA4 property set **or the GA4 property that is set is not within the account**, similar logic as implemented in the `selectAccount` action above should be implemented to also check for a matching GA4 property in the account, and set it if found.
    * The bold condition here is an important extra requirement, as the two properties selected between UA and GA4 must always be within the same Analytics account.

## Implementation Brief

* Add a new action `waitForProperties` to the properties datastore in the analytics-4 module. It should work similarly to the `waitForWebDataStreams` action from the same datastore but check for getProperties to be resolved using the `__experimentalResolveSelect` function:
[...]

PR #3367, opened 2021-05-14
Enhancement/3286 pre select ga4 property

eugene-manuilov:
## Summary

Addresses issue #3286 

## Checklist

### DIFF
--- a/assets/js/modules/analytics/datastore/accounts.js
+++ b/assets/js/modules/analytics/datastore/accounts.js
@@ -29,11 +29,23 @@ import API from 'googlesitekit-api';
 import Data from 'googlesitekit-data';
 import { createValidatedAction } from '../../../googlesitekit/data/utils';
 import { isValidAccountSelection } from '../util';
-import { STORE_NAME, ACCOUNT_CREATE, PROPERTY_CREATE, FORM_ACCOUNT_CREATE } from './constants';
+import {
+	STORE_NAME,
+	ACCOUNT_CREATE,
+	PROPERTY_CREATE,
+	FORM_ACCOUNT_CREATE,
+	PROPERTY_TYPE_UA, PROPERTY_TYPE_GA4,
+} from './constants';
 import { CORE_FORMS } from '../../../googlesitekit/datastore/forms/constants';
 import { createFetchStore } from '../../../googlesitekit/data/create-fetch-store';
 import { actions as errorStoreActions } from '../../../googlesitekit/data/create-error-store';
 import { actions as tagActions } from './tags';
+import {
+	MODULES_ANALYTICS_4,
+	PROPERTY_CREATE as GA4_PROPERTY_CREATE,
+} from '../../analytics-4/datastore/constants';
+import { isFeatureEnabled } from '../../../features';
+import { CORE_SITE } from '../../../googlesitekit/datastore/site/constants';
 const { createRegistrySelector } = Data;
 const { receiveError, clearError } = errorStoreActions;
 
@@ -138,13 +150,27 @@ const baseActions = {
 				return;
 			}
 
-			// Trigger cascading selections.
-			const properties = registry.select( STORE_NAME ).getProperties( accountID );
-			if ( properties === undefined ) {
-				return; // Selection will happen in resolver.
+			yield Data.commonActions.await( registry.dispatch( STORE_NAME ).waitForProperties( accountID ) );
+
+			const uaProperties = registry.select( STORE_NAME ).getProperties( accountID );
+			registry.dispatch( STORE_NAME ).selectProperty( uaProperties[ 0 ]?.id || PROPERTY_CREATE );
+			registry.dispatch( STORE_NAME ).setPrimaryPropertyType( PROPERTY_TYPE_UA );
+
+			if ( isFeatureEnabled( 'ga4setup' ) ) {
+				yield Data.commonActions.await( registry.dispatch( MODULES_ANALYTICS_4 ).waitForProperties( accountID ) );
+
+				const referenceURL = registry.select( CORE_SITE ).getReferenceSiteURL();
+				const ga4Properties = registry.select( MODULES_ANALYTICS_4 ).getProperties( accountID );
+				const ga4Property = registry.dispatch( MODULES_ANALYTICS_4 ).matchPropertyByURL( ga4Properties, referenceURL );
+				registry.dispatch( MODULES_ANALYTICS_4 ).selectProperty( ga4Property?._id || GA4_PROPERTY_CREATE );
+
+				if ( ga4Property?._id ) {
+					const matchedUAProperty = registry.select( STORE_NAME ).getMatchedProperty();
+					if ( ! matchedUAProperty ) {
+						registry.dispatch( STORE_NAME ).setPrimaryPropertyType( PROPERTY_TYPE_GA4 );
+					}
+				}
 			}
-			const property = properties[ 0 ] || { id: PROPERTY_CREATE };
-			registry.dispatch( STORE_NAME ).selectProperty( property.id );
 		}
 	),
 
--- a/assets/js/modules/analytics/datastore/properties.js
+++ b/assets/js/modules/analytics/datastore/properties.js
@@ -28,7 +28,7 @@ import API from 'googlesitekit-api';
 import Data from 'googlesitekit-data';
 import { createValidatedAction } from '../../../googlesitekit/data/utils';
 import { isValidAccountID, isValidPropertyID, parsePropertyID, isValidPropertySelection } from '../util';
-import { STORE_NAME, PROPERTY_CREATE, PROFILE_CREATE } from './constants';
+import { STORE_NAME, PROPERTY_CREATE, PROFILE_CREATE, PROPERTY_TYPE_UA, PROPERTY_TYPE_GA4 } from './constants';
 import { createFetchStore } from '../../../googlesitekit/data/create-fetch-store';
 import { actions as errorStoreActions } from '../../../googlesitekit/data/create-error-store';
 import { MODULES_ANALYTICS_4 } from '../../analytics-4/datastore/constants';
@@ -101,7 +101,7 @@ const baseInitialState = {
 	properties: {},
 	isAwaitingPropertiesProfilesCompletion: {},
 	matchedProperty: undefined,
-	primaryPropertyType: 'ua',
+	primaryPropertyType: PROPERTY_TYPE_UA,
 };
 
 const baseActions = {
@@ -240,7 +240,7 @@ const baseActions = {
 	 * @return {Object} Redux-style action.
 	 */
 	setPrimaryPropertyType( primaryPropertyType ) {
-		invariant( [ 'ua', 'ga4' ].includes( primaryPropertyType ), 'type must be "ua" or "ga4"' );
+		invariant( [ PROPERTY_TYPE_UA, PROPERTY_TYPE_GA4 ].includes( primaryPropertyType ), 'type must be "ua" or "ga4"' );
 
 		return {
 			payload: { primaryPropertyType },
@@ -250,21 +250,11 @@ const baseActions = {
 };
 
 const baseControls = {
-	[ WAIT_FOR_PROPERTIES ]: createRegistryControl( ( registry ) => ( { payload: { accountID } } ) => {
-		const arePropertiesLoaded = () => registry.select( STORE_NAME ).getProperties( accountID ) !== undefined;
-
-		if ( arePropertiesLoaded() ) {
-			return true;
-		}
-
-		return new Promise( ( resolve ) => {
-			const unsubscribe = registry.subscribe( () => {
-				if ( arePropertiesLoaded() ) {
-					unsubscribe();
-					resolve();
-				}
-			} );
-		} );
+	[ WAIT_FOR_PROPERTIES ]: createRegistryControl( ( { __experimentalResolveSelect } ) => {
+		return async ( { payload } ) => {
+			const { accountID } = payload;
+			await __experimentalResolveSelect( STORE_NAME ).getProperties( accountID );
+		};
 	} ),
 };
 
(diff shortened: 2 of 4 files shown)


assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -155,6 +166,35 @@ const baseActions = {
...
+			const ga4Property = yield Data.commonActions.await(
+				registry.dispatch( MODULES_ANALYTICS_4 ).matchPropertyByURL(
+					ga4Properties.map( ( property ) => property._id ),
+					registry.select( CORE_SITE ).getReferenceSiteURL(),
+				)
+			);
+
+			yield Data.commonActions.await(
+				registry.dispatch( MODULES_ANALYTICS_4 ).selectProperty( ga4Property?._id || GA4_PROPERTY_CREATE ),
+			);
```
aaemnnosttv, line 188: As in the previous PR, please update these actions to `yield {actionCreator}()` rather than awaiting registry.dispatch.

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -280,7 +320,39 @@ const baseResolvers = {
...
+		if ( ! isFeatureEnabled( 'ga4setup' ) ) {
+			return;
+		}
+
+		let ga4Property;
+		const ga4PropertyID = registry.select( MODULES_ANALYTICS_4 ).getPropertyID();
+		if ( ga4PropertyID ) {
+			ga4Property = yield Data.commonActions.await(
+				registry.__experimentalResolveSelect( MODULES_ANALYTICS_4 ).getProperty( ga4PropertyID )
+			);
```
aaemnnosttv, line 336: See above

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -280,7 +320,39 @@ const baseResolvers = {
...
+		const ga4Properties = registry.select( MODULES_ANALYTICS_4 ).getProperties( accountID );
+		ga4Property = yield Data.commonActions.await(
+			registry.dispatch( MODULES_ANALYTICS_4 ).matchPropertyByURL(
+				ga4Properties.map( ( property ) => property._id ),
+				registry.select( CORE_SITE ).getReferenceSiteURL(),
+			)
+		);
+
+		if ( ga4Property?._id ) {
+			yield Data.commonActions.await( registry.dispatch( MODULES_ANALYTICS_4 ).selectProperty( ga4Property._id ) );
```
aaemnnosttv, line 355: See above

aaemnnosttv (commented):
Thanks @eugene-manuilov – I didn't have time to give this a full review at the moment, but overall it looks good. There were a few minor things that should be updated similar to changes we made on the previous PR which I've highlighted below. Will take a more thorough look later today when I can.

assets/js/modules/analytics/datastore/accounts.js
eugene-manuilov: @aaemnnosttv this action is from the `analytics-4` module, so if i call it directly it will be applied to the `analytics` datastore, right?

assets/js/modules/analytics/datastore/accounts.js
eugene-manuilov: See my comment here: https://github.com/google/site-kit-wp/pull/3367#discussion_r636852057

assets/js/modules/analytics/datastore/accounts.js
eugene-manuilov: See my comment here: https://github.com/google/site-kit-wp/pull/3367#discussion_r636852057

assets/js/modules/analytics/datastore/accounts.js
aaemnnosttv: Yes, sorry I missed that. For some reason I was thinking this was in the GA4 store – nevermind! 😄

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -29,12 +29,23 @@ import API from 'googlesitekit-api';
 import Data from 'googlesitekit-data';
 import { createValidatedAction } from '../../../googlesitekit/data/utils';
 import { isValidAccountSelection } from '../util';
-import { STORE_NAME, ACCOUNT_CREATE, PROPERTY_CREATE, FORM_ACCOUNT_CREATE } from './constants';
+import {
+	STORE_NAME,
+	ACCOUNT_CREATE,
+	PROPERTY_CREATE,
+	FORM_ACCOUNT_CREATE,
+	PROPERTY_TYPE_UA, PROPERTY_TYPE_GA4,
```
aaemnnosttv, line 37:
```suggestion
	PROPERTY_TYPE_UA,
	PROPERTY_TYPE_GA4,
```

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -276,11 +316,46 @@ const baseResolvers = {
...
-		const accountID = registry.select( STORE_NAME ).getAccountID();
+		let accountID = registry.select( STORE_NAME ).getAccountID();
 		// Pre-select values from the matched property if no account is selected.
 		if ( matchedProperty && ! accountID ) {
-			registry.dispatch( STORE_NAME ).setAccountID( matchedProperty.accountId ); // eslint-disable-line sitekit/acronym-case
-			registry.dispatch( STORE_NAME ).selectProperty( matchedProperty.id, matchedProperty.internalWebPropertyId ); // eslint-disable-line sitekit/acronym-case
+			/* eslint-disable sitekit/acronym-case */
+			accountID = matchedProperty.accountId;
+			registry.dispatch( STORE_NAME ).setAccountID( matchedProperty.accountId );
+			yield Data.commonActions.await( registry.dispatch( STORE_NAME ).selectProperty( matchedProperty.id, matchedProperty.internalWebPropertyId ) );
```
aaemnnosttv, line 325: `selectProperty` here is an action on the local store so this one can be simplified 😄 
```suggestion
			yield selectProperty( matchedProperty.id, matchedProperty.internalWebPropertyId );
```

assets/js/modules/analytics/datastore/accounts.js
aaemnnosttv, line 188: Perhaps it would be more appropriate for this to be handled in an action on the GA4 store (e.g. `matchProperty( accountID )` rather than including it in the UA store? Then this could await the result of the single action here which would return the matched property, if any. I think that would be a little cleaner (both here and in the other store since we wouldn't need to use `registry.dispatch` there) and could be used in both places here.

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -276,11 +316,46 @@ const baseResolvers = {
...
+		const ga4Properties = registry.select( MODULES_ANALYTICS_4 ).getProperties( accountID );
+		ga4Property = yield Data.commonActions.await(
+			registry.dispatch( MODULES_ANALYTICS_4 ).matchPropertyByURL(
+				ga4Properties.map( ( property ) => property._id ),
+				registry.select( CORE_SITE ).getReferenceSiteURL(),
+			)
+		);
+
+		if ( ga4Property?._id ) {
+			yield Data.commonActions.await( registry.dispatch( MODULES_ANALYTICS_4 ).selectProperty( ga4Property._id ) );
```
aaemnnosttv, line 358: This could use the same extracted action as mentioned above, the only difference I see is that this one doesn't set `GA4_PROPERTY_CREATE` if it isn't matched so maybe the `selectProperty` call wouldn't be included, or the action could take a "default" argument for the ID to select if there was no matched property.

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -240,7 +241,7 @@ const baseActions = {
 	 * @return {Object} Redux-style action.
 	 */
 	setPrimaryPropertyType( primaryPropertyType ) {
-		invariant( [ 'ua', 'ga4' ].includes( primaryPropertyType ), 'type must be "ua" or "ga4"' );
+		invariant( [ PROPERTY_TYPE_UA, PROPERTY_TYPE_GA4 ].includes( primaryPropertyType ), 'type must be "ua" or "ga4"' );
```
aaemnnosttv, line 244:
```suggestion
		invariant(
			[ PROPERTY_TYPE_UA, PROPERTY_TYPE_GA4 ].includes( primaryPropertyType ),
			`type must be "${ PROPERTY_TYPE_UA }" or "${ PROPERTY_TYPE_GA4 }"`
		);
```

aaemnnosttv (commented):
Thanks @eugene-manuilov – this looks good overall. Just a few suggestions for you. Mostly minor things, but I think we can simplify things a bit by moving some of the common actions to the GA4 store. Otherwise this looks like it does everything we need.

assets/js/modules/analytics/datastore/accounts.js
eugene-manuilov: Updated.

assets/js/modules/analytics/datastore/accounts.js
eugene-manuilov: Updated.

assets/js/modules/analytics/datastore/accounts.js
eugene-manuilov: It's a great idea! Updated.

assets/js/modules/analytics/datastore/accounts.js
eugene-manuilov: Updated.

assets/js/modules/analytics/datastore/properties.js
eugene-manuilov: Updated.

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -155,6 +169,20 @@ const baseActions = {
...
 			yield propertyActions.selectProperty( uaProperty?.id, uaProperty?.internalWebPropertyId ); // eslint-disable-line sitekit/acronym-case
+
+			if ( ! isFeatureEnabled( 'ga4setup' ) ) {
+				return;
+			}
+
+			registry.dispatch( STORE_NAME ).setPrimaryPropertyType( PROPERTY_TYPE_UA );
+
+			const ga4MatchProperty = registry.dispatch( MODULES_ANALYTICS_4 ).matchAndSelectProperty( accountID, GA4_PROPERTY_CREATE );
+			const ga4Property = yield Data.commonActions.await( ga4MatchProperty );
```
aaemnnosttv, line 180: Is this await needed? The returned property comes from `matchPropertyByURL` which already uses this on the returned value.

assets/js/modules/analytics/datastore/accounts.test.js
```diff
@@ -230,6 +235,46 @@ describe( 'modules/analytics accounts', () => {
...
+						[ /^\/google-site-kit\/v1\/modules\/analytics-4\/data\/properties/, ga4Fixtures.properties ],
+						[ /^\/google-site-kit\/v1\/modules\/analytics-4\/data\/webdatastreams-batch/, ga4Fixtures.webDataStreamsBatch ],
+					].forEach( ( [ endpoint, body ] ) => {
+						fetchMock.get( endpoint, { body } );
+					} );
+
+					provideSiteInfo( registry );
+				} );
+
+				it( 'should select the correct ga4 property', async () => {
```
aaemnnosttv, line 256:
```suggestion
				it( 'should select the correct GA4 property', async () => {
```

assets/js/modules/analytics/datastore/accounts.test.js
```diff
@@ -230,6 +235,46 @@ describe( 'modules/analytics accounts', () => {
...
+					await registry.dispatch( STORE_NAME ).selectAccount( accountID );
+					expect( registry.select( MODULES_ANALYTICS_4 ).getPropertyID() ).toBe( ga4Fixtures.properties[ 0 ]._id );
+				} );
+
+				it( 'should select the correct ua property', async () => {
```
aaemnnosttv, line 261:
```suggestion
				it( 'should select the correct UA property', async () => {
```

assets/js/modules/analytics/datastore/accounts.test.js
```diff
@@ -425,6 +470,31 @@ describe( 'modules/analytics accounts', () => {
...
+						fetchMock.get( endpoint, { body } );
+					} );
+
+					provideSiteInfo( registry );
+
+					registry.dispatch( STORE_NAME ).receiveGetExistingTag( null );
+					registry.dispatch( MODULES_ANALYTICS_4 ).receiveGetSettings( {} );
+				} );
+
+				it( 'should select correct ga4 property', async () => {
```
aaemnnosttv, line 493:
```suggestion
				it( 'should select correct GA4 property', async () => {
```

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -241,7 +242,10 @@ const baseActions = {
 	 * @return {Object} Redux-style action.
 	 */
 	setPrimaryPropertyType( primaryPropertyType ) {
-		invariant( [ 'ua', 'ga4' ].includes( primaryPropertyType ), 'type must be "ua" or "ga4"' );
+		invariant(
+			[ PROPERTY_TYPE_UA, PROPERTY_TYPE_GA4 ].includes( primaryPropertyType ),
+			'type must be "ua" or "ga4"',
```
aaemnnosttv, line 247:
```suggestion
			`type must be "${ PROPERTY_TYPE_UA }" or "${ PROPERTY_TYPE_GA4 }"`,
```

aaemnnosttv (commented):
Looks good to me overall, just a few small details and one question about a potential extra await.

assets/js/modules/analytics/datastore/accounts.js
eugene-manuilov: Yes, this `await` is needed because we need to wait till the `matchAndSelectProperty` generator action (which returns a promise) finishes.

aaemnnosttv: I see, it's because we use `registry.dispatch` here which always returns a promise. A bit unsightly but unavoidable for now. This will be much cleaner after #1769 as we'll be able to simply `yield dispatch( store, ... )`. Thanks 👍

aaemnnosttv (approved):
Great, thanks!

Issue #3246, opened 2021-04-29
Implement dropdown component showing all UA and GA4 properties


felixarntz:
A new Analytics component `PropertySelectIncludingGA4` should be implemented in the `analytics` module which lists all Analytics properties the user has in the current account, of both UA and GA4 type. This component will be used in the "ga4-transitional" flow (see #3170) to list all of the account's UA _and_ GA4 properties.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new component `PropertySelectIncludingGA4` should be implemented within `assets/js/modules/analytics/components/common`.
* It should render a dropdown with similar appearance and high-level behavior as the existing `PropertySelect`, with the difference that it includes all UA properties and all GA4 properties in the current account (relying on the `getPropertiesIncludingGA4` selector from #3167). For example:
    * If no valid account ID is specified yet, the component should return `null` (also see related #3243).
    * If properties are still loading, the component should return a loading bar.
    * Every property should show up as its name with its ID in parentheses.
* Since `getPropertiesIncludingGA4` returns objects of different shape, this has to be catered for. Each object is either a UA property or a GA4 property, and they have their respective names and IDs in different fields (check the Analytics `PropertySelect` and the GA4 `PropertySelect` components for reference).
* The currently selected property in the dropdown is determined as follows:
    * If the `getPrimaryPropertyType` selector returns "ga4", the selected property has to come from the `modules/analytics-4` store's `getPropertyID` selector.
    * Otherwise or if the `getPrimaryPropertyType` selector returns "ua", the selected property has to come from the `modules/analytics` store's `getPropertyID` selector.
* When selecting a property from the dropdown, the following should happen:
    * If the selected property is a GA4 property, the `setPrimaryPropertyType` action should be called to indicate "ga4". Then, the `modules/analytics-4` store's `selectProperty` action should be used to set the GA4 property, while the `modules/analytics` store's `selectProperty` action should be used to empty/reset the UA property (since that then needs to later be chosen in another dropdown based on the GA4 property).
    * Otherwise or if the selected property is a UA property, the `setPrimaryPropertyType` action should be called to indicate "ua". Then, the `modules/analytics` store's `selectProperty` action should be used to set the UA property, while the `modules/analytics-4` store's `selectProperty` action should be used to empty/reset the GA4 property (since that then needs to later be chosen in another dropdown based on the GA4 property).
[...]

PR #3383, opened 2021-05-19
Enhancement/3246 implement dropdown UA GA4

danielgent:
## Summary

Addresses issue #3246

## Checklist

### DIFF
--- a/assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.js
+++ b/assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.js
@@ -28,23 +28,23 @@ import { __, sprintf } from '@wordpress/i18n';
 import Data from 'googlesitekit-data';
 import { Select, Option } from '../../../../material-components';
 import ProgressBar from '../../../../components/ProgressBar';
-import { STORE_NAME, PROPERTY_CREATE } from '../../datastore/constants';
-import { MODULES_ANALYTICS } from '../../../analytics/datastore/constants';
-import { isValidAccountID } from '../../../analytics/util';
+import { MODULES_ANALYTICS_4, PROPERTY_CREATE } from '../../../analytics-4/datastore/constants';
+import { MODULES_ANALYTICS } from '../../datastore/constants';
+import { isValidAccountID } from '../../util';
 import { trackEvent } from '../../../../util';
 const { useSelect, useDispatch } = Data;
 
 export default function PropertySelectIncludingGA4() {
 // TODO: Update this select hook to pull accountID from the modules/analytics-4 datastore when GA4 module becomes separated from the Analytics one
 	const accountID = useSelect( ( select ) => select( MODULES_ANALYTICS ).getAccountID() );
-	const properties = useSelect( ( select ) => select( STORE_NAME ).getProperties( accountID ) || [] );
-	const propertyID = useSelect( ( select ) => select( STORE_NAME ).getPropertyID() );
+	const properties = useSelect( ( select ) => select( MODULES_ANALYTICS_4 ).getProperties( accountID ) || [] );
+	const propertyID = useSelect( ( select ) => select( MODULES_ANALYTICS_4 ).getPropertyID() );
 	const isLoading = useSelect( ( select ) => (
 		! select( MODULES_ANALYTICS ).hasFinishedResolution( 'getAccounts' ) ||
-	! select( STORE_NAME ).hasFinishedResolution( 'getProperties', [ accountID ] )
+	! select( MODULES_ANALYTICS_4 ).hasFinishedResolution( 'getProperties', [ accountID ] )
 	) );
 
-	const { selectProperty } = useDispatch( STORE_NAME );
+	const { selectProperty } = useDispatch( MODULES_ANALYTICS_4 );
 
 	const onChange = useCallback( ( index, item ) => {
 		const newPropertyID = item.dataset.value;
--- a/assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.js
+++ b/assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.js
@@ -37,7 +37,7 @@ const { useSelect, useDispatch } = Data;
 export default function PropertySelectIncludingGA4() {
 // TODO: Update this select hook to pull accountID from the modules/analytics-4 datastore when GA4 module becomes separated from the Analytics one
 	const accountID = useSelect( ( select ) => select( MODULES_ANALYTICS ).getAccountID() );
-	const properties = useSelect( ( select ) => select( MODULES_ANALYTICS_4 ).getProperties( accountID ) || [] );
+	const unmappedProperties = useSelect( ( select ) => select( MODULES_ANALYTICS ).getPropertiesIncludingGA4( accountID ) || [] );
 	const propertyID = useSelect( ( select ) => select( MODULES_ANALYTICS_4 ).getPropertyID() );
 	const isLoading = useSelect( ( select ) => (
 		! select( MODULES_ANALYTICS ).hasFinishedResolution( 'getAccounts' ) ||
@@ -48,6 +48,7 @@ export default function PropertySelectIncludingGA4() {
 
 	const onChange = useCallback( ( index, item ) => {
 		const newPropertyID = item.dataset.value;
+		// TODO - see AC
 		if ( propertyID !== newPropertyID ) {
 			selectProperty( newPropertyID );
 			trackEvent( 'analytics_setup', 'property_change', newPropertyID );
@@ -57,6 +58,26 @@ export default function PropertySelectIncludingGA4() {
 	if ( isLoading ) {
 		return <ProgressBar small />;
 	}
+
+	// TEMP FIX SO TESTS PASS
+	const properties = unmappedProperties.map( ( p ) => {
+		// console.log( p );
+		// no idea why nullish... seems like mistake
+		if ( p?._id ) {
+			return {
+				...p,
+				// mapping to be like UA...
+				id: p._id,
+				name: p.displayName,
+			};
+		}
+		return p;
+	} )
+	// undefineds are leaking in somehow
+		.filter( Boolean );
+
+	// TODO - fix undefineds at end
+	// console.debug( properties );
 
 	return (
 		<Select
@@ -70,21 +91,21 @@ export default function PropertySelectIncludingGA4() {
 		>
 			{ ( properties || [] )
 				.concat( {
-					_id: PROPERTY_CREATE,
-					displayName: __( 'Set up a new property', 'google-site-kit' ),
+					id: PROPERTY_CREATE,
+					name: __( 'Set up a new property', 'google-site-kit' ),
 				} )
-				.map( ( { _id, displayName }, index ) => (
+				.map( ( { id, name }, index ) => (
 					<Option
 						key={ index }
-						value={ _id }
+						value={ id }
 					>
-						{ _id === PROPERTY_CREATE
-							? displayName
+						{ id === PROPERTY_CREATE
+							? name
 							: sprintf(
 							/* translators: 1: Property name. 2: Property ID. */
 								__( '%1$s (%2$s)', 'google-site-kit' ),
-								displayName,
-								_id
+								name,
+								id
 							)
 						}
 					</Option>
--- a/assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.js
+++ b/assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.js
@@ -45,21 +45,26 @@ export default function PropertySelectIncludingGA4() {
 	) );
 
 	const { selectProperty } = useDispatch( MODULES_ANALYTICS_4 );
+	const { setPrimaryPropertyType } = useDispatch( MODULES_ANALYTICS );
 
 	const onChange = useCallback( ( index, item ) => {
 		const newPropertyID = item.dataset.value;
 		// TODO - see AC
 		if ( propertyID !== newPropertyID ) {
+			// this should only happen if property is ga4
 			selectProperty( newPropertyID );
 			trackEvent( 'analytics_setup', 'property_change', newPropertyID );
+			// this should only happen if property is ga4
+			setPrimaryPropertyType( 'ga4' );
 		}
-	}, [ propertyID, selectProperty ] );
+	}, [ propertyID, selectProperty, setPrimaryPropertyType ] );
 
 	if ( isLoading ) {
 		return <ProgressBar small />;
 	}
 
 	// TEMP FIX SO TESTS PASS
+	// not sure if should even map! better to instead support this in the component? how will know which is which when clicked?
 	const properties = unmappedProperties.map( ( p ) => {
 		// console.log( p );
 		// no idea why nullish... seems like mistake
--- a/assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.js
+++ b/assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.js
@@ -38,28 +38,33 @@ export default function PropertySelectIncludingGA4() {
 // TODO: Update this select hook to pull accountID from the modules/analytics-4 datastore when GA4 module becomes separated from the Analytics one
 	const accountID = useSelect( ( select ) => select( MODULES_ANALYTICS ).getAccountID() );
 	const unmappedProperties = useSelect( ( select ) => select( MODULES_ANALYTICS ).getPropertiesIncludingGA4( accountID ) || [] );
-	const propertyID = useSelect( ( select ) => select( MODULES_ANALYTICS_4 ).getPropertyID() );
+	const ga4PropertyID = useSelect( ( select ) => select( MODULES_ANALYTICS_4 ).getPropertyID() );
+	const uaPropertyID = useSelect( ( select ) => select( MODULES_ANALYTICS ).getPropertyID() );
 	const isLoading = useSelect( ( select ) => (
 		! select( MODULES_ANALYTICS ).hasFinishedResolution( 'getAccounts' ) ||
 	! select( MODULES_ANALYTICS_4 ).hasFinishedResolution( 'getProperties', [ accountID ] )
 	) );
 
+	const primaryPropertyType = useSelect( ( select ) => select( MODULES_ANALYTICS ).getPrimaryPropertyType() );
+
 	const ga4Dispatch = useDispatch( MODULES_ANALYTICS_4 );
 	const uaDispatch = useDispatch( MODULES_ANALYTICS );
 
+	const propertyID = primaryPropertyType === 'ga4' ? ga4PropertyID : uaPropertyID;
+
 	const onChange = useCallback( ( index, item ) => {
 		// TODO - check this with ua properties
-		// console.debug( 'CLICK property ID ', item.dataset.value );
+		// console.debug( 'item.dataset.value', item.dataset.value );
 
 		const newPropertyID = item.dataset.value;
+		const internalID = item.dataset.internalId; // eslint-disable-line sitekit/acronym-case
 		if ( propertyID !== newPropertyID ) {
 			trackEvent( 'analytics_setup', 'property_change', newPropertyID );
 			if ( newPropertyID.startsWith( 'UA-' ) ) {
-				uaDispatch.selectProperty( newPropertyID );
+				// console.debug( 'selecting UA newPropertyID: ', newPropertyID, internalID );
+				uaDispatch.selectProperty( newPropertyID, internalID );
 				uaDispatch.setPrimaryPropertyType( 'ua' );
 
-				// TODO - this requires lots of changes BUT AC mentions it...
-				// selectPropertyGA4( null );
 				ga4Dispatch.setPropertyID( '' );
 				ga4Dispatch.setWebDataStreamID( '' );
 				ga4Dispatch.setMeasurementID( '' );
@@ -67,8 +72,6 @@ export default function PropertySelectIncludingGA4() {
 				ga4Dispatch.selectProperty( newPropertyID );
 				uaDispatch.setPrimaryPropertyType( 'ga4' );
 
-				// TODO - this requires lots of changes BUT AC mentions it...
-				// selectPropertyUA( null );
 				uaDispatch.setPropertyID( '' );
 				uaDispatch.setInternalWebPropertyID( '' );
 				uaDispatch.setProfileID( '' );
@@ -99,7 +102,9 @@ export default function PropertySelectIncludingGA4() {
 		.filter( Boolean );
 
 	// TODO - fix undefineds at end
-	// console.debug( properties );
+	// console.debug( 'propertyID ', propertyID );
+
+	// console.debug( 'properties ', properties );
 
 	return (
 		<Select
@@ -116,10 +121,11 @@ export default function PropertySelectIncludingGA4() {
 					id: PROPERTY_CREATE,
 					name: __( 'Set up a new property', 'google-site-kit' ),
 				} )
-				.map( ( { id, name }, index ) => (
+				.map( ( { id, name, internalWebPropertyId }, index ) => ( // eslint-disable-line sitekit/acronym-case
 					<Option
 						key={ index }
 						value={ id }
+						data-internal-id={ internalWebPropertyId } // eslint-disable-line sitekit/acronym-case
 					>
 						{ id === PROPERTY_CREATE
 							? name
--- a/assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.js
+++ b/assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.js
@@ -41,8 +41,8 @@ export default function PropertySelectIncludingGA4() {
 	const uaPropertyID = useSelect( ( select ) => select( MODULES_ANALYTICS ).getPropertyID() );
 	const isLoading = useSelect( ( select ) => (
 		! select( MODULES_ANALYTICS ).hasFinishedResolution( 'getAccounts' ) ||
-	! select( MODULES_ANALYTICS_4 ).hasFinishedResolution( 'getProperties', [ accountID ] ) ||
-	! select( MODULES_ANALYTICS ).hasFinishedResolution( 'getProperties', [ accountID ] )
+		! select( MODULES_ANALYTICS_4 ).hasFinishedResolution( 'getProperties', [ accountID ] ) ||
+		! select( MODULES_ANALYTICS ).hasFinishedResolution( 'getProperties', [ accountID ] )
 	) );
 
 	const primaryPropertyType = useSelect( ( select ) => select( MODULES_ANALYTICS ).getPrimaryPropertyType() );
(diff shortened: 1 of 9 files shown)


assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.js
```diff
@@ -0,0 +1,133 @@
...
+		return <ProgressBar small />;
+	}
+
+	const properties = unmappedProperties.map( ( p ) => {
+		if ( p?._id ) {
+			return {
+				...p,
+				// map GA4 properties to have fields like UA
+				id: p._id,
+				name: p.displayName,
```
danielgent, line 92: There are a few different ways we can display UA and GA4 properties together

We have tests anyway so it's easy to refactor this :+1:

assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.test.js
```diff
@@ -0,0 +1,174 @@
...
+	it( 'should render a select box with only an option to create a new property if no properties are available.', async () => {
+		const { getAllByRole } = render( <PropertySelect />, { setupRegistry: setupEmptyRegistry } );
+
+		const listItems = getAllByRole( 'menuitem', { hidden: true } );
+		expect( listItems ).toHaveLength( 1 );
+		expect( listItems[ 0 ].textContent ).toMatch( /set up a new property/i );
+	} );
+
+	it( 'should change between UA and GA4 properties', async () => {
+		const { getAllByRole, container, registry } = render( <PropertySelect />, { setupRegistry } );
```
danielgent, line 119: it's a long test but I combined a few together.  It gives me confidence anyway that the many parts of our state are being changed correctly!

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -476,6 +476,10 @@ const baseSelectors = {
 	getPropertiesIncludingGA4: createRegistrySelector( ( select ) => ( state, accountID ) => {
 		let properties = select( STORE_NAME ).getProperties( accountID );
 
+		if ( ! properties ) {
+			return undefined;
+		}
```
danielgent, line 481: This looks like an oversite on my behalf!  I also think that below it should return `undefined` if `propertiesGA4` is `undefined`  (AFAIK we cascade our `undefined` loading state like this)

assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -308,6 +308,8 @@ describe( 'modules/analytics properties', () => {
 		} );
 
 		describe( 'getPropertiesIncludingGA4', () => {
+			it.todo( 'returns undefined if UA or GA properties are loading' );
```
danielgent, line 311: Happy to flesh this out! (or create a ticket TODO this)

stories/module-analytics.stories.js
```diff
@@ -124,6 +127,44 @@ storiesOf( 'Analytics Module', module )
...
+
+		return (
+			<WithTestRegistry callback={ setupRegistry }>
+				<SetupWrap>
+					<div className="googlesitekit-setup-module__inputs">
+						<PropertySelectIncludingGA4 />
+					</div>
+				</SetupWrap>
+			</WithTestRegistry>
+		);
```
danielgent, line 166: Tested and working https://google.github.io/site-kit-wp/storybook/pull/3383/?path=/story/analytics-module--property-select-including-ga4-properties

assets/js/modules/analytics/datastore/properties.js
danielgent: To elaborate, without this fix the selector crashes as it calls `properties.concat` below. Instead I could just pass an array as a default argument, but doing this seems to fit more with what we do with loading

assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.js
```diff
@@ -0,0 +1,133 @@
...
+export default function PropertySelectIncludingGA4() {
+	const accountID = useSelect( ( select ) => select( MODULES_ANALYTICS ).getAccountID() );
+	const unmappedProperties = useSelect( ( select ) => select( MODULES_ANALYTICS ).getPropertiesIncludingGA4( accountID ) || [] );
+	const ga4PropertyID = useSelect( ( select ) => select( MODULES_ANALYTICS_4 ).getPropertyID() );
+	const uaPropertyID = useSelect( ( select ) => select( MODULES_ANALYTICS ).getPropertyID() );
+	const isLoading = useSelect( ( select ) => (
+		! select( MODULES_ANALYTICS ).hasFinishedResolution( 'getAccounts' ) ||
+	! select( MODULES_ANALYTICS_4 ).hasFinishedResolution( 'getProperties', [ accountID ] ) ||
+	! select( MODULES_ANALYTICS ).hasFinishedResolution( 'getProperties', [ accountID ] )
+	) );
```
eugene-manuilov, line 46:
```suggestion
	const isLoading = useSelect( ( select ) => (
		! select( MODULES_ANALYTICS ).hasFinishedResolution( 'getAccounts' ) ||
		! select( MODULES_ANALYTICS_4 ).hasFinishedResolution( 'getProperties', [ accountID ] ) ||
		! select( MODULES_ANALYTICS ).hasFinishedResolution( 'getProperties', [ accountID ] )
	) );
```

assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.js
```diff
@@ -0,0 +1,133 @@
...
+	const uaDispatch = useDispatch( MODULES_ANALYTICS );
+
+	const propertyID = primaryPropertyType === 'ga4' ? ga4PropertyID : uaPropertyID;
+
+	const onChange = useCallback( ( index, item ) => {
+		const newPropertyID = item.dataset.value;
+		const internalID = item.dataset.internalId; // eslint-disable-line sitekit/acronym-case
+		if ( propertyID !== newPropertyID ) {
+			trackEvent( 'analytics_setup', 'property_change', newPropertyID );
+			if ( newPropertyID.startsWith( 'UA-' ) ) {
```
eugene-manuilov, line 60: We should better check whether the `internalID` is empty or not. GA4 properties don't have it, so we can rely on this field to understand what we should do:

```suggestion
			if ( !! internalID ) {
```

assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.js
```diff
@@ -0,0 +1,133 @@
...
+		return p;
+	} );
+
+	return (
+		<Select
+			className="googlesitekit-analytics__select-property"
+			label={ __( 'Property', 'google-site-kit' ) }
+			value={ propertyID }
+			onEnhancedChange={ onChange }
+			disabled={ ! isValidAccountID( accountID ) }
```
eugene-manuilov, line 104: This line doesn't make sense anymore because we already return `null` if the `accountID` is not valid. Let's get rid of this attribute then:

```suggestion
```

assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.js
```diff
@@ -0,0 +1,133 @@
...
+			enhanced
+			outlined
+		>
+			{ ( properties || [] )
+				.concat( {
+					id: PROPERTY_CREATE,
+					name: __( 'Set up a new property', 'google-site-kit' ),
+				} )
```
eugene-manuilov, line 112: @felixarntz if a user selects this option, which property do we need to create? ua or ga4? :thinking:

assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.test.js
```diff
@@ -0,0 +1,174 @@
...
+	properties: propertiesGA4,
+	webDataStreams,
+} = analytics4Fixtures;
+const { accounts, properties: propertiesUA, profiles } = fixtures.accountsPropertiesProfiles;
+const accountID = createProperty._accountID;
+const propertyIDga4 = createWebDataStream._propertyID;
+const propertyIDua = propertiesUA[ 0 ].id;
+
+const setupRegistry = ( { dispatch } ) => {
+	dispatch( CORE_SITE ).receiveSiteInfo( { referenceSiteURL: 'http://example.com' } );
```
eugene-manuilov, line 43: Use `provideSiteInfo( registry )` instead of this.

assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.test.js
```diff
@@ -0,0 +1,174 @@
...
+	dispatch( MODULES_ANALYTICS ).receiveGetSettings( {} );
+	dispatch( MODULES_ANALYTICS_4 ).receiveGetSettings( {} );
+	dispatch( MODULES_ANALYTICS ).setAccountID( accountID );
```
eugene-manuilov, line 46:
```suggestion
	dispatch( MODULES_ANALYTICS ).receiveGetSettings( { accountID } );
	dispatch( MODULES_ANALYTICS_4 ).receiveGetSettings( {} );
```

assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.test.js
```diff
@@ -0,0 +1,174 @@
...
+	dispatch( MODULES_ANALYTICS ).finishResolution( 'getProfiles', [ accountID, propertyIDua ] );
+
+	dispatch( MODULES_ANALYTICS_4 ).receiveGetWebDataStreams( webDataStreams, { propertyID: propertyIDga4 } );
+	dispatch( MODULES_ANALYTICS_4 ).finishResolution( 'receiveGetWebDataStreams', { propertyID: propertyIDga4 } );
+};
+
+const setupEmptyRegistry = ( { dispatch } ) => {
+	dispatch( MODULES_ANALYTICS ).receiveGetSettings( {} );
+	dispatch( MODULES_ANALYTICS_4 ).receiveGetSettings( {} );
+	dispatch( MODULES_ANALYTICS ).setAccountID( accountID );
```
eugene-manuilov, line 68:
```suggestion
	dispatch( MODULES_ANALYTICS ).receiveGetSettings( { accountID } );
	dispatch( MODULES_ANALYTICS_4 ).receiveGetSettings( {} );
```

assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.test.js
```diff
@@ -0,0 +1,174 @@
...
+
+	dispatch( MODULES_ANALYTICS ).receiveGetProperties( [], { accountID } );
+	dispatch( MODULES_ANALYTICS ).finishResolution( 'getProperties', [ accountID ] );
+
+	dispatch( MODULES_ANALYTICS_4 ).receiveGetProperties( [], { accountID } );
+	dispatch( MODULES_ANALYTICS_4 ).finishResolution( 'getProperties', [ accountID ] );
+};
+
+describe( 'PropertySelectIncludingGA4', () => {
+	it( 'should render an option for each analytics property of the currently selected account.', async () => {
```
eugene-manuilov, line 81: `async` isn't needed here because we don't wait for anything in the test. Please, remove it from all tests that don't use async/await functionality.

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -476,6 +476,10 @@ const baseSelectors = {
 	getPropertiesIncludingGA4: createRegistrySelector( ( select ) => ( state, accountID ) => {
 		let properties = select( STORE_NAME ).getProperties( accountID );
 
+		if ( ! properties ) {
```
eugene-manuilov, line 479:
```suggestion
		if ( properties === undefined ) {
```

assets/js/modules/analytics/datastore/properties.js
eugene-manuilov, line 481: We also need to make sure that we return `undefined` when ga4 properties are being loaded too (when `ga4setup` feature is enabled).

assets/js/modules/analytics/datastore/properties.test.js
eugene-manuilov: Yes, let's create a test for it in the scope of this ticket.

stories/module-analytics.stories.js
```diff
@@ -124,6 +127,44 @@ storiesOf( 'Analytics Module', module )
 			</WithTestRegistry>
 		);
 	} )
+	.add( 'Property Select including GA4 properties', () => {
+		const { accounts, properties, profiles } = fixtures.accountsPropertiesProfiles;
+		// eslint-disable-next-line sitekit/acronym-case
+		const accountID = properties[ 0 ].accountId;
+		// eslint-disable-next-line sitekit/acronym-case
+		const propertyID = profiles[ 0 ].webPropertyId;
```
eugene-manuilov, line 135:
```suggestion
		/* eslint-disable sitekit/acronym-case */
		const accountID = properties[ 0 ].accountId;
		const propertyID = profiles[ 0 ].webPropertyId;
		/* eslint-enable */
```

stories/module-analytics.stories.js
```diff
@@ -124,6 +127,44 @@ storiesOf( 'Analytics Module', module )
...
+		const setupRegistry = ( { dispatch } ) => {
+			dispatch( STORE_NAME ).receiveGetAccounts( accounts );
+			dispatch( STORE_NAME ).finishResolution( 'getAccounts', [] );
+
+			// eslint-disable-next-line sitekit/acronym-case
+			dispatch( STORE_NAME ).receiveGetProperties( properties, { accountID: properties[ 0 ].accountId } );
+			dispatch( STORE_NAME ).receiveGetProfiles( profiles, {
+				accountID,
+				// eslint-disable-next-line sitekit/acronym-case
```
eugene-manuilov, line 144: We don't need this here because the next line is correct.

stories/module-analytics.stories.js
```diff
@@ -124,6 +127,44 @@ storiesOf( 'Analytics Module', module )
...
+				propertyID,
+			} );
+
+			dispatch( STORE_NAME ).receiveGetSettings( {
+				accountID,
+			} );
+			// eslint-disable-next-line sitekit/acronym-case
```
eugene-manuilov, line 151: Same here, the next line is fine, we don't need this.

eugene-manuilov (changes requested):
Thanks, @danielgent. This is a good start. Left some comments. Please, take a look.

assets/js/modules/analytics/datastore/properties.js
danielgent: Done and added tests for this

assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.test.js
danielgent: Done for all the new tests I've added @eugene-manuilov . Nice spot! :smile:

stories/module-analytics.stories.js
```diff
@@ -124,6 +127,42 @@ storiesOf( 'Analytics Module', module )
...
+		const setupRegistry = ( { dispatch } ) => {
+			dispatch( STORE_NAME ).receiveGetAccounts( accounts );
+			dispatch( STORE_NAME ).finishResolution( 'getAccounts', [] );
+
+			// eslint-disable-next-line sitekit/acronym-case
+			dispatch( STORE_NAME ).receiveGetProperties( properties, { accountID: properties[ 0 ].accountId } );
+			dispatch( STORE_NAME ).receiveGetProfiles( profiles, {
+				accountID,
+				propertyID,
+			} );
```
danielgent, line 145: need-profiles-for-each-option

How concerned are you about this @eugene-manuilov ?  

We already have console errors on storybook which, while not ideal, doesn't break anything

Like on the other story on the other PR, I need to add profiles for each `account` and `property` ID

assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.js
```diff
@@ -0,0 +1,131 @@
...
+		if ( p?._id ) {
+			return {
+				...p,
+				// map GA4 properties to have fields like UA
+				id: p._id,
+				name: p.displayName,
+			};
+		}
+		return p;
+	} );
```
eugene-manuilov, line 96: Let's simplify it a little bit:

```suggestion
	const properties = unmappedProperties.map( ( p ) => ( {
		...p,
		id: p._id || p.id,
		name: p.displayName || p.name,
	} ) );
```

assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.js
```diff
@@ -0,0 +1,131 @@
...
+				.map( ( { id, name, internalWebPropertyId }, index ) => ( // eslint-disable-line sitekit/acronym-case
+					<Option
+						key={ index }
+						value={ id }
+						data-internal-id={ internalWebPropertyId } // eslint-disable-line sitekit/acronym-case
+					>
+						{ id === PROPERTY_CREATE
+							? name
+							: sprintf(
+							/* translators: 1: Property name. 2: Property ID. */
```
eugene-manuilov, line 121:
```suggestion
								/* translators: 1: Property name. 2: Property ID. */
```

assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.js
```diff
@@ -0,0 +1,131 @@
...
+								__( '%1$s (%2$s)', 'google-site-kit' ),
```
eugene-manuilov, line 122: Let's use `_x` instead of `__` here to provide context:

```suggestion
								_x( '%1$s (%2$s)', '{property name} ({property id})', 'google-site-kit' ),
```

assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.test.js
```diff
@@ -0,0 +1,173 @@
+/**
+ * GA4 Property Select component tests.
```
eugene-manuilov, line 2:
```suggestion
 * PropertySelectIncludingGA4 component tests.
```

assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.js
```diff
@@ -0,0 +1,131 @@
+/**
+ * GA4 Property Select component.
```
eugene-manuilov, line 2:
```suggestion
 * PropertySelectIncludingGA4 component.
```

assets/js/modules/analytics/datastore/properties.js
```diff
@@ -476,8 +476,15 @@ const baseSelectors = {
 	getPropertiesIncludingGA4: createRegistrySelector( ( select ) => ( state, accountID ) => {
 		let properties = select( STORE_NAME ).getProperties( accountID );
 
+		if ( properties === undefined ) {
+			return undefined;
+		}
+
 		if ( select( MODULES_ANALYTICS_4 ) ) {
```
eugene-manuilov, line 483: I know this is not relevant for this ticket, but let's check that the `ga4setup` feature is enabled instead of checking its datastore:

```suggestion
		if ( isFeatureEnabled( 'ga4setup' ) ) {
```

assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -308,7 +309,82 @@ describe( 'modules/analytics properties', () => {
...
 		describe( 'getPropertiesIncludingGA4', () => {
-			it( 'returns a sorted list of ua and ga4 properties ', async () => {
+			it( 'returns undefined if UA properties are loading', () => {
+				const testAccountID = fixtures.profiles[ 0 ].accountId; // eslint-disable-line sitekit/acronym-case
+				const accountID = testAccountID;
+
+				fetchMock.get(
+					/^\/google-site-kit\/v1\/modules\/analytics\/data\/properties-profiles/,
+					{ body: fixtures.propertiesProfiles, status: 200 }
+				);
```
eugene-manuilov, line 319: Let's use `freezeFetch` instead of `fetchMock.get`.

assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -308,7 +309,82 @@ describe( 'modules/analytics properties', () => {
 		} );
 
 		describe( 'getPropertiesIncludingGA4', () => {
-			it( 'returns a sorted list of ua and ga4 properties ', async () => {
+			it( 'returns undefined if UA properties are loading', () => {
+				const testAccountID = fixtures.profiles[ 0 ].accountId; // eslint-disable-line sitekit/acronym-case
+				const accountID = testAccountID;
```
eugene-manuilov, line 314: We don't need `testAccountID`, just use `accountID`:
```suggestion
				const accountID = fixtures.profiles[ 0 ].accountId; // eslint-disable-line sitekit/acronym-case
```

Same for other tests.

stories/module-analytics.stories.js
eugene-manuilov: Yeah, we don't need to make it fully functional and recreate the whole environment. That error appears only when you select something. So we can ignore it.

eugene-manuilov, line 145: I believe we don't need profiles for this story, do we? Let's get rid of it if we don't need it.

eugene-manuilov (changes requested):
Thanks, @danielgent. This is much better now. Added a few more comments. Please, take a look.

assets/js/modules/analytics/datastore/properties.js
danielgent: Makes sense @eugene-manuilov !  IIRC you said to do the same change elsewhere on another PR :smile:

stories/module-analytics.stories.js
danielgent: If I comment it out then `select( MODULES_ANALYTICS_4 ).hasFinishedResolution( 'getProperties', [ accountID ] );` is always `false`

stories/module-analytics.stories.js
```diff
@@ -124,6 +128,51 @@ storiesOf( 'Analytics Module', module )
 			</WithTestRegistry>
 		);
 	} )
+	.add( 'Property Select including GA4 properties', () => {
+		// Have to import like with tests
+		enabledFeatures.add( 'ga4setup' );
```
danielgent, line 133: Due to the changes in `analytics/datastore/properties.js` to use `isFeatureEnabled`, I had to enable `ga4setup` in the story

But it only worked like this

stories/module-analytics.stories.js
```diff
@@ -124,6 +128,51 @@ storiesOf( 'Analytics Module', module )
...
+				propertiesGA4,
+				{ accountID }
+			);
+		};
+
+		return (
+			<WithTestRegistry
+				callback={ setupRegistry }
+				// This does not work
+				features={ [ 'ga4setup' ] }
```
danielgent, line 165: This does not cause the feature flag to be set in `analytics/datastore/properties.js`

assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.js
```diff
@@ -0,0 +1,127 @@
...
+					<Option
+						key={ index }
+						value={ id }
+						data-internal-id={ internalWebPropertyId } // eslint-disable-line sitekit/acronym-case
+					>
+						{ id === PROPERTY_CREATE
+							? name
+							: sprintf(
+								/* translators: 1: Property name. 2: Property ID. */
+								__( '%1$s (%2$s)', 'google-site-kit', 'google-site-kit' ),
```
eugene-manuilov, line 118: @danielgent the second parameter must be a context, let's better use something like this:

```suggestion
								__( '%1$s (%2$s)', '{property name} ({property id})', 'google-site-kit' ),
```

assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -38,6 +41,8 @@ describe( 'modules/analytics properties', () => {
 	} );
 
 	beforeEach( () => {
+		enabledFeatures.add( 'ga4setup' );
```
eugene-manuilov, line 44: This is only needed for tests in the `getPropertiesIncludingGA4` group. Let's define a new `beforeEach` in that group and enable the `ga4setup` feature there.

assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -308,7 +313,78 @@ describe( 'modules/analytics properties', () => {
...
+							accountId: '151753095',
+							id: 'UA-151753095-1',
+							name: 'troubled-tipped.example.com',
+						},
+
+					],
+					{ accountID }
+				);
+
+				fetchMock.get(
```
eugene-manuilov, line 363: We need to use `freezeFetch` if we want to test the loading state.

assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -308,7 +313,78 @@ describe( 'modules/analytics properties', () => {
...
+					{ body: ga4Fixtures.properties, status: 200 }
+				);
+
+				expect( registry.select( STORE_NAME ).getPropertiesIncludingGA4( testAccountID ) ).toBeUndefined();
+			} );
+
+			it( 'returns undefined if both UA and GA4 properties are loading', () => {
+				const testAccountID = fixtures.profiles[ 0 ].accountId; // eslint-disable-line sitekit/acronym-case
+
+				fetchMock.get(
```
eugene-manuilov, line 374: Same here, we need to use `freezeFetch` instead.

assets/js/modules/analytics/datastore/properties.test.js
```diff
@@ -308,7 +313,78 @@ describe( 'modules/analytics properties', () => {
...
+					/^\/google-site-kit\/v1\/modules\/analytics\/data\/properties-profiles/,
+					{ body: fixtures.propertiesProfiles, status: 200 }
+				);
+
+				fetchMock.get(
```
eugene-manuilov, line 379: And here. Use `freezeFetch`.

stories/module-analytics.stories.js
eugene-manuilov: It uses a slightly different approach... we may need to revisit it i think :thinking: but it's out of scope for this ticket.

eugene-manuilov (changes requested):
Thanks, @danielgent. It's almost good to go. Just a few more adjustments required. Please, take a look.

assets/js/modules/analytics/components/common/PropertySelectIncludingGA4.js
danielgent: @eugene-manuilov  Now the linting fails with
`118:9  error  Invalid text domain '{property name} ({property id})'  @wordpress/i18n-text-domain`

I'm not sure what this means and (from a quick look) I can't find any examples that have this argument (all the calls to `sprinf` seem to have just two arguments)

eugene-manuilov: @danielgent use `_x` instead of `__` and check out https://github.com/WordPress/gutenberg/tree/trunk/packages/i18n when you have some spare time.

eugene-manuilov (approved):
Ok, great. Looks good now.

PR #3386, opened 2021-05-19
Fix/3247 loading issue

eugene-manuilov:
## Summary

Addresses issue #3247 

## Checklist

### DIFF
--- a/assets/js/modules/analytics-4/datastore/api.js
+++ b/assets/js/modules/analytics-4/datastore/api.js
@@ -46,20 +46,17 @@ const baseSelectors = {
 		for ( const errorKey in state.errors ) {
 			if (
 				errorKey.startsWith( 'getProperties::' ) ||
-				errorKey.startsWith( 'getWebDataStreams::' )
+				errorKey.startsWith( 'getWebDataStreams::' ) ||
+				errorKey.startsWith( 'getAccountSummaries::' )
 			) {
 				return false;
 			}
 		}
 
-		for ( const propertyList of Object.values( state.properties ) ) {
-			if ( propertyList && propertyList.length && propertyList.length > 0 ) {
-				for ( const webdatastreamList of Object.values( state.webdatastreams ) ) {
-					if ( webdatastreamList && webdatastreamList.length && webdatastreamList.length > 0 ) {
-						return true;
-					}
-				}
-			}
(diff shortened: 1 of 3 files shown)


assets/js/modules/analytics-4/datastore/api.js
```diff
@@ -46,20 +46,17 @@ const baseSelectors = {
...
-				for ( const webdatastreamList of Object.values( state.webdatastreams ) ) {
-					if ( webdatastreamList && webdatastreamList.length && webdatastreamList.length > 0 ) {
-						return true;
-					}
-				}
-			}
+		// We expect that properties have been already pulled for one of the accounts.
+		// Make sure to call `getProperties` selector before relying on this one.
+		if ( Object.keys( state.properties ).length > 0 ) {
+			return true;
```
felixarntz, line 59: @eugene-manuilov Based on my previous feedback, why not use `hasFinishedResolution` here? I think that's more appropriate than relying on specific data which right now only works for one of those three selectors.

felixarntz (commented):
One point of feedback here, otherwise lgtm.

assets/js/modules/analytics-4/datastore/api.js
eugene-manuilov: @felixarntz, unfortunately, we can't use `hasFinishedResolution` because we don't know which arguments have been used in the `getProperties` selector call.

felixarntz: @eugene-manuilov Ah, I see now that there is no `hasFinishedResolution` alternative that works for _any_ call to that selector. `@wordpress/data` has some other _actions_ that work on the entire store (e.g. `invalidateResolutionForStoreSelector`).

Maybe we could still do that though, almost implement our own `hasFinishedResolutionForStoreSelector`: We would need to check whether [this map for the entire selector](https://github.com/WordPress/gutenberg/blob/trunk/packages/data/src/redux-store/metadata/selectors.js#L19) has no resolved calls - which in principle we could access via [the `getCachedResolvers` selector](https://github.com/WordPress/gutenberg/blob/trunk/packages/data/src/redux-store/metadata/selectors.js#L76). It would be a cleaner solution, but it would also require us to implement something that would be more appropriate to have in `@wordpress/data` itself - so what you currently have works for now.

felixarntz (approved):

Issue #3225, opened 2021-04-26
Ensure widgets in an area are combined beyond their row if all of them have the same special state


felixarntz:
This is a follow-up to #2252, covering a use-case that previously hasn't been considered but is relevant for e.g. the Site Kit module pages, see screenshot:
<img width="897" alt="Screen Shot 2021-04-26 at 10 52 34 AM" src="https://user-images.githubusercontent.com/3531426/116128224-8fe1e100-a67d-11eb-9d31-f4d71ffc13bc.png">

The logic to combine widgets in the same widget area with similar states normally only combines widgets which are visually in the same row. This should generally be maintained, however with one exception: If _all_ widgets in the widget area have the same state, they should all be combined, resulting in a single full-width widget with that state.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* If _all_ widgets within a widget area have the same special state (via `getWidgetState`, e.g. `ReportZero`, `ModuleActivateCTA`, `CompleteModuleActivationCTA`), they should all be combined into a single widget box which spans the full width of 12 columns.
    * So far, this does not happen because widgets are only combined if they are in a single row within the widget area (see screenshot). This behavior should be maintained generally, it's really only for the specific situation that all widgets in the widget area have the same state that the new behavior mentioned above should be implemented.
    * This basically requires changes to `combineWidgets`, e.g. a new loop before the main one that checks if _all_ widgets have the same state and if so applies the required behavior (use override component based on the common state for one of the widgets and set the width to 12, then hide all other widgets).

## Implementation Brief

In `combine-widgets`, before the main loop, add an additional loop which cycles through all the widgets and:
  * checks if all the widgets in the array share the same state (via `widgetStates` object, e.g. if they all return `ReportZero`, `ModuleActivateCTA`, `CompleteModuleActivationCTA`)
  * if they do all share the same state, then 
    * one of the widgets should be full width (i.e. `gridColumnWidths[i]` is `12` and the others should be hidden (i.e. `gridColumnWidths[i]` is `0`)
    * the returned `overrideComponents` array should contain the component state to be rendered, e.g. 
    ```js
    // Example where 3 widgets share the same state:
    return {
        overrideComponents: [ widgetStates[ widgets[ 0 ].slug ] ],
        gridColumnWidths: [ 12, 0, 0 ],
    };
    ```

### Test Coverage

[...]

PR #3394, opened 2021-05-24
Ensure widgets in an area are combined beyond their row if all of them have the same special state (3225).

johnPhillips:
## Summary

Addresses issue #3225

## Relevant technical choices
Although in the AC / IB it suggested inserting a check in the `combineWidgets` function I ended up pulling it out into a separate function as I felt `combineWidgets` was getting a bit long. Happy to revert if required.
I also added a `SPECIAL_WIDGET_STATES` constant as there are only certain states where we want to combine the widgets and I couldn't see another way around this.

## Checklist

### DIFF
--- a/assets/js/googlesitekit/widgets/util/combine-widgets.js
+++ b/assets/js/googlesitekit/widgets/util/combine-widgets.js
@@ -21,6 +21,11 @@
  */
 import isEqual from 'lodash/isEqual';
 
+/**
+ * Internal dependencies
+ */
+import { SPECIAL_WIDGET_STATES } from './constants';
+
 function stateAndRowMatch( stateA, stateB, rowA, rowB ) {
 	return rowA === rowB && isEqual( stateA, stateB );
 }
@@ -57,16 +62,35 @@ export function combineWidgets( widgets, widgetStates, {
 	let currentRowIndex = -1;
 	let columnWidthsBuffer = [];
 
-	const states = [];
+	const states = {};
+	/*
+	 * All the widgets here will be in the same area
+	 * We need to determine whether:
+	 * - They are all the same module
+	 * - They are all in the same special state
+	 */
 
 	widgets.forEach( ( widget ) => {
-		const state = widgetStates?.[widget.slug]?.Component?.name;
-		if ( state && ! states.includes( state ) ) {
-			states.push( state );
+		const widgetState = widgetStates?.[widget.slug];
+		const state = widgetState?.Component?.name;
+		const module = widgetState?.metadata?.moduleSlug;
+		const isSpecialState = SPECIAL_WIDGET_STATES.includes( state );
+
+		if ( ! state || ! module || ! isSpecialState ) {
+			// break;
+		}
+
+		if ( states[ module ] ) {
+			states[ module ].push( state );
+		} else {
+			states[ module ] = [ state ];
 		}
 	} );
 
-	if ( states.length === 1 && widgets.length > 1 ) {
+	const isMultipleModules = Object.keys( states ).length > 1;
+	const allStatesEqual = Object.entries( states ).every( ( [ module, states ] ) => state === states[ 0 ] );
+
+	if ( ! isMultipleModules && allStatesEqual && states.length === widgets.length ) {
 		// All widgets have the same state, so we should only render one and hide the rest.
 		const hiddenRows = Array.from( { length: widgets.length - 1 } ).fill( 0 );
 		// All the components are the same, pick the first.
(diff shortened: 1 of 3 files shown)


assets/js/googlesitekit/widgets/util/combine-widgets.js
```diff
@@ -21,11 +21,62 @@
...
+	const states = {};
+
+	for ( let i = 0; i < widgets.length; i++ ) {
+		const widget = widgets[ i ];
+		const widgetState = widgetStates?.[widget.slug];
+
+		const state = widgetState?.Component?.name;
+		const module = widgetState?.metadata?.moduleSlug;
+		const isSpecialState = SPECIAL_WIDGET_STATES.includes( state );
+		const isMultipleModules = Object.keys( states ).length > 1;
```
eugene-manuilov, line 59: It won't return `false` if the last widget is from another module: `[ analytics-widget-A, analytics-widget-B, search-console-widget-A ]`

eugene-manuilov (changes requested):
Thanks, @johnPhillips. It seems to be a good start. Please, take a look at my comments. Two of them are just minor formatting issues, the last one is more important.

assets/js/googlesitekit/widgets/util/combine-widgets.js
johnPhillips: Great catch! Thank you.

assets/js/googlesitekit/widgets/util/combine-widgets.js
```diff
@@ -21,11 +21,68 @@
...
+			states[ module ] = [ state ];
+		}
+	}
+
+	if ( Object.keys( states ).length > 1 ) {
+		// We have multiple modules, so the widgets can't be combined.
+		return false;
+	}
+
+	// const isMultipleModules = Object.keys( states ).length > 1;
```
eugene-manuilov, line 81: This is not needed, is it?
```suggestion
```

assets/js/googlesitekit/widgets/util/combine-widgets.js
```diff
@@ -21,11 +21,68 @@
...
+		if ( states[ module ] ) {
+			if ( states[ module ].includes( state ) === false ) {
+				// We have a widget from a given module in a different state from the others.
+				return false;
+			}
+
+			states[ module ].push( state );
+		} else {
+			states[ module ] = [ state ];
+		}
```
eugene-manuilov, line 73: How about this?

```suggestion
		if ( states[ module ] ) {
			if ( states[ module ] !== state ) {
				// We have a widget from a given module in a different state from the others.
				return false;
			}
		} else {
			states[ module ] = state;
		}
```

eugene-manuilov (changes requested):
Ok, great. Seems to be good as far as I see. Left two tiny comments and it should be good to go.

eugene-manuilov (approved):
LGTM!

assets/js/googlesitekit/widgets/util/constants.js
```diff
@@ -28,3 +28,8 @@ export const WIDTH_GRID_COUNTER_MAP = {
 };
 
 export const HIDDEN_CLASS = 'googlesitekit-hidden';
+export const SPECIAL_WIDGET_STATES = [
+	'ReportZero',
+	'ModuleActivateCTA',
+	'CompleteModuleActivationCTA',
```
aaemnnosttv, line 34: I don't think we should reference component names as strings here. We should be able to import those components directly for comparison where needed. E.g. https://github.com/google/site-kit-wp/blob/1198c7ed8bcaaf09dd7fc9597be17c2352dbfc97/assets/js/googlesitekit/widgets/util/is-inactive-widget-state.js#L33

We preserve the names of component functions in minified production builds, but if that were to ever change I think this name based reference would break.

assets/js/googlesitekit/widgets/util/combine-widgets.test.js
```diff
@@ -30,13 +30,64 @@ import Null from '../../../components/Null';
...
 	const getHalfWidget = ( slug ) => ( { slug, width: WIDGET_WIDTHS.HALF } );
+	const getFullWidget = ( slug ) => ( { slug, width: WIDGET_WIDTHS.FULL } );
 
 	const getRegularState = () => null;
 	const getReportZeroState = ( moduleSlug ) => ( { Component: ReportZero, metadata: { moduleSlug } } );
 	const getActivateModuleCTAState = ( moduleSlug ) => ( { Component: ActivateModuleCTA, metadata: { moduleSlug } } );
 	const getCompleteModuleActivationCTAState = ( moduleSlug ) => ( { Component: CompleteModuleActivationCTA, metadata: { moduleSlug } } );
 	const getNullState = () => ( { Component: Null, metadata: {} } );
 
+	it( 'Widgets in an area are combined beyond their row if all of them have the same special state', () => {
```
aaemnnosttv, line 41: Test names should read as a sentence starting with "it ..." but these do not. A minor detail but consistency is important.

aaemnnosttv (commented):
@eugene-manuilov – a few minor things that would be good to clean up, particularly the special widget state comparison.

Issue #3254, opened 2021-04-29
Update Analytics settings UI to include specified GA4 property (if set)


felixarntz:
If the `modules/analytics-4` store is loaded (i.e. the `ga4setup` feature flag is enabled) and a GA4 property is set, the Analytics settings panel should allow controlling the selected GA4 property.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* If the `ga4setup` feature flag is enabled, the Analytics `SettingsView` should display the value for the currently used GA4 property _if one is set_. If none is set, no information about this setting should be displayed entirely.
* The Analytics `SettingsEdit` component should contain UI to specify the GA4 property that matches the UA property currently chosen, relying on `getSetupFlowMode`:
    * If it returns "legacy" or "ga4" (the latter of which would be a special case here where the current user doesn't have access to the currently selected property), nothing about GA4 should show.
    * If it returns "ua", the [notification that is also used in the setup](https://www.figma.com/file/ix0pcdTQi1Bf9QFNBlBByB/GA4-%2B-UA-onboarding?node-id=20269%3A8621) to indicate a new GA4 property will be created should be shown. In that case, the `propertyID` setting of the `modules/analytics-4` store should automatically be set to `PROPERTY_CREATE`.
    * If it returns "ga4transitional", the notification with a GA4 property dropdown from [this Figma design](https://www.figma.com/file/ix0pcdTQi1Bf9QFNBlBByB/GA4-%2B-UA-onboarding?node-id=20269%3A9075) should be shown, relying on the `PropertySelect` component of the `analytics-4` module.
        * In other words: Differently from setup where for this case there's a dropdown with all GA4 and UA properties (see #3250), in the settings UI the UA property is for now always the "primary" property.

## Implementation Brief
* Using `SettingsView`,
    * Query the `analytics/4` data store using the `getPropertyID` selector to check for the currently used GA4 property.
    * If this is the case, then display the property ID using the `DisplaySetting` component like it's done right now with the non GA4 property.
* Using `SettingsEdit`,
    * Query the `analytics` data store using the `getSetupFlowMode` selector.
    * Based on the value returned by `getSetupFlowMode`,
        * If it returns `SETUP_FLOW_MODE_LEGACY` or `SETUP_FLOW_MODE_GA4`, nothing about GA4 should be displayed.
        * If it returns `SETUP_FLOW_MODE_UA`,
[...]

PR #3407, opened 2021-05-25
Feature/3254 settings UI ga4

ivankruchkoff:
## Summary

Addresses issue #3254

## Checklist


asvinb:
@googlebot I consent.

assets/js/modules/analytics/components/settings/SettingsView.js
```diff
@@ -37,6 +38,7 @@ const { useSelect } = Data;
 export default function SettingsView() {
 	const accountID = useSelect( ( select ) => select( STORE_NAME ).getAccountID() );
 	const propertyID = useSelect( ( select ) => select( STORE_NAME ).getPropertyID() );
+	const propertyID4 = useSelect( ( select ) => select( MODULES_ANALYTICS_4 ).getPropertyID() );
```
eugene-manuilov, line 41: We should rely on `const isGA4Enabled = useFeature( 'ga4setup' );` when we interact with the `analytics-4` module functionality. In other words we should try to pull `propertyID` only if the `isGA4Enabled` is true.

assets/js/modules/analytics/components/settings/SettingsForm.js
```diff
@@ -32,17 +37,42 @@ import {
...
+	SETUP_FLOW_MODE_UA,
+	STORE_NAME,
+} from '../../datastore/constants';
+import { MODULES_ANALYTICS_4, PROPERTY_CREATE } from '../../../analytics-4/datastore/constants';
+import SettingsNotice, { TYPE_INFO } from '../../../../components/SettingsNotice';
+import Link from '../../../../components/Link';
+const { useSelect, useDispatch } = Data;
 
 export default function SettingsForm() {
+	const ga4Enabled = isFeatureEnabled( 'ga4setup' );
```
eugene-manuilov, line 58: Let's use `useFeature` hook instead:
```suggestion
	const ga4Enabled = useFeature( 'ga4setup' );
```

assets/js/modules/analytics/components/settings/SettingsForm.js
```diff
@@ -32,17 +37,42 @@ import {
 	ProfileNameTextField,
 	ExistingGTMPropertyNotice,
 } from '../common';
+import {
+	PropertySelect as PropertySelect4,
+} from '../../../analytics-4/components/common';
```
eugene-manuilov, line 42:
```suggestion
import PropertySelect4 from '../../../analytics-4/components/common/PropertySelect';
```

assets/js/modules/analytics/components/settings/SettingsEdit.test.js
```diff
@@ -32,6 +32,10 @@ describe( 'SettingsEdit', () => {
 			/tagmanager\/data\/settings/,
 			{ body: {}, status: 200 },
 		);
+		fetchMock.getOnce(
+			/^\/google-site-kit\/v1\/modules\/analytics-4\/data\/properties/,
+			{ body: [], status: 200 },
```
eugene-manuilov, line 37: The `status` is already `200` by default, so we can omit it:
```suggestion
			{ body: [] },
```

assets/js/modules/analytics/components/settings/SettingsForm.js
```diff
@@ -32,17 +37,42 @@ import {
...
+	const { selectProperty } = useDispatch( MODULES_ANALYTICS_4 );
+	useSelect( ( select ) => {
+		// We need to call getProperties for getSetupFlowMode to work.
+		const accountID = select( STORE_NAME ).getAccountID();
+		return select( MODULES_ANALYTICS_4 ).getProperties( accountID );
+	} );
+	const setupFlowMode = useSelect( ( select ) => select( STORE_NAME ).getSetupFlowMode() );
+	if ( ga4Enabled && SETUP_FLOW_MODE_UA === setupFlowMode ) {
+		selectProperty( PROPERTY_CREATE );
+	}
```
eugene-manuilov, line 69: It should be done in the `useEffect` hook.

assets/js/modules/analytics/components/settings/SettingsForm.js
```diff
@@ -32,17 +37,42 @@ import {
...
+const { useSelect, useDispatch } = Data;
 
 export default function SettingsForm() {
+	const ga4Enabled = isFeatureEnabled( 'ga4setup' );
 	const hasExistingTag = useSelect( ( select ) => select( STORE_NAME ).hasExistingTag() );
+	const { selectProperty } = useDispatch( MODULES_ANALYTICS_4 );
+	useSelect( ( select ) => {
+		// We need to call getProperties for getSetupFlowMode to work.
+		const accountID = select( STORE_NAME ).getAccountID();
+		return select( MODULES_ANALYTICS_4 ).getProperties( accountID );
```
eugene-manuilov, line 64: We should do it only if the `ga4Enabled` is true.

assets/js/modules/analytics/components/settings/SettingsForm.js
```diff
@@ -54,7 +84,39 @@ export default function SettingsForm() {
 
 				<ProfileSelect />
 			</div>
-
+			{
+				ga4Enabled && SETUP_FLOW_MODE_GA4_TRANSITIONAL === setupFlowMode && (
```
eugene-manuilov, line 88:
```suggestion
				( ga4Enabled && SETUP_FLOW_MODE_GA4_TRANSITIONAL === setupFlowMode ) && (
```

assets/js/modules/analytics/components/settings/SettingsForm.js
```diff
@@ -54,7 +84,39 @@ export default function SettingsForm() {
...
+					<SettingsNotice type={ TYPE_INFO }>
+						<div>
+							{ __( 'You\'ll need to connect the Google Analytics 4 property that\'s associated with this Universal Analytics property', 'google-site-kit' ) }
```
eugene-manuilov, line 91:
```suggestion
							{ __( 'You’ll need to connect the Google Analytics 4 property that’s associated with this Universal Analytics property', 'google-site-kit' ) }
```

assets/js/modules/analytics/components/settings/SettingsForm.js
```diff
@@ -54,7 +84,39 @@ export default function SettingsForm() {
...
+							external
+							inherit
+						>
+							{ __( 'Learn more here.', 'google-site-kit' ) }
+						</Link>
+					</SettingsNotice>
+				)
+			}
+			{
+				ga4Enabled && SETUP_FLOW_MODE_UA === setupFlowMode && (
```
eugene-manuilov, line 107:
```suggestion
				( ga4Enabled && SETUP_FLOW_MODE_UA === setupFlowMode ) && (
```

eugene-manuilov (dismissed):
Thanks, @ivankruchkoff. It's a good start, however, there are some things that we need to improve before we can approve this PR. Please, take a look at my comments.

eugene-manuilov:
@googlebot I consent.

assets/js/modules/analytics/components/settings/SettingsForm.js
```diff
@@ -31,39 +36,78 @@ import {
...
 export default function SettingsForm() {
+	const isGA4Enabled = useFeature( 'ga4setup' );
+	useSelect( ( select ) => {
+		if ( ! isGA4Enabled ) {
+			return;
+		}
+
+		// We need to call getProperties for getSetupFlowMode to work.
+		const accountID = select( STORE_NAME ).getAccountID();
+		return select( MODULES_ANALYTICS_4 ).getProperties( accountID );
```
tofumatt, line 65: This seems a bit of a hack—relying on the side effect of a selector to call another one.

It makes me think we should be calling these selectors as part of `getSetupFlowMode`'s resolver.

Why can't we do that here?

tofumatt (changes requested):
This looks fine apart from one thing: relying on a selector's side-effect before calling another one.

Having to manage that flow in a component seems the wrong approach, as it's entirely what resolvers are for. Is there any reason we _can't_ call:

```js
const accountID = select( STORE_NAME ).getAccountID();
select( MODULES_ANALYTICS_4 ).getProperties( accountID );
```

in the resolver for `getSetupFlow`? Or even as a `createRegistrySelector` in part of its selector?

assets/js/modules/analytics/components/settings/SettingsForm.js
ivankruchkoff: My initial idea was to change that resolver, however it wasn't in the IB, otherwise this hack has to happen everywhere we call the selector.

tofumatt: As in: if we don't change the resolver this needs to be added everywhere we call the `getSetupFlowMode`'s resolver?

That's a recipe for bugs in my mind; I see it isn't in the IB, but let's modify the resolver/selector to call these dependant selectors first.

----

Actually, I'm a bit confused; I went to look at the selector for `getSetupFlowMode` and it already calls the selectors you're looking for: https://github.com/google/site-kit-wp/blob/922999e8bb12bb40108244e992b61e5c9c07f88f/assets/js/modules/analytics/datastore/setup-flow.js#L56-L74

eugene-manuilov: > Actually, I'm a bit confused; I went to look at the selector for getSetupFlowMode and it already calls the selectors you're looking for:

The `getProperties` may return undefined because it has to be resolved first. That's why we call it beforehand.

tofumatt: Once it's resolved the selectors should all update though, no?

We account for that here: https://github.com/google/site-kit-wp/blob/b252720322e3dce7883c92c79811c6d83e19f20c/assets/js/modules/analytics/datastore/setup-flow.js#L76-L78

If `undefined` is returned by `getSetupFlowMode` we should render a loading screen/null/something else until things have loaded. The selector is designed to work without needing things to resolve first, I think here we just have to account for it returning `undefined` while things load.

tofumatt (changes requested):
I think we need to adjust `SettingsForm` to allow to `getSetupFlowMode()` to return `undefined` while loading. That's part of the selector's design and I don't think we should be relying on components to call selectors in order like that.

assets/js/modules/analytics/components/settings/SettingsForm.js
ivankruchkoff: `isAdminAPIWorking()` returns undefined without it
https://github.com/google/site-kit-wp/blob/922999e8bb12bb40108244e992b61e5c9c07f88f/assets/js/modules/analytics-4/datastore/api.js#L32

which gets us to here:
https://github.com/google/site-kit-wp/blob/b252720322e3dce7883c92c79811c6d83e19f20c/assets/js/modules/analytics/datastore/setup-flow.js#L49

eugene-manuilov: @tofumatt let's create a follow-up ticket for this if you think it's worth it and not block this PR because it is not relevant to this ticket.

tofumatt (approved):
In light of the issues with the `getSetupFlowMode` selector this looks good; I'll open a follow-up PR for that selector.

Thanks!

Issue #3358, opened 2021-05-13
Implement Idea Hub notice in WP dashboard widget


felixarntz:
If there are already any saved ideas in the Idea Hub module, the Site Kit WordPress dashboard widget should display a notice that links to the "Saved" tab of the Idea Hub widget, with the notice following the [Figma design](https://www.figma.com/file/YB79eKcGOwyUjlWG8mjDZA/Prompt-Engine-%2F-Idea-Hub?node-id=20332%3A6780).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* If the `ideaHubModule` feature flag is enabled and if the Idea Hub module is active and if there are any saved ideas stored, Site Kit's WP dashboard widget should display a [notice per this Figma design](https://www.figma.com/file/YB79eKcGOwyUjlWG8mjDZA/Prompt-Engine-%2F-Idea-Hub?node-id=20332%3A6780) which links to the Idea Hub widget (more specifically, to its "Saved" tab) in the Site Kit dashboard.

## Implementation Brief
* Create new file `/assets/js/components/wp-dashboard/WPDashboardIdeaHub.js` which exports the `WPDashboardIdeaHub` functional component.
* Check if the `ideaHubModule` feature flag is enabled via the `useFeature` hook.
* Check if the `idea-hub` module is active by querying the `core/modules` data store via the `isModuleActive` selector.
* Check if there are saved ideas by querying the `idea-hub` data store via the `getSavedIdeas` selector.
* If all of the above three conditions are met, then `WPDashboardIdeaHub` should render a notice as per the designs in Figma. If not, the component should return `null`.
* Create `assets/sass/components/idea-hub/_googlesitekit-idea-hub-wpdashboard-widget.scss` which will contain all the styles for the notification. BEM methodology should be used for the class names. For e.g `.googlesitekit-idea-hub__wpdashboard--title`.
* The Google logo should be rendered as inline SVG and the existing `assets/svg/logo-g.svg` file can be used.
* The text for the notice and the `View` texts should link to the Site Kit dashboard.
    * The URL for the dashboard can be obtained by querying the `core/site` data store via the `getAdminURL` selector.
    * To automatically scroll the user to the `Saved` tab of the Idea Hub widget, append the `saved-ideas` hash to the dashboard URL.
* Create `/assets/js/components/wp-dashboard/WPDashboardIdeaHub.stories.js` which will contain the story for the component.
[...]

PR #3423, opened 2021-05-27
Idea Hub notice in WP dashboard widget - updated design (3358).

johnPhillips:
## Summary
After a round of QA we decided to adjust the design for this on smaller screens and to un-link the copy text. 

Addresses issue #3358

## Checklist


aaemnnosttv (changes requested):
@johnPhillips Let's update this to match the current design for now, even though I think this will probably be revised – we can do that in a follow-up issue.

The only thing missing for now is the alignment of the `VIEW` link – let's center align this on the row so it's inline with the logo too. Should be as simple as adding `align-items` in the right place 👍

aaemnnosttv (approved):
Thanks @johnPhillips 👍

Issue #3227, opened 2021-04-26
Do not make Google API request if default scopes are missing


felixarntz:
Currently, if the user for some reason does not complete the OAuth flow after activating a module (e.g. by just returning via the browser's Back button), they (as expected) see an error that they need to grant additional scopes. However, currently Site Kit's behavior is largely unaffected by this, still making Google API requests as usual, which results in a bunch of error API requests that could have been avoided from the start.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* If a REST data point which relies on a "service" (i.e. a Google API) is used when the current user has not granted the basic scopes required by the module that the service/API is part of, the request should fail with a permissions error before actually reaching the individual data point logic, **so that no actual request is sent to the Google API** (since that would fail anyway).
* In other words, `Module::validate_data_request` should be enhanced so that, if the REST data point relies on an API service (see `service` argument of the data point) and the module's basic scopes are not yet granted, it throws an appropriate `Insufficient_Scopes_Exception`, similar to how it is already done for specific data points which require additional scopes.
    * The error/exception message used for this scenario should be `Site Kit can’t access the relevant data from %s because you haven’t granted all permissions requested during setup.`, where the placeholder should be the module name.

## Implementation Brief

* Refactor `Module::validate_data_request` to include the following additional logic:
	* Determine if the data point relies on a service (i.e. `$definitions[ $datapoint_key ]['service']` is set)
	* Determine if the module's scopes are granted, for example:
	```php
	$scopes =  $definitions[ $datapoint_key ]['scopes'];
	$oauth_client = $this->authentication->get_oauth_client();
	
	$scopes_granted = $oauth_client->has_sufficient_scopes( $scopes )
	```
	* If the datapoint requires a service and the scopes are not granted, throw an `Insufficient_Scopes_Exception`, with the message `Site Kit can’t access the relevant data from %s because you haven’t granted all permissions requested during setup.`, passing the module name as the placeholder.
* Update the relevant tests to provide coverage for this new logic

### Test Coverage

* New test coverage should be added for the logic above; existing tests should pass.

### Visual Regression Changes

* None anticipated.

## QA Brief

1. Set up Site Kit with just Search Console connected
2. Create a second admin user, and log in to Site Kit as that user in a different browser / incognito window. So now you should be logged in as two separate admin users.
3. Verify that both users see the same Search Console data correctly
4. Now connect Site Kit to Analytics with the first user account.
[...]

PR #3434, opened 2021-05-28
Do not make Google API request if default scopes are missing (3227).

johnPhillips:
## Summary

Addresses issue #3227

## Checklist

### DIFF
--- a/includes/Core/Modules/Module.php
+++ b/includes/Core/Modules/Module.php
@@ -518,15 +518,29 @@ abstract class Module {
 			return;
 		}
 
-		$datapoint = $definitions[ $datapoint_key ];
+		$datapoint        = $definitions[ $datapoint_key ];
+		$requires_service = isset( $datapoint['service'] );
+		$scopes           = $datapoint['scopes'];
+		$oauth_client     = $this->authentication->get_oauth_client();
+		$scopes_granted   = $oauth_client->has_sufficient_scopes( $scopes );
 
-		// If the datapoint requires specific scopes, ensure they are satisfied.
-		if ( ! $this->authentication->get_oauth_client()->has_sufficient_scopes( $datapoint['scopes'] ) ) {
-			$request_scopes_message = ! empty( $datapoint['request_scopes_message'] )
-				? $datapoint['request_scopes_message']
-				: __( 'You’ll need to grant Site Kit permission to do this.', 'google-site-kit' );
+		if ( ! $scopes_granted ) {
+			if ( $requires_service ) {
+				// If the datapoint relies on a service which requires scopes and
+				// these have not been granted, fail the request with a permissions
+				// error (see issue #3227).
 
-			throw new Insufficient_Scopes_Exception( $request_scopes_message, 0, null, $datapoint['scopes'] );
+				/* translators: %s: module name */
+				$message = sprintf( __( 'Site Kit can’t access the relevant data from %s because you haven’t granted all permissions requested during setup.', 'google-site-kit' ), $this->name );
+			} else {
+				// Otherwise, if the datapoint doesn't rely on a service but requires
+				// specific scopes, ensure they are satisfied.
+				$message = ! empty( $datapoint['request_scopes_message'] )
+					? $datapoint['request_scopes_message']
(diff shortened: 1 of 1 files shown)


includes/Core/Modules/Module.php
```diff
@@ -518,15 +518,29 @@ private function validate_data_request( Data_Request $data ) {
...
 
-		$datapoint = $definitions[ $datapoint_key ];
-
-		// If the datapoint requires specific scopes, ensure they are satisfied.
-		if ( ! $this->authentication->get_oauth_client()->has_sufficient_scopes( $datapoint['scopes'] ) ) {
-			$request_scopes_message = ! empty( $datapoint['request_scopes_message'] )
-				? $datapoint['request_scopes_message']
-				: __( 'You’ll need to grant Site Kit permission to do this.', 'google-site-kit' );
+		$datapoint        = $definitions[ $datapoint_key ];
+		$requires_service = isset( $datapoint['service'] );
```
aaemnnosttv, line 522: `isset` will return `true` even for an empty string, for which most `service` definitions are set to that don't rely on one.
```suggestion
		$requires_service = ! empty( $datapoint['service'] );
```

includes/Core/Modules/Module.php
```diff
@@ -518,15 +518,29 @@ private function validate_data_request( Data_Request $data ) {
...
+		$scopes           = $datapoint['scopes'];
+		$oauth_client     = $this->authentication->get_oauth_client();
+		$scopes_granted   = $oauth_client->has_sufficient_scopes( $scopes );
```
aaemnnosttv, line 525: Let's keep these two inlined as before since they're only used once.

includes/Core/Modules/Module.php
```diff
@@ -518,15 +518,29 @@ private function validate_data_request( Data_Request $data ) {
...
+			} else {
+				// Otherwise, if the datapoint doesn't rely on a service but requires
+				// specific scopes, ensure they are satisfied.
+				$message = ! empty( $datapoint['request_scopes_message'] )
+					? $datapoint['request_scopes_message']
+					: __( 'You’ll need to grant Site Kit permission to do this.', 'google-site-kit' );
+			}
 
-			throw new Insufficient_Scopes_Exception( $request_scopes_message, 0, null, $datapoint['scopes'] );
+			throw new Insufficient_Scopes_Exception( $message, 0, null, $datapoint['scopes'] );
```
aaemnnosttv, line 543:
```suggestion
			throw new Insufficient_Scopes_Exception( $message, 0, null, $scopes );
```

includes/Core/Modules/Module.php
```diff
@@ -518,15 +518,29 @@ private function validate_data_request( Data_Request $data ) {
...
-		$datapoint = $definitions[ $datapoint_key ];
-
-		// If the datapoint requires specific scopes, ensure they are satisfied.
-		if ( ! $this->authentication->get_oauth_client()->has_sufficient_scopes( $datapoint['scopes'] ) ) {
-			$request_scopes_message = ! empty( $datapoint['request_scopes_message'] )
-				? $datapoint['request_scopes_message']
-				: __( 'You’ll need to grant Site Kit permission to do this.', 'google-site-kit' );
+		$datapoint        = $definitions[ $datapoint_key ];
+		$requires_service = isset( $datapoint['service'] );
+		$scopes           = $datapoint['scopes'];
```
aaemnnosttv, line 523: `$datapoint[scopes]` are datapoint-specific scopes, which are the on-demand write scopes (e.g. to create a property/profile, etc).

This issue is about making sure that the module's "base scopes" (i.e. non-datapoint-specific read-only scopes) are fulfilled if the datapoint depends on a service. _These scopes_ are what is returned by `Module::get_scopes` (in this context `$this->get_scopes()`).

So the existing datapoint-specific scope check should be done first, then check the base scopes.

aaemnnosttv (changes requested):
Thanks @johnPhillips – this is off to a good start, but needs a little work yet due to a detail that was missed in IBR.

aaemnnosttv (commented):
This looks good to me, it just needs a tweak to the indentation which I will apply now.

aaemnnosttv (approved):
LGTM thanks!

Issue #1997, opened 2020-09-09
Launch Widget API-based dashboard without feature flag


felixarntz:
Potential considerations before we can remove the feature flag and get rid of the legacy components for the dashboard widgets:
* Monitor how many API requests are made by the Widget API widgets. Room for optimization? For example implementing batching in `googlesitekit-api`, combining almost similar Analytics requests into one and then splitting out response data, ...
* Check all refactored widgets to ensure they include relevant checks for their API response data (**defensive coding**):
    * return if e.g. `! data || ! data.length || ! data[ 0 ].rows || ! data[ 0 ].totals` etc.
* Calls to `getDataErrorComponent` and `getNoDataComponent` should either use non-legacy components introduced for that purpose, or at least the calls should be updated to also include all the extra parameters, so that it at least works correctly (e.g. currently the full `error` object is not passed in the new widgets).
* Investigate "Maximum call stack size exceeded" error occasionally happening on refactored All Traffic widget.
* Review legacy data API - does it include _any_ logic not currently replicated by `googlesitekit-api`? Like setting global error states etc.
* Use new function instead of over-complicated legacy `getDataTableFromData` function.
* Potentially allow widgets to specify minimum and maximum width as alternative to a single width (relevant e.g. in #2023).
* Potentially modify `quarter` widget width to still display two in one row for mobile view? See last screenshots in https://github.com/google/site-kit-wp/issues/1898#issuecomment-693178796 for context.
* Use datastore selectors for URLs instead of legacy `getSiteKitAdminURL` function.
* Modify sparkline component to render a flat zero graph if it doesn't receive any data (relevant e.g. for Analytics Goals widget, see https://github.com/google/site-kit-wp/pull/1938#discussion_r485274499).
* Include bug issues in the sprint that were fixed as part of this refactoring, solely for tracking purposes and changelog:
    * #1996 

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

**Note that this is blocked by #3406. The IB can be already written though and even a PR could be started already.**

1. The `widgets.dashboard` and `widgets.pageDashboard` feature flags should be removed:
    * Everything that was previously only used with those flags active should now be unconditionally active.
    * Everything that was previously not used with those flags active should now be entirely removed.
1. All usages of the `googlesitekit.DashboardModule` and `googlesitekit.DashboardDetailsModule` filters (including the components that are only used in conjunction with those filters) should be removed.
[...]

PR #3439, opened 2021-05-28
Enhancement/1997 launch widget api based dashboard without feature flag

danielgent:
## Summary

Addresses issue #1997

## Checklist


danielgent:
@googlebot  I consent

tofumatt:
@googlebot I consent.

assets/js/modules/adsense/index.legacy.js
```diff
@@ -1,63 +0,0 @@
-/**
```
eugene-manuilov, line 1: Looks like we can also delete the `parseAFCClientID` function in the `assets/js/modules/adsense/util/parsing.js` file because we don't use it anymore.

assets/js/modules/analytics/index.legacy.js
```diff
@@ -25,56 +25,12 @@ import { addFilter } from '@wordpress/hooks';
  * Internal dependencies
  */
 import { getModulesData } from '../../util';
-import { createAddToFilter } from '../../util/helpers';
```
eugene-manuilov, line 28: We can also remove the following:

* The `countriesByTimeZone` constant in the `assets/js/modules/analytics/util/countries-timezones.js` file
* The `translateAnalyticsError` function in the `assets/js/modules/analytics/util/index.js` file

eugene-manuilov (changes requested):
Thanks, @danielgent. Mostly looks good to me. Found a few issues that we need to address. Could you please take a look?

tests/e2e/config/bootstrap.js
```diff
@@ -243,6 +243,20 @@ function observeNavigationRequest( req ) {
 	}
 }
 
+function requestHandlerIdea( request ) {
```
eugene-manuilov, line 246: @tofumatt could you please add jsdoc to this function?

eugene-manuilov: Also, looks like this function doesn't do anything. Do we really need it or we can remove it?

eugene-manuilov (changes requested):
Thanks, @tofumatt. This looks good to me. Added the last question and then this is good to go.

tests/e2e/config/bootstrap.js
tofumatt: Good catch! I commented out what it was doing because it looked like debugging left in the PR by @danielgent… then it passed and I forgot to remove this last bit 😅 

Thanks! Removed 👍🏻

eugene-manuilov:
@tofumatt this PR looks good to me but for some reason, it has one failing check. Could you please take a look at it and fix it?

tofumatt:
One E2E test was a timeout, just needed a restart.

The other one I have no clue about, but maybe it's from changing to the `main` branch or something? It's failing on lint, but the error is incorrect and the actual linting passes.

Screenshot 2021-07-27 at 15 12 20

Changing that will cause linting fails so let's ignore that failure for now? It's just the size check anyhow…

package-lock.json
```diff
@@ -25355,12 +25355,6 @@
         }
       }
     },
-    "eslint-import-resolver-alias": {
-      "version": "1.1.2",
-      "resolved": "https://registry.npmjs.org/eslint-import-resolver-alias/-/eslint-import-resolver-alias-1.1.2.tgz",
-      "integrity": "sha512-WdviM1Eu834zsfjHtcGHtGfcu+F30Od3V7I9Fi57uhBEwPkjDcii7/yW8jAT+gOhn4P/vOxxNAXbFAKsrrc15w==",
-      "dev": true
-    },
```
eugene-manuilov, line 25363: @tofumatt can the eslint error be caused by this change? Could you please revert it since this doesn't seem to be a relevant change for this PR?

eugene-manuilov (changes requested):

package-lock.json
eugene-manuilov: > Changing that will cause linting fails so let's ignore that failure for now? It's just the size check anyhow…

I am afraid that if we just ignore it for now, everyone else will have this problem in their PRs which is not what we want. :smile:

tofumatt: Oh shoot, good eye. Thanks!

tofumatt: That wasn't it, actually 😞 

I'm a bit baffled, I'll see if I can fix that error without causing other ones...

tofumatt: I don't think there's any error here: https://github.com/google/site-kit-wp/blob/cbd3f165daf78be82940d9c6456c25ad92876de2/assets/js/components/wp-dashboard/WPDashboardWidgets.js#L78

That's the line it's complaining about. I'm unclear on what's wrong—any idea @aaemnnosttv?

eugene-manuilov (approved):

Issue #3252, opened 2021-04-29
Implement GA4 logic in server-side callback for creating a new Analytics account


felixarntz:
In the server-side callback where a new Analytics account has been created, if the `ga4setup` flag is active, a corresponding GA4 property should be created as well.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* In `Analytics::handle_provisioning_callback` a new action should be triggered right before redirecting the user `googlesitekit_analytics_handle_provisioning_callback`.
* Then in the `Analytics_4` class a callback should be added to that action which creates a corresponding new GA4 property and GA4 web data stream and stores their data in the respective `Analytics_4\Settings` settings.
    * Since the respective data points for creating a GA4 property and creating a GA4 web data stream already rely on the current website (e.g. for name and URL), it's not really necessary to have any additional context on which account/property/profile were created. It is just _a new_ one for the current site.
    * If any of these API requests fails, it should be ignored / not result in a user-facing error or blocking behavior.

## Implementation Brief

* Trigger the new `googlesitekit_analytics_handle_provisioning_callback` action right before the final redirect in the `Analytics::handle_provisioning_callback()` method.
  * The action should retrieve account ID, property ID, profile ID, and internal web property ID as action arguments.
* Extract the property creation script into a new private method `create_property( $accountID )`: https://github.com/google/site-kit-wp/blob/8da6a9176fd4c5dd3ef6100d3d1b9e75ecb91635/includes/Modules/Analytics_4.php#L229-L233
* Extract the web data stream creation script into a new private method `create_webdatastream( $propertyID )`: https://github.com/google/site-kit-wp/blob/8da6a9176fd4c5dd3ef6100d3d1b9e75ecb91635/includes/Modules/Analytics_4.php#L244-L248
* Add a new private method `handle_provisioning_callback( $accountID )` to the `Analytics_4` class.
  * In the method body, create a new ga4 property and web datastream using new helper methods.
  * Save new `propertyID` and `webDataStreamID` in the Analytics4 settings.
  * Suppress any error triggered during the creation of new property and web data stream because the ga4 API is still in the alpha phase. Add a to-do note to remove error suppression once ga4 API becomes stable.
* Add a new hook to the `Analytics_4::register` method that calls the new `handle_provisioning_callback` method for the `googlesitekit_analytics_handle_provisioning_callback` action.

### Test Coverage

* N/A

### Visual Regression Changes

* N/A

## QA Brief

[...]

PR #3443, opened 2021-05-31
Enhancement/3252 ga4 for new account

eugene-manuilov:
## Summary

Addresses issue #3252

## Checklist


includes/Modules/Analytics_4.php
```diff
@@ -200,6 +202,73 @@ protected function get_datapoint_definitions() {
...
+			$web_datastream = $this->create_webdatastream( $property->_id );
+			$web_datastream = self::filter_webdatastream_with_ids( $web_datastream );
+			if ( empty( $web_datastream->_id ) ) {
+				return;
+			}
+
+			$this->get_settings()->merge(
+				array(
+					'webDataStreamID' => $web_datastream->_id,
+					'measurementID'   => $web_datastream->measurementID, // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
```
aaemnnosttv, line 264: Is this the correct case? Our fixtures show this should use the `Id` form.
```suggestion
					'measurementID'   => $web_datastream->measurementId, // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
```

includes/Modules/Analytics_4.php
```diff
@@ -200,6 +202,73 @@ protected function get_datapoint_definitions() {
...
+	 * @param string $account_id Account ID.
+	 */
+	private function handle_provisioning_callback( $account_id ) {
+		// TODO: remove this try/catch once GA4 API stabilizes.
+		try {
+			$property = $this->create_property( $account_id );
+			$property = self::filter_property_with_ids( $property );
+			if ( empty( $property->_id ) ) {
+				return;
+			}
```
aaemnnosttv, line 251: Is this possible? Wouldn't the service/API throw if this were to fail?

includes/Modules/Analytics_4.php
```diff
@@ -68,6 +69,7 @@ final class Analytics_4 extends Module
 	public function register() {
 		$this->register_scopes_hook();
 
+		add_action( 'googlesitekit_analytics_handle_provisioning_callback', $this->get_method_proxy( 'handle_provisioning_callback' ) );
```
aaemnnosttv, line 72: Can we add a test for this? We should be able to use our `FakeHttpClient` to mock responses to the Google API. It would be good to assert that the relevant entities are created and settings updated when this fires.

includes/Modules/Analytics.php
```diff
@@ -282,15 +282,25 @@ protected function handle_provisioning_callback() {
...
 			)
 		);
 
+		do_action(
+			'googlesitekit_analytics_handle_provisioning_callback',
+			$account_id,
+			$web_property_id,
+			$internal_web_property_id,
+			$profile_id
+		);
```
aaemnnosttv, line 302: Let's expect this to be called in the `AnalyticsTest::test_handle_provisioning_callback` as well.

aaemnnosttv (changes requested):
Thanks @eugene-manuilov – this looks great! Just a few questions and requests for you around tests.

includes/Modules/Analytics_4.php
eugene-manuilov: :man_facepalming: yes, you are right. I'll update it.

includes/Modules/Analytics_4.php
eugene-manuilov: This is possible if GA4 APIs change the schema of returned entities.

includes/Modules/Analytics_4.php
eugene-manuilov: Added.

includes/Modules/Analytics.php
eugene-manuilov: Added.

tests/phpunit/integration/Modules/Analytics_4Test.php
```diff
@@ -20,45 +20,142 @@
...
+				'accountID'       => '12345678',
+				'propertyID'      => '',
+				'webDataStreamID' => '',
+				'measurementID'   => '',
+			)
+		);
+
+		$handle_provisioning_callback = ( new ReflectionClass( $analytics ) )->getMethod( 'handle_provisioning_callback' );
+		$handle_provisioning_callback->setAccessible( true );
+		$handle_provisioning_callback->invoke( $this->analytics, '12345678' );
```
aaemnnosttv, line 134: Let's call this via `do_action( 'googlesitekit_analytics_handle_provisioning_callback' ... )` instead. This will require us to register the `Analytics_4` instance here as well.

tests/phpunit/integration/Modules/Analytics_4Test.php
```diff
@@ -20,45 +20,142 @@
...
+					default:
+						return new Response( 200 );
+				}
+			}
+		);
+
+		$this->analytics->get_client()->setHttpClient( $http_client );
+
+		$options   = new Options( $this->context );
+		$analytics = new Analytics_4( $this->context, $options );
```
aaemnnosttv, line 120: I think the extra instance here isn't needed; we should be able to use `$this->analytics` instead, it isn't necessary to pass this specific `Options` instance to it.

aaemnnosttv (commented):
Thanks @eugene-manuilov – this is really close, just a few final details then this will be good to go!

tests/phpunit/integration/Modules/Analytics_4Test.php
eugene-manuilov: Updated.

tests/phpunit/integration/Modules/Analytics_4Test.php
eugene-manuilov: Updated.

aaemnnosttv (approved):
Great, thanks!

Issue #3380, opened 2021-05-19
Implement main `Survey` component


aaemnnosttv:
## Feature Description

The `Survey` component is the main component for rendering a single survey. This component is responsible for managing state for answers and rendering the appropriate inner components implemented in #3379.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A `Survey` component should be created in `assets/js/components/surveys/` as a functional, hook-based, hybrid-connected component
* It should (for now) retrieve its array of `question` objects (see `.question` in the [survey JSON](https://docs.google.com/document/d/1jqUg4blDUFu6BCyHQVgzlz9smxK0UGrzsOpPojo2EJM/edit?ts=609c08e3#heading=h.2ksz9wryvy2j)) and array of `completion` objects (see `.completion` in the [survey JSON](https://docs.google.com/document/d/1jqUg4blDUFu6BCyHQVgzlz9smxK0UGrzsOpPojo2EJM/edit?ts=609c08e3#heading=h.2ksz9wryvy2j)) from the `currentSurvey` in the data store (added in #3355)
	- The component can assume that this exists if it is being rendered
* The component should keep track of the answered state of each question  
_For the initial implementation there will only be a single question, but it should be capable of handling multiple_
* The component should send a `survey_shown` event when rendered for the first time (e.g. on mount)
	- See `sendSurveyEvent` action added in #3355 for all events that are sent (i.e. not `trackEvent`)
* The component should keep track of the current question (if any are left to be answered) and render the proper `SurveyQuestion*` component based on the question type (these components are implemented in other issues)
	- Note that the only question type that will be implemented at this point will be `rating` but the proper component-per-question-type should be considered here (for any other type `null` should be rendered for now)
	- An `answerQuestion` prop should be passed to the question which when called should set and submit the answer for the question by sending a `question_answered` event which should include event data with the [`QuestionAnswer` object answer which will vary in shape based on the question type](https://docs.google.com/document/d/1jqUg4blDUFu6BCyHQVgzlz9smxK0UGrzsOpPojo2EJM/edit?pli=1#heading=h.qxbkeiyvcd7o)
	- All `question_answered` events for all question types will have the following top-level properties:
		- `question_ordinal` – from the current question
		- `answer` [object] – value varies by question type
	- For the `question_answered` event for a **`rating`** type question, the full event data should have the following shape:
		```json
		{
			"question_ordinal": 1,
			"answer": {
				"answer": {
[...]

PR #3476, opened 2021-06-03
Add Survey component, tests, and stories.

tofumatt:
## Summary

Addresses issue #3380.

Stories for this PR: https://google.github.io/site-kit-wp/storybook/pull/3476/?path=/story/components-surveys-currentsurvey--survey-single-question-story

## Relevant technical choices

As [mentioned in the issue](https://github.com/google/site-kit-wp/issues/3380#issuecomment-853984466), this differs a bit from the IB (though matches the updated ACs) in that it creates a `CurrentSurvey` component that fetches survey data from the data store directly. This means the `Survey`/`CurrentSurvey` component doesn't take _any_ props and instead is entirely reliant on the datastore.

The previous `CurrentSurvey` component has been renamed `CurrentSurveyPortal` but otherwise behaves similarly. I removed the stories in Storybook for this component as they were redundant thanks to the new connected `CurrentSurvey` stories.

This component _does_ support multi-question surveys, as can be seen in Storybook and the unit tests. It also supports "finishing the survey early" by meeting a completion trigger before all questions have been answered (also available in Storybook and the unit tests).

I used Jest snapshots in a few places where we wanted to make sure either `SurveyCompletion` or `SurveyQuestion` was rendered… this seemed more accurate than only selecting their text nodes, but like all snapshots is a bit more sensitive to DOM changes in these components.

I think that covers everything from the ACs/IBs; let me know if I'm missing anything.

## Checklist


assets/js/components/surveys/CurrentSurvey.js
```diff
@@ -16,32 +16,173 @@
...
+		return {
+			...acc,
+			[ answer.question_ordinal ]: answer.answer_ordinal,
+		};
+	}, {} ) : [];
+
+	// Use Array.some to avoid looping through all completions; once a matching
+	// completion has been found, treat the survey as complete.
+	completions?.some( ( completion ) => {
+		completion.trigger_condition.some( ( condition ) => {
```
aaemnnosttv, line 117: It seems like there is a missing condition here in that all questions need to be answered before completion logic is checked. I think you mentioned "completing early" in the description but as far as I know, that isn't a requirement.

assets/js/components/surveys/CurrentSurvey.js
```diff
@@ -16,32 +16,173 @@
...
+	if ( ! triggeredCompletion && questions?.length && currentQuestionOrdinal > questions?.length ) {
+		triggeredCompletion = completions[ 0 ];
+	}
+
+	useEffect( () => {
+		if ( triggeredCompletion && ! hasSentCompletionEvent ) {
+			sendSurveyEvent( 'completion_shown', {
+				completion_ordinal: triggeredCompletion.completion_ordinal,
+			} );
+			setHasSentCompletionEvent( true );
```
aaemnnosttv, line 146: Let's move the set state before sending the actual event above. We're not awaiting it here, but still a good precaution to take.

assets/js/components/surveys/CurrentSurvey.js
```diff
@@ -16,32 +16,173 @@
...
+			<SurveyQuestionRating
+				answerQuestion={ answerQuestion }
+				choices={ currentQuestion.question.answer_choice }
+				dismissSurvey={ dismissSurvey }
+				key={ currentQuestion.question_ordinal }
+				question={ currentQuestion.question_text }
 			/>
-		</Portal>
-	);
+		);
```
aaemnnosttv, line 178: How about we extract a `[type]: Component` map here? I believe the props will always be the same to all question types so maybe the only condition here would be "does a component for this type exist" rather than checking the specific type.

assets/js/components/surveys/CurrentSurvey.test.js
```diff
@@ -0,0 +1,412 @@
...
+				credentials: 'include',
+				headers: {
+					Accept: 'application/json, */*;q=0.1',
+					'Content-Type': 'application/json',
+				},
+				method: 'POST',
+			}
+		);
+
+		fetchMock.reset();
```
aaemnnosttv, line 94: This is probably not what we want to do right here, as [`reset`](http://www.wheresrhys.co.uk/fetch-mock/#api-lifecyclerestore_reset) "Resets fetch() to its unstubbed state".

I think what we want instead is `resetHistory()`

```suggestion
		fetchMock.resetHistory();
```

assets/js/components/surveys/CurrentSurvey.test.js
```diff
@@ -0,0 +1,412 @@
...
+					'Content-Type': 'application/json',
+				},
+				method: 'POST',
+			}
+		);
+
+		expect( getByText( 'Another question: how do you feel when it rains?' ) ).toBeVisible();
+	} );
+
+	it( 'should trigger an early completion if a trigger condition is met, even if all questions have not been answered', async () => {
```
aaemnnosttv, line 174: I don't believe this is a requirement. Can you confirm @felixarntz ?

assets/js/components/surveys/CurrentSurvey.test.js
```diff
@@ -0,0 +1,412 @@
...
+		);
+
+		fetchMock.post( /^\/google-site-kit\/v1\/core\/user\/data\/survey-event/, { body: {}, status: 200 } );
+
+		const { container } = render( <CurrentSurvey />, { registry } );
+
+		expect( container ).toMatchSnapshot();
+	} );
+
+	it( 'should trigger a completion once all questions are answered, even if no matching trigger_conditions are found', () => {
```
aaemnnosttv, line 277: This could maybe be named more explicitly?
```suggestion
	it( 'should trigger the first completion when no matching trigger_conditions are met', () => {
```

assets/js/components/surveys/CurrentSurvey.test.js
```diff
@@ -0,0 +1,412 @@
...
+					Accept: 'application/json, */*;q=0.1',
+					'Content-Type': 'application/json',
+				},
+				method: 'POST',
+			}
+		);
+
+		fetchMock.reset();
+
+		// Render again to ensure we don't send another `survey_shown` event.
```
aaemnnosttv, line 346:
```suggestion
		// Render again to ensure we don't send another `completion_shown` event.
```

assets/js/components/surveys/__fixtures__/index.js
```diff
@@ -0,0 +1,261 @@
+export const invalidQuestionTypeSurvey = {
```
aaemnnosttv, line 1: Please add a file header 🙂

assets/js/googlesitekit/datastore/user/surveys.js
```diff
@@ -168,6 +170,40 @@ const baseSelectors = {
...
+	 * @since n.e.x.t
+	 *
+	 * @return {Array|null|undefined} Current survey's completion triggers if available; undefined if loading and `null` if no questions/survey are found.
+	 */
+	getCurrentSurveyCompletions: createRegistrySelector( ( select ) => () => {
+		const currentSurvey = select( CORE_USER ).getCurrentSurvey();
+
+		if ( currentSurvey === undefined ) {
+			return undefined;
+		}
```
aaemnnosttv, line 185: `getCurrentSurvey` is a little different than other selectors. It doesn't have a resolver attached to it and so the initial state is `null`, so the `undefined` condition would likely never be true here and can be removed.

assets/js/googlesitekit/datastore/user/surveys.js
```diff
@@ -168,6 +170,40 @@ const baseSelectors = {
...
+	 * @since n.e.x.t
+	 *
+	 * @return {Array|null|undefined} Current survey's questions if available; undefined if loading and `null` if no questions/survey are found.
+	 */
+	getCurrentSurveyQuestions: createRegistrySelector( ( select ) => () => {
+		const currentSurvey = select( CORE_USER ).getCurrentSurvey();
+
+		if ( currentSurvey === undefined ) {
+			return undefined;
+		}
```
aaemnnosttv, line 202: See above.

assets/js/components/surveys/__fixtures__/index.js
```diff
@@ -0,0 +1,261 @@
...
+					{
+						question_ordinal: 1,
+						answer_ordinal: [ 4, 5 ],
+					},
+				],
+			},
+			{
+				completion_ordinal: 2,
+				completion_title: 'You answered both questions!',
+				completion_text: 'What a completionist!.',
```
aaemnnosttv, line 63:
```suggestion
				completion_text: 'What a completionist!',
```

assets/js/components/surveys/SurveyHeader.js
```diff
@@ -42,6 +43,7 @@ const SurveyHeader = ( { title, dismissSurvey } ) => (
 			<h3>{ title }</h3>
 
 			<Button
+				aria-label={ __( 'Dismiss this survey', 'google-site-kit' ) }
```
aaemnnosttv, line 46: 👌

aaemnnosttv (changes requested):
Thanks @tofumatt ! This is a super solid start, nice work!

I left a few comments below – mostly suggestions I think but there are few corrections to make. The biggest change may be the early-completion part but I've asked @felixarntz to confirm.

The tests and storybook are very thorough as well. Thanks!

assets/js/components/surveys/CurrentSurvey.js
felixarntz: Yes that is correct. "Early completion" should not be possible. The completion data is always present in the API response, but it should only be checked / possibly displayed once all questions have been answered.

assets/js/components/surveys/CurrentSurvey.js
```diff
@@ -16,32 +16,173 @@
...
+	const questions = useSelect( ( select ) => select( CORE_USER ).getCurrentSurveyQuestions() );
+	const surveySession = useSelect( ( select ) => select( CORE_USER ).getCurrentSurveySession() );
+
+	const formName = surveySession ? `survey-${ surveySession.session_id }` : null;
+
+	const answers = useSelect( ( select ) => select( CORE_FORMS ).getValue( formName, 'answers' ) );
+
+	const currentQuestionOrdinal = answers?.length ? answers.reduce( ( a, b ) => {
+		return ( a.question_ordinal > b.question_ordinal ? a.question_ordinal : b.question_ordinal );
+	} ).question_ordinal + 1 : 1;
```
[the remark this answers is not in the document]
tofumatt: The answers array will _probably_ be in the order the questions were answered, but I didn't want to introduce the potential for bugs so was thorough with checking the highest ordinal answered.

`Math.max( 0, ...answers.map( ( a ) => a.question_ordinal ) ) + 1` works for me! 👍🏻

assets/js/components/surveys/CurrentSurvey.js
```diff
@@ -16,32 +16,173 @@
...
+
+	const answerQuestion = useCallback( ( answer ) => {
+		// eslint-disable-next-line camelcase
+		if ( currentQuestion?.question_type === 'rating' ) {
+			sendSurveyEvent( 'question_answered', {
+				// eslint-disable-next-line camelcase
+				question_ordinal: currentQuestion?.question_ordinal,
+				answer: {
+					answer_ordinal: answer.answer_ordinal,
+				},
```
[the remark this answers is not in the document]
tofumatt: Fair enough; it's only restricted for the rating question type, but thanks for catching the missing `answer` nesting; I totally missed that one 👍🏻

assets/js/components/surveys/CurrentSurvey.js
tofumatt: Ahhhh, okay, that wasn't clear to me because of the `question_ordinal` attribute in the `trigger_condition` object.

My assumption in reading that was that the completion was triggered _when_ that question was answered with a particular value. But I understand now it means that "if question 1 in a three-question survey is answered with X, show this completion".

Sounds good; I'll adjust the tests and the code accordingly 👍🏻

assets/js/components/surveys/CurrentSurvey.stories.js
```diff
@@ -20,36 +20,74 @@
...
-			return (
-				<WithRegistrySetup func={ setupRegistry }>
-					<Story />
-				</WithRegistrySetup>
-			);
-		},
+		( Story ) => (
+			<div className="googlesitekit-survey">
+				<Story />
+			</div>
```
aaemnnosttv, line 90: This wrapping div + class needs to be part of the output of the component now as this is actually missing when it renders in the app now. I think you originally had just question / completion components here which is why we moved this to the decorator. This needs to be added for both questions and completions.

aaemnnosttv (commented):
Thanks @tofumatt ! This is pretty close now although there are a few things left to address. I left a few comments, some minor but there are still a few important things to add here (such as the missing `googlesitekit-survey` class).

I also just realized that the feature flag has not been added to the JSON file yet (we we wouldn't be able to force-enable it for testing) so I'll push that up now as well.

aaemnnosttv:
@googlebot I consent.

assets/js/components/surveys/CurrentSurvey.js
```diff
@@ -1,5 +1,5 @@
 /**
- * CurrentSurvey component.
+ * Survey component.
```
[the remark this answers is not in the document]
tofumatt: Hah, shoot, yes, sorry, some renaming of files and I got confused 😓

assets/js/components/surveys/CurrentSurvey.js
```diff
@@ -64,27 +66,23 @@ const CurrentSurvey = () => {
...
-		// eslint-disable-next-line camelcase
-		if ( currentQuestion?.question_type === 'rating' ) {
-			sendSurveyEvent( 'question_answered', {
-				// eslint-disable-next-line camelcase
-				question_ordinal: currentQuestion?.question_ordinal,
-				answer: {
-					answer_ordinal: answer.answer_ordinal,
-				},
-			} );
+		// console.log( 'answer', answer );
```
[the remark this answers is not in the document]
tofumatt: lol, whoops 😅

assets/js/components/surveys/CurrentSurvey.js
```diff
@@ -16,32 +16,173 @@
...
-	const currentSurvey = useSelect( ( select ) => select( CORE_USER ).getCurrentSurvey() );
+	const [ hasSentSurveyShownEvent, setHasSentSurveyShownEvent ] = useState( false );
+	const [ hasSentCompletionEvent, setHasSentCompletionEvent ] = useState( false );
+
+	useEffect( () => {
+		if ( questions?.length && ! hasSentSurveyShownEvent ) {
+			setHasSentSurveyShownEvent( true );
+			sendSurveyEvent( 'survey_shown' );
+		}
+	}, [ questions, hasSentSurveyShownEvent, sendSurveyEvent ] );
```
felixarntz, line 47: `sendSurveyEvent` is `undefined` here I think, it's only assigned to a variable further below.

assets/js/components/surveys/CurrentSurvey.js
```diff
@@ -16,32 +16,173 @@
...
+					// eslint-disable-next-line camelcase
+					question_ordinal: currentQuestion?.question_ordinal,
+					answer: { answer },
+				},
+			],
+		} );
+	}, [ answers, currentQuestion, formName, sendSurveyEvent, setValues ] );
+
+	const ctaOnClick = useCallback( () => {
+		sendSurveyEvent( 'follow_up_link_clicked' );
```
felixarntz, line 88: Here the respective `completion_ordinal` should also be passed as event data, see https://docs.google.com/document/d/1jqUg4blDUFu6BCyHQVgzlz9smxK0UGrzsOpPojo2EJM/edit?pli=1#heading=h.k0jpie3yvlqo

felixarntz (commented):
@tofumatt Did just a quick pass over this to review the specific way the survey events are sent, there's one minor thing that needs to change; and one other code bug that you'd probably encounter yourself sooner than later :)

aaemnnosttv (commented):
Looks good to me, just a few small details to clean up which I'll push real quick, and then this will be good to go 👍

Issue #3418, opened 2021-05-26
dateRangeLength must be a positive integer dashboard error


jamesozzie:
## Bug Description

There have been recent reports of the below error message, with one user stating this occurs when clicking on the Search console tab. 

```
dateRangeLength must be a positive integer.

    in LegacySearchConsoleDashboardWidgetSiteStats
    in withData(LegacySearchConsoleDashboardWidgetSiteStats)
    in Unknown
    in WithSelect(withData(LegacySearchConsoleDashboardWidgetSiteStats))
    in div
    in Layout
    in div
    in div
    in div
    in div
    in GoogleSitekitSearchConsoleDashboardWidget
    in FilteredComponent
    in WithFilters(ModuleApp)
    in ModuleApp
    in GoogleSitekitModule
    in RestoreSnapshots
    in ErrorHandler
    in Root
```

- https://wordpress.org/support/topic/site-kit-encountered-an-error-36/ | Open | [SH info provided](https://docs.google.com/forms/d/1lhjns2K7c0Ny7ME4gxTcsnYtNUtcMr8BvFpPQi23Y1E/edit?urlBuilderDomain=google.com#response=ACYDBNgzDi1jqZ-t1Hi2v1KoXxSm_VRBVhA6Zd5eV3fowA0_2dvJKGLRAfRhBiwsooJ4GSE)
- https://wordpress.org/support/topic/error-search-google-in-site-kit/ | Open | [SH info provided](https://docs.google.com/forms/d/1lhjns2K7c0Ny7ME4gxTcsnYtNUtcMr8BvFpPQi23Y1E/edit?urlBuilderDomain=google.com#response=ACYDBNix_sqSD8RDFcGKxRgshXF32xXj5nCMEeZ5u5taH8DdEqv2pHb0Z3kjc2R8Fq3CAEc)
- https://wordpress.org/support/topic/my-site-kit-has-occured-an-error/ | Open
- https://wordpress.org/support/topic/gsuite-shows-a-error-what-does-it-means | **Resolved** | SH info within topic | randomly started working as expected once more
[...]

PR #3478, opened 2021-06-04
Fix date range partitioning

aaemnnosttv:
## Summary

Addresses issue #3418

This PR fixes a number of bugs – most importantly the React error which can happen in legacy Search Console components when using `extractSearchConsoleDashboardData` with an array with only 1 row. This is possible when a site only has 1 row of data as the Search Console API does not return empty rows like Analytics can.

## Relevant technical choices

- `partitionReport` is now always called with the `dateRangeLength` coming from the current date range, so it should never be 0 or another value that would fail `invariant` validation.
- To provide this, the existing `withSelect` in `withData` was updated to provide `dateRangeLength` as a prop to the wrapped component, in addition to the `dateRange` it was providing before.
- Only doing the above fixed the error, but resulted in incorrect charts in Search Console when the `report` was shorter than 2x the date range length because it was filling the `compareRange` before the `currentRange`
- `partitionReport` has been updated to slice from the end to the beginning (reverse from how it was before) starting with the `currentRange`. For sites with incomplete data (e.g. only 14 days/rows of data) it will now show the same information for 14, 28 and 90 day ranges whereas before only the 7 day range would show data (7 + 7) but 14 day and above would show no data since the only data would fall under the rows for the compare range.
- Weekdays should still align (as tests show), but the utility no longer tries to return ranges of equal length, so the code that was previously failing due to this had to be updated to provide fallback values when compare range data was incomplete/non-existent.
- **Currently targets main to include in 1.34.0 release but will need versions updated if we decide to do so**

## Checklist


assets/js/modules/search-console/util/index.js
```diff
@@ -110,14 +108,12 @@ export const isDataZeroSearchConsole = ( data ) => {
...
 
-	const processedData = extractSearchConsoleDashboardData( data );
-
 	const {
 		totalClicks,
 		totalImpressions,
 		averageCTR,
 		averagePosition,
-	} = processedData;
+	} = reduceSearchConsoleData( data );
```
aaemnnosttv, line 116: This might not be an obvious change:
It used to use `extractSearchConsoleDashboardData` here before but now uses the simpler `reduceSearchConsoleData` function. The important difference here is that the values here are now based on the entire report, whereas before these were based on `latestData` which was incorrect in some cases as well since we always sliced the compare range first and then latest.

https://github.com/google/site-kit-wp/blob/f6f02c5e3443611ac27a9c4b4be7dda7a03f040d/assets/js/modules/search-console/util/index.js#L89-L92

Zero data state should always be based on the entire report, not just the latest. Even though this utility is only for legacy components, it's still used so this should be an improvement/fix nonetheless.

assets/js/util/partition-report.js
```diff
@@ -22,9 +22,10 @@
 import invariant from 'invariant';
 
 /**
- * Splits report data into segments of equal length for current and comparison ranges.
+ * Splits report data into segments for current and comparison ranges.
  *
  * @since 1.33.0
+ * @since n.e.x.t Updated to support incomplete `report` arrays.
```
aaemnnosttv, line 28: In case we want to include this in the upcoming release:
```suggestion
 * @since 1.34.0 Updated to support incomplete `report` arrays.
```

assets/js/modules/search-console/util/site-stats-data.js
```diff
@@ -30,19 +30,22 @@ import { __, _x, sprintf } from '@wordpress/i18n';
  * Internal dependencies
  */
 import { getLocale, numFmt, calculateChange, getChartDifferenceArrow } from '../../../util';
+import { getPreviousDate } from '../../../util/date-range/get-previous-date';
 
 /**
  * Gets data for a Google chart from a Search Console report.
  *
  * @since 1.30.0
+ * @since n.e.x.t Added `dateRangeLength` parameter.
```
aaemnnosttv, line 39:
```suggestion
 * @since 1.34.0 Added `dateRangeLength` parameter.
```

felixarntz (approved):
Awesome work - let's 🚢  in 1.34.0!

PR #3479, opened 2021-06-04
Enhancement/3248 fix qa issues

eugene-manuilov:
## Summary

Addresses issue #3248

## Relevant technical choices

* Fixed margin between dropdowns and GA4 property notice
* Fixed missing scopes issue
* Fixed property creation issue

## Checklist

### DIFF
--- a/assets/js/modules/analytics/datastore/settings.js
+++ b/assets/js/modules/analytics/datastore/settings.js
@@ -44,6 +44,7 @@ import {
 import { STORE_NAME, PROPERTY_CREATE, PROFILE_CREATE, FORM_SETUP } from './constants';
 import { createStrictSelect } from '../../../googlesitekit/data/utils';
 import { isFeatureEnabled } from '../../../features';
+import { isPermissionScopeError } from '../../../util/errors';
 
 // Invariant error messages.
 export const INVARIANT_INVALID_ACCOUNT_ID = 'a valid accountID is required to submit changes';
@@ -66,8 +67,8 @@ export async function submitChanges( { select, dispatch } ) {
 		}
 
 		propertyID = property.id;
-		await dispatch( STORE_NAME ).setPropertyID( property.id );
-		await dispatch( STORE_NAME ).setInternalWebPropertyID( property.internalWebPropertyId ); // eslint-disable-line sitekit/acronym-case
+		dispatch( STORE_NAME ).setPropertyID( property.id );
+		dispatch( STORE_NAME ).setInternalWebPropertyID( property.internalWebPropertyId ); // eslint-disable-line sitekit/acronym-case
 	}
 
 	const profileID = select( STORE_NAME ).getProfileID();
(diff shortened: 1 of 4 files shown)


includes/Modules/Analytics_4.php
```diff
@@ -226,9 +226,15 @@ protected function create_data_request( Data_Request $data ) {
...
 
+				$timezone = get_option( 'timezone_string' );
+				if ( empty( $timezone ) ) {
+					$timezone = 'UTC';
+				}
+
 				$property = new Google_Service_GoogleAnalyticsAdmin_GoogleAnalyticsAdminV1alphaProperty();
 				$property->setParent( self::normalize_account_id( $data['accountID'] ) );
 				$property->setDisplayName( wp_parse_url( $this->context->get_reference_site_url(), PHP_URL_HOST ) );
+				$property->setTimeZone( $timezone );
```
felixarntz, line 237: Is setting a timezone here necessary? I don't think we have been doing that when creating UA properties. I see why it might be nice, but it also introduces complexity and might not be as straightforward as it is (since WordPress probably supports a different list of timezone strings than Google Analytics does).

Does not having a timezone here produce an error? If it doesn't and the field is optional, we should continue to not set it instead.

assets/js/modules/analytics/datastore/settings.js
```diff
@@ -99,7 +100,10 @@ export async function submitChanges( { select, dispatch } ) {
 
 	if ( isFeatureEnabled( 'ga4setup' ) ) {
 		if ( select( MODULES_ANALYTICS_4 ).haveSettingsChanged() ) {
-			await dispatch( MODULES_ANALYTICS_4 ).submitChanges();
+			const { error } = await dispatch( MODULES_ANALYTICS_4 ).submitChanges();
+			if ( isPermissionScopeError( error ) ) {
+				return { error };
+			}
```
felixarntz, line 106: Nice!

felixarntz (changes requested):
@eugene-manuilov Mostly lgtm, but the other problem I pointed out in https://github.com/google/site-kit-wp/issues/3248#issuecomment-854270843 is certainly relevant. Maybe it's an edge case for me (in which case we can open a follow-up issue), but it even happens in Storybook (see https://google.github.io/site-kit-wp/storybook/main/?path=/story/modules-analytics-setup-setupformua--ready). Can you please look into that? It may be the same reason why I'm encountering the problem in production.

Last but not least, this PR is against `develop` but it has to be against `main`, so you might need to cherry-pick or reopen a new PR.

includes/Modules/Analytics_4.php
eugene-manuilov: Yes, it is required now. The createProperty request fails without a timezone.

https://developers.google.com/analytics/devguides/config/admin/v1/rest/v1alpha/properties#Property
developers google com_analytics_devguides_config_admin_v1_rest_v1alpha_properties

eugene-manuilov:
@felixarntz unfortunately, I can't rebase this branch, so I have created a new PR: #3483. Going to close this PR.

Issue #3291, opened 2021-05-04
Improve default Analytics property selection when selecting an Analytics account


felixarntz:
This enhancement purely relates to the existing Analytics module and the `modules/analytics` store, but it is still part of the GA4 epic since it ensures certain prerequisites to bring the same functionality to GA4 are in place.

Currently, when selecting an Analytics account (in the module setup or settings), the first property will be initially selected. This experience should be improved to be as follows:
* If one of the properties within the account matches the current site's URL, that property should be initially selected.
* If none of the properties within the account matches the current site's URL, no property should be initially selected.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new selector `getSiteURLPermutations()` should be added to the `core/site` store. It should create return all 4 "permutations" of the current site URL, including/excluding `www.` and with `http://` or `https://`. No trailing slash should be included at the end of each permutation.
* A new action `matchPropertyByURL( properties, url )` should be implemented in the `modules/analytics` store, with a purpose similar to the action of the same name in the `modules/analytics-4` store (see #3168).
    * It should simply iterate over the given list of property objects until it finds a property that has a `websiteUrl` which matches the given `url` (if `url` is a string) or matches any of the given values in `url` (if `url` is an array of strings).
* The existing logic to potentially make a cascading (UA) Analytics property selection in the `selectAccount` action in the `modules/analytics` store should be enhanced to try to find a matching property rather than just selecting the first property:
    * It should wait for the `modules/analytics` selector `getProperties` and, once properties have been fetched, rely on the new `matchPropertyByURL` action in the same store (see above) passing the `getSiteURLPermutations` array of URLs to attempt to find a matching UA property for the current site URL. If one is found, the `selectProperty` action should be called with it so that it is set.
    * In other words, at the end of this action, either the matched property should be selected or no property should be selected at all.
* The logic in the `getProperties` resolver that potentially selects the first property if none is selected yet should be removed, since it is no longer relevant - this should have always happened from within `selectAccount` as it is the appropriate action to trigger any automated sub-selections.

## Implementation Brief

* Create a new selector in the core/site store called `getSiteURLPermutations`, which should:
	* select the current site URL using the `getReferenceSiteURL` selector
	* return an array of all four permutations of the URL:
		* with and without `www.` 
		* with `http://` or `https://`
	* No trailing slash should be included at the end of the returned URLs
	* Add unit tests for this new selector
[...]

PR #3481, opened 2021-06-04
Fix/3291 default property selection

eugene-manuilov:
## Summary

Addresses issue #3291

## Checklist

### DIFF
--- a/assets/js/modules/analytics/datastore/accounts.js
+++ b/assets/js/modules/analytics/datastore/accounts.js
@@ -163,12 +163,17 @@ const baseActions = {
 
 			if ( ! uaProperty ) {
 				uaProperty = {
-					id: PROPERTY_CREATE,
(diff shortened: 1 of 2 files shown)


felixarntz (changes requested):
@eugene-manuilov This fixes the fallback selection issue as expected - the PR needs to be based on `main` and against `main` though, since it's for a release issue :)

About my other problem, have you tried to recreate the exact same situation I've described in https://github.com/google/site-kit-wp/issues/3291#issuecomment-854266366?

* Use a Google account with at least two different Analytics accounts.
* Create a property for your site URL in two different Analytics accounts. Make sure that in both cases these properties are not the first (or only) properties in the account, just to rule out that they get selected because they are the first in the list.
* Now, you should see the property being selected correctly at first. But then, when you switch to the other account that also has a property for that site URL, it does not get selected.

eugene-manuilov:
Going to close this PR in favor of #3484

Issue #2942, opened 2021-03-10
Migrate `AdminBarApp` Storybook coverage to follow new v6 best practices


felixarntz:
With Storybook now in v6, we should migrate a few particular stories to follow the new best practices established in [this doc](https://docs.google.com/document/d/1Ny2Tlo1_FxZY6s6dC2-E1TWJYi1rhp0OMEtEfXxwXjY/edit). Migrating all existing stories is not a high priority, we can keep most old stories in their existing state; they should only be migrated on demand.

This issue is about migrating a specific view component (i.e. a component consisting of multiple sub-components) to follow the new Storybook v6 best practices.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* New `*.stories.js` files should be added for the following connected admin bar components, each following the [new Storybook best practices](https://docs.google.com/document/d/1Ny2Tlo1_FxZY6s6dC2-E1TWJYi1rhp0OMEtEfXxwXjY/edit) (please check `DashboardUniqueVisitorsWidget.stories.js` from #2940 for reference):
    * `AdminBarClicks` (including decorator using `WithRegistrySetup` for populating datastores)
    * `AdminBarImpressions` (including decorator using `WithRegistrySetup` for populating datastores)
    * `AdminBarSessions` (including decorator using `WithRegistrySetup` for populating datastores)
    * `AdminBarUniqueVisitors` (including decorator using `WithRegistrySetup` for populating datastores)
    * each of them should have a decorator using `WithRegistrySetup` for populating datastores), there should be two stories (one for having data available, the other one for zero data state), and the title used should be `Views/AdminBarApp/{ComponentName}`
* `AdminBarZeroData` should also be accompanied by its own `AdminBarZeroData.stories.js` file with a simple purely presentational single story, the title should also be `Views/AdminBarApp/{ComponentName}` here.
* `AdminBarWidgets` should be accompanied by its own `AdminBarWidgets.stories.js` file, which should rely on reusing decorators declared in its sub-components (see above) wherever possible. It should include the following stories:
    * Analytics active & connected, data available for every sub-component
    * Analytics active & connected, zero data for both Analytics sub-components
    * Analytics active & connected, zero data for all sub-components
    * Analytics inactive, data available for every (Search Console) sub-component
    * Analytics inactive, zero data for every (Search Console) sub-component
* The existing Storybook coverage for `AdminBarApp` (and any of the other admin bar components) from before should be removed.

[...]

PR #3491, opened 2021-06-07
Add AdminBar stories.

asvinb:
## Summary

Addresses issue #2942 

## Relevant technical choices
* Added `common.stories.js` which contains decorators and functions used by most of the AdminBar components components.

## Checklist


assets/js/components/adminbar/AdminBarClicks.stories.js
```diff
@@ -0,0 +1,47 @@
...
+ */
+import { withWidgetComponentProps } from '../../googlesitekit/widgets/util';
+import { setupSearchConsoleMockReports, widgetDecorators } from './common.stories';
+import AdminBarClicks from './AdminBarClicks';
+import WithRegistrySetup from '../../../../tests/js/WithRegistrySetup';
+
+const WidgetWithComponentProps = withWidgetComponentProps( 'widget-slug' )( AdminBarClicks );
+
+const Template = ( { setupRegistry, ...args } ) => (
+	<WithRegistrySetup func={ setupRegistry ? setupRegistry : () => {} }>
```
tofumatt, line 30: This is a bit more concise and I'd argue a bit easier to read—`setupRegistry` appearing twice in a row is a bit tougher to parse to my eyes 🙂 
```suggestion
	<WithRegistrySetup func={ setupRegistry || () => {} }>
```

assets/js/components/adminbar/AdminBarImpressions.stories.js
```diff
@@ -0,0 +1,47 @@
...
+ */
+import { withWidgetComponentProps } from '../../googlesitekit/widgets/util';
+import { setupSearchConsoleMockReports, widgetDecorators } from './common.stories';
+import AdminBarImpressions from './AdminBarImpressions';
+import WithRegistrySetup from '../../../../tests/js/WithRegistrySetup';
+
+const WidgetWithComponentProps = withWidgetComponentProps( 'widget-slug' )( AdminBarImpressions );
+
+const Template = ( { setupRegistry, ...args } ) => (
+	<WithRegistrySetup func={ setupRegistry ? setupRegistry : () => {} }>
```
tofumatt, line 30: Same as above/for all instances here.

Also, you could do:
```suggestion
const Template = ( { setupRegistry =  () => {}, ...args } ) => (
	<WithRegistrySetup func={ setupRegistry }>
```

tofumatt (changes requested):
<del>
The stories themselves look good here, but the naming of everything seems a bit redundant. The stories end up looking like this:
</del>
Screenshot 2021-06-09 at 18 21 28
<del>
Which I think is a bit too many instances of "AdminBar".
</del>
(See my second review below, I think I'm wrong about this 😆 )

The "Ready" and "DataUnavailable" are odd names because in the case of "Zero Data" the "Ready" state is the only one used which is a bit odd to parse—I get how that works but I don't think it's immediately intuitive.

Maybe "Loaded" and "Not loaded" as that's what they are, right?

So the overall approach is good but a few coding style tweaks and maybe some improvements to naming so there's less repetition. But if that request is too vague let me know and I can help with some suggestions 😅

tofumatt (changes requested):
Upon reflection, I guess the repetition is because those are all the component names. That's actually helpful then and shouldn't be changed; I'm just not used to seeing it. 🤦🏻 

But maybe we can still adjust the "Ready" and "DataUnavailable" to be more descriptive?

assets/js/components/adminbar/common.stories.js
```diff
@@ -0,0 +1,126 @@
...
+		},
+		{
+			slug: 'analytics',
+			active: true,
+			connected: true,
+		},
+	] );
+
+	// Call story-specific setup.
+	if ( args?.setupRegistry && typeof args?.setupRegistry === 'function' ) {
```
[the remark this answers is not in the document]
asvinb: 👍

assets/js/components/adminbar/AdminBarZeroData.stories.js
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Internal dependencies
+ */
+import AdminBarZeroData from './AdminBarZeroData';
+
+const Template = ( args ) => <AdminBarZeroData { ...args } />;
+
+export const Ready = Template.bind( {} );
+Ready.storyName = 'Ready';
```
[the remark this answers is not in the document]
asvinb: @tofumatt I've renamed it to "Data Unavailable" to be consistent with how we did with other stories (new v6 ones). Let me know if that sounds good.

assets/js/components/adminbar/AdminBarWidgets.stories.js
```diff
@@ -0,0 +1,201 @@
...
+		( Story, { args } ) => {
+			const setupRegistry = ( registry ) => {
+				// Set some site information.
+				provideSiteInfo( registry, {
+					currentEntityURL: 'https://www.sitekitbygoogle.com/blog/',
+					currentEntityTitle: 'Blog test post for Google Site Kit',
+				} );
+
+				// Call story-specific setup.
+				if ( args?.setupRegistry && typeof args?.setupRegistry === 'function' ) {
```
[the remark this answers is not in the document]
asvinb: 👍

assets/js/components/adminbar/AdminBarClicks.stories.js
asvinb: It was tough to parse in my eyes as well, 🤣  however when i tried it, I was getting a syntax error. I went with the other solution where you suggested we set the default parameter value.

assets/js/components/adminbar/AdminBarImpressions.stories.js
asvinb: 👍

asvinb:
@tofumatt I kept the "Ready" and "Data Unavailable" naming convention so that it's consistent with what we did for other stories.
image

Do you have any suggestions?

tofumatt (approved):
Looks good 👍🏻 

Thanks for the clarification on the story names, makes sense to keep 'em consistent.

Thanks for the tweaks as well, looks good 👍🏻

Issue #2910, opened 2021-03-02
Remove the option to include / exclude logged in users when the Analytics snippet is not placed via Site Kit


jamesozzie:
## Bug Description

Some Site Kit users may have their Analytics snippet placed manually or via another plugin before connecting the Analytics module within Site Kit. They may have included the logged in user opt-out snippet snippet via another plugin or manually. 

While even the presence of one opt-out snippet will respect the users choice to not include logged in traffic let's consider or start a discussion on removing this option for manually inserted Analytics snippets. This would be useful in particular if there's a possibility that some plugins may insert the opt out snippet below but return a `return false` value. 

```
<!-- Google Analytics user opt-out added via Site Kit by Google -->
 <script type="text/javascript">window["_gaUserPrefs"] = { ioo : function() { return true; } }</script>
```

When that when connecting the Analytics module initially there are no additional configuration options visible - so a user would not be aware of the opt-out snippet (See #2909)

Some considerations for any changes to this configuration:
- Should this configuration option be removed for when users insert Analytics via Tag Manager
- Should these options be visible when initially setting up the Analytics module if a snippet was manually placed

## Steps to reproduce

1. Insert a manually placed Site Kit snippet
2. Connect the Analytics module within Site Kit
3. No configuration options appear but the opt out snippet exists - placed via SK
4. A user must go into their SK Analytics settings to toggle off this snippet

## Screenshots

![ex](https://user-images.githubusercontent.com/41326532/109727132-b880b980-7bab-11eb-828b-46b186edc30e.gif)

![image](https://user-images.githubusercontent.com/41326532/109715158-36889480-7b9b-11eb-9f65-f5e6c35c0a99.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* If Site Kit does not place the Analytics snippet (neither via Analytics nor via Tag Manager), the option to exclude certain users from tracking should not be displayed.
* Similarly, in that case the respective snippet should not be rendered in the frontend.
    * While it is technically possible to render it even when the snippet comes from elsewhere, this is more confusing than helpful, plus it might conflict with e.g. the other plugin already having that feature on its own.

## Implementation Brief

* Edit `assets/js/modules/analytics/components/settings/SettingsForm.js`:
  * Grab the `useSnippet` setting of Analytics and Tag Manager modules using `getUseSnippet` selectors.
[...]

PR #3494, opened 2021-06-07
Remove the option to include / exclude logged in users when the Analytics snippet is not placed via Site Kit (2910).

johnPhillips:
## Summary

Addresses issue #2910

## Checklist


tofumatt:
@googlebot I consent.

includes/Modules/Analytics.php
```diff
@@ -119,6 +119,11 @@ public function register() {
 	 * @return bool
 	 */
 	protected function is_tracking_disabled() {
+		$settings = $this->get_settings()->get();
+		if ( ! apply_filters( 'googlesitekit_allow_tracking_disabled', $settings['useSnippet'] ) ) {
```
tofumatt, line 123: It's good to mention that a filter is documented elsewhere, so something like this should do the trick:

```suggestion
		// This filter is documented below, later in this function.
		if ( ! apply_filters( 'googlesitekit_allow_tracking_disabled', $settings['useSnippet'] ) ) {
```

tofumatt (changes requested):
This looks good, but if I'm understanding the issue the aim is that _a site with an existing tag placed by something other than Site Kit does not show the "Exclude from Analytics" option_. Correct?

If so, I think the code for this story should be updated: https://google.github.io/site-kit-wp/storybook/pull/3494/?path=/story/analytics-module-settings--edit-with-existing-tag-w-access

Right now it shows the existing tag option when the user has already placed a tag, which seems wrong 🙂

assets/js/modules/analytics/components/settings/SettingsForm.js
```diff
@@ -41,6 +41,7 @@ import {
 } from '../common';
 import GA4PropertySelect from '../../../analytics-4/components/common/PropertySelect';
 import StoreErrorNotices from '../../../../components/StoreErrorNotices';
+import GA4Notice from '../common/GA4Notice';
```
tofumatt, line 44:
```suggestion
```

includes/Modules/Analytics.php
johnPhillips: You mean direct the reader to `Tag_Manager:: filter_analytics_allow_tracking_disabled`?

tofumatt: Yup! 🙂  Sorry, I thought it was in this file, but yes, pointing to there is good.

tofumatt (commented):
I just had a call with @johnPhillips and it turns out I wasn't 100% clear on the issue from reading #2910's summary/ACs/IB. Turns out the implementation here is right and the Storybook stories are still correct/fixed, but I got confused about the conditions they were looking for.

After the merge conflict is resolved I'll take another look but I think my concerns about Storybook stories from the previous review were incorrect!

tofumatt (approved):

Issue #3510, opened 2021-06-08
Ensure User Feedback CTA links open in new tab/window


eclarke1:
## Feature Description

When clicking any of the CTA buttons on the User Feedback survey, please could we ensure the links open in a new tab.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Any links that appear in user feedback survey completion prompts (visually displayed as buttons, but technically links) should always open in a new window/tab.
* The buttons (powered by the `Button` component) _should not_ visually include the "external" icon, but they _should_ include screen reader-only text "(opens in a new tab)" - this part should be included via `aria-label` using the same logic that is currently used in the `Link` component.

## Implementation Brief
* Using `assets/js/components/Button.js`,
    * Add the `aria-label` attribute using the same logic as in the `Link` component with the following change to the `getAriaLabel` function:
        * Instead of checking for the non existent `external` prop in the `Button` component, check if `target` is equal to `_blank`
* Using `assets/js/components/surveys/SurveyCompletion.js`,
    * Add the `target` prop to the `Button` component, with `_blank` as value.

### Test Coverage
* No new tests to be added.

### Visual Regression Changes
* N/A

## QA Brief

* Install and activate plugins
  * Site Kit tester plugin
  * [Site Kit Survey Tester plugin](https://app.asana.com/app/asana/-/get_asset?asset_id=1200430197238350)
* In the tester plugin settings
  * Set Force enabled feature flags to Always override
  * Enable the checkbox for the userFeedback feature flag 
* Reset Site Kit
* Go through the Site Kit setup flow
* Go to the dashboard and the survey should be displayed
* Answer the single question that is there, giving a 4 or 5 rating
* Check that the `Let's go` button link opens in a new tab

After being triggered, the survey will not be triggered again for 1 hour. To reset this delay, clear the browser cache and reload the dashboard page.

* The survey should be displayed on the dashboard again
* Answer the single question that is there, giving a 1, 2, or 3 rating
* * Check that the `Get help` button link opens in a new tab

## Changelog entry

* Update button links to show "opens in new tab" ARIA help text, and open all User Feedback links in a new tab.

wpdarren:
### QA Update: Pass ✅ 

- Verified that when you give a rating of 4 or 5, the `Let's go` button link opens in a new tab.
- Verified that when you give a rating of 1, 2, or 3, the `Get help` button link opens in a new tab.

PR #3553, opened 2021-06-10
Update to open user feedback links in a new tab.

Hazlitte:
## Summary

Update to open user feedback links in a new tab.

Addresses issue #3510

## Checklist

### DIFF
--- a/assets/js/components/Button.js
+++ b/assets/js/components/Button.js
@@ -27,6 +27,7 @@ import useMergedRef from '@react-hook/merged-ref';
  * WordPress dependencies
  */
 import { forwardRef, useCallback } from '@wordpress/element';
+import { _x } from '@wordpress/i18n';
 
 /**
  * Internal dependencies
@@ -43,6 +44,7 @@ const Button = forwardRef( ( {
 	target,
 	icon,
 	trailingIcon,
+	ariaLabel,
 	...extraProps
 }, ref ) => {
 	const buttonRef = useCallback( ( el ) => {
@@ -54,6 +56,22 @@ const Button = forwardRef( ( {
 
 	// Use a button if disabled, even if a href is provided to ensure expected behavior.
 	const SemanticButton = ( href && ! disabled ) ? 'a' : 'button';
+
+	const getAriaLabel = () => {
+		let label = ariaLabel;
+
+		if ( target === '_blank' ) {
+			return label;
+		} else if ( typeof children === 'string' ) {
+			label = label || children;
+		}
+
+		const newTabText = _x( '(opens in a new tab)', 'screen reader text', 'google-site-kit' );
+		if ( label ) {
+			return `${ label } ${ newTabText }`;
+		}
+		return newTabText;
+	};
 
 	return (
 		<SemanticButton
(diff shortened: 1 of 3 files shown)


assets/js/components/Button.js
```diff
@@ -55,6 +57,22 @@ const Button = forwardRef( ( {
...
 	const SemanticButton = ( href && ! disabled ) ? 'a' : 'button';
 
+	const getAriaLabel = () => {
+		let label = ariaLabel;
+
+		if ( target === '_blank' ) {
+			return label;
+		} else if ( typeof children === 'string' ) {
+			label = label || children;
+		}
```
tofumatt, line 67: I can see how the IB wasn't 100% clear on this, but the idea was that if `target === '_blank` then the "opens in a new tab" text would be appended.

In this case, the "opens in a new tab" string will never be used because if `target === '_blank'` it returns the `ariaLabel`, which may or may not be defined.

So this should actually be:
```suggestion
		if ( target !== '_blank' ) {
			return label;
		}
		
		if ( typeof children === 'string' ) {
			label = label || children;
		}
```

(We can remove the `else` here really, as well.)

tofumatt (changes requested):
Unfortunately as-is this won't output the "(opens in a new tab)" text that the issue intends to add.

You can see that in the updated snapshots that have the target changed to `_blank` but don't include the `aria-label`. Once the issue is implemented as-suggested in my PR comment, the aria label should appear in the snapshot test.

assets/js/components/Button.js
Hazlitte: Thanks @tofumatt 👍,
I've had a rethink on the logic here following your clarification.  I'll be interested to know if you think the changes I have made approach this better.

tofumatt: The problem with that approach is it never uses the `children` when they're a `string` as part of the label. You can see that below in the snapshot, where the `aria-label` is only `"(opens in a new tab)"`, which explains _what happens when the link is opened_ but not _what the actual link is_.

So admittedly if we have a non-string `children` value in these components they need a custom aria label to work well (meaning our approach is a bit naive), but as most of the times buttons and links _do_ have `children` or `label` props that are strings it works well.

In the case of the test below whose snapshot you updated, what you want is `aria-label="Let’s go (opens in a new tab)"`. `aria-label` is used to essentially _replace_ what a screenreader would normally read out as the contents of the link.

Imagine if you had a close button like: `<a href="#close">X</a>`. A screenreader would read that link as "X". Not very accessible/usable, but doing `<a href="#close" aria-label="Close window">X</a>` would have a screenreader read that link as "Close window" and NOT read out the "X".

So that's what we want here; your approach won't quite work for that, it'd need to be something like:

```js
	const getAriaLabel = () => {
		let label = ariaLabel;

		if ( target !== '_blank' ) {
			return label;
		}

		const newTabText = _x( '(opens in a new tab)', 'screen reader text', 'google-site-kit' );

		if ( typeof children === 'string' ) {
			label = label || children;
		}

		if ( label ) {
			return `${ label } ${ newTabText }`;
		}

		return newTabText;
	};
```

tofumatt (changes requested):
Unfortunately the proposed simplification means the label will often omit the content, so this needs adjustment.

assets/js/components/Button.js
Hazlitte: Many thanks @tofumatt 👍

Issue #2585, opened 2021-01-06
Introduce factory for AdSense response mock data


felixarntz:
After #2235, the `module-adsense-components.stories.js` file is now full of inlined fixtures, which is due to the mock data being dependent on the `referenceDate` used for the story. While we could have continued to use fixtures, doing so is in another way confusing as it decouples the two pieces entirely and obfuscates that they are connected.

We should introduce an AdSense-specific test utility, which generates AdSense response data.

We should also think about how this would affect request args definition in Storybook and tests - whether we should use fixtures there, factories, or just inline them (since those are _a lot_ simpler than response data).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a function like `getAdSenseMockResponse( requestArgs )` which generates a full AdSense response for the given request `requestArgs`.
    * This can basically replicate logic which we already have in PHP via the Site Kit tester plugin. It'll be based on some hard-coded mock values, combined with logic to arrange them accordingly to get the correct response format for the request. It should also follow any conventions introduced in #2584.
* The AdSense stories should use the above function instead of inlining response mock data.
* All AdSense fixtures for report data or args should be deleted from the plugin.

## Implementation Brief

* Add `getAdSenseMockResponse( requestArgs )` function in a new file `modules/adsense/util/data-mock.js`. This function is responsible for generating mock responses and should use the similar approach from #2584. It should also use `rxjs` to generate and process response data.
  * Generate an args hash and use it as a new seed value for the faker as we do it in the data mock function in the Analytics module (see [here](https://github.com/google/site-kit-wp/blob/develop/assets/js/modules/analytics/util/data-mock.js#L176-L187)). But for the AdSense module, let's generate a hash using JSON stringified version of the `requestArgs` variable.
  * Define a new object `data` that replicates the response data object.
  * Create a new rxjs stream with dates starting from `startDate` and ending with `endDate` if the `requestArgs` contains the `DATE` dimension (see an example [here](https://github.com/google/site-kit-wp/blob/develop/assets/js/modules/analytics/util/data-mock.js#L215-L226)). If the `DATE` dimension is not selected, then create a new stream with a single `null` value like it's done [here](https://github.com/google/site-kit-wp/blob/develop/assets/js/modules/analytics/util/data-mock.js#L245-L246).
  * Pipe following rxjs operators to the newly created stream using [`.pipe()`](https://rxjs-dev.firebaseapp.com/guide/operators) method:
[...]

PR #3548, opened 2021-06-10
Update Google API client services library with Idea Hub API and AdSense API v2

felixarntz:
## Summary

Addresses issue #3517

## Relevant technical choices

* AdSense API request behavior hasn't changed other than now using a single `parent` argument (like in the Google Analytics 4 Admin API), with the only additionally changed API endpoint being the one to generate a report (see https://developers.google.com/adsense/management/reference/rest/v2/accounts.reports/generate).
* The format of all returned objects from the API has been changed, which this PR accounts for. See documentation about new format for reference:
    * Account: https://developers.google.com/adsense/management/reference/rest/v2/accounts#Account
    * Client: https://developers.google.com/adsense/management/reference/rest/v2/accounts.adclients/list#AdClient
    * Alert: https://developers.google.com/adsense/management/reference/rest/v2/accounts.alerts/list#Alert
    * URL channel: https://developers.google.com/adsense/management/reference/rest/v2/accounts.adclients.urlchannels/list#UrlChannel
    * Ad unit: https://developers.google.com/adsense/management/reference/rest/v2/accounts.adclients.adunits#AdUnit
* The `GET:adunits` endpoint previously returned the full API response, which is against our convention, where we just return the array of objects within. This PR changes it for consistency, since it touches that part of the code anyway.
* Available alert types have changed to use lower-hyphen-case now, e.g. `GRAYLISTED_PUBLISHER` is now `graylisted-publisher`.

## Checklist

- [ ] My code is tested and passes existing unit tests.
- [ ] My code has an appropriate set of unit tests which all pass.
- [ ] My code is backward-compatible with WordPress 4.7 and PHP 5.6.
- [ ] My code follows the [WordPress](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) coding standards.
- [ ] My code has proper inline documentation.
- [ ] I have added a QA Brief on the issue linked above.
- [ ] I have signed the Contributor License Agreement (see <https://cla.developers.google.com/>).

Issue #3517, opened 2021-06-08
Update Google API PHP client services library to version with Idea Hub and AdSense V2 API


felixarntz:
Now that the [auto-generated PHP client code for the Idea Hub API](https://github.com/googleapis/google-api-php-client-services/blob/master/src/Google/Service/Ideahub.php) is available, we need to upgrade the Google API services library to include it.

The main reason for the update is the above Idea Hub API - however, related to that, the new PHP client services library also includes the new AdSense API client code, which switches from the v1.4 API to the v2 API - **with the v1.4 API getting deprecated in October this year**. The changes are minimal, from an API perspective, only the API root differs (see https://developers.google.com/adsense/management/migrating). However, the client code appears to be generated on a slightly different foundation, so some of Site Kit's usage of it needs to be adjusted. The changes are purely naming- and format-related though, no logic changes.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `google/apiclient-services` PHP library in `composer.json` should be updated to `^0.180` (latest version, which includes Idea Hub), and it should be updated accordingly (in `composer.lock` as well).
* The list of Google services to include the PHP third-party code for in the PHP scoper config `scoper.inc.php` should be modified to include `Ideahub`.
* The Google service classes for AdSense from the library now use `Adsense` instead of `AdSense` in their name, so this needs to be adjusted in `scoper.inc.php` as well.
* All usage of the AdSense PHP service classes also needs to be reviewed to follow the new names that the library code uses now.

## Implementation Brief

* Update `composer.json` as outlined in the ACs, then run `composer update google/apiclient-services` to actually update it.
* Modify `scoper.inc.php` to add `Ideahub` and replace `AdSense` with `Adsense`.
* Update code in `AdSense` module class to work with the new PHP library client code:
    * Update all library class name references from `Google_Service_AdSense` to `Google_Service_Adsense`.
    * Some resources have been renamed to more precisely follow their hierarchy, e.g. `accounts_urlchannels` is now `accounts_adclients_urlchannels` and `accounts_adunits` is now `accounts_adclients_adunits`. This applies both for the resource property name as well as for the resource class name.
[...]

PR #3555, opened 2021-06-10
Enhancement/2585 adsense mock data

eugene-manuilov:
## Summary

Addresses issue #2585

## Checklist


assets/js/modules/adsense/util/data-mock.js
```diff
@@ -0,0 +1,221 @@
...
+
+	// We set seed for every data mock to make sure that the same arguments get the same report data.
+	// It means that everyone will have the same report data and will see the same widgets in the storybook.
+	// This approach gives us additional flexibility to control randomness on a per widget basis.
+	if ( ! Number.isNaN( argsHash ) ) {
+		faker.seed( argsHash );
+	}
+
+	const metrics = ( Array.isArray( args.metrics ) ? args.metrics : [ args.metrics ] ).filter( ( metric ) => !! metric );
+	const dimensions = ( Array.isArray( args.dimensions ) ? args.dimensions : [ args.dimensions ] ).filter( ( dimension ) => !! dimension );
```
aaemnnosttv, line 176: I would prefer to use `castArray` over `Array.isArray( value ) ? value : [ value ]` which we use in a few places already. I've seen this in a few places and I think it would be a bit easier to read then a longer line with inline ternary.

assets/js/modules/adsense/util/data-mock.js
```diff
@@ -0,0 +1,221 @@
...
+
+	const data = {
+		kind: 'adsense#report',
```
aaemnnosttv, line 179: I believe this is changing with v2 (not much I don't think) so we should probably hold on merging this until #3548 is merged since that one is more important.

assets/js/modules/adsense/util/data-mock.js
```diff
@@ -0,0 +1,221 @@
...
+
+	const startDate = new Date( args.startDate );
+	const endDate = new Date( args.endDate );
+	const dayInMilliseconds = 24 * 60 * 60 * 1000;
+	const totalDays = 1 + ( ( endDate - startDate ) / dayInMilliseconds ); // +1 to include the endDate into the dates range.
+
+	// This is the list of operations that we will apply to the range (array) of numbers.
+	const ops = [
+		// Converts range number to a date string.
+		map( ( item ) => ( new Date( startDate.getTime() + ( dayInMilliseconds * item ) ) ).toISOString().split( 'T' )[ 0 ] ),
```
aaemnnosttv, line 197: It looks like this is creating a new date incremented by `item` days from the start. How about something like this instead?

```js
new Date( startDate ).setDate( startDate.getDate() + item );
```

Also for formatting, if this is our date string, we should use our `getDateString` utility instead.

assets/js/modules/adsense/util/data-mock.js
```diff
@@ -0,0 +1,221 @@
...
+			// really need mathematically correct values and can simplify the process of finding this information.
+			data.totals = [ ...( rows[ rows.length - 1 ] || [] ) ];
+			data.averages = [ ...( rows[ 0 ] || [] ) ];
+		} );
+
+	// Set the original seed value for the faker.
+	faker.seed( originalSeedValue );
+
+	return data;
+}
```
aaemnnosttv, line 221: Can we also add a `provideAdSenseMockReport` function here like we do for Analytics? That way we only need a single function to generate and receive the report which is what we'll need in most cases.

assets/js/modules/adsense/util/data-mock.js
```diff
@@ -0,0 +1,221 @@
...
+	IMPRESSIONS: METRIC_TALLY,
+	PAGE_VIEWS_CTR: METRIC_RATIO,
+};
+
+/**
+ * Generates and returns report headers.
+ *
+ * @since n.e.x.t
+ *
+ * @param {Array.<string>} metrics    Metircs list.
```
aaemnnosttv, line 43:
```suggestion
 * @param {Array.<string>} metrics    Metrics list.
```

assets/js/modules/adsense/util/data-mock.js
```diff
@@ -0,0 +1,221 @@
...
+	const endDate = new Date( args.endDate );
+	const dayInMilliseconds = 24 * 60 * 60 * 1000;
+	const totalDays = 1 + ( ( endDate - startDate ) / dayInMilliseconds ); // +1 to include the endDate into the dates range.
+
+	// This is the list of operations that we will apply to the range (array) of numbers.
+	const ops = [
+		// Converts range number to a date string.
+		map( ( item ) => ( new Date( startDate.getTime() + ( dayInMilliseconds * item ) ) ).toISOString().split( 'T' )[ 0 ] ),
+		// Add dimension and metric values.
+		map( ( date ) => [ ...generateDimensionValues( date, dimensions ), ...generateMetricValues( date, metrics ) ] ),
```
aaemnnosttv, line 199:
```suggestion
		map( ( date ) => [
			...generateDimensionValues( date, dimensions ),
			...generateMetricValues( date, metrics ),
		] ),
```

aaemnnosttv (changes requested):
Thanks @eugene-manuilov – this looks great overall. I left just a few comments mostly with suggestions, but I think we should probably wait to merge this until #3517 is merged to not delay or complicate the work in progress there.

Nice work!

assets/js/modules/adsense/util/data-mock.js
eugene-manuilov: Nice, updated.

assets/js/modules/adsense/util/data-mock.js
eugene-manuilov: Updated. :+1:

assets/js/modules/adsense/util/data-mock.js
eugene-manuilov: Added.

assets/js/modules/adsense/util/data-mock.js
eugene-manuilov: Updated.

assets/js/modules/adsense/util/data-mock.js
eugene-manuilov: Updated.

assets/js/modules/adsense/util/data-mock.js
```diff
@@ -219,3 +245,16 @@ export function getAdSenseMockResponse( args ) {
...
+ * Generates mock response for AdSense reports.
+ *
+ * @since n.e.x.t
+ *
+ * @param {wp.data.registry} registry Registry with all available stores registered.
+ * @param {Object}           options  Report options.
+ */
+export function provideAdSenseMockReport( registry, options ) {
+	registry.dispatch( STORE_NAME ).receiveGetReport( getAdSenseMockResponse( options ), { options } );
+	registry.dispatch( STORE_NAME ).finishResolution( 'getReport', [ options ] );
```
aaemnnosttv, line 259: The Analytics provider (and other `provide*` utilities) don't mark resolvers as finished, so we should probably leave this out for consistency. It shouldn't have any effect to not do this since the report is only fetched if no report exists yet for the given options.
```suggestion
```

aaemnnosttv (changes requested):
Looks good to me overall. I left one comment about `provideAdSenseMockReport` but nothing big. The main thing blocking this is the merge conflict which is a result of #3548 which is merged now. It looks like the only change with the new API for reports is the `kind` property has been removed. See https://github.com/google/site-kit-wp/pull/3548/files#diff-f99516d88faa1f82031cfc7876d0c93afeab899dcbda2572fc8ac9d24ccb3230

assets/js/modules/adsense/util/data-mock.js
eugene-manuilov: Updated.

package-lock.json
```diff
@@ -42956,7 +42956,7 @@
       "dependencies": {
         "strip-ansi": {
           "version": "3.0.1",
-          "resolved": "https://registry.npmjs.org/strip-ansi/-/strip-ansi-3.0.1.tgz",
+          "resolved": "http://registry.npmjs.org/strip-ansi/-/strip-ansi-3.0.1.tgz",
```
aaemnnosttv, line 42959: Can you see why this changed? It's minor but this is something I've noticed goes back and fourth in lock file updates. Dependabot PRs change this back to `https` as well. I'll fix it manually here to not block the PR but it's a bit odd.
```suggestion
          "resolved": "https://registry.npmjs.org/strip-ansi/-/strip-ansi-3.0.1.tgz",
```

package-lock.json
```diff
@@ -36582,7 +36582,7 @@
         },
         "strip-ansi": {
           "version": "3.0.1",
-          "resolved": "https://registry.npmjs.org/strip-ansi/-/strip-ansi-3.0.1.tgz",
+          "resolved": "http://registry.npmjs.org/strip-ansi/-/strip-ansi-3.0.1.tgz",
```
aaemnnosttv, line 36585:
```suggestion
          "resolved": "https://registry.npmjs.org/strip-ansi/-/strip-ansi-3.0.1.tgz",
```

aaemnnosttv (approved):
Awesome!

Issue #3529, opened 2021-06-09
Add shadow to help differentiate survey pop-up from the rest of the dashboard


eclarke1:
## Feature Description

Add drop shadow to the User Feedback survey to differentiate it from the rest of the dashboard

![image](https://user-images.githubusercontent.com/56588136/121335363-6bcb1e00-c912-11eb-8ded-d1b7f57befe1.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Instead of following the widget conventions for shadow, the user feedback survey box should use a more prominent shadow, particularly that shadow should match the one that is used for surveys in Google's own services: `box-shadow: rgb(0 0 0 / 14%) 0px 16px 24px 2px, rgb(0 0 0 / 12%) 0px 6px 30px 5px, rgb(0 0 0 / 20%) 0px 8px 10px -5px`
* The user feedback survey box should furthermore use a `border-radius: 8px` (also like surveys in Google's own services)

## Implementation Brief
* Using `assets/sass/components/surveys/_googlesitekit-surveys.scss`, update the `box-shadow` and add the `border-radius` CSS properties as per the AC.

### Test Coverage
* No tests to be added.

### Visual Regression Changes
* N/A

## QA Brief
**Set up**
* See the resource in Asana: https://app.asana.com/0/1200379030254833/1200379030254852
* The survey plug in is here: https://app.asana.com/app/asana/-/get_asset?asset_id=1200430197238350

**QA**
* Once you have the survey rendered, verify that it has the correct `box-shadow` and `border-radius` as per the AC.

## Changelog entry

* Update shadow used by user surveys to enhance contrast and improve visibility.

tofumatt:
IB ✅

wpdarren:
### QA Update: Pass ✅ 

- The shadow matches the AC: `box-shadow: rgb(0 0 0 / 14%) 0px 16px 24px 2px, rgb(0 0 0 / 12%) 0px 6px 30px 5px, rgb(0 0 0 / 20%) 0px 8px 10px -5px`
- The user feedback survey box should furthermore use a border-radius: 8px.
- Checked feedback survey box on all primary browsers, i.e. Safari, Firefox etc. 
- Checked feedback survey box on Android and iOS devices for smaller screens.

![image](https://user-images.githubusercontent.com/73545194/121900536-6d835000-cd1d-11eb-9e65-3bd10ec81ff5.png)

PR #3563, opened 2021-06-11
Add shadow to help differentiate survey pop-up from the rest of the dashboard (3529).

johnPhillips:
## Summary

Addresses issue #3529

## Checklist


aaemnnosttv:
@googlebot I consent.

aaemnnosttv (approved):
Looks good to me @johnPhillips (everything is as noted in the ACs) although given the extremely specific styles here, let's add a comment explaining that these are explicitly matching Google styles. In the future, maybe we'll extract this to a variable so such a comment wouldn't be needed but let's keep it here for now.

I've gone ahead and added this so we're all set, thanks!

Issue #3538, opened 2021-06-09
Can't submit the setup form when an account has UA and GA4 properties with different website URL


eclarke1:
## Bug Description

Can't submit the setup form when an account has UA and GA4 properties with different website URL. Looks like it happens because there is no matching web data stream for the current site

![cantsave](https://user-images.githubusercontent.com/873049/121718168-aab3cc00-caea-11eb-9952-a851e553bca8.png)

[Related Asana task](https://app.asana.com/0/1200316599642381/1200447371742681)

## Additional Context

 - PHP Version: 
 - Plugin Version [e.g. 22]

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* In the Analytics setup flow, if no matching web data stream for the current site already exists in the selected GA4 property, it should still be possible to submit the form.
    * In that case, the new web data stream should be created, then set upon submission.
    * This is possibly related to a similar bug in the Analytics settings screen (see #3535).

## Implementation Brief

* Merge #3565

### Test Coverage

* N/A

### Visual Regression Changes

* N/A

## QA Brief

* Create a new jurassic site
* Setup the SK plugin
* Activate the Analytics module
* On the setup form, select an account that has both GA4 and UA properties, but neither of them matches the current website URL.
* Select UA and GA4 properties and make sure that you can save the form.
* Save the form and verify that a new web data stream is created for the current website.

## Changelog entry

* Fix a problem where Analytics settings cannot be saved when selected account has UA and GA4 properties with different URLs.

aaemnnosttv:
IB ✅

wpdarren:
### QA Update: Pass ✅ 

- Verified that a new web data stream is created for the current website.

![image](https://user-images.githubusercontent.com/73545194/122087164-54050580-cdfc-11eb-8107-7fdf0c7ca35e.png)

PR #3565, opened 2021-06-11
Bug/3538 cant submit setup form

eugene-manuilov:
## Summary

Addresses issue #3538 

## Checklist

### DIFF
--- a/assets/js/modules/analytics-4/datastore/properties.js
+++ b/assets/js/modules/analytics-4/datastore/properties.js
@@ -27,7 +27,7 @@ import invariant from 'invariant';
 import API from 'googlesitekit-api';
 import Data from 'googlesitekit-data';
 import { CORE_SITE } from '../../../googlesitekit/datastore/site/constants';
-import { STORE_NAME, PROPERTY_CREATE, MAX_WEBDATASTREAMS_PER_BATCH } from './constants';
+import { STORE_NAME, PROPERTY_CREATE, MAX_WEBDATASTREAMS_PER_BATCH, WEBDATASTREAM_CREATE } from './constants';
 import { normalizeURL } from '../../../util';
 import { createFetchStore } from '../../../googlesitekit/data/create-fetch-store';
 import { isValidPropertySelection } from '../utils/validation';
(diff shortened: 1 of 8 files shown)


assets/js/modules/analytics-4/datastore/properties.js
```diff
@@ -168,6 +168,8 @@ const baseActions = {
 			if ( webdatastream ) {
 				registry.dispatch( STORE_NAME ).setWebDataStreamID( webdatastream._id );
 				registry.dispatch( STORE_NAME ).setMeasurementID( webdatastream.measurementId ); // eslint-disable-line sitekit/acronym-case
+			} else {
+				registry.dispatch( STORE_NAME ).setWebDataStreamID( WEBDATASTREAM_CREATE );
```
aaemnnosttv, line 172: Should we maybe also do this above if `PROPERTY_CREATE === propertyID` as well? Currently it just returns early.

assets/js/modules/analytics-4/datastore/settings.js
```diff
@@ -52,7 +52,7 @@ export async function submitChanges( { select, dispatch } ) {
 	}
 
 	const webDataStreamID = select( STORE_NAME ).getWebDataStreamID();
-	if ( ! isValidWebDataStreamID( webDataStreamID ) ) {
+	if ( webDataStreamID === WEBDATASTREAM_CREATE || ! isValidWebDataStreamID( webDataStreamID ) ) {
```
aaemnnosttv, line 55: Shouldn't this only happen if `webDataStreamID === WEBDATASTREAM_CREATE` like it does above for properties? It seems like this condition is accounting for this not being set somewhere.

aaemnnosttv (commented):
Thanks @eugene-manuilov – Looks good to me overall, just a few questions.

assets/js/modules/analytics-4/datastore/properties.js
eugene-manuilov: Updated to set `webDataStreamID` setting to `WEBDATASTREAM_CREATE` by default. If we find a matching web data stream, we will update the `webDataStreamID` setting to the correct value.

assets/js/modules/analytics-4/datastore/settings.js
eugene-manuilov: Updated the `submitChanges` function to use the `WEBDATASTREAM_CREATE` constant instead of an empty string, but left this condition as is.

aaemnnosttv (approved):
LGTM

Issue #3539, opened 2021-06-09
A new GA4 property has not been created for a new account created via settings page


eclarke1:
## Bug Description

A new GA4 property has not been created for a new account created via settings page. Now I see wrong GA4 property on the view settings page.

[Related Asana task](https://app.asana.com/0/1200316599642381/1200447371742693)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* When creating a new Analytics account via the setup flow, a new GA4 property for that account and the current site should also be created (in addition to the regular account creation process which already includes creating the UA property).
    * This was already implemented in #3252, but there seems to be a problem with the implementation.

## Implementation Brief

* Merge #3571

### Test Coverage

* N/A

### Visual Regression Changes

* N/A

## QA Brief

* Create a new account via setup or settings form.
* Verify that it also creates GA4 property with the appropriate web data stream.

## Changelog entry

* Update Analytics account creation to ensure a GA4 property is always created when GA4 is enabled.

eugene-manuilov:
@felixarntz @aaemnnosttv after digging into this issue it turned out that this issue caused by insufficient permissions scopes error that is thrown when we call the `create` method: https://github.com/google/site-kit-wp/blob/af36f9384d3a56fedb3daddc74bbdbbbda1ad0e6/includes/Modules/Analytics_4.php#L224

Here is the error body:
```
{
  "error": {
    "code": 403,
    "message": "Request had insufficient authentication scopes.",
    "errors": [
      {
        "message": "Insufficient Permission",
        "domain": "global",
        "reason": "insufficientPermissions"
      }
    ],
    "status": "PERMISSION_DENIED"
  }
}
```

Do you know how we can overcome it? :thinking:

felixarntz:
@eugene-manuilov Oh I assume that is because creating an account only requires the `https://www.googleapis.com/auth/analytics.provision` scope, but creating the property requires the `https://www.googleapis.com/auth/analytics.edit` scope.

I suggest we therefore include _both_ scopes as new requirement for the `POST:create-account-ticket` data point. While strictly the second scope is only needed to create the property, from a user-facing perspective, this is now a single unified process - and technically the creation of an account includes already the creation of a UA property anyway (which would already require `https://www.googleapis.com/auth/analytics.edit` if it had been done individually).

The above fix should be fairly straightforward and would resolve this problem. Probably we should only require the extra scope for now if the `ga4Setup` feature flag is enabled.

[...]

PR #3571, opened 2021-06-14
Bug/3539 new ga4 property for new account

eugene-manuilov:
## Summary

Addresses issue #3539

## Checklist


assets/js/modules/analytics/components/common/AccountCreate/index.js
```diff
@@ -58,16 +60,17 @@ export default function AccountCreate() {
 	const canSubmitAccountCreate = useSelect( ( select ) => select( STORE_NAME ).canSubmitAccountCreate() );
 	const isDoingCreateAccount = useSelect( ( select ) => select( STORE_NAME ).isDoingCreateAccount() );
 	const hasProvisioningScope = useSelect( ( select ) => select( CORE_USER ).hasScope( PROVISIONING_SCOPE ) );
+	const hasEditScope = useSelect( ( select ) => isGA4enabled ? select( CORE_USER ).hasScope( EDIT_SCOPE ) : undefined );
```
aaemnnosttv, line 63: No need to select this conditionally since we're already using the same selector above for the provisioning scope so the resolver will be invoked for `getAuthentication()` anyways. Better to keep it simple and consistent.

Also, this would also cause the `! hasEditScope` to always be `true` below if GA4 was not enabled which could cause problems later if `isGA4enabled` wasn't checked before using it.

```suggestion
	const hasEditScope = useSelect( ( select ) => select( CORE_USER ).hasScope( EDIT_SCOPE ) );
```

aaemnnosttv (commented):
Looks good to me, just one thing to change which has a suggestion that I'll apply now (see below).

aaemnnosttv (approved):
LGTM 👍

Issue #2998, opened 2021-03-22
Ensure a Web container is available for selection during Tag Manager module setup if using Standard mode AMP - when not all URLs are AMP


jamesozzie:
## Bug Description

As raised by [one user in the support forums](https://wordpress.org/support/topic/site-kit-amp-standard-mode-gtm/) some users have AMP active in standard mode. For some URLs they may have AMP disabled, as configured in the AMP plugins "Supported Templates" configuration.  If they wish to enable the Tag Manager module within Site Kit this results in no Web Container dropdown available - resulting in no GTM container for non AMP URLs.

Rather than just checking the active AMP mode the plugin should check the AMP Plugins "Supported Templates" configuration. There is a "Serve all templates as AMP?" flag as evident in a users Site Health information with the AMP plugin active, this could be used to allow the disable of a web container.

[AMP plugin GitHub PR](https://github.com/ampproject/amp-wp/pull/1235) with "Support Templates" details

## Steps to reproduce

1. Setup AMP in standard mode
2. Disable AMP for any template modes or content types (ie. homepage)
3. Setup Tag Manager
4. Web Container doesn't appear

## Screenshots

![image](https://user-images.githubusercontent.com/41326532/111974324-3410bf00-8af7-11eb-9970-8f5b795a4298.png)

![image](https://user-images.githubusercontent.com/41326532/111974451-530f5100-8af7-11eb-9ea3-6e927e948353.png)
 

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The Site Kit logic to detect the AMP mode relevant for the site should be improved as follows:
    * If the AMP plugin option "Serve all templates as AMP" is not enabled, the AMP mode returned should be "secondary" instead of "primary". That is because, even though the site is mainly in standard mode ("primary"), there may still be non-AMP content present.
    * If the Web Stories plugin is active, the AMP mode returned should be "secondary" instead of `false`. In other words, if e.g. the AMP plugin is not active at all but the Web Stories plugin is, we still need to assume the site to be "secondary" AMP (since Web Stories uses AMP, while in that case the rest of the site wouldn't).

## Implementation Brief

Update the function `get_amp_mode()` to work as follows:

https://github.com/google/site-kit-wp/blob/d6ea05554e20353df9e770abd193c98b006cdf79/includes/Context.php#L321

- Always `self::AMP_MODE_SECONDARY;` in place of `self::AMP_MODE_PRIMARY;`
[...]

PR #3577, opened 2021-06-15
Update to always return AMP mode secondary.

Hazlitte:
## Summary

Update to always return AMP mode secondary.

Addresses issue #2998

## Checklist

### DIFF
--- a/includes/Context.php
+++ b/includes/Context.php
@@ -28,7 +28,11 @@ class Context {
 	/**
 	 * Primary "standard" AMP website mode.
 	 *
-	 * @since 1.0.0
+	 * This mode is currently unused due to Tag Manager setup not showing the Web Container dropdown
+	 * when AMP is in standard mode and some urls have AMP disabled.
+	 *
+	 * @since 1.0.0 Originally introduced.
+	 * @since n.e.x.t Marked as unused, see description.
 	 * @var string
 	 */
 	const AMP_MODE_PRIMARY = 'primary';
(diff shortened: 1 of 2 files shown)


Hazlitte:
@googlebot I consent

Hazlitte:
@tofumatt As suggested I tested with the latest release of the AMP plugin and version 1.5.5.

tofumatt:
Looks like this is a new PR—we originally had the PR discussion in #3504, just curious why this new one was opened? It's usually best to keep changes in one PR so we can keep track of the discussion 🙂

includes/Context.php
```diff
@@ -348,27 +352,7 @@ public function get_amp_mode() {
...
-			}
-		} elseif ( function_exists( 'amp_is_canonical' ) ) {
-			// On older versions, if it is not primary AMP, it is definitely secondary AMP (transitional or reader mode).
-			if ( amp_is_canonical() ) {
-				return self::AMP_MODE_PRIMARY;
-			}
-
+		if ( $exposes_support_mode || function_exists( 'amp_is_canonical' ) ) {
 			return self::AMP_MODE_SECONDARY;
 		}
```
tofumatt, line 357: Let's keep the original check in here, but add `STANDARD_MODE_SLUG` in, eg:

```php
if ( $exposes_support_mode ) {
    // If recent version, we can properly detect the mode.
    if ( $amp_plugin_version_2_or_higher ) {
        $mode = AMP_Options_Manager::get_option( 'theme_support' );
    } else {
        $mode = AMP_Theme_Support::get_support_mode();
    }

    if (
        in_array( $mode, array(
            AMP_Theme_Support::STANDARD_MODE_SLUG,
            AMP_Theme_Support::TRANSITIONAL_MODE_SLUG,
            AMP_Theme_Support::READER_MODE_SLUG
        ), true )
    ) {
        return self::AMP_MODE_SECONDARY;
    }
} elseif ( function_exists( 'amp_is_canonical' ) ) {
    return self::AMP_MODE_SECONDARY;
}
```

I know it seems a lot more verbose, but if new modes are added to the AMP plugin we should need to explicitly support them I'd think, which is what our old code is doing. Otherwise `if ( in_array( $mode, array( AMP_Theme_Support::TRANSITIONAL_MODE_SLUG, AMP_Theme_Support::READER_MODE_SLUG ), true ) ) {` would be skipped and we'd always `return self::AMP_MODE_SECONDARY`, but we don't.

So let's keep those checks in.

tofumatt (changes requested):
Let's keep the existing check in place, but other than that looks good.

tofumatt (approved):
Looks good, but as I made some changes passing over to @aaemnnosttv for another look before merging into the release.

tofumatt:
@googlebot I consent.

includes/Context.php
Hazlitte: Hi @tofumatt,
A couple of days ago Evan told us "develop has been cut into main so please ensure any remaining PRs for 1.35.0 (or follow-ups from QA) are created from and updated to target main.".  I took this to mean that I needed to create a new branch from main for this ticket, transfer the changes to it and create a new PR.  I guess this was the wrong approach.  What should I have done?

aaemnnosttv: @Hazlitte that comment was only regarding issues for the release (1.35.0). #2998 was not tagged with the release, so no change was needed here.

Hazlitte: @aaemnnosttv Ah ok, thanks for clarifying 👍

Issue #3509, opened 2021-06-08
Add slide up / down animation to User Feedback survey


eclarke1:
## Feature Description

The User Feedback survey is missing the slide up / down animation to show / hide the survey.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The user input survey box should not just appear, it should have an animation when it appears and disappears:
    * When it appears, it should slide up into view from the bottom of the screen.
    * When it disappears (e.g. completion or dismissal), it should slide down out of view back "into" the bottom of the screen.
* Potentially this can technically use https://material-ui.com/components/transitions/ - this is not a requirement though, just a note.

## Implementation Brief
* Install the `@material-ui/core` package.
* Using `assets/js/components/surveys/CurrentSurvey.js`,
    * Create a new state variable, `animateSurvey` which is a `boolean`, with `false` as default value.
    * Use the `useMount` hook to set the `animateSurvey` to `true`.
    * Create new function `handleAnimationOnExited` which calls the `setValues` action (which can be found in the `dismissSurvey` function) to hide the form. Basically the form is hidden and unmounted when the animation is complete.
    * Wrap the markup for the survey questions (not when it has been completed) within the `Slide` component, imported from `@material-ui/core` with the following props:
        * `direction="up"`
        * `in={ animateSurvey }`
        * `onExited={ handleAnimationOnExited }`
    * Update the `dismissSurvey` function to :
        * set `animateSurvey` to `false`.
        * Remove the `setValues` call since we moved it to the `handleAnimationOnExited` function.

### Test Coverage
* No new tests to be added.

### Visual Regression Changes
* N/A

## QA Brief
* Run storybook locally via the `npm run storybook` command.
* Visit the "Surveys -> Current Survey" story.
* The user feedback survey should slide up upon page load and slide down when closed.

## Changelog entry

* Enhance user survey display with added animation on enter and exit.

tofumatt:
> Make calls to the `sendSurveyEvent` and `setValues` after 1 second, so that the animation has enough time to end.

I'm confused by this, why link the events and the animation? Does this mean we delay sending events until the animation is done to prevent issues with the component being unmounted after the events are completed?

I don't want to overcomplicate the issue, but that seems a bit of a brittle approach to dealing with animations for components.

[...]

PR #3578, opened 2021-06-15
Add Slide animation to user feedback survey.

asvinb:
## Summary

Addresses issue #3509 

## Checklist

### DIFF
--- a/assets/js/components/surveys/CurrentSurvey.js
+++ b/assets/js/components/surveys/CurrentSurvey.js
@@ -17,6 +17,12 @@
  */
 
 /**
+ * External dependencies
+ */
+import { useMount } from 'react-use';
+import { Slide } from '@material-ui/core';
+
+/**
  * WordPress dependencies
  */
 import { useCallback, useEffect, useState } from '@wordpress/element';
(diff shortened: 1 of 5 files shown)


assets/js/components/surveys/CurrentSurvey.js
```diff
@@ -151,23 +162,29 @@ const CurrentSurvey = () => {
 		}
 	}, [ hasSentCompletionEvent, sendSurveyEvent, triggeredCompletion ] );
 
+	useMount( () => {
+		setAnimateSurvey( true );
+	} );
```
eugene-manuilov, line 167: @asvinb why not use `true` as the default value for the `const [ animateSurvey, setAnimateSurvey ] = useState( false );`?

eugene-manuilov (changes requested):
Thanks, @asvinb. It looks good to me in general. Just have one question. Please, take a look.

assets/js/components/surveys/CurrentSurvey.js
asvinb: @eugene-manuilov When testing I noticed that the animation does not fire up when in the WordPress dashboard. The popup just shows up without any animation. So, we make sure it's only when the component is mounted, then it fires the animation.

eugene-manuilov (approved):

Issue #3543, opened 2021-06-09
Account with GA4 property only is not preselected on the setup form


eclarke1:
## Feature Description

Account with GA4 property only is not preselected on the setup form

[Related Asana ticket](https://app.asana.com/0/1200316599642381/1200447371742671)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* On pageload of the Analytics setup flow (with `ga4Setup` flag enabled): When you have an account that has only GA4 properties _one of which has a web data stream for your current site URL_ and no account has a UA property that matches your current site URL, the matching GA4 property (and its account of course) should be preselected.
    * This effectively lands you in the `ga4` version of the setup flow, since that account has only GA4 properties.
    * If you have any UA property that matches the current site URL among your accounts, that should still take precedence. This issue is particularly for the case that you have only a matching GA4 property somewhere, but no matching UA property.
* Essentially, once preselected you should land in this UI: https://google.github.io/site-kit-wp/storybook/develop/?path=/story/modules-analytics-setup-setupformga4--without-existing-tag

## Implementation Brief

* Merge #3590

### Test Coverage

* N/A

### Visual Regression Changes

* N/A

## QA Brief

* Create a new jurassic site.
* Create a new Google Analytics account that has only GA4 property with the new website URL.
* Install and activate the SK plugin.
* Set up the plugin and activate the Analytics module.
* On the Analytics setup form, verify that you see your new Google Analytics account with the appropriate property. preselected.

## Changelog entry

* Ensure account and property are preselected during setup when user has a matching GA4 property only.

aaemnnosttv:
IB ✅

wpdarren:
### QA Update: Pass ✅ 

- Verified that for a GA4 property only, on the Analytics setup form, the new Google Analytics account with the appropriate property is preselected.

![image](https://user-images.githubusercontent.com/73545194/122575957-206aeb00-d049-11eb-8614-95ae8345a7b0.png)

PR #3590, opened 2021-06-16
Bug/3543 preselect matching ga4 account

eugene-manuilov:
## Summary

Addresses issue #3543 

## Checklist

### DIFF
--- a/assets/js/modules/analytics-4/datastore/accounts.js
+++ b/assets/js/modules/analytics-4/datastore/accounts.js
@@ -21,10 +21,8 @@
  */
 import API from 'googlesitekit-api';
 import Data from 'googlesitekit-data';
-import { CORE_SITE } from '../../../googlesitekit/datastore/site/constants';
 import { STORE_NAME } from './constants';
 import { createFetchStore } from '../../../googlesitekit/data/create-fetch-store';
-const { commonActions } = Data;
 
 const fetchGetAccountSummariesStore = createFetchStore( {
 	baseName: 'getAccountSummaries',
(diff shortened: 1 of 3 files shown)


assets/js/modules/analytics-4/datastore/accounts.js
```diff
@@ -41,6 +43,35 @@ const baseInitialState = {
...
+		const ga4PropertyIDs = ( Array.isArray( accounts ) ? accounts : [] ).reduce(
+			( acc, { propertySummaries } ) => [ ...acc, ...propertySummaries.map( ( { _id } ) => _id ) ],
+			[],
+		);
+
+		const property = yield Data.commonActions.await(
+			registry.dispatch( STORE_NAME ).matchPropertyByURL( ga4PropertyIDs, url )
+		);
+
+		return property?._accountID;
```
aaemnnosttv, line 73: How about we return the whole property here, that way we can select it directly rather than matching it again with the `accountID`? Then this could be renamed (and probably moved) to `matchProperty` or maybe `findMatchedProperty`.

assets/js/modules/analytics-4/datastore/accounts.js
```diff
@@ -41,6 +43,35 @@ const baseInitialState = {
...
+		const accounts = yield Data.commonActions.await(
+			registry.__experimentalResolveSelect( STORE_NAME ).getAccountSummaries()
+		);
+
+		if ( ! Array.isArray( accounts ) || accounts.length === 0 ) {
+			return null;
+		}
+
+		const url = registry.select( CORE_SITE ).getReferenceSiteURL();
+		const ga4PropertyIDs = ( Array.isArray( accounts ) ? accounts : [] ).reduce(
```
aaemnnosttv, line 64: The `Array.isArray` check is redundant here since we already have that above.
```suggestion
		const ga4PropertyIDs = accounts.reduce(
```

assets/js/modules/analytics-4/datastore/accounts.js
```diff
@@ -41,6 +43,35 @@ const baseInitialState = {
 };
 
 const baseActions = {
+	/**
+	 * Gets an ID of an account that has a matching property.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return {string} The account ID on success, otherwise NULL.
```
aaemnnosttv, line 51:
```suggestion
	 * @return {(string|null)} The account ID on success, otherwise NULL.
```

aaemnnosttv (commented):
Looks good overall, just a few suggestions to potentially improve.

assets/js/modules/analytics-4/datastore/accounts.js
eugene-manuilov: Yep, makes sense. Updated.

assets/js/modules/analytics-4/datastore/accounts.js
eugene-manuilov: Updated.

aaemnnosttv (approved):
LGTM – Thanks @eugene-manuilov !

Issue #3545, opened 2021-06-09
Display measurement ID on the analytics settings page next to GA4 Property ID


eclarke1:
## Feature Description

Display measurement ID on the analytics settings page instead of ga4 property ID or next to it. Show measurement ID as well as Property, with the GA4 property below the blue link.

![image](https://user-images.githubusercontent.com/56588136/121396255-152e0600-c94b-11eb-9e18-c9fe2cd88bb9.png)

[Related Asana task](https://app.asana.com/0/1200316599642381/1200447371742663)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* If the `ga4Setup` feature flag is enabled and a GA4 property (and its web data stream plus the web data stream's measurement ID) is set, the measurement ID should be displayed in Analytics `SettingsView`, on the right of the "Google Analytics 4 Property" field.
    * Its label should be "Google Analytics 4 Measurement ID", but visually appear as "Measurement ID". As it's next to the "Google Analytics 4 Property" field, it will be visually clear that they're related, but we should still use a fully-descriptive label for accessibility purposes.
* The measurement ID should not appear when editing, since it is derived from the Google Analytics 4 Property selected.

## Implementation Brief
- Using `assets/js/modules/analytics/components/settings/SettingsView.js`,
    - Get the GA4 measurement ID via the `getMeasurementID` selector of the GA4 data store.
    - Duplicate the markup of the section (the `googlesitekit-settings-module__meta-item` container ) where the GA4 property ID is displayed with the following updates:
        - Label should be as per AC, prepended by `Google Analytics 4`. The latter should be wrapped between the `VisuallyHidden` so that it's only visible to screen readers to provide context for the user.
        - Value should be the GA4 measurement ID.
- Add story to reflect the state of the component

### Test Coverage
* No new tests to be added.

### Visual Regression Changes
* N/A

## QA Brief

* Enable the `ga4setup` feature
* Set up Analytics with an account that includes a GA4 property ([these instructions may help](https://app.asana.com/0/1200316599642381/1200379030254853))
* Navigate to the Analytics settings (`/wp-admin/admin.php?page=googlesitekit-settings#settings/analytics/view`)
* Verify that the correct `measurementID` is present on the settings overview as per the ACs.
* Right click the `Measurement ID` subtitle in order to inspect the source code.
[...]

PR #3597, opened 2021-06-17
Display measurement ID on analytics settings page (3545).

johnPhillips:
## Summary

Addresses issue #3545

## Checklist


assets/js/modules/analytics/components/settings/SettingsView.js
```diff
@@ -103,6 +105,17 @@ export default function SettingsView() {
 							<DisplaySetting value={ ga4PropertyID } />
 						</p>
 					</div>
+					<div className="googlesitekit-settings-module__meta-item">
+						<h5 className="googlesitekit-settings-module__meta-item-type">
+							<VisuallyHidden>
+								{ __( 'Google Analytics 4', 'google-site-kit' ) }
+							</VisuallyHidden>
+							{ __( 'Measurement ID', 'google-site-kit' ) }
```
tofumatt, line 113: Unfortunately this won't actually work as-expected, because these two strings end up being presented to translators as two disparate strings and thus can't be translated together as-intended. In a language, for instance, where this should read more like "Measurement ID for Google Analytics 4", there'd be no way to properly translate this string.

In order to have a `gettext` string with a React component in it, you'll want to use `createInterpolateElement` from `'@wordpress/element'`.

So what needs to happen is something more like:

```suggestion
							{ createInterpolateElement(
								__( '<visuallyhidden>Google Analytics 4</visuallyhidden> Measurement ID', 'google-site-kit' ),
								{
									visuallyhidden: <VisuallyHidden />,
								}
							) }
```

tofumatt (changes requested):
This needs an adjustment to the i18n approach, but otherwise looks good!

assets/js/modules/analytics/components/settings/SettingsView.js
johnPhillips: Ah - useful to know, thank you. I've added this to the PR.

tofumatt: Oh, that's so frustrating. It's happening because you aren't supplying the component type here, but an instance of it without `children`. `createInterpolateElement` then modifies the component and adds the `children` text to it.

Anyway, `children` should never be a required prop like this, I think that's an oversight. The easy fix, but also the correct one, is to mark `children` as not required by `VisuallyHidden`.

I understand the intent with marking it as a required prop, but it's not necessary.

I'll make the change so we can get this merged, thanks! 👍🏻

tofumatt:
@googlebot I consent.

tofumatt (approved):

PR #3608, opened 2021-06-21
Fix date for adsense data test.

danielgent:
## Summary

Addresses issue #2585

## Checklist

### DIFF
--- a/.github/workflows/visual-regression.yml
+++ b/.github/workflows/visual-regression.yml
@@ -12,6 +12,7 @@ on:
       - '!assets/**/__tests__/**/*.js'
       - '!assets/**/test/*.js'
       - '!assets/**/*.test.js'
+      - 'stories/**'
       - 'tests/backstop/**'
       - '.storybook/**'
       - '**.scss'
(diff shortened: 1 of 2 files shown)


.github/workflows/visual-regression.yml
```diff
@@ -12,6 +12,7 @@ on:
       - '!assets/**/__tests__/**/*.js'
       - '!assets/**/test/*.js'
       - '!assets/**/*.test.js'
+      - 'stories/**'
```
aaemnnosttv, line 15: Let's also add a pattern for matching stories written in CSF which may not be in this directory.

aaemnnosttv (commented):
LGTM – just one more thing to add.

.github/workflows/visual-regression.yml
danielgent: Makes sense @aaemnnosttv. I changed the pattern and you can see the VRTs are running!

aaemnnosttv: Thanks @danielgent, although the fact that they're running now can also be attributed to this path 😉 
https://github.com/google/site-kit-wp/blob/890e24fac0144fdcc08f4eb02c54c4170f34e757/.github/workflows/visual-regression.yml#L10

The new pattern should match them all though so +1 to this change 👍

aaemnnosttv (approved):
Great, thanks!

Issue #3519, opened 2021-06-08
Implement Idea Hub widget functionality for saving and unsaving an idea


felixarntz:
The Idea Hub widget's functionality to save and "unsave" an idea should be implemented.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The Idea Hub widget functionality for saving and unsaving an idea should be implemented:
    * When clicking on the pin icon in the "New" tab, the idea should be saved via the `saveIdea` action of the `modules/idea-hub` data store. After completing the action, both the "New" and "Saved" tabs should be "force-refreshed" via re-running the two respective API calls after clearing their client-side caches. The number on the "Saved" tab should accordingly increase by 1.
    * When clicking on the pin icon in the "Saved" tab, the idea should be unsaved via the `unsaveIdea` action of the `modules/idea-hub` data store. After completing the action, both the "New" and "Saved" tabs should be "force-refreshed" via re-running the two respective API calls after clearing their client-side caches. The number on the "Saved" tab should accordingly decrease by 1.

## Implementation Brief
All the following will be inside [assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Idea.js](https://github.com/google/site-kit-wp/blob/90cada68c0294b915f9ac71e17b780e4d98395d6/assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Idea.js).

In the `handlePin` callback, call the `saveIdea` action from the Idea Hub store. 
- Pass the `ideaName` (i.e. the component's `name` prop) and await the response. (eg. `await dispatch( MODULES_IDEA_HUB ).saveIdea( name )`)
- Clear the cache by calling [`await invalidateCache()`](https://github.com/google/site-kit-wp/blob/d778e6b404670e446c394e925c6e7a77e0caac94/assets/js/googlesitekit/api/index.js#L299) for the `new-ideas` and `saved-ideas` endpoints (i.e. call it twice, passing `'modules', 'idea-hub', 'new-ideas'` and `'modules', 'idea-hub', 'saved-ideas'` respectively).
- Force-refresh the component by dispatching the Idea Hub store's `fetchGetSavedIdeas()`  and `fetchGetNewIdeas()` actions. Verify that the number on the 'Saved' tab increases by 1.

Now do a similar thing for un-saving (unpinning) an idea.
In the `handleUnpin` callback, call the `unsaveIdea` action from the Idea Hub store. 
[...]

PR #3609, opened 2021-06-21
Add functionality to pin and unpin an idea.

asvinb:
## Summary

Addresses issue #3519 

## Relevant technical choices
Instead of dispatching the `fetchGetSavedIdeas()` and `fetchGetNewIdeas()` actions to refresh the data, the reducerCallback was used to manage the data.

## Checklist


assets/js/modules/idea-hub/datastore/idea-state.js
```diff
@@ -49,6 +49,35 @@ const fetchPostUpdateIdeaStateStore = createFetchStore( {
...
+			};
+		}
+
+		if ( idea.saved === false ) {
+			return {
+				...state,
+				savedIdeas: ( state.savedIdeas || [] ).filter( ( { name } ) => name !== idea.name ),
+			};
+		}
+	},
```
eugene-manuilov, line 80: Let's also return the `state` as is at the very end of the `reducerCallback` function.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Idea.js
```diff
@@ -49,23 +50,33 @@ import UnpinIcon from '../../../../../../svg/idea-hub-unpin.svg';
...
-	}, [ name ] );
-
-	const handleUnpin = useCallback( () => {
-		// @TODO: Implement callback.
-		global.console.log( `Unpinned: ${ name }` );
-	}, [ name ] );
+	const refreshNewSavedIdeas = useCallback( async () => {
+		await API.invalidateCache( 'modules', 'idea-hub', 'new-ideas' );
+		await API.invalidateCache( 'modules', 'idea-hub', 'saved-ideas' );
+	}, [] );
```
eugene-manuilov, line 64: Let's better move it to the `controlCallback` function of the `fetchPostUpdateIdeaStateStore` fetch store. We need to do the following:

1. Update the `controlCallback` function to return a new promise.
1. In the promise callback:
   1. Call and wait for results of the `API.set(...)` function.
   1. Call invalidate cache functions
   1. Resolve the promise passing the results of the `API.set(...)` function call.

eugene-manuilov (changes requested):
Thanks, @asvinb. It looks good to me. Added two suggestions and then it should be good to go.

assets/js/modules/idea-hub/datastore/idea-state.js
asvinb: 👍

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Idea.js
asvinb: Updated @eugene-manuilov Can you check?

eugene-manuilov: Yes, looks good to me. Thanks!

eugene-manuilov (changes requested):
Thanks, @asvinb. It looks good to me. Could you please fix js tests? Once all checks pass, I'll approve and merge this PR.

asvinb:
@eugene-manuilov Failing test fixed. Can you check whether the approach is good?
Thanks!

assets/js/modules/idea-hub/datastore/idea-state.js
```diff
@@ -32,23 +32,63 @@ const { receiveError, clearError } = errorStoreActions;
...
+
+		try {
+			const results = await API.set( 'modules', 'idea-hub', 'update-idea-state', params );
+			await API.invalidateCache( 'modules', 'idea-hub', 'new-ideas' );
+			await API.invalidateCache( 'modules', 'idea-hub', 'saved-ideas' );
+
+			resolve( results );
+		} catch ( err ) {
+			reject( err );
+		}
```
eugene-manuilov, line 52: Actually, I think we can simplify it once again and get rid of `new Promise` approach. How about doing this?

```javascript
controlCallback: ( { name, saved, dismissed } ) => {
    const params = { name };

    if ( saved !== undefined ) {
        params.saved = saved;
    } else {
        params.dismissed = dismissed;
    }

    return API.set( 'modules', 'idea-hub', 'update-idea-state', params ).then( () => async {
        await API.invalidateCache( 'modules', 'idea-hub', 'new-ideas' ).catch( () => {} );
        await API.invalidateCache( 'modules', 'idea-hub', 'saved-ideas' ).catch( () => {} );
    } );
},
```

This should also work, shouldn't it? Sorry, for the incorrect suggestion at the beginning.

eugene-manuilov (changes requested):

assets/js/modules/idea-hub/datastore/idea-state.js
asvinb: Yep, works with a few tweaks from your suggestion :)

eugene-manuilov: Ah... yes, you are right, we need to return the result :laughing:

assets/js/modules/idea-hub/datastore/idea-state.js
```diff
@@ -32,15 +32,21 @@ const { receiveError, clearError } = errorStoreActions;
 
 const fetchPostUpdateIdeaStateStore = createFetchStore( {
 	baseName: 'updateIdeaState',
-	controlCallback: ( { name, saved, dismissed } ) => {
+	controlCallback: async ( { name, saved, dismissed } ) => {
```
eugene-manuilov, line 35: This `async` isn't needed here:
```suggestion
	controlCallback: ( { name, saved, dismissed } ) => {
```

eugene-manuilov (changes requested):
Thanks, @asvinb! Looks good to me. Left one suggestion that I'll apply by myself.

eugene-manuilov (approved):

Issue #3528, opened 2021-06-09
User feedback show terms in the footer


eclarke1:
## Feature Description

Potential to add 'terms and conditions' section at the bottom of the user feedback notice as per discussion [in Asana](https://app.asana.com/0/1200379030254833/1200441839407985).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* _If_ the current user has not opted in to Site Kit's regular Analytics tracking, the user feedback survey (specifically only the first question) should also include a footer informing about the data being sent to Google.
    * The copy should be: _ By continuing, you agree to allow Google to use your answers and account info to improve services, per our Privacy & Terms._ (this is intentionally slightly different from the text in the screenshot below, as we don't send any system information)
    * The word "Privacy" should link to `https://myaccount.google.com/privacypolicy?hl={currentLocale}`
    * The word "Terms" should link to `https://policies.google.com/terms?hl={currentLocale}`
    * The locale value used for these (and in principle any URLs using the `hl` query parameter) should only include the first two segments of the actual locale (see #2105).
    * It should have a separator above with equal spacing above and below, and generally it should follow the design from the following screenshot (which shows a demo survey in one of Google's own services):
![image (3)](https://user-images.githubusercontent.com/3531426/121406634-3f63d180-c913-11eb-9a22-47d66d4ba227.png)

## Implementation Brief
* Using `assets/js/googlesitekit/datastore/site/info.js`,
    * Update the `getGoogleSupportURL` selector to have the host of the URL dynamic (default value: `support.google.com`) so that the host can be specified and not hardcoded to `support.google.com`.
* Using `assets/js/components/surveys/CurrentSurvey.js`,
    * Query the user data store via the `isTrackingEnabled` selector to check if tracking is enabled.
    * Check if `isTrackingEnabled` is still resolving, via the `hasFinishedResolution` selector of the user data store and return `null` if this is the case, meaning that the survey should not be displayed if it's still resolving.
    * Get the "Privacy" and "Terms" URL by querying the site data store via the `getGoogleSupportURL` selector passing the appropriate parameters.
    * If user tracking is explicitly set to `false`, and the question ordinal is `1`, then display the text as per the AC wrapped in a `p` tag, which is itself wrapped within a container with the `googlesitekit-survey__footer` class name. For e.g:
        ```js
        <div className="googlesitekit-survey__footer">
            <p>...</p>
        </div>
        ```
[...]

Issue #2105, opened 2020-09-25
AdSense dashboard error when Portuguese AO90 language is selected


jamesozzie:
## Bug Description

As mentioned in a [WordPress support topic](https://wordpress.org/support/topic/site-kit-message-data-error-in-adsense-and-missing-add-txt-code/) when the Portuguese (AO90) language is active the below dashboard error appears. No AdSense metrics are available and there's no obvious way to bypass this without switching language at present.

``Invalid value 'pt_PT_ao90'. Values must match the following regular expression: '[a-zA-Z_]'`

The below response appears in the console:

```
googlesitekit-dashbo…b0e46cfbfbb5db.js:4 WP Error in data response 
Object
code: 400
data: {status: 400, reason: "invalidParameter"}
message: "Invalid value 'pt_PT_ao90'. Values must match the following regular expression: '[a-zA-Z_]+'"
__proto__: Object
```
`
## Steps to reproduce

1. Setup AdSense on a site
2. Set Portuguese AO90 as you primary language 
3. Visit the main Site Kit or AdSense specific dashboard

## Screenshots

Main dashboard error
![image](https://user-images.githubusercontent.com/41326532/94273424-50fb7980-ff3c-11ea-8a71-501e412e91f4.png)

AdSense dashboard error

![image](https://user-images.githubusercontent.com/41326532/94273516-6ffa0b80-ff3c-11ea-9f2f-ada00c033179.png)

## Additional Context

- Ads don't appear to be impacted
- No warnings or errors appear when visiting the AdSense dashboard on the AdSense Web UI
- Previous discussion on locale settings in #1309

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Locales sent along with AdSense API requests should always be sanitized to only include the standard first two segments (e.g. `pt_PT` instead of `pt_PT_ao90`, or `de_DE` instead of `de_DE_formal`).

## Implementation Brief

While the API docs don't explicitly state the regex pattern for matching locale, https://developers.google.com/adsense/management/v1.4/reference/reports/generate, it's clear that it's facing issues with integers in the locale based on the screenshot.

There are two WordPress functions responsible for fetching locale, both of which will return `pt_PT_ao90` when it is selected:
get_locale - https://developer.wordpress.org/reference/functions/get_locale/
get_user_locale - https://developer.wordpress.org/reference/functions/get_user_locale/

We will create a new function on the `Context` class
[...]

PR #3610, opened 2021-06-22
Enhancement/3528 user feedback terms

eugene-manuilov:
## Summary

Addresses issue #3528

## Checklist


assets/js/googlesitekit/datastore/site/urls.js
```diff
@@ -0,0 +1,134 @@
...
+
+export const actions = {};
+
+export const controls = {};
+
+export function reducer( state ) {
+	return state;
+}
+
+export const resolvers = {};
```
tofumatt, line 34: Since these are all empty we can actually omit them, our `combineStores` utility will handle missing arguments.

I don't think there's much use in having this boilerplate here. 🤷🏻‍♂️

assets/js/googlesitekit/datastore/site/urls.js
```diff
@@ -0,0 +1,134 @@
...
+	 * @param {Object} state         Data store's state.
+	 * @param {Object} [args]        Optional arguments for the resulting URL.
+	 * @param {string} [args.path]   URL path to build complete URL with starting slash.
+	 * @param {Object} [args.query]  Object to append query to the URL.
+	 * @param {string} [args.hash]   Optional hash.
+	 * @param {string} [args.locale] Optional locale.
+	 * @return {(string|null)} The URL containing the user's locale or `null` if path is not set.
+	 */
+	getGoogleSupportURL( state, args ) {
+		return selectors.getGoogleURL( state, { ...args, website: 'https://support.google.com' } );
```
tofumatt, line 91: This is quite minor, but it's better practise to use `createRegistrySelector` here. I know in this case we don't have any resolvers to worry about, but if we _did_, this wouldn't call them.

Using `createRegistrySelector` will serve as a better reference so let's use that instead.

assets/js/googlesitekit/datastore/site/urls.js
```diff
@@ -0,0 +1,134 @@
...
+	 * @param {Object} state          Data store's state.
+	 * @param {Object} [args]         Optional arguments for the resulting URL.
+	 * @param {string} [args.website] Base URL hostname and schema.
+	 * @param {string} [args.path]    URL path to build complete URL with starting slash.
+	 * @param {Object} [args.query]   Object to append query to the URL.
+	 * @param {string} [args.hash]    Optional hash.
+	 * @param {string} [args.locale]  Optional locale.
+	 * @return {(string|null)} The URL containing the user's locale or `null` if path is not set.
+	 */
+	getGoogleURL( state, args ) {
```
tofumatt, line 51: This is a handy selector, and I know it's tested indirectly elsewhere, but can you add a few tests for it? Mainly to serve as a usage reference 🙂

assets/js/googlesitekit/datastore/site/urls.js
```diff
@@ -0,0 +1,134 @@
...
+	getGooglePrivacyPolicyURL( state ) {
+		return selectors.getGoogleURL( state, {
+			website: 'https://myaccount.google.com',
+			path: '/privacypolicy',
+			locale: getLocaleLanguage(),
+		} );
+	},
+
+	/**
+	 * Gets the Google terms URL.
```
tofumatt, line 111:
```suggestion
	 * Gets the Google terms of service URL.
```

assets/js/util/i18n.js
```diff
@@ -340,3 +340,16 @@ export const getLocale = ( _global = global ) => {
...
+ *
+ * @since n.e.x.t
+ *
+ * @param {Object} _global The global window object.
+ * @return {string} Current locale language.
+ */
+export function getLocaleLanguage( _global = global ) {
+	const locale = getLocale( _global );
+	return locale.substr( 0, 2 );
+}
```
tofumatt, line 355: These should definitely only be two-characters? For languages that share the first two characters but are quite different (Brazilian vs Continental Portuguese, Traditional vs Simplified Chinese, etc.) this will result in quite fuzzy language matching for these URLs, won't it?

I know there are redirects for `en-US` to `en`, but if that the case for other locales?

----

I actually just tested this with the Chinese example, and this removes precision:

`https://support.google.com/?hl=zh-TW` redirects to `https://support.google.com/?hl=zh-Hant`
`https://support.google.com/?hl=zh` redirects to `https://support.google.com/?hl=zh-Hans`

So there _are_ differences there; we shouldn't be using only the first two characters for locales.

assets/sass/components/surveys/_googlesitekit-surveys.scss
```diff
@@ -69,6 +69,16 @@
 			padding-top: 12px;
 			text-align: left;
 		}
+
+		.googlesitekit-survey__terms {
+			margin: 0 0 -12px;
+			padding-top: 12px;
+
+			a[target="_blank"] {
```
tofumatt, line 77: We do have a `.googlesitekit-cta-link--external` class we can target here instead of this selector, which I _think_ is a bit more efficient.
```suggestion
			.googlesitekit-cta-link--external {
```

assets/js/util/i18n.js
tofumatt: Using `getLocale` should be fine here.

tofumatt (changes requested):
Seems like we should be using `getLocale`, not creating a new function that removes some of the info from the locale string.

In some cases there are redirects from things like `en_US` to `en`, but that's not always the case, so let's not remove the second part of the locale string.

assets/js/googlesitekit/datastore/site/urls.js
eugene-manuilov: Ok, removed.

assets/js/googlesitekit/datastore/site/urls.js
eugene-manuilov: Sounds good, updated.

assets/js/googlesitekit/datastore/site/urls.js
eugene-manuilov: Added

assets/js/googlesitekit/datastore/site/urls.js
eugene-manuilov: Updated.

assets/sass/components/surveys/_googlesitekit-surveys.scss
eugene-manuilov: Updated.

assets/js/util/i18n.js
eugene-manuilov: From AC:

> The word "Privacy" should link to ...google.com/privacypolicy?hl={**currentTwoLetterLanguageBasedOnLocale**}
> The word "Terms" should link to ...google.com/terms?hl={**currentTwoLetterLanguageBasedOnLocale**}

tofumatt: @felixarntz: Can you clarify why that is required/in the ACs, because it seems like it will send users to the wrong URL/locale.

tofumatt (commented):
I see the issue is with the ACs here, but I think they're worth revisiting before merging this. I think the two-letter-codes might be an oversight, because of the redirects used for things like `en-US`. But I think we need to amend them to support differences between languages like Simplified vs Traditional Chinese.

I've assigned @felixarntz to weigh in on this as he wrote the ACs.

(Otherwise this looks good though.)

assets/js/util/i18n.js
felixarntz: @eugene-manuilov  @tofumatt Hmm that's a fair point. Apologies, I had been under the assumption that Google's different locale support only considers the language part - so maybe it's best to revert that change and use both the first segments of the locale string present. It's important though to limit it to the first two segments, not just take the whole string, which at least in PHP side can have more elements, which may cause problems in non-WordPress systems (see #2105). I wonder whether we should generally stick to getting the locale from the server eventually, rather than a mix of server and client, which may be different - we should avoid having mixed language content in Site Kit where possible.

Maybe for this issue/PR:
* Let's check what the client-side locale values can look like. If we're certain they don't include a third segment (like what is possible in WP), we can just take the full string.
* Otherwise, we should strip the client-side locale down to the first two segments only and use that.

I'll update the ACs to reflect that.

eugene-manuilov: Thanks, @felixarntz. PR is updated.

eugene-manuilov:
@tofumatt this is ready for your review.

assets/js/googlesitekit/datastore/site/urls.js
```diff
@@ -0,0 +1,122 @@
...
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {string} The Google privacy policy URL.
+	 */
+	getGooglePrivacyPolicyURL: createRegistrySelector( ( select ) => () => {
+		return select( STORE_NAME ).getGoogleLocaleAwareURL( {
+			website: 'https://myaccount.google.com',
+			path: '/privacypolicy',
+			locale: ( getLocale() || '' ).substr( 0, 5 ),
```
tofumatt, line 99: Why the `.substr( 0, 5 )` here?

A locale like `https://policies.google.com/privacy?hl=zh-hans` is valid, and any extra data will be disregarded/redirected by Google here. Examples:

https://policies.google.com/privacy?hl=en-USA redirects to https://policies.google.com/privacy?hl=en

But https://policies.google.com/privacy?hl=zh-hans and https://policies.google.com/privacy?hl=zh-cw are both valid. I don't think we should be modifying these locale strings at all, as it seems Google is able to recognise longer ones and load the correct content.

Removing any parts of the locale _after_ the first two segments makes sense (see #2105), but we shouldn't assume the second part of locale will always be two characters, because that isn't always how they work.

I guess it depends on the locale standard used by WordPress, which is where we get this data from, right? I can't actually find, [in the WP docs](https://developer.wordpress.org/reference/functions/get_locale/), what the format they use is 🤷🏻‍♂️ 

I guess this might be fine, but is there any harm to passing the results of `getLocale()` directly?

tofumatt (changes requested):
Just one question/note about the approach to trimming the locale here.

assets/js/googlesitekit/datastore/site/urls.js
felixarntz: @tofumatt @eugene-manuilov Agreed. Let's not rely on any specific character length, we should only consider the first two segments though (essentially no matter their length). Having a third segment isn't really standardized (see #2105), so I think it's safer to cut off any additional segments more than the first two.

eugene-manuilov: Ok, updated.

assets/js/googlesitekit/datastore/site/urls.js
```diff
@@ -0,0 +1,121 @@
...
+	 * @return {(string|null)} The URL containing the user's locale or `null` if path is not set.
+	 */
+	getGoogleLocaleAwareURL( state, args ) {
+		const currentLocale = getLocale();
+		const {
+			website,
+			path,
+			query,
+			hash,
+			locale = currentLocale.match( /^([a-zA-Z]+[-_]?[a-zA-Z]*)/ )?.[ 0 ] || currentLocale,
```
tofumatt, line 49: This looks great, thanks! 👍🏻

tofumatt (commented):
This looks great. I'm going to make the change mentioned here: https://github.com/google/site-kit-wp/issues/3528#issuecomment-867789231

I'm also going to add a test case for the new locale handling we're doing, but aside from that this is awesome. 👍🏻

tofumatt (approved):
Huh, I just noticed something when adding tests: [`getLocale()` itself always returns a two-character string](https://github.com/google/site-kit-wp/blob/ccc65cb6f32341ecf122132987332d880343e9e9/assets/js/util/i18n.js#L330). Google will change `zh-TW` to `zh-hant` so this is fine.

At any rate, tests added and the string mentioned by @bethanylang was updated so this is good-to-go 😄

tofumatt:
@googlebot I consent.

tofumatt:
(Merging with known failing AMP test in Nightly, see https://github.com/ampproject/amp-wp/issues/6416.)

PR #3619, opened 2021-06-23
Return secondary amp mode if this post/page is a web story.

tofumatt:
## Summary

Addresses issue #2998.

## Checklist

### DIFF
--- a/includes/Context.php
+++ b/includes/Context.php
@@ -323,13 +323,13 @@ class Context {
 	 *                     false if AMP not active, or unknown mode
 	 */
 	public function get_amp_mode() {
-		if ( ! class_exists( 'AMP_Theme_Support' ) ) {
-			// If the AMP plugin isn't enabled but this page/post is a web
-			// story, use AMP Secondary mode.
-			if ( is_singular( 'web-story' ) ) {
-				return self::AMP_MODE_SECONDARY;
-			}
+		// If the Web Stories plugin is enabled, consider the site to be running
+		// in Secondary AMP mode.
+		if ( defined( 'WEBSTORIES_VERSION' ) ) {
+			return self::AMP_MODE_SECONDARY;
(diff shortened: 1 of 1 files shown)


includes/Context.php
```diff
@@ -373,7 +373,10 @@ public function get_amp_mode() {
 			) {
 				return self::AMP_MODE_SECONDARY;
 			}
-		} elseif ( function_exists( 'amp_is_canonical' ) ) {
+		} elseif (
+			function_exists( 'amp_is_canonical' ) ||
+			is_singular( 'web-story' )
+		) {
```
aaemnnosttv, line 379: This is too late for this condition since the first line in `get_amp_mode` will return early if the AMP plugin is not installed which is not a requirement/dependency of the Web Stories plugin.

includes/Context.php
```diff
@@ -324,6 +324,12 @@ public function is_amp() {
 	 */
 	public function get_amp_mode() {
 		if ( ! class_exists( 'AMP_Theme_Support' ) ) {
+			// If the AMP plugin isn't enabled but this page/post is a web
+			// story, use AMP Secondary mode.
+			if ( is_singular( 'web-story' ) ) {
```
felixarntz, line 329: @tofumatt In addition to @aaemnnosttv's comment, per https://github.com/google/site-kit-wp/issues/2998#issuecomment-869865609 this shouldn't check whether the current context is a web story (which will _never_ be the case in WP Admin), instead it needs to check whether the Web Stories plugin is active (e.g. check for whether the `WEBSTORIES_VERSION` constant is defined (see https://github.com/google/web-stories-wp/blob/main/web-stories.php).

felixarntz (changes requested):

felixarntz (approved):
Great!

Issue #3255, opened 2021-04-29
Safari issue: Failed to initialize NumberFormat since used feature is not supported in the linked ICU version


jamesozzie:
## Bug Description

There have been reports of the below dashboard error since 1.31.0 was released. This appears when accessing the primary Site Kit dashboard. 
`Failed to initialize NumberFormat since used feature is not supported in the linked ICU version`

One user has reported this only occurs in a Safari browser, troubleshooting ongoing with the other. 

<details>
<summary>Full error details below for 1 user</summary>

```
Failed to initialize NumberFormat since used feature is not supported in the linked ICU version

    in DataBlock
    in div
    in div
    in section
    in LegacySearchConsoleDashboardWidgetOverview
    in withData(LegacySearchConsoleDashboardWidgetOverview)
    in Unknown
    in WithSelect(withData(LegacySearchConsoleDashboardWidgetOverview))
    in div
    in Layout
    in div
    in div
    in div
    in div
    in GoogleSitekitSearchConsoleDashboardWidget
    in FilteredComponent
    in WithFilters(ModuleApp)
    in ModuleApp
    in GoogleSitekitModule
    in RestoreSnapshots
    in ErrorHandler
    in Root
```
</details>

<details>
<summary>Full error for second user (slightly different - **_LegacySearchConsoleDashboardWidgetTopLevel_**)</summary>

```
Failed to initialize NumberFormat since used feature is not supported in the linked ICU version
in DataBlock
in div
in LegacySearchConsoleDashboardWidgetTopLevel
in withData(LegacySearchConsoleDashboardWidgetTopLevel)
in Unknown
in WithSelect(withData(LegacySearchConsoleDashboardWidgetTopLevel))
in FilteredComponent
in FilteredComponent
in WithFilters(LegacyDashboardSearchFunnelInner)
in div
in div
in div
in Layout
in div
in LegacyDashboardSearchFunnel
in FilteredComponent
in FilteredComponent
in FilteredComponent
in WithFilters(LegacyDashboardModule)
in div
in Row
in div
in Grid
in div
in DashboardApp
in GoogleSitekitDashboard
in RestoreSnapshots
in ErrorHandler
in Root
```

</details>

Impacted users:
[...]

PR #3620, opened 2021-06-23
Allow numberFormat to degrade gracefully.

ivankruchkoff:
## Summary

Addresses issue #3255 

## Checklist


assets/js/util/i18n.js
```diff
@@ -285,8 +289,43 @@ export const numFmt = ( number, options = {} ) => {
...
+		 * If an error is thrown, we remove some troublesome params from the formatOptions object and fallback to no formatting.
+		 *
+		 * This allows us to degrade somewhat gracefully without breaking the dashboard for users of unaffected browsers.
+		 */
+		return new Intl.NumberFormat( locale, formatOptions ).format( number );
+	} catch ( error ) {
+		logOnce( `Unable to: new Intl.NumberFormat( ${ JSON.stringify( locale ) }, ${ JSON.stringify( formatOptions ) } ).format( ${ number } );`, 'warn' );
+
+		// Remove these key/values from formatOptions and try again.
+		const removeFormatOptions = {
```
aaemnnosttv, line 306: The name is a little confusing here since it's fairly close to `reducedFormatOptions`, it makes the code below a bit difficult to read. How about `unstableFormatOptionValues` ?

Similarly, I think it would be a little easier to read in the loop to have a separate variable for the any-value options (e.g. `unstableFormatOptions`). Then we could just check `unstableFormatOptions.includes( key )` or use an object with the same `*` just to be able to check by key.

assets/js/util/i18n.js
```diff
@@ -285,8 +289,43 @@ export const numFmt = ( number, options = {} ) => {
...
+			style: 'unit',
+			signDisplay: '*', // * denotes any value.
+			compactDisplay: '*',
+		};
 
-	return new Intl.NumberFormat( locale, formatOptions ).format( number );
+		const reducedFormatOptions = {};
+
+		for ( const [ key, value ] of Object.entries( formatOptions ) ) {
+			if ( removeFormatOptions[ key ] && [ value, '*' ].includes( removeFormatOptions[ key ] ) ) {
```
aaemnnosttv, line 317: The second part of the conditional here is a little clever 😄 

Given the comments above, let's make this a bit more straightforward using separate variables so we can explicitly check by key and value, or just the key.

assets/js/util/test/numberFormat.js
```diff
@@ -87,6 +87,7 @@ describe( 'numberFormat', () => {
 		expect(
 			numberFormat( 123456789.87, { locale: 'de-DE' } )
 		).toStrictEqual( '123.456.789,87' );
+		expect( console ).not.toHaveWarned();
```
aaemnnosttv, line 90: Nice touch 👍  is it possible to add a test or two that will trigger the fallback as well?

aaemnnosttv (changes requested):
Thanks @ivankruchkoff – this looks great, just a few things to address but nothing major 👍

assets/js/util/i18n.js
ivankruchkoff: Addressed in https://github.com/google/site-kit-wp/pull/3620/commits/96d0593c48ce49b0227e787020007e9e966f7a7e

assets/js/util/i18n.js
ivankruchkoff: Updated in https://github.com/google/site-kit-wp/pull/3620/commits/96d0593c48ce49b0227e787020007e9e966f7a7e

assets/js/util/i18n.js
```diff
@@ -285,8 +289,43 @@ export const numFmt = ( number, options = {} ) => {
...
+				continue;
+			}
+			reducedFormatOptions[ key ] = value;
+		}
+
+		try {
+			return new Intl.NumberFormat( locale, reducedFormatOptions ).format( number );
+		} catch ( error2 ) {
+			return new Intl.NumberFormat( locale ).format( number );
+		}
```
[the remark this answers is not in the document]
ivankruchkoff: Updated in https://github.com/google/site-kit-wp/pull/3620/commits/fec47948962e47c7f28c9cc558fd4de7656cf038

assets/js/util/test/numberFormat.js
ivankruchkoff: Yep, addressed in https://github.com/google/site-kit-wp/pull/3620/commits/e07d41e9a513daa7c1751bdba7105e69dfc707a9

assets/js/util/test/numberFormat.js
```diff
@@ -115,5 +116,98 @@ describe( 'numberFormat', () => {
...
+
+	const NumberFormat = Intl.NumberFormat;
+
+	/*
+	 * This function mocks the implementation of certain errors identified in
+	 * https://github.com/google/site-kit-wp/issues/3255
+	 * so that we can test that we've properly handled them.
+	 */
+	const mockNumberFormat = () => {
+		global.Intl.NumberFormat = ( locales, options ) => {
```
aaemnnosttv, line 130: This will probably work but it would be more appropriate to let Jest handle the mocking rather than doing this more manually. We should be able to do so as follows:

```js
NumberFormatSpy = jest.spyOn( global.Intl, 'NumberFormat' );
NumberFormatSpy.mockImplementation( ( locales, options ) => {
```

That way, we can clear the mock in an `afterEach` so it doesn't potentially cause other problems in the event of a failure. It also allows us to do things like `expect( global.Intl.NumberFormat ).toBeCalled( ... )` if we wanted to. The main benefits are consistency and resilient cleanup though.

assets/js/util/test/numberFormat.js
```diff
@@ -115,5 +116,98 @@ describe( 'numberFormat', () => {
...
+				currencySign: 'accounting',
+				style: 'unit',
+				signDisplay: '*', // * denotes any value.
+				compactDisplay: '*',
+			};
+			for ( const [ key, value ] of Object.entries( options ) ) {
+				if ( optionsThatError[ key ] && [ value, '*' ].includes( optionsThatError[ key ] ) ) {
+					throw new Error( 'Implementation error' );
+				}
+			}
```
aaemnnosttv, line 145: Let's export these things from the module so we're not maintaining these in multiple places.

How about we extract all of this to a single function like `omitUnstableNumberFormatOptions( formatOptions ) : numberFormatOptionsWithoutTheUnstableOnes` (a la [`_.omit`](https://lodash.com/docs/4.17.15#omit))? This should be annotated as `@private`.

assets/js/util/test/numberFormat.js
```diff
@@ -115,5 +116,98 @@ describe( 'numberFormat', () => {
...
+			signDisplay: 'never', // This parameter will be removed.
+			style: 'percent',
+			maximumFractionDigits: 1,
+		} );
+
+		// Ensure we don't log more than once.
+		expect( console.warn ).toHaveBeenCalledTimes( 1 ); // eslint-disable-line no-console
+
+		// Restore the regular behaviour.
+		Intl.NumberFormat = NumberFormat;
```
aaemnnosttv, line 184: As mentioned above, this should be handled outside of the test function itself to avoid the mocked state leaking between tests.

assets/js/util/test/numberFormat.js
```diff
@@ -115,5 +116,98 @@ describe( 'numberFormat', () => {
...
+	} );
+
+	it( 'degrades gracefully when style:unit is encountered in certain scenarios', () => {
```
aaemnnosttv, line 187:
```suggestion
	it( 'degrades gracefully when the `style:unit` option is provided', () => {
```

assets/js/util/test/numberFormat.js
```diff
@@ -115,5 +116,98 @@ describe( 'numberFormat', () => {
...
+				if ( optionsThatError[ key ] && [ value, '*' ].includes( optionsThatError[ key ] ) ) {
+					throw new Error( 'Implementation error' );
+				}
+			}
+
+			return NumberFormat( locales, options );
+		};
+	};
+
+	it( 'degrades gracefully when signDisplay is encountered in certain scenarios', () => {
```
aaemnnosttv, line 151: Let's avoid ambiguity in the test names as much as we can. In this case "certain scenarios" does not add much useful context for the reader so one needs to read the whole test to know what these are. Instead, let's use the title to describe the situation(s) where the behavior does change. For this one in particular, it would be any value other than the default.
```suggestion
	it( 'degrades gracefully when `signDisplay` has any value other than the default of `auto`', () => {
```

aaemnnosttv (changes requested):
Thanks @ivankruchkoff! This is looking good, just a few points left to address then I think this should be good to go 👍

assets/js/util/test/numberFormat.js
ivankruchkoff: Updated in https://github.com/google/site-kit-wp/pull/3620/commits/98bec60f18fb859b20a62f0d978f98d2c5b30692

assets/js/util/test/numberFormat.js
ivankruchkoff: Updated in https://github.com/google/site-kit-wp/pull/3620/commits/98bec60f18fb859b20a62f0d978f98d2c5b30692

assets/js/util/test/numberFormat.js
ivankruchkoff: That won't make sense in this scenario as our test isn't intended to clean up the formatOptions, that's the role of `numberFormat`. Instead we're trying to imitate the browser behaviour and throw the appropriate error as Safari would in these instances so that it triggers the catch clause in our `numberFormat` function.

assets/js/util/test/numberFormat.js
```diff
@@ -115,5 +116,105 @@ describe( 'numberFormat', () => {
...
+		// Error message that browser throws on error.
+		const errorMessage = 'TypeError: Failed to initialize NumberFormat since used feature is not supported in the linked ICU version';
+
+		/**
+		 * This function mocks the implementation of certain errors identified in
+		 * https://github.com/google/site-kit-wp/issues/3255
+		 * so that we can test that we've properly handled them.
+		 */
+		const mockNumberFormat = () => {
+			const NumberFormat = Intl.NumberFormat;
```
aaemnnosttv, line 140: Let's save this reference at the top module level where it will always reference the original. Otherwise it could end up grabbing a reference to the mocked implementation, depending on when it's called.

assets/js/util/test/numberFormat.js
```diff
@@ -115,5 +116,105 @@ describe( 'numberFormat', () => {
...
+			NumberFormatSpy = jest.spyOn( global.Intl, 'NumberFormat' );
```
aaemnnosttv, line 141: Let's move this line to a `beforeEach` as there's no harm in setting up the spy – it will use the original implementation by default until `mockImplementation` or `mockImplementationOnce` are used. This also simplifies the tear down since we wouldn't need to check for `NumberFormatSpy.mockRestore`.

assets/js/util/test/numberFormat.js
```diff
@@ -115,5 +116,105 @@ describe( 'numberFormat', () => {
...
+					currencySign: 'accounting',
+					style: 'unit',
+					signDisplay: '*', // * denotes any value.
+					compactDisplay: '*',
+				};
+				for ( const [ key, value ] of Object.entries( options ) ) {
+					if ( optionsThatError[ key ] && [ value, '*' ].includes( optionsThatError[ key ] ) ) {
+						throw new TypeError( errorMessage );
+					}
+				}
```
aaemnnosttv, line 157: Maybe we can simplify this a bit while improving readability of the tests by introducing a little helper here like `createThrowIfOptionMatch( key, [ value ] )` which would return a function that throws if the key and optionally the value matches, otherwise forwarding through to the original implementation.

E.g.
```js
function createThrowIfOptionMatch( key, value ) {
    return ( locale, options = {} ) => {
        if ( options.hasOwnProperty( key ) ) {
            if ( value === options[ key ] || value === undefined ) {
                throw new Error( errorMessage );
            }
        }
        return originalNumberFormat( locale, options );
    }
}
```

Then in the test we could call e.g. `NumberFormatSpy.mockImplementation( createThrowIfOptionMatch( 'currencyDisplay', 'narrow' ) )` rather than `mockNumberFormat`.

What do you think?

aaemnnosttv (commented):
Thanks @ivankruchkoff – a few final thoughts for you about the tests and then we should be all set. Let me know what you think.

assets/js/util/test/numberFormat.js
aaemnnosttv: Yeah, my mistake; this loop throws rather than modifying the options 😄  Originally I was looking at just extracting the object(s) of unstable options (since these are constant) and saw the loop was duplicated from the original implementation. See my review for my final thoughts here.

assets/js/util/test/numberFormat.js
ivankruchkoff: Updated in https://github.com/google/site-kit-wp/pull/3620/commits/fd3d490c8d8864c2307c0a1e1296a8cfafd92cc0

assets/js/util/test/numberFormat.js
ivankruchkoff: Updated in https://github.com/google/site-kit-wp/pull/3620/commits/fd3d490c8d8864c2307c0a1e1296a8cfafd92cc0

assets/js/util/test/numberFormat.js
ivankruchkoff: Updated in https://github.com/google/site-kit-wp/pull/3620/commits/a2cabed123c702969fffad9ed2e122f7d9edc89c

aaemnnosttv (approved):
Awesome, thanks @ivankruchkoff !

Issue #3516, opened 2021-06-08
Introduce button tooltips to help clarify icons


felixarntz:
In much of Google UI, buttons (_particularly_ icon buttons) have tooltips with a (secondary) text label. While for regular buttons, this is not as critical, for icon buttons this is quite helpful as it helps convey the meaning of each icon. For screen readers, we already have assistive text, but that text is not accessible for e.g. browser users.

Here is a screenshot from Google Drive as example:
<img width="221" alt="Screen Shot 2021-05-26 at 1 55 59 PM" src="https://user-images.githubusercontent.com/3531426/119716300-d6139700-be19-11eb-9a4b-5d773d6d7743.png">
(originally posted in https://github.com/google/site-kit-wp/issues/3379#issuecomment-849039705)

Site Kit should support tooltips with similar appearance for its icon buttons. This is somewhat related to the user feedback epic since a few of those icon buttons are present there.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Site Kit's `Button` component should be enhanced so that, if it displays an icon (via `icon` prop) and a text label is also present (e.g. via `aria-label` prop), that text label displays as a small tooltip with white text on gray background, following the overall design present in the screenshot above.
* The following `aria-label` props related to the above enhancement should be changed to be more concise:
    * In `HelpMenu`, the help icon's `aria-label` should say just "Help" - it does not need the full actions since it appears in a list of menu items and controls where this is unnecessary.
    * In `SurveyHeader`, the dismiss icon's `aria-label` should say just "Dismiss" - the user is in the survey and its title comes just before in the DOM order, so it's unnecessary to state that they're dismissing the survey.
    * In `SurveyQuestionRatingChoice`, the choice icon's `aria-label` should just say `choice.text` - the "icon" suffix is rather useless since for screen reader users it is irrelevant that the choice is an icon, while visually it's obvious that it is an icon anyway (it's a label rather than an `alt` description).

## Implementation Brief
* Install the `@material-ui/core` package.
* Update the `Button` component,
    * Add the `aria-label` prop as it is done in the `Link` component.
    * Wrap the markup for the component within the `Tooltip` component (imported from the `@material-ui/core` package) with the `title` prop set to the `aria-label` value if there is an icon and `aria-label` and no `children`.
    * When wrapping around the `Tooltip` component, add an extra span tag around the markup or else the tooltip is not positioned correctly.
[...]

PR #3626, opened 2021-06-24
Increase z-index for tooltips.

asvinb:
## Summary

Addresses issue #3516 

## Relevant technical choices
Increase the z-index for tooltips

## Checklist


assets/sass/components/global/_googlesitekit-tooltip.scss
```diff
@@ -19,3 +19,8 @@
 .googlesitekit-tooltip.MuiTooltip-tooltipPlacementBottom {
 	margin-top: 0;
 }
+
+// High z-index since the survey modal has a high z-index as well
+.googlesitekit-tooltip-popper.MuiTooltip-popper {
+	z-index: 99999;
```
aaemnnosttv, line 25: This is the same z-index used by the survey modal (and a few others that I found).
https://github.com/google/site-kit-wp/blob/37e9cad8e29c8fa83bd8337bf9403f80243d42a4/assets/sass/components/surveys/_googlesitekit-surveys-modal.scss#L23
The comment here seems to imply that this value should be higher than the survey modal but maybe that's just how I'm reading it. I'm guessing perhaps this was intentional but just want to confirm.

I think ideally, we should have a few variables for z-indexes and then others can be defined relative to those. It's also important to know what WP uses as well to help explain where things need to be relative to one of those. It's a little odd though that most of our z-indexes are less than `15`, but then we jump right to `10000` and `99999`. Changing all of these is not within the scope here but would be great to perhaps open an issue to address more holistically.

aaemnnosttv (commented):
Thanks @asvinb – just one question for you.

assets/sass/components/global/_googlesitekit-tooltip.scss
asvinb: @aaemnnosttv Actually the portal has the same z-index as the WordPress admin bar, which was intentional, just to make sure the portal content will always be on top of any content. Do you think we should lower the z-index? 99999 was like the safest value for me.

aaemnnosttv: Okay, that's fine then, just wanted to confirm. In that case, I think there's still room for improvement overall to use variables for the z-index rather than needing to add comments in all uses to explain, but that can be for another issue 👍

aaemnnosttv (approved):

Issue #3522, opened 2021-06-08
Implement Idea Hub setup UI


felixarntz:
While the Idea Hub module does not have any settings for now (see #3521), it still needs to have a setup screen, for now with just a note about privacy and terms of service of this API that the user has to acknowledge.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The Idea Hub module should have its basic information (in `Idea_Hub::setup_info`) updated:
    * The `description` should be: _ Idea Hub suggests what you can write about next, based on searches that haven't been answered yet_
    * The `homepage` should be entirely removed.
* The Idea Hub module should have a new setting `tosAccepted` which expects a boolean. The module should only consider itself connected (via `Idea_Hub::is_connected`) if the value in that settings is `true`.
* The Idea Hub module should have a setup screen via a `SetupMain` component in `assets/js/modules/idea-hub/components/setup/SetupMain.js`.
    * It should follow the [design in the Figma file](https://www.figma.com/file/YB79eKcGOwyUjlWG8mjDZA/Prompt-Engine-%2F-Idea-Hub?node-id=20332%3A6967), with two exceptions:
        * Instead of the first sentence below the Idea Hub heading, the same description that is present above should be used (taken directly from the Idea Hub module's data).
        * Instead of "Done", the submit button should say "Continue", for consistency with e.g. the AdSense setup.
    * The checkbox on that screen should control the above `tosAccepted` setting. Until the checkbox is clicked, the submit button should be disabled.

## Implementation Brief
* Using `includes/Modules/Idea_Hub.php`,
    * Update the `description` in the `setup_info` method as per the AC and remove `homepage`.
    * Add the `tosAccepted` to the `$required_keys` array of the `is_connected` method. 
* Using `includes/Modules/Idea_Hub/Settings.php`,
    * Add `tosAccepted` to the array found in `get_default`, with `false` as default value.
* Using `assets/js/modules/idea-hub/datastore/base.js`,
    * Add `tosAccepted` to the `settingSlugs` array when calling the `createModuleStore` function.
* Create `assets/js/modules/idea-hub/components/setup/SetupMain.js` with exports the `SetupMain` functional component with the following details:
    * The component should be rendered as per the designs in Figma.
    * Most of the code can be taken from the `Optimize` `SetupMain` component `assets/js/modules/optimize/components/setup/SetupMain.js` with the following changes:
[...]

PR #3629, opened 2021-06-24
Implement Idea Hub setup UI (3522).

johnPhillips:
## Summary

Addresses issue #3522

## Checklist


assets/js/modules/idea-hub/components/setup/Setup.stories.js
```diff
@@ -0,0 +1,80 @@
...
+
+export const DefaultSetup = Template.bind( null );
+DefaultSetup.storyName = 'Set up form';
+DefaultSetup.parameters = { features };
+DefaultSetup.decorators = [
+	( Story ) => {
+		const moduleFixture = [ {
+			slug: 'idea-hub',
+			name: 'Idea Hub',
+			description: "Idea Hub suggests what you can write about next, based on searches that haven't been answered yet",
```
tofumatt, line 53: At first glance, I thought: *I think elsewhere we tend to use gettext/`__()` functions for strings like these where they'd normally be translated*.

But then I thought "okay, this is a fixture for an API response which would always return a string and be translated/have the locale handled in PHP".

So I guess I'm just noting that thought here for future reference—this is correct, but it's sort of a subtle difference because many stories use gettext, even if we can't really view translated stories right now I think.

`</matt-thoughts>` 😅

assets/js/modules/idea-hub/components/setup/Setup.stories.js
```diff
@@ -0,0 +1,80 @@
...
+						</div>
+					</div>
+				</div>
+			</section>
+		</div>
+	);
+}
+
+export const DefaultSetup = Template.bind( null );
+DefaultSetup.storyName = 'Set up form';
```
tofumatt, line 46: "Setup" is a noun, "set up" is a verb; English is annoying 😆 

```suggestion
DefaultSetup.storyName = 'Setup form';
```

assets/js/modules/idea-hub/components/setup/Setup.stories.js
```diff
@@ -0,0 +1,80 @@
...
+			connected: false,
+			dependencies: [],
+			dependants: [],
+			owner: null,
+			forceActive: true,
+		} ];
+
+		const setupRegistry = ( registry ) => {
+			registry.dispatch( CORE_MODULES ).receiveGetModules( moduleFixture );
+			registry.dispatch( STORE_NAME ).receiveGetSettings( { tosAccepted: true } );
```
tofumatt, line 67: Let's not set this to `true` in the story by default; the way the module itself works is that it defaults to `false`. This creates a bit of an odd story UX because it made me think this is the default in the actual plugin, which would be:

1. weird (the user should need to explicitly opt-in)
2. opposite the IB
3. opposite the functionality of the plugin

So a separate story for it checked (though frankly the use can check it, so not _needed_) would work, but let's not have the "Default" story start with it checked.

Not sure if you can remove this line or need:
```suggestion
			registry.dispatch( STORE_NAME ).receiveGetSettings( { tosAccepted: false } );
```
but let's do one of them 🙂

assets/js/modules/idea-hub/components/setup/SetupForm.js
```diff
@@ -0,0 +1,73 @@
...
+const { useSelect, useDispatch } = Data;
+
+export default function SetupForm() {
+	const { description } = useSelect( ( select ) => select( CORE_MODULES ).getModule( 'idea-hub' ) );
+	const tosAccepted = useSelect( ( select ) => select( STORE_NAME ).getTosAccepted() );
+	const { setTosAccepted } = useDispatch( STORE_NAME );
+
+	const onChange = useCallback( ( event ) => {
+		const { checked: accept } = event.target;
+		setTosAccepted( accept );
```
tofumatt, line 43: Why the renaming of `checked` to `accept`? I think
```suggestion
		const { checked } = event.target;
		setTosAccepted( checked );
```
is just as, if not more readable/scan-able. `isChecked` would be better, but not worth the renaming 😉

assets/js/modules/idea-hub/components/setup/SetupForm.js
```diff
@@ -0,0 +1,73 @@
...
+
+			<div className="googlesitekit-ideahub-setup__checkbox">
+				<Checkbox
+					name="tosAccepted"
+					id="tosAccepted"
+					value="1"
+					checked={ tosAccepted }
+					onChange={ onChange }
+				>
+					{ __( 'I agree that Idea Hub will track my usage of the feature in order to provide better idea suggestions', 'google-site-kit' ) }
```
tofumatt, line 60: Is there a separate Idea Hub privacy policy/terms of service we should make users aware of? This is a good checkbox to have, but I'm not sure what it means and it doesn't offer much insight for the user presented with this.

cc @marrrmarrr

assets/js/modules/idea-hub/components/setup/SetupForm.js
```diff
@@ -0,0 +1,73 @@
...
+				</Checkbox>
+			</div>
+
+			<Button
+				className="googlesitekit-ideahub-setup__button"
+				disabled={ ! tosAccepted }
+			>{
+					__( 'Continue', 'google-site-kit' )
+				}</Button>
+
```
tofumatt, line 70: Some odd whitespace here, no biggie 🙂 
```suggestion
			>
				{ __( 'Continue', 'google-site-kit' ) }
			</Button>

```

tofumatt (changes requested):
Looks great; just a few small tweaks and a question for @marrrmarrr about the Idea Hub "opt-in" checkbox.

assets/js/modules/idea-hub/components/setup/SetupForm.js
tofumatt: @marrrmarrr and I chatted about this and because the usage data tracked is minimal we settled on:
```suggestion
					{ __( 'I agree that Idea Hub will track my usage of the feature (e.g. creating a draft; saving/dismiss ideas) in order to provide better idea suggestions, 'google-site-kit' ) }
```

tofumatt:
@googlebot I consent.

assets/js/modules/idea-hub/components/setup/Setup.stories.js
johnPhillips: I went round a similar mental merrygoround 😄  🎠
Honestly I didn't know what the protocol was for things like this where, as you say, the text is coming pre-translated from the back end.

assets/js/modules/idea-hub/components/setup/Setup.stories.js
johnPhillips: I've defaulted it to `false` 👍

assets/js/modules/idea-hub/components/setup/SetupForm.js
johnPhillips: I just followed an example from elsewhere with a checkbox. No worries, I've removed it 👍

tofumatt (approved):
A PHPUnit test was failing because the TOS is now required, but I tweaked the test so it worked. 👍🏻

tofumatt (approved):

Issue #3655, opened 2021-06-30
Cached user surveys have incorrect TTL


eugene-manuilov:
## Bug Description

The current implementation of the `triggerSurvey` action does not set TTL when it caches the survey: https://github.com/google/site-kit-wp/blob/eea7808d8ab973561179d2d15d9fa327485aeb15/assets/js/googlesitekit/datastore/user/surveys.js#L109-L112

It means that all surveys are cached for an hour since it is the default value for the `setItem` function.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `triggerSurvey` action should pass the provided `ttl` to the `setItem` function call to cache surveys for intended period of time.

## Implementation Brief

* Add js tests to verify that the survey is cached for specified ttl.
* Merge #3656

### Test Coverage

* N/A

### Visual Regression Changes

* N/A

## QA Brief

* Check the `triggerSurvey` action and verify that it correctly passes the `ttl` parameter to the `setItem` function call.

## Changelog entry

* Add a configurable timeout to User Survey components.

aaemnnosttv:
Good catch @eugene-manuilov 👍  As it is currently, this wouldn't result in a change in the survey cooldown period because the default `ttl` for `setItem` is 1 hour as you mentioned (which is the correct value for now).

The IB looks good, but since we're not in a rush to merge this so let's also add test coverage for this detail which should be fairly straightforward. Currently the tests only check that the validation for the `ttl` value, not that `setItem` is actually called with this value.

eugene-manuilov:
@aaemnnosttv, yes we can add tests. IB is updated.

aaemnnosttv:
IB ✅

Hazlitte:
@eugene-manuilov @tofumat I have been working on the QA Eng for this.  I can see that `ttl` is now passed to the `setItem` call in the `triggerSurvey` action and the js tests are passing.

I enabled the `userFeedback` flag in my local environment and observed the state of local storage in the inspector. The user feedback modal is displayed and I select icon and close it.  But I never saw a key/value set.  Am I testing this wrong?

eugene-manuilov:
@Hazlitte you don't need to select icon and close the survey. Just open the dashboard page and when the user survey appears, check the session storage, it should contain a record for that survey with the appropriate ttl (1 hour).

Hazlitte:
Thanks @eugene-manuilov 👍

**QA ✅**

[...]

PR #3656, opened 2021-06-30
Bug/3655 survey ttl

eugene-manuilov:
## Summary

Addresses issue #3655

## Checklist


assets/js/googlesitekit/datastore/user/surveys.test.js
```diff
@@ -98,6 +98,26 @@ describe( 'core/user surveys', () => {
...
+							ttl = JSON.parse( options )?.ttl;
+						} catch {
+						}
+					},
+				} );
+
+				fetchMock.postOnce( surveyTriggerEndpoint, { body: { triggerID } } );
+
+				await registry.dispatch( STORE_NAME ).triggerSurvey( triggerID, { ttl: 500 } );
+				expect( ttl ).toBe( 500 );
```
tofumatt, line 119: Elsewhere we reset the backend storage mechanism after modifying it, see https://github.com/google/site-kit-wp/blob/93d2c1b3e902cd4d91aadf82e029d0d5fabaa830/assets/js/googlesitekit/api/index.test.js#L72 for instance.

So I think we'll want to do that here too:

```suggestion
				expect( ttl ).toBe( 500 );

				// Reset the backend storage mechanism.
				setSelectedStorageBackend( undefined );
```

tofumatt (changes requested):
I think it'd be safer to reset the storage mechanism after modifying it in the test, because it isn't reset in each test. Maybe it should be? But that's bigger than the scope of the issue, so we should at least reset it after modifying it in the test here 🙂 

Otherwise looks good!

assets/js/googlesitekit/datastore/user/surveys.test.js
eugene-manuilov: Yes, makes sense to me!

tofumatt (approved):

Issue #1147, opened 2020-02-19
WP Dashboard widget CTAs take up too much space


aaemnnosttv:
## Bug Description

For users who don't have Analytics connected (maybe they don't use it for example), the dashboard widget has two CTAs on it to activate Analytics (more space is used by CTAs than stats!). Ideally, there shouldn't be more than one of these in the widget, and given that this is not a Site Kit screen, it's a bit distracting for these to not be dismissible.

Possible alternatives:
* Remove module CTAs from dashboard widget - inner components/sections will appear based on module activation
* Limit CTAs displayed to a single dismissible CTA

## Steps to reproduce

1. Set up Site Kit, without connecting Analytics
1. Go to WordPress dashboard
1. See 2 CTAs to setup Analytics

## Screenshots

![image](https://user-images.githubusercontent.com/1621608/74848784-af2a0380-5340-11ea-9f30-5cb0d6f5f44d.png)

Error examples when search console isn't correctly setup:
From `/wp-admin/admin.php?page=googlesitekit-dashboard`:
<img width="1244" alt="115778007-d9bf8400-a36a-11eb-9df4-cc273e70cb32" src="https://user-images.githubusercontent.com/373931/116478972-cc573d80-a833-11eb-8cc1-8c712cd5ed44.png">

Another from `/wp-admin/index.php`
<img width="616" alt="115778176-1c815c00-a36b-11eb-98bf-92c8e055db96" src="https://user-images.githubusercontent.com/373931/116479052-e8f37580-a833-11eb-8442-b5aefa8a1e0d.png">

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `WPDashboardWidgets` component should be updated to only show a single CTA (activation/setup/zero-data) component for each module
    * Only CTAs of the same type should be consolidated
    * This will need to be done manually like we do in the admin bar now since we do not use (and will not be using) the widgets API in this location which handles aggregating these cases automatically. Unlike the widgets API logic, this consolidation should happen regardless of the widgets being directly adjacent to one another.
* Jest component tests should be added for each of these cases

### Determine widget states
Determine whether each Analytics and Search Console widget in `components/wp-dashboard/WPDashboardWidgets.js` is in a special state or not.

- Query the Widgets store's `getWidgetState` selector for each Search Console and Analytics widget, passing the widget slug. Determine whether each widget is in a special state. See `AdminBarWidgets` component for an example:
[...]

PR #3659, opened 2021-07-01
WP Dashboard widget CTAs take up too much space (1147).

johnPhillips:
## Summary

Addresses issue #1147

## Checklist


assets/js/components/wp-dashboard/WPDashboardWidgets.js
```diff
@@ -84,8 +111,39 @@ const WPDashboardWidgets = () => {
...
+					<WPDashboardUniqueVisitorsWidget />
+					<div className={ classnames(
+						'googlesitekit-wp-dashboard-stats__area-wrapper',
+						{
+							[ HIDDEN_CLASS ]: shouldCombineAnalyticsArea1,
+						} ) }>
+						<WPDashboardSessionDurationWidget />
+					</div>
+				</Fragment>
+			) }
```
felixarntz, line 126: You're changing here the order of the widgets it looks like. These two widgets used to be displayed on top if Analytics is active, that should not be changed. If Analytics is active, the two widgets should be on top - if Analytics is not active, the Analytics CTA should be below the Search Console widgets, so that still the actual data remains on top.

Can you move the updated clause back above where it used to be?

assets/js/components/wp-dashboard/WPDashboardWidgets.js
```diff
@@ -84,8 +111,39 @@ const WPDashboardWidgets = () => {
...
+
 			{ analyticsModuleActive && analyticsModuleConnected && (
-				<WPDashboardPopularPagesWidget />
+				<Fragment>
+					<WPDashboardUniqueVisitorsWidget />
+					<div className={ classnames(
+						'googlesitekit-wp-dashboard-stats__area-wrapper',
+						{
+							[ HIDDEN_CLASS ]: shouldCombineAnalyticsArea1,
+						} ) }>
```
felixarntz, line 122: Let's try to solve this without adding extra wrapper elements which may disrupt the layout in unexpected ways and increase the complexity of the CSS (see the extra CSS you needed to add to make this work).

It might be more appropriate to add a new prop to the individual widgets themselves, e.g. a `className` where you could pass through the `HIDDEN_CLASS`, so that the class is added on existing DOM elements rather than new wrapper elements.

felixarntz (changes requested):
This looks like a solid start, but we can probably remove some complexity from this fix by sticking closer to the original CSS and DOM structure.

assets/js/components/wp-dashboard/WPDashboardWidgets.js
johnPhillips: I was a bit daunted by the prospect of prop-drilling `className` down through the various widgets to their CTAs, but now I look at it again, you are right, I think this can be done simply. Thanks 👍

assets/js/components/wp-dashboard/WPDashboardWidgets.js
johnPhillips: Sure, sorry about that.

assets/js/components/wp-dashboard/WPDashboardWidgets.js
felixarntz: @johnPhillips Actually, I see what you're saying. The updated approach has its own problems since you need to update low-level components like `ReportZero` just for this "special case". Also it now results in some inconsistent behavior since passing a `className` to one of the widget components only affects the `ReportZero` version of that widget, which would be unexpected.

I think we need to do it yet another way:
* Instead of altering the classes of the widgets in place, let's render them in an entirely different location if they need to be hidden.
* Specifically, if any of the widgets need to be hidden, let's render a `div` container with `HIDDEN_CLASS` somewhere e.g. at the end of the overall dashboard widget and put all the widgets to hide in there, unmodified.
* This ensures the widgets still get "rendered", but without potentially disrupting the main parts of the DOM, and without making all those changes to pass through class names.

Sorry on the change of course here again, but your update made it clear another approach is needed. The one above is actually closest to how regular widgets behave.

johnPhillips: No problem @felixarntz 
Agreed, this feels like the more solid option. I've implemented your suggestion - let me know what you think 👍

assets/js/components/wp-dashboard/WPDashboardWidgets.js
```diff
@@ -84,9 +125,14 @@ const WPDashboardWidgets = () => {
...
-			{ analyticsModuleActive && analyticsModuleConnected && (
-				<WPDashboardPopularPagesWidget />
-			) }
+
+			{ analyticsModuleActive && analyticsModuleConnected && ! shouldCombineAnalyticsArea2 && <WPDashboardPopularPagesWidget /> }
+
+			<div className={ HIDDEN_CLASS }>
+				{ shouldCombineSearchConsoleWidgets && <WPDashboardClicksWidget /> }
+				{ shouldCombineAnalyticsArea1 && <WPDashboardSessionDurationWidget /> }
+				{ shouldCombineAnalyticsArea2 && <WPDashboardPopularPagesWidget /> }
```
felixarntz, line 134: I think in this line we still also need the conditions that Analytics is active and connected that the widget has above - otherwise this may render without Analytics being active/connected, e.g. causing API requests to be made that could result in errors.

assets/js/components/wp-dashboard/WPDashboardWidgets.js
```diff
@@ -84,9 +125,14 @@ const WPDashboardWidgets = () => {
 					) }
 				</div>
 			) }
-			{ analyticsModuleActive && analyticsModuleConnected && (
-				<WPDashboardPopularPagesWidget />
-			) }
+
+			{ analyticsModuleActive && analyticsModuleConnected && ! shouldCombineAnalyticsArea2 && <WPDashboardPopularPagesWidget /> }
+
+			<div className={ HIDDEN_CLASS }>
```
felixarntz, line 131: Nit-pick, but maybe we could only render that if it's needed (i.e. if at least one of the below conditions is true).

felixarntz (commented):
@johnPhillips This looks almost good to go, just two small things worth refining.

felixarntz (approved):
Excellent!

PR #3664, opened 2021-07-02
Update report usage for AdSense API v2

aaemnnosttv:
## Summary

Addresses issue #3517 (follow-up)

## Relevant technical choices

* Changed `TOTAL_EARNINGS` metric requested to `ESTIMATED_EARNINGS` due to zero values being shown otherwise
    * This may not be the case for all users but seems to be related to the ad client used, so estimated seems like the safer option available to the most users
    * When testing with the previous `TOTAL_EARNINGS` I was getting no-data states and this warning appearing in the returned report: "Some of the requested metrics are not available for some of the ad clients used by this report."
* Updates generated mock data for AdSense reports
* Updates usage of AdSense reports – the main changes are:
    * `totals` is now a Row object
    * `averages` is now a Row object
    * `rows` is now an array of Row objects
    * `currency` in header objects is now `currencyCode`
* Consuming code has been updated accordingly
* Added new utilities for validating dimensions and metrics requested against the list of known valid choices
* Updated domain filter application to be a single string (necessary for OR logic to work properly now)
* Approved visual changes in VRT due to changes in report options

## Testing / QA

Please check the following

* Widget and non-widget based dashboards
* Widget and non-widget based AdSense dashboard
* Both with only AdSense setup and connected and with Analytics connected using an account that is linked with AdSense

## Checklist


assets/js/modules/adsense/util/is-zero-report.test.js
```diff
@@ -21,6 +21,80 @@
...
+			name: 'PAGE_VIEWS_RPM',
+			type: 'METRIC_CURRENCY',
+			currencyCode: 'EUR',
+		},
+	],
+	rows: [
+		{
+			cells: [
+				{
+					value: '0.00',
```
aaemnnosttv, line 46: Here's an example of a real report that returns `0.00` for earnings when requesting `TOTAL_EARNINGS`. See the warning below. This is still non-zero report though because it has impressions.

felixarntz:
@aaemnnosttv I've pushed one extra commit https://github.com/google/site-kit-wp/pull/3664/commits/29d293ab63ebb13c3c6a46d95971292a646d2f04, which fixes a potential React error due to undefined `totals` field. Before it wasn't needed since it was just looking up a value directly within `totals`, but now that it's a nested lookup, we need to use `?.` to prevent errors.

felixarntz (approved):
Looks good to me. @wpdarren ~~Can you please test this PR against real AdSense data to ensure it works now?~~ Once merged, please test again with real AdSense data.

Issue #3272, opened 2021-05-03
Show Idea Hub context for post based on Idea Hub idea in WordPress block editor / Gutenberg


felixarntz:
The WordPress block editor (Gutenberg) should indicate if the current post is a draft based on an Idea Hub idea.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* When editing a WordPress post that has Idea Hub idea metadata set in the block editor (Gutenberg), a notice [like in this Figma file](https://www.figma.com/file/YB79eKcGOwyUjlWG8mjDZA/Prompt-Engine-%2F-Idea-Hub?node-id=289%3A2479) should be displayed on top of the editor.
    * The copy used should actually differ from the Figma design and say: _This post was created from an idea you picked in Site Kit's Idea Hub: {idea}_ The bold "Google Site Kit" prefix from the Figma design should not be included.
    * The notice should be dismissible. When dismissed, that dismissal should be cached temporarily for an hour, it doesn't need to be persistently stored. Editing a particular post doesn't happen all the time, and it's okay to remind the user again if they open the same post in the editor after a while.
    * This functionality should rely on the [Gutenberg Notice API](https://developer.wordpress.org/block-editor/reference-guides/components/notice/).
    * The entire integration should be implemented in a separate JS file that is loaded only in the block editor (from the `Idea_Hub` class in PHP), if necessary for the current post. Other than typical for Site Kit JS code, the code here should actually rely on WordPress's included versions of `@wordpress/*` dependencies (e.g. reference `wp.data` instead of `@wordpress/data`, and include `wp-data` in the dependencies for the asset in PHP) and not come with any of our own bundled dependencies, to not double-load dependencies and minimize the footprint in the block editor.

#### PHP

* Edit the `Asset` class: 
  * Add the following constants to the class:
    * `CONTEXT_ADMIN_GLOBAL = 'admin-global'`
    * `CONTEXT_ADMIN_POST_EDITOR = 'admin-post-editor'`
    * `CONTEXT_ADMIN_POSTS = 'admin-posts'`
    * `CONTEXT_ADMIN_SITEKIT = 'admin-sitekit'`
  * Add `'load_contexts' => array( self::CONTEXT_ADMIN_SITEKIT ),` to the array of default values for the `$args` property.
  * Add a new public method `has_context( $context )` that returns `TRUE` if the incoming `$context` exists in the `$this->args['load_contexts']` array.
[...]

PR #3666, opened 2021-07-02
Add Idea Hub Gutenberg notices.

tofumatt:
## Summary

Addresses issue #3272.

## Relevant technical choices

I needed to add the ability for the Idea Hub postmeta to be exposed in the REST API; I went with the `$args` approach as it was straightforward and quick, but obviously it moves the source of truth from the meta file to the instantiation of the meta. This might not be ideal, but I wasn't sure.

Let me know what you think on that point, we can refactor it to be something fancier but I didn't want to start modifying too many classes 😅 

Regarding the IB, I put everything in one listener rather than listening for the `hasNotices` in its own listener, because we need to wait for the postmeta requests to resolve anyway, which involves a listener. I think my approach is a bit less complex than the one proposed in the IB, and it accomplishes the same thing 🙂 

It might be nice to eventually store the "has viewed" meta on the server, as we _do_ have a TTL on these and it might get annoying to see the notice in different browsers/computers, but for now it's client-side only.

## Checklist

### DIFF
--- a/assets/js/googlesitekit-idea-hub-notice.js
+++ b/assets/js/googlesitekit-idea-hub-notice.js
@@ -58,22 +58,18 @@ const loadIdeaHubNotices = async ( _global = global ) => {
 
 		const { cacheHit } = await getItem( noticeKey );
 
-		if ( hasNotice( postID ) !== false ) {
+		// We've already shown this notice on a previous visit to this page in
+		// the editor, so don't show it again.
+		if ( cacheHit ) {
+			unsubscribeFromListener();
 			return;
 		}
 
 		// We've already shown this notice, so when it's hidden, mark it as shown
 		// so it doesn't appear again.
-		if ( shownNotices.includes( noticeKey ) ) {
+		if ( hasNotice( postID ) === false && shownNotices.includes( noticeKey ) ) {
 			// Don't show this notice for another 90 days.
 			setItem( noticeKey, true, { ttl: 108000 } );
-			unsubscribeFromListener();
-			return;
-		}
-
-		// We've already shown this notice on a previous visit to this page in
-		// the editor, so don't show it again.
-		if ( cacheHit ) {
 			unsubscribeFromListener();
 			return;
 		}
@@ -94,8 +90,7 @@ const loadIdeaHubNotices = async ( _global = global ) => {
 			);
 
 			// Mark the notice as shown locally, so if it is no longer marked as
-			// visible, we known it was dismissed (rather than having yet to be
-			// displayed).
+			// visible, we know it was dismissed.
 			shownNotices.push( noticeKey );
 		}
 	};
--- a/assets/js/googlesitekit-idea-hub-notice.js
+++ b/assets/js/googlesitekit-idea-hub-notice.js
@@ -36,13 +36,13 @@ const loadIdeaHubNotices = async ( _global = global ) => {
 	const { wp } = _global;
 
 	const hasNotice = ( postID ) => {
-		if ( wp.data.select( 'core/notices' ).getNotices() === undefined ) {
+		const notices = wp.data.select( 'core/notices' ).getNotices();
+		if ( notices === undefined ) {
 			return undefined;
 		}
 
-		return wp.data.select( 'core/notices' ).getNotices().some( ( notice ) => {
-			return notice.id === editorNoticeKey( postID );
-		} );
+		const key = editorNoticeKey( postID );
+		return notices.some( ( { id } ) => id === key );
 	};
 
 	const listener = async () => {
@@ -55,7 +55,9 @@ const loadIdeaHubNotices = async ( _global = global ) => {
 
 		const { cacheHit } = await getItem( editorNoticeKey( postID ) );
 
-		if ( hasNotice( postID ) === false ) {
+		if ( hasNotice( postID ) !== false ) {
+			return;
+		}
 			// We've already shown this notice, so when it's hidden, mark it as shown
 			// so it doesn't appear again.
 			if ( shownNotices.includes( editorNoticeKey( postID ) ) ) {
--- a/assets/js/googlesitekit-idea-hub-notice.js
+++ b/assets/js/googlesitekit-idea-hub-notice.js
@@ -26,23 +26,24 @@ import { __, sprintf } from '@wordpress/i18n';
  */
 import { getItem, setItem } from './googlesitekit/api/cache';
 
-const shownNotices = [];
-
 const editorNoticeKey = ( postID ) => {
 	return `modules::idea-hub::dismissed-editor-notice-${ postID }`;
 };
 
 const loadIdeaHubNotices = async ( _global = global ) => {
 	const { wp } = _global;
+	const shownNotices = [];
 
 	const hasNotice = ( postID ) => {
 		const notices = wp.data.select( 'core/notices' ).getNotices();
+
 		if ( notices === undefined ) {
 			return undefined;
 		}
 
-		const key = editorNoticeKey( postID );
-		return notices.some( ( { id } ) => id === key );
+		const noticeKey = editorNoticeKey( postID );
+
+		return notices.some( ( { id } ) => id === noticeKey );
 	};
 
 	const listener = async () => {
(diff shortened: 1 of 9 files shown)


includes/Modules/Idea_Hub.php
```diff
@@ -121,7 +121,7 @@ function( $maybe_empty, $postarr ) {
 		$this->post_name_setting = new Post_Idea_Name( $post_meta );
 		$this->post_name_setting->register();
 
-		$this->post_text_setting = new Post_Idea_Text( $post_meta );
+		$this->post_text_setting = new Post_Idea_Text( $post_meta, array( 'show_in_rest' => true ) );
```
tofumatt, line 124: This is what I meant by moving the source of truth. A simple `bool` for `show_in_rest` isn't a good idea either, because complex values (like JSON arrays, etc.) require a transform to handle how they should be displayed. We could add that as some extra `Post_Meta` metadata, I suppose.

includes/Core/Assets/Assets.php
```diff
@@ -80,7 +80,8 @@ public function __construct( Context $context ) {
 	/**
 	 * Registers functionality through WordPress hooks.
 	 *
-	 * @since 1.0.0
+	 * @since 1.0.0 Function introduced.
```
eugene-manuilov, line 83: No need to add a dummy comment for the initial version of the method. Just the version number is enough:

```suggestion
	 * @since 1.0.0
```

includes/Core/Modules/Module_With_Assets.php
```diff
@@ -33,7 +33,10 @@ public function get_assets();
 	/**
 	 * Enqueues all assets necessary for the module.
 	 *
-	 * @since 1.7.0
+	 * @since 1.7.0 Function introduced.
```
eugene-manuilov, line 36: Same here:
```suggestion
	 * @since 1.7.0
```

includes/Core/Modules/Module_With_Assets_Trait.php
```diff
@@ -50,15 +50,21 @@ public function get_assets() {
 	 * This default implementation simply enqueues all assets that the module
 	 * has registered.
 	 *
-	 * @since 1.7.0
+	 * @since 1.7.0 Function introduced.
```
eugene-manuilov, line 53: And here:
```suggestion
	 * @since 1.7.0
```

includes/Core/Storage/Post_Meta_Setting.php
```diff
@@ -33,15 +33,26 @@ abstract class Post_Meta_Setting {
...
 	 *
-	 * @since 1.33.0
+	 * @since 1.33.0 Function introduced.
+	 * @since n.e.x.t Added the $args option.
 	 *
 	 * @param Post_Meta_Interface $post_meta Post_Meta_Interface instance.
+	 * @param array               $args      Options to pass to `register_meta`.
 	 */
-	public function __construct( Post_Meta_Interface $post_meta ) {
+	public function __construct( Post_Meta_Interface $post_meta, $args = array() ) {
```
eugene-manuilov, line 53: @tofumatt, although this approach works, we should better solve the problem with the `show_in_rest` option using the OOP approach. What we need to do is the following:

1. Remove the new `$args` approach.
1. Add a new protected method `get_show_in_rest` that returns `false` by default.
1. Update the `register_meta` function call to be as the following:
    ```php
    register_meta(
        'post',
        static::META_KEY,
        array(
            'type'              => $this->get_type(),
            'sanitize_callback' => $this->get_sanitize_callback(),
            'single'            => true,
            'show_in_rest'      => $this->get_show_in_rest(),
        )
    );
    ```
1. Add the new method `get_show_in_rest` to the `Post_Idea_Text` class that returns `true`.

includes/Modules/Idea_Hub.php
```diff
@@ -500,6 +500,18 @@ protected function setup_assets() {
...
 			),
+			new Script(
+				'googlesitekit-idea-hub-notice',
+				array(
+					'src'           => $base_url . 'js/googlesitekit-idea-hub-notice.js',
+					'dependencies'  => array(
+						'googlesitekit-vendor',
+						'googlesitekit-api',
+						'googlesitekit-data',
+					),
```
eugene-manuilov, line 511: Seems like we don't need these dependencies for this script, do we? Let's remove them then.

assets/js/googlesitekit-idea-hub-notice.js
```diff
@@ -0,0 +1,101 @@
...
+		// eslint-disable-next-line sitekit/acronym-case
+		const postID = wp.data.select( 'core/editor' ).getCurrentPostId();
+
+		if ( ! postID ) {
+			return;
+		}
+
+		const { cacheHit } = await getItem( editorNoticeKey( postID ) );
+
+		if ( hasNotice( postID ) === false ) {
```
eugene-manuilov, line 58: Let's return early if this condition hasn't met:
```suggestion
		if ( hasNotice( postID ) !== false ) {
			return;
		}
```

assets/js/googlesitekit-idea-hub-notice.js
```diff
@@ -0,0 +1,101 @@
...
+
+	const hasNotice = ( postID ) => {
+		if ( wp.data.select( 'core/notices' ).getNotices() === undefined ) {
+			return undefined;
+		}
+
+		return wp.data.select( 'core/notices' ).getNotices().some( ( notice ) => {
+			return notice.id === editorNoticeKey( postID );
+		} );
+	};
```
eugene-manuilov, line 46:
```suggestion
	const hasNotice = ( postID ) => {
		const notices = wp.data.select( 'core/notices' ).getNotices();
		if ( notices === undefined ) {
			return undefined;
		}

		const key = editorNoticeKey( postID );
		return notices.some( ( { id } ) => id === key );
	};
```

assets/js/googlesitekit-idea-hub-notice.js
```diff
@@ -0,0 +1,101 @@
...
+
+	const listener = async () => {
+		// eslint-disable-next-line sitekit/acronym-case
+		const postID = wp.data.select( 'core/editor' ).getCurrentPostId();
+
+		if ( ! postID ) {
+			return;
+		}
+
+		const { cacheHit } = await getItem( editorNoticeKey( postID ) );
```
eugene-manuilov, line 56: The `editorNoticeKey( postID )` function call repeats four times. Please, assign it to a variable and use the variable instead of the same function calls.

assets/js/googlesitekit-idea-hub-notice.js
```diff
@@ -0,0 +1,101 @@
...
+ * WordPress dependencies
+ */
+import { __, sprintf } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import { getItem, setItem } from './googlesitekit/api/cache';
+
+const shownNotices = [];
```
eugene-manuilov, line 29: This variable belongs to the `loadIdeaHubNotices` function context. Let's move it into that function then.

eugene-manuilov (changes requested):
Thanks, @tofumatt. This is a good start. However, there are a few things (+ some nitpicks) that we need to address before we can merge this PR.

includes/Core/Storage/Post_Meta_Setting.php
tofumatt: Right on, that's nicer 🙂

assets/js/googlesitekit-idea-hub-notice.js
```diff
@@ -0,0 +1,109 @@
...
+		if ( notices === undefined ) {
+			return undefined;
+		}
+
+		const noticeKey = editorNoticeKey( postID );
+
+		return notices.some( ( { id } ) => id === noticeKey );
+	};
+
+	const listener = async () => {
```
eugene-manuilov, line 49: This function is a little bit of a mess. Please, update it to have the following flow:

1. Get postID
    * If postID is falsy, return
1. Get the item from cache
    * If cacheHit is truthy, unsubscribe and return
1. Get the hasNotice result
    * If it returns `false` and `shownNotices` contains `noticeKey`, then call `setItem`, unsubscribe and return
1. Get post meta
    * If idea_text is not empty, show the notice

eugene-manuilov (changes requested):
Thanks, @tofumatt. Added two more comments. I believe it will be good to go once we address the remaining issues.

assets/js/googlesitekit-idea-hub-notice.js
tofumatt: Thanks, that's much better! 👍🏻

assets/js/googlesitekit-idea-hub-notice.js
```diff
@@ -0,0 +1,101 @@
...
+					__( 'This post was created from an idea you picked in Site Kit’s Idea Hub: %s', 'google-site-kit' ),
+					postMeta.googlesitekitpersistent_idea_text
+				),
+				{ id: noticeKey }
+			);
+
+			// Mark the notice as shown locally, so if it is no longer marked as
+			// visible, we know it was dismissed.
+			shownNotices.push( noticeKey );
+		}
```
eugene-manuilov, line 95: @tofumatt, we need to make sure that we create a new notice only if the `shownNotices` does not include `noticeKey`.

eugene-manuilov (changes requested):
Found one more issue. Matthew, could you please take a look?

assets/js/googlesitekit-idea-hub-notice.js
tofumatt: 🤦🏻 Whoops, my bad, missed it in the refactor. Thanks for catching that!

tofumatt: Hmm, actually, I think it'd be rare that `shownNotices.includes( noticeKey )` happens here because the above `if ( hasNotice( postID ) === false && shownNotices.includes( noticeKey ) )` will return early.

But just in case I've added ` && ! shownNotices.includes( noticeKey )` to this conditional.

eugene-manuilov (approved):
Awesome! Looks good to me!

Issue #3359, opened 2021-05-13
Display Idea Hub notice above WP posts list table if there are new or saved ideas


felixarntz:
If there are any new or saved ideas in the Idea Hub module, a corresponding notice that links to the respective area of the Idea Hub widget should be displayed above the WordPress posts list admin table, following the [Figma design](https://www.figma.com/file/YB79eKcGOwyUjlWG8mjDZA/Prompt-Engine-%2F-Idea-Hub?node-id=20332%3A6816).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* If the `ideaHubModule` feature flag is enabled and the Idea Hub module is active, a WordPress admin notice should be potentially displayed above the WordPress posts list table:
    * If there are saved ideas available, the [notice should look like in this Figma file](https://www.figma.com/file/YB79eKcGOwyUjlWG8mjDZA/Prompt-Engine-%2F-Idea-Hub?node-id=20332%3A6816) except not to include G Site Kit logo.
    * If there are no saved but new ideas available, the [notice should look like in this Figma file](https://www.figma.com/file/YB79eKcGOwyUjlWG8mjDZA/Prompt-Engine-%2F-Idea-Hub?node-id=20332%3A6868) except not to include G Site Kit logo.
    * In both cases, the link should lead to the respective tab of the Idea Hub widget in the Site Kit dashboard.
    * The results for whether there are new ideas or saved ideas should be cached server-side (e.g. in a transient cached for a day).
* The notice dismissals for the two scenarios should be separately stored, both on the server-side:
    * The dismissal for the "New ideas" notice should be stored for 1 week. In other words, if the user after one week still has no saved ideas, the notice should re-appear.
    * The dismissal for the "Saved ideas" notice should be permanent, in general. However, there is one particular circumstance when it should be cleared: If the user makes a change that results in them to _no longer_ have any saved ideas (for example unsaving the only saved idea or creating a draft post from the only saved idea), the dismissal should be cleared. In other words, once the user later saves something again, the notice should re-appear.
    * To summarize: As long as the user continuously has something in "Saved", they will not see any notice again once they dismiss it.
    * Technically, this requires a new REST data point, e.g. `POST:core/user/data/dismiss-notification`, expecting a notification slug.

## Implementation Brief
- Once #3272 and #3360 have been merged.
- Create `assets/js/googlesitekit-idea-hub-post-list-notice.js`.
[...]

PR #3668, opened 2021-07-05
Feature/3359 idea hub notice

ivankruchkoff:
## Summary

Addresses issue #3359

## Relevant technical choices
Added expiration time to `dismissItem` action to accommodate for the required behaviour.

## Checklist


includes/Core/Assets/Assets.php
```diff
@@ -124,21 +124,30 @@ function() {
 			}
 		);
 
-		add_action(
-			'enqueue_block_editor_assets',
-			function() {
-				$assets = $this->get_assets();
+		// A map of asset contexts to load for specific hooks.
+		$actions_contexts_map = array(
+			'admin_print_scripts-edit.php' => Asset::CONTEXT_ADMIN_POSTS,
```
eugene-manuilov, line 129: We need to make sure that we load assets with the `CONTEXT_ADMIN_POSTS` context only on the posts list table (`$GLOBALS['typenow'] === 'post'`) and not for post list tables of other custom post types.

includes/Modules/Idea_Hub.php
```diff
@@ -128,6 +159,96 @@ function( $maybe_empty, $postarr ) {
...
+	 * Shows admin notification for idea hub ideas on post list screen.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $notices Array of admin notices.
+	 *
+	 * @return array Array of admin notices.
+	 */
+	private function googlesitekit_admin_notices( $notices ) {
+		if ( 'edit-post' !== get_current_screen()->id ) {
```
eugene-manuilov, line 172: Same here, we need to make sure that the current post type is `post` (`$GLOBALS['typenow'] === 'post'`)

includes/Modules/Idea_Hub.php
```diff
@@ -128,6 +159,96 @@ function( $maybe_empty, $postarr ) {
...
 	}
 
+	/**
+	 * Shows admin notification for idea hub ideas on post list screen.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $notices Array of admin notices.
+	 *
+	 * @return array Array of admin notices.
```
eugene-manuilov, line 169:
```suggestion
	 * @param array $notices Array of admin notices.
	 * @return array Array of admin notices.
```

includes/Modules/Idea_Hub.php
```diff
@@ -128,6 +159,96 @@ function( $maybe_empty, $postarr ) {
...
+
+		if ( false === $ideas ) {
+			$ideas = array(
+				'saved-ideas' => $this->get_data( 'saved-ideas' ),
+				'new-ideas'   => $this->get_data( 'new-ideas' ),
+			);
+
+			$this->transients->set( self::TRANSIENT_IDEAS, $ideas, DAY_IN_SECONDS );
+		}
+		$dismissed_items = array();
```
eugene-manuilov, line 200: Looks like this is not needed, right? 
```suggestion
```

assets/js/googlesitekit-idea-hub-post-list-notice.js
```diff
@@ -0,0 +1,65 @@
...
+ */
+import domReady from '@wordpress/dom-ready';
+// import { render } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import Data from 'googlesitekit-data';
+const { dispatch } = Data;
+import { CORE_USER } from './googlesitekit/datastore/user/constants';
```
eugene-manuilov, line 35:
```suggestion
import { CORE_USER } from './googlesitekit/datastore/user/constants';
const { dispatch } = Data;
```

includes/Modules/Idea_Hub.php
```diff
@@ -77,7 +101,9 @@ final class Idea_Hub extends Module
 	 * @since 1.32.0
 	 */
 	public function register() {
-		$post_meta = new Post_Meta();
+		$post_meta             = new Post_Meta();
+		$this->dismissed_items = new Dismissed_Items( $this->user_options );
+		$this->transients      = new Transients( $this->context );
```
eugene-manuilov, line 106: Looks like these instances are specific only for the `googlesitekit_admin_notices` method. In this case, let's move them into the method itself and make them be local variables for the method.

includes/Modules/Idea_Hub.php
```diff
@@ -128,6 +159,96 @@ function( $maybe_empty, $postarr ) {
...
+	/**
+	 * Shows admin notification for idea hub ideas on post list screen.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $notices Array of admin notices.
+	 *
+	 * @return array Array of admin notices.
+	 */
+	private function googlesitekit_admin_notices( $notices ) {
```
eugene-manuilov, line 171: Let's rename this method to be more specific regarding what it does.

includes/Modules/Idea_Hub.php
```diff
@@ -128,6 +159,96 @@ function( $maybe_empty, $postarr ) {
...
+		}
+		$dismissed_items = array();
+
+		foreach ( $ideas as $idea_type => &$all_ideas ) {
+			foreach ( $all_ideas as $k => $idea ) {
+				if ( in_array( $idea['name'], $dismissed_items, true ) ) {
+					unset( $all_ideas[ $k ] );
+				}
+			}
+		}
```
eugene-manuilov, line 208: I am not sure that I can follow the logic here. How come a saved or a new idea can be added to dismissed items list?

eugene-manuilov (changes requested):
Thanks, @ivankruchkoff. This looks like a good start. I have added some comments that we need to address. Please, take a look at it.

assets/js/googlesitekit-idea-hub-post-list-notice.js
```diff
@@ -0,0 +1,65 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * External dependencies
+ */
+// import './modules';
+
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/c01c856303de52f1ca08f567e329773f3b189482

assets/js/googlesitekit-idea-hub-post-list-notice.js
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * WordPress dependencies
+ */
+import domReady from '@wordpress/dom-ready';
+// import { render } from '@wordpress/element';
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/c01c856303de52f1ca08f567e329773f3b189482

assets/js/googlesitekit-idea-hub-post-list-notice.js
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/c01c856303de52f1ca08f567e329773f3b189482

assets/js/googlesitekit-idea-hub-post-list-notice.js
```diff
@@ -0,0 +1,65 @@
...
+	} else if ( savedNotice ) {
+		type = 'saved';
+		notice = savedNotice;
+	} else {
+		return;
+	}
+
+	// Button pops up only after the timeout passes
+	setTimeout( () => {
+		const button = notice.getElementsByClassName( 'notice-dismiss' )[ 0 ];
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/f172dfadcb02f6e21e42f2a385c9da418f09c6f2

assets/js/googlesitekit/datastore/user/dismissed-items.js
```diff
@@ -48,11 +48,14 @@ const fetchGetDismissedItemsStore = createFetchStore( {
 
 const fetchDismissItemStore = createFetchStore( {
 	baseName: 'dismissItem',
-	controlCallback: ( { slug } ) => API.set( 'core', 'user', 'dismiss-item', { slug } ),
+	controlCallback: ( { slug, expiresInSeconds } ) => API.set( 'core', 'user', 'dismiss-item', { slug, expiresInSeconds } ),
 	reducerCallback,
-	argsToParams: ( slug ) => ( { slug } ),
-	validateParams: ( { slug } = {} ) => {
+	argsToParams: ( { slug, expiresInSeconds = 0 } ) => {
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/f0e13258617cc1d9b85d65c16d4831aa822c4e2d

includes/Core/Assets/Assets.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/08c3d3a8da6c25bac7a7a0b4ec579fd05e30d7a7

includes/Modules/Idea_Hub.php
ivankruchkoff: In the IB:

```
The notice dismissals for the two scenarios should be separately stored, both on the server-side:
The dismissal for the "New ideas" notice should be stored for 1 week. In other words, if the user after one week still has no saved ideas, the notice should re-appear.
The dismissal for the "Saved ideas" notice should be permanent, in general. However, there is one particular circumstance when it should be cleared: If the user makes a change that results in them to no longer have any saved ideas (for example unsaving the only saved idea or creating a draft post from the only saved idea), the dismissal should be cleared. In other words, once the user later saves something again, the notice should re-appear.
```

So in this case we are dismissing the notice itself and preventing it from being shown on a subsequent load.

includes/Modules/Idea_Hub.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/918df4e4ad848360545de51b12ac4167db594c2b

includes/Modules/Idea_Hub.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/918df4e4ad848360545de51b12ac4167db594c2b

includes/Modules/Idea_Hub.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/918df4e4ad848360545de51b12ac4167db594c2b

includes/Modules/Idea_Hub.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/918df4e4ad848360545de51b12ac4167db594c2b

includes/Modules/Idea_Hub.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/918df4e4ad848360545de51b12ac4167db594c2b

includes/Modules/Idea_Hub.php
eugene-manuilov: Yes, I saw it in the IB/AC. The thing that confuses me is why do we need to check the following:

```php
$dismissed_items = $this->dismissed_items->get_dismissed_items();

...

if ( in_array( $idea['name'], $dismissed_items, true ) ) {
```

We don't have anything in the codebase that lets you as a user dismiss a particular idea using our custom dismissed items functionality. An idea may be dismissed, but the information about it is stored on the Idea Hub service, not in the WordPress database.

includes/Modules/Idea_Hub.php
```diff
@@ -128,6 +157,96 @@ function( $maybe_empty, $postarr ) {
...
 
+	/**
+	 * Shows admin notification for idea hub ideas on post list screen.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $notices Array of admin notices.
+	 * @return array Array of admin notices.
+	 */
+	private function admin_notice_idea_hub_ideas( $notices ) {
```
eugene-manuilov, line 168: I think the logic in this method needs to be reworked a little bit to leverage our Notice API. We need to do it like this:

```php
private function admin_notice_idea_hub_ideas( $notices ) {
    $notices[] = new Notice(
        'saved-ideas',
        array(
            'type'            => Notice::TYPE_INFO,
            'dismissible' => true,
            'content'       => function() {
                ob_start();
                // render notice for saved ideas
                return ob_get_clean();
            },
            'active_callback' => function() {
                global $post_type;
                
                $current_screen = get_current_screen();
                if ( is_null( $current_screen ) || 'edit-post' !== $current_screen->id || 'post' !== $post_type ) {
                    return false;
                }

                // pull saved ideas and return true if we have new saved ideas, otherwise, return false
            }
        )
    );


    $notices[] = new Notice(
        'new-ideas',
        array(
            // simmilar approach but for new ideas
        )
    );

    return $notices;
}
```

eugene-manuilov (changes requested):
Thanks, @ivankruchkoff. Added more comments. Please, take a look.

includes/Modules/Idea_Hub.php
ivankruchkoff: Perhaps I misunderstood the IB, I thought individual ideas could be dismissed too, so that we need need to make sure that there were ideas that were not dismissed prior to rendering the notice.

I based that on the following, but on re-read it talks only about whether the notices are dismissed not each idea.

```
Check if there are saved ideas by using the get_data method of the class with saved-ideas as parameter and whether the notice has not been dismissed by checking if the slug for the saved idea notice does not exist in the list of get_dismissed_items.
```

assets/js/googlesitekit-idea-hub-post-list-notice.js
```diff
@@ -0,0 +1,59 @@
...
+	} else {
+		return;
+	}
+
+	// Button pops up only after the timeout passes
+	setTimeout( () => {
+		const button = notice.querySelector( '.notice-dismiss' );
+		const slug = type === 'new' ? 'new-ideas' : 'saved-ideas';
+		const expirationType = type === 'new' ? WEEK_IN_SECONDS : 0;
+		button.addEventListener( 'click', () => dismissNotice( slug, expirationType ) );
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/c24aa017e0b12d34fc10d6fcaf5db35514cba3a9

includes/Modules/Idea_Hub.php
ivankruchkoff: I think we're talking about different ways of implementing the same thing. Our end result is at most one new notice for saved/new ideas using the Notice API

eugene-manuilov: Ah... yes, you are right. Please, disregard this comment.

includes/Core/Assets/Assets.php
```diff
@@ -124,21 +124,34 @@ function() {
...
+			'enqueue_block_editor_assets'  => Asset::CONTEXT_ADMIN_POST_EDITOR,
+		);
 
-				array_walk(
-					$assets,
-					function( $asset ) {
-						if ( $asset->has_context( Asset::CONTEXT_ADMIN_POST_EDITOR ) ) {
-							$this->enqueue_asset( $asset->get_handle() );
-						}
+		foreach ( $actions_contexts_map as $hook => $context ) {
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/0eb18faa79bab6947559fc677e52d1ebcaa624da

includes/Modules/Idea_Hub.php
```diff
@@ -128,6 +157,96 @@ function( $maybe_empty, $postarr ) {
...
+		if ( 'edit-post' !== get_current_screen()->id || 'post' !== $GLOBALS['post_type'] ) {
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/428b52d4ef0848db7d8d4385b10726711c2d2ca6

includes/Modules/Idea_Hub.php
```diff
@@ -128,6 +157,96 @@ function( $maybe_empty, $postarr ) {
...
+			return $notices;
+		}
+		$this->dismissed_items = new Dismissed_Items( $this->user_options );
+		$this->transients      = new Transients( $this->context );
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/428b52d4ef0848db7d8d4385b10726711c2d2ca6

includes/Modules/Idea_Hub.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/4e43c593495304cdce3ca1c8f2f3482608bf91d4

includes/Modules/Idea_Hub.php
```diff
@@ -146,12 +156,102 @@ function( $maybe_empty, $postarr ) {
...
+			$transients->set( self::TRANSIENT_IDEAS, $ideas, DAY_IN_SECONDS );
+		}
+
+		$has_saved_ideas = count( $ideas['saved-ideas'] ) > 0;
+		$has_new_ideas   = count( $ideas['new-ideas'] ) > 0;
+
+		if ( ! $has_saved_ideas && in_array( 'saved-ideas', $dismissed_items, true ) ) {
+			// Saved items no longer need to be dismissed as there are none currently.
+			$this->dismissed_items->delete( 'saved-ideas' );
+		}
```
eugene-manuilov, line 219: > ... there is one particular circumstance when it should be cleared: If the user makes a change that results in them to no longer have any saved ideas (for example unsaving the only saved idea or creating a draft post from the only saved idea), the dismissal should be cleared. In other words, once the user later saves something again, the notice should re-appear.

We should clear the `saved-ideas` dismissal in the appropriate endpoint handler when a user saves or removes an idea from the saved ideas list.

eugene-manuilov (changes requested):
Added a few more comments.

aaemnnosttv (changes requested):
I was also assigned on the issue so I figured my review was also requested – apologies for the delay, but just a few things in addition to @eugene-manuilov 's comments.

assets/js/googlesitekit-idea-hub-post-list-notice.js
```diff
@@ -0,0 +1,59 @@
...
+	setTimeout( () => {
+		const button = notice.querySelector( '.notice-dismiss' );
+		if ( ! button ) {
+			return;
+		}
+		const slug = type === 'new' ? 'new-ideas' : 'saved-ideas';
+		const expirationType = type === 'new' ? WEEK_IN_SECONDS : 0;
+		button.addEventListener( 'click', () => dismissNotice( slug, expirationType ) );
+	}, 1 );
+} );
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/d384a3e92762d77f1095316cd2a958697896de57

includes/Core/Assets/Assets.php
```diff
@@ -482,6 +503,7 @@ private function get_assets() {
 						'googlesitekit-data',
 						'googlesitekit-api',
 						'googlesitekit-user-data',
+						'googlesitekit-runtime',
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/92c2114d7f75cf040ab92a8305741cb5072a86bb

includes/Core/Assets/Assets.php
```diff
@@ -124,6 +124,27 @@ function() {
...
+				global $post_type;
+				if ( 'post' !== $post_type ) {
+					// For CONTEXT_ADMIN_POSTS we only load scripts for the 'post' post type.
+					return;
+				}
+				$assets = $this->get_assets();
+
+				array_walk(
+					$assets,
+					function( $asset ) {
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/92c2114d7f75cf040ab92a8305741cb5072a86bb

includes/Modules/Idea_Hub.php
```diff
@@ -146,12 +156,102 @@ function( $maybe_empty, $postarr ) {
...
+			return $notices;
+		}
+		$dismissed_items_instance = new Dismissed_Items( $this->user_options );
+		$transients               = new Transients( $this->context );
+
+		$notice_settings = array(
+			'saved' => array(
+				'text' => esc_html__( 'Need some inspiration? Revisit your saved ideas in Site Kit', 'google-site-kit' ),
+				'cta'  => esc_html__( 'See saved ideas', 'google-site-kit' ),
+				'link' => admin_url( 'admin.php?page=googlesitekit-dashboard#new-ideas' ),
```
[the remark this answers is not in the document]
ivankruchkoff: Updated in https://github.com/google/site-kit-wp/pull/3668/commits/b71c7b9f1332a49c8a232a447cc4b3a3775f7ee9

includes/Modules/Idea_Hub.php
```diff
@@ -146,12 +156,102 @@ function( $maybe_empty, $postarr ) {
...
+			),
+			'new'   => array(
+				'text' => esc_html__( 'Need some inspiration? Here are some new ideas from Site Kit’s Idea Hub', 'google-site-kit' ),
+				'cta'  => esc_html__( 'See new ideas', 'google-site-kit' ),
+				'link' => admin_url( 'admin.php?page=googlesitekit-dashboard#saved-ideas' ),
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/b71c7b9f1332a49c8a232a447cc4b3a3775f7ee9

includes/Modules/Idea_Hub.php
ivankruchkoff: Updated in https://github.com/google/site-kit-wp/pull/3668/commits/faadd204be65bbda730317644d3b527de07d8626

includes/Modules/Idea_Hub.php
```diff
@@ -146,12 +156,102 @@ function( $maybe_empty, $postarr ) {
...
+	 * @param array $notices Array of admin notices.
+	 * @return array Array of admin notices.
+	 */
+	private function admin_notice_idea_hub_ideas( $notices ) {
+		global $post_type;
+		$current_screen = get_current_screen();
+		if ( is_null( $current_screen ) || 'edit-post' !== $current_screen->id || 'post' !== $post_type ) {
+			return $notices;
+		}
+		$dismissed_items_instance = new Dismissed_Items( $this->user_options );
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/6cad7e8e04426c532fed2463773f94a17d4590a6

includes/Modules/Idea_Hub.php
```diff
@@ -146,12 +156,102 @@ function( $maybe_empty, $postarr ) {
...
+		$ideas           = $transients->get( self::TRANSIENT_IDEAS );
+
+		if ( false === $ideas ) {
+			$ideas = array(
+				'saved-ideas' => $this->get_data( 'saved-ideas' ),
+				'new-ideas'   => $this->get_data( 'new-ideas' ),
+			);
+
+			$transients->set( self::TRANSIENT_IDEAS, $ideas, DAY_IN_SECONDS );
+		}
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/85933e21f3cf5c49a859db1810cd1778e3aa707d

webpack.config.js
```diff
@@ -285,6 +285,8 @@ const webpackConfig = ( env, argv ) => {
 				'googlesitekit-i18n': './assets/js/googlesitekit-i18n.js',
 				// Analytics advanced tracking script to be injected in the frontend.
 				'analytics-advanced-tracking': './assets/js/analytics-advanced-tracking.js',
+				// Idea Hub Post List notice.
+				'googlesitekit-idea-hub-post-list-notice': './assets/js/googlesitekit-idea-hub-post-list-notice.js',
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/19b3b594bff9d7458ca455e07bf21ae5ec2ec13d

includes/Modules/Idea_Hub.php
ivankruchkoff: I think this might have to be done later, and I can remove this portion entirely:
saving / unsaving an idea would be via `POST:update-idea-state` which is currently pending.

Creating a draft post from the only saved idea, I think this may need a call to `POST:create-idea-draft-post` but that's insufficient to remove the idea from the saved ideas.

Thoughts here?

assets/js/googlesitekit/datastore/user/dismissed-items.js
```diff
@@ -66,15 +69,17 @@ const baseActions = {
...
 	 * @since 1.37.0
 	 *
-	 * @param {string} slug Item slug to dismiss.
+	 * @param {string} slug             Item slug to dismiss.
+	 * @param {number} expiresInSeconds An integer number of seconds for expiry. 0 denotes permanent dismissal.
 	 * @return {Object} Generator instance.
 	 */
 	dismissItem: createValidatedAction(
-		( slug ) => {
+		( slug, expiresInSeconds = 0 ) => {
```
aaemnnosttv, line 77: Let's make the second argument an (optional) object of `options`. Simply passing an integer as the second argument won't be obvious what it is when reading this later where it's used. We have many functions that use an object of options/settings so this would be a bit more consistent with our conventions as well. Don't forget that optional parameters should be documented with the respective `@param {type} [name] Optional. Description...` – or in this case `[options.expiresInSeconds]` 🙂

includes/Modules/Idea_Hub.php
```diff
@@ -146,12 +161,123 @@ function( $maybe_empty, $postarr ) {
...
+		global $post_type;
+		$current_screen = get_current_screen();
+		if ( is_null( $current_screen ) || 'edit-post' !== $current_screen->id || 'post' !== $post_type ) {
+			return $notices;
+		}
+		$transients               = new Transients( $this->context );
+		$dismissed_items_instance = new Dismissed_Items( $this->user_options );
+		if ( $dismissed_items_instance->is_dismissed( 'saved-ideas' ) && $dismissed_items_instance->is_dismissed( 'new-ideas' ) ) {
+			return $notices;
+		}
```
aaemnnosttv, line 194: Please move these calls to the `active_callback` for the respective notices.

includes/Modules/Idea_Hub.php
```diff
@@ -146,12 +161,123 @@ function( $maybe_empty, $postarr ) {
...
+		$dismissed_items = $dismissed_items_instance->get_dismissed_items();
```
aaemnnosttv, line 195: Let's rely on the `is_dismissed` method rather than `in_array`.

includes/Modules/Idea_Hub.php
```diff
@@ -146,12 +161,123 @@ function( $maybe_empty, $postarr ) {
...
+		$saved_ideas     = $transients->get( self::TRANSIENT_SAVED_IDEAS );
+		$new_ideas       = $transients->get( self::TRANSIENT_NEW_IDEAS );
```
aaemnnosttv, line 197: These can move into the callbacks as well. I realize there will be some additional overhead as a result, but it will only hit the storage backend once.

includes/Modules/Idea_Hub.php
```diff
@@ -146,12 +161,123 @@ function( $maybe_empty, $postarr ) {
...
+
+		$notices[] = new Notice(
+			'saved',
```
aaemnnosttv, line 200: The slug should be unique across all notices, so we should at least make it specific to the module. Similarly, we should probably use the same slug for the notice as we use for the dismissed item. How about `idea-hub_saved-ideas` and `idea-hub_new-ideas` ?

assets/js/googlesitekit-idea-hub-post-list-notice.js
```diff
@@ -0,0 +1,57 @@
...
+ */
+import Data from 'googlesitekit-data';
+import { CORE_USER } from './googlesitekit/datastore/user/constants';
+const { dispatch } = Data;
+
+const WEEK_IN_SECONDS = 3600 * 24 * 7;
+
+domReady( () => {
+	const newNotice = document.getElementById( 'googlesitekit-notice-new' );
+	const savedNotice = document.getElementById( 'googlesitekit-notice-saved' );
```
aaemnnosttv, line 35: Per my last comment about notice slugs, these IDs would need to be updated accordingly.

Also, I think we could simplify the logic here by using a combined selector:

```js
document.querySelector( '[id="googlesitekit-notice-A"],[id="googlesitekit-notice-B"]' )`
```

Then the slug can be parsed from the `el.id` that comes back by stripping off the leading `googlesitekit-notice-`, although I wouldn't be against adding the notice slug as a data attribute to `Notice`.

aaemnnosttv (commented):
Thanks @ivankruchkoff – a few final thoughts and requests here from me, otherwise this is looking really close 👍

includes/Modules/Idea_Hub.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/f96128adcfde0ffcb4f0e5f6271b129652121b8e

includes/Modules/Idea_Hub.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/c09d94eb2f3b03e8aafd02ec81e16392edcb6c05

includes/Modules/Idea_Hub.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/88d332f5f9233b223baacdddf9f131f6e79c52c1

includes/Modules/Idea_Hub.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/20feaddb833e00552af18ac924d9622784926b59

assets/js/googlesitekit/datastore/user/dismissed-items.js
```diff
@@ -66,15 +69,17 @@ const baseActions = {
...
 			invariant( slug, 'A tour slug is required to dismiss a tour.' );
+			invariant( Number.isInteger( expiresInSeconds ), 'expiresInSeconds should be an integer.' );
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/d3ff1a0bca12867e67756432a229d11cc6431d22

assets/js/googlesitekit/datastore/user/dismissed-items.js
```diff
@@ -48,11 +48,14 @@ const fetchGetDismissedItemsStore = createFetchStore( {
...
+	controlCallback: ( { slug, expiresInSeconds } ) => API.set( 'core', 'user', 'dismiss-item', { slug, expiresInSeconds } ),
 	reducerCallback,
-	argsToParams: ( slug ) => ( { slug } ),
-	validateParams: ( { slug } = {} ) => {
+	argsToParams: ( slug, expiresInSeconds = 0 ) => {
+		return ( { slug, expiresInSeconds } );
+	},
+	validateParams: ( { slug, expiresInSeconds } = {} ) => {
 		invariant( slug, 'slug is required.' );
+		invariant( Number.isInteger( expiresInSeconds ), 'expiresInSeconds should be an integer.' );
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/290e914c9575c090ce0c2086d951877847466e57

assets/js/googlesitekit-idea-hub-post-list-notice.js
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/e4a3fea58fb52114f5aef655f1ce6dfb1eeef9a0

assets/js/googlesitekit/datastore/user/dismissed-items.js
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/2b259f354141d7a53fd9ed478dd75e859c012981

includes/Modules/Idea_Hub.php
```diff
@@ -146,12 +161,126 @@ function( $maybe_empty, $postarr ) {
...
+					);
+					ob_start();
+					?>
+					<p>
+						<?php echo esc_html( $saved_notice['text'] ); ?>
+						<a href="<?php echo esc_url( $saved_notice['link'] ); ?>"
+						><?php echo esc_html( $saved_notice['cta'] ); ?></a>
+					</p>
+					<?php
+					return ob_get_clean();
```
eugene-manuilov, line 210: Let's render directly and get rid of the unnecessary `$saved_notice` variable.

includes/Modules/Idea_Hub.php
```diff
@@ -146,12 +161,126 @@ function( $maybe_empty, $postarr ) {
...
+					);
+					ob_start();
+					?>
+					<p>
+						<?php echo esc_html( $new_notice['text'] ); ?>
+						<a href="<?php echo esc_url( $new_notice['link'] ); ?>"
+						><?php echo esc_html( $new_notice['cta'] ); ?></a>
+					</p>
+					<?php
+					return ob_get_clean();
```
eugene-manuilov, line 250: Same here. Let's get rid of the `$new_notice` variable.

includes/Modules/Idea_Hub.php
```diff
@@ -146,12 +161,126 @@ function( $maybe_empty, $postarr ) {
...
+						$transients->set( self::TRANSIENT_SAVED_IDEAS, $saved_ideas, DAY_IN_SECONDS );
+					}
+					$has_saved_ideas = count( $saved_ideas ) > 0;
+					if ( ! $has_saved_ideas && $dismissed_items->is_dismissed( 'saved-ideas' ) ) {
+						// Saved items no longer need to be dismissed as there are none currently.
+						$dismissed_items->delete( 'saved-ideas' );
+					}
+					if ( $dismissed_items->is_dismissed( 'saved-ideas' ) ) {
+						return false;
+					}
```
eugene-manuilov, line 226: Let's check it at the beginning of the `activate_callback` function to return early if this notice is already dismissed.

includes/Modules/Idea_Hub.php
```diff
@@ -146,12 +161,126 @@ function( $maybe_empty, $postarr ) {
...
+						><?php echo esc_html( $new_notice['cta'] ); ?></a>
+					</p>
+					<?php
+					return ob_get_clean();
+				},
+				'type'            => Notice::TYPE_INFO,
+				'active_callback' => function() use ( $transients, $dismissed_items ) {
+					if ( $dismissed_items->is_dismissed( 'new-ideas' ) ) {
+						return false;
+					}
```
eugene-manuilov, line 256: We also need to check whether the `saved-ideas` notice is dismissed too. If either of these notices is dismissed, we should return early.

includes/Modules/Idea_Hub.php
```diff
@@ -146,12 +161,126 @@ function( $maybe_empty, $postarr ) {
...
+					$has_saved_ideas = count( $saved_ideas ) > 0;
+					if ( ! $has_saved_ideas && $dismissed_items->is_dismissed( 'saved-ideas' ) ) {
+						// Saved items no longer need to be dismissed as there are none currently.
+						$dismissed_items->delete( 'saved-ideas' );
+					}
+					if ( $dismissed_items->is_dismissed( 'saved-ideas' ) ) {
+						return false;
+					}
+
+					return $has_saved_ideas && ! $dismissed_items->is_dismissed( 'saved-ideas' );
```
eugene-manuilov, line 228: No need to check whether the `saved-ideas` notice is dismissed because we already check it above.

```suggestion
					return $has_saved_ideas;
```

eugene-manuilov (changes requested):
A few more comments from me:

includes/Modules/Idea_Hub.php
ivankruchkoff: I put it after due to the check for no items above:
```
if ( ! $has_saved_ideas && $dismissed_items->is_dismissed( 'saved-ideas' ) ) {
						// Saved items no longer need to be dismissed as there are none currently.
						$dismissed_items->delete( 'saved-ideas' );
					}
```

eugene-manuilov: Yes, you are right. Disregard this comment, please :)

includes/Modules/Idea_Hub.php
eugene-manuilov: Ok, let's use your approach for now. I think your version is the best option for now.

includes/Modules/Idea_Hub.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/850a96b2717cadb8ddfc0ee4119accb15f07b6c1

includes/Modules/Idea_Hub.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/739b0d3d19ed3d2c24eae7b424b0a37254b148d7

includes/Modules/Idea_Hub.php
ivankruchkoff: https://github.com/google/site-kit-wp/commit/850a96b2717cadb8ddfc0ee4119accb15f07b6c1

includes/Modules/Idea_Hub.php
ivankruchkoff: I don't _think_ we need to check for saved-ideas because if `saved-ideas` is dismissed, you can still get a notice for new ideas, no?

Isn't the check to avoid double notifying?

ivankruchkoff: Updated in https://github.com/google/site-kit-wp/pull/3668/commits/4a39e97f48c93a1ed4800eb5e93e15bdbbeadc6a

assets/js/googlesitekit-idea-hub-post-list-notice.js
```diff
@@ -0,0 +1,49 @@
...
+
+	// Button pops up only after the timeout passes
+	setTimeout( () => {
+		const button = notice.querySelector( '.notice-dismiss' );
+		if ( ! button ) {
+			return;
+		}
+		const expiresInSeconds = type === 'new-ideas' ? WEEK_IN_SECONDS : 0;
+		button.addEventListener( 'click', () => dispatch( CORE_USER ).dismissItem( type, { expiresInSeconds } ) );
+	}, 1 );
```
aaemnnosttv, line 48: Sorry I didn't notice this before, but it looks like the timeout here is to account for the element the selector is targeting which may not exist yet (added by WP I believe).

A safer approach would be to attach the event listener on the notice itself (which we know exists at this point) and modify the event listener to guard the dismissal by checking if the event target is a `.notice-dismiss`. That way we avoid a potential weird race condition where the dismiss button is added later so our dismiss handler wasn't bound.

includes/Modules/Idea_Hub.php
```diff
@@ -146,12 +161,118 @@ function( $maybe_empty, $postarr ) {
...
+					if ( false === $saved_ideas ) {
+						$saved_ideas = $this->get_data( 'saved-ideas' );
+						$transients->set( self::TRANSIENT_SAVED_IDEAS, $saved_ideas, DAY_IN_SECONDS );
+					}
+					$has_saved_ideas = count( $saved_ideas ) > 0;
+					if ( ! $has_saved_ideas && $dismissed_items->is_dismissed( 'saved-ideas' ) ) {
+						// Saved items no longer need to be dismissed as there are none currently.
+						$dismissed_items->delete( 'saved-ideas' );
+					}
+					if ( $dismissed_items->is_dismissed( 'saved-ideas' ) ) {
```
aaemnnosttv, line 219: As mentioned in https://github.com/google/site-kit-wp/pull/3668#discussion_r673811707 – I think we should use the same slug for the dismissed item slugs since these are not inherently scoped to anything. In the future, maybe we can even build in our dismissal handler into the notice itself rather than wiring it up manually.

This would apply to all `$dismissed_items->` calls here.

aaemnnosttv (commented):
Two more comments from me. Thanks for your patience @ivankruchkoff 🙂

includes/Modules/Idea_Hub.php
```diff
@@ -146,12 +161,118 @@ function( $maybe_empty, $postarr ) {
...
+
+					$new_ideas = $transients->get( self::TRANSIENT_NEW_IDEAS );
+					if ( false === $new_ideas ) {
+						$new_ideas = $this->get_data( 'new-ideas' );
+						$transients->set( self::TRANSIENT_NEW_IDEAS, $new_ideas, DAY_IN_SECONDS );
+					}
+
+					$has_new_ideas = count( $new_ideas ) > 0;
+
+					return $has_new_ideas && ! $dismissed_items->is_dismissed( 'new-ideas' );
```
eugene-manuilov, line 268: No need to check whether the `new-ideas` notice is dismissed here because we already checked it above.
```suggestion
					return $has_new_ideas;
```

eugene-manuilov (changes requested):
Thanks, @ivankruchkoff. Found a few more things to address.

assets/js/googlesitekit-idea-hub-post-list-notice.js
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/a189c759c53827ccac44c9bab61b9ff677f947c3

includes/Modules/Idea_Hub.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/5554b79b5bc3de3e7ffc7f31d411044e60563081

includes/Modules/Idea_Hub.php
```diff
@@ -146,12 +161,118 @@ function( $maybe_empty, $postarr ) {
...
+		$notices[] = new Notice(
+			'idea-hub_saved-ideas',
+			array(
+				'content'         => function() {
+					ob_start();
+					?>
+					<p>
+						<?php echo esc_html( esc_html__( 'Need some inspiration? Revisit your saved ideas in Site Kit', 'google-site-kit' ) ); ?>
+						<a href="<?php echo esc_url( $this->context->admin_url() . '#saved-ideas' ); ?>"
+						><?php echo esc_html( esc_html__( 'See saved ideas', 'google-site-kit' ) ); ?></a>
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/0664db90bcda5690cc03605124604e0b170dd988

includes/Modules/Idea_Hub.php
```diff
@@ -146,12 +161,118 @@ function( $maybe_empty, $postarr ) {
...
+		$notices[] = new Notice(
+			'idea-hub_new-ideas',
+			array(
+				'content'         => function() {
+					ob_start();
+					?>
+					<p>
+						<?php echo esc_html( esc_html__( 'Need some inspiration? Here are some new ideas from Site Kit’s Idea Hub', 'google-site-kit' ) ); ?>
+						<a href="<?php echo esc_url( $this->context->admin_url() . '#new-ideas' ); ?>"
+						><?php echo esc_html( esc_html__( 'See new ideas', 'google-site-kit' ) ); ?></a>
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/0664db90bcda5690cc03605124604e0b170dd988

includes/Modules/Idea_Hub.php
```diff
@@ -146,12 +161,118 @@ function( $maybe_empty, $postarr ) {
...
+				'active_callback' => function() use ( $transients, $dismissed_items ) {
+					$saved_ideas = $transients->get( self::TRANSIENT_SAVED_IDEAS );
+					if ( false === $saved_ideas ) {
+						$saved_ideas = $this->get_data( 'saved-ideas' );
+						$transients->set( self::TRANSIENT_SAVED_IDEAS, $saved_ideas, DAY_IN_SECONDS );
+					}
+					$has_saved_ideas = count( $saved_ideas ) > 0;
+					if ( ! $has_saved_ideas && $dismissed_items->is_dismissed( 'saved-ideas' ) ) {
+						// Saved items no longer need to be dismissed as there are none currently.
+						$dismissed_items->delete( 'saved-ideas' );
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/b7463e89cb5ab39cf73a75841ffb595692058c9a

includes/Modules/Idea_Hub.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/c5ab2b69e7d1b6bb5244ed52f8a5f22c071235c6

ivankruchkoff:
@googlebot I consent.

includes/Modules/Idea_Hub.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/c5ab2b69e7d1b6bb5244ed52f8a5f22c071235c6

includes/Modules/Idea_Hub.php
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3668/commits/c5ab2b69e7d1b6bb5244ed52f8a5f22c071235c6

assets/js/googlesitekit-idea-hub-post-list-notice.js
```diff
@@ -35,15 +35,12 @@ domReady( () => {
...
+	const type = notice.id.replace( 'googlesitekit-notice-', '' );
+	const expiresInSeconds = type === 'idea-hub_new-ideas' ? WEEK_IN_SECONDS : 0;
 
-	// Button pops up only after the timeout passes
-	setTimeout( () => {
-		const button = notice.querySelector( '.notice-dismiss' );
-		if ( ! button ) {
-			return;
+	notice.addEventListener( 'click', ( event ) => {
+		if ( 'notice-dismiss' === event.target.className ) {
```
aaemnnosttv, line 42: Let's make it conditional base on the presence of the class, rather than it _only having that class_ 🙂 
```suggestion
		if ( event.target.classList.contains( 'notice-dismiss' ) ) {
```

includes/Modules/Idea_Hub.php
```diff
@@ -70,6 +70,16 @@ final class Idea_Hub extends Module
...
 
+	/**
+	 * New ideas notice slug and dismissible item key.
+	 */
+	const SLUG_NEW_IDEAS = 'idea-hub_new-ideas';
+
+	/**
+	 * Saved ideas notice slug and dismissible item key.
+	 */
+	const SLUG_SAVED_IDEAS = 'idea-hub_new-ideas';
```
aaemnnosttv, line 81: Oops 😄 
```suggestion
	const SLUG_SAVED_IDEAS = 'idea-hub_saved-ideas';
```

aaemnnosttv (commented):
A few tweaks that I've added suggestions for and then that's all from me 👍

aaemnnosttv:
@googlebot I consent.

aaemnnosttv (approved):
Looks good from me – over to @eugene-manuilov in case there's anything left from him.

includes/Modules/Idea_Hub.php
```diff
@@ -146,12 +171,112 @@ function( $maybe_empty, $postarr ) {
...
+		}
+		$transients      = new Transients( $this->context );
+		$dismissed_items = new Dismissed_Items( $this->user_options );
+
+		$notices[] = new Notice(
+			self::SLUG_SAVED_IDEAS,
+			array(
+				'content'         => function() {
+					return sprintf(
+						'<p>%s <a href="%s">%s</a>',
```
eugene-manuilov, line 208:
```suggestion
						'<p>%s <a href="%s">%s</a></p>',
```

includes/Modules/Idea_Hub.php
```diff
@@ -146,12 +171,112 @@ function( $maybe_empty, $postarr ) {
...
+						esc_html__( 'Need some inspiration? Revisit your saved ideas in Site Kit', 'google-site-kit' ),
+						esc_url( $this->context->admin_url() . '#saved-ideas' ),
+						esc_html( esc_html__( 'See saved ideas', 'google-site-kit' ) )
```
eugene-manuilov, line 211:
```suggestion
						esc_html__( 'See saved ideas', 'google-site-kit' )
```

includes/Modules/Idea_Hub.php
```diff
@@ -146,12 +171,112 @@ function( $maybe_empty, $postarr ) {
...
+				},
+				'dismissible'     => true,
+			)
+		);
+		$notices[] = new Notice(
+			self::SLUG_NEW_IDEAS,
+			array(
+				'content'         => function() {
+					return sprintf(
+						'<p>%s <a href="%s">%s</a>',
```
eugene-manuilov, line 240:
```suggestion
						'<p>%s <a href="%s">%s</a></p>',
```

includes/Modules/Idea_Hub.php
```diff
@@ -146,12 +171,112 @@ function( $maybe_empty, $postarr ) {
...
+						esc_html__( 'Need some inspiration? Here are some new ideas from Site Kit’s Idea Hub', 'google-site-kit' ),
+						esc_url( $this->context->admin_url() . '#new-ideas' ),
+						esc_html( esc_html__( 'See new ideas', 'google-site-kit' ) )
```
eugene-manuilov, line 243:
```suggestion
						esc_html__( 'See new ideas', 'google-site-kit' )
```

eugene-manuilov (changes requested):

eugene-manuilov:
@googlebot I consent.

eugene-manuilov (approved):
Ok, looks good now!

Issue #3523, opened 2021-06-08
Implement Idea Hub dashboard notification informing about the module


felixarntz:
A Site Kit dashboard notification should be displayed to inform users about the new Idea Hub module if they have access to it. See the [Figma file](https://www.figma.com/file/YB79eKcGOwyUjlWG8mjDZA/Prompt-Engine-%2F-Idea-Hub?node-id=20332%3A181) for UI design.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new notification component `IdeaHubModuleNotification` should be implemented which prompts the user to activate the Idea Hub module.
    * The overall notice should look similar to [this Figma design](https://www.figma.com/file/YB79eKcGOwyUjlWG8mjDZA/Prompt-Engine-%2F-Idea-Hub?node-id=20332%3A181) (no need to follow pixel-perfect/design a new component, but follow what we already have in other notifications), _except_ for the buttons which should be left-aligned like we have in all other notifications.
    * Clicking the primary button should activate the Idea Hub module and then redirect to OAuth/module setup as usual.
    * Clicking the dismiss button should dismiss the notice persistently by storing the dismissal on the server-side. This should use the same implementation that #3360 has established, **however it must use a different identifier** (i.e. the dismissals for the notification and the widget CTA should be kept separate, even though they are for a rather similar purpose - we want to make sure people don't "accidentally" dismiss one; they should see both).
    * Notification graphic: The Idea Hub module icon.
    * Notification heading: _Get new ideas to write about based on what people are searching for_
    * Notification text: _Set up Idea Hub to get topic suggestions based on unanswered searches that match your site’s topic._
    * Notification button texts: _Set up_ / _Dismiss_
* The above notification should be shown on the Site Kit dashboard under the following conditions:
    * The `ideaHubModule` feature flag is enabled.
    * The Idea Hub module is not active.
    * The success feedback notification about having connected Site Kit or activated a module is not there (we want to avoid showing both together - it _is_ okay though to show this one alongside any remote `core/site/data/notifications` notifications).

## Implementation Brief
* Once #3360 has been merged.
* Create `assets/js/components/notifications/IdeaHubModuleNotification.js` which exports the `IdeaHubModuleNotification` functional component with the following details:
    * Create a constant for will be the predefined ID of the notification.
[...]

PR #3677, opened 2021-07-06
Add IdeaHubModuleNotification component.

asvinb:
## Summary

Addresses issue #3523 

## Checklist


assets/js/components/notifications/IdeaHubModuleNotification.js
```diff
@@ -0,0 +1,69 @@
...
+
+	const isActive = useSelect( ( select ) => select( CORE_MODULES ).isModuleActive( 'idea-hub' ) );
+	const isItemDismissed = useSelect( ( select ) => select( CORE_USER ).isItemDismissed( NOTIFICATION_ID ) );
+	const adminReauthURL = useSelect( ( select ) => select( MODULES_IDEA_HUB )?.getAdminReauthURL() );
+
+	const handleOnDismiss = useCallback( async () => {
+		await dismissItem( NOTIFICATION_ID );
+	}, [ dismissItem ] );
+
+	if ( isActive || isItemDismissed || isItemDismissed === undefined ) {
```
tofumatt, line 49: Technically, can't [`isActive` also be in an `undefined`/"loading" state](https://github.com/google/site-kit-wp/blob/65e6954e56d56167b15e218146e0e93e6ed852e5/assets/js/googlesitekit/modules/datastore/modules.js#L677-L679)? Similar to `isItemDismissed`, we might want to render `null` if it's still loading that info as well.

tofumatt:
@googlebot I consent.

tofumatt (changes requested):
I suppose because we don't have a setup screen for Idea Hub yet (but do need one), the redirect that this ends up using is a bit awkward; it redirects to this URL because Idea Hub gets setup, but there's no activation notice. But that's because we don't yet have a setup screen for it so that's fine...

Screenshot 2021-07-09 at 17 30 46

I did notice that dismissing the notification has an error in the JS console we should fix:

https://user-images.githubusercontent.com/90871/125109759-aa075900-e0db-11eb-84c4-132d81b9b543.mov

My commit efc79ee24814aecf53dd3c4994398e38a5f8fe4b fixes that.

-----

Regarding changes needed, it looks like the [storybook story isn't loading](https://google.github.io/site-kit-wp/storybook/pull/3677/?path=/story/modules-idea-hub-notifications-modulenotification--ready), because a `GET` call is failing/isn't being mocked:

Screenshot 2021-07-09 at 17 35 08

Just one other thing I think needs tweaking here, which is checking for an `undefined` return value from the `isActive()` selector.

assets/js/components/notifications/IdeaHubModuleNotification.js
asvinb: 👍

asvinb:
Thanks for catching that @tofumatt Guess I didn't check the story again after merging the latest code from develop. 🤦 
PR has been updated. Can you check again?

tofumatt (approved):
Awesome, thanks! 👍🏻

Issue #3518, opened 2021-06-08
Switch Idea Hub module REST routes to actually use Idea Hub API


felixarntz:
Now that the Idea Hub API client library code is available, Site Kit should actually start using the API.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `Idea_Hub` class should have its REST data points that are intended to proxy through requests to the Idea Hub API updated to actually use the API. **The necessary PHP client service classes are included in Site Kit via #3517.**
    * The class needs to implement the `setup_services` method, to bootstrap a single `ideahub` service, which should be [this class](https://github.com/googleapis/google-api-php-client-services/blob/master/src/Google/Service/Ideahub.php).
    * The `get_datapoint_definitions` method needs to be modified to reference the new `ideahub` service for all data points that use the API (`GET:new-ideas`, `GET:saved-ideas`, `POST:update-idea-state`).
    * The three above endpoints need to be connected to the respective Idea Hub API endpoint:
        * For `GET:new-ideas` and `GET:saved-ideas`, remove the hard-coded return values and instead use https://github.com/googleapis/google-api-php-client-services/blob/master/src/Google/Service/Ideahub/Resource/PlatformsPropertiesIdeas.php, passing `sitekit` for platform and the _URL-encoded_ current site's URL ("reference site URL") for property. In order to only return new ideas, the `filter` option needs to be used, passing a string "saved(false)". In order to only return saved ideas, the `filter` option needs to be used, passing a string "saved(true)".
        * ~~For `POST:update-idea-state`, use https://github.com/googleapis/google-api-php-client-services/blob/master/src/Google/Service/Ideahub/Resource/PlatformsPropertiesIdeaStates.php, passing platform and property as mentioned above and the received idea name as the "ideaState" placeholder in the path. The data point should expect a single variable `ideaState` which needs to be an object / associative array with `name` (idea name string, also used in the path as mentioned right above), `saved` (boolean) and `dismissed` (boolean) fields.~~

## Implementation Brief

* Add a `protected function setup_services` to the `Idea_Hub` class that creates a new instance of Google_Service_Ideahub` to be used in the instance of `Idea_Hub`.
* Modidy the existing `get_datapoint_definitions` in the `Idea_Hub` class so that the following point to `array( 'service' => 'ideahub' )`:
    * `GET:new-ideas`
    * `GET:saved-ideas`
    * ~~`POST:update-idea-state`~~
[...]

PR #3681, opened 2021-07-07
Enhancement/3518 idea hub api

eugene-manuilov:
## Summary

Addresses issue #3518

## Relevant technical choices

* Renamed incorrect `display_name` property of an idea topic object to be `displayName` as it is returned from the Idea Hub API.
* Added a new `getNewIdeasSlice` selector to return a slice from the new ideas array and fixed incorrect behavior of the `getNewIdeas` selector.
* Added a new `getSavedIdeasSlice` selector to return a slice from the saved ideas array and fixed incorrect behavior of the `getSavedIdeas` selector.
* Added a new `getDraftPostIdeasSlice` selector to return a slice from the draft post ideas array and fixed incorrect behavior of the `getDraftPostIdeas` selector.

## Checklist


includes/Modules/Idea_Hub.php
```diff
@@ -384,15 +328,38 @@ protected function create_data_request( Data_Request $data ) {
...
+
+				$parent = $this->get_parent_slug();
+				$body   = new Google_Service_Ideahub_GoogleSearchIdeahubV1alphaIdeaState();
+
+				$body->setName( $data['name'] );
+
+				if ( isset( $data['saved'] ) ) {
+					$parent = $parent . '/ideaStates/saved';
+					$body->setSaved( filter_var( $data['saved'], FILTER_VALIDATE_BOOL ) );
+				} elseif ( isset( $data['dismissed'] ) ) {
```
felixarntz, line 350: This shouldn't/doesn't need to be an either/or. It is acceptable to specify both states at the same time, i.e. it would be more appropriate to have two individual `if` clauses here. Only if none of the two is provided, the request is pointless, so returning an error then makes sense.

includes/Modules/Idea_Hub.php
```diff
@@ -299,66 +302,7 @@ protected function create_data_request( Data_Request $data ) {
...
-								array(
-									array(
-										'mid'          => '/m/07fhc',
-										'display_name' => 'Cooking',
-									),
-								),
-						),
-					);
-				};
+				return $this->fetch_ideas( 'new' );
```
felixarntz, line 305: This datapoint (as well as the one for saved ideas) also need to get a case within the `parse_data_response` method, since only the array of `ideas` within the response object should be returned, not the full response object (similar to how all our other API proxy datapoints work).

In addition to that, we also need to afterwards filter out any ideas from the response for which there already is a WordPress draft or published post - in other words, these endpoints should never return ideas that would also be included in the response from `GET:draft-post-ideas` or `GET:published-post-ideas`.

felixarntz (commented):
@eugene-manuilov Just tested the API myself using this, unfortunately no progress. Using one of the test sites that my Google account has access to, I still get an empty object back from the API.

So far, the only thing that seems to work is the following:
* If I use a site that I actually own (via Site Verification API, which Idea Hub internally uses for permission checks), I get the empty object back.
* If I use a site I don't own, I get a permission error.

So this is still blocked. I reached out to the team again, will keep you posted. In the meantime, I left two other comments here that could already be addressed now.

includes/Modules/Idea_Hub.php

includes/Modules/Idea_Hub.php
eugene-manuilov: Yes, I know it. I intentionally didn't add it yet. Will do it once I can get ideas on my end.

eugene-manuilov: Implemented.

includes/Modules/Idea_Hub.php
```diff
@@ -730,4 +728,89 @@ private function is_idea_post( $post_id ) {
...
+
+		// phpcs:ignore WordPress.DB.DirectDatabaseQuery
+		$drafted_ideas = $wpdb->get_col(
+			// phpcs:ignore WordPress.DB.PreparedSQLPlaceholders.ReplacementsWrongNumber
+			$wpdb->prepare(
+				"SELECT DISTINCT meta_value FROM {$wpdb->postmeta} WHERE meta_key = %s AND meta_value IN (" . implode( ', ', array_fill( 0, count( $names ), '%s' ) ) . ')',
+				Post_Idea_Name::META_KEY,
+				...$names
+			)
+		);
```
felixarntz, line 799: While this is clever (and I don't have any problem bypassing object cache here), I think we should use a similar approach here to what we're doing for `GET:draft-post-ideas` (and the similar `GET:published-post-ideas`).

For example, what you have here now ignores whether the post is a draft, it will get this data for _any_ post that has it - including e.g. trashed posts. We should ensure that we use a single approach for getting ideas that have a post throughout to avoid any inconsistencies.

I'd be in favor of just using the established `WP_Query` approach using a `meta_query` rather than putting together a direct database query - I don't think the performance benefit of the simplified query is worth it, and on the other end it would mean that object caching isn't used when available. How about we introduce a private method `get_idea_posts( $status )` which we then use in here, in `GET:draft-post-ideas` and in `GET:published-post-ideas`?

felixarntz (changes requested):
@eugene-manuilov This mostly looks good, just a few minor comments and one slightly more significant point of feedback about consolidating the logic to get Idea Hub posts.

Last but not least, please remove the idea state-related changes, which should be part of #3747 instead.

includes/Modules/Idea_Hub.php
```diff
@@ -10,10 +10,12 @@
 
 namespace Google\Site_Kit\Modules;
 
+use Google\Service\ShoppingContent\Resource\Pos;
```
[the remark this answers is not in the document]
eugene-manuilov: Hm.... ok, removed.

includes/Modules/Idea_Hub.php
```diff
@@ -533,15 +478,41 @@ protected function create_data_request( Data_Request $data ) {
...
+					$parent = $parent . '/ideaStates/saved';
+					$body->setSaved( filter_var( $data['saved'], FILTER_VALIDATE_BOOLEAN ) );
+				}
+
+				if ( isset( $data['dismissed'] ) ) {
+					$parent = $parent . '/ideaStates/dismissed';
+					$body->setDismissed( filter_var( $data['dismissed'], FILTER_VALIDATE_BOOLEAN ) );
+				}
+
+				return $this->get_service( 'ideahub' )->platforms_properties_ideaStates->patch( $parent, $body );
```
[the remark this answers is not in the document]
eugene-manuilov: Removed.

includes/Modules/Idea_Hub.php
```diff
@@ -345,10 +349,10 @@ protected function get_datapoint_definitions() {
 		return array(
 			'POST:create-idea-draft-post' => array( 'service' => '' ),
 			'GET:draft-post-ideas'        => array( 'service' => '' ),
-			'GET:new-ideas'               => array( 'service' => '' ),
+			'GET:new-ideas'               => array( 'service' => 'ideahub' ),
 			'GET:published-post-ideas'    => array( 'service' => '' ),
-			'GET:saved-ideas'             => array( 'service' => '' ),
-			'POST:update-idea-state'      => array( 'service' => '' ),
+			'GET:saved-ideas'             => array( 'service' => 'ideahub' ),
+			'POST:update-idea-state'      => array( 'service' => 'ideahub' ),
```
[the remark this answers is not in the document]
eugene-manuilov: Updated

includes/Modules/Idea_Hub.php
```diff
@@ -594,6 +567,14 @@ function( $post_id ) {
...
 					)
 				);
+			case 'GET:saved-ideas':
+				return $this->filter_out_drafted_ideas( $response->getIdeas() );
+			case 'POST:update-idea-state':
+				return array(
+					'name'      => $response->getName(),
+					'saved'     => $response->getSaved(),
+					'dismissed' => $response->getDismissed(),
+				);
```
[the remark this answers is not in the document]

includes/Modules/Idea_Hub.php
eugene-manuilov: > I think we should use a similar approach here to what we're doing for GET:draft-post-ideas (and the similar GET:published-post-ideas).
> We should ensure that we use a single approach for **getting ideas that have a post** throughout to avoid any inconsistencies.
> How about we introduce a private method get_idea_posts( $status ) which we then use in here, in GET:draft-post-ideas and in GET:published-post-ideas?

The problem is that `GET:draft-post-ideas` and `GET:published-post-ideas` datapoints **get posts that have an idea**, not the other way around. Therefore we can't introduce a private method that will use one algorithm for all `{draft|published}-post-ideas` and `{new|saved}-ideas` datapoints.

> I'd be in favor of just using the established WP_Query approach using a meta_query rather than putting together a direct database query - I don't think the performance benefit of the simplified query is worth it, and on the other end it would mean that object caching isn't used when available.

Currently, this method adds one SQL query per request to `new-ideas`/`saved-ideas` endpoints. If we rework it to use `WP_Query`, it will add `1 + N` SQL queries for the first request (where `N` is the number of ideas, the current limit is `100`) when object cache is not used (usually on shared hostings) and the same `1` query for sequential requests (because queries created by WP_Query are not cached) when object cache is used on the server. Plus, the query created by WP_Query will be always much slower than the direct query to the postmeta table.

So, I think it would be better if we use the direct query to the postmeta table as it is done right now. But if you still want me to rework it to use the WP_Query approach, let me know and I'll update it.

assets/js/modules/idea-hub/datastore/draft-post-ideas.js
```diff
@@ -59,26 +60,37 @@ const baseSelectors = {
...
 	 *
+	 * @param {Object} state Data store's state.
+	 * @return {(Array.<Object>|undefined)} A list of idea hub ideas; `undefined` if not loaded.
+	 */
+	getDraftPostIdeas( state ) {
+		return state.draftPostIdeas;
+	},
+
+	/**
+	 * Gets Draft Post Ideas from the Idea Hub.
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

includes/Modules/Idea_Hub.php
eugene-manuilov: > For example, what you have here now ignores whether the post is a draft, it will get this data for any post that has it - including e.g. trashed posts.

Forgot to comment on this. If I created a post and then decided to move it to trash, it means that I am not interested in that idea anymore, right? I think we don't need to care about post statuses when we filter new and saved ideas.

aaemnnosttv:
@eugene-manuilov would you please resolve the merge conflicts here?

eugene-manuilov:
@aaemnnosttv fixed.

includes/Modules/Idea_Hub.php
```diff
@@ -37,6 +38,8 @@
 use Google\Site_Kit\Modules\Idea_Hub\Post_Idea_Text;
 use Google\Site_Kit\Modules\Idea_Hub\Post_Idea_Topics;
 use Google\Site_Kit\Modules\Idea_Hub\Settings;
+use Google\Site_Kit_Dependencies\Google\Service\Ideahub as Google_Service_Ideahub;
+use Google\Site_Kit_Dependencies\Google\Service\Ideahub\GoogleSearchIdeahubV1alphaIdeaState as Google_Service_Ideahub_GoogleSearchIdeahubV1alphaIdeaState;
```
felixarntz, line 42: This is unused for now, so should be removed.

includes/Modules/Idea_Hub.php
felixarntz: @eugene-manuilov 
> > How about we introduce a private method get_idea_posts( $status ) which we then use in here, in GET:draft-post-ideas and in GET:published-post-ideas?
> 
> The problem is that `GET:draft-post-ideas` and `GET:published-post-ideas` datapoints **get posts that have an idea**, not the other way around. Therefore we can't introduce a private method that will use one algorithm for all `{draft|published}-post-ideas` and `{new|saved}-ideas` datapoints.

I don't see why that is a limitation. For `GET:new-ideas` and `GET:saved-ideas`, we could query the ("draft" or "publish") posts that have an idea, loop through them and remove them from the ideas list.

> > I'd be in favor of just using the established WP_Query approach using a meta_query rather than putting together a direct database query - I don't think the performance benefit of the simplified query is worth it, and on the other end it would mean that object caching isn't used when available.
> 
> Currently, this method adds one SQL query per request to `new-ideas`/`saved-ideas` endpoints. If we rework it to use `WP_Query`, it will add `1 + N` SQL queries for the first request (where `N` is the number of ideas, the current limit is `100`) when object cache is not used (usually on shared hostings) and the same `1` query for sequential requests (because queries created by WP_Query are not cached) when object cache is used on the server. Plus, the query created by WP_Query will be always much slower than the direct query to the postmeta table.

I'm not sure the "1 + N" is entirely accurate. When `WP_Query` queries 100 posts, it will typically make 2 queries, with or without persistent object cache:
1. Get the IDs for all posts that satisfy the query.
2. Query all data for those ideas to prime the relevant post caches. Without object cache, this is still cached in a variable, so no subsequent queries are made, i.e. that's where I think the "1 + N" is not accurate.

> Forgot to comment on this. If I created a post and then decided to move it to trash, it means that I am not interested in that idea anymore, right? I think we don't need to care about post statuses when we filter new and saved ideas.

We shouldn't use that as a signal that the idea isn't relevant anymore. Maybe the user wrote a post, thought it was bad and trashed it to write another one later. Specifically for both `GET:new-ideas` and `GET:saved-ideas`, we should filter out any ideas for which there is a post, no matter whether it's draft or publish (we also need to ensure a published post's idea doesn't still appear in "New" or "Saved"). Technically, we should only exclude trashed posts, since those cannot be edited and it should be treated like a deleted post.

I think we should go with the `WP_Query` approach, so that we
* Can take the post status into account, which we should definitely do
* Keep all queries around posts with an idea follow the same approach to avoid inconsistent data
* Follow the WP best practice to not use direct DB queries unless absolutely necessary (this is the least important point, but I don't think the performance benefit here is relevant enough, for the above reason)

I think the most flexible approach would be a private method like `query_idea_posts( array $args )` where you can pass extra arguments for querying that will be merged into defaults. The defaults could be what's currently provided in `GET:published-post-ideas`. We can then use that method as follows:
* In `GET:new-ideas` and `GET:saved-ideas` query posts of any post status other than "trash".
* In `GET:draft-post-ideas` query posts of the "draft" post status.
* In `GET:published-post-ideas` query posts of the "publish" post status.

includes/Modules/Idea_Hub.php
```diff
@@ -730,4 +693,89 @@ private function is_idea_post( $post_id ) {
...
+
+	/**
+	 * Filters out ideas for which we have already created a post.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $ideas Ideas list to filter.
+	 * @return array Filtered ideas list.
+	 */
+	private function filter_out_drafted_ideas( $ideas ) {
```
felixarntz, line 746: This method name _right now_ is technically inaccurate, since at the moment it queries post ideas regardless of "draft" post status.
But anyway, see my other comment reply, this isn't exactly what we need to do here. It should filter out ideas for which there's a post of _any_ post status other than "trash".

felixarntz (changes requested):
@eugene-manuilov This now looks good mostly, but I still think it would be better to consistently use `WP_Query` to get posts with ideas. I left a more detailed reply below.

includes/Modules/Idea_Hub.php
eugene-manuilov: Removed.

includes/Modules/Idea_Hub.php
eugene-manuilov: @felixarntz I think you are right, thanks for the detailed answer. Updated. :+1:

includes/Modules/Idea_Hub.php

aaemnnosttv (commented):
Thanks @eugene-manuilov – this looks pretty good to me. I left a few comments/suggestions but overall looking good. I'll leave it to @felixarntz to weigh in with any final thoughts here.

assets/js/modules/idea-hub/datastore/draft-ideas.js
```diff
@@ -42,6 +42,8 @@ const fetchCreateIdeaDraftPostStore = createFetchStore( {
 			{ idea }
 		);
 
+		API.invalidateCache( 'modules', 'idea-hub', 'new-ideas' );
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/idea-hub/datastore/new-ideas.test.js
```diff
@@ -51,194 +53,95 @@ describe( 'modules/idea-hub new-ideas', () => {
...
-				);
-
-				registry.select( MODULES_IDEA_HUB ).getNewIdeas( options );
-				await untilResolved( registry, MODULES_IDEA_HUB ).getNewIdeas(
-					options
-				);
-
-				expect( fetchMock ).toHaveFetchedTimes( 1 );
-
+			it( 'should adjust idea results when only limit parameter is supplied', async () => {
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/sass/components/settings/_googlesitekit-settings-notice.scss
```diff
@@ -171,6 +173,10 @@
 	&--info {
 		background-color: $c-background-notice-info;
 		color: $c-text-notice-info;
+
+		.googlesitekit-cta-link {
+			color: $c-text-notice-info;
```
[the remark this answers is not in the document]
eugene-manuilov: This is not from my branch. Looks like it had become here after rebasing my branch. I have merged `develop` into my branch and these changes have disappeared.

includes/Modules/Idea_Hub.php
```diff
@@ -730,4 +652,124 @@ private function is_idea_post( $post_id ) {
...
+	 *
+	 * @return string Parent slug.
+	 */
+	private function get_parent_slug() {
+		$reference_url = $this->context->get_reference_site_url();
+		$reference_url = rawurlencode( $reference_url );
+
+		$parent = "platforms/sitekit/properties/{$reference_url}";
+
+		return $parent;
```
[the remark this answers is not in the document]

includes/Modules/Idea_Hub.php
```diff
@@ -730,4 +652,124 @@ private function is_idea_post( $post_id ) {
...
+						'compare' => 'EXISTS',
+					),
+					array(
+						'key'     => Post_Idea_Text::META_KEY,
+						'compare' => 'EXISTS',
+					),
+					array(
+						'key'     => Post_Idea_Topics::META_KEY,
+						'compare' => 'EXISTS',
+					),
```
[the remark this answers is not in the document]
eugene-manuilov: I left only one meta query for the `Post_Idea_Name::META_KEY` key which is what we really need here.

includes/Modules/Idea_Hub.php
```diff
@@ -730,4 +652,124 @@ private function is_idea_post( $post_id ) {
...
+		if ( 'saved' === $type ) {
+			$params['filter'] = 'saved(true)';
+		} elseif ( 'new' === $type ) {
+			$params['filter'] = 'saved(false)';
+		}
+
+		$service = $this->get_service( 'ideahub' );
+		$ideas   = $service->platforms_properties_ideas;
+
+		return $ideas->listPlatformsPropertiesIdeas( $parent, $params );
```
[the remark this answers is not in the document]

felixarntz (commented):
@eugene-manuilov LGTM, except there's a few failing tests - maybe some code formatting?

eugene-manuilov:
@felixarntz broken js tests are fixed.

eugene-manuilov:
@felixarntz, please, ignore phpunit tests failure for WP Nightly this is not relevant to this PR.

felixarntz (approved):
Excellent work!

aaemnnosttv (approved):

Issue #2418, opened 2020-11-24
Deprecate legacy functions with modern alternatives and prune dead code


aaemnnosttv:
## Feature Description

There are a number of legacy functions in the codebase which have newer implementations which should be used instead for all new usage.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All legacy functions which are no longer used should be deleted
    - Imports/exports and tests do not count as usage. In this case, only references in source files/components should be considered
    - Please list these instances in the IB
* Any remaining legacy functions with a modern alternative (e.g. `getSiteKitAdminURL` - replaced by `select( CORE_SITE ).getAdminURL(...)`) should be annotated as `@deprecated n.e.x.t Use x instead.`
    - Please list these instances in the IB
    - Most (if not all) of these will likely be under `assets/js/util/index.js`

## Implementation Brief

Delete the following from `assets/js/util`:
* `refresh-authentication.js`
* `average-object-list-value.js`
* `cache-data.js`

Using `assets/js/util/index.js` delete the following functions:
* `removeURLFallBack`
* `removeURLParameter`
* `getDaysBetweenDates`
* `validateOptimizeID`

Delete the following from `assets/js/util/test`:
* `removeURLParameter.js`
* `getDaysBetweenDates.js`
* `validateOptimizeID.js`
* `getSiteKitAdminURL.js`
* `averageObjectListValue.js`

In the following components remove the import of `getSiteKitAdminURL` and replace calls to it with the store equivalent `useSelect( ( select ) => select( CORE_SITE ).getAdminURL( page , args ) )`.  In each case add imports for `useSelect` and `CORE_SITE`:
* `js/components/data-table`
* `js/components/legacy-setup/SetupUsingGCP`
* `js/components/setup/ModuleSetup`
* `js/modules/adsense/components/dashboard/LegacyAdsenseDashboardMainSummary`

In `js/util/index` remove the `getSiteKitAdminURL` function and its tests in `assets/js/util/test/getSiteKitAdminURL.js`

### Test Coverage

* run js tests and check that they pass.

### Visual Regression Changes

* n/a

## QA Brief

As per the IB, check:
* `util` files have been deleted
* functions have been removed for `index.js`
* associated tests have been deleted
* calls to `getSiteKitAdminURL` have been replaced

Also check js and VRT tests pass.

## Changelog entry

* N/A

felixarntz:
[...]

PR #3698, opened 2021-07-12
Enhancement/2418 prune dead code

Hazlitte:
## Summary

Prune dead code and replace calls to `getSiteKitAdminURL`.

Addresses issue #2418

## Checklist


assets/js/components/data-table.js
```diff
@@ -51,6 +54,7 @@ export const getDataTableFromData = ( data, headers, options ) => {
 		const cells = [];
 		const link = links && links[ j ];
 		const permaLink = link && link[ 0 ] === '/' ? getFullURL( global._googlesitekitLegacyData.admin.siteURL, link ) : link;
+		const adminURL = useSelect( ( select ) => select( CORE_SITE ).getAdminURL( 'googlesitekit-dashboard', { permaLink } ) );
```
tofumatt, line 57: I'm surprised this doesn't trigger a `rules-of-hooks` ESLint error, but this looks like it would [break the rules of hooks](https://reactjs.org/docs/hooks-rules.html) by calling a hook conditionally and/or a variable number of times. Hooks need to be called in the exact same order every time for every render, and the number of hooks called in each component should not change between renders. Placing hooks inside a loop will cause bugs/errors.

Usually in these cases, we'll want to abstract out a new, small component that selects the `adminURL` based on a `permaLink` prop. In this case, something like replacing `<PrimaryLink>` on line 77 with `<PrimaryAdminLink permaLink={permaLink}>{cell}</PrimaryAdminLink>`. That component would then call the hook, eg:

```js
const PrimaryAdminLink = ({children, permaLink}) => {
    const adminURL = useSelect( ( select ) => select( CORE_SITE ).getAdminURL( 'googlesitekit-dashboard', { permaLink } ) );

    return (
        <PrimaryLink
            className="googlesitekit-table__body-item-link"
            href={ useAdminURLs ? getSiteKitAdminURL( 'googlesitekit-dashboard', { permaLink } ) : link }
            href={ useAdminURLs ? adminURL : link }
            external={ ! useAdminURLs }
            inherit
        >
            { children }
        </PrimaryLink>
    );
};
```

tofumatt (changes requested):
Overall looks good except the usage of hooks in `data-table.js`. I made a suggestion for how to refactor this, and after that this should be good.

assets/js/components/data-table.js
Hazlitte: Thanks @tofumatt.  No this didn't trigger an ESLint error.  I have refactored this aspect as suggested 👍

assets/js/components/data-table.js
```diff
@@ -70,14 +88,14 @@ export const getDataTableFromData = ( data, headers, options ) => {
...
 						? <div className="googlesitekit-table__body-item-content">
-							<PrimaryLink
-								className="googlesitekit-table__body-item-link"
-								href={ useAdminURLs ? getSiteKitAdminURL( 'googlesitekit-dashboard', { permaLink } ) : link }
-								external={ ! useAdminURLs }
-								inherit
+							<PrimaryAdminLink
+								permaLink={ permaLink }
+								useAdminURLs={ useAdminURLs }
+								PrimaryLink={ PrimaryLink }
```
tofumatt, line 94: I was curious about this component being passed around; it turns out it isn't actually used anywhere this function is called:

* https://github.com/google/site-kit-wp/blob/dc6ccb44c39e12a968c55942273b3dc77d427562/assets/js/modules/search-console/components/dashboard/LegacySearchConsoleDashboardWidgetKeywordTable.js#L103-L109
* https://github.com/google/site-kit-wp/blob/dc6ccb44c39e12a968c55942273b3dc77d427562/assets/js/modules/search-console/components/dashboard/LegacyDashboardWidgetPopularKeywordsTable.js#L102-L108
* https://github.com/google/site-kit-wp/blob/dc6ccb44c39e12a968c55942273b3dc77d427562/assets/js/modules/analytics/components/dashboard/LegacyAnalyticsDashboardWidgetTopAcquisitionSources.js#L108-L113
* https://github.com/google/site-kit-wp/blob/dc6ccb44c39e12a968c55942273b3dc77d427562/assets/js/modules/analytics/components/dashboard/LegacyAnalyticsDashboardWidgetPopularPagesTable.js#L91-L99

So let's remove it entirely and just use the `Link` component directly rather than passing this around. 🙂

tofumatt (commented):
Just one tweak to make here and it's straightforward, so I've made the adjustment. Thanks!

tofumatt:
@googlebot I consent.

tofumatt (approved):

Issue #3614, opened 2021-06-22
Don't display the blue notice bar until a user selects either a GA4 or UA property when connecting Analytics. 


jamesozzie:
When setting up Analytics and selecting an account which has **both a UA and GA4 property** don't display the blue notice bar regarding the creation of an additional property until a user makes a selection. 

This would lead to less confusion considering the notice bar that appears on account selection assumes the user will select a UA property. 

![2021-06-22-14-03-27](https://user-images.githubusercontent.com/41326532/122929632-14da3580-d363-11eb-8de5-4be288fb6bac.gif)

_Screenshot of the current blue bar that appears upon selecting an Analytics account with both a UA and GA4 property._ 

![image](https://user-images.githubusercontent.com/41326532/122930296-c4afa300-d363-11eb-8ece-17b493afbc9c.png)

The blue notice assumes a user will select a UA property. If the user selects a GA4 property the notice does change, which is ideal however an improvement would be to only display that notice when a property selection has been made. 

![image](https://user-images.githubusercontent.com/41326532/122930857-3ee02780-d364-11eb-8d19-2a489c00273c.png)

## Steps to reproduce

1. Setup Site Kit 
2. Connect Analytics
3. Select an account that has both a UA and GA4 property
4. The blue notice bar appears before the user makes a property selection, assuming they will select a UA property. 

## Additional Context

 - PHP Version: 
 - Plugin Version [e.g. 22]

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* In the Analytics setup flow (with `ga4Setup` flag active), any of the three new blue notices around the associated other property (whether about creating a new one or selecting a matching one) should only appear when a property is already selected in the primary dropdown.
    * In other words, if no account/property is selected or if only an account has been selected, the notice should not be visible _yet_.

## Implementation Brief

* Update the `SetupFormUA` component:
  * Pull UA propertyID using the `getPropertyID` selector
  * Update the `shouldShowGA4PropertyNotice` constant to also check if the propertyID is not empty.
* Update the `SetupFormGA4` component:
  * Pull GA4 propertyID using the `getPropertyID` selector
  * Update the `GA4PropertyNotice` element to render it only when the `propertyID` is not empty.
* Update the `SetupFormGA4Transitional` component:
  * Pull GA4 propertyID using the `getPropertyID` selector of the analytics-4 module
[...]

PR #3699, opened 2021-07-13
Enhancement/3614 blue notice bar updates

eugene-manuilov:
## Summary

Addresses issue #3614

## Relevant technical choices

This PR does a few things:

* Updates the `SetupFormGA4Transitional` component to display the blue notice bar only when the primary property is not empty.
* Fixes the issue on the setup form when a  user could submit the setup form without selecting the associated GA4 property. Steps to reproduce the issue in the develop branch:
  1. Disconnect the Analytics module and enable the ga4setup feature flag.
  1. Using the developer plugin, set the current reference site URL to a website URL that is not present in any of your UA/GA4 properties
  1. Connect the Analytics module and go to the setup form
  1. Select an account that has both UA and GA4 properties
  1. Select a UA property and make sure you don't select a GA4 property and it doesn't say that it will create a new one for you
  1. Check the submit button - it is not disabled and allows you to submit your selection with GA4 propertyID not being selected
* Fixes the issue on the settings form when a GA4 property could be selected unintentionally in a situation when a user selects a different account. Steps to reproduce the issue in the develop branch:
   1. Disconnect the Analytics module and disable the ga4setup feature flag
   1. Connect the Analytics module again and go to the setup form
   1. Verify that GA4 functionality isn't working (you don't see new notices on the setup form).
   1. Select an account with UA properties only and submit the setup form
   1. Enable the ga4setup feature flag again
   1. Go to the Analytics settings and edit it
   1. Run `googlesitekit.data.select( 'modules/analytics-4' ).getSettings()` in the browser console to see analytics-4 settings. It should show empty propertyID, webDataStreamID, and measurementID settings.
   1. Select another account that has GA4 properties that match the current website URL
   1. Check the analytics-4 settings, now it has propertyID, webDataStreamID and measurementID added
   1. If you save settings, analytics-4 settings will be saved as well

## Checklist

### DIFF
--- a/assets/js/modules/analytics/components/common/UseUAandGA4SnippetSwitches.js
+++ b/assets/js/modules/analytics/components/common/UseUAandGA4SnippetSwitches.js
@@ -26,19 +26,35 @@ import { __ } from '@wordpress/i18n';
  * Internal dependencies
  */
 import Data from 'googlesitekit-data';
+import { CORE_MODULES } from '../../../../googlesitekit/modules/datastore/constants';
 import { STORE_NAME } from '../../datastore/constants';
 import { MODULES_ANALYTICS_4 } from '../../../analytics-4/datastore/constants';
 import Switch from '../../../../components/Switch';
 import { trackEvent } from '../../../../util';
+import { useFeature } from '../../../../hooks/useFeature';
 const { useSelect, useDispatch } = Data;
 
 export default function UseUAandGA4SnippetSwitches() {
+	const isGA4Enabled = useFeature( 'ga4setup' );
+
 	const useUASnippet = useSelect( ( select ) => select( STORE_NAME ).getUseSnippet() );
 	const canUseUASnippet = useSelect( ( select ) => select( STORE_NAME ).getCanUseSnippet() );
 
-	const ga4ExistingTag = useSelect( ( select ) => select( MODULES_ANALYTICS_4 ).getExistingTag() );
-	const ga4MeasurementID = useSelect( ( select ) => select( MODULES_ANALYTICS_4 ).getMeasurementID() );
-	const useGA4Snippet = useSelect( ( select ) => select( MODULES_ANALYTICS_4 ).getUseSnippet() );
+	const {
+		ga4ExistingTag,
+		ga4MeasurementID,
+		useGA4Snippet,
+	} = useSelect( ( select ) => {
+		if ( ! isGA4Enabled || select( CORE_MODULES ).isModuleConnected( 'analytics-4' ) === false ) {
+			return {};
+		}
+
+		return {
+			ga4ExistingTag: select( MODULES_ANALYTICS_4 ).getExistingTag(),
+			ga4MeasurementID: select( MODULES_ANALYTICS_4 ).getMeasurementID(),
+			useGA4Snippet: select( MODULES_ANALYTICS_4 ).getUseSnippet(),
+		};
+	} );
 
 	const { setUseSnippet: setUseUASnippet } = useDispatch( STORE_NAME );
 	const { setUseSnippet: setUseGA4Snippet } = useDispatch( MODULES_ANALYTICS_4 );
@@ -59,7 +75,7 @@ export default function UseUAandGA4SnippetSwitches() {
 		trackEvent( 'analytics_setup', useGA4Snippet ? 'analytics4_tag_enabled' : 'analytics4_tag_disabled' );
 	}, [ useGA4Snippet, setUseGA4Snippet ] );
 
-	if ( useGA4Snippet === undefined || useUASnippet === undefined ) {
+	if ( useUASnippet === undefined ) {
 		return null;
 	}
 
@@ -87,14 +103,17 @@ export default function UseUAandGA4SnippetSwitches() {
 						disabled={ ! canUseUASnippet }
 					/>
 				</div>
-				<div className="googlesitekit-settings-module__inline-item">
-					<Switch
-						label={ __( 'Place Google Analytics 4 code', 'google-site-kit' ) }
-						checked={ useGA4Snippet }
-						onClick={ onGA4Change }
-						hideLabel={ false }
-					/>
-				</div>
+
+				{ useGA4Snippet !== undefined && (
+					<div className="googlesitekit-settings-module__inline-item">
+						<Switch
+							label={ __( 'Place Google Analytics 4 code', 'google-site-kit' ) }
+							checked={ useGA4Snippet }
+							onClick={ onGA4Change }
+							hideLabel={ false }
+						/>
+					</div>
+				) }
 			</div>
 
 			<p className="googlesitekit-margin-top-0">{ message }</p>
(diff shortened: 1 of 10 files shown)


assets/js/modules/analytics/components/common/UseUAandGA4SnippetSwitches.js
```diff
@@ -87,14 +103,17 @@ export default function UseUAandGA4SnippetSwitches() {
...
-				<div className="googlesitekit-settings-module__inline-item">
-					<Switch
-						label={ __( 'Place Google Analytics 4 code', 'google-site-kit' ) }
-						checked={ useGA4Snippet }
-						onClick={ onGA4Change }
-						hideLabel={ false }
-					/>
-				</div>
+
+				{ useGA4Snippet !== undefined && (
```
felixarntz, line 107: Nit-pick: This is technically correct of course, but it reads a bit confusing. We mostly use `undefined` checks for some sort of loading state check, but here that is different. I think it would be a more self-explanatory approach to define a variable above like `const showGA4Toggle = isGA4Enabled && select( CORE_MODULES ).isModuleConnected( 'analytics-4' ) === select( CORE_MODULES ).isModuleConnected( 'analytics' )`

assets/js/modules/analytics/components/common/UseUAandGA4SnippetSwitches.js
```diff
@@ -26,19 +26,35 @@ import { __ } from '@wordpress/i18n';
...
 
-	const ga4ExistingTag = useSelect( ( select ) => select( MODULES_ANALYTICS_4 ).getExistingTag() );
-	const ga4MeasurementID = useSelect( ( select ) => select( MODULES_ANALYTICS_4 ).getMeasurementID() );
-	const useGA4Snippet = useSelect( ( select ) => select( MODULES_ANALYTICS_4 ).getUseSnippet() );
+	const {
+		ga4ExistingTag,
+		ga4MeasurementID,
+		useGA4Snippet,
+	} = useSelect( ( select ) => {
+		if ( ! isGA4Enabled || select( CORE_MODULES ).isModuleConnected( 'analytics-4' ) === false ) {
```
felixarntz, line 48: Related to what you did above, I think it would technically be more appropriate to do the following check here:
```suggestion
		if ( ! isGA4Enabled || select( CORE_MODULES ).isModuleConnected( 'analytics-4' ) !== select( CORE_MODULES ).isModuleConnected( 'analytics' ) ) {
```

Not strictly necessary right now, but e.g. if we wanted to show these toggles in the setup flow too it would become a problem.

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -177,6 +177,13 @@ const baseActions = {
 				return;
 			}
 
+			// Do nothing if UA is already connected but GA4 is not.
+			const isUAConnected = registry.select( CORE_MODULES ).isModuleConnected( 'analytics' );
+			const isGA4Connected = registry.select( CORE_MODULES ).isModuleConnected( 'analytics-4' );
+			if ( isUAConnected && ! isGA4Connected ) {
```
felixarntz, line 183: Hmm, here you have a different check than before. Is there a particular reason for this? I don't think the conditions need to differ.

Technically, the one you have here is actually more accurate even than `isUAConnected !== isGA4Connected` - since the scenario that UA is not connected while GA4 is right now is not really possible, and we can technically ignore it.

Maybe we could have a registry selector for this whole thing rather than duplicating checks (which can be error-prone as in using slightly different checks)? How about something like a `showGA4UI` selector in the `modules/analytics` store, which checks the following:
* is the `ga4setup` feature flag active?
* is the `analytics` module not connected _or_ is the `analytics-4` module connected?

That would nicely define the logic in one place.

felixarntz (commented):
Great work @eugene-manuilov! Technically this all works well, I have a few suggestions on slightly inconsistent logic, which we could combine in a single place for DRY.

assets/js/modules/analytics/components/common/UseUAandGA4SnippetSwitches.js
eugene-manuilov: Updated.

assets/js/modules/analytics/components/common/UseUAandGA4SnippetSwitches.js
eugene-manuilov: Updated.

assets/js/modules/analytics/datastore/accounts.js
eugene-manuilov: Added the `canUseGA4Controls` selector and updated my changes to rely on it.

assets/js/modules/analytics/datastore/settings.js
```diff
@@ -98,12 +98,13 @@ export async function submitChanges( { select, dispatch } ) {
...
 
-	if ( isFeatureEnabled( 'ga4setup' ) ) {
-		if ( select( MODULES_ANALYTICS_4 ).haveSettingsChanged() ) {
-			const { error } = await dispatch( MODULES_ANALYTICS_4 ).submitChanges();
-			if ( isPermissionScopeError( error ) ) {
-				return { error };
-			}
+	// Try to save GA4 settings only if the Analytics-4 module has the same connection status as the Analytics module has.
+	// This check is needed to prevent saving GA4 settings unintentionally for cases when the Analytics module has been
+	// already connected by the time the GA4 functionality is revealed.
```
felixarntz, line 103: This comment is now no longer accurate. It might be okay to just remove it.

assets/js/modules/analytics/datastore/settings.js
```diff
@@ -165,7 +166,10 @@ export function validateCanSubmitChanges( select ) {
 	// Do existing tag check last.
 	invariant( hasExistingTagPermission() !== false, INVARIANT_INSUFFICIENT_TAG_PERMISSIONS );
 
-	if ( isGA4Enabled ) {
+	// Validate GA4 settings only if the Analytics-4 module has the same connection status as the Analytics module has.
+	// This check is needed to prevent saving GA4 settings unintentionally for cases when the Analytics module has been
+	// already connected by the time the GA4 functionality is revealed.
```
felixarntz, line 171: See above.

assets/js/modules/analytics/datastore/setup-flow.js
```diff
@@ -92,6 +100,25 @@ const baseSelectors = {
 		// There are UA and GA4 properties, so use the transitional mode.
 		return SETUP_FLOW_MODE_GA4_TRANSITIONAL;
 	} ),
+
+	/**
+	 * Determines whether we can use GA4 controls or not.
```
felixarntz, line 105: Maybe:
```suggestion
	 * Determines whether GA4 controls should be displayed or not.
```

felixarntz (commented):
@eugene-manuilov Thanks for the updates. This looks good, just a few quick clean-up changes around comments that are now no longer relevant.

assets/js/modules/analytics/datastore/settings.js
eugene-manuilov: Removed.

assets/js/modules/analytics/datastore/settings.js
eugene-manuilov: Removed.

felixarntz (approved):
Great!

Issue #3706, opened 2021-07-14
Analytics setup screen loads continuously during setup - Uncaught (in promise) TypeError: Cannot read property 'replace' of undefined


jamesozzie:
## Bug Description

There have been 3 recent reports of the below console errors appearing, when trying to setup the Analytics module. The setup screen itself never progresses, remaining in a loading state:
`Uncaught (in promise) TypeError: Cannot read property 'replace' of undefined`

All sites using SK 1.36.0 with a first time Site Kit setup. This occurs after clicking on the "continue module setup" or when connecting Analytics for the first time via Site Kit. 

- https://wordpress.org/support/topic/analytics-not-connecting-after-new-site-uploaded/#post-14659590 - **Open** - [SH info](https://docs.google.com/forms/d/1lhjns2K7c0Ny7ME4gxTcsnYtNUtcMr8BvFpPQi23Y1E/edit?urlBuilderDomain=google.com#response=ACYDBNjEXYowVdi8kktvInzqGnhm6HWmlTJOomRSUyDKiYnkjXUfFD2whOGdzZc0vq2pRFU)
- https://wordpress.org/support/topic/connect-analytics-in-site-kit-loads-infinitely/ - **Open** - [SH info](https://docs.google.com/forms/d/1lhjns2K7c0Ny7ME4gxTcsnYtNUtcMr8BvFpPQi23Y1E/edit?urlBuilderDomain=google.com#response=ACYDBNjfQn1fJJBtaJwBRGYLvVGYF0cwNzXhJsIpjeTwJr7NL2PoMlz9b3IbUeBdSLL34jc)
- https://wordpress.org/support/topic/cant-activate-google-analytics/#post-14659801 - **Open** - [SH info](https://docs.google.com/forms/d/1lhjns2K7c0Ny7ME4gxTcsnYtNUtcMr8BvFpPQi23Y1E/edit#response=ACYDBNgEHqurQt0oLegbWnxUhyIeTSxGhLzc6LqGwdAdGq7UMgWotJDDV2vxZ3x1vnPFaHo)
[...]

PR #3711, opened 2021-07-14
Bug/3706 replace undefined

eugene-manuilov:
## Summary

Addresses issue #3706

## Checklist

### DIFF
--- a/assets/js/modules/analytics-4/datastore/properties.js
+++ b/assets/js/modules/analytics-4/datastore/properties.js
@@ -240,7 +240,9 @@ const baseActions = {
 	 */
 	*matchPropertyByURL( properties, url ) {
 		const registry = yield commonActions.getRegistry();
-		const urls = ( Array.isArray( url ) ? url : [ url ] ).map( normalizeURL );
+		const urls = ( Array.isArray( url ) ? url : [ url ] )
+			.filter( ( item ) => item?.length > 0 )
+			.map( normalizeURL );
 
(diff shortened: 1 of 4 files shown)


eugene-manuilov:
@googlebot I consent.

assets/js/modules/analytics-4/datastore/properties.js
```diff
@@ -240,7 +240,9 @@ const baseActions = {
 	 */
 	*matchPropertyByURL( properties, url ) {
 		const registry = yield commonActions.getRegistry();
-		const urls = ( Array.isArray( url ) ? url : [ url ] ).map( normalizeURL );
+		const urls = ( Array.isArray( url ) ? url : [ url ] )
+			.filter( ( item ) => item?.length > 0 )
```
felixarntz, line 244: Maybe use `typeof item === 'string'` instead? Things that aren't strings can still have a `length` greater than 0.

felixarntz (commented):
One comment, other than that looks good. I think we should include it in 1.37.0.

assets/js/modules/analytics-4/datastore/properties.js
aaemnnosttv: We should also add a guard to `normalizeURL` which doesn't have a safe fallback if given another type. https://github.com/google/site-kit-wp/blob/371459e27494c3c2b083249c79af390a7f1139d0/assets/js/util/urls.js#L52-L54

eugene-manuilov: Updated.

felixarntz (approved):
Great!

Issue #2054, opened 2020-09-17
The plugin should not send any data if the wp_get_environment_type() is detected as anything other than production


lukecav:
https://make.wordpress.org/core/2020/07/24/new-wp_get_environment_type-function-in-wordpress-5-5/
https://developer.wordpress.org/reference/functions/wp_get_environment_type/

No data should be sent to Google if the `WP_ENVIRONMENT_TYPE` system variable is not production.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* If the function `wp_get_environment_type()` exists (we should _not_ polyfill this one) and it it returns anything other than "production", all Site Kit tags (Analytics, AdSense, Tag Manager) that collect data should not be rendered.

## Implementation Brief

* create a new `Tag_Production_Guard` common tag guard class similar to `Tag_Verify_Guard` in the same `Core/Tags/Guards` location
* in the `can_activate` method of the new class use the function `wp_get_environment_type()` to return `TRUE` if the environment is `production` and the function exists
* also return `TRUE` if the function does not exist
* update the `register_tag` methods used by Adsense, Analytics, Analytics 4 and Tag Manager to include the new `Tag_Production_Guard`.  For example the Adsense code is in `Modules/Adsense.php`.  The additional code will be something like `$tag->use_guard( new Tag_Production_Guard() );`

### Test Coverage

Add tests to a new `Core/Tags/Guards/Tag_Product_GuardTest` file.
* if the `wp_get_environment_type()` function does not exist test that `TRUE` is returned
* if the functions does exist test what the tag guard returns based on whether `WP_ENVIRONMENT_TYPE` is set to `production` or not.  The environment variable can be set using `putenv`:
  * `putenv( 'WP_ENVIRONMENT_TYPE=development' )` sets it to `development`
  * `putenv( 'WP_ENVIRONMENT_TYPE' )` unsets the variable

### Visual Regression Changes

* n/a

## QA Brief

* Take a look at `Setting the environment type
` https://make.wordpress.org/core/2020/07/24/new-wp_get_environment_type-function-in-wordpress-5-5/
* Enable any of the modules that enable tag guard (e.g. Adsense, Analytics, Analytics 4 and Tag Manager )
* Set your environment type to `local` and save the page source for the home page of your site that end users see (e.g. http://sitekit.test) to `local.txt`
* Change it back to `production` and save the page source for the home page to `production.txt`
[...]

PR #3734, opened 2021-07-19
Add tag production env guard.

ivankruchkoff:
## Summary

Addresses issue #2054 

## Relevant technical choices
As we can't override the env on a per test basis due to https://github.com/WordPress/WordPress/blob/f5713b8d717b398efe7de40c869f77543fb2068e/wp-includes/load.php#L195

We instead define a function in the namespace to provide custom values for our tests.
https://github.com/google/site-kit-wp/pull/3734/files#diff-cd9bfad4b0a994c130af2d7fce75605bab9e19e99ccea0a3124f6ca73e69c4f4R58

## Checklist


eugene-manuilov (changes requested):
@ivankruchkoff have you seen the [`uopz_set_static`](https://www.php.net/manual/en/function.uopz-set-static.php) function? Seems like we can use it to reset static variable inside of the `wp_get_environment_type` function, can't we? Can you try it on your end and if it works update your tests to use it?

ivankruchkoff:
> @ivankruchkoff have you seen the [`uopz_set_static`](https://www.php.net/manual/en/function.uopz-set-static.php) function? Seems like we can use it to reset static variable inside of the `wp_get_environment_type` function, can't we? Can you try it on your end and if it works update your tests to use it?

Tried it and added it to our workflow.

tests/phpunit/integration/Core/Tags/Guards/Tag_Production_GuardTest.php
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Tag_Production_GuardTest
+ *
+ * @package   Google\Site_Kit\Tests\Core\Tags\Guards
+ * @copyright 2021 Google LLC
+ * @license   https://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
+ * @link      https://sitekit.withgoogle.com
+ */
+
+namespace Google\Site_Kit\Tests\Core\Tags\Guards {
```
eugene-manuilov, line 11: Let's remove curly brackets since we don't use them anywhere.
```suggestion
namespace Google\Site_Kit\Tests\Core\Tags\Guards;
```

tests/phpunit/integration/Core/Tags/Guards/Tag_Production_GuardTest.php
```diff
@@ -0,0 +1,52 @@
...
+			if ( ! function_exists( 'wp_get_environment_type' ) ) {
+				$this->markTestSkipped( 'Missing wp_get_environment_type() function.' );
+			}
+			if ( ! function_exists( 'uopz_set_static' ) ) {
+				$this->markTestSkipped( 'The uopz extension is not available.' );
+			}
+			$tagproduction = new Tag_Production_Guard();
+			uopz_set_static( 'wp_get_environment_type', array( 'current_env' => 'development' ) );
+			$this->assertFalse( $tagproduction->can_activate() );
+			uopz_set_static( 'wp_get_environment_type', array( 'current_env' => 'production' ) );
```
eugene-manuilov, line 48: Let's probably revert to original value instead of reverting to `production`:
```php
$env_type      = wp_get_environment_type();
$tagproduction = new Tag_Production_Guard();

uopz_set_static( 'wp_get_environment_type', array( 'current_env' => 'development' ) );
$this->assertFalse( $tagproduction->can_activate() );

uopz_set_static( 'wp_get_environment_type', array( 'current_env' => $env_type ) );
```

eugene-manuilov (changes requested):
That's awesome! Thanks, @ivankruchkoff! Added two minor comments and it should be good to go.

tests/phpunit/integration/Core/Tags/Guards/Tag_Production_GuardTest.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3734/commits/5d02b7481e6870acc3b4ba472ce463e116432196

tests/phpunit/integration/Core/Tags/Guards/Tag_Production_GuardTest.php
ivankruchkoff: Updated in https://github.com/google/site-kit-wp/pull/3734/commits/ff826d110dfc2bc829d668209619d53a8ad05fe0 but the way the other tests are written, if the value is anything but production, they will fail.

tests/phpunit/integration/Core/Tags/Guards/Tag_Production_GuardTest.php
```diff
@@ -0,0 +1,51 @@
...
+class Tag_Production_GuardTest extends TestCase {
+
+	public function test_can_activate_on_older_environments() {
+		// Pre WP-5.5.0
+		if ( ! function_exists( 'wp_get_environment_type' ) ) {
+			$this->markTestSkipped( 'Missing wp_get_environment_type() function.' );
+		}
+		$tagproduction = new Tag_Production_Guard();
+		$this->assertTrue( $tagproduction->can_activate() );
+	}
```
eugene-manuilov, line 25: I think if we want to test it for older versions of WP, we need to skip this test if the function exists, right?

tests/phpunit/integration/Core/Tags/Guards/Tag_Production_GuardTest.php
```diff
@@ -0,0 +1,51 @@
...
+
+	public function test_can_activate_in_production() {
+		// Post WP-5.5.0
+		if ( function_exists( 'wp_get_environment_type' ) ) {
+			$this->markTestSkipped( 'Testing legacy environments without wp_get_environment_type() function.' );
+		}
+		$tagproduction = new Tag_Production_Guard();
+		$this->assertTrue( $tagproduction->can_activate() );
+	}
```
eugene-manuilov, line 34: Here is the opposite situation, we need to skip this test if the function doesn't exist which means that it is the old version of WP.

tests/phpunit/integration/Core/Tags/Guards/Tag_Production_GuardTest.php
eugene-manuilov: Maybe let's rename the method name to be something like `test_can_activate_on_older_versions`?

tests/phpunit/integration/Core/Tags/Guards/Tag_Production_GuardTest.php
eugene-manuilov: Same here, maybe let's rename the method name to be something like `test_can_activate_on_post_5_5_version`?

eugene-manuilov (changes requested):
Ok, thanks, @ivankruchkoff. The last comments from me and it should good to go.

tests/phpunit/integration/Core/Tags/Guards/Tag_Production_GuardTest.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3734/commits/a81ec2efe6837400ef41a3b32d8f2d40b0dbb31a

tests/phpunit/integration/Core/Tags/Guards/Tag_Production_GuardTest.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3734/commits/a81ec2efe6837400ef41a3b32d8f2d40b0dbb31a

eugene-manuilov:
@ivankruchkoff looks like you missed two previous comments: https://github.com/google/site-kit-wp/pull/3734#discussion_r675531996 and https://github.com/google/site-kit-wp/pull/3734#discussion_r675532732. Please, take a look.

tests/phpunit/integration/Core/Tags/Guards/Tag_Production_GuardTest.php
ivankruchkoff: Updated commenting for both in https://github.com/google/site-kit-wp/pull/3734/commits/267859182074ef4640ca2902182371aeb6596854

First test `test_can_activate_on_post_5_5_version()` is for post 5.5.0 scenarios, and thus skips if `wp_get_environment_type()` is not defined.
Second test `test_can_activate_on_older_versions()` is for pre 5.5.0 scenarios, and thus skips if `wp_get_environment_type()` is defined.

ivankruchkoff:
> @ivankruchkoff looks like you missed two previous comments: [#3734 (comment)](https://github.com/google/site-kit-wp/pull/3734#discussion_r675531996) and [#3734 (comment)](https://github.com/google/site-kit-wp/pull/3734#discussion_r675532732). Please, take a look.

Addressed, the suggested function names were inverted, so I added clarifying comments.

eugene-manuilov (approved):
Thanks, @ivankruchkoff. This looks great to me!

Issue #2217, opened 2020-10-20
Mark Google API requests from PHP client as user-specific


felixarntz:
Almost all Google APIs generally support a request query parameter called `quotaUser`, which allows passing a string that uniquely identifies a user of the application. This parameter is recommended to be used when issuing Google API requests from a server backend, as otherwise it may be impossible for the API to distinguish where the traffic is coming from and which rate limits it applies to.

The Site Kit authentication service makes use of this parameter because otherwise all requests would be associated with the same user which obviously would result in rate limits being exceeded at that scale. We didn't implement the same behavior in the plugin originally, however there could still be a case where it helps, e.g. if a site itself has too many users.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The plugin should (in PHP) pass the `quotaUser` query parameter _or_ the `X-Goog-Quota-User` HTTP header to every Google API request with a string that uniquely identifies this user (e.g. some combination of site URL and WordPress user ID). The string doesn't need to be immutable (e.g. if the site URL changes, that's okay), it's just important that it doesn't conflict with another Site Kit user's identifier.
* This should be handled centrally, not individually in every API request (potentially via a request middleware or some extended class).

## Implementation Brief

* Write a function to get the `quotaUser` query string param, by editing the following file `/includes/Core/Authentication/Clients/OAuth_Client.php` and adding the function: `private function get_quota_user()`
* The function will return a unique identifier for a user on a given site:
`return get_home_url() . '-' . get_current_user_id();`
* Next we need to hook into every http request and add a header to our request. As we're using Guzzle 5 library, we can't make use of middlewares, so instead we use the event emitter to hook into requests - event emitter documentation https://docs.guzzlephp.org/en/5.3/http-messages.html?highlight=emitter#event-emitter
* [Adding request headers](https://docs.guzzlephp.org/en/5.3/quickstart.html#headers)
* We do that by adding to the code here: https://github.com/google/site-kit-wp/blob/8c341d37f0ba40d83ce5a8a067ac223f42c4f9d8/includes/Core/Authentication/Clients/OAuth_Client.php#L237

```php
$http_client->getEmitter()->on( 'before', function ( BeforeEvent $before ) {
        $before->getRequest()->addHeader('X-Goog-Quota-User, $this->get_quota_user() );
} );
```

[Google Header Reference
](https://cloud.google.com/apis/docs/system-parameters)

[...]

PR #3737, opened 2021-07-19
Enhancement/2217 use specified requests

eugene-manuilov:
## Summary

Addresses issue #2217

## Checklist


includes/Core/Authentication/Clients/Google_Site_Kit_Client.php
```diff
@@ -243,4 +244,23 @@ protected function fetchAuthToken( OAuth2 $auth, callable $http_handler = null )
...
+	/**
+	 * Executes deferred HTTP requests.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param RequestInterface $request Request object to execute.
+	 * @param string           $expected_class Expected class to return.
+	 * @return object An object of the type of the expected class or Psr\Http\Message\ResponseInterface.
+	 */
+	public function execute( RequestInterface $request, $expected_class = null ) {
```
aaemnnosttv, line 257: Would you please move this to be grouped with the other public methods?

includes/Core/Authentication/Clients/Google_Site_Kit_Client.php
```diff
@@ -243,4 +244,23 @@ protected function fetchAuthToken( OAuth2 $auth, callable $http_handler = null )
...
+		$request = $request->withHeader(
+			'X-Goog-Quota-User',
+			sprintf( '%s-%s', get_home_url(), get_current_user_id() )
```
aaemnnosttv, line 260: This is probably fine but just a thought – what do you think about formatting it in a more structured way? i.e. `{scheme}{id}@{url-without-scheme}` like basic auth? e.g. `https://1@example.com` rather than `https://example.com-1`.

aaemnnosttv (commented):
Thanks @eugene-manuilov – This looks great! One tiny request and one question for your consideration.

includes/Core/Authentication/Clients/Google_Site_Kit_Client.php
eugene-manuilov: Yep, updated.

includes/Core/Authentication/Clients/Google_Site_Kit_Client.php
eugene-manuilov: Yep, i like it. Updated.

includes/Core/Authentication/Clients/Google_Site_Kit_Client.php
```diff
@@ -201,6 +203,48 @@ public function fetchAccessTokenWithRefreshToken( $refresh_token = null ) {
...
+	 * @since n.e.x.t
+	 *
+	 * @return string Unique user identifier.
+	 */
+	public static function getQuotaUser() {
+		$url        = wp_parse_url( get_home_url() );
+		$quota_user = "{$url['scheme']}://";
+
+		$user        = wp_get_current_user();
+		$quota_user .= $user instanceof WP_User && $user->ID
```
aaemnnosttv, line 233: `wp_get_current_user` always returns a `WP_User` so we can simplify this condition but also, there's no harm in using `0` here if there was no current user (e.g. cron or cli) which is the `ID` when no user is logged in, so we can always rely on user `ID`.

includes/Core/Authentication/Clients/Google_Site_Kit_Client.php
```diff
@@ -201,6 +203,48 @@ public function fetchAccessTokenWithRefreshToken( $refresh_token = null ) {
...
+			? "{$user->user_login}@{$url['host']}"
+			: $url['host'];
+
+		if ( ! empty( $url['port'] ) ) {
+			$quota_user .= ":{$url['port']}";
+		}
```
aaemnnosttv, line 239: I think this part probably isn't necessary as Site Kit doesn't support non-default ports.

includes/Core/Authentication/Clients/Google_Site_Kit_Client.php
```diff
@@ -201,6 +203,48 @@ public function fetchAccessTokenWithRefreshToken( $refresh_token = null ) {
...
+	/**
+	 * Returns a string that uniquely identifies a user of the application.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string Unique user identifier.
+	 */
+	public static function getQuotaUser() {
+		$url        = wp_parse_url( get_home_url() );
+		$quota_user = "{$url['scheme']}://";
```
aaemnnosttv, line 230: Rather than building up a string in pieces with conditions for if a part exists or not, I think we could simply parse out the scheme, host, and path and then put them all together in one line rather than many.

tests/phpunit/integration/Core/Authentication/Clients/Google_Site_Kit_ClientTest.php
```diff
@@ -0,0 +1,69 @@
...
+			'basic home url'                      => array(
+				'https://example.com',
+				'user111',
+				'https://user111@example.com',
+			),
+			'home url with port and subdirectory' => array(
+				'https://example.com:8080/subdirectory',
+				'testuser',
+				'https://testuser@example.com:8080/subdirectory',
+			),
```
aaemnnosttv, line 60: Subdirectory makes sense, but again, I don't think we need to account for custom ports yet :)

aaemnnosttv (commented):
Thanks @eugene-manuilov – this is almost there. See my comments below 👍

includes/Core/Authentication/Clients/Google_Site_Kit_Client.php
```diff
@@ -201,6 +203,48 @@ public function fetchAccessTokenWithRefreshToken( $refresh_token = null ) {
...
+
+		$user        = wp_get_current_user();
+		$quota_user .= $user instanceof WP_User && $user->ID
+			? "{$user->user_login}@{$url['host']}"
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

includes/Core/Authentication/Clients/Google_Site_Kit_Client.php
eugene-manuilov: Updated.

includes/Core/Authentication/Clients/Google_Site_Kit_Client.php
eugene-manuilov: Updated.

includes/Core/Authentication/Clients/Google_Site_Kit_Client.php
eugene-manuilov: Updated.

tests/phpunit/integration/Core/Authentication/Clients/Google_Site_Kit_ClientTest.php
eugene-manuilov: Updated.

aaemnnosttv (approved):

Issue #3502, opened 2021-06-08
Configure eslint-plugin-import 


danielgent:
## Feature Description

We have `eslint-plugin-import` installed but it doesn't look like it's configured

I created a branch here where I configured this https://github.com/google/site-kit-wp/tree/eslint-import-idea (you can see from the diff that it even found one error :smile: )

The remaining errors on this branch that need fixing are:
```
error  Multiple exports of name 'getLocale'  import/export
error  Unable to resolve path to module 'tests/js/utils'  import/no-unresolved
error  Unable to resolve path to module 'assets/js/googlesitekit/datastore/site'  
```

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The ESLint configuration should be modified to extend `plugin:import/recommended`.  
_The plugin is already installed_
* Any errors or warnings raised as a result should be fixed

## Implementation Brief

* Merge linked PR https://github.com/google/site-kit-wp/pull/3745

### Test Coverage

* There should be no changes and no updates to tests needed (apart from those done to fix new linting rules!)

### Visual Regression Changes

* There should be no changes

## QA Brief

* Check that JS linting is passing correctly

## Changelog entry

* N/A

danielgent:
Just to give an example of a task I've done a few times.  I need to delete a load of files. Instead of vscode then using eslint to syntax highlight, I have to leave it building to then open the files.  It's possible, but it's just way nicer to use eslint

![no import](https://user-images.githubusercontent.com/8496063/123455042-86bfb280-d5e1-11eb-899a-837d59c111b6.png)

Also it means that I can then make a commit with a bad import that will break the build
![failing commit checked in](https://user-images.githubusercontent.com/8496063/123455068-90e1b100-d5e1-11eb-84a4-4ede630ebab5.png)

aaemnnosttv:
Thanks @danielgent – this looks like a nice enhancement to add 👍 

Looking at your branch, is there a reason why you chose the import plugins for `errors` and `warnings` over the [`recommended`](https://github.com/benmosher/eslint-plugin-import/blob/1012eb951767279ce3b540a4ec4f29236104bb5b/config/recommended.js) ruleset? It looks like that one has all the same rules in one. Any reason not to use that instead?

danielgent:
No @aaemnnosttv .  Most likely I quickly read the readme incorrectly :smile:   

Just doing `- plugin:import/recommended` is enough

danielgent:
[...]

PR #3745, opened 2021-07-20
Infrastructure/3502 configure eslint plugin import

danielgent:
## Summary

Addresses issue #3502

## Checklist


.eslintrc.json
```diff
@@ -3,7 +3,9 @@
     "plugin:@wordpress/eslint-plugin/recommended-with-formatting",
     "plugin:jest/recommended",
     "plugin:sitekit/main",
-    "plugin:react-hooks/recommended"
+    "plugin:react-hooks/recommended",
+    "plugin:import/errors",
+    "plugin:import/warnings"
```
danielgent, line 8: As commented on ticket https://github.com/google/site-kit-wp/issues/3502#issuecomment-883490130
`plugin:import/recommended` pulls in `comma-dangle` as an error.  Odd

.eslintrc.json
```diff
@@ -56,6 +58,18 @@
     ],
     "quotes": ["error", "single", { "allowTemplateLiterals": true, "avoidEscape": true }]
   },
+  "settings": {
+    "import/core-modules":
```
danielgent, line 62: This ignores the followin modules

.eslintrc.json
```diff
@@ -56,6 +58,18 @@
     ],
     "quotes": ["error", "single", { "allowTemplateLiterals": true, "avoidEscape": true }]
   },
+  "settings": {
+    "import/core-modules": 
+      [
+        "googlesitekit-data",
+        "googlesitekit-modules",
+        "googlesitekit-api",
+        "googlesitekit-widgets",
```
danielgent, line 67: I believe that we're removing these soon anyway

.eslintrc.json
```diff
@@ -56,6 +58,18 @@
...
+        "@wordpress/i18n__non-shim",
+        "@wordpress/api-fetch__non-shim",
```
danielgent, line 69: Is this ok? How come they aren't dependencies in package.json like the other @wordpress packages :thinking: 

I can look at what these are next

.eslintrc.json
```diff
@@ -56,6 +58,18 @@
...
+        "tests/js/utils"
```
danielgent, line 70: This is a quick workaround to not get lots of errors. Next, I'll look at resolvers

This workaround could work well as these are only used in tests *and* the tests simply won't run if not correctly imported *but* it's nice to have this plugin working to speed up development time when writing tests :smile:

.eslintrc.json
```diff
@@ -56,6 +58,22 @@
...
+        "googlesitekit-data",
+        "googlesitekit-modules",
+        "googlesitekit-api",
+        "googlesitekit-widgets",
+        "@wordpress/i18n__non-shim",
+        "@wordpress/api-fetch__non-shim"
+      ],
+      "import/resolver": {
+        "alias": [
+          ["tests/js/utils", "./tests/js/utils"]
```
danielgent, line 73: I was researching this for a while. I had to install another package `eslint-import-resolver-alias` and then add this config

This was the most popular way to do this online :smile:

assets/js/googlesitekit-datastore-site.js
```diff
@@ -20,6 +20,6 @@
  * Internal dependencies
  */
 import Data from 'googlesitekit-data';
-import { registerStore } from 'assets/js/googlesitekit/datastore/site';
+import { registerStore } from './googlesitekit/datastore/site';
```
danielgent, line 23: I could add a resolver for this (like above) but it seems more sensible to just change the one reference!

assets/js/googlesitekit/datastore/user/feature-tours.test.js
```diff
@@ -26,7 +26,8 @@ import {
 } from '../../../../../tests/js/utils';
 import * as CacheModule from '../../../googlesitekit/api/cache';
 import { STORE_NAME } from './constants';
-import initialState, {
+import {
+	initialState,
```
danielgent, line 30: I'm surprised this worked at all! It was importing the default export but `eslint-plugin-import` warned me
`warning  Using exported name 'initialState' as identifier for default export  import/no-named-as-default`

assets/js/components/higherorder/withData.js
```diff
@@ -36,7 +36,7 @@ import Data from 'googlesitekit-data';
 import { getModulesData, stringifyObject } from '../../util';
 import getNoDataComponent from '../legacy-notifications/nodata';
 import getDataErrorComponent from '../legacy-notifications/data-error';
-import getSetupIncompleteComponent, { getModuleInactiveComponent } from '../legacy-notifications/setup-incomplete';
+import { getSetupIncompleteComponent, getModuleInactiveComponent } from '../legacy-notifications/setup-incomplete';
```
danielgent, line 39: `eslint-plugin-import` warns about importing a default export with the same name as a named export

assets/js/components/legacy-notifications/setup-incomplete.js
```diff
@@ -41,8 +41,6 @@ export const getSetupIncompleteComponent = ( module, inGrid = false, fullWidth =
 	return ctaWrapper( cta, inGrid, fullWidth, createGrid );
 };
 
-export default getSetupIncompleteComponent;
```
danielgent, line 44: I know we're deleting this soon but I thought I'd fix it anyway :smile:

.eslintrc.json
aaemnnosttv: Not the above `googlesitekit-` ones, no.

.eslintrc.json
aaemnnosttv: This should be replaced with a normal relative import; it's an old import that we used in the past before we moved everything (except above) to relative imports.

assets/js/googlesitekit-datastore-site.js
aaemnnosttv: Yes, this is the correct change to make. Thanks!

assets/js/googlesitekit/datastore/user/feature-tours.test.js
aaemnnosttv: Probably a weird auto import that wasn't caught. Thanks!

package.json
```diff
@@ -176,6 +176,7 @@
     "css-loader": "^3.6.0",
     "cssnano": "^4.1.10",
     "eslint": "^7.20.0",
+    "eslint-import-resolver-alias": "^1.1.2",
```
aaemnnosttv, line 179: As mentioned above, we shouldn't need this.

assets/js/components/legacy-notifications/setup-incomplete.js
aaemnnosttv: Assuming this isn't used or only as a named import?

.eslintrc.json
aaemnnosttv: Very odd. I'd like to see if we can figure out why that is happening to stick with `recommended` but we don't need to spend too much time on it.

aaemnnosttv (commented):
Thanks @danielgent – a few thoughts here for you since this is still technically in IBR.

.eslintrc.json
tofumatt: As mentioned [here](https://github.com/google/site-kit-wp/issues/3502#issuecomment-884168941) that's not the case, so let's use the recommended plugin instead.

.eslintrc.json
danielgent: @aaemnnosttv what are these modules? They 404 on NPM :thinking:

.eslintrc.json
danielgent: Sounds great @aaemnnosttv 

Aliases are a great idea in principle but they cause problems in storybook and testing and etc :smile:

danielgent: This rule is removed and all failing imports have been changed to relative @aaemnnosttv  :+1:

assets/js/components/legacy-notifications/setup-incomplete.js
danielgent: Exactly that @aaemnnosttv . The two named exports are only imported once

only-used-once

`import { getSetupIncompleteComponent, getModuleInactiveComponent } from '../legacy-notifications/setup-incomplete';`

package.json
danielgent: Removed

.eslintrc.json
danielgent: Ok @aaemnnosttv 

I can see them in the webpack config e.g. `'googlesitekit-data': './assets/js/googlesitekit-data.js',`

It looks like the default import is always imported.  I could try and put all those aliases into `eslint-plugin-import` (and put the eslint resolver plugin back in)

Or we can just leave them like this and ignore them. I figure webpack will not build if this file changes

tofumatt: Sorry, I don't quite understand this comment. These are being ignored because the ESLint plugin can't resolve the path, correct?

That seems fine, but are you saying you can add the paths to the `eslint-plugin-import` config? If so that'd be arguably better—you're right, if we remove these files the plugin/webpack build will blow up, but no harm in having ESLint be aware of these paths rather than ignoring them, if it's just as easy?

assets/js/components/DeviceSizeTabBar.js
```diff
@@ -78,7 +78,7 @@ const DeviceSizeTabBar = ( {
 						{ icon }
 					</Tab>
 				);
-			}
+			},
```
tofumatt, line 81: Huh, that's kinda weird. I get it, but it's unexpected to me...

.eslintrc.json
tofumatt: These are shims we need to use to configure the WordPress API and I18n packages: https://github.com/google/site-kit-wp/blob/1757a849eae1e77e13a2a092eaaa645d3c14dff8/webpack.config.js#L120-L122

They're named this way, with aliases, I suppose to make it clear what they're referencing/working with, but I suppose that makes it a bit unclear too 😞 

Anyway, they'd also be great to reference in the ESLint config as the paths specified in the Webpack config like the `googlesitekit-*` JS files, but otherwise ignoring them is fine.

.eslintrc.json
tofumatt: Is this still an issue or did you get the resolvers working? 🙂

tofumatt (approved):
I had a look into what we'd gain from aliasing the `googlesite-*` modules and it's not much actually, so let's ship this 👍🏻

danielgent:
Cheers @tofumatt . I'll reply anyway for completeness even though the ticket has been merged (as I did push for this) :smile:

danielgent:
> I had a look into what we'd gain from aliasing the `googlesite-*` modules and it's not much actually, so let's ship this 👍🏻

Yep! My thoughts exactly.  It's imported in lots of places the same way.

assets/js/components/DeviceSizeTabBar.js
danielgent: It's the `comma-dangle` rule. Just say the word and we can get rid of this rule @tofumatt !  

I don't like it, but I do like consistency and not thinking bout this kind of stuff :smile: +

.eslintrc.json
danielgent: Yes @tofumatt , have a look at this 

https://github.com/google/site-kit-wp/pull/3745/commits/701c180e324cb67e69ef92a1aa94835c74e2c2db#diff-6884918dc8291219be508e05e28965b958c734def85324f3b53858ea4702090fR73

On a previous commit I installed `eslint-import-resolver-alias` and configured this for `test/js/utils` (but Evan said better to just make all those relative imports and drop this alias)

I should be able to just do some config like the following though (if you change your mind)

```
    "import/resolver": {
        "alias": [
          ["googlesitekit-data". "./assets/js/googlesitekit-data.js"]
 ```

Issue #990, opened 2019-12-19
Add a toggle to enable/disable the Site Kit admin bar feature


jamesozzie:
## Feature Description

As per [this](https://wordpress.org/support/topic/site-kit-admin-bar/#post-12251754) WordPress support topic some users may want to remove the Site Kit admin bar feature. A toggle to enable the Site Kit admin bar feature may be something to consider. 

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new Site Kit-wide setting `googlesitekit_admin_bar_menu_enabled` should be registered, which is a boolean with default value of `true`.
* If the setting is `false`, the admin bar menu of Site Kit should not be displayed (under any circumstance, so this should be the first check).
* In the _Admin Settings_ tab on the Site Kit settings page, a new box should be displayed on the right side of the existing _Plugin Status_ box, which should be shrinked to half width on desktop (i.e. the new box should also be half width on desktop).
    * The new box should look similar to the existing _Tracking_ box.
    * It should have the title _Plugin Settings_.
    * It should for now only contain a single toggle _Display relevant page stats in the Admin bar_, which controls the new setting.

## Implementation Brief
* Using `assets/js/components/settings/SettingsAdmin.js`,
    * Update the markup for the `Plugin Status` section to be half width replacing the html markup by the `Grid`, `Row` and `Cell` components.
    * On small screens, the `Plugin Status` should be full width, and then half width on tablet and up.
    * Add another cell adjacent to the `Plugin Status` one, 100% wide on small screens and 50% on tablet and up. The cell will host the `SettingsPlugin` component. More below.
* Create `includes/Core/Admin_Bar/Admin_Bar_Settings.php` which contains the `Admin_Bar_Settings` class that extends the `Setting` class and registers the `googlesitekit_admin_bar_menu_enabled` option name. Default value `true`.
    * Code example: `includes/Core/Authentication/Owner_ID.php`
* Using `includes/Core/Admin_Bar/Admin_Bar.php`,
    * In the `register` method,
        * create a new object from `Admin_Bar_Settings` which will register the new setting.
        * using the `googlesitekit_rest_routes` action, add a filter to add the following REST route `core/site/data/admin_bar`.
            * Code example: `includes/Core/Util/Tracking.php`.
            * In the `WP_REST_Server::CREATABLE` callback, the value for the setting should be toggled.
[...]

PR #3750, opened 2021-07-21
Enhancement/990 disable admin bar

eugene-manuilov:
## Summary

Addresses issue #990

## Checklist


eugene-manuilov:
@googlebot I consent.

assets/js/components/settings/SettingsAdmin.js
```diff
@@ -70,150 +71,97 @@ const SettingsAdmin = () => {
...
 											{ __( 'Site Kit is connected', 'google-site-kit' ) }
-											<span className="
-												googlesitekit-settings-module__status-icon
-												googlesitekit-settings-module__status-icon--connected
-											">
+											<span className="googlesitekit-settings-module__status-icon googlesitekit-settings-module__status-icon--connected">
 												<VisuallyHidden>
 													{ __( 'Connected', 'google-site-kit' ) }
 												</VisuallyHidden>
 											</span>
```
tofumatt, line 119: A bit unrelated to the changes being made, but this icon doesn't need "alt" text like this and we should remove it.

The status is already read above ("Site Kit is connected"), so this text is just floating here without context and would result in a screen reader reading: "Site Kit is connected Connected".

A follow-up issue might be better, as I see this pattern in two files in the code.

tofumatt (changes requested):
Looks great. I think it'd be nice to set the Admin Bar as enabled for the Storybook stories/reference images. If it isn't a huge pain, just because it's nice when the stories/references have an "expected" default.

This also needs a QA Brief, but after that we can merge this, and I think it'll be fine to add to the release 👍🏻

assets/js/components/settings/SettingsAdmin.js
tofumatt: Filed #3763 🙂

eugene-manuilov:
@tofumatt, storybook is updated!

tofumatt (approved):
Nice!

(The CLA error looks to be because there's a merge commit or something, it's safe to ignore.)

tofumatt (changes requested):
Oops, can you add a QA brief first?

Then we can merge this 👍🏻

tofumatt (approved):
(Actually, I'm marking this as approved because once it has a QA Brief it's fine to merge, so if you want to merge it after the QA Brief is added that's cool, no need for me to review the code again 🙂 )

Issue #2258, opened 2020-10-23
Remove legacy data API and legacy higher-order components


felixarntz:
The legacy data API and the legacy higher-order components `withData` and `addToFilter` should be removed once they are no longer used.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* `assets/js/components/data` should be entirely removed. Before, it should be verified it is no longer used anywhere.
* `assets/js/components/higherorder` should be entirely removed. Before, it should be verified it is no longer used anywhere.
* The legacy infrastructure for mocking Storybook data should be entirely removed. E.g.:
    * `.storybook/data/*`
    * `assets/js/util/cache-data.js`
* Any Storybook stories that use data from `.storybook/data/*` should be removed. Those should be individually listed in the IB, so that decisions whether they can just be removed or whether they need to be replaced can be made.

### 1. Remove `assets/js/components/data`
**collect-module-data.js**
  - Remove the file
  - Then, in `assets/js/components/Root/index.js`:
	- Remove `<CollectModuleData>`
	- Remove legacy  `dataAPI*` props
	- Go through the consumers of `<Root>`, find instances where these props are passed, and remove them:
		- assets/js/googlesitekit-user-input.js    	
		- assets/js/googlesitekit-dashboard.js    
		- assets/js/googlesitekit-dashboard-splash.js    
		- assets/js/googlesitekit-activation.js    
		- assets/js/googlesitekit-adminbar.js    
		- assets/js/googlesitekit-settings.js    
		- assets/js/googlesitekit-dashboard-details.js    
		- assets/js/googlesitekit-module.js    
		- assets/js/googlesitekit-wp-dashboard.js

**invalidate-cache-group.js** 
  - Remove the file and all usage from the following:
    - assets/js/modules/search-console/datastore/settings.js
    - assets/js/modules/tagmanager/datastore/settings.js
    - assets/js/modules/analytics/datastore/settings.js 
    - assets/js/googlesitekit/data/create-settings-store.js
  - double check the codebase and remove any remaining instances.

**constants.js**
Remove the file, and delete the imports from: 
-   assets/js/modules/analytics/datastore/settings.js
-   assets/js/modules/tagmanager/datastore/settings.js
-   assets/js/modules/search-console/datastore/settings.js
-   assets/js/googlesitekit/data/create-settings-store.js

**cache.js**
[...]

PR #3765, opened 2021-07-27
Remove legacy data API and legacy higher-order components (2258).

johnPhillips:
## Summary

Addresses issue #2258

## Checklist


assets/js/components/legacy-notifications/util.js
```diff
@@ -87,10 +88,15 @@ const removeDismissed = ( notifications ) => {
 		return notifications;
 	}
 
-	return notifications.filter( ( notification ) => {
-		const dismissed = getCache( `notification::dismissed::${ notification.id }` );
+	const promises = await notifications.map( ( notification ) => getItem( `notification::dismissed::${ notification.id }` ) );
```
aaemnnosttv, line 91: Probably an oversight but these only need to be awaited below by `Promise.all`.
```suggestion
	const promises = notifications.map( ( notification ) => getItem( `notification::dismissed::${ notification.id }` ) );
```

assets/js/components/legacy-notifications/util.js
```diff
@@ -87,10 +88,15 @@ const removeDismissed = ( notifications ) => {
...
+	const allNotifications = await Promise.all( promises );
+
+	const notDismissed = notifications.filter( ( _, index ) => {
+		const dismissed = allNotifications[ index ];
```
aaemnnosttv, line 95: This is not quite accurate as `getItem` will always resolve an object: https://github.com/google/site-kit-wp/blob/3db19cbeca41145aefc6c6c3383ee9b66e8b4121/assets/js/googlesitekit/api/cache.js#L184-L196

```suggestion
		const dismissed = allNotifications[ index ].cacheHit;
```

assets/js/components/legacy-notifications/util.js
```diff
@@ -87,10 +88,15 @@ const removeDismissed = ( notifications ) => {
 		return notifications;
 	}
 
-	return notifications.filter( ( notification ) => {
-		const dismissed = getCache( `notification::dismissed::${ notification.id }` );
+	const promises = await notifications.map( ( notification ) => getItem( `notification::dismissed::${ notification.id }` ) );
+	const allNotifications = await Promise.all( promises );
+
+	const notDismissed = notifications.filter( ( _, index ) => {
```
aaemnnosttv, line 94: IMO it's more clear to return the result of the filtered notifications here than `notDismissed` .
```suggestion
	return notifications.filter( ( _, index ) => {
```

assets/js/components/legacy-notifications/util.js
```diff
@@ -87,10 +88,15 @@ const removeDismissed = ( notifications ) => {
 		return notifications;
 	}
 
-	return notifications.filter( ( notification ) => {
-		const dismissed = getCache( `notification::dismissed::${ notification.id }` );
+	const promises = await notifications.map( ( notification ) => getItem( `notification::dismissed::${ notification.id }` ) );
+	const allNotifications = await Promise.all( promises );
```
aaemnnosttv, line 92: Maybe something like `notificationDismissals` would be a bit more clear here since the value for each is only the cache entry for the notification's dismissed status.

assets/js/components/legacy-notifications/util.js
```diff
@@ -111,8 +117,8 @@ export async function getModulesNotifications() {
 		const promise = new Promise( async ( resolve ) => {
 			const { identifier } = module;
 
-			const notifications = removeDismissed(
-				await data.get( TYPE_MODULES, identifier, 'notifications', {}, false ),
+			const notifications = await removeDismissed(
+				await API.get( 'modules', identifier, 'notifications', {}, false ),
```
aaemnnosttv, line 121: As mentioned above, `nocache: false` is the same as the default of `useCache: true` which can be removed.
```suggestion
				await API.get( 'modules', identifier, 'notifications' ),
```

assets/js/components/legacy-setup/search-console.js
```diff
@@ -74,7 +74,7 @@ class SearchConsole extends Component {
 		const { setErrorMessage } = this.props;
 		( async () => {
 			try {
-				const properties = await data.get( TYPE_MODULES, 'search-console', 'matched-sites' );
+				const properties = await API.get( 'modules', 'search-console', 'matched-sites' );
```
aaemnnosttv, line 77: Here's another example which looks harmless but remember that the default value of the parameter controlling whether to use cache or not is not the same between the two.

The legacy dataAPI defaults to bypassing cache, while the modern API defaults to using cache **(only relevant for `.get`)**
https://github.com/google/site-kit-wp/blob/3db19cbeca41145aefc6c6c3383ee9b66e8b4121/assets/js/components/data/index.js#L240

So here we'd need to set `useCache: false` to be consistent, which is something we generally do for all requests made during set up.

assets/js/components/legacy-setup/site-verification.js
```diff
@@ -72,7 +72,7 @@ class SiteVerification extends Component {
 
 		( async () => {
 			try {
-				const { verified, identifier } = await data.get( TYPE_MODULES, 'site-verification', 'verification' );
+				const { verified, identifier } = await API.get( 'modules', 'site-verification', 'verification' );
```
aaemnnosttv, line 75: See above.

assets/js/components/Root/index.js
```diff
@@ -39,9 +38,6 @@ export default function Root( {
 	children,
 	registry,
 	viewContext = null,
-	// TODO: Remove legacy dataAPI prop support once phased out.
-	dataAPIContext,
-	dataAPIModuleArgs,
```
aaemnnosttv, line 44: The prop types for these are yet to be removed still 🙂

stories/settings.stories.js
```diff
@@ -35,13 +35,13 @@ import { __ } from '@wordpress/i18n';
...
...\/qwWDyxRyr+rBY+XP+Cna7f2Gfid\\\/1z0\\\/\\\/ANONtTtYepxn\\\/BIL\\\/kzmz\\\/7Dl9\\\/NKYHu\\\/wC2N\\\/yah8YP+xT1P\\\/0megDxz\\\/gkz\\\/yZT4X\\\/AOwhqH\\\/pS9AH2HIgkQqehoAjNrGTnYM5zmgQfZYz\\\/D+poGPSNY84HJ60APoAKACgAoAKACgAoA+W\\\/wDgp5\\\/yYz8Tv+uen\\\/8ApxtqAOK\\\/4JBf8mc2f\\\/Ycvv5pQB7v+2N\\\/yah8YP8AsU9T\\\/wDSZ6APHP8Agkz\\\/AMmU+F\\\/+whqH\\\/pS9AH11qVu91aSxJGJPMUoymZouDwcMoJB9xVJ2dzOabi0vzsZ8mgRahFH9utVMiIsYK3LsQFOQd2Ac+\\\/Wk99BxulZ\\\/5kt34etb65E80bO4Ocee4XoB06dv1Pqcosml0iCaaGVoR5kW0IwlYbcHIHH+SODkUAV5fDFhPPNK9ojPNnzN0jENn1B4\\\/wD1UAPn8O2dyYd9vxDH5SKs7hQmCMYHBGD\\\/AC9BQA6x0K3066M8EW2QoUyZWbIJyck8nnnn1PrQBoZf+6v\\\/AH1\\\/9agAy\\\/8AdX\\\/vr\\\/61ABl\\\/7q\\\/99f8A1qAFUtnkAD2Of6UAfLv\\\/AAU8\\\/wCTGfid\\\/wBc9P8A\\\/TjbUAcV\\\/wAEgv8Akzmz\\\/wCw5ffzSgD3f9sb\\\/k1D4wf9inqf\\\/pM9AHjn\\\/BJn\\\/kynwv8A9hDUP\\\/Sl6APsKVmVGKgM2OATjJobsByeo\\\/EzRNHuRbX161vcFJZBGbOViyxyGNiNoOcFSf8AdG7pzXPQrwxEeaHmvubT\\\/FCTT2DVPiboejanHp9zf5upGCBILSaYB\\\/m+UlFPJ2tgdTg+ldAzf1XU10aza6uXIgQjcYrd5CMkDO1cnGTyccDJOACapRcnaJEpKCuyfT7xL+zhuI5FljlQSI6ggMpGQcHnoRU7FJ3VyzQMKACgAoAKACgD5b\\\/4Kef8mM\\\/E7\\\/rnp\\\/8A6cbagDiv+CQX\\\/JnNn\\\/2HL7+aUAe8ftjf8mofGD\\\/sVNT\\\/APSZ6APG\\\/wDgkz\\\/yZT4X\\\/wCwhq
```
aaemnnosttv, line 44: Given the absolute ginormous size of this line, I think it also justifies having its own fixture file as mentioned before – ALTHOUGH I would try testing it with the former fixture used by `resetGlobals` to see if this one is needed at all – my guess is that it isn't. I don't think there is a substantial difference in the global here that is unique to the settings page.

aaemnnosttv (dismissed):
Thank you @johnPhillips – this PR is so wonderful to see – I thought I'd never see the day! 😄 

Overall, it looks great and definitely deserves one of these...

<details>
<summary>...</summary>

image

</details>

In all seriousness though, there are a few things that need adjusting (mostly around migrating `dataAPI` to `API`) but it should be fairly straightforward to address. Let me know if you have any questions 👍

assets/js/components/legacy-notifications/notification.js
```diff
@@ -22,68 +22,87 @@
...
-		this.cardRef = createRef();
-
-		this.handleDismiss = this.handleDismiss.bind( this );
-		this.handleCTAClick = this.handleCTAClick.bind( this );
-
-		if ( 0 < this.props.dismissExpires ) {
-			this.expireDismiss();
+import { getItem, setItem, deleteItem } from '../../googlesitekit/api/cache';
+
+function Notification( {
```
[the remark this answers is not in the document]
johnPhillips: The previous class component was calling `getCache` in its constructor. Since the cache methods are now async, it needed a workaround.
I discussed with @tofumatt and we decided to just move it over to a functional component and use hooks.

aaemnnosttv (commented):
Almost there – just a few things left to address 🙂

aaemnnosttv:
@googlebot I consent.

tests/js/utils.js
```diff
@@ -428,6 +428,29 @@ export const unsubscribeFromAll = () => {
...
 
+/**
+ * Creates a function that allows extra time for registry updates to have completed.
+ *
+ * @since n.e.x.t
+ *
+ * @param {Object} registry WP data registry instance.
+ * @return {Function} Function to await all registry updates since creation.
+ */
+export const createWaitForRegistry = ( registry ) => {
```
aaemnnosttv, line 439: This ended up being unnecessary, but I think it's still worth keeping as this will likely come up again as the underlying problem of registry changes after render is still a thing.

stories/modal-dialog.stories.js
```diff
@@ -21,25 +21,23 @@
...
-import { __ } from '@wordpress/i18n';
-
 /**
  * Internal dependencies
  */
 import Dialog from '../assets/js/components/Dialog';
 
 storiesOf( 'Global', module )
 	.add( 'Modal Dialog', () => {
-		const { provides } = global._googlesitekitLegacyData.modules.analytics;
```
aaemnnosttv, line 36: The legacy data no longer has this key as it only exists in module registration now (JS) whereas before this was coming from PHP which has been removed for a while.

.storybook/__fixtures__/_googlesitekitLegacyData.js
```diff
@@ -0,0 +1,260 @@
...
+
+export default {
+	admin: {
+		siteURL: 'https://example.com',
+		siteName: 'My Great Blog',
+		siteUserID: 'a1b2c3d4e5f6a1b2c3d4e5f6a1b2c3d4',
+		adminRoot: 'https://example.com/wp-admin/admin.php',
+		assetsRoot: 'https://example.com/wp-content/plugins/google-site-kit/dist/assets/',
+		nojscache: false,
+		datacache: false,
```
aaemnnosttv, line 27: `datacache` was the key that held the huge string of JSON that made this so massive before. It isn't used anymore although the key is still set.

aaemnnosttv (commented):
Ready for another pass @tofumatt 👍

tests/js/utils.js
```diff
@@ -428,6 +428,29 @@ export const unsubscribeFromAll = () => {
...
+	const unsubscribe = subscribeWithUnsubscribe( registry, listener );
+
+	// Return a function that waits until the next tick for updates.
+	// We unsubscribe afterwards to allow for potential additions while
+	// Promise.all is resolving.
+
+	return async () => {
+		await Promise.all( updates );
+		unsubscribe();
+	};
```
tofumatt, line 451:
```suggestion
	// Return a function that waits until the next tick for updates.
	// We unsubscribe afterwards to allow for potential additions while
	// Promise.all is resolving.
	return async () => {
		await Promise.all( updates );
		unsubscribe();
	};
```

tofumatt (approved):
Awesome, looks great 👍🏻

Issue #3702, opened 2021-07-13
Provide link in Analytics settings panel to update GA4 configuration and improve UI for related UA link


felixarntz:
In the Analytics settings panel (in view-only state), there is currently a long message with a link that points to the Analytics frontend for the currently active Analytics property and its view.

Now that users can also have a GA4 property configured there, this message is confusing, as it appears below all of those settings while still only focusing on UA. We will now need two links instead of just the one. In addition, it's a bit strange to have a long sentence as a complete link anyway. Let's revisit this UI.

Screenshot:
<img width="647" alt="Screen Shot 2021-07-13 at 9 55 56 AM" src="https://user-images.githubusercontent.com/3531426/125494185-71b2e7f5-5c2d-47ea-9a3e-9563ec5f6782.png">

A potential solution would be:
* Display a small "Edit (Universal Analytics configuration) in Analytics" external link for the UA property on the right of the three UA values.
* Display a similar small "Edit (Google Analytics 4 configuration) in Analytics" external link for the GA4 property on the right of the two UA values.

cc @Pratheep-lab 

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

In the Analytics settings view:
* Remove the link "You can make changes to this view..." and instead place a short link (with external indicator) to the same URL on the right side of the "View" value.
    * It should link to `https://analytics.google.com/analytics/web/?authuser={googleEmailAddress}#/a{accountID}w{uaPropertyID}p{viewID}/admin/view/settings` (same as the previous "You can make changes to this view..." link)
    * It should say "Edit in Analytics"
    * For screen readers, it should be more specific: "Edit Universal Analytics configuration in Analytics"
* If a GA4 property is also connected, display a similar link on the right side of the "Measurement ID" value:
    * It should link to `https://analytics.google.com/analytics/web/?authuser={googleEmailAddress}#/a{accountID}p{ga4PropertyID}/admin/streams/table/{webDataStreamID}`
    * It should say "Edit in Analytics"
    * For screen readers, it should be more specific: "Edit Google Analytics 4 configuration in Analytics"
* The links should both appear in the same vertical position and with the same font-size as the individual setting values (i.e. the IDs that display right under all the labels).
* The changes should be available/previewable in Storybook.

## Implementation Brief

* Continue and merge PR here https://github.com/google/site-kit-wp/pull/3766

Using `assets/js/modules/analytics/components/settings/SettingsView.js`

In order to build the GA4 Url
[...]

PR #3766, opened 2021-07-27
Provide link in Analytics settings panel to update GA4 configuration and improve UI for related UA link 

danielgent:
## Summary

Addresses issue #3702

## Checklist


assets/js/modules/analytics/components/settings/SettingsView.js
```diff
@@ -42,6 +42,7 @@ export default function SettingsView() {
 	const isGA4Enabled = useFeature( 'ga4setup' );
 	const ga4PropertyID = useSelect( ( select ) => isGA4Enabled ? select( MODULES_ANALYTICS_4 ).getPropertyID() : '' );
 	const ga4MeasurementID = useSelect( ( select ) => isGA4Enabled ? select( MODULES_ANALYTICS_4 ).getMeasurementID() : '' );
+	const webDataStreamID = useSelect( ( select ) => isGA4Enabled ? select( MODULES_ANALYTICS_4 ).getWebDataStreamID() : '' );
```
danielgent, line 45: I think this needs double checking.  For me it was opening a web data stream for a totally different property (which makes no sense given the Url includes the correct property!)

assets/js/modules/analytics/components/settings/SettingsView.js
```diff
@@ -95,6 +100,23 @@ export default function SettingsView() {
 						<DisplaySetting value={ profileID } />
 					</p>
 				</div>
+				<div className="googlesitekit-settings-module__meta-item">
+					<h5 className="googlesitekit-settings-module__meta-item-type">
+						&nbsp;
+					</h5>
```
danielgent, line 106: Easiest and safest way to make the Links vertically aligned as per the AC (but this always feels a bit hacky!)

assets/js/modules/analytics/components/settings/SettingsView.js
```diff
@@ -95,6 +100,23 @@ export default function SettingsView() {
...
+					<Link
+						href={ editViewSettingsURL }
+						external
+						className="googlesitekit-settings-module__inline-link"
```
danielgent, line 110: This seemed the safest way to override using CSS without making cascading changes through the whole site

assets/js/modules/analytics/components/settings/SettingsView.js
danielgent: This needs hiding on mobile (as per updated AC)

assets/sass/components/settings/_googlesitekit-settings-module.scss
```diff
@@ -157,6 +157,10 @@
 		&--nomargin {
 			margin: 0;
 		}
+
+		.googlesitekit-settings-module__inline-link {
+			font-size: 1rem;
+		}
```
danielgent, line 163: Remove this class. Change of approach

assets/js/modules/analytics/components/settings/SettingsView.js
danielgent: Change of plan: remove this `className` and instead wrap the links with `<p className="googlesitekit-settings-module__meta-item-data">`

assets/js/modules/analytics/components/settings/SettingsView.js
```diff
@@ -63,6 +64,10 @@ export default function SettingsView() {
 		path: `/a${ accountID }w${ internalWebPropertyID }p${ profileID }/admin/view/settings`,
 	} ) );
 
+	const editViewSettingsGA4URL = useSelect( ( select ) => select( STORE_NAME ).getServiceURL( {
+		path: `/a${ accountID }p${ ga4PropertyID }/admin/streams/table/${ webDataStreamID }`,
```
danielgent, line 68: This should be encoded using `escapeURI`

`escapeURI`/a${ accountID }p${ ga4PropertyID }/admin/streams/table/${ webDataStreamID }`,`

danielgent: Also do this for the UA url

assets/js/modules/analytics/components/settings/SettingsView.js
```diff
@@ -63,6 +64,10 @@ export default function SettingsView() {
 		path: `/a${ accountID }w${ internalWebPropertyID }p${ profileID }/admin/view/settings`,
 	} ) );
 
+	const editViewSettingsGA4URL = useSelect( ( select ) => select( STORE_NAME ).getServiceURL( {
```
aaemnnosttv, line 67: GA4 doesn't have Views (Profiles) so we should update this to be more specific to a data stream. Data streams are also GA4-specific so we also don't need to include GA4 in the name to disambiguate either 🙂 
 ```suggestion
	const editDataStreamSettingsURL = useSelect( ( select ) => select( STORE_NAME ).getServiceURL( {
```

File: assets/js/modules/analytics/components/settings/SettingsView.js (hunk 2 above)
Comment (aaemnnosttv) [reply assets/js/modules/analytics/components/settings/SettingsView.js:106]: We should make the links sibling elements of their respective `DisplaySetting`s instead. We want the link to be clearly associated with the value and putting it in a separate element like this spaces it in the grid like a separate thing. This also has the potential to wrap on smaller screens where it would lose the positional context as to what it is referring to. Keeping them in the same parent also has the benefit of not requiring a hack to preserve the alignment 👍

File: assets/js/modules/analytics/components/settings/SettingsView.js (hunk 3 above)
Comment (aaemnnosttv) [reply assets/js/modules/analytics/components/settings/SettingsView.js:110]: See my previous comment about placement. If needed, I think the `inherit` prop of the `Link` component should be all we need to match the surrounding text.

File: assets/js/modules/analytics/components/settings/SettingsView.js [hunk 7]
```diff
@@ -95,6 +100,23 @@ export default function SettingsView() {
...
+					<h5 className="googlesitekit-settings-module__meta-item-type">
+						&nbsp;
+					</h5>
+					<Link
+						href={ editViewSettingsURL }
+						external
+						className="googlesitekit-settings-module__inline-link"
+					>
+						{ createInterpolateElement(
+							__( 'Edit <VisuallyHidden>Universal Analytics property view </VisuallyHidden>in Analytics', 'google-site-kit' ),
```
aaemnnosttv, line 113: Let's use two strings for the variants instead – this will be simpler for translators as well.
The screen reader variant can simply be set as the `aria-label` of the `Link`.

aaemnnosttv (changes requested):
Thanks @danielgent – this looks good overall. I've left a few comments here as well as replies to some of your annotations. Let me know if you have any questions 👍

assets/js/modules/analytics/components/settings/SettingsView.js
danielgent: Makes sense. Especially when going onto mobile

What do you think about the spacing? I just added one space.   


same row

assets/js/modules/analytics/components/settings/SettingsView.js
danielgent: Much better. It works

works

assets/js/modules/analytics/components/settings/SettingsView.js
danielgent: Done

assets/js/modules/analytics/components/settings/SettingsView.js
danielgent: Done :+1:

assets/js/modules/analytics/components/settings/SettingsView.js
danielgent: This is copied from before in the code.  I'll comment there now @aaemnnosttv

assets/js/modules/analytics/components/settings/SettingsView.js
```diff
@@ -170,18 +192,39 @@ export default function SettingsView() {
 								<DisplaySetting value={ ga4MeasurementID } />
 							</p>
 						</div>
+						<div className="googlesitekit-settings-module__meta-item">
+							<h5 className="googlesitekit-settings-module__meta-item-type">
+								{ createInterpolateElement(
```
danielgent, line 197: @aaemnnosttv Note the use of `VisuallyHidden` in this pattern in the elements preceding this one

assets/js/modules/analytics/components/settings/SettingsView.js
aaemnnosttv: It does look a _little_ tight but also makes it more clear what the link refers to. I don't think we should get into adding `&nbsp;`s or another class to add space as this seems a bit arbitrary and makes the link a bit more ambiguous.

What do you think about this?

image

This is how it looks if we also passed the `small` prop to the `Link`s? I realize the ACs specifically say that it should be the same size but I think it might look a bit better with a reduced size and we have an existing prop to do this. Seeing the options side-by-side would also help make the case for changing that in case we feel it looks better 🙂  We'd have to ask about changing it but I think prefer the smaller link. What do you think?

assets/js/modules/analytics/components/settings/SettingsView.js
aaemnnosttv: Ah okay, yeah I missed that. In that case then it's fine and consistent with what's already there 👍

aaemnnosttv (approved):
Thanks @danielgent – this looks good as defined. See my question about the size of the link, but otherwise this looks good to go.

assets/js/modules/analytics/components/settings/SettingsView.js
```diff
@@ -170,18 +192,39 @@ export default function SettingsView() {
 								<DisplaySetting value={ ga4MeasurementID } />
 							</p>
 						</div>
+						<div className="googlesitekit-settings-module__meta-item">
+							<h5 className="googlesitekit-settings-module__meta-item-type">
```
aaemnnosttv, line 196: Not sure how I missed this before, but this is duplicating the output for the Measurement ID
image

aaemnnosttv (changes requested):
Looks like I missed an important detail in the last review!

assets/js/modules/analytics/components/settings/SettingsView.js
aaemnnosttv: Let's keep it as it is in the ACs for now, we can always follow-up with a super simple PR if we decide to tweak the Link size later.

assets/js/modules/analytics/components/settings/SettingsView.js
danielgent: Fixed now!

oops

That's kind of funny. It even has screenshots with the duplicate Measurement Id and we still didn't spot it  :stuck_out_tongue:

assets/js/modules/analytics/components/settings/SettingsView.js
danielgent: I don't think there's anything actionable here for me now @aaemnnosttv ?

aaemnnosttv (approved):

Issue #3747, opened 2021-07-20
Switch Idea Hub module REST route for updating idea states to actually use Idea Hub API


felixarntz:
Broken out from #3518: Since the Idea Hub API route to update an idea's state is still not working correctly, the `POST:update-idea-state` REST endpoint should be connected to the actual API separately, in this issue.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `POST:update-idea-state` endpoint should use https://github.com/googleapis/google-api-php-client-services/blob/master/src/Google/Service/Ideahub/Resource/PlatformsPropertiesIdeaStates.php, passing platform and property similar to the other endpoints (see #3518) and the received idea name as the "ideaState" placeholder in the path. The data point should expect a single variable `ideaState` which needs to be an object / associative array with `name` (idea name string, also used in the path as mentioned right above), `saved` (boolean) and `dismissed` (boolean) fields.
* In other words, with this issue, saving/unsaving and dismissing an idea should actually begin to work / result in the idea no longer showing up in the new/saved tab.

## Implementation Brief

* Verify that the approach in #3767 works as intended when the corresponding Idea Hub API starts to work correctly. If it doesn't work as intended (see AC) then make necessary changes and send PR for code review

### Test Coverage

* N/A

### Visual Regression Changes

* N/A

## QA Brief

* Activate Idea Hub module
* Set the custom site URL setting in the Site Kit developer settings to `https://www.elasticpress.io/`
* Go to the dashboard and verify that you see new ideas
* Click on the pin icon to save the idea. Verify that the saved idea has been moved to the "Saved" ideas tab.
* Refresh the page and verify that the saved ideas list contains the recently saved idea and the new ideas list doesn't contain it anymore.
* Now unsave the idea and verify that the idea has been moved to the ideas list (at the end). Plus after refreshing the page you see it in the new ideas list and don't see it in the saved ideas anymore.
    * **NOTE:** While the above is what _should_ happen here, there appears to be currently a bug with the API that prevents the first sentence above from working. In other words, the (for now broken) accepted functionality is: When unsaving, at first the idea won't move back to the "New" tab. But it should at least have been moved there once you refresh the page. This likely needs to be fixed at the API level (also see https://github.com/google/site-kit-wp/pull/3767#discussion_r685573564).
* Dismiss an idea and verify that the idea doesn't re-appear after refreshing the dashboard page.

## Changelog entry

[...]

PR #3767, opened 2021-07-27
Enhancement/3747 update idea states

eugene-manuilov:
## Summary

Addresses issue #3747

## Checklist


assets/js/modules/idea-hub/datastore/idea-state.js
```diff
@@ -78,36 +78,38 @@ const fetchPostUpdateIdeaStateStore = createFetchStore( {
...
+			if ( ! ideaDetails ) {
+				ideaDetails = idea;
 			}
 
 			return {
 				...state,
-				newIdeas: newIdeas.filter( ( { name } ) => name !== idea.name ),
-				savedIdeas: [ ...savedIdeas, ...ideaDetails ],
-			};
-		} else if ( idea.saved === false ) {
```
felixarntz, line 101: Don't we still need a condition here for `idea.saved === false`? When you unsave an idea in the "Saved" tab, it should disappear from there and then new ideas should be updated so that it reappears there _if_ it still is a new idea.

felixarntz (changes requested):
@eugene-manuilov The code mostly looks good, I left a few comments. However, I'm getting a 400 error when I test the endpoint that the datapoint doesn't exist. Not sure if I'm doing something wrong, this seems really basic 🤔 
Screen Shot 2021-08-09 at 1 03 55 PM

includes/Modules/Idea_Hub.php
```diff
@@ -617,6 +668,23 @@ public function set_post_idea( $post_id, array $idea ) {
...
+	 * @since n.e.x.t
+	 *
+	 * @param Google_Model $idea Idea model.
+	 * @return \stdClass Updated model with _id attribute.
+	 */
+	public static function filter_ideas_with_id( $idea ) {
+		$name_parts = explode( '/', $idea['name'] );
+
+		$obj      = $idea->toSimpleObject();
+		$obj->_id = array_pop( $name_parts );
```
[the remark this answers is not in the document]
eugene-manuilov: Yep, makes sense to me. Updated.

includes/Modules/Idea_Hub.php
```diff
@@ -617,6 +668,23 @@ public function set_post_idea( $post_id, array $idea ) {
...
 
+	/**
+	 * Parses an idea ID, adds it to the model object and returns updated model.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Google_Model $idea Idea model.
+	 * @return \stdClass Updated model with _id attribute.
+	 */
+	public static function filter_ideas_with_id( $idea ) {
```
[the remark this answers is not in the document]

includes/Modules/Idea_Hub.php
```diff
@@ -498,7 +546,10 @@ function( $post_id ) {
 					)
 				);
 			case 'GET:saved-ideas':
-				return $this->filter_out_ideas_with_posts( $response->getIdeas() );
+				$ideas = $this->filter_out_ideas_with_posts( $response->getIdeas() );
+				return array_map( array( self::class, 'filter_ideas_with_id' ), $ideas );
+			case 'POST:update-idea-state':
+				return self::filter_ideas_with_id( $response );
```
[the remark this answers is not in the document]
eugene-manuilov: Added a new method.

composer.json
```diff
@@ -20,7 +20,7 @@
     "ext-json": "*",
     "ext-openssl": "*",
     "google/apiclient": "^2.9.1",
-    "google/apiclient-services": "^0.200",
+    "google/apiclient-services": "^0.207.0",
```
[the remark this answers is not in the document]
eugene-manuilov: No, i haven't tested other modules. So, decided to revert this update for now.

assets/js/modules/idea-hub/datastore/idea-state.js
eugene-manuilov: Yes, it worked like that. Added back `else if ( idea.saved === false )` condition to be more compatible with how it's done in the develop branch.

includes/Modules/Idea_Hub.php
```diff
@@ -439,10 +441,55 @@ protected function create_data_request( Data_Request $data ) {
...
+				);
+
+				$update_mask = array();
+
+				$body = new Google_Service_Ideahub_GoogleSearchIdeahubV1alphaIdeaState();
+				$body->setName( $idea_name );
+
+				if ( isset( $data['saved'] ) ) {
+					$body->setSaved( filter_var( $data['saved'], FILTER_VALIDATE_BOOLEAN ) );
+					$body->setDismissed( false );
```
felixarntz, line 478: Now that we've figured out the use of the update mask, is this even still relevant? I think it can now be removed.

includes/Modules/Idea_Hub.php
```diff
@@ -439,10 +441,55 @@ protected function create_data_request( Data_Request $data ) {
...
+					$update_mask[] = 'saved';
+				}
+
+				if ( isset( $data['dismissed'] ) ) {
+					$body->setDismissed( filter_var( $data['dismissed'], FILTER_VALIDATE_BOOLEAN ) );
+					$body->setSaved( false );
```
felixarntz, line 484: See above.

assets/js/modules/idea-hub/datastore/idea-state.js
felixarntz: Hmm, it looks like for some reason the response doesn't include the property if it is set to `false`. Try unsaving an idea from the saved tab - even though it sends `saved:false` seemingly correctly, the response only includes `name` and `_id`, and therefore the logic here isn't triggered. Unsaving itself actually works as expected though, on the following page load you see the idea in the right place.

Could it be that somehow the `saved: false` is stripped on the way to our data point logic, since it's `false`? If not, this may be an API bug we need to flag, unless you see another reason it's occuring. While we could work around it, I don't think that's a good idea, it should rather be fixed in the API then.

felixarntz (approved):
@eugene-manuilov There's one thing that's still not working here, but as far as I can tell it's due to the API. I'll mark this as approved since otherwise it's fine other than two nit-picks.

We'd just have to be aware during the bug bash that, once this is flagged, it's already a known bug.

includes/Modules/Idea_Hub.php
eugene-manuilov: Yes, we don't need it. Added it to #3818

includes/Modules/Idea_Hub.php
eugene-manuilov: Yep, we don't need it.

assets/js/modules/idea-hub/datastore/idea-state.js
eugene-manuilov: > Could it be that somehow the saved: false is stripped on the way to our data point logic, since it's false?
No, I don't think so. I have created a follow-up PR (#3818) that fixes this issue on our end.

Issue #3762, opened 2021-07-26
Add support for additional survey question type "open_text"


felixarntz:
When the initial user feedback survey was introduced, scope of the MVP was limited to only support questions of the "rating" type, since that was the only one used for the initial survey. This should now be followed up on, with this issue focusing on the "open_text" question type.

![Screen Shot 2021-07-26 at 3 52 03 PM](https://user-images.githubusercontent.com/3531426/127069302-e22e2a37-7db0-4dc0-940b-b50cdd71f0dd.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new `SurveyQuestionOpenText` JS component should be introduced and integrated into the user feedback survey JS infrastructure, rendering a survey question of type "open_text".
* The UI should look like the above screenshot and follow the "rating" question type where similar (e.g. font, spacing, G icon etc.).
* The available arguments to consider here from the Site Kit Service `/survey/trigger/` API response are as follows:
    * `question.question.subtitle` should, if present, be displayed right below the text input (see screenshot).
    * `question.question.placeholder` should, if present, be displayed as a text input placeholder (see screenshot).
    * The text input must be non-empty in order to proceed. The maximum character limit should be 100.
* The data object for answering the question (what needs to be passed to the `answerQuestion` function prop) should look as follows:
    * A single `answer` property with the string from the text input.
    * Example: `{ answer: 'My open text response' }`
* The new question type should have an example in Storybook where it can be inspected, similar to the existing "rating" type component.

#### Create new component

Create `assets/js/components/surveys/SurveyQuestionOpenText.js`

Base on SurveyQuestionRating.js
* wrap everything in a div with class `.googlesitekit-survey__open-text`
* Use `SurveyHeader` to match styling
* Use the `TextField`, `Input`, `HelperText` components
* Add the following string props and propTypes: `question`, `subtitle`, `placeholder`
* Add the following function props and propTypes: `answerQuestion`, `dismissSurvey`
* HelperText should be passed into `TextField`'s `helperText` prop (with the child `subtitle`)
* In order to limit user input to 100 characters, just disallow extra characters in the TextField's onChange handler (slicing the text to 100 characters)
* If the TextField is empty, set a disabled prop to true on the submit button
* When submitting, pass the answer to `answerQuestion` as per AC

In `assets/sass/components/surveys/_googlesitekit-surveys.scss`

* Add a new rule for `.googlesitekit-survey__open-text`
[...]

PR #3771, opened 2021-07-27
Enhancement/3762 survey question type open text

danielgent:
## Summary

Addresses issue #3762

## Checklist


danielgent:
Note there are no tests here. Happy to add them! :+1:

assets/js/components/surveys/CurrentSurvey.js
```diff
@@ -264,9 +266,11 @@ export default function CurrentSurvey() {
 				<SurveyQuestionComponent
 					key={ currentQuestion.question_text }
 					answerQuestion={ answerQuestion }
-					choices={ currentQuestion.question.answer_choice }
+					choices={ currentQuestion.question?.answer_choice } // eslint-disable-line camelcase
 					dismissSurvey={ dismissSurvey }
 					question={ currentQuestion.question_text }
+					subtitle={ currentQuestion.subtitle }
+					placeholder={ currentQuestion.placeholder }
```
danielgent, line 273: This part wasn't mentioned in the IB. Passing in every prop required for every type of QuestionComponent doesn't seem the ideal approach (especially when each question won't have these values)

There are better ways to do this *but* I don't want to generate merge conflicts, so I just did the minimum for the tests to pass :+1:

assets/js/components/surveys/CurrentSurvey.test.js
```diff
@@ -80,6 +81,122 @@ describe( 'CurrentSurvey', () => {
 		expect( container ).toMatchSnapshot();
 	} );
 
+	it( "should render an open text question when the `question_type` is 'open_text'", async () => {
```
danielgent, line 84: This test covers all the functionality of `SurveyQuestionOpenText`, and also its integration with `CurrentSurvey`

assets/js/components/surveys/SurveyQuestionOpenText.js
```diff
@@ -0,0 +1,95 @@
...
+} ) => {
+	const [ value, setValue ] = useState( '' );
+
+	const handleSubmit = () => {
+		answerQuestion( { answer: value } );
+	};
+
+	const onChange = useCallback(
+		( event ) => {
+			setValue( event.target.value?.slice( 0, MAXIMUM_CHARACTER_LIMIT ) );
```
danielgent, line 54: Strangely `@material/react-text-field` doesn't support the HTML attribute `maxlength`, so I have to manually program this (there are tests)
https://stackoverflow.com/questions/45939909/put-length-contraint-in-a-textfield-in-react-js

assets/js/components/surveys/CurrentSurvey.test.js
```diff
@@ -80,6 +80,120 @@ describe( 'CurrentSurvey', () => {
...
+		expect( fetchMock ).toHaveFetched(
+			'/google-site-kit/v1/core/user/data/survey-event?_locale=user',
+			{
+				body: {
+					data: {
+						event: {
+							question_answered: {
+								question_ordinal: 1,
+								answer: {
+									answer:
```
danielgent, line 174: This answer key is nested twice... :thinking: 

Seems wrong. Will check spec

danielgent: This is what `SurveyQuestionRatingChoice` does! Leaving as is :shrug:

assets/js/components/surveys/CurrentSurvey.js
tofumatt: I'm not a fan of prop spreading, but I suppose assembling an object for each question type then spreading that isn't the end of the world.

I'd be fine with that if you wanted to, but I don't really mind the explicitness of passing all of the props 😅

assets/js/components/surveys/CurrentSurvey.test.js
tofumatt: This looks good in terms of testing functionality, but this is a very long test with many `expect` statements that seem unrelated.

Let's put everything related to `SurveyQuestionOpenText` in its own `describe` block, then separate out some of these `expect` statements.

assets/js/components/surveys/CurrentSurvey.test.js
```diff
@@ -80,6 +80,120 @@ describe( 'CurrentSurvey', () => {
...
+
+		// Now enter answer text
+		fireEvent.change( getByPlaceholderText( 'Write here' ), {
+			target: { value: STRING_110_CHARACTERS },
+		} );
+
+		// Input does not have a label so this is the best way to assert the text has been entered correctly
+		expect(
+			getByDisplayValue( STRING_100_CHARACTERS )
+		).toBeInTheDocument();
```
tofumatt, line 139: Oh, why doesn't the input have a label?

assets/js/components/surveys/CurrentSurvey.test.js
```diff
@@ -80,6 +80,120 @@ describe( 'CurrentSurvey', () => {
...
+
+		// Submit button should be enabled if text has been entered
+		expect( getByRole( 'button', { name: 'Submit' } ) ).not.toHaveAttribute(
+			'disabled'
+		);
+
+		// check clearing and entering input again
```
tofumatt, line 146: These are super useful, thanks 🙂 

Nitpick, but can you make all of these comments full sentences including ending with a fullstop?

assets/js/components/surveys/CurrentSurvey.test.js
```diff
@@ -80,6 +80,120 @@ describe( 'CurrentSurvey', () => {
...
+				credentials: 'include',
+				headers: {
+					Accept: 'application/json, */*;q=0.1',
+					'Content-Type': 'application/json',
+				},
+				method: 'POST',
+			}
+		);
+
+		await findByText( 'Thanks for sharing your thoughts!' );
```
tofumatt, line 194: I know this should fail if the `await` doesn't fulfill, but that's a bit less obvious than an expectation around this… maybe:
```suggestion
		const completionText = await findByText( 'Thanks for sharing your thoughts!' );
		expect( completionText ).not.toBeEmpty();
```
or something like that?

assets/js/components/surveys/SurveyQuestionOpenText.js
tofumatt: Huh, brutal! That's quite odd!

Thanks for catching that 👍🏻

assets/js/components/surveys/SurveyQuestionOpenText.js
```diff
@@ -0,0 +1,95 @@
...
+		<div className="googlesitekit-survey__open-text">
+			<SurveyHeader title={ question } dismissSurvey={ dismissSurvey } />
+			<div className="googlesitekit-survey__body">
+				<div>
+					<TextField
+						name="open-text"
+						helperText={ <HelperText>{ subtitle }</HelperText> }
+						onChange={ onChange }
+						label={ placeholder }
+						noLabel
```
tofumatt, line 69: Why is there no label for this input? I don't see that mentioned in the ACs/IB, and doing this severely impacts accessibility of the input.

tofumatt (changes requested):
Looks good, but I think the input here needs a label (even if it's off-screen), and then the tests need to be split up.

If the data from the API/selector doesn't have an explicit `label` value, there should be some kind of useful help text to show users as a `<label>`.

If there _isn't_, then we need to tweak the API and file an issue for that.

assets/js/components/surveys/CurrentSurvey.js
danielgent: Currently, there are three open PRs all with similar changes here.

IMHO we should just have a switch statement that returns the actual component instantiated with the correct props, rather than this `SurveyQuestionComponent` trickery :smile: 

But I presume this will be fixed when 

I only use prop spreading to allow setting of HTML attributes (when creating a text input or something like that) :+1:

assets/js/components/surveys/CurrentSurvey.test.js
danielgent: Sounds good @tofumatt 

I don't think I really got the mandate from Evan to do so many tests (see comment https://github.com/google/site-kit-wp/issues/3762#issuecomment-892736060) but I :heart: comprehensive testing as I :heart:  TDD

assets/js/components/surveys/CurrentSurvey.test.js
danielgent: It's actually to stop Act warnings.  If we don't wait for this then I think we get Act warnings.  This is actually copied from another test though!  `should send a 'question_answered' event when a question is answered`, so for consistency I think we should leave it (I also have something similar on my next two open PRs) :smile:

assets/js/components/surveys/SurveyQuestionOpenText.js
danielgent: To match the design. If I remove this prop then when entering text the placeholder (which is really the label) does this

label

assets/js/components/surveys/CurrentSurvey.test.js
danielgent: Done.

assets/js/components/surveys/CurrentSurvey.test.js
danielgent: Answered below as a response to another comment :smile:

assets/js/components/surveys/CurrentSurvey.test.js
danielgent: I've broken up this test into smaller tests. Happy to make any more changes of course

assets/js/components/surveys/SurveyQuestionOpenText.js
tofumatt: Right, that is a quirk of material inputs.

Then let's add a manual `<label>` tag and connect the ID for this element and the label. We can't have inputs without labels or users of assistive technology won't be able to use the form.

assets/js/components/surveys/CurrentSurvey.test.js
```diff
@@ -80,6 +80,147 @@ describe( 'CurrentSurvey', () => {
...
+
+			// Check question_text is set by question_text prop.
+			expect(
+				getByText( 'How satisfied are you with Site Kit?' )
+			).toBeInTheDocument();
+
+			// Check subtitle is set by subtitle prop.
+			expect(
+				getByText( 'Based on your experience so far, tell us.' )
+			).toBeInTheDocument();
```
tofumatt, line 114: These tests are much better separated, thanks!

These two expect statements in particular aren't really related to the 100 character limit tests, so could also be separated. I think that'd be good, but these two (`question_text` and `subtitle` prop) could probably be in the same `it()` block.

assets/js/components/surveys/CurrentSurvey.test.js
```diff
@@ -80,6 +80,147 @@ describe( 'CurrentSurvey', () => {
...
+					credentials: 'include',
+					headers: {
+						Accept: 'application/json, */*;q=0.1',
+						'Content-Type': 'application/json',
+					},
+					method: 'POST',
+				}
+			);
+
+			await findByText( 'Thanks for sharing your thoughts!' );
```
tofumatt, line 220: Can we associate this with an `expect` statement or something? I'm guessing this is here to ensure this text gets rendered, as the test will time out without it fulfilling the `await`. But that's implicit, so let's explicitly check for that text in the DOM instead.

tofumatt (changes requested):
Looks like the input still needs a `label`, but otherwise this looks good 👍🏻

assets/js/components/surveys/CurrentSurvey.test.js
danielgent: Understood. I just didn't want to keep adding extra small tests but I'm sure these are quick to run anyway :+1:

danielgent: https://kentcdodds.com/blog/write-fewer-longer-tests

Initially I was following the react-testing-library creator's advice, but it's not a hard science :smile:

assets/js/components/surveys/CurrentSurvey.test.js
danielgent: As mentioned somewhere else (on another PR I think) this is copied from the test `should send a 'question_answered' event when a question is answered`

Git blame says you wrote this test :stuck_out_tongue:  (But I'm sure that's just because you merged the Prettier PR :see_no_evil: )


I can change it here and there and every PR *but* I only have this just to not get this error
`Can't perform a React state update on an unmounted component.`

assets/js/components/surveys/SurveyQuestionOpenText.js
danielgent: Sure thing. I'll do the same as I did on the other PRs :+1:

assets/js/components/surveys/SurveyQuestionOpenText.js
```diff
@@ -0,0 +1,104 @@
...
+		( event ) => {
+			setValue( event.target.value?.slice( 0, MAXIMUM_CHARACTER_LIMIT ) );
+		},
+		[ setValue ]
+	);
+
+	const instanceID = useInstanceId(
+		SurveyQuestionOpenText,
+		'SurveyQuestionOpenText'
+	);
```
danielgent, line 64: Made this change as per the other PRs

assets/js/components/surveys/CurrentSurvey.js
```diff
@@ -269,9 +271,11 @@ export default function CurrentSurvey() {
 				<SurveyQuestionComponent
 					key={ currentQuestion.question_text }
 					answerQuestion={ answerQuestion }
-					choices={ currentQuestion.question.answer_choice }
+					choices={ currentQuestion.question?.answer_choice } // eslint-disable-line camelcase
```
tofumatt, line 274: Why the optional chaining here?

assets/js/components/surveys/CurrentSurvey.test.js
```diff
@@ -86,6 +86,158 @@ describe( 'CurrentSurvey', () => {
...
+					triggerID: 'jestSurvey',
+				} );
+
+			fetchMock.post(
+				/^\/google-site-kit\/v1\/core\/user\/data\/survey-event/,
+				{ body: {}, status: 200 }
+			);
+		} );
+
+		it( 'should  display question text and subtitle correctly', async () => {
```
tofumatt, line 103:
```suggestion
		it( 'should display the question prompt and subtitle', async () => {
```

assets/js/components/surveys/CurrentSurvey.test.js
```diff
@@ -86,6 +86,158 @@ describe( 'CurrentSurvey', () => {
...
+			const { getByText } = render( <CurrentSurvey />, {
+				registry,
+			} );
+
+			// Check question_text is set by question_text prop.
```
tofumatt, line 108:
```suggestion
			// Check the question's prompt is set by the `question_text` prop.
```

assets/js/components/surveys/CurrentSurvey.test.js
```diff
@@ -86,6 +86,158 @@ describe( 'CurrentSurvey', () => {
...
+			expect(
+				getByText( 'How satisfied are you with Site Kit?' )
+			).toBeInTheDocument();
+
+			// Check subtitle is set by subtitle prop.
+			expect(
+				getByText( 'Based on your experience so far, tell us.' )
+			).toBeInTheDocument();
```
tofumatt, line 135: These are tested elsewhere, so can be removed from this test.

assets/js/components/surveys/CurrentSurvey.test.js
```diff
@@ -86,6 +86,158 @@ describe( 'CurrentSurvey', () => {
...
+
+			// Enter answer text.
+			fireEvent.change( getByLabelText( 'Write here' ), {
+				target: { value: STRING_110_CHARACTERS },
+			} );
+
+			// Input does not have a label so this is the best way to assert the text has been entered correctly
```
tofumatt, line 142: This isn't true anymore, so we should select this by its label.

assets/js/components/surveys/SurveyQuestionOpenText.js
```diff
@@ -0,0 +1,104 @@
...
+} ) => {
+	const [ value, setValue ] = useState( '' );
+
+	const handleSubmit = () => {
+		answerQuestion( { answer: value } );
+	};
+
+	const onChange = useCallback(
+		( event ) => {
+			setValue( event.target.value?.slice( 0, MAXIMUM_CHARACTER_LIMIT ) );
```
tofumatt, line 56: This doesn't need optional chaining though, right? I think you mentioned in another PR this wasn't needed, can we remove it here too?
```suggestion
			setValue( event.target.value.slice( 0, MAXIMUM_CHARACTER_LIMIT ) );
```

If it's actually needed, then we should do `|| ''` or this could be `undefined`.

assets/sass/components/surveys/_googlesitekit-surveys.scss
```diff
@@ -166,14 +185,11 @@
...
-				flex-grow: 1;
-				margin-left: 10px;
-			}
-
 			.mdc-text-field {
+				flex-grow: 1;
 				height: 35px;
-				width: 100%;
+				margin-left: 10px;
+				width: auto;
```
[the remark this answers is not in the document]
tofumatt: Ohhh, shoot okay thanks. I missed that 😓

tofumatt (changes requested):
Looks good, just a few test tweaks now that we have a label in the component. But looks great otherwise, thanks! 👍🏻

assets/js/components/surveys/CurrentSurvey.js
danielgent: This type of question has no choices, so this was the least merge-conflict way of solving this :smile: 

Now this is the last PR to be merged, we could make changes here.

IMHO we should get rid of the `ComponentMap` and just do a switch statement, returning the correct component with the correct props

assets/js/components/surveys/CurrentSurvey.test.js
danielgent: Ooops! Nice spot :smile:

assets/sass/components/surveys/_googlesitekit-surveys.scss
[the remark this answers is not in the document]
danielgent: I'd :heart:  to set up VRTs for every story for this reason but I know that's excessive, but they're so powerful :star_struck:

assets/js/components/surveys/SurveyQuestionOpenText.js
danielgent: Removed and the tests pass. :+1:

assets/js/components/surveys/CurrentSurvey.js
tofumatt: Ahhh, okay, right, makes sense.

Yeah, I think I agree. There's no sense in this approach, and it's less obvious.

Good call on waiting until the last PR though 🙂

I'm fine with a switch case, but if it's not a giant pain/unreadable to do `currentQuestion?.question_type === 'single_select' && <SurveyQuestionSingleSelect />` I'd slightly prefer it. Since we can't put a switch case in the JSX is all, and I find it a bit easier to read about the component output the more we can put all in the JSX that's returned.

But if it makes it really bad to read we can do the switch case instead.

Agreed, let's do that 👍🏻

tofumatt (changes requested):
100% agree on the switch statement and removing the map. Can you make that change then send this back to me? Should be good-to-go after that, since it's already good now but I agree: that's a nice improvement to the code. 👍🏻

assets/js/components/surveys/CurrentSurvey.js
danielgent: @tofumatt  Apologies but I'm not following the sentence `if it's not a giant pain/unreadable to do`

In the comment at the end of this PR you mention to just do the switch statement so I'll do that now, make sure the tests pass, and then we could sync to discuss your other idea further :smile:

danielgent: I made the change and the tests pass for me locally. Will assign to you when all the GitHub actions pass :+1:

tofumatt (approved):

tofumatt:
@googlebot I consent.

assets/js/components/surveys/CurrentSurvey.js
```diff
@@ -260,61 +269,40 @@ export default function CurrentSurvey() {
...
+						{ ...commonProps }
+						choices={ currentQuestion.question.answer_choice }
+					/>
+				) }
+				{ currentQuestion.question_type === 'single_select' && (
+					<SurveyQuestionSingleSelect
+						{ ...commonProps }
+						choices={ currentQuestion.question.answer_choice }
+					/>
+				) }
```
danielgent, line 305: Interesting change @tofumatt  :smile: 

I like the switch statement because it's a traditional programming construct *and* it's safe in that only one thing can be returned

(actually, I've just thought about that and without eslint it's really easy to not break out of a case :yum: )

Whereas with your way of doing it, everything is inline and readable without having to scroll back up.  Something I really like in React.

This could obviously be a huge nested ternary statement but I find them really unreadable. Better to just have each condition on a new line :smile: 

This is purely just bikeshedding so don't feel that you need to reply :smile:

assets/js/components/surveys/CurrentSurvey.test.js
```diff
@@ -121,7 +121,6 @@ describe( 'CurrentSurvey', () => {
 				registry,
 			} );
 
-			// Enter answer text.
```
danielgent: Makes sense. It's readable in the code now :joy: 

These started off as actual TODO notes for what I wanted to do.

Another way would be having explicitly named helper methods e.g `changeInputField()` like KCD does in https://kentcdodds.com/blog/avoid-nesting-when-youre-testing :smile:

Issue #3182, opened 2021-04-19
Introduce new integration between Storybook and Backstop


aaemnnosttv:
## Feature Description

With the update to Storybook to v6, we've started migrating a few of our stories to use the new Component Story Format in line with our documentation for how stories should be written going forward.

While Storybook is very useful by itself, we also use it as the basis for our visual regression tests using BackstopJS. Currently, all VRT scenarios are defined in a giant array in the [`.storybook/storybook-data.js`](https://github.com/google/site-kit-wp/blob/6efb4e19cbf87b3593909fa4747bd34763dc57e0/.storybook/storybook-data.js) module. This is cumbersome to update because it's completely separate from both stories and scenarios, leaving a lot to be desired.

In our legacy stories, backstop scenario configuration was set at the story level as a `options` object passed via story `parameters`. This gave the appearance that these options were actually used by Backstop although it still required adding into the `storybook-data.js` manually.

With the new CSF format, we have a lot more flexibility in making this work as expected because story definitions will no longer require running through `storiesOf()`.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Stories defined using CSF should be able to define their BackstopJS scenario definitions on the component itself, just like `args` or `parameters`, etc by defining a `scenario` property
    * This property should be a boolean or a plain object (similar to the way `parameters.options` is now)
        * If `true` the scenario for that story should be added using the default/common scenario definition (see below)
        * `false` should yield the same as if there was no `scenario` property defined (no added scenario) so there's not much reason to use it
        * If defined with an object, this configuration should extend the default/common scenario definition but it should not be able to override the `scenario.url` which must be based on the Story's `id`
            - This format is intended for stories that require custom options for `delay`, etc
* Backstop [scenario sourcing](https://github.com/google/site-kit-wp/blob/6efb4e19cbf87b3593909fa4747bd34763dc57e0/tests/backstop/scenarios.js) should be updated to additionally include scenarios defined on story exports
    - All stories should be collected according to [the paths defined in the main Storybook config](https://github.com/google/site-kit-wp/blob/6efb4e19cbf87b3593909fa4747bd34763dc57e0/.storybook/main.js#L19-L23)
    - The named exports of each module should then be inspected, and any `namedExports.scenario` should be added to the list of scenarios
        - Each `scenario` added in this way should have default values populated for its `label` and `url` based on the story which are required by Backstop
[...]

PR #3782, opened 2021-07-30
Infrastructure/3182 introduce new integration between storybook and backstop

danielgent:
## Summary

Addresses issue #3182

## Checklist


.storybook/storybook-data.js
```diff
@@ -10,73 +10,6 @@ module.exports = [
...
-		story: 'DashboardCoreSiteAlerts1',
-		parameters: {
-			fileName: './assets/js/components/legacy-notifications/DashboardCoreSiteAlerts.stories.js',
-			options: {
-				hierarchyRootSeparator: '|',
-				hierarchySeparator: {},
-				readySelector: '.googlesitekit-publisher-win',
-			},
-		},
-	},
```
danielgent, line 26: These are all now defined in `DashboardCoreSiteAlerts.stories.js`

Note that all the existing VRTs pass with no changes

assets/js/components/legacy-notifications/DashboardCoreSiteAlerts.stories.js
```diff
@@ -62,6 +62,14 @@ NotificationCTA.args = {
 		registry.dispatch( CORE_SITE ).receiveGetNotifications( [ notification1 ], {} );
 	},
 };
+NotificationCTA.scenario = 	{
+	kind: 'Global',
+	name: 'DashboardCoreSiteAlerts1',
+	story: 'DashboardCoreSiteAlerts1',
+	hierarchyRootSeparator: '|',
+	hierarchySeparator: {},
+	readySelector: '.googlesitekit-publisher-win',
```
danielgent, line 71: Ideally these three would be nested inside `options` (maybe inside `parameters`) but currently `@babel/traverse` is returning `undefined` for any nested objects (I haven't looked at this at all though!)

tests/backstop/scenarios.js
```diff
@@ -1,6 +1,104 @@
 const storybookHost = require( './detect-storybook-host' );
 const rootURL = `${ storybookHost }iframe.html?id=`;
 const storybookStories = require( '../../.storybook/storybook-data' );
+// does not work as these paths are relative to this file
+// const filePaths = require( '../../.storybook/main' );
```
danielgent, line 5: TODO - will fix this

tests/backstop/scenarios.js
```diff
@@ -1,6 +1,104 @@
...
+const glob = require( 'glob' );
+const fs = require( 'fs' );
+const parser = require( '@babel/parser' );
+const traverse = require( '@babel/traverse' ).default;
+const csf = require( '@componentdriven/csf' );
+
+const camelCaseToKebabCase = ( string ) => string
+	.replace( /([a-z0-9])([A-Z])/g, '$1-$2' )
+	.replace( /([A-Z])([A-Z])(?=[a-z])/g, '$1-$2' )
+	.toLowerCase();
```
danielgent, line 15: Needed in order

I see we have some other similar utils in some other places so happy to move this (unless I'm missing something, but as per Asvin's snippet no stories were being found.  The Id needs to be the Url)

tests/backstop/scenarios.js
```diff
@@ -1,6 +1,104 @@
...
+			} else if ( node.right.type === 'ObjectExpression' ) {
+				nodeValue = {};
+				node.right.properties.forEach( ( property ) => {
+					nodeValue[ property.key.name ] = property.value.value;
+				} );
+			}
+
+			if ( ! stories[ node.left.object.name ] ) {
+				stories[ node.left.object.name ] = {};
+			}
```
danielgent, line 50: Pretty much the only change from Asvin's snippet!

tests/backstop/scenarios.js
```diff
@@ -1,6 +1,104 @@
...
+		finalStories[ storyID ].parameters = {
+			__id: storyID,
+		};
+		if ( finalStories[ storyID ].args ) {
+			finalStories[ storyID ].parameters.__isArgsStory = true;
+		}
+
+		if ( value && value.scenario && Object.keys( value.scenario ).length > 0 && value.scenario && value.scenario.constructor === Object ) {
+			const newStory = {
+				id: storyID,
```
danielgent, line 79: Here I've changed this a lot in order that the current VRTs pass!

assets/js/components/legacy-notifications/DashboardCoreSiteAlerts.stories.js
aaemnnosttv: This object should only include properties of a [Backstop scenario](https://github.com/garris/BackstopJS#advanced-scenarios). So these three are actually the only properties that belong here 😄 

The Story `id` and `kind` should all be same based on the existing definition so we shouldn't need anything special for that here.

tests/backstop/scenarios.js
aaemnnosttv: How about we use [`lodash/kebabCase`](https://lodash.com/docs/4.17.15#kebabCase) instead?

tests/backstop/scenarios.js
```diff
@@ -1,6 +1,104 @@
...
+	for ( const [ key, value ] of Object.entries( stories ) ) {
+		const storyID = csf.toId( defaultTitle, camelCaseToKebabCase(key) ); // eslint-disable-line
+
+		finalStories[ storyID ] = { ...value };
+		finalStories[ storyID ].key = key;
+		finalStories[ storyID ].id = storyID;
+		finalStories[ storyID ].name = value.storyName;
+		finalStories[ storyID ].kind = defaultTitle;
+		finalStories[ storyID ].story = value.storyName;
+		finalStories[ storyID ].scenarios = value.scenario || {};
```
aaemnnosttv, line 68: The property here should be `scenario`, and shouldn't have a default value. Accordingly, I think this line can simply be removed as this should come through from `value`, no?

tests/backstop/scenarios.js
```diff
@@ -1,6 +1,104 @@
...
+					fileName: storyFile,
+					options: {
+						// would be better to spread?
+						// ...value.scenario,
+						// why is hierarchySeparator undefined here? presume only supports serializable values?
+						hierarchySeparator: value.scenario.hierarchySeparator || {},
+						hierarchyRootSeparator: value.scenario.hierarchyRootSeparator,
+						readySelector: value.scenario.readySelector,
+					},
+				},
```
aaemnnosttv, line 93: We shouldn't need to do anything special here – the sourcing of `scenario`s from stories should only handle stories written in CSF format. `parameters.options` is something used by the old stories which we could maybe handle in a separate issue but we shouldn't really need to since we already have all of those already defined in `storybook-data.js`.

assets/js/components/legacy-notifications/DashboardCoreSiteAlerts.stories.js
danielgent: Changed now. All VRTs passing :+1:

tests/backstop/scenarios.js
danielgent: Nice. I didn't know that existed

.storybook/storybook-data.js
```diff
@@ -10,110 +10,6 @@ module.exports = [
...
-			fileName: './assets/js/components/Button.stories.js',
-			options: {
-				hierarchyRootSeparator: '|',
-				hierarchySeparator: {},
-				hoverSelector: '.googlesitekit-button--hover',
-				postInteractionWait: 3000,
-				onReadyScript: 'mouse.js',
-			},
-		},
-	},
```
danielgent, line 116: There are no more CSF stories here

tests/backstop/scenarios.js
```diff
@@ -1,8 +1,84 @@
...
+const traverse = require( '@babel/traverse' ).default;
+const csf = require( '@componentdriven/csf' );
+const kebabCase = require( 'lodash/kebabCase' );
 
-module.exports = storybookStories.map( ( story ) => {
+const newBackstopTests = [];
+
+const storyFiles =
+	// remove relative parent ..
+	glob.sync( filePaths.stories[ 1 ].slice( 1 ) );
```
danielgent, line 16: Here I'm just using the second value from 
`stories: [ '../stories/**/*.stories.js', '../assets/js/**/*.stories.js' ],`

If I use  '../stories/**/*.stories.js' then I get an error
```
Error: Invalid kind '', must include alphanumeric characters
    at sanitizeSafe (/home/main/wp-local-docker-sites/sitekit-10up
    ```
    
  Is it a problem to leave this one out? As I don't think there are any CSF stories in `./stories/**/*.stories.js'`

File: tests/backstop/scenarios.js [hunk 11]
```diff
@@ -1,8 +1,84 @@
...
+		},
+	} );
+
+	for ( const [ key, value ] of Object.entries( stories ) ) {
+		const storyID = csf.toId( defaultTitle, kebabCase( key ) ); // eslint-disable-line sitekit/acronym-case
+		if (
+			value &&
+			value.scenario &&
+			Object.keys( value.scenario ).length > 0 &&
+			// TODO - also support bool here
```
danielgent, line 68: I've tried doing this with no luck

I'm not sure whether there is a problem with `@babel/parser` or `@babel/traverse` but if I set `scenario = true` on a CSF story, then here the object has an empty string assigned to the scenario key :thinking: 

I can look into this more but I don't know anything about these libraries (apart from a high-level overview of what they do)

danielgent: The quickest thing to do would be just to make it be an empty object.  Then we can just check truthy here and spread in an empty object if no options are set :shrug:

tests/backstop/scenarios.js
danielgent: Is this a problem? I thought there will be no CSF stories in `./stories/**/*.stories.js'` ?

I'm happy to look into this more but it could be time consuming

aaemnnosttv: `ReportError` and `ReportZero` stories are defined with CSF but don't have any scenarios defined for them.

I would prefer not to have any special handling here of the paths. Instead, we should be able to pass the [`cwd` option](https://github.com/isaacs/node-glob#options) to `glob.sync` so that they are processed relative to the `.storybook` directory.

tests/backstop/scenarios.js
```diff
@@ -1,8 +1,84 @@
 const storybookHost = require( './detect-storybook-host' );
 const rootURL = `${ storybookHost }iframe.html?id=`;
 const storybookStories = require( '../../.storybook/storybook-data' );
+const filePaths = require( '../../.storybook/main' );
```
aaemnnosttv, line 4: I realize this is a bit old but please add a file header and group imports as we do elsewhere 🙂

tests/backstop/scenarios.js
```diff
@@ -1,8 +1,84 @@
...
+		if (
+			value &&
+			value.scenario &&
+			Object.keys( value.scenario ).length > 0 &&
+			// TODO - also support bool here
+			value.scenario.constructor === Object
+		) {
+			const newBackstopTest = {
+				url: `${ rootURL }${ storyID }`,
+				...value.scenario,
```
aaemnnosttv, line 73: Let's spread this in before the `url` is defined since that shouldn't be able to be changed.

tests/backstop/scenarios.js
```diff
@@ -1,8 +1,84 @@
...
+		const storyID = csf.toId( defaultTitle, kebabCase( key ) ); // eslint-disable-line sitekit/acronym-case
+		if (
+			value &&
+			value.scenario &&
+			Object.keys( value.scenario ).length > 0 &&
+			// TODO - also support bool here
+			value.scenario.constructor === Object
+		) {
+			const newBackstopTest = {
+				url: `${ rootURL }${ storyID }`,
```
aaemnnosttv, line 72: This is missing a default value for the `label` which should be generated the same as it was before. This is a required field by Backstop but we should continue to provide it as before, while allowing it to be overridden at in the scenario definition (unlike the `url`).

tests/backstop/scenarios.js
```diff
@@ -1,8 +1,84 @@
...
+				...value.scenario,
+			};
+
+			newBackstopTests.push( newBackstopTest );
+		}
+	}
+} );
+
+const backstopTests = storybookStories.map( ( story ) => {
```
aaemnnosttv, line 81: Maybe we can map them all through in the same way here? A lot of the options below can be simplified by just spreading in `...story.parameters.options`, we'd just want to ensure that this can't override the `url` or anything defined in the `Story.scenario` definition.

tests/backstop/scenarios.js
aaemnnosttv: Yeah that was how it was initially defined but having `Story.scenario = {}` is a bit odd to enable a Backstop test as you mentioned so it would be nice to allow for `true` here as well. Not strictly necessary if it's too complicated but would be a nice to have for sure.

tests/backstop/scenarios.js
```diff
@@ -1,8 +1,84 @@
 const storybookHost = require( './detect-storybook-host' );
 const rootURL = `${ storybookHost }iframe.html?id=`;
 const storybookStories = require( '../../.storybook/storybook-data' );
```
aaemnnosttv, line 3: It might be a good idea to rename this too as it's a bit confusing. How about:
```suggestion
const legacyStorybookScenarios = require( '../../.storybook/storybook-data' );
```

aaemnnosttv (changes requested):
Thanks @danielgent – this is looking good! I left a few comments to address and then I think this should be good to go.

tests/backstop/scenarios.js
danielgent: :thinking:   But then the paths won't work with `fs.readFileSync(`, as they're still relative.

The output of `glob.sync( '../assets/js/**/*.stories.js', { cwd: './.storybook' } );` is an array of relative paths 
`[ " '../assets/js/components/adminbar/AdminBarApp.stories.js', ... ]`

tests/backstop/scenarios.js
```diff
@@ -1,8 +1,103 @@
...
+const traverse = require( '@babel/traverse' ).default;
+const csf = require( '@componentdriven/csf' );
+const kebabCase = require( 'lodash/kebabCase' );
+
+const newBackstopTests = [];
+
+const storyFiles = glob
+	.sync( filePaths.stories[ 1 ], { cwd: '.storybook' } )
+	// remove relative path from .storybook dir
+	.map( ( path ) => path.slice( 1 ) );
```
danielgent, line 35: So here I'm using `cwd` but I still need to remove the relative path `..` (as mentioned on the other comment thread)

tests/backstop/scenarios.js
danielgent: Done :+1:

tests/backstop/scenarios.js
danielgent: Done like `label: `${ defaultTitle }/${ value.storyName }`,`

https://github.com/google/site-kit-wp/pull/3782/files#diff-31cf1e014b409e892e3bd1af434eaa4ad7dec46b3b2d37480e525a993ebf6537R96

assets/js/components/Button.stories.js
```diff
@@ -30,6 +30,8 @@ DefaultButton.args = {
 	children: 'Default Button',
 	href: '#',
 };
+// Test empty scenario to create new VRT.
+DefaultButton.scenario = {};
```
danielgent, line 34: I added this to test that the code populated the label correctly.

New VRTs have been added

I can of course remove this before merging (or remove after on a new ticket) or even include another new VRT that may be needed

tests/backstop/scenarios.js
```diff
@@ -1,8 +1,109 @@
...
+const kebabCase = require( 'lodash/kebabCase' );
+const flatten = require( 'lodash/flatten' );
+
+const newBackstopTests = [];
+
+const storyFiles = flatten(
+	filePaths.stories.map( ( stories ) =>
+		glob.sync( stories, { cwd: '.storybook' } )
+	)
+).map( ( path ) => path.slice( 1 ) );
```
danielgent, line 37: I hate that flatten can't be called on the array to make this look cleaner (by being sequential with no inner loop)

tests/backstop/scenarios.js
danielgent: I just don't much about these AST parsing packages.  I'd :heart:  to learn more but it might be out-of-scope of the ticket. I could spend an hour and see where I get? @aaemnnosttv

aaemnnosttv (commented):
This looks good to go, just a few small details that I'll fix/clean up real quick and then we can merge this 👍

tests/backstop/scenarios.js
aaemnnosttv: I've updated it to expand the glob patterns into absolute path patterns, so no need to massage the relative paths anymore 👍

aaemnnosttv:
@googlebot I consent.

aaemnnosttv (approved):
Great!

tests/backstop/scenarios.js
```diff
@@ -19,22 +19,26 @@
...
 
+const storybookDir = path.resolve( __dirname, '../../.storybook' );
 const storyFiles = flatten(
-	filePaths.stories.map( ( stories ) =>
-		glob.sync( stories, { cwd: '.storybook' } )
-	)
-).map( ( path ) => path.slice( 1 ) );
+	storybookConfig.stories
+		.map( ( storiesPattern ) =>
+			path.resolve( storybookDir, storiesPattern )
```
danielgent, line 48: Nice! :+1:

Issue #3761, opened 2021-07-26
Add support for additional survey question type "multi_select"


felixarntz:
When the initial user feedback survey was introduced, scope of the MVP was limited to only support questions of the "rating" type, since that was the only one used for the initial survey. This should now be followed up on, with this issue focusing on the "multi_select" question type.

![Screen Shot 2021-07-26 at 3 51 57 PM](https://user-images.githubusercontent.com/3531426/127069280-ed227901-11b6-49b7-ac15-85c4c10801b8.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new `SurveyQuestionMultiSelect` JS component should be introduced and integrated into the user feedback survey JS infrastructure, rendering a survey question of type "multi_select".
* The UI should look like the above screenshot and follow the "rating" question type where similar (e.g. font, spacing, G icon etc.).
* The available arguments to consider here from the Site Kit Service `/survey/trigger/` API response are fairly similar to those for the "rating" type, as follows:
    * `question.question.answer_choice` contains the list of choices, with `answer_ordinal` (int) and `text` (string), like for the rating question type.
        * In addition, an extra `write_in` boolean property should be supported within each answer choice. If set to `true`, the respective answer should offer a free-text input where the user can type something custom. In that case, the `text` field should be displayed right before the text input. The text input must be non-empty in order to proceed. The maximum character limit should be 100. (This additional free-text input concept is also used in #3760, so this should be coordinated to work similarly, potentially reuse common logic where possible.)
    * In addition, `question.question.min_choices` and `question.question.max_choices` should be supported - if present, these are integers that limit the minimum required and maximum allowed choices to provide (relevant since this is a multi-select question). The default should be 1 minimum, unlimited maximum.
* The data object for answering the question (what needs to be passed to the `answerQuestion` function prop) should look as follows:
    * A single `answer` property, containing an array of objects about the selected answers, each with its `answer_ordinal` (similar to the "rating" question type), and in addition `answer_text` (string with free-text input value, if applicable for the selected answer).
    * Example: `{ answer: [ { answer_ordinal: 2 }, { answer_ordinal: 4, answer_text: 'My custom response' } ] }`
* The new question type should have an example in Storybook where it can be inspected, similar to the existing "rating" type component.

### Additional ACs after QA

[...]

PR #3788, opened 2021-07-31
Enhancement/3761 additional survey question type "multi select"

danielgent:
## Summary

Addresses issue #3761

## Checklist


assets/js/components/surveys/SurveyQuestionMultiSelect.js
```diff
@@ -0,0 +1,179 @@
...
+ */
+import Button from '../Button';
+import Checkbox from '../Checkbox';
+import SurveyHeader from './SurveyHeader';
+import { TextField, Input } from '../../material-components';
+
+const MAXIMUM_CHARACTER_LIMIT = 100;
+
+// Why have this rule enabled if we get this from the APIs *and* send back to the APIs like this?
+/* eslint-disable camelcase */
```
danielgent, line 41: From the AC it looks like this component will be receiving snake-case props.

If we don't allow this then we should convert this when we receive it from the API

assets/js/components/surveys/SurveyQuestionMultiSelect.js
```diff
@@ -0,0 +1,179 @@
...
+	const handleCheck = ( answer_ordinal ) => {
+		const newState = {
+			...selectedValues,
+			[ answer_ordinal ]: {
+				...selectedValues[ answer_ordinal ],
+				selected: ! selectedValues[ answer_ordinal ].selected,
+			},
+		};
+
+		setSelectedValues( newState );
```
danielgent, line 67: There is a lot of filtering and reducing here.  Which makes me think that
1. `useReducer` may be better (or an off-the-shelf hook from a popular React hooks library)
2. Some react-testing-library tests would be good here

assets/js/components/surveys/SurveyQuestionMultiSelect.js
```diff
@@ -0,0 +1,179 @@
...
+				title={ question }
+				dismissSurvey={ dismissSurvey }
+			/>
+
+			<div className="googlesitekit-survey__body">
+				{ choices.map( ( { answer_ordinal, text, write_in } ) => {
+					const answer = selectedValues[ answer_ordinal ];
+
+					return (
+						<div key={ text } style={ { display: 'flex' } } >
```
danielgent, line 124: We don't have any design for this AFAIK.  TBD, I just added this for now so it works

danielgent:
Note there are no tests here. I'd be happy to add react testing library tests (especially if we're thinking of refactoring :+1: )

assets/js/components/surveys/CurrentSurvey.js
```diff
@@ -267,6 +269,8 @@ export default function CurrentSurvey() {
 					choices={ currentQuestion.question.answer_choice }
 					dismissSurvey={ dismissSurvey }
 					question={ currentQuestion.question_text }
+					min_choices={ currentQuestion.min_choices }
+					max_choices={ currentQuestion.max_choices }
```
danielgent, line 273: As commented on other PR, this part was not mentioned in AC or IB https://github.com/google/site-kit-wp/pull/3771/files#r684347927

I would just have a switch statement that chooses which component to render rather than have this `SurveyQuestionComponent`

assets/js/components/surveys/CurrentSurvey.test.js
```diff
@@ -80,6 +80,158 @@ describe( 'CurrentSurvey', () => {
...
+
+		// check correct question loads
+		expect(
+			getByText( 'What are your favorite pizza toppings?' )
+		).toBeInTheDocument();
+
+		// button should be disabled until two options are selected
+		expect( getByRole( 'button', { name: 'Next' } ) ).toHaveAttribute(
+			'disabled'
+		);
```
danielgent, line 112: The tests here are pretty comprehensive.  The only extra tests I would add (if we comprehensively test the MultiSelect in its own test suite) is variations on  `min_choices` and `max_choices` and different amounts of answers

assets/js/components/surveys/SurveyQuestionMultiSelect.js
danielgent: It's too late to do this now as we're doing this everywhere, so I'm just disabling this rule per file :smile:

assets/js/components/surveys/SurveyQuestionMultiSelect.js
danielgent: Now there are comprehensive tests I can refactor this pretty easily! :+1:

assets/js/components/surveys/CurrentSurvey.test.js
```diff
@@ -80,6 +80,158 @@ describe( 'CurrentSurvey', () => {
...
+			{
+				credentials: 'include',
+				method: 'POST',
+				body: {
+					data: {
+						event: {
+							question_answered: {
+								question_ordinal: 1,
+								answer: {
+									answer: [
```
danielgent, line 207: I only just noticed that this key is nested! I'll check this again

danielgent: This is what `SurveyQuestionRatingChoice` does! Leaving as is :shrug:

danielgent: This is what `SurveyQuestionRatingChoice` does! Leaving as is :shrug:

assets/js/components/surveys/SurveyQuestionMultiSelect.js
tofumatt: We tend to disable it per-line… which yes is more effort at first but prevents us from making the mistakes elsewhere. Can you change it to be per-line?

tofumatt: Hmm, while I see the IB mentions using `max_choices`, presumably because that's what the API response looks like, let's have our actual props be the usual case style eg `maxChoices`. Especially as we're manually providing them I think that's fine.

In this case the IB wasn't perfect and I think it's fine to diverge from it 🙂

assets/js/components/surveys/SurveyQuestionMultiSelect.js
```diff
@@ -0,0 +1,199 @@
...
+const SurveyQuestionMultiSelect = ( {
+	question,
+	choices,
+	answerQuestion,
+	dismissSurvey,
+	min_choices = 1,
+	max_choices,
+} ) => {
+	const mappedChoices = choices.map( ( { answer_ordinal, write_in } ) => {
+		const optionalKeys = write_in ? { answer_text: '' } : {};
```
tofumatt, line 51: I think the point of this is to say that a "write-in" key is optional, but can you elaborate on this logic with a comment? It's just a bit unclear to me upon immediately reading this file, so a bit of explanation here would likely help future developers reading it 🙂

assets/js/components/surveys/SurveyQuestionMultiSelect.js
```diff
@@ -0,0 +1,199 @@
...
+	return (
+		<div className="googlesitekit-survey__multi-select">
+			<SurveyHeader title={ question } dismissSurvey={ dismissSurvey } />
+
+			<div className="googlesitekit-survey__body">
+				{ choices.map( ( { answer_ordinal, text, write_in } ) => {
+					const answer = selectedValues[ answer_ordinal ];
+
+					return (
+						<div key={ text } style={ { display: 'flex' } }>
```
tofumatt, line 139: Let's not have style props set on elements unless we 100% have to 😄 

Can we assign a `className` to this instead?

assets/js/components/surveys/SurveyQuestionMultiSelect.js
```diff
@@ -0,0 +1,199 @@
...
+									<Input
+										onChange={ ( event ) =>
+											handleAnswerChange(
+												event,
+												answer_ordinal
+											)
+										}
+										value={ answer.answer_text }
+										disabled={ ! answer.selected }
+										aria-label={ `Text input for option ${ text }` }
```
tofumatt, line 160: Why can't we use a regular label here? It could be wrapped in `VisuallyHidden` if needed, but a normal label tag here seems fine.

assets/js/components/surveys/SurveyQuestionMultiSelect.js
```diff
@@ -0,0 +1,199 @@
...
+					);
+				} ) }
+			</div>
+
+			<div className="googlesitekit-survey__footer">
+				<Button
+					onClick={ handleSubmit }
+					disabled={ isSubmitButtonDisabled }
+				>
+					Next
```
tofumatt, line 174: This should be localised:
```suggestion
					{ __( 'Next', 'google-site-kit' ) }
```

assets/js/components/surveys/CurrentSurvey.test.js
```diff
@@ -80,6 +80,158 @@ describe( 'CurrentSurvey', () => {
 		expect( container ).toMatchSnapshot();
 	} );
 
+	it( "should render a multi select question when the `question_type` is 'multi_select'", async () => {
```
tofumatt, line 83: [I saw you mentioned breaking these up like for #3762](https://github.com/google/site-kit-wp/issues/3762#issuecomment-895417521) 👍🏻

assets/js/components/surveys/SurveyQuestionMultiSelect.js
tofumatt: This should be localised as well, eg. `sprintf( __( 'Text input for option %s', 'google-site-kit' ), text )`

tofumatt (changes requested):
Looks like we're missing some gettext/localisation calls here, and the tests could use the same breaking up as mentioned elsewhere, but the overall approach looks good 👍🏻

assets/js/components/surveys/SurveyQuestionMultiSelect.js
danielgent: I've just put this in 12 times :grimacing:   I don't see this lint rule helping at all but it's done now :see_no_evil:

assets/js/components/surveys/SurveyQuestionMultiSelect.js
danielgent: Ooops. I left that in from some rapid development :see_no_evil: 

FIxing now

danielgent: I'm also going to match the styling of the single select more closely

@asvinb  wrote the IB for this and I like the styling he chose for the text input (we have no designs for any of this!)

**single select**

single select


**multi select now** _will copy above_

multiselect

assets/js/components/surveys/SurveyQuestionMultiSelect.js
danielgent: One thing I meant to bring up, should this be configurable with props?

As the design (the screenshot :wink: ) for 3762 (the open text) shows the text _Submit_ (which I've hardcoded)

But the design for this and the Single Select shows _Next_

I would guess that this should be a prop :thinking:

assets/js/components/surveys/CurrentSurvey.test.js
danielgent: Broken up now! :+1:

assets/js/components/surveys/SurveyQuestionMultiSelect.js
danielgent: I've added a label `VisuallyHidden`. I wasn't aware that there was a difference to screen readers

tofumatt: Screen readers _will_ use the `aria-label`, but the label tag is arguably semantically a bit more correct and easier for things to parse. It's the better default is all 🙂

assets/js/components/surveys/SurveyQuestionMultiSelect.js
tofumatt: Hmm, indeed. I don't know if it needs to be a prop exactly, but that's an implementation detail—I think in this case it would make sense to be one.

But mainly: it makes sense that it should be `Submit` when it's the last question (when the answer ordinal is the same as the total length of questions) and `Next` otherwise.

We don't mention that in the ACs or IBs, but I think it's clear how to do that—in `CurrentSurvey` we should check to see if `questions?.length === currentQuestionOrdinal`. If so, the button should be `Submit`. Otherwise: `Next`.

assets/js/components/surveys/SurveyQuestionMultiSelect.js
```diff
@@ -0,0 +1,238 @@
...
+								className="googlesitekit-survey__multi-select__choice"
+							>
+								<Checkbox
+									checked={ answer.selected }
+									onChange={ () =>
+										handleCheck( answer_ordinal )
+									}
+									// eslint-disable-next-line camelcase
+									value={ `${ answer_ordinal }` }
+									id={ text }
```
tofumatt, line 163: What does this ID end up looking like? Is `text` a long string or just a confusing attribute name? 😆

assets/js/components/surveys/SurveyQuestionMultiSelect.js
```diff
@@ -0,0 +1,238 @@
...
+								key={ text }
+								className="googlesitekit-survey__multi-select__choice"
+							>
+								<Checkbox
+									checked={ answer.selected }
+									onChange={ () =>
+										handleCheck( answer_ordinal )
+									}
+									// eslint-disable-next-line camelcase
+									value={ `${ answer_ordinal }` }
```
tofumatt, line 162: I'm guessing this prop also needs to be a string? If it's our component we could fix that, but if not let's at least leave a comment saying why we do it.

assets/js/components/surveys/SurveyQuestionMultiSelect.js
```diff
@@ -0,0 +1,238 @@
...
+		<div className="googlesitekit-survey__multi-select">
+			<SurveyHeader title={ question } dismissSurvey={ dismissSurvey } />
+
+			<div className="googlesitekit-survey__body">
+				{
+					// eslint-disable-next-line camelcase
+					choices.map( ( { answer_ordinal, text, write_in } ) => {
+						const answer = selectedValues[ answer_ordinal ];
+						// eslint-disable-next-line camelcase
+						const uniqueID = `googlesitekit-survey__multi-select-${ answer_ordinal }-${ text }`;
```
tofumatt, line 149: We can create unique instance IDs using hooks instead, we do it elsewhere: https://github.com/google/site-kit-wp/blob/ee5e21191618450ad6a10332400905e81f7a16cb/assets/js/components/Dialog.js#L60 🤷🏻‍♂️

tofumatt (changes requested):
Looks good but there's still a few tweaks and questions from me.

I think I've addressed all the questions you left, but please ping me if there's something unclear or if you want to chat about any of my comments. 👍🏻

assets/js/components/surveys/SurveyQuestionMultiSelect.js
danielgent: Done in this PR :+1: 

If you're ok with this implementation then we could merge this PR first, then I can update the other PRs

assets/js/components/surveys/CurrentSurvey.js
```diff
@@ -267,6 +269,8 @@ export default function CurrentSurvey() {
 					choices={ currentQuestion.question.answer_choice }
 					dismissSurvey={ dismissSurvey }
 					question={ currentQuestion.question_text }
+					minChoices={ currentQuestion.minChoices }
```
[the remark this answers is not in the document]
danielgent: Ooops!

assets/js/components/surveys/SurveyQuestionMultiSelect.js
danielgent: Exactly that!  I'm happy to change `Checkbox` but changing core components is something I wouldn't do lightly :smile:

assets/js/components/surveys/SurveyQuestionMultiSelect.js
danielgent: I can't use that here though as it's in a loop.  So not only would the Ids all be the same, but `eslint-plugin-react` won't allow this (and I'm not sure what would actually happen in React! :sweat_smile: )

I could create one `instanceID` and then append `answer_ordinal` to it

Alternatives:
1. Create a new inner Choice component (like on the SingleSelect PR)
1. Wrap the `TextField` inside the `label` element.

assets/js/components/surveys/SurveyQuestionMultiSelect.js
```diff
@@ -0,0 +1,236 @@
...
+									id={ text }
+									name={ text }
+								>
+									{ text }
+								</Checkbox>
+								{
+									// eslint-disable-next-line camelcase
+									write_in && (
+										// eslint-disable-next-line jsx-a11y/label-has-for
+										<label>
```
danielgent, line 171: This works. Tests pass and with some slight CSS changes it is styled the same.

`jsx-a11y/label-has-for` is  deprecated (see https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-for.md)

With `label-has-associated-control` it is possible (and recommended!) to configure it to recognise React component children as native form HTML inputs

assets/js/components/surveys/SurveyQuestionMultiSelect.js
danielgent: See other comment, I went with 2.

assets/js/components/surveys/SurveyQuestionMultiSelect.js
danielgent: As you wrote on the other PR, this can't have spaces  :smile: 

It's a required prop in `Checkbox` although from the HTML spec it doesn't look required as it's only for labels :thinking: 

For now I've just swapped the white space for dashes. Let me know if there's a better solution!

Screenshot from 2021-08-11 12-41-41

assets/js/components/surveys/SurveyQuestionMultiSelect.js
tofumatt: Ah right, whoops, didn't put together the loop piece.

We don't tend to wrap components with labels but I've not problem with that 👍🏻

tofumatt (approved):

tofumatt:
@googlebot I consent.

Issue #3378, opened 2021-05-19
Add tests for broken stories


aaemnnosttv:
## Feature Description

We use Storybook as the foundation of our visual regression tests but we also have many more stories than we have VRT scenarios. Currently VRT is the only test we have in place to prevent Storybook from being broken, but it only catches stories that have corresponding VRT scenarios which are almost entirely all legacy stories. It doesn't make sense to enable VRT for every story, but we should at least still be able to require that no stories are broken which is a problem that keeps coming up. 

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new `test:storybook` command should be created, which when run will check every story to see if it's broken or not
    * The command should list all broken stories by default although an option to fail on the first one found would be nice
    * If a broken story is found, the command should return with a non-zero exit code
    * Any new test scripts/config/files should go in a new `tests/storybook` directory
    * A broken story is simply one that results in a `sb-show-errordisplay` class on the `body` element in the preview iframe – this frame should always be loaded directly (like it is in VRT) rather than in the larger view we normally use
* The new command should be integrated into our current GitHub actions so that broken stories cause the Storybook workflow to fail 

## Implementation Brief
* Install the `@storybook/addon-storyshots-puppeteer` package as dev-dependency along with `puppeteer` as per the documentation. https://storybook.js.org/addons/@storybook/addon-storyshots-puppeteer
* Create `.storybook/Stories.test.js` which contains the code sample in the "Specifying the storybook URL" section of the documentation, using the  local static build of storybook found in `/dist` via `file://`.
* The `setupTimeout` and `testTimeout` values are `15s` by default and this can be reduced to `5s` as per the documentation.
* Update the `parameters` object in `.storybook/preview.js` to include the asynchronous `puppeteerTest` function (refer to documentation), where the function checks whether the `body` of the rendered story contains the `sb-show-errordisplay` class name. For e.g:
    ```js
    expect( await page.$eval( 'body', ( el ) => el.classList.contains( 'sb-show-errordisplay' ) ) ).toBe( false );
    ```
* Update `package.json` to include a new command, `test:storybook` which runs:
    ```
    npm run test:js -- .storybook/Stories.test.js
    ```
* Using `.github/workflows/storybook.yml`,
    * Add a new job `test-storybook` which uses the code from the existing `build-storybook` job with the following changes:
        * Instead of running `npm run build:storybook`, run the `npm run test:storybook` command.
[...]

PR #3795, opened 2021-08-03
Add tests for broken stories

danielgent:
## Summary

Addresses issue #3378

## Checklist


.storybook/stories.test.js
```diff
@@ -0,0 +1,13 @@
...
+import { puppeteerTest } from '@storybook/addon-storyshots-puppeteer';
+import path from 'path';
+
+initStoryshots( {
+	suite: 'Puppeteer storyshots',
+	test: puppeteerTest( {
+		// eslint-disable-next-line sitekit/acronym-case
+		storybookUrl: `file://${ path.resolve( __dirname, '../dist' ) }`,
+		setupTimeout: 5000,
+		testTimeout: 5000,
```
danielgent, line 11: To check! are these millseconds or seconds :thinking: 

The docs aren't clear

stories/module-analytics-settings.stories.js
```diff
@@ -604,49 +604,50 @@ storiesOf( 'Analytics Module/Settings', module )
...
-	)
-	.add(
-		'Existing Tag w/o access, GTM property w/o access',
-		usingGenerateGTMAnalyticsPropertyStory( {
-			useExistingTag: true,
-			gtmPermission: false,
-			gaPermission: false,
-		} )
-	);
+	// TODO - rewrite like above stories
```
danielgent, line 607: These must have been broken for a long time.  As discussed on the ticket, I'd rather pull out the registry setup function from `generateGTMAnalyticsPropertyStory`

tests/js/jest.storybook-config.js
```diff
@@ -0,0 +1,6 @@
+const jestConfig = require( './jest.config' );
+
+module.exports = {
+	...jestConfig,
+	testMatch: [ '<rootDir>/.storybook/?(*.)test.js' ],
```
danielgent, line 5: This seemed like the easiest solution

We need to include this path in `testMatch` to run the test AFAIK *BUT* we don't want this test ran when running them all normally

I'm sure there are other ways to do this!

stories/module-analytics-settings.stories.js
danielgent: Commenting the tests out so I can see if the PR checks pass

.storybook/stories.test.js
danielgent: It looks like milliseconds from inspecting the code (docs don't say)

https://github.com/storybookjs/storybook/search?q=setupTimeout

looks like ms

.github/workflows/storybook.yml
```diff
@@ -147,3 +147,33 @@ jobs:
           git pull --no-edit --quiet
           git commit -m "Remove storybook for pull/${PULL_REQUEST_NUMBER}."
           git push origin gh-pages
+
+  test-storybook:
+    name: Test Storybook
+    runs-on: ubuntu-latest
+    needs: build-storybook
```
danielgent, line 154: Previously the new storybook test passed using this. Now it's failing because it doesn't have the built output :thinking:

.storybook/stories.test.js
aaemnnosttv: I would expect milliseconds as well.

aaemnnosttv:
@googlebot I consent.

.github/workflows/storybook.yml
```diff
@@ -147,3 +147,38 @@ jobs:
           git pull --no-edit --quiet
           git commit -m "Remove storybook for pull/${PULL_REQUEST_NUMBER}."
           git push origin gh-pages
+
+  test-storybook:
+    name: Test Storybook
+    runs-on: ubuntu-latest
+    # Tried this but keeps failing
+    # needs: build-storybook
```
danielgent, line 155: Yesterday this step kept failing when I used the `needs` step (although I'm sure it was working before :thinking:  )  @aaemnnosttv 

I can look into this more today, and see if there's anything problematic with this, or if I'm doing something wrong 

An alternative would be to combine the build and test steps (like we combined the various JS steps).  If the test fails then I suppose we don't want to do anything else with storybook and want to bail as quickly as possible!

danielgent:
All the e2e tests are failing :thinking: 

I'll check this locally. They are passing on other PRs I've just made. Probably the puppeteer changes I made are affecting all uses of puppeteer

danielgent:
I'm not sure why the e2es are failing.  Other recent PRs have them passing :thinking: 

Locally I'm getting `develop` failing with the same errors as this

So I'm not sure why this PR and local are failing for me *but* another PR is working

package.json
```diff
@@ -200,6 +203,7 @@
 		"postcss-loader": "^3.0.0",
 		"postcss-preset-env": "^6.7.0",
 		"prettier": "npm:wp-prettier@^2.2.1-beta-1",
+		"puppeteer": "^10.1.0",
```
aaemnnosttv, line 206: Is this needed since we already have puppeteer installed? 🤔 This may be why there are problems in E2E as well.

danielgent: @aaemnnosttv Nice!  I was trying to work this out for ages!  :joy: 

That explains why everything was failing for me locally even on develop :joy:   I should have nuked `node_modules` and tried again.

All working now :+1:

danielgent:
All tests now passing :+1:

aaemnnosttv (commented):
Thanks @danielgent – this is really close, see my last comments below.

.storybook/stories.test.js
```diff
@@ -0,0 +1,13 @@
+import initStoryshots from '@storybook/addon-storyshots';
```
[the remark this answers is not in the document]
danielgent: If you're interested, I did have a quick fifteen-minute attempt at linting for this with https://www.npmjs.com/package/eslint-plugin-header

Sadly our header looks too complex, and it's not particularly easy to test out this plugin! :stuck_out_tongue:    I tried using regex to have lines that match anything (for the lines that have the component name) but I didn't want to spend more time without  it being a ticket

.storybook/stories.test.js
```diff
@@ -0,0 +1,13 @@
...
+import { puppeteerTest } from '@storybook/addon-storyshots-puppeteer';
+import path from 'path';
```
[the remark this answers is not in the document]
danielgent: Have you looked into configuring this in eslint?  Would be a nice-to-have

I had a look at https://eslint.org/docs/rules/sort-imports but that's not what we want at all

https://github.com/lydell/eslint-plugin-simple-import-sort looks interesting but I don't see how it can have the complexity that we want
* differentiating between WordPress, Node, and external dependencies, and then adding headers

.storybook/stories.test.js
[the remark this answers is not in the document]
aaemnnosttv: Yeah, it would be good to add something just to ensure the standard header is there. As for the dependency groups, there is are already rules for this in the WP ESLint plugin, but we disabled it a while back because it couldn't understand our aliases as being internal dependencies.

.storybook/stories.test.js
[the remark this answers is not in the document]
aaemnnosttv: See https://github.com/WordPress/gutenberg/blob/trunk/packages/eslint-plugin/docs/rules/dependency-group.md 😉

aaemnnosttv (approved):
Awesome! 🎉

Issue #3618, opened 2021-06-23
Cannot read property 'slug' of undefined


jamesozzie:
## Bug Description

There have been 3 reports of a `Cannot read property 'slug' of undefined` error, first occurring from 1.34.1. Unknown at what stage this occurs yet, with no additional insights. 

https://wordpress.org/support/topic/depannage-du-tableau-de-bord-google-kit-site/ - Open - No SH info
https://wordpress.org/support/topic/site-kite-error/ - Open - No SH info
https://wordpress.org/support/topic/site-kit-error-8 - Open - No SH info

## Additional Context

 - PHP Version: 
 - Plugin Version [e.g. 22]

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* 

## Implementation Brief

* 

### Test Coverage

* 

### Visual Regression Changes

* 

## QA Brief

* 

## Changelog entry

*

bethanylang:
Noting that we've had a fourth report of this today, but no real details to go on just yet. Monitoring as a possible code yellow for the week of June 28.

jamesozzie:
Based on reviewing the Site Health information provided by 1 impacted user - with their topic since resolved without any action, the only plugin with a known [configuration conflict](https://imgur.com/7OyyYyI) is iThemes. There is an additional plugin extension active (Log cleaner for iThemes Security) which is referenced in the shared console errors.  To be tested while awaiting additional insights.

abdullah1908:
@jamesozzie I were unable to reproduce this issue & I have done following activities:
- Tested with uBlock Chrome extension
- Installed Astra theme along with the plugins as per the user SH info
- Tested with iThemes Security

jamesozzie:
[Site Health information](https://docs.google.com/forms/d/1JTQmphDlk9YgHpDoFRbHXpkDqv8-QmzTT7MDR4AstyQ/edit?urlBuilderDomain=google.com#response=ACYDBNjtD7RoNTXt8n1joHe0DOz2wrVdTIgpMQMD_o2wOTjuDwGIRPo4VHOHIHYfiDzcjFo) added from [latest report](https://wordpress.org/support/topic/error-in-setup-wizard/#new-topic-0) to the main description. 

[...]

PR #3800, opened 2021-08-04
Fix error in combineWidgets when no widgets given

aaemnnosttv:
## Summary

Addresses issue #3618

## Relevant technical choices

- See comments below

## Checklist

### DIFF
--- a/assets/js/googlesitekit/widgets/util/combine-widgets.test.js
+++ b/assets/js/googlesitekit/widgets/util/combine-widgets.test.js
@@ -232,4 +232,13 @@ describe( 'combineWidgets', () => {
 			expected
 		);
 	} );
+
+	it( 'works with no widgets', () => {
+		const widgets = [];
+		const widgetStates = [];
+		const layout = getWidgetLayout( widgets, widgetStates );
+		expect( () =>
+			combineWidgets( widgets, widgetStates, layout )
+		).not.toThrow();
+	} );
 } );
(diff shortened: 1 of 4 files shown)


assets/js/googlesitekit/widgets/util/combine-widgets.test.js
```diff
@@ -232,4 +232,13 @@ describe( 'combineWidgets', () => {
...
 		);
 	} );
+
+	it( 'works with no widgets', () => {
+		const widgets = [];
+		const widgetStates = [];
+		const layout = getWidgetLayout( widgets, widgetStates );
+		expect( () =>
+			combineWidgets( widgets, widgetStates, layout )
+		).not.toThrow();
```
aaemnnosttv, line 242: You can see this test fail with the error in the issue before the early returns were added here https://github.com/google/site-kit-wp/issues/3618#issuecomment-892488288

assets/js/googlesitekit/widgets/util/get-widget-layout.js
```diff
@@ -119,12 +119,15 @@ function getNextActiveWidget( offset, widgets, widgetStates ) {
  *                  each active widget.
  */
 export function getWidgetLayout( widgets, widgetStates ) {
-	let counter = 0;
-	let rowIndex = 0;
-
 	let columnWidths = [];
 	const rowIndexes = [];
 
+	if ( ! widgets?.length ) {
```
aaemnnosttv, line 125: This return wasn't necessary to fix this but seemed like a good idea to add.

aaemnnosttv (commented):
The problem was due to `shouldCombineAllWidgets` defaulting to `true` and assuming there are always 1 or more widgets given. Neither return value really makes sense when there are no widgets though because there is nothing to combine, so the condition was added to return early before it got to that point.

tofumatt (approved):

tofumatt:
Not sure why the tests aren't running here, but running them locally seemed good 👍🏻

Issue #1703, opened 2020-06-22
1.10.0: frontent webfont configuration interferes with site font configuration


lkraav:
## Bug Description

Our UI library enqueues `'sans-serif' => 'Roboto:300,300i,400,400i,700,700i,900'`

This plugin adds hardcoded Roboto:500, which matches some of our upstream vendor components styling in an unintended fashion.

https://github.com/google/site-kit-wp/blob/8ffa689fd0f30d5ed83642c029cfcb89eb26ad32/includes/Core/Assets/Assets.php#L176-L179

## Steps to reproduce

:point_up: 

## Solution

I'm not sure.

a) I couldn't find any setting to disable loading Site Kit on frontend. I'm not sure if anybody is actually looking at it, perhaps I should be able to just disable loading.

b) Make font list filterable, so I could remove the interfering weight?

c) All of the above?

d) Any other ideas?

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The Google font families definition that Site Kit uses by default should be filterable, relying on a new `googlesitekit_font_families` filter.
* If the filter returns an empty array, the entire process of enqueuing the fonts should be short-circuited - basically in that case it should be assumed that another plugin/theme is taking care of it in their own way.

## Implementation Brief
* Using `includes/Core/Assets/Assets.php`
    * In the `register` method, add a new filter via the `add_filter` function, named `googlesitekit_font_families`.
    * The second parameter of the `add_filter` function should be an anonymous function having `$fonts` as only parameter.
        * Within the function, check if `$fonts` is not an `array` and is `null`. If that is the case, then return the existing fonts array, i.e the values of the `$font_families` variable.
        * If `$fonts` is an array and is not null, return `$fonts`.
    * In the `enqueue_fonts` method, get the values for the `$font_families` variable by applying the `googlesitekit_font_families` filter the `apply_filters` function and having `null` as second parameter.

### Test Coverage
* No new tests to be added.

### Visual Regression Changes
* N/A

## QA Brief
* Within the active theme's `functions.php`, insert the following snippet:
	```php
	add_filter( 'googlesitekit_font_families', function() {
		return null;
	} );
	```
* Check on the FE via Devtools to see ifno requests are made to load the `Roboto` and `Google Sans` fonts.

## Changelog entry

* Update Google fonts loaded by Site Kit to be filterable.

tofumatt:
IB ✅

johnPhillips:
[...]

PR #3801, opened 2021-08-04
Add filterable fonts on the FE.

asvinb:
## Summary

Addresses issue #1703 

## Checklist


includes/Core/Assets/Assets.php
```diff
@@ -212,10 +226,7 @@ public function enqueue_fonts() {
 
 		$this->fonts_enqueued = true;
 
-		$font_families = array(
-			'Google+Sans:300,300i,400,400i,500,500i,700,700i',
-			'Roboto:300,300i,400,400i,500,500i,700,700i',
-		);
```
aaemnnosttv, line 218: I don't think this line needed to change, we just want to add the filter afterwards to allow it to be filterable. Then after the filter is applied, if it's empty, we essentially abort the enqueue as mentioned in the ACs.

includes/Core/Assets/Assets.php
```diff
@@ -212,10 +226,7 @@ public function enqueue_fonts() {
 
 		$this->fonts_enqueued = true;
```
aaemnnosttv, line 227: This property should also not be set until after we know for sure that SK will be enqueuing the fonts – ie. after we've applied the filter and ensured it's not empty.

includes/Core/Assets/Assets.php
```diff
@@ -108,6 +108,20 @@ function( $tag, $handle ) {
 			2
 		);
 
+		add_filter(
+			'googlesitekit_font_families',
```
aaemnnosttv, line 112: While this is mentioned in the IB, it's not really necessary and makes it a bit more complicated. Let's keep it simple and allow the array to be filtered by providing the initial value to the filter directly.

aaemnnosttv (changes requested):
Thanks @asvinb! It seems perhaps there was some misunderstanding in the IB (which was approved so not your fault 😄 ) but there are a few changes to make.

See my comments below and let me know if you have any questions.

asvinb:
Thanks @aaemnnosttv . I've updated the PR. Can you check if everything's ok now? 😅

includes/Core/Assets/Assets.php
```diff
@@ -206,17 +206,19 @@ public function enqueue_asset( $handle ) {
 	 * @since 1.0.0
 	 */
 	public function enqueue_fonts() {
-		if ( $this->fonts_enqueued ) {
-			return;
-		}
```
aaemnnosttv, line 211: Let's keep this condition to avoid unnecessarily invoking the filter when it isn't going to be output anyways. My last comment was just about moving the setting of the property until after the filter to avoid a situation where it could be set but no fonts were actually enqueued.

aaemnnosttv (commented):
Thanks @asvinb – this is super close, just the one change left and I think this should be good to go.

includes/Core/Assets/Assets.php
asvinb: 👍 Condition restored at the beginning of the method.

aaemnnosttv (approved):
Great!

Issue #3625, opened 2021-06-23
Add Idea Hub feature tour on WordPress posts page


felixarntz:
## Feature Description

In addition to #3524, another feature tour should showcase the possible Idea Hub interactions on the WordPress posts page.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

A new tooltip tour `ideaHubPostsTable` should be displayed on the WordPress admin posts list table _if_ the `ideaHubModule` feature flag is enabled. Additional requirements to show the tooltip:
* The Idea Hub module is active.
* The user has successfully completed Idea Hub setup.
* The user has created at least one draft from Idea Hub

**Tooltip**
* Position: point to the first saved draft entry on the Posts page 
* Heading: _Get started writing about your saved ideas_ 
* Body: _Drafts you created from Idea Hub within Site Kit appear here for easy access._ 

Note that this will require to load the feature tooltip infrastructure on a non-Site Kit screen for the first time. The JS files to do that should for now only be loaded if necessary (i.e. if the above conditions are fulfilled and if the above feature tour hasn't been dismissed yet).

## Implementation Brief

Our high level approach is as follows:

1.  Add a tooltip tour and a new tooltip tour type (context) for posts list.
1.  Add dependencies to our existing JS file on the `edit.php` screen to run the tooltip tour `checkRequirements().` and to render the `<FeatureTour />`
1.  Create a container for rendering a js app.
1.  Add classes to each row of a draft idea hub post so that we can attach a tour to it.
1.  Render our tooltip tours on the edit.php screen

**Before you begin**, keep in mind that you may need to further update either the javascript file or dependencies list to add any missing javascript dependencies we need for loading the functionality as you progress with the execution of this ticket.

Here's how we achieve it:

* Rename `js/googlesitekit-idea-hub-post-list-notice.js` to `js/googlesitekit-idea-hub-post-list.js`:
  * in `includes/Modules/Idea_Hub.php`
  * in `webpack.config.js`

* In `includes/Modules/Idea_Hub.php` 
  * For the script `'googlesitekit-idea-hub-post-list-notice'` (which we just renamed to `googlesitekit-idea-hub-post-list`) add to the `dependencies`: 'googlesitekit-modules-idea-hub'. This will give our selectors in the `checkRequirements` to follow.
[...]

PR #3806, opened 2021-08-05
Enhancement/3625 idea hub feature tours on wp pages

eugene-manuilov:
## Summary

Addresses issue #3625

## Relevant technical choices

* The only thing that is missing is the storybook. Decided not to include it because it will require recreating the whole WordPress posts page which seems to be overkill.

## Checklist


assets/js/modules/idea-hub/index.js
```diff
@@ -58,6 +58,10 @@ export const registerModule = ifIdeaHubIsEnabled( ( modules ) => {
 } );
 
 export const registerWidgets = ifIdeaHubIsEnabled( ( widgets ) => {
+	if ( ! widgets ) {
+		return;
+	}
```
aaemnnosttv, line 63: Is this because we're loading the Idea Hub module on this page but not Widgets?

includes/Modules/Idea_Hub.php
```diff
@@ -176,6 +177,18 @@ public function register() {
...
+
+			/**
+			 * Adds a special class name to idea posts.
+			 */
+			add_filter( 'post_class', $this->get_method_proxy( 'update_post_classes' ), 10, 3 );
+
+			add_action(
+				'admin_footer',
+				function() {
+					echo '<div id="js-googlesitekit-post-list" class="googlesitekit-page"></div>';
```
aaemnnosttv, line 189: `Screen`s use the `googlesitekit-page` class on their render targets, but we shouldn't use that here since we're not rendering a screen. I'm not sure if there's anything actually going in here, but if there is, it should always be wrapped with a `googlesitekit-plugin` class.
```suggestion
					echo '<div id="js-googlesitekit-post-list" class="googlesitekit-plugin"></div>';
```

includes/Modules/Idea_Hub.php
```diff
@@ -650,12 +663,15 @@ protected function setup_assets() {
...
+				'googlesitekit-idea-hub-post-list',
 				array(
-					'src'           => $base_url . 'js/googlesitekit-idea-hub-post-list-notice.js',
+					'src'           => $base_url . 'js/googlesitekit-idea-hub-post-list.js',
 					'load_contexts' => array( Asset::CONTEXT_ADMIN_POSTS ),
 					'dependencies'  => array(
+						'googlesitekit-i18n',
 						'googlesitekit-datastore-user',
+						'googlesitekit-datastore-ui',
+						'googlesitekit-modules-idea-hub',
```
aaemnnosttv, line 674: Is the module needed here? I don't see it used in the tour.

includes/Modules/Idea_Hub.php
```diff
@@ -730,4 +746,37 @@ private function is_idea_post( $post_id ) {
...
+	private function update_post_classes( $classes, $class, $post_id ) {
+		global $post_type;
+
+		$current_screen = get_current_screen();
+		if ( is_null( $current_screen ) || 'edit-post' !== $current_screen->id || 'post' !== $post_type ) {
+			return $notices;
+		}
+
+		if ( $this->is_idea_post( $post_id ) ) {
+			$classes[] = 'googlesitekit-idea-hub__draft';
```
aaemnnosttv, line 768: Do we want to add a draft-specific class to all Idea Hub posts? Is this only relevant for drafts or is it for all posts created from IH ideas? Maybe it should be `post` instead or only applied if it is currently a draft?

includes/Modules/Idea_Hub.php
```diff
@@ -730,4 +746,37 @@ private function is_idea_post( $post_id ) {
...
+
+			if ( ! wp_style_is( 'googlesitekit-admin-css' ) ) {
+				$assets = new Assets( $this->context );
+				// Enqueue fonts.
+				$assets->enqueue_fonts();
+				// Enqueue base admin screen stylesheet.
+				$assets->enqueue_asset( 'googlesitekit-admin-css' );
+			}
```
aaemnnosttv, line 776: We should avoid this because the class keeps instance properties for ensuring fonts are only enqueued once. Also, `enqueue_asset` would cause all assets to be registered again for a similar reason.

Without using a `Screen` I think our best option here is to fire an action that the main `Assets` instance can respond to.

Lastly, I think it would be a good idea to separate this into two methods – one for just adding the classes, and another that handles the assets.

includes/Modules/Idea_Hub.php
```diff
@@ -730,4 +746,37 @@ private function is_idea_post( $post_id ) {
...
+	 *
+	 * @param array $classes An array of post class names.
+	 * @param array $class An array of additional class names added to the post.
+	 * @param int   $post_id The post ID.
+	 * @return array An array of post class names.
+	 */
+	private function update_post_classes( $classes, $class, $post_id ) {
+		global $post_type;
+
+		$current_screen = get_current_screen();
```
aaemnnosttv, line 762: Is this function always defined on front end page loads? I feel like it might not be sometimes.

aaemnnosttv (changes requested):
Thanks @eugene-manuilov – this is looking good, just a few questions and suggestions for you in addition to some minor things to address.

assets/js/components/TourTooltips.js
```diff
@@ -122,6 +122,10 @@ export default function TourTooltips( {
 		status,
 		type,
 	} ) => {
+		if ( ! gaEventCategory ) {
+			return;
+		}
```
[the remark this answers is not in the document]
eugene-manuilov: We don't have a ga event category for this feature tour. That's why I needed to add this. Asked Felix to provide it: https://github.com/google/site-kit-wp/issues/3625#issuecomment-895140526

assets/js/modules/idea-hub/index.js
eugene-manuilov: Yes, correct.

includes/Modules/Idea_Hub.php
eugene-manuilov: We check whether the idea hub module is active and connected.

assets/js/feature-tours/idea-hub-posts-table.js
```diff
@@ -0,0 +1,70 @@
...
+			.isModuleActive( 'idea-hub' );
+		const isIdeaHubModuleConnected = registry
+			.select( CORE_MODULES )
+			.isModuleConnected( 'idea-hub' );
+
+		if ( ! isIdeaHubModuleActive || ! isIdeaHubModuleConnected ) {
+			return false;
+		}
+
+		return true;
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

includes/Modules/Idea_Hub.php

includes/Modules/Idea_Hub.php
eugene-manuilov: Updated to use the `googlesitekit-idea-hub__post` class.

includes/Modules/Idea_Hub.php
eugene-manuilov: Updated to make sure it runs only on in the admin.

assets/js/components/Root/index.js
```diff
@@ -34,7 +34,15 @@ import { FeatureToursDesktop } from '../FeatureToursDesktop';
...
 import CurrentSurveyPortal from '../surveys/CurrentSurveyPortal';
 
-export default function Root( { children, registry, viewContext = null } ) {
+export default function Root( props ) {
+	const {
+		children,
+		registry,
+		viewContext = null,
+		withSurveys,
+		withPermissionsModal,
```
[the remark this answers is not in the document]
eugene-manuilov: We don't want to display surveys on non-sitekit pages, right? Plus, and we don't need permissions modals when we are not

assets/js/googlesitekit-idea-hub-post-list.js
```diff
@@ -47,3 +52,20 @@ domReady( () => {
...
+	const renderTarget = document.getElementById(
+		'js-googlesitekit-post-list'
+	);
+
+	if ( renderTarget ) {
+		render(
+			<Root
+				viewContext={ VIEW_CONTEXT_POSTS_LIST }
+				withPermissionsModal={ false }
+				withSurveys={ false }
```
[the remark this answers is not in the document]
eugene-manuilov: We pull the current survey and show it in the `CurrentSurveyPortal` component which will not be rendered if `withSurveys` equals `false`. So, answering your question, it will prevent the added survey from showing on this page.

includes/Modules/Idea_Hub.php
eugene-manuilov: > Without using a Screen I think our best option here is to fire an action that the main Assets instance can respond to.

Added it and updated a few other classes to use the same approach instead of repeating calling `enqueue_fonts()` and `enqueue_asset(...)` methods.

> Lastly, I think it would be a good idea to separate this into two methods – one for just adding the classes, and another that handles the assets.

In this case, these two methods will be 99% identical and will just repeat each other. Not sure whether we really need it. #DRY

assets/js/components/Root/index.js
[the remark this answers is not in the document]
aaemnnosttv: > We don't want to display surveys on non-sitekit pages, right?

Yes, but simply rendering the `<CurrentSurvey>` doesn't trigger a survey which is the more expensive part since it makes a request. Maybe we should make this controllable by a prop but it seems a bit out of scope here without a concrete need.

> Plus, we don't need permissions modals when we are not on a sitekit page.

I'm not sure about that – the permissions modal will be triggered by any insufficient scopes – not only an on-demand request to create something. Again, maybe we should make this conditional but I think we should move this to a new issue to discuss unless needed here.

> These two props help us to turn off unnecessary functionality and avoid a need to load additional assets.

We should be careful to not be too clever to optimize which assets are loaded based on what we know to be used within the component/module. Instead the module should declare it's dependencies and let WP enqueue them as needed. It sounds like maybe you're trying to avoid loading the `core/user` datastore here but I'm not sure we can do that.

assets/js/components/TourTooltips.js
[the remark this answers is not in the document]
aaemnnosttv: Thanks – I think we just need to add that to the tour then this condition can be removed. We want to make sure that prop continues to be required.

assets/js/googlesitekit-idea-hub-post-list.js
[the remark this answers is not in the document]
aaemnnosttv: My mistake, I was referring to the feature tour 🤦‍♂️ 

As mentioned above, I'm not sure these extra props are needed just yet – for surveys in particular they wouldn't be triggered here so I'm not sure rendering the `CurrentSurvey` would be a problem as there wouldn't be one without being triggered anyways 🙂

assets/js/components/Root/index.js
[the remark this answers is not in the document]
eugene-manuilov: :+1: Reverted.

assets/js/components/TourTooltips.js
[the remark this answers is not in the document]
eugene-manuilov: Added an empty property for now.

assets/js/googlesitekit-idea-hub-post-list.js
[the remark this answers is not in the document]
eugene-manuilov: Ok, removed.

includes/Modules/Idea_Hub.php
aaemnnosttv: Yes of course - what I mean is that `get_current_screen` isn't normally loaded on the front where this filter is also invoked. I see you've added a new guard for the frontend here so this is all good now 👍

includes/Core/Util/Activation_Notice.php
```diff
@@ -96,8 +96,7 @@ function( $hook_suffix ) {
 				 */
 				unset( $_GET['activate'] ); // phpcs:ignore WordPress.Security.NonceVerification, WordPress.VIP.SuperGlobalInputUsage
 
-				$this->assets->enqueue_fonts();
-				$this->assets->enqueue_asset( 'googlesitekit-admin-css' );
+				Assets::enqueue_styles( 'googlesitekit-admin-css' );
 				$this->assets->enqueue_asset( 'googlesitekit-activation' );
```
aaemnnosttv, line 100: While this solves our problem, it looks wrong in the context of other usage like this that has access to the `Assets` instance directly.

includes/Core/Assets/Assets.php
```diff
@@ -99,6 +99,16 @@ public function register() {
 		add_action( 'admin_enqueue_scripts', $register_callback );
 		add_action( 'wp_enqueue_scripts', $register_callback );
 
+		add_action(
+			'googlesitekit_enqueue_styles',
+			function( $handle ) {
+				// Enqueue fonts.
+				$this->enqueue_fonts();
+				// Enqueue base admin screen stylesheet.
+				$this->enqueue_asset( $handle );
```
aaemnnosttv, line 108: I feel like this shouldn't do more than call `->enqueue_asset( $handle )`, it isn't really clear that this also loads fonts - the problem there of course is that it doesn't currently support loading fonts as a normal asset although I think given the tools we have now we could make that possible (the problem here is that it uses a style for AMP and a script for web). I don't want to increase the scope of this issue any farther though. I suppose we could go with this for now and then come back to it in a separate issue.

One alternative would be to provide the `Assets` instance to module classes but I don't see a great way to do that.

Thoughts @felixarntz ?

includes/Core/Assets/Assets.php
```diff
@@ -99,6 +99,16 @@ public function register() {
 		add_action( 'admin_enqueue_scripts', $register_callback );
 		add_action( 'wp_enqueue_scripts', $register_callback );
 
+		add_action(
+			'googlesitekit_enqueue_styles',
```
aaemnnosttv, line 103: This may interfere with the action-related checks in `enqueue_fonts`.

assets/js/modules/idea-hub/index.js
aaemnnosttv: See my comment regarding `googlesitekit-modules-idea-hub` – we shouldn't need to load it so this would also be unnecessary I think in that case.

aaemnnosttv (commented):
Thanks @eugene-manuilov – I left a few more comments/replies for you. I'd also like to get @felixarntz 's thoughts on the change to make Assets available to the Idea Hub module. See my comments below.

includes/Modules/Idea_Hub.php
eugene-manuilov: Yes, good catch. Updated.

assets/js/components/Root/index.js
```diff
@@ -62,7 +62,7 @@ export default function Root( { children, registry, viewContext = null } ) {
 }
 
 Root.propTypes = {
-	children: PropTypes.node.isRequired,
```
aaemnnosttv, line 65: This looks like it can be reverted as there there should never be a case where we render `Root` without children.

includes/Modules/Idea_Hub.php
```diff
@@ -181,6 +182,18 @@ public function register() {
...
 			 */
 			add_filter( 'googlesitekit_admin_notices', $this->get_method_proxy( 'admin_notice_idea_hub_ideas' ) );
+
+			/**
+			 * Adds a special class name to idea posts.
+			 */
+			add_filter( 'post_class', $this->get_method_proxy( 'update_post_classes' ), 10, 3 );
+
+			add_action(
+				'admin_footer',
```
aaemnnosttv, line 192: Let's scope this to the specific screen rather than all admin screens.
```suggestion
				'admin_footer-edit.php',
```

includes/Modules/Idea_Hub.php
```diff
@@ -181,6 +182,18 @@ public function register() {
...
+				function() {
+					echo '<div id="js-googlesitekit-post-list" class="googlesitekit-plugin"></div>';
```
aaemnnosttv, line 194: Let's also only output this for posts.
```suggestion
					if ( 'post' !== get_current_screen()->post_type ) {
						return;
					}
					echo '<div id="js-googlesitekit-post-list" class="googlesitekit-plugin"></div>';
```

includes/Modules/Idea_Hub.php
```diff
@@ -741,6 +758,40 @@ private function is_idea_post( $post_id ) {
...
+	 * @since n.e.x.t
+	 *
+	 * @param array $classes An array of post class names.
+	 * @param array $class An array of additional class names added to the post.
+	 * @param int   $post_id The post ID.
+	 * @return array An array of post class names.
+	 */
+	private function update_post_classes( $classes, $class, $post_id ) {
+		global $post_type;
+
```
aaemnnosttv, line 773: Let's use the screen object for this instead.
```suggestion
```

includes/Modules/Idea_Hub.php
```diff
@@ -741,6 +758,40 @@ private function is_idea_post( $post_id ) {
...
+		// Do nothing on the frontend.
+		if ( ! is_admin() ) {
+			return $classes;
+		}
+
+		$current_screen = get_current_screen();
+		if ( is_null( $current_screen ) || 'edit-post' !== $current_screen->id || 'post' !== $post_type ) {
```
aaemnnosttv, line 780:
```suggestion
		if ( is_null( $current_screen ) || 'edit-post' !== $current_screen->id || 'post' !== $current_screen->post_type ) {
```

aaemnnosttv (commented):
@eugene-manuilov A few final tweaks to make here apart from the changes around Assets pending review by @felixarntz .

eugene-manuilov:
@googlebot I fixed it.

eugene-manuilov:
@googlebot I fixed it.

eugene-manuilov:
:rage1: @googlebot I fixed it.

includes/Modules/Idea_Hub.php

includes/Modules/Idea_Hub.php

includes/Modules/Idea_Hub.php

includes/Modules/Idea_Hub.php

assets/js/components/Root/index.js
eugene-manuilov: We have the `Root` element without children here: https://github.com/google/site-kit-wp/blob/9fe10cb8c8214391288ba078e5fa4520cca52f36/assets/js/googlesitekit-idea-hub-post-list.js#L61-L66

includes/Core/Assets/Assets.php
eugene-manuilov: Updated to use $assets instance created in the Plugins class.

includes/Plugin.php
```diff
@@ -157,17 +157,18 @@ function () {
...
 
 				$permissions = new Core\Permissions\Permissions( $this->context, $authentication );
 				$permissions->register();
 
-				$modules = new Core\Modules\Modules( $this->context, $options, $user_options, $authentication );
+				$modules = new Core\Modules\Modules( $this->context, $options, $user_options, $authentication, $assets );
 				$modules->register();
 
-				$assets = new Core\Assets\Assets( $this->context );
+				// Assets must be registered after Modules instance is registered.
```
aaemnnosttv, line 171: I see why it's necessary now but we should avoid this. It doesn't need to block this PR though and I can open a follow-up to fix it.

aaemnnosttv (approved):

Issue #3760, opened 2021-07-26
Add support for additional survey question type "single_select"


felixarntz:
When the initial user feedback survey was introduced, scope of the MVP was limited to only support questions of the "rating" type, since that was the only one used for the initial survey. This should now be followed up on, with this issue focusing on the "single_select" question type.
![Screen Shot 2021-07-26 at 3 51 49 PM](https://user-images.githubusercontent.com/3531426/127069257-05074421-4a7a-4c92-902c-4ec2e832403f.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new `SurveyQuestionSingleSelect` JS component should be introduced and integrated into the user feedback survey JS infrastructure, rendering a survey question of type "single_select".
* The UI should look like the above screenshot and follow the "rating" question type where similar (e.g. font, spacing, G icon etc.).
* The available arguments to consider here from the Site Kit Service `/survey/trigger/` API response are fairly similar to those for the "rating" type, as follows:
    * `question.question.answer_choice` contains the list of choices, with `answer_ordinal` (int) and `text` (string), like for the rating question type.
        * In addition, an extra `write_in` boolean property should be supported within each answer choice. If set to `true`, the respective answer should offer a free-text input where the user can type something custom. In that case, the `text` field should be displayed right before the text input. The text input must be non-empty in order to proceed. The maximum character limit should be 100. (This additional free-text input concept is also used in #3761, so this should be coordinated to work similarly, potentially reuse common logic where possible.)
* The data object for answering the question (what needs to be passed to the `answerQuestion` function prop) should look as follows:
    * A single `answer` property, containing an object with the selected `answer_ordinal` (similar to the "rating" question type), and in addition `answer_text` (string with free-text input value, if applicable for the selected answer).
    * Example: `{ answer: { answer_ordinal: 4, answer_text: 'My custom response' } }`
* The new question type should have an example in Storybook where it can be inspected, similar to the existing "rating" type component.

## Implementation Brief
* Create `assets/js/components/surveys/SurveyQuestionSingleSelect.js` which exports the `SurveyQuestionSingleSelect` functional component with the following props (similar to `SurveyQuestionRating`):
	* `question`: Required string
	* `choices`: Required array of objects with the `answer_ordinal`, `text` and `write_in` keys.
	* `answerQuestion`: Required function to be called when a choice has been selected.
	* `dismissSurvey`: Required function to dismiss the survey.
[...]

PR #3814, opened 2021-08-09
Enhancement/3760 additional survey question type single select

danielgent:
## Summary

Addresses issue #3760

## Checklist

### DIFF
--- a/assets/js/components/surveys/CurrentSurvey.test.js
+++ b/assets/js/components/surveys/CurrentSurvey.test.js
@@ -80,139 +80,165 @@ describe( 'CurrentSurvey', () => {
 		expect( container ).toMatchSnapshot();
 	} );
 
-	it( "should render a single select question when the `question_type` is 'single_select'", async () => {
-		expect( true ).toBe( true );
+	describe( "should render a single select question when the `question_type` is 'single_select'", () => {
+		beforeEach( () => {
+			registry
+				.dispatch( CORE_USER )
+				.receiveTriggerSurvey(
+					fixtures.singleQuestionSurveySingleSelect,
+					{
+						triggerID: 'jestSurvey',
+					}
+				);
 
-		registry
-			.dispatch( CORE_USER )
-			.receiveTriggerSurvey( fixtures.singleQuestionSurveySingleSelect, {
-				triggerID: 'jestSurvey',
-			} );
-
-		fetchMock.post(
-			/^\/google-site-kit\/v1\/core\/user\/data\/survey-event/,
-			{ body: {}, status: 200 }
-		);
-
-		const { getByText, getByRole, getByLabelText, findByText } = render(
-			<CurrentSurvey />,
-			{
-				registry,
-			}
-		);
-
-		expect( fetchMock ).toHaveBeenCalledTimes( 1 );
-
-		// check correct question loads
-		expect(
-			getByText(
-				'Based on your experience so far, how satisfied are you with Site Kit?'
-			)
-		).toBeInTheDocument();
-
-		// button should be disabled until an option is selected
-		expect( getByRole( 'button', { name: 'Next' } ) ).toHaveAttribute(
-			'disabled'
-		);
-
-		fireEvent.click( getByText( 'Unhappy' ) );
-
-		expect( getByRole( 'button', { name: 'Next' } ) ).not.toHaveAttribute(
-			'disabled'
-		);
-
-		// Check text input is disabled when option is selected
-		expect(
-			getByLabelText( `Text input for option Other` )
-		).toHaveAttribute( 'disabled' );
-		fireEvent.click( getByText( 'Other' ) );
-		expect(
-			getByLabelText( `Text input for option Other` )
-		).not.toHaveAttribute( 'disabled' );
-
-		// Toggle between Other and other options
-		fireEvent.click( getByText( 'Satisfied' ) );
-		expect(
-			getByLabelText( `Text input for option Other` )
-		).toHaveAttribute( 'disabled' );
-		fireEvent.click( getByText( 'Other' ) );
-		expect(
-			getByLabelText( `Text input for option Other` )
-		).not.toHaveAttribute( 'disabled' );
-		fireEvent.click( getByText( 'Neutral' ) );
-		expect(
-			getByLabelText( `Text input for option Other` )
-		).toHaveAttribute( 'disabled' );
-		fireEvent.click( getByText( 'Other' ) );
-		expect(
-			getByLabelText( `Text input for option Other` )
-		).not.toHaveAttribute( 'disabled' );
-
-		// Submit button should now be disabled until text is entered
-		expect( getByRole( 'button', { name: 'Next' } ) ).toHaveAttribute(
-			'disabled'
-		);
-
-		fireEvent.change( getByLabelText( `Text input for option Other` ), {
-			target: { value: 'foo' },
+			fetchMock.post(
+				/^\/google-site-kit\/v1\/core\/user\/data\/survey-event/,
+				{ body: {}, status: 200 }
+			);
 		} );
 
-		expect( getByRole( 'button', { name: 'Next' } ) ).not.toHaveAttribute(
-			'disabled'
-		);
-
-		// Text input only allows up to 100 characters of input
+		// Text input should only allow up to 100 characters of input.
 		const STRING_100_CHARACTERS =
 			'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec suscipit auctor dui, id faucibus nisl';
 
 		const STRING_110_CHARACTERS = STRING_100_CHARACTERS + ' rhoncus n';
 
-		fireEvent.change( getByLabelText( `Text input for option Other` ), {
-			target: { value: STRING_110_CHARACTERS },
+		it( 'should disable submit button when no option is selected', () => {
+			const { getByText, getByRole } = render( <CurrentSurvey />, {
+				registry,
+			} );
+
+			// check correct question loads
+			expect(
+				getByText(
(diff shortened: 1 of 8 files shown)


assets/js/components/surveys/CurrentSurvey.test.js
```diff
@@ -80,6 +80,141 @@ describe( 'CurrentSurvey', () => {
 		expect( container ).toMatchSnapshot();
 	} );
 
+	it( "should render a single select question when the `question_type` is 'single_select'", async () => {
```
tofumatt, line 83: Just calling it out for posterity, but like I mentioned in the PRs for #3761 and #3762: these tests look great but let's split them up in a `describe` block 👍🏻

assets/js/components/surveys/SurveyQuestionSingleSelect.js
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Internal dependencies
+ */
+import Button from '../Button';
+import SurveyHeader from './SurveyHeader';
+import SurveyQuestionSingleSelectChoice from './SurveyQuestionSingleSelectChoice';
+
+// const MAXIMUM_CHARACTER_LIMIT = 100;
+
+/* eslint-disable camelcase */
```
tofumatt, line 38: Where this does need to happen let's do it per-line.

assets/js/components/surveys/SurveyQuestionSingleSelect.js
```diff
@@ -0,0 +1,119 @@
...
+	choices,
+	answerQuestion,
+	dismissSurvey,
+} ) => {
+	const [ value, setValue ] = useState( '' );
+	const [ writeIn, setWriteIn ] = useState( '' );
+
+	const mappedChoices = choices.map( ( choice ) => ( {
+		...choice,
+		answer_ordinal: `${ choice.answer_ordinal }`,
```
tofumatt, line 51: Why does this need to be cast as a string here? The ordinals are numbers by default.

assets/js/components/surveys/SurveyQuestionSingleSelect.js
```diff
@@ -0,0 +1,119 @@
...
+						choice={ choice }
+					/>
+				) ) }
+			</div>
+			<div className="googlesitekit-survey__footer">
+				<Button
+					onClick={ handleSubmit }
+					disabled={ isSubmitButtonDisabled }
+				>
+					Next
```
tofumatt, line 96: This should be localised.

assets/js/components/surveys/SurveyQuestionSingleSelectChoice.js
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Internal dependencies
+ */
+import Radio from '../Radio';
+import { TextField, Input } from '../../material-components';
+
+const MAXIMUM_CHARACTER_LIMIT = 100;
+
+/* eslint-disable camelcase */
```
tofumatt, line 32: As mentioned elsewhere, we tend to do this per-line to be extra deliberate. It's more annoying to write but ensures every instance is intentional 😅

assets/js/components/surveys/SurveyQuestionSingleSelectChoice.js
```diff
@@ -0,0 +1,90 @@
...
+							setWriteIn(
+								event.target.value?.slice(
+									0,
+									MAXIMUM_CHARACTER_LIMIT
+								)
+							)
+						}
+						value={ writeIn }
+						disabled={ ! isChecked }
+						aria-label={ `Text input for option ${ text }` }
```
tofumatt, line 67: This should be localised, but also can probably be a visually hidden real `<label>` tag.

assets/js/components/surveys/SurveyQuestionSingleSelectChoice.js
```diff
@@ -0,0 +1,90 @@
...
+			</Radio>
+			{ write_in && (
+				<TextField>
+					<Input
+						onChange={ ( event ) =>
+							setWriteIn(
+								event.target.value?.slice(
+									0,
+									MAXIMUM_CHARACTER_LIMIT
+								)
```
tofumatt, line 62: Because `event.target.value?` has optional chaining, this could wind up being `undefined`, right?

But the `writeIn` value is always a string, right? Should this be:
```suggestion
								event.target.value?.slice(
									0,
									MAXIMUM_CHARACTER_LIMIT
								) || ''
```

Maybe it isn't a biggie, but just something I caught.

tofumatt (changes requested):
In addition to the comments I had for #3761/#3762 that apply here too, it looks like there is one place we might want to protect against setting a string to `undefined` and add some localisations.

assets/js/components/surveys/SurveyQuestionSingleSelect.js
danielgent: The `Radio` component requires value to be a string in its propTypes

https://github.com/google/site-kit-wp/blob/develop/assets/js/components/Radio.js#L92

assets/js/components/surveys/SurveyQuestionSingleSelect.js
danielgent: Same concern as comment here - https://github.com/google/site-kit-wp/pull/3788#discussion_r685917798 - should this be a prop anyway?

assets/js/components/surveys/SurveyQuestionSingleSelectChoice.js
danielgent: I'm not sure why I put optional chaining here!  I removed it and the tests pass

assets/js/components/surveys/SurveyQuestionSingleSelectChoice.js
danielgent: Done!

assets/js/components/surveys/SurveyQuestionSingleSelectChoice.js
danielgent: I'll do the same as on the PR. Localise *and* use a visually hidden label

assets/js/components/surveys/SurveyQuestionSingleSelect.js
danielgent: Done

assets/js/components/surveys/CurrentSurvey.test.js
danielgent: Done

assets/js/components/surveys/SurveyQuestionSingleSelect.js
tofumatt: Fair enough. Part of me thinks we should be casting that in the component 😅 

Let's just leave a comment explaining why we do this at least, as it surprised me at first.

assets/js/components/surveys/SurveyQuestionSingleSelect.js
tofumatt: As mentioned in that PR: let's make it a prop and set it based on the current question ordinal 👍🏻

assets/js/components/surveys/SurveyQuestionSingleSelectChoice.js
```diff
@@ -0,0 +1,111 @@
...
+	choice,
+} ) => {
+	const { answer_ordinal, text, write_in } = choice; // eslint-disable-line camelcase
+	const isChecked = value === answer_ordinal; // eslint-disable-line camelcase
+	const uniqueID = `googlesitekit-survey__multi-select-${ answer_ordinal }-${ text }`; // eslint-disable-line camelcase
+
+	return (
+		<div className="googlesitekit-single-select__choice">
+			<Radio
+				id={ text }
```
tofumatt, line 53: Why is `text` the ID for this? Couldn't text be multiple words with punctuation? HTML ID attributes can't have whitespace so that'd cause errors.

assets/js/components/surveys/CurrentSurvey.test.js
```diff
@@ -80,6 +80,167 @@ describe( 'CurrentSurvey', () => {
...
+						},
+					},
+					headers: {
+						Accept: 'application/json, */*;q=0.1',
+						'Content-Type': 'application/json',
+					},
+				}
+			);
+
+			await findByText( 'Thanks for sharing your thoughts!' );
```
tofumatt, line 240: We should put an `expect` call around this, something like:

```suggestion
			const completionMessage = await findByText( 'Thanks for sharing your thoughts!' );
			expect( completionMessage ).not.toBeEmpty();
```

Or something. Right now this operates as an implicit test but let's make it explicit.

tofumatt (changes requested):

assets/js/components/surveys/SurveyQuestionSingleSelectChoice.js
danielgent: TIL this!  :exploding_head: 

But other attributes can. Makes sense!

assets/js/components/surveys/SurveyQuestionSingleSelect.js
danielgent: Done :+1:

assets/js/components/surveys/SurveyQuestionSingleSelectChoice.js
danielgent: Used same fix as in other PR. Replacing whitespace with dashes

tofumatt (commented):
The tests here could still use some separating and the Submit/Next text needed to be wrapped in a `__()` call, so I added it. I think now it's good to go.

tofumatt:
@googlebot I consent.

tofumatt (approved):

Issue #3758, opened 2021-07-26
Remove legacy `getModulesData` JS function and `_googlesitekitLegacyData.modules` global


felixarntz:
Related to #3647: All usages of the legacy function `getModulesData` which relies on the `_googlesitekitLegacyData.modules` global should be refactored, and then those two pieces should be entirely removed.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All JS components currently relying on `getModulesData` should be refactored to use the `getModules` or `getModule` datastore selector instead, using React hooks.
* The `getModulesData` function should be entirely removed.
* The `_googlesitekitLegacyData.modules` global should be entirely removed, including where it is set in PHP.

#### Update `getModules` code when changing from `getModulesData`

* In any place where we change from `getModulesData` to `getModules`, note that the data returned from `getModules` is slightly different. Most of the keys/values/info is the same, but here's a map of the old to new keys:
  * `sort` (old) => `order` (new)
  * `screenID` (old) => N/A (not needed)
  * `autoActivate` (old) => `forceActive` (new)
  * `required` (old) => `dependencies` (new)
  * `settings` (old) => N/A (if any data from `settings` is needed by components using `getModulesData` it will need refactoring)
  * All other keys are the same or are only available in the new API

#### `SettingsModules` component

* Remove the `useEffect` in https://github.com/google/site-kit-wp/blob/3db19cbeca41145aefc6c6c3383ee9b66e8b4121/assets/js/components/settings/SettingsModules.js#L38 and change `modulesData` to `const moduleData = useSelect( ( select ) => select( CORE_MODULES ).getModules() );` (renaming the `modulesData` to `modules` as well).

#### Some component require changing to functional components to use hooks, but this is straightforward:

* Convert `assets/js/components/legacy-notifications/dashboard-setup-alerts.js` to a functional component. Refactor https://github.com/google/site-kit-wp/blob/3db19cbeca41145aefc6c6c3383ee9b66e8b4121/assets/js/components/legacy-notifications/dashboard-setup-alerts.js#L62-L67 to access the module in a hook before the `switch` statement.
[...]

PR #3819, opened 2021-08-10
Remove getModulesData calls.

tofumatt:
## Summary

Addresses issue #3758.

## Checklist


tofumatt:
There one test suite here failing and I'm not sure why—I'd appreciate another set of eyes in case it's something obvious 😓

aaemnnosttv (changes requested):
Thanks @tofumatt – this looks good, just a few things left to address.

assets/js/modules/optimize/index.legacy.js
```diff
@@ -1,48 +0,0 @@
...
- */
-import { getModulesData } from '../../util';
-
-const modulesData = getModulesData();
-if ( modulesData.optimize.active ) {
-	/**
-	 * Add data to the congrats setup Win Notification for display.
-	 */
-	addFilter(
-		'googlesitekit.SetupWinNotification-optimize',
```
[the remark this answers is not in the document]
aaemnnosttv: Hmm, apparently all of the legacy module indexes were no longer loaded as of https://github.com/google/site-kit-wp/commit/8d9b8102864c843accabc937190aa47cf21e90c6 as part of #1997 (which I think was somewhat of a mistake). Let's move the `addFilter` to the regular module `index` for now (as we have in ga4) to be preserved then the file can be deleted (as the only `index.legacy` left) since there's nothing else there to keep and it isn't being loaded anyways 😄 

We don't need to guard the addition of the filter as that is already handled in `DashboardSetupAlerts` and the filter itself is Optimize-specific like you said 👍

assets/js/components/legacy-notifications/dashboard-modules-alerts.js
```diff
@@ -20,91 +20,86 @@
...
 import Notification from './notification';
 import { modulesNotificationsToRequest, getModulesNotifications } from './util';
-import NotificationAlertSVG from '../../../svg/notification-alert.svg';
+const { useSelect } = Data;
 
-class DashboardModulesAlerts extends Component {
-	constructor( props ) {
-		super( props );
+function DashboardModulesAlerts() {
+	const [ alerts, setAlerts ] = useState( [] );
```
aaemnnosttv, line 42: `alerts` is defined as an array here, but `response.results` is an object below.

assets/js/components/legacy-notifications/dashboard-modules-alerts.js
```diff
@@ -20,91 +20,86 @@
...
-	render() {
-		const { data } = this.state;
+		if ( modulesWithNotifications ) {
+			const response = await getModulesNotifications(
+				modulesWithNotifications
+			);
 
-		if ( 0 === Object.keys( data ).length ) {
-			return null;
+			setAlerts( response.results );
```
aaemnnosttv, line 56: `response.results` is a map of `[slug]: notification[]`
https://github.com/google/site-kit-wp/blob/162ccaf28515e6227aeca713bbf1c6d5a84c8963/assets/js/components/legacy-notifications/util.js#L137

assets/js/components/legacy-notifications/dashboard-modules-alerts.js
```diff
@@ -20,91 +20,86 @@
...
-		if ( modules ) {
-			getModulesNotifications( modules ).then( ( response ) => {
-				this.setState( {
-					data: response.results,
-				} );
-			} );
-		}
-	}
+	useMount( async () => {
+		const modulesWithNotifications = modulesNotificationsToRequest();
```
aaemnnosttv, line 49: There's been a regression (although not from this PR) where this function will return an empty array now. This is from the filter that was removed in https://github.com/google/site-kit-wp/commit/8d9b8102864c843accabc937190aa47cf21e90c6#diff-47b6b8d7a6c2934687704e4fba19f328306558821fe7836f023021b99cd26d69L61-L67 which should have been kept. AdSense is the only module that uses/implements notifications right now.

assets/js/modules/optimize/index.js
```diff
@@ -32,6 +33,23 @@ import { MODULES_OPTIMIZE } from './datastore/constants';
...
+		( winData ) => {
+			winData.description = __(
+				'To set up experiments and see the results, go to ',
+				'google-site-kit'
+			);
+			winData.learnMore.label = 'Optimize';
+			winData.learnMore.url = 'https://optimize.withgoogle.com/';
+			return winData;
+		}
+	);
```
aaemnnosttv, line 51: This looks a bit out of place here but I like the idea of keeping the module index side-effect free.

assets/js/components/legacy-notifications/dashboard-modules-alerts.js
```diff
@@ -20,91 +20,86 @@
...
 
-	render() {
-		const { data } = this.state;
+		if ( modulesWithNotifications ) {
+			const response = await getModulesNotifications(
+				modulesWithNotifications
+			);
```
aaemnnosttv, line 54: I realize this is how it was before, but `getModulesNotifications` actually takes no arguments and instead calls `modulesNotificationsToRequest` internally 😄 
Maybe best to keep that way since the return value from that function is an odd structure.

aaemnnosttv (commented):
Thanks @tofumatt – this one is almost ready to go, just a few things to straighten out that I found. See comments below.

assets/js/modules/optimize/index.js
tofumatt: Yeah, it felt weird but like the least-worst-option. I'll do this for all modules, but I'll annotate it as legacy 👍🏻

aaemnnosttv:
@googlebot I consent.

assets/js/components/legacy-notifications/util.js
```diff
@@ -106,41 +113,36 @@ const removeDismissed = async ( notifications ) => {
  * Gets notifications from session storage, fallback to notifications API request.
  *
  * @since 1.0.0
+ * @since n.e.x.t Memoized to prevent duplicate simultaneous fetch requests from different callers.
  *
- * @return {number} Number of module notifications.
+ * @return {Promise} Object with `results` (map [slug]: notificationObject[]) and `total` (int).
  */
-export async function getModulesNotifications() {
+export const getModulesNotifications = memize( async () => {
```
aaemnnosttv, line 120: Memoizing the function works as expected here because it's essentially memoizing a promise immediately rather than the final resolved value of the function.

assets/js/components/legacy-notifications/util.js
```diff
@@ -106,41 +113,36 @@ const removeDismissed = async ( notifications ) => {
...
+	// Legacy hack: we need to use the global datastore instance here.
+	await Data.__experimentalResolveSelect( CORE_MODULES ).getModules();
+	const { isModuleActive } = Data.select( CORE_MODULES );
 
-	modules.map( async ( module ) => {
-		const promise = new Promise( async ( resolve ) => {
-			const { identifier } = module;
+	const activeModuleSlugs = modulesNotificationsToRequest().filter(
+		( slug ) => isModuleActive( slug )
+	);
```
aaemnnosttv, line 130: This was the least worst thing I could think of to fix this. In the future, we might want to expose `isModuleActive` via our `Modules` API (ie. the object provided to `registerModule`), or add support for another function arg to `registerModule` that only runs when the module is initially active (similar to `Module::register` on the PHP side).

assets/js/modules/adsense/index.js
```diff
@@ -45,10 +46,22 @@ import {
...
+	// This is called inside `registerModule` to prevent this file from having
+	// side-effects. This is used to show notifications in the AdSense dashboard.
+	/**
+	 * Add components to the Notification requests.
+	 */
+	addFilter(
+		'googlesitekit.ModulesNotificationsRequest',
+		'googlesitekit.adsenseNotifications',
+		( notificationModules ) => {
+			return notificationModules.concat( 'adsense' );
```
aaemnnosttv, line 61: As the only instance of this filter, it makes sense to simplify this to use module slugs instead of the odd object structure it had before.

aaemnnosttv (approved):
@tofumatt this looks good to me now 👍 I pushed a few commits to fix the few remaining things we discussed yesterday (duplicate fetch requests for module notifications + fetching notifications even when module is inactive) which were part regression.

I left a few comments on my changes for you to review, but otherwise I think this is good to go 👍

tofumatt (commented):
Nice, looks good, I like the improvements 👍🏻

Issue #3757, opened 2021-07-26
Idea Hub draft post ideas requested on every page load


felixarntz:
## Bug Description

This is a follow-up bug for #3733.

See https://github.com/google/site-kit-wp/issues/3733#issuecomment-886938896: Currently the request to get draft post ideas in the dashboard happens on every page load, which is undesirable for performance.

In addition, the `getNewIdeas` and `getSavedIdeas` selectors also consider draft posts for ideas (in order to exclude those from the responses), but the original fix doesn't address that. Essentially, whatever mechanism to invalidate the cache based on Idea Hub post changes `getDraftPostIdeas` gets, we also need to do for `getNewIdeas` and `getSavedIdeas`.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Whenever an Idea Hub-based post has its post status changed or is deleted, that time should be stored in a transient for an hour, essentially as a "when was any Idea Hub post last changed" value.
* On the client-side, Idea Hub caches should be invalidated for any already cached results that are older than that stored timestamp (if there is any). This should apply to any Idea Hub endpoint caches to get ideas.
* The server-side timestamp should be in a transient with one hour expiration, since it wouldn't be meaningful after more than an hour anyway, as the client-side cache expiration is also one hour.

### Generate last changed time for Idea Hub posts

In `includes/Modules/Idea_Hub.php`:

* Add a new `const IDEA_HUB_LAST_IDEA_POST_UPDATED_AT = googlesitekit_idea_hub_last_idea_post_updated_at`
* In `$this->is_connected() )` function, add a hook for `transition_post_status` action (https://codex.wordpress.org/Post_Status_Transitions )
  * Return if the post type is not an idea hub post (see `is_idea_post()`)
  * Update the transient `IDEA_HUB_LAST_IDEA_POST_UPDATED_AT` with the output of `time()` using the `set` method of the `Transient` class instance with `$expiration` param set to `0`.
* Add a new instance of the `Script_Data` class to the Idea Hub assets returned from the `setup_assets` method. It should use the `googlesitekit-idea-hub-data` handler and the following options passed as the second argument to its constructor:
  * The `global` property should be `_googlesitekitIdeaHub`
  * The `data_callback` property should be a function that returns an array with the `lastIdeaPostUpdatedAt` property.

### <del>Create an endpoint for last updated param for Idea Hub posts</del>

[...]

PR #3826, opened 2021-08-11
Idea Hub draft post ideas requested on every page load (3757).

johnPhillips:
## Summary

Addresses issue #3757

## Checklist


includes/Modules/Idea_Hub.php
```diff
@@ -770,6 +792,26 @@ private function is_idea_post( $post_id ) {
...
+	 * @param string  $old_status Previous post status.
+	 * @param WP_Post $post The post in question.
+	 */
+	private function on_idea_hub_post_status_transition( $new_status, $old_status, $post ) {
+		if ( ! $this->is_idea_post( $post->ID ) ) {
+			return;
+		}
+
+		if ( $new_status !== $old_status ) {
+			$transients = new Transients( $this->context );
```
eugene-manuilov, line 810: I see we have many places in the Idea_Hub class where we instantiate new instances of the Transients class. I think it's a good time to refactor it a little bit and add a new property for transients to this class, and then use one instance in all places.

assets/js/modules/idea-hub/datastore/module-data.js
```diff
@@ -0,0 +1,118 @@
...
+			return;
+		}
+
+		if ( ! global._googlesitekitIdeaHub ) {
+			global.console.error( ' Could not load modules/idea-hub data.' );
+			return;
+		}
+
+		yield actions.receiveIdeaHubData( global._googlesitekitIdeaHub );
+	},
```
eugene-manuilov, line 93: I know that you have implemented what has been requested in IB and it works, but I think this implementation is not flexible enough to support new properties in the future. :thinking: If we need to add one more property to the `_googlesitekitIdeaHub` object in the future, we will need to add one more resolver that will do exactly the same thing as this resolver does. This is not cool.

Let's create a new base selector that will load the Idea Hub data into the store state in its resolver and returns the whole data object. Then update the `getLastIdeaPostUpdatedAt` selector to select the whole data object and return the appropriate field from it. See how the `getSiteInfo` and other selectors work as an example: https://github.com/google/site-kit-wp/blob/develop/assets/js/googlesitekit/datastore/site/info.js.

eugene-manuilov (changes requested):
Thanks, @johnPhillips. In general, this looks good. I have added two suggestions on how we can make this code even better. Could you please take a look at it?

assets/js/modules/idea-hub/datastore/module-data.js
```diff
@@ -0,0 +1,139 @@
...
+	 * For an example see `on_idea_hub_post_status_transition` in
+	 * `includes/Modules/Idea_Hub.php`.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} ideaHubData Module data, via a PHP-to-JS global.
+	 * @return {Object} Redux-style action.
+	 */
+	receiveIdeaHubData( ideaHubData ) {
+		invariant( ideaHubData, 'Idea Hub data is required.' );
```
eugene-manuilov, line 61: Let's verify that the `ideaHubData` is a plain object using the `isPlainObject` function from the `lodash` library.

assets/js/modules/idea-hub/datastore/module-data.js
```diff
@@ -0,0 +1,139 @@
...
+		if ( select( MODULES_IDEA_HUB ).getIdeaHubData() !== undefined ) {
+			return;
+		}
+
+		if ( ! global._googlesitekitIdeaHub ) {
+			global.console.error( ' Could not load modules/idea-hub data.' );
+			return;
+		}
+
+		const { lastIdeaPostUpdatedAt } = global._googlesitekitIdeaHub;
```
eugene-manuilov, line 101: Let's just default to an empty object if we can't find the `_googlesitekitIdeaHub` in the global scope:
```suggestion
		const { lastIdeaPostUpdatedAt } = global._googlesitekitIdeaHub || {};
```

includes/Modules/Idea_Hub.php
```diff
@@ -107,6 +114,13 @@ final class Idea_Hub extends Module
 	 */
 	private $post_topic_setting;
 
+	/**
+	 * Transients instance.
+	 *
+	 * @var Transients
```
eugene-manuilov, line 120:
```suggestion
	 * @since n.e.x.t
	 * @var Transients
```

includes/Modules/Idea_Hub.php
```diff
@@ -225,7 +240,7 @@ private function admin_notice_idea_hub_ideas( $notices ) {
 			return $notices;
 		}
 
-		$transients      = new Transients( $this->context );
+		$transients      = $this->transients;
```
eugene-manuilov, line 243: No need to assign it to a variable. Just use `$this->transients->...` directly instead of `$transients->...` methods.

includes/Modules/Idea_Hub.php
```diff
@@ -339,6 +354,8 @@ public function is_connected() {
 			}
 		}
 
+		add_action( 'transition_post_status', $this->get_method_proxy( 'on_idea_hub_post_status_transition' ), 10, 3 );
+
```
eugene-manuilov, line 358: This needs to be defined in the `register` method withtin the `if ( $this->connected() )` condition.

includes/Modules/Idea_Hub.php
```diff
@@ -613,7 +630,8 @@ protected function setup_settings() {
 	 * @return Asset[] List of Asset objects.
 	 */
 	protected function setup_assets() {
-		$base_url = $this->context->url( 'dist/assets/' );
+		$transients = $this->transients;
```
eugene-manuilov, line 633: Same here, no need to assign it to a variable, just use `$this->transients->...` directly.

eugene-manuilov (changes requested):
Thanks, @johnPhillips. This looks very close to what we need. Added a few more comments.

includes/Modules/Idea_Hub.php
johnPhillips: @eugene-manuilov I assigned it to a variable because the anonymous function has a `use` which I believe can't take `$this->...`? 

```php
'active_callback' => function() use ( $transients, $dismissed_items ) {
	$saved_ideas = $transients->get( self::TRANSIENT_SAVED_IDEAS );
	if ( false === $saved_ideas ) {
		$saved_ideas = $this->get_data( 'saved-ideas' );
		$transients->set( self::TRANSIENT_SAVED_IDEAS, $saved_ideas, DAY_IN_SECONDS );
// ...
```
If I understand it correctly, in a PHP lamdba `$this` is bound to the scope where it is declared rather than when it is called (I guess like lexical `this` in JS)? 
Therefore I wasn't sure when `use` is necessary and when it isn't... in the end I think I just looked at existing examples and copied those.

This also happens further on in the file in `setup_assets` as you mention below. 

So can I remove `$transients` from the `use` in both cases and just reference `$this->transients` inside the lambda?

includes/Modules/Idea_Hub.php
johnPhillips: Please see my question above ☝️

eugene-manuilov (approved):
LGTM! Thanks, @johnPhillips!

PR #3864, opened 2021-08-12
Enhancement/3702 analytics links

eugene-manuilov:
## Summary

Addresses issue #3702

## Checklist

### DIFF
--- a/assets/js/modules/analytics/components/settings/SettingsView.js
+++ b/assets/js/modules/analytics/components/settings/SettingsView.js
@@ -151,7 +151,7 @@ export default function SettingsView() {
 				</div>
 				<div className="googlesitekit-settings-module__meta-item">
 					<div className="googlesitekit-settings-module__meta-item-type" />
-					<p className="googlesitekit-settings-module__meta-item-data">
+					<p className="googlesitekit-settings-module__meta-item-data googlesitekit-settings-module__meta-item-data--tiny">
 						<Link href={ editViewSettingsURL } external inherit>
 							{ createInterpolateElement(
 								__(
--- a/assets/js/modules/analytics/components/settings/SettingsView.js
+++ b/assets/js/modules/analytics/components/settings/SettingsView.js
@@ -146,7 +146,12 @@ export default function SettingsView() {
 						{ __( 'View', 'google-site-kit' ) }
 					</h5>
 					<p className="googlesitekit-settings-module__meta-item-data">
-						<DisplaySetting value={ profileID } />{ ' ' }
+						<DisplaySetting value={ profileID } />
+					</p>
+				</div>
+				<div className="googlesitekit-settings-module__meta-item">
+					<div className="googlesitekit-settings-module__meta-item-type" />
+					<p className="googlesitekit-settings-module__meta-item-data">
 						<Link href={ editViewSettingsURL } external inherit>
 							{ createInterpolateElement(
 								__(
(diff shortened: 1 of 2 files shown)


assets/js/modules/analytics/components/settings/SettingsView.js
```diff
@@ -189,7 +194,12 @@ export default function SettingsView() {
 								) }
 							</h5>
 							<p className="googlesitekit-settings-module__meta-item-data">
-								<DisplaySetting value={ ga4MeasurementID } />{ ' ' }
+								<DisplaySetting value={ ga4MeasurementID } />
+							</p>
+						</div>
+						<div className="googlesitekit-settings-module__meta-item">
+							<div className="googlesitekit-settings-module__meta-item-type" />
```
felixarntz, line 201: This seems hacky, along with the `:empty` CSS you have below. Can we instead omit this empty container and provide styles to fix `.googlesitekit-settings-module__meta-item-data:first-child`?

felixarntz (commented):
@eugene-manuilov Only one concern here.

assets/js/modules/analytics/components/settings/SettingsView.js
eugene-manuilov: @felixarntz using the "CSS hack" for the  `:empty` selector helps us to inherit the height of meta item headers without specifying the exact `height` in CSS. In the future, if we decide to change the font size of headers, these links will automatically re-adjust their heigh without changing anything.

If we remove empty headers for these links and use the `.googlesitekit-settings-module__meta-item-data:first-child` selector, we will have to hardcode the top margin to make sure links are correctly positioned on the layout. IMO this is a less maintainable approach than using what is in this PR right now.

felixarntz: @eugene-manuilov There should be a cleaner way to fix this without an empty container, e.g. by aligning the content towards the bottom. We could use `align-items: flex-end` on the outer `.googlesitekit-settings-module__meta-items` container.

eugene-manuilov: Ok, updated.

assets/sass/components/settings/_googlesitekit-settings-module.scss
```diff
@@ -154,6 +154,11 @@
 			margin: 0 ($grid-gap-desktop * 2) $grid-gap-desktop 0;
 		}
 
+		&--data-only {
+			align-items: flex-end;
+			display: flex;
+		}
```
felixarntz, line 160: Nice! I like having that only here while using `stretch` for the container.

assets/sass/components/settings/_googlesitekit-settings-module.scss
```diff
@@ -184,6 +189,11 @@
 		&--wrap {
 			word-break: break-all;
 		}
+
+		&--tiny a {
+			font-size: 0.8125rem;
+			font-weight: 400;
+		}
```
felixarntz, line 196: This works, but I think conceptually it would make more sense to apply this styling to the entire copy (i.e. even if it wasn't only a link). Otherwise it would look weird having a text there with a link in it, as the link would randomly appear smaller than the rest of the text around it.

felixarntz (approved):
Great work!

I made the one "tiny" tweak to apply the `--tiny` modifier styles to the entire element it is used on.

PR #3871, opened 2021-08-15
Feature/di containers

eugene-manuilov:
## Summary

Addresses issue #

## Checklist


includes/Core/DI/DI_Container.php
```diff
@@ -0,0 +1,196 @@
...
+	 */
+	public function set( $id, $entry ) {
+		if ( ! empty( $this->definitions[ $id ]['is_protected'] ) ) {
+			return false;
+		}
+
+		$this->definitions[ $id ] = array(
+			'is_factory'   => false,
+			'is_protected' => false,
+			'is_service'   => is_callable( $entry ),
```
felixarntz, line 96: This may not be entirely accurate, I think it's a bit too clever. For example, `$entry` will also be a callable for a factory, in which case it isn't a "service".

I'd prefer if the `set` method was more flexible (e.g. `$entry` would be an array with all the necessary data), and we add wrapper methods to simplify the more specific definitions, e.g. with a `set_service( $id, $create_func )` and `set_factory( $id, $create_func )` which then call the lower-level `set` accordingly. We could even introduce something like a `set_value( $id, $value )` for the most generic type which then would call `set( $id, array( 'entry' => $value )` - essentially `set` would always be the low-level method, and for each of the three types of entries there's a higher-level method to set them.

includes/Core/DI/DI_Container.php
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	/**
+	 * Sets the entry to be protected.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $id Identifier of the entry.
+	 */
+	public function set_protected( $id ) {
```
felixarntz, line 110: Somewhat related to my last comment, maybe this should be `set_is_protected` to clarify it only sets the attribute within an existing entry?

includes/Core/DI/DI_Container.php
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	/**
+	 * Sets the entry to be factory.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $id Identifier of the entry.
+	 */
+	public function set_factory( $id ) {
```
felixarntz, line 123: This may not be needed if we go with the approach of my first comment. IMO there are three types of entries:
* An entry that just returns whatever value is set. This is the default. Can be set using `set_value`.
* A service entry, which returns the return value from a callback, invoked just once. Can be set using `set_service`.
* A factory entry, which returns the return value from a callback, on every access call. Can be set using `set_factory`.

Any of those three types can be protected, so keeping a `set_protected` (or `set_is_protected`, see above) around makes sense.

includes/Core/DI/DI_Container.php
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	/**
+	 * Sets services.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $definitions Service definitions.
+	 */
+	public function set_services( array $definitions ) {
```
felixarntz, line 136: Rather than having this method, how about we go with just the singular `set_service`? Related to my above comment, that could bundle the logic to set a service. If you pass a callback, it'll use that directly, or if you pass a string, it'll interpret it as class name and create the callback by itself, like the code in here does.

Alternatively, if you feel strongly about keeping `set_services`, let's just make this method a wrapper that goes through `$definitions` and calls `set_service` for each.

includes/Core/DI/DI_Container.php
```diff
@@ -0,0 +1,196 @@
...
+		foreach ( $definitions as $service_name => $service_class ) {
+			$service_class_key = sprintf(
+				'%s_CLASS',
+				strtoupper( $service_name )
+			);
+
+			$this->set( $service_class_key, $service_class );
```
felixarntz, line 143: Why is this needed?

includes/Core/DI/DI_Container.php
```diff
@@ -0,0 +1,196 @@
...
+	 * @return mixed Entry.
+	 */
+	public function get( $id ) {
+		if ( ! isset( $this->definitions[ $id ] ) ) {
+			return null;
+		}
+
+		$definition = $this->definitions[ $id ];
+		if ( $definition['is_service'] ) {
+			if ( empty( $definition['instance'] ) || $definition['is_factory'] ) {
```
felixarntz, line 168: Related to some of my other comments, I wouldn't argue a factory is always a service. Could we handle these separately? Essentially:
* If this is a service, instantiate it _only_ if it hasn't been instantiated yet.
* If this is a factory, always instantiate it and return a new instance.

includes/Core/DI/DI_Services_Aware_Trait.php
```diff
@@ -0,0 +1,36 @@
...
+
+	/**
+	 * Gets DI service instance.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $name Service name.
+	 * @return mixed DI service on success, otherwise NULL.
+	 */
+	public function __get( $name ) {
```
felixarntz, line 28: Not a fan of using magic getter here - it's unspecific and also a pain to integrate if the class itself wants to include its own. Let's make this a dedicated method, like `get_service`. I suppose you did this to make migration of the code easier, but I don't think we should go that route just because.

An acceptable path for me would be to keep this a magic getter just to simplify the initial PR and transition, and then have a follow-up issue to change the magic getter to a dedicated method, which would need to happen together with each class using e.g. `$this->get_service( 'assets' )` instead of `$this->assets`.

includes/Core/DI/DI_Services_Aware_Trait.php
```diff
@@ -0,0 +1,36 @@
...
+		if ( $this instanceof DI_Aware_Interface ) {
+			return $this->get_di()->get( $name );
```
felixarntz, line 30: Would be great to somehow ensure here that this only returns the value if it is actually a service. The trait includes "services-aware" as a term, so it probably shouldn't allow to just access _anything_ in the DI container from the outside?

includes/Core/Modules/Module.php
```diff
@@ -171,11 +123,17 @@ final public function __isset( $key ) {
 	 * @return mixed Value for $key, or null if not available.
 	 */
 	final public function __get( $key ) {
-		if ( ! isset( $this->info[ $key ] ) ) {
-			return null;
+		$di = $this->get_di();
+		if ( $di->has( $key ) ) {
+			return $di->get( $key );
+		}
```
felixarntz, line 129: This is exactly what I was referring to above - I'd prefer if we wouldn't use magic properties for the services (or anything in the DI container fwiw).

includes/Core/REST_API/REST_Routes.php
```diff
@@ -31,57 +35,17 @@
  * @since 1.0.0
  * @access private
  * @ignore
+ *
+ * @property-read Authentication $authentication Authentication instance.
+ * @property-read Context        $context        Plugin context.
+ * @property-read Modules        $modules        Modules instance.
+ * @property-read Transients     $transients     Transients API instance.
```
felixarntz, line 42: This also confirms my comment above: Maintaining lists of those properties will be painful in the long run. The DI container itself will contain a lot more than those 4, and it will change. I'm fairly sure this will be a maintenance burden, and it won't help since we'll forget them and they'll be out of date quickly.

includes/di.php
```diff
@@ -0,0 +1,67 @@
...
+		)
+	);
+
+	$di['user_options'] = function( $di ) {
+		return new User_Options( $di['context'] );
+	};
+
+	$di['authentication'] = function( $di ) {
+		return new Authentication( $di['context'], $di['options'], $di['user_options'], $di['transients'] );
+	};
```
felixarntz, line 55: See my first comment, this looks quite intransparent, e.g. why is one passing a value, another one passing a callback, another one using `set_services` with class names?

If this was all using methods like `set_value` and `set_service`, it would be easier to understand. Also this shows our `set_service` should support passing the entire callback, so that those dependencies can be filled in. Passing the class name should be supported too, but that is only enough for the "simple" services without constructor dependencies.

includes/di.php
```diff
@@ -0,0 +1,67 @@
...
+
+	// Define protected services.
+	$di->set_protected( 'PLUGIN_CLASS' );
+	$di->set_protected( 'plugin' );
```
felixarntz, line 59: Now that I see this, I'm a bit wondering about the purpose of protected services. For production code, _all_ our services should probably be protected. The only purpose to override them is for tests. Maybe we should set all our services as protected, but only in a production environment?

felixarntz (commented):
@eugene-manuilov I left some feedback for the DI container implementation. It would probably make sense to focus purely on that implementation first and align on the approach before updating the other classes to work with it, just to save some potentially unnecessary work.

includes/Core/REST_API/REST_Routes.php
eugene-manuilov: @felixarntz this list of `@property-read` tags is not intended to list all values and services defined in the DI container. We use these annotations only for values and services that are needed in a particular class and only for the purpose of having autocomplete working in our IDEs.

eugene-manuilov: In other words, we don't need to update this list if we add a new service or a value.

includes/di.php
eugene-manuilov: > this looks quite intransparent, e.g. why is one passing a value, another one passing a callback, another one using set_services with class names?

@felixarntz this is just a demonstration of how some services that require a more complex setup can be defined in the DI container. Eventually, after refactoring, the `Authentication` class will also be added to the `set_services` definitions.

> Also this shows our set_service should support passing the entire callback, so that those dependencies can be filled in. Passing the class name should be supported too, but that is only enough for the "simple" services without constructor dependencies.

The idea for the `set_service` method was exactly to allow us to define "simple" services that can be easily instantiated without passing anything into the constructor. Perhaps we should rename that method to `set_clean_services`?

includes/Core/DI/DI_Container.php
eugene-manuilov: To let us easily override the class name that we want to use for a specific service. For example, let's assume we have these services:

```php
$di->set_services(
    array(
        'serviceA' => '\\Path\\To\\ClassA',
        'serviceB' => '\\Path\\To\\ClassB',
        ...
    )
)
```

Now, if we want to override `serviceA` and `serviceB`, and we don't use the "class name" approach, we need to do the following:

```php
$di['serviceA'] = function( $di ) {
    return new \Path\To\ClassA\SubclassA();
};

$di['serviceB'] = function( $di ) {
    return new \Path\To\ClassB\SubclassB();
};
```

To reduce the boilerplate, we can use the "class name" approach:

```php
$di['SERVICEA_CLASS'] = '\\Path\\To\\ClassA\\SubclassA';
$di['SERVICEB_CLASS'] = '\\Path\\To\\ClassB\\SubclassB';
```

includes/Core/DI/DI_Container.php
eugene-manuilov: Yes, i like this idea. Updated.

includes/Core/DI/DI_Container.php
eugene-manuilov: Yep, updated.

includes/Core/DI/DI_Container.php
eugene-manuilov: Yep, updated.

includes/Core/DI/DI_Container.php
eugene-manuilov: Updated the `set_services` method to use the `set_service` method.

includes/Core/DI/DI_Container.php
eugene-manuilov: Yep, updated.

includes/Core/DI/DI_Container.php
felixarntz: I think this is a bit overengineered. Overriding the callback function IMO isn't too much to ask, having to override the extra `%s_CLASS` key is actually less intuitive and more "magical" I think. Plus it wouldn't work anyway if a service needs constructor parameters.

IMO overriding the callback is more intuitive and more flexible, so I don't think we should store the class name in this extra key.

includes/di.php
felixarntz: > Eventually, after refactoring, the `Authentication` class will also be added to the `set_services` definitions.

Thanks, makes sense.

> The idea for the `set_service` method was exactly to allow us to define "simple" services that can be easily instantiated without passing anything into the constructor. Perhaps we should rename that method to `set_clean_services`?

I don't think this differentiation makes sense. Why does it matter whether a service requires constructor parameters or not? Either way it's a service, so it should be set through `set_service` (or the `set_services` wrapper). Why not support `string|callable` (class name or create function) for that and handle it accordingly?

includes/Core/REST_API/REST_Routes.php
felixarntz: That's fair. Still, I'd prefer to do away with the magic getter and use a more explicit approach, like a method.

includes/Core/DI/DI_Container.php
```diff
@@ -0,0 +1,258 @@
...
+			);
+
+			$this->set_value( $service_class_key, $service_class );
+			$this->set_service(
+				$service_name,
+				function( $di ) use ( $service_class_key ) {
+					return new $di[ $service_class_key ]();
+				}
+			);
+		}
```
felixarntz, line 193: It's weird to me that `set_services` works differently than `set_service`. Related to my other point, why not instead change `set_service` to accept either a class name or create function? IMO the logic should be bundled there and not only work if you call `set_services`. What if you want to set only one service where you wanna pass a class name? It's rather unintuitive that you have to call `set_services` in that case and can't use `set_service` in the same way.

includes/Core/DI/DI_Container.php
```diff
@@ -0,0 +1,258 @@
...
+	 *
+	 * @param string   $id Identifier of the entry.
+	 * @param callable $factory_func Factory function.
+	 * @return bool TRUE if the factory is added, otherwise FALSE.
+	 */
+	public function set_factory( $id, $factory_func ) {
+		return $this->set(
+			$id,
+			array(
+				'is_service' => true,
```
felixarntz, line 152: This shouldn't be set here, only `is_factory`.

felixarntz (commented):
@eugene-manuilov Reviewed the iterations, they look great to me, thanks. I left two follow-up comments, but also please review the outstanding comments from my other review. I resolved all the ones that don't need further attention.

includes/Core/DI/DI_Container.php
eugene-manuilov: Ok, updated it.

includes/Core/DI/DI_Services_Aware_Trait.php
eugene-manuilov: Yes, let's keep it, for now, to make migration easier. I have moved the functionality into the new `get_di_entry( $name )` method and marked the `__get` method as deprecated.

includes/Core/DI/DI_Services_Aware_Trait.php
eugene-manuilov: Ok, "services-aware" is probably incorrect naming for this trait. I have updated it to be `DI_Entry_Aware_Trait` to better reflect its purpose.

includes/di.php
eugene-manuilov: Ok, updated the `set_service` method to accept class names as well :+1:

includes/di.php
eugene-manuilov: Updated the `DI_Container` class to make it possible to "seal" the container and not allow modifications in the production environment.

includes/Core/DI/DI_Container.php
eugene-manuilov: Yep, updated it to allow passing class names into the `set_service` method.

includes/Core/DI/DI_Container.php
eugene-manuilov: Updated.

includes/Core/DI/DI_Entry_Aware_Trait.php
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+namespace Google\Site_Kit\Core\DI;
+
+/**
+ * Trait for DI services aware classes.
+ *
+ * @since n.e.x.t
+ */
+trait DI_Entry_Aware_Trait {
```
felixarntz, line 18: I don't think this separate trait is necessary, the `get_di_entry` method should rather be baked into `DI_Aware_Interface` and its trait. This makes the code simpler, it can just implement the `DI_Aware_Interface` by using `DI_Aware_Trait`, without needing to do something else just to easily access entries.

I've reflected this simplification in the ACs for #4192.

Issue #3817, opened 2021-08-10
Implement GA events for Idea Hub module


felixarntz:
Related to #3524 and #3625, which via feature tours introduce the first couple Idea Hub-focused GA events, this issue is about introducing various other GA events for capturing user interactions with the Idea Hub module and its features.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The event category for the main feature tour introduced in #3524 should be updated to `idea_hub_widget`.
* The following other Idea Hub widget events should be tracked using event category `idea_hub_widget`:
    * prompt_widget_view: User sees CTA widget (i.e. on scroll into view, like with PSI widget)
    * prompt_widget_setup: User clicks CTA button in CTA widget
    * prompt_widget_dismiss: User dismisses CTA widget
    * widget_view: User sees regular (non-CTA) Idea Hub widget, in any state (on scroll into view, like with PSI widget); this should always be followed by one of the following two events
        * widget_gathering_data_view: User sees version of the Idea Hub widget with zero data entirely (i.e. no new ideas, no ideas saved, no drafts; also on scroll into view, like with PSI widget)
        * default_tab_view: User sees Idea Hub widget with a certain initial tab open (this is the counterpart to the above "widget_gathering_data_view"); the event label should be accordingly either "new-ideas", "saved-ideas", or "draft-ideas"
    * tab_select: User clicks to select another tab; the event label should be accordingly either "new-ideas", "saved-ideas", or "draft-ideas"
    * start_draft: User clicks button to start an idea draft
    * view_draft: User clicks the link/button to view an existing idea draft
    * save_idea: User clicks the button to save an idea
    * unsave_idea: User clicks the button to unsave an idea
    * dismiss_idea: User clicks the button to dismiss an idea
    * click_outgoing_link: User clicks learn more link in Idea Hub widget; the event label should be (always) "idea_hub_learn_more"
* The following events should be tracked using event category `idea_hub_dashboard`:
    * prompt_notification_view: User sees Idea Hub CTA notification in the Site Kit dashboard
    * prompt_notification_setup: User clicks to activate Idea Hub module via the prompt notification
    * prompt_notification_dismiss: User dismisses the prompt notification
* The following events (other than the feature tour ones from #3625) should be tracked using event category `idea_hub_posts`:
[...]

Issue #3278, opened 2021-05-04
IntersectionObserver is not defined dashboard error when accessing Analytics tab


jamesozzie:
## Bug Description

There have been reports of the below dashboard error, both occurring from Site Kit 1.31.0 users. This seems to occur only when clicking on the Analytics tab for one user while it only occurs on the primary dashboard tab for another. 

One user reported that disconnecting Analytics results in no dashboard notice. 

Awaiting further insights.

```
Can't find variable: IntersectionObserver

    in header
    in DashboardPageSpeed
    in div
    in Layout
    in div
    in LegacyDashboardSpeed
    in FilteredComponent
    in WithFilters(LegacyDashboardModule)
    in div
    in Row
    in div
    in Grid
    in div
    in DashboardApp
    in GoogleSitekitDashboard
    in RestoreSnapshots
    in ErrorHandler
    in Root
```

https://wordpress.org/support/topic/cant-find-variable-intersectionobserver/ | **Open** | Awaiting SH info
https://wordpress.org/support/topic/site-kit-encountered-an-error-32 | **Resolved** (by user, unknown cause) | [SH info filled in although not complete](https://docs.google.com/forms/d/1lhjns2K7c0Ny7ME4gxTcsnYtNUtcMr8BvFpPQi23Y1E/edit#response=ACYDBNiRUkImypFv2sEuzamlFFnx9Kr1Xd6Sb1bWIn_EhjACze_Rh_z0lsXrr5b1rEMzoxg)
https://wordpress.org/support/topic/intersectionobserver-is-not-defined/#post-14401649 | **Open** | awaiting SH info
https://wordpress.org/support/topic/site-kit-encountered-an-error-34/ | **Open** | Awaiting SH info
https://wordpress.org/support/topic/fatal-error-3876/ | **Open** | [SH info](https://docs.google.com/forms/d/1lhjns2K7c0Ny7ME4gxTcsnYtNUtcMr8BvFpPQi23Y1E/edit#response=ACYDBNisbpvhpuTeGXncvNi1A8VkysLQ0H5Zclv09L6Xt7gHBH3uc_O-e_Ov9p_lhoBsGvQ)
[...]

PR #3891, opened 2021-08-17
Enhancement/3817 idea hub ga events

eugene-manuilov:
## Summary

Addresses issue #3817

## Checklist


assets/js/googlesitekit-idea-hub-post-list.js
```diff
@@ -46,9 +47,31 @@ domReady( () => {
...
+					: 'newposts_notification_dismiss'
+			);
+		} else if (
+			target.classList.contains( 'googlesitekit-idea-hub__saved-ideas' )
+		) {
+			trackEvent( 'idea_hub_posts', 'savedposts_notification_complete' );
+		} else if (
+			target.classList.contains( 'googlesitekit-idea-hub__new-ideas' )
+		) {
+			trackEvent( 'idea_hub_posts', 'newposts_notification_complete' );
```
aaemnnosttv, line 74: Should we await these since the click here will trigger a navigation?

assets/js/components/wp-dashboard/WPDashboardIdeaHub.js
```diff
@@ -63,12 +70,36 @@ function WPDashboardIdeaHub() {
...
+	} );
+
+	useEffect( () => {
+		if ( inView ) {
+			trackEvent(
+				'idea_hub_wpdashboard',
+				'savedposts_notification_view'
+			);
+		}
+	}, [ inView ] );
```
aaemnnosttv, line 85: Same as above – I think we should ensure this is only tracked once per page view.

includes/Modules/Idea_Hub.php
```diff
@@ -195,6 +195,23 @@ public function register() {
...
 
+			add_filter(
+				'googlesitekit_tracking_allowed',
+				function( $tracking_allowed ) {
+					if ( true === $tracking_allowed ) {
+						return $tracking_allowed;
+					}
+
+					// If tracking is not allowed yet, check whether it is posts list or dashboard page.
+					$screen = get_current_screen();
```
aaemnnosttv, line 206: We need to ensure this function exists first because the above filter will be called on the frontend as well if the SK admin bar is enabled.

includes/Core/Util/Tracking.php
```diff
@@ -117,7 +117,11 @@ public function is_active() {
 	 */
 	private function inline_js_base_data( $data ) {
 		global $hook_suffix;
-		$data['trackingAllowed'] = $this->screens->get_screen( $hook_suffix ) instanceof Screen;
+
+		$screen           = $this->screens->get_screen( $hook_suffix );
+		$tracking_allowed = $screen instanceof Screen;
+
+		$data['trackingAllowed'] = apply_filters( 'googlesitekit_tracking_allowed', $tracking_allowed );
```
aaemnnosttv, line 124: I understand why this was changed but this wasn't in the IB so let's make sure @felixarntz is on board with it.

Alternatively, we could register proper `Screens` for these even though they're not Site Kit™️ screens but that might be worth considering instead as we have a few other workarounds (such as enqueueing assets) which might be cleaner with a `Screen` as well.

aaemnnosttv (changes requested):
Thanks @eugene-manuilov ! Overall this is looking good. There were a few places that I think we want to limit the number of possible events tracked - although this was not in the ACs but I think it's a reasonable assumption to make. We can confirm with @felixarntz (there are a few other questions for him here as well)

There was one event from the ACs that that I didn't find for `click_outgoing_link` so I think that still needs to be added, otherwise all the events were accounted for.

See the rest of my comments below.

assets/js/components/wp-dashboard/WPDashboardIdeaHub.js
eugene-manuilov: @aaemnnosttv we use `triggerOnce: true` param in the `useInView` hook to make sure this is triggered only once.

assets/js/googlesitekit-idea-hub-post-list.js
eugene-manuilov: I think that even if we add `await` here, the event listener will not wait till the promise is resolved. Anyway, when I tested it on my local machine, it worked correctly.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -80,17 +88,60 @@ const DashboardIdeasWidget = ( {
...
+	const [ trackingRef, inView ] = useInView( {
+		triggerOnce: true,
+		threshold: 0.25,
+	} );
+
+	useEffect( () => {
+		if ( inView ) {
+			trackEvent( 'idea_hub_widget', 'widget_view' );
+		}
+	}, [ inView ] );
```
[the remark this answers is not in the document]
eugene-manuilov: It won't trigger multiple times because we use `triggerOnce: true` parameter in the `useInView` hook settings.

includes/Core/Util/Tracking.php
eugene-manuilov: Tracking doesn't work on WP dashboard and post list pages without these modifications.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -80,17 +88,60 @@ const DashboardIdeasWidget = ( {
...
+	useEffect( () => {
+		if ( trackedGatheringDataEvent ) {
+			return;
+		}
+
+		if (
+			newIdeas?.length === 0 &&
+			savedIdeas?.length === 0 &&
+			draftIdeas?.length === 0
+		) {
```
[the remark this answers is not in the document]
eugene-manuilov: This is needed only for this component. I have updated the selector to produce two variables: `hasNoIdeas` and `hasManyIdeas` to cover our needs and avoid duplicates.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -80,17 +88,60 @@ const DashboardIdeasWidget = ( {
...
+			draftIdeas?.length > 0
+		) {
+			trackEvent(
+				'idea_hub_widget',
+				'default_tab_view',
+				DashboardIdeasWidget.tabIDsByIndex[ activeTabIndex ]
+			);
+		}
+
+		setTrackedGatheringDataEvent( true );
```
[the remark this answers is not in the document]
eugene-manuilov: Sounds good to me. Updated.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -80,17 +88,60 @@ const DashboardIdeasWidget = ( {
...
+	}, [
+		newIdeas,
+		savedIdeas,
+		draftIdeas,
+		trackedGatheringDataEvent,
+		setTrackedGatheringDataEvent,
+		activeTabIndex,
+	] );
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

includes/Modules/Idea_Hub.php
eugene-manuilov: Updated.

assets/js/components/wp-dashboard/WPDashboardIdeaHub.js
aaemnnosttv: Ah thanks, I missed that part 👍

includes/Core/Util/Tracking.php
aaemnnosttv: Yeah, I forgot that `Screen`s don't really support using for existing/WP-core screens which was my thought there to fit within the existing logic here. I'm not sure we should introduce the ability to disable this for all screens though – that's what the opt-in is for.

Alternatively to using this filter as an on/off switch, we could do something similar to the screen check where we could introduce the concept of "modules with core screens" where those modules return an array of hook-suffixes or `WP_Screen` IDs that we could check for here. That could be useful in other ways as well 🤔  @felixarntz ?

felixarntz: I agree with @aaemnnosttv's concern about introducing a filter here. When introducing filters, we should think through whether we really want to open that respective area up for _any_ changes, since it can be done by third parties too.

In this particular case, I think Site Kit should for now remain in control of where our tracking generally works, since we also don't offer APIs do integrate Site Kit with further areas of WordPress.

I'm not sure about what would be the best API to implement here to facilitate this use-case, a generic "module with core screens" approach may also be too broad for only defining where to load tracking. I'd vote for being pragmatic here and add the Idea Hub logic (if the module is active and connected) directly into the `Tracking` class here. Clearly this is not an architecturally clean approach, but we also shouldn't build APIs if we can't assess whether something else in the future will use it or how something else in the future may want to use it - we can always add `TODO` comments to such areas to potentially revisit later.

aaemnnosttv: > I'm not sure about what would be the best API to implement here to facilitate this use-case, a generic "module with core screens" approach may also be too broad for only defining where to load tracking.

@felixarntz I think it could have other uses, not only for this. Alternatively, we could leverage the existing `Module_With_Screens` functionality and [the existing logic in place here](https://github.com/google/site-kit-wp/blob/f46f548d709f4224a7af75f40701b3b05c9c848f/includes/Core/Util/Tracking.php#L120) related to it by introducing a new subclass of `Screen` for using with core screens (i.e. screens we are not registering/adding to the menu but we could still hook into `load-$hook_suffix`). Maybe the subclass isn't the best approach here since the suggested new class would actually be simpler than the current `Screen` class. That way we'd only be making the existing API more flexible in a reasonable way for a use-case I think we can see is becoming more relevant as SK starts having a larger footprint in WP admin.

Perhaps this is still out of scope for this issue, so I wouldn't be opposed to doing something inline for the sake of the smallest change that doesn't require added filters/APIs yet and then we can refactor in a follow-up once the new approach is defined.

eugene-manuilov: Ok, I have reverted modifications in the Tracking class and leveraged the existing filter to update js inline data. Let me know if it works.

includes/Modules/Idea_Hub.php
```diff
@@ -212,6 +212,31 @@ public function register() {
...
+						! function_exists( 'get_current_screen' )
+					) {
+						return $data;
+					}
+
+					$screen = get_current_screen();
+					if ( ! is_null( $screen ) ) {
+						$data['trackingAllowed'] =
+							( 'post' === $screen->post_type && 'edit-post' === $screen->id ) ||
+							'dashboard' === $screen->id;
```
felixarntz, line 232: This is still obviously kind of hacky overriding only the JS value, but it makes sense given we don't want to introduce a dedicated filter at this point. Let's open a follow-up issue revisiting how we should treat core screens that have some very limited Site Kit usage. It's not necessarily only about this, but also about loading certain assets on core screens, where I'm sure we can improve the situation by building a proper API.

cc @aaemnnosttv

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -87,6 +118,46 @@ const DashboardIdeasWidget = ( {
...
+		if ( hasNoIdeas || hasManyIdeas ) {
+			setTrackedWidgetView( true );
+		}
+	}, [
+		hasNoIdeas,
+		hasManyIdeas,
+		trackedWidgetView,
+		setTrackedWidgetView,
+		activeTabIndex,
+	] );
```
felixarntz, line 159: This implementation is slightly different from what's described in the ACs (and what we do in the PSI widget): One of these two events should be sent right after `widget_view`. Right now they're sent regardless of whether the user scrolls to the widget or not.

assets/js/modules/idea-hub/components/dashboard/DashboardCTA.js
```diff
@@ -70,8 +90,17 @@ function DashboardCTA( { Widget, WidgetNull } ) {
 		}
 	}, [ activateModule, navigateTo, setInternalServerError ] );
 
+	const onLinkClick = useCallback( () => {
+		trackEvent( IDEA_HUB_GA_CATEGORY_WIDGET, 'click_outgoing_link' );
```
felixarntz, line 94: This event is missing the event label from the ACs.

felixarntz (changes requested):
There are two places here where the implementation doesn't fully match the ACs - other than that, this is good to go.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
eugene-manuilov: Updated.

assets/js/modules/idea-hub/components/dashboard/DashboardCTA.js
eugene-manuilov: Updated.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -50,34 +55,60 @@ import Empty from './Empty';
...
-		select( MODULES_IDEA_HUB ).getDraftPostIdeas()
+	return global.window.pageYOffset + widgetOffset - headerOffset;
+}
+
+function DashboardIdeasWidget( props ) {
+	const { defaultActiveTabIndex, Widget, WidgetReportError } = props;
+
+	const [ trackedWidgetView, setTrackedWidgetView ] = useState( false );
+
+	const { hasNoIdeas, hasManyIdeas, savedIdeas, draftIdeas } = useSelect(
```
aaemnnosttv, line 81: Let's keep the previous style of selection here as the nested select isn't needed here.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -87,6 +118,49 @@ const DashboardIdeasWidget = ( {
...
+			trackEvent(
+				IDEA_HUB_GA_CATEGORY_WIDGET,
+				'default_tab_view',
+				DashboardIdeasWidget.tabIDsByIndex[ activeTabIndex ]
+			);
+		}
+
+		if ( hasNoIdeas || hasManyIdeas ) {
+			setTrackedWidgetView( true );
+		}
```
aaemnnosttv, line 154: Let's move this call right before the `trackEvent`s above rather than duplicating the condition again here. It means duplicating the set state call once but it's a bit simpler and also makes it a little more clear what the state represents 🙂  It could go right after too, I generally prefer to set it closer to the guard and before the action (once qualified) rather than after.

aaemnnosttv (commented):
Thanks @eugene-manuilov – this is nearly good to go, just a few small things I noted.

Otherwise my only remaining concern is still https://github.com/google/site-kit-wp/pull/3891/files#r690974091 – if @felixarntz is okay with merging this first and resolving that before the release, or if he feels it's just not a big concern, then that's fine with me too.

felixarntz:
@aaemnnosttv I think it's okay to merge before #3278, since Idea Hub is also not exposed to end users just yet. Since #3278 is already being worked, it's only important we ship this before releasing Idea Hub to the public.

felixarntz (approved):
Nothing to add from my end to @aaemnnosttv's last points, lgtm!

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
eugene-manuilov: Updated.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
eugene-manuilov: Updated.

aaemnnosttv (approved):

aaemnnosttv:
VRT failure is only the buttons scenario which is unstable, so that can be ignored.

Issue #3886, opened 2021-08-17
Convert useHash to useQueryParam in Idea Hub


tofumatt:
## Feature Description

We have one remaining use of `useHash` to manage Idea Hub tabs on the dashboard—we should convert this to use React Router for better control of the link behaviour (see #3796) and to remove the `useHash` hook entirely.

Having thought about this more though, we shouldn't rely on hash URLs for non-page-level components/routing. Hash routing should handle page-wide routes, and things like widget routing/tabs should be handled in-memory. There's no reason to change the URL based on the Idea Hub tab—the main design for hash routes is to be able to load a particular tab on page load.

To do that, let's use a query param and set the Idea Hub's default route to be that param's value if detected.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `useHash` hook should be removed from usage.
* Idea Hub's tabs on the Site Kit Dashboard should use query params and not add to the browser's history.

## Implementation Brief

* Add a `MemoryRouter` to the `assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js`
* Remove the `useHash` hook, instead getting the initial Idea Hub tab position from the `useSearchParam` hook, given a `MemoryRouter` will be used.
* Replace each Tab with a format like:
```js
<Tab
  replace
  tag={ Link }
  to="/draft-ideas"
>
```
* Replace the body of the IdeaHub widget with a `<Switch><Route exact path="/new-ideas"><NewIdeas WidgetReportError={ WidgetReportError }/></Route></Switch>`, except with a route for each tab.
* If a value for the `idea-hub-tab` query param is set (from `useSearchParam('idea-hub-tab')`), use the appropriate component for the "default" route (eg. `<Route exact path="/"><SavedIdeas WidgetReportError={ WidgetReportError }/></Route>` for `idea-hub-tab=saved-ideas`). Use `<NewIdeas />` if no query param is set.
* Update `includes/Modules/Idea_Hub.php` to use the new `?idea-hub-tab=new-ideas` type URL over the existing `#new-ideas` ones

See #3897 for a proof-of-concept PR.

### Test Coverage

* Some of the tests for Idea Hub routes should be updated from `#new-drafts` to use the query params eg `?idea-hub-tab=new-drafts`.

### Visual Regression Changes

* N/A

## QA Brief

* The Idea Hub tabs should continue to work as-expected when changing between tabs.
[...]

Issue #3860, opened 2021-08-12
the idea tab doesn't show my idea in the draft tab if i return to the dashboard page by clicking on the back button.


simonchampion:
## Bug Description

the idea tab doesn't show my idea in the draft tab if i do the following: 1. click on create draft, go to the draft tab, click on view draft for the idea i just clicked on create draft, after getting to the draft compose page hit the back navigation button.  the idea for the draft i was just looking at is missing from the list of drafts and if i click on the new tab, i see it back there.  after refresh, the drafts tab show the idea again.  however, the new tab still shows the drafted ideas even though when i first clicked on the create draft, they disappeared

**Asana** https://app.asana.com/0/1200491083500938/1200760567419961

## Screenshots

https://user-images.githubusercontent.com/73545194/130076030-91783abf-2182-4060-b9e0-f96d7a89987f.mp4

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* When the user goes back to the dashboard pave using the Back button in the browser, the previous state of the Idea Hub widget should remain the same as before leaving the dashboard page.

## Implementation Brief

* This issue is solved by using React Router, which will reliably load the correct tab on page load. See #3886.

### Test Coverage

* N/A

### Visual Regression Changes

* N/A

## QA Brief

* Visit http://sitekit.10uplabs.com/wp-admin/admin.php?page=googlesitekit-dashboard&idea-hub-tab=draft-ideas
* Click on a link to another WordPress dashboard page.
* Click the browser's "Back" button.
* The Draft Ideas tab and content should be active.

## Changelog entry

* Improve Idea Hub tab behaviour when using browser navigation.

eugene-manuilov:
I can't reproduce this issue. @wpdarren can you take a look at it too?

wpdarren:
@eugene-manuilov yes, I was able to recreate this issue.  I followed the steps, here's a screencast.

https://user-images.githubusercontent.com/73545194/130076030-91783abf-2182-4060-b9e0-f96d7a89987f.mp4

eugene-manuilov:
Thanks, @wpdarren. Which browser do you use?

wpdarren:
@eugene-manuilov I used Chrome. Would you like me to try with another browser?

eugene-manuilov:
No, that's fine, just wanted to double-check it.

aaemnnosttv:
IB ✅

tofumatt:
[...]

Issue #3796, opened 2021-08-03
Browser navigation is broken after selecting different tabs in the ideas widget


eugene-manuilov:
## Bug Description

When I switch tabs in the ideas widget multiple times, my browser history is full of similar records with just different hashes. It causes a bad user experience because if I want to go back to the previous page, I'll need to click on the back button the same number of times that I switched tabs. Plus, when I click back and the URL changes to the previous one (with a different hash), the selected tab in the ideas widget doesn't change.

We need to either update tabs in the ideas widget to follow the hash changes or update how we change hashes to avoid generating duplicate records in the browser history.

https://user-images.githubusercontent.com/873049/128038754-267dd890-58a1-4116-95f1-7c1c2bcc5542.mp4

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Changing tabs on the Idea Hub dashboard widget should not add new items to the browser history
   * I.e. clicking "Back" should invoke a page navigation rather than a hash change

## Implementation Brief

Using the files from the existing open `react-use` PR https://github.com/streamich/react-use/pull/1435
 
1. Create `assets/js/hooks/useHash.js`
1. Copy the contents of https://github.com/streamich/react-use/blob/ed831e6db478d739d3638a8efca37418aba1328e/src/useHash.ts into this file
1. Remove the TypeScript typings
1. Fix the imports (need to import dependency hooks from `react-use`)
1. Fix eslint warnings (`window` should be `global`)

The code should now build.

In `assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js`
1. Import new `useHash` hook (instead of from `react-use`)
1. Where `setHash` is used, pass in `false` as a new second argument

See this all done on existing PR https://github.com/google/site-kit-wp/pull/3862

### Test Coverage

* Copying existing tests from PR to new `assets/js/hooks/useHash.test.js` https://github.com/streamich/react-use/blob/ed831e6db478d739d3638a8efca37418aba1328e/tests/useHash.test.ts
* Fix the imports (need to import `useHash` from the same folder)
* Fix eslint warnings

### Visual Regression Changes

* None

## QA Brief

* Visit any page that is not the Site Kit dashboard
* Navigate to the Site Kit dashboard by the menu bar
* Change tabs on the Idea Hub widget
* Observe that the query param has updated in the browser URL
* Use the back button
* Observe that the back button works as expected and goes back to the previously visited page

## Changelog entry

[...]

PR #3897, opened 2021-08-17
Add React Router to Idea Hub.

tofumatt:
## Summary

Addresses issue #3886.
Addresses issue #3796.

By changing the query param using our `useQueryArg` hook, this also addresses #3860.

## Checklist


assets/js/components/Link.js
```diff
@@ -27,143 +27,154 @@ import { Link as RouterLink } from 'react-router-dom';
...
-	to,
-	...otherProps
-} ) {
-	const getType = () => {
-		// Force button element if `onClick` prop is passed and there's no `href`
-		// or `to` prop.
-		if ( ! href && ! to && onClick ) {
-			if ( disabled ) {
-				return BUTTON_DISABLED;
+const Link = forwardRef(
```
tofumatt, line 38: This was needed because the `Tab` components attach a `ref` to the components they render.

In general this is a nice-to-have and doesn't affect the component in any way besides offering refs to parent components that need it.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -100,9 +119,12 @@ const DashboardIdeasWidget = ( {
 	const handleTabUpdate = useCallback(
 		( tabIndex ) => {
 			setActiveTabIndex( tabIndex );
-			setHash( DashboardIdeasWidget.tabIDsByIndex[ tabIndex ] );
+			setQueryString( {
+				'idea-hub-tab': DashboardIdeasWidget.tabIDsByIndex[ tabIndex ],
+				page: queryParamPage,
+			} );
```
aaemnnosttv, line 125: Can we not update only the one query param without writing the whole query string? Looks like we even have our own custom hook for this from a while back: https://github.com/google/site-kit-wp/blob/develop/assets/js/hooks/useQueryArg.js

package.json
```diff
@@ -150,6 +150,7 @@
 		"svgxuse": "^1.2.6",
 		"url-polyfill": "^1.1.9",
 		"use-memo-one": "^1.1.2",
+		"use-query-string": "^2.3.0",
```
aaemnnosttv, line 153: See my previous comment – I think this might not be necessary.

aaemnnosttv (changes requested):
Thanks @tofumatt! This looks good overall but it loses the important height-preservation behavior of the widget since only a single tab is rendered.

https://user-images.githubusercontent.com/1621608/131654082-066ff5a4-3221-42db-a1e8-377bdea884c9.mp4

A few other comments for you below.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
tofumatt: Hah, nice, I totally forgot about it. Let's use that then! 👍🏻

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -166,25 +188,27 @@ const DashboardIdeasWidget = ( {
...
+							</Route>
+							<Route exact path="/draft-ideas">
+								<DraftIdeas
+									WidgetReportError={ WidgetReportError }
+								/>
+							</Route>
+							<Route path="*">
+								<Redirect to={ activeTab } />
+							</Route>
+						</Switch>
```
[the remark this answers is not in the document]
tofumatt: Indeed, this does break it: 

I'll have a think on how best to approach this then.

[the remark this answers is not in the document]
tofumatt: I tried this on `develop` and I'm still seeing a varying height there, but what I've committed seems to keep the height pretty consistent between pages and tabs.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -55,25 +57,18 @@ import SavedIdeas from './SavedIdeas';
...
-function isValidHash( hash ) {
-	return getHash( hash ) in DashboardIdeasWidget.tabToIndex;
-}
-
-function getIdeaHubContainerOffset( widgetOffset ) {
-	const header =
+const getIdeaHubContainerOffset = ( ideaHubWidgetOffsetTop ) => {
+	const siteHeaderHeight =
 		document.querySelector( '.googlesitekit-header' )?.offsetHeight || 0;
-	const adminBar = document.getElementById( 'wpadminbar' )?.offsetHeight || 0;
```
aaemnnosttv, line 71: The height of the admin bar is relevant here, but only at `window.innerWidth > 600` which is when it becomes fixed and our CSS adds a `top` rule for the SK header.

I'm guessing it doesn't matter though as long as the SK header is positioned properly relative to it?

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -295,25 +301,23 @@ function DashboardIdeasWidget( props ) {
 				</div>
 
 				<div className="googlesitekit-idea-hub__body">
-					<div
-						className="googlesitekit-idea-hub__content"
-						aria-hidden={ activeTab !== 'new-ideas' }
```
aaemnnosttv, line 300: Is there a reason we removed the `aria-hidden` attribute and moved to a single wrapping `__content` div? It seems like this would be still relevant to keep.

aaemnnosttv (commented):
Thanks @tofumatt – this looks great, the only thing is the widget height which is still not preserved (I believe due to the consolidated `__content` divs into one).

Here's the current behavior in this PR compared to develop

https://user-images.githubusercontent.com/1621608/131804481-7ae014f0-9181-41bc-a088-bbb2da504dc3.mp4

develop:

https://user-images.githubusercontent.com/1621608/131804518-11c69c50-b1fe-4ae0-af60-225a84734c7f.mp4

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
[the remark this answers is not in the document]
aaemnnosttv: The only time I see a change (that we can't currently avoid) is between pages of ideas, it doesn't shift between tabs though on develop, see my screencast below.

aaemnnosttv:
@googlebot I consent.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
aaemnnosttv: Fixed this as it wasn't accounted for 👍

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
aaemnnosttv: Restored this part as the `<Route>` elements weren't needed and we need to always render each tab's content to preserve the height.

aaemnnosttv (approved):
Thanks @tofumatt – I pushed a few fixes on top of where you left off here so sending this back your way for a final pass but I think this should be good to go now.

tofumatt (commented):
While this no longer uses React Router at all, it does solve all the issues it intended to, so looks good 👍🏻

Issue #3841, opened 2021-08-12
Mobile: Height of ideas list makes for poor UX changing tabs or pagination


simonchampion:
## Bug Description

The height of the list makes it so users will need to scroll down to use the pagination and scroll up again to switch tabs which makes for a rough experience.

Darren: _Yes, just noticed this too. When you paginate, you are not sent to the top of the widget._

**Asana link:** https://app.asana.com/0/1200491083500938/1200759450014575

## Screenshots

![image](https://user-images.githubusercontent.com/83005909/129164544-46a51071-572a-4cfa-ba92-e2f467b0e30d.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* When navigating between pages in the Idea Hub widget, the viewport should scroll to the top of the current tab if it is smaller than the entire list (e.g. can happen on mobile devices).
    * The height itself should not be updated here, that is part of #3843.

## Implementation Brief

* When the page for Idea Hub pagination changes, scroll to the top of the Idea Hub widget  _if_ the Idea Hub widget header is not on-screen. 
* This behaviour should be placed in a `useUpdateEffect` hook that relies on the pagination value changing, so it doesn't trigger on first mount. It can be placed in the `assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js` file as it already has a reference to the `ideaHubContainer` ref.
* See #3898 for a proof-of-concept PR.

### Test Coverage

* N/A

### Visual Regression Changes

* N/A

## QA Brief

* On a small screen and/or mobile device, visit the Idea Hub widget on either the Site Kit dashboard or in Storybook—the behaviour should be the same in either place.
* Visit the second page of pagination results.
* Scroll such that the header of the Idea Hub widget is not on screen.
* Click a pagination link to go to the previous, taller page.
* The widget should be scrolled into view.

Following the steps above, if the widget was already entirely in view (eg on a large screen, tablet, etc.) the viewport should NOT be scrolled.

## Changelog entry

* Update Idea Hub widget pagination to scroll to widget top on page change if out of view.

danielgent:
The AC says
_"if it is smaller than the entire list (e.g. can happen on mobile devices)."_

But I don't think comparing `height of widget < height of viewport` is the correct check, because the user could have scrolled way down so that only the pagination buttons are visible.

IMHO we should just check if the header of the Ideas Hub Widget is visible, and scroll up if it's not.

This is my idea on desktop

[...]

PR #3898, opened 2021-08-17
Scroll to Idea Hub header when pagination changes.

tofumatt:
## Summary

Addresses issue #3841.

## Checklist


assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -105,6 +112,28 @@ const DashboardIdeasWidget = ( {
...
+			tabBarRectangle.bottom <=
+				( global?.innerHeight ||
+					global?.document?.documentElement?.clientHeight ) &&
+			tabBarRectangle.right <=
+				( global?.innerWidth ||
+					global?.document?.documentElement?.clientWidth );
+
+		if ( ! isOnScreen ) {
+			global.window.scrollTo( {
+				top: getIdeaHubContainerOffset( tabBarRectangle.top ),
```
aaemnnosttv, line 131: This function results in an improper scroll position on mobile due to what seems to be the added offset of the wp-admin bar height but this shouldn't be added in this case because it isn't offsetting anything. Only the sticky SK header needs to be accounted for.

Position after scrolling: 
image

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -105,6 +112,28 @@ const DashboardIdeasWidget = ( {
...
+
+		const isOnScreen =
+			tabBarRectangle.top >= 0 &&
+			tabBarRectangle.left >= 0 &&
+			tabBarRectangle.bottom <=
+				( global?.innerHeight ||
+					global?.document?.documentElement?.clientHeight ) &&
+			tabBarRectangle.right <=
+				( global?.innerWidth ||
+					global?.document?.documentElement?.clientWidth );
```
aaemnnosttv, line 127: Are all these conditional chains needed?

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -105,6 +112,28 @@ const DashboardIdeasWidget = ( {
 		[ setHash, setActiveTabIndex ]
 	);
 
+	// Any time the pagination value changes, scroll to the top of the container.
+	useUpdateEffect( () => {
+		const tabBarRectangle = tabBarHeaderRef.current.getBoundingClientRect();
```
aaemnnosttv, line 117: Is it possible the ref might not be set here? The ref isn't added in all cases as there is an early return below if there are zero total ideas.

aaemnnosttv (changes requested):
Thanks @tofumatt – this looks good overall, just a few questions and things to address.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
tofumatt: Yeah, I think it's possible, I need to protect against it 👍🏻

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
tofumatt: _Probably not, but I'm always wary of accessing things off the `global`/`window`. But these should be fine 😄

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
tofumatt: I've removed the WP Admin from the calculation and it works much better, thanks 👍🏻

tofumatt:
Merge conflict resolved, back to you 🙂

aaemnnosttv (approved):
@tofumatt this works well for me, thanks!

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -200,6 +206,32 @@ function DashboardIdeasWidget( props ) {
...
+		if ( ! tabBarRectangle ) {
+			return;
+		}
+
+		const isOnScreen =
+			tabBarRectangle.top >= 0 &&
+			tabBarRectangle.left >= 0 &&
+			tabBarRectangle.bottom <=
+				( global.innerHeight ||
+					global?.document.documentElement.clientHeight ) &&
```
aaemnnosttv, line 222: I think we can rely on `document` always being there as we have below.
```suggestion
					global.document.documentElement.clientHeight ) &&
```

Issue #3852, opened 2021-08-12
WP post list CTA language could be improved for consistency with dashboard CTA


simonchampion:
## Bug Description

[See WP dashboard CTA inconsistent with latest design and language](https://app.asana.com/0/1200491083500938/1200759450014532)

![image](https://user-images.githubusercontent.com/83005909/129230533-cad6637c-7e3d-449f-881d-6311a9d21381.png)

**Asana**: https://app.asana.com/0/1200491083500938/1200759450014538

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

**For the case where the user has already saved some ideas**
- Update text to: "Want some inspiration for a new post? Revisit your saved ideas in Site Kit"
- The link CTA is this part of the second sentence: "Revisit your saved ideas" 
- Remove the separate "See saved ideas" link CTA that currently exists. 

**For the case where the user has no saved ideas**
- Update text to: "Want some inspiration for a new post? Review your new ideas in Site Kit"
- The link CTA is this part of the second sentence: "Review your new ideas" 
- Remove the separate "See new ideas" link CTA that currently exists. 

Eg. 
![image](https://user-images.githubusercontent.com/1621608/129882609-69d5e33b-8407-413c-a86d-a66cff7ac021.png)

## Implementation Brief
* Using `includes/Modules/Idea_Hub.php`, update the notices in the `admin_notice_idea_hub_ideas` function as per the AC.

### Test Coverage
* No new tests to be added.

### Visual Regression Changes
* N/A

## QA Brief
* Check if the notices in the WP post list page are updated as per the AC when there are/are not saved ideas.

## Changelog entry

* Update CTA language for WP post list notifications.

aaemnnosttv:
@marrrmarrr I added a screenshot to the ACs, I think this is all set now 👍

eugene-manuilov:
IB :heavy_check_mark:

wpdarren:
### QA Update: ❌ 

Within idea hub I have [no saved ideas](https://user-images.githubusercontent.com/73545194/130078073-43bdc8f4-f78e-4fff-a768-3959081bcd00.png) but the message in the posts screen is: `Want some inspiration for a new post? Revisit your new ideas in Site Kit.` which does not match the AC. 

![image](https://user-images.githubusercontent.com/73545194/130078292-9c4dae09-0efa-4d71-8055-4558544bfddd.png)

**Verified:**

[...]

PR #3904, opened 2021-08-18
Update CTA language.

asvinb:
## Summary

Addresses issue #3852 

## Checklist

### DIFF
--- a/includes/Modules/Idea_Hub.php
+++ b/includes/Modules/Idea_Hub.php
@@ -253,13 +253,13 @@ final class Idea_Hub extends Module
 			self::SLUG_SAVED_IDEAS,
 			array(
 				'content'         => function() {
-					return sprintf(
-						'<p>%s <a href="%s">%s</a> %s</p>',
-						esc_html__( 'Want some inspiration for a new post?', 'google-site-kit' ),
-						esc_url( $this->context->admin_url() . '#saved-ideas' ),
-						esc_html__( 'Revisit your saved ideas', 'google-site-kit' ),
-						esc_html__( 'in Site Kit.', 'google-site-kit' )
+					$message = sprintf(
+						/* translators: %s: URL to saved ideas */
+						__( 'Want some inspiration for a new post? <a href="%s">Revisit your saved ideas</a> in Site Kit.', 'google-site-kit' ),
+						esc_url( $this->context->admin_url() . '#saved-ideas' )
 					);
+
+					return $this->escape_notice_content( $message );
 				},
 				'type'            => Notice::TYPE_INFO,
 				'active_callback' => function() use ( $dismissed_items ) {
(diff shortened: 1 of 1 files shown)


includes/Modules/Idea_Hub.php
```diff
@@ -254,10 +254,11 @@ private function admin_notice_idea_hub_ideas( $notices ) {
...
 				'content'         => function() {
 					return sprintf(
-						'<p>%s <a href="%s">%s</a></p>',
-						esc_html__( 'Need some inspiration? Revisit your saved ideas in Site Kit.', 'google-site-kit' ),
+						'<p>%s <a href="%s">%s</a> %s</p>',
+						esc_html__( 'Want some inspiration for a new post?', 'google-site-kit' ),
 						esc_url( $this->context->admin_url() . '#saved-ideas' ),
-						esc_html__( 'See saved ideas', 'google-site-kit' )
+						esc_html__( 'Revisit your saved ideas', 'google-site-kit' ),
+						esc_html__( 'in Site Kit.', 'google-site-kit' )
```
eugene-manuilov, line 261: @asvinb it should be one translatable string:

```php
$message = sprintf(
    /* translators: %s: URL to saved ideas */
    __( 'Want some inspiration for a new post? <a href="%s">Revisit your saved ideas</a> in Site Kit.', 'google-site-kit' ),
    esc_url( $this->context->admin_url() . '#saved-ideas' ),
);

return "<p>{$message}</p>"
```

Same should be done for the new ideas notification.

eugene-manuilov (changes requested):

includes/Modules/Idea_Hub.php
asvinb: 👍 Ready for another pass @eugene-manuilov

includes/Modules/Idea_Hub.php
```diff
@@ -965,4 +967,24 @@ function( $idea ) use ( $ideas_with_posts ) {
...
+		$message = wp_kses(
+			$message,
+			array(
+				'a' => array(
+					'href' => array(),
+				),
+			)
+		);
+
+		return '<p>' . $message . '</p>';
```
eugene-manuilov, line 988: @asvinb this is a side effect for this function. We probably need to rename it to something like `escape_and_wrap_notice_content` to better reflect what this method does. Or we can also make this method be an anonymous function in the `admin_notice_idea_hub_ideas` method. It can be done something like this:

```php
private function admin_notice_idea_hub_ideas( $notices ) {
    ...

    $escape_and_wrap_notice_content = function( $message ) {
        $message = wp_kses( ... );
        return $message;
    };
    
    ...
    
    $notices[] = new Notice( self::SLUG_SAVED_IDEAS, array(
        'content'         => function() use ( $escape_and_wrap_notice_content ) {
            $message = sprintf( ... );
            return $escape_and_wrap_notice_content( $message );
        },
        ...
    ) );
    
    ...
}
```

eugene-manuilov (changes requested):

includes/Modules/Idea_Hub.php
asvinb: Nice. Updated! I guess am not sure when to use an anonymous function in PHP or just have it as a method in the class? 🤔

eugene-manuilov: We don't have specific rules for anonymous functions in PHP. In general, if it is a small ad-hoc function for a specific method, then we can use an anonymous function.

Issue #3865, opened 2021-08-12
Update Idea Hub widget footer based on selected tab


aaemnnosttv:
## Feature Description

In #3773 we moved the Idea Hub widget footer into the widget itself (rather than in the "body" of the component).

This brought the widget to be more consistent with others but left a few things to be addressed.

Other related points were raised in the bug bash which can be addressed here as well.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

The following changes all apply to the Idea Hub dashboard widget

* The "Updated every 2-3 days" text in the footer should only be displayed on the tab for new ideas as these are the only ideas that are refreshed from the API
* The footer should not be displayed for zero-idea states – except on the new ideas tab (as the previously mentioned message is still relevant)
* Pagination in the footer should not be shown for zero-idea states (only applies to new ideas per above)
* The height of the widget should not change based on the selected tab (#3723)
* The "This may take 24 hours" text should be removed in all cases
* The following changes/exceptions only apply to the zero total ideas state (no new/saved/draft ideas)
	* The widget tabs should not be displayed at all (as is currently)
    * The widget footer should not be displayed at all

## Implementation Brief
* Once https://github.com/google/site-kit-wp/issues/3832 has been merged.
* Using `assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js`,
	* Check if selectors have been resolved and there are no errors for each of the following selectors: `getNewIdeas`, `getSavedIdeas` and `getDraftPostIdeas` via the `hasFinishedResolution` and `getErrorForSelector` selectors.
	* `WrappedFooter` should be a variable, with `null` as an initial value instead of a function. The reason being we are passing a function as the value to the `Footer` prop of `Widget` which implies the prop is not null. Thus, even if the function returns `null`, an empty footer will be displayed.
	* Depending on the value of `activeTab`,
		* if it's `new-ideas`, check if there are any errors for the `getNewIdeas` selector and if it has finished resolving, to assign `WrapperFooter` the `Footer` component, passing on the `activeTab` as value to the `tab` prop and `footerText` prop set to the translatable `Updated every 2-3 days` string.
		* if it's `saved-ideas` or `draft-ideas`, check if there are any errors for the `getSavedIdeas` or `getDraftPostIdeas` selectors and if they have finished resolving to assign `WrapperFooter` the `Footer` component.
[...]

PR #3915, opened 2021-08-19
Enhancement/3865 widget footer

eugene-manuilov:
## Summary

Addresses issue #3865

## Checklist


assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -64,21 +64,37 @@ const getIdeaHubContainerOffset = ( ideaHubWidgetOffsetTop ) => {
...
+			manyIdeas = true;
+		}
+
+		return {
+			hasNoIdeas: noIdeas,
+			hasManyIdeas: manyIdeas,
+			savedIdeas: saved,
+			draftIdeas: draft,
+		};
+	} );
```
eugene-manuilov, line 97: This has been added from #3891 to reduce the number of merge conficts.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -193,26 +193,26 @@ function DashboardIdeasWidget( props ) {
...
-					: false
-			}
-		/>
-	);
+	let WrappedFooter;
+	if ( activeTab === 'new-ideas' ) {
+		WrappedFooter = () => (
+			<Footer
+				tab={ activeTab }
+				footerText={ __( 'Updated every 2-3 days', 'google-site-kit' ) }
```
tofumatt, line 207: Just a question: why use a special prop here instead of `children`? Eg. `<Footer tab={ activeTab }>{ __( 'Updated every 2-3 days', 'google-site-kit' ) }</Footer>`

Not a big deal either way, but using `children` here seems fine 🤷🏻‍♂️

tofumatt (changes requested):
The IB mentions:

> Check if selectors have been resolved and there are no errors for each of the following selectors: `getNewIdeas`, `getSavedIdeas` and `getDraftPostIdeas` via the `hasFinishedResolution` and `getErrorForSelector` selectors.

But I don't see those checks added. For all tabs, I think the intention was to ensure they didn't have errors and had resolved to use the new footer, otherwise to leave it blank.

Otherwise looks good—just one minor comment about using `children` over the custom `footerText` prop but that's not a big deal.

tofumatt (approved):
Just noting that the IB here mentions explicitly checking for the resolution/error state of the selectors, but the PR solves this by checking for `savedIdeas?.length`. From @eugene-manuilov:

> we don't really need those additional checks for errors and resolved status because we verify it implicitly when we check conditions like `savedIdeas?.length > 0`. If there is an error or the request is not resolved yet, the `savedIdeas` variable will be `undefined` which will turn the condition to be `false`-y and the footer won't be displayed.

PR #3918, opened 2021-08-20
Improve multi-select behaviour by disabling options after the maximum number has been reached.

tofumatt:
## Summary

Addresses issue #3761.

## Relevant technical choices

See comment at https://github.com/google/site-kit-wp/issues/3761#issuecomment-902763594. Instead of allowing too many items to be selected and disabling the submit button, we disable options after too many are selected. This is in line with other components and a more reasonable UX.

## Checklist

### DIFF
--- a/assets/js/components/surveys/CurrentSurvey.test.js
+++ b/assets/js/components/surveys/CurrentSurvey.test.js
@@ -147,7 +147,7 @@ describe( 'CurrentSurvey', () => {
 			// Submit button should be enabled if text has been entered.
 			expect(
 				getByRole( 'button', { name: 'Submit' } )
-			).not.toHaveAttribute( 'disabled' );
+			).not.toBeDisabled();
 
 			// Clear and enter input again.
 			fireEvent.change( getByLabelText( 'Write here' ), {
@@ -162,7 +162,7 @@ describe( 'CurrentSurvey', () => {
 			} );
 			expect(
 				getByRole( 'button', { name: 'Submit' } )
-			).not.toHaveAttribute( 'disabled' );
+			).not.toBeDisabled();
 		} );
 
 		it( 'should submit answer in correct shape', async () => {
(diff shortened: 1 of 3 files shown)


assets/js/components/surveys/CurrentSurvey.test.js
```diff
@@ -447,31 +447,58 @@ describe( 'CurrentSurvey', () => {
...
-			fireEvent.click( getByText( 'Black Olives' ) );
-			fireEvent.click( getByText( 'Sweetcorn' ) );
+			fireEvent.click( getByLabelText( 'Pepperoni' ) );
+			fireEvent.click( getByLabelText( 'Sausage' ) );
+			fireEvent.click( getByLabelText( 'Mushrooms' ) );
 
-			expect( getByRole( 'button', { name: 'Submit' } ) ).toHaveAttribute(
+			// This option will be enabled because we still haven't selected the
+			// maximum number of items.
+			expect( getByLabelText( 'Sweetcorn' ) ).not.toHaveAttribute(
```
aaemnnosttv, line 463: Shouldn't we be using [`(not.)toBeDisabled`](https://github.com/testing-library/jest-dom#tobedisabled) rather than checking the presence of the attribute? The matcher (which we have available via `jest-dom`) is quite a bit more robust than this so I would prefer we use that unless we have a specific reason why we can't.

assets/js/components/surveys/SurveyQuestionMultiSelect.js
```diff
@@ -162,6 +162,12 @@ const SurveyQuestionMultiSelect = ( {
 							>
 								<Checkbox
 									checked={ answer.selected }
+									disabled={
+										hasMaximumNumberOfChoices &&
+										! answer.selected
+											? true
+											: undefined
+									}
```
aaemnnosttv, line 170: Per my last comment, this condition could be a bit simpler as a boolean rather than a ternary to use `undefined` (no attribute) which seems like an odd coupling to the test.

aaemnnosttv (commented):
Thanks @tofumatt a few questions for you but otherwise this works as expected.

One thing that stands out as a bit odd to me though is that there does not seem to be any visual difference between the choice or text between normal and disabled states (unlike User Input). Is something missing here?

tofumatt:
Good call on the disabled view, I've tweaked it to look like this:



The User Input app uses the same `0.25` opacity so it seemed appropriate.

aaemnnosttv (approved):
Great, thanks @tofumatt 👍

Issue #3839, opened 2021-08-12
Icons overlapping idea title.


asvinb:
## Bug Description
When the browser is resized to 960px, upon hovering an idea, the icons are displayed on top of the idea name.

## Steps to reproduce

1. Go to Site Kit dashboard
3. Scroll down to Idea Hub widget
4. Resize browser to 960px
5. Hover a suggested idea.

## Screenshots
<img src=https://user-images.githubusercontent.com/31818/129164142-b4c2258c-59e3-4ee2-beef-b4486b595c18.png width=400 />

## Additional Context

 - Browser: Chrome

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* For viewports where the Idea Hub widget appears in a narrow state (even if this viewport is technically not "mobile"), the buttons on each idea should appear below the idea rather than on the right of the idea, similar to how it is in the current mobile layout.

## Implementation Brief

Using `assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Idea.js` and `assets/sass/components/idea-hub/_googlesitekit-idea-hub-dashboard-ideas-widget.scss`:
* Replace the grid layout with normal `div`s:
  * Replace the `<Grid>` and `<Cell>` with `<div>`s with the same `className` attributes (`.googlesitekit-idea-hub__idea--actions` and `.googlesitekit-idea-hub__idea--details` respectively). 
  * The `<Row>` can be replaced with a div with a suitable className (`div.googlesitekit-idea-hub__idea--wrapper` for example).
  * Remove any `mdc-layout-grid-*` CSS that is now redundant.
* Refactor the layout CSS:
  * For viewports below `$bp-xlarge` (1280), use a `display: block`-based layout, such that the `.googlesitekit-idea-hub__idea--actions` wraps underneath the `.googlesitekit-idea-hub__idea--details`.
  * For viewports above `$bp-xlarge`, switch to a flex-based layout, such that the two divs are inline
[...]

Issue #3907, opened 2021-08-18
Show spinner when saving, unsaving, or dismissing an idea in Idea Hub widget


felixarntz:
There can be a delay when clicking one of the Idea Hub widget icon buttons to save, unsave, or dismiss an idea. This delay is expected due to sending the API request, however we should cater for this in the UI by showing a loading indicator.

See https://app.asana.com/0/1200491083500938/1200758448284506/f

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* When clicking on one of the icon buttons to "Save", "Dismiss" (both in the "New" tab) or "Unsave" (in the "Saved" tab) an idea, the respective icon should be replaced by a loading spinner until the action has been completed.
* Whenever that spinner is displayed, the "overlay" with the idea interaction buttons for that respective idea should be "permanently" displayed as long as the spinner is there / as long as the process/API request takes.
    * In other words, when hovering/focusing away from the idea while one of the interactions is in progress, the overlay shouldn't disappear like it usually does when there's _no_ action in progress. This can be accomplished for example by conditionally applying an extra CSS class on the overlay wrapper.

### Additional Acceptance criteria

See https://github.com/google/site-kit-wp/issues/3907#issuecomment-989000016.

* All 4 actions should have the loading spinner.
* All 4 actions should have a brief success state where they show a brief message that remains there for a brief period before the idea disappears (from the current tab) - basically this should work the same way it already does for creating a draft.
* The success messages should be:
    * _Idea saved_
    * _Idea dismissed_
    * _Idea removed from saved_
    * _Draft created_ (already exists)

## Implementation Brief

* Inside `assets/js/modules/idea-hub/datastore/constants.js`:
	* Add the following new activity constants: 
	- `IDEA_HUB_ACTIVITY_IS_DELETING`
	- `IDEA_HUB_ACTIVITY_IS_PINNING`
	- `IDEA_HUB_ACTIVITY_IS_UNPINNING`
	* Remove `IDEA_HUB_ACTIVITY_IS_PROCESSING`
* Inside `assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Idea.js`
	* Import the new constants, and remove the import of `IDEA_HUB_ACTIVITY_IS_PROCESSING`
[...]

PR #3920, opened 2021-08-23
Wrap icons when there is not enough space.

asvinb:
## Summary

Addresses issue #3839 

## Checklist


asvinb:
@johnPhillips I know you worked on the IB, but can you check my notes in the "Relevant technical choices" of this PR? 
Thanks!

johnPhillips:
Thanks @asvinb - you helped me a lot with the IB, so you are in a good position to evaluate the best approach here.

It sounds good to me, but I'll mention two things I noticed when sketching in the IB we arrived at.

1. When the idea content is verbose and the actions are *not* stacked, then the text has to wrap on hover, which is slightly janky:

Screenshot 2021-08-23 at 13 21 09

2. Although at 600 to 960px there does seem a lot of space, again bear in mind that there is no character limit on these ideas, so they can get quite long:

Screenshot 2021-08-23 at 13 20 33

So here we could have gone inline with the first two items, but the third might have looked a little strange.

In summary, I'd just suggest eyeballing your solution with some verbose ideas and seeing how it looks? Happy to go with whatever you decide 👍

asvinb:
Thanks @johnPhillips . Based on our conversation, I've tweaked the PR so that items are stacked when the viewport is < 1280px.
@aaemnnosttv @felixarntz Is there a UX reason why we are showing the icons upon hover? I'am thinking whether it'll be best to show the icons across all breakpoints, especially we have #3907 where we are replacing icons by the loading spinner? 🤔

felixarntz:
@asvinb 
> Is there a UX reason why we are showing the icons upon hover? I'am thinking whether it'll be best to show the icons across all breakpoints, especially we have #3907 where we are replacing icons by the loading spinner? 🤔 

I believe this was decided to not visually overload the user with too many action buttons. Any additional thoughts @marrrmarrr?

johnPhillips:
If it is the case that we don't want to visually overload the UI with buttons, that's reasonable. But I would point out that we are showing them all on smaller viewports where there is less space: 

Screenshot 2021-08-24 at 10 23 58

This is my first issue with using `:hover` to hide the buttons. My second is that it increases complexity for us, especially as the buttons are also sometimes inline and sometimes wrapped. For example:
- How do we treat the [spinner behaviour ](https://github.com/google/site-kit-wp/issues/3907) on a button that is sometimes hidden unless your hover over it?
- What about verbose idea text that will have to wrap when you hover and expose the buttons (see my comment above)?

Just wondering if there is a more consistent way to keep the buttons visible but to lessen their visual impact / reduce the clutter. Maybe take a page out of WordPress's book? On the Posts page they use checkboxes and a single row of buttons that allows batch operations. They are also dealing with potentially verbose text, so they put all the actions underneath and only expose them on hover:

Screenshot 2021-08-24 at 12 13 03

Batch operations are obviously a fair amount of work. But if we always wrapped the buttons and made them visible, it could solve some of the issues I alluded to above. We could also look at secondary colours for some of the buttons to mute them slightly...

| Verbose content, small viewport | Large viewport |
| --- | --- |
| Screenshot 2021-08-24 at 12 27 21 | Screenshot 2021-08-24 at 12 28 14 |
 
I'm no UX expert and I'm certainly not a designer, so these might be terrible ideas 😂. 🤷‍♂️. All I'm saying is that perhaps we can find a better way to solve the clutter concern across all viewports?

felixarntz:
@marrrmarrr @alex-Moulin Can you share your thoughts on @johnPhillips's feedback above? I tend to agree only showing the buttons on hover is causing issues here.

alex-moulin:
@felixarntz @marrrmarrr @johnPhillips I would be keen to discuss this as I don't know for sure the reasons behind some of the decisions that were taken to design these elements. However, my 2 cents would be that instead of showing all the icons by default, we could potentially show 3 vertical dots on hover, and when interacted with, it would display the 3 options: Draft, Pin, Dismiss
This would allow us to keep the interface clean on all viewports but also could be a pattern we use throughout the plugin.

If possible, we could use an animation to show what's happening with the element, i.e. Dismiss would move the element to the left out of the list, Pin would move it to the right (as going into the other tab) and Draft would magnify it to take the users to the other screen.

felixarntz:
Per https://github.com/google/site-kit-wp/issues/3839#issuecomment-910491145, we may actually want to circle back and go with the original PR approach here.

felixarntz (approved):
This actually looks good to me while testing! Paging @marrrmarrr for an additional approval.

marrrmarrr:
Just tested it too, looks good!

Issue #3294, opened 2021-05-05
Disable Google Analytics measurement per-property


aaemnnosttv:
## Feature Description

Initially mentioned in #2211 (but not limited to that issue), now that Site Kit is aware of Analytics tags in Tag Manager containers and those the property IDs must match, we should be able to be more targeted with our disabling of Analytics measurement for logged-in users by using the property-specific method rather than complete opt-out. This would be preferable because the current method will disable all Analytics measurement for all properties on the page, including Site Kit's usage tracking on the front end but also tracking by other plugins, etc. rather than only the property configured in the Analytics &/or GTM modules.

See [Disable Google Analytics measurement](https://developers.google.com/analytics/devguides/collection/gtagjs/user-opt-out)

For AMP, we may need to continue to use the same opt-out or simply change the output of the `amp-analytics` tag as it does not seem to support the same property-specific mechanism.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* For non-AMP, Site Kit should no longer use the "catch-all" Analytics opt-out snippet, but instead print a more granular opt-out snippet, which affects only the specific Analytics properties configured in Site Kit:
    * In case tracking should be disabled for the current user, the `Analytics` module should print the `window['ga-disable-GA_MEASUREMENT_ID'] = true;` snippet (see [docs](https://developers.google.com/analytics/devguides/collection/gtagjs/user-opt-out)) for the following measurement IDs:
        * The UA property ID.
        * The GA4 (web data stream) measurement ID, if one is set.
        * _If_ Tag Manager is active: The Tag Manager UA property ID (see Tag Manager's `gaPropertyID` setting), if one is set that differs from the UA property ID (currently not possible since the same one is enforced, but this will change in the future and the logic should already cater for it - no need to specifically QA it for now).
            * While this is related to the Tag Manager module, the Analytics opt-out is fully controlled by the Analytics module, so the logic should be kept together in here.
* For AMP, since a property-specific mechanism is not available, nothing should change from the current opt-out snippet.

## Implementation Brief

* In the `Analytics::print_tracking_opt_out` method:
  * Grab the current `accountID` and `propertyID` from the Analytics settings.
  * Replace `window["_gaUserPrefs"] = ...` code with `window['ga-disable-{$propertyID}'] = true;`.
  * Fire a new `googlesitekit_analytics_tracking_opt_out` action at the end of the method passing `propertyID` and `accountID` as action arguments.
[...]

PR #3923, opened 2021-08-23
Add analytics opt out updates.

ivankruchkoff:
## Summary

Addresses issue #3294

## Checklist

### DIFF
--- a/includes/Modules/Analytics.php
+++ b/includes/Modules/Analytics.php
@@ -1196,15 +1196,19 @@ final class Analytics extends Module
 		if ( ! $this->is_tracking_disabled() ) {
 			return;
 		}
+		$settings    = $this->get_settings()->get();
+		$account_id  = $settings['accountID'];
+		$property_id = $settings['propertyID'];
 
 		?>
 		<!-- <?php esc_html_e( 'Google Analytics user opt-out added via Site Kit by Google', 'google-site-kit' ); ?> -->
 		<?php if ( $this->context->is_amp() ) : ?>
 			<meta name="ga-opt-out" content="" id="__gaOptOutExtension">
 		<?php else : ?>
-			<script type="text/javascript">window["_gaUserPrefs"] = { ioo : function() { return true; } }</script>
+			<script type="text/javascript">window["ga-disable-<?php echo esc_attr( $property_id ); ?>"] = true; </script>
 		<?php endif; ?>
 		<?php
+		do_action( 'googlesitekit_analytics_tracking_opt_out', $property_id, $account_id );
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


includes/Modules/Analytics.php
```diff
@@ -1196,15 +1196,19 @@ private function print_tracking_opt_out() {
...
 		?>
 		<!-- <?php esc_html_e( 'Google Analytics user opt-out added via Site Kit by Google', 'google-site-kit' ); ?> -->
 		<?php if ( $this->context->is_amp() ) : ?>
 			<meta name="ga-opt-out" content="" id="__gaOptOutExtension">
 		<?php else : ?>
-			<script type="text/javascript">window["_gaUserPrefs"] = { ioo : function() { return true; } }</script>
+			<script type="text/javascript">window["ga-disable-<?php echo esc_attr( $property_id ); ?>"] = true; </script>
 		<?php endif; ?>
 		<?php
+		do_action( 'googlesitekit_analytics_tracking_opt_out', $property_id, $account_id );
```
aaemnnosttv, line 1211: This action should go into the web-specific `else` above. This was mentioned in the IB but this specific detail was lost. Otherwise we risk emitting invalid AMP markup in that case. This action is web-specific though as there is no way to do this property-specific opt-out in AMP.

includes/Modules/Analytics_4.php
```diff
@@ -247,6 +248,20 @@ private function create_webdatastream( $property_id ) {
 		);
 	}
 
+	/**
+	 * Handles analytics tracking opt-out.
```
aaemnnosttv, line 252:
```suggestion
	 * Handles Analytics measurement opt-out for a GA4 property.
```

includes/Modules/Analytics_4.php
```diff
@@ -247,6 +248,20 @@ private function create_webdatastream( $property_id ) {
...
+	 *
+	 * @since n.e.x.t
+	 */
+	private function analytics_tracking_opt_out() {
+		$settings       = $this->get_settings()->get();
+		$measurement_id = $settings['measurementID'];
+		?>
```
aaemnnosttv, line 259: We should also return early here if no `measurementID` is set.

includes/Modules/Tag_Manager.php
```diff
@@ -643,6 +644,20 @@ private function can_analytics_use_snippet( $original_value ) {
...
 
+	/**
+	 * Handles analytics tracking opt-out.
+	 *
+	 * @since n.e.x.t
+	 */
+	private function analytics_tracking_opt_out() {
+		$settings       = $this->get_settings()->get();
+		$ga_property_id = $settings['gaPropertyID'];
+		?>
```
aaemnnosttv, line 655: Here we should return early if no `gaPropertyID` is set, but also if the given `$propertyID` (from param) matches the `gaPropertyID` to not add a duplicate opt-out script.

includes/Modules/Tag_Manager.php
```diff
@@ -643,6 +644,20 @@ private function can_analytics_use_snippet( $original_value ) {
 		return $original_value;
 	}
 
+	/**
+	 * Handles analytics tracking opt-out.
```
aaemnnosttv, line 648:
```suggestion
	 * Handles Analytics measurement opt-out for the configured Analytics property in the container(s).
```

aaemnnosttv (changes requested):
Thanks @ivankruchkoff ! Just a few small details to address and then this should be good to go 👍

PR #3925, opened 2021-08-23
Show helper text when supplied to survey component

tofumatt:
## Summary

Addresses issue #3762.

It turns out we _had_ a subtitle in here but it was being hidden with an odd `opacity: 0` rule. This removes it and makes sure the `HelperText` component isn't rendered _if_ there isn't a `subtitle` prop.

## Checklist

### DIFF
--- a/assets/js/components/surveys/SurveyQuestionOpenText.js
+++ b/assets/js/components/surveys/SurveyQuestionOpenText.js
@@ -74,7 +74,7 @@ const SurveyQuestionOpenText = ( {
 					name={ `survey-opentext-${ instanceID }` }
 					helperText={
 						subtitle ? (
-							<HelperText>{ subtitle }</HelperText>
+							<HelperText persistent>{ subtitle }</HelperText>
 						) : undefined
 					}
 					onChange={ onChange }
--- a/assets/js/components/surveys/SurveyQuestionOpenText.js
+++ b/assets/js/components/surveys/SurveyQuestionOpenText.js
@@ -71,8 +71,12 @@ const SurveyQuestionOpenText = ( {
 					<label htmlFor={ instanceID }>{ placeholder }</label>
 				</VisuallyHidden>
 				<TextField
-					name="open-text"
-					helperText={ <HelperText>{ subtitle }</HelperText> }
+					name={ `survey-opentext-${ instanceID }` }
+					helperText={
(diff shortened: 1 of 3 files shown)


assets/sass/components/surveys/_googlesitekit-surveys.scss
```diff
@@ -57,6 +60,10 @@
 		.mdc-text-field-helper-line {
 			padding-left: 0;
 		}
+
+		.mdc-text-field-helper-text {
+			opacity: 1;
+		}
```
aaemnnosttv, line 66: This fixes the problem but not quite in the right place.

It turns out the helper text is designed to be conditionally shown and likely uses `opacity` to avoid a layout shift.

All we need to do is pass the `persistent` prop to `HelperText` to indicate that the text is to always be shown as the component is made to also support input validation.

See https://github.com/material-components/material-components-web-react/blob/master/packages/text-field/helper-text/README.md#props

You can also see this work on `develop` by toggling the prop in React devtools.

assets/js/components/surveys/SurveyQuestionOpenText.js
```diff
@@ -71,8 +71,12 @@ const SurveyQuestionOpenText = ( {
...
 				</VisuallyHidden>
 				<TextField
-					name="open-text"
-					helperText={ <HelperText>{ subtitle }</HelperText> }
+					name={ `survey-opentext-${ instanceID }` }
+					helperText={
+						subtitle ? (
+							<HelperText>{ subtitle }</HelperText>
+						) : undefined
+					}
```
aaemnnosttv, line 79:
```suggestion
					helperText={
						subtitle ? (
							<HelperText persistent>{ subtitle }</HelperText>
						) : undefined
					}
```

aaemnnosttv (changes requested):
Thanks @tofumatt – we don't actually need to change any styles here (as these come from the component). Instead we're just missing a prop 😄

assets/sass/components/surveys/_googlesitekit-surveys.scss
tofumatt: Huh, whoops, thanks! 👍🏻

aaemnnosttv (approved):
Great, thanks!

Issue #1194, opened 2020-03-02
Add link to Settings page in plugin action links


aaemnnosttv:
## Feature Description

Many plugins add a link to their settings page to the plugin row itself for convenience. 

![image](https://user-images.githubusercontent.com/1621608/75675907-cfd35100-5c90-11ea-9128-b27d91d6e1c0.png)

We should do the same to improve the discoverability of plugin settings and generally feel more consistent with other well-known plugins.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* In the WordPress plugins list table, the Site Kit plugin should have a "Settings" link (to the `googlesitekit-settings` admin screen) as first item in the links below (e.g. before "Deactivate") _if_ the current user has access to the screen (`Permissions::MANAGE_OPTIONS`).

## Implementation Brief

* Add a new class similar to `\Google\Site_Kit\Core\Admin\Plugin_Row_Meta` in the same folder called `Plugin_Action_Links`
* Add a `register` method with a `plugin_action_links` filter to the new class where the name of the filter is generated like so:
  `'plugin_action_links_' . GOOGLESITEKIT_PLUGIN_BASENAME`
* Add a closure function to the filter which:
  * Receives `$links` array as a parameter
  * Generates  a new settings url using `Context::admin_url` 
  * Adds the url to a link with 'Settings' text which is translatable
  * Adds the new settings link to the `$links` array using `array_unshift` so that the settings link is returned before the deactivate link
  * Returns the `$links` array
  * Add a conditional to check if the user has the ability to view the settings page (`Permissions::MANAGE_OPTIONS`) before the link is created and returned
* Call the method in the same way as `Plugin_Row_Meta` in the `register` method of `Plugin.php` 

## QA Brief

* The Settings Link will be shown on plugin once the plugin activated and initial setup is complete.  

## Changelog entry

* Add link to Settings page in plugin action links on plugins page.

felixarntz:
@aaemnnosttv Potentially a link to the "Dashboard" (or splash if not connected) would be more suitable here? While most plugins have a "Settings" link I'd argue this is largely because for these plugins their "Settings" is the central entry point to the overall functionality. For us however, that is the "Dashboard".

I'd say this is mostly for discoverability, rather than providing a link to the plugin's settings. Also note that if we linked to "Settings" it would only appear sometimes because the menu is only available if you're an admin and already connected.

What do you think?

aaemnnosttv:
[...]

PR #3927, opened 2021-08-24
Add Settings page to plugin action links.

kuasha420:
## Summary

Addresses issue #1194

## Checklist

### DIFF
--- a/includes/Core/Admin/Plugin_Action_Links.php
+++ b/includes/Core/Admin/Plugin_Action_Links.php
@@ -1,0 +1,71 @@
+<?php
+/**
+ * Class Google\Site_Kit\Core\Admin\Plugin_Action_Links
+ *
+ * @package   Google\Site_Kit
+ * @copyright 2021 Google LLC
+ * @license   https://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
+ * @link      https://sitekit.withgoogle.com
+ */
+
+namespace Google\Site_Kit\Core\Admin;
+
+use Google\Site_Kit\Context;
+use Google\Site_Kit\Core\Permissions\Permissions;
+
+/**
+ * Class for managing plugin action links.
+ *
+ * @since 1.41.0
+ * @access private
+ * @ignore
+ */
+class Plugin_Action_Links {
+
+	/**
+	 * Plugin context.
+	 *
+	 * @since 1.41.0
+	 * @var Context
+	 */
+	private $context;
+
+	/**
+	 * Constructor.
+	 *
+	 * @since 1.41.0
+	 *
+	 * @param Context $context      Plugin context.
+	 */
+	public function __construct(
+		Context $context
+	) {
+		$this->context = $context;
+	}
+
+	/**
+	 * Registers functionality through WordPress hooks.
+	 *
+	 * @since 1.24.0
+	 */
+	public function register() {
+		add_filter(
+			'plugin_action_links_' . GOOGLESITEKIT_PLUGIN_BASENAME,
+			function ( $links ) {
+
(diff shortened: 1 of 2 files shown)


includes/Core/Admin/Plugin_Action_Links.php
```diff
@@ -0,0 +1,71 @@
...
+
+namespace Google\Site_Kit\Core\Admin;
+
+use Google\Site_Kit\Context;
+use Google\Site_Kit\Core\Permissions\Permissions;
+
+/**
+ * Class for managing plugin action links.
+ *
+ * @since 1.41.0
```
aaemnnosttv, line 19: Since tags should always be added using the literal `n.e.x.t` placeholder which is then updated as part of the preparation for a release so you don't need to think about what release it might go in.

includes/Core/Admin/Plugin_Action_Links.php
```diff
@@ -0,0 +1,71 @@
...
+	 */
+	public function register() {
+		add_filter(
+			'plugin_action_links_' . GOOGLESITEKIT_PLUGIN_BASENAME,
+			function ( $links ) {
+
+				if ( current_user_can( Permissions::MANAGE_OPTIONS ) ) {
+					$settings_url = $this->context->admin_url( 'settings' );
+
+					$settings_link = '<a href="' . $settings_url . '">' . __( 'Settings', 'default' ) . '</a>';
```
aaemnnosttv, line 59: A few things here, but most importantly, we always* need to use the `google-site-kit` text domain for our translations. There are a few exceptions to this, but that's the general rule (our ESLint config has better checks for this). While it would probably be safe to use the core translation here, I'm not sure if that is the intention so I have to assume it is not.

In this case, let's use the same as it is defined for the menu which uses our own text domain 👍  https://github.com/google/site-kit-wp/blob/7ff89bbf73b310573a125887b80469941bbce438/includes/Core/Admin/Screens.php#L327

Second, we generally want to use WP escaping functions on any dynamic parts that we output. PHPCS will error if this is not done when outputting HTML, but I think it's not complaining here due to just building an HTML string. Let's use `esc_url` on the URL and for the translatable string, WP has a `esc_html__` version of `__` which escapes any HTML from the translated output (which is just a good practice if we don't expect there to be any).

Finally, it would be a little cleaner to use `sprintf` here to build the string rather than many concatenations but nothing wrong with that. This does make it a bit easier to do something with each part when filling.

includes/Core/Admin/Plugin_Action_Links.php
```diff
@@ -0,0 +1,71 @@
...
+	public function __construct(
+		Context $context
+	) {
+		$this->context = $context;
+	}
+
+	/**
+	 * Registers functionality through WordPress hooks.
+	 *
+	 * @since 1.24.0
```
aaemnnosttv, line 49: Always make sure that any parts you copy from another file are updated 😄

includes/Core/Admin/Plugin_Action_Links.php
```diff
@@ -0,0 +1,71 @@
...
+
+					$settings_link = '<a href="' . $settings_url . '">' . __( 'Settings', 'default' ) . '</a>';
+
+					array_unshift( $links, $settings_link );
+				};
+
+				return $links;
+			},
+			10,
+			2
```
aaemnnosttv, line 67: These two can actually be left off because we only need the first element passed to the filter (`$links`) which is the default. We also don't need a special priority so we omit them when not needed.

aaemnnosttv (changes requested):
Thanks @kuasha420 – Nice work, this looks good overall and works as intended, there are just a few details to address and then this will be good to go 👍

aaemnnosttv (approved):
Great, thanks @kuasha420 👍

PR #3930, opened 2021-08-24
Always render the same idea hub Footer component regardless of footerText.

aaemnnosttv:
## Summary

Addresses issue #3865 (follow-up)

This PR updates the Idea Hub widget footer to be hidden in zero states in a way that preserves the height of the footer but keeps it invisible. The tradeoff here is that the zero state may appear somewhat uncentered in the widget but it isn't very obvious and better than the layout shift so I think it's acceptable.

Using zero-state for adjacent Search Console widget to ensure its height does not interfere with Idea Hub widget, to show that the height is preserved across tabs by the widget itself.

https://user-images.githubusercontent.com/1621608/130656920-9f1c6489-f43f-4782-9d68-e193ed4fbc39.mp4

https://user-images.githubusercontent.com/1621608/130656930-143f4ff5-d6e4-4076-9fcb-cdd21938aa40.mp4

## Checklist

### DIFF
--- a/assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Footer.js
+++ b/assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Footer.js
@@ -28,14 +28,6 @@ import { Grid, Cell, Row } from '../../../../../material-components';
 import Pagination from './Pagination';
 
 export default function Footer( { tab, footerText } ) {
-	if ( ! footerText ) {
-		return (
-			<div className="googlesitekit-idea-hub__footer">
-				<Pagination tab={ tab } />
-			</div>
-		);
-	}
-
 	return (
 		<Grid className="googlesitekit-idea-hub__footer">
 			<Row>
--- a/assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
+++ b/assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
@@ -19,6 +19,7 @@
 /**
  * External dependencies
  */
+import classnames from 'classnames';
 import PropTypes from 'prop-types';
 import Tab from '@material/react-tab';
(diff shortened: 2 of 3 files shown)


assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Footer.js
```diff
@@ -28,14 +28,6 @@ import { Grid, Cell, Row } from '../../../../../material-components';
 import Pagination from './Pagination';
 
 export default function Footer( { tab, footerText } ) {
-	if ( ! footerText ) {
-		return (
-			<div className="googlesitekit-idea-hub__footer">
-				<Pagination tab={ tab } />
-			</div>
-		);
-	}
```
felixarntz, line 37: Kind of a nit-pick, but is there no way to solve this while still not rendering an empty div if `footerText` is empty? We could still use the `Row` to maintain the same layout but then only render the `Cell` for the pagination - I think it's possible to use specific classes to horizontally offset?

felixarntz (approved):
Great work, visually hiding it makes sense to achieve the desired UX, even though it may seem odd from a development perspective. Had just some minor feedback here, but nothing blocking.

tofumatt (dismissed):
Looks good to me, but there's one failing test that needs fixing before we merge 👍🏻

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Footer.js
aaemnnosttv: @felixarntz There probably is a way, but I don't think it's worth the effort to avoid rendering a single cell with empty content. It sounds like a micro optimization for this specific case, where it might even be more performant to keep the structure there for a smaller update when switching tabs (when there is `footerText`). It's simpler this way so I don't think we should worry about it as it is now.

Issue #3908, opened 2021-08-18
Update Idea Hub module description with more clear message


felixarntz:
The Idea Hub module description should be clarified that the suggestions come from **user-invoked** search queries.

See https://app.asana.com/0/1200491083500938/1200755649650592/f

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The Idea Hub module description should be updated to: _Idea Hub suggests what you can write about next, from actual questions people asked on Google Search_ 
* The Idea Hub widget description on the main dashboard should be updated to: _Ideas to write about, from actual questions people asked on Search_
* The hard-coded configuration in Storybook re-including the description should be updated accordingly as well. Also, that Storybook configuration includes a homepage URL which is actually incorrect since Idea Hub does not have a homepage to point to, so that entry should be removed / emptied.

## Implementation Brief

* Inside `includes/Modules/Idea_Hub.php`, change the module description text inside the `setup_info` method to be:
  _Idea Hub suggests what you can write about next, from actual questions people asked on Google Search_
* In `assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js`, update the heading text to read:
  _Ideas to write about, from actual questions people asked on Search_
  * You will probably need to add some simple CSS for larger viewports to get the `Experimental` badge to display nicely with the heading, for example by adding `display: inline-flex` to `.googlesitekit-idea-hub__title` (see images below).
* Update `assets/js/modules/idea-hub/components/setup/Setup.stories.js` as per the ACs:
  * Update the hard-coded description to match the module description above
  * Remove the `homepage` value 
 

#### Experimental badge wraps on larger viewports 
![Screenshot 2021-08-26 at 17 39 59](https://user-images.githubusercontent.com/2470911/131002708-7eb8d167-25d3-405a-a3f0-11f8b06d2be3.png)

#### Example solution with `inline-flex`
![Screenshot 2021-08-26 at 17 40 10](https://user-images.githubusercontent.com/2470911/131002709-0c249c9c-794f-4eda-a66b-3a0c34d08931.png) 

### Test Coverage

* No new test coverage required.

### Visual Regression Changes

* Some images may need updating to reflect the changes in the UI.

## QA Brief

[...]

PR #3956, opened 2021-08-28
Update Idea Hub module description with more clear message (3908).

johnPhillips:
## Summary

Addresses issue #3908

## Relevant technical choices

Although in the IB I mentioned that it might be a good idea to check the styles on the 'Experimental' badge across different viewports, I found that although the heading text and badge wrap somewhat awkwardly here and there, actually there isn't a particular viewport that can be targeted to make a uniform fix.

## Checklist

### DIFF
--- a/assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
+++ b/assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
@@ -236,7 +236,7 @@ function DashboardIdeasWidget( props ) {
 				<div className="googlesitekit-idea-hub__header">
 					<h3 className="googlesitekit-idea-hub__title">
 						{ __(
-							'Ideas to write about based on unanswered searches',
+							'Ideas to write about, from actual questions people asked on Search',
 							'google-site-kit'
 						) }
 
--- a/assets/js/modules/idea-hub/components/setup/Setup.stories.js
+++ b/assets/js/modules/idea-hub/components/setup/Setup.stories.js
@@ -52,8 +52,7 @@ DefaultSetup.decorators = [
 				slug: 'idea-hub',
 				name: 'Idea Hub',
 				description:
-					"Idea Hub suggests what you can write about next, based on searches that haven't been answered yet",
-				homepage: 'https://www.google.com/webmasters/verification/home',
+					'Idea Hub suggests what you can write about next, from actual questions people asked on Google Search',
 				internal: true,
 				order: 0,
 				active: false,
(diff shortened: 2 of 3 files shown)


eugene-manuilov (changes requested):
Thanks, @johnPhillips. This looks good but misses the following part from IB:

```
You will probably need to add some simple CSS for larger viewports to get the Experimental badge to display nicely with the heading, for example by adding display: inline-flex to .googlesitekit-idea-hub__title (see images below).
```

johnPhillips:
@eugene-manuilov See the `relevant technical choice` section.

A bit more context:

When I wrote the IB I noticed that occasionally, on some screen widths, the text would wrap and that meant the `Experimental` label was orphaned on a new line which looked not-quite-optimal. So I noted that some CSS might be required.

However, in development I had a more thorough look at the various screen widths and wasn't able to identify a consistent way to stop this, as it isn't a mobile / tablet / desktop issue. It can happen at any breakpoint if the width of the container is sufficient to cause the badge to wrap. So the options were:
- use `display: inline-flex` on all viewports (looks a little bit strange on mobile)
- write some extremely specific, whack-a-mole media queries to target the few precise widths where the badge wraps (at a given font size etc)
- do nothing and accept that occasionally the badge will wrap.

That was the logic. I'm very happy to revisit this, but didn't want to spend too long on it as it felt quite marginal.

Let me know what you think?

eugene-manuilov:
@johnPhillips if we leave it as is, it will continue looking ugly on smaller screens. For example:



We can fix it by using `display: flex; justify-content: space-between; align-items: flex-start;` (like you said) styles for the `<h3>` title and it will look much better:

johnPhillips:
@eugene-manuilov I'm happy to add that - just to flag again that it isn't really about device width. 

For example, on very wide desktops (~1400px), we also see the issue. Adding the fix can cause the last word of the heading to be orphaned in the same way the badge was, so it still looks slightly strange:

| As-is | with `display: inline-flex` |
| --- | --- | 
| Screenshot 2021-08-31 at 17 03 08 | Screenshot 2021-08-31 at 17 03 21 | 

On smaller viewports (~375px), it looks fine because the badge is inline and the text wraps. However, adding the fix makes for a slightly strange 2 column layout:

| As-is | with `display: inline-flex` |
| --- | --- | 
| Screenshot 2021-08-31 at 17 02 25 | Screenshot 2021-08-31 at 17 02 05 |

I did spend a bit of time looking at this on Friday and couldn't find a simple way to make it look good every time, so just wanted to make my thought process clear. 

Again, very happy to add the rule - your call 👍

eugene-manuilov (approved):
Ok, I think I am going to approve it without CSS changes because this is out-of-scope for this ticket. Thanks, @johnPhillips.

eugene-manuilov:
We can address styles in a separate ticket in the future.

Issue #3070, opened 2021-04-01
Changing the title of a page or post results in that URL appearing twice in the "Most Popular Content" table


jamesozzie:
## Bug Description

When making changes such as the following this results in the "Most popular content" table listing duplicate content:
- Post or page title (without any slug or permalink changes)
- Site Title changes (Settings > General)
- Slug changes

The changes above reset the statistics for that content, with that same URL appearing twice in the "Most Popular Content" table - Once with total views based on the previous title and again with the new title. 

Any changes to a sites title will result in the homepage being listed twice

![image](https://user-images.githubusercontent.com/41326532/113275967-3df4a800-92d7-11eb-952b-c4bac42f78fb.png)

## Steps to reproduce

1. Change the title or a page or post
2. Don't make any slug changes
3. Revisit the most popular content table after traffic to those pages
4. One occurrence will appear prior to the URL change, another occurrence for after the title change

## Additional Context

 - Duplicate references do not occur on Analytics directly
 - Separately any slug changes will also result in duplicate occurrences in the same table (Tested when the post is part of the permalink structure)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All _refactored_ Analytics widgets which display most popular/most earning titles and URLs (e.g. in WP dashboard widget, Popular Pages widget and Top Earning Pages widgets on SK dashboard, the same widgets on Analytics/AdSense module page) should be modified so that they don't consider a URL multiple records just because the URL's title may have changed over time. This is currently resulting in:
    * slightly incorrect numbers compared to the Analytics frontend (e.g. for one title the URL has 1000 views, while for the other title it has 4, which means in Site Kit probably only the first record makes it to the list, showing 1000, while the Analytics frontend shows 1004)
    * duplicate results, particularly on sites with low traffic (e.g. for one title the URL has 5 views, while for the other title it has 4, which means in Site Kit probably both records show in the list, but it's the same URL, while in Analytics frontend it's just one record showing 9 views)
* The main technical problem here is that what we actually want is some sort of Analytics metric like `ga:latestPageTitle`, which however doesn't exist. There is no way to get the titles from Analytics without also separating it out as a dimension - but we need it as a metric. Therefore we need a workaround, as follows:
    * Change the primary Analytics API request that is used to only query by the `ga:pagePath` dimension, instead of also including the `ga:pageTitle` dimension. This ensures that results will be grouped purely by the URL/path.
[...]

PR #3958, opened 2021-08-30
Feature/3070 most popular content table

ivankruchkoff:
## Summary

Addresses issue #3070

## Checklist


assets/js/components/wp-dashboard/WPDashboardPopularPages.js
```diff
@@ -16,6 +16,11 @@
  * limitations under the License.
  */
 
+/**
+ * External dependencies
+ */
+import { cloneDeep } from 'lodash';
```
eugene-manuilov, line 22:
```suggestion
import cloneDeep from 'lodash/cloneDeep';
```

assets/js/components/wp-dashboard/WPDashboardPopularPages.js
```diff
@@ -39,45 +44,69 @@ import { numFmt } from '../../util';
...
-		dimensions: [ 'ga:pageTitle', 'ga:pagePath' ],
-		orderby: [
-			{
-				fieldName: 'ga:pageviews',
-				sortOrder: 'DESCENDING',
-			},
-		],
-		limit: 5,
-	};
+		let pagePaths = [];
```
eugene-manuilov, line 81: This variable is not needed in this scope. Please, move it to the scope where this variable is used. The same applies to other components, please, update this variable everywehere.

assets/js/components/wp-dashboard/WPDashboardPopularPages.js
```diff
@@ -39,45 +44,69 @@ import { numFmt } from '../../util';
...
+		let hasLoadedPageTitles = true;
+		if ( reportLoaded ) {
+			const { startDate, endDate } = dateRangeDates;
+			const pageTitlesArgs = {
+				startDate,
+				endDate,
+			};
```
eugene-manuilov, line 88: We can use destruction here to make it a bit more concise:

```suggestion
			const pageTitlesArgs = {
				...dateRangeDates,
			};
```

Furthermore, we can re-struct the code in this section to avoid defining the `pageTitlesArgs` variable at all:

```js
( data.report?.[ 0 ]?.data?.rows || [] ).forEach(
    ...
);

data.titles = select( MODULES_ANALYTICS ).getPageTitles( {
    ...dateRangeDates,
    pagePaths,
} );
```

assets/js/modules/adsense/components/module/ModuleTopEarningPagesWidget/index.js
```diff
@@ -20,6 +20,7 @@
  * External dependencies
  */
 import PropTypes from 'prop-types';
+import { cloneDeep } from 'lodash';
```
eugene-manuilov, line 23:
```suggestion
import cloneDeep from 'lodash/cloneDeep';
```

assets/js/modules/analytics/components/dashboard/DashboardPopularPagesWidget.js
```diff
@@ -16,6 +16,11 @@
  * limitations under the License.
  */
 
+/**
+ * External dependencies
+ */
+import { cloneDeep } from 'lodash';
```
eugene-manuilov, line 22:
```suggestion
import cloneDeep from 'lodash/cloneDeep';
```

assets/js/modules/analytics/components/module/ModulePopularPagesWidget/index.js
```diff
@@ -20,6 +20,7 @@
  * External dependencies
  */
 import PropTypes from 'prop-types';
+import { cloneDeep } from 'lodash';
```
eugene-manuilov, line 23:
```suggestion
import cloneDeep from 'lodash/cloneDeep';
```

assets/js/components/wp-dashboard/WPDashboardPopularPages.js
```diff
@@ -39,45 +44,69 @@ import { numFmt } from '../../util';
...
+			( data.report?.[ 0 ]?.data?.rows || [] ).forEach(
+				( { dimensions } ) => {
+					pagePaths = pagePaths.concat(
+						dimensions.filter(
+							( url ) => ! pagePaths.includes( url )
+						)
+					);
+				}
+			);
```
eugene-manuilov, line 97: This is not safe to rely on the `dimensions` property as on an array with URLs only. In our case, only the first item in the dimensions array is interesting for us. Please, rework this loop to use the `dimensions[0]` element only. The same applies to other components as well, please, update them all.

eugene-manuilov (changes requested):
Thanks, @ivankruchkoff. Found a few issues to address. Could you please take a look?

assets/js/components/wp-dashboard/WPDashboardPopularPages.js
ivankruchkoff: Updated in https://github.com/google/site-kit-wp/pull/3958/commits/a72c239078c2ab4bf187dea1bc7b3ed4ce87d9d4

assets/js/components/wp-dashboard/WPDashboardPopularPages.js
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3958/commits/a52d5da689077a9cbc62b42c918026c5bc15e3bd

assets/js/modules/analytics/datastore/report.test.js
```diff
@@ -273,5 +273,62 @@ describe( 'modules/analytics report', () => {
...
+
+				const pageTitlesReport = registry
+					.select( MODULES_ANALYTICS )
+					.getReport( pageTitlesArgs );
+
+				expect( pageTitlesReport ).toEqual( fixtures.pageTitles );
+
+				registry
+					.select( MODULES_ANALYTICS )
+					.getPageTitles( { pagePaths, startDate, endDate } );
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3958/commits/fd6e765244b4f9412f081798a122859ee50a9e46

assets/js/components/wp-dashboard/WPDashboardPopularPages.js
ivankruchkoff: `dateRangeDates` contains other properties that we don't need, so the best we can do is pull out the two args `startDate` `endDate` and omit creation of `pageTitlesArgs`

ivankruchkoff: https://github.com/google/site-kit-wp/pull/3958/commits/04f7489d6dead3d6306483f17adc53dfffd1e66e

assets/js/components/wp-dashboard/WPDashboardPopularPages.js
```diff
@@ -39,45 +44,66 @@ import { numFmt } from '../../util';
...
-		dimensions: [ 'ga:pageTitle', 'ga:pagePath' ],
-		orderby: [
-			{
-				fieldName: 'ga:pageviews',
-				sortOrder: 'DESCENDING',
-			},
-		],
-		limit: 5,
-	};
+		let hasLoadedPageTitles = true;
```
eugene-manuilov, line 81: This should be `false` by default, shouldn't it?

assets/js/modules/adsense/components/module/ModuleTopEarningPagesWidget/index.js
```diff
@@ -70,15 +71,46 @@ function ModuleTopEarningPagesWidget( {
 				limit: 10,
 			};
 
+			const report = select( MODULES_ANALYTICS ).getReport( reportArgs );
+
+			let pageTitles;
+
+			let hasLoadedPageTitles = true;
```
eugene-manuilov, line 78: Same here. Should it be `false` by default?

assets/js/modules/analytics/components/dashboard/DashboardPopularPagesWidget.js
```diff
@@ -78,6 +83,34 @@ function DashboardPopularPagesWidget( {
 				limit: 10,
 			};
 
+			const report = store.getReport( args );
+			let pageTitles;
+
+			let hasLoadedPageTitles = true;
```
eugene-manuilov, line 89: And here, `false` by default?

assets/js/modules/analytics/components/module/ModulePopularPagesWidget/index.js
```diff
@@ -83,17 +84,41 @@ export default function ModulePopularPagesWidget( {
...
+			error: select( MODULES_ANALYTICS ).getErrorForSelector(
+				'getReport',
+				[ args ]
+			),
+			titles: undefined,
+		};
+		const reportLoaded = select(
+			MODULES_ANALYTICS
+		).hasFinishedResolution( 'getReport', [ args ] );
+		let hasLoadedPageTitles = true;
```
eugene-manuilov, line 100: And here, `false` by default?

eugene-manuilov (changes requested):
Thanks, @ivankruchkoff. This looks much better. Left one comment in four components. Please, take a look.

assets/js/modules/analytics/components/module/ModulePopularPagesWidget/index.js
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3958/commits/3cddcca994d8eb6a2e939ac3aca4998081ba16f2

assets/js/modules/analytics/components/dashboard/DashboardPopularPagesWidget.js
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3958/commits/3cddcca994d8eb6a2e939ac3aca4998081ba16f2

assets/js/modules/adsense/components/module/ModuleTopEarningPagesWidget/index.js
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3958/commits/3cddcca994d8eb6a2e939ac3aca4998081ba16f2

assets/js/components/wp-dashboard/WPDashboardPopularPages.js
ivankruchkoff: Updated in https://github.com/google/site-kit-wp/pull/3958/commits/3cddcca994d8eb6a2e939ac3aca4998081ba16f2

Though it could go either way, since on an initial load, there is no report, and with no report there are no pageTitles to load. Probably makes more sense for `false` though.

assets/js/modules/adsense/components/module/ModuleTopEarningPagesWidget/index.js
```diff
@@ -70,15 +71,46 @@ function ModuleTopEarningPagesWidget( {
...
+				} );
+				hasLoadedPageTitles = undefined !== pageTitles;
+			}
+
+			const hasLoaded =
+				hasLoadedPageTitles &&
+				select( MODULES_ANALYTICS ).hasFinishedResolution(
+					'getReport',
+					[ reportArgs ]
+				);
```
felixarntz, line 104: @eugene-manuilov @ivankruchkoff I'm concerned about the code complexity this implementation is introducing here throughout all these widgets, which is something I was aiming to specifically avoid when we scoped out the IB. I don't think we should require all widgets that require page titles to add all this complicated logic within the `useSelect`.

Could we abstract this somehow so that the complexity goes into a helper function, or maybe even directly into `getPageTitles`? How about this idea:
* Change `getPageTitles` to the following signature `getPageTitles( report, options )` where it would simply receive the `report` (or `undefined`) and the `options` (i.e. report arguments) from where it could get information like the `startDate` and `endDate`.
* If `report` is `undefined`, return `undefined` for the page titles as well.
* If `report` is set/loaded, extract the URLs (we don't need to pass the index within `dimensions` for it, see my other comment) from the `report` and the `startDate` and `endDate` from `options` and run the extra request.

That should simplify the implementation within each widget drastically, something like:
```js
const report = select( MODULES_ANALYTICS ).getReport( reportArgs );
const pageTitles = select( MODULES_ANALYTICS ).getPageTitles( report, reportArgs );
```

assets/js/modules/adsense/components/module/ModuleTopEarningPagesWidget/index.js
```diff
@@ -70,15 +71,46 @@ function ModuleTopEarningPagesWidget( {
...
+			let pageTitles;
+
+			let hasLoadedPageTitles = false;
+			if ( undefined !== report ) {
+				const pagePaths = [];
+				( report?.[ 0 ]?.data?.rows || [] ).forEach(
+					( { dimensions } ) => {
+						if (
+							dimensions.length &&
+							! pagePaths.includes( dimensions[ 0 ] )
```
felixarntz, line 85: Related to my other comment: The API response includes information about the column headers, i.e. under which `dimensions` index the `ga:pagePath` is stored, so we don't need to have all of this code hard-coded / repeated per widget component.

felixarntz (changes requested):
This looks good against the IB, but when looking at it, I think that IB is in parts not what we should go for - that approach introduces a ton of complex code within each widget. Complexity should be abstracted away into reusable code, rather than requiring every widget that needs page titles to deal with it.

assets/js/components/wp-dashboard/WPDashboardPopularPages.js
```diff
@@ -39,45 +44,51 @@ import { numFmt } from '../../util';
...
-				sortOrder: 'DESCENDING',
-			},
-		],
-		limit: 5,
-	};
+		const { startDate, endDate } = dateRangeDates;
+		data.titles = select( MODULES_ANALYTICS ).getPageTitles( data.report, {
+			startDate,
+			endDate,
+		} );
```
felixarntz, line 85: Why not pass the entire `reportArgs` as second parameter? It's technically unnecessary right now, but in a way more straightforward than manually passing individual pieces of information. The idea is that in certain cases other report arguments than `startDate` and `endDate` might be relevant for `getPageTitles` as well. For example, if a report argument was some flag to include certain results in the response that otherwise would be excluded by default, we would need to set such a flag in the `getPageTitles` request too.

Not strictly required at this point, but due to the above it conceptually makes sense to pass the entire `reportArgs`.

assets/js/modules/adsense/components/module/ModuleTopEarningPagesWidget/index.js
```diff
@@ -70,15 +71,33 @@ function ModuleTopEarningPagesWidget( {
...
 
+			const report = select( MODULES_ANALYTICS ).getReport( reportArgs );
+
+			const pageTitles = select( MODULES_ANALYTICS ).getPageTitles(
+				report,
+				{
+					startDate,
+					endDate,
+				}
+			);
```
felixarntz, line 82: See above, I suggest we simplify this to:
```suggestion
			const pageTitles = select( MODULES_ANALYTICS ).getPageTitles( report, reportArgs );
```

assets/js/modules/analytics/components/dashboard/DashboardPopularPagesWidget.js
```diff
@@ -78,6 +83,18 @@ function DashboardPopularPagesWidget( {
 				limit: 10,
 			};
 
+			const report = store.getReport( args );
+
+			const pageTitles = store.getPageTitles( report, {
+				startDate,
+				endDate,
+			} );
```
felixarntz, line 91: See above:
```suggestion
			const pageTitles = store.getPageTitles( report, args );
```

assets/js/modules/analytics/components/module/ModulePopularPagesWidget/index.js
```diff
@@ -83,17 +84,25 @@ export default function ModulePopularPagesWidget( {
...
+				[ args ]
+			),
+		};
+		const reportLoaded = select(
+			MODULES_ANALYTICS
+		).hasFinishedResolution( 'getReport', [ args ] );
+
+		data.titles = select( MODULES_ANALYTICS ).getPageTitles( data.report, {
+			...dates,
+		} );
```
felixarntz, line 101: See above:
```suggestion
		data.titles = select( MODULES_ANALYTICS ).getPageTitles( data.report, args );
```

assets/js/modules/analytics/datastore/report.js
```diff
@@ -183,6 +184,78 @@ const baseSelectors = {
...
+					);
+					( data?.rows || [] ).forEach( ( { dimensions } ) => {
+						if (
+							! pagePaths.includes( dimensions[ pagePathIndex ] )
+						) {
+							pagePaths.push( dimensions[ pagePathIndex ] );
+						}
+					} );
+				}
+			} );
```
felixarntz, line 228: If `pagePaths` is empty, let's return early here to avoid a useless request. Here it should probably return an empty object though to prevent an endless loading state.

assets/js/modules/analytics/components/module/ModulePopularPagesWidget/index.js
```diff
@@ -174,13 +183,19 @@ export default function ModulePopularPagesWidget( {
 		},
 	];
 
+	const rows = cloneDeep( report[ 0 ].data.rows );
+	// Combine the titles from the pageTitles with the rows from the metrics report.
+	rows.forEach( ( row ) => {
+		const url = row.dimensions[ 0 ];
+		if ( titles[ url ] ) {
+			row.dimensions.unshift( titles[ url ] );
+		}
```
felixarntz, line 192: This must have a fallback / `else` too, otherwise the data in the report would become inconsistent which could cause problems, if for some reason `titles[ url ]` is not found for example. See the ACs on the issue, this should use a translated "(unknown)" string.

Alternatively (probably better actually), we should handle this within the `getPageTitles` selector so that it doesn't need to be duplicated everywhere it's used.

assets/js/modules/analytics/datastore/report.js
```diff
@@ -183,6 +184,78 @@ const baseSelectors = {
...
+			const urlTitleMap = {};
+
+			( pageTitlesReport?.[ 0 ]?.data?.rows || [] ).forEach(
+				( { dimensions } ) => {
+					if ( ! urlTitleMap[ dimensions[ 0 ] ] ) {
+						// key is the url, value is the page title.
+						urlTitleMap[ dimensions[ 0 ] ] = dimensions[ 1 ];
+					}
+				}
+			);
```
felixarntz, line 254: See my other comment, _at some point_ we need to ensure that `urlTitleMap` has a value for _every_ URL from the initial report. This is likely the case, but if any title isn't found for some reason, we need to populate the map value with a translated "(unknown)" string, per the ACs.

felixarntz (commented):
This looks solid, left a few follow-up comments.

assets/js/modules/analytics/datastore/report.js
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3958/commits/95a9b65d0330d3479871a83f2edb031e0ecb820d

assets/js/modules/analytics/datastore/report.js
ivankruchkoff: https://github.com/google/site-kit-wp/pull/3958/commits/95a9b65d0330d3479871a83f2edb031e0ecb820d

felixarntz (approved):
Excellent work!

eugene-manuilov (approved):
LGTM to me too. Thanks, @ivankruchkoff.

Issue #3932, opened 2021-08-25
Google fonts don't load when enqueued late


aaemnnosttv:
## Bug Description

As discovered in https://github.com/google/site-kit-wp/issues/3625#issuecomment-904854320, when we try to enqueue Google fonts after `(wp|admin)_head`, the fonts end up not being loaded due to some odd legacy behavior around how they are enqueued.

We currently load our fonts primarily using the [JS-based web font loader](https://developers.google.com/fonts/docs/webfont_loader) and secondarily using CSS if the site is using AMP. This is overly complex and not necessary for our use which is largely the reason for the problem here.

We should update our fonts to be loaded as a dependency of our main/base styles using CSS as is the recommended/primary approach in the [Google font docs](https://developers.google.com/fonts/docs/getting_started#overview).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Google fonts used by Site Kit should always be loaded via a CSS stylesheet sourced from `fonts.googleapis.com/css` regardless of the AMP mode
* These fonts should be registered as a dependency of the main Site Kit stylesheet used in wp-admin and the stylesheet used by the admin-bar integration – this means they should never need to be enqueued explicitly
* The `Assets::enqueue_fonts` method should be deprecated for removal in the future but updated to continue to work similar to before (without the `WebFontConfig` variant though)
* All calls to `Assets::enqueue_fonts` should be removed
* The font families which are loaded should continue to be filterable using the current `googlesitekit_font_families` filter
* The JS-based `WebFontConfig` used in Storybook should be replaced with a stylesheet tag for consistency 

## Implementation Brief

* Finish POC branch https://github.com/google/site-kit-wp/compare/develop...followup/google-fonts-loading
    * Most of the above is already done in this branch, except deprecating `enqueue_fonts` and the changes to Storybook `preview-head.html`

### Test Coverage

* No significant changes expected, may need to update a test or two due to `enqueue_fonts` becoming deprecated

### Visual Regression Changes

* No changes – this should be a clean refactor for VRT

## QA Brief

* See https://github.com/google/site-kit-wp/issues/3625#issuecomment-904854320 related to the feature tour for Idea Hub on the post list view – ensure Site Kit fonts are loaded in this situation and the font/styles used here match that used by feature tours on the SK dashboard
* Check everywhere else where Site Kit adds an interface to ensure our fonts are used 
    * SK dashboard, page dashboard, and module pages
    * WP Admin bar
[...]

PR #3960, opened 2021-08-30
Replace Google font loading with stylesheet.

aaemnnosttv:
## Summary

Addresses issue #3932

## Relevant technical choices

- Should we load our fonts on the WP dashboard? It seems to be a very small difference, but I think we generally try to use core styles on the WP dashboard so I don't think we should but I've kept it the way it was for consistency.

## Checklist


stories/module-analytics-settings.stories.js
```diff
@@ -598,7 +598,7 @@ storiesOf( 'Analytics Module/Settings', module )
 			);
 		},
 		{
-			decorators: [ withRegistry ],
+			decorators: [ WithRegistry ],
```
aaemnnosttv, line 601: This is just a stray error that I caught looking at another issue. It was showing up in the terminal but not causing the story to fail the storybook tests probably because it caused all stories in the module to be invalid?

includes/Core/Assets/Assets.php
```diff
@@ -204,74 +196,43 @@ public function enqueue_asset( $handle ) {
...
+		if ( ! empty( $assets['googlesitekit-fonts'] ) && $assets['googlesitekit-fonts'] instanceof Asset ) {
+			$assets['googlesitekit-fonts']->enqueue();
 		}
+	}
 
+	/**
+	 * Get Google fonts src for CSS.
+	 *
+	 * @since n.e.x.t
+	 * @return false|string String URL src, or `false` if no font families to load.
```
felixarntz, line 215: Nit-pick (that would be caught by #924 😆):
```suggestion
	 * @since n.e.x.t
	 *
	 * @return false|string String URL src, or `false` if no font families to load.
```

Also, maybe use `bool|string`? I've seen `false` in other places, but it's a bit strange since it's not a type.

includes/Core/Assets/Assets.php
```diff
@@ -204,74 +196,43 @@ public function enqueue_asset( $handle ) {
...
 			'Google+Sans:300,300i,400,400i,500,500i,700,700i',
 			'Roboto:300,300i,400,400i,500,500i,700,700i',
 		);
 
 		$filtered_font_families = apply_filters( 'googlesitekit_font_families', $font_families );
 
-		if ( ! is_array( $filtered_font_families ) || empty( $filtered_font_families ) ) {
-			return;
+		if ( empty( $filtered_font_families ) ) {
+			return false;
```
felixarntz, line 226: Nit-pick, but maybe let's make this an empty string? While this works in PHP, I'm always in favor of using one return type where possible, and both `false` and `''` are false-y anyway.

felixarntz (approved):
@aaemnnosttv LGTM! Just two nit-picks, but not blocking.

felixarntz:
@aaemnnosttv e2e tests failed multiple times here (not related to this PR), did we possibly introduce something unstable recently? Or maybe just random 🤔

includes/Core/Assets/Assets.php
aaemnnosttv: Right, I think maybe the IDE generated it 🤔  Will update 👍

aaemnnosttv: Now just a string per the other comment 😄

aaemnnosttv:
> e2e tests failed multiple times here before finally passing (not related to this PR), did we possibly introduce something unstable recently?

I've only seen the 4.7 job be somewhat unstable recently. Were you seeing others as well?

includes/Core/Assets/Assets.php
```diff
@@ -204,74 +196,43 @@ public function enqueue_asset( $handle ) {
...
+		if ( ! empty( $assets['googlesitekit-fonts'] ) && $assets['googlesitekit-fonts'] instanceof Asset ) {
+			$assets['googlesitekit-fonts']->enqueue();
 		}
+	}
 
+	/**
+	 * Get Google fonts src for CSS.
+	 *
+	 * @since n.e.x.t
+	 * @return string String URL src.
```
felixarntz, line 215: @aaemnnosttv You probably missed that from the previous review, there should be a blank line here.
```suggestion
	 * @since n.e.x.t
	 *
	 * @return string String URL src.
```

felixarntz:
@aaemnnosttv 
> I've only seen the 4.7 job be somewhat unstable recently. Were you seeing others as well?

No it was indeed only that one, it took I think 4 attempts here.

Issue #3973, opened 2021-08-31
Scaffold contributor documentation structure


felixarntz:
This is the initial issue around the contributor documentation epic #2830, setting up the infrastructure and scaffolding all articles intended to be written.

All article files should be scaffolded here, so that we have clear file names figured out from the start and can e.g. cross-link to other articles, even before they are being written.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `/docs` folder should be created.
* Markdown files for all articles to be written (see issues in #2830) should be scaffolded.
* An index / TOC should be scaffolded in the `README.md` file, including links to each article file, in the order in which they are specified in the [original planning doc](https://docs.google.com/document/d/1uiLfbH6EUsOBryjHd3HdhRnwjNM6iw4-ug5Z183mduw/edit).
* Every article (in its own file and the index) should be annotated with a `(TODO)`, which should later be removed per individual article that is being written.

## Implementation Brief

* Merge #3974.

### Test Coverage

* N/A

### Visual Regression Changes

* N/A

## QA Brief

* Go to the new `/docs` directory in the GitHub repository.
* Make sure the links between the article scaffolded files work correctly.

## Changelog entry

* N/A

aaemnnosttv:
@felixarntz LGTM although I think we also need to tweak our configuration for the Plugin zips workflow not to run for these PRs. See my comment on the PR https://github.com/google/site-kit-wp/pull/3974#pullrequestreview-743682348

Otherwise IB ✅

wpdarren:
### QA Update: ✅ 

- In the new `/docs` directory the links between the article scaffolded files work correctly.

PR #3974, opened 2021-09-01
Scaffold docs directory with TOC readme file.

felixarntz:
## Summary

Addresses issue #3973

## Checklist


aaemnnosttv (commented):
Looks good to me! One thing I noticed though is that the Plugin zips workflow ran as part of this which I don't think we want to be running for these kinds of changes. Looking at the workflow, it does not seem to have any file type filters that it runs for on PRs so I think this may be as simple as adding an exclude the docs. The example in the GHA docs looks like exactly what we need (just only to, or at least additionally to exclude for PRs though, not only pushes) 🙂 

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-ignoring-paths

aaemnnosttv (approved):
:shipit:

Issue #2798, opened 2021-02-12
Block external WP HTTP requests in E2E


aaemnnosttv:
## Feature Description

In our PHPUnit tests, we block all out going HTTP requests to prevent external factors from influencing test results.

We should do the same for E2E, the difference is that we can do better to completely disconnect the server side from making external connections using Docker.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* No external server-side HTTP requests should be allowed in E2E tests via the WP HTTP API

## Implementation Brief

* Update `bin/local-env/docker-compose.yml` to set additional configuration via the `WORDPRESS_CONFIG_EXTRA` environment variable already used in the Docker service definition
	* [Set the `WP_HTTP_BLOCK_EXTERNAL` constant to `true`](https://github.com/google/site-kit-wp/pull/2799/files#diff-7f3de7c3ccd5e7d9decf2a88b626862a7edcbba443f523cb45b2a816628013fbR14), but only for non-WP-CLI commands (since they share the same config) – this is necessary to allow for the E2E provisioning scripts to work which largely use WP CLI to install/update WP, plugins, and themes which need access to the external network

### Test Coverage

* No change in coverage, may require updates to E2E if tests fail due to failing HTTP requests 

### Visual Regression Changes

* N/A

## QA Brief

* Remove the existing docker volume for the local environment using `docker volume rm googlesitekit-e2e_wordpress_data`
* Start the local environment running `npm run env:start` command
* Visit `http://localhost:9002/wp-admin/` and log in using `admin`/`password` credentials
* Verify that the WordPress site works
* Now create a new file (name doens't matter) in the `tests/e2e/mu-plugins` folder and put the following content into it:
   ```php
   <?php

   add_action(
   	'init',
   	function() {
   		$resp = wp_remote_get( 'https://api.github.com/repos/google/site-kit-wp' );
   		if ( ! is_wp_error( $resp ) ) {
   			wp_die( "WP_HTTP_BLOCK_EXTERNAL doesn't work!" );
   		}
   	}
   );
   ```
* Go back to the browser and refresh the page. Verify that you don't see the error screen with the message `WP_HTTP_BLOCK_EXTERNAL doesn't work!`.

## Changelog entry

* N/A

eugene-manuilov:
[...]

PR #3987, opened 2021-09-03
Enhancement/2798 block external requests

eugene-manuilov:
## Summary

Addresses issue #2798

## Relevant technical choices

* Renamed the volume to make sure the new constant is available on local machines

## Checklist

### DIFF
--- a/bin/local-env/docker-compose.yml
+++ b/bin/local-env/docker-compose.yml
@@ -15,7 +15,7 @@ services:
         define( 'SCRIPT_DEBUG', true );
         define( 'WP_HTTP_BLOCK_EXTERNAL', ! defined( 'WP_CLI' ) );
     volumes:
-      - wordpress_data:/var/www/html
+      - wp_data:/var/www/html
       - ../../:/var/www/html/wp-content/plugins/google-site-kit
       - ../../tests/e2e/plugins:/var/www/html/wp-content/plugins/google-site-kit-test-plugins
       - ../../tests/e2e/mu-plugins:/var/www/html/wp-content/mu-plugins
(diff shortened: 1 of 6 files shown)


bin/local-env/docker-compose.yml
```diff
@@ -44,4 +46,4 @@ services:
       MYSQL_DATABASE: wordpress_test
 
 volumes:
-  wordpress_data:
+  wp_data:
```
aaemnnosttv, line 49: Why change this volume?

tests/e2e/utils/activate-plugins.js
```diff
@@ -46,3 +48,27 @@ export async function activatePlugins( ...plugins ) {
...
+	await switchUserToAdmin();
+	await visitAdminPage( 'plugins.php' );
+
+	for ( const plugin of plugins ) {
+		await page
+			.click( `tr[data-slug="${ plugin }"] input` )
+			.catch( () => {} );
+	}
+
+	await page.select( '#bulk-action-selector-top', 'deactivate-selected' );
```
aaemnnosttv, line 69: The second argument here is the only difference between these two functions. How about we extract a bulk plugin action function that these two use?

eugene-manuilov: Yep, good idea. Updated.

bin/local-env/docker-compose.yml
aaemnnosttv: Ah that makes sense, although couldn't we delete it by running `docker-compose down -v` so that it's recreated next time?

Changing the volume name is a clever idea but maybe we should make the versioning a bit more explicit? 
```suggestion
  # Increment the volume name when the generate wp-config.php needs to be regenerated.
  wordpress_data_v2:
```

Alternatively, can we extract the extra constants into a file we mount so we don't need to recreate the volume just to add/change a constant? This could be the whole wp-config or just the extra parts defined in that env variable – then the `WORDPRESS_CONFIG_EXTRA` could simply be something like `require wp-config-extra.php;`?

eugene-manuilov: > although couldn't we delete it by running `docker-compose down -v` so that it's recreated next time?

Yes, IMO this is the best option for us. I have updated the `env:stop` command to delete volumes when we stop the environment.

aaemnnosttv (approved):

Issue #3080, opened 2021-04-02
When clicking on any AMP link in the "Most Popular Content" tab it opens a canonical "Detailed Page Stats" - which excludes AMP traffic


jamesozzie:
## Bug Description

In a users "Most Popular Content" table both AMP and non AMP URLs may be listed. When clicking on an AMP URL to see further stats via the "Detailed Page Stats" only the canonical stats appear, with the resulting data excluding AMP traffic. 

Any fix would also apply to using the admin toolbar, to display AMP URL specific insights. 

![image](https://user-images.githubusercontent.com/41326532/117869531-cf6e1700-b292-11eb-9bdd-32dc3a3b6f08.png)

## Steps to reproduce

1. Setup Site Kit on a new site with paired AMP (via the AMP plugins transitional or reader mode, preferably transitional mode in case your homepage shows a list of posts)
2. Connect the Analytics module and ensure logged in users are counted in reporting
3. Visit both your AMP homepage and your non AMP homepage 5 times each
4. Both the AMP version of your homepage and canonical AMP version appear in the "Most popular content"
5. When clicking on the AMP version it will show the canonical version data

## Screenshots

![2021-04-02-13-20-12](https://user-images.githubusercontent.com/41326532/113415018-3821c400-93b6-11eb-8d1f-fa4cdf9841bf.gif)

## Additional Context

 - [Video demonstrating issue](https://recordit.co/O8NhsdLSaY)
 - SK 1.29.0
 - When clicking on the SK link to check further insights at Analytics traffic is broken up between AMP and non AMP (see below)

![2021-04-02-13-23-30](https://user-images.githubusercontent.com/41326532/113415216-ab2b3a80-93b6-11eb-8ef5-1778b07d0f99.gif)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Secondary AMP URLs should be properly identified as a distinct entity and viewable in the single URL view.
* The `Google\Site_Kit\Core\Util\Entity` class should be enhanced:
    * It should receive a new `mode` string class property that can be set in the constructor with the other arguments. This property should by default be empty and allow to specify an identifier for a certain "sub-variant" of an entity URL (this could be for example the secondary AMP version of the entity, or the embed version of the entity).
    * A public `get_mode` method should also be present.
* The `Google\Site_Kit\Core\Util\Entity_Factory` class should be enhanced:
[...]

PR #4019, opened 2021-09-11
Bug/3080 amp canonical

kuasha420:
## Summary

Addresses issue #3080

## Checklist

### DIFF
--- a/includes/Core/Util/Entity_Factory.php
+++ b/includes/Core/Util/Entity_Factory.php
@@ -493,7 +493,7 @@ final class Entity_Factory {
 	}
 
 	/**
-	 * Gets the entity for the given URL, if available.
+	 * Converts given entity to amp entity if the given url is an amp url.
 	 *
 	 * @since n.e.x.t
 	 *
@@ -517,7 +517,7 @@ final class Entity_Factory {
 			// check if the $url has amp query param.
 			if ( array_key_exists( 'amp', $url_query_params ) ) {
 				$new_url = add_query_arg( 'amp', '1', $current_url );
-				return self::create_updated_amp_entity( $new_url, $entity );
+				return self::convert_to_amp_entity( $new_url, $entity );
 			}
 		}
 
@@ -526,12 +526,12 @@ final class Entity_Factory {
 			// That's the reason why we need to check for both version.
 			if ( '/amp' === substr( $url_parts['path'], -4 ) ) { // -strlen('/amp') is -4
 				$new_url = untrailingslashit( $url_parts['path'] ) . '/amp';
-				return self::create_updated_amp_entity( $new_url, $entity );
+				return self::convert_to_amp_entity( $new_url, $entity );
 			}
 
 			if ( '/amp/' === substr( $url_parts['path'], -5 ) ) { // -strlen('/amp/') is -5
 				$new_url = untrailingslashit( $url_parts['path'] ) . '/amp/';
-				return self::create_updated_amp_entity( $new_url, $entity );
+				return self::convert_to_amp_entity( $new_url, $entity );
 			}
 		}
 
@@ -539,7 +539,7 @@ final class Entity_Factory {
 	}
 
 	/**
-	 * Creates updated AMP Entity..
+	 * Converts given entity to amp entity by changing the entity url and adding correct mode.
 	 *
 	 * @since n.e.x.t
 	 *
@@ -547,7 +547,7 @@ final class Entity_Factory {
 	 * @param Entity $entity The initial entity.
 	 * @return Entity The new entity.
 	 */
-	private static function create_updated_amp_entity( $new_url, $entity ) {
+	private static function convert_to_amp_entity( $new_url, $entity ) {
 		$new_entity = new Entity(
 			$new_url,
 			array(
--- a/includes/Core/Util/Entity_Factory.php
+++ b/includes/Core/Util/Entity_Factory.php
@@ -11,6 +11,7 @@
 namespace Google\Site_Kit\Core\Util;
 
 use Google\Site_Kit\Context;
+use Google\Site_Kit\Plugin;
 use WP_Query;
 use WP_Post;
 use WP_Term;
@@ -62,11 +63,8 @@ final class Entity_Factory {
 		if ( $wp_the_query instanceof WP_Query ) {
 			$entity = self::from_wp_query( $wp_the_query );
 
-			if ( is_null( $entity ) || ! defined( 'AMP__VERSION' ) ) {
-				return $entity;
-			}
-
-			return self::maybe_convert_to_amp_entity($_SERVER['REQUEST_URI'], $entity); // phpcs:ignore
+			$request_uri = Plugin::instance()->context()->input()->filter( INPUT_SERVER, 'REQUEST_URI' );
+			return self::maybe_convert_to_amp_entity( $request_uri, $entity );
 		}
 
 		return null;
@@ -91,10 +89,6 @@ final class Entity_Factory {
 		$query->get_posts();
 
 		$entity = self::from_wp_query( $query );
-
-		if ( is_null( $entity ) || ! defined( 'AMP__VERSION' ) ) {
-			return $entity;
-		}
 
 		return self::maybe_convert_to_amp_entity( $url, $entity );
 	}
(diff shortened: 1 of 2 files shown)


includes/Core/Util/Entity_Factory.php
```diff
@@ -60,8 +60,54 @@ public static function from_context() {
...
-			return self::from_wp_query( $wp_the_query );
+			$entity = self::from_wp_query( $wp_the_query );
+
+			if ( null === $entity ) {
+				return null;
+			}
+
+			if ( ! defined( 'AMP__VERSION' ) ) {
+				return $entity;
+			}
```
eugene-manuilov, line 71: Let's combine these two conditions into one:

```suggestion
			if ( is_null( $entity ) || ! defined( 'AMP__VERSION' ) ) {
				return $entity;
			}
```

includes/Core/Util/Entity_Factory.php
```diff
@@ -83,7 +129,52 @@ public static function from_url( $url ) {
...
-		return self::from_wp_query( $query );
+		$entity = self::from_wp_query( $query );
+
+		if ( null === $entity ) {
+			return null;
+		}
+
+		if ( ! defined( 'AMP__VERSION' ) ) {
+			return $entity;
+		}
```
eugene-manuilov, line 140: Same here, we can combine two conditions into one.

includes/Core/Util/Entity_Factory.php
```diff
@@ -60,8 +60,54 @@ public static function from_context() {
...
+				$new_url,
+				array(
+					'id'    => $entity->get_id(),
+					'type'  => $entity->get_type(),
+					'title' => $entity->get_title(),
+					'mode'  => 'amp_secondary',
+				)
+			);
+
+			return $new_entity;
```
eugene-manuilov, line 108: Looks like this logic is almost the same for both methods. Let's extract it into a new private static method `maybe_convert_to_amp_entity( $url )` and use it in both methods.

includes/Core/Util/Entity_Factory.php
```diff
@@ -60,8 +60,54 @@ public static function from_context() {
...
+				return $entity;
+			}
+
+			$url_parts    = wp_parse_url( $_SERVER['REQUEST_URI'] ); // phpcs:ignore
+			$new_url_tail = '';
+
+			// check if the $url has amp query param.
+			if ( strpos( $url_parts['query'], 'amp' ) !== false ) {
+				$new_url_tail = '?amp=1';
+			}
```
eugene-manuilov, line 79: Checking the query argument using `strpos` is not sufficient because it can produce false positives when `amp` sequence is present in a query arg value (`?product_type=lamp`) or is a part of another query key (`?lamp=true`). Let's use the [`wp_parse_str`](https://developer.wordpress.org/reference/functions/wp_parse_str/) function to parse query arguments and check whether the `amp` key exists.

includes/Core/Util/Entity_Factory.php
```diff
@@ -60,8 +60,54 @@ public static function from_context() {
...
+
+			// check if the $url has `/amp` in path.
+			if ( '/amp' === substr( $url_parts['path'], -strlen( '/amp' ) ) ) {
+				$new_url_tail = '/amp';
+			}
+
+			// check if the $url has `/amp/` in path.
+			if ( '/amp/' === substr( $url_parts['path'], -strlen( '/amp/' ) ) ) {
+				$new_url_tail = '/amp/';
+			}
```
eugene-manuilov, line 89: Let's strip the trailing slash from the `$url_parts['path']` using the [`untrailingslashit`](https://developer.wordpress.org/reference/functions/untrailingslashit/) function and check only for `/amp` suffix.

Also, no need to use `-strlen( '/amp' )` because it always returns the same result which is `-4`. You can leave a comment to explain what `-4` means.

eugene-manuilov (changes requested):
Nice work, @kuasha420. Left some comments for you. Please, take a look.

includes/Core/Util/Entity_Factory.php
kuasha420: Done. But wouldn't only checking for non-trailing slashed version and returning non trailing slashedd url cause issues if the original url is trailing slashed? Don't we need to append `/amp` vs `/amp/` depending on trailing slash? 

Cheers.

includes/Core/Util/Entity_Factory.php
```diff
@@ -60,8 +60,15 @@ public static function from_context() {
...
 		// Otherwise, run frontend-specific `WP_Query` logic.
 		if ( $wp_the_query instanceof WP_Query ) {
-			return self::from_wp_query( $wp_the_query );
+			$entity = self::from_wp_query( $wp_the_query );
+
+			if ( is_null( $entity ) || ! defined( 'AMP__VERSION' ) ) {
+				return $entity;
+			}
+
+			return self::maybe_convert_to_amp_entity($_SERVER['REQUEST_URI'], $entity); // phpcs:ignore
```
eugene-manuilov, line 69: @kuasha420, please, use the same approach to get the `REQUEST_URI` value as we do it in here: https://github.com/google/site-kit-wp/blob/2f43c80560510e50bd5a3af914baa52f48f143c2/includes/Modules/Site_Verification.php#L98

It will eliminate a need to use `// phpcs:ignore`. To get the context instance, you can use `Plugin::instance()->context()->...`.

includes/Core/Util/Entity_Factory.php
```diff
@@ -60,8 +60,15 @@ public static function from_context() {
 
 		// Otherwise, run frontend-specific `WP_Query` logic.
 		if ( $wp_the_query instanceof WP_Query ) {
-			return self::from_wp_query( $wp_the_query );
+			$entity = self::from_wp_query( $wp_the_query );
+
+			if ( is_null( $entity ) || ! defined( 'AMP__VERSION' ) ) {
+				return $entity;
+			}
```
eugene-manuilov, line 67: Looks like this condition repeats in both methods. Let's move it to the `maybe_convert_to_amp_entity` method and put it at the very beginning of the method:

```php
private static function maybe_convert_to_amp_entity( $url, $entity ) {
    if ( is_null( $entity ) || ! defined( 'AMP__VERSION' ) ) {
        return $entity;
    }

    ...
}
```

includes/Core/Util/Entity_Factory.php
```diff
@@ -484,4 +497,49 @@ private static function prefix_title( $title, $prefix ) {
...
+	 *
+	 * @param string $url URL to determine the entity from.
+	 * @param Entity $entity The initial entity.
+	 * @return Entity The initial or new entity for the given URL.
+	 */
+	private static function maybe_convert_to_amp_entity( $url, $entity ) {
+		$url_parts    = wp_parse_url( $url );
+		$new_url_tail = '';
+
+		wp_parse_str( $url_parts['query'], $url_query_params );
```
eugene-manuilov, line 514: Let's define the `$url_query_params` variable before passing it to the function, plus we need to call the function only if the `query` property exists in the array:
```suggestion
		$url_query_params = array();
		if ( ! empty( $url_parts['query'] ) {
			wp_parse_str( $url_parts['query'], $url_query_params );
		}
```

includes/Core/Util/Entity_Factory.php
```diff
@@ -484,4 +497,49 @@ private static function prefix_title( $title, $prefix ) {
...
+
+		// check if the $url has amp query param.
+		if ( array_key_exists( 'amp', $url_query_params ) ) {
+			$new_url_tail = '?amp=1';
+		}
+
+		// check if the $url has `/amp` in path.
+		if ( '/amp' === substr( untrailingslashit( $url_parts['path'] ), -4 ) ) { // -strlen('/amp') is -4
+			$new_url_tail = '/amp';
+		}
```
eugene-manuilov, line 524: Same here, we need to make sure that the `$url_parts['path']` is not empty before working with it.

includes/Core/Util/Entity_Factory.php
```diff
@@ -484,4 +497,49 @@ private static function prefix_title( $title, $prefix ) {
...
+
+		if ( empty( $new_url_tail ) ) {
+			return $entity;
+		}
+
+		$new_url = $entity->get_url();
+		$new_url = $new_url . $new_url_tail;
```
eugene-manuilov, line 531: We can't just concatenate the entity URL and the new tail because the existing entity URL may already have query parameters which will lead to something like this: `http://example.com/?p=12345?amp=1`.

eugene-manuilov (changes requested):
Thanks, @kuasha420. This is closer to the end. Please, take a look at my comments below and make the appropriate changes. Once they are done, we should be good to go.

includes/Core/Util/Entity_Factory.php
eugene-manuilov: Yes, it's a good point, @kuasha420. I think you are right, let's append a trailing slash if it exists in the original URL.

kuasha420: Should I just revert to old logic of checking twice for trailing slash/non-trailing slash version or is there a better way too do it? @eugene-manuilov

eugene-manuilov: yes, let's revert it to the old logic and leave a comment that explains why we have it separated.

includes/Core/Util/Entity_Factory.php
```diff
@@ -484,4 +491,73 @@ private static function prefix_title( $title, $prefix ) {
 			$title
 		);
 	}
+
+	/**
+	 * Gets the entity for the given URL, if available.
```
eugene-manuilov, line 496: @kuasha420, could you, please, provide a more relevant description for this method? After reading the description, it should be clear what this method does without reading the code itself. No need to go into too many details though, just a brief overview.

includes/Core/Util/Entity_Factory.php
```diff
@@ -484,4 +491,73 @@ private static function prefix_title( $title, $prefix ) {
...
+				$new_url = untrailingslashit( $url_parts['path'] ) . '/amp/';
+				return self::create_updated_amp_entity( $new_url, $entity );
+			}
+		}
+
+		return $entity;
+	}
+
+	/**
+	 * Creates updated AMP Entity..
```
eugene-manuilov, line 542: Same here, please, add a more relevant description.

includes/Core/Util/Entity_Factory.php
```diff
@@ -484,4 +491,73 @@ private static function prefix_title( $title, $prefix ) {
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $new_url URL of the new entity.
+	 * @param Entity $entity The initial entity.
+	 * @return Entity The new entity.
+	 */
+	private static function create_updated_amp_entity( $new_url, $entity ) {
```
eugene-manuilov, line 550: Also, let's rename this method. If we have `maybe_convert_to_amp_entity` method, then this method should be named `convert_to_amp_entity`.

eugene-manuilov (changes requested):
Almost looks good to me. Just a few nitpicks and it should be good to go. @kuasha420, please, take a look at my comments.

includes/Core/Util/Entity_Factory.php
```diff
@@ -484,4 +491,73 @@ private static function prefix_title( $title, $prefix ) {
 			$title
 		);
 	}
+
+	/**
+	 * Converts given entity to amp entity if the given url is an amp url.
```
eugene-manuilov, line 496: Same here, upper case for acronyms:
```suggestion
	 * Converts given entity to AMP entity if the given URL is an AMP URL.
```

eugene-manuilov (changes requested):
@kuasha420, found two more issues with new URLs. Please, take a look at it too.

includes/Core/Util/Entity_Factory.php
```diff
@@ -484,4 +491,73 @@ private static function prefix_title( $title, $prefix ) {
...
+				$new_url = add_query_arg( 'amp', '1', $current_url );
+				return self::convert_to_amp_entity( $new_url, $entity );
+			}
+		}
+
+		if ( ! empty( $url_parts['path'] ) ) {
+			// We need to correctly add trailing slash if the original url had trailing slash.
+			// That's the reason why we need to check for both version.
+			if ( '/amp' === substr( $url_parts['path'], -4 ) ) { // -strlen('/amp') is -4
+				$new_url = untrailingslashit( $url_parts['path'] ) . '/amp';
```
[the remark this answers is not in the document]
kuasha420: YES! My bad. Fixed in next commit.

eugene-manuilov (approved):
Awesome! Thanks, @kuasha420! LGTM!

Issue #3477, opened 2021-06-04
Avoid unnecessary token refresh


aaemnnosttv:
## Feature Description

Access tokens that are generated for communicating for Google APIs are short-lived tokens which are only valid for one hour. Once expired, a new access token can be requested using the refresh token received during a user's initial connection. The refresh token is much longer lived and requires going through oAuth again to get a new one. The underlying PHP API Client library for Google APIs handles refreshing access tokens automatically on-the-fly when needed when a request is made.

As we have started to move away from batch requesting all data via Site Kit's `/data` REST endpoint and more towards a more traditional API architecture (making one request per endpoint), this has introduced the potential for multiple access tokens to be requested at the same time.

The most common situation for this would be on the WordPress dashboard, right after logging in for the first time in over an hour. If Analytics is connected, this would result in 4 separate requests to the Google API at the same time and since the access token would be expired at the time the request is made, a new token would be requested in each case, used once and only the last one would be kept.

![image](https://user-images.githubusercontent.com/1621608/120764670-10b7b680-c521-11eb-9b35-0c5983a183a6.png)

_4 consecutive requests to `/o/oauth2/token/` followed by 4 requests to Google APIs (3 for Analytics, 1 for Search Console)_

While the on-demand token refreshing is important, we should see how to avoid such cases where more tokens are requested than needed.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Instead of mainly relying on the automated on-demand refresh of access tokens that are about to expire (which is built into the Google PHP client library), Site Kit should proactively refresh the current user's access token in that case, **outside of REST API requests** which may come in multiple at a time, causing the problem outlined. (Note that this is not a major problem, so we do not need to prevent it - but we should avoid it where possible.)
    1. On pageload: If the current user has an access token which will expire in the next 30 seconds or less, that access token should be refreshed, relying on the existing `OAuth_Client::refresh_token()` method (which used to be used, but has for a while now been unused).
    2. The same logic should run for every WordPress "heartbeat" request (see `heartbeat_tick` action), so that it also works in the case that the user stays on a Site Kit screen without a page reload for longer than the access token expiry.

## Implementation Brief

* Add new hooks for `admin_init` and `heartbeat_tick` actions to  the `Authentication` class:
  * Get the current token using `$this->token->get()` method;
  * Return early if the current token is empty or it doesn't exire in less than 30 seconds;
[...]

PR #4024, opened 2021-09-13
Enhancement/3477 token refresh

eugene-manuilov:
## Summary

Addresses issue #3477

## Checklist

### DIFF
--- a/includes/Core/Authentication/Authentication.php
+++ b/includes/Core/Authentication/Authentication.php
@@ -416,7 +416,7 @@ final class Authentication {
 			}
 
 			// Do nothing if the token expires in more than 5 minutes.
-			if ( $token['created'] + $token['expires_in'] > time() + 5 * MINUTES_IN_SECONDS ) {
+			if ( $token['created'] + $token['expires_in'] > time() + 5 * MINUTE_IN_SECONDS ) {
 				return;
 			}
 
--- a/includes/Core/Authentication/Authentication.php
+++ b/includes/Core/Authentication/Authentication.php
@@ -405,6 +405,26 @@ final class Authentication {
 			add_action( 'googlesitekit_authorize_user', $set_initial_version );
 			add_action( 'googlesitekit_reauthorize_user', $set_initial_version );
 		}
+
+		// Maybe refresh token.
+		$maybe_refresh_token = function() {
+			$token = $this->token->get();
+
+			// Do nothing if the token is not set.
+			if ( empty( $token['created'] ) || empty( $token['expires_in'] ) ) {
+				return;
+			}
+
+			// Do nothing if the token expires in more than 5 minutes.
+			if ( $token['created'] + $token['expires_in'] > time() - 5 * MINUTES_IN_SECONDS ) {
+				return;
+			}
+
+			$this->get_oauth_client()->refresh_token();
+		};
+
+		add_action( 'admin_init', $maybe_refresh_token );
+		add_action( 'heartbeat_tick', $maybe_refresh_token );
 	}
 
 	/**
--- a/includes/Core/Authentication/Authentication.php
+++ b/includes/Core/Authentication/Authentication.php
@@ -407,6 +407,10 @@ final class Authentication {
 		}
 
 		$maybe_refresh_token = function() {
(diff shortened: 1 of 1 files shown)


includes/Core/Authentication/Authentication.php
```diff
@@ -405,6 +405,26 @@ function ( $user_id ) {
 			add_action( 'googlesitekit_authorize_user', $set_initial_version );
 			add_action( 'googlesitekit_reauthorize_user', $set_initial_version );
 		}
+
+		// Maybe refresh token.
```
aaemnnosttv, line 409: Nit: Let's avoid adding comments that are duplicated in the code.
```suggestion
```

includes/Core/Authentication/Authentication.php
```diff
@@ -405,6 +405,26 @@ function ( $user_id ) {
...
+		$maybe_refresh_token = function() {
+			$token = $this->token->get();
+
+			// Do nothing if the token is not set.
+			if ( empty( $token['created'] ) || empty( $token['expires_in'] ) ) {
+				return;
+			}
+
+			// Do nothing if the token expires in more than 5 minutes.
+			if ( $token['created'] + $token['expires_in'] > time() + 5 * MINUTE_IN_SECONDS ) {
```
aaemnnosttv, line 419: The AC mentions "30 seconds or less" for this period where as this uses 5 minutes.

This sounds good to me as a more proactive refresh, especially since heartbeat ticks can be a few minutes apart. @felixarntz are you okay with this change?

aaemnnosttv (approved):
LGTM! One question for @felixarntz regarding a change that doesn't meet the current ACs. See my comment below.

includes/Core/Authentication/Authentication.php
eugene-manuilov: :man_facepalming: I don't know why I set 5 minutes here although we discussed it during working on IB. Sorry about it...

includes/Core/Authentication/Authentication.php
```diff
@@ -405,6 +405,25 @@ function ( $user_id ) {
 			add_action( 'googlesitekit_authorize_user', $set_initial_version );
 			add_action( 'googlesitekit_reauthorize_user', $set_initial_version );
 		}
+
+		$maybe_refresh_token = function() {
+			$token = $this->token->get();
```
felixarntz, line 410: Maybe a good idea to add a permission check here for whether the user can generally authenticate? Not super-critical, but I think that's a good pattern. A person without permissions shouldn't have a token set anyway, but it could be relevant e.g. if your permission is downgraded from admin to editor.

Also, maybe worth checking for whether Site Kit credentials are set as well, otherwise this won't work.

includes/Core/Authentication/Authentication.php
felixarntz: Using 5 minutes sounds good to me, per Evan's point.

includes/Core/Authentication/Authentication.php
```diff
@@ -405,6 +405,25 @@ function ( $user_id ) {
...
+			// Do nothing if the token expires in more than 5 minutes.
+			if ( $token['created'] + $token['expires_in'] > time() + 5 * MINUTE_IN_SECONDS ) {
+				return;
+			}
+
+			$this->get_oauth_client()->refresh_token();
+		};
+
+		add_action( 'admin_init', $maybe_refresh_token );
+		add_action( 'heartbeat_tick', $maybe_refresh_token );
```
felixarntz, line 426: It would be _kind of_ nicer to only do this when in any Site Kit-related UI (e.g. Site Kit screen, WP dashboard, etc.). But that may require a bunch of overhead in engineering - and realistically, most people land in the WP dashboard after logging in anyway, which should be the most common scenario where this logic would trigger one way or another.

felixarntz (commented):
@eugene-manuilov Two additional points of feedback here, other than that this should be good to go.

includes/Core/Authentication/Authentication.php
eugene-manuilov: Added permissions and credentials check.

felixarntz (approved):
Great!

Issue #3980, opened 2021-09-01
Implement additional Idea Hub events for widget page navigation


felixarntz:
As a follow-up to #3817, this issue defines a few additional GA tracking events Idea Hub interactions around page navigation that should be implemented.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The following additional GA tracking events should be sent for their respective interactions within the Idea Hub widget:
    * `new_page_advance`: User clicks on the button to go to the next page in the "New" tab.
    * `new_page_return`: User clicks on the button to go to the previous page in the "New" tab.
    * `saved_page_advance`: User clicks on the button to go to the next page in the "Saved" tab.
    * `saved_page_return`: User clicks on the button to go to the previous page in the "Saved" tab.
    * `drafts_page_advance`: User clicks on the button to go to the next page in the "Drafts" tab.
    * `drafts_page_return`: User clicks on the button to go to the previous page in the "Drafts" tab.
* For all of these events, the event category should be `idea_hub_widget`, and the **current** page number (starting at 1, i.e. not index) should be provided as event label (e.g. if you go from page 2 to page 3, it should send 2).

## Implementation Brief

Inside `assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Pagination.js`:
- Import `IDEA_HUB_GA_CATEGORY_WIDGET` from `assets/js/modules/idea-hub/datastore/constants.js`
- inside the `handleNext` callback, add the following logic
  - depending on the current tab (you could use a `switch` statement, for example), the `trackEvent` function is called with the appropriate `..._page_advance` event.
  - along with the event key, pass `IDEA_HUB_GA_CATEGORY_WIDGET` as the category parameter and the current page number as the label parameter. The function signature is `trackEvent( category, action, label )`.
  - See the following gist to get a better idea:
  ```js
  if( tab === 'new-ideas' ) {
      trackEvent( IDEA_HUB_GA_CATEGORY_WIDGET, 'new_page_advance', page );
  } else if ( tab === 'saved-ideas' ) {
      trackEvent( IDEA_HUB_GA_CATEGORY_WIDGET, 'saved_page_advance', page );
  } else if ( tab === 'draft-ideas') {
      trackEvent( IDEA_HUB_GA_CATEGORY_WIDGET, 'draft_page_advance', page );
  }
  ```
[...]

PR #4060, opened 2021-09-16
Add constants for Idea Hub tab names to avoid future typos (3980).

johnPhillips:
## Summary

Addresses issue #3980

## Relevant technical choices

There was a silly typo in one of the instances where the tab names were inlined. It occurred to me that we are inlining these strings all over the place in the dashboard widget now, so to avoid this happening again I've made them constants.

## Checklist

### DIFF
--- a/assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Pagination.js
+++ b/assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Pagination.js
@@ -36,6 +36,9 @@ import {
 	IDEA_HUB_IDEAS_PER_PAGE,
 	MODULES_IDEA_HUB,
 	IDEA_HUB_GA_CATEGORY_WIDGET,
+	IDEA_HUB_TAB_NAMES_NEW,
+	IDEA_HUB_TAB_NAMES_SAVED,
+	IDEA_HUB_TAB_NAMES_DRAFT,
 } from '../../../datastore/constants';
 import { trackEvent } from '../../../../../util';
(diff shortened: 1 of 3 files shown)


aaemnnosttv (commented):
Thanks @johnPhillips !

This looks great, I just see one more instance that we should also update:
https://github.com/google/site-kit-wp/blob/1dbccbb472a2619566c285e00b7a358e9a8c726f/assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Footer.js#L57

Let's update this one too then this should be good to go!

johnPhillips:
Nice catch - updated!

assets/js/components/Button.js
```diff
@@ -118,7 +118,16 @@ const Button = forwardRef(
...
-					{ ButtonComponent }
+					{
+						// If the button is disabled it has no events for the Tooltip
+						// component to use, so it needs to be wrapped in a containing
+						// element.
+						disabled ? (
+							<span>{ ButtonComponent }</span>
+						) : (
+							ButtonComponent
+						)
```
aaemnnosttv, line 129: I'm not sure if this is the right solution here. Would you please revert this change for now and we can address it separately?

aaemnnosttv (commented):
Thanks @johnPhillips ! This is good to go, but let's handle the button change in a separate issue.

assets/js/components/Button.js
johnPhillips: Sure 👍
Just for context, this is where the solution came from: https://mui.com/components/tooltips/#disabled-elements
Happy to revisit as a separate issue.

aaemnnosttv (approved):
Great!

Issue #3909, opened 2021-08-19
Idea Hub post annotation only shown on drafts


ivankruchkoff:
## Bug Description

When you trash a draft post, the idea isn't shown in the trash.

## Steps to reproduce

1. Create a draft from an idea, notice the idea name is in the status.
2. Trash the post, idea name is not shown.
3. Restore the trashed post, and see the idea is shown again.

## Recording

https://user-images.githubusercontent.com/373931/130031511-4fe045a8-c7f4-4f7a-a4dc-d3ac43fadcb7.mp4

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The Idea Hub annotation added to Idea Hub created posts in the post list table should be shown for posts of all statuses, not only drafts (e.g. trashed, future, pending, private, etc) – how it is added depends on the post status, post state, and post title
* For `draft` posts with no title only, the state should be replaced by `Idea Hub Draft “{idea-name}”`
* For all other post statuses OR drafts that have titles, the annotation should be simplified to `inspired by Idea Hub`
* For all post statuses (except drafts with no title)
	- If there is already content for the post state (e.g. [`scheduled`](https://github.com/WordPress/WordPress/blob/baa81b67d19e29be29d0ac49c9bede316ab8f80a/wp-admin/includes/template.php#L2210)), combine the original content of the post state by appending the annotation like so: `{post-state}, {idea-hub-annotation}`
	- If no content exists for the post state, the annotation becomes the post state, rather than appending to it
* See https://github.com/google/site-kit-wp/pull/4061#discussion_r712384583 for reference

## Implementation Brief

* In `includes/Modules/Idea_Hub.php` -> `register_persistent()` -> `display_post_states()`:
* 1) Remove the check for post status:
https://github.com/google/site-kit-wp/blob/38ae6cb09c82acea56a5bde768028d93b3cc6580/includes/Modules/Idea_Hub.php#L165-L167

* 2) Instead of adding our message to the `draft` post state, add it for the current post's post state
https://github.com/google/site-kit-wp/blob/38ae6cb09c82acea56a5bde768028d93b3cc6580/includes/Modules/Idea_Hub.php#L173

```php
$post_states[$post->post_status ] = ... 
```

[...]

PR #4061, opened 2021-09-16
Idea Hub post annotation only shown on drafts.

ivankruchkoff:
## Summary

Addresses issue #3909

## Checklist


includes/Modules/Idea_Hub.php
```diff
@@ -162,15 +162,14 @@ public function register_persistent() {
...
-				if ( 'draft' !== $post->post_status ) {
-					return $post_states;
-				}
 				$idea = $this->get_post_idea( $post->ID );
 				if ( is_null( $idea ) ) {
 					return $post_states;
 				}
-				/* translators: %s: Idea Hub Idea Title */
-				$post_states['draft'] = sprintf( __( 'Idea Hub Draft “%s”', 'google-site-kit' ), $idea['text'] );
+				$post_states[ $post->post_status ] = '' === $post->post_title ?
```
aaemnnosttv, line 169: I think we may need to reconsider the implementation here as it would be currently overwriting the default states for every post status for all Idea Hub posts for which there are quite a few https://github.com/WordPress/WordPress/blob/baa81b67d19e29be29d0ac49c9bede316ab8f80a/wp-admin/includes/template.php#L2183-L2211

Ideally this would be an additive annotation rather than taking over the state entirely for IH posts. Thoughts @felixarntz?

tests/phpunit/integration/Modules/Idea_HubTest.php
```diff
@@ -135,14 +145,14 @@ public function test_draft_labels() {
...
-		$post_states2 = apply_filters( 'display_post_states', array( 'draft' => 'Draft' ), $post2 );
+		$post_states2 = apply_filters( 'display_post_states', array( 'publish' => 'Draft' ), $post2 );
 		// With a regular draft post
 		$post_states3 = apply_filters( 'display_post_states', array( 'draft' => 'Draft' ), $post3 );
 
 		$post_states1 = apply_filters( 'display_post_states', array( 'draft' => 'Draft' ), $post1 );
 
 		$this->assertEquals( $post_states1, array( 'draft' => 'Idea Hub Draft “Using Site Kit to analyze your success”' ) );
-		$this->assertEquals( $post_states2, array( 'draft' => 'Idea Hub Draft “Why Penguins are guanotelic?”' ) );
+		$this->assertEquals( $post_states2, array( 'publish' => 'inspired by Idea Hub' ) );
```
aaemnnosttv, line 155: Nit: I generally avoid using numbered variables to separate cases in a test unless it's specific to the test in some way such as asserting something is order-specific.

The quick and easy change to make here would be to use more expressive names for the variables themselves (e.g. `$idea_hub_post_published`, `$regular_post_draft`, etc). Ideally, the cases would be extracted into a data provider so that each test run is only for one case at a time (no special names needed). I think this would be the better use here since the cases here are independent of one another.

aaemnnosttv (commented):
Thanks @ivankruchkoff – this looks good as defined although I have a concern that we may want to discuss further before merging this as-is. I've pinged @felixarntz for his thoughts below.

includes/Modules/Idea_Hub.php
felixarntz: @aaemnnosttv @ivankruchkoff Good point, I agree. Let's do it like this:
* Only for the `draft` status specifically, if the title is empty, use the existing "Idea Hub Draft xyz".
* Otherwise, if there is already content for the status, combine the original content with "inspired by Idea Hub" e.g. like `sprintf( __( '%s, inspired by Idea Hub', 'google-site-kit' ), $original_content )`.
* Otherwise (i.e. not a `draft` and there is no original content for the status, set it to "inspired by Idea Hub".

includes/Modules/Idea_Hub.php
```diff
@@ -162,15 +162,14 @@ public function register_persistent() {
 		add_filter(
 			'display_post_states',
 			function( $post_states, $post ) {
-				if ( 'draft' !== $post->post_status ) {
-					return $post_states;
-				}
 				$idea = $this->get_post_idea( $post->ID );
 				if ( is_null( $idea ) ) {
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/4061/commits/bac54c95840bf513402b489ddc99f382924c9bee

includes/Modules/Idea_Hub.php
ivankruchkoff: Updated in https://github.com/google/site-kit-wp/pull/4061/commits/2204098e7c21907855e5c98c95e0b6ee4abf8ebc

includes/Modules/Idea_Hub.php
```diff
@@ -172,15 +172,16 @@ public function register_persistent() {
...
-				if ( 'draft' !== $post->post_status ) {
+				if ( ! $this->is_idea_post( $post->ID ) ) {
 					return $post_states;
 				}
 				$idea = $this->get_post_idea( $post->ID );
-				if ( is_null( $idea ) ) {
-					return $post_states;
+				if ( '' === $post->post_title && 'draft' === $post->post_status ) {
+					/* translators: %s: Idea Hub Idea Title */
+					$post_states['draft'] = sprintf( __( 'Idea Hub Draft “%s”', 'google-site-kit' ), $idea['text'] );
```
aaemnnosttv, line 181: @felixarntz given that WP already allows for multiple post states and formats them as a list, I feel like we shouldn't overwrite the `draft` state here. Why not use the same `inspired by Idea Hub` here as well? We could still include the idea name/title for this case specifically too. If it was a draft it would look like `Draft, inspired by Idea Hub "{idea-name}"`

includes/Modules/Idea_Hub.php
```diff
@@ -172,15 +172,16 @@ public function register_persistent() {
...
+				} else {
+					$post_states['idea'] = __( 'inspired by Idea Hub', 'google-site-kit' );
```
aaemnnosttv, line 183: Maybe this should be more specific to Idea Hub?
```suggestion
					$post_states['idea-hub'] = __( 'inspired by Idea Hub', 'google-site-kit' );
```

tests/phpunit/integration/Modules/Idea_HubTest.php
```diff
@@ -133,18 +143,18 @@ public function test_draft_labels() {
...
+		$post_states_for_draft_non_idea_hub_post = apply_filters( 'display_post_states', array( 'draft' => 'Draft' ), $draft_post_not_idea_hub );
 
-		$post_states1 = apply_filters( 'display_post_states', array( 'draft' => 'Draft' ), $post1 );
+		$post_states_for_draft_post_with_no_title = apply_filters( 'display_post_states', array( 'draft' => 'Draft' ), $draft_idea_hub_post_with_no_title );
 
-		$this->assertEquals( $post_states1, array( 'draft' => 'Idea Hub Draft “Using Site Kit to analyze your success”' ) );
-		$this->assertEquals( $post_states2, array( 'draft' => 'Idea Hub Draft “Why Penguins are guanotelic?”' ) );
-		$this->assertEquals( $post_states3, array( 'draft' => 'Draft' ) );
+		$this->assertEquals( $post_states_for_draft_post_with_no_title, array( 'draft' => 'Idea Hub Draft “Using Site Kit to analyze your success”' ) );
+		$this->assertEquals( $post_states_for_published_post_with_title, array( 'idea' => 'inspired by Idea Hub' ) );
```
aaemnnosttv, line 156:
```suggestion
		$this->assertEquals( $post_states_for_published_post_with_title, array( 'idea-hub' => 'inspired by Idea Hub' ) );
```

tests/phpunit/integration/Modules/Idea_HubTest.php
```diff
@@ -133,18 +143,18 @@ public function test_draft_labels() {
 
 		$this->idea_hub->register_persistent();
 		$this->idea_hub->register();
-		$this->idea_hub->set_post_idea( $post2->ID, $idea );
+		$this->idea_hub->set_post_idea( $published_idea_hub_post_with_title->ID, $idea );
 
 		// With an IdeaHub post
-		$post_states2 = apply_filters( 'display_post_states', array( 'draft' => 'Draft' ), $post2 );
+		$post_states_for_published_post_with_title = apply_filters( 'display_post_states', array( 'idea' => 'inspired by Idea Hub' ), $published_idea_hub_post_with_title );
```
aaemnnosttv, line 149:
```suggestion
		$post_states_for_published_post_with_title = apply_filters( 'display_post_states', array( 'idea-hub' => 'inspired by Idea Hub' ), $published_idea_hub_post_with_title );
```

aaemnnosttv (commented):
Thanks @ivankruchkoff ! This looks good to me, although I have one more question for @felixarntz here in light of the way post states work.

includes/Modules/Idea_Hub.php
felixarntz: @aaemnnosttv That makes sense, however I think overwriting the `draft` state here works as well, we essentially combine two into one. While your suggested approach would be a bit cleaner, I think having that one special situation is acceptable as well, which allows us to keep the copy originally defined.

includes/Modules/Idea_Hub.php
felixarntz: +1

felixarntz (commented):
This looks good to me except for the one detail @aaemnnosttv pointed out, let's replace `idea` with `idea-hub`.

ivankruchkoff:
Addressed in https://github.com/google/site-kit-wp/pull/4105/files

Issue #3208, opened 2021-04-21
Ad blocker warning message is inconsistent in different locations


Pratheep-lab:
It says " _Ad blocker detected, you need to disable it in order to set up AdSense_."

However, the AdSense setup has already been done and I can see that AdSense is working properly. 

Maybe the message should say "Ad blocker detected, you need to disable it in order for AdSense to work properly."

## Steps to reproduce

1. Setup AdSense on Site Kit without any Ad Blockers
2. Enable an adblocker. In this instance, I used https://adblockplus.org/
3. Open the AdSense details page. 
4. You can see the above error. Also seen on the AdSense module on the dashboard.

## Screenshots

![image](https://user-images.githubusercontent.com/81642523/115597990-d1ffd100-a2d1-11eb-895d-8cf6fddc9bb7.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The ad blocker warning should be consistent in all locations under the same conditions. Currently the locations are as follows:
	* Setup CTA (from SK success notification on main dashboard)
	* Setup module (on _Connect More Services_ tab of settings) 
	* AdSense settings view (when setup is incomplete)
	* AdSense setup
	* Dashboard summary widget
	* Dashboard top earning pages widget (only shown when Analytics is also activated and linked with the connected AdSense account)
	* **The module page will be covered separately in #4178**
* A new `getAdBlockerWarningMessage` selector should be added to the AdSense datastore to be used in both places
* The logic in the `AdBlockerWarning` component for [determining which message to use](https://github.com/google/site-kit-wp/blob/f54254f2bf2e71d6660c4f5f3510c89027a048bd/assets/js/modules/adsense/components/common/AdBlockerWarning.js#L56) should be simplified to come from the new `getAdBlockerWarningMessage` selector rather than using its own logic based on site and account setup completion.

## Implementation Brief

Inside `assets/js/modules/adsense/datastore/adblocker.js`
- Add a new selector, `getAdBlockerWarningMessage` with the following logic:
  - if there is no ad blocker, return `null`
  - if the AdSense module is connected and there is an ad blocker, return '_Ad blocker detected, you need to disable it to get the AdSense latest data_.'
  - if the AdSense module is not connected and there is an ad blocker, return '_Ad blocker detected, you need to disable it in order to set up AdSense_.' 
  - Make sure to use i18n for the messages, i.e. 
[...]

PR #4065, opened 2021-09-16
Fix Ad Blocker Warning Message based on modules connected state.

kuasha420:
## Summary

Addresses issue #3208

## Checklist

### DIFF
--- a/assets/js/modules/adsense/datastore/adblocker.js
+++ b/assets/js/modules/adsense/datastore/adblocker.js
@@ -147,7 +147,9 @@ export const selectors = {
 	 * @since n.e.x.t
 	 *
 	 * @param {Object} state Data store's state.
-	 * @return {string|null} The error message or null if ad blocker isn't detected.
+	 * @return {(string|null|undefined)} The error message string if an ad blocker is active,
+	 *                                   `null` if an ad blocker isn't detected,
+	 *                                   `undefined` if ad blocker detection has not completed yet.
 	 */
 	getAdBlockerWarningMessage: Data.createRegistrySelector(
 		( select ) => () => {
--- a/assets/js/modules/adsense/datastore/adblocker.js
+++ b/assets/js/modules/adsense/datastore/adblocker.js
@@ -17,6 +17,11 @@
  */
 
 /**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
+/**
  * External dependencies
  */
 import invariant from 'invariant';
@@ -27,6 +32,7 @@ import { detectAnyAdblocker } from 'just-detect-adblock';
  */
 import Data from 'googlesitekit-data';
 import { MODULES_ADSENSE } from './constants';
+import { CORE_MODULES } from '../../../googlesitekit/modules/datastore/constants';
 
 // Actions
 const CHECK_ADBLOCKER = 'CHECK_ADBLOCKER';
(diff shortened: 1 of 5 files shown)


assets/js/modules/adsense/index.js
```diff
@@ -75,20 +75,17 @@ export const registerModule = ( modules ) => {
 			__( 'AdSense insights through Site Kit', 'google-site-kit' ),
 		],
 		checkRequirements: async ( registry ) => {
-			const adBlockerActive = await registry
+			const adBlockerActiveMsg = await registry
```
tofumatt, line 78: Let's not use abbreviations for variable names like "msg". We can afford the bytes to write `adBlockerActiveMessage` instead 😉 

```js
			const adBlockerActiveMsg = await registry
```

tofumatt: Alternatively: `message` is fine and then you can use it in the object return below 🙂

assets/js/modules/adsense/datastore/adblocker.js
```diff
@@ -134,6 +140,46 @@ export const selectors = {
...
+				return null;
+			}
+
+			const isModuleConnected = select( CORE_MODULES ).isModuleConnected(
+				'adsense'
+			);
+
+			if ( isModuleConnected ) {
+				return __(
+					'Ad blocker detected, you need to disable it to get the AdSense latest data.',
```
tofumatt, line 172: I know this string was originally written this way, but it's awkward. Let's take this opportunity to improve it, because "AdSense latest data" is not very clearly written:

```suggestion
					'Ad blocker detected, you need to disable it to get the latest AdSense data.',
```

assets/js/modules/adsense/datastore/adblocker.js
```diff
@@ -134,6 +140,46 @@ export const selectors = {
...
+					'google-site-kit'
+				);
+			}
+
+			return __(
+				'Ad blocker detected, you need to disable it in order to set up AdSense.',
```
tofumatt, line 178: This string can be clearer too:

```suggestion
				'Ad blocker detected, you need to disable it to set up AdSense.',
```

assets/js/modules/adsense/index.js
```diff
@@ -75,20 +75,17 @@ export const registerModule = ( modules ) => {
...
 				.__experimentalResolveSelect( MODULES_ADSENSE )
-				.isAdBlockerActive();
+				.getAdBlockerWarningMessage();
```
tofumatt, line 80: This is the reason VRT and E2E tests were failing; `resolveSelect` will return a value after the selector's resolver has completed, but because the IB suggested changing the selector used, this returns immediately because `getAdBlockerWarningMessage` doesn't _have_ an associated selector.

There are two options here:

1. continue using `isAdBlockerActive` here but getting the message later on if `isAdBlockerActive`  is true
2. add a (redundant) resolver to `getAdBlockerWarningMessage`

I added a commit that does the former. It's not 100% in line with the IB but still fulfills the ACs and behaves as-intended.

tofumatt (changes requested):
This needs some changes because it doesn't actually behave as-intended—see the issue I mentioned with the `resolveSelect` not working anymore. The E2E and Visual Regression tests are failing as a result.

I'll add a commit that should fix it though.

tofumatt:
@googlebot I consent.

tofumatt (approved):
The rest of this works for me; I think there are a few string changes we should make but I'll leave it for the reviewer to confirm.

I'm moving this to merge review to get another look at the approach though.

aaemnnosttv:
@googlebot I consent.

aaemnnosttv:
I pushed one commit to fix a doc block but otherwise this LGTM 👍

I like the changes to the wording as well but I'll leave this for @felixarntz to sign off on.

felixarntz (approved):

Issue #4052, opened 2021-09-15
Trigger remote survey about Idea Hub usage


felixarntz:
A new remote survey for Idea Hub should be supported, specifically on the plugin side this means that the triggering logic needs to be added. Also the new survey relies on #4051 - that is only a blocker for full testing though, it shouldn't block the implementation of this issue.

The requirements on the Site Kit Service side to trigger this survey are the following:
* You have to be on at least Site Kit version 1.42.0.
* You have to have first activated Idea Hub at least 2 days (48 hours) ago.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The Idea Hub module should internally (on the server-side) introduce a single counter per user that counts the following interactions with the feature:
    * idea dismissal
    * idea save
    * idea draft creation
    * In other words, when any of these actions is performed, the count for the current user should be increased. To clarify, it should be one counter that aggregates all three interactions, we don't need 3 individual counters.
* The counter value for the current user should be provided from PHP to JS so that it can be used there. This can be as simple as possible, e.g. a PHP-to-JS property is preferred unless there's a good reason to use a REST data point.
* In JS: Whenever the current user performs one of the above three actions on the Idea Hub widget, if their current counter value is greater than 5, a survey trigger request for a new trigger `interact_idea_hub` should be issued. If a survey is returned it should be displayed as usual.
    * This requires that JS somehow keeps track of the value too, also when it changes (which is for example relevant if the user makes 5 interactions in a single page load). So maybe a REST data point is more reliable after all, or the JS code would need to "duplicate" and do the counting itself.
* The timeout for `/survey/trigger/` and `/notifications/` requests to the proxy should be increased to 15 seconds.

## Implementation Brief

**The API infrastructure is currently being built. Depending on how quickly this gets implemented, the Site Kit Service part may not be ready yet. Please coordinate testing with @felixarntz.**

### PHP

* Add a new `Google\Site_Kit\Modules\Idea_Hub\Idea_Interaction_Count` class which extends `User_Setting`
    - Slug `googlesitekit_idea-hub_interaction_count`
    - type `integer`
    - default value `0`
    - Add an additional public method `increment` which simply takes the current value, increases it by `1` and saves it
    - Add a test case for this class, mostly for the added method, everything else is covered by the tests for the base class
* Update `Idea_Hub::__construct` to initialize `Idea_Interaction_Count` to a protected `interactionCount` property similar to how is currently done with post name/text/topic classes
[...]

PR #4072, opened 2021-09-17
Enhancement/4052 trigger remote survey

eugene-manuilov:
## Summary

Addresses issue #4052

## Relevant technical choices

* Decided not to create a new hook because there is no sense in it since we would use it only for one component.

## Checklist

### DIFF
--- a/assets/js/components/ModulesList.js
+++ b/assets/js/components/ModulesList.js
@@ -17,6 +17,11 @@
  */
 
 /**
+ * External dependencies
+ */
+import PropTypes from 'prop-types';
+
+/**
  * WordPress dependencies
  */
 import { useCallback } from '@wordpress/element';
@@ -24,15 +29,16 @@ import { useCallback } from '@wordpress/element';
 /**
  * Internal dependencies
  */
-import { trackEvent } from '../util';
 import Data from 'googlesitekit-data';
-import ModulesListItem from './ModulesListItem';
 import { CORE_SITE } from '../googlesitekit/datastore/site/constants';
 import { CORE_MODULES } from '../googlesitekit/modules/datastore/constants';
 import { CORE_LOCATION } from '../googlesitekit/datastore/location/constants';
+import { VIEW_CONTEXT_DASHBOARD } from '../googlesitekit/constants';
+import { trackEvent } from '../util';
+import ModulesListItem from './ModulesListItem';
 const { useSelect, useDispatch } = Data;
 
-function ModulesList( { moduleSlugs } ) {
+export default function ModulesList( { moduleSlugs } ) {
 	const { activateModule } = useDispatch( CORE_MODULES );
 	const { navigateTo } = useDispatch( CORE_LOCATION );
 	const { setInternalServerError } = useDispatch( CORE_SITE );
@@ -53,7 +59,11 @@ function ModulesList( { moduleSlugs } ) {
 				return null;
 			}
 
-			await trackEvent( `${ slug }_setup`, 'module_activate', slug );
+			await trackEvent(
+				`${ VIEW_CONTEXT_DASHBOARD }_authentication-success-notification`,
+				'activate_module',
+				slug
+			);
 
 			// Redirect to ReAuthentication URL
 			navigateTo( response.moduleReauthURL );
@@ -85,6 +95,7 @@ function ModulesList( { moduleSlugs } ) {
 			( module ) => ! module.internal && 0 === module.dependencies.length
 		)
 		.sort( ( a, b ) => a.order - b.order );
+
 	return (
 		<div className="googlesitekit-modules-list">
 			{ modulesToShow.map( ( module ) => (
@@ -98,4 +109,6 @@ function ModulesList( { moduleSlugs } ) {
 	);
 }
 
-export default ModulesList;
+ModulesList.propTypes = {
+	moduleSlug: PropTypes.arrayOf( PropTypes.string ).isRequired,
+};
--- a/assets/js/components/ModulesList.js
+++ b/assets/js/components/ModulesList.js
@@ -110,5 +110,5 @@ export default function ModulesList( { moduleSlugs } ) {
 }
 
 ModulesList.propTypes = {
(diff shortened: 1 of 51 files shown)


assets/js/modules/idea-hub/datastore/module-data.js
```diff
@@ -65,18 +66,48 @@ export const actions = {
...
 				ideaHubData: {
 					lastIdeaPostUpdatedAt,
+					interactionCount,
+				},
+			};
+		}
+		case INCREMENT_INTERACTION_COUNT: {
+			const { ideaHubData = {} } = state;
+
+			let interactionCount = parseInt( ideaHubData.interactionCount );
```
aaemnnosttv, line 101:
```suggestion
			let interactionCount = parseInt( ideaHubData.interactionCount, 10 );
```

assets/js/modules/idea-hub/datastore/module-data.js
```diff
@@ -65,18 +66,48 @@ export const actions = {
...
+			if ( Number.isNaN( interactionCount ) ) {
+				interactionCount = 0;
+			}
```
aaemnnosttv, line 104: We could use `0` as a fallback instead, e.g. `parseInt(...) || 0`?

assets/js/modules/idea-hub/datastore/module-data.test.js
```diff
@@ -0,0 +1,134 @@
...
+	createTestRegistry,
+	unsubscribeFromAll,
+	untilResolved,
+} from '../../../../../tests/js/utils';
+import { enabledFeatures } from '../../../features';
+import { MODULES_IDEA_HUB } from './constants';
+
+describe( 'modules/idea-hub module-data', () => {
+	const baseInfoVar = '_googlesitekitIdeaHub';
+	const baseInfo = {
```
aaemnnosttv, line 32: These look like they were copied from the `core/site` tests. Let's rename these to be more specific to this test.
```suggestion
	const ideaHubDataGlobal = '_googlesitekitIdeaHub';
```
As for `baseInfo` here, it looks like it's only used in the one test for selectors, so how about we inline it there instead?

includes/Modules/Idea_Hub.php
```diff
@@ -106,20 +107,23 @@ final class Idea_Hub extends Module
...
 	 *
+	 * @since n.e.x.t
 	 * @var Post_Idea_Text
 	 */
 	private $post_text_setting;
 
 	/**
 	 * Post_Idea_Topics instance.
 	 *
+	 * @since n.e.x.t
```
aaemnnosttv, line 126: Nice catch! Let's update these to use the correct version though as `n.e.x.t` would not be accurate.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -82,15 +84,19 @@ function DashboardIdeasWidget( props ) {
...
-	const newIdeas = useSelect( ( select ) =>
-		select( MODULES_IDEA_HUB ).getNewIdeas()
+	const [ triggeredSurvey, setTriggeredSurvey ] = useState( false );
+
+	const { newIdeas, savedIdeas, draftIdeas, interactionsCount } = useSelect(
+		( select ) => ( {
+			newIdeas: select( MODULES_IDEA_HUB ).getNewIdeas(),
+			savedIdeas: select( MODULES_IDEA_HUB ).getSavedIdeas(),
+			draftIdeas: select( MODULES_IDEA_HUB ).getDraftPostIdeas(),
+			interactionsCount: select( MODULES_IDEA_HUB ).getInteractionCount(),
```
aaemnnosttv, line 94: Let's make this singular like it is in the selector (and PHP).

```suggestion
			interactionCount: select( MODULES_IDEA_HUB ).getInteractionCount(),
```

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.js
```diff
@@ -82,15 +84,19 @@ function DashboardIdeasWidget( props ) {
 	const { defaultActiveTabIndex, Widget, WidgetReportError } = props;
 
 	const [ trackedWidgetView, setTrackedWidgetView ] = useState( false );
-
-	const newIdeas = useSelect( ( select ) =>
-		select( MODULES_IDEA_HUB ).getNewIdeas()
+	const [ triggeredSurvey, setTriggeredSurvey ] = useState( false );
+
+	const { newIdeas, savedIdeas, draftIdeas, interactionsCount } = useSelect(
```
aaemnnosttv, line 89: This also pending the result of our call later.

assets/js/components/ModulesList.js
```diff
@@ -110,5 +110,5 @@ export default function ModulesList( { moduleSlugs } ) {
 }
 
 ModulesList.propTypes = {
-	moduleSlug: PropTypes.arrayOf( PropTypes.string ).isRequired,
+	moduleSlugs: PropTypes.arrayOf( PropTypes.string ).isRequired,
```
aaemnnosttv, line 113: This just fixes an unrelated prop type error, correct?

aaemnnosttv (commented):
Thanks @eugene-manuilov ! This looks solid to me, just a few minor things to touch up and then this should be good to go 👍

felixarntz (changes requested):
@eugene-manuilov I found one problem with the implementation while testing.

Regarding your question: In order to get a survey to be returned, you need to technically be on Site Kit version 1.42.0 or higher. In order to do that (and have the proxy recognize it), you'll need to e.g. manually update the version numbers in your Site Kit codebase, and then either wait for the refresh token call to trigger or just disconnect/reset and reconnect. Any of those will update the proxy data with the current Site Kit version used on the site.

eugene-manuilov:
@felixarntz @aaemnnosttv going to close this PR because it contains unwanted changes from the develop branch. Please, take a look at #4100. It contains all changes and addresses code review feedback added for this PR.

PR #4083, opened 2021-09-20
SwG PoC: Gutenberg plugin

ChrisAntaki:
## Summary

Addresses issue #

## Checklist


assets/js/modules/subscribe-with-google/components/plugins/AccessSelector.js
```diff
@@ -0,0 +1,84 @@
...
+import { Component } from '@wordpress/element';
+
+export default class AccessSelector extends Component {
+	constructor( props ) {
+		super( props );
+
+		const { select } = global.wp.data;
+		const postMeta = select( 'core/editor' ).getEditedPostAttribute(
+			'meta'
+		);
```
tofumatt, line 32: This should be done with the [`withSelect` higher-order component](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-data/#withselect), or this data will not update when the data store does.

assets/js/modules/subscribe-with-google/components/plugins/AccessSelector.js
```diff
@@ -0,0 +1,84 @@
...
+		this.state = {
+			access: postMeta.sitekit__reader_revenue__access,
+		};
+	}
+
+	setAccess( access ) {
+		// Update post meta field.
+		const { dispatch } = global.wp.data;
```
tofumatt, line 40: Similarly, this should use [`withDispatch`](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-data/#withdispatch) and not the global dispatch.

eugene-manuilov:
@ChrisAntaki this might be helpful to you: https://developer.wordpress.org/block-editor/how-to-guides/metabox/

assets/js/modules/subscribe-with-google/components/plugins/AccessSelector.js
ChrisAntaki: Good call, updated!

assets/js/modules/subscribe-with-google/components/plugins/AccessSelector.js
```diff
@@ -0,0 +1,93 @@
...
+	withDispatch( ( dispatch ) => ( {
+		setAccess: ( access ) =>
+			dispatch( 'core/editor' ).editPost( {
+				meta: { sitekit__reader_revenue__access: access },
+			} ),
+	} ) ),
+	withSelect( ( select ) => ( {
+		access: select( 'core/editor' ).getEditedPostAttribute( 'meta' )
+			.sitekit__reader_revenue__access,
+		products: global.googlesitekit.data.select( STORE_NAME ).getProducts(),
```
ChrisAntaki, line 91: Hey @tofumatt ! Just wondering, do we have a good solution for using Site Kit's `withSelect` for a Gutenberg component? I tried to use Site Kit's `withSelect` alongside WordPress', in this call to `compose`, but then WordPress crashed and mentioned something about two versions of React. What do you think?

ChrisAntaki: cc @aaemnnosttv and @felixarntz

tonyjemba:
checks still failing

assets/js/modules/subscribe-with-google/components/plugins/AccessSelector.js
aaemnnosttv: >  I tried to use Site Kit's withSelect alongside WordPress', in this call to compose, but then WordPress crashed and mentioned something about two versions of React. What do you think?

Hey @ChrisAntaki, not sure if this is still an issue for you but the reason this is due to `withSelect` using hooks internally which requires that the same version of React is used and Site Kit's will always reference the version of React we bundle. So when using within the context of Gutenberg, we'll always need to use `withSelect` via the `wp` global for the same reason we need to use class-based components here.

If we don't already have an issue for using the two data stores together we'll need to create one for that.

ChrisAntaki: Thanks for the explanation, @aaemnnosttv !

Hey @felixarntz , just checking, did you already create an issue for this a while back ?

felixarntz: @aaemnnosttv I created #4107 (and related #4108 + #4109) around this, but it doesn't actually encompass the ability to use stores from both registries in a single component yet I believe. Can you check where the new issue fits in with those?

aaemnnosttv: @felixarntz @ChrisAntaki I opened https://github.com/google/site-kit-wp/issues/4357. With this we would use the Site Kit registry for accessing both our stores and WP core stores – we would still need to use the HOCs via `wp` though as mentioned before.

ChrisAntaki:
This issue causes a JS test failure: https://github.com/google/site-kit-wp/issues/4831

PR #4100, opened 2021-09-21
Enhancement/4052 trigger remote survey

eugene-manuilov:
## Summary

Addresses issue #4052

## Relevant technical choices

* Decided not to create a new hook because there is no sense in it since we would use it only for one component.

## Checklist


includes/Core/Notifications/Notifications.php
```diff
@@ -134,7 +134,12 @@ private function get_rest_routes() {
 								return new WP_REST_Response( array() );
 							}
 
-							$response = wp_remote_get( $endpoint ); // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.wp_remote_get_wp_remote_get
+							$response = wp_remote_get( // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.wp_remote_get_wp_remote_get
+								$endpoint,
+								array(
+									'timeout' => 15, // phpcs:ignore WordPressVIPMinimum.Performance.RemoteRequestTimeout.timeout_timeout
```
felixarntz, line 140: This timeout should also be applied to getting surveys, which is still missing from here and therefore still results in timeout errors.

Let's just add it to the `Google_Proxy::request` method, so that it applies to any of those requests to the proxy. It shouldn't do any harm, and it's safer to have it at 15 than at 5 (of course we should still improve the service performance to make this faster).

felixarntz (changes requested):
@eugene-manuilov This looks good to me, just one crucial thing that's missing.

includes/Core/Notifications/Notifications.php
eugene-manuilov: Looks like i didn't copy over it from the previous PR: https://github.com/google/site-kit-wp/pull/4072/files#diff-e3ac1fdfcae738afa70efaf4637416ebfae905a5fa75bfcbc459d7e172745890R492

I'll add the default timeout to the request method then :+1:

eugene-manuilov: Ok, updated.

aaemnnosttv (approved):
Thanks @eugene-manuilov, this all looks good to me. I tested the items @felixarntz called out in the last review as well and confirmed that the `trigger-survey` requests aren't timing out after 5 sec anymore. The trigger for the interactions is also confirmed to not be invoked on mount.

Issue #4111, opened 2021-09-23
Experimental label too close to Idea Hub name in some places


aaemnnosttv:
## Bug Description

The distance between the Idea Hub module title and the experimental badge is inconsistent, and in some places there is none at all.

## Screenshots

**Settings > Connected Services**
![image](https://user-images.githubusercontent.com/1621608/134498772-4f892b52-4c48-4462-8459-c5a7740e20a2.png)

**Settings > Connect More Services**
![image](https://user-images.githubusercontent.com/1621608/134492184-19b271ea-0e6f-493e-beb8-282390e057c4.png)

**Module setup**
![image](https://user-images.githubusercontent.com/1621608/134492405-e0a593ba-f460-4dfd-9d1e-ebc5ccd91516.png)

**Widget (for reference: 12px)**
![image](https://user-images.githubusercontent.com/1621608/134492503-745c6615-fe9a-467f-9acc-153599cffc15.png)

Also note that the font weight of the badge differs in some places (e.g. settings and module setup) where it inherits a heavier weight rather than remaining consistent.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be 12px between the Idea Hub module title and experimental badge in all instances
    * Note: the `Badge` component is only used for this so that makes it easy to find all instances, but the styles should not be added to the badge component/class without additional context (i.e. not to `.googlesitekit-badge` alone)

## Implementation Brief

* Update `Badge` component to accept additional class names via prop
* Add a new class for the Idea Hub badge, giving it 12px left margin
* Apply the new class on all Idea Hub badge components
* Remove now unnecessary `padding-right` on widget title
* Merge #4115 

### Test Coverage

* No change

## QA Brief

* Check the Idea Hub badges in the following places to make sure they all have consistent spacing with the experimental label
    * Settings > Connect More Services
    * Settings > Connected Services (once activated)
    * Module set up
    * Widget title

## Changelog entry

* N/A

wpdarren:
### QA Update: ✅ 

Checked that the Idea Hub badges in the places below, all have consistent spacing with the experimental label:

Settings > Connect More Services  - [Screenshot](https://user-images.githubusercontent.com/73545194/134640567-d2d1c700-2c2e-4224-a41c-3f894af16eaa.png)
[...]

PR #4115, opened 2021-09-23
Update Idea Hub badges for consistency

aaemnnosttv:
## Summary

Addresses issue #4111

## Checklist

### DIFF
--- a/assets/sass/components/global/_googlesitekit-badge.scss
+++ b/assets/sass/components/global/_googlesitekit-badge.scss
@@ -23,6 +23,7 @@
 	display: inline-block;
 	font-family: $f-primary;
 	font-size: 10px;
+	font-weight: 500;
 	line-height: 1;
 	padding: 6px 12px;
 	text-transform: uppercase;
--- a/assets/sass/components/idea-hub/_googlesitekit-idea-hub-common.scss
+++ b/assets/sass/components/idea-hub/_googlesitekit-idea-hub-common.scss
@@ -1,5 +1,5 @@
 /**
- * Idea Hub styles.
+ * Idea Hub common styles.
  *
  * Site Kit by Google, Copyright 2021 Google LLC
  *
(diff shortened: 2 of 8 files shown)


assets/sass/components/global/_googlesitekit-badge.scss
```diff
@@ -23,6 +23,7 @@
 	display: inline-block;
 	font-family: $f-primary;
 	font-size: 10px;
+	font-weight: 500;
```
aaemnnosttv, line 26: This was added to avoid variation when left undefined where it would be (and was) inherited. In titles this usually left it with a heavier weight and other places it had a normal weight. Looking at similar components in the wild (in GitHub for example), a heavier weight is used, likely for better contrast/readability, especially with darker backgrounds so this makes sense to have on the base class.

assets/sass/components/idea-hub/_googlesitekit-idea-hub-common.scss
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+.googlesitekit-plugin {
+
+	.googlesitekit-idea-hub__badge {
+		margin-left: 12px;
+	}
```
eugene-manuilov, line 23: @aaemnnosttv, unfortunately, it doesn't work well on small devices. I believe `padding-right: 12px` was added on purpose to make sure that the badge won't have an empty margin on the left when it gets wrapped to the next line. See how `develop` looks on small mobiles:


And how your branch:

eugene-manuilov (changes requested):

assets/sass/components/idea-hub/_googlesitekit-idea-hub-common.scss
aaemnnosttv: Good catch! I was wondering why it was done that way 😄  I've updated this one instance to keep the same styles as before without the added class on the badge as an exception but with an added comment to explain.

eugene-manuilov (approved):
LGTM

Issue #4030, opened 2021-09-13
Scaffold DashboardMainApp and DashboardEntityApp


tofumatt:
## Feature Description

The [`GoogleSiteKitDashboard` component](https://github.com/google/site-kit-wp/blob/01faf4805f14baa86d60314106b955c71a0776c1/assets/js/googlesitekit-dashboard.js#L37) is the entry-point to the dashboard and can be set to render differently based on the [`unifiedDashboard ` feature flag](https://github.com/google/site-kit-wp/issues/4029). When the feature flag for `unifiedDashboard` is disabled, the existing `DashboardApp` component should render. When the feature flag for `unifiedDashboard` is enabled, we should instead render either `DashboardMainApp` or `DashboardEntityApp` based on the presence of the `permaLink` query param.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Two new components should be scaffolded in the `assets/js/components` folder: `DashboardMainApp` and `DashboardEntityApp`. They should only render the Site Kit Header (`assets/js/components/Header.js`) component for now.
* The [`GoogleSiteKitDashboard` component](https://github.com/google/site-kit-wp/blob/01faf4805f14baa86d60314106b955c71a0776c1/assets/js/googlesitekit-dashboard.js#L37) should, when `unifiedDashboard` feature flag is enabled: render `DashboardMainApp` when there is no `permaLink` query param and render `DashboardEntityApp` when the `permaLink` query param exists.
* If the `unifiedDashboard` feature flag is not enabled: `GoogleSiteKitDashboard` should behave as it currently does today.

## Implementation Brief

* 

### Test Coverage

* 

### Visual Regression Changes

* 

## QA Brief

* 

## Changelog entry

*

aaemnnosttv:
IB ✅

johnPhillips:
> The GoogleSiteKitDashboard component should, when unifiedDashboard feature flag is enabled: render DashboardMainApp when there is no permaLink query param and render DashboardEntityApp when the permaLink query param exists.

@aaemnnosttv I think we might have missed a step, looking at this more closely. 

It seems [if the `permaLink` query var is present, the PHP layer loads an entirely separate div](https://github.com/google/site-kit-wp/blob/861db7225e67b969e69a3afd7f851016096a7859/includes/Core/Admin/Screens.php#L295): `googlesitekit-dashboard-details`, into which the `DashboardDetailsApp` is loaded, so it never even hits 

[...]

PR #4117, opened 2021-09-24
Scaffold out DashboardMainApp and DashboardEntityApp (4030).

johnPhillips:
## Summary

Addresses issue #4030

## Checklist


tofumatt (changes requested):
I noticed in the IB it says to put these files in folders, eg. `components/dashboard-entity` and `components/dashboard-main`, but I don't think that's necessary as most dashboard files will be shared and there shouldn't even be that many. And eventually we might even want to combine them.

If you want to put them in a folder at all we could use "dashboard", but the ACs for this issue say to put `DashboardMainApp.js` and `DashboardEntityApp.js` right in the `components` folder, so let's do that.

Then we can 🚢

tofumatt (approved):
Awesome, looks good 👍🏻

Issue #4078, opened 2021-09-20
Unified Dashboard: Register existing widgets in new Content area


tofumatt:
## Feature Description

The widgets that should appear in the Content area (created in #4031):

### `mainDashboardContentPrimary` area:
1. Top keywords widget (half-width widget) (`DashboardPopularKeywordsWidget`)
2. Idea Hub widget (half-width) (`DashboardIdeasWidget`)
3. New widget with overall page metrics (*new widget, so shouldn't be registered by this issue*)
4. Modified most popular content widget (`ModulePopularPagesWidget`)

### `entityDashboardContentPrimary` area:
1. Top keywords widget (full width) (DashboardPopularKeywordsWidget)
2. New widget with overall page metrics (*new widget*, so shouldn't be registered by this issue*)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

These widgets should be registered to the following areas and appear on the Main Dashboard and Entity Dashboard respectively, in this order:

### `mainDashboardContentPrimary` area:
* Top keywords widget (half-width widget) (`DashboardPopularKeywordsWidget`)
* Idea Hub widget (half-width) (`DashboardIdeasWidget`)
* Modified most popular content widget (`ModulePopularPagesWidget`)

### `entityDashboardContentPrimary` area:
* Top keywords widget (full width) (DashboardPopularKeywordsWidget)

## Implementation Brief

* 

### Test Coverage

* 

## QA Brief

* 

## Changelog entry

*

eugene-manuilov:
@johnPhillips, sending this ticket back to you to make adjustments discussed in #4079

eugene-manuilov:
Thanks, @johnPhillips. IB :heavy_check_mark:

kuasha420:
QA: :warning: 

I confirmed that. 

* The widgets from the AC has been registered to Main and Entity Dashboard and are showing up. 
* Idea Hub Widget shows the CTA to connect and once connected, shows ideas. 

Things I noticed that needs clarification:

* Padding and margin on half width `DashboardPopularKeywordsWidget` is off, specially the top padding seems unaligned with the idea hub module. 
* If Analytics module is not connected, the `ModulePopularPagesWidget` shows loading skeleton  forever instead of a CTA to activate analytics module. Not sure if it falls outside of the scope this ticket. 

Waiting for clarification. 

CC @eugene-manuilov

PR #4131, opened 2021-09-27
Enhancement/4078 content area widgets

eugene-manuilov:
## Summary

Addresses issue #4078

## Checklist


assets/js/modules/analytics/index.js
```diff
@@ -118,11 +119,11 @@ export const registerWidgets = ( widgets ) => {
 		'analyticsPopularPages',
 		{
 			Component: DashboardPopularPagesWidget,
-			width: widgets.WIDGET_WIDTHS.HALF,
+			width: [ widgets.WIDGET_WIDTHS.HALF, widgets.WIDGET_WIDTHS.FULL ],
```
johnPhillips, line 122: @eugene-manuilov I was wondering: will this mean that this widget could start appearing full width on the current dashboard and cause some regressions? 
If so, maybe we should use the feature flag to conditionally apply them?
Let me know what you think.

eugene-manuilov: Technically, it could appear full width on the current dashboard, but it will happen only if there are no other widgets fit into that row. So, I don't think we need to use the feature flag here.

PR #4150, opened 2021-09-28
SwG PoC: Bulk Edit

ChrisAntaki:
## Summary

Addresses issue #

## Checklist


assets/js/googlesitekit-subscribe-with-google-bulk-edit.js
```diff
@@ -0,0 +1,66 @@
...
+		bulkActionSelector.addEventListener(
+			'change',
+			handleBulkActionSelectorChange
+		);
+	}
+} );
+
+function handleBulkActionSelectorChange( e ) {
+	const hidden = e.target.value !== 'sitekit-swg-access';
+	renderDropdowns( { hidden } );
```
felixarntz, line 61: Instead of re-rendering the dropdowns on every bulk action change (which seems like quite some overhead), I think we could simplify this to only `render` (inject into the DOM) once, in the above `for` loop, and then rely on non-React code to visually hide it? E.g. we could conditionally add a CSS class to hide `sitekit-swg-access-selector-container`.

assets/js/modules/subscribe-with-google/components/posts/AccessSelector.js
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+/**
+ * Internal dependencies
+ */
+import Data from 'googlesitekit-data';
+import { STORE_NAME } from '../../datastore/constants';
+const { useSelect } = Data;
+
+export default function AccessSelector( { hidden, onChange, selectedOption } ) {
```
felixarntz, line 26: So far I don't think this component actually needs `onChange` and `selectedOption` props. It would simplify the code to fully internalize that into this component, basically rely on `useState` to capture the current option and then implement an internal `onChange` that sets the selected option in state.

assets/js/googlesitekit-subscribe-with-google-bulk-edit.js
```diff
@@ -0,0 +1,66 @@
...
+}
+
+function handleChange( { target: { value } } ) {
+	renderDropdowns( { selectedOption: value } );
+}
```
felixarntz, line 66: Related to my comment in `AccessSelector`, I think this function shouldn't be needed. It also relates to my comment right above, I'd argue re-calling `render` here is even more unnecessary than above since this logic could be internalized by `AccessSelector` so that React would take care of it itself.

assets/js/googlesitekit-subscribe-with-google-bulk-edit.js
```diff
@@ -0,0 +1,66 @@
...
+domReady( () => {
+	const bulkActionSelectors = [
+		...document.querySelectorAll(
+			'#bulk-action-selector-top, #bulk-action-selector-bottom'
+		),
+	];
+	for ( const bulkActionSelector of bulkActionSelectors ) {
+		const accessSelectorContainer = document.createElement( 'span' );
+		accessSelectorContainer.classList.add(
+			'sitekit-swg-access-selector-container'
```
felixarntz, line 48: Nit-pick, but we typically prefix things with `googlesitekit` instead of `sitekit`.

assets/js/googlesitekit-subscribe-with-google-bulk-edit.js
```diff
@@ -0,0 +1,66 @@
+/**
+ * WordPress dependencies
+ */
+import domReady from '@wordpress/dom-ready';
+
+/**
+ * External dependencies
+ */
+import { render } from 'react-dom';
```
felixarntz, line 9: This should be imported from `@wordpress/element` instead which wraps React. In general, we should never import anything from React directly unless it's something that `@wordpress/element` does not support (in which case we should question though whether that should be used in the first place).

assets/js/modules/subscribe-with-google/components/posts/AccessSelector.js
```diff
@@ -0,0 +1,58 @@
...
+		<option key={ option } value={ option }>
+			{ option === 'openaccess' ? '— Free —' : option }
+		</option>
+	) );
+
+	return (
+		<select
+			value={ selectedOption }
+			name="sitekit-swg-access-selector"
+			className="sitekit-swg-access-selector"
```
felixarntz, line 51: See above, let's use `googlesitekit`.

includes/Modules/Subscribe_With_Google.php
```diff
@@ -47,7 +47,71 @@ public function register() {
 			return;
 		}
 
-		// TODO: Bring back SwG functionality after #3120 is merged.
+		// Register "access" meta field.
+		register_post_meta(
+			'',
+			'sitekit__reader_revenue__access',
```
felixarntz, line 53: See above, this shouldn't use `sitekit_` as prefix. For settings and post meta, we typically use `googlesitekit_`, there is however a nuance to it:
* Site Kit allows to reset, wiping _basically_ everything related to Site Kit in the WordPress site.
* Resetting is generally regarded as a non-sensitive action. While we _do_ show a warning modal before resetting, Site Kit stores almost no data that wouldn't be easy to get back to after setting up the plugin again afterwards - that is because most of the _actual_ data comes from Google APIs.
* For configuration like this here though, it's different. If you configured 25 posts in a certain way for SwG and then reset Site Kit, you probably don't want to do that again, and you may not even remember how you configured them.
* For the above type of settings/meta, we have a special `googlesitekitpersistent_` prefix - any setting/meta that uses that won't get wiped on reset.

So for this meta field here, I'd suggest we use `googlesitekitpersistent_`.

Last but not least (nit-pick), let's only use single underscores to separate words.

includes/Modules/Subscribe_With_Google.php
```diff
@@ -47,7 +47,71 @@ public function register() {
...
+				'single'       => true,
+				'type'         => 'string',
+			)
+		);
+
+		// Add "Set access to..." bulk edit option.
+		add_filter(
+			'bulk_actions-edit-post',
+			function( $bulk_actions ) {
+				$bulk_actions['sitekit-swg-access'] = __( 'Set access to...', 'google-site-kit' );
```
felixarntz, line 65: Nit-pick, but I think WordPress generally prefers `&hellip:` instead of `...`.

includes/Modules/Subscribe_With_Google.php
```diff
@@ -47,7 +47,71 @@ public function register() {
...
+					if ( $access && 'openaccess' !== $access ) {
+						echo esc_html( $access );
+					} else {
+						esc_html_e( '— Free —', 'google-site-kit' );
+					}
+				}
+			},
+			10,
+			2
+		);
```
felixarntz, line 114: No technical concern with this code, but from a UX perspective I wonder if that is the most appropriate way to show the access information for a post in the table. A whole column takes a lot of space, and while users can hide columns in the UI, I think there may be more subtle ways that would also show this information closer to e.g. the post title. Especially since not every post is expected to have a special state (thinking about sites that would monetize let's say 5 out of 100 posts), I'm not sure this is worth a column.

For Idea Hub for example, we display something additional right _in_ the post title column. Maybe that could be a path.

You probably don't need to change that here unless you feel strongly about that, but I think this may get flagged from a UX perspective eventually.

felixarntz (commented):
Left some feedback. I think there's some room for simplification in the React to non-React JS "communication", but otherwise this looks great.

includes/Modules/Subscribe_With_Google.php
ChrisAntaki: Good to know, updated

includes/Modules/Subscribe_With_Google.php
ChrisAntaki: Sounds good, updated

includes/Modules/Subscribe_With_Google.php
ChrisAntaki: I don't feel strongly either way, just wanna build something you and the Site Kit team are happy with

Issue #4054, opened 2021-09-15
Implement/refactor GA tracking events for plugin setup


felixarntz:
## Feature Description

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* 

## Implementation Brief

* 

### Test Coverage

* 

## QA Brief

* 

## Changelog entry

*

eugene-manuilov:
Thanks, @asvinb. IB almost looks good. There is one thing missing as far as I can tell right now: we need to track new setup events (`start_user_setup` and `start_site_setup`) in the `SetupUsingGCP` component as well.

> * Using assets/js/components/setup/SetupUsingProxy.js,
>   * Within the onButtonClick function,
>     * Implement the start_site_setup and start_user_setup tracking calls as mentioned above for assets/js/components/activation/activation-app.js.

It's also probably worth mentioning that for the `SetupUsingProxy` component we should always use `proxy` as the event label and `custom-oauth` for the `SetupUsingGCP` component because we display these components based on the `isUsingProxy` selector results: https://github.com/google/site-kit-wp/blob/496ebb7b1968ca4f84c1f0aefd6ba5f4ba7271df/assets/js/components/dashboard-splash/DashboardSplashApp.js#L29-L39

> * Use the hasConnectedAdmins selector of the site data store to check if there are connected admins.
> * Use the useMount hook:
>    * if slug is null and there are connected admins, call trackEvent with complete_site_setup as action and passing proxy as label if the result of isUsingProxy is true, else pass custom-oauth as label.

We probably need to introduce a new PHP-to-JS variable that will determine how many connected admins we have. It should work the same way as `hasConnectedAdmins` works, but return the number of admins instead of `true/false`. cc @felixarntz

asvinb:
@eugene-manuilov IB updated to include `SetupUsingGCP`. For the connected admins, I think we can use the `hasMultipleAdmins` selector which returns `false` if there's only 1 admin which is pretty much what we want :)
https://github.com/google/site-kit-wp/blob/43e3866a3a5a523f12ebb27938cc178a6e281a02/assets/js/googlesitekit/datastore/site/connection.js#L178

eugene-manuilov:
Actually, yes, you are right, @asvinb, we can use the `hasMultipleAdmins` selector :+1:. IB :heavy_check_mark:

kuasha420:
QA: :warning:  

The following tracking events are not firing consistently. 

[...]

PR #4162, opened 2021-09-29
 Implement/refactor GA tracking events for plugin setup (4054).

johnPhillips:
## Summary

Addresses issue #4054

## Checklist

### DIFF
--- a/assets/js/components/activation/activation-app.js
+++ b/assets/js/components/activation/activation-app.js
@@ -17,6 +17,11 @@
  */
 
 /**
+ * External dependencies
+ */
+import { useMount } from 'react-use';
+
+/**
  * WordPress dependencies
  */
 import { Fragment, useCallback } from '@wordpress/element';
@@ -26,6 +31,7 @@ import { __ } from '@wordpress/i18n';
  * Internal dependencies
  */
 import Data from 'googlesitekit-data';
+import { VIEW_CONTEXT_ACTIVATION } from '../../googlesitekit/constants';
 import { trackEvent } from '../../util';
 import { ActivationMain } from './activation-main';
 import NotificationCounter from '../legacy-notifications/notification-counter';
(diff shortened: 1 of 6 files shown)


assets/js/components/legacy-setup/SetupUsingGCP.js
```diff
@@ -389,6 +411,8 @@ export default compose(
 					notification: 'authentication_success',
 				}
 			),
+			isUsingProxy: select( CORE_SITE ).isUsingProxy(),
```
tofumatt, line 414: This component is only every used, as I understood things, when the proxy is _not_ in use, so there shouldn't be a need to use this selector (or the subsequent ternary logic in the component) here.

If this component is being used we're definitely **not** using the proxy, so we can ignore these checks.

tofumatt (changes requested):
Looks like there's an issue with some `useSelect` calls here, but also I have some questions about our usage of `isUsingProxy()` selectors in the GCP Setup component.

assets/js/components/activation/activation-app.js
```diff
@@ -52,6 +58,16 @@ export function ActivationApp() {
 	const canViewDashboard = useSelect( ( select ) =>
 		select( CORE_USER ).hasCapability( PERMISSION_VIEW_DASHBOARD )
 	);
+	const isConnected = useSelect( ( select ) => {
+		select( CORE_SITE ).isConnected();
+	} );
+	const isUsingProxy = useSelect( ( select ) => {
+		select( CORE_SITE ).isUsingProxy();
+	} );
```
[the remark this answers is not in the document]
johnPhillips: Eek - good catch!

assets/js/components/legacy-setup/SetupUsingGCP.js
johnPhillips: Hmm... This might be one for @asvinb to advise on as he wrote the IB 🤔 . I'm happy to admit this falls outside of my knowledge.
I'll make the adjustments you suggest, but it might be worth checking with Asvin in due course if there was a reason why the IB suggested this? 👍

tofumatt:
@googlebot I consent.

Issue #3304, opened 2021-05-06
Admin Settings: Tracking section creates layout shift on load


aaemnnosttv:
## Bug Description

The admin settings section for Tracking contains a single component for controlling the user's opt-in to anonymous usage tracking. In most cases the component shows as expected, but under certain conditions can result in a layout shift due to an unhandled loading state.

https://user-images.githubusercontent.com/1621608/117275969-435d8900-ae67-11eb-836d-f52e7a330a67.mp4

## Steps to reproduce

1. Clear session storage to wipe any existing request cache
2. Navigate to the settings page and wait 1 or more seconds before switching to the Admin Settings tab
3. Notice the contents of the Tracking section are blank for some period of time while loading

## Additional Context

This is due to preloading of REST routes, for which `core/user/data/tracking` is included, however the preloaded data is only valid for a very short period of time, after which a request will be made as usual. That combined with the `OptIn` component's behavior to render nothing when loading create this layout shift/flicker on load.

https://github.com/google/site-kit-wp/blob/a194a8f034d72bf2776746399182b3c7f1068662/assets/js/components/OptIn.js#L61-L63

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `OptIn` component should render a loading state using a `PreviewBlock` as a placeholder of equal size for the content of the component while it is loading

## Implementation Brief

* 

### Test Coverage

* 

### Visual Regression Changes

* 

## QA Brief

* 

## Changelog entry

*

aaemnnosttv:
Anything to add/change here @felixarntz ?

felixarntz:
@aaemnnosttv Good catch! Should we maybe use a progress bar like we have in setup flows instead of `PreviewBlock`? We only use `PreviewBlock` for widgets, not in setup/settings areas, so that would be a bit inconsistent.

aaemnnosttv:
@felixarntz I thought about the progress bar too, but that seems a bit busy for what will eventually be a single checkbox with some text. I think using a `PreviewBlock` would be more appropriate because it fills a specific amount of space whereas I think the progress bar may still shift the layout without additional wrapping/styles.

[...]

PR #4163, opened 2021-09-29
 Admin Settings: Tracking section creates layout shift on load (3304).

johnPhillips:
## Summary

Addresses issue #3304

## Checklist


assets/js/components/settings/SettingsPlugin.js
```diff
@@ -65,6 +65,9 @@ export default function SettingsPlugin() {
 										checked={ showAdminBar }
 										onChange={ onAdminBarToggle }
 										disabled={ showAdminBar === undefined }
+										loading={
+											typeof showAdminBar !== 'boolean'
+										}
```
aaemnnosttv, line 70: Let's use the same condition here that we use for the `disabled` prop above as it's essentially the same logic.
```suggestion
										loading={ showAdminBar === undefined }
```

assets/sass/vendor/_mdc-checkbox.scss
```diff
@@ -31,4 +31,8 @@
 		width: 100%;
 		z-index: 1;
 	}
+
+	.spinner {
+		margin: 0;
+	}
```
aaemnnosttv, line 37: I realize this was in the IB but let's apply this using an inline style (which are already being used within `Spinner` otherwise I'd opt for a utility class here). We can just merge the given `style` object with the inner one.

This stylesheet is more for calibrating the material styles for SK and most exceptions that we might need for it should be doable with utilities where needed 👍

assets/js/components/Checkbox.js
```diff
@@ -16,12 +16,22 @@
  * limitations under the License.
  */
 
+/**
+ * WordPress dependencies
```
aaemnnosttv, line 20: Please move this group between external and internal groups below 🙂

aaemnnosttv (changes requested):
Thanks @johnPhillips, just a few minor tweaks here and then this will be good to go!

assets/sass/vendor/_mdc-checkbox.scss
johnPhillips: I think that's the method I originally wrote in the IB but I changed it after this feedback from @eugene-manuilov: https://github.com/google/site-kit-wp/issues/3304#issuecomment-920241907

CCing him in case he had any specific concerns.

I'll switch it to use the `style` prop for now 👍

aaemnnosttv: We generally don't use inline styles in components although there are a few instances left. Rather than unpacking the existing inline styles on `Spinner` here I think it's okay to extend them since they'd take precedence over what would be defined in a stylesheet. I'll take a note to refactor the `Spinner` later 👍

aaemnnosttv (approved):
Great, thanks @johnPhillips !

Issue #4086, opened 2021-09-20
Ensure Analytics widgets only show "Gathering data" box when API is actually gathering data


felixarntz:
Currently, Site Kit widgets are configured to show the "Analytics gathering data" box whenever the API returns a response with all 0s. This is somewhat confusing and an incorrect approach, since it could also mean that the site actually has 0 data, even though data is already being aggregated properly.

This should be adjusted, and the "Analytics gathering data" box should only be shown for 0 data if the API is actually _likely_ gathering data - _likely_ since this needs to be based on heuristics as there is no API flag that clearly states whether or not data is now ready to be served.

Related: #4085, #4087

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The Analytics JS store should offer a new parameter-less selector `isGatheringData()` which issues a GA report request to the API only to see if it already has gathered any data, returning a boolean or `undefined` (on load).
    * The request must be for the `ga:date` dimension.
    * The request should preferably match another already existing request that is fired in the Site Kit dashboard, just to not result in yet another request. Of course it is impossible to reliably prevent that, but let's make sure it uses the same report arguments like another request that is typically fired. For example, we may need to also take the current date range into account in this request.
    * The selector should return `true` if the report does not contain any `data[0].data.rows` / is empty. As soon as there is at least one row returned, the selector should return `false`.
        * While Site Kit's Analytics requests always have the `include_empty_rows` argument enabled, it looks like that doesn't apply when there are no rows at all. When testing with the current Site Kit codebase for a new site, `data[0].data.rows` was `undefined`, so it looks like this works even with that argument enabled.
* All Analytics-based widgets (including the applicable AdSense-focused widgets) should, in case that it would currently display the "zero data" UI, also check if Analytics `isGatheringData` - only if that's the case, the widget should display the "zero data" UI. Otherwise it should display the regular widget content, actually "showing the 0s".

## Implementation Brief

* Add a new selector `isGatheringData()` to the `analytics` module:
  * Get the current `startDate` and `endDate` values using the `getDateRangeDates` selector.
  * Get the current entity URL using the `getCurrentEntityURL` selector.
  * Get the result of the `getReport` selector call with the following arguments:
    * `dimensions` should be `[ 'ga:date' ]`
    * `metrics` should be `[ { expression: 'ga:users' } ]`
    * `startDate` should have the `startDate` variable value
    * `endDate` should have the `envDate` variable value
[...]

PR #4186, opened 2021-10-04
Enhancement/4092 analytics gathering data

kuasha420:
## Summary

Addresses issue #4086

## Checklist

### DIFF
--- a/assets/js/modules/analytics/datastore/report.js
+++ b/assets/js/modules/analytics/datastore/report.js
@@ -32,21 +32,24 @@ import { __ } from '@wordpress/i18n';
  */
 import API from 'googlesitekit-api';
 import Data from 'googlesitekit-data';
-import { MODULES_ANALYTICS } from './constants';
-import { stringifyObject } from '../../../util';
 import { createFetchStore } from '../../../googlesitekit/data/create-fetch-store';
+import { CORE_SITE } from '../../../googlesitekit/datastore/site/constants';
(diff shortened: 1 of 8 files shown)


assets/js/modules/analytics/datastore/report.js
```diff
@@ -274,6 +277,42 @@ const baseSelectors = {
...
+			args.url = url;
+		}
+
+		const report = select( MODULES_ANALYTICS ).getReport( args );
+
+		if ( report === undefined ) {
+			return undefined;
+		}
+
+		if ( Array.isArray( report ) && report.length === 0 ) {
```
eugene-manuilov, line 310: I think the condition here should be if **not** an array **or** an empty array then return `true`.

eugene-manuilov (changes requested):
Thanks, @kuasha420. This looks solid to me. Have one comment and then it should be good to go.

assets/js/modules/analytics/datastore/report.js
kuasha420: @eugene-manuilov really? Is my reading comprehension this bad or the IB is wrong? Can you have a look at the IB of this ticket?

eugene-manuilov: @kuasha420 Yes, IB says that it should only check whether it is an empty array, but we should also return `true` if the report data is `NULL` which is possible if I remember correctly.

Could you please check failed e2e and vrt checks?

eugene-manuilov (approved):
LGTM

PR #4221, opened 2021-10-13
Enhancement/4086 fix dashboard issues

eugene-manuilov:
## Summary

Addresses issue #4086

## Checklist


eugene-manuilov:
@felixarntz i think it should be good now. Please, take a look.

felixarntz:
@eugene-manuilov 
> I left it intentionally because it contains a call-to-action link to create new goals. I thought that it was more useful to users because they would be able to create new goals by clicking one link rather than observing zeros.

Hmm, I see your point, but it also makes it inconsistent. By what you're saying, we could also show it when there _is_ data, just to provide users with a link to add more goals. I'd prefer to make it consistent and show the 0s if there are already any goals set. We're not showing it for the "Data unavailable" state either, which I think is correct.

Issue #3807, opened 2021-08-05
Add support for enabling GA4 for UA-only Analytics users


aaemnnosttv:
## Feature Description

In https://github.com/google/site-kit-wp/issues/3586 we changed the settings edit view to not show GA4 fields for sites which already had Analytics connected using classic UA Analytics which has been available in Site Kit since the very beginning. The GA4 fields were removed to not force users to set up GA4 just to make a change to the module settings.

In doing so, the experience for these users was diminished as there is no clear path to enable GA4 support if they want it without disconnecting and reconnecting the module.

Users who only have classic UA Analytics configured in Site Kit should have the option to enable support for GA4 from the settings edit view.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* For sites which already have GA4 set up, the Analytics settings edit UI should be reorganized as in [this Figma design](https://www.figma.com/file/VILRieNSLg2DOVyEgFBkXe/%5BMVP%5D-Generic-Dashboard?node-id=295%3A1535), with the GA4 property dropdown and snippet toggle more clearly separated from the UA ones.
* For sites which don't have GA4 set up yet, and if GA4 "works" (see `isAdminAPIWorking`), the Analytics settings edit UI should look like in [this Figma design](https://www.figma.com/file/VILRieNSLg2DOVyEgFBkXe/%5BMVP%5D-Generic-Dashboard?node-id=795%3A4154), with a disabled GA4 property dropdown. The toggle below should allow to "enable" the GA4 part of the UI, at which point the UI changes to the one in the above bullet point (e.g. the toggle turns into the snippet toggle).
    * In other words, it shouldn't be possible to "disable" that UI again in the same screen. If you decide you don't want to set up GA4, you can still cancel out or just leave the page.
    * Preferably, the GA4 dropdown should not have a value as long as it is disabled and only run its typical prepopulate logic once it is enabled.
    * The "Learn more" link should point to `https://sitekit.withgoogle.com/documentation/ga4-analytics-property/`.
    * Once the GA4 UI is enabled here, saving the settings should also require completing the GA4 configuration, i.e. it should run the typical validation logic for checking GA4 property ID and web data stream ID, including creation of a new web data stream if it doesn't exist yet.

#### Analytics `SettingsForm`

* The fields should be reorganized in all cases to group UA fields and GA4 fields separately, divided by a horizontal rule
    - UA fields (account, property, view), with the snippet control moved onto the row below
    - line
[...]

Issue #3775, opened 2021-07-28
Remove ga4setup feature flag


eugene-manuilov:
## Feature Description

The GA4 support has been rolled out a few releases before but the `ga4setup` feature flag is still used across our codebase. While this is not a big deal, we need to remove it from our codebase and make a functionality that has been hidden by that feature flag work unconditionally.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `ga4setup` feature flag should be removed
* All code that has been activated when the `ga4setup` feature flag is enabled should work unconditionally now.
* All code that has been deactivated when the `ga4setup` feature flag is enabled should be removed.

## Implementation Brief
* Update `feature-flags.json` to remove `ga4setup`
* Search for the `ga4setup` string within the code base and review code where the `ga4setup` is present. Any logic/component which was rendered conditionally when the `ga4setup` feature was enabled should be rendered unconditionally now. Any code which was rendered when the feature flag was disabled should be removed.
* Use of the `useFeature` hook with `ga4setup` as a parameter should be removed.
* Stories using the `ga4setup` feature flag should be updated as well to no longer rely on it.
* Changes should apply to PHP files as well. For e.g `includes/Core/Modules/Modules.php` where the `Analytics_4` should be loaded unconditionally.

### Test Coverage
* No new tests to be added.

### Visual Regression Changes
* N/A

## QA Brief

* Check code changes are done as per AC
    * Pay attention to user-facing code behind or in front of flags
* Check all tests are passing
* Check modified stories are not broken
_Note "Analytics Module/Settings" are already crashing on develop:_ https://google.github.io/site-kit-wp/storybook/develop/?path=/story/analytics-module-settings--no-tag-gtm-property-w-access

## Changelog entry

* N/A

eugene-manuilov:
IB :heavy_check_mark:

danielgent:
I think this could well go over 7 hours.  I've only removed the flag from a few files and there are so many failing tests.  **Lots** of endpoints need mocking, many of which haven't been done before (so may involve manual inspection of endpoints)

danielgent:
Got to the point where there are a few failing tests and e2es. 3hrs 40 mins spent so far. I can pick this up at a later date

danielgent:
[...]

PR #4229, opened 2021-10-15
Enhancement/3807 enable ga4 support

eugene-manuilov:
## Summary

Addresses issue #3807

## Relevant technical choices

* Had to rework how settings are canceled to be able to roll back GA4 settings when UA settings are canceled.
* Removed the `SettingsForm.stories.js` file because it doesn't make any sense.

## Checklist


assets/js/modules/analytics-4/components/common/PropertyNotice.js
```diff
@@ -34,7 +34,7 @@ import SettingsNotice, {
 } from '../../../../components/SettingsNotice';
 import Link from '../../../../components/Link';
 
-export default function GA4PropertyNotice( { notice, children } ) {
+export default function PropertyNotice( { notice, children } ) {
```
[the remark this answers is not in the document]
felixarntz: My reasoning here is that for now the `analytics-4` module is a bit of a "future-compatible" early version, since at some point it will be a standalone module. Right now though, it's only facilitating a few functionalities that are still all controlled in the `analytics` module. And I'd prefer if we only put those things _into_ the `analytics-4` module right now, which are `analytics`-agnostic and only rely on `analytics-4`'s own features.

assets/js/modules/analytics/components/common/AccountCreate/index.js
```diff
@@ -196,7 +195,6 @@ export default function AccountCreate() {
 
 	return (
 		<div>
-			{ ! isGA4enabled && <GA4Notice /> }
```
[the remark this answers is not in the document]
felixarntz: @eugene-manuilov Right, that's in the separate issue though. We typically remove the parts of code which are behind the `! enabled` check for such feature flags as part of that issue, so I think we should do that in that other issue too.

assets/js/modules/analytics-4/components/common/UseSnippetSwitch.js
```diff
@@ -0,0 +1,80 @@
...
+	}, [ useSnippet, setUseSnippet ] );
+
+	if ( useSnippet === undefined ) {
+		return null;
+	}
+
+	return (
+		<div className="googlesitekit-analytics-usesnippet">
+			<Switch
+				label={ __( 'Place GA4 code on your site', 'google-site-kit' ) }
```
felixarntz, line 55: Let's use "Place Google Analytics 4 code" here, to match the text in the UA snippet switch.

assets/sass/components/settings/_googlesitekit-settings-module.scss
```diff
@@ -127,6 +127,16 @@
...
 
+	.googlesitekit-settings-module__fields-group {
+		border-bottom: 1px solid $c-mercury;
+		margin-bottom: 1em;
+
+		&:last-child {
+			border-bottom: none;
+			margin-bottom: 0;
+		}
+	}
```
felixarntz, line 138: This is probably more related to a problem in the component code, but I see two Storybook stories where there is still an undesired double border now:
* https://google.github.io/site-kit-wp/storybook/pull/4229/?path=/story/analytics-module-settings--existing-tag-w-access-gtm-property-w-access
* https://google.github.io/site-kit-wp/storybook/pull/4229/?path=/story/analytics-module-settings--existing-tag-w-access-gtm-property-w-o-access

Could you check those?

felixarntz (commented):
@eugene-manuilov This mostly looks good now, a few remaining comments, about a bit of a mix up/overlap with #3775, as well as some minor things and Storybook.

One overarching thing: Would it be possible to trigger the property matching logic we typically run in the moment the user activates the GA4 toggle? When I tested this PR, it just showed me the empty GA4 dropdown, although I have properties that already apply to my current site. Those should preferably be preselected like we would also do in the setup flow.

assets/js/modules/analytics/components/common/AccountCreate/index.js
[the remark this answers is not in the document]
eugene-manuilov: Ok, reverted.

assets/js/modules/analytics-4/datastore/properties.js
```diff
@@ -85,6 +85,10 @@ const fetchGetPropertiesStore = createFetchStore( {
 		);
 	},
 	reducerCallback( state, properties, { accountID } ) {
+		if ( ! Array.isArray( properties ) ) {
+			return state;
+		}
```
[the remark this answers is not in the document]
eugene-manuilov: Hm... i don't remember why i added it... Removed.

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
```diff
@@ -0,0 +1,103 @@
...
+						label={ __(
+							'Google Analytics 4 Property',
+							'google-site-kit'
+						) }
+						value="disabled"
+						disabled
+						enhanced
+						outlined
+					>
+						<Option value="disabled" />
```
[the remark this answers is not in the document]
eugene-manuilov: Updated it to display the matching ga4 property. If matched property is not found, nothing will be displayed.

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
```diff
@@ -0,0 +1,103 @@
...
+					</Select>
+				) }
+			</div>
+
+			{ isDisabled && <GA4ActivateSwitch /> }
+
+			<div className="googlesitekit-setup-module__inputs googlesitekit-setup-module__inputs--multiline">
+				{ ! isDisabled && <UseSnippetSwitch /> }
+			</div>
```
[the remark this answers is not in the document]
eugene-manuilov: Don't remember why i did it this way. Updated to put the check around the entire container.

assets/js/modules/analytics/components/settings/SettingsForm.stories.js
```diff
@@ -1,157 +0,0 @@
-/**
```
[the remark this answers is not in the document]
eugene-manuilov: I decided to remove it because after i took a look at it, i hadn't seen any value in this file. It contains just one story that shows the same state of the form that is shown by similar legacy stories. Furthermore with all changes for this ticket the one story that exists in this file doesn't make any sense anymore because we don't have one component for UA and GA4 switches anymore.

I know that we plan to use storybook v6 approach only, but from my point of view it doesn't mean that we need to add storybook v6 files "just because". I think that if we want to start using v6 for the SettingsForm component, then we need to move all settings-form related stories from the legacy file to this file all together and only then add new stories to this file. 

Reverted for now. I can delete it again if you agree with me 😅

assets/js/modules/analytics-4/components/common/UseSnippetSwitch.js
eugene-manuilov: Updated.

assets/js/modules/analytics-4/datastore/properties.js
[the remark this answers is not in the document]
eugene-manuilov: Ah... remembered! To fix storybook issues 😄, I'll use another approach to address those issues.

assets/sass/components/settings/_googlesitekit-settings-module.scss
eugene-manuilov: Ok, I think it should be good now.

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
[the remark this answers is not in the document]
felixarntz: Overall this is a great solution. However, I'm a little wary of having all this extra code in `GA4SettingsControls`, especially since it happens together with the `! isAdminAPIWorking` check. Can we ensure that these GA4 API-based selectors are only called if `isAdminAPIWorking`?

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
```diff
@@ -0,0 +1,130 @@
...
+	);
+
+	const { matchAccountProperty, setPropertyID } = useDispatch(
+		MODULES_ANALYTICS_4
+	);
+
+	useEffect( () => {
+		setMatchedProperty( undefined );
+		matchAccountProperty( accountID ).then( setMatchedProperty );
+	}, [ accountID, setMatchedProperty, matchAccountProperty ] );
```
felixarntz, line 71: See my other comment in https://github.com/google/site-kit-wp/pull/4229#discussion_r741547616, I think we should also account for `isAdminAPIWorking` here and only trigger the logic if it does work.

assets/js/modules/analytics/components/settings/SettingsForm.js
```diff
@@ -89,75 +57,36 @@ export default function SettingsForm() {
...
-				<PropertySelect />
-				<ProfileSelect />
+			<div className="googlesitekit-settings-module__fields-group">
+				<SettingsControls />
 			</div>
 
-			{ profileID === PROFILE_CREATE && (
-				<div className="googlesitekit-setup-module__inputs googlesitekit-setup-module__inputs--multiline">
-					<ProfileNameTextField />
+			{ isValidAccountID( accountID ) && (
```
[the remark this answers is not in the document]
felixarntz: Great - now that #3775 is already merged, this is good to be removed.

assets/js/modules/analytics/components/settings/SettingsForm.stories.js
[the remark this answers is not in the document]
felixarntz: It makes sense to not duplicate stories, however in that case we should rather remove that story from the file with the legacy stories. Preferably we should of course eventually migrate everything, but that's a very time-consuming effort. Definitely worth doing if at any time there's less work on Site Kit ... which is hard to get 🙃  I agree it's a bit more messy while we're in the transition state with some stories in the legacy files and others next to the component, but it's a transition state, so it should be fine - even if it's a _long_ transition state.

For now, I think we should migrate stories to the new format on demand, e.g. when adding a new one or substantially enhancing an existing one where it would require notable changes even in the legacy version. Since the story here has already been migrated we definitely shouldn't remove it. If there is a matching story among the legacy stories, let's rather remove that (and maybe add a comment that the story is already covered in its appropriate CSF file).

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
```diff
@@ -0,0 +1,130 @@
...
+
+	const onActivate = useCallback( () => {
+		setPropertyID( matchedProperty?._id );
```
felixarntz, line 74: We probably need to set the web data stream ID here correctly too? See my overarching review comment / screenshot.

felixarntz (commented):
This mostly looks good now, except for one thing that's not yet working as expected: It currently prompts me for the permissions to create a web data stream for my GA4 property even though there already is one.
Screen Shot 2021-11-02 at 5 39 05 PM

I also have one other comment (https://github.com/google/site-kit-wp/pull/4229#pullrequestreview-796024815) worth reconsidering.

assets/js/modules/analytics/components/settings/SettingsForm.stories.js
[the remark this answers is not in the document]
eugene-manuilov: Ok, i think it would be better not to touch stories in this ticket because we have limited time. We can get to it later in another ticket, i think 🤔. Ok?

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
eugene-manuilov: @felixarntz, I have added a check to the `useEffect` hook to make sure `isAdminAPIWorking` is true. I like this approach better than moving this logic to the `onActivate` callback because if we do it, there will be a lag between actual activation and the end of the property matching process which is worse user experience.

assets/js/modules/analytics/components/settings/SettingsForm.js
```diff
@@ -84,74 +57,30 @@ export default function SettingsForm() {
...
-				<PropertySelect />
-				<ProfileSelect />
-			</div>
-
-			{ profileID === PROFILE_CREATE && (
-				<div className="googlesitekit-setup-module__inputs googlesitekit-setup-module__inputs--multiline">
-					<ProfileNameTextField />
-				</div>
-			) }
+			<SettingsControls />
```
aaemnnosttv, line 73: Nice 👌

assets/js/modules/analytics/components/settings/SettingsView.js
```diff
@@ -109,6 +114,11 @@ export default function SettingsView() {
 		} )
 	);
 
+	const { setValues } = useDispatch( CORE_FORMS );
+	useMount( () => {
+		setValues( FORM_SETUP, { enableGA4: undefined } );
```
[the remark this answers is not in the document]
aaemnnosttv: This should be handled by `rollbackChanges` https://github.com/google/site-kit-wp/pull/4229/files#diff-fae8a9fa73da5c087c836c0f21fb09033bc08d1fc32fe0c1bb6476f2a36d36bcR107

Do we need to add the same to the other rollback function?

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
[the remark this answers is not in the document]
eugene-manuilov: Sure, added it.

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
eugene-manuilov: Oups... good catch. Fixed.

assets/sass/components/settings/_googlesitekit-settings-module.scss
```diff
@@ -127,6 +127,16 @@
 		margin: 0 0 8px 0;
 	}
 
+	.googlesitekit-settings-module__fields-group {
+		border-bottom: 1px solid $c-mercury;
```
aaemnnosttv, line 131: Maybe an added `<hr>` would be more appropriate? I think it might also be a good idea to apply the same max-width (1098) to it as we do for the GA4 notice to keep it from spanning the entire area. What do you think?

aaemnnosttv (dismissed):
Thanks @eugene-manuilov ! This looks great overall, just a few things I found left to address.

One question: if the account does not have a GA4 property when activating GA4, should the property select not be selecting property_create automatically? When tested, it remained empty for me.

image

assets/js/googlesitekit/data/create-settings-store.js
```diff
@@ -400,6 +400,25 @@ export function makeDefaultSubmitChanges( slug, storeName ) {
...
+/**
+ * Creates a default rollbackChanges control function.
+ *
+ * @since n.e.x.t
+ *
+ * @param {string} slug      Module slug.
+ * @param {string} storeName Datastore slug.
+ * @return {Function} Control function to rollback settings changes.
+ */
+export function makeDefaultRollbackChanges( slug, storeName ) {
```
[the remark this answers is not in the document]
eugene-manuilov: Yep, not needed... Decided to add it for consistency with the `makeDefaultSubmitChanges` function. Updated.

assets/js/googlesitekit/modules/create-submit-changes-store.js
```diff
@@ -82,6 +87,21 @@ export function createSubmitChangesStore( {
 				payload: {},
 			};
 
+			return result;
+		},
+		/**
+		 * Rollbacks changes.
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/googlesitekit/modules/datastore/settings.js
```diff
@@ -39,16 +41,36 @@ export const actions = {
...
-				payload: { slug },
 				type: SUBMIT_MODULE_CHANGES,
+				payload: { slug },
 			};
-		} )();
-	},
+		}
+	),
+	/**
+	 * Rollbacks all changes for a module.
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/analytics-4/components/common/UseSnippetSwitch.js
```diff
@@ -0,0 +1,83 @@
...
+					</span>
+				) }
+				{ ! useSnippet && (
+					<span>
+						{ __(
+							'Site Kit will not add the GA4 code to your site.',
+							'google-site-kit'
+						) }
+					</span>
+				) }{ ' ' }
```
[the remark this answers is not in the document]
eugene-manuilov: Yes, seems like it is not needed. Updated.

assets/js/modules/analytics/components/common/GA4ActivateSwitch.js
```diff
@@ -0,0 +1,77 @@
+/**
+ * Analytics Use Snippet Switch component.
```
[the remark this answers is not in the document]
eugene-manuilov: Updated.

assets/js/modules/analytics/components/common/GA4ActivateSwitch.js
```diff
@@ -0,0 +1,77 @@
...
+ */
+import Data from 'googlesitekit-data';
+import { CORE_FORMS } from '../../../../googlesitekit/datastore/forms/constants';
+import { FORM_SETUP } from '../../datastore/constants';
+import Switch from '../../../../components/Switch';
+import Link from '../../../../components/Link';
+const { useSelect, useDispatch } = Data;
+
+export default function GA4ActivateSwitch( props ) {
+	const { onActivate = () => {} } = props;
```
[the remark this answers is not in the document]
eugene-manuilov: Updated the `onChange` callback to check if the `onActivate` is a function before calling it.

assets/sass/components/settings/_googlesitekit-settings-module.scss
eugene-manuilov: I considered it at the beginning but it didn't work well.

assets/js/modules/analytics/components/settings/SettingsView.js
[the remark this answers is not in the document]
aaemnnosttv: This still looks like it shouldn't be necessary. @eugene-manuilov can you confirm?

assets/js/modules/analytics/components/common/GA4ActivateSwitch.js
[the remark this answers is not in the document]
aaemnnosttv: Looks like maybe this didn't get pushed?

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
```diff
@@ -0,0 +1,170 @@
...
+					/>
+				) }
+				{ isDisabled && (
+					<Select
+						className="googlesitekit-analytics__select-property"
+						label={ __(
+							'Google Analytics 4 Property',
+							'google-site-kit'
+						) }
+						value={ matchedProperty?._id || '' }
```
aaemnnosttv, line 141: The value isn't really important here, it should always be "selected" though because otherwise the floating label moves into the input area and is cut off.
```suggestion
						value="1"
```

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
```diff
@@ -0,0 +1,170 @@
...
+						disabled
+						enhanced
+						outlined
+					>
+						<Option value={ matchedProperty?._id || '' }>
```
aaemnnosttv, line 146: This makes it so the one value is always selected.
```suggestion
						<Option value="1">
```

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
```diff
@@ -0,0 +1,170 @@
...
+		matchedProperty,
+		matchedWebDataStream,
+		setPropertyID,
+		setWebDataStreamID,
+		setMeasurementID,
+	] );
+
+	if ( ! isAdminAPIWorking ) {
+		return null;
+	}
```
aaemnnosttv, line 119: This check results in a layout shift when the edit screen is loading which would be good to avoid. How about instead of returning `null` here, we conditionally apply a `visibility: hidden` style while `isAdminAPIWorking !== true` ?

I would even be open to rendering the div at 50% opacity + disabled until the API is confirmed to be working to show that something is there rather than empty space, but either of these would be preferable to the shift we have now.

Thoughts @felixarntz ?

aaemnnosttv (commented):
Thanks @eugene-manuilov just a few final comments from me, otherwise this is very close I think.

One question for @felixarntz about a UX detail here as well.

assets/js/modules/analytics/components/settings/SettingsView.js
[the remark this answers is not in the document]
eugene-manuilov: @aaemnnosttv, actually, you are right, @aaemnnosttv. Updated.

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
eugene-manuilov: @aaemnnosttv see this comment: https://github.com/google/site-kit-wp/pull/4229#discussion_r739488882

> It should instead show its label within the dropdown field, which is the default behavior if there is no selection being made - that's why value needs to be empty.

assets/js/modules/analytics/components/common/GA4ActivateSwitch.js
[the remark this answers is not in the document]
eugene-manuilov: @aaemnnosttv it should be updated now.  😅

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
felixarntz: Keeping this in `useEffect` sounds good!

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
felixarntz: @aaemnnosttv Yes the label should be within the select when it's empty, this is the expected "empty" behavior of all our selects. So I think this is correct as is.

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
felixarntz: See above, I think this is correct as is. If anything, it may be even better to not even render any `Option` if there is no matching property, since in that case that `Option` is irrelevant. Still, this is more of a nitpick than a blocker. cc @eugene-manuilov

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
felixarntz: @aaemnnosttv Good point, however I'd prefer a loading indicator here rather than showing the UI. @eugene-manuilov Could we return a `ProgressBar` here that matches the height of the overall regular GA4 settings controls content?

Regarding the layout shift, it cannot be 100% avoided here: For example, if the GA4 API is not working, we don't want to keep displaying a disabled or blank-space area here, so in that case there would be a layout shift. With that said though, that should be more of the edge-case situation, so we can optimize for the common case. That's why I'm suggesting to display a progress bar with the height of the GA4 settings controls, to avoid the layout shift in case things work as expected.

If we can get this in today, let's do it - otherwise I'm also okay with fixing this in a separate follow-up issue (conscious of time here since this is a big PR and the release is close).

felixarntz (approved):
LGTM from my end!

@eugene-manuilov If you can realistically implement the changes in https://github.com/google/site-kit-wp/pull/4229#pullrequestreview-784779544 and https://github.com/google/site-kit-wp/pull/4229#discussion_r742176280 today, let's get them in. I think they are both worth doing as a UX enhancement. At the same time, they both aren't blockers for the first version though, so alternatively we can do both of them in follow-up issues. Let me know your thoughts! cc @aaemnnosttv

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
eugene-manuilov: @aaemnnosttv @felixarntz I have added the progress bar. Please, take a look at it and let me know if this looks good to you.

eugene-manuilov:
@felixarntz @aaemnnosttv, requested changes have been made. Please, take another look. Thanks.

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
```diff
@@ -0,0 +1,175 @@
...
+	}, [
+		matchedProperty,
+		matchedWebDataStream,
+		setPropertyID,
+		setWebDataStreamID,
+		setMeasurementID,
+	] );
+
+	if ( isAdminAPIWorking === undefined ) {
+		return <ProgressBar height={ 158 } small />;
```
felixarntz, line 119: Where do you get the `158` from? In my testing:
* If `enableGA4`, that entire area is `184` on desktop and `168` on smaller viewports.
* If `! enableGA4`, that area is `152` on desktop and `128` on smaller viewports.
* In all cases, the bottom margin is `16px`.

What if we set this to `200` (`184 + 16`) if `enableGA4` and to `168` (`152 + 16`) if not? It still doesn't get _all_ cases, but at least I would think there is no layout shift on desktop? This is a bit of a nit-picking exercise though, so maybe we should just let it be for now 🤔

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
```diff
@@ -0,0 +1,175 @@
...
+	}, [
+		accountID,
+		matchAccountProperty,
+		matchWebDataStream,
+		isAdminAPIWorking,
+	] );
+
+	const onActivate = useCallback( () => {
+		setPropertyID( matchedProperty?._id || PROPERTY_CREATE );
+		setWebDataStreamID( matchedWebDataStream?._id || WEBDATASTREAM_CREATE );
```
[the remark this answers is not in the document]
eugene-manuilov: Yes, makes sense. Updated.

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
eugene-manuilov: `158` is the height of the `GA4SettingsControls` component when `! enableGA4` in my browser. If this is critical, let's create a follow-up issue for this and leave `158` for now. What do you think?

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
aaemnnosttv: Yes, but for this specific select it looks broken because the text doesn't fit.
image

aaemnnosttv: @felixarntz this isn't particularly blocking from me but IMO it looks better and is a bit more consistent with the rest of the interface when the label is over the input here.

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
aaemnnosttv: @felixarntz I think this is the last important thing to finalize here and then this is good to go on my end.

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
felixarntz: @aaemnnosttv Right, but then a better fix would be to extend the width of the dropdown. A very long property ID for example would also be cut off like that, so I don't think we should fix it by enforcing the label to appear above, which would be inconsistent with the intended design.

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
felixarntz: I just tested this again and pushed another commit https://github.com/google/site-kit-wp/pull/4229/commits/dfe5fd63c6feb38b950fd902a13d5577513a002f that fixes it for me (at least partially, i.e. on desktop viewports).

It now uses `180` and `212` for the heights, because of the following:
* Height of the actual UI for the disabled and enabled state on desktop are `168` and `200` respectively, per what I've described above.
* There's an additional `16px` margin from the container in the actual UI, so this would lead to `184` and `216` respectively.
* The `ProgressBar` itself is `4px` tall, so whatever height you pass to it will actually not be entirely correctly reflected, the overall height will be `4px` taller because the height passed as prop is only used for the margins. This is a problem in the `ProgressBar` component we should probably fix, but not worth doing here. Anyway, because of those `4px` we need to subtract them, leading to `180` and `212`.
* Of course this still doesn't account for smaller viewports. We can't properly fix that though because the `ProgressBar` component doesn't support multiple heights per viewport - another thing to eventually fix. Smaller viewports are much less commonly used for WordPress and Site Kit though, so we should focus on desktop for now.

@eugene-manuilov @aaemnnosttv Let me know if this sounds good to you for now.

felixarntz (approved):

felixarntz:
@eugene-manuilov @aaemnnosttv This is good from my end, I pushed one more commit that fixes the height/layout shift issue for me, see https://github.com/google/site-kit-wp/pull/4229#discussion_r742373385. Please review, if it's good from your end feel free to merge.

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
aaemnnosttv: > A very long property ID for example would also be cut off like that

Actually it does not cut off the actual value in the select, it will stretch to show the whole thing. So the truncated text seems to only apply to the label. More text can fit when it is above due to the smaller size though so this is less likely to happen.
image

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
aaemnnosttv: The layout shift is gone for me now as well 🎉

aaemnnosttv (changes requested):
I noticed one thing left which is missing regarding the ACs:

> Once the GA4 UI is enabled here, saving the settings should also require completing the GA4 configuration, i.e. it should run the typical validation logic for checking GA4 property ID and web data stream ID, including creation of a new web data stream if it doesn't exist yet.

As it is, if GA4 is enabled it is possible to submit changes under some circumstances without any selected GA4 property option.

image

To reproduce: use a UA configuration that has no matching GA4 property, but does have at least 1 GA4 property (requiring user to choose) 
- delete any GA4 settings and then edit the module
- activate GA4 (submit changes will be disabled here, but only because no settings have changed)
- Toggle off the use snippet setting for GA4
- Changes can now be submitted even though no property choice was selected

aaemnnosttv:
@googlebot I consent.

aaemnnosttv (approved):
LGTM, thanks @eugene-manuilov ! 🚀

Issue #4208, opened 2021-10-11
Add support for auto-configuring Analytics based on service API response


felixarntz:
To facilitate the onboarding experience for Google Ads, Site Kit's authentication service is now integrating with their backend, providing custom behavior for users that set up Site Kit as part of the Google Ads onboarding flow. Specifically, based on the site URL, the service will know whether the user comes from the Google Ads onboarding flow and in that case send Analytics configuration data back to Site Kit, so that the module can be auto-activated and auto-configured, without the user even having to go through the Analytics setup UI.

The authentication service will internally handle adding the OAuth scope to the list of requested scopes, so that the user does not need to go to the OAuth consent screen a second time (like they have to today when setting up Analytics after the plugin setup). The experience for users coming from the Google Ads frontend to set up Site Kit will effectively be the following: After going through the plugin setup, in addition to the regular Search Console, Analytics will already be active and connected, without the user having to do anything extra. Due to that, this feature does not encompass any UI changes, everything happens automatically.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Listening to the `googlesitekit_authorize_user` hook, Site Kit should, if the response includes an `analytics_configuration` key, activate the Analytics module (if it's not active yet) and set the data from the associative array within as settings for the module (if it's not connected yet).
    * This bypasses both the GA4 and the existing tag check parts from the regular setup, which is acceptable for this feature.
* In case the Analytics module wasn't active before, it means that the `https://www.googleapis.com/auth/analytics.readonly` scope was stored in the "additional scopes" user setting instead of the "regular scopes" user setting. Therefore the handler that conditionally activates the Analytics module should also migrate that specific scope from the "additional scopes" user setting to the "regular scopes" user setting.

## Implementation Brief

* Note that while we only need to hook into `googlesitekit_authorize_user`, the IB splits that into two callbacks because one is Analytics-internal and needs to run regardless of whether Analytics is already active or will be activated on-the-fly, while the other one needs to activate Analytics on-the-fly, which is a core behavior and wouldn't be appropriate to trigger from within the Analytics module code.
* Implement a new public `Google\Site_Kit\Modules\Analytics::handle_token_response_data( $token_response )`. In there, if the `$token_response` includes a non-empty `analytics_configuration` key and the module is not already connected:
[...]

PR #4273, opened 2021-10-26
Enhancement/4208 auto configure analytics

eugene-manuilov:
## Summary

Addresses issue #4208 

## Checklist


includes/Core/Modules/Modules.php
```diff
@@ -244,6 +245,41 @@ function ( Module $module ) {
...
+		add_action(
+			'googlesitekit_authorize_user',
+			function( $token_response ) {
+				if ( empty( $token_response['analytics_configuration'] ) ) {
+					return;
+				}
+
+				if ( ! $this->is_module_active( Analytics::MODULE_SLUG ) ) {
+					$this->activate_module( Analytics::MODULE_SLUG );
+				}
```
felixarntz, line 257: I think it would be more accurate to bail early here if the module is already active. All logic that there is below should only run when the module isn't active:
* The call to `handle_token_response_data` would happen already if the module is active, due to the hook in the `Analytics` class.
* The migration of the extra scope is also only relevant in case the module isn't active when receiving the scopes.
```suggestion
				if ( $this->is_module_active( Analytics::MODULE_SLUG ) ) {
					return;
				}
				
				$this->activate_module( Analytics::MODULE_SLUG );
```

includes/Core/Modules/Modules.php
```diff
@@ -244,6 +245,41 @@ function ( Module $module ) {
...
+
+				try {
+					$analytics = $this->get_module( Analytics::MODULE_SLUG );
+					$analytics->handle_token_response_data( $token_response );
+				} catch ( Exception $e ) {
+					return;
+				}
```
felixarntz, line 264: Nit-pick, but I think this block should be moved to the bottom of this method. The part of migrating the scope is directly tied to the module not having been active before, it's basically "cleanup" around that.

The call here however just invokes the actual functionality to handle the response data, which would otherwise happen via the hook in the `Analytics` class on priority `10`, so it makes sense for this to happen last.

Technically, we could even just call the `Analytics::register` method here and then `handle_token_response_data` could remain private/protected, but I agree with the approach of calling it directly here since it's easier to follow and comes without the other registration side effects.

felixarntz (commented):
@eugene-manuilov Mostly looks good, just two small comments. This will also need to be refreshed against `develop` and have its merge conflicts fixed.

includes/Core/Modules/Modules.php
eugene-manuilov: Updated.

includes/Core/Modules/Modules.php
eugene-manuilov: Updated.

includes/Core/Modules/Modules.php
```diff
@@ -252,17 +252,13 @@ function( $token_response ) {
...
+				// Do nothing if the Analytics module is already activated.
+				if ( $this->is_module_active( Analytics::MODULE_SLUG ) ) {
 					return;
 				}
 
+				$this->activate_module( Analytics::MODULE_SLUG );
+
 				$extra_scopes = $this->user_options->get( OAuth_Client::OPTION_ADDITIONAL_AUTH_SCOPES );
 				if ( is_array( $extra_scopes ) && in_array( Analytics::READONLY_SCOPE, $extra_scopes, true ) ) {
 					unset( $extra_scopes[ array_search( Analytics::READONLY_SCOPE, $extra_scopes, true ) ] );
```
felixarntz: Nit-pick: We could _only_ call `array_search` instead of both `array_search` and `in_array` here, and the condition could then just check if the found index is `> -1`. But nothing blocking.

includes/Core/Modules/Modules.php
```diff
@@ -244,6 +245,44 @@ function ( Module $module ) {
...
+				// Do nothing if the Analytics module is already activated.
+				if ( $this->is_module_active( Analytics::MODULE_SLUG ) ) {
+					return;
+				}
+
+				$this->activate_module( Analytics::MODULE_SLUG );
+
+				$extra_scopes = $this->user_options->get( OAuth_Client::OPTION_ADDITIONAL_AUTH_SCOPES );
+				if ( is_array( $extra_scopes ) && in_array( Analytics::READONLY_SCOPE, $extra_scopes, true ) ) {
+					unset( $extra_scopes[ array_search( Analytics::READONLY_SCOPE, $extra_scopes, true ) ] );
```
felixarntz, line 264: Nit-pick: We could only call `array_search` instead of both `array_search` and `in_array` here, and the condition could then just check if the found index is `> -1`. But nothing blocking.

felixarntz (approved):
Excellent! Just one tiny comment, but not blocking.

includes/Core/Modules/Modules.php
eugene-manuilov: Ok, updated.

eugene-manuilov:
@felixarntz do you mind taking the final check and merging if you are good with all changes?

felixarntz (approved):
Excellent!

Issue #4219, opened 2021-10-12
Enhance GA event dimensions sent


felixarntz:
The custom dimensions sent with each GA tracking event should be slightly modified and enhanced, to provide more useful contextual information.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

The dimensions currently sent with every GA tracking event (see `trackEvent` method implementation) should be updated as follows:

* The value that is sent for `dimension2` should no longer be whether the user is the "first admin". Instead, it should be a comma-separated list of the current user's roles (typically only one role, but it is possible to have multiple roles e.g. via a plugin facilitating that, so this should be covered here as well).
* A `dimension6` should be added, which should be a comma-separated list of all the active modules' slugs. They should be in the order defined per the module order field.
* The two values needed for this can be added to the `_googlesitekitBaseData` global (we can't rely on the datastore here as the tracking still needs to work without loading most of Site Kit). The `isFirstAdmin` value should be removed completely.

## Implementation Brief

In `Google\Site_Kit\Core\Assets\Assets` class:

* In the `inline_data` array of `get_inline_user_data` method, add `roles` with the value `$current_user->roles`.

In the `register` method of `Google\Site_Kit\Core\Modules\Modules` class:

* add a new filter to `googlesitekit_inline_base_data` hook that adds `activeModules`  to the `data` array and passes the keys of `active_modules` as the value of it. 

In `Google\Site_Kit\Core\Authentication\Authentication` class:

* remove `$data['isFirstAdmin']` from the `inline_js_base_data` method.

In `assets/js/util/tracking/index.js` file:

* Remove `isFirstAdmin` from the destructuring assignment of `global._googlesitekitBaseData` and replace it with `activeModules`.
* get the roles of current user from `global._googlesitekitUserData`. ie. `const currentUserRoles = global._googlesitekitUserData?.user?.roles ?? [];`
* Pass both `activeModules` and `currentUserRoles` to the `initialConfig` object.

In `assets/js/util/tracking/createTracking.js` file:

* Update `DEFAULT_CONFIG` by removing `isFirstAdmin: false,` and adding `activeModules` and `currentUserRoles` with empty array as their values.

In `assets/js/util/tracking/createTrackEvent.js` file:

* Remove `isFirstAdmin` from the destructuring assignment of `config`.
* Destructure `activeModules` and `currentUserRoles` from the `config` object.
[...]

PR #4279, opened 2021-10-26
 Enhance GA event dimensions sent

kuasha420:
## Summary

Addresses issue #4219

## Checklist

### DIFF
--- a/assets/js/util/tracking/createTrackEvent.js
+++ b/assets/js/util/tracking/createTrackEvent.js
@@ -57,11 +57,11 @@ export default function createTrackEvent(
 			event_label: label,
 			value,
 			dimension1: referenceSiteURL,
-			dimension2: currentUserRoles.join(),
+			dimension2: currentUserRoles.join( ',' ),
 			dimension3: userIDHash,
 			dimension4: global.GOOGLESITEKIT_VERSION || '',
 			dimension5: Array.from( enabledFeatures ).join( ', ' ),
-			dimension6: activeModules.join(),
+			dimension6: activeModules.join( ',' ),
 		};
 
 		return new Promise( ( resolve ) => {


assets/js/util/tracking/createTrackEvent.js
```diff
@@ -56,10 +57,11 @@ export default function createTrackEvent(
 			event_label: label,
 			value,
 			dimension1: referenceSiteURL,
-			dimension2: isFirstAdmin ? 'true' : 'false',
+			dimension2: currentUserRoles.join( ',' ),
 			dimension3: userIDHash,
 			dimension4: global.GOOGLESITEKIT_VERSION || '',
 			dimension5: Array.from( enabledFeatures ).join( ', ' ),
+			dimension6: activeModules.join( ',' ),
```
felixarntz, line 64: @kuasha420 A bit of a nit-pick, but can we align on a consistent way of concatenating comma-separated values here? Right now it's mixed between `', '` and `','` (without space). I don't have a strong preference, I just think it'd be better for it to be consistent. @aaemnnosttv @tofumatt @eugene-manuilov Let me know if you have any strong feelings towards one or the other.

kuasha420: @eugene-manuilov can elaborate on this better, initially I went with the `', '`, matching `dimension5`.

felixarntz: Ah, I see now. @eugene-manuilov In that case, how about we use only the comma throughout?

eugene-manuilov: @felixarntz, yes, just commas is better because this format is intended for machine-reading whilst `, ` is what we use in UI. @kuasha420 can you create a follow-up PR that updates the `dimension5` value as well?

kuasha420: @eugene-manuilov @felixarntz I will create a follow up PR against `main`. Thanks for the clarification!

aaemnnosttv: +1 to omitting the added space. FWIW this is how `Array.toString` formats the output as well 👍

Issue #3201, opened 2021-04-20
Allow (programmatically) to do a full reset of Site Kit


felixarntz:
When resetting Site Kit, most data that the plugin creates (all `googlesitekit_` options and user options) are wiped. However, there is some essential data that cannot easily be restored which therefore does not get wiped on reset (prefixed with `googlesitekitpersistent_`).

We should allow to do a "hard reset" that even wipes those options. This caters for users that truly want to start fresh or that want to get rid of Site Kit data entirely.

Since this is not a generally encouraged usage, for now this should only be allowed programmatically.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a new filter applied `googlesitekit_hard_reset_enabled`, which expects a boolean. The default value should be `false`.
* If the filtered value is `true`, anytime a Site Kit reset is triggered, it should also wipe all data that starts in `googlesitekitpersistent_`:
    * options
    * user options
    * post meta (as of #3154)

## Implementation Brief

In `includes/Core/Util/Reset.php`:
 * Create a new `public function maybe_hard_reset()`
   * Return if `apply_filters( 'googlesitekit_hard_reset_enabled', false )` is `false`.
   * Instantiate a new `Reset_Persistent` with `$this->context`
   * Call `$reset_persistent->all()`
 * In the rest route handler, after the call to `$this->all()` add a call to `$this->maybe_hard_reset().`
 * Add the same call to `$this->maybe_hard_reset()` in the `handle_reset_action()` after `$this->all()`.

https://github.com/google/site-kit-wp/blob/develop/includes/Core/Util/Reset.php#L240-L241

### Test Coverage

* Update `tests/phpunit/integration/Core/Util/ResetTest.php`

### Visual Regression Changes

* N/A

## QA Brief

* Test the filter works in both the action `admin_action_googlesitekit_reset` and the reset button of the interface `window.googlesitekit.data.stores["core/site"].actions.fetchReset()`
* Check for the presence / absence of `persistent` meta as appropriate `wp user meta list admin|grep googlesitekitpersistent`
* 
## Changelog entry

* Introduce a new filter that allows for persistent data to be deleted during reset.

felixarntz:
@aaemnnosttv Is this waiting for anything? I don't recall, but I'd be okay with moving this forward to IB as is.

eugene-manuilov:
[...]

PR #4292, opened 2021-10-28
Enhancement/3201 full reset

eugene-manuilov:
## Summary

Addresses issue #3201

## Checklist

### DIFF
--- a/includes/Core/Util/Reset.php
+++ b/includes/Core/Util/Reset.php
@@ -288,6 +288,18 @@ class Reset {
 	 * @since n.e.x.t
 	 */
 	public function maybe_hard_reset() {
+		/**
+		 * Filters the hard reset option, which is `false` by default.
+		 *
+		 * By default, when Site Kit is reset it does not delete "persistent" data
+		 * (options prefixed with `googlesitekitpersistent_`). If this filter returns `true`,
+		 * all options belonging to Site Kit, including those with the above "persistent"
+		 * prefix, will be deleted.
+		 *
+		 * @since n.e.x.t
+		 *
+		 * @param bool $hard_reset_enabled If a hard reset is enabled. `false` by default.
+		 */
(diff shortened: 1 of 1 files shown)


includes/Core/Util/Reset.php
```diff
@@ -278,4 +281,20 @@ private function handle_reset_action( $nonce ) {
...
 		exit;
 	}
+
+	/**
+	 * Performs hard reset if it is enabled programmatically.
+	 *
+	 * @since n.e.x.t
+	 */
+	public function maybe_hard_reset() {
+		$hard_reset_enabled = apply_filters( 'googlesitekit_hard_reset_enabled', false );
```
tofumatt, line 291: Let's add documentation to this filter like we do elsewhere. How about:
```suggestion
		/**
		 * Filters the hard reset option, which is `false` by default.
		 *
		 * By default, when Site Kit is reset it does not delete "persistent" data
		 * (options prefixed with `googlesitekitpersistent_`). If this filter returns `true`,
		 * all options belonging to Site Kit, including those with the above "persistent"
		 * prefix, will be deleted.
		 *
		 * @since n.e.x.t
		 *
		 * @param bool $hard_reset_enabled If a hard reset is enabled. `false` by default.
		 */
		$hard_reset_enabled = apply_filters( 'googlesitekit_hard_reset_enabled', false );
```

tofumatt (changes requested):
Looks good, but I think we should add a doc block to the filter introduced here. If you think my suggestion looks good feel free to commit it then assign this back to me for another look/to merge 👍🏻

includes/Core/Util/Reset.php
eugene-manuilov: Thanks, @tofumatt. Looks good to me.

tofumatt (approved):

Issue #4133, opened 2021-09-27
Update PageSpeed Insights to be active by default for new installs


tofumatt:
## Feature Description

As per the [Unified Dashboard Design](https://docs.google.com/document/d/1StGx76XXisE9MfwM1916D9fn7-H9BRiRgizRejTsecI/edit?pli=1#heading=h.t2gbzjvuj834), the PageSpeed Insights should be an always-activated/auto-activated module that can't be disabled.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

**Despite this technically being part of the _Unified dashboard_ epic, this can be released decoupled, without the feature flag, i.e. as soon as it is ready.**

* The PSI module should be activated by default for new users (or users that reset and then re-setup the plugin). In other words, it should really just be activated per the default value of the option, which would mean existing sites that already have the option set won't be affected.
* The CTA versions of the PSI widget on the dashboard should be removed (since once users consciously deactivate it, we shouldn't prompt them to activate it again).

### Previous ACs (outdated)

* The PageSpeed Insights Module should be changed to be a "force-active" module, i.e. it should always be active regardless of the active modules option, and it should not allow to be disabled.
* The CTA to activate PageSpeed Insights should be removed, since that state will no longer be relevant. The PageSpeed widget can be modified to render nothing if PageSpeed Insights is inactive (which is not realistically possible, but should still be covered here, similar to how we do for Search Console widgets).
* The special conditions for displaying alternate content when having successfully activated `pagespeed-insights` in the success notification `DashboardSetupAlerts` should be removed, since it will no longer be possible to land there.
* A new tooltip tour `pagespeedInsightsForceActive` should be added, to inform users of sites that didn't have the module activated before about that it's now active (it should not be shown on sites where the module was already active).
    * It should show if on the Site Kit dashboard.
    * It should show if the initial Site Kit version is older than whatever release version this issue gets planned for.
    * It should show (using `checkRequirements`) if the site **does not** have `pagespeed-insights` in the list of active modules from the option. Note that this somehow needs to look at the option value explicitly, since when checking module activation status from JS regularly, it would always return that the module is active (due to the module being force-active).
    * The GA event category should be `${ VIEW_CONTEXT_DASHBOARD }_pagespeed-widget-force-active`.
    * There should be a single tooltip that targets the overall PageSpeed Insights widget.
        * Title: _PageSpeed Insights is now active for everyone_
[...]

PR #4305, opened 2021-11-02
Feature/4133 always enable pagespeed insights

ivankruchkoff:
## Summary

Addresses issue #4133

## Relevant technical choices

- One item not covered in IB was covered in subsequent comments https://github.com/google/site-kit-wp/issues/4133#issuecomment-956505945
- 
## Checklist


assets/js/feature-tours/pagespeed-insights-force-active.js
```diff
@@ -0,0 +1,70 @@
...
+	PAGE_SPEED_INSIGHTS_GA_CATEGORY_FORCE_ACTIVE,
+	STRATEGY_DESKTOP,
+	STRATEGY_MOBILE,
+} from '../modules/pagespeed-insights/datastore/constants';
+import { CORE_SITE } from '../googlesitekit/datastore/site/constants';
+
+const pagespeedInsightsForceActive = {
+	slug: 'pagespeedInsightsForceActive',
+	contexts: [ VIEW_CONTEXT_DASHBOARD ],
+	version: '1.43.0', // @TODO version
```
ivankruchkoff, line 39: what version should we set this to?

It can't be `n.e.x.t.` as we get a non semver error.

includes/Core/Assets/Assets.php
```diff
@@ -668,20 +669,22 @@ private function get_inline_base_data() {
...
+			'adminURL'             => esc_url_raw( trailingslashit( admin_url() ) ),
+			'assetsURL'            => esc_url_raw( $this->context->url( 'dist/assets/' ) ),
+			'blogPrefix'           => $wpdb->get_blog_prefix(),
+			'ampMode'              => $this->context->get_amp_mode(),
+			'isNetworkMode'        => $this->context->is_network_mode(),
+			'timezone'             => get_option( 'timezone_string' ),
+			'siteName'             => get_bloginfo( 'name' ),
+			'enabledFeatures'      => Feature_Flags::get_enabled_features(),
+			'webStoriesActive'     => defined( 'WEBSTORIES_VERSION' ),
+			'psiManuallyActivated' => $modules->manually_enabled( 'pagespeed-insights' ),
```
eugene-manuilov, line 687: Instead of adding this information here, let's better update the appropriate hook in the Modules class: https://github.com/google/site-kit-wp/blob/740fa3262c465dafd097a7b8d4100bf68c6a52c4/includes/Core/Modules/Modules.php#L247-L262

includes/Core/Util/Migration_1_45_0.php
```diff
@@ -0,0 +1,120 @@
...
+ * @access private
+ * @ignore
+ */
+class Migration_1_45_0 {
+	/**
+	 * Target DB version.
+	 */
+	const DB_VERSION = '1.45.0';
+
+	const OPTION_KEY_PSI_UPDATED = 'googlesitekit_psi_updated';
```
eugene-manuilov, line 38: > A potential way to solve this is could be:
> * Include an upgrade routine that sets a very simple option in the DB like googlesitekit_psi_updated if the site is currently connected.
> * **Show the tooltip tour only if that flag is set (and if PSI is not in the active modules option).**
> * On reset, delete this option so that it's gone.

We also need to check this flag when we determine a value for the `psiManuallyActivated` property.

includes/Modules/PageSpeed_Insights.php
```diff
@@ -17,6 +17,7 @@
 use Google\Site_Kit\Core\Modules\Module_With_Deactivation;
 use Google\Site_Kit\Core\Modules\Module_With_Scopes;
 use Google\Site_Kit\Core\Modules\Module_With_Scopes_Trait;
+use Google\Site_Kit\Core\Modules\Modules;
```
eugene-manuilov, line 20: Is this needed here?

includes/Core/Util/Migration_1_45_0.php
```diff
@@ -0,0 +1,120 @@
...
+use WP_Error;
+
+/**
+ * Class Migration_1_45_0
+ *
+ * @since 1.45.0
+ * @access private
+ * @ignore
+ */
+class Migration_1_45_0 {
```
eugene-manuilov, line 32: We need to instantiate this class in the Plugins class as we do it for other migration classes: https://github.com/google/site-kit-wp/blob/740fa3262c465dafd097a7b8d4100bf68c6a52c4/includes/Plugin.php#L194-L195

includes/Core/Util/Migration_1_45_0.php
```diff
@@ -0,0 +1,120 @@
...
+	 * Target DB version.
+	 */
+	const DB_VERSION = '1.45.0';
+
+	const OPTION_KEY_PSI_UPDATED = 'googlesitekit_psi_updated';
+
+	/**
+	 * Context instance.
+	 *
+	 * @since 1.45.0
```
eugene-manuilov, line 43: All `@since` phpdoc tags in this file should be `n.e.x.t` as well as the `DB_VERSION` constant value.

includes/Core/Util/Migration_1_45_0.php
```diff
@@ -0,0 +1,120 @@
+<?php
+/**
+ * Migration for 1.45.0
```
eugene-manuilov, line 3: I think it should be `1.46.0`, shouldn't it? The whole file and the class name should be renamed to `Migration_1_46_0`.

eugene-manuilov (changes requested):
Thanks, @ivankruchkoff. This is a good start. Added some comments for you. Please, take a look.

includes/Core/Util/Migration_1_45_0.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/4305/commits/95aec903472e378b824d99b9e3a98f9b390d8a3b

assets/js/modules/pagespeed-insights/components/dashboard/DashboardPageSpeedWidget.js
```diff
@@ -33,10 +33,4 @@ function DashboardPageSpeedWidget( { Widget } ) {
 
 export default whenActive( {
 	moduleName: 'pagespeed-insights',
-	FallbackComponent: ( { WidgetActivateModuleCTA } ) => (
-		<WidgetActivateModuleCTA moduleSlug="pagespeed-insights" />
-	),
-	IncompleteComponent: ( { WidgetCompleteModuleActivationCTA } ) => (
-		<WidgetCompleteModuleActivationCTA moduleSlug="pagespeed-insights" />
-	),
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/4305/commits/7ed480fe70a177cc0254af2c556f68528838c3a4

assets/js/feature-tours/pagespeed-insights-force-active.js
```diff
@@ -0,0 +1,70 @@
...
+	STRATEGY_DESKTOP,
+	STRATEGY_MOBILE,
+} from '../modules/pagespeed-insights/datastore/constants';
+import { CORE_SITE } from '../googlesitekit/datastore/site/constants';
+
+const pagespeedInsightsForceActive = {
+	slug: 'pagespeedInsightsForceActive',
+	contexts: [ VIEW_CONTEXT_DASHBOARD ],
+	version: '1.44.0',
+	gaEventCategory: PAGE_SPEED_INSIGHTS_GA_CATEGORY_FORCE_ACTIVE,
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/4305/commits/72b81242ce456373491ec98d81bcbb2c46391b68

includes/Modules/PageSpeed_Insights.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/4305/commits/999698b4485ce25311ef90750c157566e1e7ada6

includes/Core/Assets/Assets.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/4305/commits/a76c3de66869fc93743b5ac7b48a1bdf044d7d26

includes/Core/Assets/Assets.php
```diff
@@ -11,6 +11,7 @@
 namespace Google\Site_Kit\Core\Assets;
 
 use Google\Site_Kit\Context;
+use Google\Site_Kit\Core\Modules\Modules;
```
eugene-manuilov, line 14: Changes in this file aren't needed. Please, remove it.

eugene-manuilov (changes requested):
Thanks, @ivankruchkoff. This is closer now, but still require a bit of work. Please, see my comment below and two more comments from the previous review: https://github.com/google/site-kit-wp/pull/4305#discussion_r745055355 and https://github.com/google/site-kit-wp/pull/4305#discussion_r745057502

includes/Core/Assets/Assets.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/4305/commits/9b103db948a1d0793a69c00688625e918fde282d (another unused import went with it).

includes/Core/Util/Migration_1_45_0.php
ivankruchkoff: I thought I had done this... https://github.com/google/site-kit-wp/pull/4305/commits/50a92f41a8c0e6ac53fc8667ce803c739f6a0e37

includes/Core/Util/Migration_1_45_0.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/4305/commits/43021fd8f08fdf722074790b62021424c4a7db18

includes/Core/Assets/Assets.php
ivankruchkoff: https://github.com/google/site-kit-wp/commit/9b103db948a1d0793a69c00688625e918fde282d

assets/js/feature-tours/pagespeed-insights-force-active.js
```diff
@@ -0,0 +1,70 @@
...
+	MODULES_PAGESPEED_INSIGHTS,
+	STRATEGY_DESKTOP,
+	STRATEGY_MOBILE,
+} from '../modules/pagespeed-insights/datastore/constants';
+import { CORE_SITE } from '../googlesitekit/datastore/site/constants';
+
+const pagespeedInsightsForceActive = {
+	slug: 'pagespeedInsightsForceActive',
+	contexts: [ VIEW_CONTEXT_DASHBOARD ],
+	version: '1.45.0',
```
eugene-manuilov, line 38: I think this should be `1.46.0` since we are going to display this feature tour only for users who have installed the plugin earlier.

includes/Core/Util/Migration_1_46_0.php
```diff
@@ -0,0 +1,120 @@
...
+use WP_Error;
+
+/**
+ * Class Migration_1_46_0
+ *
+ * @since 1.46.0
+ * @access private
+ * @ignore
+ */
+class Migration_1_46_0 {
```
eugene-manuilov, line 32: I have been testing this on my end and it seems like we don't need this migration class at all. The tooltip tour won't re-appear after reset because dismissed feature tours are saved in the persistent option. Thus I think we can remove this class.

eugene-manuilov (changes requested):
@ivankruchkoff, added three more comments. Please, take a look.

includes/Core/Util/Migration_1_46_0.php
eugene-manuilov: cc @felixarntz for visibility.

ivankruchkoff: @eugene-manuilov yes, in that scenario it would be fine, but we're trying to avoid a different one:

https://github.com/google/site-kit-wp/issues/4133#issuecomment-958647430

assets/js/feature-tours/pagespeed-insights-force-active.js
ivankruchkoff: https://github.com/google/site-kit-wp/pull/4305/commits/e127eb1e08d6fc09cfb94ff99bde0f7761f56520

assets/js/feature-tours/pagespeed-insights-force-active.js
```diff
@@ -0,0 +1,70 @@
...
+import {
+	MODULES_PAGESPEED_INSIGHTS,
+	STRATEGY_DESKTOP,
+	STRATEGY_MOBILE,
+} from '../modules/pagespeed-insights/datastore/constants';
+import { CORE_SITE } from '../googlesitekit/datastore/site/constants';
+
+const pagespeedInsightsForceActive = {
+	slug: 'pagespeedInsightsForceActive',
+	contexts: [ VIEW_CONTEXT_DASHBOARD ],
```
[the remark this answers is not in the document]
ivankruchkoff: https://github.com/google/site-kit-wp/pull/4305/commits/e127eb1e08d6fc09cfb94ff99bde0f7761f56520

assets/js/feature-tours/pagespeed-insights-force-active.js
felixarntz: @ivankruchkoff @eugene-manuilov Sounds good - we just need to ensure then that this actually ships with 1.46.0. I've therefore set this release on the issue to mark it already as "requirement".

Overall, where does the semver error come from? It would be great if we could specify `n.e.x.t` here as we do elsewhere - maybe we can allow for a special case (e.g. `n.e.x.t` could _always_ consider be greater than any other version you compare it too). Nothing to break our heads about here, but maybe worth a separate issue.

assets/js/feature-tours/pagespeed-insights-force-active.js
```diff
@@ -0,0 +1,73 @@
...
+		// Wait until page speed insights module has loaded.
+		const referenceURL = registry
+			.select( CORE_SITE )
+			.getCurrentReferenceURL();
+		await registry
+			.__experimentalResolveSelect( MODULES_PAGESPEED_INSIGHTS )
+			.getReport( referenceURL, STRATEGY_MOBILE );
+		await registry
+			.__experimentalResolveSelect( MODULES_PAGESPEED_INSIGHTS )
+			.getReport( referenceURL, STRATEGY_DESKTOP );
```
felixarntz, line 54: Is this really needed? We point to the overall widget area, so I don't think it matters for the tooltip where the widget is fully loaded. I'd say let's remove this, less code to maintain.

includes/Core/Modules/Modules.php
```diff
@@ -468,6 +470,19 @@ public function activate_module( $slug ) {
...
 
+	/**
+	 * Checks whether the module identified by the given slug is enabled by the option.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param string $slug Unique module slug.
+	 * @return bool True if module has been manually enabled, false otherwise.
+	 */
+	public function manually_enabled( $slug ) {
```
felixarntz, line 481: Let's not make this public - it's not a method intended for others to call, and it doesn't need to be public for the code here either.

includes/Core/Util/Migration_1_46_0.php
felixarntz: @eugene-manuilov @ivankruchkoff is correct - this extra flag is needed to indicate that the plugin was already connected before. In combination with the check for whether the module is not manually activated, this tells us that:
* the site was using Site Kit before and was already connected
* the site didn't have PSI connected

Because of those conditions we need to show the tooltip tour. On reset, this option here will get wiped, which is intentional because after reset the site starts in a fresh state. In that situation, we shouldn't ever show the tooltip tour because activating PSI after setup is more of an expected scenario, while if the site already _was_ connected, it would be a more "random" experience where we have to explain why it's suddenly active.

includes/Core/Modules/Modules.php
```diff
@@ -256,7 +257,8 @@ function( Module $module ) {
 					}
 				);
 
-				$data['activeModules'] = array_keys( $non_internal_active_modules );
+				$data['activeModules']        = array_keys( $non_internal_active_modules );
+				$data['psiManuallyActivated'] = $this->manually_enabled( 'pagespeed-insights' ) && $this->options->get( Migration_1_46_0::OPTION_KEY_PSI_UPDATED );
```
felixarntz, line 261: I think the logic here is slightly off, in combination with https://github.com/google/site-kit-wp/pull/4305/files#diff-b0980b9f904136a052c4068889aa1973f11d6a7fc5d1c784bf37329ac8c13f79R56. For example, right now it will evaluate to `false` when a site that is just starting with 1.46.0 (i.e. no migration) hasn't manually activated PSI, which for that version will always be the case (since you can't activate it anymore).

This will lead to the tooltip tour potentially showing, which is not correct.

I suggest we simplify this: We don't actually care for whether PSI was manually activated. We care for whether PSI is **not** manually activated and the current version was migrated to from a pre-1.46.0 version. I think this here could simply become:
```suggestion
				$data['showPSIForceActiveTour'] = ! $this->manually_enabled( 'pagespeed-insights' ) && $this->options->get( Migration_1_46_0::OPTION_KEY_PSI_UPDATED );
```

Then https://github.com/google/site-kit-wp/pull/4305/files#diff-b0980b9f904136a052c4068889aa1973f11d6a7fc5d1c784bf37329ac8c13f79R56 should check for whether that value is `true`.

felixarntz (changes requested):
@ivankruchkoff Left a few comments here. This mostly looks good, but I believe there's one critical logic problem.

assets/js/feature-tours/pagespeed-insights-force-active.js
ivankruchkoff: See my comment in https://github.com/google/site-kit-wp/issues/4133#issuecomment-965912725 for the issues with each approach.

includes/Core/Modules/Modules.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/4305/commits/6e07da780fa8d2be9e85a194dc55592b879d4b07

assets/js/feature-tours/pagespeed-insights-force-active.js
ivankruchkoff: https://github.com/google/site-kit-wp/pull/4305/commits/6e07da780fa8d2be9e85a194dc55592b879d4b07 addresses the CR comment.

includes/Core/Modules/Modules.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/4305/commits/96b968af5c7efc192fb6ddd4c5e3269ad0f1debe

includes/Core/Util/Migration_1_46_0.php
eugene-manuilov: Thanks, @felixarntz and @ivankruchkoff.

includes/Core/Modules/Modules.php
```diff
@@ -256,7 +257,9 @@ function( Module $module ) {
 					}
 				);
 
-				$data['activeModules'] = array_keys( $non_internal_active_modules );
+				$data['activeModules']          = array_keys( $non_internal_active_modules );
+				$data['psiManuallyActivated']   = $this->manually_enabled( 'pagespeed-insights' ) && $this->options->get( Migration_1_46_0::OPTION_KEY_PSI_UPDATED );
```
eugene-manuilov, line 261: @ivankruchkoff, this line is not needed anymore, right?

eugene-manuilov (changes requested):

includes/Core/Modules/Modules.php
ivankruchkoff: https://github.com/google/site-kit-wp/pull/4305/commits/04da3b7f7f3fdb28dea93771cf83620555d4c817

eugene-manuilov (approved):
Ok, this looks good to me. Thanks, @ivankruchkoff! @felixarntz leaving this for you to approve and merge if you have nothing more to add.

felixarntz (approved):
LGTM given that we go with the approach from https://github.com/google/site-kit-wp/issues/4133#issuecomment-966515572.

Before merging this, please review https://github.com/google/site-kit-wp/issues/4133#issuecomment-966598285 though - if we can find a solution like that, let's adjust.

felixarntz (approved):

Issue #4106, opened 2021-09-22
Show loading state on secondary GA dropdown when "auto-populate" logic is still running


felixarntz:
When setting up Analytics and you already have both a matching UA and GA4 property configured, there is currently an odd user experience where the secondary dropdown (the one in the blue notice, usually the one for GA4) first shows the empty state where the user needs to select a property, just to update this 1-2 seconds later with the auto-populated property.

It could easily result in problems where the user quickly selects a property and then their choice is overwritten because the auto-population logic completed. Essentially, this dropdown should be editable until the auto-population logic has completed (whether or not it eventually auto-populates anything).

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* While the logic to find a matching UA or GA4 property is running, the respective dropdown should be in a loading state (the typical "small" progress bar) instead of showing in empty state, which makes the user think they could already use it.
    * In other words: Whether or not the logic to find a matching property eventually actually finds and sets one or not, this isn't known in advance so the dropdown should be in loading state.
* This is likely already correct in several places, but particularly in the case of having GA4 as the secondary property it is currently resulting in a bad user experience (see above).

## Implementation Brief
* Using `assets/js/modules/analytics-4/datastore/properties.js`,
	* Add a new state boolean variable, `isMatchingProperty` with `false` as default value.
	* Add a corresponding action, reducer and selector to set and get the value of `isMatchingProperty`
	* Update the `matchAndSelectProperty` action to dispatch the action defined above to set `isMatchingProperty` to `true`. This should happen before calling `matchAccountProperty`.
	* Then before returning `property`, set `isMatchingProperty` to `false` by dispatching the newly added action.
* Using `assets/js/modules/analytics-4/components/common/PropertySelect.js`,
	* Query the `analytics-4` data store via the newly added selector in `assets/js/modules/analytics-4/datastore/properties.js` to see if we are waiting for a matched GA4 property.
	* Add the value returned by the selector to the `isLoading` constant.

### Test Coverage
* Tests to be added for the newly added action and selector in `assets/js/modules/analytics-4/datastore/properties.js`

## QA Brief

* Set up the Analytics module with UA and GA4 properties.
* Go to the Analytics module settings and edit them.
* Make sure you see the progress bar for the GA4 property when you select a different account.

## Changelog entry

* Improve loading state for GA4 Analytics dropdown in settings.

wpdarren:
[...]

PR #4310, opened 2021-11-03
Enhancement/4106 loading state on secondary ga4 dropdown follow up

eugene-manuilov:
## Summary

Addresses issue #4106

## Checklist


eugene-manuilov:
Need to fix tests issue.

assets/js/modules/analytics-4/components/common/PropertySelect.js
```diff
@@ -82,11 +83,7 @@ export default function PropertySelect( { label } ) {
 		[ propertyID, selectProperty ]
 	);
 
-	if ( ! isValidAccountID( accountID ) ) {
-		return null;
-	}
```
felixarntz, line 87: Does this really need to be changed? If no account is selected, this component still shouldn't render at all, so this change seems off to me.

assets/js/modules/analytics-4/components/common/PropertySelect.test.js
```diff
@@ -92,48 +89,6 @@ describe( 'PropertySelect', () => {
 		expect( listItems ).toHaveLength( properties.length + 1 );
 	} );
 
-	it( 'should not render if account ID is not valid', async () => {
```
felixarntz, line 95: See above, I'm not sure why this behavior needs to be removed here? I think it should still be the case. The property dropdown shouldn't display in any way if no account is selected yet.

assets/js/modules/analytics-4/datastore/properties.js
```diff
@@ -137,11 +137,10 @@ const fetchCreatePropertyStore = createFetchStore( {
 
 // Actions
 const WAIT_FOR_PROPERTIES = 'WAIT_FOR_PROPERTIES';
-const START_MATCHING_PROPERTY = 'START_MATCHING_PROPERTY';
-const STOP_MATCHING_PROPERTY = 'STOP_MATCHING_PROPERTY';
+const FINISH_MATCHING_PROPERTY = 'FINISH_MATCHING_PROPERTY';
 
 const baseInitialState = {
-	isMatchingProperty: false,
+	finishedMatchingProperty: undefined,
```
felixarntz, line 143: Shouldn't this be `false` right on load? I don't think this makes sense to be `undefined` initially, since it's definitely not finished on load. The `undefined` here would prevent the strict `=== false` check above from being truthy.

assets/js/modules/analytics-4/datastore/properties.js
```diff
@@ -287,8 +286,10 @@ const baseActions = {
 	 */
 	*matchAndSelectProperty( accountID, fallbackPropertyID = '' ) {
 		yield {
-			type: START_MATCHING_PROPERTY,
-			payload: {},
+			type: FINISH_MATCHING_PROPERTY,
+			payload: {
+				finished: false,
+			},
```
felixarntz, line 292: This is semantically a bit odd. Why don't we keep the `START_MATCHING_PROPERTY` and `STOP_MATCHING_PROPERTY` as it was before? I don't think sending a `FINISH_MATCHING_PROPERTY` action before and after the action makes sense. Also particularly since it was like that before, I don't see the value in changing it since it doesn't really affect the underlying logic.

felixarntz (commented):
@eugene-manuilov A few points of feedback here, some of the changes here look questionable to me. I may be missing something though, so please clarify why you made those changes where applicable.

assets/js/modules/analytics-4/components/common/PropertySelect.js
eugene-manuilov: You are right, this should stay as is. Reverted.

assets/js/modules/analytics-4/components/common/PropertySelect.test.js
eugene-manuilov: Agree. Reverted.

eugene-manuilov:
@felixarntz, I have re-thought how this should work, and looks like the only way to make it reliable and work correctly in all cases is to use an indicator that signals when the account selection is finished. Changes that have been added to the properties datastore are not needed, thus I have removed them. Could you please take another look?

felixarntz (approved):
@eugene-manuilov Great idea - this indeed solves it in a cleaner and reliable way. The new approach looks all good to me!

Issue #4267, opened 2021-10-25
Analytics CTA Setup link causes the widget contents to "jump" during redirect to Auth Screen


tofumatt:
## Bug Description

When you click on the Analytics CTA widget (for instance, in the Unified Dashboard Search Funnel Widget without Analytics being connected/active), the contents of the widget changes as it goes from "entirely unconnected" to "setup not complete".

Here's a discussion for context: https://github.com/google/site-kit-wp/pull/4223#pullrequestreview-785958429

## Steps to reproduce

1. Go to the Unified Dashboard by enabling the "Unified Dashboard" feature flag in the Tester Plugin
2. Make sure Analytics is not connected
3. Click on the "Set up Analytics" link in the blue CTA on the top-right side of the screen/Search Funnel widget at the top of the page
4. See the content change from:

<img width="579" alt="Screenshot 2021-10-25 at 21 13 49" src="https://user-images.githubusercontent.com/90871/138764227-ce342b80-1748-4745-ad5d-2f64dc1e11e5.png">

to

<img width="663" alt="Screenshot 2021-10-25 at 21 13 54" src="https://user-images.githubusercontent.com/90871/138764247-1dbb6443-8a20-4401-99ab-1165f40ad09a.png">

before the page redirects.

We should _improve the UX here so that the content does not change as drastically or a progress bar with no user CTA is shown for this state_. Eg. If the user clicks on this setup link, they shouldn't be presented with the same or a different call to action while the page is redirecting to the auth flow. The next time they visit the dashboard, it's fine to show the "Setup incomplete" CTA (as seen in the second screenshot), but it should not appear between when the user clicks the link and the redirect.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* When clicking on the "Set up Analytics" button link in the widget above, the widget should not change to the "Complete Analytics Setup" copy while the page redirects to the Authentication URL.
* A progress bar should replace the "Set up Analytics" link/button in the bottom of the CTA widget, so there is no CTA button/link in the widget and the user gets a visual indication that something is happening before/during the redirect
* When the dashboard is visited again (eg. after a full-page refresh), the "Complete Analytics Setup" content for the Analytics CTA widget _should_ appear as usual.

## Implementation Brief
* Using `assets/js/components/ActivateModuleCTA.js`,
[...]

PR #4324, opened 2021-11-08
Add flag in the core/ui data store to indicate when we are activating…

asvinb:
… a module.

## Summary

Addresses issue #4267 

## Checklist

### DIFF
--- a/assets/js/components/notifications/CTA.js
+++ b/assets/js/components/notifications/CTA.js
@@ -56,7 +56,7 @@ const CTA = ( {
 
 		{ inProgress && (
 			<div className="googlesitekit-cta__progress">
-				<ProgressBar />
+				<ProgressBar small />
 			</div>
 		) }
 
--- a/assets/js/components/notifications/CTA.js
+++ b/assets/js/components/notifications/CTA.js
@@ -26,6 +26,7 @@ import classnames from 'classnames';
  * Internal dependencies
  */
 import Link from '../Link';
+import ProgressBar from '../ProgressBar';
 
 const CTA = ( {
 	title,
@@ -35,6 +36,7 @@ const CTA = ( {
 	ctaLinkExternal,
 	error,
 	onClick,
+	inProgress,
 	'aria-label': ariaLabel,
 } ) => (
 	<div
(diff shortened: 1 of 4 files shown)


eugene-manuilov (changes requested):
Thanks, @asvinb. Looks like I missed it during reviewing IB but we need to implement it slightly different. Check what AC says:

> A progress bar should replace the "Set up Analytics" link/button in the bottom of the CTA widget, so there is no CTA button/link in the widget and the user gets a visual indication that something is happening before/during the redirect

It means that we need to display the progress bar instead of the setup button in the ActivateModuleCTA component itself and we don't need the progress bar in the Overview component of the SearchFunnelWidget widget.

Could you please update your PR to match requirements from AC?

asvinb:
@eugene-manuilov Thanks for catching that. I have updated the PR to reflect that in the following [commit](https://github.com/google/site-kit-wp/pull/4324/commits/2676d3026e5ceb4f2e1bb3753363a61c5227aec4). I have added a new `inProgress` prop to the `CTA` component since it's what being used. My main worry right now however is how that change can affect the whole site, meaning whenever we have a CTA to activate a module, the progress bar will be displayed. I'am perfectly fine with that but curious whether the impact of it.
cc @tofumatt 

Let me know what you think.

assets/js/components/notifications/CTA.js
```diff
@@ -51,7 +53,14 @@ const CTA = ( {
 				{ description }
 			</div>
 		) }
-		{ ctaLabel && (
+
+		{ inProgress && (
+			<div className="googlesitekit-cta__progress">
+				<ProgressBar />
```
eugene-manuilov, line 59: Let's add the `small` attribute to the progress bar to make it look nicer: 
```suggestion
				<ProgressBar small />
```

eugene-manuilov (changes requested):
Awesome! Thanks, @asvinb. This looks almost ideal to me. Added one suggestion after which we can approve this.

assets/js/components/notifications/CTA.js
asvinb: 👍

eugene-manuilov (approved):
Thanks, @asvinb. Looks good to me.

felixarntz:
@asvinb What's the state of this PR here? Without having looked to closely, I'm slightly confused because this is still open, but the issue #4267 is successfully closed including QA 🤔 

Was there an alternative PR that fixed this?

asvinb:
@felixarntz Apologies for the late reply, it's only now i saw your comment. I'am puzzled as well. Most likely another PR fixed the issue but I'll dig into it and find out what exactly happened.

aaemnnosttv:
Closing as the related issue has already been addressed.

Issue #3822, opened 2021-08-10
Ensure the plugin inserts the Optimize snippet in all scenarios between GA4 and UA


jamesozzie:
## Bug Description

If a user has manually or used a third party plugin to insert their Analytics snippet (UA, GA4 or both) and then connects the Optimize module there are situations whereby the plugin won't insert the Optimize container ID - as it can't attach the container ID to a non Site Kit placed snippet. 

This can be confusing for users, who may not be aware that the plugin no longer or doesn't insert the Optimize container ID, despite both services being connected. It can come about in the following circumstances:
- A user manually inserts both a UA and GA4 snippet and then connected Optimize
- A user may revisit their Analytics settings after setting up Optimize and doesn't allow Site Kit to insert a GA4 or UA snippet (see below)

![image](https://user-images.githubusercontent.com/41326532/128904396-6ac11535-e862-4c8f-8733-534511479d83.png)

To improve awareness or resolve this issue the following could be considered:

1. Don't allow the Optimize module to be connected if Site Kit has not inserted either a UA or GA4 Analytics snippet - while also disconnecting Optimize (with a warning) if a user removes Site Kit inserted Analytics snippets while keeping the Analytics module connected
2. Insert the Optimize snippet to a users site [independent of the Analytics snippet](https://i.imgur.com/LDaKVx7.png)

## Screenshots of the various possibilities at present:
- UA and GA4 snippet placed via SK - **no issue** ([Optimize snippet added to UA tag](https://i.imgur.com/0595KOf.png))
- UA manually inserted with GA4 snippet added via SK - **no issue** ([Optimize snippet added to GA4 tag](https://i.imgur.com/j7dgI50.png)), vice versa for a GA4 manually inserted tag
- UA manually inserted with no GA4 snippet added via SK - **problem** ([Optimize container is not added via Site Kit but the service remains connected](https://i.imgur.com/cZkxG59.png))
- UA and GA4 manually inserted - **problem**([Optimize container is not added via Site Kit but the service remains connected](https://i.imgur.com/P7NfkA8.png))
 
## To reproduce the issue

1. Setup Analytics within Site Kit although with a manually inserted GA4 & UA property
2. Setup Optimize within the plugin
3. Check your sites source code, the Optimize container is not added.  

## Additional Context

 - These potential confusing issues started since Site Kit inserted GA4 snippets 
[...]

PR #4344, opened 2021-11-11
Add notice for when userSnippet is disabled in GA4/UA.

kuasha420:
## Summary

Addresses issue #3822

## Checklist

### DIFF
--- a/assets/js/modules/optimize/components/common/OptimizeSnippetNotice.js
+++ b/assets/js/modules/optimize/components/common/OptimizeSnippetNotice.js
@@ -19,7 +19,6 @@
 /**
  * WordPress dependencies
  */
-import { useMemo } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 
 /**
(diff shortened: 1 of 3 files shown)


assets/js/modules/optimize/components/common/OptimizeSnippetNotice.js
```diff
@@ -0,0 +1,81 @@
...
+
+	const useGA4Snippet = useSelect( ( select ) =>
+		select( MODULES_ANALYTICS_4 ).getUseSnippet()
+	);
+
+	const useUASnippet = useSelect( ( select ) =>
+		select( MODULES_ANALYTICS ).getUseSnippet()
+	);
+
+	const notice = useMemo( () => {
```
eugene-manuilov, line 48: This `useMemo` is not needed. Let's remove it.

eugene-manuilov (changes requested):
Thanks, @kuasha420. This looks good in general. There is one comment for you. Once it's addressed, this PR should be good to go.

assets/js/modules/optimize/components/common/OptimizeSnippetNotice.js
kuasha420: Removed the `useMemo` and instead used a more imperative approach. Question, why's the use of `useMemo` bad here? Thanks!

eugene-manuilov (approved):
LGTM

assets/js/modules/optimize/components/common/OptimizeSnippetNotice.js
eugene-manuilov: @kuasha420 the question here is not why it is bad, the question is why the use of `useMemo` here is needed? The `useMemo` hook is needed to avoid expensive calculations on every render. Translating a string is not an expensive calculation from my point of view, thus we don't need this hook.

Issue #4074, opened 2021-09-20
Cannot read properties of undefined (reading 'load')


jamesozzie:
## Bug Description

There have been four users who referenced the below dashboard error, all within the same week:
`Cannot read properties of undefined (reading 'load')`

- https://wordpress.org/support/topic/site-kit-encountered-an-error-49/ | **Open** | [SH info](https://docs.google.com/forms/d/1lhjns2K7c0Ny7ME4gxTcsnYtNUtcMr8BvFpPQi23Y1E/edit#response=ACYDBNgXnVLGZ6D0PN6f3y2RcJfNcA5kAiP5_FKtcFit9hTxXSO9N6zEpMO0VDcnKITQtXk)
- https://wordpress.org/support/topic/error-detected-3/#post-14888496 | **Open** | No SH info at this point
- https://wordpress.org/support/topic/cannot-read-properties-of-undefined-reading-load | **Open** | No SH info at this point

While awaiting details from one impacted user on the first report this only occurs when selecting specific reporting periods on the primary Site Kit dashboard - all expect the 28 day reporting period. This occurs consistently for this user, and doesn't occur in troubleshooting mode (using the Health Check plugin). 

This is likely as a result of a plugin or theme conflict. Awaiting for checks and details from impacted users
 
## Additional Context

 - Plugin Version: 1.41.0
 - This only started to occur from 1.41.0
 - No previous reports until last week
 - [One user](https://wordpress.org/support/topic/site-kit-encountered-an-error-49/) (not the original reporter) stated that this occurred once PageSpeed Insights was activated, within one of the support forum threads. Once they "cleared the error" it no longer occurred
 - The issue does still remain incognito mode
 - The issue no longer occurs in troubleshooting mode using the Health Check & Troubleshooting plugin for one user (performing other checks at present to try and pinpoint the issue)
- No browser console errors provided yet

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All Site Kit dashboard screens should proactively define the `window.google` as undefined to avoid a potential error when charts are initializing. 
See https://github.com/google/site-kit-wp/issues/4074#issuecomment-967364462 for more information

<details>
<summary>Original AC</summary>

* The Google charts library should be rolled back to the hard-coded `49` version, as the newer versions are still causing a high number of errors.
    * In a future issue we should explore updating again but rule out that those errors happen. It will probably be a much larger effort, but this is a critical problem that we need to get out for the upcoming release.

[...]

PR #4354, opened 2021-11-15
Make sure `window.google` is undefined before loading Charts.

kuasha420:
## Summary

Addresses issue #4074

## Checklist

### DIFF
--- a/assets/js/components/GoogleChart.js
+++ b/assets/js/components/GoogleChart.js
@@ -17,6 +17,9 @@
  */
 /* eslint-disable react-hooks/exhaustive-deps */
 
+/* Ensures `google` global is undefined before loading `react-google-charts` library */
+import '../util/initialize-google-global';
+
 /**
  * External dependencies
  */
--- a/assets/js/util/initialize-google-global.js
+++ b/assets/js/util/initialize-google-global.js
@@ -1,0 +1,21 @@
+/**
+ * Side Effect to initialize `google` global.
+ *
+ * Site Kit by Google, Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


aaemnnosttv:
Thanks @kuasha420, this looks great. If it's not too much effort, can we add a test for this as well as described in the IB? We don't need to render any specific charts that Site Kit uses, but we should ensure that a chart that renders after charts have already been loaded does not raise this error. We could use something simple as is shown in the [Quick Start](https://github.com/rakannimer/react-google-charts#quick-start).

kuasha420:
@aaemnnosttv should I add a unit test for the component and/or a storybook test? 

Cheers!

aaemnnosttv:
@kuasha420 just a component test should suffice as this is more of a timing issue. The visual state of the component is irrelevant. Thanks 👍

kuasha420:
@aaemnnosttv Hi Evan, I don't think unit testing the loaded state of charts library is possible in jest/jsdom as the charts never load no matter how much I wait. I also found [this](https://github.com/rakannimer/react-google-charts/issues/214) which reaffirms the point. The two charts remain in loading state forever in Jest. As there is no way to see if the `undefined` error is happening with/without applying the fix in this PR via jest, is there any value in adding the test like this? I think not. 

```js
// GoogleChart.test.js
import { Fragment } from '@wordpress/element';
import GoogleChart from './GoogleChart';
import { render } from '../../../tests/js/test-utils';

describe( 'GoogleChart', () => {
	it( 'renders two consecutive google charts without throwing when there is a `div` with id `google` in dom', () => {
		const divGoogle = document.createElement( 'div' );
		divGoogle.id = 'google';
		document.body.appendChild( divGoogle );

		const { container } = render(
			<Fragment>
				<GoogleChart
					chartType="LineChart"
					data={ [
						[ 'Age', 'Weight' ],
						[ 4, 5.5 ],
						[ 8, 12 ],
					] }
					width="100%"
					height="400px"
					legendToggle
				/>
				<GoogleChart
					chartType="LineChart"
					data={ [
						[ 'Age', 'Weight' ],
						[ 4, 5.5 ],
						[ 8, 12 ],
					] }
					width="100%"
					height="400px"
					legendToggle
				/>
			</Fragment>
		);
		expect(
			container.querySelectorAll( '.googlesitekit-chart' ).length
		).toEqual( 2 );
	} );
} );
```

aaemnnosttv (approved):
Thanks @kuasha420 ! I was hoping we could have added a test but unfortunately that isn't going to work like you said. This works great though 👍

Issue #4385, opened 2021-11-22
PHPUnit tests fail on local


eugene-manuilov:
## Bug Description

When I run phpunit tests on local, I see 3 tests fail because of discrepancy in used and expected domain names for the testing site:

```
1) Google\Site_Kit\Tests\Core\Authentication\AuthenticationTest::test_verify_proxy_setup_nonce
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'The link you followed has expired.</p><p><a href="http://example.org/wp-admin/admin.php?page=googlesitekit-splash">Please try again.</a>'
+'The link you followed has expired.</p><p><a href="http://example.com/wp-admin/admin.php?page=googlesitekit-splash">Please try again.</a>'

/var/www/html/wp-content/plugins/google-site-kit/tests/phpunit/integration/Core/Authentication/AuthenticationTest.php:257

2) Google\Site_Kit\Tests\Core\Authentication\AuthenticationTest::test_handle_proxy_permissions
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'The link you followed has expired.</p><p><a href="http://example.org/wp-admin/admin.php?page=googlesitekit-splash">Please try again.</a>'
+'The link you followed has expired.</p><p><a href="http://example.com/wp-admin/admin.php?page=googlesitekit-splash">Please try again.</a>'

/var/www/html/wp-content/plugins/google-site-kit/tests/phpunit/integration/Core/Authentication/AuthenticationTest.php:807

3) Google\Site_Kit\Tests\Core\Authentication\AuthenticationTest::test_invalid_nonce_error_sitekit_action
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'The link you followed has expired.</p><p><a href="http://example.org/wp-admin/admin.php?page=googlesitekit-splash">Please try again.</a>'
+'The link you followed has expired.</p><p><a href="http://example.com/wp-admin/admin.php?page=googlesitekit-splash">Please try again.</a>'

/var/www/html/wp-content/plugins/google-site-kit/tests/phpunit/integration/Core/Authentication/AuthenticationTest.php:905
```

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* PHPUnit tests should pass on local

## Implementation Brief

* In the `tests/phpunit/wp-tests-config.php`:
[...]

PR #4386, opened 2021-11-22
Bug/4385 phpunit tests on local

eugene-manuilov:
## Summary

Addresses issue #4385 

## Relevant technical choices

* Also updated the phpunit setup script to allow it to read the database connection information from environment variables used to run a wordpress container.

## Checklist

### DIFF
--- a/tests/phpunit/wp-tests-config.php
+++ b/tests/phpunit/wp-tests-config.php
@@ -23,15 +23,33 @@ define( 'ABSPATH', dirname( dirname( __DIR__ ) ) . '/vendor/roots/wordpress/' );
  * DO NOT use a production database or one that is shared with something else.
  */
 
+/**
+ * Returns an environment variable value if it exists, otherwise the default value.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $name Environment variable name.
+ * @param mixed $default A default value to use if the env variable is not set.
+ * @return mixed The environment variable value if it exsits, otherwise the default value.
+ */
+function get_env( $name, $default ) {
+	$value = getenv( $name );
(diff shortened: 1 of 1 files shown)


tests/phpunit/wp-tests-config.php
```diff
@@ -23,15 +23,33 @@
...
+/**
+ * Returns an environment variable value if it exists, otherwise the default value.
+ *
+ * @since n.e.x.t
+ *
+ * @param string $name Environment variable name.
+ * @param mixed $default A default value to use if the env variable is not set.
+ * @return mixed The environment variable value if it exsits, otherwise the default value.
+ */
+function get_env( $name, $default ) {
```
aaemnnosttv, line 35: Since this is only intended to be used here, how about we define this as a closure instead? Alternatively, we could add a namespace to the file. What do you think?

eugene-manuilov: > ...how about we define this as a closure instead?

@aaemnnosttv do you want to hide it from the global scope? Something like this?

```php
$get_env = function( $name, $default ) {
    ...
};

define( 'DB_USER', $get_env( 'WORDPRESS_DB_USER', 'root' ) );
```

aaemnnosttv: Yes that's what I meant 🙂 Either defining it as more of a temporary local function like this or namespacing the file to avoid adding a global function, but this looks fine to me 👍

eugene-manuilov: 👍 Updated.

aaemnnosttv (approved):
Great thanks @eugene-manuilov !

Issue #4289, opened 2021-10-27
Ensure pills for widget contexts that are empty/inactive are not displayed


felixarntz:
#4053 introduced a pill navigation for the unified dashboard that points to individual widget context "section" with anchor links. Per #4136, widget contexts can contextually be empty ("inactive") though, and in that case the respective pill should not be displayed.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The pill navigation introduced in #4053 should be updated as follows: For each of the pills, it should be checked whether the respective widget context is active, using the `isWidgetContextActive` selector from #4136. If it is not active, the pill should be skipped/not rendered.
    * For that purpose, each pill needs to be associated with the respective widget context (see `default-contexts.js`).
    * When on the main dashboard, the contexts should be the respective `CONTEXT_MAIN_DASHBOARD_*`.
    * When on the entity dashboard, the contexts should be the respective `CONTEXT_ENTITY_DASHBOARD_*`.

## Implementation Brief
* Using `assets/js/component/DashboardNavigation`,
	* use the `useDashboardType` hook to check whether we are on the main dashboard or entity dashboard.
	* import the `CONTEXT_MAIN_DASHBOARD_*` and `CONTEXT_ENTITY_DASHBOARD_*` constants from `assets/js/googlesitekit/widgets/default-contexts.js`
	* Based on the dashboard type, from the result of `useDashboardType`, query the `core/widgets` data store via the `isWidgetContextActive` selector, passing either the `CONTEXT_MAIN_DASHBOARD_*` or `CONTEXT_ENTITY_DASHBOARD_*` constant for every pill. Note that every pill is associated with the appropriate context. For e.g `CONTEXT_MAIN_DASHBOARD_TRAFFIC` and `CONTEXT_ENTITY_DASHBOARD_TRAFFIC` are associate to the `Traffic` tab.
	* Render the pill only if its appropriate context is active.

### Test Coverage
* No new tests to be added.

## QA Brief

* Enable unifiedDashboard feature flag.
* Make sure Adsense Module is not connected.
* In Main Dashboard, you should not see the Monetization Chip
* Connect Adsense Module. 
* The Monetization Chip Should be visible now.

## Changelog entry

* Update dashboard navigation to hide navigation chips for empty areas.

eugene-manuilov:
IB :heavy_check_mark:

eugene-manuilov:
@kuasha420 could you please add QAB?

[...]

PR #4388, opened 2021-11-22
Remove chips that have no corresponding widgets.

kuasha420:
## Summary

Addresses issue #4289 

## Checklist

### DIFF
--- a/assets/js/components/DashboardNavigation.js
+++ b/assets/js/components/DashboardNavigation.js
@@ -30,14 +30,69 @@ import { __ } from '@wordpress/i18n';
 /**
  * Internal dependencies
  */
+import Data from 'googlesitekit-data';
 import {
 	ANCHOR_ID_CONTENT,
 	ANCHOR_ID_MONETIZATION,
 	ANCHOR_ID_SPEED,
 	ANCHOR_ID_TRAFFIC,
 } from '../googlesitekit/constants';
+import { CORE_WIDGETS } from '../googlesitekit/widgets/datastore/constants';
+import {
+	CONTEXT_ENTITY_DASHBOARD_TRAFFIC,
+	CONTEXT_ENTITY_DASHBOARD_CONTENT,
+	CONTEXT_ENTITY_DASHBOARD_SPEED,
+	CONTEXT_ENTITY_DASHBOARD_MONETIZATION,
+	CONTEXT_MAIN_DASHBOARD_TRAFFIC,
+	CONTEXT_MAIN_DASHBOARD_CONTENT,
+	CONTEXT_MAIN_DASHBOARD_SPEED,
+	CONTEXT_MAIN_DASHBOARD_MONETIZATION,
+} from '../googlesitekit/widgets/default-contexts';
+import useDashboardType, {
+	DASHBOARD_TYPE_MAIN,
+} from '../hooks/useDashboardType';
+
+const { useSelect } = Data;
 
 export default function DashboardNavigation() {
+	const dashboardType = useDashboardType();
+
+	const [ showTraffic, showContent, showSpeed, showMonitization ] = useSelect(
+		( select ) => {
+			if ( dashboardType === DASHBOARD_TYPE_MAIN ) {
+				return [
+					select( CORE_WIDGETS ).isWidgetContextActive(
+						CONTEXT_MAIN_DASHBOARD_TRAFFIC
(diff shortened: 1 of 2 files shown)


assets/js/components/DashboardNavigation.js
```diff
@@ -30,14 +30,69 @@ import { __ } from '@wordpress/i18n';
...
+import useDashboardType, {
+	DASHBOARD_TYPE_MAIN,
+} from '../hooks/useDashboardType';
+
+const { useSelect } = Data;
 
 export default function DashboardNavigation() {
+	const dashboardType = useDashboardType();
+
+	const [ showTraffic, showContent, showSpeed, showMonitization ] = useSelect(
```
eugene-manuilov, line 60: Thanks, @kuasha420. This works as expected, but we don't encourage useSelect hooks with multiple selects (although we have some exception in the codebase). Could you, please, split it into four useSelect hooks for each individual chips?

```js
const showTraffic = useSelect( ( select ) => select( CORE_WIDGETS ).isWidgetContextActive(
    dashboardType === DASHBOARD_TYPE_MAIN
        ? CONTEXT_MAIN_DASHBOARD_TRAFFIC
        : CONTEXT_ENTITY_DASHBOARD_TRAFFIC
) );
```

eugene-manuilov (changes requested):

assets/js/components/DashboardNavigation.js
kuasha420: Done, Is there any reason for not reusing the useSelect call? Is there any correctness issue here? Using multiple select for same store seems kinda odd and verbose at first glance.

assets/js/components/DashboardNavigation.stories.js
```diff
@@ -19,13 +19,160 @@
...
+		<ViewContextProvider value={ viewContext }>
+			<DashboardNavigation { ...args } />
+		</ViewContextProvider>
+	</WithRegistrySetup>
+);
 
 export const DefaultDashboardNavigation = Template.bind( {} );
-DashboardNavigation.storyName = 'DashboardNavigation Default State';
-DashboardNavigation.args = {};
+DefaultDashboardNavigation.storyName = 'DashboardNavigation Default State';
```
eugene-manuilov, line 44: No need to repeat the `Dashboardnavigation` group name in every story title. Just use the `Default State` name for this story and `Monetization Hidden State` for the other story.

assets/js/components/DashboardNavigation.stories.js
```diff
@@ -19,13 +19,160 @@
...
+		} );
+		registry
+			.dispatch( CORE_WIDGETS )
+			.assignWidgetArea( 'SpeedArea', CONTEXT_MAIN_DASHBOARD_SPEED );
+		registry.dispatch( CORE_WIDGETS ).registerWidget( 'SpeedWidget', {
+			Component: () => <div>Speed Widget</div>,
+		} );
+		registry
+			.dispatch( CORE_WIDGETS )
+			.assignWidget( 'SpeedWidget', 'SpeedArea' );
```
eugene-manuilov, line 93: Looks like this part repeats for both stories. Let's extract it into a helper function (create one at the top of this file, after all imports), and use it in both stories to reduce duplicate code.

eugene-manuilov (changes requested):
Thanks, @kuasha420. This looks better to me. Added two small comments for the stories file. Once they are addressed, this PR should be good to go.

assets/js/components/DashboardNavigation.js
eugene-manuilov: There is nothing wrong with using multiple selects in the same hook. However, we have a convention to use a separate hook for each select because it makes it easier to read the code. If you think this should be changed, feel free to raise this question during our engineering sync call and we can discuss it.

assets/js/components/DashboardNavigation.stories.js
```diff
@@ -19,13 +19,123 @@
 /**
  * Internal dependencies
  */
+
```
eugene-manuilov, line 22:
```suggestion
```

eugene-manuilov (changes requested):
Thanks, @kuasha420. Looks good to me. Added one suggestion to remove an empty line. I'll apply it and approve this PR.

eugene-manuilov (approved):
LGTM

PR #4396, opened 2021-11-23
Fix AdSense widget registration.

tofumatt:
## Summary

Addresses issue #4348.

This ensures the AdSense widgets are registered when the Unified Dashboard feature flag is enabled.

Note: This diverges from the IB in that the `unifiedDashboard` feature flag checks are still intact. The reason for this is that some widgets share the same slug in both the Unified and "old" dashboard: https://github.com/google/site-kit-wp/blob/53a768c6a3bc353966f95371e6c03c816a61ec79/assets/js/modules/adsense/index.js#L164

Separating out the registrations allows us to change the sizes, priorities, etc. easier if needed for only one "kind" of dashboard, plus it makes the code _much_ more scannable/readable so we know which widgets should appear in which dashboard. It'll also make refactoring/removal of old dashboard code easier when the feature flag is removed.

So I've opted to keep the feature checks for everything except the adblocker, because it doesn't cause any harm now that the registration is fixed.

## Checklist

### DIFF
--- a/assets/js/modules/adsense/index.js
+++ b/assets/js/modules/adsense/index.js
@@ -97,11 +97,7 @@ export const registerWidgets = ( widgets ) => {
 			priority: 1,
 			wrapWidget: false,
 		},
-		[
-			isFeatureEnabled( 'unifiedDashboard' )
-				? AREA_MAIN_DASHBOARD_MONETIZATION_PRIMARY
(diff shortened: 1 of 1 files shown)


eugene-manuilov (approved):
LGTM! Thanks, @tofumatt.

Issue #2044, opened 2020-09-16
Register site on proxy service before redirecting to OAuth


felixarntz:
As part of the new user input flow, the onboarding on the proxy service will slightly change. Most importantly, the initial "Welcome" screen will no longer be present. The form on that screen normally registers or looks up a new site if that site is not (already) in possession of site ID and site secret. Per the removal of that step, the plugin must now take care of this registration call to the proxy service and afterwards redirect to the OAuth endpoint as part of the setup flow on its own.

**This is closely related to #4383 and should preferably be worked on in the same sprint.**

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

In the new version of the setup flow (`serviceSetupV2`), there is no longer a "Welcome" screen present on the authentication service when a user gets there for the Site Kit plugin setup; instead they land right away on the OAuth consent screen (before reaching the UI with the 3 steps as usual). The "Welcome" screen used to call another API route to register or look up the site on the service, so since that is no longer going to happen, this call has to be issued from the plugin going forward. A proof-of-concept implementation for parts of this and #2045 already exists [here](https://github.com/google/site-kit-wp/compare/poc/service-setup-v2-migration), so that branch will be referenced in a few places in the ACs below.

**All changes below should only apply if the `serviceSetupV2` feature flag is enabled.**

* The `Google_Proxy::OAUTH2_SITE_URI` endpoint on the service should going forward be used to register or look up the site whenever the plugin setup is started (i.e. on `Google_Proxy::ACTION_SETUP`).
    * Instead of using this route to only sync site fields if they are out of sync, this route should now be called _in any case_.
        * If the site is not connected, the route needs to be called to register or look up the site based on the site fields. From the plugin site there isn't much of a difference for whether the site gets registered (i.e. it has never used the service before) or just looked up (i.e. it has used the service before, but then e.g. was reset so that it doesn't have the site credentials anymore).
        * If the site is already connected, the route needs to be called with the site credentials, in which case it will have pretty much the same purpose as in the existing usage, to update any site fields out of sync.
        * The reason that this call now needs to _always_ happen, even in the "already connected" case above, is that the route will now return the proper URI for where to actually redirect the user to for the setup. It has always done that for the service, but now we need to use the same URI in the plugin. In other words, the new flow is to call that URI, retrieve the redirect URI returned, and then redirect the user to that URI (i.e. no longer use the `Google_Proxy::setup_url()` here).
[...]

PR #4411, opened 2021-11-26
Enhancement/4342 analytics step mode

eugene-manuilov:
## Summary

Addresses issue #4342 

## Checklist


.github/workflows/php-tests-wp-4-7-php-5-6.yml
```diff
@@ -71,7 +71,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-composer-
       - name: Composer Install
-        run: docker run --rm -v "$PWD:/app" -v "${{ steps.composer-cache.outputs.dir }}:/tmp/cache" composer install --no-interaction --no-progress # Run inside docker so composer install doesn't error...
+        run: docker run --rm -v "$PWD:/app" -v "${{ steps.composer-cache.outputs.dir }}:/tmp/cache" composer:2.1.11 install --no-interaction --no-progress # Run inside docker so composer install doesn't error...
```
eugene-manuilov, line 74: Looks like the newest version of Composer (released ~5 hours ago) contains backward-incompatible changes that cause a fatal error:

image

So, I have to update this command to make sure we use the latest compatible version of the `compose` image.

includes/Modules/Analytics.php
```diff
@@ -1343,4 +1345,21 @@ protected function find_property( array $properties, $property_id = '', array $u
...
+	/**
+	 * Adds mode=analytics-step to the proxy params if the serviceSetupV2 feature flag is enabled.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $params An array of Google Proxy setup URL parameters.
+	 * @return array Updated array with the mode=analytics-step parameter.
+	 */
+	private function update_propxy_setup_mode( $params ) {
+		if ( Feature_Flags::enabled( 'serviceSetupV2' ) ) {
```
felixarntz, line 1358: I believe this condition also needs to include the check for whether the module is not connected (see ACs). If the module is already connected, the `mode=analytics-step` should not be included.

includes/Modules/Analytics.php
```diff
@@ -1343,4 +1345,21 @@ protected function find_property( array $properties, $property_id = '', array $u
...
+
+	/**
+	 * Adds mode=analytics-step to the proxy params if the serviceSetupV2 feature flag is enabled.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $params An array of Google Proxy setup URL parameters.
+	 * @return array Updated array with the mode=analytics-step parameter.
+	 */
+	private function update_propxy_setup_mode( $params ) {
```
felixarntz, line 1357: Typo here.

felixarntz (changes requested):
@eugene-manuilov Two small things to address here, then it will be good to merge.

Aside: This will eventually be adjusted as part of #2044 - where a new filter in `Google_Proxy::get_metadata_fields()` should be used instead of the `googlesitekit_proxy_setup_url_params` filter (which should then be dropped).

includes/Modules/Analytics.php

includes/Modules/Analytics.php
eugene-manuilov: Good catch! Updated!

felixarntz (approved):
Great!

PR #4414, opened 2021-11-29
Update Pull Request template.

mitogh:
## Summary

Addresses issue #4424 

By using the list structure instead of a single line structure allow expanding into a better way automatically when GH expands the issue ID.

Image with the example of the automatic expansion from Github, the top link is the issue when using a list, the bottom is the current implementation. 

## Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,7 +1,9 @@
 ## Summary
 
-<!-- Please reference the issue this PR addresses. -->
-Addresses issue #
+<!-- Please reference the issue this PR addresses in the following list. -->
+Addresses issue:
+
+- #
 
 ## Relevant technical choices
 


aaemnnosttv:
Thanks for the PR @mitogh ! This looks like a nice change to me but one that we have to coordinate a little bit if we want to adopt it as we have a script for generating our changelogs for each release which would need to be updated first.

mitogh:
Makes sense I can do some review on those if we are looking to move forward with this change. @aaemnnosttv

felixarntz:
@mitogh Great idea, but like @aaemnnosttv says we need to make adjustments in our script for that first. Also, there should always be an issue to discuss a new enhancement, the pull-request is then only for writing and reviewing the code after it is defined. Could you open an issue where we can proceed with the communication on this?

mitogh:
Thanks @felixarntz @aaemnnosttv 

I've created a new issue to keep track of this:

- #4424

aaemnnosttv (changes requested):
Marking this as requiring changes to make it clear that we're not ready to merge this yet 👍

felixarntz (approved):
Good to merge now from my end, I'll assign to you @aaemnnosttv to eventually do so.

aaemnnosttv (approved):

Issue #4286, opened 2021-10-27
Learn More Link on Link Analytics and AdSense CTA opens in same window


wpdarren:
## Bug Description

On the `Link Analytics and AdSense` CTA, there's a `Learn more` link and it currently opens the page in the same window rather than opening a new one like our other CTAs.  

![image.png](https://images.zenhubusercontent.com/5f99965247ef5a1da6985d7d/f3a7da5a-2752-4cc7-88b2-d08549c4ae41)

## Steps to reproduce

1. Connect `AdSense` but do not link Analytics.
2. Go to `Site Kit Dashboard` and scroll down to `Earnings` widget.
3. See the CTA `Link Analytics and AdSense`
4. Click on the `Learn more` link and see behaviour.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The Learn more link on the CTA to link AdSense and Analytics should open in a separate window/tab, like other "Learn more" links.

## Implementation Brief

* Add the `ctaLinkExternal` prop to the `CTA` component rendered in `AdSenseLinkCTA`.
https://github.com/google/site-kit-wp/blob/1be77237454c18764f51aee4935f4dfac8232e12/assets/js/modules/analytics/components/common/AdSenseLinkCTA.js#L40-L48

### Test Coverage

* No new tests needed.
* Existing tests should pass.

## QA Brief

1. Connect `AdSense` but do not link Analytics.
2. Go to `Site Kit Dashboard` and scroll down to `Earnings` widget.
4. Click on the `Learn more` link and observe the link now it opens in a separate window.

## Changelog entry

* Update the "Learn More" links of Analytics and AdSense CTAs to open in a new window.

eugene-manuilov:
IB ✔️

wpdarren:
### QA Update: ✅ 

Verified: when you click on the Learn more link, it opens in a separate window.

PR #4425, opened 2021-11-29
`Learn More` links on the Adsense modules opens in a new window.

mitogh:
## Summary

Add property `ctaLinkExternal` to open the link in a separate window.

Addresses issue #4286 

## Relevant technical choices

- Updated the `CTA` component in order to register the property `ctaLinkExternal` for external callers.
- Added a default value for the property `ctaLinkExternal`  as `false` in order to be backwards compatible.

## Checklist

### DIFF
--- a/assets/js/components/notifications/CTA.js
+++ b/assets/js/components/notifications/CTA.js
@@ -76,6 +76,7 @@ CTA.propTypes = {
 	'aria-label': PropTypes.string,
 	error: PropTypes.bool,
 	onClick: PropTypes.func,
+	ctaLinkExternal: PropTypes.bool,
 };
 
 CTA.defaultProps = {
@@ -85,6 +86,7 @@ CTA.defaultProps = {
 	ctaLabel: '',
 	error: false,
 	onClick: () => {},
+	ctaLinkExternal: false,
 };
 
 export default CTA;
--- a/assets/js/components/notifications/CTA.js
+++ b/assets/js/components/notifications/CTA.js
@@ -72,11 +72,11 @@ CTA.propTypes = {
 	title: PropTypes.string.isRequired,
 	description: PropTypes.oneOfType( [ PropTypes.string, PropTypes.node ] ),
 	ctaLink: PropTypes.string,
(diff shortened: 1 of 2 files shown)


assets/js/modules/analytics/components/common/AdSenseLinkCTA.js
```diff
@@ -45,6 +45,7 @@ export default function AdSenseLinkCTA() {
 			) }
 			ctaLink={ supportURL }
 			ctaLabel={ __( 'Learn more', 'google-site-kit' ) }
+			ctaLinkExternal={ true }
```
eugene-manuilov, line 48: No need to explicitly pass `true` to properties, we can add property without a value to implicitly pass `true` to it.
```suggestion
			ctaLinkExternal
```

assets/js/components/notifications/CTA.js
```diff
@@ -76,6 +76,7 @@ CTA.propTypes = {
 	'aria-label': PropTypes.string,
 	error: PropTypes.bool,
 	onClick: PropTypes.func,
+	ctaLinkExternal: PropTypes.bool,
```
eugene-manuilov, line 79: Thanks for adding this :+1:. Could you please move it to go after the `ctaLink` prop?

assets/js/components/notifications/CTA.js
```diff
@@ -85,6 +86,7 @@ CTA.defaultProps = {
 	ctaLabel: '',
 	error: false,
 	onClick: () => {},
+	ctaLinkExternal: false,
```
eugene-manuilov, line 89: No need to set `false` by default for this property. It should be `undefined` if not set.

eugene-manuilov (changes requested):
Thanks, @mitogh. This looks almost good to me. Added a few nitpicks. Could you please check them out?

assets/js/components/notifications/CTA.js
mitogh: Sure.

eugene-manuilov (approved):
Thanks, @mitogh. LGTM

Issue #4107, opened 2021-09-22
Create separate Webpack configuration for Gutenberg-related entrypoints


felixarntz:
Site Kit has recently added its first integration with the Gutenberg editor (for the Idea Hub notice) and will soon require much more substantial integrations (particularly for the SwG module). This brings a number of new challenges based on the following:
* Site Kit is bundling its own versions of `@wordpress/*` dependencies to be cross-compatible and not rely on what WordPress bundles.
* Gutenberg specifically relies strongly on WordPress core's own bundled versions of those.
* In order to integrate with Gutenberg, those parts of Site Kit will need to rely on WordPress core's bundled versions as well - both for compatibility (so that it even works), but also from a best-practices perspective (e.g. preferably we shouldn't load our own React there).

This issue is the first out of a few that tackle the related problems and set us up for those integrations. It focuses on creating a separate Webpack configuration for any Gutenberg-only JS assets from Site Kit.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a separate Webpack configuration for entry points that integrate with Gutenberg/the block editor.
    * Those scripts should be able to access common dependencies bundled by both Site Kit and WordPress core (in potentially diferent versions). Potentially this part could even be helpful for non-Gutenberg scripts, but it's not crucial there for now at least. If it's more straightforward to fix this globally though, let's do it.
    * Code that imports `@wordpress/*` should continue to rely on Site Kit's bundled version (i.e. the dependency that ends up in the Site Kit bundle).
    * Code that imports `@wordpress-core/*` should rely on WordPress core's bundled version (i.e. in production code this would be an external for `wp.*`). In the source code, `@wordpress-core/*` is essentially an alias for the respective `@wordpress/*` dependency coming from WordPress core. When running the non-production code (e.g. in tests), this should rely on those actual dependencies.
        * It would be nice to "double-bundle" them with different aliases (specifically related to #4108, where it would be nice to have e.g. Jest tests run with WP 5.0-bundled versions of those dependencies), but it's unclear whether that's possible.
    * Non of those scripts should have React injected.
    * In general, any of those scripts that only rely on `@wordpress-core/*` imports should not have any other Site Kit script loaded as a dependency.
* The `googlesitekit-idea-hub-notice` script (which right now is the only one that this applies to) should be adjusted to use the above Webpack configuration. The code of the file itself should also be updated to rely on `@wordpress-core/*` imports instead of `global.wp`.
[...]

PR #4435, opened 2021-12-01
Update webpack config for Gutenberg entry points

kuasha420:
## Summary

Addresses issue #4107.

## Checklist


kuasha420:
WIP: This is as far as I've come so far. I think I have messed up somewhere in the config as the idea hub notice is not showing up in the Editor at the moment, but can't quite figure out what's wrong. Any pointer would be appreciated and I will add QAB later.

webpack.config.js
```diff
@@ -150,6 +150,171 @@ const GOOGLESITEKIT_VERSION = googleSiteKitVersion
...
+				} ),
+			],
+			/*
+                 The runtimeChunk value 'single' creates a runtime file to be shared for all generated chunks.
+                 Without this, imported modules are initialized for each runtime chunk separately which
+                 results in duplicate module initialization when a shared module is imported by separate entries
+                 on the same page.
+                 See: https://v4.webpack.js.org/configuration/optimization/#optimizationruntimechunk
+             */
+			// runtimeChunk: 'single',
```
tofumatt, line 298:
```suggestion
```suggestion
		// The runtimeChunk value 'single' creates a runtime file to be shared for all generated chunks.
		// Without this, imported modules are initialized for each runtime chunk separately which
		// results in duplicate module initialization when a shared module is imported by separate entries
		// on the same page.
		// See: https://v4.webpack.js.org/configuration/optimization/#optimizationruntimechunk
		runtimeChunk: 'single',
```

Why is this commented out? I assume with the explanation above this was intended to stay? Or should this all be removed?

webpack.config.js
```diff
@@ -461,5 +626,15 @@ module.exports.default = ( env, argv ) => {
 		} );
 	}
 
+	// Build Gutenberg Entrypoints
```
tofumatt, line 629:
```suggestion
	// Build Gutenberg entrypoints.
```

tofumatt (changes requested):
This is looking good but I'm not sure about a few things that are commented out, and I'm not sure what still needs doing.

Can you leave a comment in the PR with what remains or what you want help with? Happy to take a look but I'm not quite clear on what's remaining or what you want another set of eyes on 🙂

kuasha420:
@tofumatt thank you for reviewing. As you've noticed, the PR has all the bits that are mentioned in the AC/IB. Only major problem is that it is not working :stuck_out_tongue_winking_eye: 

:heavy_check_mark: the modules are generated and enqueued correctly. 
:heavy_check_mark: site kit dashboard works fine.
:heavy_multiplication_x: the gutenberg idea hub integration does not work, does not show any error in console either. 

I am not sure whats wrong here, I have tried different things without success. If you have any question let me know. We can maybe jump on a small call to figure this out if you can manage time. 

Back to you. Cheers.

tofumatt (commented):
Just a note that I can see this script loading in the network tab in Chrome, and it's being properly compiled by Webpack, but it doesn't ever load _anything_ in the script.

I'll look into this more to find out what's up; because the code here seems good. 🤔 

EDIT: Looks like it might be that a lot of stuff was duplicated, including the `vendor` chunk code here: https://github.com/google/site-kit-wp/pull/4435/files#diff-1fb26bc12ac780c7ad7325730ed09fc4c2c3d757c276c3dacc44bfe20faf166fR283-R302

I'll push an update that fixes it as soon as I confirm that doesn't break anything else 😅

tofumatt (commented):
I've made some changes to the webpack config so the Gutenberg code still runs properly.

I think it looks good, but best for someone else to review this as I've changed a fair bit of the PR at this point 🙂

webpack.config.js
```diff
@@ -150,6 +150,51 @@ const GOOGLESITEKIT_VERSION = googleSiteKitVersion
...
+	gutenbergExternals[ `@wordpress-core/${ name }` ] = [
+		'wp',
+		camelCaseDash( name ),
+	];
+} );
+
+const gutenbergEntryPoints = {
+	'googlesitekit-idea-hub-notice':
+		'./assets/js/googlesitekit-idea-hub-notice.js',
+};
```
eugene-manuilov, line 196: Same here. Why do we need it here? Let's move it to the config declaration to be consistent with the approach we use for other entries.

webpack.config.js
```diff
@@ -404,6 +445,74 @@ function* webpackConfig( env, argv ) {
...
 		],
 	};
+
+	// Build the Gutenberg entry points.
+	yield {
+		entry: gutenbergEntryPoints,
+		externals: gutenbergExternals,
+		output: {
+			filename:
+				mode === 'production' ? '[name]-[contenthash].js' : '[name].js',
```
eugene-manuilov, line 455: I see we have many `mode === 'production'` occurrences in this file. How about assigning it to a variable and using the variable instead of repeating this condition every time?

```js
const isProduction = mode === 'production;

...

filename: isProduction ? '[name]-[contenthash].js' : '[name].js',
chunkFilename: isProduction ? '[name]-[chunkhash].js' : '[name].js',

...
```

webpack.config.js
```diff
@@ -404,6 +445,74 @@ function* webpackConfig( env, argv ) {
...
+						// We preserve function names that start with capital letters as
+						// they're _likely_ component names, and these are useful to have
+						// in tracebacks and error messages.
+						keep_fnames: /__|_x|_n|_nx|sprintf|^[A-Z].+$/,
+						output: {
+							comments: /translators:/i,
+						},
+					},
+					extractComments: false,
+				} ),
```
eugene-manuilov, line 511: Looks like this plugin is identical to the one we already use in the main config. Let's assign TerserPlugin options to a variable and use it for both instances. Or maybe create a function that will return a new instance of the TerserPlugin plugin. Up to you.

eugene-manuilov (changes requested):
Thanks, @kuasha420 and @tofumatt. This looks great. Added some nitpicks. @kuasha420 could you, please, take a look at them? Once addressed, this will be good to go.

eugene-manuilov (approved):
Thanks, @kuasha420. Looks good to me.

Issue #4028, opened 2021-09-13
Fix too much spacing between Optimize input field and errors


felixarntz:
## Bug Description

See https://google.github.io/site-kit-wp/storybook/main/?path=/story/optimize-module-setup--start-with-invalid-values: The space between the Optimize ID field and the error text below is way too much. This is likely happening because the error text is placed outside the `.googlesitekit-setup-module__inputs` wrapper, which shouldn't be the case as it includes extra margin etc.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* In the Optimize setup (and settings as well), any input error text should have the correct space from its associated input field, specifically an error for the Optimize ID shouldn't have the extra margin that it currently has (compared to e.g. an error for the AMP experiment JSON field).

## Implementation Brief

* The extra space is caused by the following CSS Rules in the `assets/sass/components/setup/_googlesitekit-setup-module.scss` file. 
* https://github.com/google/site-kit-wp/blob/8bff83f9f11742dee8aca49b04f0f55121d040e1/assets/sass/components/setup/_googlesitekit-setup-module.scss#L109-L115
* Add new CSS Rules that overrides and sets the bottom margin (ie. `margin-bottom: 0`) to the `div.mdc-text-field--error` and the following helper text `div.mdc-text-field-helper-line` elements inside `div.googlesitekit-setup-module__inputs` of `form.googlesitekit-optimize-setup__form`.
    * We can keep the change limited to the Optimize ID input field by targeting the div with `[name="optimizeID"]` attribute selector. 

### Test Coverage

* N/A

### Visual Regression Changes

* May need updating. 

## QA Brief
* Check whether spacing has been reduced in the following story: https://google.github.io/site-kit-wp/storybook/main/?path=/story/optimize-module-setup--start-with-invalid-values

## Changelog entry

* Fix styles for errors on Optimize setup and settings forms.

eugene-manuilov:
Moving back to the IB column since it has been moved to the Execution column by accident.

eugene-manuilov:
@kuasha420 let's not remove the existing css rules because it will affect something else, instead, left's add one more css rule that will fix the issue for the Optimize module. We can use `.googlesitekit-optimize-setup__form` and `.googlesitekit-setup-module--optimize` selectors to restrict style changes to work only for the Optimize module.

kuasha420:
[...]

PR #4437, opened 2021-12-01
Use negative margin to position error text correctly.

asvinb:
## Summary

Addresses issue #4028 

## Relevant technical choices
If we set `margin-bottom: 0` to the elements as per the IB, this causes UI issues since there might be other elements besides the input which we removed the margin as noted by QA. The solution is to use negative margin on the error text to make sure we are not introducing regressions.

## Checklist

- [z] I have added a QA Brief on the issue linked above.


eugene-manuilov (approved):
LGTM

PR #4444, opened 2021-12-02
Add individual loading icons for Idea Hub interaction.

asvinb:
## Summary

Addresses issue #3907 

## Checklist


assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.stories.js
```diff
@@ -235,6 +240,44 @@ DataUnavailableAll.decorators = [
...
+				);
+			registry
+				.dispatch( MODULES_IDEA_HUB )
+				.setActivity(
+					'ideas/7612031899179595408',
+					IDEA_HUB_ACTIVITY_CREATING_DRAFT
+				);
+		};
+
+		enabledFeatures.clear();
```
aaemnnosttv, line 268: This shouldn't be necessary as enabled features should be cleared before every story automatically 🙂

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.stories.js
```diff
@@ -235,6 +240,44 @@ DataUnavailableAll.decorators = [
...
+		enabledFeatures.add( 'ideaHubModule' );
+		mockEndpoints();
+
+		const registry = bootstrapRegistry();
+		return (
+			<WithTestRegistry registry={ registry } callback={ setupRegistry }>
```
aaemnnosttv, line 274: This should use `WithRegistrySetup` as `WithTestRegistry` is more for Jest now (also perhaps why features may not be enabled as expected).

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.stories.js
asvinb: Thanks for catching that @aaemnnosttv . I updated all the stories.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Idea.js
```diff
@@ -253,6 +329,16 @@ Idea.propTypes = {
 	buttons: PropTypes.arrayOf( PropTypes.string ).isRequired,
 };
 
+Idea.notices = {
```
aaemnnosttv, line 332: I like the idea of using a map for this but let's not attach it to the component as a static property unless it's needed outside of the component (in which case it could also maybe just be an import).

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.stories.js
```diff
@@ -72,189 +77,184 @@ const mockEndpoints = ( args ) => {
...
+			.setActivity(
+				'ideas/7612031899179595408',
+				IDEA_HUB_ACTIVITY_DRAFT_CREATED
+			);
+		registry
+			.dispatch( MODULES_IDEA_HUB )
+			.setActivity(
+				'ideas/2285812891948871921',
+				IDEA_HUB_ACTIVITY_UNPINNED
+			);
```
aaemnnosttv, line 223: I think we've shyed away from doing this in the past but it's quite common in Gutenberg.
Let's grab the `setActivity` action creator once rather than calling `dispatch` repeatedly. The same could be applied elsewhere here too. 

```suggestion
		const { setActivity } = registry.dispatch( MODULES_IDEA_HUB );
		setActivity(
			'ideas/17450692223393508734',
			IDEA_HUB_ACTIVITY_DELETED
		);
		setActivity(
			'ideas/14025103994557865535',
			IDEA_HUB_ACTIVITY_PINNED
		);
		setActivity(
			'ideas/7612031899179595408',
			IDEA_HUB_ACTIVITY_DRAFT_CREATED
		);
		setActivity(
			'ideas/2285812891948871921',
			IDEA_HUB_ACTIVITY_UNPINNED
		);
```

assets/js/modules/idea-hub/datastore/idea-state.js
```diff
@@ -280,6 +249,66 @@ const baseActions = {
...
+
+	/**
+	 * Removes an idea from the list of newIdeas state variable.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {string} name Idea name.
+	 * @return {Object} Redux-style action.
+	 */
+	removeIdeaFromNewIdeas( name ) {
```
aaemnnosttv, line 301: I think we can simplify this a little bit – at least in terms of the actions that go through the reducer and overall duplication here.

How about this:
- Add lower level actions here for simply `addIdeaToList(name, list)` and `removeIdeaFromList(name, list)` where `list` would be the state key (`newIdeas`, `savedIdeas` as a string)
- Refactor these higher level action creators to use these add/remove actions inside (since every move is a remove(name, list-a) + add(name, list-b)), but we also have an action here which simply removes
- Update the reducer in terms of the primitive action creators only as the higher functions here would simply yield the primitive

E.g. `moveIdeaFromSavedIdeasToNewIdeas` could be reimplemented like so

```js
function *moveIdeaFromSavedIdeasToNewIdeas( name ) {
	yield removeIdeaFromList( name, 'savedIdeas' );
	yield addIdeaToList( name, 'newIdeas' );
}
```

The only tricky part here is that the validation won't really work in a generator, which is why we have `createValidatedAction` 👍

aaemnnosttv (changes requested):
Thanks @asvinb ! This looks good and works well overall. I left a few comments about how we can clean things up a bit more now that the activity buttons are more/less the same as well as a few other suggestions. Nice work!

assets/js/modules/idea-hub/datastore/idea-state.js
asvinb: @aaemnnosttv Good suggestion. For removing ideas from a list, that's all good. However for adding an idea to a list, it'll require some extra steps to know from which list the idea comes from so that we can grab all the details, since we are only passing the idea name around. So technically, adding an idea would require 3 parameters, the list it's currently in, the list it'll be moved to and the name. So somewhere we must still specify the list names, and I'am wondering whether it'll be simpler to keep the newly added actions from the PR. Let me know what you think 😁

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Idea.js
```diff
@@ -66,36 +72,80 @@ export default function Idea( props ) {
...
 		removeIdeaFromNewAndSavedIdeas,
+		moveIdeaFromSavedIdeasToNewIdeas,
+		moveIdeaFromNewIdeasToSavedIdeas,
 	} = useDispatch( MODULES_IDEA_HUB );
 
 	const activity = useSelect( ( select ) =>
 		select( MODULES_IDEA_HUB ).getActivity( name )
 	);
 
+	const clearActivity = useCallback(
```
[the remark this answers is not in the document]
asvinb: 👍 I thought maybe it'll reduce the amount of duplicated code by actually having `removeActivity` within the function, but updated as per your suggestion!

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Idea.js
```diff
@@ -104,32 +154,55 @@ export default function Idea( props ) {
...
+		if ( ! showNotice ) {
+			return null;
+		}
+
+		return (
+			<div className="googlesitekit-idea-hub__loading-notice">
+				<p>{ Idea.notices[ activity ] }</p>
+			</div>
+		);
+	};
```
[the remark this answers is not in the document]
asvinb: 👍

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Idea.js
```diff
@@ -145,29 +218,21 @@ export default function Idea( props ) {
...
-									activity === IDEA_HUB_ACTIVITY_IS_PROCESSING
+									activity === IDEA_HUB_ACTIVITY_IS_DELETING
 								}
-								icon={ <DeleteIcon /> }
+								icon={ getIcon(
+									IDEA_HUB_ACTIVITY_IS_DELETING,
+									DeleteIcon
+								) }
 								title={ __( 'Dismiss', 'google-site-kit' ) }
 							/>
```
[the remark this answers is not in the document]
asvinb: `IdeaActivityButton` component added. However I'am not sure it really reduces the amount of duplicated code. Can you take a look and let me know what you think?

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/index.stories.js
asvinb: Good catch. Thanks @aaemnnosttv

assets/js/modules/idea-hub/datastore/idea-state.js
aaemnnosttv: Ah good point, it looked like the name was enough so I thought maybe the lists were only using the name as references to the objects.

It's still possible as proposed, the only change would be we would want the lower level actions to pass the `idea` object rather than names. The higher level action can use `name` instead, with an additional action + selector for getting the `idea` by name. I wasn't proposing we do away with the actions we have now, but simplify it with composition to avoid duplication and extra logic in the reducers.

I created a quick POC with this proposal here https://github.com/google/site-kit-wp/compare/enhancement/3907-loading-icons...poc/3907-variant

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/utils.js
```diff
@@ -0,0 +1,43 @@
...
+
+const activityIconMap = {
+	[ IDEA_HUB_BUTTON_DELETE ]: <DeleteIcon />,
+	[ IDEA_HUB_BUTTON_PIN ]: <PinIcon />,
+	[ IDEA_HUB_BUTTON_UNPIN ]: <UnpinIcon />,
+	[ IDEA_HUB_BUTTON_CREATE ]: <CreateIcon />,
+};
+
+export const getIconFromActivity = ( activity ) => {
+	return activityIconMap[ activity ] || Null;
```
aaemnnosttv, line 42: The map returns an element but the fallback here returns a component type.
```suggestion
	return activityIconMap[ activity ] || <Null />;
```

aaemnnosttv (commented):
Thanks @asvinb! I left a few more comments below, [one of which](https://github.com/google/site-kit-wp/pull/4444/files#r768891619) has a branch I made from yours as a POC for thought.

Let me know what you think?

assets/js/modules/idea-hub/datastore/idea-state.test.js
```diff
@@ -213,5 +217,113 @@ describe( 'modules/idea-hub idea-state', () => {
...
+					.receiveGetSavedIdeas( [], {} );
+
+				expect(
+					registry.stores[ MODULES_IDEA_HUB ].store.getState()
+						.newIdeas
+				).toEqual( fixtures.newIdeas );
+				expect(
+					registry.stores[ MODULES_IDEA_HUB ].store.getState()
+						.savedIdeas
+				).toEqual( [] );
```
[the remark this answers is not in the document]
asvinb: 👍

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Idea.js
```diff
@@ -42,19 +43,34 @@ import {
...
+		'google-site-kit'
+	),
+	IDEA_HUB_ACTIVITY_DELETED: __( 'Idea dismissed', 'google-site-kit' ),
+};
+const waitForActivity = () =>
+	new Promise( ( resolve ) => {
+		setTimeout( () => {
+			resolve();
+		}, ACTIVITY_TIMER );
+	} );
```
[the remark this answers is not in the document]
asvinb: Thanks for catching that. I missed it due to the annotation 🤦 
image

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Idea.js
```diff
@@ -203,13 +248,14 @@ export default function Idea( props ) {
...
 								className="googlesitekit-idea-hub__actions--create"
-								onClick={ handleCreate }
-								disabled={
-									activity === IDEA_HUB_ACTIVITY_IS_PROCESSING
+								inProgress={
+									activity ===
+									IDEA_HUB_ACTIVITY_CREATING_DRAFT
 								}
-								icon={ <CreateIcon /> }
+								onClick={ handleCreate }
```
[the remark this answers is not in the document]
asvinb: @aaemnnosttv I think adding different activity button might result in having some unnecessary extra files. I updated the PR so that `IdeaActivityButton` is a bit more smart, getting most of its information from maps. We can even have the handlers defined in a map and then call them accordingly, but I think we'll just moving most of the code from `Idea` to `IdeaActivityButton` which will result in the same issue, i.e a large component.
Let me know what you think, happy to go with your decision 😁 even though am not a big fan of having those individual button components.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/utils.js
asvinb: 👍

assets/js/modules/idea-hub/datastore/idea-state.js
asvinb: Thanks @aaemnnosttv . Makes sense! I've merged your POC into the current branch.

aaemnnosttv (approved):
Thanks @asvinb! This looks good to me. I pushed a few small changes around the parts from my POC and added a few missing tests.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Idea.js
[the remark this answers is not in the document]
aaemnnosttv: I was thinking about this initially too. This works well for now I think although I'm still somewhat thinking keeping the components separate might be better. I don't feel very strongly about it though so we can always revisit if needed 👍

assets/js/modules/idea-hub/datastore/idea-state.js
```diff
@@ -280,8 +247,69 @@ const baseActions = {
...
+	};
+}
+
+function removeIdeaFromList( idea, list ) {
+	return {
+		payload: { idea, list },
+		type: REMOVE_IDEA_FROM_LIST,
+	};
+}
+
```
eugene-manuilov, line 312: These functions should be actions. Please, move them to the corresponding baseActions objects.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Idea.js
```diff
@@ -66,36 +66,69 @@ export default function Idea( props ) {
...
 	const activity = useSelect( ( select ) =>
 		select( MODULES_IDEA_HUB ).getActivity( name )
 	);
 
 	const handleDelete = useCallback( async () => {
-		setActivity( name, IDEA_HUB_ACTIVITY_IS_PROCESSING );
+		setActivity( name, IDEA_HUB_ACTIVITY_IS_DELETING );
 		await dismissIdea( name );
-		removeActivity( name );
+		setActivity( name, IDEA_HUB_ACTIVITY_DELETED );
```
eugene-manuilov, line 82: Now we have separate activities per action and can move `setActivity` calls to the corresponding actions:

```diff
 	*dismissIdea( ideaName ) {
 		invariant( typeof ideaName === 'string', 'ideaName must be a string.' );

 		yield clearError( 'dismissIdea', [ ideaName ] );
+		yield baseActions.setActivity( ideaName, IDEA_HUB_ACTIVITY_IS_DELETING );

 		const { response, error } = yield baseActions.updateIdeaState( {
 			name: ideaName,
 			dismissed: true,
 		} );

 		if ( error ) {
 			yield receiveError( error, 'dismissIdea', [ ideaName ] );
 		}

+		yield baseActions.setActivity( ideaName, IDEA_HUB_ACTIVITY_DELETED );

 		return { response, error };
 	},
```

Same for other activities.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Idea.js
```diff
@@ -66,36 +66,69 @@ export default function Idea( props ) {
...
 
 		trackEvent( IDEA_HUB_GA_CATEGORY_WIDGET, 'dismiss_idea' );
-	}, [ name, dismissIdea, setActivity, removeActivity ] );
+
+		await waitForActivity();
+		removeActivity( name );
```
eugene-manuilov, line 87: Probably this is also can be moved into the actions as well.

eugene-manuilov (changes requested):
Great work, @asvinb @aaemnnosttv. In general, looks good to me. Added minor comments regarding where things should be defined. @asvinb, could you, please, take a look?

assets/js/modules/idea-hub/datastore/idea-state.js
aaemnnosttv: These are low-level utility actions that we don't particularly want/need to expose externally. This way we can still use them for our needs here without making them into fully bound action creators.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Idea.js
aaemnnosttv: My general thought is that it's preferable to keep delays like this in the component as it's more of a presentational concern rather than waiting for data. Moving it into the store would require creating a control for it as well.

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/Idea.js
aaemnnosttv: @eugene-manuilov this looks like a great enhancement 👍  Would we also need to set/restore the activity if there was an error updating the idea state?

asvinb: @aaemnnosttv @eugene-manuilov I think it'll be good if we reset the activity in case there's an error. I've updated the PR to cater for that.

asvinb:
@eugene-manuilov PR updated and I'll let you review @aaemnnosttv 's comments above as well.

eugene-manuilov (changes requested):

assets/js/modules/idea-hub/datastore/idea-state.js
eugene-manuilov: > we don't particularly want/need to expose externally

@aaemnnosttv why? We have similar utility actions in other modules defined as regular actions. I think there is nothing wrong in keeping these utility functions as actual actions.

aaemnnosttv: They're not any less actions as `baseActions`, they're just not registered as bound action creators (i.e. `dispatch(...)`). We have quite a few of these in the codebase but they're using an inline literal form (search for `yield {`) – these just have a function.

I'm not strongly opposed to adding them to be dispatchable like the others, but it isn't a novel concept/pattern in the datastore. If we were to move them then they should also get tests for each which is something every bound action should get. Right now they're covered implicitly by the higher actions that use them, so it isn't as if they're untested.

Thoughts @felixarntz ?

assets/js/modules/idea-hub/datastore/idea-state.js
```diff
@@ -168,8 +144,11 @@ const baseActions = {
 
 		if ( error ) {
 			yield receiveError( error, 'saveIdea', [ ideaName ] );
+			yield baseActions.removeActivity( ideaName );
 		}
 
+		yield baseActions.setActivity( ideaName, IDEA_HUB_ACTIVITY_PINNED );
```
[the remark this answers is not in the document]
asvinb: 👍

assets/js/modules/idea-hub/components/dashboard/DashboardIdeasWidget/utils.js
```diff
@@ -0,0 +1,90 @@
...
+	IDEA_HUB_ACTIVITY_UNPINNED: __(
+		'Idea removed from saved',
+		'google-site-kit'
+	),
+	IDEA_HUB_ACTIVITY_DELETED: __( 'Idea dismissed', 'google-site-kit' ),
+};
+
+export const getIconFromActivity = ( activity ) => {
+	return activityIconMap[ activity ] || <Null />;
+};
```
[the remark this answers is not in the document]
asvinb: I agree with you @eugene-manuilov . The `getIconFromActivity` function is redundant and we can have the same logic within the component. Updated!

eugene-manuilov (approved):
Ok, looks good to me. Thanks, @asvinb.

Issue #4383, opened 2021-11-19
Add support for new proxy `action` names used by v2 proxy


felixarntz:
Historically, the proxy has used the `Google_Proxy::ACTION_SETUP` name for the `action` query parameter that it uses to pass certain data back to Site Kit as part of some intermediate redirects during the plugin setup flow. This has been somewhat problematic since it uses a single `action` for multiple interactions, and it's even worse because the same action is also used by the plugin internally to redirect the user from an "internal" setup request to the proxy setup.

Since we are now making a clean break by migrating to a `v2/site-management/setup/` action, this is a great point to resolve this mess. The proxy will now, whenever the new route is used, send this data through distinct `action` query parameters, which this issue aims at adding support for.

**The work in this issue is closely related to #2044 and should preferably be worked on in the same sprint.**

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Instead of handling all intermediate proxy redirects via the `googlesitekit_proxy_setup` (`Google_Proxy::ACTION_SETUP`) action (which is also used by the plugin internally for another purpose), this should now be taken care of in a more granular way _if_ the `serviceSetupV2` feature flag is enabled:
    * For the request to verify the site (includes verification token), the proxy now sends a new `googlesitekit_proxy_verify` action.
    * For the request to purely exchange the site code for the actual site credentials, the proxy now sends a new `googlesitekit_proxy_exchange_site_code` action.
* The logic that is currently in place to handle both of these interactions can pretty much remain the same. No parameters sent are changed, it's purely the `action` query parameter that has the names updated as above. However, having these two distinct action allows us to clean up the code quite a bit and split the handlers more appropriately - up to the IB to solve this nicely. :)
* The only other behavioral change that needs to be covered here is the following: Both actions should check for an optional `step` query parameter sent from the proxy. If that is received, its value should be added to the redirect URL to send back to as well. (This is typically going to be `verification` for the `googlesitekit_proxy_verify` action and `delegation_consent` for the `googlesitekit_proxy_exchange_site_code` action.)
* [This commit in the proof-of-concept branch](https://github.com/google/site-kit-wp/commit/ad097bf0269eb270d9a805e4266d1ee24a28abb2) pretty much contains the changes needed here, however in a somewhat quick and hacky implementation - looking at this can still serve as inspiration or to better clarify the requirements.
[...]

PR #4451, opened 2021-12-03
Extract setup into classes for v1 + v2 and support for new action names

aaemnnosttv:
## Summary

Addresses issue #4383 

* The changes here look a bit larger than they are, most of the diff are things moving from `Authentication` and/or its test into the new classes

## Relevant technical choices

* Adds a new helper `subscribe_to_wp_http_requests` method to the base `TestCase` which takes a function to be called for each request made via WP HTTP to help with making assertions
* Made a tiny fix to the phpcs config to avoid linting files it shouldn't and ensure the `Manifest` class is no longer ignored
* Tweaked the `Fake_Site_Connection_Trait` to add methods for getting fake credentials without applying them (e.g. for use in a response stub from the proxy)

## Checklist


includes/Core/Assets/Manifest.php
```diff
@@ -38,7 +38,7 @@ class Manifest {
 	 */
 	public static function get_filename( $handle ) {
 		if ( is_null( self::$assets ) ) {
-			$path = Plugin::instance()->context()->path( "dist/manifest.php" );
+			$path = Plugin::instance()->context()->path( 'dist/manifest.php' );
```
aaemnnosttv, line 41: Previously this class was being ignored by the phpcs config and I noticed it was still ignored (even though it's no longer generated) when phpcs was failing on the now generated `dist/manifest.php` file (which is now ignored 😄 )

includes/Core/Authentication/Setup_V1.php
```diff
@@ -0,0 +1,86 @@
...
+
+		$this->handle_verification();
+
+		if ( $code && $site_code ) {
+			$this->handle_site_code( $code, $site_code );
+		}
+
+		if ( $redirect_url ) {
+			$this->user_options->set( OAuth_Client::OPTION_REDIRECT_URL, $redirect_url );
+		}
```
aaemnnosttv, line 82: ⚠️ This is not yet implemented in v2 as I wasn't sure which action it should be added to or if it's even needed at all.

tests/phpunit/includes/Fake_Site_Connection_Trait.php
```diff
@@ -20,52 +20,84 @@
 trait Fake_Site_Connection_Trait {
```
aaemnnosttv, line 20: Changes to this trait were not _strictly_ necessary but I think this is a worthwhile enhancement that makes the implementation a bit cleaner while adding methods for getting credentials without "applying" them.

tests/phpunit/integration/Core/Authentication/AuthenticationTest.php
```diff
@@ -239,93 +239,6 @@ public function test_register_allowed_redirect_hosts() {
 		$this->assertEquals( 'https://sitekit.withgoogle.com', wp_validate_redirect( 'https://sitekit.withgoogle.com' ) );
 	}
 
-	public function test_verify_proxy_setup_nonce() {
```
aaemnnosttv, line 242: These moved to the V1Test more/less unchanged.

aaemnnosttv:
@felixarntz please see [my comment about the redirect URL above](https://github.com/google/site-kit-wp/pull/4451#discussion_r762017859), otherwise this should be ready for your review!

includes/Core/Authentication/Setup_V1.php
felixarntz: @aaemnnosttv Yes this is still needed. It will be required only for the action to start the setup, not for the intermediate proxy redirects.

includes/Core/Authentication/Setup.php
```diff
@@ -0,0 +1,196 @@
...
+						$params['site_code'] = $site_code;
+						return $params;
+					}
+				);
+				return;
+			}
+
+			$this->user_options->set( OAuth_Client::OPTION_ERROR_CODE, $error_message );
+			wp_safe_redirect( $this->context->admin_url( 'splash' ) );
+			exit;
```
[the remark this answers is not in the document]
aaemnnosttv: The problem with inlining this into the handler is that it's used in both code exchange and verify handlers in v2 (as well as v1) so I don't think we want to duplicate this in each.

I think throwing different exceptions here might be cleaner 🤔

includes/Core/Authentication/Setup_V2.php
```diff
@@ -0,0 +1,148 @@
...
+		if ( ! current_user_can( Permissions::SETUP ) ) {
+			wp_die( esc_html__( 'You don\'t have permissions to set up Site Kit.', 'google-site-kit' ), 403 );
+		}
+
+		if ( ! $code ) {
+			wp_die( esc_html__( 'Invalid request.', 'google-site-kit' ), 400 );
+		}
+
+		if ( ! $verification_token || ! $verification_method ) {
+			wp_die( 'Verifying site ownership requires a token and verification method.', 400 );
```
felixarntz, line 93: This should probably also be wrapped in `esc_html__`?

includes/Core/Authentication/Setup_V2.php
```diff
@@ -0,0 +1,148 @@
...
+
+		$this->handle_verification( $verification_token, $verification_method );
+
+		// If the site does not have a site ID yet, a site code will be passed.
+		// Handling the site code here will save the extra redirect from the proxy if successful.
+		if ( $site_code ) {
+			try {
+				$this->handle_site_code( $code, $site_code );
+			} catch ( Missing_Verification_Exception $exception ) {
+				$this->redirect_to_proxy( $code, compact( 'site_code' ) );
```
felixarntz, line 104: I think here it shouldn't redirect right away, but instead add the `site_code` to the query parameters list. Otherwise, it would omit the `step` query parameter that's only added in the redirect below.

includes/Core/Authentication/Setup_V2.php
```diff
@@ -0,0 +1,148 @@
...
+		}
+
+		if ( ! $code || ! $site_code ) {
+			wp_die( esc_html__( 'Invalid request.', 'google-site-kit' ), 400 );
+		}
+
+		try {
+			$this->handle_site_code( $code, $site_code );
+		} catch ( Missing_Verification_Exception $exception ) {
+			$this->redirect_to_proxy( $code, compact( 'site_code' ) );
```
felixarntz, line 141: See above.

felixarntz (commented):
@aaemnnosttv Almost lgtm! Only a few minor things worth addressing.

includes/Core/Authentication/Setup.php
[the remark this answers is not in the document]
felixarntz: +1 to using exceptions, that's indeed cleaner!

felixarntz (approved):
@aaemnnosttv Approving this already to not block it, but please make the few changes above :)

includes/Core/Authentication/Setup_V2.php
aaemnnosttv: Ah okay. I wasn't sure if it should include the step here or not as I was thinking that maybe it was omitted to restart the process in this case. I'll add it in 👍

includes/Core/Authentication/Setup_V2.php
aaemnnosttv: Oops! Done 👍

includes/Core/Authentication/Setup_V2.php
aaemnnosttv: Added `step` here too 👍

aaemnnosttv:
Final feedback points have been addressed; merging as per @felixarntz 's guidance back-channel 🚀

Issue #4466, opened 2021-12-06
Idea Hub notice script throws errors on the widgets page


eugene-manuilov:
## Bug Description

If you go to the widgets page and open the browser console, you will see a lot of errors thrown by the Idea Hub notice script that is intended to run only when editing a post. It happens because, in the modern versions of WordPress, the widgets page has been reworked to use Gutenberg for editing widgets as well. We need to update the `assets/js/googlesitekit-idea-hub-notice.js` script to make sure it does nothing if it's executed on the wrong page.

## Screenshots

![image](https://user-images.githubusercontent.com/873049/144853807-02b202ac-f1cc-497c-b76a-b96c73fdd934.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `googlesitekit-idea-hub-notice.js` script does nothing if the `core/editor` datastore is not loaded.

## Implementation Brief

* In `assets/js/googlesitekit-idea-hub-notice.js`:
  * Update the `listener` function to return early if the `wp.data.select( 'core/editor' )` call returns `null` (or `undefined`).

### Test Coverage

* N/A

## QA Brief

* Go to the widgets page and verify that you don't see errors thrown from the `googlesitekit-idea-hub-notice.js` script in the browser console.

## Changelog entry

* Fix browser console errors related to Idea Hub on the Appearance > Widgets page in the WordPress Admin.

tofumatt:
IB ✅

wpdarren:
### QA Update: ✅ 

**Verified:**

- The console areas no longer appear on the widgets page when Idea Hub is activated.  

![image](https://user-images.githubusercontent.com/73545194/145407973-963ea6e6-2dd6-4bce-ba42-d02b2836630c.png)

PR #4483, opened 2021-12-08
Set defaults when selecting a post ID from widgets page.

mitogh:
## Summary

When selecting the function `getCurrentPostId` from the`wp.data.select( 'core/editor' )` make sure that a default value is set to the store in case it returns `null` or `undefined` in this case an object is set to the default value, and a no operation value is set to the function `getCurrentPostId` if does not exists on the object.

Addresses issue #4466 

## Checklist

### DIFF
--- a/assets/js/googlesitekit-idea-hub-notice.js
+++ b/assets/js/googlesitekit-idea-hub-notice.js
@@ -47,8 +47,13 @@ const loadIdeaHubNotices = async ( _global = global ) => {
 	};
 
 	const listener = async () => {
+		// Added a no operation function as default value if the function does not exist on the store.
+		const noOperation = () => {};
+		const editorData = wp.data.select( 'core/editor' ) || {};
 		// eslint-disable-next-line sitekit/acronym-case
-		const postID = wp.data.select( 'core/editor' ).getCurrentPostId();
+		const { getCurrentPostId = noOperation } = editorData;
+		// eslint-disable-next-line sitekit/acronym-case
+		const postID = getCurrentPostId();
 
 		if ( ! postID ) {
 			return;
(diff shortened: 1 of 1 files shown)


assets/js/googlesitekit-idea-hub-notice.js
```diff
@@ -47,8 +47,13 @@ const loadIdeaHubNotices = async ( _global = global ) => {
...
 
 	const listener = async () => {
+		// Added a no operation function as default value if the function does not exist on the store.
+		const noOperation = () => {};
+		const editorData = wp.data.select( 'core/editor' ) || {};
 		// eslint-disable-next-line sitekit/acronym-case
-		const postID = wp.data.select( 'core/editor' ).getCurrentPostId();
+		const { getCurrentPostId = noOperation } = editorData;
+		// eslint-disable-next-line sitekit/acronym-case
+		const postID = getCurrentPostId();
```
tofumatt, line 56: This can be done a bit more straightforwardly, I think. The IB mentions checking to see if `wp.data.select( 'core/editor' ) === null`, which it will be when that store isn't loaded. We can return earlier here and not need to create/assign/run the no-op function, which I think is clearer logic to follow, eg:

```js
if ( wp.data.select( 'core/editor' ) === null ) {
	return;
}
```

tofumatt (changes requested):
I tested this and it works well, but let's simplify the return early logic here 🙂

assets/js/googlesitekit-idea-hub-notice.js
mitogh: I do agree, wanted to ask / suggest going with the optional chaining operator instead so we don't introduce a new conditional.

```js
const postID = wp.data.select( 'core/editor' )?.getCurrentPostId();
```

Then the next (existing) conditional would return as `postID` would be `undefined`. Let me know what do you think about this approach.

Thanks for the review.

tofumatt: Oh, yes, that's solid. We use optional chaining elsewhere so this makes total sense. Good catch! 👍🏻

tofumatt (approved):
So much nicer, thanks! 👍🏻

PR #4509, opened 2021-12-08
Remove unnecessary wrapper around OptimizeID field.

asvinb:
## Summary

Addresses issue #4028 

## Checklist

### DIFF
--- a/assets/sass/components/setup/_googlesitekit-setup-module.scss
+++ b/assets/sass/components/setup/_googlesitekit-setup-module.scss
@@ -109,7 +109,7 @@
 		> div {
 			margin: $grid-gap-phone $grid-gap-phone $grid-gap-phone 0;
 
-			@media (min-width: $bp-desktop) {
+			@media (min-width: $bp-wpAdminBarTablet) {
 				margin: $grid-gap-desktop $grid-gap-desktop $grid-gap-desktop 0;
 			}
 		}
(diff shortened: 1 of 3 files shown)


assets/sass/components/setup/_googlesitekit-setup-module.scss
```diff
@@ -126,9 +126,22 @@
...
+	}
+
+	div[name="optimizeID"] {
+		margin-top: 20px;
+		width: 100%;
+
+		@media (min-width: $bp-wpAdminBarTablet) {
+			min-width: 350px;
+			width: auto;
+		}
```
aaemnnosttv, line 138: Why make it full-width in smaller viewports? It looks odd particularly because the textarea is not as wide. 
image

assets/sass/components/setup/_googlesitekit-setup-module.scss
```diff
@@ -126,9 +126,22 @@
 		> .googlesitekit-setup-module__input {
 			margin: 1em 0;
 		}
+	}
+
+	div[name="optimizeID"] {
+		margin-top: 20px;
```
aaemnnosttv, line 132: We should avoid adding any Optimize-specific styles here. I don't think this module is particularly special compared to the others so we should be able to make it work in a way which is consistent with the others.

Looking at some of the other setup forms, there is some inconsistency there already so we shouldn't get off track and try to solve that here.

assets/sass/components/setup/_googlesitekit-setup-module.scss
```diff
@@ -126,9 +126,22 @@
...
+			min-width: 350px;
+			width: auto;
+		}
+
+		+ .mdc-text-field-helper-line {
 
-		+ .googlesitekit-error-text {
-			margin-top: -24px;
+			.mdc-text-field-helper-text {
+				font-size: 0.75rem;
```
aaemnnosttv, line 143: Looks like this is the correct font-size for helper-text but we should be applying it consistently throughout rather than only here; let's address that in another issue though as it's beyond the scope of this one.

aaemnnosttv (changes requested):
Thanks @asvinb. There are some good ideas here but I think these are mostly out of scope for this issue.

Given the AC, I think the simplest change we can make here would be to conditionally render the `helperText` and make it `persistent` as we do in other places to avoid the added visual gap when it isn't being shown. The error text already adds a layout shift anyways so keeping this text won't prevent that. Let's finish the spacing issue here and then we can look at how we can best enhance the layout of the setup forms more holistically 👍

Issue #4255, opened 2021-10-21
Analytics with zero data shows +100% on the Analytics page audience overview graph


cole10up:
## Bug Description

While testing https://github.com/google/site-kit-wp/issues/4226 I noticed that within the Analytics page, on the audience overview graph, if you hover over a date you'll see +100%

## Steps to reproduce

1. Setup a Jurassic Ninja site with the Site Kit plugin installed
2. Activate the Analytics module 
3. Wait a few days and then navigate to the Analytics page
4. Hover over a date in the Audience overview graph

Notice:
![image](https://user-images.githubusercontent.com/57115907/138333866-72dd63f4-9dd8-496a-b9e4-2c6e58e42f7b.png)

Expecting this functionality to match search console:
![image](https://user-images.githubusercontent.com/57115907/138334471-f42abaff-1533-4c8f-919a-6516ed5462f8.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The change percentage "calculation" between current and previous value should always work as follows:
    * If both values >0, like today.
    * If both values ==0, it should be "0%".
    * If current value ==0 and previous value >0, it should be "-100%".
    * If current value >0 and previous value ==0, it should be "100%".
* Specifically, this should also behave that way in chart tooltips with comparison percentages (see screenshot).

## Implementation Brief
* Using `assets/js/util/chart.js`,
	* Create new function `calculateDifference` which accepts 2 parameters, both integers.
		* `currentValue`
		* `previousValue`
	* The function should return the value as per the AC based on the `currentValue` and `previousValue`.
		* Please note that when both values are > 0, the logic should be the same as in the `extractAnalyticsDashboardData` from `assets/js/modules/analytics/util/index.js`
* Using
	* `assets/js/modules/adsense/util/site-stats-data.js`,
	* `assets/js/modules/analytics/util/index.js`,
	* `assets/js/modules/search-console/util/site-stats-data.js`,
		* update the `extractAnalyticsDashboardData` function to use the new function `calculateDifference` to get the value for the `difference` constant.
		* If `difference` is `0`, no SVG icon should be displayed.

### Test Coverage
* Add tests for the new `calculateDifference` function.

## QA Brief

* Follow step to reproduce in a new instawp site. 
* audience graph should no longer show +100% when zero data. 

## Changelog entry

* Fix the incorrect difference for chart values when the current and the previous values equal zero.

eugene-manuilov:
[...]

PR #4518, opened 2021-12-09
Fix zero data showing +100% percentage.

kuasha420:
## Summary

Addresses issue #4255

## Checklist

### DIFF
--- a/assets/js/util/chart.js
+++ b/assets/js/util/chart.js
@@ -75,15 +75,18 @@ export const isSingleSlice = ( report ) => {
 };
 
 /**
- * Calculates difference between two values.
+ * Calculates difference between two chart values.
  *
  * @since n.e.x.t
  *
- * @param {number} currentValue  The difference which can be positive or negative.
- * @param {number} previousValue Options to be passed to the ChangeArrow component.
+ * @param {number} currentValue  Current chart value.
+ * @param {number} previousValue Previous chart value.
  * @return {number} The difference.
  */
-export const calculateDifference = ( currentValue, previousValue ) => {
+export const calculateDifferenceBetweenChartValues = (
+	currentValue,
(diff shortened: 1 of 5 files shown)


assets/js/util/chart.js
```diff
@@ -73,3 +73,25 @@ export const isSingleSlice = ( report ) => {
...
 	return false;
 };
+
+/**
+ * Calculates difference between two values.
+ *
+ * @since n.e.x.t
+ *
+ * @param {number} currentValue  The difference which can be positive or negative.
+ * @param {number} previousValue Options to be passed to the ChangeArrow component.
```
eugene-manuilov, line 83: @kuasha420 these descriptions look irrelevant to what these parameters are. Can you adjust it?

assets/js/util/chart.js
```diff
@@ -73,3 +73,25 @@ export const isSingleSlice = ( report ) => {
...
+ * @return {number} The difference.
+ */
+export const calculateDifference = ( currentValue, previousValue ) => {
```
eugene-manuilov, line 86: Let's probably rename this function to be more specific. Something like `calculateDifferenceBetweenChartValues` will better explain what the function does. What do you think?

eugene-manuilov (changes requested):
Thanks, @kuasha420. Almost looks good to me. Added two small nitpicks. Once they are fixed, I'll approve this PR.

assets/js/util/chart.js
kuasha420: Yes, it makes more sense and will communicate the fact that it's specifically for finding difference in chart value and not a general solution. Updated.

eugene-manuilov (approved):
Awesome! Thanks, @kuasha420.

Issue #4380, opened 2021-11-19
Improve behaviour when PageSpeed 'visit the dashboard` link is clicked in Site Kit settings


wpdarren:
## Bug Description

When you click on the `visit the dashboard` link on PageSpeed Insights settings (`wp-admin/admin.php?page=googlesitekit-settings#/connected-services/pagespeed-insights`) you are redirected to the top of the Site Kit Dashboard. It would be more user-friendly to jump down to the PSI widget.  I realise we're working on the Unified dashboard but feel this still applies.

![image](https://user-images.githubusercontent.com/73545194/142656048-4d60b642-e221-414d-91c2-66581b0ae8ac.png)

## Steps to reproduce

1. Go to 'Site Kit settings'
2. Click on 'PageSpeed Insights.'
3. Observe the `visit the dashboard` link
4. See behaviour.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `visit the dashboard` link shown in the PageSpeed Insights settings view should be updated to link to the PageSpeed section of the dashboard
    * For the Unified Dashboard, this should link to the respective `#speed` anchor
    * For the current/non-unified dashboard, this should link to the same `#googlesitekit-pagespeed-header` used by the jump link today

## Implementation Brief
* Using `assets/js/modules/pagespeed-insights/components/settings/SettingsView.js`,
	* Use the `useFeature` hook and check whether the `unifiedDashboard` feature flag is enabled.
	* If the above feature flag is enabled,
		* append `#speed`
		* else append `#googlesitekit-pagespeed-header` to `dashboardPermalink`.
* Using `assets/js/modules/pagespeed-insights/components/dashboard/DashboardPageSpeed.js`,
	* Use the `useFeature` hook and check whether the `unifiedDashboard` feature flag is enabled.
	* Using the `useMount` hook, if the `unifiedDashboard` feature flag is not enabled, check for the `googlesitekit-pagespeed-header` hash.
		* If the hash is present, then scroll to the widget taking the height of the header into account.

### Test Coverage
* No new tests to be added.

## QA Brief

**Testing on Unified Dashboard**
- Enable unified dashboard via the tester plugin
- Go to `Site Kit settings` and click on `PageSpeed Insights`
- Click `visit the dashboard` link
- Verify that you are navigated to the dashboard `Speed` section
- Verify `#speed` is appended to the URL

**Testing when the Unified Dashboard is disabled**
[...]

PR #4520, opened 2021-12-09
Fix scroll to page speed section when the visit the dashboard link is clicked

hussain-t:
## Summary

Addresses issue #4380 

## Checklist


assets/js/modules/pagespeed-insights/components/dashboard/DashboardPageSpeed.js
```diff
@@ -62,9 +62,44 @@ import {
...
+	/**
+	 * This is to show the PSI header title.
+	 * Otherwise it scrolls to the PSI header (tabs).
+	 */
+	const marginBottom = 75;
+
+	const headerOffset = headerHeight + marginBottom;
+
+	return contextTop + global.scrollY - headerOffset;
+};
```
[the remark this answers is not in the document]
hussain-t: Thanks for the review @eugene-manuilov. I was thinking to make it as a utility function. However, the behaviour slightly changes i.e. the `getContextScrollTop` function in the `DashboardNavigation` component takes an account of the unified dashboard navigation bar and calculates the result. However, I can calculate it conditionally by checking `.googlesitekit-navigation` selector is available or not.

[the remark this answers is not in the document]
hussain-t: > extract that function into its own file in the utils folder

Can I create the function in this directory `assets/js/util`?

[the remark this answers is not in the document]
eugene-manuilov: > However, the behaviour slightly changes i.e. the getContextScrollTop function in the DashboardNavigation component takes an account of the unified dashboard navigation bar and calculates the result.

Yes, you will need to update that function a little bit to be more universal. Either approach works for me, probably the first one is more preferable, but I will let you select the one you think works the best 😄 

> Can I create the function in this directory assets/js/util?

Yep.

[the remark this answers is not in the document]
hussain-t: Awesome, thanks. Let me choose the first approach since it's more readable for either cases 👍

assets/js/util/get-context-scroll-top.js
```diff
@@ -0,0 +1,40 @@
+import { ANCHOR_ID_TRAFFIC } from '../googlesitekit/constants';
```
eugene-manuilov, line 1: We need to add the file header comment as we do it in other files.

assets/js/util/get-context-scroll-top.js
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @since n.e.x.t
+ *
+ * @param {string} contextID  The ID of the context element to scroll to.
+ * @param {string} breakpoint The current breakpoint.
+ * @return {number} The offset to scroll to.
+ */
+const getContextScrollTop = ( contextID, breakpoint ) => {
+	const contextElement = document.getElementById( contextID );
+	if ( contextID === ANCHOR_ID_TRAFFIC || ! contextElement ) {
```
eugene-manuilov, line 14: I think we can move this contextID check back to the `DashboardNavigation` component since it's only makes sense there.

```diff
- if ( contextID === ANCHOR_ID_TRAFFIC || ! contextElement ) {
+ if ( ! contextElement ) {
```

Then in the `assets/js/components/DashboardNavigation.js` file:

```diff
 global.scrollTo( {
-	top: getContextScrollTop( hash, breakpoint ),
+	top: hash !== ANCHOR_ID_TRAFFIC ? getContextScrollTop( hash, breakpoint ) : 0,
 	behavior: 'smooth',
 } );
```

eugene-manuilov (changes requested):
Thanks, @hussain-t. This is almost good to go. Two more comments and we should be good. Please, take a look.

assets/js/util/get-context-scroll-top.js
hussain-t: Perfect!

assets/js/components/DashboardNavigation.js
```diff
@@ -160,10 +132,13 @@ export default function DashboardNavigation() {
 		if ( global.location.hash !== '' ) {
 			setTimeout( () => {
 				global.scrollTo( {
-					top: getContextScrollTop(
-						global.location.hash.substr( 1 ),
-						breakpoint
-					),
+					top:
+						global.location.hash !== ANCHOR_ID_TRAFFIC
```
eugene-manuilov, line 136: The location hash contains `#` at the beginning, so it will never be equal `ANCHOR_ID_TRAFFIC`.
```suggestion
						global.location.hash.substr( 1 ) !== ANCHOR_ID_TRAFFIC
```

eugene-manuilov: Probably worth assigning the hash to a variable and then use it in the condition and in the function call:

```js
const hash = global.location.hash.substr( 1 );
global.scrollTo( {
    top: hash !== ANCHOR_ID_TRAFFIC ? getContextScrollTop( hash, breakpoint ) : 0
    ...
} );

eugene-manuilov (changes requested):

assets/js/components/DashboardNavigation.js
hussain-t: Apologies I missed it. Good catch 👍

eugene-manuilov (approved):
LGTM! Thanks, @hussain-t.

Issue #4428, opened 2021-11-29
Implement new entity dashboard header


felixarntz:
Per https://www.figma.com/file/VILRieNSLg2DOVyEgFBkXe/%5BMVP%5D-Unified-Dashboard?node-id=1330%3A9080, a few late updates have been made to give the URL search and entity dashboard header for the `unifiedDashboard` a better user experience. This is the third out of three issues which replaces #4147. The others are:
* #4426
* #4427

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Originally, an `EntityHeaderBanner` component was implemented as part of the `unifiedDashboard` feature to indicate which current entity the user is looking at on the entity dashboard. This design has now been discarded in favor of a new design that takes less space and is displayed in a more appropriate location.
* The `EntityHeaderBanner` component and its usage should be entirely removed.
* Instead, a new component (e.g. just called `EntityHeader`, since it's technically no longer a banner) should be implemented that follows [the entity header UI right below the pills in this Figma design](https://www.figma.com/file/VILRieNSLg2DOVyEgFBkXe/%5BMVP%5D-Unified-Dashboard?node-id=1330%3A9083) (see [this Figma design for mobile](https://www.figma.com/file/VILRieNSLg2DOVyEgFBkXe/%5BMVP%5D-Unified-Dashboard?node-id=1330%3A12574)):
    * There should be a "Back to dashboard" link on the left that points back to the main dashboard.
    * The current entity title and URL should be displayed on the right.
    * If there is no current entity, it shouldn't be displayed at all.
    * The component should have an explicit gray background and drop shadow when scrolled past so that it overlaps the other content when scrolled down (see [here in the Figma file](https://www.figma.com/file/VILRieNSLg2DOVyEgFBkXe/%5BMVP%5D-Unified-Dashboard?node-id=1330%3A9909)). In other words, the component should also be displayed fixed (on both desktop and mobile), and when scrolled past should overlap the content, hence it needs the same gray as in the overall background color.
* The new component should be integrated into the entity dashboard like in the Figma file (right below the pills navigation).
* Extra requirement: The `BannerNotifications` must never be displayed within the entity dashboard, but it is implemented like that now. This was likely a misunderstanding in a previous issue, `BannerNotifications` should only be present on the main dashboard (the same way it is today).

## Implementation Brief
* Create `assets/js/components/EntityHeader.js` which exports the `EntityHeader` functional component.
[...]

PR #4528, opened 2021-12-09
Enhancement/4428 New Entity Dashboard Header

jimmymadon:
## Summary

Addresses issue #4428

## Relevant technical choices

* Preferred to re-add the background SVG to the `.googlesitekit-entity-header__details` class instead of using a CSS filter to change the colour of `googlesitekit-cta-link--external`.

## Checklist


assets/js/components/EntityHeader.js
```diff
@@ -37,9 +38,21 @@ import { CORE_SITE } from '../googlesitekit/datastore/site/constants';
...
+const EntityHeader = () => {
+	const [ hasScrolled, setHasScrolled ] = useState( false );
+	useEffect( () => {
+		global.window.onscroll = function () {
+			if ( global.window.scrollY > 5 ) {
+				setHasScrolled( true );
+			} else {
+				setHasScrolled( false );
+			}
+		};
```
eugene-manuilov, line 53: We don't use the old approach of defining event listeners. If you need to add one, use the `addEventListener` method accompanied with the corresponding `removeEventListener` call in the clean-up function to avoid memory leaks and unexpected console errors.

Also, `global` already means `window` when run in the browser. So using `global.window....` doesn't make sense. Just use `global.addEventListener( ... )`.

assets/js/components/Header.js
```diff
@@ -84,14 +84,13 @@ const Header = ( { children, subHeader, showNavigation } ) => {
...
-				<div className="googlesitekit-subheader">
-					<EntityHeaderBanner />
-					{ subHeader }
-				</div>
+				<div className="googlesitekit-subheader">{ subHeader }</div>
 			) }
 
 			{ showNavigation && <DashboardNavigation /> }
 
+			{ <EntityHeader /> }
```
eugene-manuilov, line 92: This should be added only when `unifiedDashboardEnabled` is truthy.

assets/sass/components/global/_googlesitekit-entity-header.scss
```diff
@@ -0,0 +1,122 @@
...
+		height: auto;
+		letter-spacing: 0.3px;
+		line-height: 1;
+		margin: 4px 0 0;
+		min-width: auto;
+		text-transform: none;
+
+		@media (min-width: $bp-tablet) {
+			font-size: 16px;
+			margin: 4px 0 12px;
```
eugene-manuilov, line 69: I think the bottom margin should be equal `4px` too because currently the back button looks vertically aligned incorrectly.

image

eugene-manuilov (changes requested):
Thanks, @jimmymadon. This is very good start. Added some comments for you. Please, take a look.

assets/js/components/EntityHeader.js
```diff
@@ -37,9 +38,21 @@ import { CORE_SITE } from '../googlesitekit/datastore/site/constants';
 import BackspaceIcon from '../../svg/keyboard-backspace.svg';
 import { CORE_LOCATION } from '../googlesitekit/datastore/location/constants';
 import Link from './Link';
+import classnames from 'classnames';
```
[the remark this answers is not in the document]
jimmymadon: Oops!

assets/js/components/EntityHeader.js
jimmymadon: Thanks for the tips - will keep these in mind!

assets/js/components/EntityHeader.js
```diff
@@ -37,9 +38,21 @@ import { CORE_SITE } from '../googlesitekit/datastore/site/constants';
...
+	const [ hasScrolled, setHasScrolled ] = useState( false );
+	useEffect( () => {
+		global.window.onscroll = function () {
+			if ( global.window.scrollY > 5 ) {
+				setHasScrolled( true );
+			} else {
+				setHasScrolled( false );
+			}
+		};
+	}, [] );
```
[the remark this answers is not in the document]
jimmymadon: Sweet - there's a hook for everything! :)

assets/sass/components/global/_googlesitekit-entity-header.scss
```diff
@@ -0,0 +1,122 @@
...
+			text-overflow: ellipsis;
+
+			@media (min-width: $bp-tablet) {
+				font-size: 14px;
+				line-height: 20px;
+				margin: 0;
+			}
+		}
+
+		a {
```
eugene-manuilov, line 109: Can we add text overflow for the link as well to make it look better on mobile devices? Currently if the link is long, I don't see an icon at the end and the last part of the link.

image

I think ideally it would be better if we "collapse" the central part of the URL to be something like `http://mysite.com/.../page-slug.html` instead of `http://mysite.com/categoryA/...`. 

cc @felixarntz

eugene-manuilov: Also the link looks too small and too thin on bigger screens. Maybe we should make it a bit bigger and to use the same styles as we use for source links?

image

eugene-manuilov (changes requested):
Thanks, @jimmymadon. This PR almost looks good. Left some comments regarding UI changes. Please, take a look.

assets/sass/components/global/_googlesitekit-entity-header.scss
felixarntz: @eugene-manuilov @jimmymadon Agreed, can we add an ellipsis to the URL instead of actually cutting it off because of overflow issues? I also agree that it'd be a better experience putting that ellipsis somewhere in between, to try to keep the last segment visible. Let's do that, unless that would require a significantly more complex approach.

cc @marrrmarrr

assets/sass/components/global/_googlesitekit-entity-header.scss
```diff
@@ -0,0 +1,122 @@
...
+				top: 180px;
+			}
+
+			@media ( min-width: $bp-desktop ) {
+				top: 196px;
+			}
+		}
+
+		@media (min-width: $bp-tablet) {
+			padding: 16px 34px;
```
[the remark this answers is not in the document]
jimmymadon: Thanks - good to see we use these pre-defined padding / transition vars. Will use them in the future.

assets/sass/components/global/_googlesitekit-entity-header.scss
jimmymadon: @eugene-manuilov @felixarntz I've added the ellipsis in between the `entityURL` string by splitting the strings in JSX and then using CSS to add the ellipsis. The current solution always adds the ellipsis 10 characters before the end of the URL. This way, most of the initial string is visible (which usually contains the domain name, date folders, categories etc) as well as the last 10 chars. 
Screenshot 2021-12-14 at 14 09 30

This happens on all viewports for very long URLs - including mobile viewports. Since CSS takes care of the overflow now, do we still need a maximum character limit?

On a side note, I do agree with @eugene-manuilov that the font-size of the URL is tiny on hi-res (qHD+) large screens. So far I've just followed the Figma designs but let me know if I should increase this.

assets/js/components/EntityHeader.js
```diff
@@ -66,44 +70,43 @@ const EntityHeaderBanner = () => {
...
+					{ /* Split the entityURL into two strings which are separated by an ellipsis (...) using CSS */ }
+					<span className="start">
+						{ entityURL.substr( 0, entityURL.length - 10 ) }
+					</span>
+					<span className="end">
+						{ entityURL.substr(
+							entityURL.length - 10,
+							entityURL.length
+						) }
+					</span>
```
eugene-manuilov, line 105: Hm... This is not going to work because we need to display ellipsis after the origin so the user can see the last part of the URL. We need to programmatically calculate the number of characters that we can display there and manually insert `...` into the URL.

Here is PoC of how it can be done:

```js
const linkRef = useRef();
const [ url, setURL ] = useState( entityURL );

useEffect( () => {
	const calculate = () => {
		const width = linkRef.current.clientWidth;
		const fontSize = global
			.getComputedStyle( linkRef.current, null )
			.getPropertyValue( 'font-size' );
		const size = parseFloat( fontSize );

		const paddingRight = global
			.getComputedStyle( linkRef.current, null )
			.getPropertyValue( 'padding-right' );
		const padding = parseFloat( paddingRight );

		// 2.28 is font's character constant: https://pearsonified.com/characters-per-line/
		const maxChars = ( ( width - padding ) * 2.28 ) / size; 

		const url2 = new URL( entityURL );

		if ( maxChars < url2.pathname.length ) {
			const newlen = maxChars - url2.origin.length;
			url2.pathname =
				'/...' +
				url2.pathname.substr( url2.pathname.length - newlen );
		}

		setURL( url2.toString() );
	};

	calculate();

	global.addEventListener( 'resize', calculate );
	return () => {
		global.removeEventListener( 'resize', calculate );
	};
}, [ entityURL, linkRef, setURL ] );

...

return (
	<div
		className={ classnames( 'googlesitekit-entity-header', {
			'googlesitekit-entity-header--has-scrolled': y > 1,
		} ) }
	>
		...

		<div className="googlesitekit-entity-header__details">
			<p>{ currentEntityTitle }</p>
			<Link ref={ linkRef } href={ entityURL } external inherit>
				{ url }
			</Link>
		</div>
	</div>
);
```

You will also need to remove flex styles from the link and add `display: block` property. Use this as a starting point. We still need to figure out how to make this work the best to have a good user experience.

cc @asvinb in case you can help with this 😄

eugene-manuilov (changes requested):

assets/sass/components/global/_googlesitekit-entity-header.scss
jimmymadon: @felixarntz Eugene rightly pointed out that the ending part of the URL is what is more useful for users to view as the initial part of the URLs in a site are all going to be similar. So do we have any strong preferences as to where the ellipsis should be inserted? Should it be inserted right after the origin/hostname, clipping the initial sections of the pathname?

felixarntz: @jimmymadon I think it would be best to try to specifically keep the last path segment intact.

We could either put the ellipsis right before that, or maybe right after the initial slash after the domain.

assets/js/components/EntityHeader.js
jimmymadon: @eugene-manuilov I realised that the `clientWidth`, `offsetWidth` and `getBoundingClientRect.width()` were all not returning the correct width of the visible rendered <a> element. This is because React loads the <a> element with the full `entityURL` first giving it a longer width than what it has space for. Fetching the widths of the parent give the same result as the `googlesitekit-entity-header__details` is within the `googlesitekit-entity-header` Flex-box container. So I would potentially have to calculate the width of the entire `googlesitekit-entity-header` and then subtract the “Back to dashboard” link, paddings, etc. which could get quite complicated for what we want to achieve. Any thoughts? 

@tofumatt had a look too with me at trying to calculate the available width - but he suggested to tweak the CSS only solution in my previous commit to truncate the protocol/www etc. I could keep the domain name and the last segment fixed and insert the ellipsis just before the start of the last path segment? But there are too many possibilities with URLs and it will be hard to cater for all types (different ratios of origin : pathname : queryStrings : hash, etc).

eugene-manuilov: @jimmymadon in this case let's define the maximum width for that `<a>` tag for all breakpoints and use those widths to do calculations.

eugene-manuilov: Try to select the best suiting width for each breakpoint and hardcode it in the component + add it to the scss file.

jimmymadon: @eugene-manuilov I managed to tweak your POC to calculate the width of the `entity-header` and derive the width of the url. I've also ensured that URLs with short paths but long query params, etc are covered.

assets/js/components/EntityHeader.js
```diff
@@ -0,0 +1,164 @@
...
+			const urlWidth = headerDetailsWidth - 20 - 16; // padding: entity-header__details = 20px, url <a> = 16px
+
+			const urlFontSize = global
+				.getComputedStyle( headerRef.current.lastChild.lastChild, null ) // url <a> element
+				.getPropertyValue( 'font-size' );
+			const size = parseFloat( urlFontSize );
+
+			// 2 is appox. the min character constant for sans-serif fonts:
+			// https://pearsonified.com/characters-per-line/
+			const maxChars = ( urlWidth * 2 ) / size;
```
eugene-manuilov, line 90: @jimmymadon why do we need all these calculations? Why not use the `clientWidth` property of the link to get the width of the element?

eugene-manuilov (changes requested):
Thanks, @jimmymadon. This looks definitely better. Left a few comments for you. Please, take a look.

assets/js/components/EntityHeader.js
jimmymadon: @eugene-manuilov That's what we started with but as I mentioned [here](https://github.com/google/site-kit-wp/pull/4528#discussion_r769687358), the clientWidth, offsetWidth and getBoundingClientRect.width() were all not returning the correct width of the visible rendered <a> element. This is because React loads the element with the full entityURL first pretending to give it a longer width than what it has space for.

eugene-manuilov: Ah... ok, thanks for the clarification. Then, I think, if we add `max-width: 100%` property to the link styles, it will return the correct width even if the initial link is longer, right?

assets/sass/components/global/_googlesitekit-entity-header.scss
```diff
@@ -0,0 +1,122 @@
...
+				margin: 0;
+			}
+		}
+
+		a {
+			// override default .googlesitekit-cta-link--external SVG's color
+			background-image: url("data:image/svg+xml,%3Csvg width='12' height='12' viewBox='0 0 12 12' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M10.6667 10.6667H1.33333V1.33333H6V0H1.33333C0.979711 0 0.640573 0.140476 0.390524 0.390524C0.140476 0.640573 0 0.979711 0 1.33333V10.6667C0 11.0203 0.140476 11.3594 0.390524 11.6095C0.640573 11.8595 0.979711 12 1.33333 12H10.6667C11.4 12 12 11.4 12 10.6667V6H10.6667V10.6667ZM7.33333 0V1.33333H9.72667L3.17333 7.88667L4.11333 8.82667L10.6667 2.27333V4.66667H12V0H7.33333Z' fill='%232998E8'/%3E%3C/svg%3E");
+			background-size: 12px;
+			color: $c-picton-blue;
+			display: block;
```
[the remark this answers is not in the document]
jimmymadon: Thanks - didn't notice that! Tested with quite a few different URLs.

assets/js/components/EntityHeader.js
```diff
@@ -0,0 +1,158 @@
...
+				const origin = shortenedURL.origin;
+				const restOfURL = shortenedURL.toString().replace( origin, '' );
+				const newRestOfURL = '/...' + restOfURL.substr( extraChars );
+				setURL( origin + newRestOfURL );
+			} else {
+				setURL( entityURL );
+			}
+		};
+
+		// Use throttled version only on window resize
```
eugene-manuilov, line 94: Single line comments should start with a capital letter and end with a full stop:

```suggestion
		// Use throttled version only on window resize.
```

assets/js/components/EntityHeader.js
```diff
@@ -0,0 +1,158 @@
...
+	const headerDetailsRef = useRef();
+	const [ url, setURL ] = useState( entityURL );
+
+	useEffect( () => {
+		const shortenURL = () => {
+			if ( ! headerDetailsRef.current ) {
+				return;
+			}
+
+			// Remove 40 px for margins + SVG at the end of the URL link
```
eugene-manuilov, line 69:
```suggestion
			// Remove 40 px for margins + SVG at the end of the URL link.
```

assets/js/components/EntityHeader.js
```diff
@@ -0,0 +1,158 @@
...
+			const fontSize = parseFloat( urlFontSize );
+
+			// 2 is appox. the minimum character constant for sans-serif fonts:
+			// https://pearsonified.com/characters-per-line/
+			const maxChars = ( availableWidth * 2 ) / fontSize;
+
+			const shortenedURL = new URL( entityURL );
+
+			if ( maxChars < entityURL.length ) {
+				const extraChars = entityURL.length - maxChars + 4; // length of "/..." = 4
```
eugene-manuilov, line 84:
```suggestion
				const extraChars = entityURL.length - maxChars + 4; // 4 is length of "/...".
```

assets/js/components/EntityHeader.js
```diff
@@ -0,0 +1,158 @@
...
+			const urlFontSize = global
+				.getComputedStyle( headerDetailsRef.current.lastChild, null )
+				.getPropertyValue( 'font-size' );
+			const fontSize = parseFloat( urlFontSize );
+
+			// 2 is appox. the minimum character constant for sans-serif fonts:
+			// https://pearsonified.com/characters-per-line/
+			const maxChars = ( availableWidth * 2 ) / fontSize;
+
+			const shortenedURL = new URL( entityURL );
```
eugene-manuilov, line 81: Let's define this variable just before using it:

```js
if ( maxChars < entityURL.length ) {
    const shortenedURL = new URL( entityURL );
    ...
} else {
    setURL( entityURL );
}
```

eugene-manuilov (changes requested):
Thanks, @jimmymadon. This is almost good to go. Added a few more tiny nitpicks. Please, take a look. Once they are fixed, I'll approve this PR.

jimmymadon:
@eugene-manuilov Thanks for your inputs and help on this!

eugene-manuilov (approved):
Thanks, @jimmymadon. This looks good to me.

felixarntz (approved):
Excellent work, lgtm!

The font sizing for the title and URL look a bit small to me on desktop, but I realize that's how it is in the Figma design as well. I guess it still sticks out enough especially given that you can only land on the entity dashboard intentionally, so good to go from my point of view.

We should probably do a final UI review with @marrrmarrr and @samitron7 before launch.

PR #4541, opened 2021-12-10
Add auto hide to entity search input on unfocus.

kuasha420:
## Summary

Addresses issue #4427

## Checklist

### DIFF
--- a/assets/js/components/PostSearcherAutoSuggest.js
+++ b/assets/js/components/PostSearcherAutoSuggest.js
@@ -31,7 +31,7 @@ import {
 /**
  * WordPress dependencies
  */
-import { useState, useEffect, useCallback } from '@wordpress/element';
+import { useState, useEffect, useCallback, useRef } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import { ENTER, ESCAPE } from '@wordpress/keycodes';
 
@@ -69,6 +69,8 @@ export default function PostSearcherAutoSuggest( {
 		select( CORE_SITE ).getCurrentEntityTitle()
 	);
 
+	const postTitle = useRef( null );
+
 	const onFocus = useCallback( () => {
 		setIsActive( true );
 	}, [ setIsActive ] );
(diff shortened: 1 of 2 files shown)


aaemnnosttv (changes requested):
Thanks @kuasha420 – this looks good overall. One quirk that I noticed is that when selecting a choice the input is also cleared, which I don't think should happen according to the AC as this is by definition a "valid search result" 😄 

https://user-images.githubusercontent.com/1621608/145789782-20c17ba9-3b15-4015-aa73-fe3188f2eb94.mp4

kuasha420:
@aaemnnosttv IB didn't cover it so I had to improvise a solution. Let me know if it's correct!

assets/js/components/PostSearcherAutoSuggest.js
```diff
@@ -67,6 +69,19 @@ export default function PostSearcherAutoSuggest( {
...
+
+	const onFocus = useCallback( () => {
+		setIsActive( true );
+	}, [ setIsActive ] );
+
+	const onBlur = useCallback( () => {
+		setTimeout( () => {
+			setIsActive( false );
+			setSearchTerm( postTitle.current ?? currentEntityTitle ?? '' );
+		}, 100 );
```
aaemnnosttv, line 82: Why the delay here? 🤔

kuasha420: When clicked on a result, the onBlur is fired before onSelect, setTimeout is there to make sure logic for onSelect runs first. That's the other reason why I had to use the useRef hook to get the updated post title from onSelectCallback.

aaemnnosttv: Could this maybe be done using an effect that runs when `isActive` changes rather than using a timeout?

kuasha420: Hi, 

Tried to use `useEffect` and only set `isActive` on the `onBlur` handler. As the visibility of dropdown is dependent on isActive, the dropdown gets out of the way when a result is clicked, missing the onSelect handler completely. 

I have added a comment explaining the timeout. 

Cheers.

Issue #3993, opened 2021-09-06
Ability to programmatically disable modules


henrywright:
## Feature Description

In a network install, each site admin doesn't have many technical capabilities. They are unable to perform many of the actions necessary to optimise a website for speed. For example, PageSpeed Insights includes some very useful suggestions but the suggestions aren't actionable by the site admin. They are, however, actionable by the super admin or a person with developer access.

It would be neat if the super admin could enable and disable the various modules at the network level based on whether they are useful for the site admins in the network.

By modules I'm referring to:

- Google AdSense
- Google Analytics
- Google Search Console
- PageSpeed Insights
- Tag Manager
- Optimize

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* When registering the Site Kit modules internally in the Site Kit registry, the list of module class names should be filtered via a new filter `googlesitekit_available_modules`. **The filter should only allow removing items, not adding any that aren't in the unfiltered list.**

## Implementation Brief
* Using `includes/Core/Modules/Modules.php`:
  * Update the `$core_modules` array to be an associative array where keys are module slugs and values are class names.
  * Update the `setup_registry` method:
    * to apply the `googlesitekit_available_modules` filter to keys of the `$this->core_modules` array.
    * then in the `foreach` loop register modules if their slug is in the list of the available modules returned from the filter or the module is force active.

### Test Coverage
* Update `tests/phpunit/integration/Core/Modules/ModulesTest.php` to include tests for the `googlesitekit_available_modules` filter.

## QA Brief

List with the slug of all the modules.

- `site-verification`
- `search-console`
- `analytics`
- `optimize`
- `tag-manager`
- `adsense`
- `pagespeed-insights`
- `analytics-4`

Modules added via featured flags. 

- `idea-hub`
- `subcribe-with-google`

Add the following snippets of code into your `functions.php` of your active theme or in `mu-plugins/overrides.php` and place the code there.

**Disable All Modules**

```
<?php
add_filter(
	'googlesitekit_available_modules',
	function( $modules ) {
		return array();
	}
);
```

**Enable Only Search Console*

```
<?php
add_filter(
	'googlesitekit_available_modules',
	function( $modules ) {
		return array( 'search-console' );
	}
);
```

## Changelog entry

* Add the ability to programmatically disable modules using the `googlesitekit_available_modules` filter.

jamesozzie:
Hi @henrywright,

[...]

PR #4545, opened 2021-12-11
Add filter to toggle modules programmatically.

mitogh:
## Summary

A new filter was added `googlesitekit_available_modules` the filter allow customizing programmatically which modules are present on the site.

Each module now has a unique key in order to identify each module during the filter just by passing a slug of the module.

Addresses issue #3993

## Checklist

### DIFF
--- a/includes/Core/Modules/Module.php
+++ b/includes/Core/Modules/Module.php
@@ -794,11 +794,11 @@ abstract class Module {
 	}
 
 	/**
-	 * Default behavior for all the modules, is not to be forced to be active.
+	 * Determines whether the current module is forced to be active or not.
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @return bool Modules are not forced to be active by default. `false` is returned by default.
+	 * @return bool TRUE if the module forced to be active, otherwise FALSE.
 	 */
 	public static function is_force_active() {
 		return false;
--- a/includes/Core/Modules/Module.php
+++ b/includes/Core/Modules/Module.php
@@ -687,7 +687,7 @@ abstract class Module {
 				'homepage'     => '',
 				'feature'      => '',
 				'depends_on'   => array(),
-				'force_active' => false,
+				'force_active' => static::is_force_active(),
 				'internal'     => false,
 			)
 		);
@@ -793,4 +793,14 @@ abstract class Module {
 		return $items;
 	}
 
+	/**
+	 * Default behavior for all the modules, is not to be forced to be active.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return bool Modules are not forced to be active by default. `false` is returned by default.
+	 */
+	public static function is_force_active() {
+		return false;
+	}
 }
--- a/includes/Core/Modules/Modules.php
+++ b/includes/Core/Modules/Modules.php
@@ -576,13 +576,13 @@ final class Modules {
 		 *
 		 * @since n.e.x.t
 		 *
-		 * @retun array An array of filtered module slugs.
+		 * @return array An array of filtered module slugs.
 		 */
 		$available_modules = (array) apply_filters( 'googlesitekit_available_modules', array_keys( $this->core_modules ) );
 		$modules           = array_fill_keys( $available_modules, true );
 
 		foreach ( $this->core_modules as $slug => $module ) {
-			if ( isset( $modules[ $slug ] ) ) {
+			if ( isset( $modules[ $slug ] ) || call_user_func( array( $module, 'is_force_active' ) ) ) {
 				$registry->register( $module );
 			}
 		}
--- a/includes/Core/Modules/Modules.php
+++ b/includes/Core/Modules/Modules.php
@@ -566,7 +566,18 @@ final class Modules {
 	 * @return Module_Registry
 	 */
 	protected function setup_registry() {
-		$registry          = new Module_Registry();
+		$registry = new Module_Registry();
+		/**
+		 * Filters an array with the list of all the available modules slugs, each slug present on this array would
+		 * be registered for inclusion. If a module is `force_activate` it would be included even if the module is
+		 * removed from this filter.
+		 *
+		 * @param array $available_modules An array with list of module's slug available on this installation.
+		 *
+		 * @since n.e.x.t
+		 *
+		 * @retun array<string> An array with the list of available slugs.
+		 */
 		$available_modules = (array) apply_filters( 'googlesitekit_available_modules', array_keys( $this->core_modules ) );
 
 		foreach ( $this->core_modules as $slug => $module ) {
--- a/includes/Core/Modules/Modules.php
+++ b/includes/Core/Modules/Modules.php
@@ -126,13 +126,13 @@ final class Modules {
 	 * @var string[] Core module class names.
 	 */
 	private $core_modules = array(
-		'site-verification'  => Site_Verification::class,
-		'search-console'     => Search_Console::class,
-		'analytics'          => Analytics::class,
-		'optimize'           => Optimize::class,
-		'tagmanager'         => Tag_Manager::class,
-		'adsense'            => AdSense::class,
-		'pagespeed-insights' => PageSpeed_Insights::class,
+		'site-verification'             => Site_Verification::class,
+		'search-console'                => Search_Console::class,
+		Analytics::MODULE_SLUG          => Analytics::class,
+		Optimize::MODULE_SLUG           => Optimize::class,
+		Tag_Manager::MODULE_SLUG        => Tag_Manager::class,
+		AdSense::MODULE_SLUG            => AdSense::class,
+		PageSpeed_Insights::MODULE_SLUG => PageSpeed_Insights::class,
 	);
 
 	/**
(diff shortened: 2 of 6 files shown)


eugene-manuilov (changes requested):
Thanks, @mitogh. This is a very good start. Added two comments for you. Please, take a look.

includes/Core/Modules/Modules.php
```diff
@@ -566,10 +566,13 @@ protected function get_registry() {
 	 * @return Module_Registry
 	 */
 	protected function setup_registry() {
-		$registry = new Module_Registry();
+		$registry          = new Module_Registry();
+		$available_modules = (array) apply_filters( 'googlesitekit_available_modules', array_keys( $this->core_modules ) );
```
felixarntz, line 570: We need to add a documentation block here (all actions/filters should have one). As mentioned in parts in https://github.com/google/site-kit-wp/issues/3993#issuecomment-992629451, let's also include in that documentation that force-active modules cannot be modified by this (see also comment below).

includes/Core/Modules/Modules.php
```diff
@@ -126,13 +126,13 @@ final class Modules {
...
-		Tag_Manager::class,
-		AdSense::class,
-		PageSpeed_Insights::class,
+		'site-verification'  => Site_Verification::class,
+		'search-console'     => Search_Console::class,
+		'analytics'          => Analytics::class,
+		'optimize'           => Optimize::class,
+		'tagmanager'         => Tag_Manager::class,
+		'adsense'            => AdSense::class,
+		'pagespeed-insights' => PageSpeed_Insights::class,
```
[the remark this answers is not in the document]
mitogh: Makes sense, that's a good flag, thanks.

includes/Core/Modules/Modules.php
```diff
@@ -566,10 +566,13 @@ protected function get_registry() {
...
 	 */
 	protected function setup_registry() {
-		$registry = new Module_Registry();
+		$registry          = new Module_Registry();
+		$available_modules = (array) apply_filters( 'googlesitekit_available_modules', array_keys( $this->core_modules ) );
 
-		foreach ( $this->core_modules as $core_module ) {
-			$registry->register( $core_module );
+		foreach ( $this->core_modules as $slug => $module ) {
+			if ( in_array( $slug, $available_modules, true ) ) {
```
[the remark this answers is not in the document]
mitogh: Hey @eugene-manuilov when looking at this based on the current IB at `setup_registry` the modules were not yet set so this filter should run instead at `get_available_modules()` method or at `get_active_modules()` as at this point all the modules were created and we know to have access to the property `force_active` on each module.

Let me know what approach do you think is best here and how we should go about updating the IB of this ticket.

Thank you.

includes/Core/Modules/Modules.php
```diff
@@ -566,7 +566,18 @@ protected function get_registry() {
 	 * @return Module_Registry
 	 */
 	protected function setup_registry() {
-		$registry          = new Module_Registry();
+		$registry = new Module_Registry();
+		/**
+		 * Filters an array with the list of all the available modules slugs, each slug present on this array would
+		 * be registered for inclusion. If a module is `force_activate` it would be included even if the module is
```
henrywright, line 572: Very tiny issue

Looking through the repository, the search-console and site-verification are set to `force_active`, not `force_activate`.

includes/Core/Modules/Modules.php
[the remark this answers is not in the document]
henrywright: just an idea

If you define `force_active` as a public static class variable you could get the `force_active` value without needing a class instance. That said, I'm not a huge fan of using them because they act like global variables, albeit namespaced global variables. There is likely a cleaner way to do it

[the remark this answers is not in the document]
eugene-manuilov: ugh... I forgot about it :smile:. Yes, let's go with the static approach in this case. But let's use static methods instead of variables. We will need to do the following:

1. Define a new public static method `is_force_active()` in the `Module` class that returns just `false` by default.
2. Update the `Module::parse_info` method to use `static::is_force_active()` method to set the value for the `force_active` property instead of hardcoded `false`: https://github.com/google/site-kit-wp/blob/e6b6890a69213a970cec7c1ca1837e72df9f1150/includes/Core/Modules/Module.php#L690
3. Add `is_force_active()` methods that return `true` to `Search_Console` and `Site_Verification` classes and update the appropriate `setup_info` methods in those classes to use the new static method.
4. Update `Modules::setup_registry` method to check whether the module is active using the new static method.

cc @felixarntz to chime in if he has a better idea.

eugene-manuilov (changes requested):
Thanks, @mitogh. Added a few more comments.

includes/Core/Modules/Modules.php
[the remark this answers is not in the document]
henrywright: good idea to use methods instead of properties @eugene-manuilov

includes/Core/Modules/Modules.php
mitogh: Looks like a bad on my side it should be `force_active` .

includes/Core/Modules/Module.php
```diff
@@ -793,4 +793,14 @@ function( $item ) {
 		return $items;
 	}
 
+	/**
+	 * Default behavior for all the modules, is not to be forced to be active.
```
eugene-manuilov, line 797:
```suggestion
	 * Determines whether the current module is forced to be active or not.
```

includes/Core/Modules/Module.php
```diff
@@ -793,4 +793,14 @@ function( $item ) {
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return bool Modules are not forced to be active by default. `false` is returned by default.
```
eugene-manuilov, line 801:
```suggestion
	 * @return bool TRUE if the module forced to be active, otherwise FALSE.
```

includes/Modules/Search_Console.php
```diff
@@ -553,4 +558,15 @@ protected function setup_assets() {
 		);
 	}
 
+	/**
+	 * Flag to indicate this module should be force activated regardless
+	 * if a user removes the module by filters.
```
eugene-manuilov, line 563:
```suggestion
	 * Returns TRUE to indicate that this module should be always active.
```

includes/Modules/Site_Verification.php
```diff
@@ -499,4 +504,17 @@ private function serve_verification_file( $verification_token ) {
 
 		// If the user does not have the necessary permissions then let the request pass through.
 	}
+
+
+	/**
+	 * Flag to indicate this module should be force activated regardless
+	 * if a user removes the module by filters.
```
eugene-manuilov, line 511:
```suggestion
	 * Returns TRUE to indicate that this module should be always active.
```

eugene-manuilov (changes requested):
Thanks, @mitogh. This almost looks good to me. Added a few more nitpicks that I'll apply by myself and will approve this PR. Nice work!

eugene-manuilov (approved):

includes/Core/Modules/Modules.php
```diff
@@ -567,9 +567,24 @@ protected function get_registry() {
...
-			$registry->register( $core_module );
+		/**
+		 * Filters core module slugs before registering them in the module registry. Each slug presented on this array will
+		 * be registered for inclusion. If a module is forced to be active, then it will be included even if the module slug is
+		 * removed from this filter.
+		 *
+		 * @param array $available_modules An array of core module slugs available for registration in the module registry.
+		 *
+		 * @since n.e.x.t
+		 *
```
felixarntz, line 578: A bit of a nit-pick, but `@param` should always come right before `@return` (without blank line), and `@since` should be before `@param` (with blank line in between).

includes/Modules/Search_Console.php
```diff
@@ -553,4 +558,14 @@ protected function setup_assets() {
...
+	/**
+	 * Returns TRUE to indicate that this module should be always active.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return bool Returns `true` indicating that this module should be activated all the time.
+	 */
+	public static function is_force_active() {
+		return true;
+	}
```
felixarntz, line 570: Now that this method is used to mark a module as `force_active`, we should probably remove the duplicate directive within the `Search_Console::setup_info` method. It will already be force-active because of this new method here, and otherwise we'll have to maintain this in two places.

includes/Modules/Site_Verification.php
```diff
@@ -499,4 +504,16 @@ private function serve_verification_file( $verification_token ) {
...
+	/**
+	 * Returns TRUE to indicate that this module should be always active.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return bool Returns `true` indicating that this module should be activated all the time.
+	 */
+	public static function is_force_active() {
+		return true;
+	}
```
felixarntz, line 518: Same as above, now that this is here, we should probably remove the duplicate directive in `Site_Verification::setup_info`.

felixarntz (commented):
A few minor points that IMO are still worth addressing (found as part of the _Approval_ process).

Issue #2697, opened 2021-01-27
Update feature flags to use PHP based configuration


eugene-manuilov:
## Feature Description

Currently, the Feature Flags feature relies on the JSON based configuration file and uses the `JSON_File` class to load available features into PHP. Overall it's okay solution, however, it has a very minimal but steady negative impact on the overall performance of the site because the plugin reads the json file from the server on every request. FS operations are I/O bound and are slower than including PHP scripts because PHP can cache scripts in OPcache and reuse it for the next requests.

We should consider reworking the way how we load feature flags to rely on the PHP based configuration. Ideally, we should update our webpack config to generate and emit a PHP script with feature flags configuration and import it in the PHP.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* 

## Implementation Brief

* 

### Test Coverage

* 

### Visual Regression Changes

* 

## QA Brief

* 

## Changelog entry

*

aaemnnosttv:
@felixarntz maybe we don't even need `config.json` anymore? We should still expose available features but maybe there's a cleaner way to do this.

felixarntz:
@aaemnnosttv @eugene-manuilov It's a good point we shouldn't need a JSON file anymore. I think a central configuration file is still beneficial, especially since it can be generated. Maybe we should just make it a PHP file, e.g. `config.php`.

danielgent:
I had a go with this for a few hours but I think I'm spending too much time on it so I'll work on this sprint first, then ask for help


So this seems to work fine locally running the website.  It worked straight away with my solution here https://github.com/google/site-kit-wp/pull/3308

**BUT** requiring a `.php` file is a big change.  As opposed to loading a `config.json` file at run time, we're now requiring a `config.php` file in the `dist` folder all the time.

I managed to get the PHP tests working by building with npm first (which surely must add a lot of time to the tests!)

But I've not been successful with the e2e tests. I managed to get the e2e tests to start (by building with webpack before activating the Site Kit), but the tests try and view the PHP file which complains (this is where the require is to a presumably non-existent file)
![hang-here-e2e](https://user-images.githubusercontent.com/8496063/117319432-96e1ce00-ae8b-11eb-8390-f979ff1fd063.png)

danielgent:
I'm unassigning myself from this as I believe the scope of this ticket is much larger than initially anticipated :smile:

eugene-manuilov:
[...]

PR #4547, opened 2021-12-13
Ticket/2697 configuration from json to php

mitogh:
## Summary

 Create a `config.php` file instead of a `.json` one
    
Following the IB from #2697 a new configuration file was created as a `.php` type of file instead of a `.json` one.

Removal of non unsed files any more as well. Testing is and JSON parsing file are no longer required.

Addresses issue #2697

## Checklist


tests/phpunit/includes/TestCase.php
```diff
@@ -29,7 +28,13 @@ public static function setUpBeforeClass() {
 		parent::setUpBeforeClass();
 
 		if ( ! self::$featureFlagsConfig ) {
-			self::$featureFlagsConfig = new JSON_File( GOOGLESITEKIT_PLUGIN_DIR_PATH . 'feature-flags.json' );
+			self::$featureFlagsConfig = array();
+			if ( file_exists( GOOGLESITEKIT_PLUGIN_DIR_PATH . 'feature-flags.json' ) ) {
```
aaemnnosttv, line 32: We don't need to check if this file exists because it's under version control (it's only not present in the built zip, but neither are tests).

webpack.config.js
```diff
@@ -35,6 +35,9 @@ const { BundleAnalyzerPlugin } = require( 'webpack-bundle-analyzer' );
 const CreateFileWebpack = require( 'create-file-webpack' );
 const ManifestPlugin = require( 'webpack-manifest-plugin' );
 const features = require( './feature-flags.json' );
+const formattedFeaturesToPHPArray = Array.isArray( features )
+	? features.map( ( feature ) => `'${ feature }'` ).join( ',' )
+	: '';
```
aaemnnosttv, line 40: We don't need the array check here because it's our own import which isn't dynamic so we can rely on it always being an array. If it wasn't something is probably wrong and it's probably better for it to fail here 👍

aaemnnosttv (changes requested):
Thanks @mitogh – this looks good, just a few minor changes to address 👍

tests/phpunit/includes/TestCase.php
mitogh: In tests, this would create some errors if it does not exist as the wrapper `JSON_File` was doing that for us if we include it tests would fail due to the warnings.

aaemnnosttv: I tried reproducing this in the tests but wasn't able to. Can you show me the failure you're getting without this?

As an aside, I think most of this flag resetting here is unnecessary as this array only sets the list of available flags now which doesn't have any affect on feature activation (except that it requires the list to be non-empty). We can probably actually remove most of this but let's not do that just yet.

webpack.config.js
```diff
@@ -252,11 +255,8 @@ function* webpackConfig( env, argv ) {
 			} ),
 			new CreateFileWebpack( {
 				path: './dist',
-				fileName: 'config.json',
-				content: JSON.stringify( {
-					buildMode: flagMode,
-					features,
-				} ),
+				fileName: 'config.php',
+				content: `<?php return array( 'buildMode' => '${ flagMode }', 'features' => array( ${ formattedFeaturesToPHPArray } ), ); ?>`,
```
felixarntz, line 259: Leaving a note here after talking through this with @aaemnnosttv: Let's create a "properly" formatted file here, with the license header - it might be good to rely on a template in JS, like we're doing already for the manifest file.

mitogh: Sounds like a plan :100:

tests/phpunit/includes/TestCase.php
mitogh: This was a mistake on my end by confusing this file with the file from the main plugin file.

I just updated it by removing this conditional. Thanks for the feedback and follow-up patch, that we can introduce in the upcoming ticket.

aaemnnosttv (approved):
LGTM, thanks @mitogh !

PR #4551, opened 2021-12-14
Reduce spacing for optimize setup.

asvinb:
## Summary

Addresses issue #4028 

## Relevant technical choices
The styling is pretty specific to the Optimize setup form since we have `flex-wrap: wrap` on the flex items. This means we don't have much control over when the elements are stacked in order to remove the extra margin.

## Checklist

### DIFF
--- a/assets/js/modules/optimize/components/common/OptimizeIDField.js
+++ b/assets/js/modules/optimize/components/common/OptimizeIDField.js
@@ -24,7 +24,7 @@ import classnames from 'classnames';
 /**
  * WordPress dependencies
  */
-import { Fragment, useCallback } from '@wordpress/element';
+import { useCallback } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 
 /**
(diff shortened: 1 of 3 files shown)


assets/sass/components/setup/_googlesitekit-setup.scss
```diff
@@ -114,6 +114,26 @@
 		padding-left: 10px;
 		padding-right: 10px;
 	}
+
+	.googlesitekit-optimize-setup__form {
```
aaemnnosttv, line 118: I think these styles may not actually be necessary. Most of the styling appears to be fixed by wrapping the optimize field with a `div`.

aaemnnosttv (changes requested):
Thanks @asvinb !

As per my last comment below, I found that the main problem here seems to be the missing wrapping div for the `OptimizeIDField`. All other similar components `AdsConversionIDTextField`, `ProfileNameTextField`, `WebContainerNameTextField`, etc all have wrapping divs (some with classes) but for some reason `OptimizeIDField` uses a `Fragment` which is what causes the `googlesitekit-setup-module__inputs > div` style to add too much margin where it shouldn't.

Let's add this missing wrapper and then only add any additional styles as needed.

assets/js/modules/optimize/components/common/OptimizeIDField.js
```diff
@@ -61,7 +61,7 @@ export default function OptimizeIDField() {
 				name="optimizeID"
 				onChange={ onChange }
 				helperText={
-					<HelperText>
+					<HelperText persistent>
```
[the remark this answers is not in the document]
asvinb: 👍

assets/sass/components/setup/_googlesitekit-setup-module.scss
```diff
@@ -128,7 +147,11 @@
 		}
 
 		+ .googlesitekit-error-text {
-			margin-top: -24px;
```
aaemnnosttv, line 131: 👍

aaemnnosttv (approved):
LGTM 👍  Thanks @asvinb !

Issue #4247, opened 2021-10-19
Print or display an old site URL or provide insights along with the "Looks like your URL has changed."


jamesozzie:
## Bug Description

Many users open support topics when they encounter the "_Looks like your URL has changed_" notice, in many cases unsure of such a URL change. This can occur in the following situations:
- When a site has both it's http and https site available, without redirection
- When using a language plugin that may add a /en or other suffix to a homepage
- For site migrations or domain changes

In addition to the error message it might be useful to print the old site URL, in the form of a console error or using a description along the lines of the below:
_..Looks like your URL has changed from ***** to *****._ 

This may help users determine a plugin related issue or identify a https status change. 

[Example support topic](https://wordpress.org/support/topic/looks-like-the-url-of-your-site-has-changed-in-order-to-continue-using-site-kit-2/#post-14957342)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

Update the current notice to include the old and new URLs:

_Reconnect Site Kit_
_Looks like the URL of your site has changed. In order to continue using Site Kit, you’ll need to reconnect, so that your plugin settings are updated with the new URL._
_- Old URL: https://www.example.com_
_- New URL: https://www.updatedexample.com_

## Implementation Brief

The final implementation changed from the original definition. See https://github.com/google/site-kit-wp/pull/4565

<details>
	<summary>Original IB</summary>
	
* Create `includes/Core/Authentication/Previous_Connected_Proxy_URL.php` which contains the `Previous_Connected_Proxy_URL` class.
	* The class extends the `Connected_Proxy_URL` class.
	* Set the `OPTION` to the `OPTION` of the parent class (`Connected_Proxy_URL::OPTION`), with `_previous` appended.
	* Add the `register` public function which:
		* is hooked to the `update_option_{$option}` action where `$option` is the `Connected_Proxy_URL::OPTION`.
			* The callback function accepts 1 parameter which is the old value for that option and uses the `set` method to set the value of the option to the old value of the parent option.
* Using `includes/Core/Authentication/Authentication.php`,
	* Add a new protected property `previous_connected_proxy_url`.
	* In the constructor, set `previous_connected_proxy_url` to a new instance of the `Previous_Connected_Proxy_URL` class.
	* In the `register` function, call the `register` method of `previous_connected_proxy_url`
	* Update the `get_rest_routes` method to include the above options in the array returned for the `core/user/data/authentication` REST route. The keys can be:
[...]

PR #4565, opened 2021-12-15
Change order of conditions.

ivankruchkoff:
## Summary

Addresses issue #4247

## Checklist


includes/Core/Authentication/Authentication.php
```diff
@@ -980,12 +980,41 @@ private function get_reconnect_after_url_mismatch_notice() {
...
+							sprintf(
+							/* translators: 1: Previous URL. 2: Current URL */
+								esc_html__( 'Looks like the URL of your site has changed. In order to continue using Site Kit, you’ll need to reconnect, so that your plugin settings are updated with the new URL. Old url was %1$s, new url is %2$s.', 'google-site-kit' ),
+								esc_url( $connected_url ),
+								esc_url( get_home_url() )
+							),
+							esc_url( $this->get_proxy_setup_url() ),
+							esc_html__( 'Reconnect', 'google-site-kit' )
+						);
+					}
```
felixarntz, line 1015: This can be a lot simplified by going with my suggestion from https://github.com/google/site-kit-wp/issues/4247#issuecomment-995226116. We don't need to entirely separate conditions, we can keep the `p` tag the same no matter what. Then, conditionally only append the `ul` list with the old and new URL.

felixarntz (commented):
@ivankruchkoff The splash UI changes look good, however please have a look at my feedback for the new admin notice UI in https://github.com/google/site-kit-wp/issues/4247#issuecomment-995226116, which I think could be improved. That would furthermore also allow us to not use the more complicated logic you've added here (also see comment below).

includes/Core/Authentication/Authentication.php
ivankruchkoff: Yes, I can make that update, but the question in the ticket is more around the logic for which URLs to show.  In JS, we only show a message if there is a value set for both connected url and previous connected url https://github.com/google/site-kit-wp/blob/681fc336694ab2cc15221c6952e24792c75b17e0/assets/js/components/setup/SetupUsingProxy.js#L286-L293 

In PHP the approach covers an additional scenario:

> The issue we could now face is that I update my site kit plugin with the new version which introduces Previous Connected Proxy URL class, but has no value set, I then change my URL via the above filter, and get to a scenario where the url has changed, that previous_connected_url is not set (false), hence this alternate codepath

aaemnnosttv: I've pushed a few commits to simplify things here as it turns out the previous connected proxy URL wasn't actually needed as this is what the connected proxy URL would have. We always have access to the current URL (connected or not).

felixarntz (approved):
LGTM!

Issue #4562, opened 2021-12-15
After pressing enter on a URL Entity Search Result, the loading bar appears before redirect


tofumatt:
## Bug Description

When the URL Entity Search input is activated (either by clicking on an autocomplete result sometimes, or consistently by selecting one with the keyboard then pressing the "Enter"/"Return" key), the loading bar that shows when searching appears again underneath the URL Entity Search input, even though it shouldn't be searching and is actually redirecting the user to a new Entity Dashboard page.

**EDIT: Upon further testing, this seems to be intermittently reproducible with the mouse, but always reproducible with the keyboard.**

## Steps to reproduce

1. Enable `unifiedDashboard` feature flag
2. Enter a valid page/post title/URL into the URL Entity Search input
3. Select the result using the arrow keys + Enter/Return
4. See the search loading indicator below the input element whilst the redirect is taking place

## Screenshots

![2021-12-15 15 34 38](https://user-images.githubusercontent.com/90871/146216559-6ea59974-6320-4d3f-b5ed-f3e7f86ba8d4.gif)

## Additional Context

It looks like this was introduced sometime after 48288d66879a0b2ff93407dc87a9fb1c5259f704, at least. It's possible that #4486 introduced code that would cause `isLoading` to be set after the results are selected, causing it to appear like it's loading again. 🤔 

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The search loading indicator should not appear after selecting an already-loaded result and clicking on it.

## Implementation Brief

Eventually the logic and amount of state/`setState` data/functions being passed through here could be improved, but for now (while we need to support both the `unifiedDashboard` and non-Unified Dashboard searches), this approach should fix the bug above without breaking anything or requiring complicated refactoring.

* Pass `match={ match }` to [`<PostSearcherAutoSuggest />` in the `EntitySearchInput.js` component](https://github.com/google/site-kit-wp/blob/b0c598be8bdc2fe1216d9763af8afa9f09095366/assets/js/components/EntitySearchInput.js#L115)
* Accept `match` as a prop in `PostSearcherAutoSuggest`.
  * Add a check for `debouncedValue !== match?.post_title` in the [`useEffect` that sets `setIsLoading?.( true );`](https://github.com/google/site-kit-wp/blob/b0c598be8bdc2fe1216d9763af8afa9f09095366/assets/js/components/PostSearcherAutoSuggest.js#L119)

[...]

PR #4573, opened 2021-12-16
Remove loading indicator after pressing enter on a URL Entity Search Result from the auto complete

hussain-t:
## Summary

Addresses issue #4562 

## Checklist


assets/js/components/PostSearcherAutoSuggest.js
```diff
@@ -59,7 +60,25 @@ export default function PostSearcherAutoSuggest( {
...
+	 * `debouncedValue !== postTitleFromMatch` to set the loading state to true.
+	 * However, the `debouncedValue` is always delayed. Hence this condition is always true.
+	 * Hence the loading state is to true.
+	 * Therefore, we need to set the debounce delay/timer value to `0` if the
+	 * `searchTerm === postTitleFromMatch` to not to set the loading state to true.
+	 */
+	const debouncedValue = useDebouncedState(
+		searchTerm,
+		searchTerm === postTitleFromMatch ? 0 : 200
+	);
```
eugene-manuilov, line 80: I think we don't need this. What we really need is to make sure that we compare lower cased versions of strings. It will fix the problem even without changing the debounce delay.

```js
useEffect( () => {
	if (
		debouncedValue !== '' &&
		debouncedValue?.toLowerCase() !== currentEntityTitle?.toLowerCase() &&
		debouncedValue?.toLowerCase() !== postTitleFromMatch.toLowerCase()
	) {
		...
	}
} );
```

I think we need to update all checks in this file to make sure we compare strings correctly. Also, let's make sure we don't have tons of `.toLowerCase()` calls in this component. Let's cast all strings to the lower cases and assign them to variables.

eugene-manuilov (changes requested):

assets/js/components/PostSearcherAutoSuggest.js
hussain-t: Thanks @eugene-manuilov. I changed all the checks and tested as per the comment. However, it didn’t solve the issue. Could you explain how does comparing the lower case strings solve the issue?

eugene-manuilov: Ok, I think you are right, we need this. Looks like I spotted another issue that is fixed by comparing lowercased versions of strings.

Try to enter `hello world!` (the exact match for the `Hello world!` article title) and hit enter, you will see that the progress bar is shown again, but it shouldn't. That happens because `postTitleFromMatch` doesn't equal `debouncedValue` since one is `Hello world!` and another one is `hello world!`.

hussain-t: @eugene-manuilov Yes, that's right 👍 However, in that scenario the progress bar didn’t appear but `No results found` appeared. By setting the lower cased string fixes the issue. I think we need lower cased string only for this condition:

```javascript
debouncedValue?.toLowerCase() !== postTitleFromMatch?.toLowerCase()
```

eugene-manuilov (approved):
Awesome! Thanks, @hussain-t. LGTM.

PR #4580, opened 2021-12-17
Feat/pr checklist

eugene-manuilov:
## Summary

Addresses issue #4581

## Checklist

---------------

## Code Review Checklist

- [x] Run the code.
- [x] Ensure the acceptance criteria are covered.
- [x] Stay open-minded.
- [x] Ensure CI checks pass.

## Merge Review Checklist


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -16,3 +16,29 @@ Addresses issue #
 - [ ] My code has proper inline documentation.
 - [ ] I have added a QA Brief on the issue linked above.
 - [ ] I have signed the Contributor License Agreement (see <https://cla.developers.google.com/>).
+
+## Code Review Checklist
```
felixarntz, line 20: We should probably visually separate the two new areas more from the rest of the template, almost like we have in the issue template for the workflow sections. It needs to be clear that the person opening a PR doesn't need to do anything here (and also shouldn't remove them).

Also, the original checklist above (which the PR author _should_ use) is now still only called "Checklist", we may want to clarify that, now that there are other checklists.

.github/workflows/code-review-checklist.yml
```diff
@@ -0,0 +1,20 @@
...
+      - reopened
+      - edited
+
+jobs:
+  checklist:
+    name: Checklist Checks
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check non-resolved tasks
+        if: contains(github.event.pull_request.body, '- [ ]')
```
felixarntz, line 19: That's great to have in either way 👍

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -16,3 +16,29 @@ Addresses issue #
...
+
+- [ ] **Run the code.** 
+  <details><em>Do not just look at the code (the “minimum” code review), but also briefly test it in your development environment, especially if the pull request touches production code. This should explicitly not be QA, but a basic “sanity check” that the PR is not significantly off or even broken.</em></details>
```
felixarntz, line 23: Why wrap all the explanation in `em`? Is there a problem with the formatting we'd get otherwise?

felixarntz (commented):
@eugene-manuilov +1 to the idea here, it's important to consider all of these steps in every review!

I'm not sure we need to put the full explanation for each step in here, as that makes the template quite bulky. Since all engineers are expected to read our code review guidelines, I think that part can be omitted. What we should rather do then is add them somewhere else to GitHub (e.g. in `docs` or the Wiki) and link to the full doc from here, to get all the details.

.github/PULL_REQUEST_TEMPLATE.md
eugene-manuilov: Just wanted to differentiate checklist titles and descriptions. Anyway removed descriptions.

.github/PULL_REQUEST_TEMPLATE.md
eugene-manuilov: Yep, updated.

eugene-manuilov:
> I'm not sure we need to put the full explanation for each step in here, as that makes the template quite bulky. Since all engineers are expected to read our code review guidelines, I think that part can be omitted. What we should rather do then is add them somewhere else to GitHub (e.g. in docs or the Wiki) and link to the full doc from here, to get all the details.

Yep, makes sense. I have updated the template to have only checklist items without descriptions. Not added it to GitHub though, not sure what should be there and where it should be. Probably out-of-scope for this ticket?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -16,3 +16,21 @@ Addresses issue #
 - [ ] My code has proper inline documentation.
 - [ ] I have added a QA Brief on the issue linked above.
 - [ ] I have signed the Contributor License Agreement (see <https://cla.developers.google.com/>).
+
+---------------
```
felixarntz, line 20: Let's also add something like this here, similar to how we have in issues:
```suggestion
---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._
```

Maybe "code reviewers" would be more precise here than saying "moderators", but I don't have a strong opinion - I think the term "moderators" has the benefit that it clarifies that a 3P contributor shouldn't touch this section. Either works for me.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -16,3 +16,21 @@ Addresses issue #
...
+
+## Code Reviewer Checklist
+
+- [ ] Run the code.
+- [ ] Ensure the acceptance criteria are covered.
+- [ ] Stay open-minded.
```
felixarntz, line 26: I think this reads a bit weird now as its own requirement - you should always stay open minded. :) This point was more specifically about following the IB where appropriate, but also allowing for diversion from the IB if it makes sense.

Maybe something like "Check whether the approach from the IB still makes sense here in practice"? @aaemnnosttv Any wording suggestion?

felixarntz (commented):
@eugene-manuilov Two more comments here, other than that lgtm!

cc @aaemnnosttv for feedback/ideas on the wording

.github/PULL_REQUEST_TEMPLATE.md
eugene-manuilov: Thanks. Either works for me too. Accepted your suggestion.

.github/PULL_REQUEST_TEMPLATE.md
aaemnnosttv: The points in the handbook have changed, so let's update this accordingly.
```suggestion
- [ ] Reassess the implementation with the IB.
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -16,3 +16,23 @@ Addresses issue #
...
 - [ ] I have signed the Contributor License Agreement (see <https://cla.developers.google.com/>).
+
+---------------
+
+_Do not alter or remove anything below. The following sections will be managed by moderators only._
+
+## Code Reviewer Checklist
+
+- [ ] Run the code.
+- [ ] Ensure the acceptance criteria are covered.
```
aaemnnosttv, line 27:
```suggestion
- [ ] Ensure the acceptance criteria are satisfied.
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -16,3 +16,23 @@ Addresses issue #
...
+- [ ] Stay open-minded.
+- [ ] Ensure CI checks pass.
```
aaemnnosttv, line 29:
```suggestion
- [ ] Ensure no unrelated changes are included.
- [ ] Ensure CI checks pass.
```

.github/workflows/code-review-checklist.yml
```diff
@@ -0,0 +1,20 @@
...
+
+on:
+  pull_request:
+    branches:
+      - develop
+      - main
+    types:
+      - opened
+      - reopened
+      - edited
```
aaemnnosttv, line 11: How long will this job take to run? It looks pretty minimal but it also looks like this will trigger a job every time someone clicks a checkbox 😬 

Also, do we want this running on draft PRs?

aaemnnosttv (changes requested):
Thanks @eugene-manuilov – this looks good, just a few updates are needed for consistency with the latest docs. Also a few questions below.

.github/workflows/code-review-checklist.yml
eugene-manuilov: > How long will this job take to run? 

A second or two.

> It looks pretty minimal but it also looks like this will trigger a job every time someone clicks a checkbox

Yes, you are right. This is not a big deal I think, but the fact that it will generate a ton of emails is a little bit annoying 😄. Unfortunately, I haven't found a way to disable notifications for a certain job. If this is a problem, then I think we need to remove this job. cc @felixarntz 

> Also, do we want this running on draft PRs?

Good point, we don't need to run if for draft PRs. Updated it to run only for PRs that are ready for review.

aaemnnosttv: > Yes, you are right. This is not a big deal I think, but the fact that it will generate a ton of emails is a little bit annoying 😄

I think it could be problematic as each job will fill up our queue of jobs to run and then run in order they were added which adds some amount of overhead for each separate job, when really only the last one is important.

I like the idea here, but I'm not sure it's worth adding this as an automated check, especially since we don't require all checks to pass before merging anyways. If checking the boxes becomes significant in that a check will fail as a result, engineers will be somewhat incentivized to check them all ahead of time without really paying attention to them. I would say let's keep this managed manually and try to use this as a useful reminder, unless of course @felixarntz would prefer to keep it 😄

aaemnnosttv (commented):
Template changes LGTM, but I have some concerns about the added check/action. Leaving for @felixarntz to decide.

If anything, maybe we should keep this issue limited to the template and reassess a potential automation in a new issue.

.github/workflows/code-review-checklist.yml
felixarntz: Yeah, I would prefer to keep this only a checklist for awareness, we don't need to enforce anything at the automation level.

felixarntz (commented):
@eugene-manuilov The PR template LGTM, however I think we don't need a GH action to enforce this.

eugene-manuilov:
@felixarntz @aaemnnosttv sounds good to me. The workflow file is removed.

felixarntz (approved):
Great!

aaemnnosttv (approved):
:shipit:

Issue #4457, opened 2021-12-03
Update items in the header for consistency


aaemnnosttv:
## Feature Description

Currently we have some inconsistency in the layout of items that are present in the main Site Kit header. This is most noticeable in mobile where the items are reduced to circular icon buttons.

### Current header in mobile

![image](https://user-images.githubusercontent.com/1621608/144454353-9813bc2c-61a8-431c-bcbd-fdb2c3d2668c.png)
* User menu is farther from the edge than the Google logo on the other side

| User menu | Help menu |
| --- | --- |
| ![image](https://user-images.githubusercontent.com/1621608/144454858-05a48d05-7b2c-4ec0-8c7c-236bb78833cd.png) | ![image](https://user-images.githubusercontent.com/1621608/144455051-63fea0a6-75b8-43aa-b289-416ac97c7680.png) |

* User menu active background is rectangular rather than matching the circular shape of the button in mobile (like the Help menu)

![image](https://user-images.githubusercontent.com/1621608/144455441-2883a58b-3428-41e3-ad34-0b0ccc1f1f90.png)
* User menu button is too wide in mobile which throws off the spacing
 
### New header with `unifiedDashboard`

![image](https://user-images.githubusercontent.com/1621608/144645032-15020973-2d80-4b0d-b55f-6b8179709e95.png)
![image](https://user-images.githubusercontent.com/1621608/144645040-c4999b76-4376-484f-a74a-de87d8022856.png)

* Chevron removed but extra space for it seems to still be there

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* First and last items in the header should have a consistent distance to the closest edge (i.e. user menu should be the same distance from the right edge as the Google logo is to the left) which is visually consistent with the margin/gap for the viewport size
	* The distance should be measured to the closest visual edge when inactive (the larger area occupied by the background when active should not be considered here)
	* The chevron for the date range dropdown would also count as an edge when present
* Items in the header should have the same distance between them using the responsive grid gap for the viewport size
* Icon buttons (icon only, no text) should always have a circular background

## Implementation Brief

* 

### Test Coverage

* 

## QA Brief

* 

## Changelog entry

*

aaemnnosttv:
@felixarntz anything to add or change about the ACs? If not, let's move it forward 👍

felixarntz:
Thanks @aaemnnosttv, ACs lgtm!

kuasha420:
[...]

PR #4587, opened 2021-12-20
Fix styles in the header mobile view

hussain-t:
## Summary

Addresses issue #4457 

## Checklist


assets/sass/components/global/_googlesitekit-entity-search.scss
```diff
@@ -42,7 +42,17 @@
...
+			/**
+			*	We need to set 5px left and right padding for
+			*	larger screens since we have set the close button
+			*	icon size 20 for consistency.
+			*/
+			padding: 0 5px;
+
+			@media (max-width: $bp-mobileOnly ) {
+				padding: 0;
+			}
```
hussain-t, line 55: If we don’t want these changes, we need to revert the `CloseDark` icon to have the width `30`.

```javascript
<CloseDark width="20" height="20" />
```

assets/sass/components/global/_googlesitekit-header.scss
```diff
@@ -90,8 +90,8 @@
 			.mdc-button__icon,
 			.mdc-button__icon--image {
 				// increase default button image size
-				height: 28px;
-				width: 28px;
+				height: 24px;
```
asvinb, line 93: @hussain-t As per the AC, the width and height of the image should be `28px`. You can set the padding to the parent button (`.googlesitekit-user-selector button`) to `4px` so that the total width is `36px` (28 + 4 + 4).

assets/svg/magnifying-glass.svg
```diff
@@ -1 +1,3 @@
-<svg width="16" height="16" fill="none" xmlns="http://www.w3.org/2000/svg"><path fill-rule="evenodd" clip-rule="evenodd" d="M10.758 9.395 16 14.637 14.637 16l-5.242-5.242c-.979.705-2.159 1.135-3.449 1.135a5.946 5.946 0 1 1 5.947-5.947c0 1.29-.43 2.47-1.135 3.45ZM5.946 1.83A4.111 4.111 0 0 0 1.83 5.946a4.111 4.111 0 0 0 4.116 4.117 4.111 4.111 0 0 0 4.117-4.117A4.111 4.111 0 0 0 5.946 1.83Z" fill="currentColor"/></svg>
\ No newline at end of file
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path fill-rule="evenodd" clip-rule="evenodd" d="M16 9.5C16 10.9069 15.5495 12.2145 14.7819 13.2827L20.49 19L19 20.49L13.2827 14.7819C12.2145 15.5495 10.9069 16 9.5 16C5.91 16 3 13.09 3 9.5C3 5.91 5.91 3 9.5 3C13.09 3 16 5.91 16 9.5ZM5 9.5C5 11.99 7.01 14 9.5 14C11.99 14 14 11.99 14 9.5C14 7.01 11.99 5 9.5 5C7.01 5 5 7.01 5 9.5Z" fill="#5F6368"/>
```
asvinb, line 2: fill should be set to `currentColor` and why the file was updated? 🤔

assets/js/components/EntitySearchInput.js
```diff
@@ -137,8 +137,9 @@ function EntitySearchInput() {
 				<div className="googlesitekit-entity-search__actions">
 					<Button
 						onClick={ onClose }
-						trailingIcon={ <CloseDark width="30" height="20" /> }
+						trailingIcon={ <CloseDark width="20" height="20" /> }
 						className="googlesitekit-entity-search__close"
+						title="Close"
```
asvinb, line 142: @hussain-t Strings must be localized.
```suggestion
						title={ __( 'Close', 'google-site-kit' ) }
```

assets/js/components/EntitySearchInput.js
```diff
@@ -149,10 +150,12 @@ function EntitySearchInput() {
 	return (
 		<div className="googlesitekit-entity-search">
 			<Button
+				className="googlesitekit-border-radius-round--phone googlesitekit-button-icon--phone"
 				onClick={ onOpen }
 				text
 				ref={ buttonRef }
-				trailingIcon={ <MagnifyingGlass width="16" height="16" /> }
+				title="Search"
```
asvinb, line 157: Same here:
```suggestion
				title=__( 'Search, 'google-site-kit' )
```

assets/js/components/DateRangeSelector.js
```diff
@@ -100,13 +100,15 @@ export default function DateRangeSelector() {
...
+					'googlesitekit-border-radius-round--phone',
+					'googlesitekit-button-icon--phone',
 					{
 						'googlesitekit-header__date-range-selector-menu--has-unified-dashboard': unifiedDashboardEnabled,
 					}
 				) }
 				text
 				onClick={ handleMenu }
-				icon={ <DateRangeIcon width="18" height="20" /> }
+				icon={ <DateRangeIcon width="20" height="20" /> }
```
asvinb, line 111: The button for the date range selector has incorrect width since it has `padding-right: 8px` which sets the width to: `44px`. Can you review to make the button `36px` on small screens as per the AC?

asvinb (changes requested):
Nice job @hussain-t . I left a few comments in your PR, can you check them out? Also I think we can further improve things in some places. Right now, the direct children of `.googlesitekit-header__children` individually set the spacing among themselves, either using `margin-left` or `margin-right` which is not consistent. Either we use `margin-left` for all, or `margin-right` for all. For e.g `.googlesitekit-entity-search` has `margin-right: 16px`, `.googlesitekit-date-range-selector` does not have any set. The help button and user menu have `margin-left` set. I think we can simplify all this by removing the margins set for each class and rather have something like this:
```css
.googlesitekit-header__children > div:not(:first-child) {
    margin-left: 16px;

    @media (min-width: $bp-desktop) {
        margin-left: 24px;
    }
}
```

assets/sass/components/global/_googlesitekit-entity-search.scss
```diff
@@ -42,7 +42,17 @@
 		.googlesitekit-entity-search__close {
```
asvinb, line 42: @hussain-t Can we have the close button rounded as well? The button should be `36x36` on small screens and `32x32` on large screens to be consistent with the height of the input text.

assets/sass/components/global/_googlesitekit-entity-search.scss
asvinb: @hussain-t Let's set it to 30 but see my previous comment about its dimensions.

assets/sass/components/global/_googlesitekit-entity-search.scss
```diff
@@ -42,7 +42,17 @@
...
 			height: 32px;
 			min-width: 36px;
-			padding: 0;
+
+			/**
+			*	We need to set 5px left and right padding for
+			*	larger screens since we have set the close button
+			*	icon size 20 for consistency.
+			*/
+			padding: 0 5px;
```
asvinb, line 51: @hussain-t The way you have it makes a harder to read and understand. Basically you want padding set to `0` on small screens only which can be a bit confusing when reading the code, because you set  `padding: 0 5px` and then you set it to `0` for mobile. In the mobile first approach, it's better to have the following:
```css
padding: 0;

@media (min-width: $bp-tablet) {
     padding: 0 5px;
}
```
It's easier to read and understand 😁 This may be redundant if we set the dimensions of the icon to `30` but worth noting for the future 👍

assets/sass/components/global/_googlesitekit-header.scss
```diff
@@ -115,6 +115,21 @@
 
 			.mdc-button--dropdown {
 				padding-right: $grid-gap-desktop;
+
+				@media (max-width: $bp-mobileOnly) {
+					padding-right: 8px;
+				}
+			}
+
+			@media (max-width: $bp-mobileOnly) {
```
asvinb, line 124: I would move this block to be just under `.googlesitekit-date-range-selector` in the file, i.e before the styles for `svg`. Once again, there is nothing wrong but it's easier to read and you know immediately those styles are for `.googlesitekit-date-range-selector`. Where you have it right now, i must scroll back to know for which selector those styles apply for.

asvinb (changes requested):
Some more CSS changes 👍

assets/svg/magnifying-glass.svg
hussain-t: The existing SVG icon was inappropriate size. Styling it didn’t work. Hence, I have exported a new one from the Figma design.

assets/sass/components/global/_googlesitekit-entity-search.scss
```diff
@@ -40,9 +39,16 @@
 		}
 
 		.googlesitekit-entity-search__close {
+			border-radius: 50%;
 			height: 32px;
-			min-width: 36px;
+			min-height: 36px;
+			min-width: 36px !important;
```
asvinb, line 45: @hussain-t Let's avoid using `!important` unless we don't have any options. For e.g if there are inline styles. What's happening here is that `min-width` is getting overridden by a the `min-width` of a more specific selector, which is: `.googlesitekit-plugin .googlesitekit-entity-search .mdc-button`. So in this case, we can look into using the existing selector, or add more specificity to `.googlesitekit-plugin .googlesitekit-entity-search .googlesitekit-entity-search__close` by doing something like that: `.googlesitekit-plugin .googlesitekit-entity-search .mdc-button.googlesitekit-entity-search__close`

assets/sass/components/global/_googlesitekit-entity-search.scss
```diff
@@ -40,9 +39,16 @@
...
 			padding: 0;
+
+			@media ( min-width: $bp-wpAdminBarTablet ) {
+				min-height: 32px;
+				min-width: 32px !important;
```
asvinb, line 50: @hussain-t See my comment above.

assets/sass/components/global/_googlesitekit-entity-search.scss
```diff
@@ -40,9 +39,16 @@
...
 
 		.googlesitekit-entity-search__close {
+			border-radius: 50%;
 			height: 32px;
-			min-width: 36px;
+			min-height: 36px;
+			min-width: 36px !important;
 			padding: 0;
+
+			@media ( min-width: $bp-wpAdminBarTablet ) {
```
asvinb, line 48: @hussain-t Can you check the breakpoints closely. I see sometime we are using `$bp-tablet` and sometimes `$bp-wpAdminBarTablet` 👍

assets/sass/components/global/_googlesitekit-entity-search.scss
asvinb: @hussain-t At `783px`, the search button is out of the input text. can you look into that?
image

Also, I forgot to mention, the button should be round on small screens only since it looks weird when it's round within the input text. When it's within the input text, let's make it square, to fill in the area within the input text.

asvinb:
> Nice job @hussain-t . I left a few comments in your PR, can you check them out? Also I think we can further improve things in some places. Right now, the direct children of `.googlesitekit-header__children` individually set the spacing among themselves, either using `margin-left` or `margin-right` which is not consistent. Either we use `margin-left` for all, or `margin-right` for all. For e.g `.googlesitekit-entity-search` has `margin-right: 16px`, `.googlesitekit-date-range-selector` does not have any set. The help button and user menu have `margin-left` set. I think we can simplify all this by removing the margins set for each class and rather have something like this:
> 
> ```css
> .googlesitekit-header__children > div:not(:first-child) {
>     margin-left: 16px;
> 
>     @media (min-width: $bp-desktop) {
>         margin-left: 24px;
>     }
> }
> ```

@hussain-t On small screens, the margin is consistent. However as from `783px`, the margin is not consistent. `googlesitekit-entity-search` has margin-right set to `24px` and a mixture of `margin-left` `24px` and `16px`. Can you check again? I would solely rely on `.googlesitekit-header__children > div:not(:first-child) {` for all the spacings across viewpoints and remove all the individual margins for each element. Let me know it you want to jump on a call 👍

asvinb (changes requested):
@hussain-t I left some more comments, can you check them out? Thanks!

assets/sass/components/global/_googlesitekit-header.scss
```diff
@@ -184,4 +216,12 @@
 		display: flex;
 		justify-content: flex-end;
 	}
+
+	.googlesitekit-header__children > div:not(:first-child) {
+		margin-left: $grid-gap-phone;
+
+		@media (min-width: $bp-desktop) {
+			margin-left: $grid-gap-desktop; // There is a large margin between the search icon and the date range
```
asvinb: @hussain-t I'am not sure the comment is still applicable. Can you remove it if it's the case? 😁

asvinb (approved):
@hussain-t There were a couple of minor tweaks which i applied myself. Basically there was [those lines](https://github.com/google/site-kit-wp/blob/develop/assets/sass/components/global/_googlesitekit-header.scss#L101-L107) which were redundant and causing some alignment mismatch on the tablet breakpoint.

hussain-t:
> @hussain-t There were a couple of minor tweaks which i applied myself. Basically there was [those lines](https://github.com/google/site-kit-wp/blob/develop/assets/sass/components/global/_googlesitekit-header.scss#L101-L107) which were redundant and causing some alignment mismatch on the tablet breakpoint.

Thanks a bunch @asvinb!

asvinb:
@hussain-t Can you check the failing VRT?

hussain-t:
Thanks @asvinb. It was failing due to the previous [commit](https://github.com/google/site-kit-wp/pull/4587/commits/cf7051ef9f8342ae001e41cf8ec0aa75a9b87d88). I have updated the VRT file 👍 

Screenshot 2022-01-05 at 2 38 59 PM

assets/sass/vendor/_mdc-button.scss
```diff
@@ -87,4 +95,19 @@
...
 		margin-left: 8px;
 	}
+
+	&.googlesitekit-button-icon {
+		min-height: auto;
+		min-width: auto;
+		padding: 8px;
+	}
+
+	&.googlesitekit-button-icon--phone {
```
aaemnnosttv, line 105: Let's not mix these style definitions in the vendor stylesheets which are more for implementing/adapting Material styles. These probably belong in a new `assets/sass/components/global/_googlesitekit-button-icon.scss` stylesheet.

assets/sass/components/global/_googlesitekit-header.scss
```diff
@@ -184,4 +209,12 @@
 		display: flex;
 		justify-content: flex-end;
 	}
+
+	.googlesitekit-header__children > div:not(:first-child) {
+		margin-left: $grid-gap-phone;
+
+		@media (min-width: $bp-desktop) {
+			margin-left: $grid-gap-desktop;
```
aaemnnosttv, line 217: Can we do this with `gap` on the `googlesitekit-header__children` instead?

Here's with the margin
image

Here's with `gap`
image

aaemnnosttv (changes requested):
Thanks @hussain-t ! This one is looking good and really close I think there are just a few details to touch up.

I left a few comments below, but I also wanted to highlight this odd sizing around the help menu icon. The inspector shows it is 36px square, but its parent is only 32px wide. Can you see why?

image

image

---

Also, in the non-unified dashboard I noticed a bit of a quirk with the width of the date range icon in mobile as well.

image

---

Finally, please resolve the merge conflict with the magnifying glass svg which recently moved 👍

aaemnnosttv:
Thanks @hussain-t ! This is almost there, just one thing I noticed in the tablet viewport on the non unified dashboard where the user menu button is in the mobile size but too wide.

image

The user menu button should be just as it is in mobile here (date range can stay as is though)

image

Finally, VRT is failing as a result of some of the last changes so please make sure these are up to date as just about any change here will likely cause them to need updating as well.

assets/js/components/UserMenu.js
```diff
@@ -161,18 +161,19 @@ export default function UserMenu() {
...
-				) }
+				className="googlesitekit-user-selector googlesitekit-dropdown-menu googlesitekit-dropdown-menu__icon-menu mdc-menu-surface--anchor"
 			>
 				<Button
-					className="googlesitekit-header__dropdown mdc-button--dropdown"
+					className={ classnames(
+						'googlesitekit-header__dropdown',
+						'mdc-button--dropdown',
+						// Adds circle background to the non-unified dashboard mobile view
+						'googlesitekit-border-radius-round--phone',
```
aaemnnosttv, line 171: This leaves out an odd in-between viewport in tablet where it appears as a icon button but still has a rectangular background.

aaemnnosttv (commented):
Thanks @hussain-t – this is so close, I just found one thing that was off in tablet which I'll push a quick fix for.

aaemnnosttv (approved):
Great!

Issue #4571, opened 2021-12-16
New Search Funnel widget still showing legacy blue zero data box


felixarntz:
When there is no Search Console data to display (to clarify: "zero data", not "gathering data"), the Search Funnel widget still shows the legacy blue CTA - which per #4085 and #4086 should only be shown if the service is actually gathering data. This should be fixed - it is especially problematic in the case that Analytics isn't activated yet, in which case that CTA will not be shown either.

In other words, the below UI should _only_ be shown if Search Console is actually gathering data. Currently, it is also being shown when Search Console report has zero data, regardless of the "gathering data" situation.
<img width="1499" alt="Screen Shot 2021-12-16 at 9 37 56 AM" src="https://user-images.githubusercontent.com/3531426/146422802-16944877-8d23-48d1-96ad-fdc3f9d7a713.png">

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `SearchFunnel` widget should be adjusted so that it only renders in the "blue zero data box" state if the Search Console report is zero _and_ Search Console `isGatheringData`.
* Otherwise, the widget should be rendered as usual, with the Search Console metrics displaying 0s and empty charts as needed.
* For the Analytics data in the widget, it should work as follows:
    * If Analytics is not active/connected, render the CTA the same as already happens today.
    * Otherwise if the Analytics report is zero _and_ Analytics `isGatheringData`, render the blue `ReportZero` component in there, similar to the Analytics CTA, in place of the Analytics-specific tabs which therefore cannot be accessed.
    * If the Analytics report is zero but Analytics is no longer "gathering data", the metrics should be rendered as usual, like for Search Console, displaying 0s and empty charts as needed.

## Implementation Brief
* Using `assets/js/modules/search-console/components/dashboard/SearchFunnelWidget/Overview.js`,
	* Query the `search-console` and `analytics` data store separately via the `isGatheringData` selector to see if the data is being gathered for each module.
	* For Search Console,
		* If the selector returns `true` and the report data is zero (via the `isZeroReport` function from `assets/js/modules/search-console/util/is-zero-report.js`), render `WidgetReportZero`, with `search-console` as value for the `moduleSlug` prop.
		* The `WidgetReportZero` component should be rendered in place of the first 2 data blocks (which are Search Console specific), similar to how it's done for Analytics.
	* For Analytics,
		* Check if the analytics module is active and connected, if not, return `false` when using the `isGatheringData` selector.
[...]

PR #4603, opened 2021-12-23
Show Search Funnel widget gathering data CTA appropriately

hussain-t:
## Summary

Addresses issue #4571 

## Checklist


tofumatt (changes requested):
The Search Funnel widget in Storybook looks wrong here, it seems like the logic for displaying the cells allows for too many to be shown. See here: http://localhost:9001/?path=/story/modules-searchconsole-widgets-searchfunnelwidget--search-console-gathering-data

CleanShot 2022-01-04 at 23 49 26

It happens when both Search Console & Analytics are gathering data:
CleanShot 2022-01-04 at 23 50 46

Can you check the logic for how these are displayed? Let me know if you want a hand looking through the logic, but from those screenshots/stories it looks like Analytics reports with zero data are still be shown even when "isGatheringData" is `true`, which shouldn't happen.

assets/js/modules/search-console/components/dashboard/SearchFunnelWidget/Overview.js
```diff
@@ -158,33 +174,49 @@ const Overview = ( {
...
-						title={ __( 'Total Impressions', 'google-site-kit' ) }
-						datapoint={ totalImpressions }
-						change={ totalImpressionsChange }
-						changeDataUnit="%"
-						context="button"
-						selected={ selectedStats === 0 }
-						handleStatSelection={ handleStatsSelection }
-					/>
-				</Cell>
+				{ isSearchConsoleGatheringData && ! hasSearchConsoleData && (
```
aaemnnosttv, line 177: This condition can be simplified as it isn't possible for the module to be gathering data and have data. Gathering data means there is no data available yet.
```suggestion
				{ isSearchConsoleGatheringData && (
```

assets/js/modules/search-console/components/dashboard/SearchFunnelWidget/Overview.js
```diff
@@ -158,33 +174,49 @@ const Overview = ( {
...
+					<Cell { ...halfCellProps }>
+						<WidgetReportZero moduleSlug="search-console" />
+					</Cell>
+				) }
+
+				{ ! isSearchConsoleGatheringData && hasSearchConsoleData && (
```
aaemnnosttv, line 183: The `DataBlock` below should be shown so long as it is not gathering data, even if it is  zero data. It's a bit confusingly named perhaps, but gathering data == WidgetReportZero

assets/js/modules/search-console/components/dashboard/SearchFunnelWidget/Overview.js
```diff
@@ -218,101 +250,92 @@ const Overview = ( {
...
-									) }
-									datapoint={ analyticsVisitorsDatapoint }
-									change={ analyticsVisitorsChange }
-									changeDataUnit="%"
-									context="button"
-									selected={ selectedStats === 2 }
-									handleStatSelection={ handleStatsSelection }
-								/>
-							</Cell>
+				{ isAnalyticsGatheringData && ! hasAnalyticsData && ! error && (
```
aaemnnosttv, line 253: See above
```suggestion
				{ isAnalyticsGatheringData && ! error && (
```

assets/js/modules/search-console/components/dashboard/SearchFunnelWidget/Overview.js
```diff
@@ -218,101 +250,92 @@ const Overview = ( {
...
-											change={ analyticsGoalsChange }
-											changeDataUnit="%"
-											context="button"
-											selected={ selectedStats === 3 }
-											handleStatSelection={
-												handleStatsSelection
-											}
-										/>
-									) }
+				{ ! isAnalyticsGatheringData && hasAnalyticsData && ! error && (
```
aaemnnosttv, line 259:
```suggestion
				{ ! isAnalyticsGatheringData && ! error && (
```

assets/js/modules/search-console/components/dashboard/SearchFunnelWidget/index.stories.js
```diff
@@ -121,6 +121,51 @@ Ready.args = {
...
+			options: searchConsoleArgs,
+		} );
+
+		for ( const options of analyticsArgs ) {
+			registry
+				.dispatch( MODULES_ANALYTICS )
+				.receiveGetReport( [], { options } );
+		}
+	},
+};
```
aaemnnosttv, line 167: This story seems to be a duplicate of `DataUnavailable` below. The 'Data Unavailable' story is one that is a bit of a convention throughout so let's keep that one instead for this case.

assets/js/modules/search-console/components/dashboard/SearchFunnelWidget/index.stories.js
```diff
@@ -121,6 +121,51 @@ Ready.args = {
...
+
```
aaemnnosttv, line 168: Would you please add a new story showing a non-gathering data zero data state? In the past gathering data and zero data were treated the same so we want to see these separately now 👍

aaemnnosttv (changes requested):
Thanks @hussain-t – this is almost ready to go, just a few things to address before we can merge it 👍  Let me know if you have any questions.

assets/js/modules/search-console/components/dashboard/SearchFunnelWidget/Overview.js
hussain-t: What about?
```suggestion
				{ hasAnalyticsData && ! error && (
```

Since we are checking `isSearchConsoleGatheringData === false` in `hasAnalyticsData`.

aaemnnosttv: Right, but we still want to render this expression if there is no data as mentioned before. Zero data is okay here but it shouldn't be gathering.

aaemnnosttv (changes requested):
Thanks @hussain-t, this is almost there 🙂 

As mentioned before, there were some commits added here from develop that are for the next release, so let's recreate this PR without those.

Also, there is at least one tweak still needed in the logic for what is shown regarding the Analytics elements as it is possible for too many things to be showing as you can see here:

image

I also reproduced this on a site 
image

Ideally it would be great to get a few more stories here but we'll have to do those in a follow-up.

hussain-t:
Closing this PR since it has the develop branch changes however, we need to merge it to the main branch. A new [PR](https://github.com/google/site-kit-wp/pull/4639) is created instead of this.

PR #4616, opened 2022-01-04
Enhancement/4428 - New Dashboard Entity Header 

jimmymadon:
## Summary

Addresses issue:

- #4428

## Checklist

### DIFF
--- a/assets/js/components/EntityHeader.js
+++ b/assets/js/components/EntityHeader.js
@@ -46,6 +46,7 @@ import { CORE_SITE } from '../googlesitekit/datastore/site/constants';
 import BackspaceIcon from '../../svg/keyboard-backspace.svg';
 import { CORE_LOCATION } from '../googlesitekit/datastore/location/constants';
 import Link from './Link';
+import { normalizeURL } from '../util/urls';
 const { useSelect, useDispatch } = Data;
 
 const EntityHeader = () => {
@@ -79,12 +80,17 @@ const EntityHeader = () => {
 			const maxChars = ( availableWidth * 2 ) / fontSize;
 
 			if ( maxChars < entityURL.length ) {
-				const extraChars = entityURL.length - maxChars + 4; // 4 is the length of "/...".
 				const shortenedURL = new URL( entityURL );
+				// We will further shorten the URL below by removing protocol/www (normalizing).
+				const extraChars =
+					normalizeURL( shortenedURL.toString() ).length -
+					maxChars +
+					4; // 4 is the length of "/...".
 				const origin = shortenedURL.origin;
 				const restOfURL = shortenedURL.toString().replace( origin, '' );
 				const newRestOfURL = '/...' + restOfURL.substr( extraChars );
-				setURL( origin + newRestOfURL );
+
+				setURL( normalizeURL( origin ) + newRestOfURL );
 			} else {
 				setURL( entityURL );
 			}
--- a/assets/js/components/EntityHeader.js
+++ b/assets/js/components/EntityHeader.js
@@ -138,7 +138,12 @@ const EntityHeader = () => {
 				className="googlesitekit-entity-header__details"
 			>
 				<p>{ currentEntityTitle }</p>
-				<Link href={ entityURL } external inherit>
+				<Link
+					href={ entityURL }
+					aria-label={ entityURL }
+					external
+					inherit
+				>
 					{ url }
 				</Link>
 			</div>
--- a/assets/js/components/EntityHeader.js
+++ b/assets/js/components/EntityHeader.js
@@ -46,7 +46,7 @@ import { CORE_SITE } from '../googlesitekit/datastore/site/constants';
 import BackspaceIcon from '../../svg/keyboard-backspace.svg';
 import { CORE_LOCATION } from '../googlesitekit/datastore/location/constants';
 import Link from './Link';
-import { normalizeURL } from '../util/urls';
+import { shortenURL } from '../util/urls';
 const { useSelect, useDispatch } = Data;
 
 const EntityHeader = () => {
@@ -62,7 +62,7 @@ const EntityHeader = () => {
 	const [ url, setURL ] = useState( entityURL );
 
 	useEffect( () => {
-		const shortenURL = () => {
+		const shortenEntityURL = () => {
 			if ( ! headerDetailsRef.current ) {
 				return;
 			}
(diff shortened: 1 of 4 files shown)


assets/js/components/EntityHeader.js
```diff
@@ -79,12 +80,17 @@ const EntityHeader = () => {
...
+				const extraChars =
+					normalizeURL( shortenedURL.toString() ).length -
+					maxChars +
+					4; // 4 is the length of "/...".
 				const origin = shortenedURL.origin;
 				const restOfURL = shortenedURL.toString().replace( origin, '' );
 				const newRestOfURL = '/...' + restOfURL.substr( extraChars );
-				setURL( origin + newRestOfURL );
+
+				setURL( normalizeURL( origin ) + newRestOfURL );
```
felixarntz, line 93: I think it would be cleaner to move this into a utility function like `shortenURL( url, maxChars )`, including Jest tests for it.

Also, pending the response on https://github.com/google/site-kit-wp/issues/4428#issuecomment-1010995527, we may want to strip the protocol and domain entirely at first and only shorten further if it then _still_ doesn't fit.

assets/sass/components/global/_googlesitekit-entity-header.scss
```diff
@@ -93,33 +99,28 @@
 
 		p {
 			font-family: $f-secondary;
-			font-size: 11px;
-			line-height: 14px;
 			margin: 0;
 			overflow: hidden;
 			text-overflow: ellipsis;
 
+			@include responsive-text("body3", "body3");
```
felixarntz, line 106: Not a problem per-se, but if we agree on https://github.com/google/site-kit-wp/issues/4428#issuecomment-1010995527, this would need to also use `body2` for the second value, similar to the CSS for the "Back to dashboard" button above.

felixarntz (changes requested):

assets/js/util/urls.js
```diff
@@ -85,3 +85,27 @@ export function normalizeURL( incomingURL ) {
...
+	}
+
+	const urlObject = new URL( url );
+	const shortenedURL = url.replace( urlObject.origin, '' );
+	if ( shortenedURL.length < maxChars ) {
+		return shortenedURL;
+	}
+
+	const extraChars = shortenedURL.length - Math.floor( maxChars ) + 3; // 3 is the length of "...".
+	return '...' + shortenedURL.substr( extraChars );
```
tofumatt, line 110: Arguably this should at least be an actual ellipsis:

```suggestion
	return '…' + shortenedURL.substr( extraChars );
```

But it's worth noting that this might not localise extremely well. Given it's a URL the RTL considerations probably don't matter too much, but what if there are locales where we don't trail off with "…"? That said, a quick web search shows this StackOverflow question that indicates a lot of locales _will_ use ellipsis, and WordPress itself seems to include it and it gets kept in the localisation: https://stackoverflow.com/questions/48718469/how-is-ellipsis-localized

Seems fine, but we should use the actual ellipsis character over three full-stops I think. 🙂

assets/js/util/urls.test.js
```diff
@@ -123,3 +129,39 @@ describe( 'isHashOnly', () => {
...
+	it.each( [
+		[
+			'https://www.very-long-domain-name.com/some-directory/some-file',
+			30,
+			'/some-directory/some-file',
+		],
+		[
+			'https://www.domain.com/some-directory/some-directory/some-file',
+			30,
+			'...ry/some-directory/some-file',
```
tofumatt, line 149: These tests will need updating for the new ellipsis as well.

assets/js/util/urls.js
```diff
@@ -85,3 +85,27 @@ export function normalizeURL( incomingURL ) {
...
+		return url;
+	}
+
+	const urlObject = new URL( url );
+	const shortenedURL = url.replace( urlObject.origin, '' );
+	if ( shortenedURL.length < maxChars ) {
+		return shortenedURL;
+	}
+
+	const extraChars = shortenedURL.length - Math.floor( maxChars ) + 3; // 3 is the length of "...".
```
tofumatt, line 109: Seems like this might need to be changed to reflect the actual "ellipsis" character here?

tofumatt (changes requested):
The code here looks good, but I think we should be using the actual ellipsis character as suggested instead of three separate full-stops.

Looks like this was based off `main`, which is great, because I was worried we'd need to do some git-foo to not have to make a new PR. But I changed the target branch from `develop` to `main` and it worked. 👍🏻 

One other thing to note that was missed in the original review:
CleanShot 2022-01-12 at 21 40 48

An `aria-label={ entityURL }` prop should be supplied to the `Link` component. It will take care of adding the "opens in a new tab" suffix, as shown here: https://github.com/google/site-kit-wp/blob/91d02933ac30e7f40ecff857b2a4cc6bae3a1156/assets/js/components/Link.js#L101

But as-is the `aria-label` has an ellipsis added to it because it uses the `{ url }` supplied as the `children` prop to create the `aria-label`, which is confusing and not necessary.

assets/js/util/urls.js
```diff
@@ -85,3 +85,27 @@ export function normalizeURL( incomingURL ) {
...
+ * @param {string} url      The original URL to shorten.
+ * @param {number} maxChars The maximum length of the URL.
+ * @return {string} The shortened URL.
+ */
+export function shortenURL( url, maxChars ) {
+	if ( url.length <= maxChars ) {
+		return url;
+	}
+
+	const urlObject = new URL( url );
```
aaemnnosttv, line 103: `new URL` will throw if an invalid URL is provided to it which could happen if a selector that provides a valid URL hasn't resolved yet for example.

Let's wrap this in a try/catch or add a guard using `isURL` from `@wordpress/url` (essentially the same thing in a utility).

jimmymadon:
@tofumatt Thanks for the feedback. Have implemented your requested changes. Just a note on the aria-labels behaviour:

1. If I do not add the `aria-label` explicitly, it uses the child {url} like you mentioned, which is the shortened URL that is displayed on the page. It doesn't truncate the aria-label further when I tested this using long URLs. So not sure why it adds its own ellipsis in your screenshot.
Screenshot 2022-01-13 at 10 57 09
Screenshot 2022-01-13 at 10 02 35

2. If I explicitly set `aria-label={entityURL}`, it will have the full URL in the aria-label but the truncated form is displayed on the page.
Screenshot 2022-01-13 at 10 56 34

I have explicitly added the aria-label for now which does behaviour 2 for me above.

tofumatt (approved):
Looks good and works great for me! 👍🏻

Issue #4502, opened 2021-12-08
Box shadow animation when scrolling top of the page


eclarke1:
## Bug Description

When scrolling to the top of the main dashboard, the box shadow animation is weird, with 2 borders showing up momentarily.

https://user-images.githubusercontent.com/1621608/146362433-86d77e36-df43-4847-ab89-2252ed31ee9c.mp4

A quick inspection shows the problem is related to the added space here which is only shown when scrolled to the top due to the sticky position of the header. https://github.com/google/site-kit-wp/blob/85a5e02c0cabf732b651a2fd5ebdff6c113e28e7/assets/sass/components/global/_googlesitekit-header.scss#L173

https://app.asana.com/0/1201438453642602/1201480648448644

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The Site Kit header should not appear to change in height between scrolled at the top-most position and below in the context of the unified dashboard. It also should be ensured that there is no visual change in the box-shadow at the moment of the "transition" between regular and sticky header.

## Implementation Brief

- Remove the `shadow` mixin from `.googlesitekit-subheader` and the `min-height` property.
- Add a new pseudo element at `.googlesitekit-navigation`, specifically `before` and make sure is visible using the full width of the container.
  - Set the position to `absolute`
  - Add the `shadow` mixin
  - Set the height to `3px`.
  - Set the `top` position to `-3px`.

### Test Coverage

* 

## QA Brief
* Enable the `unifiedDashboard` feature flag.
* Scroll up and down on the main dashboard.
* There should be no weird behaviors with the box-shadow for the header.

## Changelog entry

* N/A

eclarke1:
@felixarntz @aaemnnosttv this was one of our 'Nice to have' for launch issues, which is sitting in Triage right now. Any chance we could move this over to AC?

aaemnnosttv:
@felixarntz I've added some preliminary AC, let me know what you think as it's a bit odd to define. cc: @asvinb thoughts here?

felixarntz:
@aaemnnosttv Makes sense - looks like it's indeed just a height change and not actually a change in the box shadow itself. I still added a note that there should not be any notable box shadow style change.

eclarke1:
@mitogh please could you kindly add an estimate here, considering engineering, CR and QA? Thanks

mitogh:
Added an estimate @eclarke1

aaemnnosttv:
[...]

PR #4621, opened 2022-01-06
Add box shadow to navigation.

asvinb:
## Summary

Addresses issue:

- #4502 

## Relevant technical choices
* Something that was missed in the IB and IB review was having 2 box-shadows visible when scrolling the page. A fix has been pushed where we detect when the navigation is sticky to hide one of the box-shadows

## Checklist


aaemnnosttv (commented):
Thanks @asvinb ! This looks good for the most part, there is just a bit of a weird pulse that I'm seeing when transitioning between sticky and "stuck. I'm guessing this is due to some overlap in property transitions. Is there anything we can do to avoid this? It looks like maybe the transition durations may not be in sync or might need some tweaking. What do you think?

https://user-images.githubusercontent.com/1621608/148413195-a35b7feb-a89d-49cf-8675-756995589c3d.mp4

Alternatively, since this "stuck" view is only activated once the window is scrolled, is there any way to trick the element to always be "stuck"? That way we'd only ever need to deal with the one shadow. Just a thought. What do you think?

asvinb:
Thanks @aaemnnosttv . I reworked the logic a little bit where I added a check to know whether we have a subheader or not. I think UI wise it's much better. While working on this, I got a small issue with those lines: https://github.com/google/site-kit-wp/pull/4621/files#diff-8e2e443764d5f50185dd0c6ff11d41ad4a0b3cbaed0a1bee306bdb6a494b480cR53-R57 Basically if I try to get the child nodes count without the timeout, it returns `0` which is weird. Do you happen to know what might be the reason?

aaemnnosttv:
> I reworked the logic a little bit where I added a check to know whether we have a subheader or not. I think UI wise it's much better.

Thanks @asvinb – in the browser this looks great! I (somehow) noticed that there is a quirk at 960px though where there's a gap in the header

image



> While working on this, I got a small issue with those lines: [#4621 (files)](https://github.com/google/site-kit-wp/pull/4621/files#diff-8e2e443764d5f50185dd0c6ff11d41ad4a0b3cbaed0a1bee306bdb6a494b480cR53-R57) Basically if I try to get the child nodes count without the timeout, it returns `0` which is weird. Do you happen to know what might be the reason?

That's a bit odd, maybe those nodes don't exist yet when the ref is called? Maybe @tofumatt may know. Why can't we just use the presence of the `subHeader` prop instead? If a timeout there really is needed, does it really need to be 1s? That seems quite long, I would think we may even be able to get away with no value there (i.e. fastest possible timeout). If it really requires a longer amount of time I would be concerned that this might vary too much by browser leading to a potential race condition.

aaemnnosttv:
@asvinb looks like Jest is failing now, can you take a look?
https://github.com/google/site-kit-wp/runs/4759168698?check_suite_focus=true#step:10:4315

asvinb:
Yep @aaemnnosttv Looking at it right now

asvinb:
@aaemnnosttv Failing tests have been updated. Looks like when the delay is set to `0`, Jest complains about it. Have you come across this before?

assets/js/components/Header.js
```diff
@@ -48,12 +48,19 @@ const Header = ( { children, subHeader, showNavigation } ) => {
 		select( CORE_USER ).isAuthenticated()
 	);
 	const { y } = useWindowScroll();
+	const [ hasSubheader, setHasSubheader ] = useState( false );
+
+	const subHeaderElementRef = useCallback( ( node ) => {
+		setTimeout( () => {
+			setHasSubheader( node?.childNodes.length );
+		}, 1 ); // technically should be 0 but for failing Jest tests.
+	}, [] );
```
aaemnnosttv, line 57: After thinking about this a bit more and doing some testing, I found that this approach is a bit too simple to cover our needs as we can't rely on the child nodes being there as soon as the ref is set, or even 1sec later. Some may come from remote requests, and it's possible that the notification may appear for a split second only to be hidden/removed.

The key point here is that we need a solution that will update when children are added and removed, whenever that happens, not just once.

I found a solution that seems to work well using mutation observer and added a few tests, so I'll put together a quick PR for your review on top of this one.

asvinb: 👍

assets/js/components/DashboardNavigation.js
```diff
@@ -63,6 +64,9 @@ const { useSelect } = Data;
 
 export default function DashboardNavigation() {
 	const dashboardType = useDashboardType();
+	const elementRef = useRef();
+	const { y } = useWindowScroll();
```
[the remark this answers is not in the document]
aaemnnosttv: Thoughts @eugene-manuilov ?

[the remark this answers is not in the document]
eugene-manuilov: @aaemnnosttv @asvinb We already use this same approach in 2 components: 
https://github.com/google/site-kit-wp/blob/bcc13f819025c14da699050fb432eff34b8c0d9a/assets/js/components/EntityHeader.js#L105 https://github.com/google/site-kit-wp/blob/037120d66e456e274b1677527295bd08a3789ea1/assets/js/components/Header.js#L50

Maybe it worth moving this functionality to one of parent components to avoid having it in multiple places? In this case that parent component will add a CSS class (i.e. `googlesitekit-plugin--scrolled`) to the `<div class="googlesitekit-plugin">` element that we can use in CSS to activate certain styles for required elements. 

```scss
.googlesitekit-entity-header {
    .googlesitekit-plugin--scrolled & {
        ...
    }
}
```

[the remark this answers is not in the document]
asvinb: Thanks @eugene-manuilov . I've simplified the approach for `EntityHeader` based on the latest updates for `DashboardNavigation` and not using the `useWindowScroll` hook within it. Can you review and let me know what you think?

assets/js/components/Header.js
```diff
@@ -48,12 +49,16 @@ const Header = ( { children, subHeader, showNavigation } ) => {
 		select( CORE_USER ).isAuthenticated()
 	);
 	const { y } = useWindowScroll();
```
aaemnnosttv, line 51: @asvinb you removed this from the `EntityHeader` but is still used here. Is it still needed here?

aaemnnosttv (commented):
Thanks @asvinb – this LGTM, just one question.

assets/js/components/Header.js
asvinb: @aaemnnosttv Good question. Actually I think we can rely on the scroll event in `DashboardNavigation` to toggle the `googlesitekit-header--has-scrolled` class to the `Header` component and thus we no longer need to use the `useWindowScroll` hook. *However* I'am not too comfortable with this approach since `DashboardNavigation` is kinda updating the `Header` component whereas I would prefer to have each component changing its own markup/class names. What do you think about that?

aaemnnosttv: Thanks @asvinb – I agree, we shouldn't have components mutating the DOM rendered by another component.

I agree with @eugene-manuilov's idea above that we should be able to centralize this (or most of it) perhaps into a common parent? Thoughts @tofumatt ?

tofumatt: 100% agree we shouldn't be updating other component's DOM. 👍🏻 

Eugene's suggestion above https://github.com/google/site-kit-wp/pull/4621#discussion_r784785585 makes sense; I'm not sure if the hook is smart enough to combine listeners for performance, but if all we're doing is wanting to know if a scroll happened somewhere, either:

1. a custom hook (`useHasScrolled`)
2. Eugene's CSS approach

would be best. I prefer the CSS one as it's more flexible (it will apply to any children on the page)/less duplicated code.

tofumatt (commented):
Left a comment about the `useWindowScroll` as requested by @aaemnnosttv 🙂

assets/js/components/Header.js
aaemnnosttv: Thanks @tofumatt – `useWindowScroll` uses `useRafState` internally which does not share listeners from what I can see https://github.com/streamich/react-use/blob/3685b7502a4d1980953deed11c3dee31b220c89d/src/useRafState.ts?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D

aaemnnosttv:
@asvinb back to you to resolve the merge conflicts and revise the scroll related behavior. See https://github.com/google/site-kit-wp/pull/4621#discussion_r788131754

assets/js/components/Header.js
asvinb: @tofumatt @aaemnnosttv Just so that we're clear, even if we have the scroll logic to toggle a class name in the `Dashboard` component or `Root`, we will still need to have another scroll listener in `DashboardNavigation` for the chips to be active and also to know when the navigation becomes sticky. 
I can certainly see the benefits of moving the `--has-scrolled` class higher up the DOM since we can easily add any styles to any components.
Let me know if you're still on board with that approach.

tofumatt: Fair enough, I guess the `DashboardNavigation` is a bit more specialised.

To me, that's fine, shame that `useWindowScroll` is creating multiple listeners but if we move the more generic "has scrolled" behaviour to `Root` then we at least won't make more for individual components in the future.

Let's still go with that approach, I'm on-board 🙂

asvinb: @aaemnnosttv After talking to @tofumatt we were planning to add the scroll logic in the `Root` component where the `googlesitekit-plugin--has-scrolled` class would have been added to `body`. However this would have resulted in unnecessarily adding scroll event listeners to pages where we won't ever use them, for e.g the settings page. Thus the best approach is going with the `useHasScrolled` hook where we just add the hook to the `DashboardMainApp` and `DashboardEntityApp` where we will be using them. Over for you to review 😁

assets/js/hooks/useHasScrolled.test.js
```diff
@@ -0,0 +1,54 @@
...
+			0,
+			false,
+		],
+		,
+	] )( '%s', async ( _, args, expected ) => {
+		let result;
+
+		await act( async () => {
+			( { result } = await renderHook( () => {
+				Object.defineProperty( global.window, 'pageYOffset', {
```
aaemnnosttv, line 43: I'm pretty sure we don't need to use window here since `global` is basically an alias for `window` but double check first 😄 
```suggestion
				Object.defineProperty( global, 'pageYOffset', {
```

assets/js/hooks/useHasScrolled.js
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Returns whether the user has scrolled the page and adds/remove the googlesitekit-plugin--has-scrolled class to the body.
+ *
+ * @since n.e.x.t
+ *
+ * @return {boolean} `true` if the user has scrolled the page, `false` otherwise.
+ */
+export const useHasScrolled = () => {
+	const { y } = useWindowScroll();
```
aaemnnosttv, line 32: Looks like we shouldn't use this hook, especially at a high-level as it causes any consuming components to be rerendered every time `y` changes, even when using `useEffect` as mentioned here and rerendering the dashboard apps adds an additional 30-40ms in the profiler for me.

image

Even an alternative using a listener on `scroll` is more performant, but the rerender of the whole dashboard is something we should avoid.

This gives more control over when a rerender will happen but still had the same expense in terms of the rerendering of the dashboard, just less often.

I found that putting this hook in an empty component that is rendered in the dashboard solved this problem although I would imagine there's probably a more idomatic way to do this in React to avoid triggering unnecessary rerenders in the consuming component. cc: @tofumatt ?

image
image

image

aaemnnosttv (changes requested):
Thanks @asvinb – I think we're really close here but we just need to iron out the last details. See comments below.

assets/js/hooks/useHasScrolled.js
tofumatt: In my experience that's the idiomatic way, hah! In this case we have a component with side-effects and placing it on its own and outside of other component structures is the way-to-go. I've seen similar things done with stuff like `<SetupRequests />` or `<GetUserProfile />` components before.

It's a bit _odd_ to see but it's totally fine and I think makes sense. That'd be the approach I would've taken here as well, so I say let's go with that approach 👍🏻

assets/js/components/DashboardEntityApp.js
```diff
@@ -55,9 +55,13 @@ import { Cell, Grid, Row } from '../material-components';
 import PageHeader from './PageHeader';
 import Layout from './layout/Layout';
 import { CORE_WIDGETS } from '../googlesitekit/widgets/datastore/constants';
+import { useHasScrolledEffect } from '../hooks/useHasScrolledEffect';
+import ScrollEffect from './ScrollEffect';
 const { useSelect } = Data;
 
 function DashboardEntityApp() {
+	useHasScrolledEffect();
```
aaemnnosttv, line 63: This should only be used in the `ScrollEffect` component, as you have in the `DashboardMainApp` below 👍

assets/js/components/ScrollEffect.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * ScrollEffect component.
+ *
+ * Site Kit by Google, Copyright 2021 Google LLC
```
aaemnnosttv, line 4:
```suggestion
 * Site Kit by Google, Copyright 2022 Google LLC
```

assets/js/hooks/useHasScrolledEffect.test.js
```diff
@@ -0,0 +1,55 @@
+/**
+ * `useHasScrolled` hook tests.
```
aaemnnosttv, line 2: Please update references to use the updated name.

assets/sass/components/global/_googlesitekit-entity-header.scss
```diff
@@ -44,21 +49,21 @@
...
-		@media (min-width: $bp-tablet) {
-			padding-left: $grid-gap-desktop * 1.5;
-			padding-right: $grid-gap-desktop * 1.5;
-		}
 	}
 
-	.googlesitekit-entity-header--has-scrolled {
-		box-shadow: 0 3px 3px rgb(0 0 0 / 10%);
-		transition: box-shadow $t-speed-2;
+	/* stylelint-disable no-descending-specificity */
```
aaemnnosttv, line 58: Is this necessary? I would think we could merge this block with the `.googlesitekit-entity-header` block above, no? 🤔

aaemnnosttv (commented):
Thanks @asvinb – we're super close here, thanks for your patience on this one 😄

assets/js/components/DashboardEntityApp.js
asvinb: 🤦 Forgot to remove that.

assets/js/components/ScrollEffect.js
asvinb: 👍

assets/js/hooks/useHasScrolledEffect.test.js
asvinb: 👍

assets/sass/components/global/_googlesitekit-entity-header.scss
asvinb: @aaemnnosttv Actually it was pretty much impossible to have the desired generated CSS since when using the `&` to reference the parent, it references the compiled parent selector. In order to avoid the "hacky" like stylelint disable, I've moved the styles to the related also `assets/sass/components/global/_googlesitekit-dashboard-navigation.scss` where we don't need the stylelint comments.

aaemnnosttv (approved):
Great! Thanks @asvinb 👍

PR #4624, opened 2022-01-06
Disable CSS animations in VRT

aaemnnosttv:
## Summary

Addresses issue:

- #4623

## Checklist

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -20,7 +20,7 @@ module.exports = {
 	stories: [ '../stories/**/*.stories.js', '../assets/js/**/*.stories.js' ],
 	addons: [ '@storybook/addon-viewport', '@storybook/addon-postcss' ],
 	previewHead: ( head ) => {
-		if ( ! process.env.VRT ) {
+		if ( process.env.VRT !== '1' ) {
 			return head;
 		}
 
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -19,4 +19,20 @@
 module.exports = {
 	stories: [ '../stories/**/*.stories.js', '../assets/js/**/*.stories.js' ],
 	addons: [ '@storybook/addon-viewport', '@storybook/addon-postcss' ],
+	previewHead: ( head ) => {
+		if ( ! process.env.VRT ) {
+			return head;
+		}
+
+		return `${ head }
+<style>
+	#root *,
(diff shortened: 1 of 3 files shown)


.storybook/main.js
```diff
@@ -19,4 +19,20 @@
...
+	previewHead: ( head ) => {
+		if ( ! process.env.VRT ) {
+			return head;
+		}
+
+		return `${ head }
+<style>
+	#root *,
+	#root *:before,
+	#root *:after {
```
tofumatt, line 31: These tend to work, but technically these should be `::before` and `::after`, right? `:before` and `:after` are CSS2 🙂  https://developer.mozilla.org/en-US/docs/Web/CSS/::before

Also, do we want to cover things like `:hover`, just-in-case?

```suggestion
	#root *::before,
	#root *::after,
	#root *:hover,
	#root *:focus,
	#root *:active,
	#root *:visited {
```

.storybook/main.js
```diff
@@ -19,4 +19,20 @@
 module.exports = {
 	stories: [ '../stories/**/*.stories.js', '../assets/js/**/*.stories.js' ],
 	addons: [ '@storybook/addon-viewport', '@storybook/addon-postcss' ],
+	previewHead: ( head ) => {
+		if ( ! process.env.VRT ) {
```
tofumatt, line 23: Not strictly necessary, but this might be a bit more user-friendly as something like:

```suggestion
		if ( ! process.env.VRT?.length && process.env.VRT !== '0' ) {
```

Otherwise won't `VRT= npm run build:storybook` or `VRT=0 npm run build:storybook` cause this to be truthy? That would be a bit unexpected—even though we don't use it now it might be nice to anticipate that usage and handle it here 🙂

tofumatt (changes requested):
Looks good, but maybe we should protect against CSS pseudo-classes having animations too.

And I left a minor note about checking the `VRT` env variable. Then this should be good-to-go. 👍🏻

.storybook/main.js
aaemnnosttv: Thanks! I don't think adding `:hover` is necessary (I tested with this originally) as it should be covered by the base selector, especially since we're using `!important`.

tofumatt (approved):
Nice! 👍🏻

PR #4629, opened 2022-01-10
Update modules to be share aware for dashboard sharing

hussain-t:
## Summary

Addresses issue: #4521 

## Checklist

### DIFF
--- a/includes/Core/Modules/Module.php
+++ b/includes/Core/Modules/Module.php
@@ -826,9 +826,9 @@ abstract class Module {
 	 */
 	public function is_shareable() {
 		if ( $this instanceof Module_With_Owner && $this->is_connected() ) {
-			$definitions = $this->get_datapoint_definitions();
-			foreach ( $definitions as $modules => $details ) {
-				if ( array_key_exists( 'shareable', $details ) && true === $details['shareable'] ) {
+			$datapoint = $this->get_datapoint_definitions();
+			foreach ( $datapoint as $details ) {
+				if ( ! empty( $details['shareable'] ) ) {
 					return true;
 				}
 			}
--- a/includes/Core/Modules/Module.php
+++ b/includes/Core/Modules/Module.php
@@ -826,8 +826,8 @@ abstract class Module {
 	 */
 	public function is_shareable() {
 		if ( $this instanceof Module_With_Owner && $this->is_connected() ) {
-			$datapoint = $this->get_datapoint_definitions();
-			foreach ( $datapoint as $details ) {
+			$datapoints = $this->get_datapoint_definitions();
+			foreach ( $datapoints as $details ) {
 				if ( ! empty( $details['shareable'] ) ) {
 					return true;
 				}
--- a/includes/Core/Modules/Module.php
+++ b/includes/Core/Modules/Module.php
@@ -816,4 +816,24 @@ abstract class Module {
 	public static function is_force_active() {
 		return false;
 	}
+
+	/**
+	 * Checks whether the module is shareable.
+	 *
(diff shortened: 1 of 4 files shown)


includes/Core/Modules/Module.php
```diff
@@ -816,4 +816,24 @@ function( $item ) {
...
+	 * Checks whether the module is shareable.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return bool True if module is shareable, false otherwise.
+	 */
+	public function is_shareable() {
+		if ( $this instanceof Module_With_Owner && $this->is_connected() ) {
+			$definitions = $this->get_datapoint_definitions();
+			foreach ( $definitions as $modules => $details ) {
```
eugene-manuilov, line 830: Keys in the definitions list are a conjunction of HTTP methods and datapoint names like `GET:report`. So calling them `$modules` doesn't make sense. Would better if we use `$datapoint` name because it gives an exact clue of what it is. However we won't need keys in this loop and can just omit it:

```php
foreach ( $definitions as $details ) {
    ...
}
```

includes/Core/Modules/Module.php
```diff
@@ -816,4 +816,24 @@ function( $item ) {
...
+				if ( array_key_exists( 'shareable', $details ) && true === $details['shareable'] ) {
+					return true;
+				}
```
eugene-manuilov, line 833: How about something like this:
```suggestion
				return ! empty( $details['shareable'] );
```

eugene-manuilov (changes requested):
Thanks, @hussain-t. Almost looks good to me. Added two nitpicks. Once they are addressed, it should be good to go.

includes/Core/Modules/Module.php
hussain-t: @eugene-manuilov The unit test for `test_get_shareable_modules` method fails for this approach.
```php
$shareable_active_modules = array_map( 'get_class', $modules->get_shareable_modules() );
```
Since `shareable_active_modules` returns an empty array in the test.

However, having an explicit if condition works 👍 
```php
if ( ! empty( $details['shareable'] ) ) {
	return true;
}
```

eugene-manuilov (changes requested):
Thanks, @hussain-t. This looks good to me. Added one suggestion that I'll apply by myself.

includes/Core/Modules/Module.php
eugene-manuilov: Yes, my suggestion was wrong. Sorry, @hussain-t. :man_facepalming:

tests/phpunit/integration/Core/Modules/ModulesTest.php
```diff
@@ -472,4 +472,19 @@ public function provider_feature_flag_modules() {
...
+	public function test_get_shareable_modules() {
+		$modules = new Modules( new Context( GOOGLESITEKIT_PLUGIN_MAIN_FILE ) );
+
+		$this->enable_feature( 'dashboardSharing' );
+
+		$shareable_active_modules = array_map( 'get_class', $modules->get_shareable_modules() );
+
+		$this->assertEqualSets(
+			array(
+				'search-console' => 'Google\\Site_Kit\\Modules\\Search_Console',
```
eugene-manuilov, line 485: @hussain-t looks like we need to update this test to also include the `Google\Site_Kit\Modules\PageSpeed_Insights` module too: https://github.com/google/site-kit-wp/runs/4779620869?check_suite_focus=true

eugene-manuilov (changes requested):

eugene-manuilov (approved):
LGTM

Issue #4488, opened 2021-12-08
Chip doesn't do anything when in the section


eclarke1:
## Bug Description

As per bug bash tickets:
https://app.asana.com/0/1201438453642602/1201478352201031
https://app.asana.com/0/1201438453642602/1201453960164093
https://app.asana.com/0/1201438453642602/1201479361460033

Upon initial dashboard load, clicking the 'Traffic' chip doesn't do anything because the user is already in the Traffic section. If you scroll down the page after initial load and then click the Traffic chip, it does work. Perhaps the issue here is that the Traffic chip appears like it isn't working because the user is already at that anchor on the page.

Potential suggestion to make the chips active when scrolling to make it clear to users what section they are in on the dashboard.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `DashboardNavigation` component is updated to use a custom implementation of chips instead of using the functionality from the `@material/react-chips` package. The updated component works as following:
  * The selected chip can't be unselected
  * When the user clicks on a non-selected chip, the page scrolls to the correct location, the clicked chip becomes selected instead of the previously selected one, and the current URL hash changes to the newly selected chip anchor.
  * When the user scrolls the page, an appropriate chip becomes selected and the URL hash changes to the new anchor if the top of the corresponding area reaches the upper visible position of the widgets area (below the navigation bar).
    ![sitekit 10uplabs com_wp-admin_admin php_page=googlesitekit-dashboard](https://user-images.githubusercontent.com/873049/145206333-c6cc3085-fac5-43f6-b19e-f9afd679c797.png)
  * When the user has scrolled to the very top of the page and the first area is still behind due to a notification banner, the first chip is selected anyway. In other words, if the first widget area is lower than the navigation chips, the first chip should be selected.

## Implementation Brief

* Edit the `DashboardNavigation` component:
  * Update the `selectedIds` state variable to be a single string instead of an array and rename the variable to be just `selectedId`. Rename the companion `setSelectedIds` function to `setSelectedId`.
  * Update the `handleSelect` callback that should get "hash" from the `.target.id` property of the `event` argument passed to the callback function instead of the `selections`.
  * Replace the `ChipSet` element with a simple `<div>` using the same class name.
  * Update all `Chip` elements:
    * Add `onClick={ handleSelect }` property.
    * Add `selected={ selectedId === ... }` property.
  * Add a new `useEffect` hook that does the following:
[...]

PR #4644, opened 2022-01-12
Bug/4488  fix chips scroll 2

eugene-manuilov:
## Summary

Addresses issue:

- #4488

## Checklist


assets/js/components/DashboardNavigation.js
```diff
@@ -99,76 +99,128 @@ export default function DashboardNavigation() {
...
-			handleSelect={ handleSelect }
-			choice
-		>
+		<div className="googlesitekit-navigation mdc-chip-set">
 			{ showTraffic && (
 				<Chip
 					id={ ANCHOR_ID_TRAFFIC }
 					label={ __( 'Traffic', 'google-site-kit' ) }
 					leadingIcon={ <NavTrafficIcon width="18" height="16" /> }
+					onClick={ handleSelect.bind( null, ANCHOR_ID_TRAFFIC ) }
```
aaemnnosttv, line 204: We should use `useCallback` for these functions as is our convention.

aaemnnosttv (changes requested):
Thanks @hussain-t @eugene-manuilov – this looks good to me, just one minor detail to adjust then this should be good to go.

assets/js/components/DashboardNavigation.js
aaemnnosttv: @hussain-t `handleSelect` now uses `useCallback` but using `bind` here will create a new function on every render. Let's extract an additional callback for each (e.g. `handleSelectTraffic`, `handleSelectContent`) here.

aaemnnosttv (approved):
Great, thanks @hussain-t 👍

assets/js/components/DashboardNavigation.js
```diff
@@ -99,76 +98,129 @@ export default function DashboardNavigation() {
...
-					{},
-					'',
-					removeQueryArgs( global.location.href )
-				);
-			}
-			setSelectedIds( selections );
+		( { target } ) => {
+			const chip = target.classList.contains( 'mdc-chip' )
+				? target
+				: target.parentNode;
```
aaemnnosttv, line 109: This will select the wrong chip if the click happens on an inner element of the svg resulting in Traffic being selected instead.

image

```suggestion
			const chip = target.closest( '.mdc-chip' );
```

This works in my testing ^

aaemnnosttv (changes requested):
Thanks @eugene-manuilov – this is a nice simplification, it just needs a little tweaking to ensure the intended item is always selected.

Issue #1592, opened 2020-05-26
Dashboard search should support URIs that don't resolve to individual posts


ryanwelcher:
Currently, any URIs that are entered to the input must resolve to a related piece of content, or "No results found" are displayed. In the case of directly entering valid URIs that don't resolve such as tag archive or the homepage when there is no Page assigned.

The was discovered while working on #1434 and will be more pronounced with the work done there, as now the View Data button is not enabled unless there is a matched piece of content.

We should be able to submit valid URIs directly as long as they match the site domain.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a new class `Google\Site_Kit\Core\Util\REST_Entity_Search_Controller`, which has a `register` method that adds an additional REST endpoint `core/search/data/entity-search`.
    * The endpoint should essentially be a revised version of the existing `core/search/data/post-search` endpoint, with a few differences, as listed below.
    * Instead of returning an array of post objects, it should return an array of entity objects, with the following fields:
        * `id`
        * `title`
        * `url`
        * `type`
    * If a URL is passed and an entity is found, the result should always be returned, not only if it is a post.
    * If the provided value is not a URL, the behavior should be similar to today, relying on only post search, since WordPress does not have a solid search mechanism for other entities. In that case, the posts found just need to be transformed to the entity format (see `Entity_Factory::create_entity_for_post`).
* The new class should be initialized (and registered) in the `Plugin` main class.
* All code (mostly in JS) which currently relies on `core/search/data/post-search` should be switched to the new endpoint instead, including the updated response format of course.
* The then unued `core/search/data/post-search` endpoint should be removed.

## Implementation Brief

* Create a new file `includes/Core/Util/includes/Core/Util/REST_Entity_Search_Controller.php` with
a class `Google\Site_Kit\Core\Util\REST_Entity_Search_Controller`.
  * Add a `register()` function which adds a `new REST_Route` `core/search/data/entity-search` based on `core/search/data/post-search` from the following:
  https://github.com/google/site-kit-wp/blob/4713411/includes/Core/REST_API/REST_Routes.php#L171-L222

  * The returned value will be an array of entities, see the AC for the format. The entities will be fetched either via `create_entity_for_post()` or from the call to `get_reference_entity_from_url()`: 
  https://github.com/google/site-kit-wp/blob/4713411/includes/Core/Util/Entity_Factory.php#L169-L186

[...]

PR #4646, opened 2022-01-13
New `entity-search` api endpoint. 

kuasha420:
## Summary

Addresses issue:

- #1592 

## Checklist

### DIFF
--- a/includes/Core/Util/Entity_Factory.php
+++ b/includes/Core/Util/Entity_Factory.php
@@ -177,7 +177,7 @@ final class Entity_Factory {
 	 * Creates the entity for a given post object.
 	 *
 	 * @since 1.15.0
-	 * @since n.e.x.t method access modifier changed to public.
+	 * @since n.e.x.t Method access modifier changed to public.
 	 *
 	 * @param WP_Post $post A WordPress post object.
 	 * @return Entity The entity for the post.
--- a/includes/Core/Util/Entity_Factory.php
+++ b/includes/Core/Util/Entity_Factory.php
@@ -177,11 +177,12 @@ final class Entity_Factory {
 	 * Creates the entity for a given post object.
 	 *
 	 * @since 1.15.0
+	 * @since n.e.x.t method access modifier changed to public.
 	 *
 	 * @param WP_Post $post A WordPress post object.
 	 * @return Entity The entity for the post.
 	 */
-	private static function create_entity_for_post( WP_Post $post ) {
+	public static function create_entity_for_post( WP_Post $post ) {
 		return new Entity(
 			get_permalink( $post ),
 			array(
--- a/includes/Core/Util/REST_Entity_Search_Controller.php
+++ b/includes/Core/Util/REST_Entity_Search_Controller.php
@@ -108,14 +108,7 @@ class REST_Entity_Search_Controller {
 								if ( ! empty( $posts ) ) {
 									$entities = array_map(
 										function( $post ) {
-											$entity = new Entity(
-												get_permalink( $post ),
-												array(
-													'type' => 'post',
-													'title' => $post->post_title,
-													'id'   => $post->ID,
-												)
-											);
+											$entity = Entity_Factory::create_entity_for_post( $post );
 											return array(
 												'id'    => $entity->get_id(),
 												'title' => $entity->get_title(),
--- a/includes/Core/Util/REST_Entity_Search_Controller.php
+++ b/includes/Core/Util/REST_Entity_Search_Controller.php
@@ -18,7 +18,7 @@ use WP_REST_Request;
 use WP_REST_Response;
 
 /**
- * Class for handling entity search rest routes.
+ * Class for handling entity search REST routes.
  *
  * @since n.e.x.t
  * @access private
--- a/includes/Core/Util/REST_Entity_Search_Controller.php
+++ b/includes/Core/Util/REST_Entity_Search_Controller.php
@@ -3,7 +3,7 @@
  * Class Google\Site_Kit\Core\Util\REST_Entity_Search_Controller
  *
  * @package   Google\Site_Kit\Core\Util
- * @copyright 2021 Google LLC
+ * @copyright 2022 Google LLC
  * @license   https://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
  * @link      https://sitekit.withgoogle.com
  */
--- a/includes/Core/Util/REST_Entity_Search_Controller.php
+++ b/includes/Core/Util/REST_Entity_Search_Controller.php
@@ -1,0 +1,148 @@
+<?php
+/**
+ * Class Google\Site_Kit\Core\Util\REST_Entity_Search_Controller
+ *
+ * @package   Google\Site_Kit\Core\Util
+ * @copyright 2021 Google LLC
+ * @license   https://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
+ * @link      https://sitekit.withgoogle.com
+ */
+
+namespace Google\Site_Kit\Core\Util;
+
+use Google\Site_Kit\Context;
+use Google\Site_Kit\Core\Permissions\Permissions;
+use Google\Site_Kit\Core\REST_API\REST_Route;
+use WP_REST_Server;
+use WP_REST_Request;
+use WP_REST_Response;
+
+/**
+ * Class for handling entity search rest routes.
+ *
+ * @since n.e.x.t
+ * @access private
+ * @ignore
+ */
+class REST_Entity_Search_Controller {
+
+	/**
+	 * Plugin context.
+	 *
+	 * @since n.e.x.t
+	 * @var Context
+	 */
+	private $context;
+
+	/**
+	 * Constructor.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Context $context        Plugin context.
+	 */
+	public function __construct( Context $context ) {
+		$this->context = $context;
+	}
+
+	/**
+	 * Registers functionality through WordPress hooks.
+	 *
+	 * @since n.e.x.t
+	 */
+	public function register() {
+		add_filter(
+			'googlesitekit_rest_routes',
+			function ( $routes ) {
+				return array_merge( $routes, $this->get_rest_routes() );
+			}
+		);
+	}
+
+	/**
+	 * Gets REST route instances.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return REST_Route[] List of REST_Route objects.
+	 */
+	protected function get_rest_routes() {
+		$can_authenticate = function() {
+			return current_user_can( Permissions::AUTHENTICATE );
+		};
+
+		return array(
+			new REST_Route(
+				'core/search/data/entity-search',
+				array(
+					array(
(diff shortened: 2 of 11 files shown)


includes/Core/Util/REST_Entity_Search_Controller.php
```diff
@@ -0,0 +1,148 @@
...
+									'update_post_term_cache' => false,
+									'post_status'     => array( 'publish' ),
+								);
+
+								$posts = ( new \WP_Query( $args ) )->posts;
+
+								if ( ! empty( $posts ) ) {
+									$entities = array_map(
+										function( $post ) {
+											$entity = new Entity(
```
techanvil, line 111: Given the IB mentions using [`create_entity_for_post`](https://github.com/google/site-kit-wp/blob/4713411/includes/Core/Util/Entity_Factory.php#L169-L186), it looks like we should change the access modifier for `create_entity_for_post` to `public` and then use it here rather than instantiating `Entity` directly. I.e.

```php
$entity = Entity_Factory::create_entity_for_post( $post );
```

tests/phpunit/integration/Core/Util/REST_Entity_Search_ControllerTest.php
```diff
@@ -0,0 +1,100 @@
...
+namespace Google\Site_Kit\Tests\Core\Util;
+
+use Google\Site_Kit\Context;
+use Google\Site_Kit\Core\REST_API\REST_Routes;
+use Google\Site_Kit\Core\Util\REST_Entity_Search_Controller;
+use Google\Site_Kit\Tests\TestCase;
+use WP_REST_Request;
+use WP_REST_Server;
+
+class REST_Entity_Search_ControllerTest extends TestCase {
```
techanvil, line 20: It would be good to see test(s) which verify the behaviour of the endpoint, i.e. check the correct entities are returned for a search.

techanvil (changes requested):

includes/Core/Util/REST_Entity_Search_Controller.php
kuasha420: Thanks for spotting it, I initially wanted to do it, but if felt drastic to change access modifiers. Let's hope changing it doesn't break anything. :pray:

tests/phpunit/integration/Core/Util/REST_Entity_Search_ControllerTest.php
kuasha420: Added two more tests. Cheers.

kuasha420:
@techanvil applied your suggestions and rebased from `develop`. Ready for another look.

tests/phpunit/integration/Core/Util/REST_Entity_Search_ControllerTest.php
```diff
@@ -0,0 +1,157 @@
...
+			)
+		);
+		$response = rest_get_server()->dispatch( $request );
+
+		$this->assertNotEquals( 200, $response->get_status() );
+		$this->assertArrayHasKey( 'code', $response->get_data() );
+		$this->assertEquals( 'rest_forbidden', $response->get_data()['code'] );
+	}
+
+
```
techanvil, line 67: Please remove these two extra blank lines.

techanvil (changes requested):

tests/phpunit/integration/Core/Util/REST_Entity_Search_ControllerTest.php
kuasha420: As much as I love the idea, I feel like this is overkill as at that point we will be testing WP's core search functionality more than our API endpoint and see very little value in it. The main thing to test here is whether it returns post for general keyword and entity for direct links as that's our implementation. Isn't it right?

kuasha420: I have fixed the empty line and added more assertions to make sure it returns the correct entity. What do you think?

techanvil: > As much as I love the idea, I feel like this is overkill as at that point we will be testing WP's core search functionality more than our API endpoint and see very little value in it. The main thing to test here is whether it returns post for general keyword and entity for direct links as that's our implementation. Isn't it right?

Not really, IMHO. As things stand we are not verifying that we are even doing a search - the backend might simply be returning all of the posts. The integration test is not aware that the underlying implementation is using the WP search API. It should treat the code it's testing as a bit of a black box and simply verify the desired behaviour.

kuasha420: re. testing core wp functionality: maybe someday wp search will become too smart and consider hello and howdy similar word and return howdy results as well for hello keyword, so that will break the test although It's not really broken.

techanvil: Again, I don't think this is a case where we should be considering what is core WP functionality. We should just be writing a test that verifies the endpoint behaviour. Besides, if the underlying search behaviour were to change like it, it would be a breaking change and as such I'd expect to see some test failures.

techanvil: Maybe we should ask for a second opinion here?

includes/Core/Util/REST_Entity_Search_Controller.php
```diff
@@ -0,0 +1,141 @@
+<?php
+/**
+ * Class Google\Site_Kit\Core\Util\REST_Entity_Search_Controller
+ *
+ * @package   Google\Site_Kit\Core\Util
+ * @copyright 2021 Google LLC
```
eugene-manuilov, line 6:
```suggestion
 * @copyright 2022 Google LLC
```

tests/phpunit/integration/Core/Util/REST_Entity_Search_ControllerTest.php
```diff
@@ -0,0 +1,158 @@
+<?php
+/**
+ * REST_Entity_Search_ControllerTest
+ *
+ * @package   Google\Site_Kit\Tests\Core\Util
+ * @copyright 2021 Google LLC
```
eugene-manuilov, line 6:
```suggestion
 * @copyright 2022 Google LLC
```

includes/Core/Util/REST_Entity_Search_Controller.php
```diff
@@ -0,0 +1,141 @@
...
+								}
+							} else {
+								$args = array(
+									'posts_per_page'  => 10,
+									'google-site-kit' => 1,
+									's'               => $query,
+									'no_found_rows'   => true,
+									'update_post_meta_cache' => false,
+									'update_post_term_cache' => false,
+									'post_status'     => array( 'publish' ),
```
eugene-manuilov, line 103: @kuasha420, could you, please, fix the indentation?

tests/phpunit/integration/Core/Util/REST_Entity_Search_ControllerTest.php
eugene-manuilov: > How about, rather than creating these posts/term inline in the tests, create them all in setUpBeforeClass.

IMO, that's fine to keep those posts and terms creation in test methods since they are needed only there. If we move them to the `setUpBeforeClass` method, all those posts and terms will be created for each test in this file, which is redundant.

> Also add a couple more which won't be matched.

Yes, it would be nice to have at least one more entity that doesn't match to make sure we get the correct number of entities in the response.

> we will be testing WP's core search functionality more than our API endpoint and see very little value in it.

Although we use core functions in the handler, we don't test WP's core search functionality here. As Tom said, we should treat our handler as a black box and verify that it returns the correct results for a certain request. In this case, the correct result is the number of returned entities and the fact that every entity in the returned response contains correct properties (`id`/`url`/`title`/`type`).

> maybe someday wp search will become too smart and consider hello and howdy similar word and return howdy results as well for hello keyword, so that will break the test although It's not really broken.

That's absolutely fine and the reason why we add these tests. If something changes in the core and breaks our tests, then we will know about it immediately and will be able to react to those changes in the core.

techanvil: Thanks @eugene-manuilov, appreciate you chipping in. For me, the key thing here is to make the tests more robust, which adding some extra post/terms and making the assertions more explicit will cover. The use of `setUpBeforeClass` is more of an organisational issue and as such I don't have such a strong preference either way.

@kuasha420, based on the above, are you happy to make changes to add an extra post/term and also verify the entity data in the response?

includes/Core/Util/REST_Entity_Search_Controller.php
kuasha420: @eugene-manuilov Apparently this is the right indentation according to phpcs. Very odd indeed. 

Screenshot_20220127_182940

kuasha420:
Thank you both @eugene-manuilov @techanvil ! I have updated the PR accordingly.

tests/phpunit/integration/Core/Util/REST_Entity_Search_ControllerTest.php
```diff
@@ -0,0 +1,172 @@
...
+			)
+		);
+
+		$response = rest_get_server()->dispatch( $request );
+		$data     = $response->get_data();
+
+		$this->assertEquals( 200, $response->get_status() );
+		$this->assertEquals( 2, count( $data ) );
+		$this->assertEquals( 'post', $data[0]['type'] );
+		$this->assertArrayHasKey( 'id', $data[0] );
```
techanvil, line 138: Hey @kuasha420, thanks for making this change. However, this test is still not asserting that the correct posts have been returned. Could you please either take an approach similar to the `test_entity_search` where you verify the ID matches, or otherwise test on the title?

techanvil (changes requested):

techanvil (changes requested):

techanvil (approved):
LGTM, nice one @kuasha420!

includes/Core/Util/Entity_Factory.php
```diff
@@ -177,11 +177,12 @@ public static function from_wp_query( WP_Query $query ) {
 	 * Creates the entity for a given post object.
 	 *
 	 * @since 1.15.0
+	 * @since n.e.x.t method access modifier changed to public.
```
tofumatt, line 180:
```suggestion
	 * @since n.e.x.t Method access modifier changed to public.
```

includes/Core/Util/REST_Entity_Search_Controller.php
```diff
@@ -0,0 +1,141 @@
...
+
+use Google\Site_Kit\Context;
+use Google\Site_Kit\Core\Permissions\Permissions;
+use Google\Site_Kit\Core\REST_API\REST_Route;
+use WP_REST_Server;
+use WP_REST_Request;
+use WP_REST_Response;
+
+/**
+ * Class for handling entity search rest routes.
```
tofumatt, line 21:
```suggestion
 * Class for handling entity search REST routes.
```

tofumatt (approved):
Looks great, just a few wording tweaks in the comments and it's good 👍🏻

Issue #4475, opened 2021-12-07
Create `Module_Sharing_Settings`


aaemnnosttv:
## Feature Description

In preparation for the interface for managing dashboard sharing settings, we need a new class for managing the value like we have for other settings and options.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a new `Module_Sharing_Settings` class in the `Core\Modules` namespace, extending the `Core\Storage\Setting` class
* The option it name it manages is `googlesitekit_dashboard_sharing`
* The default value is an empty array
* The type it is registered with should be `object`
* Its `sanitize_callback` should enforce the following structure:  
	```
	[module-slug]: SharingSettings
	```
	where `SharingSettings` has the following shape
	```
	{
		sharedRoles : string[],
		management  : string,
	}
	```
	`management` can only be one of the following: `owner` or `all_admins` (default: `owner`)

## Implementation Brief

* Create a new class `Core\Modules\Module_Sharing_Settings` that extends the `Core\Storage\Setting` class.
  * Add the `OPTION` constant with the `googlesitekit_dashboard_sharing` value.
  * Add a new protected method `get_default()` that returns an empty array.
  * Add a new protected method `get_type()` that returns `'object'` string.
  * Add a new protected method `get_sanitize_callback()` that returns an anonymous function that accepts `$option` argument and enforces the structure defined in AC.

### Test Coverage

* Add phpunit tests to check the `Module_Sharing_Settings` class.

## QA Brief

QA: Switch on Dashboard Sharing and perform simple smoke testing.

QA:Eng: Additional code review to verify the requirements are met.

## Changelog entry

* N/A

aaemnnosttv:
IB ✅

jimmymadon:
@aaemnnosttv Regarding the `get_sanitize_callback()`, I think you suggested only casting/removing elements rather than modifying/adding default elements. For the `sharedSetting[ $module_slug ][ 'management' ]` key, we said we could modify the `get_default()` method in the future to enforce its value to be `owner` by default. However, the `default` value for the setting is probably only used when the option is not set. So as @eugene-manuilov [suggests](https://github.com/google/site-kit-wp/pull/4668#discussion_r787772641 ), would it make sense to enforce the AC in this callback? Can you confirm what you had in mind for this method in [this PR](https://github.com/google/site-kit-wp/pull/4668)? Thanks!

[...]

PR #4668, opened 2022-01-18
Implement the new module sharing settings class.

jimmymadon:
## Summary

Addresses issue: #4475 

## Relevant technical choices
- Sanitize callback method prevents the addition of invalid wordpress roles.

## Checklist

### DIFF
--- a/includes/Core/Modules/Module_Sharing_Settings.php
+++ b/includes/Core/Modules/Module_Sharing_Settings.php
@@ -13,7 +13,7 @@ namespace Google\Site_Kit\Core\Modules;
 use Google\Site_Kit\Core\Storage\Setting;
 
 /**
- * Base class for module sharing settings.
+ * Class for module sharing settings.
  *
  * @since n.e.x.t
  * @access private
@@ -50,7 +50,7 @@ class Module_Sharing_Settings extends Setting {
 	 *
 	 * @since n.e.x.t
 	 *
-	 * @return callable|null
+	 * @return callable Callback method that filters or type casts invalid setting values.
 	 */
 	protected function get_sanitize_callback() {
 		return function( $option ) {
@@ -75,7 +75,7 @@ class Module_Sharing_Settings extends Setting {
 	}
 
 	/**
-	 * Filter empty or non-string elements from a given array.
+	 * Filters empty or non-string elements from a given array.
 	 *
 	 * @since n.e.x.t
 	 *
--- a/includes/Core/Modules/Module_Sharing_Settings.php
+++ b/includes/Core/Modules/Module_Sharing_Settings.php
@@ -97,4 +97,23 @@ class Module_Sharing_Settings extends Setting {
 		return array_values( $filtered_elements );
 	}
 
+	/**
+	 * Gets the settings after filling in default values.
+	 *
+	 * {@inheritDoc}
+	 *
+	 * @since n.e.x.t
+	 */
+	public function get() {
+		$settings = parent::get();
+
+		foreach ( $settings as $module_slug => $sharing_settings ) {
+			if ( ! isset( $sharing_settings['management'] ) || ! in_array( $sharing_settings['management'], array( 'all_admins', 'owner' ), true ) ) {
+				$settings[ $module_slug ]['management'] = 'owner';
+			}
+		}
+
+		return $settings;
+	}
+
 }
--- a/includes/Core/Modules/Module_Sharing_Settings.php
+++ b/includes/Core/Modules/Module_Sharing_Settings.php
@@ -59,7 +59,11 @@ class Module_Sharing_Settings extends Setting {
 			}
 			$sanitized_option = array();
 			foreach ( $option as $module_slug => $sharing_settings ) {
-				$sanitized_option[ $module_slug ]['sharedRoles'] = $this->sanitize_string_list( $sharing_settings['sharedRoles'] );
+				$sanitized_option[ $module_slug ] = array();
+
+				if ( isset( $sharing_settings['sharedRoles'] ) ) {
+					$sanitized_option[ $module_slug ]['sharedRoles'] = $this->sanitize_string_list( $sharing_settings['sharedRoles'] );
+				}
 
 				if ( isset( $sharing_settings['management'] ) ) {
 					$sanitized_option[ $module_slug ]['management'] = (string) $sharing_settings['management'];
@@ -108,6 +112,9 @@ class Module_Sharing_Settings extends Setting {
 		$settings = parent::get();
 
 		foreach ( $settings as $module_slug => $sharing_settings ) {
+			if ( ! isset( $sharing_settings['sharedRoles'] ) || ! is_array( $sharing_settings['sharedRoles'] ) ) {
+				$settings[ $module_slug ]['sharedRoles'] = array();
+			}
 			if ( ! isset( $sharing_settings['management'] ) || ! in_array( $sharing_settings['management'], array( 'all_admins', 'owner' ), true ) ) {
 				$settings[ $module_slug ]['management'] = 'owner';
 			}
--- a/includes/Core/Modules/Module_Sharing_Settings.php
+++ b/includes/Core/Modules/Module_Sharing_Settings.php
@@ -104,9 +104,9 @@ class Module_Sharing_Settings extends Setting {
 	/**
 	 * Gets the settings after filling in default values.
 	 *
-	 * {@inheritDoc}
-	 *
 	 * @since n.e.x.t
+	 *
+	 * @return array Value set for the option, or registered default if not set.
 	 */
 	public function get() {
 		$settings = parent::get();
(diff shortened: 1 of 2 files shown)


includes/Core/Modules/Module_Sharing_Settings.php
```diff
@@ -0,0 +1,99 @@
...
+	 */
+	protected function get_sanitize_callback() {
+		return function( $option ) {
+			if ( is_array( $option ) ) {
+				foreach ( $option as &$sharing_settings ) {
+					if ( isset( $sharing_settings['sharedRoles'] ) ) {
+						$sharing_settings['sharedRoles'] = $this->filter_valid_roles( $sharing_settings['sharedRoles'] );
+					}
+					if ( isset( $sharing_settings['management'] ) ) {
+						$sharing_settings['management'] = (string) $sharing_settings['management'];
```
eugene-manuilov, line 63: AC says:

> `management` can only be one of the following: `owner` or `all_admins` (default: `owner`)

Let's make sure that `$sharing_settings['management']` is either `owner` or `all_admins` with a fallback to `owner`.

tests/phpunit/integration/Core/Modules/Module_Sharing_SettingsTest.php
```diff
@@ -0,0 +1,67 @@
+<?php
+/**
+ * Module_Sharing_SettingsTest
+ *
+ * @package   Google\Site_Kit\Tests\Core\Modules
+ * @copyright 2021 Google LLC
```
eugene-manuilov, line 6:
```suggestion
 * @copyright 2022 Google LLC
```

includes/Core/Modules/Module_Sharing_Settings.php
```diff
@@ -0,0 +1,99 @@
...
+	 * Gets the callback for sanitizing the setting's value before saving.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return callable|null
+	 */
+	protected function get_sanitize_callback() {
+		return function( $option ) {
+			if ( is_array( $option ) ) {
+				foreach ( $option as &$sharing_settings ) {
```
eugene-manuilov, line 58: Let's better define a new options array and copy information into it instead of modifying array elements using references.

eugene-manuilov (changes requested):
Thanks, @jimmymadon. This is a good start. We still need to make some adjustments before we can merge this PR. Please, take a look at my comments.

includes/Core/Modules/Module_Sharing_Settings.php
jimmymadon: I had originally implemented this but when I had a call with @aaemnnosttv on this and another issue, he suggested I either typecast or remove invalid elements and not modify elements or set default elements in the `sanitize_callback()`.

aaemnnosttv: @jimmymadon correct, sanitizing should be neutral or subtractive, not additive. I think the other part of that that we talked about was potentially adding a method for getting the settings for a specific module, which is where any missing parts would be filled in with defaults. This method could be internal for now and called in `get()` as defaults should be applied in the return value as well.

jimmymadon: @aaemnnosttv @eugene-manuilov So should I create this new method to get settings per module and override `Setting::get()` in `Module_Sharing_Settings` to use this new method?

eugene-manuilov: @jimmymadon, yes.

jimmymadon: @eugene-manuilov @aaemnnosttv Creating a new method to individually get settings per module required me to call `get()` anyways. So for now, I have updated the PR to only override the `Setting::get()` method that fills in a default value for the `management` key. This method can then be used to fetch settings for an individual module in the future.

includes/Core/Modules/Module_Sharing_Settings.php
```diff
@@ -0,0 +1,119 @@
...
+	 * @since n.e.x.t
+	 */
+	public function get() {
+		$settings = parent::get();
+
+		foreach ( $settings as $module_slug => $sharing_settings ) {
+			if ( ! isset( $sharing_settings['management'] ) || ! in_array( $sharing_settings['management'], array( 'all_admins', 'owner' ), true ) ) {
+				$settings[ $module_slug ]['management'] = 'owner';
+			}
+		}
```
aaemnnosttv, line 114: @jimmymadon this is great but please also fill in the `sharedRoles` as well.

includes/Core/Modules/Module_Sharing_Settings.php
aaemnnosttv: @jimmymadon left a comment on the method there, but please add the default for `sharedRoles` as well.

includes/Core/Modules/Module_Sharing_Settings.php
```diff
@@ -0,0 +1,119 @@
...
+	 * @return callable|null
+	 */
+	protected function get_sanitize_callback() {
+		return function( $option ) {
+			if ( ! is_array( $option ) ) {
+				return array();
+			}
+			$sanitized_option = array();
+			foreach ( $option as $module_slug => $sharing_settings ) {
+				$sanitized_option[ $module_slug ]['sharedRoles'] = $this->sanitize_string_list( $sharing_settings['sharedRoles'] );
```
eugene-manuilov, line 62: There is no `$sanitized_option[ $module_slug ]` item in the array at the moment of this assignment. PHP will have to guess the type correctly and implicitly add a new array to the options list at the module slug key. It would be better to explicitly define the correct type for the new item. Also we would need to make sure that `sharedRoles` exists in the `$sharing_settings`.

```php
$sanitized_option[ $module_slug ] = array();
if ( isset( $sharing_settings['sharedRoles'] ) ) {
    $sanitized_option[ $module_slug ]['sharedRoles'] = $this->sanitize_string_list( $sharing_settings['sharedRoles'] );
}
```

includes/Core/Modules/Module_Sharing_Settings.php
```diff
@@ -0,0 +1,119 @@
...
+			}
+		);
+		// Avoid index gaps for filtered values.
+		return array_values( $filtered_elements );
+	}
+
+	/**
+	 * Gets the settings after filling in default values.
+	 *
+	 * {@inheritDoc}
```
eugene-manuilov, line 103: We don't use `{@inheritDoc}` in our classes from the `includes` folder, thus let's get rid of it.
```suggestion
```

includes/Core/Modules/Module_Sharing_Settings.php
```diff
@@ -0,0 +1,119 @@
...
+	 *
+	 * @since n.e.x.t
```
eugene-manuilov, line 105: We also need to add the `@return` tag.

tests/phpunit/integration/Core/Modules/Module_Sharing_SettingsTest.php
```diff
@@ -0,0 +1,111 @@
...
+use Google\Site_Kit\Context;
+use Google\Site_Kit\Core\Modules\Module_Sharing_Settings;
+use Google\Site_Kit\Core\Storage\Options;
+use Google\Site_Kit\Tests\Modules\SettingsTestCase;
+
+class Module_Sharing_SettingsTest extends SettingsTestCase {
+
+	/**
+	 * @inheritDoc
+	 */
```
eugene-manuilov, line 22: @aaemnnosttv why do we need this phpdoc tag in setting tests?

eugene-manuilov (changes requested):
Thanks, @jimmymadon. Added a few more comments.

includes/Core/Modules/Module_Sharing_Settings.php
jimmymadon: I thought that since we were sanitizing `sharedRoles` when saving them and adding a default empty `array()`, we wouldn't need to add defaults when fetching. But makes sense to be consistent and safe here. 👍

includes/Core/Modules/Module_Sharing_Settings.php
jimmymadon: Cool - I saw them in the tests and thought it helped to make it clear that we are overriding. But good to stay consistent - have removed this and added the `@return` tag.

tests/phpunit/integration/Core/Modules/Module_Sharing_SettingsTest.php
aaemnnosttv: @eugene-manuilov I don't think it's needed, but it seems to be used in several other places.

E.g.
- https://github.com/google/site-kit-wp/commit/cfc7a33a54d6579834ed8e4d5133d8acc786e18d#diff-cf79c1c0564dc8b2b41a66557570d042535518601c25f05c567bc500942c43e3R49
- https://github.com/google/site-kit-wp/commit/266d067efd7bd8f5dd15db5f799d59fff4c971b5#diff-3abda76dcad4f6e726d3e69df503252dc93dcb1b7334846332bd0bf1cddb002cR57

These are actually yours 😄  Could be something that someone's IDE added automatically and we didn't think about it too much. It doesn't seem to offer much benefit (at least here) so I'd be happy to remove it; not a big deal either way though.

jimmymadon:
@eugene-manuilov Thanks for the meticulous feedback!

eugene-manuilov (approved):
LGTM

Issue #4527, opened 2021-12-09
Expose recoverable modules to client


aaemnnosttv:
## Feature Description

[Module recovery](https://docs.google.com/document/d/19OtWe7jXKC6dIUfs6J4gDrXujHgCuXnksxBM1uUWmOQ/edit#heading=h.184kxp23btyr) is a new concept in dashboard sharing whereby a shared module can be recovered in the event that the module owner is lost. Later, we'll use this data to alert admins and prompt those who are eligible to recover the module.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The [`googlesitekit-dashboard-sharing-data` script data](https://github.com/google/site-kit-wp/issues/4473) should be extended to include a new `recoverableModules` property as a list of module slugs (`string[]`) that are recoverable
* A recoverable module is a [shareable module](https://docs.google.com/document/d/19OtWe7jXKC6dIUfs6J4gDrXujHgCuXnksxBM1uUWmOQ/edit#heading=h.j8mstfogtglx) that meets any of the following criteria:
	- Has no owner identified by its `ownerID`
	- The user identified by the module's `ownerID` no longer (one of the following)
		- exists
		- is authenticated
		- lacks the necessary capability to share (i.e. `AUTHENTICATE`)
* Note one or more of these cases could probably be covered by some overarching logic but it's important that all are accounted for

## Implementation Brief

In `Core/Assets/Assets.php` update `get_inline_dashboard_sharing_data()` `$inline_data` add a property `recoverableModules` which will be an array of module strings.
First fetch the list of shareable modules `get_shareable_modules()` (introduced in #4521),

For each shareable module get the owner id `get_owner_id()`
https://github.com/google/site-kit-wp/blob/cf9badad4e23c528719aea864e5d74bc5ca6c890/includes/Core/Modules/Module_With_Owner.php#L29

If the owner ID doesn't correspond to a valid user ( `get_user_by( 'ID', $owner_id )` ) 
https://developer.wordpress.org/reference/functions/get_user_by/
Add the module slug to the `recoverableModules` array.

If the owner is not authenticated, add the module slug to the `recoverableModules` array.

See https://github.com/google/site-kit-wp/blob/cf9badad4e23c528719aea864e5d74bc5ca6c890/includes/Core/Authentication/Authentication.php#L625

Also ensure you're checking the module owner, and not the current user, code snippet:
[...]

PR #4670, opened 2022-01-18
Enhance/4527 expose recoverable modules to the client

hussain-t:
## Summary

Addresses issue: #4527 

## Checklist

## Screenshot
Screenshot 2022-01-20 at 11 28 54 PM


includes/Core/Assets/Assets.php
```diff
@@ -781,6 +789,39 @@ private function get_inline_dashboard_sharing_data() {
 			}
 		}
 
+		// Expose recoverable modules to the client.
+		$modules           = new Modules( $this->context );
+		$shareable_modules = $modules->get_shareable_modules();
```
eugene-manuilov, line 794: Instead of creating a new instance of the Modules class here, let's move this logic to a new method `get_recoverable_modules` of the Modules class and add a hook for the `googlesitekit_dashboard_sharing_data` filter to call the new method.

includes/Core/Assets/Assets.php
```diff
@@ -781,6 +789,39 @@ private function get_inline_dashboard_sharing_data() {
...
+
+		foreach ( $shareable_modules as $module ) {
+			$owner_id = $module->get_owner_id();
```
eugene-manuilov, line 797: Not all modules have this method. We need to make sure that `$module` is an instance of the `Module_With_Owner` class, before invoking the `get_owner_id` method.

includes/Core/Assets/Assets.php
```diff
@@ -781,6 +789,39 @@ private function get_inline_dashboard_sharing_data() {
...
+			$user = get_user_by( 'ID', $owner_id );
+
+			// If the User doesn't exists - push the module slug to the recoverableModules array.
+			if ( false === $user ) {
+				$inline_data['recoverableModules'][] = $module->slug;
+				continue;
+			}
+
+			$module_owner_user_options   = new User_Options( $this->context, $user->ID );
+			$module_owner_authentication = new Authentication( $this->context, null, $module_owner_user_options );
```
eugene-manuilov, line 817: You can use `$this->authentication->is_authenticated()` instead of creating new instances once this code is moved to the Modules class. Also you will need to switch the user before checking whether they are authenticated:

```php
// Modules.php
$restore_user = $this->user_options->switch_user( $module_owner_id );
if ( ! $this->authentication->is_authenticated() ) {
    ...
}
$restore_user();
```

eugene-manuilov (changes requested):
Thanks, @hussain-t. This is a good start. Added a few comments for you. Please, take a look.

includes/Core/Assets/Assets.php
hussain-t: That's a good idea 👍

includes/Core/Assets/Assets.php
```diff
@@ -766,8 +766,12 @@ private function get_inline_user_data() {
...
 	 */
 	private function get_inline_dashboard_sharing_data() {
-		$all_roles   = wp_roles()->roles;
-		$inline_data = array( 'roles' => array() );
+		$inline_data = array(
+			'roles' => array(),
+		);
+
+		// Expose roles to the client.
+		$all_roles = wp_roles()->roles;
```
eugene-manuilov, line 774: I think these changes aren't needed anymore. Let's revert it.

includes/Core/Modules/Modules.php
```diff
@@ -1119,4 +1127,45 @@ function( Module $module ) {
...
+	public function get_recoverable_modules() {
+		$recoverable_modules = array();
+		$shareable_modules   = $this->get_shareable_modules();
+
+		foreach ( $shareable_modules as $module ) {
+			$owner_id = $module instanceof Module_With_Owner && $module->get_owner_id();
+
+			// If no owner identified by its owner_id OR lacks the AUTHENTICATE capability -
+			// push the module slug to the recoverableModules array.
+			if ( false === $owner_id || ! user_can( $owner_id, Permissions::AUTHENTICATE ) ) {
```
eugene-manuilov, line 1146: The `get_owner_id()` method can return `0` which also means that there is no owner assigned.
```suggestion
			if ( empty( $owner_id ) || ! user_can( $owner_id, Permissions::AUTHENTICATE ) ) {
```

includes/Core/Modules/Modules.php
```diff
@@ -1119,4 +1127,45 @@ function( Module $module ) {
...
+			}
+
+			// If the owner doesn't have the AUTHENTICATE permission - push the module slug to the recoverableModules array.
+			$user = get_user_by( 'ID', $owner_id );
+
+			// If the User doesn't exists - push the module slug to the recoverableModules array.
+			if ( false === $user ) {
+				$recoverable_modules[] = $module->slug;
+				continue;
+			}
```
eugene-manuilov, line 1158: I think we don't need this check anymore because the `user_can` function verifies that the user exists. If the user doesn't exist anymore, the `user_can` will return `false` and the previous condition will be satisfied. So we can remove this section.

eugene-manuilov (changes requested):
Thanks, @hussain-t. Added a few more comments. Please, take a look.

includes/Core/Assets/Assets.php
hussain-t: @eugene-manuilov Do you mean we don't need to return `roles` array at all to the client?

eugene-manuilov (approved):
LGTM. Thanks, @hussain-t.

includes/Core/Modules/Modules.php
```diff
@@ -1119,4 +1127,38 @@ function( Module $module ) {
...
+	 * @since n.e.x.t
+	 *
+	 * @return array List of recoverable modules.
+	 */
+	public function get_recoverable_modules() {
+		$recoverable_modules = array();
+		$shareable_modules   = $this->get_shareable_modules();
+
+		foreach ( $shareable_modules as $module ) {
+			$owner_id = $module instanceof Module_With_Owner && $module->get_owner_id();
```
aaemnnosttv, line 1142: @hussain-t this expression results in a boolean rather the value of the owner ID as you would get in JavaScript. This results in the permission checks failing because there is no user with an ID of `true`.

To fix this, we can actually remove the `Module_With_Owner` condition here as all shareable modules are modules with an owner. Instead, we can update the return value of `get_shareable_modules` to be `Module_With_Owner[]` so that IDEs are aware that this method is available on `$module`.

cc: @eugene-manuilov

Issue #4510, opened 2021-12-08
Helper text has incorrect font size


asvinb:
## Bug Description
We have a global CSS rule for paragraphs where we set the font size to `1rem`. However, in some cases, for e.g helper texts which are `p` having the `mdc-text-field-helper-text` class name should have a smaller font size, `0.75rem` rather than `1rem` which looks quite big.

## Screenshots

Font size set to `1rem`:
![image](https://user-images.githubusercontent.com/31818/145214543-4b276e05-7aa5-498d-beb4-bdee814b13e4.png)

Font size set to `0.75rem`:
![image](https://user-images.githubusercontent.com/31818/145214600-d358eb74-065f-41dd-a90e-1b237b7d3e0c.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The [default font-size for Site Kit `p` tags](https://github.com/google/site-kit-wp/blob/a27f08668b1d738f72672bfc5d7c61e90bef981c/assets/sass/base/_defaults.scss#L144-L147) should not override material styles

## Implementation Brief

* Update `p` to not match `p` tags that contain a material class prefix (i.e. `p:not([class*="mdc-"])` in the default styles https://github.com/google/site-kit-wp/blob/a27f08668b1d738f72672bfc5d7c61e90bef981c/assets/sass/base/_defaults.scss#L144-L147

### Test Coverage

* May require VRT updates

## QA Brief

- Go to the Site Kit settings page
- Edit Analytics
- Type an incorrect value in the `Ads Conversion ID` text field
- Verify the `font-size` of the helper text should be `0.75rem`
- Verify the text length is smaller than the text field

## Changelog entry

* Update sizing of input helper text to be consistent with Material styles.

aaemnnosttv:
cc: @felixarntz

asvinb:
IB 👍

wpdarren:
@aaemnnosttv as discussed on [Slack](https://10up.slack.com/archives/CBKKQEBR9/p1643720952525489) yesterday, 

> I have noticed that the ‘connect service’ font size is different. The font size in main is as expected so doesn’t impact 1.50.0. Develop is 16px where main is 12px. 

It was confirmed that the font size change above was because of the work in this ticket. It's possible we could have other font size issues across the plugin, so in my opinion we need to compare `develop` to the latest release.  Do you feel any differences should be addressed in this ticket or create new one?  Thoughts appreciated! 

[...]

PR #4679, opened 2022-01-19
Enhance/4510 fix helper text font size new

hussain-t:
## Summary

Addresses issue: #4510 

## Checklist

## Screenshot
Screenshot 2022-01-19 at 10 13 15 AM


assets/sass/base/_defaults.scss
```diff
@@ -142,7 +142,7 @@
 	}
 
 	legend,
-	p {
+	p:not([class*="mdc-"]) {
```
asvinb, line 145: @hussain-t Looking at the VRT updates, I think the intended change by the CSS update should not affect those screens. To fix that, we have 2 solutions:

1. Either we explicitly set `p.mdc-text-field-helper-text` to be `0.75rem` which will prevent a lot of regressions.
2. Or we tweak the styles for other components so that there are no regressions. For e.g adding more specificity to the following selectors and exclude `p.mdc-dialog__lead` from setting the font size to `1rem`:
	* `.googlesitekit-settings-module p.googlesitekit-settings-module__status`
	* `.googlesitekit-settings-connect-module p.googlesitekit-settings-connect-module__text`
	* `.googlesitekit-settings-connect-module p.googlesitekit-settings-connect-module__cta`
	* `.googlesitekit-settings-module p.googlesitekit-settings-module__status`
	* `.googlesitekit-wizard-progress-step p.googlesitekit-wizard-progress-step__text`
	

@aaemnnosttv  I personally prefer the first option since we don't have VRT for all screens and it'll hard/take a lot of time to differentiate between any unwanted changes. Let me know what you think?

asvinb (changes requested):

assets/sass/base/_defaults.scss
aaemnnosttv: @asvinb I would be okay with styling `p.mdc-text-field-helper-text` if this issue were specific to helper-text only, but the problem is a larger one that font-size for `p` tags is interfering with Material styles, so this would not work with the current AC.

As to the VRTs, some of the changes I'm seeing are due to stylesheets for some components not including `.googlesitekit-plugin` at the top level, which all our styles should have.

Let's check for missing `.googlesitekit-plugin` (e.g. `_googlesitekit-cta-link.scss`) and see if adding those balances out the number of VRT failures and additional fixes that are needed.

asvinb: @hussain-t Unfortunately the changes in the PR introduce more regressions. For e.g, you can compare the position of the tooltip when you hover the last button in the following story:
Your PR: https://google.github.io/site-kit-wp/storybook/pull/4679/?path=/story/components-button--vrt-story
Develop: https://google.github.io/site-kit-wp/storybook/main/?path=/story/components-button--vrt-story
Notice how it's spaced incorrectly.

I've pushed a new POC branch where I've added the `.googlesitekit-plugin` prefix only where needed.
https://github.com/google/site-kit-wp/compare/poc/4510-helper-text?expand=1
I tested locally and there don't seem to be any regressions.
@aaemnnosttv I think we can create a separate issue to update all the *.scss files to always have the `.googlesitekit-plugin` prefix and tackle the UI issues there since it can be very time consuming, especially to QA. Let me know what you think and if you're ok with it, i'll create the ticket. @hussain-t Feel free to take a look at my POC and let me know your thoughts.

aaemnnosttv: @asvinb we can create a new issue for the correction to scss files, sure. Would that then be a prerequisite to finishing this issue or are you proposing something else?

As an aside to avoid large diffs, please prefix the first/top-level selector rather than wrap it with `.googlesitekit-plugin` (where possible). 

E.g.


```diff
- .googlesitekit-plugin {
-     .googlesitekit-wizard-progress-step {
+ .googlesitekit-plugin .googlesitekit-wizard-progress-step {
```

asvinb: @aaemnnosttv So to finish this issue, I suggest we only add the `.googlesitekit-plugin` prefix only to files where there are regressions pertaining to the current issue, like I did in the [branch](https://github.com/google/site-kit-wp/compare/poc/4510-helper-text?expand=1) I shared. Then a separate issue will tackle making sure all SCSS files are prefixed correctly. Unless you are happy with doing it in the current ticket and we'll be way above the estimates since it'll require a lot of testing and QA.

About the large diffs, totally agree with you, however for the sake of consistency, I would prefer if we have the first/top-level selector wrapped and immediately it's clear that everything is wrapped within `.googlesitekit-plugin`. It's more obvious 😁 
Let me know what you think!

aaemnnosttv: > So to finish this issue, I suggest we only add the .googlesitekit-plugin prefix only to files where there are regressions pertaining to the current issue, like I did in the branch I shared. Then a separate issue will tackle making sure all SCSS files are prefixed correctly. 

@asvinb would this require changing the AC or we could still complete it as defined but in a more simplified way?

---

As to the future approach, can we maybe apply the wrapping `.googlesitekit-plugin` class at a higher level so we don't need to worry about it within stylesheet partials or would that mess up logic with `&` in selectors? E.g.

```scss
.googlesitekit-plugin {
	@import "components/...";
	@import "components/...";
	@import "components/...";
	// ...
}
```

asvinb: @EvanHerman We can still complete it as defined in a more simplified way in the sense that we don't prefix all the files for now. WE only prefix the files which are causing regressions and then we can open a new ticket to prefix all the files :)

The wrapper at a higher level is awesome! But I think this should go in a separate ticket 😁

hussain-t: Thanks @aaemnnosttv. Wrapping the `.googlesitekit-plugin` class is a great idea 👍 
As @asvinb suggested let's do that in a separate ticket.

asvinb (approved):

aaemnnosttv (approved):
Great, thanks all!

Issue #4677, opened 2022-01-19
GA4 issues when connecting Analytics, unable to select or place snippet. 


jamesozzie:
## Bug Description

There seem to be multiple connected issues during GA4 property selection and also with viewing the existing property, as [highlighted by one user in the forums](https://wordpress.org/support/topic/ga4-settings-disappear/). This is reproducible in support, with the plugin not handling GA4 properties as it did before. 

**The main issue is new users are unable to select or place a GA4 property snippet using Site Kit.** 

A summary of issue findings including the below, occur on all test sites at present (3 tested so far):
- If selecting a GA4 property initially after selecting an account that has an existing GA4 property the screen reloads to only display the UA properties ([video](https://recordit.co/JnSqJ4xUIe))
- If an existing non SK placed GA4 snippet exists users are permitted to select a different GA4 property
- In both cases above when revisiting the GA4 settings overview there are no GA4 reference ([screenshot](https://i.imgur.com/JtwKywZ.png))
- When editing the Analytics settings there is a greyed out GA4 property box ([screenshot](https://i.imgur.com/X0smh04.png))
-  Despite granting all GA4 permissions previously users are asked to grant further permissions each time. 

![2022-01-19-11-10-25](https://user-images.githubusercontent.com/41326532/150118825-4d0738db-28f1-44cf-a67f-4c8cd021bea8.gif)

**Console Errors with the greyed out GA4 box when revisiting (_see screenshot of these errors further down_)**

```
Error handling response: TypeError: self.processResponse is not a function
    at chrome-extension://cmkdbmfndkfgebldhnkbfhlneefdaaip/js/notification.js:154:10
/wp-json/google-site-kit/v1/modules/analytics-4/data/webdatastreams-batch?propertyIDs%5B0%5D=279129939&_locale=user:1 Failed to load resource: the server responded with a status of 500 ()
googlesitekit-api-26ce8eaec5dba8b68ca9.js:1 Google Site Kit API Error method:GET datapoint:webdatastreams-batch type:modules identifier:analytics-4 error:"<p>There has been a critical error on this website.</p><p><a href="https://wordpress.org/support/article/faq-troubleshooting/">Learn more about troubleshooting WordPress.</a></p>"
(anonymous) @ googlesitekit-api-26ce8eaec5dba8b68ca9.js:1
```

**Video 1: Exhibiting GA4 issues during setup: [Video](https://recordit.co/GVeSITKkqU)**
- A new burner site is setup
- SK is setup
[...]

PR #4690, opened 2022-01-20
Update webdatastreams to be web datastreams

aaemnnosttv:
## Summary

Addresses issue:

- #4677

## Relevant technical choices

Web dataStreams now come back from the API like this:

```json
{
      "name": "properties/123456789/dataStreams/9999999",
      "type": "WEB_DATA_STREAM",
      "displayName": "https://example.com - Example!",
      "createTime": "2021-09-09T09:17:05.968Z",
      "updateTime": "2021-09-09T09:17:05.968Z",
      "webStreamData": {
        "measurementId": "G-XXXXXXXXXX",
        "defaultUri": "http://example.com"
      }
},
```

The main change that impacts us is that web datastream data `.measurementId` and `.defaultUri` are now under the `webDataStream` data property rather than part of the top-level object and the object is a generic `dataStream` where the type is determined by the `type` property (`"type": "WEB_DATA_STREAM"`).

## Checklist


tests/phpunit/integration/Core/Authentication/TokenTest.php
```diff
@@ -100,8 +100,11 @@ public function test_set() {
 		$this->assertFalse( $token->set( array( 'expires_in' => 3600 ) ) );
 
 		// Setting only access token also sets reasonable defaults for created and expires_in.
+		$created_before = time();
 		$this->assertTrue( $token->set( array( 'access_token' => 'a1b2c3d4e5' ) ) );
-		$this->assertEquals( time(), $this->user_options->get( OAuth_Client::OPTION_ACCESS_TOKEN_CREATED ) );
+		$created_at = $this->user_options->get( OAuth_Client::OPTION_ACCESS_TOKEN_CREATED );
+		$this->assertGreaterThanOrEqual( $created_before, $created_at );
+		$this->assertLessThanOrEqual( time(), $created_at );
```
aaemnnosttv, line 107: This fixes a random failure where the time changed in between the lines causing a failure.
https://github.com/google/site-kit-wp/runs/4895422306?check_suite_focus=true#step:11:37

I verified this works as expected by testing with an added delay.

image

assets/js/modules/analytics-4/datastore/__fixtures__/webdatastreams.json
```diff
@@ -1,24 +1,28 @@
 [
-  {
-    "_id": "2000",
-    "_propertyID": "1000",
-    "name": "properties/1000/webDataStreams/2000",
-    "measurementId": "1A2BCD345E",
-    "firebaseAppId": "",
```
aaemnnosttv, line 7: As shown in the json in the PR description, `firebaseAppId` is not included here. This is also now part of the `webStreamData` object but not included in the JSON representation when not set.

eugene-manuilov (approved):
LGTM

assets/js/modules/analytics-4/datastore/settings.js
```diff
@@ -87,7 +87,7 @@ export async function submitChanges( { select, dispatch } ) {
 		dispatch( MODULES_ANALYTICS_4 ).setWebDataStreamID( webdatastream._id );
 		dispatch( MODULES_ANALYTICS_4 ).setMeasurementID(
 			// eslint-disable-next-line sitekit/acronym-case
-			webdatastream.measurementId
+			webdatastream.webStreamData.measurementId
```
felixarntz, line 90: Maybe safer to use `?.` here? Not strictly necessary since we only deal with web data streams, but you _do_ use it below, so not sure why we shouldn't have it here.
```suggestion
			webdatastream.webStreamData?.measurementId
```

felixarntz (approved):
@aaemnnosttv Great, thanks for fixing this so promptly! Just one tiny comment, but nothing blocking.

assets/js/modules/analytics-4/datastore/settings.js
aaemnnosttv: `webdatastream` here comes from our `createWebDataStream` action so we should be able to trust that this is web dataStream object as long as there was no error as we do here. Adding the conditional chain here would move the error raised to `setMeasurementID` because `invariant` within will not allow an `undefined` value to be passed. So I think it might be more useful to let this error here if `webStreamData` was missing for some reason although it never should be if `createWebDataStream` was successful. We could add a check in that action perhaps?

The other instance where a conditional chain was used already had one, so I was following what was already there. I'll leave a separate comment on that one.

assets/js/modules/analytics/components/settings/GA4SettingsControls.js
```diff
@@ -116,7 +116,10 @@ export default function GA4SettingsControls() {
 
 		setPropertyID( newPropertyID );
 		setWebDataStreamID( newWebDataStreamID );
-		setMeasurementID( matchedWebDataStream?.measurementId || '' ); // eslint-disable-line sitekit/acronym-case
+		setMeasurementID(
+			// eslint-disable-next-line sitekit/acronym-case
+			matchedWebDataStream?.webStreamData?.measurementId || ''
```
aaemnnosttv, line 121: @felixarntz in this case the `matchedWebDataStream` may not be set, but when it is that value comes from `matchWebDataStream` so it should always be a web dataStream. So the condition on `webStreamData?` shouldn't really be needed here but it is intended to be fail-safe due to the unreliability of `matchedWebDataStream`.

So this could be removed for consistency but doesn't hurt due to the fallback
```suggestion
		setMeasurementID(
			// eslint-disable-next-line sitekit/acronym-case
			matchedWebDataStream?.webStreamData.measurementId || ''
```

felixarntz: I guess my point is in neither case having this hurts :) I see what you mean about the `matchedWebDataStream` potentially not being set, but like you're saying even then that second `?.` is not necessary. So I think we should consistently go with one or the other approach:
* Either never use `?.` where we don't expect the initial value to not be set
* Or be defensive and always use `?.` on "sub-objects" from 3P API

I think I prefer the first approach (so basically what you are suggesting above), since at least we would get a clear failure at this point _if_ the shape of the object was suddenly different from what we're expecting.

aaemnnosttv: @felixarntz Yeah that sounds good. Being overly defensive here would mask a critical change in the API response (if it were to change unexpectedly) so I think we should only use the conditional access when it is expected to be necessary. I'll apply my suggestion above 👍

felixarntz (approved):
Great!

PR #4720, opened 2022-01-25
Enhancement/2044 register site on proxy

eugene-manuilov:
## Summary

Addresses issue:

- #2044

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist

### DIFF
--- a/includes/Core/Authentication/Google_Proxy.php
+++ b/includes/Core/Authentication/Google_Proxy.php
@@ -375,6 +375,8 @@ class Google_Proxy {
 		 * Filters the setup mode.
 		 *
 		 * @since n.e.x.t
+		 *
+		 * @param string An initial setup mode.
 		 */
 		$metadata['mode'] = apply_filters( 'googlesitekit_proxy_setup_mode', $metadata['mode'] );
 
@@ -455,10 +457,11 @@ class Google_Proxy {
 	 *
 	 * @since n.e.x.t
 	 *
+	 * @param string $mode Sync mode.
 	 * @return string|WP_Error Redirect URL on success, otherwise an error.
 	 */
-	public function register_site() {
-		return $this->sync_site_fields( null, null );
+	public function register_site( $mode = 'async' ) {
+		return $this->send_site_fields( null, $mode );
 	}
 
 	/**
@@ -471,7 +474,20 @@ class Google_Proxy {
 	 * @param string      $mode        Sync mode.
 	 * @return string|WP_Error Redirect URL on success, otherwise an error.
 	 */
-	public function sync_site_fields( Credentials $credentials = null, $mode = 'async' ) {
+	public function sync_site_fields( Credentials $credentials, $mode = 'async' ) {
+		return $this->send_site_fields( $credentials, $mode );
+	}
+
+	/**
+	 * Sends site fields to the proxy.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param Credentials $credentials Credentials instance.
+	 * @param string      $mode        Sync mode.
+	 * @return string|WP_Error Redirect URL on success, otherwise an error.
+	 */
+	private function send_site_fields( Credentials $credentials = null, $mode = 'async' ) {
 		$response = $this->request(
 			self::OAUTH2_SITE_URI,
 			$credentials,
@@ -483,7 +499,7 @@ class Google_Proxy {
 					$this->get_user_fields(),
 					$this->get_metadata_fields(),
 					array(
-						'scopes' => implode( ' ', $this->required_scopes ),
+						'scope' => implode( ' ', $this->required_scopes ),
 					)
 				),
 			)
--- a/includes/Core/Authentication/Google_Proxy.php
+++ b/includes/Core/Authentication/Google_Proxy.php
@@ -44,6 +44,7 @@ class Google_Proxy {
 	const ACTION_PERMISSIONS        = 'googlesitekit_proxy_permissions';
 	const ACTION_VERIFY             = 'googlesitekit_proxy_verify';
 	const NONCE_ACTION              = 'googlesitekit_proxy_nonce';
+	const HEADER_REDIRECT_TO        = 'Redirect-To';
 
 	/**
 	 * Plugin context.
@@ -54,6 +55,14 @@ class Google_Proxy {
 	private $context;
 
 	/**
+	 * Required scopes list.
+	 *
+	 * @since n.e.x.t
+	 * @var array
+	 */
+	private $required_scopes;
+
+	/**
 	 * Google_Proxy constructor.
 	 *
 	 * @since 1.1.2
@@ -62,6 +71,17 @@ class Google_Proxy {
 	 */
 	public function __construct( Context $context ) {
 		$this->context = $context;
+	}
+
+	/**
+	 * Sets required scopes to use when the site is registering at proxy.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array $scopes List of scopes.
+	 */
+	public function with_scopes( array $scopes ) {
+		$this->required_scopes = $scopes;
 	}
 
 	/**
--- a/includes/Core/Authentication/Google_Proxy.php
+++ b/includes/Core/Authentication/Google_Proxy.php
@@ -356,8 +356,8 @@ class Google_Proxy {
 	 */
 	public function get_metadata_fields() {
 		$metadata = array(
-			'supports'         => implode( ',', $this->get_supports() ),
-			'nonce'            => wp_create_nonce( self::ACTION_SETUP ),
+			'supports'         => implode( ' ', $this->get_supports() ),
+			'nonce'            => wp_create_nonce( self::NONCE_ACTION ),
 			'mode'             => '',
 			'hl'               => $this->context->get_locale( 'user' ),
 			'application_name' => self::get_application_name(),
--- a/includes/Core/Authentication/Google_Proxy.php
+++ b/includes/Core/Authentication/Google_Proxy.php
@@ -60,7 +60,7 @@ class Google_Proxy {
 	 * @since n.e.x.t
 	 * @var array
 	 */
-	private $required_scopes;
+	private $required_scopes = array();
 
 	/**
 	 * Google_Proxy constructor.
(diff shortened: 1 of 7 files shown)


includes/Core/Authentication/Google_Proxy.php
```diff
@@ -330,6 +354,33 @@ public function get_site_fields() {
...
+			'mode'             => '',
+			'hl'               => $this->context->get_locale( 'user' ),
+			'application_name' => self::get_application_name(),
+			'service_version'  => Feature_Flags::enabled( 'serviceSetupV2' ) ? 'v2' : '',
+		);
+
+		/**
+		 * Filters the setup mode.
+		 *
+		 * @since n.e.x.t
```
felixarntz, line 377: This is missing a `$param` doc for the filtered variable. In WordPress hooks you can just put any variable name there, e.g. `$mode`, it doesn't have to match the name of what's really passed to it.

includes/Core/Authentication/Google_Proxy.php
```diff
@@ -399,24 +450,59 @@ public function unregister_site( Credentials $credentials ) {
...
-				'mode' => $mode,
-				'body' => $this->get_site_fields(),
+				'return' => 'response',
+				'mode'   => $mode,
+				'body'   => array_merge(
+					$this->get_site_fields(),
+					$this->get_user_fields(),
+					$this->get_metadata_fields(),
+					array(
+						'scopes' => implode( ' ', $this->required_scopes ),
```
felixarntz, line 486: Similar to OAuth, this parameter has to be called `scope` (singular), even though it's multiple scopes.

includes/Core/Authentication/Google_Proxy.php
```diff
@@ -399,24 +450,59 @@ public function unregister_site( Credentials $credentials ) {
...
+	 * @since n.e.x.t Updated the function to return redirect URL.
 	 *
 	 * @param Credentials $credentials Credentials instance.
 	 * @param string      $mode        Sync mode.
-	 * @return array|WP_Error Response of the wp_remote_post request.
+	 * @return string|WP_Error Redirect URL on success, otherwise an error.
 	 */
-	public function sync_site_fields( Credentials $credentials, $mode = 'async' ) {
-		return $this->request(
+	public function sync_site_fields( Credentials $credentials = null, $mode = 'async' ) {
```
felixarntz, line 474: This method should _always_ be called with credentials being passed, to sync site fields they need to be present.

I like your idea of reusing the same logic, but could we do it in a way that both public methods "make sense" by themselves? How about we introduce a private method with the parts that is common between them which can be controlled with whatever parameters are needed, but the public method should "suggest" the right parameters that are needed.

Basically this method's parameters should remain like they were before, with `Credentials` instance being required.

includes/Core/Authentication/Google_Proxy.php
```diff
@@ -399,24 +450,59 @@ public function unregister_site( Credentials $credentials ) {
...
 
+	/**
+	 * Registers the site on the proxy.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return string|WP_Error Redirect URL on success, otherwise an error.
+	 */
+	public function register_site() {
+		return $this->sync_site_fields( null, null );
```
felixarntz, line 461: See my other comment, maybe better to use a private "low-level" method here. Also, let's either also add the `$mode = 'async'` parameter so that it can be controlled from the outside, or we need to specify `sync` here because the only use-case we have for it now needs to be `sync`.

For parity with the other method, I prefer having a parameter for it that can be controlled from the outside.

includes/Core/Authentication/Setup_V2.php
```diff
@@ -54,15 +54,23 @@ public function handle_action_setup_start() {
...
 		}
 
-		if ( $this->credentials->has() ) {
-			$this->google_proxy->sync_site_fields( $this->credentials, 'sync' );
+		$required_scopes = $this->authentication->get_oauth_client()->get_required_scopes();
+		$this->google_proxy->with_scopes( $required_scopes );
+
+		$oauth_setup_redirect = $this->credentials->has()
+			? $this->google_proxy->sync_site_fields( $this->credentials, 'sync' )
+			: $this->google_proxy->register_site();
```
felixarntz, line 62: Just pointing out here that, related to my comments above, we may need to provide `'sync'` here if we introduce a parameter for it - I think that would be more consistent.

includes/Modules/Analytics.php
```diff
@@ -121,6 +121,13 @@ public function register() {
 		add_action( 'template_redirect', $this->get_method_proxy( 'register_tag' ) );
 
 		add_filter( 'googlesitekit_proxy_setup_url_params', $this->get_method_proxy( 'update_proxy_setup_mode' ) );
```
felixarntz, line 123: This filter call (and also the entire `googlesitekit_proxy_setup_url_params` filter being applied as well) should be removed, per the ACs.

includes/Modules/Analytics.php
```diff
@@ -1404,11 +1411,25 @@ public function handle_token_response_data( $token_response ) {
...
 	private function update_proxy_setup_mode( $params ) {
-		if ( Feature_Flags::enabled( 'serviceSetupV2' ) && ! $this->is_connected() ) {
-			$params['mode'] = 'analytics-step';
+		$mode = $this->get_proxy_setup_mode();
+		if ( ! empty( $mode ) ) {
+			$params['mode'] = $mode;
 		}
 
 		return $params;
 	}
```
felixarntz, line 1420: See above, this method can then be removed.

tests/phpunit/integration/Core/Authentication/Google_ProxyTest.php
```diff
@@ -334,14 +334,22 @@ public function test_sync_site_fields() {
...
+				'scopes',
+				'service_version',
+				'site_id',
+				'site_secret',
+				'supports',
+				'url',
+				'user_roles',
 			),
 			array_keys( $this->request_args['body'] )
 		);
```
felixarntz, line 355: While this would be somewhat similar, could you also add one simple test for the `register_site` method?

includes/Core/Authentication/Setup_V2.php
```diff
@@ -54,15 +54,23 @@ public function handle_action_setup_start() {
...
+		if ( is_wp_error( $oauth_setup_redirect ) || ! filter_var( $oauth_setup_redirect, FILTER_VALIDATE_URL ) ) {
+			wp_die( esc_html__( 'The request to the authentication proxy has failed. Please, try again later.', 'google-site-kit' ) );
 		}
 
 		if ( $redirect_url ) {
 			$this->user_options->set( OAuth_Client::OPTION_REDIRECT_URL, $redirect_url );
 		}
 
-		$this->redirect_to_proxy();
+		wp_redirect( $oauth_setup_redirect ); // phpcs:ignore WordPress.Security.SafeRedirect.wp_redirect_wp_redirect
```
felixarntz, line 72: Can't we use `wp_safe_redirect` here?

felixarntz (commented):
@eugene-manuilov Looks great, only a few points to review.

includes/Core/Authentication/Google_Proxy.php
eugene-manuilov: Updated.

includes/Core/Authentication/Google_Proxy.php
eugene-manuilov: Updated.

includes/Core/Authentication/Google_Proxy.php
eugene-manuilov: Yep, makes sense. Updated.

includes/Core/Authentication/Google_Proxy.php
eugene-manuilov: Updated.

includes/Core/Authentication/Setup_V2.php
eugene-manuilov: Updated.

includes/Modules/Analytics.php
eugene-manuilov: Removed.

includes/Modules/Analytics.php
eugene-manuilov: Removed.

tests/phpunit/integration/Core/Authentication/Google_ProxyTest.php
eugene-manuilov: Added.

includes/Core/Authentication/Setup_V2.php
eugene-manuilov: No, because the redirect URL has a different domain.

includes/Modules/Analytics.php
felixarntz: You've now removed the filter _usage_ here, can you also remove the entire filter itself please? The filter hook shouldn't even exist anymore.

felixarntz (commented):
@eugene-manuilov New changes lgtm, but please review the two comment replies above. I resolved all other comments.

includes/Modules/Analytics.php
eugene-manuilov: Ok, got it. Removed the filter too.

includes/Core/Authentication/Setup_V2.php
eugene-manuilov: I don't know because I can't verify it on my local. When I try to set up the plugin with the V2 setup flow, I get the following error: 

```
We were unable to send credentials to your WordPress site. Please go back to your site and restart the setup. The following error was returned: No route was found matching the URL and request method.
```

Is it something wrong on the WP side or you need to adjust something on the proxy side?

includes/Core/Authentication/Google_Proxy.php
```diff
@@ -330,6 +347,35 @@ public function get_site_fields() {
...
+	 * Gets metadata fields.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return array Metadata fields array.
+	 */
+	public function get_metadata_fields() {
+		$metadata = array(
+			'supports'         => implode( ',', $this->get_supports() ),
+			'nonce'            => wp_create_nonce( self::ACTION_SETUP ),
```
felixarntz, line 360: This has to use `self::NONCE_ACTION` for the action per the changes in #4383 and #4376.

includes/Core/Authentication/Setup_V2.php
```diff
@@ -54,15 +54,23 @@ public function handle_action_setup_start() {
...
+		exit;
 	}
```
felixarntz, line 75: There is another problem further down in this file that I discovered during testing, but it's more of an oversight from #4376 than it belongs to this issue. Still, let's fix it in here: In `Setup_V2::handle_action_exchange_site_code`, if the request is successful, we currently redirect back to the proxy without adding the `site_id` query parameter which causes an exception.

includes/Core/Authentication/Setup_V2.php
felixarntz: I just tested this (note that you have to use the staging proxy) and found a few other small, but crucial problems that were preventing the flow from working. But yes, I then verified that `wp_safe_redirect` works here as expected, since we allowlist the host.

felixarntz (changes requested):
@eugene-manuilov Found a few minor, but critical problems during my testing (see below).

Since I've already addresed them locally, I'll push these updates here myself.

felixarntz (approved):
With the additional changes, lgtm!

Will assign this back to you @eugene-manuilov for a final sanity check on my additional changes. I've already added a QAB for this.

eugene-manuilov:
> With the additional changes, lgtm!
> 
> Will assign this back to you @eugene-manuilov for a final sanity check on my additional changes. I've already added a QAB for this.

Thanks, @felixarntz. All looks good to me. I can't approve mine PR. So assigning this back to you.

Issue #4724, opened 2022-01-26
Refine logic for recoverable modules


aaemnnosttv:
## Feature Description

Building on the initial logic defined in #4527, the set of recoverable modules should be refined to only include modules that are shared.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The definition of a recoverable module should be refined (in addition to the existing criteria) to  modules which are currently shared with one or more user roles via the dashboard sharing settings.  
_This is particularly important to avoid "all admin modules" (such as PSI) to appear recoverable right away as they only get their owner ID by association to their dashboard sharing settings_
* This logic in `get_recoverable_modules` should be extracted into a new public `is_module_recoverable` method on the `Modules` class
	* This method should receive added test coverage

## Implementation Brief

- Extract the callback method from `Modules::get_recoverable_modules()` into a new public `is_module_recoverable` method in the same class.
- Modify this method by checking the incoming parameter and if it is a `string`, it should try to get the module instance using the `Modules::get_module()` method. This way the method should be able to accept either an instance of `Module` or a module slug string. 
- In the same method, after checking if `$module instanceof Module_With_Owner`, add a check to see if the module has any `sharedRoles` using the `Module_Sharing_Settings` class. If there are no `sharedRoles` set for the Module, then the method should `return false`.

### Test Coverage

* Add PHPUnit tests for the new `is_module_recoverable()` method. Cover the already existing cases within this method as well.

## QA Brief

**QA:Eng**
- Make sure the `dashboardSharing` feature is enabled
- Open the browser developer console and type `window._googlesitekitDashboardSharingData`
- Verify this refactor doesn’t break the existing functionality i.e. the `recoverableModules` array should be available in the console
- Verify the changes met the AC
- Verify the newly added tests are getting passed 

## Changelog entry

* N/A

aaemnnosttv:
@felixarntz anything else we should consider here?

felixarntz:
@aaemnnosttv Hmm, why is that condition needed? Wouldn't it technically also make sense to recover a module that is not shared? For example if there are two admins, one sets up Analytics, then disappears. I guess the other admin should then still be able to recover the module, or not? Maybe I'm missing something here.

aaemnnosttv:
> Wouldn't it technically also make sense to recover a module that is not shared? For example if there are two admins, one sets up Analytics, then disappears. I guess the other admin should then still be able to recover the module, or not?

[...]

PR #4725, opened 2022-01-26
Fix expose recoverable modules incorrect condition - Follow up

hussain-t:
## Summary

Addresses issue:

- #4527 

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist

### DIFF
--- a/includes/Core/Modules/Modules.php
+++ b/includes/Core/Modules/Modules.php
@@ -1131,14 +1131,14 @@ final class Modules {
 	 *
 	 * @since 1.50.0
 	 *
-	 * @return array List of recoverable modules.
+	 * @return Module_With_Owner[] array List of recoverable modules.
 	 */
 	public function get_recoverable_modules() {
 		$recoverable_modules = array();
 		$shareable_modules   = $this->get_shareable_modules();
 
 		foreach ( $shareable_modules as $module ) {
-			$owner_id = $module instanceof Module_With_Owner && $module->get_owner_id();
+			$owner_id = $module->get_owner_id();
 
 			// 1. If no owner identified by its owner_id
 			// 2. Lacks the AUTHENTICATE Permissions


includes/Core/Modules/Modules.php
```diff
@@ -1131,14 +1131,14 @@ function( Module $module ) {
 	 *
 	 * @since 1.50.0
 	 *
-	 * @return array List of recoverable modules.
+	 * @return Module_With_Owner[] array List of recoverable modules.
```
aaemnnosttv, line 1134: @hussain-t this return type was intended for the `get_shareable_modules` method – this method currently returns a list of strings rather than module objects.

This wasn't spelled out in the AC or IB but `get_*_modules` methods on this class return an array of `Module` instances, but this one currently returns a list of slugs.

Let's update this method to return a filtered list of `Module` objects instead (like `get_shareable_modules` does only with the logic we have here) in which case this return type can stay.

Then we can `wp_list_pluck` the slugs from the list of modules above for the dashboard sharing data.

includes/Core/Modules/Modules.php
```diff
@@ -1131,14 +1131,14 @@ function( Module $module ) {
...
 	 */
 	public function get_recoverable_modules() {
```
aaemnnosttv, line 1136: As a public method, this should have added tests for it by convention. Since we don't have a need for it publicly yet, let's make it `protected` for now. This will receive added testing in https://github.com/google/site-kit-wp/issues/4724.

aaemnnosttv (commented):
@hussain-t This PR fixes the main problem we were having before but still needs some minor reworking. Also I see this PR was merged into `develop` but it should be targeting `main` at this point since this PR is a follow-up to another that is for the 1.50.0 release (currently in main).

Please open another PR to address the points below from main and against main (you can cherry pick this commit into it as a starting point). Once merged, we'll merge it back into develop again so you don't need to worry about that.

cc: @eugene-manuilov

PR #4748, opened 2022-01-31
Fix Analytics Goals chart when in Zero State.

tofumatt:
## Summary

Addresses issue: #4272.

- #

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist

### DIFF
--- a/assets/js/modules/search-console/components/dashboard/SearchFunnelWidget/AnalyticsStats.js
+++ b/assets/js/modules/search-console/components/dashboard/SearchFunnelWidget/AnalyticsStats.js
@@ -27,7 +27,10 @@ import PropTypes from 'prop-types';
 import Data from 'googlesitekit-data';
 import { Grid, Row, Cell } from '../../../../../material-components';
 import { CORE_MODULES } from '../../../../../googlesitekit/modules/datastore/constants';
-import { extractAnalyticsDashboardData } from '../../../../analytics/util';
+import {
+	extractAnalyticsDashboardData,
+	isZeroReport,
+} from '../../../../analytics/util';
 import GoogleChart from '../../../../../components/GoogleChart';
 const { useSelect } = Data;
 
@@ -84,6 +87,12 @@ const AnalyticsStats = ( {
 			},
 		},
 	};
+
+	if ( isZeroReport( data ) ) {
+		options.vAxis.viewWindow.max = 1;
+	} else {
+		options.vAxis.viewWindow.max = undefined;
+	}
 
 	return (
 		<Grid className="googlesitekit-analytics-site-stats">


assets/js/modules/search-console/components/dashboard/SearchFunnelWidget/AnalyticsStats.js
```diff
@@ -85,6 +88,12 @@ const AnalyticsStats = ( {
 		},
 	};
 
+	if ( isZeroReport( data ) ) {
+		options.vAxis.viewWindow.max = 1;
+	} else {
+		options.vAxis.viewWindow.max = undefined;
+	}
+
```
tofumatt, line 96: This is the same approach used elsewhere; if this isn't used then it's possible the chart heights get messed up, I think because of either loading states but possibly/also because of how the Chart component works. The IB omitted this but it's needed.

I also opted to use the `isZeroReport` function rather than the custom function in the IB, because this worked fine to identify a non-zero report.

assets/js/modules/search-console/components/dashboard/SearchFunnelWidget/AnalyticsStats.js
```diff
@@ -85,6 +88,12 @@ const AnalyticsStats = ( {
 		},
 	};
 
+	if ( isZeroReport( data ) ) {
```
aaemnnosttv, line 91: This doesn't work as expected because `data` includes metrics for both goals _and_ bounce rate (which is shown in place of goals on the entity dashboard). `isZeroReport` is only true if _all_ the report data is zero, which means that it will return false if the bounce rate is non-zero.

We need to handle things a bit differently based on which stat is selected (goals or bounce rate) which can be determined by the `selectedStats` prop:

image
image

assets/js/modules/search-console/components/dashboard/SearchFunnelWidget/AnalyticsStats.js
```diff
@@ -85,6 +88,12 @@ const AnalyticsStats = ( {
...
+		options.vAxis.viewWindow.max = 1;
```
aaemnnosttv, line 92: The `max` here should be set with a more appropriate value to use for the max y axis value. Using a value of `1` results in fractional increments between `0` and `1` which doesn't make sense for Goals because you can't partially complete a goal and it isn't displaying an average.

Similarly, bounce rate is a percentage so it would naturally have a maximum of `100` (also used elsewhere).

assets/js/modules/search-console/components/dashboard/SearchFunnelWidget/AnalyticsStats.js
aaemnnosttv: Unfortunately we can't rely on `isZeroReport` due to reasons mentioned above. I think for these two, we can simply look at the respective index (`selectedStats`) of each item in `data.0.data.totals`.

aaemnnosttv (changes requested):
Thanks @tofumatt – turns out this is a little more complicated than initially thought due to the multiple metrics included in the same data set while we only show one of them at a time on the chart. See below.

assets/js/modules/search-console/components/dashboard/SearchFunnelWidget/AnalyticsStats.js
tofumatt: Ah, right on, makes sense. 👍🏻

aaemnnosttv (changes requested):
Thanks @tofumatt – this is almost there, just needs a few tweaks.

aaemnnosttv (approved):
LGTM 👍

PR #4750, opened 2022-02-01
Tag Manager Existing Tag Simplification

kuasha420:
## Summary

Addresses issue:

- #4709

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [x] Run the code.
- [x] Ensure CI checks pass.

## Merge Reviewer Checklist


felixarntz:
@aaemnnosttv @tofumatt @eugene-manuilov Just flagging here to please not merge this PR into `develop`. As discussed recently, we'll need to have all changes for the _Existing tag simplification_ epic in a separate "feature" branch. This will also apply to all upcoming issues in the epic.

I just chat with @aaemnnosttv, and he'll look into creating and defining the branch approach.

aaemnnosttv:
@felixarntz I've created that branch and updated the PR here to target it. I've also updated all the issues for this epic to include a note about this detail in the ACs.

As a result, PR checks won't run for this until #4779 is merged but I would think we can get that through fairly soon 👍

stories/modules-tagmanager-settings.stories.js
```diff
@@ -149,70 +149,6 @@ storiesOf( 'Tag Manager Module/Settings', module )
 			decorators: [ withRegistry ],
 		}
 	)
-	.add(
```
techanvil, line 152: Rather than delete both these _View..._ stories, I think we should keep one of them - say the _with access_ one, remove the call to `receiveGetTagPermission` and remove `(with access)` from the description. Otherwise we don't have a story to show the "existing tag" behaviour at all.

stories/modules-tagmanager-settings.stories.js
```diff
@@ -389,104 +325,6 @@ storiesOf( 'Tag Manager Module/Settings', module )
 		{
 			decorators: [ withRegistry ],
 		}
-	)
-	.add(
```
techanvil, line 393: Likewise, let's keep one of these _Edit..._ stories and remove the access related stuff from it.

stories/modules-tagmanager-setup.stories.js
```diff
@@ -256,59 +256,6 @@ storiesOf( 'Tag Manager Module/Setup', module )
 			padding: 0,
 		}
 	)
-	.add(
```
techanvil, line 259: Same, here - let's keep one of these _Existing tag..._ stories and remove the access stuff from it.

techanvil (changes requested):
Hey @kuasha420. This is looking good, always a pleasure to see so much code removed, less is more, as they say!

Just a few related comments about the removed stories, and also a resolution to your own [question](https://github.com/google/site-kit-wp/issues/4709#issuecomment-1028342561) and this should be good to go...

techanvil:
Hi @kuasha420, while testing this branch I've noticed the existing GTM tag flow is now a bit broken - the dropdowns for Account and Container are not longer populated, and the Confirm... button is disabled. This is evident on Storybook and also in testing on my local install.

image

image

I think the flow should remain as seen in the current "with access" story:

image

kuasha420:
@techanvil thank you for spotting the issues. 

For not auto selecting correct container, I had to loop over all the accounts and try to find the account that has the container that's present in the existing tag, and if found, we can select the account and container. I am not sure if that's the best approach here. 

For the inputs being disabled, that will be solved in #4713 as AC of that issue covers it. 

For the Submit button being disabled, The button remains disabled if a valid account is not selected. See `validadeCanSubmitChange`. I'm not sure if that should be changed here. I think it's out of scope and will need a dedicated issue or should be fixed in  #4713 as well. 

Cheers.

techanvil:
> @techanvil thank you for spotting the issues.
> 
> For not auto selecting correct container, I had to loop over all the accounts and try to find the account that has the container that's present in the existing tag, and if found, we can select the account and container. I am not sure if that's the best approach here.
> 
> For the inputs being disabled, that will be solved in #4713 as AC of that issue covers it.
> 
> For the Submit button being disabled, The button remains disabled if a valid account is not selected. See `validadeCanSubmitChange`. I'm not sure if that should be changed here. I think it's out of scope and will need a dedicated issue or should be fixed in #4713 as well.
> 
> Cheers.

Hi @kuasha420, thanks for the info here. In that case, this looks good to go and we can assess the button behaviour if it's not covered in #4713. Luckily this branch is being merged into `feature/existing-tag-simplification`, so there's no issue with interrupting the setup flow on `develop` :)

techanvil (approved):
LGTM, nice one @kuasha420! :tada:

assets/js/modules/tagmanager/datastore/containers.js
```diff
@@ -404,6 +404,39 @@ const baseSelectors = {
...
 		return Object.values( state.isFetchingCreateContainer ).some( Boolean );
 	},
+	/**
+	 * Gets all containers for all accounts.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return {(Array|undefined)} Array of containers, or `undefined` if not loaded yet.
+	 */
+	getAllContainers: createRegistrySelector( ( select ) => () => {
```
aaemnnosttv, line 414: I believe this was added to be able to find the account (and container) in the presence of an existing tag which was previously done on the server side as part of the permission check. Going forward we won't be auto-selecting values for a detected existing tag (the inputs will be controlled by the saved value only) so this won't be necessary.

There are a few other things that aren't quite right in this selector but I'll omit those details as we won't be keeping it.

assets/js/modules/tagmanager/hooks/useExistingTagEffect.js
```diff
@@ -38,11 +38,9 @@ export default function useExistingTagEffect() {
 	const containerID = useSelect( ( select ) =>
 		select( MODULES_TAGMANAGER ).getContainerID()
 	);
-	const existingTagPermission = useSelect( ( select ) =>
-		select( MODULES_TAGMANAGER ).getTagPermission( existingTag )
-	);
-	const hasExistingTagPermission = useSelect( ( select ) =>
-		select( MODULES_TAGMANAGER ).hasExistingTagPermission()
+
+	const allContainers = useSelect( ( select ) =>
```
aaemnnosttv, line 42: As mentioned before, this selector isn't necessary. While the IB rightly calls for removing all permission related logic from this hook, the hook itself was dependent on the permissions data for the account ID to select. This hook was essentially slated for removal as part of https://github.com/google/site-kit-wp/issues/4713 but it may as well happen here as there is nothing it can or should do now that the permissions data has been removed.

Let's also remove this hook here rather than patching it up only to be removed later.

aaemnnosttv (changes requested):
Thanks @kuasha420 – this looks great overall, there was just one thing that was added that we don't need and allows us to remove a bit more here. I realize some of this overlaps with https://github.com/google/site-kit-wp/issues/4713 but may have been a bit of an oversight in the definition that it should have happened as part of this one instead.

assets/js/modules/tagmanager/datastore/containers.js
techanvil: Thanks for spotting this @aaemnnosttv, I must confess that I didn't spot @kuasha420's recent commit. I had misread this paragraph of Arafat's [comment](https://github.com/google/site-kit-wp/pull/4750#issuecomment-1046185340).

> For not auto selecting correct container, I had to loop over all the accounts and try to find the account that has the container that's present in the existing tag, and if found, we can select the account and container. I am not sure if that's the best approach here.

I mistakenly read that as a question and the rest of the comment as essentially saying that we don't need to do this now as it will be addressed in #4713. I therefore approved the PR on the basis of my previously review and testing, thinking nothing had been added.

kuasha420:
@aaemnnosttv @techanvil yes the AC/IB was kind of vague on the part about auto selection. I have removed that code as well as the disabling of Fields when there's existing tag as part of this PR. I'm still not sure about the useSnippet filter but I saw the conversation in the issue. I will await for that to be resolved/decided but we can roll it with 4713 or another issue to not block this one.. Sending it for another review. 

Cheers.

aaemnnosttv (approved):
LGTM, thanks @kuasha420 !

assets/js/modules/tagmanager/components/common/AccountSelect.js
```diff
@@ -77,7 +74,6 @@ export default function AccountSelect() {
 			label={ __( 'Account', 'google-site-kit' ) }
 			value={ accountID }
 			onEnhancedChange={ onChange }
-			disabled={ hasExistingTag }
```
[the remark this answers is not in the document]
kuasha420: Yep tom raised the issue of having this not done here. ie. Setup flow broken and not being able to continue. Would be more confusing for the QA so I put it here :D

Issue #4729, opened 2022-01-27
PHP tests broken on PHP 8 with WP latest


aaemnnosttv:
## Bug Description

https://github.com/google/site-kit-wp/runs/4963741256?check_suite_focus=true#step:11:11

```
> phpunit
PHP Warning:  require_once(/tmp/wordpress-tests-lib/includes/phpunit7/MockObject/Builder/NamespaceMatch.php): Failed to open stream: No such file or directory in /home/runner/work/site-kit-wp/site-kit-wp/tests/phpunit/bootstrap.php on line 56
PHP Fatal error:  Uncaught Error: Failed opening required '/tmp/wordpress-tests-lib/includes/phpunit7/MockObject/Builder/NamespaceMatch.php' (include_path='.:/usr/share/php') in /home/runner/work/site-kit-wp/site-kit-wp/tests/phpunit/bootstrap.php:56
Stack trace:
#0 /home/runner/work/site-kit-wp/site-kit-wp/vendor/phpunit/phpunit/src/Util/FileLoader.php(57): include_once()
#1 /home/runner/work/site-kit-wp/site-kit-wp/vendor/phpunit/phpunit/src/Util/FileLoader.php(45): PHPUnit\Util\FileLoader::load()
#2 /home/runner/work/site-kit-wp/site-kit-wp/vendor/phpunit/phpunit/src/TextUI/Command.php(1058): PHPUnit\Util\FileLoader::checkAndLoad()
#3 /home/runner/work/site-kit-wp/site-kit-wp/vendor/phpunit/phpunit/src/TextUI/Command.php(863): PHPUnit\TextUI\Command->handleBootstrap()
#4 /home/runner/work/site-kit-wp/site-kit-wp/vendor/phpunit/phpunit/src/TextUI/Command.php(173): PHPUnit\TextUI\Command->handleArguments()
#5 /home/runner/work/site-kit-wp/site-kit-wp/vendor/phpunit/phpunit/src/TextUI/Command.php(162): PHPUnit\TextUI\Command->run()
#6 /home/runner/work/site-kit-wp/site-kit-wp/vendor/phpunit/phpunit/phpunit(61): PHPUnit\TextUI\Command::main()
#7 /home/runner/work/site-kit-wp/site-kit-wp/vendor/bin/phpunit(120): include('...')
#8 {main}
  thrown in /home/runner/work/site-kit-wp/site-kit-wp/tests/phpunit/bootstrap.php on line 56

[...]

PR #4752, opened 2022-02-01
Update PHP tests for compatibility with PHP 8 and WP 5.9+ and WP < 5.2

aaemnnosttv:
## Summary

Addresses issue:

- #4729 

The "wp-phpunit" portion of WordPress core changed with WP 5.9 to allow for compatibility with newer versions of PHPUnit which are incompatible with PHP below PHP 7 due to the addition of return types on fixture methods (e.g. `setUp`, `tearDown` etc). Omitting these return types raises errors due to incompatible definitions which forced WP core to adopt the [PHPUnit Polyfills](https://github.com/Yoast/PHPUnit-Polyfills) library. This allows for authoring test cases that are forward-compatible with newer versions of PHPUnit.

This does not come for free however and requires us to rewrite our fixture methods to match the snake-case (cross-version compatible) names used by WordPress core's base test case.

This is further complicated by the changes made to the WP core base test case only being back ported to WP 5.2 which leaves the new methods uncalled in tests based on wp-phpunit for WP below 5.2.

The solution here was to introduce an additional polyfill layer between Site Kit's base test case and the `WP_UnitTestCase` by using a new `WP_UnitTestCase_Polyfill` class which works very similar to way the PHPUnit polyfill test case works only a bit simpler. If the `WP_UnitTestCase` class does not have the newer methods, we load a polyfill version of the class which is basically a copy of the same class used in WP 5.2 of the library. Otherwise, we load an empty "pass-through" version of the class; both then extend the `WP_UnitTestCase` class. This allows us to restore the same functionality for our test case while maintaining support for wp-phpunit with our current minimum supported version of WP (4.7).

The added polyfill has been implemented as a local path-type Composer package, to make it easier to extract/replace later.

The rest of the changes here are to fix deprecation warnings with v8 and v9 of PHPUnit which otherwise fail, even though no test failures occur. Using PHPUnit 8 seemed to be required as PHPUnit 7 was failing with parse errors ("unexpected token match"). The lift for compatibility from PHPUnit 8 to 9 was trivial so this was included as well. 

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist

### DIFF
--- a/.github/workflows/php-tests-wp-nightly-php-7-4.yml
+++ b/.github/workflows/php-tests-wp-nightly-php-7-4.yml
@@ -58,7 +58,7 @@ jobs:
       - uses: shivammathur/setup-php@v2
         with:
           extensions: mysqli, runkit7, uopz
-          tools: composer:2.1
+          tools: composer:2.1, phpunit-polyfills
           php-version: '7.4'
       - name: Get Composer Cache Directory
         id: composer-cache
--- a/.github/workflows/php-tests-wp-nightly-php-7-4.yml
+++ b/.github/workflows/php-tests-wp-nightly-php-7-4.yml
@@ -58,7 +58,7 @@ jobs:
       - uses: shivammathur/setup-php@v2
         with:
           extensions: mysqli, runkit7, uopz
-          tools: composer:2.1, phpunit-polyfills
+          tools: composer:2.1
           php-version: '7.4'
       - name: Get Composer Cache Directory
         id: composer-cache
--- a/.github/workflows/php-tests-wp-nightly-php-7-4.yml
+++ b/.github/workflows/php-tests-wp-nightly-php-7-4.yml
@@ -75,4 +75,4 @@ jobs:
       - name: Set up PHP test data
         run: tests/bin/install-wp-tests.sh ${MYSQL_DATABASE} ${MYSQL_USER} ${MYSQL_PASSWORD} ${DB_HOST}:${DB_PORT} ${WP_VERSION}
       - name: Run Unit Tests
-        run: composer test
+        run: composer test -- --verbose
--- a/tests/phpunit/bootstrap.php
+++ b/tests/phpunit/bootstrap.php
@@ -9,8 +9,6 @@
  */
 
 define( 'TESTS_PLUGIN_DIR', dirname( dirname( __DIR__ ) ) );
-// Ensure Composer autoloader is available.
-require_once TESTS_PLUGIN_DIR . '/vendor/autoload.php';
 
 if ( false !== getenv( 'WP_PLUGIN_DIR' ) ) {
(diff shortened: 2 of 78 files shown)


.github/workflows/php-tests-wp-nightly-php-7-4.yml
```diff
@@ -75,4 +75,4 @@ jobs:
       - name: Set up PHP test data
         run: tests/bin/install-wp-tests.sh ${MYSQL_DATABASE} ${MYSQL_USER} ${MYSQL_PASSWORD} ${DB_HOST}:${DB_PORT} ${WP_VERSION}
       - name: Run Unit Tests
-        run: composer test
+        run: composer test -- --verbose
```
felixarntz, line 78: Not opposed, but curious why you changed this throughout.

tests/phpunit/bootstrap.php
```diff
@@ -9,6 +9,8 @@
  */
 
 define( 'TESTS_PLUGIN_DIR', dirname( dirname( __DIR__ ) ) );
+// Ensure Composer autoloader is available.
+require_once TESTS_PLUGIN_DIR . '/vendor/autoload.php';
```
felixarntz, line 13: Moving this here seems a bit unrelated, was there a specific problem that led you to doing that? In any case, I agree having that here outside of the condition below makes more sense.

felixarntz (approved):
@aaemnnosttv Amazing work!

While it obviously has to be a bit clunky, this implementation is as clean as it can be and works reliably. I left two minor comments, but nothing blocking.

.github/workflows/php-tests-wp-nightly-php-7-4.yml
aaemnnosttv: Mostly because it adds information about the runtime to the run:

```
Runtime:       PHP 8.0.14
Configuration: /home/runner/work/site-kit-wp/site-kit-wp/phpunit.xml.dist
```

It also adds things like the names of tests that were incomplete or have been skipped.

If we really wanted to make our output more verbose (like Jest) we could use `--testdox`, but `--verbose` doesn't really add that much. I mainly wanted to expose the specific version of PHP used.

tests/phpunit/bootstrap.php
aaemnnosttv: It's now needed in all cases (not only when using wp-phpunit locally) since we use it to load the PHPUnit Polyfills library, but also the added `packages/wp-phpunit-polyfill` package which is loaded by Composer.

aaemnnosttv: Turns out this actually isn't needed at all because PHPUnit loads it automatically 😄

aaemnnosttv:
Thanks @felixarntz if there's anything else, let's address it in a follow-up PR/issue 👍

PR #4830, opened 2022-02-09
Add `Module_With_Service_Entity`.

kuasha420:
## Summary

Addresses issue:

- #4579 

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist


kuasha420:
This is ready for review and feedback. A few things:

1. I am not sure if the Analytics_4 implementation of the method is correct. Rest seemed to work from my testing.. 

2. I am not sure how to unit test this method as it relies on API calls to Google Server. Any pointers would be appreciated. 

Cheers.

aaemnnosttv (changes requested):
Thanks @kuasha420 this looks like a solid start, just a few comments on what we have so far.

Regarding the tests, it might make sense to pair with someone else to finish this part.

When adding tests for an interface like we are here, we have a convention for testing these using a "contract tests" type approach where we write the tests as a trait which is then used by the various module test cases to ensure all implementations fulfill to the contract. Abstract methods can be used to require things from the test case that are specific to the module if needed. See `*_ContractTests` in the codebase for examples. The HTTP mocking we need here is very simple as the test is only really concerned with the response code. See usage of `FakeHttpClient` for how this is done. As always, let me know if you have any questions 👍

includes/Modules/AdSense.php
```diff
@@ -883,4 +885,34 @@ private function render_platform_meta_tags() {
...
+			$request = $this->create_adsense_earning_data_request( $data_request );
+
+			if ( is_wp_error( $request ) ) {
+				return $request;
+			}
+		} catch ( Exception $e ) {
+			if ( $e->getCode() === 403 ) {
+				return false;
+			}
+			return $this->exception_to_error( $e, 'check_service_entity_access' );
```
aaemnnosttv, line 912: For the `$datapoint` let's change this parameter to be optional, defaulting to an empty string and remove that here as `check_service_entity_access` isn't a datapoint. It's a bit odd that the argument is required since it isn't used in the base method. Please add a new `@since` annotation for this change as well. Note also, this will need to be updated in all subclasses that implement `exception_to_error` (Analytics, and FakeModule).

Then this can simply be called with just the exception.

tests/phpunit/includes/Core/Modules/Module_With_Service_Entity_ContractTests.php
```diff
@@ -0,0 +1,94 @@
...
+	 * @return Module|Module_With_Service_Entity
+	 */
+	abstract protected function get_module_with_service_entity();
+
+	public function test_check_service_entity_access() {
+
+		$testcase = $this->get_testcase();
+		$module   = $this->get_module_with_service_entity();
+
+		$this->setup_preconditions( $module );
```
aaemnnosttv, line 36: We should be careful to add generic methods like this in such test traits because these will be exposed in the consuming class as well.

Also, it's a bit unclear what to expect as a result of calling this method. It would be better to delegate any module-specific details to the consumer with added abstract methods.

tests/phpunit/includes/Core/Modules/Module_With_Service_Entity_ContractTests.php
```diff
@@ -0,0 +1,94 @@
...
+
+		$fake_http_client->set_request_handler( $this->request_handler_success() );
+		$testcase->assertEquals( true, $module->check_service_entity_access() );
+
+		// the `check_service_entity_access` method of Optimize module will always return true.
+		// This is because Optimize module does not have an api to check service entity.
+		// Therefore, we can't test for the other response codes.
+		if ( $module instanceof Optimize ) {
+			return;
+		}
```
aaemnnosttv, line 49: As mentioned above, we should avoid module-specific details in the tests here which are intended to be abstract.

For Optimize in particular as a special case, we should be able to mark these tests as skipped instead.

tests/phpunit/includes/Core/Modules/Module_With_Service_Entity_ContractTests.php
```diff
@@ -0,0 +1,94 @@
...
+
+		$fake_http_client->set_request_handler( $this->request_handler_forbidden() );
+		$testcase->assertEquals( false, $module->check_service_entity_access() );
+
+		$fake_http_client->set_request_handler( $this->request_handler_error() );
+		$testcase->assertEquals( true, is_wp_error( $module->check_service_entity_access() ) );
```
aaemnnosttv, line 55: Let's break this into separate tests for the different cases – then we can either have separate setup methods for each, or we can have a single `setup_test_check_service_entity_access` with a parameter to indicate the variation (success, no_access, or error)

This way, `Optimize` should be able to call `markTestSkipped` for the cases that don't apply to it.

tests/phpunit/includes/Core/Modules/Module_With_Service_Entity_ContractTests.php
```diff
@@ -0,0 +1,94 @@
...
+		return function () {
+			return new Response( 403 );
+		};
+	}
+
+	protected function request_handler_error() {
+		return function () {
+			return new Response( 401 );
+		};
+	}
```
aaemnnosttv, line 92: As mentioned above, let's avoid adding generic helper methods into the trait here. These are small enough that we could inline them into the tests which, as mentioned should be separated into the 3 cases as well so the context would already be clear for each.

tests/phpunit/includes/Core/Modules/Module_With_Service_Entity_ContractTests.php
```diff
@@ -0,0 +1,94 @@
+<?php
+/**
+ * Module_With_Service_Entity_ContractTests
+ *
+ * @package   Google\Site_Kit\Tests\Core\Modules
+ * @copyright 2021 Google LLC
```
aaemnnosttv, line 6:
```suggestion
 * @copyright 2022 Google LLC
```

aaemnnosttv (commented):
Thanks @kuasha420! This is looking good, just a few things to address and then this should be good to go. Let me know if you have any questions 👍

tests/phpunit/includes/Core/Modules/Module_With_Service_Entity_ContractTests.php
```diff
@@ -0,0 +1,71 @@
...
+
+		$testcase = $this->get_testcase();
+		$module   = $this->get_module_with_service_entity();
+
+		$this::setup_test_check_service_entity_access( $module, 401 );
+
+		$testcase->assertWPError( $module->check_service_entity_access() );
+	}
+
+	public static function setup_test_check_service_entity_access( Module $module, $status_code ) {
```
aaemnnosttv, line 58: Having this as a static method makes it a bit harder to follow. It would be a bit easier to separate the mocking from the module set up.

aaemnnosttv (commented):
This is almost good to go. I'll push a few finishing touches to the tests but great work!

aaemnnosttv (approved):
Great!

aaemnnosttv:
PHP test failure on 5.6 is related to a failure to load some PHP module unrelated to this PR.

Issue #4702, opened 2022-01-21
Remove Analytics tag permission related components and infra


aaemnnosttv:
## Feature Description

The biggest change for Analytics as part of the simplification of existing tags is the removal of behavior based on a user's access to a detected existing tag. Accordingly all such behavior and infrastructure related to checking access to an existing tag (property) can be removed.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All components, hooks, selectors, actions, utilities, tests, module REST endpoints, and conditions related to **tag permission** for the Analytics module should be removed

**The PR for this issue should target the [`feature/existing-tag-simplification`](https://github.com/google/site-kit-wp/tree/feature/existing-tag-simplification) branch**

## Implementation Brief

**For the Analytics module only**

* Remove all use of `hasExistingTagPermission` and `hasTagPermission` and the use of `ExistingTagError` and `ExistingGTMPropertyError` components in components, removing any conditions that would render an `ExistingTagError` or `ExistingGTMPropertyError` entirely
	* Also render `StoreErrorNotices` in `SettingsView` unconditionally
	* Also remove the `ExistingTagError`, `ExistingGTMPropertyError` component files
* Remove use of `hasExistingTagPermission` and `hasTagPermission` from `validateCanSubmitChanges`, removing these checks entirely
* Remove logic around existing tags from `getAccounts` resolver entirely, i.e. no longer pass the two params `existingPropertyID` and `existingAccountID`, and remove support for those two parameters from the `GET:accounts-properties-profiles` and `GET:properties-profiles` REST endpoints as well
* Remove all use of `getTagPermission` from `useExistingTagEffect` and drastically simplify it so that all that it does going forward is disable the plugin snippet if the existing tag is the same as the currently configured property ID
* Remove `hasExistingTagPermission`, `hasTagPermission`, and `getTagPermission` store selectors, also the related fixtures
* Remove the `GET:tag-permission` REST endpoint (including the `has_access_to_property` method only used there)

**Storybook**

* Update Analytics setup and settings stories related to existing tags specific to a tag "without access" (i.e. without permissions) to properly reflect the correct non-blocking state
	* Since there will no longer be a concept of an existing tag with/without access these stories may be removed if duplicates of existing stories

### Test Coverage

* Related test coverage should be removed and updated accordingly
[...]

PR #4835, opened 2022-02-10
Remove existing tag code from Analytics

tofumatt:
## Summary

Addresses issue:

- #4702

I'm spinning my wheels a bit on the last failing E2E test (ignore the AMP one as I think it's just a timeout): https://github.com/google/site-kit-wp/runs/5223116819?check_suite_focus=true#step:14:249

I've not made any changes that I can see to that scenario, and looking at it in Puppeteer the relevant screen appears and the "re-fetch" button is clicked. Even when I ran it in interactive mode, the request wasn't fired. Normally I'd think that was because I messed up a component that does the re-fetching… but I haven't touched those components at all from what I can tell. 🤔 

I'll take another look tomorrow with fresh eyes, but I've spent a lot of time trying to debug that and feel like I might just be missing something obvious as there's a lot of code/changes in this PR.

Aside from that, though, I think this is all set for review and seems to work well.

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [x] Run the code.
- [x] Ensure CI checks pass.

## Merge Reviewer Checklist


tofumatt:
Yeah, I mentioned the failing test in the PR summary—I went through it and can't see what's changed/why it's failing; I've sunk a lot of time into it but nothing jumps out at me, so I figured it'd be good to have someone else take a look 😅 

If it's still failing, could you take a look or see if something in the PR is the obvious cause of the failure? I've not changed  code related to that failing test as far as I know, so I'm a bit perplexed 🤔

techanvil:
> Yeah, I mentioned the failing test in the PR summary—I went through it and can't see what's changed/why it's failing; I've sunk a lot of time into it but nothing jumps out at me, so I figured it'd be good to have someone else take a look sweat_smile
> 
> If it's still failing, could you take a look or see if something in the PR is the obvious cause of the failure? I've not changed code related to that failing test as far as I know, so I'm a bit perplexed thinking

Thanks @tofumatt, and sorry for not spotting your comment in the PR summary. I've taken a look, and, although it took a while, I have found the issue (a one-liner, as can be expected!). I've pushed the fix in this [commit](https://github.com/google/site-kit-wp/pull/4835/commits/44587e47fae78619ce8844d9a1c96f2df0247b9d).

assets/js/modules/analytics/datastore/accounts.js
```diff
@@ -56,6 +56,7 @@ const fetchGetAccountsPropertiesProfilesStore = createFetchStore( {
 			'modules',
 			'analytics',
 			'accounts-properties-profiles',
+			null,
```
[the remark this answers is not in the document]
techanvil: No problem! I must say, this is one of those bugs where I do miss working with TypeScript...

assets/js/modules/analytics/hooks/useExistingTagEffect.js
```diff
@@ -28,97 +28,20 @@ import { useEffect } from '@wordpress/element';
...
-		return data;
-	} );
+	const hasExistingTag = useSelect( ( select ) =>
+		select( MODULES_ANALYTICS ).hasExistingTag()
+	);
 
 	useEffect( () => {
-		if ( existingTag ) {
-			// Disable the plugin snippet
+		if ( hasExistingTag ) {
```
techanvil, line 41: Reading this point in the IB, it seems to suggest that `useExistingTagEffect` should disable the snippet if there is an existing tag _and_ the ID is the same as the currently configured property ID. However, I can see that you have updated `UseSnippetSwitch` to contain this logic. This seems like it's probably the right thing to do, and also the IB is a little ambiguous in its phrasing. So, I just want to highlight this and make sure this is the right way to go about things here.

> Remove all use of getTagPermission from useExistingTagEffect and drastically simplify it so that all that it does going forward is disable the plugin snippet if the existing tag is the same as the currently configured property ID

techanvil:
Hey @tofumatt, I have noticed this PR is targeting `develop` but it should in fact be targeting `feature/existing-tag-simplification`, and as such will likely need a rebase onto the feature branch too.

techanvil:
I've noticed that some of the related Storybook stories are now in a bit of an unusable state, however they should be reassessed once this branch has been rebased onto `feature/existing-tag-simplification`. These stories have both the dropdowns and button disabled:

**Analytics Module -> Settings -> No Tag, GTM property w/ access**
image

**Analytics Module -> Settings -> No Tag, GTM property w/o access**
image

**Analytics Module -> Setup -> No Tag, GTM property w/ access**
image

**Analytics Module -> Setup -> No Tag, GTM property w/o access**
image

techanvil (changes requested):
Hi @tofumatt, nice work! This broadly looks good, with just a few issues that do/may need addressing, please see the included comments. Cheers

assets/js/modules/analytics/hooks/useExistingTagEffect.js
tofumatt: Yeah, I thought this was the best place to put it but I did notice that. Sorry I forgot to mention it here 👍🏻

includes/Modules/Analytics.php
```diff
@@ -700,25 +699,6 @@ function ( $metric_def ) {
...
-							array( 'status' => 400 )
-						);
-					}
-					$property_id = $data['propertyID'];
-					return array_merge(
-						array(
-							'accountID'  => '', // Set the accountID to be an empty string and let has_access_to_property handle determining actual ID.
-							'propertyID' => $property_id,
-						),
-						$this->has_access_to_property( $property_id )
```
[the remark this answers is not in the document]
tofumatt: Good catch, thanks! 👍🏻

assets/js/modules/analytics/hooks/useExistingTagEffect.js
techanvil: Makes sense, and no problem!

techanvil:
> I've noticed that some of the related Storybook stories are now in a bit of an unusable state, however they should be reassessed once this branch has been rebased onto `feature/existing-tag-simplification`. These stories have both the dropdowns and button disabled:

Good to see these are all sorted now!

techanvil (approved):
LGTM, nice one @tofumatt!

assets/js/components/notifications/SetupSuccessBannerNotification.js
```diff
@@ -169,10 +173,6 @@ function SetupSuccessBannerNotification() {
 				}
 			}
 
-			if ( serviceSetupV2Enabled ) {
-				winData.description = '';
-			}
```
felixarntz, line 174: The changes in this file seem to have gotten here accidentally? They seem unrelated.

assets/js/modules/analytics/components/common/UseSnippetSwitch.js
```diff
@@ -52,7 +63,13 @@ export default function UseSnippetSwitch() {
 		);
 	}, [ useSnippet, setUseSnippet, viewContext ] );
 
-	if ( useSnippet === undefined ) {
+	useEffect( () => {
+		if ( toggleSwitchDisabled ) {
+			setUseSnippet( false );
+		}
+	}, [ setUseSnippet, toggleSwitchDisabled ] );
```
felixarntz, line 70: Can you explain this change? Why do we set the snippet to `false` here? That seems new and isn't specifically mentioned in the ACs (or IB) as far as I can tell.

assets/js/modules/analytics/components/common/UseSnippetSwitch.js
```diff
@@ -66,7 +83,7 @@ export default function UseSnippetSwitch() {
 				checked={ useSnippet }
 				onClick={ onChange }
 				hideLabel={ false }
-				disabled={ ! canUseSnippet }
+				disabled={ ! canUseSnippet || toggleSwitchDisabled }
```
felixarntz, line 86: See above. It seems unnecessary to disable the snippet control here? In the setup we currently don't show it all right? So while we need to set `useSnippet` to `false` during setup if there's an existing tag, I'm not sure why it's necessary to disable the toggle. Can you clarify? Maybe I missed something from another conversation.

assets/js/modules/analytics/components/dashboard/DashboardAllTrafficWidget/DimensionTabs.js
```diff
@@ -143,5 +149,6 @@ export default function DimensionTabs( { dimensionName, loaded } ) {
 
 DimensionTabs.propTypes = {
 	dimensionName: PropTypes.string.isRequired,
+	gatheringData: PropTypes.bool,
```
felixarntz, line 152: The changes in this file seem unrelated?

assets/js/modules/analytics/hooks/useExistingTagEffect.js
felixarntz: I'd like to follow up on this thread, as @techanvil mentioned above, the `useSnippet` setting should indeed only be set to `false` if the existing tag is also the same as the currently set property ID. So it looks to me that this hook here needs that update still.

Regarding doing that in `UseSnippetSwitch`, that seems problematic to me since that component isn't necessarily used, for example we only display it in Analytics settings right now, but not in the Analytics setup. The logic to conditionally hide the snippet should apply in either case though.

felixarntz (commented):
@tofumatt I left a few comments for this PR, but it also seems that there are a bunch of unrelated changes included here. Can you double-check the base branch and your branch to make sure only the relevant changes are included?

assets/js/modules/analytics/components/common/UseSnippetSwitch.js
tofumatt: Yeah, this wasn't mentioned in the ACs, but if the switch to use our snippet is disabled, I thought it best to disable our snippet. Otherwise, if the snippet were disabled but the setting was previously set to `true`, then the user would be forced to output our snippet _and_ the existing snippet, with no way to disable ours because the snippet setting cannot be changed.

But actually I think this needs some further thought, you're right 👍🏻

tofumatt:
Thanks for that review @felixarntz. I've actually updated the code so now the snippet is only disabled when the existing tag matches the propertyID set. I've added a test for this as well.

Because I made this change in the `useExistingTagEffect` hook, it works everywhere. The snippet code is set not to insert by virtue of setting the setting during setup. 👍🏻

assets/js/modules/analytics/hooks/useExistingTagEffect.js
```diff
@@ -28,97 +28,40 @@ import { useEffect } from '@wordpress/element';
...
-			// GTM container has GA tag and user has access to it, force select it.
-			setAccountID( gtmAnalyticsAccountID );
-			selectProperty( gtmAnalyticsPropertyID );
-			// We no longer set `useSnippet` to false here as this is handled using a filter
-			// on the server side now so that Analytics will take over the snippet again
-			// if Tag Manager or its snippet is disabled.
+			// Disable the Analytics snippet if there is an existing tag that
+			// matches the currently selected property.
+			setUseSnippet( false );
+			setCanUseSnippet( false );
```
felixarntz, line 58: Why the `setCanUseSnippet` here? That is a different option to enforce the value of `useSnippet`. We don't set that today, and it shouldn't happen here either. For example, if the `UseSnippetSwitch` toggle is present in the UI, it shouldn't be disabled in such a case, we should only change its value (via `setUseSnippet`).

felixarntz (commented):
@tofumatt A few follow-up comments here: We shouldn't need to touch `canUseSnippet`. Also, since the snippet placement now depends on the property selected, we also need to ensure we consider the case where the user switches from a property that matches the existing tag to a property that doesn't (in which case the snippet should be placed again).

assets/js/modules/analytics/hooks/useExistingTagEffect.js
tofumatt: Ah, okay, I thought we would disable it only _if_ the property matched. Instead we'll just disable it, sounds good 👍🏻

Issue #4673, opened 2022-01-19
Add a hasZeroData selector for Analytics


tofumatt:
## Feature Description

For the new Zero States features to be properly implemented, we need a way to explicitly check if a module has "zero data" as opposed to "gathering data". Obviously any site that's gathering data will also have "zero data", but we can use logic tailored to each "module" (eg Google Service) that Site Kit uses to check if there's no data available from the Google API because the site is new or if there's data because there's been nothing of note.

Essentially, we want to be able to tell the difference in Analytics: "Do I have no data in my Analytics report because my site is very new/my Analytics account is very new or is there is no data because my site is not popular/has received no visits?"

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* There should be a `hasZeroData()` selector for the Analytics module's datastore that checks to see if there is no data for Analytics despite `isGatheringData()` returning `false`.
* This selector should `return false` if any data is found. The selector should `return true` if no data for a "representative" Analytics request—likely whatever the default dashboard request would be (eg. 28 days)—is found but `isGatheringData() === false`.

## Implementation Brief

- In `assets/js/modules/analytics/datastore/report.js`:
    - Add a new selector `hasZeroData` that checks for zero data state as described in the AC
    - The `hasZeroData` selector will have the following logic:
        - Get the report data from `select( MODULES_ANALYTICS ).getReport( args )` selector:
        - Reuse the logic for getting the `args` to pass to `getReport( args )`  from `isGatheringData`, see:
        https://github.com/google/site-kit-wp/blob/27b8e040ec85ab9fb94efebebf44defc544a26d3/assets/js/modules/analytics/datastore/report.js#L296-L315
        - Return `undefined` if the `report` data is `undefined`
        - Get the gathering data state from the `isGatheringData()` selector
        - Get whether the report data has zero state or not, by using the `isZeroReport( report )` helper function from `assets/js/modules/analytics/util/is-zero-report.js`
        - Return `undefined` if either `isGatheringData() === undefined || getReport() === undefined`
        - Return `false` if `isGatheringData() === false && isZeroReport() === false`
        - Return `true` otherwise (essentially, if `isGatheringData() === true || isZeroReport() === true`). Note, this can be the default return value, to ensure we always return from the function

### Test Coverage

- Add test cases for the `hasZeroData` selector

## QA Brief

**QA:Eng**
- Verify the changes met AC and IB
[...]

PR #4848, opened 2022-02-15
Infrastructure/4673 - Add `hasZeroData` selector for Analytics

hussain-t:
## Summary

Addresses issue:

- #4673 

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [x] Run the code.
- [x] Ensure CI checks pass.

## Merge Reviewer Checklist


hussain-t:
Thanks @techanvil

assets/js/modules/analytics/datastore/report.test.js
```diff
@@ -31,6 +31,7 @@ import {
 import * as fixtures from './__fixtures__';
 
 describe( 'modules/analytics report', () => {
+	const analyticsRegexp = /^\/google-site-kit\/v1\/modules\/analytics\/data\/report/;
```
tofumatt, line 34:
```suggestion
	const analyticsRegexp = /^\/google-site-kit\/v1\/modules\/analytics\/data\/report/;
```

I know this reduces duplication in the code, but I actually think this is an anti-pattern for tests.

In a component or a file this kind of constant would be handy, allowing us to consolidate code and use this request regex throughout the file without typing it much.

But for a test, it is _much_ more readable to have a test and as much of its data as possible be readable within the test spec's function. Granted, our fixtures require looking at them in a different file (and in modern editors like VSCode getting to this regex is an `Alt` + Click away), but having the URL expected right in the test can help when diagnosing failures and requires less mental energy to inspect/debug/reason about.

The aim of a test spec and test files in general should not be DRY ("don't repeat yourself"). The test should be as illustrative as possible when it fails, and including as much info as close to the test as possible allows for this. Yes, it makes _test refactoring_ more work, but it makes _code refactoring less work_, which is at least part of the benefits of good tests.

So I think this should be removed and these regex's placed back into the test functions themselves. Yes, it's more verbose and repetitive, but it means you can easily, on a screen, get a good sense of what's happening in the test, which URLs we're mocking, etc.

assets/js/modules/analytics/datastore/report.test.js
```diff
@@ -425,5 +419,72 @@ describe( 'modules/analytics report', () => {
...
+					.hasZeroData();
+
+				expect( hasZeroData ).toBeUndefined();
+			} );
+
+			it.each( [
+				[ 'an empty array', [ { data: { rows: [] } } ] ],
+				[ 'null', null ],
+			] )(
+				'should return TRUE if report data in isGatheringData OR isZeroReport is %s',
```
tofumatt, line 438: Where in this test is `isGatheringData` vs `isZeroReport` being tested?

It looks like you're testing `isGatheringData` with the `null` report and `isZeroReport` with the empty array. Is that correct?

If so, these should be split up into two separate tests with the relevant result for each one.

Again, see my above comment about DRY in tests; it's tempting to use this to test both options with fewer lines of code, but minimising LOC in a test file is counter-intuitive, because when this test fails it won't be as clear why/which is failing. Doubly-so because the test spec names which selector is being tested in opposite order to the `it.each` array, I think.

assets/js/modules/analytics/datastore/report.test.js
```diff
@@ -425,5 +419,72 @@ describe( 'modules/analytics report', () => {
...
+
+					const zeroData = registry
+						.select( MODULES_ANALYTICS )
+						.hasZeroData();
+
+					expect( zeroData ).toBe( true );
+				}
+			);
+
+			it( 'should return FALSE if isGatheringData and isZeroReport return FALSE', async () => {
```
tofumatt, line 464: Same for these tests as for the above one; I think they're actually two different tests and should be split into two tests.

tofumatt (changes requested):
This looks good, but I think the test code needs some work to be more maintainable/easier-to-use. I left some comments/questions but the approach itself in the code looks right, just need to tidy up these tests. 👍🏻

assets/js/modules/analytics/datastore/report.test.js
hussain-t: Agreed 👍

hussain-t: >It looks like you're testing isGatheringData with the null report and isZeroReport with the empty array. Is that correct?
Yes.

assets/js/modules/analytics/datastore/report.test.js
```diff
@@ -425,5 +447,147 @@ describe( 'modules/analytics report', () => {
...
+				);
+
+				const zeroData = registry
+					.select( MODULES_ANALYTICS )
+					.hasZeroData();
+
+				expect( zeroData ).toBe( true );
+			} );
+
+			it( 'should return FALSE if isGatheringData return FALSE', async () => {
```
tofumatt, line 526:
```suggestion
			it( 'should return FALSE if isGatheringData returns FALSE', async () => {
```

assets/js/modules/analytics/datastore/report.test.js
```diff
@@ -425,5 +447,147 @@ describe( 'modules/analytics report', () => {
...
+				] );
+
+				const noZeroData = registry
+					.select( MODULES_ANALYTICS )
+					.hasZeroData();
+
+				expect( noZeroData ).toBe( false );
+			} );
+
+			it( 'should return FALSE if isZeroReport return FALSE', async () => {
```
tofumatt, line 564:
```suggestion
			it( 'should return FALSE if isZeroReport returns FALSE', async () => {
```

tofumatt (approved):
Looks good, just a few small tweaks so I'll commit them then merge 👍🏻

tofumatt:
It seems like 4.7 and 4.9 tests are failing across the repo so it's not this change. I'll merge then.

Issue #4858, opened 2022-02-16
Provide additional context data with remote-controlled features request


felixarntz:
Following #4856, additional high-level criteria should be included in the `/site-management/features/` request, to allow better heuristics for whether to grant a feature for a site.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* In addition to the existing body parameters `platform` and `version`, the following parameters should be sent with the `/site-management/features/` request:
    * `platform_version`: The WordPress version used.
    * `active_modules`: Space-separated list of the slugs of all active modules (including the force-active modules), sorted by the modules' `order`.
    * `connected_modules`: Space-separated list of the slugs of all (active and) connected modules (including the force-active modules), sorted by the modules' `order`.
    * `user_count`: Number of all WordPress user accounts on the site.
    * `connectable_user_count`: Number of WordPress user accounts on the site with permission to connect (i.e. currently that is the number of administrators).
    * `connected_user_count`: Number of WordPress user accounts on the site that are actually connected (i.e. that are authenticated/have an access token).

## Implementation Brief

* Update the `Google_Proxy::get_features()` method:
  * Create a new variable `$body` and assign an empty array to it.
  * Pass the $body array through the `googlesitekit_features_request_data` filter.
  * Assign the existing body properties to the `$body` array.
  * Add the `platform_version` property to the body array with the global `$wp_version` variable as value.
  * Get the number of registered users and pass it as the `user_count` property to the body array.
  * Get the number of administrators and pass it as the `connectable_user_count` property to the body array.
  * Get the number of connected users (those who have `OAuth_Client_Base::OPTION_ACCESS_TOKEN` meta) and pass it as the `connected_user_count` property to the body array. Please, use the `User_Options::get_meta_key` method to property prefix the meta key.
  * Set the $body array as the actual `body` property of the request.
* Add a hook for the `googlesitekit_features_request_data` filter to the `Modules` class that adds `active_modules` and `connected_modules` information to the filtered array.
  * `active_modules` should be a space-separated list of the slugs of all active modules (including the force-active modules), sorted by the modules' `order`.
    * `connected_modules` should be a space-separated list of the slugs of all (active and) connected modules (including the force-active modules), sorted by the modules' `order`. 

### Test Coverage

* Update the `Google_ProxyTest::test_get_features` test to verify new request params.

### QA Eng
[...]

PR #4942, opened 2022-03-14
Add more context data with getFeatures request.

jimmymadon:
## Summary

Addresses issue:

- #4858

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist

### DIFF
--- a/includes/Core/Authentication/Authentication.php
+++ b/includes/Core/Authentication/Authentication.php
@@ -1390,9 +1390,9 @@ final class Authentication {
 		}
 
 		// The experimental features (ideaHubModule and swgModule) are checked within Modules::construct() which
-		// runs before Modules::register() where the `googlesitekit_features_request_data` filter is run. Without
-		// this filter, some necessary context data is not sent with a request to Google_Proxy::get_features().
-		// Hence we skip making this request and solely check the database to see if these features are enabled.
+		// runs before Modules::register() where the `googlesitekit_features_request_data` filter is registered.
+		// Without this filter, some necessary context data is not sent when a request to Google_Proxy::get_features()
+		// is made. So we skip making this request and solely check the database to see if these features are enabled.
 		if ( in_array( $feature_name, array( 'ideaHubModule', 'swgModule' ), true ) ) {
 			$features = $this->transients->get( $transient_name );
 			if ( false === $features ) {
--- a/includes/Core/Authentication/Authentication.php
+++ b/includes/Core/Authentication/Authentication.php
@@ -1409,7 +1409,7 @@ final class Authentication {
 
 		$features = $this->transients->get( $transient_name );
 		if ( false === $features ) {
-			$features = $this->google_proxy->get_features( $this->credentials );
+			$features = $this->google_proxy->get_features( $this->credentials, $this->get_oauth_client() );
 			if ( is_wp_error( $features ) ) {
 				$this->transients->set( $transient_name, array(), HOUR_IN_SECONDS );
 			} else {
--- a/includes/Core/Authentication/Authentication.php
+++ b/includes/Core/Authentication/Authentication.php
@@ -1399,9 +1399,6 @@ final class Authentication {
 			if ( false === $features ) {
 				$active_modules = $this->options->get( Modules::OPTION_ACTIVE_MODULES );
 				if ( ! is_array( $active_modules ) ) {
-					$active_modules = $this->options->get( 'googlesitekit-active-modules' );
-				}
-				if ( ! is_array( $active_modules ) ) {
 					return false;
 				}
 				if ( 'ideaHubModule' === $feature_name ) {
--- a/includes/Core/Authentication/Authentication.php
+++ b/includes/Core/Authentication/Authentication.php
@@ -31,6 +31,8 @@ use WP_REST_Response;
 use Exception;
 use Google\Site_Kit\Core\Modules\Modules;
 use Google\Site_Kit\Core\Util\BC_Functions;
+use Google\Site_Kit\Modules\Idea_Hub;
+use Google\Site_Kit\Modules\Subscribe_With_Google;
 
 /**
  * Authentication Class.
@@ -1394,7 +1396,11 @@ final class Authentication {
 			return $feature_enabled;
 		}
 
-		$features = $this->get_transient_features();
+		if ( in_array( $feature_name, array( 'ideaHubModule', 'swgModule' ), true ) ) {
+			$features = $this->get_transient_features( $feature_name );
+		} else {
+			$features = $this->get_transient_features();
+		}
 
 		if ( ! is_wp_error( $features ) && isset( $features[ $feature_name ]['enabled'] ) ) {
 			return filter_var( $features[ $feature_name ]['enabled'], FILTER_VALIDATE_BOOLEAN );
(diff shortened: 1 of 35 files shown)


includes/Core/Modules/Modules.php
```diff
@@ -378,6 +378,23 @@ function( $setting, $setting_key ) use ( $old_values, $module_slug, &$changed_se
 			10,
 			2
 		);
+
+		add_filter(
+			'googlesitekit_features_request_data',
+			function( $body ) {
+				$active_modules    = $this->sort_modules( $this->get_active_modules() );
```
felixarntz, line 385: Maybe it would be better to add the sorting right into the `get_active_modules()` method? I think it's a reasonable expectation that the return values of that method should be sorted already, similar to how the `get_available_modules()` return values are already sorted.

That makes the code easier to use as you don't have to remember to think about order everywhere, it just works out of the box.

You could change the `get_active_modules()` implementation to use a single `array_filter` call on the available modules instead of `array_merge` with two arrays, this way you wouldn't even need to sort again since the available modules are already sorted correctly.

felixarntz (changes requested):
@jimmymadon Left a few points of feedback on the implementation here while reviewing this PR following your question in https://github.com/google/site-kit-wp/issues/4858#issuecomment-1066896017.

includes/Core/Modules/Modules.php
jimmymadon: Beautiful - all _sorted_! 😃

includes/Core/Authentication/Google_Proxy.php
```diff
@@ -585,21 +586,29 @@ public function exchange_site_code( $site_code, $undelegated_code ) {
...
-		return $this->request(
-			self::FEATURES_URI,
-			$credentials,
-			array(
-				'body' => array(
-					'platform' => $platform . '/google-site-kit',
-					'version'  => GOOGLESITEKIT_VERSION,
-				),
-			)
+	public function get_features( Credentials $credentials, OAuth_Client $oauth_client ) {
```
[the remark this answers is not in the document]
jimmymadon: Agreed! I didn't want to add `User_Options` (needed to `get_meta_key()`) as a dependency to `Google_Proxy`. Also, the function is quite generic - so wanted to move it to either `Authentication` or `OAuth_Client` which both have User_Options and deal with 'connected users'. @eugene-manuilov and I were already discussing about the best place to put this and he was going to have a proper look in CR. 

Using a filter definitely sounds perfect. I think we can move the method to `Authentication` and call it in a filter within `register()` as it happens very early on? Going to update the PR - let me know if this is fine?

The only downside to having it as a separate method is we might have to consider caching this query if this method is used outside `get_transient_features()` which is technically caching anyways.

includes/Core/Authentication/Authentication.php
```diff
@@ -1401,14 +1415,34 @@ private function filter_features_via_proxy( $feature_enabled, $feature_name ) {
...
+					return $features;
+				}
+				if ( 'ideaHubModule' === $feature_name ) {
+					$features['ideaHubModule']['enabled'] = in_array( Idea_Hub::MODULE_SLUG, $active_modules, true );
+				}
+				if ( 'swgModule' === $feature_name ) {
+					$features['swgModule']['enabled'] = in_array( Subscribe_With_Google::MODULE_SLUG, $active_modules, true );
+				}
+				return $features;
+			}
```
felixarntz, line 1444: I think handling this feature-specific behavior in this method is particularly hacky, that's why my original proposal was to do it in the `filter_features_via_proxy` method. We can do it in _only that_ method, without modifying this private method here, which makes the hack less problematic since at least it's only in one place.

I think it would be cleaner to add the same `$this->transients->get` call to `filter_features_via_proxy` in case of `ideaHubModule` and `swgModule`, That is not an expensive call, so it's fine especially since it's only for these two modules. From there we then already know if the transient is set or not, so we can put this entire block of logic in only `filter_features_via_proxy`, leaving `get_transient_features` clean (especially the additional parameter here is problematic, since that doesn't only make the method internals "hacky", but also the method signature).

includes/Core/Authentication/Authentication.php
```diff
@@ -1386,7 +1396,11 @@ private function filter_features_via_proxy( $feature_enabled, $feature_name ) {
 			return $feature_enabled;
 		}
 
-		$features = $this->get_transient_features();
+		if ( in_array( $feature_name, array( 'ideaHubModule', 'swgModule' ), true ) ) {
+			$features = $this->get_transient_features( $feature_name );
+		} else {
+			$features = $this->get_transient_features();
+		}
```
felixarntz, line 1403: See above, the entire "hacky" logic should be moved here, and we can leave `get_transient_features` as is then.

includes/Core/Modules/Modules.php
```diff
@@ -391,6 +408,27 @@ public function get_module_sharing_settings() {
...
+			$modules,
+			function( Module $a, Module $b ) {
+				if ( $a->order === $b->order ) {
+					return 0;
+				}
+				return ( $a->order < $b->order ) ? -1 : 1;
+			}
+		);
+		return $modules;
+	}
```
felixarntz, line 430: Now that this is only used in one place inside the class, let's either remove this method or make it private, since it shouldn't be necessary to call it from the outside.

includes/Core/Authentication/Authentication.php
jimmymadon: Sorry I missed making an additional call to $this->transients->get() within filter_features_via_proxy() in your previous comment.

includes/Core/Authentication/Authentication.php
```diff
@@ -373,6 +375,14 @@ function( $user ) {
 			}
 		);
 
+		add_filter(
+			'googlesitekit_features_request_data',
+			function( $body ) {
+				$body['connected_user_count'] = $this->count_connected_users();
```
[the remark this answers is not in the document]
jimmymadon: True, I agree that not having a separate filter will keep things together and simpler. However, in addition to needing a new `User_Options()` instance, I thought `count_connected_users()` sits better as a reusable public method within `Authentication` instead of in `Google_Proxy`. But I have moved the method into `GoogleProxy` for now.

includes/Core/Authentication/Authentication.php
```diff
@@ -1386,6 +1389,29 @@ private function filter_features_via_proxy( $feature_enabled, $feature_name ) {
...
+		// runs before Modules::register() where the `googlesitekit_features_request_data` filter is registered.
+		// Without this filter, some necessary context data is not sent when a request to Google_Proxy::get_features()
+		// is made. So we skip making this request and solely check the database to see if these features are enabled.
+		if ( in_array( $feature_name, array( 'ideaHubModule', 'swgModule' ), true ) ) {
+			$features = $this->transients->get( $transient_name );
+			if ( false === $features ) {
+				$active_modules = $this->options->get( Modules::OPTION_ACTIVE_MODULES );
+				if ( ! is_array( $active_modules ) ) {
+					$active_modules = $this->options->get( 'googlesitekit-active-modules' );
+				}
```
felixarntz, line 1402: Good catch to include this here, but I think we can get rid of it to have less duplicate code here, since when this old option was being used, neither `ideaHubModule` nor `swgModule` existed anyways.

I'm not even opposed to removing the same line from the `Modules` class itself maybe, but especially here we don't need to support it.

includes/Core/Authentication/Authentication.php
```diff
@@ -1386,6 +1389,29 @@ private function filter_features_via_proxy( $feature_enabled, $feature_name ) {
 			return $feature_enabled;
 		}
 
+		// The experimental features (ideaHubModule and swgModule) are checked within Modules::construct() which
+		// runs before Modules::register() where the `googlesitekit_features_request_data` filter is registered.
+		// Without this filter, some necessary context data is not sent when a request to Google_Proxy::get_features()
+		// is made. So we skip making this request and solely check the database to see if these features are enabled.
```
felixarntz, line 1395: Maybe worth briefly explaining the `$this->transients->get()` call in the code here as well: Since a request is only being made if the transient isn't set, we only need to avoid _that_ scenario. If the data is already cached in a transient, we still need to use the data from that, i.e. fall through to `$this->get_transient_features()`.

felixarntz (approved):
@jimmymadon Great work!

Just two minor comments here. Nothing blocking, but would be great if you could review them and address (with feedback or code 😄).

includes/Core/Authentication/Authentication.php
```diff
@@ -1386,6 +1389,27 @@ private function filter_features_via_proxy( $feature_enabled, $feature_name ) {
...
+				if ( ! is_array( $active_modules ) ) {
+					return false;
+				}
+				if ( 'ideaHubModule' === $feature_name ) {
+					return in_array( Idea_Hub::MODULE_SLUG, $active_modules, true );
+				}
+				if ( 'swgModule' === $feature_name ) {
+					return in_array( Subscribe_With_Google::MODULE_SLUG, $active_modules, true );
+				}
+			}
```
tofumatt, line 1410: This feels a bit cramped 😅 

```suggestion
			if ( false === $features ) {
				$active_modules = $this->options->get( Modules::OPTION_ACTIVE_MODULES );

				if ( ! is_array( $active_modules ) ) {
					return false;
				}

				if ( 'ideaHubModule' === $feature_name ) {
					return in_array( Idea_Hub::MODULE_SLUG, $active_modules, true );
				}

				if ( 'swgModule' === $feature_name ) {
					return in_array( Subscribe_With_Google::MODULE_SLUG, $active_modules, true );
				}
			}
```

tofumatt (approved):

Issue #4901, opened 2022-03-01
Remove serviceSetupV2 feature flag


eclarke1:
## Feature Description

Issue to remove the servicev2SetupV2 feature flag following rollout to 100% scheduled for March 14, 2022. Target date for removing the feature flag is March 28, 2022.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `serviceSetupV2` feature flag should be removed.
* All code paths that are currently run when the flag is not enabled should be removed, and all code structures (PHP classes and methods, JS components etc.) that are only used in such removed code should also be removed.
* The `Setup_V2` PHP class should be merged into its base class `Setup` so that going forward only that `Setup` class remains and is used (obviously no longer an abstract class then).
    * The PHPUnit tests should reflect that as well.
* The `Google_Proxy::SETUP_URI_V2` constant should be renamed to `Google_Proxy::SETUP_URI` (with the original `Google_Proxy::SETUP_URI` being deleted per the above).

## Implementation Brief

**Do not merge the PR for this before the 1.70.0 release has gone out, since this must wait for the 1.71.0 release.**

* Remove the `serviceSetupV2` feature flag from `feature-flags.json`
* Remove `useFeature( 'serviceSetupV2' )` from `SetupSuccessBannerNotification` and `SetupUsingProxy`. Remove all conditional checks for the flag and only render the component as if `serviceSetupV2` were enabled. Code paths where `serviceSetupV2Enabled === false` can be removed.
* Remove `createServiceSetupV2Variant` and usage from `SetupSuccessBannerNotification.stories.js`, as now all stories will default to using the V2 setup as it's the default after the above change to components. VRTs will likely need updating.
* Remove any `serviceSetupV2` feature flag setup in `SetupUsingProxy.stories.js`.
* Remove the `serviceSetupV2` feature flag usage in PHP files and remove any code paths where `Feature_Flags::enabled( 'serviceSetupV2' ) === false` in:
  * `includes/Plugin.php`
  * `includes/Core/Authentication/Authentication.php`
  * `includes/Core/Authentication/Google_Proxy.php`
  * `includes/Core/Authentication/Setup.php`
  * `includes/Core/Authentication/Clients/OAuth_Client.php`
  * `includes/Core/Modules/Module.php`
  * `includes/Modules/Analytics.php`
* Move everything from `Setup_V2` PHP class into its base class `Setup`. `Setup` no longer needs to be an abstract class.
[...]

PR #4950, opened 2022-03-16
Remove serviceSetupV2 feature flag.

tofumatt:
## Summary

Addresses issue:

- #4901

## Relevant technical choices

I've spent a wild amount of time trying to get this failing test working: https://github.com/google/site-kit-wp/pull/4950/files#diff-f60fee9e4d0635bf0b187460fdc3ae78bcc0ed97c1bd2f91e7256bc402b5616dR152

It seems the E2E plugins that should simulate a success OAuth sign-in aren't working for this one test. I thought all the other tests were maybe using the GCP sign-in flow, but that doesn't seem to be the case. But this one test always tries to authenticate against the real proxy and returns an error when I `var_dump()` the response:

image

I'm totally lost as the E2E plugins and OAuth code has been tough for me to follow and work backward through. Even using the site manually didn't help much.

If anyone has ideas that'd be great, but I think I need to step back from it as I'm just spinning my wheels. If fresh eyes can fix it that'd be great, but otherwise I've just skipped the tests after manually seeing that the User Input will still work after setup.

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist


aaemnnosttv (changes requested):
Thanks @tofumatt – this looks good overall although I noticed one place where the feature flag is still used (may be from the last merge). It's only used in a test but that test is also failing as a result. See `AuthenticationTest::test_filter_features_via_proxy`

There was also one place where a condition unrelated to the flag shouldn't have been removed; see below.

Regarding the issue with User Input + E2E, I'll take a closer look at that now but if we need to leave that skipped for the time being that should be ok since it isn't being rolled out yet anyways. It would be preferable of course to fix instead though 😄

includes/Core/Authentication/Authentication.php
```diff
@@ -1403,16 +1397,6 @@ private function filter_features_via_proxy( $feature_enabled, $feature_name ) {
 				}
 			}
 
-			if ( ! $this->credentials->has() ) {
```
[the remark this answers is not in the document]
tofumatt: Argh, yes, I think I did the merge wrong, thanks 👍🏻

includes/Core/Authentication/Authentication.php
```diff
@@ -1404,13 +1399,12 @@ private function filter_features_via_proxy( $feature_enabled, $feature_name ) {
...
 			if ( ! $this->credentials->has() ) {
-				// For the 'serviceSetupV2' feature, continue to check for the legacy persistent option which used to
-				// be set when remote features were cached as transients.
-				if ( 'serviceSetupV2' === $feature_name && $this->options->get( $service_setup_v2_option_name ) ) {
+				// Use a persistent option so that it remains active even if the
+				// site is not connected. This is crucial to provide a
+				// consistent setup flow experience.
+				if ( $this->options->get( $service_setup_v2_option_name ) ) {
 					return true;
 				}
```
aaemnnosttv, line 1407: I think the option is still no longer necessary here because the feature is on 100% of the time now. As is, this would enable all features if the user had this option set which is not what we want 😄 

The important thing to keep here is that we don't attempt to fetch remote features if there are no credentials as that request would fail.

I'm about to push the fixes for user input now so I'll fix this as well 👍

tofumatt: Cheers, I think after the `main` merge this code sorta confused me 😅

aaemnnosttv:
I fixed the E2E test that was being skipped before for user input. The problem wasn't that it was trying to make a request to the proxy (these are all blocked by a WP constant) but that because it was blocked, there was no `Redirect-To` header in the response which also triggers the same error message that was being shown. In order to provide this, I had to add an e2e plugin to stub the response for this request which made the previous client-side request interception no longer necessary as well. In reality it wasn't quite that simple to figure out and I ran into a really weird bug with the plugin activation utility which is still there (I was able to work around it by changing the order of plugins to activate for now), so this is also a note to address this in the future.

There are still a few E2E-related failures but these appear to be the AMP stability-related issues mentioned earlier. This should be good to go now 👍

aaemnnosttv (approved):
LGTM 👍

PR #4953, opened 2022-03-17
Enhance/4713 follow up

kuasha420:
## Summary

Addresses issue:

- #4713 

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [x] Run the code.
- [x] Ensure CI checks pass.

## Merge Reviewer Checklist

### DIFF
--- a/assets/js/modules/tagmanager/hooks/useExistingTagEffect.js
+++ b/assets/js/modules/tagmanager/hooks/useExistingTagEffect.js
@@ -41,11 +41,11 @@ export default function useExistingTagEffect() {
 	const { setUseSnippet } = useDispatch( MODULES_TAGMANAGER );
 
 	useEffect( () => {
-		if (
-			existingTag &&
-			containerID &&
-			containerID !== initialContainerID.current
-		) {
+		if ( initialContainerID.current ) {
+			initialContainerID.current = null;
+			return;
+		}
+		if ( existingTag && containerID ) {
 			if ( existingTag === containerID ) {
 				setUseSnippet( false );
 			} else {
--- a/assets/js/modules/tagmanager/hooks/useExistingTagEffect.js
+++ b/assets/js/modules/tagmanager/hooks/useExistingTagEffect.js
@@ -19,7 +19,7 @@
 /**
  * WordPress dependencies
  */
-import { useEffect } from '@wordpress/element';
+import { useEffect, useRef } from '@wordpress/element';
 
 /**
  * Internal dependencies
@@ -36,10 +36,16 @@ export default function useExistingTagEffect() {
 		select( MODULES_TAGMANAGER ).getContainerID()
 	);
 
+	const initialContainerID = useRef( containerID );
+
 	const { setUseSnippet } = useDispatch( MODULES_TAGMANAGER );
 
 	useEffect( () => {
-		if ( containerID && existingTag ) {
+		if (
+			existingTag &&
+			containerID &&
+			containerID !== initialContainerID.current
+		) {
 			if ( existingTag === containerID ) {
 				setUseSnippet( false );
 			} else {
--- a/assets/js/modules/tagmanager/hooks/useExistingTagEffect.js
+++ b/assets/js/modules/tagmanager/hooks/useExistingTagEffect.js
@@ -36,13 +36,13 @@ export default function useExistingTagEffect() {
 		select( MODULES_TAGMANAGER ).getContainerID()
 	);
 
-	const initialContainerID = useRef( containerID );
+	const skipEffect = useRef( !! containerID );
 
 	const { setUseSnippet } = useDispatch( MODULES_TAGMANAGER );
 
 	useEffect( () => {
-		if ( initialContainerID.current ) {
-			initialContainerID.current = null;
+		if ( skipEffect.current ) {
+			skipEffect.current = false;
 			return;
 		}
 		if ( existingTag && containerID ) {
--- a/assets/js/modules/tagmanager/hooks/useExistingTagEffect.test.js
+++ b/assets/js/modules/tagmanager/hooks/useExistingTagEffect.test.js
@@ -40,7 +40,7 @@ describe( 'useExistingTagEffect', () => {
 		// eslint-disable-next-line sitekit/acronym-case
 		const accountID = account.accountId;
 		const containers = factories.buildContainers(
-			3,
+			2,
 			// eslint-disable-next-line sitekit/acronym-case
 			{ accountId: account.accountId, usageContext: [ CONTEXT_WEB ] }
 		);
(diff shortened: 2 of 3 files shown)


techanvil:
Hey @kuasha420, I've tested out your followup PR and can confirm it does fix the issue of the switch not displaying the correct setting for the current snippet, as described in your [comment](https://github.com/google/site-kit-wp/issues/4713#issuecomment-1064822505).

> Setup with Same Container -> Snippet is not inserted by default -> Settings -> Enable Snippet -> Snippet is inserted -> Edit Settings -> Snippet toggle goes back to Disabled (Submit button Enabled) -> Turn the Snippet On -> Submit button becomes deactivated.
>
> Setup with Different Container -> Snippet is inserted by default -> Settings -> Disable Snippet -> Snippet is not inserted -> Edit Settings -> Snippet toggle goes back to Enabled (Submit button Enabled) -> Turn the Snippet Off -> Submit button becomes deactivated.

There is a side effect though, whereby the switch will now persist the toggle state when changing the dropdown back to the current container. I've recorded a couple of video clips to illustrate:

When editing the settings with an existing snippet, the current container is preselected with the dropdown toggled on. Then when selecting a different container, toggling the snippet off, and selected the current container again, the snippet off state is persisted.

https://user-images.githubusercontent.com/18395600/159283450-57302403-5204-4f64-a5cb-4e7ead12d98c.mp4

However when selecting a different container, toggling the snippet off, and selecting another different container, the snippet state changes back to on.

https://user-images.githubusercontent.com/18395600/159283476-b50f461e-6144-4c8a-bb11-dcda18595458.mp4

This is a bit of an edge case, and I'm not sure if we need to address it. @felixarntz, please could you take a look and confirm whether this is OK to proceed with?

That aside, this all looks good, with just a couple of trivial comments from me on the test. Nice one!

assets/js/modules/tagmanager/hooks/useExistingTagEffect.test.js
```diff
@@ -99,4 +99,153 @@ describe( 'useExistingTagEffect', () => {
...
 	} );
+
+	it( 'does not set change the useSnippet value when there is already a container ID on page load (container ID is same as existing tag)', async () => {
+		const account = factories.accountBuilder();
+		// eslint-disable-next-line sitekit/acronym-case
+		const accountID = account.accountId;
+		const containers = factories.buildContainers(
+			2,
+			// eslint-disable-next-line sitekit/acronym-case
+			{ accountId: account.accountId, usageContext: [ CONTEXT_WEB ] }
```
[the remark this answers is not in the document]
kuasha420: `sitekit/acronym-case` screws us here. `accountId`. `accountID`, `account.accountId` all are present in this file with multiple cases of `eslint-ignore`-ing the error. Maybe we can enable the ignore file wide with an explanation and tidy up these assignments.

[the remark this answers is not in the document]
techanvil: Apologies for not spotting the difference in case there, and thanks for pointing this out!

kuasha420:
@felixarntz Good idea, but only slight difference is that on settings page load, the `containerID` is never `undefined` on page load but the currently saved one. From my understanding, we can't run the logic initially when the containerID is the same as the same one, but we should run it if someone changes to a different container ID and back to the initially saved one. 

That's why I think the solution is to ignore the initial render like this: 

```js
	const initialContainerID = useRef( containerID );

	useEffect( () => {
		if ( initialContainerID.current ) {
			initialContainerID.current = null;
			return;
		}
		if ( existingTag && containerID ) {
			if ( existingTag === containerID ) {
				setUseSnippet( false );
			} else {
				setUseSnippet( true );
			}
		}
	}, [ containerID, existingTag, setUseSnippet ] );
```

I have also updated the tests to consider scenarios that Tom raised in this comment. @techanvil can you have another look at the implementation and specially the test? 

Cheers. 

cc @felixarntz @aaemnnosttv

assets/js/modules/tagmanager/hooks/useExistingTagEffect.test.js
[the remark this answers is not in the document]
kuasha420: Cool. We do have [precedence](https://github.com/google/site-kit-wp/blob/52dc809d41a8f33f9c8d42218b4c3f68f538a2e0/assets/js/util/ip-cidr.js#L19-L24) of setting file-wide eslint ignore with an explanation to avoid repeating the ignore in every line. However, I will leave this one as is for now. Cheers.

assets/js/modules/tagmanager/hooks/useExistingTagEffect.test.js
```diff
@@ -99,4 +99,189 @@ describe( 'useExistingTagEffect', () => {
...
+		} );
+
+		expect( registry.select( MODULES_TAGMANAGER ).getUseSnippet() ).toBe(
+			false
+		);
+
+		act( () => {
+			// Set useSnippet to true to simulate pressing the toggle.
+			registry.dispatch( MODULES_TAGMANAGER ).setUseSnippet( true );
+			// Change to existing tag container. This should not change the useSnippet value.
```
techanvil, line 235: Same as above:

```suggestion
			// Change to existing tag container. This should change the useSnippet value.
```

assets/js/modules/tagmanager/hooks/useExistingTagEffect.test.js
```diff
@@ -99,4 +99,189 @@ describe( 'useExistingTagEffect', () => {
...
+		} );
+
+		expect( registry.select( MODULES_TAGMANAGER ).getUseSnippet() ).toBe(
+			true
+		);
+
+		act( () => {
+			// Set useSnippet to false to simulate pressing the toggle.
+			registry.dispatch( MODULES_TAGMANAGER ).setUseSnippet( false );
+			// Change to another container. This should not change the useSnippet value.
```
techanvil, line 150: Seeing as the snippet has just been set to `true`, and is tested to be `false` at the next `expect`, this comment is a little confusing, maybe the "not" should be removed?

```suggestion
			// Change to existing tag container. This should change the useSnippet value.
```

assets/js/modules/tagmanager/hooks/useExistingTagEffect.js
```diff
@@ -36,10 +36,16 @@ export default function useExistingTagEffect() {
 		select( MODULES_TAGMANAGER ).getContainerID()
 	);
 
+	const initialContainerID = useRef( containerID );
```
techanvil, line 39: It might be a little clearer to rename this and make it a boolean:

```js
	const skipEffect = useRef( !! containerID );
	
	// ...
	
	useEffect( () => {
		if ( skipEffect.current ) {
			skipEffect.current = false;
			return;
		}
```

techanvil (changes requested):

techanvil:
> I have also updated the tests to consider scenarios that Tom raised in this comment. @techanvil can you have another look at the implementation and specially the test?

Thanks @kuasha420, this LGTM, with just a couple of minor comments to consider.

assets/js/modules/tagmanager/hooks/useExistingTagEffect.test.js
kuasha420: +1. Updated.

techanvil (approved):
LGTM, nice one @kuasha420!

felixarntz:
@kuasha420 
> Good idea, but only slight difference is that on settings page load, the `containerID` is never `undefined` on page load but the currently saved one.

Hmm initially the API-based selectors should always return `undefined` at first, at least for one "cycle", even when API data is preloaded, because only that will trigger the resolver to populate the data. Isn't that the case anymore? cc @techanvil @tofumatt 

Potentially the issue here is that the `useEffect` doesn't fire when the values are still `undefined`? Would be great to hear from one of you folks more savvy in React than myself. I think the approach eventually taken here makes sense, I'd just like to understand this better :)

kuasha420:
@felixarntz from my testing, the `containerID` is already populated in Settings Edit screen, probably because it's loaded before clicking he `Edit` button in the `SettingsView` component. However, directly going to the `Edit` screen causes it to be `undefined` at first ie. http://sitekit.10uplabs.com/wp-admin/admin.php?page=googlesitekit-settings#/connected-services/tagmanager/edit resulting another edge case where the default snippet toggle behavior is applied right away once the `containerID` is loaded. Is it something we need to fix? 

Cheers.

felixarntz:
@kuasha420 Ah, that's a good point, the data is already populated once the component is mounted. Since it's possible to link directly to the "Edit" version of these settings panels, I think it would be good to cover that use-case as well. Essentially, once it changes from `undefined` to something else, consider that as the moment that the original value has loaded.

kuasha420:
@felixarntz sadly that logic will not work for the setup form, as it's always will be `undefined` on setup form and when a container is selected via dropdown for the first time, the hook will not run. 

The most "clean" way I've found is relying on the fact that:

* On Setup Flow:
    * The containerID is `undefined` initially. 
    * The containerID then resolves to an empty string (''). - from this point onwards, we should run the logic every time container changes. 
* On SettingsView -> SettingsForm
    * The containerID is populated from the get go. - We need to ignore this change and run the logic on subsequent containerID change. 
* On Direct Accessing SettingsForm
    * The containerID is initially `undefined`
    * Then it gets populated. - We need to ignore this change and run the logic on subsequent containerID change. 

Based on that, I have created another follow up [PR](https://github.com/google/site-kit-wp/pull/4965) where I updated the hook and added couple of more tests to also cover for the new scenarios. You can have a look here. I have also did some quick smoke testing and all seems to be working correctly. 

Let's continue the discussion there. 

Cheers.

felixarntz:
@kuasha420 
> as it's always will be `undefined` on setup form

It will be `undefined` at the start, but then it'll become an empty string as soon as it's loaded, like you're also pointing out later. So even if nothing is selected yet, we can identify once it has been loaded.

The PR #4965 looks good to me, I see it's already been merged.

One follow-up question: Do we already have an issue to do the same enhancement for the Analytics `useExistingTagEffect`? The same problem would likely apply there.

Issue #4925, opened 2022-03-07
Make timeout for `view_dashboard` surveys more reliable and extend it


felixarntz:
It is currently possible to see a few too many surveys in short time frame, which isn't that great from a user experience perspective. This is particularly related to the `view_dashboard` trigger, which several surveys get attached to. While there is already a 1 hour timeout, it is only stored on the client-side and it is regularly invalidated e.g. when activating a new module, so that can result in an overload of surveys being shown in little time.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The implementation of the timeout mechanism for remote-controlled surveys should be modified to rely on a server-side storage (i.e. a storage which persists beyond user sessions in the browser). In practice, this only affects the `view_dashboard` trigger, which is the only survey trigger ID currently relying on a timeout, but the implementation should be generic enough to support other triggers in the future.
    * The storage should rely on user meta, so that this is stored per user.
    * The storage should support an undefined number of trigger IDs, so it shouldn't be a single meta key just for the `view_dashboard` trigger.
    * A REST API endpoint to set a trigger timeout should be added, which allows specifying the trigger ID and requested timeout.
    * The client has to then receive awareness of which trigger IDs are currently "on timeout" per user, replacing the session storage lookup for this (potentially with a preloaded REST API endpoint to get the list of trigger timeouts).
* The timeout for the `view_dashboard` trigger specifically should be extended to 24 hours (instead of 1 hour as it is now).

### Create new REST endpoints

Note: The Dismissed Items endpoints and related code under `includes/Core/Dismissals` can serve as a reference and broadly be copied here.

Working in the directory `includes/Core/User_Surveys`:

- Create a new class called `Survey_Timeouts` to represent the survey timeouts as a user setting.
  - This can mostly be a copy of [`includes/Core/Dismissals/Dismissed_Items.php`](https://github.com/google/site-kit-wp/blob/372a83b9ac570045311cefa1397c0b6f4230c454/includes/Core/Dismissals/Dismissed_Items.php), renamed for survey timeouts.
  - `OPTION` should be `googlesitekit_survey_timeouts`.
  - `get_dismissed_items` and `filter_dismissed_items` should be renamed `get_survey_timeouts` and `filter_survey_timeouts`.
  - There is no equivalent need for the method `is_dismissed` and this can be ignored/removed in `Survey_Timeouts`.
[...]

PR #5003, opened 2022-03-30
Enhancement/4925 survey timeouts

eugene-manuilov:
## Summary

Addresses issue:

- #4925

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist


assets/js/googlesitekit/datastore/user/surveys.js
```diff
@@ -233,20 +333,70 @@ const baseSelectors = {
...
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {(string[]|undefined)} Array of surveys slugs, `undefined` if not resolved yet.
+	 */
+	getSurveyTimeouts( state ) {
+		return state.surveyTimeouts;
+	},
+
+	/**
+	 * Determines whether the survey is timedout or not.
```
tofumatt, line 350:
```suggestion
	 * Determines whether the survey is timed out or not.
```

assets/js/googlesitekit/datastore/user/surveys.js
```diff
@@ -113,28 +190,41 @@ const baseActions = {
...
+			const isTimingOut = select( CORE_USER ).isTimingOutSurvey(
 				triggerID,
-			} );
-			const { cacheHit } = yield Data.commonActions.await(
-				getItem( cacheKey )
+				ttl
 			);
 
-			if ( false === cacheHit ) {
+			if ( ! isTimedOut && ! isTimingOut ) {
```
tofumatt, line 206: Should these be checks for `isTimedOut === false` rather than `! isTimedOut`? Because `isTimedOut` could be `undefined`… do we want this to behave the same way if it's `false` and `undefined`?

```suggestion
			if ( ! isTimedOut && ! isTimingOut ) {
```

tofumatt (commented):
This largely looks good, just a few questions and small comments. I think after that this can get merged 👍🏻 

I tested it locally and it worked well 👍🏻

assets/js/googlesitekit/datastore/user/surveys.js
eugene-manuilov: Those variables are always resolved by the time we get them. Added a comment that explains it.

tofumatt: Ah, awesome, thanks! I think in that case too it's fine to be less explicit with the checks as anyone reading this for prior art will see the comment and know why we're doing a false-y rather than `=== false` check. 👍🏻

tofumatt (approved):
Awesome, thanks for the comments. Works great when tested locally 🚢

assets/js/components/surveys/SurveyViewTrigger.js
```diff
@@ -19,41 +19,47 @@
...
-	useMount( () => {
-		if ( usingProxy ) {
+	const shouldTriggerSurvey = isTimedOut === false && isTimingOut === false;
+
+	useEffect( () => {
+		if ( shouldTriggerSurvey && usingProxy ) {
 			triggerSurvey( triggerID, { ttl } );
 		}
-	} );
+	}, [ shouldTriggerSurvey, usingProxy, triggerSurvey, triggerID, ttl ] );
```
felixarntz, line 57: Why is all this needed? It looks like the `triggerSurvey` action already includes these checks and also ensures the API-based bits are resolved. Why duplicate these conditions in here?

I'm not sure why it's needed, and it makes me slightly worried relying on `useEffect` instead of `useMount`, since this really should only happen once.

felixarntz: @tofumatt I'll leave it to you and @eugene-manuilov to make a decision on this, but my thinking is that the additional code here is somewhat of a duplicate since the action is already resilient enough to "make sure it works", plus the point about `useEffect` theoretically being more error-prone to calling the action multiple times - while I agree in that I cannot see a point where this happens using the current logic, it feels more error-prone than using `useMount`.

If you still feel the current approach is more resilient, we can leave it, but if you agree, please open a follow-up PR to address this.

Issue #4966, opened 2022-03-23
Blue Box CTA shows to notify user to complete Analytics set up if user do not complete or cancel Analytics set up. 


mohitwp:
## Feature Description

If user cancels or does not complete the Analytics set up then "Blue box CTA" display to notify user about completing the Analytics set up. As we are removing 'Blue Box widget' and implementing new designs on other places. I think this should be change too.

Steps: 
1. Using the tester plugin, force the 'gathering data' state or 'zero' data state.
2. Activate the "zeroDataStates" feature flag
3. Go to Site Kit Dashboard.
4. Click on set up Analytics.
5. Cancel or do not complete Analytics module set up.
6. See "Blue box widget' on Main Dashboard, Entity dashboard and WP dashboard.

## Screenshots

Main Dashboard

![image](https://user-images.githubusercontent.com/94359491/159644201-62a02925-e565-4ce3-bbdb-3863e152af31.png)

Entity Dashboard

![image](https://user-images.githubusercontent.com/94359491/159644112-0503995d-8201-4af9-99a5-ea8428f472e3.png)

WP Dashboard 
![image](https://user-images.githubusercontent.com/94359491/159644276-fb08a6ad-2b47-4a80-9ea6-318bc8b06a1e.png)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* All "Complete module activation" CTAs for Analytics should look exactly the same as the respective new "Activate Analytics" CTA, with the only difference being that the CTA button says "Complete setup".
    * Preferably, the same component should be used under the hood for the very similar UI.
* This change should happen everywhere this CTA can be present, i.e. the Site Kit dashboard, the WP dashboard widget, and the WP admin bar menu.
* Storybook for the UI in "Complete module activation" state should be added.

## Implementation Brief

- Create a new hook - `useCompleteModuleActivationCallback.js` which accepts `moduleSlug` as a parameter:
    - Move the entire logic of computing the `onCTAClick` callback from `CompleteModuleActivationCTA.js` into this new hook.
      https://github.com/google/site-kit-wp/blob/4baca515e05bddbe5dc26c6302789588d4a0d9d5/assets/js/components/CompleteModuleActivationCTA.js#L44-L61 
    - Rename the `onCTAClick` const to `completeModuleActivationCallback` and return this value. 
    - This refactor is similar to the one done in [#4694](https://github.com/google/site-kit-wp/issues/4694) which can be used as a reference. 
[...]

PR #5028, opened 2022-04-04
Enhancement/4966 Complete Analytics CTA

jimmymadon:
## Summary

Addresses issue:

- #4966

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist


assets/js/components/ActivateAnalyticsCTA.js
```diff
@@ -30,15 +30,26 @@ import { __ } from '@wordpress/i18n';
  * Internal dependencies
  */
 import useActivateModuleCallback from '../hooks/useActivateModuleCallback';
+import useCompleteModuleActivationCallback from '../hooks/useCompleteModuleActivationCallback';
 import Button from './Button';
 
-export default function ActivateAnalyticsCTA( { children } ) {
+export default function ActivateAnalyticsCTA( {
+	children,
+	isCompleteSetup = false,
```
aaemnnosttv, line 38: Please ensure prop types are updated when adding props 😊 

More importantly, this prop name is a bit confusing as it could easily be interpreted to mean "setup is complete" in which case the CTA shouldn't be shown. Instead, I think it would be more clear to use the inverse – `incomplete` – as the name. It's value should be equal to `active && !connected` so that "Complete setup" only shows if `incomplete` is truthy. This would also read a bit more intuitively in the logic to determine the callback to use.

assets/js/components/ActivateAnalyticsCTA.js
```diff
@@ -30,15 +30,26 @@ import { __ } from '@wordpress/i18n';
...
+		'analytics'
+	);
 
 	if ( ! activateModuleCallback ) {
 		return null;
 	}
 
+	if ( isCompleteSetup && ! completeModuleActivationCallback ) {
+		return null;
+	}
```
aaemnnosttv, line 51: These conditions could be simplified into one by computing the onClick callback above rather than inline below. Then the guard here is a simple check for the onClick callback (regardless of the case), which is more intuitive 🙂

assets/js/modules/search-console/components/dashboard/SearchFunnelWidget/index.js
```diff
@@ -385,6 +385,21 @@ const SearchFunnelWidget = ( {
 					</Grid>
 				) }
 
+			{ ! isAnalyticsActive &&
+				isAnalyticsConnected &&
```
aaemnnosttv, line 389: It's not possible for a module to be inactive and connected so this condition will never render.

Instead of adding a mostly duplicate block here, let's update the above to render if `! isAnalyticsActive || ! isAnalyticsConnected` (could really be simplified to the latter but perhaps easier to understand this way) rather than only for the inactive case. Then the `incomplete` prop would be set as above `active && ! connected` to cover both cases.

aaemnnosttv (changes requested):
Thanks @jimmymadon – This is off to a good start!

Aside from my comments below, this seems to be missing the added stories for Storybook called for in the AC+IB. Let me know if you have any questions 👍

assets/js/components/ActivateAnalyticsCTA.js
jimmymadon: @aaemnnosttv In my head, I was considering "completeSetup" as a mode or state (="is in completeSetup mode?"), rather than making it sound like the English question "is setup complete?" But I can see how this can be misinterpreted. Also, I normally prefer positive boolean flags over negative (e.g. `isValid` rather than `isInvalid`). But in this case, I agree - it will sound alright and perhaps even more intuitive to have the inverse. Finally, I would prefer booleans to be a question - so would `isSetupIncomplete` sound better than just `incomplete` here?

jimmymadon: Actually, `isSetupIncomplete` is technically also the case when the plugin is inactive and not connected (if you generally consider a plugin setup to be both). So something like `isConnectionIncomplete` might sound better.

assets/js/modules/search-console/components/dashboard/SearchFunnelWidget/index.js
jimmymadon: Beautiful!

assets/js/components/ActivateAnalyticsCTA.js
aaemnnosttv: I probably spent too much time thinking about this as well. You're right that `incomplete` can be thought to apply to the initial inactive state as well. From another perspective, incomplete implies that something is started but unfinished.

For a single prop solution, the other name I was considering would be something like `activated` as in, the module is active, but since the CTA is still showing that must mean that it's not completed yet. That would be odd to read though as well that a `ActivateAnalyticsCTA` component has an `activated` prop, one might think it shouldn't be shown in that case as well.

In a way, the confusion comes from using props I think – in the past we would use `useSelect` for `isModuleActive` and `isModuleConnected` and everything would be very explicit. This makes it a bit more complicated to set up in Storybook as recently discussed though. Seeing as we're already selecting from the store in other hooks though, why not do the same here?

assets/js/components/ActivateAnalyticsCTA.js
```diff
@@ -29,13 +29,27 @@ import { __ } from '@wordpress/i18n';
...
+		select( CORE_MODULES ).isModuleActive( 'analytics' )
+	);
+
+	const onClickCallback = analyticsModuleActive
+		? completeModuleActivationCallback
+		: activateModuleCallback;
 
-	if ( ! activateModuleCallback ) {
+	if ( ! onClickCallback ) {
 		return null;
```
aaemnnosttv, line 53: Much better! 🙂

assets/js/components/adminbar/AdminBarWidgets.stories.js
```diff
@@ -156,15 +161,43 @@ AnalyticsInactiveNew.storyName = 'Inactive: Analytics New CTA';
...
+			<WithRegistrySetup func={ setupRegistry }>
+				<Story />
+			</WithRegistrySetup>
+		);
+	},
+];
+AnalyticsInactiveNew.parameters = {
+	features: [ 'zeroDataStates' ],
+};
+AnalyticsInactiveNew.scenario = {
```
aaemnnosttv, line 179: This issue doesn't call for added VRT scenarios, only Storybook. Let's hold on adding these as we haven't really established criteria for what should be covered or not for new scenarios.

assets/js/components/adminbar/AdminBarWidgets.stories.js
```diff
@@ -156,15 +161,43 @@ AnalyticsInactiveNew.storyName = 'Inactive: Analytics New CTA';
...
+		const setupRegistry = ( registry ) => {
+			// Set up the search console and analytics module store but provide no data.
 			provideModules( registry, [
 				{
-					slug: 'search-console',
+					slug: 'analytics',
 					active: true,
-					connected: true,
+					connected: false,
+					// storeName: 'modules/analytics',
```
aaemnnosttv, line 196: Is this omission intentional or was this left in by mistake?

assets/js/components/adminbar/AdminBarWidgets.stories.js
```diff
@@ -175,9 +208,14 @@ AnalyticsInactiveNew.decorators = [
...
 	},
 ];
-AnalyticsInactiveNew.parameters = {
+AnalyticsInactiveNewCompleteActivation.parameters = {
 	features: [ 'zeroDataStates' ],
 };
+AnalyticsInactiveNewCompleteActivation.scenario = {
+	label:
+		'Views/AdminBarApp/AdminBarWidgets/AnalyticsInactiveNewCompleteActivation',
+	delay: 3000,
```
aaemnnosttv, line 217: Same as above, but also I would hope we wouldn't need to add a delay for such simple components.

assets/js/hooks/useCompleteModuleActivationCallback.test.js
```diff
@@ -0,0 +1,101 @@
...
+			( { result } = await renderHook(
+				() => useCompleteModuleActivationCallback( 'analytics' ),
+				{ registry }
+			) );
+		} );
+
+		await result.current();
+
+		expect( global.location.assign ).toHaveBeenCalledWith(
+			'http://example.com/wp-admin/admin.php?page=googlesitekit-module-analytics&slug=analytics&reAuth=true'
```
aaemnnosttv, line 68: Using the literal URL here makes this test a bit fragile to changes to the reauth URL. How about we select this value to use in the assertion instead?

aaemnnosttv (commented):
Thanks @jimmymadon – this looks good to me now, just a few minor things to clean up.

assets/js/components/adminbar/AdminBarWidgets.stories.js
jimmymadon: Oops!

assets/js/hooks/useActivateModuleCallback.test.js
```diff
@@ -114,9 +114,10 @@ describe( 'useActivateModuleCallback', () => {
 			'analytics'
 		);
 
-		expect( global.location.assign ).toHaveBeenCalledWith(
-			'http://example.com/wp-admin/admin.php?page=googlesitekit-module-analytics&slug=analytics&reAuth=true'
-		);
+		const reauthURL = registry
+			.select( 'modules/analytics' )
```
aaemnnosttv, line 118: Please use the constant for the store name as we do elsewhere 👍

assets/js/modules/search-console/components/dashboard/SearchFunnelWidget/index.stories.js
```diff
@@ -217,6 +218,37 @@ ReadyWithActivateAnalyticsCTA.scenario = {
...
+	},
+};
+ReadyWithCompleteAnalyticsActivationCTA.parameters = {
+	features: [ 'zeroDataStates' ],
+};
+ReadyWithCompleteAnalyticsActivationCTA.scenario = {
+	label:
+		'SearchConsole/SearchFunnelWidget/ReadyWithCompleteAnalyticsActivationCTA',
+	delay: 3000,
+};
```
aaemnnosttv, line 250: Please remove for now.

aaemnnosttv (changes requested):
Thanks @jimmymadon – this is pretty close I think, just a few comments left to resolve. Also, there are still added VRT reference images that should be removed. If any VRT updates were necessary, that would be fine but these all look like additions from the scenarios that were removed (or should be). Thanks!

assets/js/modules/search-console/components/dashboard/SearchFunnelWidget/index.stories.js
jimmymadon: I removed the VRT tests for the AdminBar and their corresponding reference images as these were not added for any version of the CTA - and it is likely this CTA will be modified very soon. The reason I didn't remove the new `SearchFunnelWidget` and the `WPDashboard` tests was because they have VRTs for the `ActivateAnalyticsCTA` already. So just wanted to maintain consistency. Also, the `WPDashboard` VRTs were named at the folder level, so I renamed the existing VRTs (3) and added the new `CompleteActivationCTA` tests as well (3). I can get rid of the `delays` - I copied them over from the `ActivateAnalyticsCTA` story and thought this might be needed to due to other report data loading too for surrounding widgets?

aaemnnosttv: > The reason I didn't remove the new `SearchFunnelWidget` and the `WPDashboard` tests was because they have VRTs for the `ActivateAnalyticsCTA` already. So just wanted to maintain consistency.

@jimmymadon Thanks for clarifying – that makes sense, but we don't usually add scenarios for each variant of a component. In this case, there is a rather small difference between this story and the existing inactive variant as the only observable difference would be the wording. It's something we definitely want a Storybook story for, but doesn't need an added scenario.

aaemnnosttv (commented):
Thanks @jimmymadon – see my comment below. I'll push the final change here for the sake of time, otherwise this should be good to go 👍

aaemnnosttv (approved):
👍

Issue #5008, opened 2022-03-30
Incorrect dashboard notification briefly shows before correct one is loaded


wpdarren:
## Bug Description

When one module (e.g. Analytics) is gathering data while the other module (e.g. Search Console) has just zero data available, the Site Kit dashboard may briefly flash the notification about the zero data module (Search Console) before showing the notification about the gathering data module (Analytics).

It is expected for the latter to have precedence, so that's all good, but we should ensure we don't show the other one first to then replace it. Instead, the components should ensure the necessary data has been loaded for both notifications to determine which one to show before showing any.

To replicate that, configure the modules as described in the first paragraph and load the Site Kit dashboard. You may see the one notification about the zero data module briefly show up before the other one takes over.

From the [bug bash asana ticket](https://app.asana.com/0/1201999381911618/1202030688961883/f)

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Before showing the banner notification for either "gathering data" or "zero data", it must be ensured that _all_ relevant data affecting their conditions has been loaded (i.e. is not `undefined`).
    * It should no longer be possible for one of the two notifications to briefly pop up just to be replaced by the other one a second later (as the check determining the other one is needed only completed a bit later).

## Implementation Brief

* Add `=== undefined` checks for all "gathering data" and "zero data" `useSelect` results in `ZeroDataStateNotifications.js`, eg:

```js
if (
	analyticsGatheringData === undefined ||
	searchConsoleGatheringData === undefined ||
	analyticsHasZeroData === undefined ||
	searchConsoleHasZeroData === undefined
) {
	return null;
}
```

This ensures no notification is shown until all data is ready.

That said, it might also be worth only displaying the zero data notification if "gathering data" is `false`, given the "gathering data" scenario "takes precedence" over "zero data", so there's no sense outputting even the _component_ if gathering data is true.

### Test Coverage

* No new tests are needed.

## QA Brief

- Ensure `zeroDataStates` feature flag is enabled.
- Using the tester plugin, enable the `gathering` data state for Analytics and `zero` data state for Search Console.
- Clear the session and reload the page.
- Ensure the `zero` data state notification doesn’t pop up briefly and disappear.
- Ensure the `gathering` data state notification appears as expected.

## Changelog entry

* Fix bug that could cause the incorrect notification to briefly appear when Analytics or Search Console is gathering data.

aaemnnosttv:
[...]

PR #5032, opened 2022-04-05
Bug/5008 - Incorrect dashboard notification briefly shows before correct one is loaded

hussain-t:
## Summary

Addresses issue:

- #5008 

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist

### DIFF
--- a/assets/js/components/notifications/ZeroDataStateNotifications.js
+++ b/assets/js/components/notifications/ZeroDataStateNotifications.js
@@ -55,6 +55,15 @@ export default function ZeroDataStateNotifications() {
 	);
 
 	if (
+		analyticsGatheringData === undefined ||
+		searchConsoleGatheringData === undefined ||
+		analyticsHasZeroData === undefined ||
+		searchConsoleHasZeroData === undefined
+	) {
+		return null;
+	}
+
+	if (
 		! analyticsGatheringData &&
 		! searchConsoleGatheringData &&
 		! analyticsHasZeroData &&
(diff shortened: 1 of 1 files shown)


assets/js/components/notifications/ZeroDataStateNotifications.js
```diff
@@ -99,26 +108,28 @@ export default function ZeroDataStateNotifications() {
...
-					learnMoreLabel={ __( 'Learn more', 'google-site-kit' ) }
-					learnMoreURL="https://sitekit.withgoogle.com/documentation/using-site-kit/dashboard-data-display/"
-					dismiss={ __( 'Remind me later', 'google-site-kit' ) }
-					isDismissible
-					dismissExpires={ getTimeInSeconds( 'day' ) }
-					SmallImageSVG={ ZeroStateIcon }
-				/>
-			) }
+			{ ( analyticsGatheringData === false ||
+				searchConsoleGatheringData === false ) &&
```
tofumatt, line 117: This should be an `AND` conditional, shouldn't it? Because otherwise if only one of these is `false` and the other one `true`, eg. `analyticsGatheringData === false && searchConsoleGatheringData === true`, `searchConsoleHasZeroData` would also be `true` and React would render this.

As mentioned in the IB: our UI won't actually be updated because we only show one notification at a time, but in this case the bug this attempts to solve could still happen if we weren't returning early like above.

This shouldn't ever happen, but these conditions _imply_ it'd be okay if they did… or even that maybe we think they _should_. So this should probably be:

```suggestion
			{ ( analyticsGatheringData === false &&
				searchConsoleGatheringData === false ) &&
```

tofumatt (changes requested):
I think we should update the conditions for showing the zero state here, but otherwise looks good.

(I added a comment for the logic here because I had it in my editor when I was proof-of-concept-ing this IB 😅 )

assets/js/components/notifications/ZeroDataStateNotifications.js
hussain-t: >This should be an AND conditional, shouldn't it?

Thanks, @tofumatt. You are right 💯  

Let me update the condition, and let's have it since it explicitly states when to render the zero data notifications 👍

hussain-t:
Thanks for adding the [comments](https://github.com/google/site-kit-wp/pull/5032/commits/c01aa36a3d9d86743a5ffbd61074062493c05f1e) 👍

Issue #4846, opened 2022-02-15
Update event tracking to allow segmentation by authentication


aaemnnosttv:
## Feature Description

Since the beginning, using Site Kit has always been limited to authenticated admins. Starting with dashboard sharing, this limitation will no longer apply to all users using Site Kit, making it possible for non-authenticated users to use the dashboards in a read-only way. Such users will also be able to opt-in to our anonymous usage tracking which helps us to improve the product through various tracked events, such as errors.

Going forward, it will be important to be able to segment this information by which users are authenticated (with Site Kit) and which are not.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The event data in Site Kit's `trackEvent` function should be extended with a new custom dimension based on the current user's authentication status (whether or not they have an access token or not)
	- The dimension should be `dimension7`
	- The value should be `1` if authenticated, otherwise `0`

## Implementation Brief

* Update the hook for the `googlesitekit_user_data` filter in the `Authentication` class to add the `authenticated` property to the user data. Use the result of the `is_authenticated()` method call as the value for the new property.
* Add the `dimension7` property to the `eventData` object in the `createTrackEvent` function. Use `global._googlesitekitUserData?.user?.authenticated ? 1 : 0` as the value for the new dimension.
* A comment explaining what this dimension is for should be added, as it isn't self-evident from the name `dimension7`.

### Test Coverage

* N/A

## QA Brief

This issue adds a new custom dimension (`dimension7`) to all internal events tracked. This will be either `1` if the user is signed in with Google or `0` if not.

* Enable tracking in Settings -> Admin Settings
* Go to the main dashboard
* Scroll down the page
* Most GA events which are firing will have `dimension7` set to `1` since right now we only have authenticated users.
* One event for a non-authenticated user is starting the setup from the splash page (may need to [set console output to be persistent](https://developer.chrome.com/docs/devtools/console/reference/#persist) to observe this)

## Changelog entry

* Update internal event tracking to include user authentication state.

wpdarren:
QA Update: ⚠️ 

@asvinb @aaemnnosttv could you expand on how we might test this for a non-authenticated user. In the QAB it mentions that we may need to [set console output to be persistent](https://developer.chrome.com/docs/devtools/console/reference/#persist) to observe this) but I am unsure how that will allow us to test this.  Am I missing something here?  Any additional information would be appreciated.

aaemnnosttv:
[...]

PR #5089, opened 2022-04-14
Add custom dimension for authenticated or not users.

asvinb:
## Summary

Addresses issue:

- #4846 

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist


includes/Core/Authentication/Authentication.php
```diff
@@ -364,11 +364,11 @@ function( $user ) {
 					$user['user']['email']   = $profile_data['email'];
 					$user['user']['picture'] = $profile_data['photo'];
 				}
-
-				$user['connectURL']     = esc_url_raw( $this->get_connect_url() );
-				$user['initialVersion'] = $this->initial_version->get();
-				$user['userInputState'] = $this->user_input_state->get();
-				$user['verified']       = $this->verification->has();
+				$user['user']['authenticated'] = $this->is_authenticated();
```
aaemnnosttv, line 367: The `user` key/property is for data specific to the WP User (with the exception of the email and picture 😄 ). Let's put this on the top-level, as a sibling of `verified` instead.

assets/js/util/tracking/createTrackEvent.js
```diff
@@ -64,6 +64,10 @@ export default function createTrackEvent(
 			dimension4: global.GOOGLESITEKIT_VERSION || '',
 			dimension5: Array.from( enabledFeatures ).join( ',' ),
 			dimension6: activeModules.join( ',' ),
+			// Custom dimension to differentiate between authenticated and non authenticated users.
+			dimension7: global._googlesitekitUserData?.user?.authenticated
```
aaemnnosttv, line 68: I realize this was done once before but we should avoid reaching for globals directly here. [We use a number of globals already](https://github.com/google/site-kit-wp/blob/460a21e240c91f51f6a22ca17b1ba67b8162236e/assets/js/util/tracking/index.js#L24-L31) but these are passed through in the `config`, so let's take the same approach here.

Would you please do the same with the `currentUserRoles` above for consistency?

We still reference `console` and `_gaUserPrefs` on the global below but even then we use the injected `_global`, but these are the only exceptions we should need.

Also – please add/update the tests for `trackEvent` to ensure this dimension is covered.

aaemnnosttv (changes requested):
Thanks @asvinb – this one is mostly there, just a few small things to address 👍

assets/js/util/tracking/createTrackEvent.js
asvinb: I've also updated the use of `GOOGLESITEKIT_VERSION` as well.

includes/Core/Authentication/Authentication.php
asvinb: 👍

assets/js/util/tracking/index.js
```diff
@@ -30,13 +30,23 @@ const {
...
+const { GOOGLESITEKIT_VERSION: pluginVersion } = global;
+
 const initialConfig = {
 	activeModules,
 	trackingEnabled,
 	trackingID,
 	referenceSiteURL,
 	userIDHash,
 	isSiteKitScreen,
+	currentUserRoles,
```
aaemnnosttv, line 47: We can leave off the `current` as everything related to the user here is already specific to the current user (e.g. `userIDHash`)
```suggestion
	userRoles,
```

assets/js/util/tracking/index.js
```diff
@@ -30,13 +30,23 @@ const {
...
+	authenticatedUser,
```
aaemnnosttv, line 48: Let's use `isAuthenticated` for consistency with other uses/references in JS.
```suggestion
	isAuthenticated,
```

assets/js/util/tracking/createTrackEvent.js
```diff
@@ -61,9 +61,11 @@ export default function createTrackEvent(
 			dimension1: referenceSiteURL,
 			dimension2: currentUserRoles.join( ',' ),
 			dimension3: userIDHash,
-			dimension4: global.GOOGLESITEKIT_VERSION || '',
+			dimension4: pluginVersion || '',
 			dimension5: Array.from( enabledFeatures ).join( ',' ),
 			dimension6: activeModules.join( ',' ),
+			// Custom dimension to differentiate between authenticated and non authenticated users.
```
aaemnnosttv, line 67: I know the IB mentions this but I'm not sure the comment adds much value as it reads to be rather self-documenting IMO 😄

assets/js/util/tracking/index.test.js
```diff
@@ -85,6 +85,9 @@ describe( 'trackEvent', () => {
 			trackingID: 'UA-12345678-1',
 			activeModules: [],
 			trackingEnabled: true,
+			currentUserRoles: [ 'administrator' ],
+			authenticatedUser: 1,
+			pluginVersion: global.GOOGLESITEKIT_VERSION,
```
aaemnnosttv, line 90: Let's use an inline value here for the test rather than use the global.
```suggestion
			pluginVersion: '1.2.3',
```

includes/Core/Assets/Assets.php
```diff
@@ -452,6 +452,7 @@ private function get_assets() {
 					'src'          => $base_url . 'js/googlesitekit-base.js',
 					'dependencies' => array(
 						'googlesitekit-base-data',
+						'googlesitekit-user-data',
```
aaemnnosttv, line 455: Why was this added? The base JS doesn't use `trackEvent` or reference the user global.

aaemnnosttv (commented):
Thanks @asvinb – just a few minor things left to address 👍

assets/js/util/tracking/index.js
asvinb: 👍

assets/js/util/tracking/index.js
asvinb: 👍

assets/js/util/tracking/createTrackEvent.js
asvinb: 😁 Removed!

includes/Core/Assets/Assets.php
asvinb: @aaemnnosttv It's because upon page load, `window._googlesitekitUserData` is `undefined`

aaemnnosttv: I see what you mean. I wonder if these properties belong in base data instead then? 🤔 We probably should avoid loading all user data on all of wp-admin.

@felixarntz do you think we should keep all the metadata for custom dimensions in the base data or load an extra data dependency?

felixarntz: @aaemnnosttv @asvinb Let's put it in `_googlesitekitBaseData`, like all other tracking dependency data. `userIDHash` for example is also technically user information, but part of the base data. That data isn't sensitive and should preferably be exlusively used for the tracking, so I think that's preferable over loading the overall user data everywhere.

aaemnnosttv: Thanks @felixarntz that sounds good to me 👍 

@asvinb please move `userRoles` to the base data as well so that tracking configuration only depends on the base data for now. In the future, we'll update it to use a separate global just for tracking data.

aaemnnosttv (commented):
Thanks @asvinb – this is almost there, just the one change left to make 😄 See Felix's comment here https://github.com/google/site-kit-wp/pull/5089/files#r854463054

includes/Core/Assets/Assets.php
asvinb: Sounds good @aaemnnosttv . I've updated the PR. Can you take another look?

aaemnnosttv:
Thanks @asvinb this LGTM, just a few minor places where a few of the moved values were still referenced. I'll push these last small tweaks and then this should be good to go 👍

aaemnnosttv (approved):

Issue #4794, opened 2022-02-04
Implement new actions for sharing settings


aaemnnosttv:
## Feature Description

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

The `core/modules` store should be extended with the following actions:

* `setSharingManagement(moduleSlug: string, management: enum)`  
Sets the value for a module's sharing management where `management` can be one of `all_admins` or `owner`
* `setSharedRoles(moduleSlug: string, roles: string[])`  
Sets the list of roles (`role.id`[]) the module is shared with
* `saveSharingSettings()`  
Saves all sharing settings via `POST:core/modules/data/sharing-settings` using `createFetchStore`
	- See https://github.com/google/site-kit-wp/issues/4481 for the response shape
	- The current and saved settings should be received/refreshed in the state from the response
	- Any `newOwnerIDs` included in the response should update the modules in state
* The state for settings should work similar to `createSettingsStore` which stores state for saved settings separately from the current working/draft values that selectors and actions touch
	- Accordingly, an action for `receiveGetSharingSettings` should also be created which should update both
* Test coverage should be added for all new actions

## Implementation Brief
* Create `assets/js/googlesitekit/modules/datastore/sharing-settings.js` which will contain our actions and reducers.
* The `baseInitialState` should contain the `sharingSettings` and `savedSharingSettings` property, both initialized to an empty object.
* Create `fetchSaveSharingSettingsStore` using the `createFetchStore` function with the following details:
	* `baseName` should be set to `saveSharingSettings`
	* `controlCallback` should:
		* require one parameter, an object having the same shape as `state.savedSharingSettings`.
		* make a POST request to `core/modules/data/sharing-settings` using `API.set`, and passing the `state.savedSharingSettings` as the body.
		* the `argsToParams` and `validateParams` should be set as well. Refer to how it's done in other data partials.
		* For the `reducerCallback`, 
			* `state.sharingSettings` should be updated to the response from the API, i.e both `state.sharingSettings` and `state.saveSharingSettings` should be equal.
			* check if `newOwnerIDs` from the API response is not empty. Refer to https://github.com/google/site-kit-wp/issues/4481 for the structure. In this case, loop through the array and update each module owner ID using the `setOwnerID` action.
* Similar to other data store partials in the folder, add the following `actions` within the `baseAction` object:
	* `setSharingManagement`
[...]

PR #5127, opened 2022-04-25
Enhance/4794 - Implement new actions for sharing settings

hussain-t:
## Summary

Addresses issue:

- #4794 

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist


assets/js/googlesitekit/modules/datastore/sharing-settings.js
```diff
@@ -0,0 +1,301 @@
...
+						management,
+					},
+				},
+				savedSharingSettings: {
+					...state.savedSharingSettings,
+					[ moduleSlug ]: {
+						...state.savedSharingSettings[ moduleSlug ],
+						management,
+					},
+				},
```
eugene-manuilov, line 213: @hussain-t, only `sharingSettings` should be updated because `savedSharingSettings` object represents the original sharing settings and it shouldn't be changed until settings are saved. Same applies for other actions in this reducer, except `RECEIVE_GET_SHARING_SETTINGS`.

assets/js/googlesitekit/modules/datastore/sharing-settings.js
```diff
@@ -0,0 +1,301 @@
...
+	 */
+	setOwnerID( newOwnerIDs ) {
+		invariant( newOwnerIDs, 'newOwnerIDs is required.' );
+		return {
+			payload: {
+				newOwnerIDs,
+			},
+			type: SET_OWNER_ID,
+		};
+	},
```
eugene-manuilov, line 86: By `update each module owner ID using the setOwnerID action` in IB, we meant to call the `setOwnerID` action of individual modules that updates `ownerID` setting. So we don't need this action. What we need to do is to get a store name for each module that receives new owner ID and then dispatch the appropriate `setOwnerID` action using the store name received for the module. There is a dedicated selector in the `CORE_MODULES` datastore to get a store name in the module.

eugene-manuilov (changes requested):
Thanks, @hussain-t. This is a very solid start. Added two comments with some things that we need to update in this PR before we can approve it. Please, take a look.

assets/js/googlesitekit/modules/datastore/sharing-settings.js
```diff
@@ -0,0 +1,253 @@
...
+			return {
+				...state,
+				sharingSettings: {
+					...state.sharingSettings,
+					...sharingSettings,
+				},
+				savedSharingSettings: {
+					...state.savedSharingSettings,
+					...sharingSettings,
+				},
```
eugene-manuilov, line 226: It shouldn't merge `sharingSettings` and `savedSharingSettings`, it should override it:
```suggestion
				sharingSettings,
				savedSharingSettings: sharingSettings,
```

assets/js/googlesitekit/modules/datastore/sharing-settings.js
```diff
@@ -0,0 +1,253 @@
...
+};
+
+const fetchSaveSharingSettingsStore = createFetchStore( {
+	baseName: 'saveSharingSettings',
+	controlCallback: ( { savedSharingSettings } ) => {
+		return API.set(
+			'core',
+			'modules',
+			'sharing-settings',
+			{ ...savedSharingSettings },
```
eugene-manuilov, line 49: No need to destruct the `savedSharingSettings` object, just pass it as is:
```suggestion
			savedSharingSettings,
```

assets/js/googlesitekit/modules/datastore/sharing-settings.test.js
```diff
@@ -0,0 +1,367 @@
...
+
+			it( 'requires the management param', () => {
+				expect( () => {
+					registry
+						.dispatch( CORE_MODULES )
+						.setSharingManagement( 'analytics' );
+				} ).toThrow( 'management must be one of: all_admins, owner.' );
+			} );
+
+			it( 'receives management and sets it to the sharing settings modules', () => {
```
eugene-manuilov, line 90: This test seems to be overly complicated. All what we need to do here are these 3 steps:

1. Set initial sharing settings using the `receiveGetSharingSettings` selector.
2. Dispatch the `setSharingManagement` action for the `analytics` module.
3. Check that `store.getState().sharingSettings.analytics.management` equals to passed value.

assets/js/googlesitekit/modules/datastore/sharing-settings.test.js
```diff
@@ -0,0 +1,367 @@
...
+
+			it( 'requires the roles param', () => {
+				expect( () => {
+					registry
+						.dispatch( CORE_MODULES )
+						.setSharedRoles( 'analytics' );
+				} ).toThrow( 'roles must be an array of strings.' );
+			} );
+
+			it( 'receives roles and sets it to the sharing settings modules', () => {
```
eugene-manuilov, line 161: Same here. No need to overcomplicate it. Just set the original settings, dispatch the `setSharedRoles` action for one module and check that the appropriate value is set in the state.

assets/js/googlesitekit/modules/datastore/sharing-settings.test.js
```diff
@@ -0,0 +1,367 @@
...
+
+				expect( store.getState().sharingSettings ).toMatchObject( {
+					...state,
+					...sharingSettingsWithSharedRoles,
+				} );
+			} );
+		} );
+
+		describe( 'saveSharingSettings', () => {
+			it( 'does not require any params', () => {
```
eugene-manuilov, line 200: This test doesn't seem to be needed. If the action doesn't use params, then no need to test for it.

assets/js/googlesitekit/modules/datastore/sharing-settings.test.js
```diff
@@ -0,0 +1,367 @@
...
+								},
+								{
+									slug: 'pagespeed-insights',
+									name: 'PageSpeed Insights',
+									storeName: 'modules/pagespeed-insights',
+								},
+							],
+							status: 200,
+						}
+					);
```
eugene-manuilov, line 271: No need to use fetchMock here, we can load modules using the `provideModules` function.

assets/js/googlesitekit/modules/datastore/sharing-settings.test.js
```diff
@@ -0,0 +1,367 @@
...
+
+					fetchMock.get(
+						/^\/google-site-kit\/v1\/modules\/search-console\/data\/settings/,
+						{ body: { setting1: 'value' }, status: 200 }
+					);
+
+					fetchMock.get(
+						/^\/google-site-kit\/v1\/modules\/pagespeed-insights\/data\/settings/,
+						{ body: { setting1: 'value' }, status: 200 }
+					);
```
eugene-manuilov, line 286: Same here, no need to use fetchMock for this, just use appropriate `receiveGetSettings` actions to set settings for modules.

assets/js/googlesitekit/modules/datastore/sharing-settings.test.js
```diff
@@ -0,0 +1,367 @@
...
+								},
+							],
+							status: 200,
+						}
+					);
+
+					fetchMock.get(
+						/^\/google-site-kit\/v1\/core\/site\/data\/settings/,
+						{ body: { setting1: 'value' }, status: 200 }
+					);
```
eugene-manuilov, line 276: Is it really needed? I am not sure that we have such endpoint. Where is it used?

assets/js/googlesitekit/modules/datastore/sharing-settings.test.js
```diff
@@ -0,0 +1,367 @@
...
+						}
+					);
+
+					const initialModules = registry
+						.select( CORE_MODULES )
+						.getModules();
+					// The modules info will be its initial value while the modules
+					// info is fetched.
+					expect( initialModules ).toBeUndefined();
+					await untilResolved( registry, CORE_MODULES ).getModules();
```
eugene-manuilov, line 304: This shouldn't be needed once you use `provideModules`.

assets/js/googlesitekit/modules/datastore/sharing-settings.test.js
```diff
@@ -0,0 +1,367 @@
...
+					.receiveGetSharingSettings( sharingSettings.settings );
+
+				expect( store.getState().sharingSettings ).toMatchObject( {
+					...state,
+					...sharingSettings.settings,
+				} );
+				expect( store.getState().savedSharingSettings ).toMatchObject( {
+					...state,
+					...sharingSettings.settings,
+				} );
```
eugene-manuilov, line 363: Also unnecessary complicated. Just dispatch the `receiveGetSharingSettings` action and check that the state contains the same `sharingSettings` and `savedSharingSettings` objects.

eugene-manuilov (changes requested):
Thanks, @hussain-t. Added more comments for you, mostly for tests. Shouldn't take too long to update them.

assets/js/googlesitekit/modules/datastore/sharing-settings.test.js
hussain-t: We do have several tests that check for `does not require any params`. Shall we keep it, WDYT?

eugene-manuilov: Not sure about other tests, but here we don't need it.

hussain-t: Sure, I will remove it.

assets/js/googlesitekit/modules/datastore/sharing-settings.test.js
hussain-t: It was required when I was using the `fetchMock`. However, we don't need it anymore when we dispatch the `receiveGetSettings` action 👍

eugene-manuilov (approved):
LGTM. Thanks, @hussain-t.

assets/js/googlesitekit/modules/datastore/sharing-settings.js
```diff
@@ -0,0 +1,247 @@
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {string} moduleSlug Module slug.
+	 * @param {string} management New management role for module, one of: all_admins, owner.
+	 * @return {Object} Action for SET_SHARING_MANAGEMENT.
+	 */
+	setSharingManagement( moduleSlug, management ) {
+		invariant( moduleSlug, 'moduleSlug is required.' );
+		const managementRoles = [ 'all_admins', 'owner' ];
```
aaemnnosttv, line 80: Nit: this doesn't need to be defined inside the action as it never changes. Also, "roles" is a bit of a misnomer as these aren't really roles (as are used elsewhere) so that could be a bit confusing. How about a `validManagementValues` defined at the top?

assets/js/googlesitekit/modules/datastore/sharing-settings.js
```diff
@@ -0,0 +1,247 @@
...
+			payload: {
+				moduleSlug,
+				management,
+			},
+			type: SET_SHARING_MANAGEMENT,
+		};
+	},
+
+	/**
+	 * Sets the sharing settings management role of given module.
```
aaemnnosttv, line 95: This is the same description as `setSharingManagement` which is incorrect.

assets/js/googlesitekit/modules/datastore/sharing-settings.js
```diff
@@ -0,0 +1,247 @@
...
+
+		const {
+			response,
+			error,
+		} = yield fetchSaveSharingSettingsStore.actions.fetchSaveSharingSettings(
+			sharingSettings
+		);
+
+		// Update module owner IDs in the sharing settings modules.
+		if ( Object.keys( response?.newOwnerIDs ).length ) {
```
aaemnnosttv, line 145: We don't need the conditional chain for accessing `newOwnerIDs` as it will always be in the response, but also if it were missing this code would still raise an error because you can't get keys from `undefined`
image

Instead, we need to first check that there is no `error` after which `response.newOwnerIDs` can be safely accessed.

assets/js/googlesitekit/modules/datastore/sharing-settings.js
```diff
@@ -0,0 +1,247 @@
...
+			for ( const [ slug, ownerID ] of Object.entries(
+				response.newOwnerIDs
+			) ) {
+				const storeName = registry
+					.select( CORE_MODULES )
+					.getModuleStoreName( slug );
+
+				yield registry.dispatch( storeName ).setOwnerID( ownerID );
```
aaemnnosttv, line 153: Looks like we have a few other instances of this, but only raw actions can be yielded from a generator. All actions returned from `dispatch` are bound action creators that return a promise so this will still invoke the action, but it won't be "awaited" like a normal yielded action would be.

This should be safe to dispatch without awaiting it as it should run synchronously but something to keep in mind.

```suggestion
				registry.dispatch( storeName ).setOwnerID( ownerID );
```

assets/js/googlesitekit/modules/datastore/sharing-settings.test.js
```diff
@@ -0,0 +1,252 @@
...
+				],
+				[ 'should not', {}, 1 ],
+			] )(
+				'dispatches a request to save sharing settings and %s dispatch setOwnerID action based on the `newOwnerIDs` availability',
+				async ( _, newOwnerIDs, ownerID ) => {
+					provideModules( registry, [
+						{
+							slug: 'search-console',
+							name: 'Search Console',
+							storeName: 'modules/search-console',
```
aaemnnosttv, line 176: Please use the proper constants where available.

aaemnnosttv (commented):
Thanks @hussain-t – looks like there's a few things left to do here. Would you please open a follow-up PR to address them?

PR #5245, opened 2022-05-22
Enhance/4795 - Implement new selectors for sharing settings

hussain-t:
## Summary

Addresses issue:

- #4795 

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist

### DIFF
--- a/assets/js/googlesitekit/modules/datastore/sharing-settings.js
+++ b/assets/js/googlesitekit/modules/datastore/sharing-settings.js
@@ -128,11 +128,10 @@ const baseActions = {
 	 */
 	*saveSharingSettings() {
 		const registry = yield Data.commonActions.getRegistry();
-		// TODO: Refactor to use `getSharingSettings` selector
-		// to obtain the `sharingSettings` from the state in 4795.
-		const { sharingSettings } = registry.stores[
-			CORE_MODULES
-		].store.getState();
+
+		const sharingSettings = registry
+			.select( CORE_MODULES )
+			.getSharingSettings();
 
 		const {
 			response,
--- a/assets/js/googlesitekit/modules/datastore/sharing-settings.js
+++ b/assets/js/googlesitekit/modules/datastore/sharing-settings.js
@@ -412,7 +412,7 @@ const baseSelectors = {
 			const storeName = select( CORE_MODULES ).getModuleStoreName(
 				moduleSlug
 			);
-			return !! select( storeName )?.canSubmitSharingChanges?.();
+			return !! select( storeName )?.canSubmitChanges?.();
 		}
 	),
 };
--- a/assets/js/googlesitekit/modules/datastore/sharing-settings.js
+++ b/assets/js/googlesitekit/modules/datastore/sharing-settings.js
@@ -277,6 +277,19 @@ const baseSelectors = {
 		const { sharingSettings } = state;
 		return sharingSettings;
 	},
+
+	/**
+	 * Gets the current dashboard shareable roles.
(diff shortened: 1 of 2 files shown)


assets/js/googlesitekit/modules/datastore/sharing-settings.js
```diff
@@ -221,15 +247,175 @@ const baseReducer = ( state, { type, payload } ) => {
...
+	 */
+	canSubmitSharingChanges: createRegistrySelector(
+		( select ) => ( state, moduleSlug ) => {
+			invariant( moduleSlug, 'moduleSlug is required.' );
+			const storeName = select( CORE_MODULES ).getModuleStoreName(
+				moduleSlug
+			);
+			return !! select( storeName )?.canSubmitChanges?.();
+		}
+	),
```
eugene-manuilov, line 417: Looks like there is an inaccuracy in IB for this selector. This selector shouldn't call `canSubmitChanges` selector for a module, it doesn't make sense. What we need to do here is to check whether the shared settings have changed and to use the approach similar to one that we use for `canSubmitChanges` where we have "safe" and "dangerous" versions of the `canSubmitChanges` selector.

Learn how we use the `createValidationSelector` function for the `createSubmitChangesStore` factory and the `makeDefaultCanSubmitChanges` function for default implementation. The main difference is that for this PR the `canSubmitSharingChanges` implementation should be module agnostic.

Also, pay attention that you will need to add a selector similar to `isDoingSubmitChanges` to check whether we are submitting sharing changes or not. Once again, you can see it's implementation in the `createSubmitChangesStore` factory.

cc @aaemnnosttv in case I miss something.

eugene-manuilov (changes requested):
Thanks, @hussain-t. This is a very good start. The majority of new functions look good, except the `canSubmitSharingChanges` one. Please, check my comment and update the selector.

assets/js/googlesitekit/modules/datastore/sharing-settings.js
aaemnnosttv: > Looks like there is an inaccuracy in IB for this selector. This selector shouldn't call `canSubmitChanges` selector for a module, it doesn't make sense. What we need to do here is to check whether the shared settings have changed and to use the approach similar to one that we use for `canSubmitChanges` where we have "safe" and "dangerous" versions of the `canSubmitChanges` selector.

@eugene-manuilov correct, the ACs and IB only mention to use `canSubmitChanges` as an example, it isn't relevant to call this here as you said.

> Also, pay attention that you will need to add a selector similar to `isDoingSubmitChanges` to check whether we are submitting sharing changes or not.

I'm not sure if we need this for sharing settings yet. This is more relevant for module settings because there may be multiple requests needed in order to submit all the changes. In the case of sharing settings, we should be able to get by with just the `isFetching*` selector generated by the `saveSharingSettings` fetch store. We can always wrap it later if needed with the same kind of abstraction if needed.

aaemnnosttv: Chatting with @hussain-t about this, we decided to keep the `isDoingSubmitChanges` selector as the wrapping actions are already there in the save action, it just needs the reducer cases and selector as a simple getter. Relying on the `isFetching` action could be problematic as owner IDs will change after the request is complete if owners change.

eugene-manuilov (approved):
Awesome! Thanks, @hussain-t. Great work!

PR #5246, opened 2022-05-22
Enhancement/4825 Restrict edit settings

jimmymadon:
## Summary

Addresses issue:

- #4825 

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist

### DIFF
--- a/assets/js/components/settings/EntityOwnershipChangeNotice.js
+++ b/assets/js/components/settings/EntityOwnershipChangeNotice.js
@@ -1,0 +1,86 @@
+/**
+ * EntityOwnershipChangeNotice component.
+ *
+ * Site Kit by Google, Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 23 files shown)


aaemnnosttv (commented):
@jimmymadon this looks like a good start but I think the approach here is a bit too broad and has resulted in more changes than should be necessary. It seems like there were some details missed in IBR but here are the main things that stand out to me so far:

- The changes here should only apply within the module **settings edit** views. This means that most of the changes should live in settings-specific components. Any changes needed in common components should ideally be controlled by props rather than trying to make them aware of which view they're in, that way the parent view-specific component can control that
- Module access checks should only be done for users other than the owner as this issue is specific to logic around change of ownership. This is largely the reason so many tests + E2E needed updating because the selector is being invoked unconditionally (again, in a common component)
- We need to account for the user-without-access case and potentially use a different component in place of the current `*Select` because we won't have the data in the store to populate its choices – we can instead use a placeholder component that renders the raw setting value rather than a "rich" choice that often includes the entity display name which we won't have as that comes from the API

I think much of what you have here is still usable, but just about all of the current tests should not need changes apart from maybe adding a call to `provideUserInfo`.

Let's see how we can implement this with minimal changes to existing components and tests.

Issue #5236, opened 2022-05-19
Add "Retry" button to ReportError component


tofumatt:
## Feature Description

Currently when the user encounters an error that causes `ReportError` to be displayed, there is no way to retry the request. We should add a "retry" button that invalidates the request's resolver and forces the request to be tried again.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A "Retry" button should be added to the `ReportError` component whenever the return value from `getErrorForSelector ` contains the `selectorData` object.
* Other than that, there are a few exceptions where the button should **not** be shown if any of the following is true:
    * If the `selectorName` is not `getReport` (because then it could be any other type of error that is not from the report API request) - @eugene-manuilov raised this earlier today, and it makes sense to me.
    * If the error is an "insufficient permissions error", i.e. the user doesn't have access to this data. In that case, they will have to coordinate with the admin who does have access so that they can get access to. That's not exactly a quick task (unless they're sitting right next to them), so we probably shouldn't show the Retry button there as it's not going to help.
    * If the error is a "permission scope error", i.e. the user hasn't granted certain OAuth scopes. In that case they will see a banner notification to fix this anyway, and retrying the request won't do it.
    * If the error has a `reconnectURL` with it, i.e. there is already a CTA URL present which can fix the problem - again retrying the request won't do it in this case.
* When activated, the button should use this code:

```js
dispatch(selectorData.storeName)
  .invalidateResolution(
    selectorData.name,
    selectorData.args
  );
```

to clear the error.

## Implementation Brief

* In `assets/js/components/notifications/CTA.js` component:
    * Take two new props `retry` which is a boolean and `onRetry` which is a function. 
        * Both props are optional.
    * When the `retry` prop is `true`, render a `Button` component with the following:
        * `onClick` should be `onRetry`.
        * Children should be translated string `Retry`. 
* In `assets/js/components/ReportError.js` component: 
    * When the passed `error` prop has a `selectorData` object-
        * Define a callback function for the retry handler which dispatches the `invalidateResolution` action like it is shown in the AC. 
        * Pass `retry` to the `CTA` element.
        * Pass the aforementioned retry handler function to the `onRetry` prop of  the `CTA` element. 

### Test Coverage

* Add story with the Retry Button on `assets/js/components/ReportError.stories.js`. We can do it without waiting for the blocker issue by passing an object of same shape as the `error` prop.

## QA Brief

[...]

PR #5337, opened 2022-06-12
Feat/5236 - Add Retry button to ReportError component

hussain-t:
## Summary

Addresses issue:

- #5236 

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist


felixarntz (commented):
@hussain-t Looks good so far. I left a few suggestions here for your consideration.

assets/js/components/ReportError.js
```diff
@@ -67,7 +70,27 @@ export default function ReportError( { moduleSlug, error } ) {
 		purify.sanitize( message, { ALLOWED_TAGS: [] } )
 	);
 
-	return <CTA title={ title } description={ description } error />;
+	const retry = !! error?.selectorData;
+
+	const handleRetry = useCallback( () => {
+		if ( retry ) {
+			const { selectorData } = error;
+			dispatch( selectorData.storeName ).invalidateResolution(
```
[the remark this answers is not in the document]
hussain-t: `storeName` is a required param in `createErrorStore`. If I'm not wrong it will be always set. However, we can do the check here 👍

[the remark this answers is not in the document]
felixarntz: Correct, but better safe than sorry :)

assets/js/components/notifications/CTA.js
```diff
@@ -51,7 +53,10 @@ const CTA = ( {
 				{ description }
 			</div>
 		) }
-		{ ctaLabel && (
+		{ ctaLabel && ctaType === 'button' && (
+			<Button onClick={ onClick }>{ ctaLabel }</Button>
+		) }
+		{ ctaLabel && ctaType === 'link' && (
```
felixarntz, line 59: Let's keep using the existing condition where currently the `Link` is printed. The `Button` component largely supports the same props (it can also be used for a semantic link that _looks_ like a button).

For example, we could do something like this:
```js
const Tag = ctaType === 'button' ? Button : Link;
```

And then render that via `<Tag ...` instead of the hard-coded `<Link ...` below.

felixarntz (commented):
@hussain-t One more follow-up comment, other than that lgtm.

@tofumatt Would be great if you do a quick follow-up review here as well.

assets/js/components/notifications/CTA.js
hussain-t: @felixarntz, we get warnings for the invalid props passed to the `Button` component. IMO, we could keep it separate to avoid these warnings. @tofumatt, WDYT?

Screenshot 2022-06-15 at 11 15 35 AM

hussain-t:
@felixarntz, @tofumatt, should we check for `const retry = error?.selectorData?.storeName === 'getReport'` to render the retry button?

felixarntz (commented):
@hussain-t Left a reply on your feedback. Also please review the higher-level comment with potential refinement in https://github.com/google/site-kit-wp/issues/5236#issuecomment-1156725774.

felixarntz:
@hussain-t 
> @felixarntz, @tofumatt, should we check for `const retry = error?.selectorData?.storeName === 'getReport'` to render the retry button?

No, `getReport` refers to the _selector_ name. So we need to check it like:
```js
const retry = !! error?.selectorData?.storeName && error?.selectorData?.selectorName === 'getReport';
```

assets/js/components/ReportError.js
```diff
@@ -67,7 +70,28 @@ export default function ReportError( { moduleSlug, error } ) {
 		purify.sanitize( message, { ALLOWED_TAGS: [] } )
 	);
 
-	return <CTA title={ title } description={ description } error />;
+	const retry = !! error?.selectorData?.storeName;
```
tofumatt, line 73: This variable name could be a bit more descriptive, and maybe check for all the arguments required to invalidate a selector, eg:

```suggestion
	const showRetry = error?.selectorData?.storeName && error?.selectorData?.storeName && error?.selectorData?.name && selectorData?.args;
```

assets/js/components/ReportError.js
```diff
@@ -67,7 +70,28 @@ export default function ReportError( { moduleSlug, error } ) {
...
+
+	const handleRetry = useCallback( () => {
+		if ( retry ) {
+			const { selectorData } = error;
+			dispatch( selectorData.storeName ).invalidateResolution(
+				selectorData.name,
+				selectorData.args
+			);
+		}
```
tofumatt, line 82: I don't think this check here is required, because this callback is only set when `retry` is truth-y. This can be removed, eg.

```suggestion
		const { selectorData } = error;
		
		dispatch( selectorData.storeName ).invalidateResolution(
			selectorData.name,
			selectorData.args
		);
```

assets/js/components/ReportError.test.js
```diff
@@ -97,4 +98,38 @@ describe( 'ReportError', () => {
 			'Insufficient permissions in Test Module'
 		);
 	} );
+
+	it( 'renders the `Retry` button if the error has `selectorData` property', () => {
```
tofumatt, line 102: This test spec isn't quite accurate, as we check for attributes _on_ the `selectorData` object, not just for the presence of `selectorData` at all. Arguably a test that has `selectorData: {}` (shouldn't ever happen, but just in case) should _not_ show the retry button and would be worth having.

This test should be named a bit more explicitly.

assets/js/components/notifications/CTA.js
tofumatt: Based on the type you could pass in the correct args, eg. assemble the relevant props by doing something like:

```js
const ctaProps = {};

if (ctaType === 'button') {
  ctaProps.onClick = onClick;
  ctaProps.href = ctaLink;
}

if (ctaType === 'link') {
  ctaProps.href = ctaLink;
  ctaProps.hideExternalIndicator = ctaLinkExternal;
}
```

etc.

That said I think this is a more roundabout approach, I'd prefer just outputting the components separately as is done here, because then all the props and how they map to each component is more explicit.

That said, at least the `aria-label` and `href` props seem like they should be passed to `Button` as it should accept them.

tofumatt (changes requested):
I had a look through and I think there are still a few changes needed to the code, but overall this looks like the right approach.

assets/js/components/notifications/CTA.js
hussain-t: Honestly, I prefer to keep the components separately and pass the props explicitly instead of passing them as 
`{ ...ctaProps }`. Let me pass the `aria-label` and `href` props to the `Button` component.

hussain-t:
> @hussain-t
> 
> > @felixarntz, @tofumatt, should we check for `const retry = error?.selectorData?.storeName === 'getReport'` to render the retry button?
> 
> No, `getReport` refers to the _selector_ name. So we need to check it like:
> 
> ```js
> const retry = !! error?.selectorData?.storeName && error?.selectorData?.selectorName === 'getReport';
> ```

Sorry, that's what I meant. Thanks 👍

assets/js/components/ReportError.js
hussain-t: I think checking for the `storeName` and the `name === 'getReport'` should be sufficient like so:
```js
const showRetry = !! (
	error?.selectorData?.storeName &&
	error.selectorData?.name === 'getReport'
);
```

felixarntz (approved):
@hussain-t Great work, thanks! Just one comment worth addressing, but not a blocker.

Let's also wait for @tofumatt's follow-up review.

assets/js/components/ReportError.js
```diff
@@ -67,7 +74,35 @@ export default function ReportError( { moduleSlug, error } ) {
 		purify.sanitize( message, { ALLOWED_TAGS: [] } )
 	);
 
-	return <CTA title={ title } description={ description } error />;
+	const showRetry =
+		!! (
+			error?.selectorData?.storeName &&
+			error.selectorData?.name === 'getReport'
+		) &&
```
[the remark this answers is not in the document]
hussain-t: That's a good catch, @felixarntz 👍 I have addressed and pushed the changes.

assets/js/util/errors.test.js
```diff
@@ -115,10 +116,37 @@ describe( 'Error Utilities', () => {
...
+					reason: 'dailyLimitExceeded',
+				},
+			};
+
+			expect( isAuthError( error ) ).toBe( false );
+		} );
+
+		it( 'should return FALSE if the passed object does not have the code property', () => {
+			expect( isAuthError( { message: 'Not Found' } ) ).toBe( false );
+		} );
```
tofumatt, line 142: Let's make the declaration of the error inline with the other tests here so it's easier to parse:

```suggestion
		it( 'should return FALSE if the passed object does not have the `data` property', () => {
			const error = {
				message: 'Not Found',
			};

			expect( isAuthError( error ) ).toBe( false );
		} );
```

tofumatt (commented):
Looks good! Just a few small tweaks to make for readability and such; I'll commit them, test out the branch, then get this merged 🙂

hussain-t:
@tofumatt, thanks for applying the changes. However, the JS tests are failing due to this change. We will have to escape the single quote character.

Screenshot 2022-06-22 at 6 11 30 PM

tofumatt (approved):
Tested locally and worked a treat! 👍🏻

Issue #5356, opened 2022-06-15
Incorrect Analytics Properties defaulted when Site Kit is Set up, then Reset/Setup 


wpdarren:
## Bug Description

Google Analytics connect service screen does not automatically add the correct UA/GA4 properties when you disconnect/reset Site Kit after setting up Site Kit and Analytics. 

![analytics.png](https://images.zenhubusercontent.com/5f99965247ef5a1da6985d7d/fc5f3920-a1cb-4c2d-ab69-9e9de96f3c4e)

When I switched to the latest release (1.75.0) the connect service page loads as expected.

No feature flags set with tester plugin.

Please note that even if you do not include Analytics within the Site Kit Set up (from the splash screen) but instead set up Analytics via the CTA on Site Kit Dashboard, the same issue occurs. 

## Steps to reproduce

Here are the steps I took:

1. New site. Set up Site Kit with Analytics.  All set up fine, no issues.
2. Disconnect/Reset Site Kit and redirected to Set up Site Kit Splash page.
3. Reconnect Site Kit with Analytics, and when you get to the connect service page, the primary property is GA4 and nothing in secondary property selected.

I was expecting to see primary property as UA property and secondary property as GA4.  

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* After resetting Site Kit and connecting Analytics again, if a UA account/property/profile that matches the site is found it should be selected and should appear as the "primary" analytics property selected (eg. at the top of the form).

## Implementation Brief

The problem is related to a flaw in the data request infrastructure that caused non-shared requests to have the client defer set improperly before the `parse_data_response` was called.
* Merge https://github.com/google/site-kit-wp/pull/5367 

### Test Coverage

* No changes

## QA Brief

* Follow the steps to reproduce above, the primary matched property should match a UA Analytics property and the secondary property should also match a respective GA4 property

## Changelog entry

* Fix bug that could cause Analytics properties not to be pre-selected properly during setup.

felixarntz:
I see you self-assigned this @tofumatt but just looked at this too, from my point of view IB + CR ✅

tofumatt:
IB ✅

wpdarren:
### QA Update: ✅ 

**Verified:**

- When a site is set up with Site Kit and Analytics, then reset. The primary matched property defaults in correctly and matches the UA Analytics property and the secondary property also matches a respective GA4 property.
[...]

PR #5367, opened 2022-06-15
Fix datapoint request execution

aaemnnosttv:
## Summary

Addresses issue:

- #5356

## Relevant technical choices

This PR fixes the reported issue by fixing a flaw in the datapoint request execution which resulted in the client having a "stacked" defer that resulted in the request for Analytics account summaries in the `accounts-properties-profiles` parse response method being deferred and thus not executed.
https://github.com/google/site-kit-wp/blob/abeb47e26e5d4ffea8849fa1ec44796780a28337/includes/Modules/Analytics.php#L781

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist

### DIFF
--- a/includes/Core/Modules/Module.php
+++ b/includes/Core/Modules/Module.php
@@ -343,7 +343,7 @@ abstract class Module {
 			// even if a different client will be the one to execute the request because
 			// the default instance is what services are setup with.
 			$restore_defers[] = $this->get_client()->withDefer( true );
-			if ( $this->get_client() !== $oauth_client ) {
+			if ( $this->authentication->get_oauth_client() !== $oauth_client ) {
 				$restore_defers[] = $oauth_client->get_client()->withDefer( true );
 			}
 


includes/Core/Modules/Module.php
```diff
@@ -343,7 +343,7 @@ final protected function execute_data_request( Data_Request $data ) {
 			// even if a different client will be the one to execute the request because
 			// the default instance is what services are setup with.
 			$restore_defers[] = $this->get_client()->withDefer( true );
-			if ( $this->get_client() !== $oauth_client ) {
+			if ( $this->authentication->get_oauth_client() !== $oauth_client ) {
```
aaemnnosttv, line 346: The purpose of this condition is to avoid stacking the defer; each callback in the array is intended to reference a restore for a different client.

This logical flaw resulted in a condition that would always be true, and so non-shared datapoint requests would be restored improperly. This broke the request in the `Analytics::parse_data_response` method which was expected to run without deferral.

felixarntz: This makes me think, we may wanna do away with that `get_client()` method completely and instead just have a method like `get_oauth_client_for_current_user()`. That would fit more nicely with `get_oauth_client_for_datapoint()` in that both would return an `OAuth_Client_Base` instance rather than one returning a `Google_Site_Kit_Client` instance which is completely different.

felixarntz (approved):
@aaemnnosttv Great catch, lgtm!

tofumatt (approved):
Looks good, though I liked Felix's idea here: https://github.com/google/site-kit-wp/pull/5367/files#r898482631

Maybe we should have a follow-up issue for that? 😄

includes/Core/Modules/Module.php
tofumatt: I filed #5368 for this change and left it in AC for now, but I like that idea 👍🏻

PR #5504, opened 2022-07-05
Add error handling to `<ModuleRecoveryAlert />`

nfmohit:
## Summary

Addresses issue:

- #5318 

## Relevant technical choices

This PR adds an `<Error />` component to `<ModuleRecoveryAlert />` which looks for errors in the API response of module recovery and displays them in the component.

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist

### DIFF
--- a/assets/js/components/dashboard-sharing/ModuleRecoveryAlert/Error.js
+++ b/assets/js/components/dashboard-sharing/ModuleRecoveryAlert/Error.js
@@ -1,0 +1,53 @@
+/**
+ * ModuleRecoveryAlert Error component.
+ *
+ * Site Kit by Google, Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Fragment } from '@wordpress/element';
+import { sprintf, __ } from '@wordpress/i18n';
+
+export default function Error( { recoverableModules, error } ) {
+	return (
+		<div className="googlesitekit-module-recovery-error">
+			{ Object.keys( error ).length > 1 ? (
+				<Fragment>
+					<p>
+						{ __(
+							'The following modules failed to be recovered:',
+							'google-site-kit'
+						) }
+					</p>
+					<ul>
+						{ Object.keys( error ).map( ( module ) => (
+							<li key={ module }>
+								{ recoverableModules[ module ].name }:{ ' ' }
+								{ error[ module ].message }
+							</li>
+						) ) }
+					</ul>
+				</Fragment>
+			) : (
+				<p>
+					{ sprintf(
+						/* translators: %s: Error message */
+						__( 'Error: %s', 'google-site-kit' ),
+						Object.values( error )[ 0 ].message
+					) }
+				</p>
+			) }
+		</div>
+	);
+}
--- a/assets/js/components/dashboard-sharing/ModuleRecoveryAlert/Error.js
+++ b/assets/js/components/dashboard-sharing/ModuleRecoveryAlert/Error.js
@@ -16,6 +16,9 @@
  * limitations under the License.
  */
 
+/**
+ * WordPress dependencies
+ */
 import { Fragment } from '@wordpress/element';
 import { sprintf, __ } from '@wordpress/i18n';
 
--- a/assets/js/components/dashboard-sharing/ModuleRecoveryAlert/Error.js
+++ b/assets/js/components/dashboard-sharing/ModuleRecoveryAlert/Error.js
@@ -25,7 +25,17 @@ import { sprintf, __ } from '@wordpress/i18n';
 export default function Error( { recoverableModules, error } ) {
 	return (
 		<div className="googlesitekit-module-recovery-error">
-			{ Object.keys( error ).length > 1 ? (
+			{ Object.keys( error ).length === 1 && (
+				<p>
+					{ sprintf(
+						/* translators: %s: Error message */
+						__( 'Error: %s', 'google-site-kit' ),
+						Object.values( error )[ 0 ].message
+					) }
+				</p>
+			) }
+
+			{ Object.keys( error ).length > 1 && (
 				<Fragment>
 					<p>
 						{ __(
@@ -42,14 +52,6 @@ export default function Error( { recoverableModules, error } ) {
 						) ) }
 					</ul>
 				</Fragment>
-			) : (
-				<p>
-					{ sprintf(
-						/* translators: %s: Error message */
-						__( 'Error: %s', 'google-site-kit' ),
-						Object.values( error )[ 0 ].message
-					) }
-				</p>
 			) }
 		</div>
 	);
--- a/assets/js/components/dashboard-sharing/ModuleRecoveryAlert/Error.js
+++ b/assets/js/components/dashboard-sharing/ModuleRecoveryAlert/Error.js
@@ -17,6 +17,11 @@
  */
 
 /**
+ * External dependencies
+ */
+import PropTypes from 'prop-types';
+
+/**
  * WordPress dependencies
  */
 import { Fragment } from '@wordpress/element';
@@ -56,3 +61,8 @@ export default function Error( { recoverableModules, error } ) {
 		</div>
 	);
 }
+
+Error.propTypes = {
+	recoverableModules: PropTypes.object,
+	error: PropTypes.object,
+};
--- a/assets/js/components/dashboard-sharing/ModuleRecoveryAlert/Errors.js
+++ b/assets/js/components/dashboard-sharing/ModuleRecoveryAlert/Errors.js
@@ -51,8 +51,12 @@ export default function Errors( { recoveryErrors } ) {
 					<ul>
 						{ Object.keys( recoveryErrors ).map( ( module ) => (
 							<li key={ module }>
-								{ recoveryErrors[ module ].name }:{ ' ' }
-								{ recoveryErrors[ module ].message }
+								{ sprintf(
+									/* translators: 1: Module name, 2: Error message */
+									__( '%1$s: %2$s', 'google-site-kit' ),
+									recoveryErrors[ module ].name,
+									recoveryErrors[ module ].message
+								) }
 							</li>
 						) ) }
 					</ul>
--- a/assets/js/components/dashboard-sharing/ModuleRecoveryAlert/Errors.js
+++ b/assets/js/components/dashboard-sharing/ModuleRecoveryAlert/Errors.js
@@ -27,20 +27,20 @@ import PropTypes from 'prop-types';
 import { Fragment } from '@wordpress/element';
 import { sprintf, __ } from '@wordpress/i18n';
 
-export default function Error( { recoverableModules, error } ) {
+export default function Errors( { recoveryErrors } ) {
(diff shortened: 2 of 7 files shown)


hussain-t (changes requested):
Thanks, @nfmohit. This is a good foundation. I have left a few comments to consider. Additionally, it would be better to add two stories. One is for single module recovery, and the other is for multiple module recovery.

assets/js/components/dashboard-sharing/ModuleRecoveryAlert/Error.js
```diff
@@ -0,0 +1,56 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * WordPress dependencies
+ */
+import { Fragment } from '@wordpress/element';
+import { sprintf, __ } from '@wordpress/i18n';
+
+export default function Error( { recoverableModules, error } ) {
```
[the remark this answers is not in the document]
nfmohit: Added. Thank you!

assets/js/components/dashboard-sharing/ModuleRecoveryAlert/Error.js
```diff
@@ -0,0 +1,56 @@
...
+	return (
+		<div className="googlesitekit-module-recovery-error">
+			{ Object.keys( error ).length > 1 ? (
```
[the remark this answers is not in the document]
nfmohit: Understood. Removed ternary and used `&&` with separate conditions. Thank you!

assets/js/components/dashboard-sharing/ModuleRecoveryAlert/index.js
```diff
@@ -26,18 +26,20 @@ import { sprintf, __ } from '@wordpress/i18n';
...
+import ProgressBar from '../../ProgressBar';
+import Spinner from '../../Spinner';
+import Error from './Error';
 
 const { useDispatch, useSelect } = Data;
 
 export default function ModuleRecoveryAlert() {
 	const [ checkboxes, setCheckboxes ] = useState( null );
 	const [ recoveringModules, setRecoveringModules ] = useState( false );
+	const [ error, setError ] = useState( false );
```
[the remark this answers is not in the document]
nfmohit: Updated. Thank you!

assets/js/components/dashboard-sharing/ModuleRecoveryAlert/index.js
```diff
@@ -80,15 +82,25 @@ export default function ModuleRecoveryAlert() {
 	);
 
 	const handleRecoverModules = useCallback( () => {
+		// Reset error state.
+		if ( error ) {
+			setError( false );
```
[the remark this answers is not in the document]
nfmohit: Updated. Thank you!

nfmohit:
> Thanks, @nfmohit. This is a good foundation. I have left a few comments to consider. Additionally, it would be better to add two stories. One is for single module recovery, and the other is for multiple module recovery.

Thank you for the kind review @hussain-t! I've addressed the suggestions. Please let me know if it doesn't look good.

assets/js/components/dashboard-sharing/ModuleRecoveryAlert/index.js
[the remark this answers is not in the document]
aaemnnosttv: As an aside, we generally use `null` to be an explicit empty/none value.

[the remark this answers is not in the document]
nfmohit: Understood and duly noted. I'll use `null` in such cases from now on. Thank you for the explanation @aaemnnosttv!

assets/js/components/dashboard-sharing/ModuleRecoveryAlert/index.stories.js
```diff
@@ -153,6 +159,88 @@ MultipleRecoverableModuleNoAccess.scenario = {
...
+				{ slug: 'analytics' }
+			);
+	},
+};
+MultipleRecoverableModuleErrors.decorators = [
+	( Story ) => {
+		setTimeout( async () => {
+			await waitForElement( () => document.querySelector( 'button' ) );
+			document.querySelector( 'button' ).click();
+		}, 0 );
```
aaemnnosttv, line 238: This isn't something we've done in other stories AFAIK mostly because errors are stored in the datastore automatically by our fetch store infrastructure, which makes it much more straightforward to put a component in such a state.

Let's source the error(s) from the store instead rather than component state using the `getErrorForAction` selector which every store should have.

assets/js/components/dashboard-sharing/ModuleRecoveryAlert/Error.js
```diff
@@ -0,0 +1,68 @@
...
+					</ul>
+				</Fragment>
+			) }
+		</div>
+	);
+}
+
+Error.propTypes = {
+	recoverableModules: PropTypes.object,
+	error: PropTypes.object,
```
aaemnnosttv, line 67: Here the prop type shows `error` is not required, but it is expected to be provided above and `Object.keys` will throw if it does not receive a compatible type.

Secondarily, `error` may not be the appropriate prop here as this object is expected to have a shape which is different than most other errors (e.g. `{ code, message }`). Perhaps `recoveryErrors` may be more appropriate?

Another idea would be to pass only the slugs of the modules which failed and the errors can be selected for each by this component. I'm open to other ideas too, just a thought.

assets/js/components/dashboard-sharing/ModuleRecoveryAlert/index.js
```diff
@@ -26,18 +26,20 @@ import { sprintf, __ } from '@wordpress/i18n';
...
+import ProgressBar from '../../ProgressBar';
+import Spinner from '../../Spinner';
+import Error from './Error';
 
 const { useDispatch, useSelect } = Data;
 
 export default function ModuleRecoveryAlert() {
 	const [ checkboxes, setCheckboxes ] = useState( null );
 	const [ recoveringModules, setRecoveringModules ] = useState( false );
+	const [ error, setError ] = useState( null );
```
aaemnnosttv, line 42: As mentioned before, let's source this from the datastore instead.

assets/js/components/dashboard-sharing/ModuleRecoveryAlert/Error.js
```diff
@@ -0,0 +1,68 @@
...
+ */
+import PropTypes from 'prop-types';
+
+/**
+ * WordPress dependencies
+ */
+import { Fragment } from '@wordpress/element';
+import { sprintf, __ } from '@wordpress/i18n';
+
+export default function Error( { recoverableModules, error } ) {
```
aaemnnosttv, line 30: Let's rename this as plural as it can handle many, it only handles a single error as a special case.

aaemnnosttv (changes requested):
Thanks @nfmohit ! This is almost there – after taking a closer look, I realized this should use a slightly different approach for the errors for consistency and to avoid duplicating state that's already available.

assets/js/components/dashboard-sharing/ModuleRecoveryAlert/index.stories.js
nfmohit: I have addressed this so that the error can be sourced directly from the reducer. Thank you!

assets/js/components/dashboard-sharing/ModuleRecoveryAlert/Error.js
nfmohit: I went with `recoveryErrors`. It was a bit tricky to keep track of the modules which failed to recover without storing in a local state, but I was able to accomplish it with a different approach. Please let me know if it looks good. Thank you!

assets/js/components/dashboard-sharing/ModuleRecoveryAlert/index.js
nfmohit: Updated, thank you!

assets/js/components/dashboard-sharing/ModuleRecoveryAlert/Error.js
nfmohit: Updated, thank you!

nfmohit:
> Thanks @nfmohit ! This is almost there – after taking a closer look, I realized this should use a slightly different approach for the errors for consistency and to avoid duplicating state that's already available.

Thank you so much for the detailed review @aaemnnosttv! I have addressed the suggestions and updated the error handling so that the errors are sourced from the datastore instead of storing them in a local state. Please let me know if it looks good.

Thank you!

assets/js/components/dashboard-sharing/ModuleRecoveryAlert/Errors.js
```diff
@@ -0,0 +1,67 @@
...
+						{ __(
+							'The following modules failed to be recovered:',
+							'google-site-kit'
+						) }
+					</p>
+					<ul>
+						{ Object.keys( recoveryErrors ).map( ( module ) => (
+							<li key={ module }>
+								{ recoveryErrors[ module ].name }:{ ' ' }
+								{ recoveryErrors[ module ].message }
```
aaemnnosttv, line 55: Oops, not sure if I missed this before or not, but this should use a translatable string for the concatenation for i18n.

assets/js/components/dashboard-sharing/ModuleRecoveryAlert/index.js
```diff
@@ -84,11 +111,13 @@ export default function ModuleRecoveryAlert() {
...
 		);
-		recoverModules( modulesToRecover ).finally( () => {
-			setRecoveringModules( false );
-			setCheckboxes( null );
+		clearErrors( 'recoverModule' ).then( () => {
+			recoverModules( modulesToRecover ).finally( () => {
+				setRecoveringModules( false );
+				setCheckboxes( null );
+			} );
 		} );
```
aaemnnosttv, line 119: Please use async/await for consistency. The `finally` shouldn't be necessary either as `recoverModules` will only throw if it's given invalid input. It won't throw if the request fails etc. `clearErrors` also won't throw. It isn't even async but all bound actions return promises as part of the wp data flow.

assets/js/components/dashboard-sharing/ModuleRecoveryAlert/Errors.js
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * WordPress dependencies
+ */
+import { Fragment } from '@wordpress/element';
+import { sprintf, __ } from '@wordpress/i18n';
+
+export default function Errors( { recoveryErrors } ) {
+	return (
+		<div className="googlesitekit-module-recovery-error">
```
aaemnnosttv, line 32: Minor detail, but let's make the class name plural for consistency with the component but also because it includes a list of errors 😄

aaemnnosttv (commented):
Thanks @nfmohit ! A handful of minor things left but this looks great 👍

assets/js/components/dashboard-sharing/ModuleRecoveryAlert/Errors.js
nfmohit: My bad! Thank you for spotting. Updated.

assets/js/components/dashboard-sharing/ModuleRecoveryAlert/index.js
nfmohit: Thank you for pointing out a best practice here! I'll keep it in mind for future cases. Updated it to use async/await.

assets/js/components/dashboard-sharing/ModuleRecoveryAlert/Errors.js
nfmohit: Sharp spot! Apologies for missing this. Updated.

nfmohit:
> Thanks @nfmohit ! A handful of minor things left but this looks great 👍

Eagle eyes! Thank you for spotting these @aaemnnosttv and my apologies for missing them. I've addressed the suggestions. Please let me know if this looks good now.

Thank you!

aaemnnosttv (approved):
Great, thanks @nfmohit !

Issue #5382, opened 2022-06-16
Extend Dashboard Sharing feature tour to include settings interface


aaemnnosttv:
## Feature Description

In #5328, we added a basic feature tour for dashboard sharing which informs an authenticated admin about the added ability to share the dashboard with other users but does not elaborate on how to use it.

This issue is about extending this tour with additional steps to walk the user through the sharing settings interface and better explain how to share the dashboard and or delegate sharing management.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

The feature tour for dashboard sharing should be extended as follows:
* The tour should get additional steps to walk the user through the sharing settings interface
	* A second step should be added:
		* Title: `Manage view access for other roles` 
		* Body: `Grant access to the view-only dashboard for each service for the specific roles you want. Users will see the Site Kit dashboard with only the services that have been shared with them without needing to sign-in with Google.`
		* On display, the dashboard sharing settings modal should be toggled open
		* Focus: "Who can view" column
	* A third step should be added but only if the site has multiple admins (otherwise this column will not be displayed/relevant)
		* Title: `Share management with other admins`
		* Body: `By default only the user who connects a service can control who it is shared with. This setting optionally lets these users allow any other admin signed in with Google to manage the roles a service is shared with.`
		* Focus: "Who can manage" column
	* On dismiss/completion the dashboard sharing settings modal should be closed
* If the user dismisses the feature tour on the first step (showing settings button) they should still see the remaining tour steps when viewing the sharing settings interface for the first time (i.e. on demand)
	* Dismissing on step 2 or 3 should dismiss the full tour as usual

Note: these could technically be separate tours with shared steps, etc – it is not strictly required that everything is defined in a single tour.

## Implementation Brief

* Update `DashboardSharingSettingsButton` to use the `CORE_UI` store for the `dialogOpen` state (e.g. `dashboardSharingDialogOpen`)
* Update `_googlesitekitDashboardSharingData` to include the value of `hasMultipleAdmins` from `Has_Multiple_Admins`
* Define a new `dashboardSharingSettings` feature tour
	- The first step should be defined as the "second" step in the ACs
		- The `target` should use `.googlesitekit-dashboard-sharing-settings__main .googlesitekit-dashboard-sharing-settings__column--view` to highlight the first row's user role select control
[...]

PR #5516, opened 2022-07-07
Extend Dashboard Sharing sharing feature tour with more steps and on-demand part

aaemnnosttv:
## Summary

Addresses issue:

- #5382

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist


assets/js/components/dashboard-sharing/DashboardSharingSettingsButton.js
```diff
@@ -95,6 +101,15 @@ export default function DashboardSharingSettingsButton() {
 		}
 	}, [ dialogOpen, haveSettingsChanged, rollbackSharingSettings ] );
 
+	const triggeredTourRef = useRef();
```
aaemnnosttv, line 104: Used a ref here because we don't need a re-render when this changes, it's only to prevent triggering a second time.

assets/js/feature-tours/dashboard-sharing-settings.js
```diff
@@ -0,0 +1,65 @@
...
+		),
+		placement: 'auto',
+	} );
+}
+
+export default {
+	slug: 'dashboardSharingSettings',
+	// TODO: This version should be changed when the feature flag is removed
+	// to the actual upcoming version.
+	version: '1.85.0',
```
aaemnnosttv, line 61: As an on-demand-only tour right now, this currently isn't checked but thinking about it now, it should probably still be supported but optional (for on demand only).

assets/js/feature-tours/dashboard-sharing-settings.js
```diff
@@ -0,0 +1,65 @@
...
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
+/*
+ * Internal dependencies
+ */
+import { isFeatureEnabled } from '../features';
+
+const { hasMultipleAdmins } = global._googlesitekitUserData || {};
```
aaemnnosttv, line 29: This state was initially planned as going in the dashboard sharing global but that one loads too late to be usable here without making the user store depend on it, which seems wrong. This state isn't specific to the current user but is user-related, and also the `core/user` store is where feature tours are built into, so from that perspective I think this makes sense.

assets/sass/components/feature-tours/_googlesitekit-dashboard-sharing.scss
```diff
@@ -0,0 +1,43 @@
+/**
+ * Dashboard Sharing Feature Tour styles.
```
aaemnnosttv, line 2: This is a workaround I came up with to solve an unforseen problem with Joyride which doesn't seem to support dynamically revealing a next step's `target` element while transitioning from the first. Even with the events it provides, I couldn't find one that allowed for this to happen without making the entire tour `controlled`. More details below, but this essentially tricks Joyride into thinking the target is visible while keeping it hidden, but only while the main DS tour is active.

There may be more needed here for accessibility.

aaemnnosttv (commented):
@tofumatt in the interest of time, this is ready for an initial review. Everything works pretty well but there may be a thing or to we want to clean up or tweak before merging so I've left it as a draft for now.

tofumatt (changes requested):
Works great (no more bugs with completing the tour), and I tested a bunch of scenarios re: completing the tour, stopping halfway through, going back-and-forth through the tour/modal… all worked excellently and very seamlessly.

I noticed this no longer allows us to click on the Dashboard Sharing button to activate its modal and show the next step of the tour, but that's minor and arguably not needed:

CleanShot 2022-07-07 at 15 05 02

It was in a proof-of-concept earlier but I actually don't think it's required. The "Next" button is fine for that and works great.

I'm checking through any potential accessibility issues with the Joyride workaround proposed here, but otherwise only minor comments and then this is good-to-merge. Awesome stuff, thanks so much! 👍🏻

tofumatt (commented):
Just as an update: this seems to work fine when using screen-readers. So just the minor points I raised and then this is good-to-go in my view. 👍🏻

assets/sass/components/feature-tours/_googlesitekit-dashboard-sharing.scss
tofumatt: Update: seems good; doesn't pose any issues with JAWS or macOS VoiceOver.

assets/js/googlesitekit/datastore/user/feature-tours.js
```diff
@@ -204,6 +261,26 @@ const baseControls = {
 			return true;
 		}
 	),
+	[ CHECK_ON_DEMAND_TOUR_REQUIREMENTS ]: createRegistryControl(
+		( registry ) => async ( { payload } ) => {
+			const { tour } = payload;
+			// Check if the tour has already been dismissed.
+			// Here we need to first await the underlying selector with the asynchronous resolver.
```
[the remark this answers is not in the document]
aaemnnosttv: It's copied from the existing control for checking non-on-demand tours 😄

assets/js/feature-tours/dashboard-sharing-settings.js
aaemnnosttv: In the interest of time I've opted to remove this from the tour definition for now as we don't have a clear case for it. Perhaps this could be baked into its `checkRequirements` function later as well if we needed it. Otherwise this undermines the utility of an on-demand tour. We can always add this in later if needed.

assets/sass/components/feature-tours/_googlesitekit-dashboard-sharing.scss
tofumatt: (Also, for posterity, `aria-hidden` if it _did_ pose an issue would work, not sure why I didn't immediately mention/think of that at the time. But it's fine without 🎉 )

tofumatt (approved):
This looks great! Tested again to be sure (and as mentioned in a previous comment: tested the accessibility and it is good).

Let's 🚢 !

Issue #5453, opened 2022-06-27
Introduce new `getAccountChooserURL` selector


aaemnnosttv:
## Feature Description

For quite some time, Google services have supported multiple sign-in to allow for users to be signed into multiple Google accounts at the same time. This makes deep-linking to external Google service UIs challenging because a user's initial/default Google account may not be the one a user is signed in with in Site Kit which can lead to problems related to mismatched accounts, lack of access, etc.

To solve this, Google has an "AccountChooser" endpoint which essentially acts as a proxy to another service URL. It works by sending the user to the chooser URL along with the user's Google email address and the destination service URL. If the user is already signed in with the given account, it will forward the user on to the service with the proper Google account selected for the given email address. If not, the user is prompted to sign in before continuing on to the final URL.

In Site Kit, we currently use a `?authuser={email}` workaround in service URLs but this is not supported by all services.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* A new `getAccountChooserURL` selector should be added to the `core/user` store which takes a single (required) argument: the destination URL, and returns a new URL in the following format:   
	```
	https://accounts.google.com/accountchooser?continue={url-encoded-destination-url}&Email={google-email-address}
	```
	* Note the `Email` parameter name is case-sensitive
	* The `Email` value should be the URL-encoded email address of the current user's Google account
* If the email address isn't available yet, it should return `undefined`

## Implementation Brief

- In `assets/js/googlesitekit/datastore/user/user-info.js`:
    - Create a new property in the `selectors` object named `getAccountChooserURL` which uses `createRegistrySelector` to create a selector function.
	- The new function should accept a string parameter named `destinationURL`.
	- The function should start with an `invariant` function call with `destinationURL` as the first parameter and an error message as the second, e.g. `destinationURL is required.`
	- Use the `createRegistrySelector`'s `select` function to select the `core/user` store and its `getEmail()` selector to query the Google email of the current user.
	- If this email is `undefined`, return `undefined` from this selector.
	- Use `encodeURIComponent` to encode both the `destinationURL` and the queried `email`.
	- From the selector, return a template literal (string) containing a URL with the described format in the AC replacing the `{url-encoded-destination-url}` with the encoded URL and `{google-email-address}` with the encoded email.

### Test Coverage

* Add a test case in `assets/js/googlesitekit/datastore/user/user-info.test.js` that describes `getAccountChooserURL`.

## QA Brief

[...]

PR #5557, opened 2022-07-14
Add getAccountChooserURL selector.

makiost:
## Summary

Addresses issue:

- #5453 

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist

### DIFF
--- a/assets/js/googlesitekit/datastore/user/user-info.js
+++ b/assets/js/googlesitekit/datastore/user/user-info.js
@@ -398,6 +398,26 @@ export const selectors = {
 	} ),
 
 	/**
+	 * Gets the url with the users email appended.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {(string|undefined)} The concatenated url if an email is present, undefined else.
+	 */
+	getAccountChooserURL: createRegistrySelector(
+		( select ) => ( state, destinationURL ) => {
+			invariant( destinationURL, 'destinationURL is required' );
+			const userEmail = select( CORE_USER ).getEmail();
+			return userEmail !== undefined
+				? `${ encodeURIComponent(
+						destinationURL
+				  ) }&Email=${ encodeURIComponent( userEmail ) }`
+				: undefined;
+		}
+	),
+
+	/**
 	 * Gets the initial version that the user used Site Kit with.
 	 *
 	 * @since 1.27.0
--- a/assets/js/googlesitekit/datastore/user/user-info.test.js
+++ b/assets/js/googlesitekit/datastore/user/user-info.test.js
@@ -364,5 +364,53 @@ describe( 'core/user userInfo', () => {
 				expect( global[ userDataGlobal ] ).not.toEqual( undefined );
 			} );
 		} );
+
+		describe( 'getAccountChooserURL', () => {
+			it( 'throws an error if a destinationURL is not given', () => {
+				expect( () => {
+					registry.select( CORE_USER ).getAccountChooserURL();
+				} ).toThrow( 'destinationURL is required' );
+			} );
+
+			it( 'returns the encoded destination url with the email appended', async () => {
+				global[ userDataGlobal ] = userData;
+
+				const testURL = 'test destination url';
+				const userEmail = global[ userDataGlobal ].user.email;
+
+				registry.select( CORE_USER ).getAccountChooserURL( testURL );
+
+				await subscribeUntil(
+					registry,
+					() =>
+						registry
+							.select( CORE_USER )
+							.getAccountChooserURL( testURL ) !== undefined
+				);
+
+				const accountChooserURL = registry
+					.select( CORE_USER )
+					.getAccountChooserURL( testURL );
+
+				expect( accountChooserURL ).toBe(
+					`${ encodeURIComponent(
+						testURL
+					) }&Email=${ encodeURIComponent( userEmail ) }`
+				);
+			} );
+
+			it( 'should return undefined when no data is available', async () => {
+				expect( global[ userDataGlobal ] ).toEqual( undefined );
+				const testURL = 'test destination url';
+
+				const userInfo = registry
+					.select( CORE_USER )
+					.getAccountChooserURL( testURL );
+
+				const { user } = initialState;
+				expect( userInfo ).toEqual( user );
+				expect( console ).toHaveErrored();
+			} );
+		} );
 	} );
 } );


assets/js/googlesitekit/datastore/user/user-info.js
```diff
@@ -397,6 +397,26 @@ export const selectors = {
 		return user !== undefined ? user.picture : user;
 	} ),
 
+	/**
+	 * Gets the url with the users email appended.
```
tofumatt, line 401:
```suggestion
	 * Gets an account chooser url with the current user's email.
```

assets/js/googlesitekit/datastore/user/user-info.js
```diff
@@ -397,6 +397,26 @@ export const selectors = {
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param {Object} state Data store's state.
+	 * @return {(string|undefined)} The concatenated url if an email is present, undefined else.
```
tofumatt, line 406:
```suggestion
	 * @return {(string|undefined)} The concatenated url if an email is present; otherwise undefined.
```

assets/js/googlesitekit/datastore/user/user-info.js
```diff
@@ -397,6 +397,26 @@ export const selectors = {
...
+	 */
+	getAccountChooserURL: createRegistrySelector(
+		( select ) => ( state, destinationURL ) => {
+			invariant( destinationURL, 'destinationURL is required' );
+			const userEmail = select( CORE_USER ).getEmail();
+			return userEmail !== undefined
+				? `${ encodeURIComponent(
+						destinationURL
+				  ) }&Email=${ encodeURIComponent( userEmail ) }`
+				: undefined;
```
tofumatt, line 416: Let's use a bit more spacing here to give sections of code more readability. Especially things like `invariant` calls can use some whitespace after them to separate out different logic/aspects of the function.

Also: the case-sensitive `Email` param is a bit odd, so I think it's worth adding a comment in case anyone notices it and is curious about it in the future. 🙂 

Oh, and one last thing: let's try to avoid ternary operators whenever possible. I think they tend to make code a lot less readable, and in this case we can do a return early rather than ternary that makes the logic/intent clearer:

```suggestion
			invariant( destinationURL, 'destinationURL is required' );

			const userEmail = select( CORE_USER ).getEmail();
			
			if ( userEmail !== undefined ) {
				return undefined;
			}

			// The `Email` query parameter is case-sensitive; a capital "E"
			// is required for the account chooser URL.
			return ` https://accounts.google.com/accountchooser?continue=${ encodeURIComponent(
				destinationURL
			) }&Email=${ encodeURIComponent( userEmail ) }`;
```

assets/js/googlesitekit/datastore/user/user-info.test.js
```diff
@@ -364,5 +364,53 @@ describe( 'core/user userInfo', () => {
...
+
+				const accountChooserURL = registry
+					.select( CORE_USER )
+					.getAccountChooserURL( testURL );
+
+				expect( accountChooserURL ).toBe(
+					`${ encodeURIComponent(
+						testURL
+					) }&Email=${ encodeURIComponent( userEmail ) }`
+				);
```
tofumatt, line 399: This test isn't correct, because the ACs show this should include the account chooser URL and should be placing the "destination URL" in a `continue=` query param:

CleanShot 2022-07-15 at 12 49 14

The right `expect` call here would be:

```suggestion
				expect( accountChooserURL ).toBe(
					`https://accounts.google.com/accountchooser?continue=${ encodeURIComponent(
						testURL
					) }&Email=${ encodeURIComponent( userEmail ) }`
				);
```

(Note: this will fail, but that should be the value here and the tests and code should be updated to match 🙂 

Otherwise, the URL we're creating and expecting here would be: `test destination url&Email=myemail@test.com`.

assets/js/googlesitekit/datastore/user/user-info.test.js
```diff
@@ -364,5 +364,53 @@ describe( 'core/user userInfo', () => {
...
+			it( 'throws an error if a destinationURL is not given', () => {
+				expect( () => {
+					registry.select( CORE_USER ).getAccountChooserURL();
+				} ).toThrow( 'destinationURL is required' );
+			} );
+
+			it( 'returns the encoded destination url with the email appended', async () => {
+				global[ userDataGlobal ] = userData;
+
+				const testURL = 'test destination url';
```
tofumatt, line 378: Let's use a "more real" value that we'd expect here, like an actual URL instead of what looks like just a generic string.

```suggestion
				const testURL = 'https://analytics.google.com/dashboard/';
```

or something like that. I'm always a fan of using as-close-to-real data in tests so they're more readable and understandable at a glance. Plus: then we get to test the URI encoding here.

tofumatt (changes requested):
Currently this doesn't meet the ACs of the issue—maybe the IB was a bit unclear because it said to reference the ACs for the values output, but please my comments and take another look at the issue to see what's missing. We want to include the account choose URL in the return value from this selector 🙂

assets/js/googlesitekit/datastore/user/user-info.test.js
```diff
@@ -364,5 +364,53 @@ describe( 'core/user userInfo', () => {
...
+
+				const accountChooserURL = registry
+					.select( CORE_USER )
+					.getAccountChooserURL( testURL );
+
+				expect( accountChooserURL ).toBe(
+					`https://accounts.google.com/accountchooser?continue=${ encodeURIComponent(
+						testURL
+					) }&Email=${ encodeURIComponent( userEmail ) }`
+				);
```
tofumatt: I feel like this test would be a bit nicer to read if the end-result text was here as a string. Maybe we could use an inline snapshot test here instead, so we aren't just replicating the code here that's in the selector itself, eg:

```
				expect( accountChooserURL ).toMatchInlineSnapshot();
```

assets/js/googlesitekit/datastore/user/user-info.test.js
```diff
@@ -364,5 +364,53 @@ describe( 'core/user userInfo', () => {
...
+			} );
+
+			it( 'should return undefined when no data is available', async () => {
+				expect( global[ userDataGlobal ] ).toEqual( undefined );
+				const testURL = 'test destination url';
+
+				const userInfo = registry
+					.select( CORE_USER )
+					.getAccountChooserURL( testURL );
```
tofumatt, line 405: Technically this test would pass even with a defined user data global, because the first time the user selects this data the selector will return `undefined`. This test should `await` resolution of the `select( CORE_USER ).getUser()` selector, eg:

```js
await subscribeUntil( registry, () =>
	registry
		.select( CORE_USER )
		.hasFinishedResolution( 'getUser' )
);
```

tofumatt (approved):
Thanks, looks good now!

A few suggestions to improve the tests but I've made the changes to save the back/forth on another code review 🙂

Issue #5572, opened 2022-07-19
WebFont global existing causes Google Charts with custom fontName to throw exception `Cannot read properties of null (reading '__eventTarget')`


jamesozzie:
## Bug Description

We've had 3 reports of the following error after updating to the latest version of the plugin, 1.79.0
`Cannot read properties of null (reading '__eventTarget')`

This appears as a dashboard error, with more details below:

Cannot read properties of null (reading '__eventTarget')

```
    in GoogleChart
    in div
    in div
    in UserDimensionsPieChart
    in div
    in Cell
    in div
    in ForwardRef
    in div
    in ForwardRef
    in div
    in div
    in Widget
    in WithWidgetSlug(Widget)
    in DashboardAllTrafficWidget
    in WhenAnalyticsActive(DashboardAllTrafficWidget)
    in WidgetRenderer
    in div
    in Cell
    in WidgetCellWrapper
    in div
    in ForwardRef
    in div
    in div
    in ForwardRef
    in WidgetAreaRenderer
    in div
    in WidgetContextRenderer
    in DashboardMainApp
    in DashboardEntryPoint
    in RestoreSnapshots
    in ErrorHandler
    in StrictMode
    in Root
```

<details><summary>Console Errors</summary>

```
TypeError: Cannot read properties of null (reading ‘__eventTarget’)
at gvjs_Zm (jsapi_compiled_default_module.js:412:29)
at Object.gvjs_Ym [as addListener] (jsapi_compiled_default_module.js:411:392)
at googlesitekit-modules-search-console-227f86b858fbc87c820e.js:31:11241
at Ui (googlesitekit-vendor-4167c42429c777851281.js:7:724946)
at va (googlesitekit-vendor-4167c42429c777851281.js:7:747869)
at t.unstable_runWithPriority (googlesitekit-vendor-4167c42429c777851281.js:7:767492)
at Mr (googlesitekit-vendor-4167c42429c777851281.js:7:692598)
at Yj (googlesitekit-vendor-4167c42429c777851281.js:7:744298)
at Lj (googlesitekit-vendor-4167c42429c777851281.js:7:734658)
at googlesitekit-vendor-4167c42429c777851281.js:7:692889
```

</details>

## Steps to reproduce

The below are the only known steps to consistently recreate so far:
1. Set up a new WP site, with SK set up and Google Analytics also connected
2. Install and activate [the Futurio Extra plugin](https://wordpress.org/plugins/futurio-extra/)
[...]

PR #5580, opened 2022-07-20
Only try to load custom fonts in Google Charts when Google WebFont is not loaded

tofumatt:
## Summary

Addresses issue:

- #5572

## Relevant technical choices

This is essentially a "don't run the code that causes the bug" rather than "fix the root cause" fix/workaround, because the underlying error is quite unclear and the bug here is somewhat niche.

This fixes an error that only happens when another file creates/overwrites the `window.WebFont` global. There’s no straightforward fix as the errors happening are very opaque, and the visual difference in this bugfix approach is quite subtle AND only affects users who even have the Google WebFont loader script in their theme (not a default).

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist

### DIFF
--- a/assets/js/components/GoogleChart.js
+++ b/assets/js/components/GoogleChart.js
@@ -225,24 +225,44 @@ export default function GoogleChart( props ) {
 		}
 	}
 
+	// Note the `typeof WebFont === 'undefined'` check for every `fontName`
+	// property. If the Google `WebFont` global exists and we try to use
+	// the `fontName` property in a Google Chart, we'll encounter an
+	// exception and the entire dashboard will crash.
+	//
+	// Our "workaround" is not to load custom fonts when the `WebFont`
+	// global exists, as these are usually very small UI elements
+	// and there's no clear way to get them to load without errors
+	// when the `WebFont` global/script is loaded.
+	//
+	// See: https://github.com/google/site-kit-wp/issues/5572
 	merge( chartOptions, {
 		hAxis: {
 			textStyle: {
-				fontName: 'Google Sans Text',
+				fontName:
+					typeof WebFont === 'undefined'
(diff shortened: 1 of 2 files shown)


assets/js/components/GoogleChart.js
```diff
@@ -225,24 +225,44 @@ export default function GoogleChart( props ) {
...
+	//
+	// See: https://github.com/google/site-kit-wp/issues/5572
 	merge( chartOptions, {
 		hAxis: {
 			textStyle: {
-				fontName: 'Google Sans Text',
+				fontName:
+					typeof WebFont === 'undefined'
+						? '"Google Sans Text", "Helvetica Neue", Helvetica, Arial, sans-serif'
+						: undefined,
```
aaemnnosttv, line 245: How about we extract this to a helper function? This would be  a bit cleaner in each instance and then we also don't need to duplicate the reasoning behind it in each use. Most people will copy whatever we already use somewhere else anyways so it should be safe for future use as well.

aaemnnosttv (commented):
Thanks @tofumatt ! This seems to fix the issue (although I see a different error locally `A cross-origin error was thrown`) is this expected?

Left one suggestion to improve this but it seems to do the trick! 👍

tofumatt:
Hmm, I didn't notice an error in testing but I'll check. Thanks 👍🏻

aaemnnosttv:
Sorry, I meant when reproducing the original issue I got the different error than what was originally reported. I didn't get any errors when trying the code in the PR.

aaemnnosttv (approved):
Great, thanks @tofumatt !

PR #5600, opened 2022-07-22
5550 Bug / Stack ModuleRecoveryAlert Notifications

jimmymadon:
## Summary

Addresses issue:

- #5550 

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist


assets/js/components/dashboard-sharing/ModuleRecoveryAlert.js
```diff
@@ -79,15 +81,18 @@ export default function ModuleRecoveryAlert() {
...
-		recoverModules( modulesToRecover ).finally( () => {
-			setRecoveringModules( false );
-			setCheckboxes( null );
-		} );
+		await recoverModules( modulesToRecover );
+
+		setRecoveringModules( false );
+		setCheckboxes( null );
+
+		return { dismissOnCTAClick: false };
```
aaemnnosttv, line 95: If we want to make this a bit more explicit, I think `dismissCTA` would be ok here, but including `OnClick` is unnecessary detail. There isn't anything particularly special about clicks here except that is the prop this function will be passed to. Let's avoid coupling with clicks where it isn't needed.

assets/js/components/dashboard-sharing/ModuleRecoveryAlert.js
```diff
@@ -218,9 +213,18 @@ export default function ModuleRecoveryAlert() {
...
+			onCTAClick={ onCTAClick }
+			ctaLink={ ctaLink }
 			description={ description }
 			learnMoreURL="https://sitekit.withgoogle.com/documentation/using-site-kit/dashboard-sharing/"
 			learnMoreLabel={ __( 'Learn more', 'google-site-kit' ) }
+			dismiss={
+				isLoading
+					? undefined
+					: __( 'Remind me later', 'google-site-kit' )
+			}
```
aaemnnosttv, line 226: Is this condition necessary here? If `isLoading` is true then the `BannerNotification` wouldn't be shown at all but would show the `ProgressBar` instead, correct?

assets/sass/components/global/_googlesitekit-publisher-wins.scss
```diff
@@ -165,6 +165,10 @@
 		margin-left: $grid-gap-desktop;
 	}
 
+	.mdc-button + .spinner + .googlesitekit-cta-link {
+		margin-left: $grid-gap-desktop;
+	}
+
```
aaemnnosttv, line 171: We can simplify this by combining the block here with the one above it and removing the `mdc-button` from the beginning since the `spinner` is rendered unconditionally (the CTA button would never be directly adjacent to the cta link anymore). So really I think the only selector we need for this is `.spinner + .googlesitekit-cta-link`.

assets/js/components/notifications/BannerNotification/index.js
```diff
@@ -153,11 +157,14 @@ function BannerNotification( {
 	async function handleCTAClick( e ) {
 		e.persist();
 
+		let dismissOnCTAClick = true;
 		if ( onCTAClick ) {
-			await onCTAClick( e );
+			setIsAwaitingCTAResponse( true );
+			( { dismissOnCTAClick } = ( await onCTAClick( e ) ) || {} );
```
aaemnnosttv, line 163: This will break dismissible CTAs that don't return this new property because the declared `dismissOnCTAClick` default above will be overridden with `undefined` here.

We need to be sure the default `true` is preserved even when it's not returned as proposed in the IB.

assets/js/components/dashboard-sharing/ModuleRecoveryAlert.js
```diff
@@ -195,20 +199,11 @@ export default function ModuleRecoveryAlert() {
...
-					}
-				>
-					{ __( 'Recover', 'google-site-kit' ) }
-				</Button>
-				<Spinner isSaving={ recoveringModules } />
 			</Fragment>
 		);
+		ctaLink = '#';
+		ctaLabel = __( 'Recover', 'google-site-kit' );
+		onCTAClick = handleRecoverModules;
```
aaemnnosttv, line 206: These seem to be duplicated unnecessarily. I think we can set the `ctaLink` and `ctaLabel` with these values below so long as there is an `onCTAClick` defined.

aaemnnosttv (changes requested):
Thanks @jimmymadon, this looks great and works well in my testing. Just a few things to address and then this should be good to go 👍

assets/js/components/dashboard-sharing/ModuleRecoveryAlert.js
jimmymadon: Well the new flag is defined and used solely in the `onCTAClick` handler - so perhaps the name of this flag is fine? I was being specific mainly because we already have `dismiss` as a prop in `BannerNotification` for the actual dismiss link/CTA. `dismissCTA` could also sound confusing as yet another dismiss CTA link / button instead of just a flag that would prevent the default dismissal of a BannerNotification. Any other name suggestions perhaps? e.g `preventCTADismissal`?

assets/js/components/dashboard-sharing/ModuleRecoveryAlert.js
jimmymadon: The `<ProgressBar>` is passed into the `children` prop of `BannerNotification` instead of the whole `BannerNotification` itself. This is to allow the `title` of the notification to be rendered and only show the rest of the component as loading. So the `dismiss` link still shows if passed as its own prop below the progress bar.

assets/js/components/notifications/BannerNotification/index.js
jimmymadon: Yikes - was trying something and removed the default.

assets/sass/components/global/_googlesitekit-publisher-wins.scss
jimmymadon: Good catch - I thought the component was conditional based on `isSaving` but its hidden only via CSS.

assets/js/components/dashboard-sharing/ModuleRecoveryAlert.js
aaemnnosttv: Ah in that case, I think it would be more appropriate to pass  `isDismissible` as `false` which we could set in the same condition where `children` is set – is there a reason we can't do that instead?

assets/js/components/dashboard-sharing/ModuleRecoveryAlert.js
aaemnnosttv: It's probably fine as is; it sticks out to me as a leaked detail though because the function here is `handleRecoverModules` rather than `handleOnCTAClick` where `handleRecoverModules` could be bound to other events than a click. In that context, I would opt for something a bit more generic for the response. It sounds like you're thinking of the name as a standalone concept in the component where it could be confused with other props. I'm actually less concerned with the name where it's used, and thinking about it more as an API of the function. So the name you have now would be fine (i.e. as an alias when destructuring), or left on the object it could be a bit more self documenting `onCTAResult.dismissCTA` that kind of thing.

Again, it doesn't matter that much right now because there is only the one use right now and everything else that wasn't using it is unaffected by this change, just sharing my perspective 👍

assets/js/components/dashboard-sharing/ModuleRecoveryAlert.js
aaemnnosttv: Is this something we can simplify?

aaemnnosttv (commented):
Thanks @jimmymadon – this looks good, just a few small questions left.

aaemnnosttv:
Also, I don't recall if this was there before but I'm seeing now there is a merge conflict 😄

jimmymadon:
@aaemnnosttv Have addressed your feedback - thanks for the thorough CR! Have also merged the changes coming from #5318 and tested it all out - looks good. Have updated the QAB to re-test that too.

aaemnnosttv (approved):
Great, thanks @jimmymadon !

Issue #5529, opened 2022-07-08
Implement TwG setup UI for customization


felixarntz:
Follow-up to #5455: The actual UI for the `SetupCustomize` component should be implemented based on the Figma design.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The `SetupCustomize` component should have its proper UI implemented, from [this Figma design](https://www.figma.com/file/t5zDpQ28Lo7ZyRcOpGdqgu/TwG?node-id=1205%3A4865), using the `ImageRadio` component from #5459 for its controls for "Type", "Prominence", and "Color".
* There is one part of the UI that needs to display conditionally: If the "Type" is set to "Overlay", the "Prominence" UI below should be displayed. Otherwise ("Type" set to "Fixed"), the other UI about auto vs manual placement from [this Figma design](https://www.figma.com/file/t5zDpQ28Lo7ZyRcOpGdqgu/TwG?node-id=1205%3A5852) should be displayed below.
* Changing any of the controls should update the corresponding TwG settings on the client-side. Only clicking the primary submit button should send the API request to actually set the settings on the server-side. The controls map to the different TwG settings as follows:
    * "Type" and the control below (whichever of the alternate two) affect the `buttonPlacement` setting:
        * "Type" set to "Overlay" and "Prominence" set to "Low" --> `dynamic_low`
        * "Type" set to "Overlay" and "Prominence" set to "High" --> `dynamic_high`
        * "Type" set to "Fixed" and below placement set to "Auto" --> `static_auto`
        * "Type" set to "Fixed" and below placement set to "Manual > Above the post" --> `static_above-content`
        * "Type" set to "Fixed" and below placement set to "Manual > Below the post" --> `static_below-content`
        * "Type" set to "Fixed" and below placement set to "Manual > Below the 1st paragraph" --> `static_below-first-paragraph`
    * "Color" maps to the `colorTheme` setting and should set the selected color's identifier (not the human-readable label).
    * "Post types" maps to the `buttonPostTypes` setting and should set the array of post type identifiers/slugs (not the human-readable labels).
* Since the TwG `SettingsEdit` UI will basically display the same content (see #5531, also [here in Figma]()), reusable "common" components for the TwG settings controls should be implemented as much as possible so that they can be reused for the UI in #5531.

## Implementation Brief

* In `assets/js/modules/thank-with-google/components/common`, create the following functional React components:
    - `TypeRadio`,
[...]

PR #5606, opened 2022-07-25
Enhancement/5529 setup UI customization

eugene-manuilov:
## Summary

Addresses issue:

- #5529

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist


assets/js/modules/thank-with-google/components/common/PostTypesSelect.js
```diff
@@ -0,0 +1,100 @@
...
+			chipCheckmark={ <ChipCheckmark /> }
+			data-chip-id={ slug }
+			id={ slug }
+			key={ slug }
+			label={ label }
+			onClick={ toggleChip }
+			onKeyUp={ toggleChip }
+			selected={ buttonPostTypes?.includes( slug ) }
+			className="googlesitekit-twg-post-type-select__chip"
+		/>
```
tofumatt, line 83: I guess this is how the component is supposed to behave, but I find the transition between checked and unchecked a bit jarring, eg:

CleanShot 2022-07-27 at 20 19 05

Maybe worth a later discussion as it seems that's how the actual component behaves and it's not something special we're doing. 🤷🏻‍♂️

assets/js/modules/thank-with-google/components/common/ColorRadio.js
```diff
@@ -0,0 +1,76 @@
...
+			const { value: color = 'blue' } = target || {};
+			setColorTheme( color );
+		},
+		[ setColorTheme ]
+	);
+
+	const colors = getColorThemes()?.map(
+		( { colorThemeID, name, svg: SVG } ) => (
+			<ImageRadio
+				key={ `${ currentColor }::${ colorThemeID }` }
```
tofumatt, line 50: `currentColor` being part of the key causes re-rendering of these components (and accessibility issues with focus) because the key's value changes between every React render.

Because the `key` here changes every time `currentColor` changes (and `currentColor` changes whenever a new colour is selected with the arrow keys):

CleanShot 2022-07-27 at 20 20 54

Fortunately, the fix is straightforward:

```suggestion
				key={ `${ name }::${ colorThemeID }` }
```

or even just
```suggestion
				key={ colorThemeID }
```

works. When tested locally this then behaves properly with fewer DOM re-renders and no keyboard issues:

CleanShot 2022-07-27 at 20 25 46

assets/js/modules/thank-with-google/components/common/PositionRadio.js
```diff
@@ -0,0 +1,128 @@
...
+			</div>
+			<div className="googlesitekit-twg-position-radio__option">
+				<Radio
+					id="button-placement-manual"
+					name="button-placement"
+					value=""
+					readonly
+				>
+					{ __( 'Manual', 'google-site-kit' ) }
+				</Radio>
```
tofumatt, line 80: I just checked and can see this is indeed what was [laid out in the Figma design](https://www.figma.com/file/t5zDpQ28Lo7ZyRcOpGdqgu/TwG?node-id=1205%3A7472), but we should change it 🙂 

Unfortunately this radio button doesn't _look_ or _behave_ like a read-only button for both sighted keyboard users and sighted mouse users—it looks and behaves just like the "Auto" button but never changes. At best this is buggy and awkward—I haven't tried using it with a screenreader—in that case the `readonly` attribute might prevent SOME confusion, but given it can _never_ be interactive and is functioning more like a heading/separator than a button, we shouldn't use a `Radio` button component here at all. Otherwise it behaves like this which is quite confusing:

CleanShot 2022-07-27 at 20 31 45

So let's deviate from the design and make this not a radio button so it doesn't appear in the keyboard order, doesn't _look_ like it can be clicked, etc.

tofumatt (changes requested):
Overall this looks and works great to me 👍🏻 

A few small tweaks/comments (and some SCSS files that could use their copyright year updated since they're new files and we use the current year for new files).

Two main issues:

1. The re-rendering/focus-loss on the colour picker. I've suggested a fix that worked locally when I tested.
2. The "Manual" radio button should instead be a heading (and sub-heading) that is _styled_ like the radio buttons there but is not a button. I left more detailed comments about it inline, but currently it's odd UX and presents a not-great-accessibility experience too. I know it's what's in the Figma mock, but we should change it as an improvement 🙂 

After that I think this is good-to-go, looks and works great 👍🏻

assets/js/modules/thank-with-google/datastore/constants.js
```diff
@@ -17,3 +17,15 @@
...
+export const STATE_ACTIVE = 'ACTIVE';
+export const STATE_ACTION_REQUIRED = 'ACTION_REQUIRED';
+export const STATE_PENDING_VERIFICATION = 'PENDING_VERIFICATION';
+
+export const BUTTON_PLACEMENT_DYNAMIC_LOW = 'dynamic_low';
+export const BUTTON_PLACEMENT_DYNAMIC_HIGH = 'dynamic_high';
+export const BUTTON_PLACEMENT_STATIC_AUTO = 'static_auto';
+export const BUTTON_PLACEMENT_STATIC_ABOVE_CONTENT = 'static_above-content';
+export const BUTTON_PLACEMENT_STATIC_BELOW_CONTENT = 'static_below-content';
+export const BUTTON_PLACEMENT_STATIC_BELOW_1ST_P =
```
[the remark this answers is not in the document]
eugene-manuilov: Sure, we can 👍. Updated.

assets/js/modules/thank-with-google/components/common/PostTypesSelect.js
eugene-manuilov: Yeah, probably worth creating a new ticket if we want to discuss if further.

assets/js/modules/thank-with-google/components/common/ColorRadio.js
eugene-manuilov: Yes, you are right. Updated.

assets/js/modules/thank-with-google/components/common/PositionRadio.js
eugene-manuilov: I have updated it to be disabled which makes it inactive. Does it look good to you now?

assets/js/components/ImageRadio.js
```diff
@@ -32,20 +32,22 @@ import { useCallback } from '@wordpress/element';
...
+		value,
+		checked = false,
+		tabIndex = undefined,
+		onKeyDown = null,
+		image,
+		label = '',
+		children,
+		description,
+		ariaLabel,
+	} = props;
```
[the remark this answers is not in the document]
eugene-manuilov: IMO keeping the long list of destructured properties in the function declaration makes function declaration cumbersome and hard to read. Yes, we should probably add an eslint rule for it, but this is out of scope for this ticket.

assets/js/modules/thank-with-google/components/common/PositionRadio.js
tofumatt: It's _better_ that way, especially for keyboard users, but visually it's still "off" to me, because there's a cursor change when hovering over the radio button (implying it can be clicked) and it visually looks like it's a button.

If a button is disabled 100% of the time, it shouldn't be a button.

But to prevent this issue from blocking other issues I've filed a follow-up issue to fix this: #5623. I think it's an issue with the source design and the `disabled` hack is certainly good enough for now. 👍🏻

tofumatt (approved):
Awesome, looks and works great.

I've made #5623 as a follow-up issue, but no reason it needs to block this and the follow-up issues 👍🏻

Issue #5548, opened 2022-07-12
Refactor `getServiceURL` selectors to use `getAccountChooserURL`


aaemnnosttv:
## Feature Description

In #5453 we introduced a new selector for using the proper service proxy for resolving the appropriate URL for a service and a given user email address.

Until now, this has been implemented manually in our `getServiceURL` selectors by injecting the user's email into different places in the URL based on the service. This issue is about delegating that part to the AccountChooser service.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* `getServiceURL` selectors for all modules should no longer pass an `authuser` parameter or `/u/{email}` path segment
	* Instead, these selectors should construct the respective user-agnostic URL for the target and return their result using the new `getAccountChooserURL` selector (see #5453)
	* One exception here is PageSpeed Insights which does not have a user-specific service URL

## Implementation Brief

- Locate the `getServiceURL` selectors for all services (except `pagespeed-insights`).
- They can be found in:
    - `assets/js/modules/adsense/datastore/service.js`
    - `assets/js/modules/analytics/datastore/service.js`
    - `assets/js/modules/optimize/datastore/service.js`
    - `assets/js/modules/search-console/datastore/service.js`
    - `assets/js/modules/tagmanager/datastore/service.js`
- Refactor the selectors:
    - The query for user email should be removed.
    - If the `path` parameter is defined, sanitize it and append it to the service `baseURI` with a `#` (this should be already available).
    - If the `query` parameter is defined, use `addQueryArgs` to add them the `baseURI` (this should be already available).
    - Completely remove the addition of `authuser` property.
    - Use the `createRegistrySelector`'s `select` function to select the `core/user` store and its `getAccountChooserURL()` selector with the updated `baseURI` passed to query the account chooser URL.
    - If the account chooser URL is `undefied`, return `undefined` from the selector.
    - Otherwise, return the queried account chooser URL.

### Test Coverage

* Update the test cases for the updated `getServiceURL` selectors. They should be expected to match the return value of `getAccountChooserURL`.

## QA Brief

This issue changes the way "deep links" to external service dashboards are constructed to leverage the "AccountChooser" proxy service, which essentially ensures that you are sent to the destination service URL with the right Google account "selected".

This issue affects all modules with service URLs (except PageSpeed): AdSense, Analytics, Optimize, Search Console, & Tag Manager.

[...]

PR #5631, opened 2022-07-31
Refactor to getAccountChooser links, update tests.

makiost:
## Summary

Addresses issue:

- #5548 

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist


assets/js/modules/analytics/datastore/service.js
```diff
@@ -50,22 +50,28 @@ export const selectors = {
...
+
+			const accountChooserBaseURI = select(
+				CORE_USER
+			).getAccountChooserURL( appendedURL );
+
+			if ( accountChooserBaseURI ) {
+				return accountChooserBaseURI;
+			}
+
+			return undefined;
```
tofumatt, line 74: This isn't really a proper "return early" type of approach.

In this case, our "ideal" or "happy path" is not the final return value, which makes this a bit harder to parse. In most of our code we will return early when we don't have the proper amount of data to return the value a selector is designed to return. This makes it easier to follow the logic, but also allows us to comment out specific reasons why we "bail early" in a function and not add many conditions to the return of the "proper" value.

So this would be better written as:

```suggestion
			if ( accountChooserBaseURI === undefined ) {
				return undefined
			}

			return accountChooserBaseURI;
```

assets/js/modules/analytics/datastore/service.js
```diff
@@ -50,22 +50,28 @@ export const selectors = {
...
-			const userEmail = select( CORE_USER ).getEmail();
+			const baseURI = 'https://analytics.google.com/analytics/web/';
 
-			if ( userEmail === undefined ) {
-				return undefined;
+			let appendedURL = baseURI;
+
+			if ( query ) {
+				appendedURL = addQueryArgs( baseURI, query );
 			}
```
tofumatt, line 59: Given we only ever use `baseURI` to define `appendedURL` and then later to redefine that variable entirely, I think it isn't really needed as a variable here. If we changed `appendedURL` to something more descriptive like `serviceURL` we could get rid of `baseURI` and the rest of the code might be a bit more descriptive, as we'd augment `serviceURL` with any query params, if present. Eg.:

```js
let serviceURL = 'https://analytics.google.com/analytics/web/';

if ( query ) {
	serviceURL = addQueryArgs(serviceURL, query);
}

if ( path ) {
	const sanitizedPath = `/${ path.replace( /^\//, '' ) }`;
	serviceURL = `${serviceURL}#${sanitizedPath}`;
}

const accountChooserServiceURL = select(CORE_USER).getAccountChooserURL(serviceURL);
```

(Also, we're mixing and matching URI and URL… we usually just use "URL" in our code… I know they're technically different, but it's probably easier to just stick with `URL`, especially as it's part of the selector's name 😅. I can see there's existing inconsistency in our codebase though, so it's not a big deal, especially re: `baseURI`. 🤷🏻‍♂️  )

tofumatt (changes requested):
The note I made in the analytics code here applies to all of the modules re: simplifying the `baseURI` + `appendedURL` variables into a single `let serviceURL` variable we redefine.

Also we shouldn't be returning early "on success", so all of the selectors should not `return undefined` as their final statement.

------

The `mockAccountChooserURL` and `decodeServiceURL` helper functions in the tests are fine, but their logic is essentially duplicated across every module's test code with only the `baseURI` changing. It'd be better to move that into a test helper file that either took a `baseURI` as an argument, or even a "function factory" that accepted a baseURL and returned a helper for each service to use, eg:

```js
const mockAccountChooserURL = createAccountChooserMock( 'https://optimize.google.com/optimize/home/' );

// [...]

const expectedURL = mockAccountChooserURL( '/test/path/to/deelink' );

expect( serviceURLWithLeadingSlash ).toEqual( expectedURL );
```

The same probably goes for `decodeServiceURL`.

That said, it might not be awful to use `toMatchInlineSnapshot` for those values instead of the helper functions. It's much less code/complexity, and it allows the tests to show the developer _exactly_ what the string output should look like. Huge snapshots are lazy test-writing, but this is actually a case where the `toMatchInlineSnapshot` function is really useful. It'll make what the end result to be clear in one line in the test, and reduce the helper functions we need in all of the tests.

tofumatt (approved):

tofumatt (approved):

tofumatt (changes requested):
Looks like this is failing E2E tests—probably because some URLs that were previously not using `getAccountChooser` links are now using them. Could you please update the E2E tests that are failing to use the correct URLs?

I think that's what is going on since they've been failing for the history of this PR's commits and they aren't failing elsewhere.

Feel free to ping me or another engineer on Slack if you have any questions 😄

tofumatt (approved):
Awesome, looks good 👍🏻

Issue #5650, opened 2022-08-03
Install the SwG client API library


aaemnnosttv:
## Feature Description

Thank with Google will rely on the `subscribewithgoogle` API which will not be have its client library available via the main https://github.com/googleapis/google-api-php-client-services package until after launch. As a temporary workaround, we can generate and install the same client library bindings for PHP which we can later remove once available in the official client services. 

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The PHP client library files for the `subscribewithgoogle` v1 API should be generated and added as a new Composer package under `/packages/apiclient-services-subscribewithgoogle`
	* A single `SubscribewithGoogle.php` will be generated which can live in the root
	* The package name should be `google/apiclient-services-subscribewithgoogle`
* The plugin should be updated to require the new package as non-dev Composer dependency
* The added client library files should be prefixed just like any other client service, with the production code using the generated prefixed versions under `/third-party/google/`

## Implementation Brief

* Generate client library file `SubscribewithGoogle.php` under the path in the AC and create the necessary `composer.json` in the same directory
* Add a new definition under the main `composer.json` for the plugin to define a new `path` type Composer repository for the new package
* Update PHP Scoper configuration to include the added package

### Test Coverage

* Add a basic test ensuring the new prefixed dependencies are present

## QA Brief

* N/A – automated tests should be sufficient for this

## Changelog entry

* Add Thank with Google PHP client library.

tofumatt:
Evan and I discussed this a bit offline as well; he'll handle the build so assigning it to him.

IB ✅

tofumatt:
Moving directly to approval since no QA is required.

PR #5651, opened 2022-08-03
Add subscribewithgoogle API client library

aaemnnosttv:
## Summary

Addresses issue:

- #5650

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,7 @@
     "ext-openssl": "*",
     "google/apiclient": "^2.12.2",
     "google/apiclient-services": "0.243.*",
+    "google/apiclient-services-subscribewithgoogle": "^0.1.0",
     "guzzlehttp/guzzle": "^5.3.3",
     "true/punycode": "^2.0 <2.1.1"
   },
@@ -56,6 +57,13 @@
     {
       "type": "path",
       "url": "packages/wp-phpunit-polyfill"
+    },
+    {
+      "type": "path",
+      "url": "packages/apiclient-services-subscribewithgoogle",
+      "options": {
+        "symlink": false
+      }
     }
   ],
   "scripts": {
--- a/packages/apiclient-services-subscribewithgoogle/SubscribewithGoogle.php
+++ b/packages/apiclient-services-subscribewithgoogle/SubscribewithGoogle.php
@@ -1,0 +1,461 @@
+<?php
+/*
+ * Copyright 2014 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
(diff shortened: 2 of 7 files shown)


composer.json
```diff
@@ -56,6 +57,13 @@
     {
       "type": "path",
       "url": "packages/wp-phpunit-polyfill"
+    },
+    {
+      "type": "path",
+      "url": "packages/apiclient-services-subscribewithgoogle",
+      "options": {
+        "symlink": false
```
aaemnnosttv, line 65: This option causes the package to be mirrored/copied into the `vendor` directory instead of using a symlink which is the default. This was necessary to avoid strange breaking behavior with our dependency prefixing.

packages/apiclient-services-subscribewithgoogle/SubscribewithGoogle.php
```diff
@@ -0,0 +1,461 @@
+<?php
```
aaemnnosttv, line 1: Note the contents of this file are entirely generated and should be treated as an external dependency. It's inclusion here is temporary until it is available as part of the `apiclient-services` package.

scoper.inc.php
```diff
@@ -110,6 +116,10 @@ function( $file_path, $prefix, $contents ) {
 				$contents = str_replace( "'Google_", "'" . $prefix . '\Google_', $contents );
 				$contents = str_replace( '"Google_', '"' . $prefix . '\Google_', $contents );
 			}
+			if ( false !== strpos( $file_path, 'apiclient-services-subscribewithgoogle' ) ) {
+				// Rewrite "Class_Name" to Class_Name::class to inherit namespace.
+				$contents = preg_replace( '/"(Google_[^"]+)"/', '\\1::class', $contents );
```
aaemnnosttv, line 121: Without this, a fatal error will be raised in the result of a successful request because the Google collection will try to map the response into the respective models but using the wrong (unprefixed) class name, which won't exist. This was originally missed because the API wasn't working yet 😄

tofumatt (approved):

packages/apiclient-services-subscribewithgoogle/SubscribewithGoogle.php
```diff
@@ -0,0 +1,461 @@
...
+ * <a href="https://developers.google.com/news/subscribe/guides/overview" target="_blank">Documentation</a>
+ * </p>
+ *
+ * @author Google, Inc.
+ */
+class Google_Service_SubscribewithGoogle extends Google_Service
+{
+  /** See your primary Google Account email address. */
+  const USERINFO_EMAIL =
+      "https://www.googleapis.com/auth/userinfo.email";
```
felixarntz, line 38: @aaemnnosttv This isn't used anywhere, is it? Just asking due to the lack of the actual SwG publications-related scope, which apparently is necessary.

If having this here actually matters, we may want to hard-code it? But I don't see currently why we would need to.

felixarntz (commented):
Leaving an LGTM here after the fact - just left one minor question :)

packages/apiclient-services-subscribewithgoogle/SubscribewithGoogle.php
aaemnnosttv: No, it's a mistake somewhere I believe as _using_ the publications API does indeed require the SwG scope. I raised this with the API team and they're following up on that internally, but it shouldn't affect us as we don't rely on the discovery document to determine the scopes we need.

PR #5700, opened 2022-08-17
Add dropdown & toggle to GA4 Activation Banner

nfmohit:
## Summary

Addresses issue:

- #5276 

## Relevant technical choices

This PR adds the Property Select dropdown and tag placement toggle to the GA4 Activation Banner.

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist

### DIFF
--- a/assets/js/modules/analytics-4/components/dashboard/ActivationBanner/SetupBanner.js
+++ b/assets/js/modules/analytics-4/components/dashboard/ActivationBanner/SetupBanner.js
@@ -45,7 +45,7 @@ export default function SetupBanner( { onCTAClick } ) {
 		select( MODULES_ANALYTICS_4 ).getExistingTag()
 	);
 
-  useExistingTagEffect();
+	useExistingTagEffect();
 
 	const referenceDateString = useSelect( ( select ) =>
 		select( CORE_USER ).getReferenceDate()
@@ -136,7 +136,7 @@ export default function SetupBanner( { onCTAClick } ) {
 			onCTAClick={ onCTAClick }
 			footer={ <p>{ footer }</p> }
 			dismiss={ __( 'Cancel', 'google-site-kit' ) }
-      dismissExpires={ getBannerDismissalExpiryTime(
+			dismissExpires={ getBannerDismissalExpiryTime(
 				referenceDateString
 			) }
 		>
--- a/assets/js/modules/analytics-4/components/dashboard/ActivationBanner/SetupBanner.js
+++ b/assets/js/modules/analytics-4/components/dashboard/ActivationBanner/SetupBanner.js
@@ -32,15 +32,10 @@ import {
 	UseSnippetSwitch,
 } from '../../../../analytics-4/components/common';
 import useExistingTagEffect from '../../../../analytics-4/hooks/useExistingTagEffect';
-import { MODULES_ANALYTICS } from '../../../../analytics/datastore/constants';
 import { MODULES_ANALYTICS_4 } from '../../../datastore/constants';
 const { useSelect } = Data;
 
 export default function SetupBanner( { onCTAClick } ) {
-	// We need this useSelect hook to trigger starting getAccounts resolution which is needed to properly
-	// display the Property Select.
-	useSelect( ( select ) => select( MODULES_ANALYTICS ).getAccounts() );
-
 	const ga4MeasurementID = useSelect( ( select ) =>
 		select( MODULES_ANALYTICS_4 ).getMeasurementID()
 	);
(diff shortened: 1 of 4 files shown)


assets/js/modules/analytics-4/components/dashboard/ActivationBanner/SetupBanner.js
```diff
@@ -19,28 +19,41 @@
...
+} from '../../../../analytics-4/components/common';
+import useExistingTagEffect from '../../../../analytics-4/hooks/useExistingTagEffect';
+import { MODULES_ANALYTICS } from '../../../../analytics/datastore/constants';
 import { MODULES_ANALYTICS_4 } from '../../../datastore/constants';
 const { useSelect } = Data;
 
 export default function SetupBanner( { onCTAClick } ) {
+	// We need this useSelect hook to trigger starting getAccounts resolution which is needed to properly
+	// display the Property Select.
+	useSelect( ( select ) => select( MODULES_ANALYTICS ).getAccounts() );
```
tofumatt, line 42: Hmm, this is… kinda odd. I guess it makes sense because our selectors have side-effects and it's what we're relying on here… though it's a bit odd the data or `hasFinishedResolution` check isn't being used. If this is needed to render something in this component's output, why aren't we checking to make sure it's defined?

Or if another component needs this data, shouldn't _it_ be the one making sure this data is loaded?

tofumatt (changes requested):
Looks like the version with a toggle has slightly different margins:

CleanShot 2022-08-21 at 18 14 06

This should be tweaked if possible so they're all consistent.

-----

Looks good overall, but I have a question about the `useSelect` side-effect and the styles for the version with a "place snippet" toggle.

------

I couldn't get the toggle to appear from the QA steps but maybe I'm missing something. Or maybe it's not working for me as-expected 🤔

nfmohit:
Thank you for the kind review @tofumatt!

> Looks like the version with a toggle has slightly different margins

I had changed this to match the Figma designs, but you're right. I've added the margin back to keep it consistent.

> I couldn't get the toggle to appear from the QA steps but maybe I'm missing something. Or maybe it's not working for me as-expected 🤔

I did the following for Site Kit to detect an existing tag. I added the following to my theme's `header.php` file:
* Inside the `<head>` tag:
```
<script async src="https://www.googletagmanager.com/gtag/js?id={ Your GA4 Property ID }"></script>
```
* After the `<body>` tag starts:
```
<script>
	window.dataLayer = window.dataLayer || [];

	function gtag() {
		dataLayer.push(arguments);
	}
	gtag('js', new Date());

	gtag('config', '{ Your GA4 Property ID }');
</script>
```

After that, the toggle appeared for me:
image

assets/js/modules/analytics-4/components/dashboard/ActivationBanner/SetupBanner.js
nfmohit: Excellent spot, @tofumatt! These were actually my questions as well when I couldn't make the `<PropertySelect />` show up without calling this selector.

I did quite some backtracking and [it looked like](https://github.com/google/site-kit-wp/commit/69408425b855f407d88ef1b23c85056ab2af8da3#diff-d7f93ad4ddb90e5e27d32343c264d4d6a6f4e50d7fea61c4bef0a9df85d81377R58-R71) this check for accounts in `PropertySelect` existed from the initiation of the component, but couldn't figure out a reason why it was actually needed. I even tried removing the condition and couldn't notice any errors/side effects.

But I didn't want to risk it and noticed a few other places where a similar implementation was done ([example](https://github.com/google/site-kit-wp/blob/a1c48f185466353a5296c8ef4a818125b0321785/assets/js/modules/search-console/components/settings/SettingsEdit.js#L31-L35)), which is why I just went with calling the selector like this.

Do you possibly have an idea why do we actually need this `getAccounts` selector call in `PropertySelect`?

Thank you!

tofumatt: The reason for "needing" it seems to be this `isLoading` check:

https://github.com/google/site-kit-wp/blob/286fc782e27170b3af93f18b06ee8611a47c6094/assets/js/modules/analytics-4/components/common/PropertySelect.js#L69-L71

But it checks for the resolution without actually calling it. We _could_ do what you added and saw elsewhere in the repo, but that introduces selector dependencies for rendering a component which is actually really weird to me.

Let's instead just add the selector right into `PropertySelect`. It's harmless to do from my point-of-view and if the component considers it a required field, we should use it 🙂

tofumatt (approved):
Thanks for the QA brief update, that totally worked. 👍🏻 

In terms of the `getAccounts` selector being needed, I left a comment about it and committed a different solution that means the component doesn't have "selector dependencies", which I found a bit weird.

I didn't observe any issues with this approach, but I've added notes to the QA Brief to ensure the change is tested.

nfmohit:
Brilliant, thank you @tofumatt!

PR #5727, opened 2022-08-23
Update validation for analytics shared requests

nfmohit:
## Summary

Addresses issue:

- #5711 

## Relevant technical choices

This is a follow-up PR addressing [this code review](https://github.com/google/site-kit-wp/pull/5721#pullrequestreview-1082615447).

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist


includes/Core/Modules/Module.php
```diff
@@ -338,6 +338,9 @@ final protected function execute_data_request( Data_Request $data ) {
 			$datapoint    = $this->get_datapoint_definition( "{$data->method}:{$data->datapoint}" );
 			$oauth_client = $this->get_oauth_client_for_datapoint( $datapoint );
 
+			// Set request as using shared credentials to false.
```
aaemnnosttv, line 341: Please use comments to add context and explain why we're doing something that may otherwise not make sense. This comment is redundant as-is because it states what it does rather than why, which would be good to call out here especially with the default value and `finally` below.
```suggestion
			// Always reset this property first to ensure it is only set true for the current request.
```

includes/Core/REST_API/Exception/Invalid_Dimensions_Exception.php
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Class Google\Site_Kit\Core\REST_API\Exception\Invalid_Dimensions_Exception
+ *
+ * @package   Google\Site_Kit\Core\REST_API\Exception
+ * @copyright 2022 Google LLC
+ * @license   https://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
+ * @link      https://sitekit.withgoogle.com
+ */
+
+namespace Google\Site_Kit\Core\REST_API\Exception;
```
aaemnnosttv, line 11: Let's move this to (a new) `Core\Validation\Exception` namespace as `REST_API` is rather specific to that integration, although this validation is applicable more generally (e.g. if SK API requests were possible via WP-CLI).

TBH this could apply to the datapoint exception too, but let's leave that be for now 😄

includes/Core/REST_API/Exception/Invalid_Dimensions_Exception.php
```diff
@@ -0,0 +1,75 @@
...
+use WP_Error;
+
+/**
+ * Exception thrown when dimensions are invalid for a report request.
+ *
+ * @since n.e.x.t
+ * @access private
+ * @ignore
+ */
+class Invalid_Dimensions_Exception extends Exception implements WP_Errorable {
```
aaemnnosttv, line 24: Let's extend PHP's [`DomainException`](https://www.php.net/manual/en/class.domainexception.php) as a more appropriate base instead of the top-level exception.

aaemnnosttv (changes requested):
Thanks @nfmohit - this looks great, just a few minor points to address 👍

includes/Modules/Analytics.php
```diff
@@ -1420,7 +1389,7 @@ protected function validate_report_metrics( $metrics ) {
 	 *
 	 * @param array $dimensions The dimensions to validate.
```
aaemnnosttv, line 1390: While this works as-is, let's update the type to expect an array of the appropriate instance.
```suggestion
	 * @param Google_Service_AnalyticsReporting_Dimension[] $dimensions The dimensions to validate.
```

includes/Modules/Analytics.php
```diff
@@ -1372,7 +1344,7 @@ public function check_service_entity_access() {
 	 *
 	 * @param array $metrics The metrics to validate.
```
aaemnnosttv, line 1345:
```suggestion
	 * @param Google_Service_AnalyticsReporting_Metric[] $metrics The metrics to validate.
```

aaemnnosttv (commented):
One more thing 😄

nfmohit:
Thank you for the kind reviews @aaemnnosttv! I've addressed them.

includes/Modules/Analytics.php
```diff
@@ -1395,32 +1367,29 @@ protected function validate_report_metrics( $metrics ) {
...
-				/* translators: %s is replaced with a comma separated list of the invalid metrics. */
-				_n(
-					'Unsupported metric requested: %s',
-					'Unsupported metrics requested: %s',
-					count( $invalid_metrics ),
-					'google-site-kit'
-				),
-				implode( ', ', $invalid_metrics )
-			);
+			throw new Invalid_Metrics_Exception( '', 0, null, $invalid_metrics );
```
aaemnnosttv, line 1381: As mentioned before, let's give the message to the exception rather than have it generate its own. This makes the instantiation much simpler as well as we don't need to provide it with the metrics (although maybe this would make sense to do in the future), it's not needed yet.

aaemnnosttv (commented):
Thanks @nfmohit ! I think we were a bit ambitious with the addition of the exceptions here and after giving it another look, I think it was a bit simpler before and while I still think they are good to use, we can aim for a bit of a middle ground here.

Apologies for the iterations – I've gone ahead and made the final adjustments (mostly undoing some of the additions here) to wrap this up for the release, but see my comments below.

aaemnnosttv (approved):

nfmohit:
> Thanks @nfmohit ! I think we were a bit ambitious with the addition of the exceptions here and after giving it another look, I think it was a bit simpler before and while I still think they are good to use, we can aim for a bit of a middle ground here.
> 
> Apologies for the iterations – I've gone ahead and made the final adjustments (mostly undoing some of the additions here) to wrap this up for the release, but see my comments below.

I've looked at all the mentioned suggestions, the implemented changes, and I agree with all of them. Thank you very much for showing me the right track here and implementing the appropriate changes, @aaemnnosttv!

Issue #5410, opened 2022-06-22
Minor UX bug on AdSense Settings When Publisher ID loading


wpdarren:
## Bug Description

Minor UX bug when you go to AdSense settings. If you click on the `Check your site status` link while the Publisher ID is loading (this can take a couple of seconds),  the settings page reloads rather than to AdSense sites page.  It's only when the Publisher ID has loaded that the link works as expected.

Not a huge issue, but wondering if it's possible to improve the UX here? 

![image.png](https://images.zenhubusercontent.com/5f99965247ef5a1da6985d7d/e7768def-49ab-44a3-b119-d6a2f51c75f9)

## Steps to reproduce

1. Go to 'Site Kit Settings'
2. Click on 'AdSense'
3. Click the 'Check your site status' straightaway and before Publisher ID loads.

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* The link to `Check your site status` should be unclickable until the URL for it is available (initially it is `undefined` but this isn't accounted for)

## Implementation Brief

* In `assets/js/modules/adsense/components/settings/SettingsView.js`:
  * If `siteStatusURL` is undefined, add the 'disabled' prop to the `<Link>` component so it is not clickable until it is loaded.

### Test Coverage

* No changes expected

## QA Brief

* Follow steps to reproduce, perhaps mocking a slower internet connection and verify the link is 'disabled'/'unclickable'.
* IMPORTANT: This issue "fixes" a bug in all Links which output the `<a>` tag. So smoke test various links within the plugin - especially links with the "external" image/icon.

## Changelog entry

* Disable AdSense "Check your site status" link while settings are loading to prevent unexpected behavior.

jimmymadon:
@aaemnnosttv The AC suggests rendering the `<Link>` but making it "Visually Hidden". Does this mean wrapping the entire component within the `<VisuallyHidden>` component? I feel this bug can be resolved easily if we just disable the link until the URL loads which is what I have suggested in the IB for now. In my testing, the loading is super quick on most sites. Or should we just show a loading bar for the whole component like we do for other settings views? This will avoid any colour changes or link appearing flickers?

eugene-manuilov:
IB ✔️

aaemnnosttv:
[...]

PR #5790, opened 2022-09-05
Disable AdSense Site Status link when loading.

jimmymadon:
## Summary

Addresses issue:

- #5410 

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist

### DIFF
--- a/assets/js/components/Link.js
+++ b/assets/js/components/Link.js
@@ -157,7 +157,9 @@ const Link = forwardRef(
 				} ) }
 				disabled={ disabled }
 				href={
-					type === LINK || type === EXTERNAL_LINK ? href : undefined
+					( type === LINK || type === EXTERNAL_LINK ) && ! disabled
+						? href
+						: undefined
 				}
 				onClick={ onClick }
 				rel={
--- a/assets/js/modules/adsense/components/settings/SettingsView.js
+++ b/assets/js/modules/adsense/components/settings/SettingsView.js
@@ -240,6 +240,7 @@ export default function SettingsView() {
 							href={ siteStatusURL }
 							className="googlesitekit-settings-module__cta-button"
 							external
+							disabled={ siteStatusURL === undefined }
 						>
 							{ siteStatusLinkLabel }
 						</Link>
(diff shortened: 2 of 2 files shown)


tofumatt (changes requested):
Unfortunately looks like the suggestion made in the issue doesn't really work here: a "disabled" link is still clickable as it still creates a link to the value in the `href` prop: 

CleanShot 2022-09-05 at 23 08 30

It looks like we should probably be rendering that component as an entirely different element when we don't want it to be a link. Arguably this is a bug in our `Link` component, where we shouldn't be rendering an `<a>` tag when it's disabled. 🤔

Personally I'd be fine fixing that bug as part of this issue, but we should thoroughly QA it and check through the repo for existing `<Link disabled />` usage to make sure there aren't any regressions.

Oh, and the "External" link indicator should either not be shown for a disabled link or the colour should match the disabled link colour. I lean toward the latter, but I think either would work.

aaemnnosttv:
Good catch @tofumatt. I think all we need to do is swap the `Link` out for a `Button` which should look the same and have the disabled behavior we want. This may have changed with GM2 though as our story looks the same as a regular button 🤔 https://google.github.io/site-kit-wp/storybook/develop/?path=/story/components-button--link-button

jimmymadon:
Yup - nice catch @tofumatt!

1. Within Link, we currently render a simple `<a>` in most cases, which does not support the `disabled` attribute (even though we effectively pass the prop as an attribute to the tag). 
2. We render `<button>` (which does support `disabled`) ONLY when the `onClick()` prop is passed AND the `href` prop is NOT passed. 

So as @tofumatt pointed out, a simple fix would be to render a `<button>` if the `disabled` prop is passed - [which is exactly what we have done in the corresponding Button component](https://github.com/google/site-kit-wp/blob/1324367229a56f31b10dddac3f656a38c9f95b00/assets/js/components/Button.js#L65-L66). However, the rendered `<button>` would not support the `external` prop. So instead, I've done another simple fix which is to remove the `href` if the link is `disabled`. And I have used the existing `hideExternalIndicator` prop to hide the icon rather than use a CSS mask to change the colour of the background svg.

tofumatt (changes requested):
So the changes to `Link` and such all look good, but this PR is targeting `main` and looks like it's had `develop` merged into it after the branch was split.

There are changes in here from some other, unrelated PRs not in the release.

@jimmymadon Can you rebase these changes to include only your commits?

tofumatt (approved):

Issue #5145, opened 2022-04-27
Show Measurement ID instead of Property ID in Analytics 4 Property Select Dropdown when there is an existing GA4 Tag


kuasha420:
## Feature Description

When there's an existing GA Tag, it is added as and shown to the UI as the Measurement ID. However, In the property select dropdown, there is only the Property ID present. This is confusing UX, as even though the toggle behaves correctly (Turning on/off) based on the selected properties underlying measurement ID, to the user it seems to be happening randomly as there's no indication that the selected property has/doesn't have the same existing tag measurement ID.

![Screenshot_20220428_002201](https://user-images.githubusercontent.com/6187401/165593745-08537f69-3f5c-4bf1-a105-f79a7778f92a.png)

Showing the measurement ID in place or in addition to the property ID would make things clearer and also help user to select the correct property. 

---------------

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance criteria

* Google Analytics 4 properties in the GA4 property dropdown/select should show their name and measurement ID, rather than their name and property ID (as they currently do).
* The REST API endpoints for GA `properties` and `property` should be updated to include the measurement ID, which is not currently included in the response. (See this comment: https://github.com/google/site-kit-wp/issues/5145#issuecomment-1166382009)

## Implementation Brief

In `assets/js/modules/analytics-4/components/common/PropertySelect.js`:
* Get the `measurementID`s of all the `properties` using the `getWebDataStreamsBatch` selector. 
    * if `properties` is not loaded, return `null`.
    * Get the `propertyIDs` array by mapping over the `properties`.
    * Get `webDataStreams` for the above `propertyIDs` using the `getWebDataStreamsBatch` selector.
    * if `webDataStreams` is not loaded, return `null`.
    * Return a `propertyID:measurementID` object by mapping over all the webstreams and picking up `webDataStream[ 0 ].webStreamData.measurementId`. 
* Update `isLoading` to be `true` when `measurementID` is null. 
* In the JSX, Update the `Option` element to use the `measurementID` like this-
```jsx
<Option key={ index } value={ _id }>
      { _id === PROPERTY_CREATE
        ? displayName
        : sprintf(
        /* translators: 1: Property name. 2: Measurement ID. */
        _x(
          '%1$s (%2$s)',
        'Analytics property name and ID',
        'google-site-kit'
    ),
      displayName,
      measurementIDs[ _id ]
  ) }
</Option>
```

### Test Coverage

* Not needed.

[...]

PR #5849, opened 2022-09-15
Show measurement ID instead of property ID.

makiost:
## Summary

Addresses issue:

- #5145 

## Relevant technical choices

One suite of the unit tests is still failing, due to the measurement ID's not being loaded in time. Haven't managed to solve it, even with help from the channel, so hoping maybe a fresh set of eyes in the CR can see what's wrong.

## PR Author Checklist

---------------

## Code Reviewer Checklist

- [ ] Run the code.
- [ ] Ensure CI checks pass.

## Merge Reviewer Checklist


assets/js/modules/analytics-4/components/common/PropertySelect.js
```diff
@@ -88,6 +88,37 @@ export default function PropertySelect( {
...
+			}
+
+			const measurementID =
+				dataStream[ currentPropertyID ][ 0 ].webStreamData
+					.measurementId; // eslint-disable-line sitekit/acronym-case
+			isLoading = false;
+			return ( measurementIDs[ currentPropertyID ] = measurementID );
+		} );
+
+		return propertyMeasurementMapping;
```
hussain-t, line 119: This doesn't do anything since you aren't utilizing the returned value. The `useSelect` hook just maps the `measurementId` to above-declared object `measurementIDs`.

assets/js/modules/analytics-4/components/common/PropertySelect.js
```diff
@@ -88,6 +88,37 @@ export default function PropertySelect( {
...
+	} );
+
```
hussain-t, line 121: The better API would be to return the mapped object from the `useSelect` hook and use it like this:
```suggestion
	const measurementIDs = useSelect( ( select ) => {
		if ( ! properties.length ) {
			return null;
		}

		return properties.reduce( ( acc, property ) => {
			const currentPropertyID = property._id;

			const dataStream =
				select( MODULES_ANALYTICS_4 ).getWebDataStreamsBatch(
					currentPropertyID
				);

			if ( ! Object.keys( dataStream ).length ) {
				isLoading = true;
				return null;
			}

			const measurementID =
				dataStream[ currentPropertyID ][ 0 ].webStreamData
					.measurementId; // eslint-disable-line sitekit/acronym-case
			isLoading = false;
			return {
				...acc,
				[ currentPropertyID ]: measurementID,
			};
		}, {} );
	} );
```

hussain-t (changes requested):
Thanks, @makiost. I tested your changes, and it works; however, I have left a comment for improvements. LMK WDYT?

Apart from that, the `PropertySelect` component tests are failing. Could you fix it?

makiost:
Should be ok now @hussain-t, I've addressed the points in your CR. Any pointers on the failing unit test? 🤔

hussain-t:
> Should be ok now @hussain-t, I've addressed the points in your CR. Any pointers on the failing unit test? 🤔

@makiost, I have fixed the failing tests. Basically, you have to provide mock data for `getWebDataStreamsBatch` using `receiveGetWebDataStreamsBatch` action. Please refer to this [commit](https://github.com/google/site-kit-wp/pull/5849/commits/13249c995b738955ecc772ae27e5e3f27780c9c3).

There was one other potential issue if the user doesn't have module access, the `properties` will be `null`. In that case, the `properties.length` check will throw an error checking `null.length`. I have fixed that as well. Please refer to this [commit](https://github.com/google/site-kit-wp/pull/5849/commits/5290ac9fc5830d78866043701d90ec3ec9caa784).

Can you check why the storybook test build is failing?

assets/js/modules/analytics-4/components/common/PropertySelect.js
```diff
@@ -88,6 +88,34 @@ export default function PropertySelect( {
...
+			return null;
+		}
+		return properties.reduce( ( acc, property ) => {
+			const currentPropertyID = property._id;
+			const dataStream =
+				select( MODULES_ANALYTICS_4 ).getWebDataStreamsBatch(
+					currentPropertyID
+				);
+			if ( ! Object.keys( dataStream ).length ) {
+				isLoading = true;
```
hussain-t, line 105: There is a problem here. If we delete `googlesitekit_analytics-4_settings` from the wp_options table, `getWebDataStreamsBatch` selector will return an empty object. Hence the following condition will be `true`. So, the loader is displayed all the time. Please refer to the recording below:

https://user-images.githubusercontent.com/24408261/195570652-e2c095ae-cc50-4a5e-88aa-d0f5466ec3bb.mov

We shouldn't set the loader to `true` here to avoid the above issue. Rather we should show the property select dropdown without the property being selected:

Screenshot 2022-10-13 at 3 37 19 PM


To make it work, we should conditionally get the measurement ID from the object `measurementIDs?.[ _id ]` in the `Option` element.

WDYT, @aaemnnosttv @tofumatt?

hussain-t (changes requested):
@makiost, I have fixed the failing tests and another issue. However, there is one potential issue please refer to my comments. Could you also look why the storybook test is failing?

assets/js/modules/analytics-4/components/common/PropertySelect.js
tofumatt: Yes, let's not run `useSelect` calls for side-effects as in the original code. @hussain-t's suggestion/approach makes sense: we should always try to use values returned from hooks and not have a `useSelect` mutate outside variables; it's harder to keep track of.

tofumatt: Just noticing one other thing: I don't know if the `isLoading` mutation in there is exactly what we want—maybe that should be a state variable?

hussain-t: @tofumatt, my point is not to set the loading state in that place. Please refer to this [comment](https://github.com/google/site-kit-wp/pull/5849#discussion_r994457037).

assets/js/modules/analytics-4/components/common/PropertySelect.js
tofumatt: Getting the measurement ID from `measurementIDs?.[ _id ]` makes sense to me 👍🏻

hussain-t: Thanks, @tofumatt

makiost:
Hey @hussain-t, fixed the story now, was the same issue as in the unit test, just had to add some data to `getWebDataStreamsBatch`  😄  I've also sent you a PM for some clarity on the last point you've made.

hussain-t:
> Hey @hussain-t, fixed the story now, was the same issue as in the unit test, just had to add some data to `getWebDataStreamsBatch` 😄 I've also sent you a PM for some clarity on the last point you've made.

As we discussed you can set `measurementIDs` ? current implementation : set it to `Select new property`

hussain-t (approved):
LGTM!

techanvil (approved):
LGTM!

